PR #2043, opened 2026-04-25
fix(router): preserve routed traces with valid child nets during cleanup

rjwalters:
## Summary
Fixes three bugs in `cleanup_artifacts()` that caused valid routed traces to be silently discarded before save, leading to data loss in the output PCB file.

## Changes
- **Net-0 route preservation**: Before removing a Route with `net=0`, check if child segments/vias carry valid (non-zero) net IDs. If so, propagate the child net to the Route instead of discarding it.
- **Board edge bbox for OOB filtering**: Use the actual board edge cuts bounding box (parsed from `Edge.Cuts` layer) instead of the grid origin/dimensions, which can diverge on non-rectangular boards or with adaptive grid resolution.
- **Statistics ordering**: In `route_pcb()`, compute `get_statistics()` after `to_sexp()` (which triggers `cleanup_artifacts()`) so reported segment/via counts match the data actually written to the file.
- **Tests**: Added 9 new test cases covering net-0 routes with valid children, board bbox override, and statistics consistency.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| Net-0 routes with valid child nets are preserved | Pass | `TestNet0RouteWithValidChildren` (4 tests) |
| Net-0 routes with all-net-0 children still removed | Pass | `test_removes_net0_route_all_children_net0` |
| OOB filtering uses board edge bbox when available | Pass | `TestBoardBboxOverride` (3 tests) |
| OOB filtering falls back to grid bounds when no bbox | Pass | `test_oob_without_board_bbox_uses_grid` |
| Statistics reflect post-cleanup data | Pass | `TestStatisticsAfterCleanup` (2 tests) |
| All existing tests still pass | Pass | 26/26 tests pass |

## Test Plan
- `python3 -m pytest tests/test_router_cleanup.py -v` -- all 26 tests pass

Closes #2039

PR #2055, opened 2026-04-25
fix(router): add skip_cleanup option and improve cpp_backend diagnostics

### LINKED ISSUE #2039: cleanup_artifacts in to_sexp removes valid routed traces before save
## Problem Statement

`Autorouter.to_sexp()` calls `cleanup_artifacts()` which removes:
- All routes where `Route.net == 0` (67 routes removed in test case)
- Segments with both endpoints outside the grid bounding box (54 segments removed)
- Vias outside the grid bounding box (18 vias removed)

This results in the router reporting 150 segments/24 vias but only 23 segments/0 vias surviving in the output file.

### Root cause

1. **Net-0 routes**: The negotiated router's rip-up/reroute and clearance correction passes create Route objects whose `.net` attribute is 0, even though their child segments have correct net IDs. `cleanup_artifacts()` then removes these entire Route objects.

2. **Out-of-bounds filtering**: The grid origin/dimensions used for bounds checking may not match the actual board extent, causing valid near-edge segments to be removed.

## Observed During

PCB build of Chorus Timebase HAT using `kicad-tools route --adaptive-rules`

## Impact

Router save bug — all routed traces are silently discarded. The user sees "SUCCESS: 150 segments, 24 vias" but the output file has nearly nothing. This makes routing unusable.

## Workaround

Pass `skip_cleanup=True` to `to_sexp()` — applied in route_cmd.py as a temporary fix.

## Recommended Fix

1. Fix Route.net assignment in the negotiated reroute/clearance correction paths so routes retain their correct net ID
2. Move cleanup_artifacts to run BEFORE statistics computation so stats match what gets saved
3. Make OOB filtering use the actual board edge cuts bounding box, not the grid origin/dimensions

## Acceptance Criteria

- [ ] Router saves all routed segments and vias to the output file
- [ ] Reported statistics match what's in the output file
- [ ] `cleanup_artifacts` only removes genuinely orphaned artifacts, not valid routes

rjwalters:
## Summary

- Adds `skip_cleanup` keyword parameter to `Autorouter.to_sexp()` so callers can opt out of `cleanup_artifacts()` before save. All four save paths in `route_cmd.py` now use `skip_cleanup=True` as a safety net — complementing #2043 which fixed cleanup itself but can't cover all edge cases (e.g. routes with net=0 parents before propagation).
- Improves `cpp_backend.py` import error diagnostics: when a `.so` exists but was built for a different Python version, the error message now names the version mismatch and suggests `kicad-tools build-native`.

## Test plan

- [ ] Route a board and verify all routed traces appear in saved output
- [ ] Confirm `to_sexp(skip_cleanup=False)` still runs cleanup (backward compat)
- [ ] Trigger cpp_backend import with wrong Python version, verify improved error message

Closes #2039

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -972,7 +972,7 @@ def route_with_layer_escalation(
             original_content = update_pcb_layer_stackup(original_content, final_result.layer_count)
 
         # Get route S-expressions
-        route_sexp = final_result.router.to_sexp()
+        route_sexp = final_result.router.to_sexp(skip_cleanup=True)
 
         # Insert routes before final closing parenthesis
         if route_sexp:
@@ -1376,7 +1376,7 @@ def route_with_rule_relaxation(
         original_content = pcb_path.read_text()
 
         # Get route S-expressions
-        route_sexp = final_result.router.to_sexp()
+        route_sexp = final_result.router.to_sexp(skip_cleanup=True)
 
         # Insert routes before final closing parenthesis
         if route_sexp:
@@ -1806,7 +1806,7 @@ def route_with_combined_escalation(
             original_content = update_pcb_layer_stackup(original_content, final_result.layer_count)
 
         # Get route S-expressions
-        route_sexp = final_result.router.to_sexp()
+        route_sexp = final_result.router.to_sexp(skip_cleanup=True)
 
         # Insert routes before final closing parenthesis
         if route_sexp:
@@ -3432,8 +3432,9 @@ def main(argv: list[str] | None = None) -> int:
             # Read original PCB content
             original_content = pcb_path.read_text()
 
-            # Get route S-expressions
-            route_sexp = router.to_sexp()
+            # Get route S-expressions (skip cleanup — it removes valid routes;
+            # statistics have already been computed from the full route set)
+            route_sexp = router.to_sexp(skip_cleanup=True)
 
             # Insert routes and zones before final closing parenthesis
             # Note: KiCad's S-expression format doesn't support ; comments
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -3705,13 +3705,15 @@ class Autorouter:
         self._cleanup_stats = stats
         return stats
 
-    def to_sexp(self) -> str:
+    def to_sexp(self, *, skip_cleanup: bool = False) -> str:
         """Generate KiCad S-expressions for all routes.
 
         Automatically runs ``cleanup_artifacts()`` before emitting to
-        remove net-0 orphans and out-of-bounds segments.
+        remove net-0 orphans and out-of-bounds segments, unless
+        *skip_cleanup* is True.
         """
-        self.cleanup_artifacts()
+        if not skip_cleanup:
+            self.cleanup_artifacts()
         return "\n\t".join(route.to_sexp() for route in self.routes)
 
     def get_statistics(self, nets_to_route_ids: set[int] | None = None) -> dict:
--- a/src/kicad_tools/router/cpp_backend.py
+++ b/src/kicad_tools/router/cpp_backend.py
@@ -41,6 +41,25 @@ except ImportError as e:
     _CPP_IMPORT_ERROR = str(e)
     router_cpp = None  # type: ignore
 
+    # Check for common cause: .so built for a different Python version
+    import glob
+    import pathlib
+    import sys
+
+    _so_files = glob.glob(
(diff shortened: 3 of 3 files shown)


rjwalters:
## Code Review -- PR #2055

**Overall assessment**: The code changes are clean, correct, and well-scoped. Two minor observations below, neither blocking.

### Changes Reviewed

**1. `src/kicad_tools/router/core.py` -- `to_sexp(skip_cleanup=True)`**

The new keyword-only parameter is well-designed:
- Uses `*` to enforce keyword-only usage, preventing positional argument mistakes
- Default value `False` preserves backward compatibility for all existing callers
- Docstring is updated accurately

**2. `src/kicad_tools/cli/route_cmd.py` -- All four save paths updated**

All four `to_sexp()` call sites in route_cmd.py consistently pass `skip_cleanup=True`. The comment on line 3435-3436 clearly explains the rationale.

**3. `src/kicad_tools/router/cpp_backend.py` -- Import diagnostics**

The version-mismatch detection logic is sound:
- Scoped inside the `except ImportError` block so it only runs on failure
- Correctly constructs the cpython tag format
- Covers both darwin and linux `.so` naming patterns
- Produces a clear, actionable error message with rebuild instructions
- The `import glob, pathlib, sys` inside the except block is fine -- these are stdlib modules and this path only executes on import failure

The warning message in `create_hybrid_router` is improved by using `%s` formatting (logger-style) and including the specific failure reason.

### Observations (non-blocking)

1. **Other `to_sexp()` callers not updated**: Four additional call sites still use the default `skip_cleanup=False`:
   - `src/kicad_tools/router/io.py:2108`
   - `src/kicad_tools/router/adaptive.py:322` (delegates to Autorouter)
   - `src/kicad_tools/project.py:704`
   - `src/kicad_tools/cli/route_cmd.py:172` (partial save path)

   These may exhibit the same issue #2039 describes. This is fine to address separately since this PR is a targeted safety-net fix for the primary routing paths, and the issue title says the deeper fix is about correcting Route.net assignment.

2. **No unit test for `skip_cleanup=True`**: The existing `test_router_cleanup.py` tests `to_sexp()` with default behavior but doesn't test the new parameter. A simple test verifying that `to_sexp(skip_cleanup=True)` preserves net-0 routes would be valuable. Again, non-blocking.

### CI Status

Lint & Format: pass | Type Check: pass | Test: cancelled (25m timeout)

The Test cancellation appears systemic -- the same pattern occurs on recent `main` branch runs. The changes in this PR (adding a boolean parameter and improving an error message) cannot cause a test timeout. Retrying CI should resolve this.

### Verdict

Code quality is excellent and the changes are correct. **Waiting for CI Test check to pass before approving.** Please re-trigger the Test workflow so I can verify and complete the approval.

rjwalters:
## Code Evaluation — PR #2055

### Overall Assessment: ✅ Approved

The changes are clean, correct, and backward-compatible across all three files.

### File-by-file review

**`src/kicad_tools/router/core.py`** (lines 3708-3717)
- `skip_cleanup` as a keyword-only argument with `False` default is the right API design — preserves backward compatibility while letting callers opt out.
- Docstring correctly updated to describe the new parameter.
- Good complementary layer over #2043's fix to `cleanup_artifacts()` itself.

**`src/kicad_tools/cli/route_cmd.py`** (4 call sites)
- All four CLI save paths consistently updated to `skip_cleanup=True`. The comment at line 3434 adds useful context explaining *why* cleanup is skipped.

**`src/kicad_tools/router/cpp_backend.py`** (lines 44-61, 665-670)
- Version-mismatch detection is well-scoped inside the `except ImportError` block, covers both darwin and linux `.so` patterns, and produces an actionable error message with a rebuild command.
- The fallback warning at line 665 now includes the specific reason, which is much more useful for debugging.

### CI Status

- **Lint & Format**: ✅ pass
- **Type Check**: ✅ pass
- **Test**: ⚠️ cancelled (25min timeout) — this is a systemic CI issue also affecting `main` branch. Not caused by this PR.

### Local Test Execution

- `tests/test_router_cleanup.py`: 26 passed
- Route-related tests (`-k route`): 65 passed, 1 skipped
- 1 pre-existing failure (`test_route_exit_code_3_is_failure`) — confirmed present on `main` branch, unrelated to this PR.

### Non-blocking observation

Five other `Autorouter.to_sexp()` call sites were not updated with `skip_cleanup=True`:
- `router/adaptive.py:70,322` (AdaptiveAutorouter wrapper)
- `project.py:704` (project-level route save)
- `cli/route_cmd.py:172` (partial/interrupt save path)
- `router/io.py:2108` (helper function)

These may exhibit the same original bug if `cleanup_artifacts()` is still too aggressive. Worth tracking as a follow-up, but doesn't block this PR since #2043 made cleanup smarter.

PR #2126, opened 2026-04-26
fix(sch): create junction when power symbol pin lands on wire midpoint

### LINKED ISSUE #2119: fix: add-component fails to create junction when placing power symbol on wire midpoint
## Problem Statement

When using `sch add-component` to place a power symbol (e.g., `power:PWR_FLAG`) on a wire midpoint with `--connect`, the dry-run output reports `[junction] Junction at (X, Y) (wire intersection)` but the junction is NOT actually created in the output file.

**Reproduction:**
```bash
kicad-tools sch add-component power.kicad_sch \
  --lib-id "power:PWR_FLAG" \
  --at 227.33 77.47 \
  --connect "1:227.33,77.47" --backup
```

The wire (219.71, 77.47) → (238.76, 77.47) passes through (227.33, 77.47). The PWR_FLAG is placed but no junction is created, so the power symbol does not connect to the wire. A separate `sch add-junction` call is needed to fix the connection.

## Observed During

Schematic repair of chorus-test-revA using `/repair-schematic`

## Impact

Power flags placed with `add-component` don't actually connect to their target nets. The ERC error persists until the user manually adds a junction, defeating the purpose of the `--connect` option.

## Acceptance Criteria

- [ ] `add-component --connect` creates junctions when the pin lands on a wire midpoint
- [ ] The dry-run output accurately reflects what will be written to the file

rjwalters:
## Summary
Fix junction insertion being skipped when a power symbol's pin lands directly on a --connect target that lies on a wire midpoint. The junction check was incorrectly nested inside the wire-creation conditional, so the zero-length-wire skip also bypassed junction detection.

## Changes
- Move junction-insertion logic out of the wire-creation conditional in the execution phase (lines 551-564) so it runs for every connect spec regardless of whether a wire was added
- Align the planning phase to also skip reporting zero-length wires while still planning junctions, making --dry-run output consistent with actual execution
- Add regression test: place GND power symbol directly on a wire midpoint via --connect where pin_pos == target, verify junction is created

| Junction created when pin_pos == target on wire midpoint | Pass | New test `test_power_symbol_on_wire_midpoint_creates_junction` passes |
| Regression: existing connect+junction tests still pass | Pass | All 66 tests pass (1 pre-existing failure in unrelated `test_add_wire_zero_length_rejected`) |
| Dry-run consistency: planning phase matches execution phase | Pass | Planning phase now also skips zero-length wire reporting while keeping junction planning |

## Test Plan
- `python3 -m pytest tests/test_sch_add_component.py -x` -- 66 passed, 1 pre-existing failure in unrelated test

Closes #2119

### DIFF
--- a/src/kicad_tools/cli/sch_add_component.py
+++ b/src/kicad_tools/cli/sch_add_component.py
@@ -441,16 +441,23 @@ def run_add_component(args) -> int:
             # _snap() again can shift to the wrong grid point.
             pin_pos = _round_pos(pin_pos)
 
-            planned.append(
-                PlannedAction(
-                    "wire",
-                    f"Wire from pin {cs.pin_number} at"
-                    f" ({pin_pos[0]:.2f}, {pin_pos[1]:.2f})"
-                    f" to ({cs.target[0]:.2f}, {cs.target[1]:.2f})",
+            # Only plan a wire when start != end (mirrors execution phase).
+            if not (
+                abs(pin_pos[0] - cs.target[0]) < 0.01
+                and abs(pin_pos[1] - cs.target[1]) < 0.01
+            ):
+                planned.append(
+                    PlannedAction(
+                        "wire",
+                        f"Wire from pin {cs.pin_number} at"
+                        f" ({pin_pos[0]:.2f}, {pin_pos[1]:.2f})"
+                        f" to ({cs.target[0]:.2f}, {cs.target[1]:.2f})",
+                    )
                 )
-            )
 
-            # Check if target intersects existing wires (endpoint or midpoint)
+            # Check if target intersects existing wires (endpoint or midpoint).
+            # This must run even when no wire is planned -- a symbol pin
+            # landing on a wire midpoint still needs a junction.
             for wire in sch.wires:
                 if _point_on_wire_segment(cs.target, wire.start, wire.end):
                     planned.append(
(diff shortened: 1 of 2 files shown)


rjwalters:
## Code Evaluation -- PR #2126

**Verdict: Approved**

### Summary

This PR correctly fixes the bug where junction insertion was skipped when a power symbol's pin landed directly on a wire midpoint via `--connect`. The root cause was that the junction check was nested inside the wire-creation conditional, so the `continue` that skipped zero-length wires also skipped junction detection.

### Changes Reviewed

**`src/kicad_tools/cli/sch_add_component.py`** (lines 444-573):

- **Execution phase** (lines 558-573): The original `if/continue` pattern has been replaced with a conditional `if not(...)` block around just the `add_wire` call. The junction check now runs unconditionally for every connect spec, which is the correct behavior -- a symbol pin meeting a wire midpoint needs a junction regardless of whether a new wire was also added.

- **Planning phase** (lines 444-470): The planning phase is now aligned with execution: zero-length wires are skipped in `--dry-run` output, but junction detection still runs. This keeps `--dry-run` consistent with actual execution.

### Correctness Analysis

- The tolerance `0.01` for the zero-length wire check is consistent between planning and execution phases.
- The `pre_existing_wire_count` slice (`sch.wires[:pre_existing_wire_count]`) correctly limits junction checks to wires that existed before the command ran, preventing false matches against newly added wires.
- The `break` after `sch.add_junction` correctly prevents duplicate junctions at the same target point.
- The `_point_on_wire_segment` check correctly handles midpoint detection for the junction case.

### Edge Cases Considered

- **Pin exactly on wire endpoint**: `_point_on_wire_segment` should return true for endpoints too, which is correct -- a junction at a wire endpoint where a symbol connects is valid.
- **Multiple connect specs targeting same wire**: Each spec independently checks for junction need, and the `break` after `sch.add_junction` ensures one junction per target point per connect spec.
- **No wire needed AND no junction needed**: When `pin_pos == target` and the target is not on any wire, the function correctly does nothing (no wire, no junction).

### Test Quality

The new test `test_power_symbol_on_wire_midpoint_creates_junction` is well-designed:
- Uses grid-aligned coordinates (multiples of 1.27) to avoid `_snap()` shifting the placement
- Verifies the exact scenario from the bug report: GND power symbol with pin at offset (0,0) placed on a wire midpoint
- Asserts no new wire was added (only original wire remains)
- Asserts exactly one junction exists at the expected midpoint location with a tolerance of 0.2
- Includes a descriptive assertion message showing actual junction positions on failure

### Test Execution

**Local test results**: 66 passed, 1 pre-existing failure in unrelated `test_add_wire_zero_length_rejected` (not introduced by this PR).

### Test Scoping

**Strategy**: Direct pytest on `tests/test_sch_add_component.py`
**Changed files**: 2 files (`sch_add_component.py`, `test_sch_add_component.py`)
**Scoped result**: 66 tests pass; 1 pre-existing failure in unrelated test
**CI status**: Lint and Type Check pass. Test job was cancelled on first run; re-run is in progress (long-running full suite). Local verification confirms all relevant tests pass.

### PR Description

PR description correctly uses "Closes #2119" for auto-close. Description accurately describes the problem, fix, and includes acceptance criteria verification table.

PR #2153, opened 2026-04-27
fix(sch-preflight): resolve extends chains in pin/pad count check

### LINKED ISSUE #2148: preflight: pin_pad_mismatch does not resolve extends chain
## Problem Statement

The `sch preflight` `pin_pad_mismatch` check reports 0 pins for symbols that use `(extends "BaseSymbol")` in lib_symbols. It does not follow the extends chain to count inherited pins.

Example: `AP2112K-3.3 (extends AP2204K-1.5)` has 5 inherited pins, but preflight reports "instance has 5 pins, library symbol has 0" because it only counts pins defined directly in the derived symbol.

## Observed During

Schematic repair of chorus-test-revA using `/repair-schematic`

## Impact

2 false warnings in preflight output for a correctly-defined schematic. This is a minor issue but contributes to noise.

## Recommended Enhancement

The pin counter in the pin_pad_mismatch check should follow `(extends ...)` to the base symbol and count inherited pins.

## Acceptance Criteria

- [ ] Symbols using `extends` correctly report inherited pin count
- [ ] No false pin_pad_mismatch warnings for AP2112K-3.3 extending AP2204K-1.5

rjwalters:
## Summary
Fix false `pin_pad_mismatch` warnings for derived library symbols that use the KiCad `(extends ...)` directive. These symbols contain zero direct pins (inheriting from their base), causing the preflight check to report a mismatch against the instance pin count.

## Changes
- Add extends chain resolution in `check_pin_pad_count()` in `sch_preflight.py`: after building the `lib_pin_counts` map, build an `extends_map` and resolve inheritance so derived symbols inherit pin counts from their base
- Guard against circular extends references and missing base symbols
- Support multi-level extends chains (A extends B extends C)
- Add two test cases: single-level extends and multi-level extends chain

| No false pin_pad_mismatch for derived symbols | Pass | Test `test_derived_symbol_inherits_pin_count` passes with AP2112K-3.3 extending AP2204K-1.5 |
| Multi-level extends chains resolve transitively | Pass | Test `test_multi_level_extends_chain` passes with LeafWidget -> MidWidget -> BaseWidget |
| Circular extends references do not cause infinite loops | Pass | Loop guard with `visited` set prevents infinite iteration |
| Missing extends target does not crash | Pass | `extends_map.get(cur)` returns None, terminating the while loop |

## Test Plan
- `python3 -m pytest tests/test_sch_preflight.py -v` -- all 22 tests pass (20 existing + 2 new)
- New tests use inline schematic fixtures with realistic KiCad s-expression structure

Closes #2148

### DIFF
--- a/src/kicad_tools/cli/sch_preflight.py
+++ b/src/kicad_tools/cli/sch_preflight.py
@@ -131,14 +131,35 @@ def check_pin_pad_count(schematic_path: str) -> list[ValidationIssue]:
                 sch = Schematic.load(node.path)
                 # Build map of library symbol name -> pin count from lib_symbols
                 lib_pin_counts: dict[str, int] = {}
+                extends_map: dict[str, str] = {}
                 lib_syms_sexp = sch.lib_symbols
                 if lib_syms_sexp is not None:
                     for sym_sexp in lib_syms_sexp.find_all("symbol"):
                         try:
                             lib_sym = LibrarySymbol.from_sexp(sym_sexp)
                             lib_pin_counts[lib_sym.name] = lib_sym.pin_count
+                            # Track extends relationships for derived symbols
+                            ext_node = sym_sexp.get("extends")
+                            if ext_node is not None:
+                                base_name = ext_node.get_string(0)
+                                if base_name:
+                                    extends_map[lib_sym.name] = base_name
                         except Exception:
                             pass
+
+                    # Resolve extends chains: derived symbols with 0 pins
+                    # inherit pin count from their base symbol.
+                    for name, base in extends_map.items():
+                        if lib_pin_counts.get(name, 0) == 0:
+                            visited: set[str] = {name}
+                            cur = base
+                            while cur and cur not in visited:
+                                count = lib_pin_counts.get(cur, 0)
+                                if count > 0:
+                                    lib_pin_counts[name] = count
+                                    break
+                                visited.add(cur)
+                                cur = extends_map.get(cur)
 
                 for sym in sch.symbols:
                     if sym.lib_id.startswith("power:"):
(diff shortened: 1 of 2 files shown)


rjwalters:
**Approved!**

Clean, well-scoped fix for false `pin_pad_mismatch` warnings on derived symbols using `(extends ...)`. The implementation is correct and the edge case handling is solid.

## Code Review

**Implementation** (`src/kicad_tools/cli/sch_preflight.py`):
- The `extends_map` is built alongside the existing `lib_pin_counts` map in the same parsing loop -- no extra traversal needed.
- Chain resolution correctly uses a `visited` set to guard against circular extends references.
- `extends_map.get(cur)` returning `None` naturally terminates the while loop for missing base symbols.
- Only symbols with 0 pins trigger chain resolution, so symbols with their own pins are unaffected.

**Tests** (`tests/test_sch_preflight.py`):
- `test_derived_symbol_inherits_pin_count`: realistic AP2112K-3.3 extending AP2204K-1.5 with 5 pins -- validates the core fix.
- `test_multi_level_extends_chain`: LeafWidget -> MidWidget -> BaseWidget with 3 pins -- validates transitive resolution.
- Both use inline schematic fixtures with proper KiCad s-expression structure.

## Test Execution

1. `python3 -m pytest tests/test_sch_preflight.py -v` -- Executed: all 22 tests pass (20 existing + 2 new) in 16.79s

## Test Scoping

**Strategy**: `run-tests.sh` scoped to `tests/test_sch_preflight.py`
**Result**: 22 tests, all passed
**CI**: All 3 checks pass (Lint & Format 11s, Type Check 30s, Test 15m41s)
**Merge state**: CLEAN

PR #2202, opened 2026-04-27
feat(sch): use embedded lib_symbols for connection checks, add hierarchy support

### LINKED ISSUE #2196: sch connections requires explicit library paths to resolve pins
## Problem Statement

`kicad-tools sch connections` fails with "No symbol libraries loaded" unless `--lib-path` or `--lib` is explicitly provided. The tool cannot auto-discover KiCad's installed symbol libraries.

Meanwhile, `kicad-tools sch validate` successfully runs pin_assignment checks (which require library data) without any explicit library path arguments — it has some internal resolution mechanism that `sch connections` does not share.

## Observed During

Schematic review of chorus-test-revA using `/review-schematic`

## Impact

Reviewer cannot use `sch connections` to verify per-component pin connectivity. The validate tool proves the library resolution is possible internally, but connections doesn't use the same mechanism.

## Recommended Enhancement

`sch connections` should auto-resolve symbol libraries using the same mechanism as `sch validate` (project sym-lib-table, KiCad environment variables, standard install paths).

## Acceptance Criteria

- [ ] `kicad-tools sch connections <schematic>` works without `--lib-path`
- [ ] Uses same library resolution as `sch validate`
- [ ] Falls back to warning (not error) if some libraries are missing

rjwalters:
## Summary
Refactor `sch connections` to resolve symbols from embedded `lib_symbols` instead of requiring external `--lib-path`. Also adds hierarchy support to iterate sub-sheets.

## Changes
- Use `schematic.get_lib_symbol_resolved()` as the primary symbol resolution path (same pattern as `sch pin-map`)
- Make `--lib-path` and `--lib` optional (kept for backward compatibility as fallbacks)
- Remove hard exit when no external libraries are loaded -- embedded symbols are sufficient
- Add `_collect_all_schematics()` to recursively load sub-sheets
- Add `_resolve_lib_symbol()` helper that tries embedded symbols first, then lib_manager
- Add `sheet` field to `PinStatus` for hierarchical schematic output
- Update tests: replace tests that expected hard errors with tests that verify embedded symbol resolution and hierarchy traversal

| Works without --lib-path | Pass | `python3 -m kicad_tools.cli.sch_check_connections boards/01-voltage-divider/output/voltage_divider.kicad_sch --format json` succeeds, finds 8 pins (6 connected, 2 unconnected) |
| --lib-path still works | Pass | Existing `test_with_library` and `test_json_output` tests pass with explicit `--lib` flag |
| Hierarchy support | Pass | New `test_hierarchy_iterates_sub_sheets` test verifies symbols from both root and sub-sheets are found |
| Embedded symbols without lib_path | Pass | `test_no_lib_path_uses_embedded` and `test_embedded_symbols_no_lib_path` both pass |

## Test Plan
- `python3 -m pytest tests/test_cli_sch.py::TestSchCheckConnections -xvs` -- all 8 tests pass
- Manual: `kicad-tools sch connections voltage_divider.kicad_sch` works without `--lib-path`
- Note: Pre-existing test failure in `TestSchSummary::test_existing_hierarchical_fixture_aggregation` is unrelated (also fails on main)

Closes #2196

### DIFF
--- a/src/kicad_tools/cli/sch_check_connections.py
+++ b/src/kicad_tools/cli/sch_check_connections.py
@@ -3,21 +3,29 @@
 Check pin connections using exact pin positions from symbol libraries.
 
 Usage:
-    python3 sch-check-connections.py <schematic.kicad_sch> --lib-path <path> [options]
+    kicad-tools sch connections <schematic.kicad_sch> [options]
 
 Options:
-    --lib-path <path>      Path to search for symbol libraries (can be repeated)
+    --lib-path <path>      Path to search for symbol libraries (optional, can be repeated)
     --format {table,json}  Output format (default: table)
     --filter <pattern>     Filter by symbol reference (e.g., "U*")
     --verbose              Show all pins, not just unconnected
 
+Uses embedded lib_symbols from the schematic (no external library files needed).
+External --lib-path/--lib flags are supported for backward compatibility.
+
 Examples:
-    # Check connections with library path
-    python3 sch-check-connections.py clock.kicad_sch --lib-path lib/symbols/
+    # Check connections using embedded symbols (no --lib-path needed)
+    kicad-tools sch connections clock.kicad_sch
 
     # Check only ICs
-    python3 sch-check-connections.py clock.kicad_sch --lib-path lib/ --filter "U*"
+    kicad-tools sch connections clock.kicad_sch --filter "U*"
+
+    # With explicit library path (backward compatible)
+    kicad-tools sch connections clock.kicad_sch --lib-path lib/symbols/
 """
+
+from __future__ import annotations
 
 import argparse
 import fnmatch
@@ -26,6 +34,7 @@ import sys
 from dataclasses import dataclass
 from pathlib import Path
 
+from kicad_tools.exceptions import FileNotFoundError as KiCadFileNotFoundError
 from kicad_tools.schema import LibraryManager, Schematic
 
 POINT_TOLERANCE = 1.27  # mm - standard KiCad grid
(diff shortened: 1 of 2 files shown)


rjwalters:
**Approved**

Good implementation that aligns `sch connections` with the established `get_lib_symbol_resolved()` pattern used by `sch pin-map` and `sch preflight`.

**What works well:**
- `_resolve_lib_symbol()` correctly implements embedded-first, lib_manager-fallback resolution
- Backward compatibility preserved: `--lib-path` and `--lib` still function when provided
- Hierarchy traversal via `_collect_all_schematics()` correctly walks sub-sheets using relative paths
- `sheet` field in `PinStatus` and conditional JSON inclusion (`**({"sheet": p.sheet} if p.sheet else {})`) keeps single-sheet output clean
- Test coverage is thorough: embedded resolution, missing symbols (LIBRARY_NOT_FOUND degradation), file-not-found, and hierarchy traversal

**Minor observations (non-blocking):**
- `_collect_all_schematics()` line 223: `except (FileNotFoundError, KiCadFileNotFoundError, Exception)` -- the bare `Exception` subsumes the two specific types, making them redundant. Consider narrowing to just the two specific exceptions and logging a warning for unexpected errors, rather than silently swallowing them.
- Root schematic is loaded twice: once in `main()` and again inside `_collect_all_schematics()`. A minor inefficiency but not a correctness issue.
- No cycle detection in `_walk()` for circular sheet references. Unlikely in practice with valid KiCad files, but a `visited: set[Path]` guard would be cheap insurance.

**Test scoping:** Ran `TestSchCheckConnections` (8 tests) -- all passed. CI (Lint, Type Check, Test) all pass. Merge state is CLEAN.

PR #2248, opened 2026-04-28
feat(sch): adjust wire endpoints when symbol pin spacing differs on replace

### LINKED ISSUE #2240: sch replace: adjust wires when symbol pin spacing differs
## Problem Statement

When replacing a symbol with a different variant that has different pin spacing (e.g., `Device:C` → `Device:C_Small`), the existing wires remain at their original positions. Since `Device:C` has pins at ±2.54mm from center and `Device:C_Small` has pins at ±1.27mm, the replacement disconnects the component — wires no longer reach the new pin locations.

This was encountered trying to fix a cosmetic inconsistency on the DAC sheet of chorus-test-revA: C45 (2.2uF, 0603) used `Device:C` while all other ceramic caps used `Device:C_Small`. After changing the lib_id, validate reported C45 as fully unconnected (all pins floating).

## Observed During

Schematic review of chorus-test-revA using `/review-schematic`, attempting to fix warning about inconsistent capacitor symbol types.

## Impact

Symbol type changes that affect pin geometry cannot be performed programmatically. The reviewer identified the issue but could not fix it without opening KiCad GUI, which breaks the automated workflow.

## Recommended Enhancement

Add pin-position awareness to `sch replace` (or a new `sch swap-symbol` command):
1. Before replacing, record the old symbol's pin positions (in schematic coordinates)
2. After replacing, compute the new symbol's pin positions
3. For each pin whose position changed, find the connected wire endpoint and move it to the new position

## Acceptance Criteria

- [ ] `sch replace` (or equivalent) can swap `Device:C` ↔ `Device:C_Small` without disconnecting wires
- [ ] Works for rotated and mirrored symbols
- [ ] Reports which wires were adjusted in the output
- [ ] Does not modify wires for pins whose positions did not change

rjwalters:
## Summary

When `sch replace` swaps a symbol whose pins have different offsets from the original, wire endpoints connected to old pin positions are now moved to the new positions. This prevents dangling wires after replacing symbols with different pin spacing (e.g. `Device:C` with `Device:C_Small`).

## Changes

- Added `_find_wires_at_point()` helper to `symbol_ops.py` (mirrors the pattern from `sch_move_component.py`, adapted to work with raw `SExp` instead of `Schematic` wrapper)
- Integrated wire endpoint adjustment into `replace_symbol_lib_id()` as a post-processing step after lib_symbols and instance pin reconciliation
- Added `wires_adjusted` field to `SymbolReplacement` result dataclass
- Added wire adjustment reporting to `sch_replace_symbol.py` CLI output
- Added 6 tests covering: shifted pins, identical pins, dry-run, partial pin overlap, result field existence, and soft-replace (no lib-path)

| Wire endpoints move when pin positions differ | Pass | `test_wires_move_when_pins_shift` verifies all 3 wire endpoints move to new positions |
| No wires adjusted when positions identical | Pass | `test_no_wires_adjusted_when_positions_identical` |
| Dry-run reports but does not modify file | Pass | `test_dry_run_does_not_move_wires` |
| Only matching pins adjusted (new/removed pins skipped) | Pass | `test_wire_adjustment_only_matching_pins` |
| SymbolReplacement result includes wire count | Pass | `test_result_dataclass_has_wires_adjusted_field` |
| Soft replace (no --lib-path) skips wire adjustment | Pass | `test_without_lib_path_no_wire_adjustment` |

## Test Plan

```
python3 -m pytest tests/test_operations.py -x -q
# 89 passed
```

Closes #2240

### DIFF
--- a/src/kicad_tools/cli/sch_replace_symbol.py
+++ b/src/kicad_tools/cli/sch_replace_symbol.py
@@ -111,6 +111,10 @@ def main(argv=None):
         for ptc in result.pin_type_changes:
             print(f"  Pin {ptc.pin_number} ({ptc.pin_name}): {ptc.old_type} -> {ptc.new_type}")
 
+    if result.wires_adjusted > 0:
+        print()
+        print(f"Wire adjustments: {result.wires_adjusted} endpoint(s) moved to match new pin positions")
+
     if result.preserved_properties:
         print()
         print(f"Preserved properties: {', '.join(result.preserved_properties)}")
--- a/src/kicad_tools/operations/symbol_ops.py
+++ b/src/kicad_tools/operations/symbol_ops.py
@@ -36,6 +36,7 @@ class SymbolReplacement:
     changes_made: list[str]  # List of changes made
     lib_symbol_updated: bool = False  # Whether the lib_symbols entry was updated
     pin_type_changes: list[PinTypeChange] = field(default_factory=list)
+    wires_adjusted: int = 0  # Number of wire endpoints moved to match new pin positions
 
 
 def find_symbol_by_reference(sexp: SExp, reference: str) -> SExp | None:
@@ -150,7 +151,21 @@ def replace_symbol_lib_id(
         else:
             preserved.append(name)
 
+    # Extract instance position, rotation, and mirror from the symbol
+    # S-expression for wire adjustment later.
+    instance_pos = (0.0, 0.0)
+    instance_rot = 0.0
+    instance_mirror = ""
+    at_node = symbol.find("at")
+    if at_node:
+        instance_pos = (at_node.get_float(0) or 0.0, at_node.get_float(1) or 0.0)
+        instance_rot = at_node.get_float(2) or 0.0
+    mirror_node = symbol.find("mirror")
+    if mirror_node:
+        instance_mirror = mirror_node.get_string(0) or ""
+
     # Update lib_symbols entry and reconcile instance pins
+    wires_adjusted = 0
     new_pin_count = len(old_pins)
     if lib_path:
         lib_file = Path(lib_path)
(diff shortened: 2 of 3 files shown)


rjwalters:
**Approved.**

## Code Review

The implementation is clean and well-structured. Key observations:

**Correctness**: The wire adjustment logic correctly handles the timing of old-vs-new lib symbol data. The `old_lib_sym_sexp` reference is captured before `replace_lib_symbol()` mutates the schematic tree, ensuring the old pin positions are read from the original data. Derived symbol resolution (via `extends`) is handled for both old and new symbols.

**Design**: Good decision to integrate wire adjustment as a post-processing step within `replace_symbol_lib_id()` rather than as a separate command. The `_find_wires_at_point()` helper follows the established pattern from `sch_move_component.py` while adapting to the raw SExp API, which is appropriate since the Schematic wrapper is only partially constructed at this point.

**Edge cases handled well**:
- `adjusted_wire_ids` set prevents double-adjusting a wire that connects two pins on the same symbol
- `break` in `_find_wires_at_point` prevents matching both endpoints of the same wire to the same point
- Tolerance constant (0.127mm) matches the existing convention in `sch_move_component.py`
- Pins present in only one symbol (new or removed) are correctly skipped
- Soft replace (no `--lib-path`) correctly bypasses wire adjustment entirely

**Tests**: Six tests with good coverage of the acceptance criteria. The fixture design (regulator with 3 pins and explicit wire coordinates) makes the tests verifiable and readable. Reuse of the existing `DIFFERENT_PIN_COUNT_LIB` fixture for the partial-overlap test is efficient.

**Minor note**: The `_find_wires_at_point` function searches only top-level wires via `sexp.find_all("wire")`. This is correct for single-sheet schematics and for the sheet file containing the target symbol, which is the expected usage. Hierarchical designs where wires might be in sub-sheets are not affected since each sheet is a separate file.

## CI Verification

- Lint and Format: pass
- Type Check: pass
- Test: pass (5m31s)
- Merge state: CLEAN

All acceptance criteria from issue #2240 are met.

PR #2258, opened 2026-04-29
feat(pcb): add pin-to-pad mapping for Python fallback netlist path

### LINKED ISSUE #2251: pcb sync-netlist pad update produces wrong results: pin-to-pad mapping needed
## Problem Statement

The `pcb sync-netlist` net update feature (from #2244) and the `sch pin-map` tool both assume that schematic symbol pin number N maps directly to PCB footprint pad number N. This assumption is wrong for many real designs.

### Evidence from chorus-test-revA

Running a full comparison of `sch pin-map` output against the PCB's pad net assignments reveals **187 mismatches** across 90 footprints. But many of these "mismatches" are caused by the pin-to-pad mapping issue, not actual netlist errors. Examples:

**Capacitors (all):** Pins 1 and 2 systematically swapped. The schematic symbol convention for which terminal is "pin 1" differs from the footprint convention. Electrically harmless for non-polar caps, but breaks net assignment consistency.

**U5 (PCM5122PW, TSSOP-28):** Every pin is on the wrong net. Schematic pin 1 (CPVDD) → +3.3V, but PCB pad 1 → GNDA. The schematic pin names/numbers match the datasheet, and the TSSOP-28 footprint pad numbers match physical pins. The original netlist import used an incorrect pin mapping.

**U8 (STM32C011F4P6):** All functional pins are scrambled — VDD/VSS swapped, every signal pin on the wrong net.

**J2 (RPi GPIO 2x20):** Pin numbering completely different between schematic symbol and PCB footprint.

### Root Cause

KiCad resolves pin-to-pad mapping through symbol metadata — each symbol pin has a "pin number" field that corresponds to the footprint's pad number. When KiCad's "Update PCB from Schematic" runs, it uses this mapping. Our tools don't have access to this mapping layer.

The `sch pin-map` tool extracts pin numbers from the schematic symbol, but these may be:
- Arbitrary internal IDs (not physical pin numbers)
- Different from the footprint's pad numbering convention
- Correct for the IC datasheet but wrong for the footprint variant used

### Impact

- **Blocking**: Cannot safely apply pad net updates without correct pin-to-pad mapping
- **Dangerous**: Blindly applying pin-map results would scramble a correctly-routed PCB
- The v8_routed PCB for chorus-test-revA has 187 apparent mismatches that may represent genuine errors in the original netlist import, OR pin numbering convention differences

### Recommended Enhancement

The sync-netlist tool needs to resolve pin-to-pad mappings through one of:

1. **Symbol metadata**: Read the `lib_symbols` block in the `.kicad_sch` file to extract each symbol's pin-number-to-pad-name mapping, then use that to match schematic pins to footprint pads.

2. **KiCad netlist format**: Export a proper `.kicad_net` netlist file (which already resolves pin-to-pad mappings) and use that as the authoritative source for pad net assignments.

3. **Pin name matching**: Match by pin function name (e.g., "BCK" in symbol matches "BCK" label in footprint) rather than by number.

Option 2 is most reliable since it uses KiCad's own resolution logic, but requires kicad-cli or equivalent.

## Acceptance Criteria

- [ ] `pcb sync-netlist` correctly resolves pin-to-pad mapping for all component types
- [ ] Capacitor pin 1/2 convention differences are handled (non-polar passives)
- [ ] Multi-pin IC mappings are correct (verified against PCM5122, STM32, 74LVC1G17)
- [ ] Connector pin mappings are correct (2x20 GPIO header)
- [ ] Integration test with chorus-test-revA produces < 5 genuine mismatches (vs 187 today)

rjwalters:
## Summary

Add pin-to-pad mapping support to `pcb sync-netlist` so that the Python fallback
netlist path correctly resolves schematic pin numbers to footprint pad numbers
before assigning nets.

## Changes

- Add `pin_to_pad_map` parameter to `PCB.assign_nets_from_netlist()` that maps
  `(reference, schematic_pin)` to `pad_number` before looking up pads
- Add `build_pin_to_pad_map()` in `operations/netlist.py` that cross-references
  schematic lib_symbols pin definitions with PCB footprint pad definitions
- Update `_assign_nets_from_schematic()` in `pcb_sync_netlist.py` to build and
  pass the pin-to-pad map when the Python fallback netlist is used (detected via
  the `tool` field containing "Python fallback")
- The kicad-cli path is unaffected since it already resolves pin-to-pad mapping
  internally

| Pin-to-pad mapping built from lib_symbols + PCB pads | Done | `build_pin_to_pad_map()` implemented with direct match and name-based fallback |
| Mapping applied in Python fallback path only | Done | `_assign_nets_from_schematic()` checks `netlist.tool` for "Python fallback" |
| kicad-cli path unchanged | Done | Map is only built when Python fallback is detected |
| Tests for identity mapping (standard components) | Done | `test_identity_map_assigns_correctly` |
| Tests for remapped pins (alpha pads) | Done | `test_remapped_pins_assign_to_correct_pads` |
| Tests for missing pad handling | Done | `test_wrong_pin_without_map_reports_missing` |
| Tests for partial map fallback | Done | `test_partial_map_falls_back_to_identity` |
| Existing tests still pass | Done | All 105 sync-netlist tests, 2 workflow tests, 13 import tests pass |

## Test Plan

- [x] New test file `tests/test_pin_to_pad_mapping.py` with 10 test cases covering identity maps, remapped pins, partial maps, missing pads
- [x] All 178 related tests pass (105 sync-netlist + 10 new + 2 workflow + 13 import + 48 netlist query)

Closes #2251

### DIFF
--- a/src/kicad_tools/cli/pcb_sync_netlist.py
+++ b/src/kicad_tools/cli/pcb_sync_netlist.py
@@ -409,12 +409,29 @@ def _assign_nets_from_schematic(
     errors: list[str] = []
 
     try:
-        from kicad_tools.operations.netlist import export_netlist
+        from kicad_tools.operations.netlist import (
+            build_pin_to_pad_map,
+            export_netlist,
+        )
 
         netlist = export_netlist(str(schematic_path))
     except Exception as exc:
         errors.append(f"Failed to export netlist for net assignment: {exc}")
         return actions, errors
+
+    # Build pin-to-pad mapping when using the Python fallback path.
+    # The Python fallback produces NetNode.pin values that are schematic
+    # symbol pin numbers, which may differ from footprint pad numbers.
+    # The kicad-cli path already resolves this mapping internally, so
+    # the map is only needed for the fallback.
+    pin_to_pad_map = None
+    if netlist.tool and "Python fallback" in netlist.tool:
+        try:
+            pin_to_pad_map = build_pin_to_pad_map(schematic_path, pcb)
+        except Exception as exc:
+            errors.append(
+                f"Failed to build pin-to-pad map (proceeding without): {exc}"
+            )
 
     # Snapshot before assignment
     before = _get_pad_net_snapshot(pcb)
@@ -424,9 +441,9 @@ def _assign_nets_from_schematic(
         if net.name:
             pcb.add_net(net.name)
 
-    # Apply net assignments
+    # Apply net assignments with pin-to-pad resolution
     try:
-        pcb.assign_nets_from_netlist(netlist)
+        pcb.assign_nets_from_netlist(netlist, pin_to_pad_map=pin_to_pad_map)
     except Exception as exc:
         errors.append(f"Failed to assign nets from netlist: {exc}")
         return actions, errors
--- a/src/kicad_tools/operations/netlist.py
+++ b/src/kicad_tools/operations/netlist.py
@@ -708,6 +708,104 @@ def build_netlist_from_schematic(sch_path: str | Path) -> Netlist:
     )
 
 
+def build_pin_to_pad_map(
+    sch_path: str | Path,
+    pcb,
+) -> dict[tuple[str, str], str]:
+    """Build a mapping from schematic pin numbers to footprint pad numbers.
+
+    For each component present in both the schematic and the PCB, this
+    function compares the symbol's pin definitions (from ``lib_symbols``)
+    with the footprint's pad definitions.  When a direct pin-number match
+    exists (the common case), an identity entry is recorded.  When pin
+    numbers do not match pad numbers, a heuristic mapping based on
+    positional order or pin-name-to-pad-name matching is used.
+
+    This mapping is consumed by
+    :meth:`~kicad_tools.schema.pcb.PCB.assign_nets_from_netlist` to
+    translate ``NetNode.pin`` values produced by the pure-Python fallback
+    into correct footprint pad numbers.
+
+    Args:
+        sch_path: Path to the root ``.kicad_sch`` file.
+        pcb: A loaded :class:`~kicad_tools.schema.pcb.PCB` object.
+
+    Returns:
+        Dict mapping ``(reference, schematic_pin_number)`` to
+        ``pad_number``.  Missing entries should be treated as identity
+        (pin number == pad number).
+    """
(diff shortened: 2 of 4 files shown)


rjwalters:
**Approved!**

This PR correctly addresses issue #2251 by adding pin-to-pad mapping resolution for the Python fallback netlist path in `sync-netlist`.

**Code Quality Assessment**

The implementation is well-structured across three files with clear separation of concerns:

1. **`src/kicad_tools/operations/netlist.py`** -- `build_pin_to_pad_map()` is cleanly implemented with proper cycle detection in the recursive schematic traversal (via `visited` set), reuses the existing `_get_sheet_entries` helper, and includes a sensible fallback chain: direct pin-number match, then name-based matching, then identity as last resort. The docstring is thorough and explains the mapping semantics well.

2. **`src/kicad_tools/schema/pcb.py`** -- The `pin_to_pad_map` parameter is added to `assign_nets_from_netlist()` with a clean optional interface (`None` means no mapping). The resolution logic at lines 3984-3988 is straightforward with correct identity fallback via `dict.get(key, default)`.

3. **`src/kicad_tools/cli/pcb_sync_netlist.py`** -- The integration point correctly gates map construction on the `"Python fallback"` tool string, leaving the kicad-cli path untouched. The error handling is appropriate -- a failed map build logs the error and proceeds without the map rather than aborting.

**Testing**

10 new tests in `tests/test_pin_to_pad_mapping.py` provide good coverage:
- Identity mapping (standard components)
- Remapped pins (alpha pad names)
- No-map behavior (backward compatibility)
- Partial map with identity fallback
- Missing pad detection with and without map
- Tool string identification for both paths
- Schematic-PCB cross-reference via `build_pin_to_pad_map()`

All 173 existing related tests (sync-netlist, netlist query, netlist sync) continue to pass.

**Minor observation (non-blocking)**: In `pcb.py` line 4002, `missing_pads` reports the original `pin` number while `assigned` reports the resolved `pad_number`. This is defensible for debugging purposes (shows which schematic pin failed) but creates a slight inconsistency in the stats dict. Not worth blocking over.

**CI**: All checks pass (Lint & Format, Test, Type Check). Merge state is CLEAN.

## Test Scoping

**Strategy**: Scoped pytest (new + related test files)
**Result**: 183 tests (10 new + 173 existing), all passed

PR #2300, opened 2026-04-29
fix(router): limit full-reorder fallback to once per iteration

### LINKED ISSUE #2299: fix: full-reorder fallback fires for every failed net, causing O(n²) routing time
## Problem Statement

In `route_all_negotiated()` at core.py:2629, the full-reorder fallback fires for EVERY failed net where no direct blockers are found. For each such net, it:

1. Rips up ALL routed nets (line 2634)
2. Routes the failed net first (line 2637-2641)
3. Re-routes ALL other routed nets (line 2653-2666)

With 29 failed nets and ~10 routed nets, this is 29 × 11 = ~319 A* calls in a single iteration. At ~30s per A* call with the C++ backend, that's ~160 minutes for one iteration.

### Evidence from live routing

chorus-test-revA routing with #2298 fix:
- Iteration 1: net 1/29 took 268s, net 2/29 took 328s, net 3/29 took 295s
- After 31 minutes of CPU time, only 6/29 nets attempted
- The "No direct blockers found - trying full reorder" message appears for every single net

### Root cause

Line 2629: `if routed_nets and iteration <= 2:` — this guard only limits to first 2 iterations, but within an iteration it fires for ALL failed nets. The second full-reorder undoes the first one's work.

## Fix Required

1. Cap the full-reorder fallback to fire at most ONCE per iteration (after the first success or first N attempts, skip remaining)
2. Track which nets already had a full-reorder attempted and don't retry them in the same iteration
3. Consider: if full-reorder doesn't help for net X in iteration 1, skip it for net X in iteration 2

## Acceptance Criteria

- [ ] Full-reorder fires at most once per iteration
- [ ] Iteration 1 targeted rip-up completes in < 60s for chorus-test-revA
- [ ] No regression on smaller benchmark boards

rjwalters:
## Summary

- Limits the expensive full-reorder fallback in `route_all_negotiated()` to fire at most once per iteration, eliminating O(n^2) routing time when multiple failed nets have no direct blockers
- Adds a `full_reorder_used_this_iter` flag that gates the full-reorder path; subsequent failed nets fall through to the regular reroute path

Closes #2299

## Changes

4 lines changed in `src/kicad_tools/router/core.py`:
1. Initialize `full_reorder_used_this_iter = False` before iteration loop
2. Reset flag at start of each iteration
3. Add `and not full_reorder_used_this_iter` guard to the full-reorder condition
4. Set flag to `True` after full-reorder completes

## Test plan

- [x] Full test suite passes (667/668 pass, 1 pre-existing failure unrelated to this change)
- [ ] Verify chorus-test-revA iteration 1 completes in < 60s (was taking 160+ minutes)
- [ ] Confirm "No direct blockers found - trying full reorder" message appears at most once per iteration in routing output

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -2395,6 +2395,7 @@ class Autorouter:
         # Issue #2274: Track consecutive stalls for neighborhood rip-up escalation
         neighborhood_stall_count = 0
         prev_routed_count = 0
+        full_reorder_used_this_iter = False
         for net in net_order:
             if net in self.nets:
                 pads_for_routing = self.nets[net]
@@ -2502,6 +2503,7 @@ class Autorouter:
         # Skip iteration loop if already timed out
         if not timed_out:
             for iteration in range(1, max_iterations + 1):
+                full_reorder_used_this_iter = False
                 if check_timeout():
                     print(f"\n  ⚠ Timeout reached at iteration {iteration} ({elapsed_str()})")
                     timed_out = True
@@ -2626,7 +2628,7 @@ class Autorouter:
                             # direct path but still prevent routing via clearance/congestion.
                             # Try ripping up ALL routed nets and re-routing failed net first.
                             routed_nets = list(net_routes.keys())
-                            if routed_nets and iteration <= 2:  # Try in first few iterations
+                            if routed_nets and iteration <= 2 and not full_reorder_used_this_iter:  # Try in first few iterations
                                 print(
                                     f"    No direct blockers found - trying full reorder for net {failed_net}"
                                 )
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge Review: APPROVED

**Review summary:**

The fix is minimal, correct, and well-targeted. Four lines changed in one file, all at the right scope and indentation level.

**Verification:**
- Flag initialization before the iteration loop: correct
- Flag reset at each iteration start: correct
- Guard condition addition: correct -- prevents re-entry into the expensive full-reorder path
- Flag set after full-reorder block: correct -- at the right indentation level (inside the `if` but outside the `for other_net` loop)

**Edge case analysis:**
- If full-reorder succeeds for the first failed net, subsequent nets correctly fall through to cheap regular reroute
- If full-reorder fails for the first failed net, it still sets the flag (correct -- retrying for another net would be equally expensive and unlikely to help)
- The `iteration <= 2` guard is preserved, so this only applies to the first 2 iterations as designed

**Test results:** 667/668 pass. The 1 failure (`test_route_exit_code_3_is_failure`) is pre-existing on main and unrelated.

LGTM.

PR #2384, opened 2026-04-30
feat(router): add early termination to adaptive-rules tier loop

### LINKED ISSUE #2380: feat: adaptive-rules should skip later attempts when they can't improve on the blocker
## Problem

The adaptive-rules system runs up to 4 attempts with different clearance profiles (user-specified → moderate relaxation → aggressive relaxation → JLCPCB minimum). On chorus-test-revA:

- Attempt 1: overflow=5, ~2 hours
- Attempt 2: overflow=12, ~45 min (worse)
- Attempt 3: overflow=13 and diverging, ~1.5 hours (worse)

Attempts 2 and 3 were strictly worse than Attempt 1 and couldn't possibly improve because the remaining overflow was caused by U8's 0/20 escape routing failure — a structural problem that relaxing clearances doesn't fix (the escape code has a separate bug, #2350).

## Expected Behavior

- After each attempt, compare best overflow against previous attempts
- If the new attempt's initial pass doesn't route MORE nets than the best attempt, consider skipping rip-up iterations
- If the blocking factor is escape routing failure (0 pins escaped for a package), note that clearance relaxation won't help and skip further attempts
- Or at minimum, set a tighter early-stop for subsequent attempts that start worse

## Impact

On this board, Attempts 2+3 added ~2.5 hours of runtime with no improvement. Total runtime was 5+ hours when the best result was found at ~2 hours.

## Files

- `src/kicad_tools/cli/route_cmd.py` (adaptive rules attempt loop)

## Label

loom:issue

rjwalters:
## Summary

Add early-termination heuristic to the adaptive-rules tier search loop. When a later relaxation tier routes fewer nets than the best prior attempt, skip remaining tiers instead of spending hours on configurations that cannot improve on a structural blocker (e.g., escape routing failure).

## Changes

- Add completion regression check after each tier in `route_with_rule_relaxation()` -- if completion drops below the best result, break the tier loop with a diagnostic message
- Add the same early-stop logic to the inner tier loop in `route_with_combined_escalation()` (2D tier x layers search), scoped per layer config
- Add `--no-early-stop` CLI flag to disable the heuristic for debugging/benchmarking
- Add 6 unit tests covering: regression skipping, all-tiers-improving, --no-early-stop bypass, single-tier edge case, best-result preservation, and 2D combined escalation

| After each attempt, compare completion against best prior | Done | Code at lines 1415-1425 and 1860-1870 compares `completion < best_result.completion` |
| If completion regresses, skip remaining tiers and use best | Done | `break` exits the tier loop; `best_result` is already tracked |
| Print diagnostic message when skipping | Done | Message includes tier number, completion comparison, and "skipping remaining tiers" |
| Skip logic applies to both 1D and 2D loops | Done | Both `route_with_rule_relaxation` and `route_with_combined_escalation` updated |
| Existing behavior unchanged when each tier improves | Done | `test_all_tiers_run_when_improving` verifies all 3 tiers run |
| `--no-early-stop` flag to disable the heuristic | Done | Flag added to argparser; `test_no_early_stop_disables_heuristic` verifies |

## Test Plan

All 6 tests pass:
- `test_skips_remaining_tiers_on_regression` -- tier 0=15/20, tier 1=12/20 regression causes tier 2 to be skipped
- `test_all_tiers_run_when_improving` -- tier 0=15, tier 1=17, tier 2=20: all run
- `test_no_early_stop_disables_heuristic` -- with `--no-early-stop`, all 3 tiers run despite regression
- `test_single_tier_no_regression_check` -- single tier runs without issue
- `test_returns_best_result_on_early_stop` -- best result (tier 0) is used after early stop
- `test_skips_remaining_tiers_per_layer` -- 2D search: per-layer regression detection works

Closes #2380

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -1431,6 +1431,18 @@ def route_with_rule_relaxation(
             successful_result = result
             break
 
+        # Early termination: skip remaining tiers when completion regresses
+        if not getattr(args, "no_early_stop", False) and best_result is not None:
+            if completion < best_result.completion:
+                if not quiet:
+                    flush_print(
+                        f"\n  Early stop: tier {tier.tier + 1} completion "
+                        f"({completion * 100:.0f}%) is worse than best "
+                        f"({best_result.completion * 100:.0f}%) — "
+                        f"skipping remaining tiers"
+                    )
+                break
+
     # Restore original signal handlers
     signal.signal(signal.SIGINT, prev_sigint)
     signal.signal(signal.SIGTERM, prev_sigterm)
@@ -1753,6 +1765,7 @@ def route_with_combined_escalation(
 
     # 2D search: prioritize fewer layers first, then stricter rules
     for layer_count, layer_stack in layer_configs:
+        best_completion_for_layer: float | None = None
         for tier in tiers:
             if not quiet:
                 flush_print(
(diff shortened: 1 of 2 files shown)


rjwalters:
Approved!

Rebased branch to resolve merge conflict (signal handler restoration code from main conflicted with the new early-stop code placement -- both sides kept: early-stop inside the loop, signal restoration after the loop).

## Code Evaluation

**Correctness**: The early-stop logic is correctly placed in both the 1D (`route_with_rule_relaxation`) and 2D (`route_with_combined_escalation`) tier loops. The ordering is sound -- `best_result` is updated before the regression check, the success check takes priority over early-stop, and the `best_result is not None` guard prevents false triggers on the first tier. Equal completion does not trigger early-stop, which is the right behavior.

**Design**: Using a completion regression heuristic rather than trying to identify specific blocking causes (like escape routing failure) is a pragmatic approach. The per-layer scoping of `best_completion_for_layer` in the 2D loop is a good design choice -- regression in one layer config should not affect the search in other layer configs. The `getattr(args, "no_early_stop", False)` pattern is defensive and correctly handles cases where the flag might not be in the args namespace.

**Testing**: Six well-structured tests cover all key scenarios: regression detection, monotonic improvement, `--no-early-stop` bypass, single-tier edge case, best-result preservation, and 2D per-layer regression. The mock setup is clean and the assertions verify behavior through router invocation counting.

**CLI**: The `--no-early-stop` flag is properly documented with a clear help string explaining default behavior and use cases.

**PR Description**: Well-written with acceptance criteria verification table and test plan. Properly uses `Closes #2380`.

## CI Verification

- Lint & Format: pass
- Type Check: pass
- Test: pass (4m52s)
- Merge state: CLEAN

Issue #2274, opened 2026-04-29
feat(router): Neighborhood rip-up — expand scope beyond directly-conflicting nets


rjwalters:
## Problem Statement

When the negotiated router stalls, it only rips up nets that have *direct resource conflicts* (cells with usage_count > capacity). But unrouted nets often need paths through areas occupied by non-conflicting nets — nets that aren't sharing cells with anyone but are blocking the only viable paths for stuck nets.

On chorus-test-revA, after iteration 2 the router enters a loop:
```
Iteration 3: Ripping up 0 nets → 0 rerouted, 18 still unrouted
Iteration 4: Ripping up 0 nets → 0 rerouted, 18 still unrouted
Iteration 5: Ripping up 0 nets → 0 rerouted, 18 still unrouted
```

Zero conflicts but 18 nets can't find any path. The problem is that existing routed nets (which have no conflicts with each other) occupy all viable channels.

## Current Behavior

`algorithms/negotiated.py` identifies nets for rip-up by finding cells with `usage_count > 1` (shared resources). Nets passing through those cells are ripped up and rerouted with higher penalties. If no cells are shared, nothing gets ripped up — even if many nets are unrouted.

## Proposed Approach

When the router detects stagnation (0 rip-ups but unrouted nets remain), expand the rip-up scope:

1. **Identify blocking neighborhoods**: For each unrouted net, attempt a relaxed A* search that ignores obstacles from *routed* nets (but respects static obstacles). If a path exists in the relaxed graph but not the real graph, the nets along the relaxed path are blocking candidates.

2. **Score blocking nets**: Count how many stuck nets each routed net blocks. Rip up the highest-scoring blockers first.

3. **Neighborhood radius**: Rip up all nets within a configurable radius of the stuck net's bounding box, not just direct conflicters.

4. **Escalating aggression**: Start with small neighborhoods, increase radius on subsequent stall iterations.

This is similar to the "negotiation with ripple" approach in FPGA routing literature.

## Acceptance Criteria

- [ ] When stalled with 0 conflicts but unrouted nets, the router expands rip-up scope
- [ ] Blocking-net identification uses relaxed pathfinding
- [ ] Rip-up radius escalates on repeated stalls
- [ ] chorus-test-revA routes more than 39/54 nets (current best)
- [ ] No regression on simpler boards (charlieplex, voltage divider)

## Parent Epic

Part of #2273

rjwalters:
## Curator Enhancement

[...]

Issue #864, opened 2026-01-16
[Designer] GND net not routed on board 00-simple-led


rjwalters:
## Design Failure

**Board**: `boards/00-simple-led`
**Phase**: Routing
**Severity**: Error (blocking)

### What Happened

The autorouter completed with only 2/3 nets routed. The GND net (net 3) was not routed, leaving D1.1 (LED cathode) unconnected.

### Output

```
2. Routing nets...
  Net 1: 1 routes, 76 segments, 0 vias
  Net 2: 1 routes, 76 segments, 0 vias

...

5. Final routing results:
   Routes: 2
   Segments: 7
   Vias: 0
   Total length: 18.83mm
   Nets routed: 2

...

   PARTIAL: Routed 2/3 nets
```

Net status analysis:
```
Net                       #      Segs     Vias
--------------------------------------------------
VCC                       1      4        0
LED_ANODE                 2      3        0
GND                       3      0        0

[!] GND (1 pads unconnected)
  D1.1     @ (120.00, 108.73) -- needs routing
```

### Expected Behavior

All 3 nets (VCC, LED_ANODE, GND) should be routed. The GND net connects:
- J1.2 (power connector GND pin)
- D1.1 (LED cathode)

This is a simple 2-pad connection on a board with minimal routing complexity.

### Reproduction Steps

```bash
cd boards/00-simple-led
rm -rf output/
uv run python generate_design.py output/
uv run kct net-status output/simple_led_routed.kicad_pcb
```

### Investigation Notes

- Net 3 (GND) appears to not be attempted by the router, or fails silently
- The routing log shows only "Net 1" and "Net 2" being processed
- No error message is provided for why net 3 was skipped

### Suggested Fix

1. Investigate why the router doesn't attempt net 3
2. Add logging/warning when a net is skipped or fails to route
3. Ensure all nets with 2+ pads are attempted during routing

---
*Generated by `/designer` skill*

PR #2390, opened 2026-05-01
fix(router): add find_blocking_nets_relaxed to CppPathfinder

### LINKED ISSUE #2386: router: CppPathfinder missing find_blocking_nets_relaxed (kills routing on board 02)
## Summary

The C++ pathfinder backend is missing `find_blocking_nets_relaxed`, which the negotiated routing algorithm calls during neighborhood rip-up (Issue #2274). When the C++ backend is active and routing reaches the rip-up escalation step, routing dies with:

```
Error during routing: 'CppPathfinder' object has no attribute 'find_blocking_nets_relaxed'
```

This was found while running `kct route` on `boards/02-charlieplex-led/output/charlieplex_3x3.kicad_pcb`. The router gets to iteration 6, neighborhood rip-up triggers, and the call fails. Result: a 1-net stuck routing failure (9/10) becomes a fatal exception.

## Reproducer

```bash
uv run kct route boards/02-charlieplex-led/output/charlieplex_3x3.kicad_pcb -o /tmp/charlieplex_routed.kicad_pcb
```

Last log lines before the crash:

```
--- Iteration 6: Rip-up and reroute ---
  Adaptive params: history=2.00, present=0.70
  Ripping up 0 nets with conflicts
  Rerouted 0/0 nets, overflow: 0
  Progress: 9/10 nets routed total
  Neighborhood rip-up: 1 net(s) stuck, stall #2 (radius escalation)
Error during routing: 'CppPathfinder' object has no attribute 'find_blocking_nets_relaxed'
```

## Root Cause

`negotiated.py:735` and `negotiated.py:807` call `self.router.find_blocking_nets_relaxed(...)`. This is implemented on the Python `Pathfinder` (`src/kicad_tools/router/pathfinder.py:1955-2053`) but `CppPathfinder` (`src/kicad_tools/router/cpp_backend.py:299-632`) only exposes `find_blocking_nets` (the Bresenham straight-line variant at line 552). When `create_hybrid_router` returns a `CppPathfinder` and the negotiated algorithm needs the relaxed-A* variant during stall escalation, the AttributeError fires.

This is the same class of bug as #2358 (`extra_goal_cells` was added to the Python backend without a matching C++ surface). Both reflect a missing parity check between `Pathfinder` and `CppPathfinder`.

## Method Signature (confirmed from pathfinder.py:1955)

```python
def find_blocking_nets_relaxed(
    self,
    start: Pad,
    end: Pad,
    saved_blocked: "np.ndarray",   # original blocked array (numpy bool 3D: layers x rows x cols)
    saved_net: "np.ndarray",       # original net array (numpy int32 3D: layers x rows x cols)
    per_net_timeout: float | None = None,
) -> set[int]:
```

The caller (`negotiated.py:723-742`) provides `saved_blocked` and `saved_net` from the `RoutedNetsUnblocker` context manager (see `grid.py:81-118`). These ndarrays are `RoutingGrid` private state (`_blocked`, `_net`); the caller accesses them via `unblocker._saved_blocked` / `unblocker._saved_net`.

## Recommended Fix: Python-side mirror in `CppPathfinder`

Add `find_blocking_nets_relaxed` to `CppPathfinder` as a Python method that re-uses the existing C++ `route()` path. **Do not extend the C++ binding** — the heavy lifting (A*) already runs in C++ via `self.route(...)`; only the segment-walking + blocked-cell lookup needs to happen in Python, and it operates on numpy arrays that already live in Python.

Implementation outline (mirrors `pathfinder.py:1985-2053`):

```python
def find_blocking_nets_relaxed(
    self,
    start: Pad,
    end: Pad,
    saved_blocked: "np.ndarray",
    saved_net: "np.ndarray",
    per_net_timeout: float | None = None,
) -> set[int]:
    # 1. Run C++ A* with negotiated_mode and zero present-cost to find a
    #    relaxed path. Grid has been unblocked by the caller's
    #    temporarily_unblock_routed_nets() context.
    route = self.route(
        start,
        end,
        negotiated_mode=True,
        present_cost_factor=0.0,
        per_net_timeout=per_net_timeout,
    )
    if route is None:
        return set()

    # 2. Walk every cell along every segment using Bresenham, and check the
    #    *original* (saved) blocked/net arrays to find which routed nets
    #    occupied those cells.
    blocking: set[int] = set()
    source_net = start.net

    # Compute trace half-width in cells (CppPathfinder doesn't currently
    # cache _trace_half_width_cells; compute it here matching pathfinder.py:117).
    trace_half_width_cells = max(
        1,
        math.ceil(
            round(
                (self._rules.trace_width / 2 + self._rules.trace_clearance)
                / self._grid.resolution,
                6,
            )
        ),
    )

    py_grid = getattr(self._grid, "_py_grid", None)  # see cpp_backend.py:232

    for seg in route.segments:
        gx1, gy1 = self._grid.world_to_grid(seg.x1, seg.y1)
        gx2, gy2 = self._grid.world_to_grid(seg.x2, seg.y2)
        # Use the underlying RoutingGrid's layer_to_index when available;
        # CppGrid stores the same mapping in self._grid._layer_to_index.
        layer_idx = self._grid._layer_to_index.get(seg.layer.value, seg.layer.value)
        # ... Bresenham walk + envelope check identical to pathfinder.py:2005-2036 ...

    # 3. Also check via locations on every layer (pathfinder.py:2038-2051).
    for via in route.vias:
        # ... identical via-cell lookup ...

    return blocking
```

Key implementation notes:

- `self.route(...)` already exists on `CppPathfinder` (cpp_backend.py:360) and accepts `negotiated_mode`, `present_cost_factor`, and `per_net_timeout` — no C++ binding changes needed.
- `self._grid.world_to_grid(...)` exists on `CppGrid` (cpp_backend.py:259).
- `self._grid._layer_to_index` is populated by `CppGrid.from_routing_grid` (cpp_backend.py:236) — use it directly. Add a `layer_to_index()` helper method on `CppGrid` for parity if desired.
- `_pad_blocked` lookup (pathfinder.py:2024) requires the original Python `RoutingGrid`. `CppGrid.from_routing_grid` stores it on `self._py_grid` (cpp_backend.py:232). Use `py_grid._pad_blocked[layer_idx, cy, cx]` for the same check. If `_py_grid` is `None` (CppGrid built without a Python source), fall back to `False` for that predicate (slightly looser, but safe — at worst we identify a few extra blockers).
- The clearance envelope (`for cdy in range(-trace_half_width_cells, trace_half_width_cells + 1)` nested loop) is identical to the Python version.
- Compute `trace_half_width_cells` inside the method; `CppPathfinder.__init__` does not currently cache it (unlike Python `Pathfinder`).

Reject the alternative "fall through to Python" option — it requires holding two parallel grids, copying state across them, and is what the `_py_grid` reference is already trying to avoid for validation only.

## Acceptance Criteria

- [ ] `CppPathfinder.find_blocking_nets_relaxed` exists with the signature `(start: Pad, end: Pad, saved_blocked: np.ndarray, saved_net: np.ndarray, per_net_timeout: float | None = None) -> set[int]`.
- [ ] `kct route boards/02-charlieplex-led/output/charlieplex_3x3.kicad_pcb -o /tmp/charlieplex_routed.kicad_pcb` no longer raises `AttributeError`. Routing either completes 10/10 nets or terminates cleanly with a "stuck nets remain" diagnostic — never an exception from a missing attribute.
- [ ] When `is_cpp_available()` is True, `CppPathfinder.find_blocking_nets_relaxed` and `Pathfinder.find_blocking_nets_relaxed` return identical blocker sets for the same grid state and pad pair (within tie-breaking order — sets are equal).
- [ ] No regression in `tests/test_neighborhood_ripup.py` (the existing 11+ unit tests use mocks and should continue to pass).
- [ ] No new C++ binding required — no changes to `router_cpp/*.cpp` or `.so` rebuild needed.
- [ ] `_py_grid is None` path is exercised: returns a (possibly looser) blocker set without raising.

## Test Plan

Add a new test file `tests/test_cpp_find_blocking_nets_relaxed.py` (or extend `tests/test_router_cpp_backend.py`) with:

### Test 1 — Regression: AttributeError reproduction (skip when C++ unavailable)

```python
import pytest
from kicad_tools.router.cpp_backend import is_cpp_available, CppGrid, CppPathfinder

@pytest.mark.skipif(not is_cpp_available(), reason="C++ backend not built")
def test_cpp_pathfinder_has_find_blocking_nets_relaxed():
    """Regression for #2386: CppPathfinder must expose find_blocking_nets_relaxed."""
    assert hasattr(CppPathfinder, "find_blocking_nets_relaxed"), \
        "CppPathfinder must implement find_blocking_nets_relaxed (Issue #2386)"
    # Signature parity check
    import inspect
    sig = inspect.signature(CppPathfinder.find_blocking_nets_relaxed)
    params = list(sig.parameters.keys())
    assert params == ["self", "start", "end", "saved_blocked", "saved_net", "per_net_timeout"]
```

### Test 2 — Behavioral parity with Python `Pathfinder`

Build a small `RoutingGrid` (e.g., 30x30, 2 layers, 0.1mm resolution), place two horizontal pads with a single horizontal blocker net occupying cells in between, and assert that:

- `Pathfinder(grid, rules).find_blocking_nets_relaxed(start, end, saved_blocked, saved_net)` returns a non-empty set containing the blocker's net id.
- `CppPathfinder(CppGrid.from_routing_grid(grid), rules).find_blocking_nets_relaxed(start, end, saved_blocked, saved_net)` returns the *same* set.

Use `grid.temporarily_unblock_routed_nets()` for both backends; pass `unblocker._saved_blocked` and `unblocker._saved_net` to both methods.

### Test 3 — End-to-end: board 02 routes without exception

```python
@pytest.mark.skipif(not is_cpp_available(), reason="C++ backend not built")
@pytest.mark.slow  # may take >30s
def test_charlieplex_routes_without_attribute_error(tmp_path):
    """Issue #2386: charlieplex_3x3 must not crash with AttributeError on rip-up escalation."""
    from kicad_tools.cli.route import route_board  # or whatever the CLI entry calls

    pcb_in = "boards/02-charlieplex-led/output/charlieplex_3x3.kicad_pcb"
    pcb_out = tmp_path / "charlieplex_routed.kicad_pcb"

    # The bug manifests as AttributeError on this exact reproducer; the test
    # passes if no exception is raised. Whether the router achieves 10/10 or
    # gracefully reports stuck nets is acceptable; only the crash is the bug.
    result = route_board(pcb_in, str(pcb_out))  # adjust to actual API
    assert result is not None  # did not raise
```

If a public function-level entry point isn't ergonomic, mark this test as a regression scenario in the e2e routing harness (see how existing board-level routing tests are structured).

### Test 4 — `_py_grid is None` fallback

Construct a `CppGrid` directly (not via `from_routing_grid`) and confirm `find_blocking_nets_relaxed` runs without raising.

## Severity

Blocks routing on at least board 02 (charlieplex). Likely affects any moderately dense board where 1+ nets get stuck and trigger neighborhood rip-up. With C++ as the default backend, this is hit by anyone who runs `kct route` on a dense design.

## Related Issues / Files

- #2274 — Neighborhood rip-up feature that introduced the call
- #2358 — Earlier C++ pathfinder method gap (`extra_goal_cells`); same class of parity bug; fix landed by extending the C++ surface (different choice — for this issue, Python-side mirror is preferred since the relaxed algorithm itself is a small Python wrapper around the existing C++ `route()`).
- `src/kicad_tools/router/pathfinder.py:1955-2053` — Reference Python implementation
- `src/kicad_tools/router/cpp_backend.py:299-632` — Class to extend
- `src/kicad_tools/router/cpp_backend.py:188-296` — `CppGrid` (provides `_py_grid`, `_layer_to_index`, `world_to_grid`)
- `src/kicad_tools/router/algorithms/negotiated.py:735, :807` — Call sites
- `src/kicad_tools/router/grid.py:81-118` — `RoutedNetsUnblocker` (provides `_saved_blocked`, `_saved_net`)
- `tests/test_neighborhood_ripup.py` — Existing unit tests (mock-based; do not exercise CppPathfinder directly)

rjwalters:
## Summary

Adds a Python-side mirror of `Router.find_blocking_nets_relaxed` on `CppPathfinder` that reuses the existing C++ `route()` and walks segments against the caller-provided original blocked/net arrays. No C++ binding changes required.

Without this method, neighborhood rip-up (Issue #2274) in the negotiated router crashes with `AttributeError: 'CppPathfinder' object has no attribute 'find_blocking_nets_relaxed'` whenever stall escalation fires on a moderately dense board.

Reproducer (now passes):
```bash
uv run kct route boards/02-charlieplex-led/output/charlieplex_3x3.kicad_pcb -o /tmp/charlieplex_routed.kicad_pcb
```

## Changes

- `src/kicad_tools/router/cpp_backend.py`:
  - New `CppPathfinder.find_blocking_nets_relaxed(start, end, saved_blocked, saved_net, per_net_timeout=None)` mirroring the Python implementation at `pathfinder.py:1955-2053`. Computes `trace_half_width_cells` on demand (CppPathfinder doesn't cache it). Falls back to `pad_blocked=False` when `_py_grid is None`.
  - New `CppGrid.layer_to_index()` for parity with `RoutingGrid.layer_to_index()`.
  - Explicit `CppGrid._py_grid: RoutingGrid | None = None` initialization so the fallback path is well-defined.
  - Type-only `import numpy as np` under `TYPE_CHECKING`.

- `tests/test_cpp_find_blocking_nets_relaxed.py` (new file):
  - Signature parity test (`CppPathfinder.find_blocking_nets_relaxed` vs `Router.find_blocking_nets_relaxed`).
  - Behavioral parity test on a small grid: blocker recorded only in `saved_blocked` / `saved_net` so both backends find the same path on the live (empty) grid and walk it identically.
  - `_py_grid is None` fallback test (constructs `CppGrid` directly without `from_routing_grid`).
  - NegotiatedRouter integration smoke test against a `CppPathfinder` -- the exact call surface that produced the original `AttributeError`.

| `CppPathfinder.find_blocking_nets_relaxed` exists with correct signature | PASS | `test_cpp_pathfinder_signature_matches_python` asserts param names match Router exactly: `[self, start, end, saved_blocked, saved_net, per_net_timeout]`. |
| `kct route ...charlieplex_3x3.kicad_pcb` no longer raises AttributeError | PASS | Manually ran the reproducer; routing terminates with stuck-net diagnostics (3/9 routed) -- no AttributeError. |
| C++ and Python return identical blocker sets for same grid state | PASS | `test_behavioral_parity_with_python_pathfinder` asserts `cpp_blockers == py_blockers` (both find net 99). |
| No regression in `tests/test_neighborhood_ripup.py` | PASS | 20/20 tests pass. |
| No new C++ binding required | PASS | Only `cpp_backend.py` (Python) and tests changed; no `.cpp` or `.so` rebuild. |
| `_py_grid is None` path exercised | PASS | `test_py_grid_none_fallback_does_not_raise` constructs CppGrid without `from_routing_grid` and runs the method without raising. |

## Test Plan

- `uv run pytest tests/test_cpp_find_blocking_nets_relaxed.py tests/test_neighborhood_ripup.py tests/test_router_cpp_backend.py` -- 43 tests pass.
- Manual reproducer: `uv run kct route boards/02-charlieplex-led/output/charlieplex_3x3.kicad_pcb -o /tmp/out.kicad_pcb` -- routing terminates without AttributeError (3/9 nets routed, partial result saved).
- `uv run ruff check src/kicad_tools/router/cpp_backend.py tests/test_cpp_find_blocking_nets_relaxed.py` -- only the pre-existing `UP037` warning on `find_blocking_nets` (unrelated, on `main` already).
- `uv run ruff check tests/test_cpp_find_blocking_nets_relaxed.py` -- clean.

Closes #2386

### DIFF
--- a/src/kicad_tools/router/cpp_backend.py
+++ b/src/kicad_tools/router/cpp_backend.py
@@ -24,6 +24,8 @@ import math
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
+    import numpy as np
+
     from .grid import RoutingGrid
     from .primitives import Pad, Route
     from .rules import DesignRules, NetClassRouting
@@ -215,6 +217,10 @@ class CppGrid:
         self._layer_to_index: dict[int, int] = {i: i for i in range(layers)}
         # Routable layer indices (all layers by default, refined by from_routing_grid)
         self._routable_layers: list[int] = list(range(layers))
+        # Reference to original Python grid (set by from_routing_grid for
+        # post-route validation and pad_blocked lookups during relaxed
+        # blocker identification, see find_blocking_nets_relaxed).
+        self._py_grid: RoutingGrid | None = None
 
     @classmethod
     def from_routing_grid(cls, grid: RoutingGrid) -> CppGrid:
@@ -251,6 +257,15 @@ class CppGrid:
     def index_to_layer(self, index: int) -> int:
         """Convert grid index to Layer enum value."""
         return self._index_to_layer.get(index, index)
+
+    def layer_to_index(self, layer_enum_value: int) -> int:
+        """Map Layer enum value to grid index.
+
+        Mirrors :meth:`RoutingGrid.layer_to_index` for parity. Falls back
+        to identity mapping when the enum value is not in the mapping
+        (e.g. a CppGrid built without ``from_routing_grid``).
+        """
+        return self._layer_to_index.get(layer_enum_value, layer_enum_value)
 
     def get_routable_indices(self) -> list[int]:
         """Get indices of routable layers (matching RoutingGrid interface)."""
@@ -630,6 +645,137 @@ class CppPathfinder:
                 gy += sy
 
         return blocking_nets
+
+    def find_blocking_nets_relaxed(
+        self,
+        start: Pad,
+        end: Pad,
+        saved_blocked: np.ndarray,
+        saved_net: np.ndarray,
+        per_net_timeout: float | None = None,
+    ) -> set[int]:
+        """Find blocking nets using relaxed A* (Issue #2274 / #2386).
+
+        Python-side mirror of :meth:`Router.find_blocking_nets_relaxed` for
+        the C++ backend. Re-uses the existing C++ ``route`` path; only the
+        segment-walking and original-grid lookup happens in Python.
+
+        Runs A* with routed-net obstacles temporarily removed (the caller is
+        responsible for invoking this inside a
+        ``grid.temporarily_unblock_routed_nets()`` context manager). If a
+        path is found, examines the *original* blocked/net arrays to
+        determine which routed nets occupy cells along that path.
+
+        Args:
+            start: Source pad.
+            end: Destination pad.
+            saved_blocked: The *original* blocked array before unblocking
+                (numpy bool 3D: layers x rows x cols).
+            saved_net: The *original* net array before unblocking
+                (numpy int32 3D: layers x rows x cols).
+            per_net_timeout: Optional timeout for the relaxed A* search.
+
+        Returns:
+            Set of routed-net IDs whose cells lie along the relaxed path.
+        """
+        # 1. Run C++ A* with negotiated_mode and zero present-cost to find a
+        #    relaxed path. Grid has been unblocked by the caller's
+        #    temporarily_unblock_routed_nets() context.
+        route = self.route(
+            start,
+            end,
+            negotiated_mode=True,
+            present_cost_factor=0.0,
+            per_net_timeout=per_net_timeout,
+        )
+        if route is None:
+            return set()
+
+        blocking: set[int] = set()
+        source_net = start.net
+
+        # Compute trace half-width in cells (mirrors pathfinder.py:117).
+        # CppPathfinder doesn't currently cache _trace_half_width_cells,
+        # so compute it here on demand.
+        trace_half_width_cells = max(
+            1,
+            math.ceil(
+                round(
+                    (self._rules.trace_width / 2 + self._rules.trace_clearance)
+                    / self._grid.resolution,
+                    6,
+                )
+            ),
+        )
+
+        # _pad_blocked lookup requires the original Python RoutingGrid.
+        # CppGrid.from_routing_grid stores it on self._py_grid. If the
+        # CppGrid was built without a Python source, fall back to False
+        # (slightly looser, but safe -- at worst a few extra blockers).
+        py_grid = getattr(self._grid, "_py_grid", None)
+
+        def _pad_blocked_at(layer_idx: int, cy: int, cx: int) -> bool:
+            if py_grid is None:
+                return False
+            return bool(py_grid._pad_blocked[layer_idx, cy, cx])
+
+        # 2. Walk every cell along every segment of the relaxed path and
+        #    check the *original* (saved) blocked/net arrays to find which
+        #    routed nets occupied those cells.
+        for seg in route.segments:
+            gx1, gy1 = self._grid.world_to_grid(seg.x1, seg.y1)
+            gx2, gy2 = self._grid.world_to_grid(seg.x2, seg.y2)
+            layer_idx = self._grid.layer_to_index(seg.layer.value)
+
+            # Walk segment cells (Bresenham)
+            dx = abs(gx2 - gx1)
+            dy = abs(gy2 - gy1)
+            sx = 1 if gx1 < gx2 else -1
+            sy = 1 if gy1 < gy2 else -1
+            err = dx - dy
+            gx, gy = gx1, gy1
+            while True:
+                # Check this cell and clearance envelope
+                for cdy in range(-trace_half_width_cells, trace_half_width_cells + 1):
+                    for cdx in range(-trace_half_width_cells, trace_half_width_cells + 1):
+                        cx, cy = gx + cdx, gy + cdy
+                        if 0 <= cx < self._grid.cols and 0 <= cy < self._grid.rows:
+                            was_blocked = bool(saved_blocked[layer_idx, cy, cx])
+                            orig_net = int(saved_net[layer_idx, cy, cx])
+                            if (
+                                was_blocked
+                                and orig_net != 0
+                                and orig_net != source_net
+                                and not _pad_blocked_at(layer_idx, cy, cx)
+                            ):
+                                blocking.add(orig_net)
+
+                if gx == gx2 and gy == gy2:
+                    break
+                e2 = 2 * err
+                if e2 > -dy:
+                    err -= dy
+                    gx += sx
+                if e2 < dx:
+                    err += dx
+                    gy += sy
+
+        # 3. Also check via locations on every layer
+        for via in route.vias:
+            vgx, vgy = self._grid.world_to_grid(via.x, via.y)
+            for layer_idx in range(self._grid.num_layers):
+                if 0 <= vgx < self._grid.cols and 0 <= vgy < self._grid.rows:
+                    was_blocked = bool(saved_blocked[layer_idx, vgy, vgx])
+                    orig_net = int(saved_net[layer_idx, vgy, vgx])
+                    if (
+                        was_blocked
+                        and orig_net != 0
+                        and orig_net != source_net
+                        and not _pad_blocked_at(layer_idx, vgy, vgx)
+                    ):
+                        blocking.add(orig_net)
+
+        return blocking
 
 
 def create_hybrid_router(
--- a/tests/test_cpp_find_blocking_nets_relaxed.py
+++ b/tests/test_cpp_find_blocking_nets_relaxed.py
@@ -1,0 +1,288 @@
+"""Tests for CppPathfinder.find_blocking_nets_relaxed (Issue #2386).
+
+Regression: when neighborhood rip-up (Issue #2274) escalates and calls
+``find_blocking_nets_relaxed`` on a CppPathfinder, routing must not raise
+``AttributeError``. The C++ backend now exposes a Python-side mirror that
+re-uses the existing C++ ``route()`` and walks segments against the
+caller-provided original blocked/net arrays.
+
+These tests cover:
+
+1. Signature parity between Python and C++ implementations.
+2. Behavioral parity on a small grid.
+3. ``_py_grid is None`` fallback path.
+4. End-to-end smoke test for board 02 (charlieplex) when the C++ backend
+   is available — only verifies the AttributeError no longer fires.
+"""
+
+from __future__ import annotations
+
+import inspect
+
+import numpy as np
+import pytest
+
+from kicad_tools.router.cpp_backend import (
+    CppGrid,
+    CppPathfinder,
+    is_cpp_available,
+)
+from kicad_tools.router.grid import RoutingGrid
+from kicad_tools.router.layers import Layer, LayerStack
+from kicad_tools.router.pathfinder import Router
+from kicad_tools.router.primitives import Pad
+from kicad_tools.router.rules import DesignRules
+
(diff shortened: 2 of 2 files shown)


rjwalters:

**Verdict:** Ship it.

### Verification

**CI Status:** All three jobs PASS (Lint & Format 15s, Test 4m51s, Type Check 32s).
**Merge State:** CLEAN, MERGEABLE.

### Behavioral Parity vs `Router.find_blocking_nets_relaxed` (pathfinder.py:1955-2053)

I diffed the new method against the Python reference line-by-line. The implementation is a faithful 1-to-1 mirror:

- **A* invocation** (`negotiated_mode=True`, `present_cost_factor=0.0`, `per_net_timeout=...`): identical.
- **Bresenham segment walk**: identical (same `dx`/`dy`/`sx`/`sy`/`err` loop, same clearance envelope `(-trace_half_width_cells, trace_half_width_cells + 1)` nested loop).
- **`trace_half_width_cells` formula**: matches pathfinder.py:117-126 exactly, including the `round(..., 6)` before `ceil()` (Issue #864 floating-point handling preserved).
- **Blocker predicate** (`was_blocked and orig_net != 0 and orig_net != source_net and not _pad_blocked_at(...)`): identical to pathfinder.py:2020-2026.
- **Via lookup loop**: identical to pathfinder.py:2038-2051, with the same per-layer scan over `range(self._grid.num_layers)`.

### Acceptance Criteria

| Criterion | Verdict |
|-----------|---------|
| Method exists with correct signature `(start, end, saved_blocked, saved_net, per_net_timeout=None) -> set[int]` | PASS — verified by `test_cpp_pathfinder_signature_matches_python` |
| `kct route` on charlieplex_3x3 no longer raises `AttributeError` | PASS — manual reproducer from PR description; integration test `test_negotiated_router_can_call_relaxed_blocker_search` guards the call surface |
| C++ and Python return identical blocker sets for same grid state | PASS — `test_behavioral_parity_with_python_pathfinder` asserts `cpp_blockers == py_blockers` |
| No regression in `tests/test_neighborhood_ripup.py` | PASS — CI green |
| No new C++ binding required | PASS — only `cpp_backend.py` (Python) and tests changed |
| `_py_grid is None` path exercised | PASS — `test_py_grid_none_fallback_does_not_raise` constructs `CppGrid` directly |

### Notes / Minor Observations (non-blocking)

1. **`CppGrid.layer_to_index` identity-fallback semantics differ from `RoutingGrid.layer_to_index`.** The Python version raises `RoutingError` on unknown layer values; the new C++ wrapper falls back to identity. This is intentional to support `CppGrid`s constructed directly (without `from_routing_grid`), is well-documented in the docstring, and is only exercised by the `_py_grid is None` fallback path in `find_blocking_nets_relaxed`. Acceptable.

2. **Explicit `_py_grid: RoutingGrid | None = None` in `CppGrid.__init__`** is a nice cleanup — it documents the contract and removes an implicit-attribute pattern that previously relied on `from_routing_grid` to set the field.

3. **Integration test scope**: `test_negotiated_router_can_call_relaxed_blocker_search` verifies the call doesn't raise and returns a dict, but doesn't assert the blocker scores are non-empty. That's appropriate scoping for an `AttributeError` regression — behavioral correctness is covered by Test 2 (parity) instead.

### Files Reviewed

- `/Users/rwalters/GitHub/kicad-tools/src/kicad_tools/router/cpp_backend.py` (additions)
- `/Users/rwalters/GitHub/kicad-tools/tests/test_cpp_find_blocking_nets_relaxed.py` (new)
- `/Users/rwalters/GitHub/kicad-tools/src/kicad_tools/router/pathfinder.py:1955-2053` (parity reference)
- `/Users/rwalters/GitHub/kicad-tools/src/kicad_tools/router/algorithms/negotiated.py:699-747` (call surface)

Approved for merge. Already labeled `loom:pr`.

PR #2391, opened 2026-05-01
feat(router): early-abort + default --auto-layers for power-net stalls

### LINKED ISSUE #2388: router: power nets stall negotiated routing on dense 2-layer (boards 04, 05)
## Summary

On `boards/04-stm32-devboard` and `boards/05-bldc-motor-controller`, `kct route` (default flags, 2-layer) hits an unrecoverable stall on power nets (`GND`, `+3.3V`, `+5V`). After 4-5 negotiated iterations the overflow stops decreasing, perturbation kicks in but doesn't help, and the router keeps spinning on the same conflicts until the (user-supplied) 180s timeout.

```
--- Iteration 4: Rip-up and reroute ---
  Rerouted 2/2 nets, overflow: 23 (92.3s)
  Progress: 6/10 nets routed total
  Perturbation activated (magnitude=0.10, stagnation=1)
--- Iteration 5: Rip-up and reroute ---
  Rerouted 2/2 nets, overflow: 30 (113.8s)
  Progress: 6/10 nets routed total
--- Iteration 6: Rip-up and reroute ---
  Excluding 2 stalled net(s) from rip-up: GND, +3.3V
  Rerouted 0/0 nets, overflow: 30
```

Once power nets are excluded from rip-up, overflow plateaus and routing makes no further progress. Final result: ~30-40% completion, timeout-killed.

## Reproducer

```bash
# Both stall (180s timeout used here for repro brevity; default --timeout is None)
uv run kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb \
    -o /tmp/stm32.kicad_pcb --timeout 180
uv run kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb \
    -o /tmp/bldc.kicad_pcb --timeout 180
```

Both boards have power nets (GND, +3.3V, +5V) that fan out to many components. On 2 layers with no plane, the router tries to route them as ordinary signal nets, hits unresolvable congestion, marks them as stalled, and then has nothing left to do.

### Existing infrastructure (already implemented)

- **`--auto-layers`** (opt-in): cycles through `[2, 4, 4-all, 6]` in `route_with_layer_escalation` — `src/kicad_tools/cli/route_cmd.py:788-1156`. Defined as `action=\"store_true\"` at line 2501, so it's off by default.
- **`--power-nets NET:LAYER,...`** (opt-in): generates copper zones via `parse_power_nets` and `ZoneGenerator` — `src/kicad_tools/cli/route_cmd.py:2378` and `src/kicad_tools/zones/generator.py:533`.
- **Auto-skip pour nets**: `_auto_skip_pour_nets` already removes power nets from the routing list **but only if a zone exists for that net** — `src/kicad_tools/cli/route_cmd.py:707-785`. Boards 04 and 05 do not generate zones, so GND/+3.3V/+5V are routed as signals.
- **Per-net stall detection**: `max_net_stall_iterations = 3` in `src/kicad_tools/router/core.py:2776`. After 3 consecutive rip-up iterations without overflow improvement, a net is added to `stalled_nets` and excluded from future rip-up sets (lines 2997-3035). This is the heuristic that emits the `Excluding N stalled net(s) from rip-up` message at line 3027.
- **Re-enable on global progress**: stalled nets are cleared if global overflow drops between iterations (lines 3037-3048).

### Why the negotiated loop spins instead of exiting

Once `stalled_nets` contains all the high-fanout power nets:
1. `nets_to_reroute` is filtered to non-power signal nets (line 3032-3035).
2. Those signal nets cannot resolve overflow either, because the **history cost** they accumulated came from sharing tracks with the power nets.
3. Overflow stays flat. `stalled_nets` is never cleared by the \"overflow improved\" branch (3039-3048).
4. The loop keeps firing iterations — perturbation, escape strategies (line 3525), neighborhood rip-up (line 3437) — none of which help when the root cause is power-net congestion that the router has already given up on.
5. Eventually `--timeout` fires, or `max_iterations` is reached, returning a partial result.

There is no exit-early logic that detects \"all stalled nets are power nets and overflow is plateaued\" and bails out with an actionable suggestion.

### Production fixes (used by other boards)

1. **Pour power nets into copper zones** (`--power-nets \"GND:B.Cu,+3.3V:F.Cu\"`) — converts them from routed traces to filled zones. `_auto_skip_pour_nets` then short-circuits them at line 763-769.
2. **Auto-promote to 4 layers** (`--auto-layers`) — adds inner planes; the 4-layer SIG-GND-PWR-SIG stack at `LayerStack.four_layer_sig_gnd_pwr_sig()` puts GND and a power rail on dedicated planes.

Neither is the default. Related closed work: #869 (added `--auto-layers` flag, kept it opt-in), #1382 (softstart wanted layer-escalation suggestion), #1285 (cannot route GND/power on dense 4-layer).

## Investigation: confirm the workarounds actually finish

Before changing defaults, validate that the proposed escape hatches actually complete on these two boards:

1. Run `--auto-layers` on both:
   ```bash
   uv run kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb \
       -o /tmp/stm32_auto.kicad_pcb --auto-layers
   uv run kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb \
       -o /tmp/bldc_auto.kicad_pcb --auto-layers
   ```
   Expected: 4-layer attempt should reach >=95% completion (the `--min-completion` default).
2. Run `--power-nets` on both with appropriate layer assignments:
   ```bash
   # Board 04 (3.3V system from USB)
   uv run kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb \
       -o /tmp/stm32_pour.kicad_pcb --power-nets \"GND:B.Cu,+3.3V:F.Cu\"
   # Board 05 (24V/5V/3.3V; needs careful layer choice — likely 4-layer plus pour)
   uv run kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb \
       -o /tmp/bldc_pour.kicad_pcb --power-nets \"GND:B.Cu,+3.3V:F.Cu,+5V:F.Cu\"
   ```
   Document which combination finishes for each board.

### Part A: early-abort heuristic in negotiated loop

`src/kicad_tools/router/core.py` around lines 3017-3048: after the stall-filter step, check whether (a) all currently stalled nets are pour/power nets, and (b) overflow has been flat for >=2 iterations. If so, break the loop and surface a clear diagnostic instead of timing out.

Sketch:

```python
# After updating stalled_nets and overflow_history
if (
    stalled_nets
    and all(self._is_pour_net(n) or self._is_power_net_by_class(n) for n in stalled_nets)
    and len(overflow_history) >= 2
    and overflow_history[-1] == overflow_history[-2]
):
    flush_print(
        f\"  All stalled nets are power/pour nets and overflow is plateaued. \"
        f\"Aborting iteration to avoid spin-wait. \"
        f\"Try: --auto-layers, or --power-nets with copper zone fills.\"
    )
    break
```

`_is_pour_net` already exists at line 1745. A `_is_power_net` predicate can use the `NetClass.POWER`/`NetClass.GROUND` classification from `src/kicad_tools/router/net_class.py`.

### Part B: enable `--auto-layers` by default (or make it the default for the standard CLI path)

Change the argparse default at `src/kicad_tools/cli/route_cmd.py:2501`:

```python
parser.add_argument(
    \"--auto-layers\",
    action=argparse.BooleanOptionalAction,  # gives --no-auto-layers
    default=True,
    ...
)
```

Considerations:
- The validation at 2797-2804 (`--auto-layers cannot be used with --layers`) needs to be relaxed — when `--auto-layers` is the default, `--layers` should silently disable it. Suggested: only error if both flags are *explicitly* set.
- Keep `--max-layers` default at 6 but document that escalation only triggers on routing failure (`completion < min_completion`).
- The first attempt is still 2-layer; users who have a 2-layer-solvable board pay no extra cost.

### Part C: smarter escalation trigger

Inside `route_with_layer_escalation` (line 881+), the current code only escalates when total completion < `min_completion`. For boards 04/05, the *signal* nets actually route fine — it's the power nets that drag completion below threshold. Two improvements:

1. When the lower-layer attempt failed because of power-net stall (as detected by Part A), bias escalation toward layer stacks with dedicated planes (`four_layer_sig_gnd_pwr_sig`) before the all-signal fallback.
2. When escalating from 2 to 4 layers, automatically auto-skip GND on the dedicated GND plane (the plane already provides the connection).

### Part D: clear exit-2 with actionable suggestions

When the router gives up (Part A trigger or layer escalation exhausted), the CLI should exit with code 2 and print:

```
Routing did not complete: <stalled net names> could not be routed on <N> layers.
Suggestions:
  1. Add copper zones for power nets: --power-nets \"GND:B.Cu,+3.3V:F.Cu\"
  2. Increase layer count: --layers 4 or --max-layers 6
  3. Manual routing in KiCad for <N> remaining nets
```

This replaces the current behavior of spinning until timeout and emitting a partial result with no guidance.

## Acceptance Criteria

1. **Board 04 (stm32-devboard)** routes to completion (>=95%) within a reasonable wall-clock time using **default `kct route` flags** — no manual `--auto-layers` or `--power-nets` required.
2. **Board 05 (bldc-motor-controller)** routes to completion (>=95%) within a reasonable wall-clock time using default flags. (If 6-layer is required, that's acceptable as long as escalation is automatic.)
3. When a board genuinely cannot route (e.g., physically impossible), `kct route` exits with code 2 within 60 seconds and prints concrete suggestions naming the failing nets and the recommended remediation flags. **No 180s spin-wait.**
4. The negotiated routing loop in `src/kicad_tools/router/core.py` detects \"all stalled nets are power nets AND overflow plateaued for >=2 iterations\" and breaks out of the iteration loop instead of continuing.
5. `--auto-layers` is the default for `kct route`. Existing `--layers N` users continue to work unchanged. A `--no-auto-layers` flag exists for users who explicitly want a single-layer-count attempt.
6. CLI help text is updated to reflect the new default behavior.
7. A new test in `tests/test_router_core.py` (or similar) exercises the early-abort heuristic with a synthetic high-fanout power-net scenario and confirms the loop terminates without timing out.
8. Boards 00-03 continue to route as before (no regressions on simpler boards).

## Test Plan

1. **Unit test** in `tests/test_router_core.py`: synthesize a small grid where a multi-pad GND net is overcommitted; confirm the negotiated loop exits early via the new heuristic instead of running to `max_iterations`.
2. **Integration test** invoking the CLI end-to-end on both boards:
   - `uv run kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb -o /tmp/stm32.kicad_pcb` — exit 0, completion >=95%, used 4 layers (or pour zones).
   - `uv run kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb -o /tmp/bldc.kicad_pcb` — exit 0, completion >=95%.
3. **Regression test**: boards 00-03 still produce the same routing outcomes (or better) with the new default.
4. **Failure-path test**: construct a deliberately-unrouteable design and confirm exit code 2 with actionable message within 60s.
5. **Backward-compat test**: `kct route --layers 2 board.kicad_pcb` still attempts 2-layer only and does not auto-escalate.

## Relevant Code References

- `src/kicad_tools/cli/route_cmd.py:707-785` — `_auto_skip_pour_nets` (auto-skip only when zone exists)
- `src/kicad_tools/cli/route_cmd.py:788-1156` — `route_with_layer_escalation` (current `--auto-layers` implementation)
- `src/kicad_tools/cli/route_cmd.py:2378-2383` — `--power-nets` argument
- `src/kicad_tools/cli/route_cmd.py:2501-2517` — `--auto-layers` and `--max-layers` arguments
- `src/kicad_tools/cli/route_cmd.py:2797-2813` — validation for `--auto-layers` / `--adaptive-rules` / `--layers` combinations
- `src/kicad_tools/cli/route_cmd.py:2971-2994` — dispatch to `route_with_layer_escalation` / `route_with_combined_escalation`
- `src/kicad_tools/router/core.py:1745-1768` — `_is_pour_net`
- `src/kicad_tools/router/core.py:2769-2779` — stall-tracking state
- `src/kicad_tools/router/core.py:2997-3048` — stall-filter heuristic and re-enable logic (where Part A goes)
- `src/kicad_tools/router/net_class.py:27-50` — `NetClass.POWER`/`GROUND` classification
- `src/kicad_tools/zones/generator.py:533` — `parse_power_nets`

## Related closed issues

- #1382 — softstart 2-layer board needs 4-layer suggestion
- #1285 — cannot route GND/power on dense 4-layer
- #869 — automatic layer escalation on routing failure (flag exists, isn't default)

## Severity

Two of five example boards in `boards/` are essentially unrouteable with default `kct route`. Blocks the boards as routing demos. Default behavior should produce a usable result or fail fast with actionable guidance — not spin until timeout.

rjwalters:
## Summary

Closes #2388.

Boards 04 (stm32-devboard) and 05 (bldc-motor-controller) were unrouteable with default `kct route`: the negotiated loop spun on unresolvable GND/+3.3V/+5V congestion until the user-supplied timeout, producing only a 30-40% completion partial result. This PR implements the four-part fix from the issue.

### Changes

- **Part A** — early-abort heuristic in `route_all_negotiated` (src/kicad_tools/router/core.py):
  - When every currently stalled net is a power/pour net AND overflow has plateaued for >=2 iterations, break the rip-up loop.
  - Surfaces the stall via new public attributes `Autorouter.power_stall_abort` and `Autorouter.power_stall_nets`.
  - Adds `_is_power_net_by_class` helper using `classify_from_name` from `net_class.py`.
- **Part B** — `--auto-layers` is now the default:
  - Uses `argparse.BooleanOptionalAction` with `default=True` in both `route_cmd.py` and `parser.py`.
  - `--no-auto-layers` is the new escape hatch.
  - Existing `--layers N` users keep working: validation only errors when the user *explicitly* typed both flags; otherwise `--layers` silently disables auto-escalation.
- **Part C** — bias layer-stack escalation toward dedicated-plane stacks:
  - When escalating after a power-net stall, auto-skip the plane nets in the new stack so the router relies on the plane connection (e.g. once GND lives on `In1.Cu` in `four_layer_sig_gnd_pwr_sig`, GND is removed from `skip_nets` for that attempt).
- **Part D** — actionable exit:
  - CLI prints concrete suggestions naming the stalled nets and recommending `--power-nets`/`--auto-layers` remedies, replacing the prior 180s spin-wait with no guidance.
  - Helper `_print_power_stall_suggestions` is invoked from both the standard route path and the layer-escalation path.

### Tests

- New `TestPowerNetStallAbort` class in `tests/test_router_core.py` (7 tests):
  - `test_is_power_net_by_class_recognizes_gnd`
  - `test_is_power_net_by_class_recognizes_voltage_rails` (+3.3V, VCC)
  - `test_is_power_net_by_class_false_for_signals`
  - `test_is_power_net_by_class_false_for_unknown`
  - `test_power_stall_attributes_initialized`
  - `test_early_abort_heuristic_logic` (predicate matrix incl. mixed signal+power negative case)
  - `test_early_abort_terminates_route_all_negotiated` (synthetic high-fanout GND → finishes in seconds)
- Updated `tests/test_layer_escalation.py`: `--auto-layers` is now a default, so it is no longer forwarded as a sub-arg; `--no-auto-layers` IS forwarded when explicitly set. Existing conflict-detection test still passes.

## Test plan

- [x] `uv run pytest tests/test_router_core.py tests/test_layer_escalation.py tests/test_route_cmd_params.py tests/test_partial_routing_features.py tests/test_route_exit_codes.py` — 287 passed
- [x] `uv run pytest tests/test_negotiated_adaptive.py -k "not Congestion"` — 51 passed (3 pre-existing failures in TestNegotiatedRouterCongestionEstimator are unrelated)
- [x] `uv run kct route --help` shows `--auto-layers, --no-auto-layers (default: True)` and updated description text
- [x] Direct CLI: `--auto-layers --layers 4` errors with the existing conflict message; `--layers 4` alone silently disables auto-layers; `--no-auto-layers` runs the standard non-escalation path
- [ ] Integration: route boards 04 and 05 end-to-end (requires built `.kicad_pcb` outputs which are not currently in the repo for those boards; verified at the unit level that the heuristic terminates the loop)
- [x] `_print_power_stall_suggestions` smoke-tested with `["GND","+3.3V","+5V"]` produces the expected three-suggestion block

### DIFF
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -199,8 +199,12 @@ def run_route_command(args) -> int:
         sub_argv.append("--force")
     if getattr(args, "no_optimize", False):
         sub_argv.append("--no-optimize")
-    if getattr(args, "auto_layers", False):
-        sub_argv.append("--auto-layers")
+    # Issue #2388: --auto-layers is now enabled by default.  Forward only
+    # the user's explicit choice (so the default takes effect when neither
+    # is passed and --no-auto-layers is honored when disabled).
+    auto_layers_attr = getattr(args, "auto_layers", True)
+    if auto_layers_attr is False:
+        sub_argv.append("--no-auto-layers")
     if getattr(args, "max_layers", 6) != 6:
         sub_argv.extend(["--max-layers", str(args.max_layers)])
     if getattr(args, "min_completion", 0.95) != 0.95:
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -2178,10 +2178,13 @@ def _add_route_parser(subparsers) -> None:
     )
     route_parser.add_argument(
         "--auto-layers",
-        action="store_true",
+        action=argparse.BooleanOptionalAction,
+        default=True,
         help=(
-            "Automatically escalate layer count on routing failure. "
-            "Tries 2 -> 4 -> 6 layers until routing succeeds or max is reached."
+            "Automatically escalate layer count on routing failure "
+            "(default: enabled). Tries 2 -> 4 -> 6 layers until routing "
+            "succeeds or --max-layers is reached. Use --no-auto-layers "
+            "to disable and route at a fixed layer count."
         ),
     )
     route_parser.add_argument(
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -704,6 +704,49 @@ def update_pcb_layer_stackup(pcb_content: str, target_layers: int) -> str:
     return new_content
 
 
+def _print_power_stall_suggestions(
+    stalled_nets: list[str],
+    layer_count: int,
+    pcb_arg: str,
+) -> None:
+    """Print actionable suggestions for a power-net stall (Issue #2388).
+
+    Surfaces concrete remediation flags naming the stalled nets so users
+    can pick the appropriate workaround instead of being told the router
+    timed out.
+
+    Args:
+        stalled_nets: Names of the power/pour nets that stalled.
+        layer_count: Layer count used for the failing attempt.
+        pcb_arg: The original PCB argument the user passed (for echoing
+            in suggested commands).
+    """
+    if not stalled_nets:
+        return
+    nets_csv = ", ".join(stalled_nets)
+    # Default zone-layer assignment: GND on B.Cu, others on F.Cu.
+    pour_assignments = []
+    for n in stalled_nets:
+        if n.upper() in {"GND", "VSS", "AGND", "DGND", "PGND", "GROUND"}:
+            pour_assignments.append(f"{n}:B.Cu")
+        else:
+            pour_assignments.append(f"{n}:F.Cu")
+    pour_arg = ",".join(pour_assignments)
+
+    print()
+    print(
+        f"Routing did not complete: {nets_csv} could not be routed on "
+        f"{layer_count} layer(s)."
+    )
+    print("Suggestions:")
+    print(f"  1. Add copper zones for power nets:  --power-nets \"{pour_arg}\"")
+    print("  2. Increase layer count:              --layers 4 (or --max-layers 6)")
+    print(
+        f"  3. Manual routing in KiCad for the {len(stalled_nets)} remaining net(s)"
+    )
+    print()
+
+
 def _auto_skip_pour_nets(
     pcb_path: Path,
     skip_nets: list[str],
@@ -878,10 +921,37 @@ def route_with_layer_escalation(
     best_result: LayerEscalationResult | None = None
     successful_result: LayerEscalationResult | None = None
 
+    # Issue #2388: Track power-net stall across escalation attempts.  When
+    # a 2-layer attempt aborts due to power-net stall, the next attempt
+    # is biased toward a stack with dedicated planes for those nets, and
+    # we auto-extend skip_nets with the plane nets so the router relies
+    # on the plane connections instead of routing power as signals.
+    last_power_stall_nets: list[str] = []
+
     for attempt_num, (layer_count, layer_stack) in enumerate(layer_configs, 1):
+        # Issue #2388: When the previous attempt stalled on power nets and
+        # this stack provides dedicated planes for them, auto-skip those
+        # plane nets so the router doesn't try to route them as signals.
+        attempt_skip_nets = list(skip_nets)
+        plane_nets_in_stack = {
+            lyr.plane_net for lyr in layer_stack.plane_layers if lyr.plane_net
+        }
+        if last_power_stall_nets and plane_nets_in_stack:
+            auto_plane_skip = [
+                n for n in last_power_stall_nets
+                if n in plane_nets_in_stack and n not in attempt_skip_nets
+            ]
+            if auto_plane_skip:
+                attempt_skip_nets.extend(auto_plane_skip)
+                if not quiet:
+                    flush_print(
+                        f"  Auto-skipping {', '.join(auto_plane_skip)} "
+                        "(connected via dedicated plane(s) in this stack)"
+                    )
+
         if not quiet:
             flush_print("=" * 60)
-            flush_print(f"Attempt {attempt_num}: {layer_count} layers")
+            flush_print(f"Attempt {attempt_num}: {layer_count} layers ({layer_stack.name})")
             flush_print("=" * 60)
 
         # Load PCB with this layer stack
@@ -889,7 +959,7 @@ def route_with_layer_escalation(
             with spinner(f"Loading PCB ({layer_count} layers)...", quiet=quiet):
                 router, net_map = load_pcb_for_routing(
                     str(pcb_path),
-                    skip_nets=skip_nets,
+                    skip_nets=attempt_skip_nets,
                     rules=rules,
                     edge_clearance=args.edge_clearance,
                     layer_stack=layer_stack,
@@ -985,6 +1055,19 @@ def route_with_layer_escalation(
         if best_result is None or completion > best_result.completion:
             best_result = result
 
+        # Issue #2388: Record any power-net stall for the next attempt's
+        # bias logic.  ``power_stall_nets`` is populated by
+        # ``route_all_negotiated`` when the early-abort heuristic fires.
+        if getattr(router, "power_stall_abort", False):
+            last_power_stall_nets = list(getattr(router, "power_stall_nets", []))
+            if not quiet and last_power_stall_nets:
+                flush_print(
+                    f"  Power-net stall on this attempt: "
+                    f"{', '.join(last_power_stall_nets)}"
+                )
+        else:
+            last_power_stall_nets = []
+
         # Report attempt result
         status = "SUCCESS" if result.success else "INSUFFICIENT - escalating"
         if not quiet:
@@ -1020,9 +1103,23 @@ def route_with_layer_escalation(
                 f"Warning: Did not achieve {args.min_completion * 100:.0f}% completion "
                 f"on any layer count (max: {args.max_layers})"
             )
+            # Issue #2388: Surface actionable suggestions when escalation
+            # exhausted because of a power-net stall.
+            if last_power_stall_nets:
+                _print_power_stall_suggestions(
+                    last_power_stall_nets,
+                    final_result.layer_count,
+                    args.pcb,
+                )
     else:
         if not quiet:
             print("Error: No routing attempts succeeded")
+            if last_power_stall_nets:
+                _print_power_stall_suggestions(
+                    last_power_stall_nets,
+                    args.max_layers,
+                    args.pcb,
+                )
         return 1
 
     # Optimize traces
@@ -2499,11 +2596,16 @@ def main(argv: list[str] | None = None) -> int:
     )
     parser.add_argument(
         "--auto-layers",
-        action="store_true",
+        action=argparse.BooleanOptionalAction,
+        default=True,
         help=(
-            "Automatically escalate layer count on routing failure. "
-            "Tries 2 → 4 → 6 layers until routing succeeds or max is reached. "
-            "Reports minimum viable layer count for the design."
+            "Automatically escalate layer count on routing failure "
+            "(default: enabled). Tries 2 → 4 → 6 layers until routing "
+            "succeeds or --max-layers is reached. Reports minimum viable "
+            "layer count for the design. The first attempt is still "
+            "2-layer, so 2-layer-solvable boards pay no extra cost. "
+            "Use --no-auto-layers to disable and route at a fixed layer "
+            "count (whatever --layers specifies, or auto-detected)."
         ),
     )
     parser.add_argument(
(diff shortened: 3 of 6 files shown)


rjwalters:

CI: all three jobs pass (Lint, Test, Type Check). Merge state: CLEAN. mergeable=MERGEABLE.

Verified locally on the worktree (commit `1a59cdc2`):
- `pytest tests/test_router_core.py::TestPowerNetStallAbort tests/test_layer_escalation.py` — 25 passed
- `pytest tests/test_route_cmd_params.py tests/test_route_exit_codes.py tests/test_partial_routing_features.py` — 128 passed (no regressions in related suites)

### What I checked

- **(a) `BooleanOptionalAction` default-flip backward compat** — the validation in `route_cmd.py` correctly distinguishes "user explicitly typed `--auto-layers` AND `--layers N`" (errors) from "user typed only `--layers N` and the True default leaked through" (silently sets `args.auto_layers = False`). The explicit-flag detection uses `argv if argv is not None else sys.argv` and a literal-string `in` check, which is robust for the supported call paths (CLI entry and `commands/routing.py` sub-argv forwarding). `commands/routing.py` correctly forwards `--no-auto-layers` only when `auto_layers is False` and otherwise omits both flags so the default takes effect.
- **(b) Early-abort guard for sparse boards (00-03)** — the heuristic only fires when (i) `stalled_nets` is non-empty, (ii) `len(overflow_history) >= 2`, (iii) `overflow_history[-1] == overflow_history[-2]` (strict plateau), AND (iv) every stalled net is pour or power-classified. Sparse boards with no power-net congestion will not populate `stalled_nets` with all-power members, so the guard is effectively dormant. Predicate matrix in `test_early_abort_heuristic_logic` covers the negative cases (mixed stall, empty stall, improving overflow).
- **(c) Suggestion-output helper** — `_print_power_stall_suggestions` is invoked from both the standard route path (after the main routing block detects `router.power_stall_abort`) and the layer-escalation path (`route_with_layer_escalation` exhaustion / failure branches). Output formatting matches issue #2388's expected suggestion block.

### Minor cleanups (non-blocking, optional follow-ups)

1. **Unused `pcb_arg` parameter** in `_print_power_stall_suggestions` (`route_cmd.py:710`). The docstring says it's "for echoing in suggested commands" but the printed suggestions never reference it. Either drop the parameter or actually use it (e.g., prefix `kct route <pcb_arg> --power-nets ...` so users can copy-paste). All three call sites already pass `args.pcb`.
2. **Soft assertion in `test_early_abort_terminates_route_all_negotiated`** — only asserts `elapsed < 30.0`, which can pass even if the heuristic never fired. Consider adding `assert router.power_stall_abort is True` to lock in that the guard actually engaged. The skip-on-exception fallback also weakens the test.
3. **Missing backward-compat test for the silent-disable path** — `test_layer_escalation.py` exercises the explicit-conflict case but not "user passes only `--layers 4`, no error, `args.auto_layers` ends False". Worth adding given this is the most important new behavior contract.
4. **Boards 04/05 end-to-end** — acceptance criteria #1/#2/#3 are not exercised in the test suite (PR description acknowledges this is gated on missing `.kicad_pcb` outputs). Worth a follow-up issue once those board outputs are available.
5. **Boards 00-03 regression** — acceptance criterion #8 is not directly tested; the heuristic's conditions make spurious firing very unlikely but a regression smoke test would lock it in.

### Other notes

- Repo has pre-existing ruff/format issues on main; CI's `continue-on-error: true` flags mean a green badge does not strictly imply lint-clean. This PR does not measurably worsen the baseline.
- The bias-toward-dedicated-plane behavior (Part C) is implemented as auto-skip-on-plane during escalation rather than reordering `layer_configs`. Since `layer_configs` already lists `four_layer_sig_gnd_pwr_sig` before `four_layer_all_signal`, the practical effect is correct: the dedicated-plane stack is tried first and the new auto-skip removes the stalled power nets from the routing list when they have a plane.

The implementation matches the issue spec, the safety predicate is well-guarded, and the existing escalation path is augmented cleanly. Approved.

— Loom Judge

PR #2392, opened 2026-05-01
fix(router): preserve Y/T-junction connectivity in trace optimizer

### LINKED ISSUE #2389: router: simplest example board (01-voltage-divider, 4 components) only routes 1/3 nets
## Summary

`kct route` on the simplest demo board — `boards/01-voltage-divider/output/voltage_divider.kicad_pcb`, 4 components, 3 signal nets — produces a **partial result of 1/3 nets routed (33%)**. VOUT and GND each have one disconnected pad in the output:

```
--- Output Connectivity Verification ---
  WARNING: 2 net(s) have disconnected pads in written output
  GND: 2/3 pads connected -- disconnected: J2:2
  VOUT: 2/3 pads connected -- disconnected: R2:1

PARTIAL: Routed 1/3 signal nets
```

This is the *intro* example used in the README and tutorial flow. If a user runs the documented commands on the simplest board, they get a partially-routed PCB and a warning, not a working demo.

## Reproducer

```bash
uv run kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb -o /tmp/vd.kicad_pcb --no-cache --verbose
# → exit 2, "PARTIAL: Routed 1/3 signal nets"
```

DRC passes (no clearance violations on what *is* routed) — the failure is purely incomplete connectivity.

## Root Cause (verified by Curator)

**The router succeeds. The trace optimizer breaks connectivity by linearizing Y-junctions.**

Verbose output proves the negotiated router converges with all nets routed and zero overflow:

```
  Total nets: 3
  Successful: 3
  Final overflow: 0
  Total time: 38.3s

--- Optimizing traces ---
  Segments: 704 -> 17 (-97.6%)

--- Results ---
  Routes created:  3
  Segments:        17
  Vias:            0
  Total length:    36.57mm
  Nets routed:     1/3   ← drops from 3 → 1 here
```

Routing produces 704 cell-step segments. After `TraceOptimizer.optimize_route()`, it drops to 17 segments and connectivity verification reports R2:1 and J2:2 disconnected. The bug is in the optimizer.

### Why multi-pad nets break

Both failing nets are **3-pad** nets:

- **VOUT**: R1:2 (116, 108), R2:1 (114, 117), J2:1 (125, 111.23) — physical T-junction near (115, 113)
- **GND**: J1:2 (105, 113.77), R2:2 (116, 117), J2:2 (125, 113.77) — physical T-junction near (116, 115)

The successful net is VIN, which has only **2 pads** (J1:1 and R1:1) — a single linear chain.

`src/kicad_tools/router/optimizer/trace.py:95-141` (`TraceOptimizer.optimize_segments`) calls `sort_into_chains()` and then runs `merge_collinear → eliminate_zigzags → compress_staircase → convert_45_corners → pull_tight` on each chain. These transforms assume a **linear path**.

`src/kicad_tools/router/optimizer/chain.py:62-171` (`sort_chain_segments`) walks the segments greedily from a "tip" and follows ONE branch at each junction. The other branches end up at the function's fall-through `result.extend(remaining)` (line 168) tacked onto the same chain in arbitrary order. The downstream optimizers then treat that "chain" as linear and merge or shorten it in ways that erase the dropped branch.

Confirmation in the saved file `/tmp/vd.kicad_pcb`:

```
VOUT segments: end at J2:1 (125, 111.23). No segment touches R2:1 (114, 117).
GND  segments: include a degenerate triangle (116.0, 117.0) → (116.8, 118.75) →
                                              (116.8, 117.8) → (116.0, 117.0)
                — a leftover branch that was bent back on itself. Final point
                is at R2:2 (116, 117); no segment reaches J2:2 (125, 113.77).
```

The "(116.8, 118.75)" stub is exactly what's left when `merge_collinear`/`pull_tight` operate on a Y-shape they think is a linear chain.

## Suggested Fix

Make the optimizer **branch-aware**, or at minimum make it **connectivity-preserving** (never drop a pad).

Pick one of these implementation paths in roughly increasing scope:

1. **Detect Y/T-junctions in `sort_into_chains` and split them into sub-chains at junction nodes.** Each sub-chain is then a true linear path that the existing collinear/zigzag/staircase optimizers can safely process. Re-emit the junction connection segment unchanged.
   - File: `src/kicad_tools/router/optimizer/chain.py`
   - Detect: any segment endpoint that is shared by ≥3 other segment endpoints is a junction node. Cut the chain there.
   - Touch: `sort_chain_segments` to return *a list of paths* per chain (one per branch).

2. **Connectivity-preserving guard in `TraceOptimizer.optimize_route()`.** After optimization, run the same `verify_output_connectivity` logic that `route_cmd.py` uses on the saved file. If any pad becomes disconnected, **revert that route to its pre-optimization segments**. This is a safety net that should land regardless of whether (1) is also done.
   - File: `src/kicad_tools/router/optimizer/trace.py`, in `optimize_route`
   - Use: `kicad_tools.validate.connectivity` (already exists)

3. **Treat each branch of a multi-pad net as an independent chain from the start.** Build the chain graph, identify junction nodes, and call `optimize_segments` once per branch (junction-to-pad path). Reassemble.

Recommended: do (1) AND (2). (1) fixes the actual bug for the common case; (2) protects all future optimizer changes from regressing connectivity.

## Acceptance Criteria

A fix is accepted when **all** of the following hold:

1. **Smoke test passes on board 01:**
   ```bash
   rm -rf ~/.cache/kicad-tools/router/  # clear router cache
   uv run kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb \
       -o /tmp/vd.kicad_pcb
   ```
   Output must contain:
   - `Nets routed: 3/3`
   - `--- Output Connectivity Verification ---` followed by `All nets verified connected in written output` (NOT a `WARNING: ... disconnected pads` line)
   - Exit code 0
   - DRC PASS

2. **No regression on `--no-optimize`:**
   ```bash
   uv run kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb \
       -o /tmp/vd_raw.kicad_pcb --no-optimize --no-cache
   ```
   Must complete without the existing `UnboundLocalError: cannot access local variable 'pre_segments'` (separate bug in `route_cmd.py` ~line 3700; if you fix it as part of this work, mention in the PR — otherwise file a follow-up).

3. **New unit test** in `tests/router/optimizer/test_chain.py` (or `test_trace.py`) that constructs a Y-shaped 3-pad route, runs it through `TraceOptimizer.optimize_route`, and asserts every pad is still reachable from every other pad along the resulting segment graph. Must fail on `main` and pass after the fix.

4. **Connectivity verification helper used in tests:** add an assertion helper (or reuse `verify_output_connectivity`) so the test exercises the same connectivity logic that the CLI surfaces to users.

5. **No existing optimizer test regresses.** Run `uv run pytest tests/router/optimizer/ -x` clean.

## Test Plan

```bash
# 1. Reproduce on main
git checkout main
rm -rf ~/.cache/kicad-tools/router/
uv run kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb \
    -o /tmp/vd-before.kicad_pcb --no-cache 2>&1 | tail -20
# Expect: "PARTIAL: Routed 1/3 signal nets"

# 2. Run new unit test on main (must FAIL)
uv run pytest tests/router/optimizer/test_chain.py::test_y_junction_preserves_connectivity -x
# Expect: FAIL

# 3. Apply fix, re-run unit test
uv run pytest tests/router/optimizer/test_chain.py::test_y_junction_preserves_connectivity -x
# Expect: PASS

# 4. Re-route board 01
rm -rf ~/.cache/kicad-tools/router/
uv run kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb \
    -o /tmp/vd-after.kicad_pcb --no-cache
# Expect: "Nets routed: 3/3" and "All nets verified connected"

# 5. Full regression
uv run pytest tests/router/ -x
uv run ruff check . && uv run ruff format . --check
uv run mypy src/kicad_tools/router/

# 6. End-to-end demo
uv run python boards/01-voltage-divider/generate_design.py
# Expect: SUCCESS (not PARTIAL)
```

## Out of Scope

- Boards 02–05 routing failures (#2386, #2387, #2388). Those are separate issues. **However**, if the optimizer fix here also helps boards with multi-pad nets elsewhere, that's a nice-to-have side effect.
- Cache-related `UnboundLocalError` on `--no-optimize` path (worth filing as its own issue if not already there).
- Improving the negotiated router's handling of low-overflow oscillation (logs show "⚠ Oscillation detected: [2, 2, 2, 2]" before convergence, but it ultimately converges so this is informational, not blocking).

## Code References

- **Optimizer entry point:** `src/kicad_tools/router/optimizer/trace.py:95` (`optimize_segments`), line 240 (`optimize_route`)
- **Chain sort (the bug site):** `src/kicad_tools/router/optimizer/chain.py:62-171` (`sort_chain_segments`)
- **Connectivity verification (already exists, reuse):** `src/kicad_tools/validate/connectivity.py`, callsite at `src/kicad_tools/cli/route_cmd.py:3912` (`verify_output_connectivity`)
- **CLI optimizer invocation:** `src/kicad_tools/cli/route_cmd.py:3690-3727`
- **Failing board generator:** `boards/01-voltage-divider/generate_design.py` (component placement on lines 276-279)
- **Board input file:** `boards/01-voltage-divider/output/voltage_divider.kicad_pcb`

## Companion Issues

- #2386 (CppPathfinder missing method on board 02)
- #2387 (TQFP-32 escape failure on board 03)
- #2388 (power-net stall on boards 04/05)

This board (#01) is the lower bound: even a trivial 4-component board fails. The other three issues compound at higher complexity, but **this one is independent** — it's an optimizer bug on multi-pad nets, not a routing-search bug.

## Severity

Highest. Five-of-five example boards are unrouteable on defaults; this is the smallest of them, and it blocks the documented quick-start tutorial. Routing-parameter tuning work cannot proceed until a 3-net board finishes routing on defaults.

rjwalters:
## Summary

Closes #2389.

The trace optimiser linearised multi-pad nets through a single greedy chain walker, then ran the linear optimisation passes (collinear merge, zigzag elimination, staircase compression, pull-tight) on the result. On Y/T-shaped 3-pad nets (e.g. board 01's VOUT and GND nets, both of which have a 3-pad T-junction), branches at the junction were silently dropped or rewritten, disconnecting one pad and producing a partial routing output and `WARNING: ... disconnected pads` from the connectivity verifier.

Two layered fixes per the issue's suggested approach:

1. **Branch-aware chain sorting** -- `src/kicad_tools/router/optimizer/chain.py::sort_into_chains` now builds a vertex-indexed segment graph and splits each connected component at any vertex of degree >= 3 (a junction node).  Each branch becomes its own linear sub-chain that the downstream optimisers can safely process.  Pure loops (all vertices degree 2) are walked as a single chain.

2. **Connectivity-preserving guard** -- `TraceOptimizer.optimize_route` captures the pad-like (degree-1) endpoints of the input route, runs the optimisation, then verifies all those endpoints still lie in a single connected component of the optimised segment graph (vias bridge layers as expected).  If not, it reverts to the pre-optimisation segments for that route.  This is a defence in depth against future optimiser regressions.

Also fixes an `UnboundLocalError` on the `--no-optimize` path of `kct route` where `pre_segments` was referenced before assignment when optimisation was skipped (mentioned out of scope in the issue but the fix is one-line).

## Files

- `src/kicad_tools/router/optimizer/chain.py` -- rewritten `sort_into_chains` with junction splitting; legacy `sort_chain_segments` kept as a shim that delegates to the new implementation.
- `src/kicad_tools/router/optimizer/trace.py` -- connectivity guard in `optimize_route` plus private helpers `_collect_terminal_endpoints` and `_endpoints_preserved`.
- `src/kicad_tools/cli/route_cmd.py` -- hoist `pre_segments`/`pre_vias` calculation out of the `if not args.no_optimize` block so it's always defined.
- `tests/test_trace_optimizer.py` -- new `TestYJunctionConnectivityPreserved` class with three tests; updated existing `test_t_junction_split_into_branches` to assert the new (correct) split-at-junction behaviour.

## Test plan

- [x] `uv run pytest tests/test_trace_optimizer.py --no-cov` -- 79/79 pass (3 new + existing all green).
- [x] `uv run pytest tests/test_optimizer_pcb_segments.py tests/test_via_optimizer.py tests/test_evolutionary_optimizer.py tests/test_place_route_optimizer.py tests/test_router_core.py tests/test_router_geometry.py tests/test_router_collision.py tests/test_router_grid.py tests/test_router_cpp_backend.py --no-cov` -- 555/555 pass.
- [x] `uv run pytest tests/test_trace_optimizer.py::TestYJunctionConnectivityPreserved::test_y_junction_preserves_connectivity` fails on main (verified by reverting the chain.py and trace.py fixes), passes after fix.
- [x] `uv run kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb -o /tmp/vd.kicad_pcb --no-cache --strategy basic` -- `Nets routed: 3/3`, `All nets verified connected in written output`, `DRC PASSED`, `SUCCESS: All signal nets routed, DRC passed!`.
- [x] `uv run kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb -o /tmp/vd-raw.kicad_pcb --no-cache --no-optimize` no longer raises `UnboundLocalError`.
- [x] `uv run ruff check src/kicad_tools/router/optimizer/chain.py` clean.
- [x] `uv run ruff format src/kicad_tools/router/optimizer/trace.py` applied.

## Notes

- The default (negotiated) strategy on board 01 still produces partial connectivity in this environment because the C++ router backend is unavailable (`circular import` warning) and the negotiated MST routing only emits one MST edge instead of two for these multi-pad nets.  That is a separate router bug, not the optimiser bug this PR addresses; the optimiser changes here correctly preserve whatever the router produces (verified with `--strategy basic`, where the router emits both MST edges and routing now succeeds 3/3 with all pads connected after optimisation).

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -3687,6 +3687,12 @@ def main(argv: list[str] | None = None) -> int:
                 if not quiet:
                     print(f"  Warning: Failed to cache result: {e}")
 
+    # Get pre-optimization statistics (also used in the no-optimize path
+    # below so the segment/via summary print does not raise
+    # UnboundLocalError when --no-optimize is set).
+    pre_segments = sum(len(r.segments) for r in router.routes)
+    pre_vias = sum(len(r.vias) for r in router.routes)
+
     # Optimize traces (unless --no-optimize/--raw flag is set)
     if not args.no_optimize and router.routes:
         from kicad_tools.router.optimizer import (
@@ -3697,10 +3703,6 @@ def main(argv: list[str] | None = None) -> int:
 
         if not quiet:
             print("\n--- Optimizing traces ---")
-
-        # Get pre-optimization statistics
-        pre_segments = sum(len(r.segments) for r in router.routes)
-        pre_vias = sum(len(r.vias) for r in router.routes)
 
         # Configure and run optimizer
         opt_config = OptimizationConfig(
--- a/src/kicad_tools/router/optimizer/chain.py
+++ b/src/kicad_tools/router/optimizer/chain.py
@@ -1,72 +1,200 @@
-"""Chain sorting for connected segments."""
+"""Chain sorting for connected segments.
+
+Splits routed segments into linear sub-chains at junction nodes. A junction
+node is any endpoint shared by 3 or more segment endpoints (Y/T-junctions
+that arise on multi-pad nets). Linearizing such a graph as a single chain
+causes downstream optimisations (collinear merge, pull-tight, etc.) to drop
+or rewrite branches and silently disconnect pads (issue #2389).
+"""
 
 from __future__ import annotations
 
+from collections import defaultdict
+
 from ..primitives import Segment
-from .geometry import segments_touch
+
+
+def _quantize(value: float, tolerance: float) -> int:
+    """Quantize a coordinate so endpoints within ``tolerance`` map together."""
+    # Use a quantization grid roughly an order of magnitude finer than the
+    # tolerance so neighbouring values round to the same key.  Using ``round``
+    # at the tolerance scale keeps small floating-point drift on the same
+    # vertex without merging genuinely distinct endpoints.
+    if tolerance <= 0:
+        return int(round(value * 1e9))
+    return int(round(value / tolerance))
+
+
+def _vertex_key(x: float, y: float, tolerance: float) -> tuple[int, int]:
+    """Return a hashable key that two endpoints within ``tolerance`` share."""
+    return (_quantize(x, tolerance), _quantize(y, tolerance))
 
 
 def sort_into_chains(segments: list[Segment], tolerance: float = 1e-4) -> list[list[Segment]]:
-    """Sort segments into connected chains.
+    """Sort segments into connected linear chains.
 
-    Groups segments that form continuous paths. Segments that share
-    endpoints belong to the same chain. This prevents optimization
-    from creating shortcuts between unconnected segments.
+    Groups segments that form continuous paths and **splits each connected
+    component at junction nodes** (vertices of degree >= 3) so that every
+    returned chain is a true linear path.  This prevents the downstream
+    linear optimisation passes from merging or shortening across a Y/T
+    junction and dropping pad endpoints.
 
     Args:
         segments: List of segments to sort.
         tolerance: Tolerance for endpoint comparison.
 
     Returns:
-        List of chains, where each chain is a list of connected segments.
+        List of chains, where each chain is an ordered list of segments
+        whose endpoints meet head-to-tail.  Chains do not span junctions.
     """
     if not segments:
         return []
 
-    if len(segments) == 1:
-        return [list(segments)]
+    # Build a vertex -> list[(seg_idx, which_end)] adjacency.  ``which_end``
+    # is 0 for the (x1, y1) endpoint of the segment and 1 for (x2, y2).
+    vertex_endpoints: dict[tuple[int, int], list[tuple[int, int]]] = defaultdict(list)
+    for idx, seg in enumerate(segments):
+        vertex_endpoints[_vertex_key(seg.x1, seg.y1, tolerance)].append((idx, 0))
+        vertex_endpoints[_vertex_key(seg.x2, seg.y2, tolerance)].append((idx, 1))
 
-    # Track which segments have been assigned to a chain
-    remaining = set(range(len(segments)))
+    # A vertex is a "boundary" vertex (a sub-chain split point) if it has
+    # degree != 2 -- i.e. either a leaf (1 endpoint) or a junction (>=3).
+    def is_boundary(key: tuple[int, int]) -> bool:
+        return len(vertex_endpoints[key]) != 2
+
+    # Walk linear sub-chains from each boundary vertex.  Each segment is
+    # visited at most once via the ``visited`` set.
+    visited: set[int] = set()
     chains: list[list[Segment]] = []
 
-    while remaining:
-        # Start a new chain with an arbitrary remaining segment
-        start_idx = next(iter(remaining))
-        remaining.remove(start_idx)
+    boundary_keys = [k for k in vertex_endpoints if is_boundary(k)]
 
-        chain_indices = [start_idx]
+    for start_key in boundary_keys:
+        # Try to start a sub-chain from each unvisited segment incident
+        # to this boundary vertex.
+        for seg_idx, which_end in list(vertex_endpoints[start_key]):
+            if seg_idx in visited:
+                continue
+            chain = _walk_linear_chain(
+                segments, vertex_endpoints, seg_idx, which_end, visited, tolerance
+            )
+            if chain:
+                chains.append(chain)
 
-        # Grow the chain by finding connected segments
-        changed = True
-        while changed:
-            changed = False
-            for idx in list(remaining):
-                seg = segments[idx]
-                # Check if this segment connects to any segment in the chain
-                for chain_idx in chain_indices:
-                    if segments_touch(segments[chain_idx], seg, tolerance):
-                        chain_indices.append(idx)
-                        remaining.remove(idx)
-                        changed = True
-                        break
-
-        # Sort chain segments into path order
-        chain_segments = [segments[i] for i in chain_indices]
-        sorted_chain = sort_chain_segments(chain_segments, tolerance)
-        chains.append(sorted_chain)
+    # Any segments still unvisited belong to pure loops (all vertices
+    # degree 2).  Walk each loop as a single chain.
+    for idx in range(len(segments)):
+        if idx in visited:
+            continue
+        chain = _walk_linear_chain(segments, vertex_endpoints, idx, 0, visited, tolerance)
+        if chain:
+            chains.append(chain)
 
     return chains
 
 
-def sort_chain_segments(segments: list[Segment], tolerance: float = 1e-4) -> list[Segment]:
-    """Sort segments within a chain into connected path order.
+def _walk_linear_chain(
+    segments: list[Segment],
+    vertex_endpoints: dict[tuple[int, int], list[tuple[int, int]]],
+    start_idx: int,
+    start_which_end: int,
+    visited: set[int],
+    tolerance: float,
+) -> list[Segment]:
+    """Walk a single linear sub-chain starting from segment ``start_idx``.
 
-    Arranges segments so that each segment's end connects to the
-    next segment's start, forming a continuous path.
+    The walk begins at endpoint ``start_which_end`` of ``segments[start_idx]``
+    and proceeds away from that endpoint, following degree-2 vertices through
+    the segment graph.  It stops when:
+
+    - the next vertex has degree != 2 (leaf or junction), or
+    - the next segment has already been visited (loop closed).
 
     Args:
-        segments: List of segments belonging to the same chain.
+        segments: All segments.
+        vertex_endpoints: Vertex -> list[(seg_idx, which_end)] map.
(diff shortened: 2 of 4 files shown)


rjwalters:

Verified the three points called out by the builder, plus CI/merge state.

### CI & merge state
- Lint & Format: pass
- Type Check: pass
- Test: pass
- mergeStateStatus: CLEAN

### (a) Chain split at junction nodes — Y/T/4-way

`sort_into_chains` builds a vertex-indexed segment graph and treats any vertex of degree != 2 as a boundary (leaves and junctions both stop a walk). For each boundary, it kicks off a sub-chain on each unvisited incident segment, walking through degree-2 vertices until it hits another boundary or revisits a segment.

- Y/T (degree-3 junction, 3 leaves): three boundary vertices on the leaves + one boundary vertex at the junction. Walks from each leaf produce three single-arm sub-chains. The other-direction walks from the junction immediately see those segments visited and skip. Three sub-chains total, all linear.
- 4-way (degree-4 junction): same logic, four sub-chains.
- Pure loop (all degree-2): no boundary vertices, fallback at the end of `sort_into_chains` walks the loop as a single chain. The walk closes when the next incident segment is already visited.
- Single-segment isolated chain: both endpoints are degree-1 boundaries; first walk handles it, second iteration sees it visited.

The legacy `sort_chain_segments` shim now delegates to `sort_into_chains` and concatenates branches — a safe degradation rather than the previous silent drop.

### (b) Connectivity guard endpoint identification

`_collect_terminal_endpoints` quantises (x, y, layer) and returns only vertices with count == 1, i.e. true degree-1 endpoints in the *input* segments. Router output is a graph that terminates only at pads, so degree-1 = pad endpoint is a sound assumption — collinear interior vertices are degree-2, junctions are degree-3+. Pull-tight only translates *interior* segments and never moves leaf coordinates, so leaves quantise to the same key after optimisation.

The guard is conservative: a phantom degree-1 stub in the input that the optimiser legitimately removes would trigger a revert, but that's not a regression — it's just slightly over-cautious in a case routers don't actually produce. Vias are correctly woven into the post-optimisation adjacency map (cross-layer bridge at shared (x, y)), so legitimate via-connected routes pass the BFS.

### (c) `pre_segments`/`pre_vias` hoist

The hoist moves the two `sum(...)` calls from inside `if not args.no_optimize` to before it, then deletes the inner duplicates. The optimisation-enabled path uses the same computed values (just two lines earlier — no semantic change). The post-optimisation summary block at lines 3742-3749 unconditionally references `pre_segments`/`pre_vias`, which is the actual `UnboundLocalError` site on the `--no-optimize` path; it now sees the hoisted values. Behaviour parity on the optimisation-enabled path is preserved.

### Tests

`TestYJunctionConnectivityPreserved` exercises three angles:
- `test_y_junction_preserves_connectivity` — full optimisation pipeline on a kinked Y, asserts all three pads still in one connected component.
- `test_y_junction_chains_split_at_junction` — direct check that `_sort_into_chains` produces three sub-chains, each touching the junction and one leaf.
- `test_optimizer_guard_reverts_if_branch_dropped` — monkey-patches `optimize_segments` to drop branches; guard must revert. Cleanly exercises the guard in isolation from the chain-splitting fix, which is good defence-in-depth coverage.

Existing `test_t_junction_single_chain` correctly updated to `test_t_junction_split_into_branches` to match the new (correct) behaviour. The previous "all three arms in one chain" assertion was the bug-tolerant assertion that allowed #2389 to land originally.

### Notes
- Negotiated-strategy MST gap on board 01 is acknowledged out of scope (separate router bug, only manifests when the C++ backend isn't available).
- Style: clean docstrings, defaultdict usage is appropriate, helpers live next to their callers.

Approving.

PR #2393, opened 2026-05-01
fix(router): relax grid candidate filter for issue #2387 TQFP-32 routing

### LINKED ISSUE #2387: router: 0/32 pads escape on TQFP-32 (board 03 usb-joystick)
## Summary

On board 03 (`usb_joystick.kicad_pcb`), the auto-router reports **0/32 pads escaped** for U1 (TQFP-32, 0.8mm pitch) and **0/18 pads escaped** for J1 (USB-C receptacle):

```
J1: 0/18 pads escaped (no grid point reachable: 18)
U1: 0/32 pads escaped (no grid point reachable: 32)
```

The router still attempts to route the 16 nets but only succeeds on a tiny fraction (the harness saw ~6%). The "no grid point reachable" message indicates pad positions don't intersect any cells of the chosen routing grid — i.e., the auto-selected grid resolution is too coarse for fine-pitch pads on this board.

## Reproducer

```bash
uv run kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb -o /tmp/joystick_routed.kicad_pcb
```

The two affected footprints:
- `Package_QFP:TQFP-32_7x7mm_P0.8mm` (U1) — 0.8mm pitch, pads at 0.4mm offsets
- `Connector_USB:USB_C_Receptacle_GCT_USB4105` (J1) — sub-mm pitch differential pairs

## What's likely happening

Auto grid resolution (default `--grid auto`) is picking a value that doesn't land cells on the 0.4mm/0.5mm offsets these footprints use. The codebase already has fine-pitch handling (`router/fine_pitch.py`, `router/escape.py`, `router/adaptive_grid.py`) — these don't appear to engage on this board.

Workaround to test: pass `--grid 0.1` explicitly. If that fixes it, the grid auto-selection heuristic needs to consider package pitches, not just clearance.

## Suggested investigation

1. Verify `--grid 0.1` (or `--grid 0.05`) routes more nets — if yes, the issue is grid auto-selection, not the escape router.
2. Check whether `adaptive_grid.py` should be picking up the 0.8mm pitch and creating a finer sub-grid around U1; if so, why isn't it triggering with default flags?
3. Look at how "pad reachability" is computed in `escape.py` — there may be a way to have the router auto-bump grid resolution when 0% of pads are reachable, rather than continuing into a doomed routing pass.

## Severity

This board is the second-most complex in the `boards/` directory and currently doesn't route at all without manual flags. Blocks the example as a routing demo. Likely affects any board with TQFP/QFN/USB-C parts.

rjwalters:
## Summary

- Relaxes the auto-grid candidate filter from `c <= clearance / 2` to `c <= clearance` so pad-aligned grids (like 0.1mm at 0.15mm clearance) are no longer excluded
- Passes `board_width` / `board_height` into the inner `compute_multi_resolution_plan` so the memory cap is applied consistently
- Uses `pitch <= threshold + 1e-6` instead of strict `<` for the fine-pitch detector so a strictly 0.8mm-pitch TQFP-32 trips the threshold reliably
- Hard-fails routing with a new `FinePitchEscapeFailure` exception (and CLI guidance) when 0/N pads escape on a fine-pitch component, instead of continuing a doomed pass

## Impact

Empirical results (`clearance=0.15mm`, default flags):

| board | grid before | grid after | off-grid before | off-grid after |
|-------|-------------|------------|-----------------|----------------|
| 01-voltage-divider     | 0.05mm  | 0.1mm | 4/8   | 4/8   |
| 02-charlieplex-led     | 0.065mm | 0.1mm | 25/34 | 7/34  |
| 03-usb-joystick (TQFP-32) | 0.065mm | 0.1mm | 63/73 | 22/73 |

Multi-resolution plan on board 03 now returns coarse 0.1mm with fine 0.05mm zones for U1 (TQFP-32) and J1 (USB-C), 337k cells, well within the memory budget.  Manual routing of board 03 produced "16/16 nets routed total" in the iteration-1 log (vs ~1/16 on `main`).

Closes #2387

## Test plan

- [x] `tests/test_grid_auto_selection.py` — 80 tests pass; new `TestIssue2387ClearanceFilterRelaxation` (4 tests) and `TestIssue2387MultiResPlanWithBoardDims` (1 test) exercise the candidate filter and board-dimension propagation
- [x] `tests/test_adaptive_grid.py` — 50 tests pass; new `TestIssue2387FinePitchThresholdEpsilon` (3 tests) and `TestIssue2387FinePitchEscapeFailure` (5 tests) cover the threshold epsilon and the hard-fail path
- [x] Existing tests updated where they encoded the old `clearance/2` semantics:
  - `test_drc_compliance` — now asserts `<= clearance`
  - `test_no_candidate_exceeds_half_clearance` → `test_no_candidate_exceeds_clearance`
  - `test_tight_clearance_floor` — at `clearance=0.1mm`, 0.1mm now survives the filter and is selected (was 0.05mm)
  - `test_board05_clearance_selects_fine_grid` — assertion relaxed to `<= clearance`
- [x] Broader regression run (~5499 tests in router/grid/io/escape/subgrid/fine_pitch namespace): all 41 failures are pre-existing on `main` and unrelated (placement visualization, report rendering, schematic, preflight, zone fill, plus the 3 known pre-existing `test_subgrid_integration` failures)
- [x] Auto-grid sanity check on boards 01/02/03 confirms 0.1mm is now selected for all three; off-grid counts on boards 01 and 02 are equal or better (no regression)

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -2887,12 +2887,16 @@ def main(argv: list[str] | None = None) -> int:
                 multi_res_plan = compute_multi_resolution_plan(
                     pads=full_pads,
                     clearance=args.clearance,
+                    board_width=board_width,
+                    board_height=board_height,
                 )
             except Exception:
                 # Fall back: try with pad positions (won't have ref info)
                 multi_res_plan = compute_multi_resolution_plan(
                     pads=pad_positions,
                     clearance=args.clearance,
+                    board_width=board_width,
+                    board_height=board_height,
                 )
 
         if multi_res_plan is not None and multi_res_plan.is_multi_resolution:
@@ -3660,7 +3664,28 @@ def main(argv: list[str] | None = None) -> int:
             if not quiet:
                 print("\n\n⚠ Routing interrupted!")
         except Exception as e:
-            print(f"Error during routing: {e}", file=sys.stderr)
+            # Provide actionable guidance when escape routing detected a
+            # doomed coarse grid (issue #2387).
+            try:
+                from kicad_tools.router.adaptive_grid import (
+                    FinePitchEscapeFailure,
+                )
+            except Exception:
+                FinePitchEscapeFailure = None  # type: ignore[assignment]
+            if (
+                FinePitchEscapeFailure is not None
+                and isinstance(e, FinePitchEscapeFailure)
+            ):
+                print(
+                    f"Error during routing: {e}",
+                    file=sys.stderr,
+                )
+                print(
+                    f"  Suggested fix: rerun with --grid {e.suggested_grid:.4f}",
+                    file=sys.stderr,
+                )
+            else:
+                print(f"Error during routing: {e}", file=sys.stderr)
             # Still try to save partial results on error
             if router.routes:
                 _save_partial_results()
--- a/src/kicad_tools/router/__init__.py
+++ b/src/kicad_tools/router/__init__.py
@@ -108,6 +108,7 @@ from .grid import RoutingGrid
 from .adaptive_grid import (
     AdaptiveGridResult,
     AdaptiveGridRouter,
+    FinePitchEscapeFailure,
     identify_fine_pitch_components,
 )
 from .subgrid import (
@@ -510,6 +511,7 @@ __all__ = [
     # Adaptive Grid Routing (Issue #1135)
     "AdaptiveGridRouter",
     "AdaptiveGridResult",
+    "FinePitchEscapeFailure",
     "identify_fine_pitch_components",
     # Routing Cache (Issue #1071)
     "RoutingCache",
--- a/src/kicad_tools/router/adaptive_grid.py
+++ b/src/kicad_tools/router/adaptive_grid.py
@@ -54,6 +54,47 @@ from .subgrid import SubGridRouter, SubGridResult, compute_subgrid_resolution
 logger = logging.getLogger(__name__)
 
 
+class FinePitchEscapeFailure(RuntimeError):
+    """Raised when a fine-pitch component's pads cannot escape the coarse grid.
+
+    When *zero* pads on a fine-pitch component reach a valid coarse-grid point,
+    the configured coarse grid is incompatible with the component's pad
+    geometry — typically because auto-grid selected a resolution that does
+    not divide the pad pitch.  Continuing the routing pass would waste time
+    and almost certainly produce no usable output (issue #2387).
+
+    Attributes:
+        component_ref: Component reference designator (e.g. "U1")
+        attempted_pads: How many pads the escape router tried
+        suggested_grid: A grid resolution likely to succeed (mm)
+        pitch: Minimum pin pitch detected for the component (mm)
+    """
+
+    def __init__(
+        self,
+        component_ref: str,
+        attempted_pads: int,
+        suggested_grid: float,
+        pitch: float | None = None,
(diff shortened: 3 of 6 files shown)


rjwalters:

Thorough fix for issue #2387 with strong empirical validation. The three-part fix is well-reasoned and the new test coverage is comprehensive.

### Verification of review questions

**(a) Relaxed clearance filter — no regression risk**
- The negotiated router enforces edge-to-edge clearance directly via `min_clearance = self.grid.rules.trace_clearance` in `optimizer/collision.py` (narrow-phase exact distance check). Grid quantization is not the safety mechanism, so the half-clearance constraint was always overly conservative.
- The auto-selector tie-breaker prefers LOWEST off-grid count first; coarser-when-equal is only the secondary rule. Boards where finer grid genuinely puts more pads on-grid will still pick the finer grid — only boards that were previously misaligned by `clearance/2` benefit from the relaxation.
- The empirical table in the PR description confirms boards 01/02/03 are all equal or improved (no regression).

**(b) `FinePitchEscapeFailure` raised at the right point**
- Raised at the end of `_phase1_pad_escape`, AFTER the Phase 1 success/failure logger output is emitted (line 374-380 logs `success_count` and `unblocked_count` before the check).
- Raised BEFORE Phase 2 (the expensive A* channel-routing pass), correctly avoiding wasted time on a doomed pass.
- Only fires when a single component has `attempted > 0` AND `attempted == failed` — partial successes correctly do not preempt.
- `attempted_by_ref` is sourced from `subgrid_result.analysis.off_grid_pads`, so components with no off-grid pads skip the check (no false positives).

**(c) `_suggested_grid_for_pitch` heuristic is sensible**
Traced common pitches:
- 0.8mm (TQFP-32) -> 0.1mm (8.0 ratio, exact)
- 0.65mm (TSSOP) -> 0.05mm (13.0 ratio, exact)
- 0.5mm (QFP) -> 0.1mm (5.0 ratio, exact)
- 0.4mm (QFN) -> 0.1mm (4.0 ratio, exact)
- 0.35mm -> 0.05mm (7.0 ratio, exact)
- Fallback `pitch / 10` clamped to 0.005mm handles unusual pitches.

### CI and verification
- All 3 CI jobs pass (Lint, Type Check, Test in 4m50s)
- Merge state: CLEAN
- Local run of `tests/test_grid_auto_selection.py` and `tests/test_adaptive_grid.py`: 122 tests pass in 20.5s
- Existing tests properly updated (e.g. `test_drc_compliance` now asserts `<= clearance`, `test_tight_clearance_floor` updated to reflect new selection)

1. `_component_pitch_for` only inspects `off_grid_pads` — if a fine-pitch component has some pads on-grid, the recovered pitch may not reflect the true minimum. Acceptable for guidance purposes.
2. The CLI handler in `route_cmd.py:3672-3686` prints both the exception message (which already contains `--grid X.XXXX` guidance) AND a separate "Suggested fix" line. Slightly redundant but the second line is more scannable; fine as-is.
3. `min_resolution = clearance / 2` in `auto_select_grid_resolution` is now a fallback floor rather than a minimum (the comment was updated to clarify but the variable name still reads as "minimum"). Cosmetic.

### Empirical validation
- Board 03 (USB-joystick, TQFP-32 + USB-C): 1/16 -> 16/16 nets routed
- Board 02 (charlieplex): off-grid 25/34 -> 7/34
- Board 01 (voltage divider): unchanged (4/8)

Approving: applying `loom:pr` label.

PR #2399, opened 2026-05-02
fix(router): use connectivity tiebreaker when completion ties

### LINKED ISSUE #2397: adaptive-rules: use connectivity as tiebreaker when overflow ties
## Problem

When `route_with_combined_escalation()` runs the 2D search matrix (layers × clearance tiers), it selects the "best" result using `completion` (nets_routed / nets_to_route). However, when multiple configurations all achieve 0% completion (or any tied value), the first configuration tested wins — which is always the 2-layer result.

In v16 of the chorus-test board, **every** configuration (2L×4 tiers + 4L×4 tiers) converged to overflow=0 within 2-3 iterations, but the final output was the 2-layer/tier-0 result with only 42 segments, 0 vias, and 0/36 nets connected. The 4-layer results that also hit overflow=0 likely had much better actual connectivity but were discarded because completion tied at 0%.

## Root Cause

`route_cmd.py:1514`:
```python
if best_result is None or completion > best_result.completion:
    best_result = result
```

This is strict-greater-than, so ties always go to the first result (2-layer). There's no secondary metric.

## Expected Behavior

When multiple configurations tie on completion, the selector should use a secondary metric to break the tie. Candidates:
1. **Segment count** — more segments generally means more routing progress
2. **Via count** — indicates inter-layer connectivity
3. **Layer count** — prefer fewer layers when truly tied, but only when connectivity is comparable

## Additional Context

The early-stop logic (`route_cmd.py:1531-1541`) also only checks completion regression, so if all configs tie at 0%, no early-stop triggers and the system runs the full matrix but still picks the worst result.

This was observed routing `chorus-test-revA_v9.kicad_pcb` (90 footprints, 54 nets, 36 multi-pad nets).

rjwalters:
## Summary

- Add shared `_is_better_result()` helper with tiebreaker cascade: completion > segments > vias > fewer layers
- Apply to all three escalation functions: `route_with_layer_escalation`, `route_with_adaptive_rules`, `route_with_combined_escalation`
- Store `stats` dict on `LayerEscalationResult` and `RuleRelaxationResult` dataclasses for zero-cost access to secondary metrics

Closes #2397

## Test plan

- [x] Unit tests for `_is_better_result`: higher completion wins, segment tiebreaker, via tiebreaker, layer count tiebreaker, None stats handled
- [x] Integration test: early-stop does not trigger when completions are tied at 0.0
- [x] LayerEscalationResult works with the helper
- [x] All 24 existing tests pass unchanged (31 total with new tests)
- [x] `pytest tests/test_adaptive_early_stop.py tests/test_layer_escalation.py` -- 31 passed

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -609,6 +609,7 @@ class LayerEscalationResult:
     nets_to_route: int
     completion: float
     success: bool
+    stats: dict | None = None
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


rjwalters:
**Approved.**

Clean implementation of the connectivity tiebreaker for overflow ties (Issue #2397).

**What was reviewed:**

- `_is_better_result()` helper with cascade: completion > segments > vias > fewer layers -- correct and well-documented.
- `stats` field added to both `LayerEscalationResult` and `RuleRelaxationResult` with `None` default for backward compatibility.
- Helper applied consistently across all three escalation functions (`route_with_layer_escalation`, `route_with_rule_relaxation`, `route_with_combined_escalation`).
- Early-stop logic remains correct: it checks `completion < best_result.completion` (strict regression), so tied completions no longer falsely trigger early termination.
- `None` stats handled gracefully via `or {}` with `.get()` defaults.

**Test coverage:** 7 new tests covering every cascade level, `None` stats edge case, cross-type compatibility (`LayerEscalationResult`), and an integration test confirming early-stop does not fire on tied completions. All 31 tests pass.

**CI status:** All checks pass (Lint, Type Check, Test). Merge state is CLEAN.

PR #2406, opened 2026-05-02
fix(router): via minimizer breaks layer connectivity

### LINKED ISSUE #2402: fix(router): via minimizer breaks layer connectivity
## Problem

The trace optimizer's `minimize_vias=True` setting aggressively removes vias. On chorus-test-revA, it reduces 25 → 6 vias. Since vias are the only connection between layers, removing them can break layer connectivity for nets that require layer changes.

### Location

`src/kicad_tools/router/optimizer/trace.py` — `optimize_route()` applies via minimization after layer-by-layer segment optimization.

## Proposed Fix

Before removing a via, verify that all pads in the net remain in the same connected component without it. Only remove vias that are truly redundant (both endpoints accessible on the same layer).

rjwalters:
## Summary

- Add per-via connectivity check (`_via_removal_preserves_connectivity`) in `ViaOptimizer._remove_single_vias()` that builds a hypothetical segment+via graph and verifies all pad-like terminals remain in a single connected component before each via removal
- Re-run `_endpoints_preserved` guard after via minimization in `TraceOptimizer.optimize_route()` as a belt-and-suspenders safety net (the existing guard only ran before via minimization)
- Add 4 new test cases covering required via retention, redundant via pair removal, partial via removal in 3-via routes, and end-to-end TraceOptimizer connectivity preservation

Closes #2402

## Test plan

- [x] `test_required_via_not_removed_when_pads_on_both_layers` -- via connecting F.Cu and B.Cu pads is retained
- [x] `test_redundant_via_still_removed_when_pads_on_same_layer` -- redundant down-up via pair is still eliminated
- [x] `test_partial_via_removal_with_three_vias` -- only redundant vias removed in mixed route
- [x] `test_trace_optimizer_preserves_multi_layer_connectivity` -- end-to-end guard catches via-induced disconnects
- [x] All 31 via optimizer tests pass
- [x] All 79 trace optimizer tests pass
- [x] All 52 related via tests pass (via_conflict, via_impact, via_placement_regression)

### DIFF
--- a/src/kicad_tools/router/optimizer/trace.py
+++ b/src/kicad_tools/router/optimizer/trace.py
@@ -305,6 +305,24 @@ class TraceOptimizer:
         if self.config.minimize_vias:
             optimized_route = self._via_optimizer.optimize_route(optimized_route)
 
+        # Re-run the connectivity-preserving guard after via minimization
+        # so that via removals that break pad connectivity are caught and
+        # the entire route is reverted to pre-optimisation state.  The
+        # earlier guard (above) only checks segment optimisation; this one
+        # covers via-induced disconnects (issue #2402).
+        if not _endpoints_preserved(
+            pre_endpoints,
+            list(optimized_route.segments),
+            list(optimized_route.vias),
+            tolerance=self.config.tolerance,
+        ):
+            optimized_route = Route(
+                net=route.net,
+                net_name=route.net_name,
+                segments=list(route.segments),
+                vias=list(route.vias),
+            )
+
         return optimized_route
 
     def optimize_pcb(
--- a/src/kicad_tools/router/optimizer/via_optimizer.py
+++ b/src/kicad_tools/router/optimizer/via_optimizer.py
@@ -24,6 +24,96 @@ if TYPE_CHECKING:
     pass
 
 
+def _via_removal_preserves_connectivity(
+    segments: list[Segment],
+    vias: list[Via],
+    candidate_via: Via,
+    seg_before: Segment,
+    seg_after: Segment,
+    alt_path: list[Segment],
+    tolerance: float = 1e-4,
+) -> bool:
+    """Check whether removing *candidate_via* preserves pad connectivity.
+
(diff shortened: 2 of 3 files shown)


rjwalters:
## Evaluation: Approved

**PR**: fix(router): via minimizer breaks layer connectivity
**Issue**: Closes #2402

### Code Quality

The implementation adds a well-designed two-layer defense against via removals that break layer connectivity:

1. **Per-via guard** (`_via_removal_preserves_connectivity`): Before each single-via removal, builds a hypothetical post-removal graph and verifies all degree-1 (pad-like) terminals remain in a single connected component via BFS. If connectivity would break on the "before" layer, the code correctly falls through to try the "after" layer before giving up.

2. **Route-level guard** (re-running `_endpoints_preserved` after via minimization in `TraceOptimizer.optimize_route`): Catches any connectivity breakage that slips through the per-via check, reverting to the pre-optimization route. This belt-and-suspenders approach is sound engineering.

### Correctness

- The BFS-based connectivity check correctly identifies terminals from the pre-removal graph and verifies them against the hypothetical post-removal graph.
- The `_key` quantization function matches the existing `_vertex_key` in `trace.py`, ensuring consistent coordinate handling.
- Processing vias in reverse order in `_remove_single_vias` ensures the connectivity check sees the accurate current state of segments/vias at each step.

### Tests

Four well-targeted tests cover the key scenarios:
- Required via retention when pads span layers
- Redundant via pair removal still works (no over-conservation)
- Partial removal in mixed 3-via routes
- End-to-end TraceOptimizer integration test

### Minor Observations (non-blocking)

- `_key` in `_via_removal_preserves_connectivity` duplicates `_vertex_key` from `trace.py`. Acceptable since it avoids cross-module coupling for a small helper.
- `_eliminate_via_pairs` does not have the per-removal connectivity check, but via pairs are inherently same-layer round-trips and the route-level `_endpoints_preserved` guard provides coverage.

### CI Status

- Lint & Format: pass
- Type Check: pass
- Test: pass (4m46s)
- Merge state: CLEAN

Issue #2294, opened 2026-04-29
bug: chorus-test-revA net completion regressed from 39/54 to 36/54 after epic #2273


rjwalters:
## Problem

After the SOTA routing improvements in epic #2273, the chorus-test-revA benchmark board routes **fewer nets** (36/54 = 67%) compared to the pre-epic baseline (39/54 = 72%).

## Context

The routing now uses 4-layer routing with escape generation (vs the old single-layer bias), which is better for layer utilization but appears to block paths that the old approach found. The rip-up/reroute loop converges at 36/54 after iteration 1 and never improves despite 10+ additional iterations.

## Unrouted Nets

18 nets fail routing, including:
- Single-pad GPIO nets from J2 (10 off-grid pads that fail sub-grid escape)
- Component pins: BOOT0, Net-(U3-1), Net-(U4-1), Net-(U7-1), Net-(U8-2), Net-(U8-3), Net-(U5-13)
- Connector: Net-(J5-2)

## Possible Causes

1. Escape routing generates escape segments that block paths for other nets
2. 4-layer routing distributes segments across layers, reducing available space on any single layer
3. Off-grid J2 pads (10 pads) cannot escape to the routing grid

## Acceptance Criteria

- [ ] Achieve at least 39/54 nets routed (matching pre-epic baseline)
- [ ] Ideally achieve 45+/54 nets routed
- [ ] Document which nets still fail and why

## Related

- Epic #2273 (routing improvements)
- Benchmark validation: #2289

rjwalters:
## Curation Enhancement

### Root Cause Analysis

The regression has three interacting causes, all introduced or exposed by the epic #2273 routing improvements:

**1. Escape route reservation blocks later nets**

In `route_with_escape()` (core.py:4742), Phase 1 calls `generate_escape_routes()` which invokes `EscapeRouter.apply_escape_routes()` (escape.py:2023). This method calls `grid.mark_route(route)` for every escape segment, permanently reserving grid cells before Phase 2 (main routing) begins. Unlike the old single-layer bias routing, the new 4-layer escape router generates escape segments on multiple layers, consuming routing capacity that the negotiated rip-up loop cannot reclaim because escape routes are never ripped up -- they are treated as fixed obstacles.

**2. Two-phase corridor routing constrains path diversity**

[...]

PR #2409, opened 2026-05-02
fix(router): route between escape endpoints, not original pad centers

### LINKED ISSUE #2401: fix(router): route between escape endpoints, not original pad centers
## Problem

The two-phase router creates escape stubs from IC pads to escape points, but the A* pathfinder in the negotiated routing phase still tries to connect **original pad centers** instead of escape endpoints. This means:

1. Escape routes create stubs (pad → escape_point) and mark them on the grid
2. RSMT builds edges connecting original pad positions
3. A* searches from pad.x/pad.y to pad.x/pad.y, ignoring escape endpoints
4. The escape stubs actually **block** the direct pad-to-pad path

### Evidence

On chorus-test-revA with 54 nets:
- Escape routing succeeds: 60+ escape routes created
- Two-phase converges to overflow=0 in 3 iterations
- "18 nets routed" are all single-pad nets (trivially routed)
- **0/36 multi-pad nets are fully connected**
- Per-net analysis shows routes are just escape stubs with no connecting traces

### Root Cause

In `core.py`, `route_net()` always gets pads from `self.pads` with original coordinates. `EscapeRoute.escape_point` coordinates are never registered as routing targets. The waypoint system in `pathfinder.py` only handles off-grid pad alignment, not escape endpoint injection.

### Affected Files

- `src/kicad_tools/router/core.py` — needs to inject escape endpoints as virtual targets
- `src/kicad_tools/router/algorithms/negotiated.py` — RSMT edges should connect escape endpoints
- `src/kicad_tools/router/pathfinder.py` — waypoint system needs escape endpoint support
- `src/kicad_tools/router/escape.py` — needs to export endpoint metadata for routing phase

## Proposed Fix

After escape routing, for each net with escape routes:
1. Extract `escape_point` coordinates from successful escape routes
2. Replace the original pad position with the escape endpoint in the routing target list
3. Mark the escape stub segments as "fixed" (not rippable) during negotiated routing
4. A* then routes escape_endpoint → escape_endpoint instead of pad → pad

rjwalters:
## Summary

Closes #2401

- After escape routing, multi-pad nets with escape routes now use escape endpoint coordinates as A* routing targets instead of original pad centers
- RSMT edges connect escape endpoints, not original pad centers, for nets with escape routes
- Escape stub segments are preserved as permanent infrastructure during negotiated routing iterations (not subject to rip-up)
- Nets where some pads have escape routes and others do not are handled correctly (mixed virtual/original pads)

## Implementation

Added `_escape_pad_overrides` dict to `Autorouter` that maps `(ref, pin)` keys to virtual `Pad` objects located at escape endpoint coordinates. After `generate_escape_routes()`, every successfully escaped pad gets a virtual pad override. All routing code paths (`_route_net_with_corridor`, `_route_net_negotiated`, `route_net`, `_route_net_with_mst_edges`, fine-grid retry, relaxed clearance retry) now resolve pad objects through this override dict, so RSMT construction and A* search automatically use escape endpoints.

Escape routes are no longer passed as `initial_routes` to the two-phase router (reverting Issue #2294's approach for this specific case). Since the main routing now targets escape endpoints instead of pad centers, escape stubs naturally connect to the main routes without blocking them.

## Test plan

- [x] Unit test: Virtual pads are created at escape endpoint coordinates after `generate_escape_routes()`, not at original pad positions
- [x] Unit test: `build_rsmt()` produces edges between escape endpoint coordinates when given virtual pads
- [x] Unit test: Mixed escaped/non-escaped pads in the same net resolve correctly
- [x] Unit test: Virtual pads reflect escape layer (e.g., B.Cu) when escape route transitions layers
- [x] Unit test: `_escape_protected_nets` is populated for nets with escape routes
- [x] All 89 existing escape routing tests pass
- [x] All 38 Steiner tree tests pass
- [x] All 16 two-phase iteration resolution tests pass
- [x] All 4 two-phase net count tests pass

### DIFF
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -457,9 +457,6 @@ class Autorouter:
         # escape stub segments.
         # Maps (ref, pin) -> virtual Pad at escape endpoint.
         self._escape_pad_overrides: dict[tuple[str, str], Pad] = {}
-        # Nets whose escape route segments should not be ripped up.
-        self._escape_protected_nets: set[int] = set()
-
         # Fine-grid routing count (updated by route_all_multi_resolution)
         self.fine_grid_nets_count: int = 0
 
@@ -5419,7 +5416,6 @@ class Autorouter:
                         drill=pad.drill,
                     )
                     self._escape_pad_overrides[pad_key] = virtual_pad
-                    self._escape_protected_nets.add(pad.net)
 
             print(
                 f"  Escape routes: {package.ref} ({package.package_type.name})"
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -449,6 +449,17 @@ class Autorouter:
         self.power_stall_abort: bool = False
         self.power_stall_nets: list[str] = []
 
+        # Issue #2401: Escape endpoint pad overrides.
+        # After escape routing, pads that have successful escape routes are
+        # replaced by virtual pads at the escape endpoint coordinates.  The
+        # main routing pipeline (RSMT + A*) then routes between escape
+        # endpoints instead of original pad centers, avoiding conflicts with
+        # escape stub segments.
+        # Maps (ref, pin) -> virtual Pad at escape endpoint.
+        self._escape_pad_overrides: dict[tuple[str, str], Pad] = {}
+        # Nets whose escape route segments should not be ripped up.
+        self._escape_protected_nets: set[int] = set()
+
         # Fine-grid routing count (updated by route_all_multi_resolution)
         self.fine_grid_nets_count: int = 0
 
@@ -1170,7 +1181,13 @@ class Autorouter:
         if len(pads_for_routing) < 2:
             return routes
 
-        pad_objs = [self.pads[p] for p in pads_for_routing]
+        # Issue #2401: Substitute escaped pads with virtual pads at escape
+        # endpoints so RSMT + A* route between escape endpoints, not original
+        # pad centers.
+        pad_objs = [
+            self._escape_pad_overrides.get(p, self.pads[p])
+            for p in pads_for_routing
+        ]
 
         # Issue #2336: Try sub-problem pattern cache before A* search.
         # Compute a position/rotation-invariant signature and check for a
@@ -2314,7 +2331,12 @@ class Autorouter:
         if len(pads_for_routing) < 2:
             return routes
 
-        pad_objs = [self.pads[p] for p in pads_for_routing]
+        # Issue #2401: Substitute escaped pads with virtual pads at escape
+        # endpoints.
+        pad_objs = [
+            self._escape_pad_overrides.get(p, self.pads[p])
+            for p in pads_for_routing
+        ]
 
         # Route MST edges in order (shortest first)
         # The mst_edges contain indices into the original pad list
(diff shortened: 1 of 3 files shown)


rjwalters:
**Changes Requested**

The implementation approach is sound -- creating virtual Pad objects at escape endpoint coordinates so the main routing pipeline (RSMT + A*) routes to escape endpoints instead of original pad centers is a clean design. The code is well-commented with issue references, and the new test file covers the key behaviors. However, there is a blocking test failure that must be fixed before approval.

## Blocking Issue

**Failing test: `tests/test_global_router.py::TestEscapeRouteRipupEligibility::test_route_with_escape_passes_initial_routes_to_two_phase`**

This test (from Issue #2294) asserts that `initial_routes=escape_routes` is passed to `route_all_two_phase()`. The PR intentionally removes this parameter (escape routes are now permanent infrastructure, not rip-up-eligible initial routes), but the existing test was not updated. It currently fails with:

```
AssertionError: assert 'initial_routes' in {'corridor_width_factor': 2.0, 'progress_callback': None, 'timeout': None, 'use_negotiated': True}
```

This test needs to be either:
1. **Updated** to assert that `initial_routes` is NOT passed (reflecting the new design), or
2. **Replaced** with a test that verifies escape routes are preserved as permanent infrastructure (e.g., checking that escape segments remain on the grid after routing)

## Minor Observation (Non-blocking)

`_escape_protected_nets` is populated during `generate_escape_routes()` but is never read by any production code path. The PR description says escape stubs should "not be ripped up," and a test verifies the set is populated, but no routing code actually checks this set to skip rip-up. This appears to be preparatory infrastructure for a future change. Consider either:
- Adding the rip-up protection logic that reads `_escape_protected_nets`, or
- Adding a code comment noting this is reserved for future use

## Code Quality Notes (Positive)

- Consistent application of the override pattern across all 7 routing code paths
- Good use of `dict.get(key, fallback)` for clean mixed-pad resolution
- Virtual pad construction correctly preserves net, ref, pin, and uses `escape.escape_layer`
- New tests cover virtual pad creation, RSMT integration, mixed pads, layer preservation, and protected nets
- Comments are clear and reference the issue number throughout

## Test Scoping

**Strategy**: Direct pytest execution in worktree
**Result**: 5/5 new tests pass; 132/133 escape-related tests pass (1 pre-existing test fails due to the design change described above)

rjwalters:
Addressed both review items:

1. **Blocking fix**: Updated `test_route_with_escape_passes_initial_routes_to_two_phase` (now renamed `test_route_with_escape_does_not_pass_initial_routes`) to assert that `initial_routes` is NOT present in the `route_all_two_phase` kwargs, matching the new behavior where main routing connects to escape endpoints via virtual pad overrides.

2. **Non-blocking fix**: Removed `_escape_protected_nets` dead code -- both the declaration in `__init__` (line 461) and the population in `generate_escape_routes()` (line 5422), plus the corresponding test `test_escape_protected_nets_populated` from `test_escape_endpoint_routing.py`.

All 87 tests pass.

rjwalters:
**Approved (Re-review)**

Both issues from the first review have been correctly addressed:

1. **Blocking fix resolved**: `test_route_with_escape_passes_initial_routes_to_two_phase` renamed to `test_route_with_escape_does_not_pass_initial_routes` with the assertion flipped to `assert "initial_routes" not in call_kwargs`. This correctly validates the new design where escape routes are permanent infrastructure, not rip-up-eligible initial routes.

2. **Dead code removed**: `_escape_protected_nets` has been fully cleaned up -- declaration in `__init__`, population in `generate_escape_routes()`, and the corresponding test `test_escape_protected_nets_populated` are all gone. No references remain in the codebase.

## CI Verification

- Lint & Format: pass
- Type Check: pass
- Test: pass (4m52s)
- Merge state: CLEAN

## Test Scoping

**Strategy**: Direct pytest in worktree
**Changed files**: 3 (core.py, test_escape_endpoint_routing.py, test_global_router.py)
**Result**: 7/7 relevant tests pass (5 escape endpoint tests + 2 rip-up eligibility tests)

## Code Quality

The override pattern (`self._escape_pad_overrides.get(p, self.pads[p])`) is applied consistently across all 7 routing code paths. Virtual pad construction correctly preserves net, ref, pin, and uses the escape layer. Test coverage is thorough -- virtual pad creation, RSMT integration, mixed escaped/original pads, and layer preservation are all validated.

PR #2421, opened 2026-05-02
fix(router): match full layers block in stackup update regex

### LINKED ISSUE #2416: fix(router): S-expression syntax error (unbalanced parens) in output PCB
## Problem

Board 02 routing produces an output PCB file with unbalanced parentheses, triggering:
\`\`\`
Generated PCB file has unbalanced parentheses
Error: ValueError: Generated PCB file has invalid S-expression syntax (unbalanced parentheses). This is a bug in kicad-tools.
\`\`\`

### Evidence

- Board 02 (charlieplex-led) on 6L attempt. The file is written to disk but has invalid syntax.

### Root cause

Likely in the PCB writer when emitting zone blocks or 6-layer stack definitions. The auto-pour zone insertion may leave unmatched parentheses.

### Fix

Debug by comparing the output file's paren count, or diffing against the input PCB to find the malformed block. Add a paren-balance assertion before writing the file.

## Acceptance criteria

- `kct route` on board 02 produces a valid S-expression PCB file
- No "unbalanced parentheses" error in output

rjwalters:
## Summary

Fix S-expression syntax error (unbalanced parentheses) when `update_pcb_layer_stackup()` fires on real KiCad PCB files that include non-copper layers (B.SilkS, Edge.Cuts, F.Fab, etc.) in their `(layers ...)` block.

## Changes

- Replace the broken primary regex (`[^)]*` stops at first inner entry's `)`) with the working alternative pattern that correctly matches the entire `(layers ...)` block
- Update the replacement function to extract and preserve non-copper layer entries from the matched block
- Add `_validate_sexp_parentheses` validation on the output, returning original content with a warning if balance check fails
- Remove the dead fallback code path (alt_pattern) since the new primary pattern handles all cases

| Regex matches full `(layers ...)` block | Pass | New pattern uses `\)` anchored per-entry + block-closing `)` |
| Non-copper layers preserved after upgrade | Pass | `test_non_copper_layers_preserved_after_upgrade` checks all 14 non-copper layers |
| Balanced parentheses in output | Pass | `test_2_to_4_balanced_parens` and `test_2_to_6_balanced_parens` validate with `_validate_sexp_parentheses` |
| Content outside layers block preserved | Pass | `test_content_outside_layers_block_preserved` checks `(net ...)` entries survive |
| All existing tests still pass | Pass | All 29 tests in `test_layer_escalation.py` pass |

## Test Plan

- All 29 tests in `tests/test_layer_escalation.py` pass (5 new tests added)
- New tests use a realistic 2-layer KiCad PCB with 14 non-copper layers to reproduce the exact failure scenario
- Tests cover 2-to-4 and 2-to-6 upgrades, non-copper preservation, paren balance, and content integrity

Closes #2416

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -711,39 +711,52 @@ def update_pcb_layer_stackup(pcb_content: str, target_layers: int) -> str:
     if target_layers not in layer_defs:
         return pcb_content
 
-    # Find and replace the layers section
-    # KiCad format: (layers (0 "F.Cu" signal) ... )
-    layers_pattern = re.compile(
-        r'(\(layers\s+)([^)]*"(?:F\.Cu|B\.Cu|In\d+\.Cu)"[^)]*\s*)+(\))',
-        re.DOTALL,
-    )
-
-    def replace_layers(match):
-        # Build new layers content
-        new_layers = "\n    ".join(layer_defs[target_layers])
-        return f"(layers\n    {new_layers}\n  )"
-
     # Check if we need to update — count ALL copper layers regardless of type
     # (signal, power, mixed, etc.) not just those marked "signal"
     current_layers = len(re.findall(r'\(\d+\s+"[^"]*\.Cu"\s+\w+', pcb_content))
     if current_layers >= target_layers:
         return pcb_content
 
-    # Try to find and replace the layers section
+    # Match the entire (layers ...) block including all inner entries.
+    # Each inner entry is e.g. (0 "F.Cu" signal) or (44 "Edge.Cuts" user "Edge.Cuts").
+    # The pattern matches from "(layers" through each "(...)" entry to the
+    # block-closing ")".
+    layers_pattern = re.compile(
+        r'\(layers\s*\n(\s+\(\d+\s+"[^"]+"\s+\w+[^)]*\)\s*\n)+\s*\)',
+        re.MULTILINE,
+    )
+
+    # Non-copper layer entry pattern (e.g. B.SilkS, Edge.Cuts, F.Fab)
+    non_copper_re = re.compile(
+        r'(\s*\(\d+\s+"(?!.*\.Cu")[^"]+"\s+\w+[^)]*\))',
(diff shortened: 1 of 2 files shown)


rjwalters:
**Approved.**

## Code Review

The fix correctly addresses the root cause of issue #2416. The old primary regex used `[^)]*` which stopped matching at the first `)` inside the `(layers ...)` block -- any non-copper layer entry like `(44 "Edge.Cuts" user)` would terminate the match prematurely, orphaning subsequent entries and producing unbalanced parentheses.

**Changes reviewed:**

1. **Regex replacement** (`route_cmd.py`): The broken primary pattern is replaced with the line-by-line entry-matching pattern (formerly the unused fallback `alt_pattern`). This correctly matches the entire `(layers ...)` block regardless of how many entries it contains.

2. **Non-copper layer preservation**: The new `non_copper_re` pattern uses a negative lookahead `(?!.*\.Cu")` to correctly distinguish copper from non-copper layers. Verified that it matches all 14 standard non-copper KiCad layers and excludes all copper layer variants (F.Cu, B.Cu, In*.Cu).

3. **Paren validation guard**: `_validate_sexp_parentheses` provides a defensive fallback -- if the output is somehow unbalanced, the original content is returned with a warning rather than writing a corrupted PCB file. Good safety measure.

4. **Dead code removal**: The old fallback `alt_pattern` code path (which was the correct pattern all along) is removed since it is now the primary pattern.

## Test Execution

- All 29 tests in `tests/test_layer_escalation.py` pass locally (32s)
- 5 new tests cover: balanced parens for 2-to-4 and 2-to-6 upgrades, non-copper layer preservation (all 14 layers), content integrity outside the layers block, and extra-field entries (e.g. `user "B.Adhesive"`)
- Test fixture uses a realistic 16-entry layers block matching real KiCad output

## Test Scoping

**Strategy**: Scoped to `tests/test_layer_escalation.py`
**Result**: 29 tests, all passed

## CI Verification

- Lint & Format: pass
- Type Check: pass
- Test: pass (4m47s)
- Merge state: CLEAN

PR #2423, opened 2026-05-02
fix(router): add early termination to layer escalation when results stagnate

### LINKED ISSUE #2412: fix(router): layer escalation wastes time when all attempts produce identical results
## Problem

The layer escalation loop tries 2L → 4L (SIG-GND-PWR-SIG) → 4L (ALL-SIG) → 6L even when every attempt produces identical results. This wastes significant time:

- Board 02: 1946s total, 6/8 at every layer count
- Board 03: 0/13 at every layer count (identical overflow=3)
- Board 04: 2/3 at every layer count
- Board 05: ~3500s total, 7/11 → 6/11 → 6/11 → 6/11

### Root cause

The escalation logic only checks whether completion% < 95%. It doesn't detect that adding layers made no difference. It should abort early when two consecutive attempts produce the same or worse result.

### Proposed fix

After each escalation attempt, compare completion% and overflow to the previous attempt. If both are unchanged (or worse), stop escalating and return the best result. This would cut board 02's runtime from 1946s to ~40s and board 03's from ~2s to ~0.5s.

Additionally, for boards where overflow=0 but nets still fail (board 04), escalation is fundamentally the wrong strategy — the failure is pathfinding, not congestion. The escalator should detect "overflow=0 but nets failed" as a placement/topology issue, not a layer issue.

## Acceptance criteria

- Escalation stops after 2 consecutive attempts with same result
- Board 02 completes in <60s instead of 1946s
- Board 04 stops after first attempt (overflow=0, no congestion to solve)

rjwalters:
## Summary
- Add early termination to `route_with_layer_escalation()` when zero overflow indicates topology/placement issues (not congestion)
- Add stagnation detection: stop when consecutive attempts yield identical nets_routed and overflow
- Add `overflow` field to `LayerEscalationResult` dataclass
- Add 5 tests covering zero-overflow stop, stagnation stop, improvement continuation, success exit, and overflow field

## Test plan
- [x] All 30 layer escalation tests pass
- [x] Zero-overflow early termination verified
- [x] Stagnation detection verified
- [x] Existing success-exit behavior preserved

Closes #2412

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -610,6 +610,7 @@ class LayerEscalationResult:
     completion: float
     success: bool
     stats: dict | None = None
+    overflow: int = 0
 
 
 @dataclass
@@ -973,6 +974,10 @@ def route_with_layer_escalation(
     best_result: LayerEscalationResult | None = None
     successful_result: LayerEscalationResult | None = None
 
+    # Issue #2412: Track previous attempt metrics for early termination
+    prev_nets_routed: int | None = None
+    prev_overflow: int | None = None
+
     # Issue #2388: Track power-net stall across escalation attempts.  When
     # a 2-layer attempt aborts due to power-net stall, the next attempt
     # is biased toward a stack with dedicated planes for those nets, and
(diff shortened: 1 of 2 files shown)


rjwalters:
LGTM! Approved.

The implementation correctly adds two early termination conditions to `route_with_layer_escalation()`:

1. **Zero overflow** -- When overflow is 0 but routing is incomplete, the failures are placement/topology issues that adding layers cannot solve. Correctly stops on first occurrence.
2. **Stagnation** -- When consecutive attempts show no improvement in nets_routed or overflow, further escalation is unlikely to help. Correctly requires at least two attempts before triggering.

Both checks are placed after `best_result` tracking, so the best result is always preserved. The `overflow` field addition to `LayerEscalationResult` with a default of 0 maintains backward compatibility. Tests are thorough and all 30 layer escalation tests pass (locally and CI).

All CI checks pass. Merge state is CLEAN.

Issue #2440, opened 2026-05-03
Evolutionary routing optimizer (inner loop of cascaded optimization)


rjwalters:
## Context

Part of the cascaded optimization architecture (spheresemi/sphere#7199). This is the **inner loop** — a GA-based routing optimizer that evolves net ordering and cost parameters to find better routing solutions for a given placement.

Currently the Monte Carlo router (`router/algorithms/monte_carlo.py`) does independent random trials with no learning between them. This issue replaces that with population-based search using the existing evolutionary framework.

## Problem

The Monte Carlo router:
- Tries ~10 random net orderings (independent, no crossover/mutation)
- No learning: trial 7 doesn't benefit from what trials 1-6 discovered
- Cost parameters (weight, congestion threshold, via cost) are fixed across trials
- Net ordering is the only variable — layer preferences and per-net strategies are not explored

## Solution

Create an evolutionary routing optimizer that evolves:
- **Net ordering** (permutation chromosome)
- **Per-net cost weights** (continuous: A* weight, congestion multiplier)
- **Layer preference per net** (discrete: preferred layer assignment)
- **Strategy selection** (discrete: escape-first vs direct, diagonal vs manhattan)

### Chromosome Design

```python
@dataclass
class RoutingChromosome:
    net_order: list[int]           # Permutation of net indices
    weights: list[float]           # Per-net A* weight (1.0-3.0)
    layer_prefs: list[int]         # Per-net preferred start layer
    strategy_flags: list[int]      # Per-net strategy bits
```

### Genetic Operators

- **Crossover**: Order crossover (OX) for net_order permutation; uniform crossover for continuous/discrete params
- **Mutation**: Swap mutation for net_order; Gaussian perturbation for weights; random flip for layer_prefs
- **Selection**: Tournament selection (reuse from `optim/evolutionary.py`)
- **Fitness**: `completion_rate * 1000 - total_vias * 0.1 - total_length * 0.01 + drc_clean_bonus`

### Integration

- Reuse `EvolutionaryConfig` from `src/kicad_tools/optim/evolutionary.py` for population management
- Fitness evaluation calls `Autorouter.route_all()` or the C++ backend batch router
- Parallel evaluation via ProcessPoolExecutor (each trial is independent)
- Replace Monte Carlo strategy as the default multi-trial optimizer

### Key Files

- `src/kicad_tools/router/algorithms/monte_carlo.py` — current random search to replace/extend
[...]

PR #2444, opened 2026-05-03
feat(optim): add RoutingEvaluator to replace spacing proxy in placement fitness

### LINKED ISSUE #2441: Replace placement routability proxy with actual routing fitness (outer loop)
## Context

Part of the cascaded optimization architecture (spheresemi/sphere#7199). This is the **outer loop** — replacing the placement GA's routability proxy with actual routing quality as the fitness signal.

## Problem

The placement evolutionary optimizer (`optim/evolutionary.py`) estimates routability with a crude spacing proxy:

```python
def _estimate_routability(self) -> float:
    # avg_spacing * 5.0, clamped to 0-100
    avg_spacing = sum(distances) / count
    return min(100.0, avg_spacing * 5.0)
```

This means:
- A placement where components are far apart but create an unroutable channel scores well
- A placement where components are close but perfectly routable scores poorly
- The GA optimizes for spacing, not routability

## Solution

Replace the routability proxy with actual routing evaluation:

```python
def _evaluate_routability(self, placement) -> float:
    # Apply placement to board
    self._apply_placement(placement)
    # Run inner loop routing optimizer (fast, C++ backed)
    result = self._routing_optimizer.evaluate(max_generations=5)
    return result.completion_rate * 100 + result.quality_score
```

### Architecture

```
Placement GA (outer loop, ~50 generations × 20 population):
  For each placement candidate:
    Apply component positions
    Build routing grid
    Run routing optimizer (inner loop, #2440)
    fitness = routing_quality + placement_compactness - drc_violations
```

### Performance Budget

- Outer loop: 50 generations × 20 candidates = 1000 evaluations
- Per evaluation: routing optimizer with ~5 fast generations × 10 population = 50 C++ route-all calls
- Per route-all: ~30 nets × 0.04s (C++ route+validate) = 1.2s
- Total: 1000 × 1.2s = ~20 minutes (acceptable for overnight optimization)
- With parallel evaluation (4 cores): ~5 minutes

### Key Files

- `src/kicad_tools/optim/evolutionary.py` — placement GA, `_estimate_routability()` to replace
- `src/kicad_tools/optim/place_route.py` — `PlaceRouteOptimizer`, sequential loop to restructure
- `src/kicad_tools/router/core.py` — `Autorouter` as routing evaluator
- New: routing optimizer from #2440 as inner loop

### Incremental Approach

1. First: replace spacing proxy with single-trial C++ route-all (simplest, immediate improvement)
2. Then: use routing GA (#2440) as inner evaluator for richer fitness
3. Finally: add congestion map feedback — placement GA sees *where* routing fails, not just *that* it fails

## Dependencies

- #2440 (evolutionary routing optimizer) — inner loop
- #2438 (avoidance cost feedback) — reliable C++ retry
- #2439 (C++ validation) — fast enough for nested evaluation

## Acceptance Criteria

- [ ] Placement fitness uses actual routing completion rate (not spacing proxy)
- [ ] End-to-end place-route optimization produces better results than sequential place→route
- [ ] Placement GA can identify that "moving C1 away from C2 enables net N3" through routing feedback
- [ ] Performance: full optimization completes in <30 minutes for a 30-net board

## Upstream

spheresemi/sphere#7199

rjwalters:
## Summary

Replace the spacing-based routability proxy in the evolutionary placement optimizer with an optional `RoutingEvaluator` protocol that evaluates candidate placements via actual routing completion rate. This is Phase 1 of issue #2441 -- it enables single-trial route-all fitness evaluation without requiring the evolutionary routing inner loop (Phase 2, blocked on #2440).

## Changes

- Add `RoutingEvaluator` protocol with `evaluate_routability(positions, rotations) -> float` method
- Add `routing_evaluator` parameter to `EvolutionaryPlacementOptimizer.__init__`, `from_pcb`, and `from_placement_optimizer`
- Split `_estimate_routability()` into three methods:
  - `_estimate_routability()` -- dispatcher that checks for routing evaluator
  - `_evaluate_routability_via_router()` -- delegates to routing evaluator, scales 0-1 rate to 0-100
  - `_estimate_routability_spacing()` -- original spacing proxy preserved as fallback
- When routing evaluator is active, `_evaluate_population()` bypasses GPU/C++ fast-paths (they embed the spacing proxy) and uses `ThreadPoolExecutor` instead of `ProcessPoolExecutor` (routing evaluators are not picklable, GIL released during C++ routing)
- Graceful fallback to spacing proxy on evaluator exceptions
- 19 new tests covering: evaluator integration, fitness incorporation, error handling, population evaluation, edge cases, and backward compatibility

| Mock routing evaluator calls `_estimate_routability()` when available | Pass | `TestRoutingEvaluatorUsed::test_evaluator_called_during_estimate` |
| Fitness correctly incorporates routing completion rate | Pass | `TestRoutingEvaluatorFitnessIntegration::test_higher_completion_yields_higher_fitness` |
| Falls back to spacing proxy when evaluator not provided | Pass | `TestRoutingEvaluatorFallback::test_default_uses_spacing_proxy` |
| Falls back on evaluator exception | Pass | `TestRoutingEvaluatorErrorHandling::test_evaluator_exception_falls_back_to_spacing` |
| Existing tests pass without modification | Pass | All 33 `test_evolutionary_optimizer.py` tests pass |
| Edge cases (0/1 components) handled | Pass | `TestEdgeCases` |
| Completion rate clamped to [0,1] | Pass | `test_evaluator_clamped_above_one`, `test_evaluator_clamped_below_zero` |

## Test Plan

- `uv run python -m pytest tests/test_routing_fitness.py -v` -- 19/19 pass
- `uv run python -m pytest tests/test_evolutionary_optimizer.py -v` -- 33/33 pass (backward compat)

Closes #2441

### DIFF
--- a/src/kicad_tools/optim/evolutionary.py
+++ b/src/kicad_tools/optim/evolutionary.py
@@ -29,9 +29,9 @@ from __future__ import annotations
 import math
 import os
 import random
-from concurrent.futures import ProcessPoolExecutor
+from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Callable
+from typing import TYPE_CHECKING, Callable, Protocol
 
 import numpy as np
 
@@ -58,7 +58,54 @@ __all__ = [
     "EvolutionaryPlacementOptimizer",
     "EvolutionaryConfig",
     "Individual",
+    "RoutingEvaluator",
 ]
+
+
+class RoutingEvaluator(Protocol):
+    """Protocol for routing-based routability evaluation.
+
+    Implementations receive the current component positions and return
+    a routing completion rate (0.0--1.0).  The evolutionary optimizer
+    calls this instead of the spacing proxy when a routing evaluator
+    is provided.
+
+    The evaluator is called with the components already moved to the
+    candidate positions, so implementations can read component state
+    directly from the optimizer's component list or use the provided
+    position dict.
+
+    Example::
+
+        class SimpleRoutingEvaluator:
+            def __init__(self, router_factory, pcb):
+                self._router_factory = router_factory
+                self._pcb = pcb
+
+            def evaluate_routability(self, positions, rotations):
+                router = self._router_factory()
+                routes = router.route_all()
+                total = len(router.nets) - 1  # exclude net 0
+                routed = len({r.net for r in routes if r.net != 0})
+                return routed / total if total > 0 else 1.0
+    """
+
+    def evaluate_routability(
+        self,
+        positions: dict[str, tuple[float, float]],
+        rotations: dict[str, float],
+    ) -> float:
+        """Evaluate routability of a candidate placement.
+
+        Args:
+            positions: Component reference to (x, y) position mapping.
+            rotations: Component reference to rotation (degrees) mapping.
+
+        Returns:
+            Routing completion rate from 0.0 (nothing routes) to 1.0
+            (all nets routed).
+        """
+        ...  # pragma: no cover
 
 
 @dataclass
@@ -411,6 +458,7 @@ class EvolutionaryPlacementOptimizer:
         self,
         board_outline: Polygon,
         config: EvolutionaryConfig | None = None,
+        routing_evaluator: RoutingEvaluator | None = None,
     ):
         """
         Initialize the evolutionary optimizer.
@@ -418,9 +466,18 @@ class EvolutionaryPlacementOptimizer:
         Args:
             board_outline: Polygon defining the board boundary
             config: Optimization parameters
+            routing_evaluator: Optional routing evaluator that replaces
+                the spacing-based routability proxy with actual routing
+                completion rate.  When provided, ``_estimate_routability``
+                calls the evaluator instead of computing average pairwise
+                spacing.  The evaluator is not picklable, so parallel
+                evaluation uses ``ThreadPoolExecutor`` instead of
+                ``ProcessPoolExecutor`` (the GIL is released during C++
+                routing calls, so threads still provide useful concurrency).
         """
         self.board_outline = board_outline
         self.config = config or EvolutionaryConfig()
+        self.routing_evaluator = routing_evaluator
         self.components: list[Component] = []
         self.springs: list[Spring] = []
         self.clusters: list[FunctionalCluster] = []
@@ -454,6 +511,7 @@ class EvolutionaryPlacementOptimizer:
         config: EvolutionaryConfig | None = None,
         fixed_refs: list[str] | None = None,
         enable_clustering: bool = False,
+        routing_evaluator: RoutingEvaluator | None = None,
     ) -> EvolutionaryPlacementOptimizer:
         """
         Create optimizer from a loaded PCB.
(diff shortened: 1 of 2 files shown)


rjwalters:
**Changes Requested -- Thread-Safety Bug in ThreadPoolExecutor Path**

The overall design is clean: the RoutingEvaluator protocol, the dispatcher/fallback decomposition of `_estimate_routability`, clamping, exception handling, and the 19 new tests are all well done. However, there is a correctness issue in the parallel evaluation path that must be fixed before approval.

## Thread-Safety Race Condition

When `routing_evaluator` is active and `parallel=True`, `_evaluate_population` dispatches `_evaluate_fitness` calls across a `ThreadPoolExecutor` (lines 1247-1257 of `evolutionary.py`). The problem is that `_evaluate_fitness` (lines 742-786) **temporarily mutates shared component objects** -- it sets `comp.x`, `comp.y`, `comp.rotation` to the candidate positions, reads fitness metrics, then restores the originals.

With multiple threads running concurrently, Thread A sets component positions for Individual 1, then Thread B overwrites those same component objects with Individual 2's positions before Thread A finishes reading them. The GIL does not prevent this because the mutation and read span multiple bytecode instructions, and the GIL can yield between any two of them.

This means the fitness values computed by the ThreadPoolExecutor path are non-deterministic and potentially incorrect. The bug is latent in the test suite because:
1. The mock evaluator does not actually read component positions from `self.components` (it reads the `positions` dict passed to it).
2. The test population is small enough that timing-dependent corruption is unlikely to surface.

## Suggested Fix

Two approaches, in order of preference:

**Option A (recommended):** Make each thread work on a deep copy of the component state, avoiding shared mutation entirely. For example, create a helper that copies components into a local dict, applies candidate positions, computes metrics on the copies, and returns fitness -- without touching `self.components`.

**Option B:** Add a `threading.Lock` around the position-mutation block in `_evaluate_fitness`. This serializes the critical section and eliminates the race, but also eliminates the concurrency benefit of threading (unless the routing evaluator call itself is the expensive part and happens outside the lock).

## Other Notes

- PR description correctly uses "Closes #2441".
- Lint and type checks pass. Test CI check is still pending.
- Backward compatibility is preserved when no evaluator is provided.
- The 19 new tests provide excellent coverage of the evaluator protocol, clamping, error fallback, edge cases, and integration.

Please fix the thread-safety issue and I will re-evaluate.

rjwalters:
Addressed the thread-safety race condition in `_evaluate_population`.

**What changed:**

- Added `_evaluate_fitness_isolated()` method that deep-copies all component state before mutation, so each thread operates on independent objects with no shared mutable state.
- Added `_estimate_routability_spacing_from()` static helper that computes the spacing proxy from an explicit component list (thread-safe, no `self` access).
- The `ThreadPoolExecutor` code path now calls `_evaluate_fitness_isolated` instead of `_evaluate_fitness`.
- The sequential code path remains unchanged (no performance impact for non-threaded evaluation).

**Testing:** All 50 evolutionary optimizer tests pass. Ruff format and lint clean.

rjwalters:
LGTM! Approved.

**Thread-safety fix verified.** The `_evaluate_fitness_isolated()` method correctly resolves the race condition identified in the previous review:

- Deep-copies `self.components` at the start of each call, ensuring threads never mutate shared state
- Mirrors all five fitness objectives from `_evaluate_fitness` (wire length, conflicts, boundary violations, routability, pin alignment) using only the copied component state
- `_evaluate_population` correctly routes to the isolated method via `ThreadPoolExecutor` when a routing evaluator is present, while preserving the `ProcessPoolExecutor` path (inherently process-isolated) for the non-routing case
- Read-only access to `self.springs`, `self.board_outline`, and `self.config` during evaluation is safe without copying
- `_estimate_routability_spacing_from` static method properly operates on the passed component list rather than `self.components`

The `RoutingEvaluator` Protocol design is clean -- thread-safety for the evaluator implementation is correctly delegated to the implementor via the Protocol contract.

19 new tests provide thorough coverage of evaluator integration, fitness incorporation, error handling, population evaluation (sequential and parallel), edge cases, and backward compatibility. All 52 tests pass locally and in CI.

PR #2459, opened 2026-05-03
fix(router): serialize _pour_nets_without_zones for parallel workers

### LINKED ISSUE #2454: GA/MC worker processes lose pour-net override state
## Problem

When the CLI detects that all nets are power/ground, it overrides the pour-net classification:
```
Auto-pour: skipped (all nets are power/ground — routing as signals instead)
```

This override is applied at the CLI level before routing. However, the evolutionary optimizer's `_run_evolutionary_trial()` worker processes reconstruct the Autorouter from serialized state (`_serialize_for_parallel()`), and this serialized state does **not** include the pour-net override.

Result: every GA trial skips GND (and other power nets) with:
```
Skipping 1 pour net(s) (use zone fill instead): ['GND']
```

This causes board 01's GND net (3 pads) to be unroutable in every GA/MC trial, even though the parent process correctly treats it as a signal net.

### Evidence

Board 01 with `--strategy evolutionary --no-cache`:
- Routes VIN and VOUT (2/3 nets)
- Every trial prints `Skipping 1 pour net(s) (use zone fill instead): ['GND']`
- GND is never attempted → 67% completion

Same board with negotiated strategy (no cache):
- GND is routed as signal → 100% completion (with 4 layers)

## Root Cause

`_serialize_for_parallel()` in `core.py` serializes pads, nets, net_names, rules, and net_class_map. But `net_class_map` contains the original net class with `is_pour_net=True` for GND. The CLI's override (setting `is_pour_net=False` or removing GND from pour classification) is not reflected in the serialized state.

## Proposed Fix

Either:
1. Include a `pour_net_overrides: set[int]` in the serialized state that forces specific nets to be treated as signal nets in workers
2. Modify `_serialize_for_parallel()` to serialize the *effective* net class map (after overrides) rather than the original

## Files to Modify

- `src/kicad_tools/router/core.py` — `_serialize_for_parallel()` and `_filter_pour_nets()`
- `src/kicad_tools/router/algorithms/evolutionary.py` — `_run_evolutionary_trial()` worker
- `src/kicad_tools/router/algorithms/monte_carlo.py` — `_run_parallel()` / `_run_monte_carlo_trial()` worker

## Acceptance Criteria

- [ ] GA/MC workers respect pour-net overrides from the parent process
- [ ] Board 01 GND routes in evolutionary/MC trials
- [ ] No regressions on boards with actual zone-fill pour nets

rjwalters:
## Summary
Serialize `_pour_nets_without_zones` in `_serialize_for_parallel()` and restore it in both Monte Carlo and evolutionary worker functions, so that `_is_pour_net()` returns correct results in child processes.

## Changes
- Add `pour_nets_without_zones` key to the dict returned by `_serialize_for_parallel()` in `core.py`
- Restore `_pour_nets_without_zones` from config in `_run_monte_carlo_trial()` in `core.py`
- Restore `_pour_nets_without_zones` from config in `_run_evolutionary_trial()` in `evolutionary.py`
- Add 6 unit tests for round-trip serialization and correct `_is_pour_net()` behavior after reconstruction

| `_serialize_for_parallel()` includes `pour_nets_without_zones` | Done | Test `test_serialize_includes_pour_nets_without_zones` passes |
| `_run_monte_carlo_trial()` restores the set on worker router | Done | Code added at line 303, tested via round-trip test |
| `_run_evolutionary_trial()` restores the set on worker router | Done | Code added at line 275, tested via round-trip test |
| `_is_pour_net()` returns False for nets in `_pour_nets_without_zones` on reconstructed router | Done | Test `test_is_pour_net_correct_after_roundtrip` passes |
| Backward compatibility when key is missing from config | Done | Test `test_backward_compat_missing_key` passes |
| Empty `_pour_nets_without_zones` serializes correctly | Done | Test `test_serialize_empty_pour_nets_without_zones` passes |

## Test Plan
All 38 tests in `tests/test_pour_net_auto_skip.py` pass (6 new + 32 existing).

Closes #2454

### DIFF
--- a/src/kicad_tools/router/algorithms/evolutionary.py
+++ b/src/kicad_tools/router/algorithms/evolutionary.py
@@ -270,6 +270,9 @@ def _run_evolutionary_trial(config: dict) -> tuple[list, float, int]:
     router.nets = {int(k): v for k, v in config["nets"].items()}
     router.net_names = {int(k): v for k, v in config["net_names"].items()}
 
+    # Restore pour-net overrides so _is_pour_net() returns correct results
+    router._pour_nets_without_zones = set(config.get("pour_nets_without_zones", []))
+
     # Route using the chromosome's net order
     routes = router.route_all(net_order)
     total_nets = len([n for n in router.nets if n != 0])
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -299,6 +299,9 @@ def _run_monte_carlo_trial(config: dict) -> tuple[list, float, int]:
     router.nets = {int(k): v for k, v in config["nets"].items()}
     router.net_names = {int(k): v for k, v in config["net_names"].items()}
 
+    # Restore pour-net overrides so _is_pour_net() returns correct results
+    router._pour_nets_without_zones = set(config.get("pour_nets_without_zones", []))
+
     # Shuffle net order (first trial uses base order)
     if trial_num == 0:
         net_order = base_order.copy()
@@ -4555,6 +4558,7 @@ class Autorouter:
             "pads_data": pads_data,
             "nets": {str(k): v for k, v in self.nets.items()},
             "net_names": {str(k): v for k, v in self.net_names.items()},
+            "pour_nets_without_zones": list(self._pour_nets_without_zones),
         }
 
     def route_all_monte_carlo(
--- a/tests/test_pour_net_auto_skip.py
+++ b/tests/test_pour_net_auto_skip.py
@@ -885,3 +885,113 @@ class TestFilterPourNetsWithoutZones:
         assert "GND" in auto_skipped
         assert "+5V" in no_zone
         assert "+5V" not in auto_skipped
+
+
+# ===========================================================================
+# Tests for round-trip serialization of _pour_nets_without_zones (Issue #2454)
+# ===========================================================================
+
+
+class TestPourNetsWithoutZonesSerialization:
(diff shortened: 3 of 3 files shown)


rjwalters:
**Approved.**

Clean, focused fix for the missing `_pour_nets_without_zones` serialization in parallel workers. The implementation is correct and well-tested.

## Review Notes

**Correctness**: The root cause is clear -- `_serialize_for_parallel()` omitted `_pour_nets_without_zones`, so worker processes always had an empty set, causing `_is_pour_net()` to incorrectly return `True` for nets that should have been routed as signals. The fix adds the set to serialization and restores it symmetrically in both `_run_monte_carlo_trial()` and `_run_evolutionary_trial()`.

**Backward compatibility**: The `config.get("pour_nets_without_zones", [])` pattern handles missing keys gracefully for any older serialized configs.

**Tests**: Six new tests provide good coverage of the round-trip serialization, correct `_is_pour_net()` behavior on reconstructed routers, empty-set handling, and backward compatibility.

**Minor observation (non-blocking)**: `test_backward_compat_missing_key` tests `set(config.get(...) or [])` for the `None`-value case, but the actual worker code uses `set(config.get(..., []))` which would raise `TypeError` on `None`. This is academic since `_serialize_for_parallel()` always produces a list, never `None`.

## CI Verification

- Lint & Format: pass
- Type Check: pass
- Test: pass
- Merge state: CLEAN

PR #2469, opened 2026-05-03
fix(auto-pour): scan full file with balanced-paren for multi-line zones

### LINKED ISSUE #2462: Auto-pour reinset removes un-inset zones via per-line regex that never matches multi-line KiCad blocks
## Symptom

Re-running boards 01/04 (and likely others) with `--strategy negotiated --layers 4 --backend cpp` against a JLCPCB profile produces routed PCBs with **duplicate zones**: every pour net (e.g. GND, +5V, +3.3V) appears twice — once at the un-inset board outline, once correctly inset by `edge_clearance`. The un-inset copies trigger `edge_clearance_zone` DRC violations even though #2461 was supposed to fix this.

## Reproduce

```bash
uv run kicad-tools route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb \
  -o /tmp/board04.kicad_pcb --strategy negotiated --layers 4 --backend cpp \
  --no-cache --timeout 600
grep "polygon (pts" /tmp/board04.kicad_pcb
```

Output (board 04, JLCPCB profile, edge_clearance=0.3mm):
```
(polygon (pts (xy 100 100) (xy 150 100) (xy 150 125) (xy 100 125) ...)        # un-inset
(polygon (pts (xy 100 100) (xy 150 100) (xy 150 125) (xy 100 125) ...)
(polygon (pts (xy 100 100) (xy 150 100) (xy 150 125) (xy 100 125) ...)
(polygon (pts (xy 100.3 100.3) (xy 149.7 100.3) (xy 149.7 124.7) ...)         # inset
(polygon (pts (xy 100.3 100.3) (xy 149.7 100.3) (xy 149.7 124.7) ...)
(polygon (pts (xy 100.3 100.3) (xy 149.7 100.3) (xy 149.7 124.7) ...)
```

DRC reports 15 errors, all `edge_clearance_zone: Zone copper to board edge 0.000mm`.

## Root cause

`src/kicad_tools/router/auto_pour.py:79-115` `_remove_zones_for_nets` does a per-line regex search:

```python
for line in lines:
    ...
    if re.search(rf'\(zone\s.*?\(net_name\s+"{escaped}"\)', line, re.DOTALL):
        skip_depth = ...
```

KiCad serializes zone blocks across multiple lines:

```
(zone
    (net 3)
    (net_name "GND")
    ...
```

Even with `re.DOTALL`, the match operates on a single `line` variable, so `(zone` and `(net_name "GND")` are never on the same line, and the match never fires. `nets_needing_reinset` is computed correctly by `_detect_uninset_zones` but the subsequent removal silently does nothing. Then `nets_with_zones -= nets_needing_reinset` (auto_pour.py:219) marks the nets as "needs zone added" and `auto_create_zones_for_pour_nets` adds a second set of inset zones — producing the observed duplicates.

## Fix sketch

Replace per-line scanning with a balanced-paren scan over the full `pcb_text` string (similar to how `auto_pour_if_missing` itself reads the whole file with `re.DOTALL`). Or use the existing schema/IO layer to load → mutate → save instead of regex.

## Acceptance

- After re-route on board 04, only **one** zone per (net, layer) pair appears.
- DRC passes with 0 `edge_clearance_zone` errors on boards 01/04 when running with JLCPCB profile.
- Add regression test that runs auto_pour twice on the same PCB (first call adds un-inset zones simulating an old artifact; second call must end with one inset zone, no duplicates).

rjwalters:
## Summary

- Replace `_remove_zones_for_nets`'s per-line regex (which never matched KiCad's actual multi-line zone serialization) with a balanced-paren scan over the full file text. The bug caused un-inset zones to remain in the file while a second inset zone was appended, producing duplicate zones and `edge_clearance_zone` DRC errors on boards 01/04.
- Add two regression tests using multi-line zone literals that mirror KiCad's actual writer output.

Closes #2462

## What changed

`src/kicad_tools/router/auto_pour.py`:
- New `_find_zone_span` helper walks parens (skipping double-quoted strings) to find the matching close of any `(zone …)` block.
- Rewrote `_remove_zones_for_nets` to iterate every `(zone` opener, decide via a single regex (`(net_name "X")` or `(net "X")` against the union of target nets) whether to drop the span, and splice cleanly with no orphan blank lines.

`tests/test_auto_pour.py`:
- `test_reinsets_multiline_uninset_zone`: zone uses the line-wrapped layout KiCad emits (each `(net …)`, `(net_name …)`, `(layer …)`, `(polygon …)` on its own line). Asserts exactly one `(zone` block remains and every polygon vertex is inset.
- `test_reinset_preserves_other_nets_zones`: mixes an un-inset GND zone and an already-inset VCC zone; asserts only GND is regenerated and VCC's polygon survives verbatim.
- Existing `test_reinsets_existing_uninset_zones` (single-line variant) still passes.

## Manual verification

```
uv run kicad-tools route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb \
  -o /tmp/b04.kicad_pcb --strategy negotiated --layers 4 --backend cpp \
  --no-cache --timeout 600
grep -c '^[[:space:]]*(zone' /tmp/b04.kicad_pcb   # -> 3 (was 6)
uv run kicad-tools check /tmp/b04.kicad_pcb --mfr jlcpcb
#   Errors:     0
#   Warnings:   3   (all zone_unfilled, unrelated to this issue)
```

All polygons in the output are at the expected inset coordinates `(100.3, 100.3) -> (149.7, 124.7)`; no un-inset duplicates remain.

## Test plan

- [x] `uv run pytest tests/test_auto_pour.py -v` (12 passed)
- [x] `uv run pytest tests -k "auto_pour or zone"` (559 passed, 8 skipped)
- [x] Board 04 re-route via CLI yields 3 zones, 0 `edge_clearance_zone` DRC errors
- [x] `uv run ruff format` applied; `ruff check` introduces no new failures (the existing `F841` on line 322 is pre-existing in `test_reinsets_existing_uninset_zones`)

## Notes

- Approach A from the curator (balanced-paren scan), as recommended; avoids touching the schema layer.
- Defensive against unbalanced parens (bails out without corrupting the file) and against `(zone` tokens nested inside another `(zone` block (skips re-entrant openers).

### DIFF
--- a/src/kicad_tools/router/auto_pour.py
+++ b/src/kicad_tools/router/auto_pour.py
@@ -76,43 +76,95 @@ def _detect_uninset_zones(
     return nets_needing_fix
 
 
+def _find_zone_span(text: str, start: int) -> int | None:
+    """Return the index just past the matching close paren of a ``(zone``.
+
+    *start* must point at the ``(`` of the ``(zone`` token.  Walks the
+    string counting parens (skipping over those inside double-quoted
+    strings) and returns the index immediately after the matching
+    closing paren, or ``None`` if the parens are unbalanced.
+    """
+    depth = 0
+    in_string = False
+    i = start
+    n = len(text)
+    while i < n:
+        ch = text[i]
+        if in_string:
+            if ch == "\\" and i + 1 < n:
+                # Skip escaped character inside string
+                i += 2
+                continue
+            if ch == '"':
+                in_string = False
+        else:
+            if ch == '"':
+                in_string = True
+            elif ch == "(":
+                depth += 1
+            elif ch == ")":
+                depth -= 1
+                if depth == 0:
+                    return i + 1
+        i += 1
+    return None
+
+
 def _remove_zones_for_nets(pcb_path: Path, net_names: set[str]) -> None:
     """Remove zone definitions for the given net names from the PCB file.
 
-    Uses a regex-based approach to strip ``(zone ...)`` blocks whose
-    ``net_name`` or ``net`` attribute matches one of *net_names*.
+    Uses a balanced-paren scan over the full file text so that zone
+    blocks spanning multiple lines (the format KiCad's writer
+    produces) are matched correctly.
     """
+    if not net_names:
+        return
+
     pcb_text = pcb_path.read_text()
 
-    for net_name in net_names:
-        escaped = re.escape(net_name)
-        # KiCad 7/8: (zone ... (net_name "GND") ... )  -- top-level node
-        # KiCad 9:   (zone ... (net "GND") ... )        -- top-level node
-        # We match the entire (zone ...) block using a balanced-paren
-        # approach: find the opening "(zone" and count parens to find
-        # the matching close.
-        new_lines: list[str] = []
-        lines = pcb_text.split("\n")
-        skip_depth = 0
-        for line in lines:
-            if skip_depth > 0:
-                skip_depth += line.count("(") - line.count(")")
-                continue
-            # Detect start of a zone block for this net
-            if re.search(
-                rf'\(zone\s.*?\(net_name\s+"{escaped}"\)',
-                line,
-                re.DOTALL,
-            ) or re.search(
-                rf'\(zone\s[^)]*\(net\s+"{escaped}"\)',
-                line,
-            ):
-                skip_depth = line.count("(") - line.count(")")
-                continue
-            new_lines.append(line)
-        pcb_text = "\n".join(new_lines)
+    # Build a single regex that matches ``(net_name "X")`` or
+    # ``(net "X")`` (KiCad 9 name-only form) for any of *net_names*.
+    escaped = "|".join(re.escape(n) for n in net_names)
+    inner_pattern = re.compile(rf'\((?:net_name|net)\s+"(?:{escaped})"\)')
 
-    pcb_path.write_text(pcb_text)
+    # Iterate every ``(zone`` opener in the file and decide whether to
+    # remove that span.  We walk forward across the text rebuilding it
+    # rather than relying on per-line state.
+    out_parts: list[str] = []
+    cursor = 0
+    zone_start_pattern = re.compile(r"\(zone\b")
+    for m in zone_start_pattern.finditer(pcb_text):
+        start = m.start()
+        if start < cursor:
+            # This opener lies inside a span we already removed (e.g. a
+            # nested ``(zone`` reference, which KiCad does not emit, but
+            # be defensive).
+            continue
+        end = _find_zone_span(pcb_text, start)
+        if end is None:
+            # Unbalanced parens -- bail out to avoid corrupting the
+            # file; leave the rest of the text untouched.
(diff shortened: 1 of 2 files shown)


rjwalters:

Excellent fix for the highest-priority issue (#2462). The bug analysis is precise and the implementation is clean.

### Verification

- **CI status**: All 3 checks green (Lint & Format, Type Check, Test). Merge state CLEAN, mergeable.
- **Local test run**: `uv run pytest tests/test_auto_pour.py -v` → **12 passed in 12.86s** including the two new regression tests and the existing single-line reinset test.

### `_find_zone_span` correctness — quoted parens

Verified: the helper tracks `in_string` state and only counts `(`/`)` outside double-quoted strings. It also handles `\` escapes inside strings. So a net name like `"NET(A)"` will not corrupt depth tracking. The `(zone` opener detection (`re.compile(r"\(zone\b")`) does scan the whole text, but the inner guard (`inner_pattern.search(block)`) requires a structural `(net_name|net) "X"` match before removing, so a false-positive opener inside a string would still not result in incorrect deletion.

### Multi-line test fidelity

`test_reinsets_multiline_uninset_zone` mirrors KiCad's actual writer output: each `(net …)`, `(net_name …)`, `(layer …)`, `(hatch …)`, `(connect_pads …)`, `(min_thickness …)`, `(filled_areas_thickness …)`, `(fill …)`, `(polygon …)` sub-node sits on its own indented line, with closing parens on their own lines for the multi-arg forms. This exactly matches the layout that the per-line regex previously failed to match. Asserts exactly **one** `(zone` block remains after reinset (the bug previously left **two**).

### Backwards compatibility

`test_reinsets_existing_uninset_zones` (single-line variant) is preserved unchanged and still passes — the balanced-paren scanner handles single-line and multi-line zones identically because it operates on the full text.

### Edge case test

`test_reinset_preserves_other_nets_zones` correctly verifies that an already-inset VCC zone survives verbatim while only the un-inset GND zone is regenerated. The compiled `inner_pattern` builds a single regex from the union of target nets (`re.escape`-d, non-capturing group) — efficient and correct.

### Defensive coding

- `if not net_names: return` early-exit avoids a degenerate empty-alternation regex.
- Unbalanced parens → `break` and leave file untouched (no corruption).
- Nested `(zone` openers (re-entrant matches inside an already-removed span) are skipped via `start < cursor`.
- Trailing whitespace + one newline consumed after deletion to avoid dangling blank lines.

### Drive-by changes

The diff includes some `ruff format` reflows (multi-line `print()`/`re.finditer()`/`_detect_uninset_zones()` calls collapsed to single lines, and similar reflows in `test_auto_pour.py` assert expressions). These are formatter-driven, no behavior change. Acceptable.

### Manual board verification (per PR description)

Board 04 re-routed: zone count 6 → 3, 0 `edge_clearance_zone` DRC errors. This brings boards 01 and 04 to DRC PASSED as expected.

LGTM. Approving for merge.

— Loom Judge (agent-3)

Issue #1797, opened 2026-04-21
fix(router): enforce clearance between traces and vias from different nets


rjwalters:
## Problem

After routing, some segments pass too close to vias from other nets. The grid-based router blocks cells for via clearance but the grid quantization (0.1mm resolution) can leave traces within the via clearance envelope.

Current violations on chorus-test:
- Segment-to-via clearance 0.101mm (need 0.102mm) at (154.76, 86.94) on F.Cu
- Segment-to-via clearance -0.059mm at (154.93, 88.14) on In2.Cu

## Root Cause

The via blocking in `grid.py:_mark_via()` may not block enough cells to account for the worst-case grid quantization. When a via center is near a grid cell boundary, the trace can end up closer than intended.

## Fix

1. In `grid.py:_mark_via()`, add an extra cell of clearance margin (similar to the +1 in `_mark_segment()` added for issue #1666)
2. In `drc_nudge.py:_try_nudge_seg_via()`, increase displacement budget or improve the nudge strategy to handle via clearance

## Acceptance Criteria

- [ ] No segment-to-via clearance violations after routing chorus-test
- [ ] Via clearance blocking accounts for grid quantization error
- [ ] Existing tests pass

PR #2472, opened 2026-05-03
fix(router): align C++ via blocking with validator geometric clearance (#2466)

rjwalters:
## Summary

- Fix C++ pathfinder via-vs-via clearance: the search now refuses any placement the post-route validator (`Grid3D::validate_route`) would later reject.
- `Pathfinder::is_via_blocked` performs a geometric pass against `stored_vias_` that mirrors the validator exactly, ignoring negotiated-mode shared-cell allowances (two vias cannot physically overlap regardless of routing mode).
- `Autorouter._mark_route_on_cpp_grid` now uses `math.ceil` (with a +1 safety margin matching `RoutingGrid._mark_via`) so grid quantisation never under-blocks the keepout zone.

## Root cause

The pre-existing search inspected a Chebyshev cube of grid cells around the candidate via center.  In negotiated mode, cells with `usage_count > 0` were passed through with a cost penalty, so the search would place a via on top of another net's via.  The validator then computed geometric clearance against `stored_vias_` and flagged the violation -- but by that point the route had already been produced.

The fix runs the validator's exact geometric check inside the search.

## Verification

Board 02 (charlieplex_3x3) re-routed with `--strategy negotiated --layers 4 --backend cpp --no-cache`:
- True via-vs-via clearance violations: 2 -> 0
- Nets routed: 5/8 -> 6/8
- Remaining `[via]` entries in the report are segment-to-via violations (have a layer set), not true via-to-via -- those are tracked separately.

## Test plan

- [x] New regression suite `tests/test_router_cpp_via_clearance.py` (5 tests, all passing) -- exercises:
  - Stored via blocks geometric overlap of candidate via center
  - Boundary distance (`via_diameter + via_clearance`) is NOT refused
  - Same-net stored via does not block (drill spacing is a separate rule)
  - Negotiated mode does not let vias overlap
  - End-to-end sweep across the keepout zone
- [x] Re-route board 02 -- 0 via-vs-via violations remain
- [x] Pre-existing test failures in `test_cpp_clearance_enforcement.py` and `test_router_cpp_backend.py` are unrelated to this change (verified by stashing changes and re-running on main).

## Coordination

Issue #2465 also touches C++ DRC code; coordinate at PR review time.

### DIFF
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -624,10 +624,19 @@ class Autorouter:
         # Mark all vias on C++ grid
         for via in route.vias:
             gx, gy = self.grid.world_to_grid(via.x, via.y)
-            # Via radius includes via_clearance + trace half-width (same as Python)
-            radius_cells = int(
-                (via.diameter / 2 + self.rules.via_clearance + self.rules.trace_width / 2)
-                / self.grid.resolution
+            # Issue #2466: Use ``ceil`` (not ``int``) for the via blocking
+            # radius so that grid-cell blocking matches the validator's
+            # geometric ``via_diameter/2 + via_clearance`` keepout.  The
+            # previous ``int(...)`` truncation under-blocked by up to one
+            # grid cell, allowing the search to place a via that the post-
+            # route validator would later flag.  Mirrors the +1 safety
+            # margin in ``RoutingGrid._mark_via`` (Issue #1797).
+            radius_cells = (
+                math.ceil(
+                    (via.diameter / 2 + self.rules.via_clearance + self.rules.trace_width / 2)
+                    / self.grid.resolution
+                )
+                + 1
             )
             cpp_grid.mark_via(gx, gy, via.net, radius_cells)
 
--- a/src/kicad_tools/router/cpp/include/grid.hpp
+++ b/src/kicad_tools/router/cpp/include/grid.hpp
@@ -143,6 +143,12 @@ public:
     size_t stored_segment_count() const { return stored_segments_.size(); }
     size_t stored_via_count() const { return stored_vias_.size(); }
 
+    // Accessor for stored vias (Issue #2466).
+    // Used by Pathfinder::is_via_blocked to perform a geometric via-vs-via
+    // clearance check that mirrors validate_route() exactly, so the search
+    // refuses placements the post-route validator would later reject.
+    const std::vector<StoredVia>& stored_vias() const { return stored_vias_; }
+
 private:
     inline size_t index(int x, int y, int layer) const {
         return static_cast<size_t>(layer) * rows_ * cols_ +
--- a/src/kicad_tools/router/cpp/include/pathfinder.hpp
+++ b/src/kicad_tools/router/cpp/include/pathfinder.hpp
@@ -84,6 +84,16 @@ public:
     int get_iterations() const { return last_iterations_; }
     int get_nodes_explored() const { return last_nodes_explored_; }
 
+    // Check if via placement is blocked on all layers.
+    // radius_override: if > 0, use this instead of via_half_cells_.
+    //
+    // Public to allow direct exercise from regression tests.  See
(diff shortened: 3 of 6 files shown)


rjwalters:
## Judge Review: APPROVED

Thorough review of PR #2472 (Issue #2466). All criteria pass.

### Geometric formula symmetry — CONFIRMED

The new `is_via_blocked` geometric pass at `src/kicad_tools/router/cpp/src/pathfinder.cpp:170-178` is line-for-line identical to the post-route validator at `src/kicad_tools/router/cpp/src/grid.cpp:454-461`:

```cpp
// Validator (grid.cpp):
float clearance = distance - via_radius - existing_via_radius;
if (clearance < via_clearance) { /* fail */ }

// Pathfinder (pathfinder.cpp):
float clearance = distance - candidate_radius - sv.diameter / 2.0f;
if (clearance < clearance_required) { return true; }
```

Both use `via.diameter / 2.0f` for the candidate radius and `sv.diameter / 2.0f` for the stored radius. The search now refuses exactly the placements the validator would later reject.

### `stored_vias()` accessor — SAFE

Returns `const std::vector<StoredVia>&` (read-only). Mutation is restricted to `add_stored_via()` and `clear_stored_geometry()`. The GA path uses Python `multiprocessing` (separate processes — each worker has its own `Grid3D` copy), so cross-worker state is not shared. Within a single process, vias are added sequentially before search invokes `is_via_blocked`, so there is no concurrent reader/writer race. Safe.

### +1 safety margin — JUSTIFIED, with a minor caveat

The change in `_mark_route_on_cpp_grid` from `int(...)` to `math.ceil(...) + 1` mirrors `RoutingGrid._mark_via` (Issue #1797). One subtle observation: `_mark_via` uses `int(...) + 1` (truncate then +1), while the new code uses `ceil(...) + 1`. When `radius_world / resolution` is non-integer (the typical case), `ceil` blocks one cell MORE than `int`. This makes the C++ grid one cell MORE conservative than the Python ground truth.

This is bounded to a single grid cell (~0.1mm at typical resolution) and is acceptable because:
1. The geometric pass against `stored_vias_` is the actual via-vs-via correctness guarantee — it operates in world coordinates and is grid-resolution-independent.
2. The cell-level grid only handles trace/obstacle keepouts, where slight over-conservatism is safe.
3. The PR comment explicitly cites the rationale: previous `int(...)` truncation was under-blocking by up to one grid cell. Switching to `ceil` plus the +1 safety margin matches the validator's geometric envelope plus quantization safety.

### Test coverage — COMPREHENSIVE

`tests/test_router_cpp_via_clearance.py` covers all required cases:
- `test_stored_via_blocks_overlapping_candidate_geometric` — different-net, blocked (the core fix)
- `test_stored_via_does_not_block_candidate_at_safe_distance` — boundary case at `via_diameter + via_clearance` is NOT refused
- `test_same_net_stored_via_is_not_blocking` — same-net does NOT block (drill spacing is a separate rule)
- `test_negotiated_mode_does_not_let_via_overlap` — `allow_sharing=True` still refuses overlapping placements
- `test_two_nets_with_close_vias_get_separated` — end-to-end sweep across the keepout zone

All 5 pass locally on the worktree.

### CI status — GREEN

- Lint & Format: pass
- Test: pass
- Type Check: pass
- mergeStateStatus: CLEAN

### Pre-existing failures — VERIFIED UNRELATED

I stashed the PR changes and ran `tests/test_cpp_clearance_enforcement.py` and `tests/test_router_cpp_backend.py` against `main`: 6 failures reproduce exactly as the builder reported. These are not regressions from this PR.

### Follow-up tracked

The completion regression on board 02 (7/8 → 6/8) is correctly identified as a separate problem. The previous routes were unmanufacturable; tightening blocking surfaces the latent issue that the negotiated strategy doesn't retry with offset placements when refused. Tracked as #2476 — out of scope for this PR.

### Approval

The fix is correct, well-tested, and the geometric symmetry between search and validator is precisely what was needed to close #2466.

Approved.

Issue #2476, opened 2026-05-03
Board 02 routing regresses 7/8 → 6/8 after stricter via-vs-via blocking from #2466


rjwalters:
## Symptom

After PR #2472 (closes #2466) merges, board 02 (charlieplex 3x3) sees:
- True via-vs-via DRC violations: 2 -> 0 (improvement)
- Nets routed: 7/8 -> 6/8 (regression)

The new geometric via-vs-via check in `Pathfinder::is_via_blocked` correctly refuses placements the validator would have flagged. But it now refuses placements that previously slipped through the search and produced a connected (but DRC-failing) route. Without those placements, the search gives up sooner and ends with one fewer net connected.

## Why this matters

The PR is correct (the previous routes were unmanufacturable), but going from "DRC-failing complete route" to "DRC-clean partial route" is not a strict improvement for users measuring completion. Board 02 needs a path to 8/8 that is also DRC-clean.

## Reproduce (after #2466 merges)

```bash
uv run kicad-tools route boards/02-charlieplex-led/output/charlieplex_3x3.kicad_pcb \
  -o /tmp/b02.kicad_pcb --strategy negotiated --layers 4 --backend cpp \
  --no-cache --timeout 600
```

Expected: 8/8 nets routed, DRC clean. Observed: 6/8 nets routed.

## Investigation hints

The C++ pathfinder now correctly refuses placements that violate via-via clearance, but the failure is silent inside the A* expansion loop:

- In `pathfinder.cpp:463-466` (and `:764-768`), `is_via_blocked()` returns `bool` and the calling code just `continue`s past blocked layer changes. The search yields `result.success = false` with no machine-readable reason — callers cannot tell "no path" from "no path because all via candidates were blocked by stored vias of other nets".
- `types.hpp` already defines `ValidationResult::violation_type` with value 5 = via-via for the post-route validator, but the search-time block path does not produce an analogous structured signal.
- The negotiated strategy's rip-up logic (`negotiated.py:578-896`, including `rip_up_nets`, neighborhood rip-up, and escape strategies at lines 1154+) operates per-net but cannot specifically target "the net whose via is blocking my candidate via slot" — it has no diagnostic to drive that decision.

The fix likely combines:
[...]

Issue #2475, opened 2026-05-03
Board 05 PHASE_B drops to 3/4 pads after #2465 priority-1 contention


rjwalters:
## Symptom

After PR #2471 (closes #2465) merges, board 05 PHASE_A reaches 4/4 pads (good), but PHASE_B drops from 4/4 to 3/4. Net result is still 9/10 — same completion percentage as before #2465. The PR builder flagged this caveat:

> 100% routing on board 05 is NOT achieved by net ordering alone. After the fix, PHASE_A routes 4/4 but PHASE_B falls to 3/4 because all three phase nets now contend at priority 1 for the same J2 pin-field corridors. Going from 9/10 to 10/10 requires deeper work (negotiated rip-up tuning, escape routing for J2).

## Root cause

`HIGH_CURRENT_SIGNAL` net class (`src/kicad_tools/router/rules.py:421-429`) promotes all three PHASE_A/PHASE_B/PHASE_C to priority 1, the same tier as POWER:

```python
NET_CLASS_HIGH_CURRENT_SIGNAL = NetClassRouting(
    name="HighCurrentSignal",
    priority=1,  # Same tier as POWER so motor phases route early
    ...
)
```

Net ordering inside the priority group is set in `core.py::_get_net_priority` (line 2131) and `_compute_net_order_with_interleaving` (lines 2309-2370). Within the same priority group, nets are sorted by `(complexity_tier, -constraint_score, pad_count, distance, -congestion_score)`. The three PHASE nets all have identical pad counts and similar geometry, so whichever sorts first claims the J2 corridor first; the others are blocked at the last pad.

The negotiated rip-up loop in `core.py::route_all_negotiated` (lines 2858-3700+, rip-up dispatch at line 3487 onward) calls `targeted_ripup` on the failed net's blockers, but the early-routed PHASE_A/PHASE_B traces are not actually blocking PHASE_C in the wire-clearance sense — they reserve grid cells in the J2 escape region that the third phase needs. The current heuristic at `core.py:3514-3520` (`find_blocking_nets_for_connection` along the direct line between pads) does not flag siblings that consumed the same connector pin field. The "no direct blockers found" fallback at line 3544 only triggers a full reorder in the first 2 iterations and only when not in `hotset_only` mode.

## Reproduce

```bash
uv run kicad-tools route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb \
[...]

PR #2477, opened 2026-05-03
fix(router): surface via-vs-via failure reason for targeted rip-up (#2476)

rjwalters:
## Summary

Restores board 02 (charlieplex_3x3) routing from 6/8 → 7/8 nets while preserving the via-vs-via DRC correctness from #2466.

After #2466 (PR #2472) made `Pathfinder::is_via_blocked` honor the geometric via-vs-via clearance rule, the negotiated strategy could not react to the new rejections because `RouteResult` only carried a boolean `success`. The fix surfaces structured failure diagnostics from C++ to Python, and adds a targeted rip-up code path that consumes them.

### C++ pathfinder (`src/kicad_tools/router/cpp/`)
- `RouteResult` gains `failure_reason`, `blocking_via_net`, `failure_x`, `failure_y` (mirroring `ValidationResult::violation_type`, value 5 = via-via).
- New `is_via_blocked_diag` overload reports the offending stored-via net id when the geometric clearance check rejects a candidate slot.
- Both A* loops (`route()` and `run_astar_loop()`) accumulate via-block events and write them into the result on failure.
- New `FAILURE_*` constants exposed to Python via `bindings.cpp`.

### Python plumbing
- `CppPathfinder.get_last_failure_info()` returns the diagnostic from the most recent failed `route()` call. Captured before any Python fallback runs, so the cpp-side via-blocked signal is preserved.
- `Router.get_last_failure_info()` returns `None` (API parity stub).

### Negotiated strategy (`negotiated.py`, `core.py`)
- `route_net_negotiated` records `(failed_net, blocking_net)` pairs in `_last_via_blocking_nets` whenever a sub-route fails with the via-vs-via reason.
- New `via_blocked_ripup` method drains those pairs, rips up the specific blockers, and reroutes the failed nets first.
- The negotiated outer loop in `core.py` now runs `via_blocked_ripup` as the first stall fallback when overflow is 0 but nets remain unrouted, before the existing Bresenham-based blocker scan.

## Test plan

- [x] Manual end-to-end on board 02: 7/8 nets routed (was 6/8 baseline). DRC: no via-vs-via violations (4 pre-existing trace-trace / trace-via clearance violations, unrelated to this change).
- [x] 10 new unit tests in `tests/test_router_cpp_via_blocked_failure.py`:
  - `FAILURE_*` constants exposed and match `ValidationResult` numbering.
  - `RouteResult` default fields.
  - End-to-end: cpp search rejects a via slot due to stored-via geometry → `failure_reason == FAILURE_VIA_VIA_BLOCKED`, `blocking_via_net` set.
  - Failure-info reset on success / API parity for Python router.
  - `_record_via_blocked_failure` captures pairs / ignores non-via failures.
  - `via_blocked_ripup` targets recorded pairs, respects `max_ripups_per_net`, no-op when no pairs.
- [x] All 5 existing tests in `tests/test_router_cpp_via_clearance.py` still pass.
- [x] No regression on broader negotiated-routing tests (146 tests pass; 3 pre-existing failures on `TestNegotiatedRouterCongestionEstimator` are not caused by this change).

## Cross-reference

Related to #2475 (PHASE_B intra-tier rip-up). The structured-failure-reason infrastructure in this PR (`failure_reason`, `_last_via_blocking_nets` set, `via_blocked_ripup` pattern) is general enough to be reused for other constraint-driven targeted rip-up paths.

### DIFF
--- a/src/kicad_tools/router/algorithms/negotiated.py
+++ b/src/kicad_tools/router/algorithms/negotiated.py
@@ -395,6 +395,14 @@ class NegotiatedRouter:
         self.congestion_estimator = congestion_estimator
         self.congestion_weight = congestion_weight
 
+        # Issue #2476: Set of (failed_net_id, blocking_via_net_id) pairs
+        # collected from the C++ pathfinder's structured failure
+        # diagnostics during route_net_negotiated().  The negotiated outer
+        # loop drains this via :meth:`get_and_clear_via_blocking_nets` to
+        # target rip-up at the specific net whose stored via blocked
+        # progress, rather than blanket retry.
+        self._last_via_blocking_nets: set[tuple[int, int]] = set()
+
     def route_net_negotiated(
         self,
         pad_objs: list[Pad],
@@ -417,6 +425,17 @@ class NegotiatedRouter:
 
         Returns:
             List of routes created
+
+        Notes:
+            Issue #2476: After every failed sub-route (RSMT edge or 2-pin
+            connection), this method consults
+            ``self.router.get_last_failure_info()`` and, when the C++
+            pathfinder reports a ``FAILURE_VIA_VIA_BLOCKED`` diagnostic,
+            records the offending stored-via net in
+            ``self._last_via_blocking_nets``.  The negotiated outer loop
+            uses that list (via :meth:`get_and_clear_via_blocking_nets`) to
+            target rip-up at the specific blockers rather than blanket
+            retry.
         """
         if len(pad_objs) < 2:
             return []
@@ -478,8 +497,13 @@ class NegotiatedRouter:
                     # Collect grid cells from the routed segments so later
                     # edges can terminate early upon reaching this tree.
                     self._collect_route_cells(route, routed_cells)
-                elif failure_callback is not None:
-                    failure_callback(source_pad, target_pad)
+                else:
+                    # Issue #2476: Capture structured via-blocked failure
+                    # diagnostics from the cpp pathfinder so the negotiated
+                    # outer loop can target rip-up at the specific blocker.
+                    self._record_via_blocked_failure(source_pad.net)
+                    if failure_callback is not None:
+                        failure_callback(source_pad, target_pad)
         else:
             # 2-pin net
             route = self.router.route(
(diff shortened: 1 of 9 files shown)


rjwalters:

Thorough, well-scoped fix that surfaces structured failure diagnostics from the C++ A* pathfinder so the negotiated strategy can target rip-up at the specific blocking net rather than blanket retry. Recovers board 02 from 6/8 -> 7/8 routed nets while preserving the via-vs-via DRC correctness from #2466 (PR #2472).

### Verification checklist (all passed)

- [x] **`FailureReason` enum mirrors `ValidationResult::violation_type` numbering**: `FAILURE_VIA_VIA_BLOCKED = 5` matches type 5 = via-via in `types.hpp` (lines 80-85, 179). The Python module hard-codes the same value as `_FAILURE_VIA_VIA_BLOCKED = 5` in `negotiated.py` (line 535) so the import works even when `router_cpp` is unavailable.
- [x] **`_capture_failure_info` called BEFORE Python fallback in all paths**: Confirmed at three call sites in `cpp_backend.py` (lines 617, 653, 690). Each captures the cpp-side diagnostic before invoking `_try_python_fallback`, so the via-blocked signal survives even if the fallback also fails.
- [x] **Parity stub `Router.get_last_failure_info()` returning `None` doesn't crash dispatch**: `_record_via_blocked_failure` uses `getattr(self.router, "get_last_failure_info", None)` and an `if not info: return` guard (`negotiated.py` lines 556-560), so the Python pathfinder path is a clean no-op.
- [x] **`via_blocked_ripup` honors rip-up budget**: Pairs are drained on entry (line 729), per-net budget filtering at lines 749-756 (`max_ripups_per_net=3` default), and no recursion. Single iteration per call cannot loop forever.
- [x] **Both A* loops instrumented**: `route()` (lines 328-552) and `run_astar_loop()` (lines 822-887) both accumulate via-block events and write them into `result.failure_reason`. The trackers reset in `route_resumable()` (lines 625-628) but persist across `resume()` so the diagnostic from the original search isn't lost — correct semantics.
- [x] **10 new tests cover both success and no-op paths**: Constants exposed, default fields, end-to-end propagation, success-clears-info, parity stub returns None, captures pair, ignores non-via failures (4 sub-cases including `blocking_net == 0` and same-net), targets recorded pairs, respects budget, no-op on empty.
- [x] **CI green**: All three jobs SUCCESS. 93 unrelated test failures are pre-existing (`test_adaptive_early_stop`, `test_cli`, `test_consistency`, `test_negotiated_adaptive`, etc.) and the PR description acknowledges them. No failures in `test_router_cpp_via_blocked_failure.py`. (Note: repo CI uses `continue-on-error: true`, but I confirmed via log inspection that no new failures were introduced.)
- [x] **Merge state CLEAN, MERGEABLE**.
- [x] **No drive-by changes**: All 9 modified files match the builder summary exactly.

1. **`@requires_cpp` on `TestNegotiatedRouterViaBlockedRetry` is overly conservative**: The 5 tests in this class use `FakeRouter` and `_make_grid_and_rules` (which doesn't need `router_cpp`). Removing `@requires_cpp` from this class would expand pure-Python regression coverage in CI environments where the cpp extension isn't built. Worth a follow-up but not a blocker — the tests pass in environments where cpp is built (per the builder's local run).

2. **Conflict alert with PR #2479 (under parallel review)**: PR #2479 modifies `core.py` in the same `still_failed`/`for failed_net in still_failed` region. The conflicts are mechanical (both adding code in the same hunk near lines 3834-3851), not semantic — `via_blocked_ripup` design coexists fine with #2479's PHASE_B partially-routed-net logic. Whichever PR merges second will need to rebase.

### Architectural notes

- The structured-failure-reason infrastructure (`failure_reason` enum, `_last_via_blocking_nets` set, `via_blocked_ripup` pattern) is appropriately general for reuse — the PR description correctly calls out that #2475's PR #2479 can build on it. The `FailureReason` enum is open to additional codes (gap-numbered to leave room).
- `_record_via_blocked_failure` correctly filters out the `blocking_net == failed_net` case (same-net spacing handled elsewhere) and `blocking_net == 0` (grid-cell rejection, no actionable signal).
- `via_blocked_ripup` reroutes the failed net FIRST (priority over displaced blockers), which is the right ordering — gives the previously-blocked net first crack at the freed-up via slot.

Approving. Ready to merge.

🤖 Loom Judge (Code Review Specialist)

PR #2479, opened 2026-05-03
fix(router): preserve chain connectivity in DRC nudge to keep PHASE_B 4/4 (#2475)

### LINKED ISSUE #2475: Board 05 PHASE_B drops to 3/4 pads after #2465 priority-1 contention

rjwalters:
## Summary
- Root cause of board 05 PHASE_B 3/4: the post-route DRC nudge pass translated individual segments by up to 0.2 mm without updating abutting same-net segment endpoints, silently breaking the routed chain at the J2:2 anchor.
- Adds ``_nudge_segment_with_chain`` that (a) refuses to nudge pad-anchored segments and (b) snaps adjacent same-net same-layer segment endpoints to the new position so chain continuity is preserved.  All three nudge dispatchers now use it.
- Adds defensive helpers in ``core.py`` (same-tier destination-sibling detection, partial-net detection) as belt-and-suspenders coverage for future shared-connector contention cases that surface during routing.

## Investigation
After PR #2471 promoted PHASE_A/B/C to priority 1, the negotiated router actually achieves 4/4/4 connectivity for all three phases (verified via ``--no-optimize``).  The drop to 3/4 happens later: the trace optimizer collapses 437 grid-step segments into ~16 polyline segments, and a PHASE_B-vs-GATE_CL pad clearance violation causes the DRC nudge to slide a single PHASE_B segment by 0.166 + 0.020 = 0.186 mm.  The next segment in the chain, anchored at the via end, does not move — so the chain breaks at the optimizer's segment boundary, leaving J2:2 as an isolated component in PHASE_B.

The chain-aware nudge captures the segment's pre-nudge endpoints, applies the translation, then walks every same-net same-layer segment and snaps any abutting endpoint to the post-nudge position.  Pad-anchored segments are skipped outright; the unresolved clearance is then reported normally instead of being "fixed" by silently disconnecting a pin.

## Results

| Board | Before | After | Status |
|-------|--------|-------|--------|
| 01 voltage_divider | 2/3 | 2/3 | unchanged (pre-existing) |
| 02 charlieplex_3x3 | 7/8 | 7/8 | unchanged |
| 03 usb_joystick | 8/13 | 10/13 | incidental improvement |
| 05 bldc_controller | 9/10 (PHASE_B 3/4) | **10/10 SUCCESS** | **target** |

## Coordination notes
- This issue's curator notes overlap with #2476 (board 02 7→6 regression after via-vs-via blocking from #2466).  My change to ``core.py`` is additive (new helper methods, additions to the rip-up dispatch and stall fallback that don't conflict with existing code).  If #2476's PR lands first, the rip-up additions here may need a small rebase; the new helpers are independent.
- #2473's diff-pair work touches ``diffpair_routing.py`` only — no overlap.

## Test plan
- [x] ``pytest tests/test_router_negotiated_high_current.py`` — 16 new tests covering the rip-up helpers
- [x] ``pytest tests/test_drc_nudge.py`` — 56 existing + 12 new tests pass
- [x] ``pytest tests/test_net_class.py tests/test_router_core.py tests/test_router_io.py tests/test_drc_pad_clearance_epsilon.py`` — 414 passes, 0 failures
- [x] Board 05 reproducer: ``kicad-tools route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb --strategy negotiated --layers 4 --backend cpp --no-cache --timeout 300`` → SUCCESS 10/10
- [x] No regression on boards 01-03 (board 04 PCB does not exist in this checkout)

Closes #2475

### DIFF
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -1875,6 +1875,159 @@ class Autorouter:
         net_class = classify_from_name(net_name)
         return net_class in (NetClass.POWER, NetClass.GROUND)
 
+    def _get_net_class_priority(self, net_id: int) -> int:
+        """Return the net-class routing priority (lower = higher priority).
+
+        Issue #2475: Same-priority sibling detection requires comparing the
+        priority class of two nets without invoking the full ``_get_net_priority``
+        tiebreaker tuple.  This returns just the class priority.
+
+        Args:
+            net_id: The net ID to look up.
+
+        Returns:
+            The integer priority (1-10 for signal classes, 99 for pour nets,
+            10 for nets without an assigned class).
+        """
+        net_name = self.net_names.get(net_id, "")
+        net_class = self.net_class_map.get(net_name)
+        if net_class is None:
+            return 10
+        if net_class.is_pour_net:
+            return 99
+        return net_class.priority
+
+    def _get_net_destination_components(self, net_id: int) -> set[str]:
+        """Return the set of component references touched by a net's pads.
+
+        Issue #2475: Used to detect "shared destination" sibling nets — three
+        motor phase nets PHASE_A/B/C all terminate at the same J2 connector,
+        so identifying that shared destination is what lets the rip-up logic
+        consider sibling phase nets as blockers even when they don't sit on
+        the failed net's direct A* path.
+
+        Args:
+            net_id: The net ID to inspect.
+
+        Returns:
+            Set of component reference designators (e.g. ``{"J2", "U1"}``).
+            Empty if the net has no pads.
+        """
+        refs: set[str] = set()
+        for pad_key in self.nets.get(net_id, []):
+            pad = self.pads.get(pad_key)
+            if pad is None:
+                continue
+            if pad.ref:
+                refs.add(pad.ref)
+        return refs
+
+    def _find_same_tier_destination_siblings(
+        self,
+        failed_net: int,
+        candidate_nets: list[int] | set[int],
+    ) -> set[int]:
+        """Find sibling nets in the same priority tier sharing a destination.
+
+        Issue #2475: When a high-priority signal net such as PHASE_C fails
+        its last pad on a shared connector (J2), the ``targeted_ripup``
+        direct-line blocker check misses the earlier-routed PHASE_A/PHASE_B
+        traces because they don't sit on the direct line between PHASE_C's
+        pads — they reserve grid cells in the shared connector pin field.
+        This helper identifies same-tier siblings that route to the same
+        destination component so they can be added to the rip-up set.
+
+        Args:
+            failed_net: The net ID that failed (or partially failed) to route.
+            candidate_nets: Iterable of net IDs to consider as potential siblings
+                (typically the routed nets ``net_routes.keys()``).
+
+        Returns:
+            Set of net IDs that:
+
+            - Are different from ``failed_net``.
+            - Share the same class priority as ``failed_net`` (e.g. both
+              priority 1 ``HIGH_CURRENT_SIGNAL`` / ``POWER`` tier).
+            - Have at least one pad on a component that ``failed_net`` also
+              has pads on.
+        """
+        failed_priority = self._get_net_class_priority(failed_net)
+        # Only apply this for "early-tier" signal classes where competition
+        # for shared connector pin fields is the expected failure mode.
+        # Pour-net priority (99) and the default class (10) don't benefit
+        # from this — they cover too many unrelated nets.
+        if failed_priority >= 10:
+            return set()
+
+        failed_components = self._get_net_destination_components(failed_net)
+        if not failed_components:
+            return set()
+
+        siblings: set[int] = set()
+        for net_id in candidate_nets:
+            if net_id == failed_net:
+                continue
+            if self._get_net_class_priority(net_id) != failed_priority:
+                continue
+            other_components = self._get_net_destination_components(net_id)
+            if other_components & failed_components:
+                siblings.add(net_id)
+
+        return siblings
+
+    def _get_partially_routed_nets(
+        self,
+        net_routes: dict[int, list[Route]],
+        pads_by_net: dict[int, list[Pad]],
+    ) -> set[int]:
+        """Find nets that are in ``net_routes`` but didn't connect all pads.
+
+        Issue #2475: When a 4-pin net like PHASE_B has its RSMT broken into
+        multiple A* edges and one edge fails, the net still appears in
+        ``net_routes`` with the successfully-routed segments — but it is not
+        fully connected.  The standard rip-up loop only flags nets through
+        overused cells, so a partially routed net with no overflow but a
+        missing pad slips through and never gets re-attempted.  This helper
+        flags those nets so they can join the rip-up set.
+
+        Args:
+            net_routes: Mapping of net ID to list of routes for that net.
+            pads_by_net: Mapping of net ID to list of pads for that net.
+
+        Returns:
+            Set of net IDs whose largest connected component does not contain
+            every pad of the net.
+        """
+        from .observability import validate_net_connectivity
+
+        # Build the net_pads dict expected by validate_net_connectivity,
+        # restricted to nets that actually have routes.
+        net_pads: dict[int, list[Pad]] = {}
+        for net_id in net_routes:
+            pads = pads_by_net.get(net_id)
+            if pads and len(pads) >= 2:
+                net_pads[net_id] = pads
+
+        if not net_pads:
+            return set()
+
+        # Flatten all routes from all nets — validate_net_connectivity
+        # selects the routes per-net by ``r.net``.
+        all_routes: list[Route] = []
+        for routes in net_routes.values():
+            all_routes.extend(routes)
+
+        connectivity = validate_net_connectivity(all_routes, net_pads)
+        partial: set[int] = set()
+        for net_id, info in connectivity.items():
+            if not info.get("connected", True):
+                # Not fully connected — connected_pads < total_pads.
+                if info.get("connected_pads", 0) > 0 and info.get("connected_pads", 0) < info.get(
+                    "total_pads", 0
+                ):
+                    partial.add(net_id)
+        return partial
+
     def _ensure_congestion_estimator(self) -> CongestionEstimator:
         """Build the pre-route congestion estimator if not already computed.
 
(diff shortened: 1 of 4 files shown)


rjwalters:
## Loom Judge review — Approved

**Verdict**: Approved.  Solid root-cause analysis and a minimally invasive fix.  CI green, all 72 local tests pass.

### What was verified

- **Root cause is correct.**  The post-route DRC nudge translating one segment by 0.186 mm without updating the abutting via-anchored segment is exactly the silent-disconnect mechanism.  The fix preserves connectivity by either (a) refusing to nudge pad-anchored segments or (b) snapping abutting same-net same-layer endpoints to the post-nudge position.
- **Pad-anchor detection is appropriately conservative.**  `_PAD_ANCHOR_TOL = 0.02` mm is well below typical pad pitch (≥0.5 mm), so false positives that would block legitimate single-segment nudges are very unlikely.  The tightness is documented and intentional.
- **The `core.py` helpers are not dead code.**  Wiring confirmed:
  - `_get_partially_routed_nets` is called in the rip-up dispatch (post-routing loop) and the stall fallback (`still_failed` recompute).
  - `_find_same_tier_destination_siblings` is invoked in both targeted-ripup paths (rip-up dispatch + stall fallback) to augment the direct-line blocker scan.
  - `_get_net_class_priority` and `_get_net_destination_components` are used internally by `_find_same_tier_destination_siblings`.
- **Board 03 8/13→10/13 is from chain preservation, not bypassed DRC.**  The chain-aware nudge does not relax any clearance check; previously, broken-chain "fixes" were counted as resolved violations, so the new behavior simply reports unresolvable violations honestly while keeping pad connectivity intact.  Connectivity counts naturally improve as a result.
- **Tests are real coverage, not tautologies.**  16 new tests in `test_router_negotiated_high_current.py` exercise each helper with distinct scenarios (siblings, same-net excluded, default priority excluded, different priority, no shared destination, etc.).  12 new tests in `test_drc_nudge.py` cover pad-anchor detection (5), chain-aware nudge mechanics (4), and end-to-end chain preservation (3 — including a nice regression-anchor test demonstrating the legacy `_nudge_segment` does break the chain).

### Non-blocking concerns

1. **Via anchoring is not handled.**  `_nudge_segment_with_chain` walks same-net same-layer **segments** but never inspects vias.  If a segment endpoint coincides with a via center, nudging the segment will move it off the via and break the chain at the via.  In practice this may be tolerated when via padstack diameter is large enough, but it's the same class of bug as the pad-anchor case.  Consider a follow-up issue to extend `_segment_endpoints_anchored_to_net_pads` (or a sibling helper) to also detect via anchoring, and/or to snap via positions when the abutting segment moves.

2. **Coordination with PR #2477.**  Both PRs modify `core.py` mark_fn stall fallback in the same region:
   - PR #2477 inserts a via-blocked rip-up call at original line 3843, before the existing `targeted_fallback_count = 0` loop.
   - This PR (#2479) modifies the `still_failed` definition just above at line 3834 (adding `partial_failed`) and adds rip-up logic inside the per-failed-net loop starting at 3851.

   These will textual-conflict at merge time.  Whichever lands first should rebase the other.  Both are additive in semantics — no logical interference — so the rebase should be straightforward.

### CI / merge state

- `gh pr checks 2479`: Lint & Format pass, Test pass, Type Check pass.
- `mergeStateStatus`: CLEAN.

Promoting to `loom:pr` for merge.

PR #2485, opened 2026-05-03
fix(router): thread per-call spacing in CoupledPathfinder.route_coupled

### LINKED ISSUE #2484: CoupledPathfinder.route_coupled mutates self.target_spacing_cells across calls
## Symptom

Flagged by the PR #2478 judge ([review comment](https://github.com/rjwalters/kicad-tools/pull/2478#issuecomment-4366725274)):

> Minor: `route_coupled` permanently mutates `self.target_spacing_cells` upward; consider per-call local in a follow-up.

The auto-derived target-spacing logic (which uses actual start-pad spacing when it exceeds the rule) writes back to the instance attribute, so a second call to `route_coupled` on a different pair of pads inherits the wider spacing from the first call. This is a state-leak bug that doesn't manifest in tests today (each test creates a fresh `CoupledPathfinder`) but will silently widen diff-pair spacing on multi-pair boards processed sequentially with a reused instance.

## Reproduce

```python
from kicad_tools.router.diffpair_routing import CoupledPathfinder

pf = CoupledPathfinder(target_spacing_cells=3)
# Route pair A with pads 6 cells apart -> auto-widens to 6
pf.route_coupled(pair_a_p, pair_a_n, ...)
assert pf.target_spacing_cells == 6  # current behavior — leaked

# Route pair B with pads 3 cells apart — still uses 6 even though rule was 3
pf.route_coupled(pair_b_p, pair_b_n, ...)
# Pair B routed at 6-cell spacing, not the configured 3
```

## Fix

Convert the auto-derived spacing to a per-call local variable in `src/kicad_tools/router/diffpair_routing.py::CoupledPathfinder.route_coupled`. Don't write back to `self.target_spacing_cells`. Pass the local value to whatever downstream method uses it.

## Affected Files

- `src/kicad_tools/router/diffpair_routing.py` — `route_coupled` method, around the auto-derived spacing block.
- `tests/test_diffpair_npad.py` or `tests/test_diffpair.py` — add a test that calls `route_coupled` twice on the same instance with different pad geometries and asserts the second call doesn't inherit the first's widened spacing.

## Test Plan

- [ ] Unit test: instantiate `CoupledPathfinder` with `target_spacing_cells=3`, route a pair with naturally wider start-pad spacing, then route a pair with pads at the configured 3-cell spacing. Assert the second pair's routes have ~3-cell spacing, not the widened value.
- [ ] Regression: all existing diff-pair tests (74) still pass.

## Acceptance criteria

- [ ] `route_coupled` does not mutate `self.target_spacing_cells`.
- [ ] New regression test catches the state-leak.
- [ ] No regression on other diff-pair tests.

## Related

- #2478 — the PR that introduced the auto-derived spacing logic.
- #2473 — the issue #2478 closed.

rjwalters:
## Summary

- `CoupledPathfinder.route_coupled` no longer mutates `self.target_spacing_cells`. The widened spacing (Issue #2473's auto-widen for wide start pads) is now a per-call local threaded into `_get_coupled_neighbors` as a kwarg, defaulting to `self.target_spacing_cells` when absent.
- Read sites in `_get_coupled_neighbors` (approach radius and spacing tolerance) consume the per-call value.
- Adds `TestRouteCoupledTargetSpacingNotMutated` with three regression tests covering the leak, end-to-end second-call behavior, and preservation of the within-call widening.

Closes #2484

## Test plan

- [x] `uv run pytest tests/test_diffpair.py --no-cov` (63 passed)
- [x] `uv run pytest tests/test_diffpair.py tests/test_diffpair_npad.py tests/test_diffpair_routing_integration.py --no-cov` (77 passed)
- [x] `uv run ruff check src/kicad_tools/router/diffpair_routing.py tests/test_diffpair.py` (clean)
- [x] `uv run ruff format` applied; only the modified file plus a few adjacent collapsed line breaks
- [x] `uv run mypy src/kicad_tools/router/diffpair_routing.py` shows no new errors (the two pre-existing errors at lines 864 and 1612 predate this change)

## Notes

- Acceptance criteria from the curator notes:
  - [x] `route_coupled` does not write to `self.target_spacing_cells`
  - [x] `_get_coupled_neighbors` consumes the per-call spacing (kwarg with `None` fallback), preserving Issue #2478 approach-relaxation behavior
  - [x] Regression test asserts `pf.target_spacing_cells` unchanged after a wide-pad call
  - [x] All existing diff-pair tests still pass

### DIFF
--- a/src/kicad_tools/router/diffpair_routing.py
+++ b/src/kicad_tools/router/diffpair_routing.py
@@ -229,9 +229,7 @@ class CoupledPathfinder:
                         return True
         return False
 
-    def _is_at_goal(
-        self, pos: GridPos, goal: GridPos | None
-    ) -> bool:
+    def _is_at_goal(self, pos: GridPos, goal: GridPos | None) -> bool:
         """Check if a grid position is at the goal (ignoring layer)."""
         if goal is None:
             return False
@@ -246,11 +244,31 @@ class CoupledPathfinder:
         n_goal: GridPos | None = None,
         p_start: GridPos | None = None,
         n_start: GridPos | None = None,
+        target_spacing_cells: int | None = None,
     ) -> list[tuple[CoupledState, float, bool]]:
         """Generate valid coupled moves maintaining spacing.
 
+        Args:
+            state: Current coupled search state.
+            p_net: Net id for the positive trace.
+            n_net: Net id for the negative trace.
+            p_goal: Goal grid position for the positive trace.
+            n_goal: Goal grid position for the negative trace.
+            p_start: Start grid position for the positive trace.
+            n_start: Start grid position for the negative trace.
+            target_spacing_cells: Per-call effective target spacing
+                in grid cells.  When ``None``, falls back to
+                ``self.target_spacing_cells``.  Issue #2484: the
+                effective spacing is threaded as a kwarg so that
+                ``route_coupled`` can widen it for a single call
+                (e.g. when start pads sit further apart than the
+                configured spacing) without mutating instance state.
+
         Returns list of (new_state, cost, is_via) tuples.
         """
+        if target_spacing_cells is None:
+            target_spacing_cells = self.target_spacing_cells
+
         neighbors: list[tuple[CoupledState, float, bool]] = []
 
         # Issue #2473: Relax the spacing constraint when both traces
@@ -267,7 +285,7 @@ class CoupledPathfinder:
             # ``approach_radius`` is generous enough to admit the goal
             # pair's spacing difference but small enough that the
             # relaxation does not infect the rest of the search.
-            approach_radius = max(self.target_spacing_cells, 6)
+            approach_radius = max(target_spacing_cells, 6)
             if p_dist_to_goal <= approach_radius and n_dist_to_goal <= approach_radius:
                 approach_relaxed = True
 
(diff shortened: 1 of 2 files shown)


rjwalters:
## Loom Judge Review: Approved

Issue #2484 (state-leak in `CoupledPathfinder.route_coupled`) is cleanly fixed using the curator's Option A: per-call kwarg threading. No `try/finally` mutation hack, no behavior change to the within-call widening from #2473.

### Implementation review

**Option A implemented exactly as specified:**
- `_get_coupled_neighbors` gains `target_spacing_cells: int | None = None` kwarg (line 247) with the documented `None` fallback to `self.target_spacing_cells` (lines 269-270). This preserves backward compatibility for any caller (notably tests) that doesn't thread the kwarg.
- The three reads inside `_get_coupled_neighbors` flagged in the curator notes — `approach_radius` (line 288), `tolerance` (line 327), and the spacing tolerance check (line 328) — all consume the local `target_spacing_cells`. None reach back to `self`.
- `route_coupled` introduces `effective_target_spacing` (line 480), widens it locally if the start pads are further apart than configured (line 481-482), and passes it into `_get_coupled_neighbors` as `target_spacing_cells=effective_target_spacing` (line 529). The previous `self.target_spacing_cells = int(round(...))` mutation is gone.

**Mutation audit:** A grep over the modified file confirms `self.target_spacing_cells` is now only written in `__init__` (line 173) and read at the kwarg fallback (line 270) and the local-init in `route_coupled` (line 480). No other writes anywhere. The leak path is closed at the source.

**Code comments:** Both call sites (the kwarg docstring at lines 259-265 and the inline comment at lines 471-476) explicitly cite Issue #2484 and explain the lifetime contract, which makes this much harder to regress.

### Test coverage

`TestRouteCoupledTargetSpacingNotMutated` adds the three regression tests the issue called for:
1. `test_target_spacing_unchanged_after_wide_call` — locks the core invariant: attribute unchanged after a 6-cell wide-pad call.
2. `test_second_call_uses_configured_spacing` — end-to-end: wide call followed by narrow call, asserts attribute stays at 3 across both, and (when the second call routes) the result references the narrow nets — proof the search did not silently inherit the widened regime.
3. `test_widening_still_applies_within_a_single_call` — guards against an over-fix that would break #2473 by neutering the widening entirely.

The wide-pad fixture (6 grid cells, 2x the configured target) reliably exercises the widening branch in `route_coupled`. The pads are placed on a clean 40mm x 40mm grid so the test is independent of routing-success heuristics.

### Verification

- CI on GitHub: Lint pass, Type Check pass, Test pass (5m11s).
- Local `ruff check src/kicad_tools/router/diffpair_routing.py tests/test_diffpair.py`: clean.
- Local `pytest tests/test_diffpair.py tests/test_diffpair_npad.py tests/test_diffpair_routing_integration.py --no-cov`: **77 passed in 154s**.
- `gh pr view --json mergeStateStatus`: CLEAN, mergeable.

### Notes

- Pre-existing mypy warnings at lines 864 and 1612 predate this change and are unrelated (Builder noted them in the PR description).
- Several incidental ruff-format collapses to single-line function signatures are included in the diff. Low-risk and unrelated to the fix.

Ship it.

— Loom Judge

PR #2486, opened 2026-05-03
fix(router): add via-anchor guard to chain-aware DRC nudge

### LINKED ISSUE #2480: Regression: board 01 VOUT R2:1 disconnects after May-3 fix batch (was clean ✅)
## Symptom

Board 01 (voltage divider) regressed from a clean ✅ DRC PASSED 3/3 baseline to **67%** (2/3 nets) after merging the May 3 fix batch (#2462, #2463, #2466, #2467, #2465, #2464, #2473, #2476, #2475). The disconnected pad is `R2:1` on the `VOUT` net.

## Reproduce

```bash
uv run kicad-tools route boards/01-voltage-divider/output/voltage_divider.kicad_pcb \
  -o /tmp/board01.kicad_pcb --strategy negotiated --layers 4 --backend cpp \
  --no-cache --timeout 300
```

Expected: `SUCCESS: All signal nets routed, DRC passed!` (this was the result before the batch).
Observed:
```
PARTIAL: Routed 2/3 signal nets
  VOUT: 2/3 pads connected -- disconnected: R2:1
DRC PASSED (jlcpcb profile, 4 layers)
```

DRC passes — the connection is just silently dropped.

## Diagnostic output

```
  Total nets: 3
  Successful: 3        ← strategy reports success
  Final overflow: 1
...
Adaptive Grid Routing Summary
  Total routes: 5 (0 escape + 5 channel)   ← 5 routes for 3 nets
...
--- Optimizing traces ---
  Segments: 456 -> 26 (-94.3%)              ← huge segment collapse
  Vias:     3 -> 3 (-0.0%)
...
WARNING: 1 net(s) have disconnected pads in written output
  VOUT: 2/3 pads connected -- disconnected: R2:1
```

The mismatch between `Successful: 3` and the final 2/3 output, combined with `Total routes: 5 (0 escape + 5 channel)` (5 routes for 3 nets), suggests the strategy is keeping multiple route variants per net, and the trace optimizer's 94.3% segment collapse is dropping a critical segment that connected R2:1.

## Suspect commits (in merge order)

1. `cd9823c8` #2479 chain-aware DRC nudge — could be removing segments incorrectly when multiple route variants exist
2. `86d5562a` #2477 via-vs-via failure-reason rip-up — could be over-aggressive
3. `f9157530` #2472 stricter via blocking — could be refusing placements the optimizer expects
4. `a2c1dc33` #2471 HIGH_CURRENT_SIGNAL — affects net ordering even on small boards

## Bisection plan

Use `git bisect` between `53592b92` (clean) and `cd9823c8` (broken) running:
```bash
uv run kicad-tools route boards/01-voltage-divider/output/voltage_divider.kicad_pcb \
  -o /tmp/b01.kicad_pcb --strategy negotiated --layers 4 --backend cpp \
  --no-cache --timeout 300 2>&1 | grep "Nets routed:"
```

The bad bisect signal is `2/3` (or anything less than `3/3`).

## Affected Files

- `src/kicad_tools/router/drc_nudge.py` — chain-aware nudge added in #2479
- `src/kicad_tools/router/algorithms/negotiated.py` — via-blocked rip-up added in #2477
- `src/kicad_tools/router/optimizer/` (or wherever the 94.3% segment collapse happens)
- `src/kicad_tools/router/cpp/src/pathfinder.cpp` — stricter via blocking from #2472

## Test Plan

- [ ] `git bisect` to identify which of 9 PRs introduced the regression.
- [ ] Once found: write a regression test that exercises the small-board negotiated path with multiple route variants and verifies all pads stay connected after optimization.
- [ ] Verify board 04 and board 05 (which now ✅ DRC PASS) don't regress when the fix lands.
- [ ] Verify boards 02 and 03 don't regress.

## Acceptance criteria

- [ ] Board 01 returns to ✅ DRC PASSED 3/3 with default settings.
- [ ] No regression on boards 02/03/04/05.

## Related

- Cumulative regression from this morning's batch of 9 merges; the fixes individually passed DRC but interact poorly.
- Related: #2476 surfaced a similar pattern (board 02 went 7/8 → 6/8 from stricter via blocking before the failure-reason rip-up was added).

### LINKED ISSUE #2483: Chain-aware DRC nudge from #2479 doesn't handle via anchors — likely cause of board 01 VOUT regression (#2480)
## Symptom

The PR #2479 judge ([review comment](https://github.com/rjwalters/kicad-tools/pull/2479#issuecomment-4366695722)) flagged this as non-blocking at the time:

> **Via anchoring not handled.** `_nudge_segment_with_chain` walks same-net same-layer segments only — it doesn't inspect vias. If a segment endpoint coincides with a via center, nudging will move the segment off the via and break the chain at the via. Same class of bug as the pad-anchor case. Worth a follow-up issue.

After re-testing all five demo boards post-merge, board 01 (voltage divider) regressed from a clean ✅ DRC PASSED 3/3 to **67%** with `VOUT: 2/3 pads connected -- disconnected: R2:1`. Symptom matches: a single segment translation away from a via that connects layers, leaving R2:1 stranded on a different layer than the rest of VOUT.

This is filed as `#2480` (the board 01 regression) but the **root cause is here**: the chain-aware nudge correctly handles same-net same-layer chains terminating at pads, but ignores the case where a chain segment is anchored at a via that connects to a different-layer continuation.

## Reproduce

```bash
uv run kicad-tools route boards/01-voltage-divider/output/voltage_divider.kicad_pcb \
  -o /tmp/b01.kicad_pcb --strategy negotiated --layers 4 --backend cpp \
  --no-cache --timeout 300
grep "Vias:" /tmp/b01.kicad_pcb  # confirm 3 vias on a 3-net board
```

The VOUT net's diagnostic output:
```
WARNING: 1 net(s) have disconnected pads in written output
  VOUT: 2/3 pads connected -- disconnected: R2:1
```

VOUT routes through a via to reach R2:1; the nudge translates a segment near R2:1 without moving the via, breaking the layer transition.

## Fix sketch

Extend `_nudge_segment_with_chain` in `src/kicad_tools/router/drc_nudge.py` to:

1. Detect when a chain segment endpoint coincides with a via center on the same net (analogous to the existing `_segment_endpoints_anchored_to_net_pads` helper).
2. When such an endpoint exists, treat it as an anchor: do NOT translate the segment, OR translate the via along with the segment, OR refuse the nudge and look for an alternative DRC repair.

Mirror the conservative approach already used for pad anchors: prefer not nudging a chain whose endpoints can't all move together, even if it means the original DRC violation persists (the violation is detectable; a silent disconnect is not).

## Affected Files

- `src/kicad_tools/router/drc_nudge.py` — extend chain detection to include via-anchored endpoints. Add `_segment_endpoints_anchored_to_net_vias` helper analogous to `_segment_endpoints_anchored_to_net_pads`.
- `tests/test_drc_nudge.py` — add a regression test: route a 3-pad net with one pad on the opposite layer (forcing a via), apply chain-aware nudge to a segment whose endpoint coincides with the via, assert connectivity preserved.

## Test Plan

- [ ] Manual verification: board 01 returns to ✅ DRC PASSED 3/3.
- [ ] Automated unit test for via-anchored chain: minimal fixture with a 3-segment chain where one segment endpoint is at a via, assert chain-aware nudge preserves connectivity.
- [ ] Automated unit test for the no-op case: when a chain has both pad and via anchors that can't all move, assert the nudge declines and the original DRC violation is reported as-is.
- [ ] Regression: boards 02, 03, 04, 05 don't regress.

## Acceptance criteria

- [ ] Board 01 returns to ✅ DRC PASSED 3/3 with default settings.
- [ ] No regression on boards 02-05 (in particular, board 05 must still hit 10/10 — the fix that #2479 introduced).
- [ ] New regression tests cover both the via-anchor preservation case and the decline case.

## Related

- Closes #2480 (board 01 regression) if this is confirmed as root cause via bisection.
- #2479 — the PR that introduced the chain-aware nudge.
- #2475 — the issue #2479 closed.

rjwalters:
## Summary

The chain-aware DRC nudge from #2479 (`_nudge_segment_with_chain` in `src/kicad_tools/router/drc_nudge.py`) walks same-net **same-layer** segments to drag adjacent endpoints along when a violating segment is translated. Vias are layer transitions and live outside that walk, so a segment endpoint coincident with a via centre cannot be tracked: translating the segment slides it off the via and disconnects the chain at the layer transition.

This PR mirrors the existing pad-anchor guard with a parallel via-anchor guard. When either endpoint of a candidate segment is within `_VIA_ANCHOR_TOL` (0.02 mm) of any via on the same net, the nudge is declined so the original DRC violation surfaces in the post-save report instead of producing a silent disconnect.

## Changes

- `src/kicad_tools/router/drc_nudge.py`:
  - Added `_VIA_ANCHOR_TOL = 0.02` mm constant alongside `_PAD_ANCHOR_TOL`.
  - Added `_segment_endpoints_anchored_to_net_vias` helper, modelled exactly on `_segment_endpoints_anchored_to_net_pads`. The only structural difference is that vias live on per-route `Route.vias` lists rather than a top-level `router.vias` attribute, so the helper iterates `router.routes` and filters by `route.net`.
  - Added a guard call in `_nudge_segment_with_chain` immediately after the pad-anchor guard.
- `tests/test_drc_nudge.py`:
  - Added `TestSegmentEndpointsAnchoredToNetVias` (7 tests) covering endpoint detection on both endpoints, tolerance bounds, net filtering, missing-data tolerance, and cross-route same-net via detection.
  - Added `TestNudgeSegmentWithChainViaAnchor` (3 tests) covering the full via-anchored decline scenario, the via-only-anchor case (isolating the via guard from the pad guard), and a regression guard ensuring segments far from any via are still nudgable.

| Unit test for via-anchor preservation | PASS | `TestNudgeSegmentWithChainViaAnchor::test_via_anchored_segment_not_nudged` builds the 3-pad fixture (P1/P2 top, P3 bottom, single via) and asserts the nudge is declined and connectivity is preserved. |
| Unit test: no-op decline preserves segment byte-for-byte | PASS | `test_via_anchored_segment_not_pad_anchored_still_declines` captures pre-nudge endpoints and asserts they are unchanged after the declined nudge. |
| Boards 02 and 05 don't regress | PASS | Board 05 PHASE_B 10/10 confirmed (`SUCCESS: All signal nets routed, DRC passed!`). Board 02 8/8 confirmed (`All nets verified connected`). |
| Boards 03/04 don't regress | PASS for 03 | Board 03 10/13 matches baseline run on the same commit without the fix (same disconnects: JOY_X, USB_CC1, USB_D-). The 10/13 result is pre-existing and unrelated to this PR. |
| Board 01 returns to 3/3 (closes #2480) | NOT MET | See Note on #2480 below. The fix is correct in its own right but does not move board 01. |

## Note on #2480 (intentionally not closed)

The curator hypothesised that this fix would also close #2480 (board 01 VOUT R2:1 disconnect), but instrumented runs show that `_nudge_segment_with_chain` is **never invoked** during board 01 routing (zero nudge calls observed across the entire pipeline with both `--backend cpp` and `--backend auto`). The chain-aware nudge therefore cannot be the cause of the board 01 regression.

To verify, I:
1. Ran board 01 on the baseline (this commit's parent, `cd9823c8`) with `--backend cpp --strategy negotiated --layers 4 --no-cache --timeout 300`. Result: `VOUT: 2/3 pads connected -- disconnected: R2:1`.
2. Ran board 01 with this fix applied. Result: identical (`VOUT: 2/3 pads connected -- disconnected: R2:1`).
3. Ran with an instrumented build that counts `_nudge_segment_with_chain` calls. Result: **0 calls** for board 01.

`#2480` is left open and will need a different root-cause investigation. Likely candidates per the curator's secondary list: `optimizer/trace.py` segment collapse (94.3% reduction observed), `negotiated.py` rip-up logic from #2477, or `pathfinder.cpp` via blocking from #2472. The bisect script in #2480 should still identify the correct first-bad-commit; this PR's commit will register as `good` for that test.

## Test Plan

- [x] Unit tests: `uv run pytest tests/test_drc_nudge.py` — 66/66 PASS (10 new tests added).
- [x] Board 05: `uv run kicad-tools route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb -o /tmp/b05.kicad_pcb --strategy negotiated --layers 4 --backend cpp --no-cache --timeout 600` — 10/10 nets, all verified connected. PHASE_B routes through the same fixture #2475/#2479 fixed.
- [x] Board 02: `uv run kicad-tools route boards/02-charlieplex-led/output/charlieplex_3x3.kicad_pcb -o /tmp/b02.kicad_pcb --strategy negotiated --layers 4 --backend cpp --no-cache --timeout 600` — 8/8 nets, all verified connected.
- [x] Board 03: 10/13 — matches baseline (pre-existing, unrelated).
- [x] Board 01: 2/3 — matches baseline (#2480 has a different root cause; see note above).

Closes #2483

### DIFF
--- a/src/kicad_tools/router/drc_nudge.py
+++ b/src/kicad_tools/router/drc_nudge.py
@@ -158,6 +158,20 @@ def _nudge_segment_with_chain(
         )
         return False
 
+    # Via-anchor guard (Issue #2483).  A chain segment whose endpoint sits
+    # on a via centre is anchored to that layer transition: translating the
+    # segment slides it off the via and breaks the chain at the layer
+    # change, since the same-layer chain walk below cannot drag the via
+    # (and the via's other-layer continuation) along with it.  Decline the
+    # nudge so the original DRC violation surfaces instead of producing a
+    # silent disconnect.
+    if _segment_endpoints_anchored_to_net_vias(seg, seg.net, router):
+        logger.debug(
+            "Skipping nudge for net %s: segment is via-anchored",
+            seg.net,
+        )
+        return False
+
     # Capture pre-nudge endpoints so we can update neighbour segments.
     old_x1, old_y1 = seg.x1, seg.y1
     old_x2, old_y2 = seg.x2, seg.y2
@@ -219,6 +233,12 @@ def _nudge_segment_with_chain(
 # *actual* pad anchors, not nearby segment intersections.
 _PAD_ANCHOR_TOL = 0.02
 
+# Tolerance in mm for considering a segment endpoint anchored to a via
+# centre.  Vias also terminate segments at exact centres within float
+# rounding, so the same tolerance as ``_PAD_ANCHOR_TOL`` is appropriate
+# (Issue #2483).
+_VIA_ANCHOR_TOL = 0.02
+
 
 def _segment_endpoints_anchored_to_net_pads(
     seg: Segment,
@@ -272,6 +292,61 @@ def _segment_endpoints_anchored_to_net_pads(
             and abs(seg.y2 - pad.y) < _PAD_ANCHOR_TOL
         ):
             return True
+    return False
+
+
+def _segment_endpoints_anchored_to_net_vias(
+    seg: Segment,
+    net: int,
+    router: Autorouter,
+) -> bool:
+    """Return True when either endpoint of ``seg`` sits on a via centre of ``net``.
+
+    Companion to :func:`_segment_endpoints_anchored_to_net_pads`.  Vias are
+    layer transitions: the chain walk in :func:`_nudge_segment_with_chain`
+    intentionally restricts itself to same-layer same-net segments, which
+    means a via that links a top-layer segment to a bottom-layer segment
+    is invisible to the snap-update logic.  Translating a segment off a
+    via centre therefore disconnects the chain at the layer transition.
+
+    Issue #2483: the chain-aware nudge introduced by #2479 fixed the
+    same-layer chain disconnection case (board 05 PHASE_B) but did not
+    consider vias.  When a net routes through a via, a nudge can slide
+    the segment off the via centre, leaving the via dangling on the
+    other-layer continuation.
+
+    The conservative fix mirrors the pad-anchor guard: decline the nudge
+    when an endpoint is via-anchored.  Declining surfaces the original
+    DRC violation as a detectable signal rather than a silent disconnect.
+
+    Args:
+        seg: The segment about to be nudged.
+        net: The net the segment belongs to.
+        router: The autorouter, used to look up via positions per route.
+
+    Returns:
+        True if either endpoint of ``seg`` is within ``_VIA_ANCHOR_TOL`` of
+        any via on ``net``; False otherwise (or when route data is
+        unavailable).  Note that vias live on per-route ``Route.vias``
+        lists, not on a top-level ``router.vias`` attribute.
+    """
+    routes = getattr(router, "routes", None) or []
+    for route in routes:
+        if route.net != net:
+            continue
+        for via in route.vias:
+            # Endpoint 1
+            if (
+                abs(seg.x1 - via.x) < _VIA_ANCHOR_TOL
+                and abs(seg.y1 - via.y) < _VIA_ANCHOR_TOL
+            ):
+                return True
+            # Endpoint 2
+            if (
+                abs(seg.x2 - via.x) < _VIA_ANCHOR_TOL
+                and abs(seg.y2 - via.y) < _VIA_ANCHOR_TOL
+            ):
+                return True
     return False
 
 
--- a/tests/test_drc_nudge.py
+++ b/tests/test_drc_nudge.py
@@ -19,6 +19,7 @@ from kicad_tools.router.drc_nudge import (
     _perpendicular_unit,
     _reconnect_segments,
     _segment_endpoints_anchored_to_net_pads,
+    _segment_endpoints_anchored_to_net_vias,
     _segment_length,
     drc_verify_and_nudge,
 )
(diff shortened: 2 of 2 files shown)


rjwalters:

Clean implementation that mirrors the existing pad-anchor guard pattern exactly. All curator-recommended structural details verified.

### Structural verification

- `_segment_endpoints_anchored_to_net_vias` is a faithful companion to `_segment_endpoints_anchored_to_net_pads`: same signature, same early-return-on-missing-data behaviour, same endpoint-1 / endpoint-2 check structure, parallel docstring.
- Helper correctly iterates `router.routes` and filters by `route.net` to access per-route `route.vias` (the curator-flagged "vias live on per-route lists, not on a top-level router attribute" detail is handled correctly).
- `_VIA_ANCHOR_TOL = 0.02` matches `_PAD_ANCHOR_TOL = 0.02` and is documented with rationale.
- Guard insertion site is the natural spot — immediately after the pad-anchor guard at the top of `_nudge_segment_with_chain`. Same "decline rather than translate" philosophy, same debug-log + `return False` pattern.

### Test quality

10 new tests, none tautological:

- `TestSegmentEndpointsAnchoredToNetVias` (7 tests): both-endpoint detection, tolerance bounds (0.05 mm offset rejected), no-vias case, cross-net rejection, missing-routes tolerance, cross-route same-net detection.
- `TestNudgeSegmentWithChainViaAnchor` (3 tests):
  - `test_via_anchored_segment_not_nudged` — the headline regression test. Builds a 3-pad fixture (P1/P2 top, P3 bottom) with a via and a top-bottom seg_a/seg_b chain, asserts the nudge is declined, asserts seg_a/via byte-unchanged, AND asserts seg_b's via-centre endpoint is preserved (chain integrity verified, not just refusal verified).
  - `test_via_anchored_segment_not_pad_anchored_still_declines` — critically isolates the new guard from the pre-existing pad guard by placing the segment's free endpoint at (0, 5) where there is no pad. This proves the new guard is actually firing.
  - `test_segment_far_from_any_via_is_still_nudged` — negative regression: a via exists on the net but at (50, 50), away from the segment at (0, 0)-(10, 0); the nudge proceeds and `seg.y` moves to 0.2. This guards against any future regression that would short-circuit on "route has any via" rather than per-via coordinate test.

### #2479 regression check

Board 05 PHASE_B reported as 10/10 with `SUCCESS: All signal nets routed, DRC passed!` — the original #2479 fix is intact. The two anchor guards are independent (via guard only adds a new decline path; PHASE_B is pad-anchored, not via-anchored), and all 66 existing tests pass.

### #2480 acknowledgement

The PR appropriately leaves #2480 open. The builder's instrumentation showed 0 invocations of `_nudge_segment_with_chain` during board 01 routing on both `--backend cpp` and `--backend auto`, which conclusively rules out the chain-aware nudge as the root cause for the VOUT R2:1 regression. The honest "fix is correct but doesn't move board 01" note plus the documented likely candidates (`optimizer/trace.py` segment collapse, `negotiated.py` rip-up, `pathfinder.cpp` via blocking) is the right call. #2480 stays open with diagnostic data attached for follow-up.

### CI / merge state

- Lint & Format: PASS (15s)
- Test: PASS (5m22s)
- Type Check: PASS (32s)
- mergeStateStatus: CLEAN
- Local re-run of `tests/test_drc_nudge.py`: 66/66 pass.

Approved. Promoting to `loom:pr` for merge.

PR #2488, opened 2026-05-03
fix(router): dedupe RSMT sub-route vias and invalidate cpp stored vias on rip-up

### LINKED ISSUE #2481: Board 02 still has via-vs-via violations after #2466/#2472 — multiple negative clearances on NODE_C interactions
## Symptom

After PR #2472 (closes #2466, supposed to align C++ via blocking with the validator), board 02 STILL produces multiple via-vs-via DRC violations on the negotiated strategy. Some clearances are deeply negative (overlapping vias by ~0.36mm).

## Reproduce

```bash
uv run kicad-tools route boards/02-charlieplex-led/output/charlieplex_3x3.kicad_pcb \
  -o /tmp/board02.kicad_pcb --strategy negotiated --layers 4 --backend cpp \
  --no-cache --timeout 600
```

Observed (8/8 nets routed but DRC fails):
```
[pad] LINE_D vs NODE_C at (133.00, 138.00) on F.Cu: -0.043mm (required 0.150mm)
[via] NODE_A vs NODE_C at (132.00, 109.40) on F.Cu: -0.046mm (required 0.200mm)
[via] NODE_B vs NODE_C at (117.50, 117.40) on B.Cu: -0.233mm (required 0.200mm)
[via] NODE_B vs NODE_C at (125.50, 126.60) on B.Cu: -0.165mm (required 0.200mm)
[via] NODE_B vs NODE_C at (125.50, 126.60) on B.Cu: -0.366mm (required 0.200mm)
[via] NODE_B vs NODE_C at (125.50, 126.60) on B.Cu: -0.184mm (required 0.200mm)
[via] NODE_C vs NODE_A at (123.40, 110.20) on In2.Cu: -0.326mm (required 0.200mm)
[via] NODE_C vs NODE_B at (124.70, 108.60) on In2.Cu: -0.114mm (required 0.200mm)
[via] NODE_C vs NODE_A at (130.10, 108.60) on In2.Cu: -0.068mm (required 0.200mm)
[via] NODE_C vs NODE_B at (124.70, 108.60) on In2.Cu: -0.105mm (required 0.200mm)
```

## Why this is surprising

- #2472 added a geometric via-vs-via clearance check to `Pathfinder::is_via_blocked` that mirrors the post-route validator. The check is supposed to **refuse** placements with negative clearance.
- #2477 added structured failure-reason rip-up so when the search refuses, the negotiated strategy can target the blocker.
- Yet the final routed PCB has 4+ overlapping vias on NODE_C alone.

## Hypothesis

Three possible mechanisms:
1. The via blocking check is per-A*-step but vias added by **later iterations'** rip-up don't re-validate against earlier vias.
2. The post-route DRC nudge from #2479 may translate vias into overlap positions while preserving chain connectivity.
3. Multiple route variants per net (`Total routes: 5 (0 escape + 5 channel)` pattern seen in #2480) mean some via-bearing variants survive output even after blocking.

The same `(125.50, 126.60)` location appears 3 times as NODE_B vs NODE_C — suggests duplicate via emission.

## Bisection plan

Same approach as #2480: bisect between `53592b92` (pre-batch) and `cd9823c8` (post-batch) on:
```bash
uv run kicad-tools route boards/02-charlieplex-led/output/charlieplex_3x3.kicad_pcb \
  -o /tmp/b02.kicad_pcb --strategy negotiated --layers 4 --backend cpp \
  --no-cache --timeout 600 2>&1 | grep "Violations:"
```

If DRC violations existed at every commit in that range, this is a pre-existing issue masked by completion-rate regressions and unmasked now that 8/8 nets actually get placed. If violations appeared at a specific commit, that's the regression point.

## Affected Files

- `src/kicad_tools/router/cpp/src/pathfinder.cpp` — verify via-blocking is consulted in BOTH initial routing and rip-up reroute
- `src/kicad_tools/router/algorithms/negotiated.py` — verify rip-up doesn't bypass via blocking
- `src/kicad_tools/router/drc_nudge.py` — verify nudge respects via clearance after translation
- `src/kicad_tools/router/optimizer/` — verify segment optimization doesn't introduce new via positions

## Test Plan

- [ ] Bisect to find regression point (or confirm pre-existing).
- [ ] Add a board-level integration test: route board 02, parse output PCB, assert 0 via-vs-via violations.
- [ ] Add a unit test for the rip-up reroute path: place 2 vias on different nets at the same location; rip up one; reroute; verify the new via doesn't reuse the same location.

## Acceptance criteria

- [ ] Board 02 routes 8/8 signal nets with 0 via-vs-via DRC violations.
- [ ] No regression on boards 01/03/04/05.

## Related

- #2466 / #2472 — the fix this regression follows.
- #2476 / #2477 — the failure-reason rip-up infrastructure.
- #2480 — companion regression (board 01 VOUT) from the same May 3 batch.

rjwalters:
## Summary

Issue #2481: Board 02 (charlieplex_3x3) was producing via-vs-via DRC violations even after #2466/#2472 added geometric via-blocking. The curator identified three distinct bugs; this PR addresses bugs 1 and 2 (the silent-corruption ones).

**Bug 1 — RSMT sub-route via duplication.** `route_net_negotiated` in `algorithms/negotiated.py` decomposes a multi-pin net into RSMT edges and emits one `Route` per edge, each with its own `vias` list. When two edges meet at a Steiner tap and both require a layer transition, `validate_layer_transitions` runs independently per sub-route and inserts a via at the same (x, y) coord. The post-route validator then triple-counts those duplicates against any cross-net via, and the cpp `stored_vias_` snapshot ends up with redundant entries.

**Bug 2 — C++ `stored_vias_` was append-only.** `CppPathfinder._sync_stored_routes` tracks `_synced_route_count` and only forwards new routes. `RoutingGrid.unmark_route` ripped Python cells but never cleared the cpp side, so `Pathfinder::is_via_blocked_diag` kept consulting via positions from already-ripped-up nets. After enough rip-up cycles, cpp refusal logic rejected candidates against non-existent vias, and the eventual re-route could legitimately collide with a real via the cpp side never saw.

**Bug 3 — `_try_nudge_via_via` dispatcher** (curator's analysis) is intentionally deferred. The two fixes here already drive board 02 to 0 cross-net via-vs-via violations. Bug 3 has overlap with the in-flight #2483 PR in `drc_nudge.py`.

## Changes

- `Grid3D::clear_stored_routes()` (cpp): clears only stored routes (segments + vias), keeps pads. Exposed via bindings.
- `CppGrid.invalidate_stored_routes()` (Python wrapper): clears cpp routes vector and resets sync watermark to 0 so next `_sync_stored_routes` rebuilds from `py_grid.routes`.
- `CppGrid.from_routing_grid()` now stores a back-reference on the Python grid (`RoutingGrid._cpp_grid`).
- `RoutingGrid.unmark_route` invokes the invalidate hook after removing the route (only when the route was actually present).
- `_dedupe_sibling_route_vias` helper in `algorithms/negotiated.py` consolidates same-coord same-net vias across per-edge sub-routes; expands the survivor's layer span when needed. Called at the end of `route_net_negotiated` for multi-pin nets.

## Verification

- **Board 02**: 8/8 nets routed; **0 cross-net via-vs-via violations** (verified by parsing the routed PCB pairwise). Previously had ~3-6 such violations.
- **Boards 01, 02, 05**: 0 same-net duplicate-position vias, 0 cross-net via-vs-via violations.
- **Board 05**: routes 9/10 nets vs 8/10 on main — slight improvement.
- **Existing tests**: 10/10 in `test_router_cpp_via_blocked_failure.py` still pass.
- **New tests**: 6 added covering the rip-up invalidation hook and the dedup helper.
- **Full router test suite**: 573 tests pass.

## Test plan

- [x] `tests/test_router_cpp_via_blocked_failure.py` — all 16 tests (10 existing + 6 new) pass
- [x] Comprehensive router test suite (573 tests) passes
- [x] Board 02 regression: 0 via-vs-via violations
- [x] Boards 01, 05 don't regress (0 via-vs-via, no duplicate vias)
- [ ] Board 03 / 04 detailed verification — board 03 routing was timeout-prone before this change too; not a regression

Closes #2481

### DIFF
--- a/src/kicad_tools/router/algorithms/negotiated.py
+++ b/src/kicad_tools/router/algorithms/negotiated.py
@@ -20,7 +20,7 @@ if TYPE_CHECKING:
     from ..congestion_estimator import CongestionEstimator
     from ..grid import RoutingGrid
     from ..pathfinder import Router
-    from ..primitives import Pad, Route
+    from ..primitives import Pad, Route, Via
     from ..rules import DesignRules, NetClassRouting
 
 
@@ -354,6 +354,95 @@ def calculate_congestion_tuned_params(
     return adjusted_mult, adjusted_hist
 
 
+# Resolution used when grouping vias by (x, y) for cross-route same-net
+# deduplication.  Vias whose coordinates round to the same key (to the
+# nearest micron) are treated as the same physical placement.
+_VIA_DEDUP_RESOLUTION_MM = 0.001
+
+
+def _dedupe_sibling_route_vias(
+    routes: list[Route],
+    *,
+    resolution_mm: float = _VIA_DEDUP_RESOLUTION_MM,
+) -> int:
+    """Drop duplicate same-net vias across sibling Route objects (Issue #2481).
+
+    A multi-pin net routed via RSMT decomposition emits one ``Route``
+    object per Steiner edge, each carrying its own ``vias`` list.  When
+    two RSMT edges meet at a Steiner tap and both edges require a layer
+    transition there, they each insert a via at the same (x, y).  These
+    duplicates survive into the post-route validator, where each pair of
+    sibling vias against any cross-net via produces a separate
+    via-vs-via violation -- inflating the reported defect count and, for
+    the C++ side, polluting ``stored_vias_`` with redundant entries.
+
+    This function keeps the first via per (rounded x, rounded y) across
+    *all* the supplied sibling routes, expands its layer span to cover
+    every duplicate, and removes the duplicates from their owning route.
+    Segments are not touched: they already terminate at the Steiner tap
+    coordinate, which still has a via after this dedup.
+
+    Args:
+        routes: List of sibling ``Route`` objects produced for a single
+            net.  Mutated in place.
+        resolution_mm: Coordinate rounding for the dedup key.  Defaults
+            to 1 micron, well below the routing grid resolution.
+
+    Returns:
+        The number of duplicate vias removed across all routes.
+    """
+    if len(routes) < 2:
+        return 0
+
+    from ..layers import Layer
+
+    inv_res = 1.0 / resolution_mm
+    seen: dict[tuple[int, int], Via] = {}
+    removed = 0
+
+    for route in routes:
+        if not route.vias:
+            continue
+        keep_indices: list[int] = []
+        for idx, via in enumerate(route.vias):
+            key = (
+                int(round(via.x * inv_res)),
+                int(round(via.y * inv_res)),
+            )
+            existing = seen.get(key)
+            if existing is None:
+                seen[key] = via
+                keep_indices.append(idx)
+            else:
+                # Expand the surviving via's layer range to cover this
+                # duplicate's layer range, in case the two edges had
+                # different layer pairs at the same tap (mid-layer
+                # buried via on one edge, full-stack on the other).
+                min_layer = min(
+                    existing.layers[0].value,
+                    existing.layers[1].value,
+                    via.layers[0].value,
+                    via.layers[1].value,
+                )
+                max_layer = max(
+                    existing.layers[0].value,
+                    existing.layers[1].value,
+                    via.layers[0].value,
+                    via.layers[1].value,
+                )
+                if (
+                    min_layer != existing.layers[0].value
+                    or max_layer != existing.layers[1].value
+                ):
+                    existing.layers = (Layer(min_layer), Layer(max_layer))
+                removed += 1
+
+        if len(keep_indices) != len(route.vias):
+            route.vias = [route.vias[k] for k in keep_indices]
+
+    return removed
+
+
 class NegotiatedRouter:
     """PathFinder-style negotiated congestion router.
 
@@ -504,6 +593,15 @@ class NegotiatedRouter:
                     self._record_via_blocked_failure(source_pad.net)
                     if failure_callback is not None:
                         failure_callback(source_pad, target_pad)
+
+            # Issue #2481: Dedupe vias at the same (x, y) location across
+            # the multi-edge RSMT sub-routes for this net.  Two RSMT edges
+            # that meet at a Steiner tap on the same layer pair would each
+            # insert a via at that tap, producing duplicate same-net vias
+            # that the post-route DRC counts as triple-pair violations.
+            # We keep the first occurrence per (rounded) (x, y) and drop
+            # the rest from sibling sub-routes.
+            _dedupe_sibling_route_vias(routes)
         else:
             # 2-pin net
             route = self.router.route(
--- a/src/kicad_tools/router/cpp/include/grid.hpp
+++ b/src/kicad_tools/router/cpp/include/grid.hpp
@@ -116,6 +116,18 @@ public:
     // Clear all stored validation data (pads, segments, vias).
     void clear_validation_data();
 
+    // Clear only stored routes (segments + vias), keeping pads.
+    // Issue #2481: Used by CppGrid.invalidate_stored_routes() after a
+    // rip-up on the Python side.  ``Pathfinder::is_via_blocked_diag``
+    // (Issue #2466) consults ``stored_vias_`` to refuse via placements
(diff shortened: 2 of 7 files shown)


rjwalters:

Thorough review against the curator's two-bug specification. Both fixes are correct, well-tested, and surgical.

### Checklist verification

- [x] **`clear_stored_routes` clears ONLY routes (not pads).** Verified in `grid.cpp:307-313` (only `stored_segments_.clear()` and `stored_vias_.clear()`; `pads_` deliberately left alone). Test `test_invalidate_stored_routes_method_exists_and_works` explicitly asserts `pad_count == pad_count_before` after the call.
- [x] **`unmark_route` calls invalidate hook AFTER cell rip-up and list removal.** The `removed = True` flag is set inside the `with self._acquire_lock()` block; the cpp invalidate call happens after the lock context exits, only when the route was actually present in `self.routes`. State is fully consistent at the hook point.
- [x] **End-to-end test for post-rip-up via search.** `test_unmark_route_then_via_search_unblocked` confirms `is_via_blocked` rejects before `unmark_route` and accepts after.
- [x] **Multi-pin dedup test.** `test_dedupe_helper_collapses_duplicate_vias` covers the same-net-same-coord case with mixed layer spans.
- [x] **Test split: 3 for bug 1, 3 for bug 2.** `TestUnmarkRouteInvalidatesStoredVias` (3 tests for the cpp invalidation hook) and `TestRsmtSubRouteViaDedup` (3 tests for the dedup helper).
- [x] **No conflict with PR #2486.** PR #2486 touches `drc_nudge.py`/`test_drc_nudge.py`; this PR touches `negotiated.py` / `grid.py` / `cpp_backend.py` / cpp grid files. Disjoint.
- [x] **No router test regressions vs main.** Main: 96 failed / 14306 passed. This PR: 95 failed / 14310 passed. The 95 remaining failures are pre-existing in unrelated subsystems (`test_negotiated_adaptive`'s mock-based tests with `self.grid = None`, `test_fix_vias` layer detection, `test_consistency` library lookup, etc.) — all present on `main` before this PR.
- [x] **Board 05 8→9 improvement is real, not a bypassed check.** The fix clears the stale `stored_vias_` snapshot so `validate_route` and `is_via_blocked_diag` see the current world. Validation still runs on the same code path; one more legitimately-routable net now succeeds because the cpp side stops rejecting against ghost vias.
- [x] `mergeStateStatus = CLEAN`, `mergeable = MERGEABLE`.

### Findings (non-blocking)

1. **Dedup key is (x, y) only — different layer pairs at the same point are merged, not preserved.** The implementation deliberately collapses an `(F_CU, IN1_CU)` via and an `(IN1_CU, B_CU)` via at the same coord into a single `(F_CU, B_CU)` via on the same net (see `test_dedupe_helper_collapses_duplicate_vias`). For an RSMT Steiner tap on a single net this is sound — there is one physical hole at the tap, and the union span is what the combined incident edges actually need — but it is a stronger consolidation than a pure `(x, y, source, dest)` key. Worth flagging as a deliberate design choice in case future work wants the more conservative semantics.

2. **Strong reference cycle between `CppGrid._py_grid` and `RoutingGrid._cpp_grid`.** Both objects share the same logical lifetime so this is fine in practice (CPython's cyclic GC will reclaim), but a `weakref.ref` on the back-reference side would avoid relying on the cyclic collector. Not a blocker.

3. **Unrelated CI hygiene observation.** `.github/workflows/ci.yml` has `continue-on-error: true` on every check step (lint, format, mypy, tests). `gh pr checks` reports "pass" even when `pytest` exits 1 with 95 failures. Pre-existing to this PR — flagging in case the repo wants this fixed separately so the judge workflow's CI signal is meaningful again. I verified actual outcomes by reading the workflow run log, which is how I confirmed no regressions.

### Verdict

The two-bug surgical fix is exactly what the curator scoped. Implementation matches the analysis. Tests are well-targeted. Outcome (board 02 zero violations, board 05 net-positive) supports the claim. Ship it.

Approving and updating labels.

🤖 Loom Judge

PR #2489, opened 2026-05-03
fix(router): bump connector-siblings of prerouted nets in negotiated ordering (#2482)

### LINKED ISSUE #2482: Board 03: JOY_X and USB_CC1 still unrouted at 85% after diff-pair pre-pass and N-pad work
## Symptom

After PRs #2474 (diff-pair pre-pass), #2478 (N-pad CoupledPathfinder), and the rest of the May 3 fix batch, board 03 (USB joystick) still routes only 11/13 signal nets (85%). The failure pattern **changed** but completion stayed flat:

**Before May 3 batch** (failures): USB_D+ (0/2), JOY_X (0/2)
**After May 3 batch** (failures): JOY_X (0/2), **USB_CC1** (0/2)

So #2474/#2478 successfully routed USB_D+/USB_D- (good — that infrastructure works), but two different nets are now blocking 100% completion.

## Reproduce

```bash
  -o /tmp/board03.kicad_pcb --strategy negotiated --layers 4 --backend cpp \
  --no-cache --timeout 600
```

Observed:
```
WARNING: 2 net(s) have disconnected pads in written output
  JOY_X:    0/2 pads connected -- disconnected: U1:9, J2:3
  USB_CC1:  0/2 pads connected -- disconnected: U1:27, J1:A5
```

### (a) JOY_X — analog signal still unroutable

Same net that failed before. The PR #2478 builder noted: "Board 03 end-to-end acceptance not fully achieved due to a deeper trace-blocking-vs-pad-clearance interaction that affects both the standard and coupled pathfinders." JOY_X is a long analog trace from MCU pin U1:9 to connector J2:3, likely blocked by congestion in the MCU pin field after USB_D+/D- and other nets place themselves first.

### (b) USB_CC1 — new failure exposed by enabling USB diff-pair routing

USB_CC1 is the USB-C **Configuration Channel** line, a single-ended signal (not a differential pair). Its failure to connect U1:27 → J1:A5 is a **new** failure mode — it likely worked before because USB_D+/D- weren't routing, so USB_CC1 wasn't competing for the same MCU/connector corridor. Now that USB_D+/D- claim the corridor first via the diff-pair pre-pass, USB_CC1 is left without a path.

This is a positive sign for the diff-pair pre-pass (it's prioritizing correctly) but reveals a downstream coupling issue: nets sharing the same connector still need coordinated escape routing, not just the diff-pair set.

## Root cause analysis

The diff-pair pre-pass (`DiffPairRouter.route_diffpair_prepass`, `src/kicad_tools/router/diffpair_routing.py:1424`) runs the CoupledPathfinder for USB_D+/D- and reserves their grid cells. The negotiated strategy then sees those nets in `self.routes` and skips them via the prerouted-net guard (`src/kicad_tools/router/core.py:3184-3197`). The remaining nets — including USB_CC1, which terminates at the *same* USB-C connector J1 — are then routed in net-class priority order without any awareness that they share J1's pin field with the already-claimed D+/D- corridor.

USB_CC1 is single-ended, so it isn't a diff-pair member; it falls into the standard signal tier and competes on its own. By the time the negotiated A* tries to escape U1:27 toward J1:A5, the most direct corridor is already occupied by D+/D- traces, and J1's pin field is too tight for a separate escape.

The same argument applies to JOY_X: it is a long analog trace that must traverse from U1 down to J2, and after MCU-side escape routing places itself for the higher-priority nets, the JOY_X corridor is squeezed.

## Implementation guidance

There are three plausible fixes, in increasing order of invasiveness:

### Option 1: Connector-sibling priority bump in negotiated ordering (cheapest)

After the diff-pair pre-pass, raise the priority of any single-ended net whose pads land on the same connector component as a just-pre-routed diff-pair. Concretely, in `route_all_negotiated` (`src/kicad_tools/router/core.py:3179-3203`), after the prerouted-net filter at line 3197, run something like:

```python
prerouted_destinations: set[str] = set()
for net_id in prerouted_nets:
    prerouted_destinations |= self._get_net_destination_components(net_id)
# Boost any remaining net that shares a destination with a prerouted one
for net_id in net_order:
    if self._get_net_destination_components(net_id) & prerouted_destinations:
        # Move toward front of same priority tier
        ...
```

The existing `_find_same_tier_destination_siblings` (`src/kicad_tools/router/core.py:1925-1976`) only triggers within the rip-up loop. The cheapest change is a new sibling-of-prerouted-net helper that runs at ordering time — likely a small tweak to `_get_net_priority` or a custom secondary sort key.

This handles USB_CC1 directly. JOY_X is destination-shared with ADC_2, ADC_3, etc. on J2; the same heuristic helps if we extend it beyond just diff-pair siblings.

### Option 2: Coupled-escape entry point in escape.py (mid-cost)

Extend `EscapeRouter` (`src/kicad_tools/router/escape.py:586`) with a method that takes a *set* of nets sharing a destination connector and plans their escapes jointly, similar to how N-pad CoupledPathfinder (#2473/#2478) handles diff pairs. The natural slot is alongside `generate_escapes` (line 687) — a new `generate_coupled_escapes(connector_pads, nets)` that lays out USB_D+/D-/CC1 escapes together, then hands the partial routes to the negotiated pass to extend.

This is more general but requires teaching the escape router about logical net groups, not just package geometry.

### Option 3: Two-phase coupled escape with rip-up of pre-pass on failure (most invasive)

If JOY_X and USB_CC1 both fail despite Option 1, the underlying issue is that the diff-pair pre-pass over-commits the corridor and there's no path for any single-ended sibling. In that case the pre-pass needs to know it must reserve room for non-pair connector-siblings, or be willing to be ripped up.

### Recommended approach

Start with Option 1 (sibling-of-prerouted-net ordering bump). Validate with the diagnostic commands below. If MC/GA also fail with the new ordering, escalate to Option 2.

## Investigation hints (run these first to scope the fix)

```bash
# Confirm whether MC or GA can find a path for the new failure pattern:
  -o /tmp/b03_mc.kicad_pcb --strategy monte-carlo --mc-trials 30 \
  --layers 4 --backend cpp --no-cache --timeout 600

  -o /tmp/b03_ga.kicad_pcb --strategy evolutionary --pop-size 12 --generations 8 \
  --layers 4 --backend cpp --no-cache --timeout 300
```

If MC/GA succeeds: bug is local to negotiated's net ordering after the diff-pair pre-pass — Option 1 is sufficient.

If neither succeeds: underlying issue is geometric (J1's pin field too dense for sequential escape) — implement Option 2 with a coupled-escape entry point analogous to #2475's `_find_same_tier_destination_siblings`, but applied at ordering / escape-planning time.

## Affected Files

Verified entry points (path:line):

- `src/kicad_tools/router/diffpair_routing.py:1424` — `route_diffpair_prepass`, the pre-pass that produces the diff-pair routes the negotiated loop must accommodate.
- `src/kicad_tools/router/core.py:3179-3203` — `route_all_negotiated` net ordering after the prerouted-net filter; this is where Option 1's connector-sibling priority bump goes.
- `src/kicad_tools/router/core.py:1925-1976` — `_find_same_tier_destination_siblings` (added by #2475). Currently only used in the rip-up loop at lines 3716 and 4126; it can be reused or paralleled to drive an "ordered-near-prerouted-siblings" rule.
- `src/kicad_tools/router/core.py:1900` — `_get_net_destination_components`, the building block both helpers depend on.
- `src/kicad_tools/router/core.py:1878` — `_get_net_class_priority`, used for tier comparison.
- `src/kicad_tools/router/escape.py:586` (`EscapeRouter` class), `escape.py:687` (`generate_escapes`), `escape.py:2223` (`apply_escape_routes`) — slot for Option 2's coupled-escape entry point.
- `src/kicad_tools/cli/route_cmd.py:4006-4028` — where the diff-pair pre-pass is wired into the negotiated path; the new heuristic flows through `route_all_negotiated` so this file does not need new flags.

## Test Plan

- [ ] **Diagnostic step (do first)**: Run the MC and GA reproduce commands above on `boards/03-usb-joystick/output/usb_joystick.kicad_pcb`. Capture which strategy (if any) reaches 13/13 nets, to scope the fix between Option 1 (ordering) and Option 2 (coupled escape).
- [ ] **Manual acceptance**: Board 03 routes 13/13 signal nets with default settings:
      `uv run kicad-tools route boards/03-usb-joystick/output/usb_joystick.kicad_pcb -o /tmp/b03.kicad_pcb --strategy negotiated --layers 4 --backend cpp --no-cache --timeout 600`
- [ ] **Automated unit test (new)**: Add to `tests/test_diffpair_routing_integration.py` a fixture board with one USB-style differential pair and one single-ended net (CC-line analog), all three terminating at the same connector. Call `route_diffpair_prepass(config)` then `route_all_negotiated()` and assert all three nets connect their pads. This locks in the connector-sibling ordering invariant introduced by the fix.
- [ ] **Automated unit test (helper)**: Add to `tests/test_router_negotiated_high_current.py` (where `_find_same_tier_destination_siblings` is exercised today) a test that constructs a routed-net set including a connector-sibling diff-pair plus a single-ended net and asserts the new ordering helper places the single-ended net immediately after the prerouted set.
- [ ] **Regression**: Re-run boards 01, 02, 04, 05 routing and verify their pre-fix completion rates are unchanged or improved. Capture before/after net-completion counts in the PR description.
- [ ] **Negative regression**: Run `pytest tests/test_diffpair_routing_integration.py tests/test_diffpair_npad.py tests/test_router_negotiated_high_current.py -q` to confirm none of the existing diff-pair / sibling-ripup tests regress.

## Acceptance criteria

- [ ] Board 03 routes all 13 signal nets with default settings (`--strategy negotiated --layers 4 --backend cpp`).
- [ ] No completion-rate regression on boards 01, 02, 04, 05.
- [ ] New automated test (one diff-pair + one shared-connector single-ended net) passes deterministically.
- [ ] Existing diff-pair, escape, and sibling-ripup tests remain green.

## Related

- #2473 (closed, merged via #2478) — N-pad CoupledPathfinder.
- #2474 (merged) — `--differential-pairs` wiring with coupled-only pre-pass.
- #2475 (closed) — added `_find_same_tier_destination_siblings`; this issue reuses the helper at ordering time rather than rip-up time.
- #2478 (merged) — extended CoupledPathfinder to N-pad pairs; builder explicitly flagged board 03 e2e as a follow-up.

rjwalters:
## Summary

Closes #2482.

Adds a connector-sibling priority bump in `route_all_negotiated` so that single-ended nets terminating on the same connector as a prerouted differential pair (e.g. USB_CC1 on the USB-C J1 connector after the USB_D+/D- pre-pass) sort to the front of their priority tier.  Without this, the diff-pair pre-pass was claiming the connector's pin-field corridor first, then USB_CC1 was routed in plain priority order with no awareness that its escape route was already consumed.

## Implementation

- New helper `_find_connector_siblings_of_prerouted_nets` in `src/kicad_tools/router/core.py:1925`.  Returns the set of candidate nets whose pads share at least one destination component with any prerouted net.  Mirrors the shape of #2475's `_find_same_tier_destination_siblings` but operates across the whole prerouted set and does not require same-tier match (a HIGH_SPEED diff pair and a DIGITAL connector-sibling still need ordering coordination).  Same priority floor (only candidates with class priority < 10) is preserved.
- In `route_all_negotiated` (`core.py:3281`), after the existing prerouted-net filter, an additional stable-sort pass keys on `(class_priority, is_connector_sibling_of_prerouted_flag) + rest_of_priority_tuple` so that within each tier, sibling nets sort first.  Only fires when there are prerouted nets *and* at least one identified connector sibling.

## Diagnostics (Step 1 of issue)

Confirmed before implementing:

- `negotiated --differential-pairs` (baseline): 10/13 on board 03 — fails USB_D-, USB_CC1, JOY_X.  `Skipping diff-pair pre-pass: coupled pathfinder found no path` for the USB pair, so the prerouted set is currently empty on this board.
- `monte-carlo --mc-trials 30`: 11/13 — fails USB_CC1, JOY_X.
- `evolutionary --pop-size 12 --generations 8`: 11/13 — fails USB_CC1, JOY_X.

MC and GA both plateau at 11/13, so per the issue's diagnostic decision-tree this is **partly ordering** (MC's randomization beats negotiated by 1 net) **and partly geometric** (USB_CC1 and JOY_X are not reachable through pure reordering).  Option 1 was implemented as the issue requested as the smallest viable fix; the geometric piece (Option 2 / coupled escape) is left for a follow-up.

## Test Plan

- [x] New helper unit tests: `tests/test_router_negotiated_high_current.py::TestFindConnectorSiblingsOfPreroutedNets` — 7 cases covering empty / positive / cross-tier / unrelated-destination / default-priority filter / exclusion-of-prerouted / end-to-end ordering invariant.
- [x] New integration test: `tests/test_diffpair_routing_integration.py::test_diffpair_prepass_then_negotiated_routes_connector_sibling` — exercises the full pre-pass + negotiated flow on a fixture with one diff pair and one shared-connector single-ended net (all three nets must connect).
- [x] All existing diff-pair / npad / sibling-ripup tests still pass: `test_diffpair_routing_integration.py` (5 pass), `test_diffpair_npad.py` (10 pass), `test_router_negotiated_high_current.py` (23 pass + 7 new = 30).
- [x] Regression on board 05 (BLDC controller): SUCCESS, 10/10, DRC PASSED.
- [x] Regression on boards 01, 02, 03: completion rates unchanged (b01 2/3, b02 7/8, b03 10/13 are all baseline pre-fix numbers, confirmed by stashing the change and re-running).

## Limitations / Follow-up

- Board 03 end-to-end acceptance (13/13 with USB_CC1 and JOY_X) is **not** achieved by this change.  The diff-pair pre-pass on board 03 fails before claiming any cells, so the connector-sibling bump never fires there.  MC/GA also fail to reach 13/13, indicating the underlying issue is geometric (J1's pin field too dense for sequential escape).  This is the situation the issue describes as Option 2's territory — a coupled-escape entry point in `escape.py`.  A follow-up issue should be filed (or this one updated) for that work; per the curator's guidance, this PR keeps to Option 1 as the smallest viable fix.
- JOY_X is an analog signal that doesn't terminate at the USB-C connector, so even if the connector-sibling bump did fire it wouldn't help JOY_X.  JOY_X needs the destination-shared sibling logic extended beyond the diff-pair set, also Option 2 territory.

## Test plan checklist

- [x] `pytest tests/test_router_negotiated_high_current.py tests/test_diffpair_routing_integration.py tests/test_diffpair_npad.py -q --no-cov` — all green
- [x] `kicad-tools route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb ...` — SUCCESS 10/10, DRC PASSED
- [x] Boards 01, 02, 03: no regression vs baseline

### DIFF
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -1922,6 +1922,82 @@ class Autorouter:
                 refs.add(pad.ref)
         return refs
 
+    def _find_connector_siblings_of_prerouted_nets(
+        self,
+        prerouted_nets: set[int] | list[int],
+        candidate_nets: list[int] | set[int],
+    ) -> set[int]:
+        """Find candidate nets sharing a destination component with prerouted nets.
+
+        Issue #2482: When the differential-pair pre-pass routes USB_D+/USB_D-
+        before the negotiated loop, it claims grid cells in the destination
+        connector's pin field (e.g. J1).  Single-ended nets that also
+        terminate on the same connector (e.g. USB_CC1) are then routed in
+        plain priority order with no awareness that their escape corridor
+        has just been consumed by the diff-pair members, and they can
+        deadlock against the just-laid traces.
+
+        This helper identifies candidate nets that share at least one
+        destination component with any prerouted net.  These siblings need
+        to be routed *first within their tier* so they get a fair chance
+        to claim the remaining corridor space before lower-priority,
+        unrelated nets do.
+
+        Unlike :meth:`_find_same_tier_destination_siblings`, this helper:
+
+        - Does NOT require the candidate net to be in the same priority
+          tier as the prerouted net.  A diff-pair member may be HIGH_SPEED
+          (priority 2) while its connector sibling is DIGITAL (priority 4)
+          — both still need ordering coordination because they share the
+          same physical pin field.
+        - Returns siblings of the *whole* prerouted set, not of a single
+          failed net.  This is the natural shape for ordering-time use:
+          we have a fixed set of pre-pass routes and need to find which
+          remaining nets care about them.
+
+        The helper still applies the priority floor from
+        :meth:`_find_same_tier_destination_siblings` to avoid promoting
+        every random default-class net that happens to share a generic
+        component (e.g. the MCU U1, which most of a board's nets touch):
+        only candidates whose own class priority is < 10 are considered.
+
+        Args:
+            prerouted_nets: Set/list of net IDs that have already been
+                routed by a pre-pass (e.g. the diff-pair pre-pass).
+            candidate_nets: Iterable of net IDs to consider as siblings
+                (typically the nets remaining after the prerouted-set
+                filter in :meth:`route_all_negotiated`).
+
+        Returns:
+            Set of candidate net IDs whose pads touch at least one
+            component that some prerouted net's pads also touch, and
+            which carry a non-default net class (priority < 10).
+        """
+        prerouted_set = set(prerouted_nets)
+        if not prerouted_set:
+            return set()
+
+        # Collect destination components for the entire prerouted set.
+        prerouted_components: set[str] = set()
+        for net_id in prerouted_set:
+            prerouted_components |= self._get_net_destination_components(net_id)
+        if not prerouted_components:
+            return set()
+
+        siblings: set[int] = set()
+        for net_id in candidate_nets:
+            if net_id in prerouted_set:
+                continue
+            # Skip default/unclassified nets (priority 10) — they cover too
+            # many unrelated nets to use a shared-component heuristic.
+            if self._get_net_class_priority(net_id) >= 10:
+                continue
+            other_components = self._get_net_destination_components(net_id)
+            if other_components & prerouted_components:
+                siblings.add(net_id)
+
+        return siblings
+
     def _find_same_tier_destination_siblings(
         self,
         failed_net: int,
@@ -3201,6 +3277,43 @@ class Autorouter:
         self._detect_and_apply_matrix_preferences(net_order)
         # Re-sort after matrix priority boost
         net_order = sorted(net_order, key=lambda n: self._get_net_priority(n))
+
+        # Issue #2482: Bump connector-siblings of prerouted nets to the
+        # front of their priority tier.
+        #
+        # When the diff-pair pre-pass routes USB_D+/USB_D- the negotiated
+        # loop sees them via ``self.routes`` and skips them above.  But
+        # the pre-pass *also* claimed grid cells in the destination
+        # connector pin field.  Single-ended nets terminating at the same
+        # connector (e.g. USB_CC1 on the USB-C J1) must route *before*
+        # lower-priority unrelated nets in their own tier or they'll
+        # find their escape corridor already consumed.
+        #
+        # We use a stable sort keyed on
+        # ``(class_priority, is_connector_sibling_of_prerouted)`` so that
+        # within each priority tier, sibling nets sort first while the
+        # rest of the priority tuple (complexity, constraint, congestion)
+        # provides the secondary tiebreaker for nets in the same group.
+        if prerouted_nets and net_order:
+            connector_siblings = self._find_connector_siblings_of_prerouted_nets(
+                prerouted_nets, net_order
+            )
+            if connector_siblings:
+                flush_print(
+                    f"  Connector-siblings of prerouted nets bumped to "
+                    f"front of tier: {len(connector_siblings)} (Issue #2482)"
+                )
+
+                def _sibling_aware_priority(net_id: int) -> tuple:
+                    base = self._get_net_priority(net_id)
+                    # Inject a 0/1 flag right after the class priority so
+                    # ties within the same tier prefer connector siblings
+                    # (flag=0) over non-siblings (flag=1).  The remainder of
+                    # the original 6-tuple becomes the tertiary tiebreaker.
+                    flag = 0 if net_id in connector_siblings else 1
+                    return (base[0], flag) + base[1:]
+
+                net_order = sorted(net_order, key=_sibling_aware_priority)
 
         total_nets = len(net_order)
 
--- a/tests/test_diffpair_routing_integration.py
+++ b/tests/test_diffpair_routing_integration.py
@@ -206,3 +206,100 @@ def test_negotiated_skips_prerouted_diffpair_nets():
         assert post_count == pre_count, (
             f"Net {net_id} was re-routed: had {pre_count} routes, now {post_count}"
         )
+
+
+def _diffpair_with_connector_sibling_router() -> Autorouter:
+    """Build a fixture with a 2-pad diff pair plus a single-ended connector sibling.
+
+    Issue #2482: The diff-pair pre-pass routes USB_D+/USB_D- and reserves
+    grid cells in J1's pin field.  USB_CC1 terminates at the *same*
+    connector (J1) and must still be able to route after the pre-pass.
+
+    Layout:
+        U1 (left) <--> J1 (right)
+        - USB_D+   on U1:1, J1:1
+        - USB_D-   on U1:2, J1:2
+        - USB_CC1  on U1:3, J1:3
+
+    All three nets share the connector J1 so the connector-sibling
+    ordering bump (Issue #2482) must place USB_CC1 before any other
+    same-tier non-sibling net.
+    """
+    from kicad_tools.router.rules import NET_CLASS_HIGH_SPEED
+
+    rules = DesignRules(
+        trace_width=0.2,
+        trace_clearance=0.15,
+        grid_resolution=0.1,
+    )
+    net_class_map = {
+        "USB_D+": NET_CLASS_HIGH_SPEED,
+        "USB_D-": NET_CLASS_HIGH_SPEED,
+        "USB_CC1": NET_CLASS_HIGH_SPEED,
+    }
+    router = Autorouter(width=30.0, height=10.0, rules=rules,
+                       net_class_map=net_class_map)
+
+    # Diff pair pads
+    p_y = 5.0 - 0.4
+    n_y = 5.0 + 0.4
+    cc_y = 1.5  # USB_CC1 routes above the diff pair
+
+    router.add_component("U1", [
+        {"number": "1", "x": 5.0, "y": p_y, "width": 0.4, "height": 0.4,
+         "net": 1, "net_name": "USB_D+"},
+        {"number": "2", "x": 5.0, "y": n_y, "width": 0.4, "height": 0.4,
+         "net": 2, "net_name": "USB_D-"},
+        {"number": "3", "x": 5.0, "y": cc_y, "width": 0.4, "height": 0.4,
+         "net": 3, "net_name": "USB_CC1"},
+    ])
+    router.add_component("J1", [
+        {"number": "1", "x": 25.0, "y": p_y, "width": 0.4, "height": 0.4,
+         "net": 1, "net_name": "USB_D+"},
+        {"number": "2", "x": 25.0, "y": n_y, "width": 0.4, "height": 0.4,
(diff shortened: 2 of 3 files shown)


rjwalters:

Reviewed the connector-sibling priority bump (Issue #2482, Option 1). Implementation is clean and correct.

### What I verified

**Helper `_find_connector_siblings_of_prerouted_nets` (`core.py:1925`)**
- Iterates only over `prerouted_nets` to collect destination components, not over `self.routes` or all nets — correct scope.
- Excludes prerouted members from the candidate output (line 1989).
- Empty-prerouted-nets case returns `set()` early (lines 1976-1978); empty-prerouted-components case also returns early (1984-1985). No crash paths.
- Priority floor (`>= 10` filter, line 1993) prevents indiscriminate bumping of default/unclassified nets that happen to share a generic component (e.g. MCU U1).

**Bump in `route_all_negotiated` (`core.py:3281`)**
- Fires AFTER the prerouted-net filter (line 3273) and AFTER the matrix-preference re-sort (line 3279) — correct ordering of operations.
- Guarded by `if prerouted_nets and net_order:` so it never fires when there's nothing to bump.
- Uses `sorted()` (stable Timsort) with key `(base[0], flag) + base[1:]`.  Primary key is `base[0]` = class priority, so **cross-tier siblings cannot be promoted across tier boundaries**.  A DIGITAL/4 sibling stays behind all HIGH_SPEED/2 nets and only sorts to the front of *its own* tier — exactly the intended Option 1 semantics.
- `base[1:]` preserves the rest of the 6-tuple (complexity, constraint, pad-count, distance, congestion) as the tertiary tiebreaker, so within the sibling group and within the non-sibling group of each tier, the existing ordering is preserved verbatim.  Boards 01/02/04/05 should see no ordering change unless they actually have prerouted nets with same-tier siblings.

**Tests — all 7 helper tests + 1 integration test reviewed**
- `test_returns_empty_when_no_prerouted` — empty input safety. ✓
- `test_finds_sibling_sharing_connector` — happy path. ✓
- `test_excludes_prerouted_nets_themselves` — self-exclusion. ✓
- `test_excludes_default_priority_candidates` — priority floor critical case, leaves net unclassified (priority 10) and asserts `result == set()`. ✓
- `test_excludes_unrelated_destination` — note the test name is slightly misleading: it actually asserts JOY_X IS returned because U1 is a shared destination, documenting the conservative behavior in a comment. Behavior is intentional but the test name could trip future readers — minor nit, fine to address in a follow-up.
- `test_cross_tier_sibling_detected` — critical case: prerouted HIGH_SPEED, candidate DIGITAL, asserts cross-tier siblings ARE detected (so the sort step can bump within their own tier). ✓
- `test_negotiated_ordering_places_sibling_before_other_tier_members` — end-to-end ordering invariant simulating prerouted state. ✓
- Integration test gracefully skips if the pre-pass yields no routes (line 218-222), which is the right call since the bump only matters when a pre-pass actually populates `self.routes`.

**CI**
- Lint & Format: pass
- Test: pass
- Type Check: pass

**Conflict check**
- 493 insertions, 0 deletions — pure additive change.
- Helper is inserted directly above `_find_same_tier_destination_siblings` (#2475), with no overlap.
- The bump in `route_all_negotiated` sits between matrix-preference re-sort and `total_nets` assignment.  No overlap with #2479 (DRC nudge in chain connectivity) or #2476 (via-vs-via failure surfacing).

### Acknowledging the partial-fix scope

PR description and the curator are aligned: Option 1 alone CAN'T solve board 03 end-to-end on the current geometry.  MC and GA both plateau at 11/13, which the issue's diagnostic decision-tree confirms is "partly ordering, partly geometric."  The connector-sibling bump is the foundation; the geometric residual (USB_CC1, JOY_X) is left to follow-up #2490 (Option 2: coupled-escape) and the underlying issue that the diff-pair pre-pass currently fails outright on board 03's J1 pin field.  When #2490 lands and the pre-pass succeeds, the connector-sibling siblings will follow correctly because of the foundation laid here.  This is the right shape for a stacked landing.

### Verdict

Approved. Promoting to `loom:pr` for the merge queue.

— Loom Judge

PR #2502, opened 2026-05-04
fix(build-cpp): use uv venv interpreter and pass Python_EXECUTABLE

### LINKED ISSUE #2500: scripts/build-cpp.sh ignores uv venv, fails 'nanobind not installed' even after uv sync
## Problem
`scripts/build-cpp.sh` invokes bare `python3` for the nanobind detection step (and passes that path to cmake's `PYTHON_EXECUTABLE`). When the project is installed via `uv sync --extra dev` (the documented setup), nanobind sits in `.venv/lib/python3.13/site-packages/nanobind/` — not visible to bare `python3`. The script aborts with:
```
Error: nanobind not installed
Install with: pip install nanobind
```
…and the user has to remember to `PATH=.venv/bin:$PATH ./scripts/build-cpp.sh` to work around it.

## Reproduction
```bash
uv sync --extra dev       # nanobind installed in .venv
./scripts/build-cpp.sh    # fails 'nanobind not installed'
```

## Root cause / technical context

Two separate bugs interact here:

### Bug 1 — script ignores the project venv
`scripts/build-cpp.sh` calls bare `python3` in three places:
- `scripts/build-cpp.sh:41` — `check` action probing the installed module
- `scripts/build-cpp.sh:55` — nanobind preflight (`python3 -c "import nanobind"`)
- `scripts/build-cpp.sh:75` — `-DPYTHON_EXECUTABLE="$(which python3)"` passed to cmake

None of them prefer `$PROJECT_ROOT/.venv/bin/python3` even though the script already computes `PROJECT_ROOT` at line 20.

### Bug 2 — `-DPYTHON_EXECUTABLE=...` is silently ignored by the per-module CMakeLists
`src/kicad_tools/router/cpp/CMakeLists.txt:7-11` only honors a caller-supplied `PYTHON_EXECUTABLE` when `SKBUILD` is set:
```cmake
if(SKBUILD)
    set(Python_EXECUTABLE "${PYTHON_EXECUTABLE}")
    set(Python_INCLUDE_DIR "${PYTHON_INCLUDE_DIR}")
    set(Python_LIBRARY "${PYTHON_LIBRARY}")
endif()
...
find_package(Python COMPONENTS Interpreter ${DEV_MODULE} REQUIRED)
```
The manual `build-cpp.sh` path does not set `SKBUILD`, so the `-DPYTHON_EXECUTABLE=...` from line 75 of the script is dropped on the floor. `find_package(Python ...)` then resolves whatever `python3` is first on `PATH` — i.e. system python again, where nanobind is missing — and configuration fails inside the `execute_process` that runs `python -m nanobind --cmake_dir`.

The same bug exists in the placement and drc CMakeLists:
- `src/kicad_tools/placement/cpp/CMakeLists.txt:7-11`
- `src/kicad_tools/drc/cpp/CMakeLists.txt:7-11`

Even after Bug 1 is fixed, the proper fix for cmake to honor the chosen interpreter on the manual path is to use the variable cmake's `FindPython` module documents as input — `Python_EXECUTABLE` (no `PYTHON_` prefix) — and either pass that from the script or let the `if(SKBUILD)` block fall through to also accept it when set from the command line.

### scripts/build-cpp.sh
Pick the venv interpreter once at the top of the script and use it everywhere `python3` appears today:
```bash
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
VENV_PY="$PROJECT_ROOT/.venv/bin/python3"
if [ -x "$VENV_PY" ]; then
    PY="$VENV_PY"
else
    PY="$(command -v python3)"
fi
```
Then replace:
- line 41 (`python3 -c "..."`) → `"$PY" -c "..."`
- line 55 (`python3 -c "import nanobind"`) → `"$PY" -c "import nanobind"`
- line 75 (`-DPYTHON_EXECUTABLE="$(which python3)"`) → `-DPython_EXECUTABLE="$PY"` (note: `Python_` not `PYTHON_`, see below)

### CMakeLists (optional bonus, fixes the silent-ignore)
Either:
- (a) Have the script pass `-DPython_EXECUTABLE="$PY"` (the variable `find_package(Python …)` actually consumes), so the existing `if(SKBUILD)` branch is irrelevant for the manual path, or
- (b) Loosen the guard in each per-module `CMakeLists.txt` so it honors a caller-supplied `Python_EXECUTABLE` outside SKBUILD too:
  ```cmake
  if(NOT DEFINED Python_EXECUTABLE AND DEFINED PYTHON_EXECUTABLE)
      set(Python_EXECUTABLE "${PYTHON_EXECUTABLE}")
  endif()
  ```

(a) is the smaller, more localized fix and matches the issue's "bonus" suggestion.

## Acceptance criteria
- [ ] `./scripts/build-cpp.sh` succeeds in a fresh `uv sync --extra dev` checkout with no PATH munging or shell activation.
- [ ] Falls back to system `python3` when `.venv/bin/python3` does not exist, preserving the non-uv workflow.
- [ ] `./scripts/build-cpp.sh check` uses the same interpreter as the build step (no skew between probe and build).
- [ ] cmake's resolved `Python_EXECUTABLE` (visible in the `message(STATUS "Python_EXECUTABLE: ...")` line emitted by each per-module CMakeLists) matches `$PROJECT_ROOT/.venv/bin/python3` when invoked from a uv venv. Today it does not — the script's `-DPYTHON_EXECUTABLE` is silently dropped because `SKBUILD` is unset.
- [ ] `./scripts/build-cpp.sh clean` continues to work (no Python invocation involved — just sanity).
- [ ] Existing `pip install .[native]` / scikit-build path is unchanged (the `if(SKBUILD)` branch must keep working).

## Test plan
1. Fresh checkout, no system nanobind:
   ```bash
   uv sync --extra dev
   ./scripts/build-cpp.sh clean
   ./scripts/build-cpp.sh           # must succeed
   ./scripts/build-cpp.sh check     # must report Available: True
   ```
2. Verify cmake picked the venv interpreter — grep configure output for the `Python_EXECUTABLE:` status line and confirm it points into `.venv/`.
3. Non-venv fallback: in a directory with no `.venv`, with system python3 + nanobind present, the script still builds.
4. scikit-build path still works:
   ```bash
   uv pip install '.[native]'  # exercises CMakeLists via scikit-build-core, SKBUILD=1
   ```
5. `clean` still removes `build/` and `router_cpp*.so`/`*.pyd` artifacts.

## Files
- `scripts/build-cpp.sh` — primary change (~5 lines added, 3 sites updated)
- `src/kicad_tools/router/cpp/CMakeLists.txt` — only if option (b) is chosen
- `src/kicad_tools/placement/cpp/CMakeLists.txt` — only if option (b) is chosen
- `src/kicad_tools/drc/cpp/CMakeLists.txt` — only if option (b) is chosen (note: not currently descended into from the root `CMakeLists.txt`, but should still stay consistent)

## Out of scope
- Adding the same `.venv` autodetect to other helper scripts (only one shell script under `scripts/` today, but worth flagging if more are added).
- Refactoring the three per-module CMakeLists to share a common Python-discovery helper.
- Wiring `src/kicad_tools/drc/cpp` into the root `CMakeLists.txt` — it currently only descends into router and placement; tracked separately if needed.

rjwalters:
## Summary
`scripts/build-cpp.sh` invoked bare `python3` for the nanobind preflight, the `check` command, and `-DPYTHON_EXECUTABLE=...` passed to cmake. When the project is installed via `uv sync --extra dev`, nanobind lives in `.venv/lib/...` and is invisible to system `python3`, so the script aborted with "nanobind not installed" even after a clean install. Compounding this, the per-module `CMakeLists.txt` only honors caller-supplied `PYTHON_EXECUTABLE` under `if(SKBUILD)` — on the manual path it was silently dropped and `find_package(Python ...)` fell back to whatever `python3` was first on PATH.

This PR adopts the curator's recommended **option (a)** fix in `scripts/build-cpp.sh` only: detect `.venv/bin/python3` once, use it everywhere, and pass cmake the lowercase `Python_EXECUTABLE` variable (the one its FindPython module actually consumes). No CMakeLists are changed, so the existing `pip install .[native]` / scikit-build path is unaffected.

## Changes
- `scripts/build-cpp.sh`: pick `$PROJECT_ROOT/.venv/bin/python3` if executable, otherwise fall back to `command -v python3`.
- Replace bare `python3` invocations in the `check` action and the nanobind preflight with the resolved `"$PY"`.
- Switch the cmake configure flag from `-DPYTHON_EXECUTABLE="$(which python3)"` to `-DPython_EXECUTABLE="$PY"` so the chosen interpreter actually survives `find_package(Python ...)` on the manual (non-SKBUILD) path. Inline comment documents why the lowercase form is required.

| `./scripts/build-cpp.sh` succeeds in a fresh `uv sync --extra dev` checkout with no PATH munging | Yes | Ran `clean` then `build` against the project venv; full build, install, and self-verify (`Available: True`) completed end-to-end |
| Falls back to system `python3` when `.venv/bin/python3` does not exist | Yes | Worktree without a `.venv` symlink resolves `PY` from `command -v python3`; `bash -n` clean and `check` runs through to the import (the only failure mode is the system python missing the `kicad_tools` package, which is the expected non-uv user environment) |
| `./scripts/build-cpp.sh check` uses the same interpreter as the build step | Yes | Both branches use the single `$PY` resolved at the top of the script |
| cmake's resolved `Python_EXECUTABLE` matches `$PROJECT_ROOT/.venv/bin/python3` when invoked from a uv venv | Yes | Configure log: `-- Python_EXECUTABLE: /Users/rwalters/GitHub/kicad-tools/.loom/worktrees/issue-2500/.venv/bin/python3` (matches the venv) |
| `./scripts/build-cpp.sh clean` continues to work | Yes | Ran `clean` twice during validation; removes `build/` and `router_cpp*.so` |
| Existing `pip install .[native]` / scikit-build path unchanged | Yes | No CMakeLists modified; `if(SKBUILD)` branch in the per-module CMakeLists is untouched |

## Test Plan
- `bash -n scripts/build-cpp.sh` — syntax check
- `shellcheck scripts/build-cpp.sh` — lint clean
- `./scripts/build-cpp.sh clean` — removes prior artifacts
- `./scripts/build-cpp.sh build` — configure log shows `Python_EXECUTABLE: .../.venv/bin/python3`, both `placement_cpp` and `router_cpp` link, module is copied into `src/kicad_tools/router/`
- `./scripts/build-cpp.sh check` — reports `Backend: cpp`, `Available: True`, `Version: 1.0.0`

Closes #2500

### DIFF
--- a/scripts/build-cpp.sh
+++ b/scripts/build-cpp.sh
@@ -21,6 +21,17 @@ PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
 BUILD_DIR="$PROJECT_ROOT/build"
 ROUTER_DIR="$PROJECT_ROOT/src/kicad_tools/router"
 
+# Prefer the project's uv-managed virtualenv interpreter when present so that
+# nanobind (installed via `uv sync --extra dev`) is visible to both the
+# preflight import check and the cmake configure step. Fall back to the first
+# `python3` on PATH so non-uv workflows continue to work.
+VENV_PY="$PROJECT_ROOT/.venv/bin/python3"
+if [ -x "$VENV_PY" ]; then
+    PY="$VENV_PY"
+else
+    PY="$(command -v python3)"
+fi
+
 # Colors for output
 RED='\033[0;31m'
 GREEN='\033[0;32m'
@@ -38,7 +49,7 @@ case "${1:-build}" in
 
     check)
         echo "Checking C++ router backend availability..."
-        python3 -c "
+        "$PY" -c "
 from kicad_tools.router.cpp_backend import is_cpp_available, get_backend_info
 info = get_backend_info()
 print(f'Backend: {info[\"backend\"]}')
@@ -52,7 +63,7 @@ if info['available']:
         echo -e "${YELLOW}Building C++ router extension...${NC}"
 
         # Check for nanobind
-        if ! python3 -c "import nanobind" 2>/dev/null; then
+        if ! "$PY" -c "import nanobind" 2>/dev/null; then
             echo -e "${RED}Error: nanobind not installed${NC}"
             echo "Install with: pip install nanobind"
             exit 1
@@ -69,10 +80,16 @@ if info['available']:
         mkdir -p "$BUILD_DIR"
 
         # Configure with CMake
+        # NOTE: pass `-DPython_EXECUTABLE` (the variable cmake's FindPython
+        # module actually consumes), not `-DPYTHON_EXECUTABLE`. The per-module
+        # CMakeLists only forwards `PYTHON_EXECUTABLE` -> `Python_EXECUTABLE`
+        # under `if(SKBUILD)`, which is unset on this manual path, so the
+        # uppercase form is silently dropped and find_package(Python ...)
+        # falls back to whatever python3 is first on PATH.
         echo "Configuring..."
         cmake -B "$BUILD_DIR" -S "$PROJECT_ROOT" \
             -DCMAKE_BUILD_TYPE=Release \
-            -DPYTHON_EXECUTABLE="$(which python3)"
+            -DPython_EXECUTABLE="$PY"
 
         # Build
         echo "Building..."


rjwalters:
## Judge Re-Review (post-approval audit)

Re-reviewed at user request. Confirming the existing `loom:pr` approval stands.

### Verification
- **CI**: all three checks (`Lint & Format`, `Type Check`, `Test`) PASSED on commit 8e82b106
- **Merge state**: CLEAN
- **Diff scope**: single file (`scripts/build-cpp.sh`, +20/-3) — matches curator's option (a) exactly

### Correctness checks
1. **`Python_EXECUTABLE` vs `PYTHON_EXECUTABLE`** — verified against `src/kicad_tools/router/cpp/CMakeLists.txt` on `main`. The `if(SKBUILD)` block only forwards `PYTHON_EXECUTABLE` -> `Python_EXECUTABLE` under scikit-build; on the manual path `find_package(Python ...)` consumes only the lowercase-prefix form. The script's switch to `-DPython_EXECUTABLE="$PY"` is the minimum-change fix that actually takes effect without touching CMakeLists.

2. **Single-source `$PY` resolution** — both `check` and `build` actions use the same resolved interpreter, satisfying the "no skew between probe and build" acceptance criterion.

3. **Fallback** — when `.venv/bin/python3` is absent, `command -v python3` keeps non-uv workflows working. Good (also a small upgrade from the original `which python3`).

4. **SKBUILD path untouched** — no `CMakeLists.txt` modified, so `pip install .[native]` / scikit-build-core continues to work via the existing `if(SKBUILD)` branch.

5. **Inline NOTE comment** — the forensic comment on lines 83–88 explaining why the lowercase form is required is exactly the kind of regression-protection comment that should accompany a non-obvious variable-name fix. Recommended pattern.

### Adherence to curator
Implements option (a) verbatim (the curator's preferred minimal fix). Out-of-scope items (other helper scripts, per-module CMakeLists refactor, drc wiring) are correctly deferred.

### Decision
Approval maintained — `loom:pr` retained, `loom:review-requested` not present. Ready for human merge.

PR #2505, opened 2026-05-04
feat(build): smoke-check emitted PCB after each write step

### LINKED ISSUE #2495: build pipeline: fail fast when emitted PCB is unloadable by kicad-cli
## Problem
Today's broken-export bugs (kct_marking + generator_version) showed the same symptom on every demo board: the build pipeline ran 7-8 steps successfully (route, ERC, DRC, sync), produced a routed PCB, then failed at the manufacturing-export step with `kicad-cli failed (exit code 3): Failed to load board`. There was no earlier signal attributing the rejection to the writer.

The pipeline trusted its own output enough to run pure-python DRC on it, but never asked kicad-cli "can you actually load this?"

## Reproduction (pre-fix)
```bash
kct build boards/00-simple-led --force --verbose
```
6 steps green, then the export step fails with the kicad-cli rejection — no indication of which writer produced the bad output.

## Proposal
Insert a `kicad-cli pcb drc --output /tmp/precheck.rpt <pcb>` smoke check immediately after each PCB write step (PCB generation, routing). If kicad-cli fails to load:
- Stop the pipeline
- Attribute the failure to the writer that just ran
- Print the kicad-cli stderr verbatim
- Suggest `head -10 <pcb>` to inspect

Skip when kicad-cli is unavailable (warn).

## Acceptance criteria
- [ ] After PCB write, build runs `kicad-cli pcb drc` (lightest load-check available).
- [ ] On failure, error message names the producing step (`Output of <step> rejected by kicad-cli`).
- [ ] Skipped with single warning line when kicad-cli not installed.

rjwalters:
## Summary

- Add `_smoke_check_pcb` helper to `cli/build_cmd.py` that runs `kicad-cli pcb drc --schematic-parity off` against the current PCB after each PCB-write step (PCB, OUTLINE, PLACEMENT, ZONES, SILKSCREEN, ROUTE).
- Discriminate "Failed to load board" parse rejections (which we want to surface) from real DRC rule violations (which we explicitly do not want to misattribute to the writer).
- Wire the helper into `main()` after the existing per-step success/failure print and early-break, gated on `result.success`, `ctx.smoke_check`, and not `--dry-run`.
- Emit a single warning when `find_kicad_cli()` returns None and skip subsequent checks for the rest of the build.
- Add a `--no-smoke-check` opt-out flag for users who want the prior behaviour.
- Add `tests/test_build_smoke_check.py` with 9 tests covering the happy path, load-failure attribution, DRC-violation discrimination, the missing-kicad-cli warn-once contract, dry-run skip, opt-out flag, and pipeline halt-on-failure.

Closes #2495

## Test plan

- [x] `uv run pytest tests/test_build_smoke_check.py --no-cov` (9/9 pass)
- [x] `uv run pytest tests/test_build_*.py --no-cov` (116/117 pass; the one failing test `test_route_exit_code_3_is_failure` is pre-existing on main and unrelated to this change)
- [x] `uv run ruff check tests/test_build_smoke_check.py` clean
- [x] `uv run ruff format src/kicad_tools/cli/build_cmd.py tests/test_build_smoke_check.py --check` clean (the only ruff error in build_cmd.py is the pre-existing unused `math` import on line 17)
- [x] `kct build --help` shows `--no-smoke-check` flag
- [x] `from kicad_tools.cli import build_cmd` imports cleanly with `_smoke_check_pcb` and `_PCB_WRITE_STEPS` exported

## Implementation notes

- Followed the curator's recommended Option A (single guard wrapper) - keeps the per-step `_run_step_*` helpers untouched and adds one dispatch site in `main()`.
- Reuses existing `find_kicad_cli` and `run_drc` from `cli/runner.py` - no new subprocess plumbing.
- Uses a temp DRC report file (`tempfile.mkstemp`) cleaned up in a `finally` so we don't pollute `output/`.
- The "kicad-cli not installed" warning is gated by a `_kicad_cli_warning_emitted` flag on `BuildContext` so it prints exactly once per build.
- Failure messages include `head -10 <pcb_path>` per the issue's acceptance criteria so the user has an immediate next step.

Generated with Claude Code

### DIFF
--- a/src/kicad_tools/cli/build_cmd.py
+++ b/src/kicad_tools/cli/build_cmd.py
@@ -77,7 +77,9 @@ class BuildContext:
     quiet: bool = False
     force: bool = False
     optimize_placement: bool = False
+    smoke_check: bool = True
     _executed_scripts: set[Path] | None = None
+    _kicad_cli_warning_emitted: bool = False
 
     def mark_script_executed(self, script: Path) -> None:
         """Mark a script as having been executed."""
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge re-review (loom:approved confirmation)

Re-reviewed per request. PR state remains valid for merge.

### Verification
- CI: Lint & Format / Type Check / Test all green (run 25316935409).
- mergeStateStatus: CLEAN.
- 9/9 smoke-check tests pass locally (`uv run pytest tests/test_build_smoke_check.py --no-cov` — 0.25s).
- `--no-smoke-check` flag is wired into argparse and visible in `kct build --help`.
- Pre-existing `test_route_exit_code_3_is_failure` failure confirmed to also fail on `origin/main` — unrelated to this change.

### Smoke-check helper assessment (`_smoke_check_pcb`)
- Correctly delegates to `find_kicad_cli()` + `run_drc(..., schematic_parity=False, ...)`. Reuses existing infrastructure cleanly.
- Temp report file via `tempfile.mkstemp` cleaned up with `contextlib.suppress(OSError)` in a `finally` — no `output/` pollution; safe even if cleanup races.
- `BuildContext._kicad_cli_warning_emitted` flag correctly gates the "kicad-cli not installed" warning to once per build.
- Clean signature returning `BuildResult | None`, exactly the curator's recommended Option A.

### Stderr discrimination (the load-failure / DRC-violation distinction)
- The combined-stream check (`combined = (stderr or "") + "\n" + (stdout or "")`) correctly looks in *both* streams for the `"Failed to load board"` marker, which addresses the curator's note that kicad-cli sometimes writes the marker to stdout.
- `test_ignores_drc_violations` proves a non-zero exit (return_code=5) with no marker is correctly treated as success — the writer is not falsely blamed.
- `test_returns_failure_on_load_error` proves return_code=3 + `"Failed to load board"` stderr produces the attributed `BuildResult(step="silkscreen", success=False, ...)`. Message format includes the producing-step name, the verbatim stderr/stdout, and the `head -10 <pcb_path>` hint required by the issue's acceptance criteria.

### `--no-smoke-check` opt-out
- Flag definition reads cleanly; default behaviour is smoke-check enabled (`smoke_check=not args.no_smoke_check`). This matches the curator's "default True, opt-out" guidance.
- `test_no_smoke_check_flag_disables_helper` confirms the helper is not invoked when the flag is passed.

### Pipeline integration
- Insertion site (after the existing `if not result.success` early-break, gated on `result.success and ctx.smoke_check and not ctx.dry_run and step in _PCB_WRITE_STEPS`) is exactly the curator's recommended location.
- `_PCB_WRITE_STEPS` set correctly excludes SCHEMATIC, VERIFY, and EXPORT and includes the six PCB-emitting steps.
- On smoke-check failure: the failed `BuildResult` is appended to `results`, a red `[FAIL]` line is printed, and the loop breaks before later steps run — so attribution lands on the last writer that ran. `test_pipeline_breaks_on_smoke_check_failure` confirms exit code 1.
- `test_smoke_check_skipped_on_dry_run` confirms `--dry-run` short-circuits the helper.

- The use of underscore-prefixed `BuildContext._kicad_cli_warning_emitted` is read/written from outside the dataclass. It's a private-by-convention attribute used as a sentinel; this is fine, though a public name (or a small `mark_warning_emitted()` method) would be slightly more idiomatic. Not worth blocking on.
- The lazy `import contextlib, os, tempfile` inside `_smoke_check_pcb` is reasonable to keep cold-path imports out of module load. Consistent with the existing pattern in `runner.py` (`import os` lazily inside `run_drc`).
- A future enhancement could surface `result.return_code` in the failure message (right now only stderr/stdout text is included), but the current message is already actionable and matches the issue's acceptance criteria.

### Conclusion
loom:approved. Label `loom:pr` is already applied from the prior judge pass; no label change required. Ready for human merge.

PR #2506, opened 2026-05-04
fix(router): exclude skipped pour/CC nets from routing summary failures

rjwalters:
Closes #2498

## Summary

- Skipped pour nets (VBUS/VCC/GND) and named-skip nets (USB_CC1/USB_CC2) were appearing under "Unrouted nets:" with the generic "No path found" message on boards 02 and 03.
- Root cause: the demo route scripts called `show_routing_summary` without `nets_to_route_ids`, and the fallback branch iterated `net_map` (which retains skipped nets for diagnostics) instead of `router.nets` (which reflects post-skip state after `load_pcb_for_routing` rewrites those pads to net=0).
- Fix per curator analysis: (a) pass `nets_to_route_ids=multi_pad_net_ids` from both `route_demo.py` scripts, and (b) make the `None`-filter fallback in `show_routing_summary` / `get_routing_diagnostics_json` use `router.nets` restricted to multi-pad nets (mirroring the convention used by `cli/route_cmd.py`).

## Files changed

- `src/kicad_tools/router/output.py` — fallback now uses `router.nets` (multi-pad only), with backward-compat fallback to `net_map` when `router.nets` isn't a real dict (e.g. legacy MagicMock callers).
- `boards/03-usb-joystick/route_demo.py` — builds `multi_pad_net_ids` and threads it through `show_routing_summary`.
- `boards/02-charlieplex-led/route_demo.py` — same pattern.
- `tests/test_routing_output.py` — new `TestIssue2498SkipNetFallback` class with 4 regression tests.

## Test plan

- [x] `uv run pytest tests/test_routing_output.py tests/test_output_connectivity.py tests/test_reroute_progress_output.py` — 50 passed
- [x] `uv run pytest tests/test_route_cmd_params.py tests/test_route_cmd_power_nets.py` — 57 passed
- [x] `uv run ruff check` and `uv run ruff format --check` on touched files — all clean
- [x] `uv run --with mypy mypy --ignore-missing-imports src/kicad_tools/router/output.py` — no new errors (3 pre-existing unrelated to this change)
- [ ] Smoke test on hardware: `kct build boards/03-usb-joystick --force --verbose` should no longer print VBUS/VCC/GND/USB_CC1/USB_CC2 under "Unrouted nets:" (JOY_X may still show — separate concern, out of scope per curator).

## Notes

- The legacy MagicMock-based test in `test_legacy_no_filter_shows_all_unrouted` still passes because the fallback safely detects when `router.nets` isn't a real dict.
[...]

PR #2504, opened 2026-05-04
feat(silkscreen): track marking identity in *.kct.json sidecar

rjwalters:
## Summary

Replace the fragile visible-text idempotency heuristic in `SilkscreenGenerator.add_board_markings` with a sibling `*.kct.json` sidecar keyed by `gr_text` UUID. PCB content stays strictly KiCad-conformant (no `kct_*` custom S-expression children); marking identity now survives project renames, revision bumps, and accidental visible-text collisions with user-edited `gr_text` nodes.

- New sidecar `<pcb>.kct.json` is loaded on construction and atomically rewritten on `save()`.
- Rename / revision bump removes the stale `gr_text` and writes a fresh one with a new UUID.
- User-added `gr_text` whose visible string matches a marking value is no longer mistaken for the generator's marking.
- Missing sidecar is silent; corrupt sidecar logs a warning and proceeds with an empty registry.
- Removed dead `_MARKING_PREFIX = "kct:"` module-level constant.

The original kicad-cli-loadability blocker was already fixed in 30256d40; this PR closes the structural follow-up flagged in the issue.

## Test plan

- [x] `uv run pytest tests/test_silkscreen_generator.py` — 27 passed
- [x] `uv run pytest tests/ -k silkscreen` — 115 passed, 5 skipped
- [x] `uv run ruff check src/kicad_tools/silkscreen/generator.py tests/test_silkscreen_generator.py` — only pre-existing errors remain
- [x] `uv run mypy --ignore-missing-imports src/kicad_tools/silkscreen/generator.py` — clean
- [x] New sidecar tests cover: revision-bump dedup, project-rename dedup, sidecar survives save/reload, missing sidecar graceful, corrupt sidecar graceful, user-edited colliding gr_text not skipped, no kct_* subfields, registry survives gr_text deletion.
- [x] Dead-constant regression test asserts `_MARKING_PREFIX` is gone.

Closes #2494

Generated with Claude Code

PR #2503, opened 2026-05-04
fix(build): pass manufacturer edge_clearance to auto-pour zone step

rjwalters:
## Summary

Fixes the `kct build` zones step so that zone copper is inset from the
board outline by the manufacturer's minimum edge clearance, eliminating
`edge_clearance_zone` DRC violations on JLCPCB-targeted boards (e.g.
board 01).

## Changes

- `src/kicad_tools/cli/build_cmd.py` — In `_run_step_zones`, look up
  `MfrLimits.min_edge_clearance` from `ctx.mfr` and forward it to
  `auto_create_zones_for_pour_nets(..., edge_clearance=...)`. Mirrors the
  auto-fill pattern at `route_cmd.py:3300-3313`. Falls back to `None` for
  unknown manufacturers (preserving prior behaviour).
- `tests/test_build_zones_step.py` — New `TestRunStepZonesEdgeClearance`
  class with three regression tests:
  - `test_jlcpcb_inset_to_min_edge_clearance` — every zone vertex is at
    least 0.3 mm from the outline.
  - `test_oshpark_uses_larger_clearance` — confirms the value is plumbed
    through (0.381 mm), not hard-coded.
  - `test_unknown_manufacturer_falls_back_to_no_inset` — unknown mfr does
    not crash and zones still get created.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| `kct build boards/01-voltage-divider --force` produces 0 `edge_clearance_zone` errors | Verified by unit test | `test_jlcpcb_inset_to_min_edge_clearance` asserts every zone vertex is >= 0.3 mm from the outline; without the fix the test fails (vertex at 0.0). |
| Zone polygon vertices inset from outline by the manufacturer's clearance | Verified by unit test | Two parametrised tests cover jlcpcb (0.3) and oshpark (0.381). |
| New regression test in `tests/test_build_zones_step.py` | Done | `TestRunStepZonesEdgeClearance` class with 3 tests. |
| Test loads PCB via `PCB.load` and asserts zone vertex distance from outline bounding box | Done | `_assert_inset` helper compares each vertex to the outline bbox with epsilon. |
| Test parametrised over multiple manufacturers (jlcpcb, oshpark) | Done | Two separate test methods, one per mfr. |

## Test Plan

- [x] `uv run pytest tests/test_build_zones_step.py --no-cov` -> 17 passed
[...]

PR #2507, opened 2026-05-04
test(ci): add kicad-cli round-trip smoke test on every emitted PCB

### LINKED ISSUE #2492: ci: add kicad-cli round-trip smoke test on every emitted PCB
## Problem
Two latent bugs that broke every demo board's manufacturing export went undetected for an unknown amount of time because no test in the suite ever invokes `kicad-cli pcb drc` (or any other kicad-cli pcb subcommand) on a kicad-tools-emitted PCB:

1. `(generator_version "8.0")` — kicad-cli 10.x rejects load (every PCB)
2. `(kct_marking ...)` non-KiCad subfield embedded in `gr_text` — kicad-cli rejects load (every PCB with silkscreen markings)

Both produced identical `Failed to load board` on `kicad-cli pcb export gerbers`, exit 3.

## Reproduction
Before today's fixes, every `kct build boards/<NN>` failed at the `export` step with `kicad-cli failed (exit code 3): Failed to load board`. There was no test that would have caught this in CI.

## Expected vs actual
**Expected:** any PCB the toolchain writes is loadable by the installed `kicad-cli`. CI fails fast if not.
**Actual:** the only signal was the manufacturing-export step at runtime on real boards.

## Acceptance criteria
- [ ] A pytest fixture / integration test that runs `kicad-cli pcb drc --output /tmp/junk.rpt <pcb>` on each demo board's emitted `*.kicad_pcb` (or a small in-suite fixture).
- [ ] Test fails with a clear message identifying the writer if kicad-cli refuses the file.
- [ ] Skipped gracefully when `kicad-cli` is not installed; warning emitted.

## Impact
Without this gate, broken PCB emission can silently re-land in main. Both fixes today were diagnosed by hand-bisecting the emitted file.

rjwalters:
Closes #2492.

## Summary

- Adds `tests/test_kicad_cli_roundtrip.py` (7 tests) — a regression gate that emits a PCB via kicad-tools' writers (`PCB.create`/`PCB.save`, `gr_text_node`, `zone_node`, `SilkscreenGenerator.add_board_markings`, plus a `parse_string` + `to_string` footprint round-trip) and verifies kicad-cli accepts the file via `run_drc()`. Closes the bug class where unloadable PCBs (commit `30256d40`: unquoted `generator_version`, embedded `kct_marking`) reach `main` undetected.
- Adds a new `kicad-cli-smoke` job to `.github/workflows/ci.yml` that `apt-get install`s kicad on `ubuntu-latest`. Kept separate from the main test job so the apt install does not slow down the broader test lane.

Implementation follows the curator's recommended Option A: in-suite library-level emitter test (no demo-board generation, no external footprint libs needed). Skips with a single warning when `find_kicad_cli()` returns `None`, via module-level `pytestmark`.

## CI design choice

Per the issue's acceptance criteria the PR must "document the choice" — I chose a **separate `kicad-cli-smoke` job** rather than adding `apt-get install kicad` to the existing test job. Rationale:

- The smoke surface is small (~20 s of test runtime). The dominant cost is the apt install (~2-3 min).
- Adding it to the main test job would slow every PR by 2-3 min for a check that runs in a tightly-scoped 7-test file.
- A separate job lets the smoke run in parallel and is independently re-runnable when CI flakes.
- A failure of this job is unambiguously about emitted-PCB validity, not about the broader suite.

## Acceptance-criteria verification (manual)

All four acceptance verifications from the issue's test plan were performed locally:

| Step | Result |
|---|---|
| `uv run pytest tests/test_kicad_cli_roundtrip.py -v` on current `main` post-fix | All 7 pass |
| Temporarily comment out `"generator_version"` from `_FORCE_QUOTED_STRING_VALUE` | 6 of 7 tests fail with attributed error: producer name, PCB path, return code, and verbatim kicad-cli stderr (`Failed to load board`). Restored. |
| Temporarily reintroduce `(kct_marking ...)` subfield in `SilkscreenGenerator` | `TestSilkscreenGeneratorRoundtrip` fails with the pre-kicad-cli regression-guard message that names `src/kicad_tools/silkscreen/generator.py`. Restored. |
| Skip behavior when kicad-cli not on PATH | Implemented via `pytestmark = pytest.mark.skipif(find_kicad_cli() is None, ...)`, matching `tests/test_pcb_import_from_schematic.py:296`. |

Example of the attributed error message format produced by `_format_load_failure`:

```
kicad-cli rejected PCB emitted by kicad-tools writer.
  Producer: PCB.save (blank)
  PCB: /tmp/.../blank.kicad_pcb
  kicad-cli return code: 3
  kicad-cli stderr:
    Failed to load board

Inspect the emitted file: head -20 /tmp/.../blank.kicad_pcb
```

## Test plan

- [x] All 7 new tests pass locally with kicad-cli 10.x installed (~20 s).
- [x] Failure mode produces clear, attributed error messages naming the producing component and printing kicad-cli stderr.
- [x] Verified that reverting either of the two recent fixes (`generator_version` quoting and `kct_marking` removal) makes the test fail.
- [x] Verified the test cleans up tmp PCB output via `tmp_path` fixture (no manual cleanup).
- [x] CI yaml validates as parseable yaml.
- [ ] CI run on this PR exercises the new `kicad-cli-smoke` job against a runner with kicad-cli installed (will be observable on the PR's checks tab).

## Notes for follow-ups

- #2495 (runtime build pipeline gate) can reuse the load-check primitive — `run_drc(pcb_path)` returning `success=True` iff the file loads.
- A thinner `kicad_cli_load_check(pcb_path) -> KiCadCLIResult` helper (suggested by the curator) was deliberately not added in this PR to minimize scope; the existing `run_drc` already provides this contract via `result.success`. If #2495 finds it ergonomic to extract, that's a small follow-up.
- Option B (smoke test on checked-in board outputs) and Option C (full demo-board generation) were left for follow-ups per the curator's recommendation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,3 +149,39 @@ jobs:
       - name: Smoke test C++ router
         run: |
           uv run pytest tests/test_router_cpp_backend.py::TestCppBuildVersionGuard -v --no-cov
+
+  kicad-cli-smoke:
+    name: kicad-cli Round-trip Smoke
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    # Separate job: a regression gate that PCB files emitted by kicad-tools'
+    # writers are loadable by kicad-cli. Two writer bugs (unquoted
+    # generator_version, embedded kct_marking subfield) shipped to main and
+    # broke every demo board's manufacturing export — this gate prevents
+    # that class of regression from re-landing.
+    #
+    # Runs as its own job so the apt install of KiCad does not slow down
+    # the main test job. The smoke surface itself is small (<30 s of test
+    # runtime); the dominant cost is the kicad apt install.
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install KiCad (kicad-cli)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends kicad
+          kicad-cli --version
+
+      - uses: astral-sh/setup-uv@v4
+        with:
+          version: "latest"
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: uv sync --extra dev
+
+      - name: Run kicad-cli round-trip smoke tests
+        run: uv run pytest tests/test_kicad_cli_roundtrip.py -v --no-cov
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,8 +166,16 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Install KiCad (kicad-cli)
+      - name: Install KiCad 9 (kicad-cli)
+        # ubuntu-latest (24.04 noble) ships KiCad 7.0.11 in the default apt
+        # repos, which rejects PCBs emitted by kicad-tools writers because we
+        # emit `(version 20240108)` (KiCad 8 file format epoch) and
+        # `(generator_version "9.0")` (KiCad 9). Installing from the KiCad 9
+        # PPA gets us a kicad-cli that matches the writer's emitted format.
+        # If the writer ever bumps to a newer major version, bump the PPA in
+        # lockstep.
         run: |
+          sudo add-apt-repository --yes ppa:kicad/kicad-9.0-releases
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends kicad
           kicad-cli --version
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,33 +160,48 @@ jobs:
     # broke every demo board's manufacturing export — this gate prevents
     # that class of regression from re-landing.
     #
-    # Runs as its own job so the apt install of KiCad does not slow down
-    # the main test job. The smoke surface itself is small (<30 s of test
-    # runtime); the dominant cost is the kicad apt install.
+    # Uses the official kicad/kicad:9.0 Docker image rather than installing
+    # KiCad via apt+PPA. Earlier attempts to install from
+    # ppa:kicad/kicad-9.0-releases on ubuntu-latest were unreliable: both
+    # api.launchpad.net and ppa.launchpadcontent.net failed back-to-back from
+    # the Azure GitHub runner, and apt then silently fell back to KiCad 7
+    # from noble/universe — producing confusing "DRC produced no output"
+    # errors instead of an attributable infrastructure failure. The official
+    # container ships kicad-cli 9 already and is immune to Launchpad outages.
+    container:
+      image: kicad/kicad:9.0
+      # Run as root so uv/pip installs go to the system site-packages and
+      # actions/checkout can write into $GITHUB_WORKSPACE without permission
+      # juggling.
+      options: --user 0:0
     steps:
+      # The kicad/kicad image is Ubuntu-based and ships Python 3 + git, but
+      # we install ca-certificates and curl defensively in case the image
+      # variant we land on is missing one of them (uv's installer needs curl,
+      # and HTTPS pulls need ca-certificates).
+      - name: Ensure container has prerequisites
+        run: |
+          set -euo pipefail
+          apt-get update
+          apt-get install -y --no-install-recommends ca-certificates curl git
+          # Sanity: the container should have kicad-cli 9 already.
+          kicad-cli --version | tee /tmp/kicad-cli-version
+          grep -qE '^9\.' /tmp/kicad-cli-version || {
+            echo "::error::kicad/kicad:9.0 container does not provide kicad-cli 9.x"
+            exit 1
+          }
+
       - uses: actions/checkout@v4
 
-      - name: Install KiCad 9 (kicad-cli)
-        # ubuntu-latest (24.04 noble) ships KiCad 7.0.11 in the default apt
-        # repos, which rejects PCBs emitted by kicad-tools writers because we
-        # emit `(version 20240108)` (KiCad 8 file format epoch) and
-        # `(generator_version "9.0")` (KiCad 9). Installing from the KiCad 9
-        # PPA gets us a kicad-cli that matches the writer's emitted format.
-        # If the writer ever bumps to a newer major version, bump the PPA in
-        # lockstep.
+      - name: Install uv
+        # setup-uv@v4 should work inside the container, but installing via
+        # the official one-liner avoids any action-host coupling and keeps
+        # the install deterministic across image refreshes.
         run: |
-          sudo add-apt-repository --yes ppa:kicad/kicad-9.0-releases
-          sudo apt-get update
-          sudo apt-get install -y --no-install-recommends kicad
-          kicad-cli --version
-
-      - uses: astral-sh/setup-uv@v4
-        with:
-          version: "latest"
-
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.12"
+          set -euo pipefail
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          # uv installs to ~/.local/bin by default; surface it for later steps.
+          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
 
       - name: Install dependencies
         run: uv sync --extra dev
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,6 +174,15 @@ jobs:
       # actions/checkout can write into $GITHUB_WORKSPACE without permission
       # juggling.
       options: --user 0:0
+    defaults:
+      run:
+        # The kicad/kicad:9.0 image links /bin/sh to dash, which does not
+        # support `set -o pipefail`. GitHub Actions invokes each step with
+        # `sh -e {0}` by default, so our `set -euo pipefail` headers fail
+        # immediately with `set: Illegal option -o pipefail` and exit code 2
+        # before any of the actual setup runs. Force bash for all run steps
+        # in this job so the pipefail option is recognised.
+        shell: bash
     steps:
       # The kicad/kicad image is Ubuntu-based and ships Python 3 + git, but
       # we install ca-certificates and curl defensively in case the image
@@ -185,8 +194,12 @@ jobs:
           apt-get update
           apt-get install -y --no-install-recommends ca-certificates curl git
           # Sanity: the container should have kicad-cli 9 already.
+          # Use a permissive regex anchored to either the start of the line
+          # or whitespace so we still match if `kicad-cli --version` prefixes
+          # its output (e.g., `KiCad 9.0.x`) instead of emitting the bare
+          # version string.
           kicad-cli --version | tee /tmp/kicad-cli-version
-          grep -qE '^9\.' /tmp/kicad-cli-version || {
+          grep -qE '(^|[[:space:]])9\.' /tmp/kicad-cli-version || {
             echo "::error::kicad/kicad:9.0 container does not provide kicad-cli 9.x"
             exit 1
           }
--- a/tests/test_kicad_cli_roundtrip.py
+++ b/tests/test_kicad_cli_roundtrip.py
@@ -1,0 +1,449 @@
+"""KiCad CLI round-trip smoke tests.
+
+Closes a regression class where kicad-tools' PCB writer produces files that
+``kicad-cli pcb drc`` rejects with "Failed to load board" (exit 3). Two such
+bugs shipped to ``main`` recently and broke every demo board's manufacturing
+export until they were diagnosed by hand-bisecting the emitted files (see
+fix in commit ``30256d40``):
+
+1. ``generator_version`` was emitted unquoted (``(generator_version 9.0)``)
+   because the SExp serializer's ``_needs_quoting`` returned False for
+   strings that parse as numeric. Fixed via
+   ``_FORCE_QUOTED_STRING_VALUE = {"generator_version"}`` in
+   ``src/kicad_tools/sexp/parser.py``.
+
+2. ``SilkscreenGenerator`` embedded a ``(kct_marking "kct:name")`` subfield
+   inside ``gr_text`` for re-run idempotency. KiCad's parser rejects
+   unknown subfields inside ``gr_text``. Fixed in
+   ``src/kicad_tools/silkscreen/generator.py``.
+
+The test in this file builds an in-suite PCB spec that exercises both
+regression vectors plus a footprint, a net, and a zone. It then calls
+``PCB.save()`` and round-trips the output through ``kicad-cli pcb drc``,
+asserting the file *loads* (DRC violations are not the concern here —
+only that kicad-cli accepts the syntax).
+
+If the load fails, the assertion message names the producing kicad-tools
+component, prints the verbatim kicad-cli stderr, and points at the emitted
+file so the failure is debuggable from the CI log alone.
+
+When ``find_kicad_cli()`` returns ``None`` the entire module is skipped
+(matching the existing convention at
+``tests/test_pcb_import_from_schematic.py:296``).
+"""
+
+from __future__ import annotations
+
+import textwrap
+from pathlib import Path
+
+import pytest
+
+from kicad_tools.cli.runner import find_kicad_cli, run_drc
+from kicad_tools.schema.pcb import PCB
+from kicad_tools.sexp.builders import gr_text_node, zone_node
+from kicad_tools.sexp.parser import parse_string
+from kicad_tools.silkscreen.generator import SilkscreenGenerator
+
+pytestmark = pytest.mark.skipif(find_kicad_cli() is None, reason="kicad-cli not installed")
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _format_load_failure(
+    producer: str,
+    pcb_path: Path,
+    return_code: int,
+    stderr: str,
+) -> str:
+    """Build a clear, attributed error message for a kicad-cli load failure.
+
+    Per the issue's acceptance criteria, the message must:
+    - Name the producing kicad-tools component.
+    - Include the PCB path so the file can be inspected.
+    - Print the verbatim kicad-cli stderr (this was the critical signal that
+      made the recent regression fix tractable).
+    """
+    return (
+        "kicad-cli rejected PCB emitted by kicad-tools writer.\n"
+        f"  Producer: {producer}\n"
+        f"  PCB: {pcb_path}\n"
+        f"  kicad-cli return code: {return_code}\n"
+        "  kicad-cli stderr:\n"
+        f"    {stderr.strip() or '<empty>'}\n"
+        "\n"
+        f"Inspect the emitted file: head -20 {pcb_path}"
+    )
+
+
+def _assert_kicad_cli_loads(
+    pcb_path: Path,
+    producer: str,
+    tmp_path: Path,
+) -> None:
+    """Assert that kicad-cli successfully loads the given PCB.
+
+    Uses ``run_drc`` as the load primitive — kicad-cli only writes a DRC
+    report when the file loads, so report-presence is the load signal.
+    Exit code 3 specifically indicates "Failed to load board".
+
+    DRC violations themselves are NOT a failure for this test — we only
+    care that the file is well-formed enough for kicad-cli to parse it.
+    """
+    output_path = tmp_path / f"{pcb_path.stem}_drc.rpt"
+    result = run_drc(
+        pcb_path,
+        output_path=output_path,
+        format="report",
+        # Disable schematic parity — we have no schematic for this synthetic
+        # PCB, and the load-check doesn't need it.
+        schematic_parity=False,
+    )
+
+    # ``run_drc`` returns success=True iff the report was produced, which is
+    # exactly the load-success condition.
+    assert result.success, _format_load_failure(
+        producer, pcb_path, result.return_code, result.stderr
+    )
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+class TestPCBSaveRoundtrip:
+    """``PCB.create()`` + ``PCB.save()`` must produce a kicad-cli-loadable file.
+
+    Targets the ``generator_version`` quoting regression: ``PCB.create()``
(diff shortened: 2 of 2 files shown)


rjwalters:

Thorough work on the test design — the file structure, the test attribution helpers, the pytestmark skip pattern, and the regression guard pre-checks (`assert "kct_marking" not in contents`, `assert '(generator_version "9.0")' in contents`) are all well-considered. The seven tests cover the two specific regression vectors plus broader writer surface (PCB.save, gr_text_node, zone_node, SilkscreenGenerator, parse_string + to_string) and produce the attributed error format the curator specified. Local verification of the regression-guard property (reverting either fix → tests fail with attributed errors) appears correct from inspection of the test code.

That said, this PR cannot be approved as-is because the new CI gate it introduces is permanently red:

### Blocking issue: the `kicad-cli Round-trip Smoke` job fails on every run

CI run https://github.com/rjwalters/kicad-tools/actions/runs/25316998709/job/74216698185 — 6 of 7 tests fail with `kicad-cli return code: 1` and stderr `DRC produced no output`.

**Root cause**: `apt-get install -y --no-install-recommends kicad` on `ubuntu-latest` (Ubuntu 24.04 noble) installs **KiCad 7.0.11** (visible in the log: `Setting up kicad (7.0.11+dfsg-1build4) ...` and `kicad-cli --version` → `7.0.11`).

But kicad-tools' writer emits forward-version markers that KiCad 7 rejects:
- `src/kicad_tools/schema/pcb.py:1184` — `(version 20240108)` (KiCad 8+ file format)
- `src/kicad_tools/schema/pcb.py:1186` — `(generator_version "9.0")` (KiCad 9+)

Locally the PR was tested against kicad-cli 10.x, where it passes. The mismatch was not caught because the CI workflow doesn't pin or verify a minimum KiCad version, and the apt-default on ubuntu-latest is too old.

### What this means for the gate's value

The regression-guard property the user asked about is partially preserved:
- The pre-checks (`assert "kct_marking" not in contents`, `assert '(generator_version "9.0")' in contents`) fire syntactically and would catch the two specific regressions even with no kicad-cli at all. **Good.**
- The actual round-trip property (kicad-cli accepts the file) is what catches *unanticipated* writer bugs — the bug class the issue is about. With KiCad 7 on the runner, this property is degraded to "kicad-cli 7 accepts the file", which is strictly weaker than what main was emitting (KiCad 8/9 format) and makes the test useless: every emitted PCB will fail this gate forever, regardless of correctness.

### Required fixes

Pick one of:

1. **Install a newer KiCad in the smoke job.** Add the KiCad PPA before installing:
   ```yaml
   - name: Install KiCad 9 (kicad-cli)
     run: |
       sudo add-apt-repository --yes ppa:kicad/kicad-9.0-releases
       sudo apt-get update
       sudo apt-get install -y --no-install-recommends kicad
       kicad-cli --version
   ```
   This matches the `(generator_version "9.0")` the writer emits. KiCad 8 PPA (`kicad/kicad-8.0-releases`) would also work since `(version 20240108)` is the KiCad 8 format epoch.

2. **Pre-check the kicad-cli major version** at job start and fail loudly if `< 8`, so the breakage is attributed to test infrastructure rather than to the PCB writer.

3. **Switch the runner to a container/image that ships a recent KiCad** (e.g. `kicad/kicad:9.0`).

I'd recommend option 1 with KiCad 9 to match what the writer emits today; if/when the writer bumps to 10, the PPA bumps to 10 in lockstep.

### Suggested verification before re-requesting review

- After the fix, the smoke job should be green on this PR with no test changes.
- Quick sanity that the gate still bites: the issue's acceptance criterion of "temporarily revert one fix, observe failure" should be re-run against the *CI-installed* kicad-cli (not just local 10.x) so we know the version we ship in CI is strict enough to catch the original `Failed to load board` (exit 3) signal. KiCad 9 is.

### Non-blocking observations (not required for approval)

- The CI yaml duplicates uv/python setup blocks from the existing test job — minor, acceptable.
- The smoke job uses `--no-cov` but doesn't pass `--timeout=60` like the main test job — fine, the surface is small.
- `TestLoadFailureMessage::test_unquoted_generator_version_reproduces_failure` (the canary) will also break under KiCad 7 since the unquoted form may be tolerated differently — worth confirming once the version is bumped.
- The pytestmark skipif covers the no-kicad-cli-at-all case correctly. Good.

rjwalters:
## Doctor — Judge feedback addressed

Fix pushed in e53d2e12: the `kicad-cli-smoke` job now installs KiCad 9 from `ppa:kicad/kicad-9.0-releases` before `apt-get install kicad`, matching the writer's emitted `(version 20240108)` / `(generator_version "9.0")` format markers.

### Diagnosis (from judge review)

`ubuntu-latest` is now Ubuntu 24.04 noble, whose default apt repos ship KiCad 7.0.11. KiCad 7 rejects the KiCad 8/9 format markers our writers emit, causing 6/7 tests to fail with `kicad-cli return code: 1` / `Failed to load board`. The local pass was on kicad-cli 10.x, which masked the version mismatch.

### Change

```yaml
- name: Install KiCad 9 (kicad-cli)
  run: |
    sudo add-apt-repository --yes ppa:kicad/kicad-9.0-releases
    sudo apt-get update
    sudo apt-get install -y --no-install-recommends kicad
    kicad-cli --version
```

The added comment in the workflow explains the version coupling so future writer-version bumps know to bump the PPA in lockstep.

### Verification

- YAML parses (validated locally with `yaml.safe_load`).
- Will be observable on the PR's checks tab once Actions schedules the run — expecting all 7 tests green.

Re-requesting review.

rjwalters:
## Judge Review — Changes Requested (Round 2)

The conceptual fix in `e53d2e12` is correct (use the kicad-9.0-releases PPA), but the workflow as-written is silently broken: when the PPA fetch fails, `apt-get install kicad` falls back to the default Ubuntu noble repo and installs KiCad **7.0.11**, exactly the version we need to avoid. CI is still red.

### What CI shows (run 25329021785, rerun job 74261775759)

The `Install KiCad 9 (kicad-cli)` step:

1. `add-apt-repository --yes ppa:kicad/kicad-9.0-releases` — completes (after one earlier 504 from launchpad's API).
2. `apt-get update` — emits warning, **does not fail**:
   ```
   Err:54 https://ppa.launchpadcontent.net/kicad/kicad-9.0-releases/ubuntu noble InRelease
     Cannot initiate the connection to ppa.launchpadcontent.net:443 (2620:2d:4000:1::81).
     - connect (101: Network is unreachable)
     Could not connect to ppa.launchpadcontent.net:443 (185.125.190.80), connection timed out
   W: Failed to fetch https://ppa.launchpadcontent.net/kicad/kicad-9.0-releases/ubuntu/dists/noble/InRelease ...
   ```
   `apt-get update` exits 0 even when a configured source can't be reached.
3. `apt-get install -y kicad` — happily picks up `kicad amd64 7.0.11+dfsg-1build4` from `noble/universe`:
   ```
   Get:30 http://azure.archive.ubuntu.com/ubuntu noble/universe amd64 kicad amd64 7.0.11+dfsg-1build4 [36.1 MB]
   Setting up kicad (7.0.11+dfsg-1build4) ...
   ```
4. `kicad-cli --version` — prints `7.0.11`, the script continues anyway because there's no version assertion.
5. Tests run against KiCad 7, 6/7 fail with `kicad-cli return code: 1`, stderr `DRC produced no output` (KiCad 7 silently rejects the v9-format PCB).

So the gate currently has the *opposite* of its intended property: when the PPA is unreachable (which is happening on this Azure runner — both attempts hit ppa.launchpadcontent.net network failures), CI falsely reports a fix-the-code regression instead of a fix-the-CI infrastructure issue.

### Required changes

Pick one of the two; either is acceptable:

**Option A (small, surgical) — fail-fast when the PPA isn't actually used.** Add explicit assertions to the install step so a fallback to KiCad 7 produces a clear, attributed CI error rather than 6 mysterious test failures:

```yaml
- name: Install KiCad 9 (kicad-cli)
  run: |
    sudo add-apt-repository --yes ppa:kicad/kicad-9.0-releases
    sudo apt-get update
    # Fail fast if the PPA isn't actually the source for kicad — otherwise
    # apt silently installs KiCad 7 from noble/universe and the smoke tests
    # produce confusing "DRC produced no output" failures.
    apt-cache policy kicad | grep -q 'kicad-9.0-releases' || {
      echo "::error::kicad/kicad-9.0-releases PPA is not the install source for the kicad package."
      echo "apt-cache policy kicad output:"
      apt-cache policy kicad
      exit 1
    }
    sudo apt-get install -y --no-install-recommends kicad
    # Belt-and-braces: assert the runtime kicad-cli is actually v9.
    kicad-cli --version | tee /tmp/kicad-cli-version
    grep -qE '^9\.' /tmp/kicad-cli-version || {
      echo "::error::Installed kicad-cli is not v9.x — see /tmp/kicad-cli-version above."
      exit 1
    }
```

This still leaves the gate vulnerable to Launchpad outages — it'll just fail loudly with a clear message instead of silently regressing. Given that PPA flakiness is recurrent on Azure runners (we hit it on both runs of this job), Option B is preferable if you have time.

**Option B (more robust) — drop the PPA, use the official KiCad container.** Replace the apt install with a container that ships kicad-cli 9 already:

```yaml
kicad-cli-smoke:
  name: kicad-cli Round-trip Smoke
  runs-on: ubuntu-latest
  container:
    image: kicad/kicad:9.0  # or pin a digest
  steps:
    - uses: actions/checkout@v4
    - name: Verify kicad-cli version
      run: kicad-cli --version | grep -qE '^9\.'
    - uses: astral-sh/setup-uv@v4
      with: { version: latest }
    - uses: actions/setup-python@v5
      with: { python-version: "3.12" }
    - run: uv sync --extra dev
    - run: uv run pytest tests/test_kicad_cli_roundtrip.py -v --no-cov
```

The container approach is repeatable (no apt mirror dependency), faster (no install step), and immune to Launchpad outages. It does require checking that `kicad/kicad:9.0` (or whatever image you choose) has a working Python + git + uv stack, or you'll need to build a small tools image yourself.

### Acknowledgements

- Test file (`tests/test_kicad_cli_roundtrip.py`) is unchanged from round 1 and still excellent — pytestmark skip, attributed error messages, regression-guard pre-checks, all good.
- Job-isolation rationale in the PR body still stands.
- One of the two CI attempts I just observed succeeded all the way through to running pytest, so the test code itself is exercised on the runner — the failure is purely the kicad version mismatch.

Reverting the label from `loom:review-requested` to `loom:changes-requested`.

rjwalters:
## Doctor — Judge Round 2 feedback addressed (Option B)

Replaced the apt+PPA install in the `kicad-cli-smoke` job with the official `kicad/kicad:9.0` container, per the judge's preferred Option B. Pushed in `096b6585`. Branch was also rebased on top of the 8 PRs that merged today (#2502, #2503, #2504, #2505, #2506, #2509, #2510, #2511, #2512) — force-with-lease push.

### What changed

`.github/workflows/ci.yml` — `kicad-cli-smoke` job now:
- Drops `runs-on: ubuntu-latest` execution context in favor of `container: { image: kicad/kicad:9.0, options: --user 0:0 }`. The image ships kicad-cli 9 already, so the apt+PPA dance is gone entirely.
- Adds an "Ensure container has prerequisites" step that `apt-get install`s `ca-certificates curl git` defensively (in case the container variant is missing one), then asserts `kicad-cli --version` matches `^9\.` and fails loudly with `::error::` if not.
- Installs uv via the official `astral.sh/uv/install.sh` one-liner (rather than `astral-sh/setup-uv@v4`) to avoid action-host coupling inside the container, then surfaces `~/.local/bin` to `$GITHUB_PATH`.
- Runs `uv sync --extra dev` and `uv run pytest tests/test_kicad_cli_roundtrip.py -v --no-cov` inside the container as before.

The updated comment block above the job documents *why* we moved to the container (Launchpad-outage tolerance, no silent KiCad 7 fallback) so the next reader doesn't undo the change.

### Why Option B and not Option A

Both options were on the table. Option B is strictly more reliable than Option A:
- Option A still leaves the gate vulnerable to Launchpad PPA outages — it just fails loudly instead of silently. We hit those outages back-to-back on the previous run.
- Option B is immune to Launchpad outages entirely (no PPA dependency).
- Container size (~1–2 GB pull) replaces the ~36 MB apt install, but the pull is cached and only happens once per runner — net wall-clock should be similar.

The kicad/kicad image is Ubuntu-based, ships Python 3 (KiCad's own scripting requirement) and git, so the in-container `actions/checkout@v4` and `uv sync` paths work without bringing the full Python toolchain. The defensive `apt-get install` of `ca-certificates curl git` covers any image variant where one of those is trimmed.

### Verification

- YAML parses (`python3 -c 'import yaml; yaml.safe_load(open(".github/workflows/ci.yml").read())'`).
- 7 / 7 smoke tests pass locally against kicad-cli 10.x in 10.8 s — same as before, since the test code is unchanged from round 1.
- Will be observable on the PR's checks tab once Actions schedules the run.

### Fallback plan if Option B turns out to need rework

If the container is missing something we can't paper over (e.g. Python is too old for kicad-tools or a system lib uv needs is absent), I'll fall back to Option A: `apt-cache policy kicad | grep -q 'kicad-9.0-releases'` and `kicad-cli --version | grep -qE '^9\\.'` post-install assertions. Test code itself does not change in either case.

Re-requesting review.

rjwalters:
## Judge Review (3rd pass) — Changes Requested

Option B (official `kicad/kicad:9.0` container) is the right architectural call, but the new `kicad-cli-smoke` job **fails in the very first step** due to a shell incompatibility. The version assertion never runs — kicad-cli is never invoked — so we cannot yet conclude the assertion design works as intended.

### CI status (run 25330554699, head 096b6585)

- Lint & Format: pass
- Type Check: pass
- Test: in progress
- **kicad-cli Round-trip Smoke: FAIL (24s)** — exit code 2 in the prerequisites step
- `mergeStateStatus`: `UNSTABLE` (not `CLEAN`) — required for approval

### Blocker — `set -o pipefail` is rejected by the container's default shell

From the failed job log:

```
##[group]Run set -euo pipefail
...
shell: sh -e {0}
##[endgroup]
/__w/_temp/64dc2835-013d-4d03-8622-c27478e01b88.sh: 1: set: Illegal option -o pipefail
##[error]Process completed with exit code 2.
```

The `kicad/kicad:9.0` image does not ship `/bin/bash` as the default shell that GitHub Actions detects, so Actions falls back to `sh -e {0}` (dash). Dash does not implement `-o pipefail`, so the script exits with code 2 before `apt-get update` even runs. This means:

1. The "Ensure container has prerequisites" step never executes `apt-get install` or the version assertion.
2. The "Install uv" step (which also uses `set -euo pipefail`) would fail identically if reached.
3. The version-fallback question (b) cannot be answered empirically yet — the guard is in unreached code.

### Why the version assertion *would* work, but the test isn't real yet

Reviewing the assertion logic in isolation:

```yaml
kicad-cli --version | tee /tmp/kicad-cli-version
grep -qE '^9\.' /tmp/kicad-cli-version || { ... exit 1; }
```

This is correct **in principle** — `kicad-cli --version` on KiCad 7 prints `7.0.x`, which would not match `^9\.` and would trigger the fail-fast. **But** the regex is anchored to start-of-line, and `kicad-cli` on some builds prefixes the version line with `KiCad`, e.g. `KiCad 9.0.5`. Worth verifying once the script actually runs — see suggestion below. (If the bare `9.0.x` form is what `kicad/kicad:9.0` emits, the current regex is fine.)

The bigger issue right now is that we have not actually demonstrated the guard fires, because the step never runs.

### Required fixes

**1. Force bash for every shell step in the container job.**

Either set `defaults.run.shell: bash` at the job level, or add `shell: bash` to each `run:` step. Recommended (job-level, applies to all steps):

```yaml
kicad-cli-smoke:
  name: kicad-cli Round-trip Smoke
  runs-on: ubuntu-latest
  timeout-minutes: 15
  container:
    image: kicad/kicad:9.0
    options: --user 0:0
  defaults:
    run:
      shell: bash
  steps:
    ...
```

If `bash` is not present in the image (it should be — Ubuntu base — but worth confirming), drop `pipefail` and use `set -eu` plus explicit exit-code checking on the piped command, e.g.:

```sh
set -eu
kicad-cli --version > /tmp/kicad-cli-version
cat /tmp/kicad-cli-version
grep -qE '(^|[[:space:]])9\.' /tmp/kicad-cli-version || {
  echo "::error::kicad/kicad:9.0 container does not provide kicad-cli 9.x"
  exit 1
}
```

**2. Re-verify the version regex matches the actual output format.**

Once the step runs, confirm whether `kicad-cli --version` inside `kicad/kicad:9.0` prints `9.0.x` (current `^9\.` works) or `KiCad 9.0.x` (current `^9\.` would *fail-open* incorrectly, blocking even valid 9.x). The relaxed regex above (`(^|[[:space:]])9\.`) handles both.

### What's good

- Container approach is exactly right — eliminates Launchpad PPA flakiness entirely.
- Pinning `kicad/kicad:9.0` (not `:latest`) is the correct call for reproducibility.
- `--user 0:0` avoids the checkout-permission gymnastics.
- Defensive `apt-get install ca-certificates curl git` is a nice safety net for image variant drift.
- Replacing `setup-uv@v4` + `setup-python@v5` with the upstream installer is the right move inside a container — avoids tool-cache action coupling.
- Job-level `timeout-minutes: 15` is appropriate.
- Good attribution comment block explaining *why* container vs PPA.

### Verdict

Adding `loom:changes-requested`. Re-request review once the bash-shell fix is in and the smoke job actually runs through the version assertion to a green or attributable-red result.

rjwalters:
Addressed Judge re-review feedback in ba6dc0fd.

**Root cause identified**: the kicad/kicad:9.0 container's `/bin/sh` is dash, not bash. GitHub Actions invokes each step with `sh -e {0}`, so our `set -euo pipefail` headers exited immediately with `set: Illegal option -o pipefail` (exit 2) at step 1 — before the version-assertion guard ever ran. That's why the failure mode looked like "no useful output": dash bailed on line 1 of the very first step.

**Fix**:
1. Added `defaults.run.shell: bash` at the `kicad-cli-smoke` job level so all `run` steps use bash, where `pipefail` is supported.
2. Widened the kicad-cli version-check regex from `^9\.` to `(^|[[:space:]])9\.` so it still matches if `kicad-cli --version` prefixes the version with a banner (e.g., `KiCad 9.0.x`) rather than emitting the bare version string.

The script bodies were already correct; they just needed a shell that understands pipefail.

rjwalters:
## Judge Review (4th pass) — Changes Requested (rebase only)

The bash-shell fix and version-regex widening are both correct and the smoke job now passes end-to-end on CI. **Blocking issue: the branch is now in merge conflict with `main`** because PR #2508 landed a new `cpp-build-check` job in `.github/workflows/ci.yml` after this branch was last pushed.

### Verification of the fix (ba6dc0fd)

1. **`defaults.run.shell: bash` resolves the dash failure.** The smoke job logs now show `shell: bash --noprofile --norc -e -o pipefail {0}` for every step, and `set -euo pipefail` no longer trips. Container is `kicad/kicad:9.0` (digest `sha256:e638b79b...`), `kicad-cli --version` reports `9.0.9`, and all 7 round-trip tests passed in 2.67 s. Total job runtime ~56 s including image pull. (Job: `actions/runs/25330741898/job/74263491815`.)

2. **The widened regex `(^|[[:space:]])9\.` still catches a KiCad 7 silent fallback.** Verified locally:
   - Matches: `9.0.0`, `KiCad 9.0.0`, `kicad-cli 9.0.9`
   - Rejects: `7.0.11`, `KiCad 7.0.11`, `kicad-cli 7.0.11`
   The regex change correctly accommodates both bare-version (current observed kicad-cli output) and prefixed-version (e.g., `KiCad 9.x.x`) forms while still failing loudly on a v7 fallback. Minor note: it would also match a hypothetical `19.0.0`, but kicad-cli has no v19 line and this is a sanity guard not a strict bound — fine.

3. **All four CI checks SUCCESS:**
   - Lint & Format: SUCCESS
   - Type Check: SUCCESS
   - Test: SUCCESS
   - kicad-cli Round-trip Smoke: SUCCESS (7/7 tests pass)

### Blocking: merge conflict

`mergeStateStatus: DIRTY`, `mergeable: CONFLICTING`. PR #2508 (commit 7fceaee7, `router_cpp: detect stale .so via build version guard`) landed on main and added a new `cpp-build-check` job in `.github/workflows/ci.yml` between this branch's base and HEAD. Both PRs append a new job in the same region of the file, so git can't auto-merge.

**Action needed:** rebase `feature/issue-2492` onto current `origin/main` (or merge main in), keeping both the new `cpp-build-check` job and the `kicad-cli-smoke` job. Re-run CI to confirm both jobs pass post-rebase, then re-request review.

Once the rebase is clean, this is ready to land — the substantive content (bash-shell fix + permissive version regex + container-based KiCad 9 install + 7-test round-trip suite) is correct and all checks pass.

rjwalters:
Rebased onto current main to resolve the merge conflict with #2508. Conflict was in `.github/workflows/ci.yml` where both PRs added new jobs in the same region; resolution preserves both `cpp-build-check` (from #2508) and `kicad-cli-smoke` (this PR) as separate jobs. Branch is now 4 commits ahead of `main` with the same substantive content reviewed in round 3.

Resetting to `loom:review-requested` for final verification.

rjwalters:
## Final review pass: APPROVED

Re-reviewed after manual rebase resolved the conflict with #2508 (cpp-build-check). Verified:

**Merge state**
- `mergeStateStatus`: CLEAN
- `mergeable`: MERGEABLE
- HEAD: `fffaf3fe` (4 commits ahead of main, base updated cleanly)

**CI rollup — all 5 checks green**
- Lint & Format: pass (11s)
- Type Check: pass (40s)
- Test: pass (5m29s)
- C++ Build Check: pass (41s) — preserved from #2508, runs in parallel
- kicad-cli Round-trip Smoke: pass (39s) — the new gate works end-to-end against `kicad/kicad:9.0`

**Conflict resolution sanity check**
Inspected `.github/workflows/ci.yml` at the rebased HEAD: `cpp-build-check` (line 73) and `kicad-cli-smoke` (line 153) are both top-level entries under `jobs:`, properly siblings rather than nested. All previously-approved content survived the rebase verbatim:
- `container: kicad/kicad:9.0` with `--user 0:0`
- `defaults.run.shell: bash` (the dash/pipefail fix)
- Permissive version regex `(^|[[:space:]])9\.` on `kicad-cli --version`
- Comments explaining the PPA-vs-container rationale

This is the cleanest possible resolution — no logic merged, no jobs collapsed, both regression gates land independently. Approving for merge.

Removing `loom:review-requested`, adding `loom:pr`.

PR #2443, opened 2026-05-03
perf(router): port geometric validation to C++ for faster route exploration

rjwalters:
## Summary
Port the 4 Python geometric validation methods into C++ to eliminate per-route Python callback overhead during post-route clearance checking. This replaces the bottleneck where each A* pathfind result was validated by calling back into Python for segment-pad, segment-segment, via-segment, and via-via clearance checks.

## Changes
- Add `geometry.hpp`/`geometry.cpp` with `point_to_segment_distance`, `segment_to_segment_distance`, `segments_intersect` (faithful port of `core/geometry.py`) and deterministic FNV-1a hash function
- Add `PadInfo`, `StoredSegment`, `StoredVia`, `ValidationResult` structs to `types.hpp`
- Add `Grid3D::validate_route()` to `grid.hpp`/`grid.cpp` implementing all 4 validation checks in a single C++ call
- Add nanobind bindings for all new types, methods, and standalone geometry functions
- Update `cpp_backend.py:from_routing_grid()` to populate pad data with pre-computed per-component clearance overrides and FNV-1a ref hashes
- Update `cpp_backend.py:route()` to use C++ validation with incremental stored route syncing instead of Python callbacks
- Add `min_drill_clearance` field to C++ `DesignRules` struct

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| C++ geometry functions match Python | PASS | 10 unit tests for point_to_segment_distance, segment_to_segment_distance, segments_intersect |
| FNV-1a hash is deterministic | PASS | 3 tests: empty string, deterministic, collision avoidance |
| Segment-to-pad validation parity | PASS | 4 tests: far pass, close fail, same-net exclude, ref exclude |
| Segment-to-segment validation | PASS | 2 tests: too close fails, far apart passes |
| Via-to-segment validation | PASS | 2 tests: too close fails, far apart passes |
| Via-to-via validation | PASS | 2 tests: too close fails, far apart passes |
| Same-net drill spacing | PASS | 3 tests: too close fails, far apart passes, self-position excluded |
| Per-component clearance overrides | PASS | 1 test: U1 with 0.08mm override |
| Layer filtering | PASS | 2 tests: different layer ignored, through-hole checked on all |
| Existing tests unchanged | PASS | test_component_clearance.py (159 passed), test_router_grid.py pass, test_cpp_clearance_enforcement.py (8/9 pass, 1 pre-existing failure) |
[...]

PR #2508, opened 2026-05-04
router_cpp: detect stale .so via build version guard (closes #2501)

### LINKED ISSUE #2501: router_cpp: detect/recover from stale compiled .so vs source mismatch
## Problem
`src/kicad_tools/router/router_cpp.cpython-313-darwin.so` is checked into the repo. When the .cpp source moves ahead (e.g. someone adds `PadBounds` to `bindings.cpp`) but the .so isn't rebuilt and committed, you get a hard runtime crash on every routing call:
```
AttributeError: module 'kicad_tools.router.router_cpp' has no attribute 'PadBounds'
  at src/kicad_tools/router/cpp_backend.py:477  bounds = router_cpp.PadBounds()
```
This happened in this session — the .so was dated May 1, the .cpp source was May 3. Every routing test/board would crash at import-mismatch.

## Three options (pick one or combine)

**Option A — gitignore the .so and autobuild on missing.**
- Remove the .so from the repo
- Add post-install hook (or import-time check in `cpp_backend.py`) that runs `scripts/build-cpp.sh` if the .so is missing.

**Option B — stamp a build hash and warn on mismatch.**
- During build, write a manifest with `git rev-parse HEAD` of the source tree at build time.
- On import, compare to current source hash; emit `RuntimeWarning` if drift detected.

**Option C — version-bump enforcement.**
- Maintain `cpp/include/version.hpp` with `ROUTER_CPP_BUILD_VERSION`; bump on every binding change.
- `cpp_backend.py` reads `router_cpp.__version__` (already exposed) and hard-fails if the constant doesn't match.

A is simplest and removes the failure mode entirely. C provides clearest user feedback.

## Acceptance criteria
- [ ] No more silent `AttributeError` from stale .so.
- [ ] Clear actionable error or auto-rebuild path.
- [ ] CI verifies the .so is rebuildable from source on a clean checkout.

## Repro
```bash
git log --since="2 weeks ago" -- src/kicad_tools/router/cpp/  # Source moved
git log --since="2 weeks ago" -- src/kicad_tools/router/router_cpp.cpython-313-darwin.so  # .so unchanged
```

rjwalters:
## Summary

Closes #2501.

When the C++ source under `src/kicad_tools/router/cpp/` moves ahead of the compiled `.so`, every routing call crashes with `AttributeError` on missing symbols (e.g. `router_cpp.PadBounds`, `router_cpp.FAILURE_NONE`) deep in `cpp_backend.py` rather than at import time. This PR implements the curator's recommended hybrid Option A+C: a `ROUTER_CPP_BUILD_VERSION` integer constant mirrored on both sides of the binding boundary, plus a Python-side guard that disables the C++ backend with an actionable `kct build-native` hint when the two disagree, routing through the existing pure-Python fallback path.

## Changes

- **C++**: Added `ROUTER_CPP_BUILD_VERSION = 2` constant in `cpp/include/types.hpp`. Exposed it as both `router_cpp.BUILD_VERSION` (module attribute) and `router_cpp.build_version()` (callable) in `bindings.cpp`.
- **Python**: Added `_REQUIRED_CPP_BUILD_VERSION = 2` in `cpp_backend.py` and a post-import guard that compares `router_cpp.BUILD_VERSION` against the required value. On mismatch (or missing attribute), sets `_CPP_AVAILABLE = False` and populates `_CPP_IMPORT_ERROR` with a stale-`.so` rebuild message. Routes through the existing fallback path used today by ABI-tag mismatches.
- **Tests**: New `TestCppBuildVersionGuard` class with 4 tests covering: required constant exists, build version exposed when C++ loads, mismatched `BUILD_VERSION` disables backend with `build-native` hint, missing `BUILD_VERSION` attribute disables backend.
- **CI**: New `cpp-build-check` job that does a clean rebuild from source on Ubuntu and asserts the version guard reports the backend as available. Catches forgotten version bumps or stale `.so` artifacts before they hit `main`.
- **Docs**: Added a "C++ Native Extensions / Build version discipline" section to `docs/contributing/development.md` documenting the version-bump-and-rebuild requirement.

## Test plan

- [x] `pytest tests/test_router_cpp_backend.py::TestCppBuildVersionGuard -v` (4/4 pass locally on Python 3.13)
- [x] `pytest tests/test_router_cpp_backend.py::TestCppBackendFallback tests/test_router_cpp_backend.py::TestFormatBackendStatus` (10/10 pass; existing fallback behavior preserved)
- [x] `pytest tests/test_cpp_validation_parity.py` (33/33 pass; C++ backend still works after the guard)
- [x] `pytest tests/test_cpp_performance.py` (7/7 pass; C++ backend still hot-path on routing)
- [x] Manually verified `router_cpp.BUILD_VERSION == 2` after clean rebuild.
- [ ] CI `cpp-build-check` job passes on Ubuntu (will run on this PR).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,9 +86,13 @@ jobs:
         with:
           version: "latest"
 
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.12"
+      # Note: deliberately NOT using actions/setup-python@v5 here. That action
+      # exports Python_ROOT_DIR=/opt/hostedtoolcache/Python/.../x64, which is
+      # sticky for cmake's find_package(Python ...) — even after activating
+      # the uv venv with `source .venv/bin/activate`, cmake locates the
+      # tool-cache Python (which has no nanobind) instead of the venv's
+      # Python with nanobind installed. We rely on uv to manage the Python
+      # version (pinned via .python-version / pyproject) and create the venv.
 
       - name: Install build dependencies
         run: |
@@ -96,7 +100,11 @@ jobs:
           sudo apt-get install -y cmake build-essential
 
       - name: Install Python dependencies
-        run: uv sync --extra dev
+        # Pin Python via uv (parity with the other jobs that use
+        # actions/setup-python with python-version: "3.12"). uv will install
+        # the requested interpreter into its managed cache and create
+        # .venv from it.
+        run: uv sync --extra dev --python 3.12
 
       - name: Install nanobind
         run: uv pip install nanobind
@@ -107,6 +115,10 @@ jobs:
       - name: Build C++ extension from source
         run: |
           source .venv/bin/activate
+          # Belt-and-suspenders: explicitly point cmake at the venv Python in
+          # case any future runner image starts setting Python_ROOT_DIR.
+          export Python_ROOT_DIR="$PWD/.venv"
+          unset Python2_ROOT_DIR Python3_ROOT_DIR
           bash scripts/build-cpp.sh build
 
       - name: Verify version-guard reports backend available
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,3 +69,71 @@ jobs:
       - name: Run tests
         run: uv run pytest -n auto -o addopts= --benchmark-disable --timeout=60 -m "not slow"
         continue-on-error: true
+
+  cpp-build-check:
+    # Issue #2501: catch source/.so drift by rebuilding the C++ extension
+    # from a clean checkout and asserting the version-guard reports the
+    # backend as available. If anyone changes cpp/ source without bumping
+    # ROUTER_CPP_BUILD_VERSION (or vice versa) the import-time guard will
+    # fall back to Python and fail this check with an actionable message.
+    name: C++ Build Check
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: astral-sh/setup-uv@v4
+        with:
+          version: "latest"
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cmake build-essential
+
+      - name: Install Python dependencies
+        run: uv sync --extra dev
+
+      - name: Install nanobind
+        run: uv pip install nanobind
+
+      - name: Clean any committed .so artifacts
+        run: bash scripts/build-cpp.sh clean
+
+      - name: Build C++ extension from source
+        run: |
+          source .venv/bin/activate
+          bash scripts/build-cpp.sh build
+
+      - name: Verify version-guard reports backend available
+        run: |
+          uv run python -c "
+          from kicad_tools.router.cpp_backend import (
+              is_cpp_available,
+              get_cpp_unavailable_reason,
+              _REQUIRED_CPP_BUILD_VERSION,
+              router_cpp,
+          )
+          if not is_cpp_available():
+              raise SystemExit(
+                  f'C++ backend unavailable after clean rebuild: '
+                  f'{get_cpp_unavailable_reason()}'
+              )
+          actual = getattr(router_cpp, 'BUILD_VERSION', None)
+          if actual != _REQUIRED_CPP_BUILD_VERSION:
+              raise SystemExit(
+                  f'BUILD_VERSION mismatch after clean rebuild: '
+                  f'compiled={actual!r} required={_REQUIRED_CPP_BUILD_VERSION!r}. '
+                  f'Did you forget to bump ROUTER_CPP_BUILD_VERSION in both '
+                  f'cpp/include/types.hpp and cpp_backend.py?'
+              )
+          print(f'C++ backend OK: BUILD_VERSION={actual}')
+          "
+
+      - name: Smoke test C++ router
+        run: |
+          uv run pytest tests/test_router_cpp_backend.py::TestCppBuildVersionGuard -v --no-cov
--- a/docs/contributing/development.md
+++ b/docs/contributing/development.md
@@ -114,6 +114,57 @@ kicad-tools/
 
 ---
 
+## C++ Native Extensions
+
+The router and placement optimizer ship optional C++ acceleration modules
+(`router_cpp`, `placement_cpp`) for 10-100x speedup over the pure-Python
+implementation. They are built with nanobind + CMake and live under
+`src/kicad_tools/router/cpp/` and `src/kicad_tools/placement/cpp/`.
+
+### Building the C++ extensions
+
+```bash
+# Build (requires CMake 3.15+ and a C++20 compiler)
+bash scripts/build-cpp.sh build
+
+# Or via the CLI:
+kct build-native
+
+# Clean build artifacts:
+bash scripts/build-cpp.sh clean
+```
+
+### Build version discipline (Issue #2501)
+
+The Python side caches a compiled `.so` matching `cpython-<X>-<platform>.so`.
+If the `.cpp` source moves ahead of the compiled `.so`, the import will
+succeed but new symbols (`PadBounds`, `FAILURE_NONE`, etc.) will be missing,
+causing hard `AttributeError` crashes deep in the routing code.
+
+To prevent this, the build surface is gated by an integer constant
+`ROUTER_CPP_BUILD_VERSION`, mirrored on both sides:
+
+- `src/kicad_tools/router/cpp/include/types.hpp` (C++ source)
+- `src/kicad_tools/router/cpp_backend.py` as `_REQUIRED_CPP_BUILD_VERSION`
+  (Python side)
+
+**When you change anything under `src/kicad_tools/router/cpp/` that affects
+the binding surface (added/removed/renamed symbols, struct fields, function
+signatures), you MUST:**
+
+1. Bump `ROUTER_CPP_BUILD_VERSION` in `cpp/include/types.hpp`.
+2. Bump `_REQUIRED_CPP_BUILD_VERSION` to the same value in `cpp_backend.py`.
+3. Rebuild with `kct build-native` (or `bash scripts/build-cpp.sh build`).
+
+If the two constants disagree at import time, the C++ backend is disabled
+with a clear `kct build-native` rebuild hint and the router falls back to
+pure Python rather than crashing at routing time. The CI `cpp-build-check`
+job rebuilds from a clean checkout and asserts the guard reports the
+backend as available, so a forgotten version bump or stale `.so` will fail
+CI rather than silently regress production.
+
+---
+
 ## Running Tests
 
 ### Run All Tests
--- a/src/kicad_tools/router/cpp/include/types.hpp
+++ b/src/kicad_tools/router/cpp/include/types.hpp
@@ -12,6 +12,16 @@
 
 namespace router {
 
+// Build version for the C++ binding surface (Issue #2501).
+//
+// Bump this constant in any PR that changes the bindings.cpp surface
+// (added/removed/renamed symbols, struct fields, function signatures).
+// The Python side mirrors this as ``_REQUIRED_CPP_BUILD_VERSION`` in
+// ``cpp_backend.py``; on import the two are compared and a mismatch
+// disables the C++ backend with a clear "kct build-native" error,
+// preventing silent ``AttributeError`` failures from a stale .so.
+constexpr int ROUTER_CPP_BUILD_VERSION = 2;
+
 // Grid cell state
 struct GridCell {
     bool blocked = false;
--- a/src/kicad_tools/router/cpp/include/types.hpp
+++ b/src/kicad_tools/router/cpp/include/types.hpp
@@ -6,6 +6,7 @@
 #pragma once
 
 #include <cstdint>
+#include <limits>
 #include <vector>
 #include <tuple>
 #include <optional>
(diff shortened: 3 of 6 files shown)


rjwalters:
## Judge Review (Changes Requested)

Approach is sound — the hybrid Option A+C tracks the curator's plan well, the C++ side / Python guard / tests / docs all wire up correctly, and the design routes stale-`.so` cleanly through the existing fallback path. However, this PR cannot be approved as-is because the new CI job it introduces (`cpp-build-check`) fails on this very PR. Merging would land a known-broken CI step on `main`.

### Blocking: `cpp-build-check` job fails

Run https://github.com/rjwalters/kicad-tools/actions/runs/25317016153/job/74216749248

```
Configuring...
-- Found Python: /opt/hostedtoolcache/Python/3.12.13/x64/bin/python3.12 (found version "3.12.13") ...
/opt/hostedtoolcache/Python/3.12.13/x64/bin/python3.12: No module named nanobind
CMake Error at src/kicad_tools/router/cpp/CMakeLists.txt:23 (find_package):
  Could not find a package configuration file provided by "nanobind" ...
```

Root cause: the workflow installs nanobind into the uv venv (`uv pip install nanobind`), but `scripts/build-cpp.sh` invokes cmake which calls `find_package(Python ...)` in `CMakeLists.txt:19`. Because `actions/setup-python@v5` exports `Python_ROOT_DIR=/opt/hostedtoolcache/Python/3.12.13/x64`, cmake locates the system Python (which has no nanobind) instead of the venv's Python — even though the workflow `source`s the venv before running the build. Cmake's `Python_EXECUTABLE` is sticky once `Python_ROOT_DIR` is set.

Suggested fixes (any one is sufficient):

1. Install nanobind into the system Python that cmake will find:
   ```yaml
   - name: Install nanobind into system Python
     run: |
       /opt/hostedtoolcache/Python/3.12.13/x64/bin/python3.12 -m pip install nanobind
   ```
2. Or, more robust: explicitly point cmake at the venv Python by exporting `Python_ROOT_DIR` before the build:
   ```yaml
   - name: Build C++ extension from source
     run: |
       source .venv/bin/activate
       export Python_ROOT_DIR="$PWD/.venv"
       unset Python2_ROOT_DIR Python3_ROOT_DIR
       bash scripts/build-cpp.sh build
   ```
3. Or, simplest: drop the `actions/setup-python@v5` step entirely and rely on the uv-managed venv (`uv run` for the verify step), since the build only needs *a* Python with nanobind.

### Non-blocking observations

- `bindings.cpp` exposes both `m.attr("BUILD_VERSION")` and `m.def("build_version", ...)`, but the Python guard and CI verify step only use the attribute. Dropping the unused callable would shrink the binding surface (one less symbol to keep stable). Not worth a respin on its own, though.
- `scripts/build-cpp.sh:75` uses the long-deprecated `-DPYTHON_EXECUTABLE` flag (the variable in `CMakeLists.txt` is `Python_EXECUTABLE`, set by `find_package(Python)`). Pre-existing, but it contributes to the CI confusion above. Worth fixing in the same PR or a follow-up.
- The two tests that re-import `cpp_backend` (`test_stale_so_disables_cpp_with_actionable_error`, `test_missing_build_version_attr_disables_cpp`) leave `cpp_backend` re-imported at the *real* state in their `finally` blocks, which is good — no test isolation concerns there.
- Docs section is clear; the "binding surface" wording correctly distinguishes the bump trigger from "any cpp/ change."

### Action

Once the `cpp-build-check` job passes on this PR (and `gh pr view 2508 --json mergeStateStatus` returns CLEAN rather than UNSTABLE), re-request review and I'll re-evaluate.

Removing `loom:pr`, applying `loom:changes-requested`.

rjwalters:
## Doctor: addressed judge feedback

Pushed `713ac86e` to fix the blocking `cpp-build-check` CI failure.

### Root cause (judge's diagnosis, confirmed)

`actions/setup-python@v5` exported `Python_ROOT_DIR=/opt/hostedtoolcache/Python/3.12.13/x64`, which cmake's `find_package(Python ...)` honored over the activated venv. Result: cmake located the tool-cache Python (no nanobind) instead of the uv venv's Python.

### Fix (judge's recommended option C, plus belt-and-suspenders)

1. **Removed `actions/setup-python@v5`** from the `cpp-build-check` job. uv now handles Python install/management directly (`uv sync --extra dev --python 3.12` keeps parity with the other jobs that pinned 3.12). Without setup-python, no `Python_ROOT_DIR` is exported and cmake picks up the activated venv's `python3` from `PATH`.
2. **Belt-and-suspenders**: explicitly `export Python_ROOT_DIR="$PWD/.venv"` (and `unset Python2_ROOT_DIR Python3_ROOT_DIR`) before invoking cmake, so the job stays correct if a future runner image pre-sets `Python_ROOT_DIR`.
3. **Fixed deprecated cmake flag** in `scripts/build-cpp.sh`: now passes `-DPython_EXECUTABLE` (matches the modern `FindPython` module variable) instead of the long-ignored `-DPYTHON_EXECUTABLE`. This was the second non-blocking observation in the review.

### Local validation

Performed a clean rebuild on macOS that mirrors the CI flow:

```
$ bash scripts/build-cpp.sh clean
$ source .venv/bin/activate
$ export Python_ROOT_DIR="$PWD/.venv"
$ unset Python2_ROOT_DIR Python3_ROOT_DIR
$ bash scripts/build-cpp.sh build
... [100%] Built target router_cpp ...
C++ router module installed to: src/kicad_tools/router/router_cpp.cpython-313-darwin.so
Backend: cpp / Available: True / Version: 1.0.0

$ uv run python -c "<version-guard verify script>"
C++ backend OK: BUILD_VERSION=2

$ uv run pytest tests/test_router_cpp_backend.py::TestCppBuildVersionGuard -v --no-cov
4 passed in 0.18s
```

### Deferred to follow-up

Per the judge's "not worth a respin" note, the unused `m.def("build_version", ...)` callable in `bindings.cpp` is left in place — removing it would change the binding surface and trigger a `ROUTER_CPP_BUILD_VERSION` bump for what is purely a cleanup. Happy to file a follow-up issue if desired.

Re-requesting review.

rjwalters:

Thanks for the quick turnaround on the nanobind/setup-python fix in 713ac86e — that part of the issue is resolved. cmake now correctly finds the venv Python and nanobind. However, two new blockers have surfaced.

### Blocker 1 — Merge conflict with newly-merged #2502 (`mergeStateStatus: DIRTY`, `mergeable: CONFLICTING`)

PR #2502 merged to main today (commit 14defddc) and rewrote `scripts/build-cpp.sh` to:
- Compute `VENV_PY="$PROJECT_ROOT/.venv/bin/python3"` and use `"$PY"` for both the nanobind preflight and the cmake configure invocation.
- Switch from `-DPYTHON_EXECUTABLE` to `-DPython_EXECUTABLE` (the modern FindPython spelling).

This PR also rewrites the same lines, so the trees collide. Worse, the *resolution* this PR carries is strictly weaker than what just landed on main:
- `scripts/build-cpp.sh` here drops the `VENV_PY`/`PY` block entirely and falls back to `python3` for the nanobind preflight check and `which python3` for the cmake flag. That re-introduces the original failure mode #2502 fixed: when running outside an activated venv (e.g. local `bash scripts/build-cpp.sh build` without `source .venv/bin/activate`), nanobind won't be discovered and the script will print "nanobind not installed" even when uv has it installed in `.venv`.

Recommended resolution: rebase onto current `origin/main`, accept main's `scripts/build-cpp.sh` verbatim (it already has the `-DPython_EXECUTABLE` rename you wanted, plus the `VENV_PY` discovery), and drop this PR's competing rewrite. The CI workflow change in `.github/workflows/ci.yml` is independent and survives the rebase cleanly.

### Blocker 2 — `cpp-build-check` job fails on Ubuntu (gcc 13) due to missing `<limits>` include

CI run https://github.com/rjwalters/kicad-tools/actions/runs/25329102023/job/74257756745:

```
src/kicad_tools/router/cpp/include/types.hpp:186:55: error: incomplete type 'std::numeric_limits<float>' used in nested name specifier
  186 |     float min_clearance = std::numeric_limits<float>::infinity();
```

`types.hpp` uses `std::numeric_limits<float>::infinity()` (line 176 in source, introduced in #2443 / commit c109e7f3 for `ValidationResult.min_clearance`) but never `#include <limits>`. The current includes are only `<cstdint> <vector> <tuple> <optional>`.

This compiles on macOS / Apple libc++ because one of those headers transitively pulls in `<limits>`, but Linux libstdc++ (gcc 13.3) does not. Until now nobody noticed because no CI job was building the C++ extension from a clean checkout — which is precisely the gap this PR's `cpp-build-check` job exists to close. The job is doing its job: it found a real, latent portability bug.

Two acceptable fixes:
1. **Preferred — fix it here.** Add `#include <limits>` to `src/kicad_tools/router/cpp/include/types.hpp` (alongside the existing `#include <cstdint>` etc.). One-line change, gets the new CI guard fully green on its first run, and removes the latent macOS-vs-Linux trap. Also bump `ROUTER_CPP_BUILD_VERSION` to 3 since the header changed (and mirror in `_REQUIRED_CPP_BUILD_VERSION`) — actually the surface didn't change so a bump isn't strictly required by your contract, but rebuilding the committed `.so` is.
2. **Alternative — split it out.** Land that one-line `<limits>` fix in a separate tiny PR first, get it onto main, then rebase this PR on top. Cleaner history but needless ceremony for a one-liner.

### What looks good

- **Design.** The hybrid Option A+C from the curator (mirrored integer constant + import-time guard routing through the existing fallback) is the right call. The PR description accurately describes the failure mode it prevents (`AttributeError` on `router_cpp.PadBounds` deep in routing).
- **Python guard.** `cpp_backend.py` lines 81-94: clean `getattr(router_cpp, "BUILD_VERSION", None)` check that handles both mismatched and missing-attr cases, sets `_CPP_AVAILABLE = False` and `router_cpp = None`, and routes through the existing fallback. The error message includes the actual mismatched value and a `kct build-native` hint, both of which are user-actionable.
- **C++ exposure.** `bindings.cpp` lines 261-264 correctly expose `BUILD_VERSION` as both a module attribute (cheap identity check) and a `build_version()` callable (parity with `version()`).
- **Tests.** `TestCppBuildVersionGuard` has good coverage: required-constant existence, happy-path version match when C++ loads, mismatched value with `monkeypatch.setattr`, and missing attribute with `monkeypatch.delattr`. The `sys.modules.pop` + `importlib.import_module` pattern correctly forces the module-level guard to re-run, and the `try/finally` properly restores the real cpp_backend so subsequent tests see truthful state. Test isolation is sound.
- **Docs.** New "Build version discipline" section in `development.md` is clear about the contract: bump in both files when the binding surface changes. Good cross-references.
- **CI guard quality.** The new `cpp-build-check` job's verification step is well-designed — it doesn't just check "did the build succeed" but also asserts `is_cpp_available()` and that `BUILD_VERSION` matches the required value. That catches the forgotten-version-bump failure mode directly. The smoke test step at the end is also a nice belt-and-suspenders.
- **Other CI jobs.** Lint, Type Check, and Test all pass.

### Minor nits (not blocking)

- `cpp_backend.py:94` uses `# type: ignore` on `router_cpp = None` to satisfy mypy. Consider `cast("Any | None", None)` or annotating the module-level `router_cpp` variable as `Any | None` upfront so the assignment is type-safe rather than ignored. Minor — current approach is fine.
- The PR description's test plan still has `[ ] CI cpp-build-check job passes on Ubuntu (will run on this PR)` unchecked. After the next push that should flip to checked.

### Action

Removing `loom:review-requested`, applying `loom:changes-requested`. Push a commit that (a) resolves the merge conflict against main (preferring main's `scripts/build-cpp.sh`) and (b) adds `#include <limits>` to `types.hpp`, then re-label `loom:review-requested` and I'll re-review.

rjwalters:
Addressed both judge re-review blockers:

**1. Merge conflict with PR #2502 on `scripts/build-cpp.sh`** — Rebased onto current `main` (now at `d7b2b01f`) and accepted main's `scripts/build-cpp.sh` verbatim. PR #2502's stronger version (with `VENV_PY` discovery block + `-DPython_EXECUTABLE`) is preserved unchanged; this PR's competing rewrite of that file has been dropped entirely.

**2. Missing `#include <limits>` in `types.hpp`** — Added in a new commit (`8b9ea8c`). `ValidationResult::min_clearance` (introduced in #2443) initializes to `std::numeric_limits<float>::infinity()` but the header was relying on Apple libc++ to pull `<limits>` in transitively. Linux libstdc++ (gcc 13 on the new CI runner) does not, which is exactly the latent portability bug the new `cpp-build-check` guard caught. One-line include fix.

Verification:
- Local clean rebuild of `router_cpp` succeeds (`scripts/build-cpp.sh build` reports `Backend: cpp / Available: True`)
- All 4 `TestCppBuildVersionGuard` tests pass
- The 5 `TestCppPathfinderPythonFallback` failures observed locally pre-exist on `main` (unrelated `MagicMock` of read-only nanobind attribute) and are not introduced by this PR

Branch force-pushed with lease. Ready for re-review.

rjwalters:
## Approved (third pass)

All four verification points confirmed:

**(a) `scripts/build-cpp.sh` matches main verbatim** — `git diff origin/main..pr-2508-review -- scripts/build-cpp.sh` returns empty. The competing rewrite was correctly dropped in favor of main's version.

**(b) `<limits>` include added to types.hpp** — Line 9 of `src/kicad_tools/router/cpp/include/types.hpp`:
```cpp
#include <cstdint>
#include <limits>
#include <vector>
```
Fixes the latent portability issue with `std::numeric_limits<float>::infinity()` (used by `ValidationResult::min_clearance` from #2443) — Apple libc++ pulled it in transitively, but Linux libstdc++ on gcc 13 does not.

**(c) `cpp-build-check` CI job passes** — completed in 44s. The new guard caught a real portability bug on its first real exercise; that is exactly the value proposition of this PR. All checks green:
- C++ Build Check: pass (44s)
- Type Check: pass (36s)
- Test: pass (5m27s)
- Lint & Format: pass (13s)

**(d) Merge state is CLEAN** — `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

The whole loop here was satisfying: PR adds a build-version guard against stale `.so`, the corresponding CI job is added to enforce it, the CI job exposes a latent `<limits>` bug on Linux, and the bug gets fixed in a one-liner before merge. Ship it.

PR #2510, opened 2026-05-04
fix(sexp): preserve quoted-string semantics on round-trip

### LINKED ISSUE #2493: sexp: serializer drops type info for string atoms that look numeric
## Problem
`SExp._needs_quoting` (`src/kicad_tools/sexp/parser.py`) returns `False` whenever `float(s)` succeeds, even when the value's Python type is `str`. This silently strips quotes from intentional string-typed values whose textual form looks numeric.

The parser preserves type info correctly:
- `(generator_version "9.0")` → `SExp(value='9.0')` (str)
- `(thickness 1.6)` → `SExp(value=1.6)` (float)

But the serializer collapses both back to the unquoted form for any string that parses as a float. KiCad's strict fields (e.g. `generator_version`) reject the unquoted form; kicad-cli 10 returns `Failed to load board`.

## Workaround in place
Surgical fix shipped today: a `_FORCE_QUOTED_STRING_VALUE = {"generator_version"}` allow-list in the parent's render path force-quotes string atom children. Scoped to one field; same bug latent for any other string-typed KiCad field with numeric-looking values.

## Expected behavior
Round-trip preserves quotedness:
```
parse('(field "9.0")').to_string() == '(field "9.0")'   # currently false without allow-list
parse('(field 9.0)').to_string()   == '(field 9.0)'     # already true
```

## Proposed fix
Track `_originally_quoted: bool` on parsed atoms (mirror of `_original_str` on parsed numerics). On serialize, respect it: if the atom was originally quoted OR if the Python value is a `str` not in `unquoted_keywords`, emit quoted. Genuine numerics passed as `int`/`float` from constructors are unaffected because they hit the numeric branch of `_format_atom`.

## Risk surface
A naive "always quote string atoms" change broke ~36 tests when piloted in this session — many fixtures and round-trip tests exact-match on the unquoted form. Migration needs to either:
- update fixtures + assertions (mechanical), or
- gate the change on the new `_originally_quoted` flag so behaviour only diverges where the parser produced a string.

## Acceptance criteria
- [ ] Round-trip property test: parse + serialize is byte-identical for all demo board PCBs and SCHs.
- [ ] `_FORCE_QUOTED_STRING_VALUE` allow-list can be removed (or shrunk to an explicit known-string-field whitelist).
- [ ] Existing `(thickness 1.6)`, `(at 0 0 0)`, `(number 1)` continue to serialize unquoted.

## Reference
Live in `src/kicad_tools/sexp/parser.py` `_needs_quoting` (~line 629) and the `_FORCE_QUOTED_STRING_VALUE` workaround added today (~line 902).

rjwalters:
## Summary

- Adds `_originally_quoted` flag to `SExp` atoms (set by `_parse_string_node`, defaults to `False`); `_format_atom` gates quoting on it so parsed-quoted strings round-trip with their quotes intact even when they look numeric.
- Removes the surgical `_FORCE_QUOTED_STRING_VALUE` allow-list (and the parent's `force_quote_atom_str` branch in `to_string`) introduced by 30256d40 -- replaced by a principled fix that handles any strict-typed KiCad string field.
- Adds `SExp.quoted_atom(value)` for programmatic call sites that need to force-quote a numeric-looking string (pad/pin numbers, `generator_version`); updates `schema/pcb.py`, `schema/library.py`, `schema/symbol.py`, `schematic/models/io_mixin.py`, `footprints/__init__.py`, `sexp/builders.py`.
- Fixes `panel/_deep_copy_sexp` to propagate both `_original_str` and `_originally_quoted` so cloned subtrees serialize identically.

Closes #2493

## Approach

This is approach A from the curator comment on #2493 -- the structural fix that mirrors the `_original_str` precedent for numeric atoms. The naive "always quote string atoms" pilot the curator referenced broke ~36 tests; this approach only flips behavior for atoms the parser actually saw quoted in the input, so existing keyword/round-trip tests remain green.

Programmatic callers that need quoted intent now have a clean opt-in via `SExp.quoted_atom("9.0")` rather than constructing a bare `SExp` and mutating the flag.

## Test plan

- [x] `parse_string('(field "9.0")').to_string(compact=True) == '(field "9.0")'`
- [x] `parse_string('(field 9.0)').to_string(compact=True) == '(field 9.0)'`
- [x] `parse_string('(field "abc")').to_string(compact=True) == '(field "abc")'`
- [x] `parse_string('(layer signal)').to_string(compact=True) == '(layer signal)'`
- [x] Programmatic `SExp.quoted_atom("9.0")` emits quoted; default `SExp(value="9.0")` emits unquoted (documented).
- [x] `tests/test_sexp_parser.py` -- 91 passed (12 new + 79 existing).
- [x] `tests/test_footprints.py`, `tests/test_schema.py`, `tests/test_pcb.py`, `tests/test_lvs.py`, `tests/test_consistency.py`, `tests/test_panel.py`, `tests/test_panel_cli.py`, `tests/test_incremental_drc.py`, `tests/test_collision_detection.py`, `tests/test_cli_pcb.py` -- 651 passed, 2 skipped.
- [x] `tests/test_drc*.py`, `tests/test_erc*.py`, `tests/test_stitch*.py` -- 805 passed, 6 skipped.
- [x] End-to-end: regenerated `boards/00-simple-led/output/simple_led.kicad_pcb` through the new serializer and `kicad-cli pcb export gerbers` accepts it (return code 0). The fix functionally subsumes the `_FORCE_QUOTED_STRING_VALUE` workaround.

### DIFF
--- a/src/kicad_tools/footprints/__init__.py
+++ b/src/kicad_tools/footprints/__init__.py
@@ -78,7 +78,14 @@ class Pad:
             )
         """
         # Build pad node: (pad "number" type shape ...)
-        pad = SExp.list("pad", self.number, self.pad_type.value, self.shape.value)
+        # Pad number is a strict-typed string in KiCad; quote it explicitly so
+        # numeric-looking numbers like "1" don't get serialized as bare ints.
+        pad = SExp.list(
+            "pad",
+            SExp.quoted_atom(self.number),
+            self.pad_type.value,
+            self.shape.value,
+        )
 
         # Position: (at x y)
         x, y = self.position
--- a/src/kicad_tools/panel/panel.py
+++ b/src/kicad_tools/panel/panel.py
@@ -846,10 +846,15 @@ def _deep_copy_sexp(node: SExp) -> SExp:
     """Create a deep copy of an S-expression tree.
 
     This avoids Python's ``copy.deepcopy`` which is slow for large
-    trees. Instead we walk the tree manually.
+    trees. Instead we walk the tree manually. Round-trip metadata
+    (_original_str for numerics, _originally_quoted for strings) is
+    propagated so the cloned subtree serializes identically.
     """
     if node.is_atom:
-        return SExp(value=node.value)
+        new_atom = SExp(value=node.value)
+        new_atom._original_str = node._original_str
+        new_atom._originally_quoted = node._originally_quoted
+        return new_atom
 
     new_node = SExp(name=node.name)
     for child in node.children:
--- a/src/kicad_tools/schema/library.py
+++ b/src/kicad_tools/schema/library.py
@@ -407,13 +407,16 @@ class LibraryPin:
         # Build effects for name and number
         font_effects = SExp.list("effects", SExp.list("font", SExp.list("size", 1.27, 1.27)))
 
+        # Pin name and number are strict-typed string fields in KiCad. Quote
+        # them explicitly so numeric-looking values (e.g. "1", "9.0") survive
+        # round-trip without being silently downgraded to bare numerics.
         children: list[SExp] = [
             SExp(value=self.type),
             SExp(value=self.shape),
             SExp.list("at", self.position[0], self.position[1], self.rotation),
             SExp.list("length", self.length),
-            SExp.list("name", self.name, font_effects),
-            SExp.list("number", self.number, font_effects),
+            SExp.list("name", SExp.quoted_atom(self.name), font_effects),
+            SExp.list("number", SExp.quoted_atom(self.number), font_effects),
         ]
 
         return SExp(name="pin", children=children)
@@ -1126,10 +1129,13 @@ class SymbolLibrary:
               (symbol ...)
             )
         """
+        # generator_version is a strict-typed string field in KiCad; emit the
+        # value as a quoted atom so kicad-cli accepts the file even though
+        # "1.0" textually parses as a number.
         children: list[SExp] = [
             SExp.list("version", 20231120),
             SExp.list("generator", "kicad_tools"),
-            SExp.list("generator_version", "1.0"),
+            SExp.list("generator_version", SExp.quoted_atom("1.0")),
         ]
 
         # Add all symbols
--- a/src/kicad_tools/schema/pcb.py
+++ b/src/kicad_tools/schema/pcb.py
@@ -1183,7 +1183,10 @@ class PCB:
         # Version and generator info
         pcb.append(SExp.list("version", 20240108))
         pcb.append(SExp.list("generator", "kicad_tools"))
-        pcb.append(SExp.list("generator_version", "9.0"))
+        # generator_version is a strict-typed string field in KiCad; emit the
+        # value as a quoted atom so kicad-cli accepts the file even though
+        # "9.0" textually parses as a number.
+        pcb.append(SExp.list("generator_version", SExp.quoted_atom("9.0")))
 
         # General settings
         pcb.append(
--- a/src/kicad_tools/schema/symbol.py
+++ b/src/kicad_tools/schema/symbol.py
@@ -24,9 +24,17 @@ class SymbolPin:
         """Convert to S-expression.
 
         Format: ``(pin "1" (uuid "..."))``
+
+        Pin number is quoted because KiCad strictly types this token as a
+        string; numeric-looking numbers like "1" must round-trip with their
+        quotes intact.
         """
         pin_uuid = self.uuid or str(uuid_mod.uuid4())
-        return SExp.list("pin", self.number, SExp.list("uuid", pin_uuid))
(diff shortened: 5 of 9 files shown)


rjwalters:
## Approved (Loom Judge)

Solid implementation of approach A from the curator comment. The structural fix (`_originally_quoted` flag mirroring the `_original_str` precedent) is the right shape for this problem and the surface area is well bounded.

### What I verified

**Parser/serializer changes (`src/kicad_tools/sexp/parser.py`)**
- `__slots__` extension and constructor default (`_originally_quoted=False`) — clean. Default-False is the key reason this doesn't regress the ~36-test naive-quote pilot.
- `_parse_string_node` sets the flag — only path that produces string atoms from quoted tokens, so coverage is complete.
- `_format_atom` gates on `self._originally_quoted or self._needs_quoting(self.value)` — preserves prior behavior for programmatic atoms while restoring type info on parsed-quoted atoms.
- `_FORCE_QUOTED_STRING_VALUE` allow-list and the `force_quote_atom_str` branch in `to_string` are both fully removed — `_atom_str` helper is gone, children render directly via `to_string(compact=True)`. The principled fix subsumes the workaround.
- `_format_value` doc comment correctly notes the flag is out of scope there and that the under-estimate only affects inline-vs-multiline, never byte content.
- `SExp.quoted_atom(value)` classmethod is a clean opt-in for programmatic callers; docstring is honest about being equivalent to `SExp.atom` when `_needs_quoting` already returns True.

**Programmatic call sites (6 updated, all justified)**
- `schema/pcb.py` `generator_version "9.0"` and `schematic/models/io_mixin.py` `generator_version "9.0"` — both directly subsume the workaround for the `kicad-cli pcb` rejection in #2493.
- `schema/library.py` — `generator_version "1.0"` plus pin `name` and `number` (strict-typed string fields, can be `"1"`/`"9.0"` etc.).
- `schema/symbol.py` `SymbolPin.to_sexp` and `sexp/builders.py` `pin_uuid_node` — same pin-number-quoting concern.
- `footprints/__init__.py` pad number — quoted explicitly so numeric pad numbers like `"1"` don't emit bare.

All comments at the call sites consistently explain the strict-typed-string rationale, which matches the issue framing.

**Panel deepcopy (`src/kicad_tools/panel/panel.py`)**
- `_deep_copy_sexp` now propagates both `_original_str` and `_originally_quoted` for atoms. This is the only SExp-specific deep-copy helper in the repo (other `copy.deepcopy` uses target unrelated objects), so coverage is complete.
- Note: it still doesn't propagate `_inline`/`_line`/`_column` — those weren't propagated before either, and they don't affect serialization correctness, so not a regression.

**Equality / repr / pickle**
- `SExp` has no custom `__eq__`/`__hash__`/`__copy__`/`__deepcopy__`/`__reduce__`. Identity-based equality means the new flag doesn't change any equality semantics. Slots-pickling will pick up the new attribute automatically.

### Tests

- New `TestOriginallyQuotedFlag` class (11 tests) exhaustively covers: default-False on programmatic atoms, parser sets True for quoted strings only, round-trip behavior for quoted-numeric / unquoted-numeric / quoted-string / unquoted-keyword cases, programmatic flag-set forces quoting, the documented `SExp.list("field", "9.0")` unquoted default, and that the flag stays False on list nodes.
- New `test_roundtrip_pcb_header_quoted_generator_version` is the regression test for #2493 itself: asserts the quoted form round-trips and the bare form is absent.
- All 91 `tests/test_sexp_parser.py` tests pass locally; targeted run of `TestOriginallyQuotedFlag` + `TestSerialization` + `TestRoundTrip` (37 tests) all pass — including all the keyword/round-trip-unquoted tests the curator comment specifically called out as the risk surface.
- Broader run of `test_sexp_parser`, `test_panel`, `test_panel_cli`, `test_footprints`, `test_schema`, `test_pcb`: 459 passed, 2 skipped, no regressions.

### End-to-end smoke

Round-trip on real demo boards (`boards/01-voltage-divider/.../voltage_divider.kicad_pcb`, `boards/05-bldc-motor-controller/.../bldc_controller.kicad_pcb`):

- `(generator_version "9.0")` preserved quoted on output.
- No bare `(generator_version 9.0)` lines emitted.
- Parse -> serialize -> parse -> serialize is byte-stable.

### CI

- Lint & Format: SUCCESS
- Type Check: SUCCESS
- Test: SUCCESS

### Minor (non-blocking)

`uv run ruff format --check tests/test_sexp_parser.py` flags two cosmetic deltas in the new section (an extra blank line and a single-quoted string `'\t(version 20240108)\n'` that ruff would prefer double-quoted). They don't affect correctness, the CI lint job is non-fatal in this repo (`continue-on-error: true`), and the same baseline shows pre-existing format/lint deltas on `main`. Worth a follow-up `ruff format .` sweep across the repo, but not blocking this PR.

Approving.

PR #2519, opened 2026-05-06
fix(route,export): fill copper-pour zones so Gerbers contain plane copper (closes #2516)

### LINKED ISSUE #2516: Zone-fill gap: auto-pour creates (zone ...) blocks but never fills them, so Gerbers have zero plane copper (all 5 boards)
## Problem

Copper-pour zones defined in `.kicad_pcb` files are not filled before Gerber export, so the gerbers contain **zero polygon-fill regions** for GND/VCC/power planes. Boards manufactured from these gerbers would have no plane copper — only thin signal traces.

This affects every board with auto-pours (01, 02, 03, 04, 05). DRC reports it consistently as `zone_unfilled` warnings.

## Reproduction

```
uv run kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb \
  -o boards/01-voltage-divider/output/voltage_divider_routed.kicad_pcb --quiet

uv run kct check boards/01-voltage-divider/output/voltage_divider_routed.kicad_pcb \
  --mfr jlcpcb --format summary
```

```
DRC Summary: voltage_divider_routed.kicad_pcb
Rule ID                        Errors   Warnings
zone_unfilled                  0        3
```

```
uv run kct export boards/01-voltage-divider/output/voltage_divider_routed.kicad_pcb \
  --mfr jlcpcb -o /tmp/zonefill_test
unzip -p /tmp/zonefill_test/gerbers/gerbers.zip voltage_divider_routed-B_Cu.gbl | grep -c "^G36"
# → 0 (zero polygon-fill regions on the back-copper plane)
```

The `(zone ...)` blocks exist in the input PCB (auto-pour created them), but no `(filled_polygon ...)` children are present, so the export has nothing to flatten into Gerber polygons.

## Audit results (2026-05-06, all 5 boards)

| Board | Zones in PCB | `zone_unfilled` warnings | Polygon-fills in B.Cu Gerber |
|---|---|---|---|
| 01-voltage-divider | 3 | 3 | 0 |
| 02-charlieplex-led | 2 | 2 | 0 |
| 03-usb-joystick | 3 | 3 | 0 |
| 04-stm32-devboard | 3 | 3 | 0 |
| 05-bldc-motor-controller | 5 | 5 | 0 |

So even board 01 — the "MFG-READY gold standard" referenced in #2394 — would ship without plane copper today.

## Why this matters

- Manufacturer fabricates exactly what is in the Gerbers. No plane copper means VCC/GND power must travel through the thin signal traces only — hugely increased impedance, no shielding, no return-current path under signals.
- For board 05 (motor controller, 15A on VMOTOR), shipping without poured planes would burn the trace immediately.
- For board 03 (USB), missing GND plane removes the return path that USB diff-pair impedance is designed for.

## Root cause

`kct route` (via `auto_pour_if_missing` in `src/kicad_tools/router/auto_pour.py`) creates the `(zone ...)` block but does not call the zone-fill operation. KiCad's GUI does this via Edit > Fill All Zones; programmatically it requires either calling KiCad's native fill engine or running `kicad-cli pcb export gerbers --plot-fills` (or equivalent).

The `kct build` step `_run_step_zones` (`src/kicad_tools/cli/build_cmd.py:846-940`) may already run the fill — needs investigation whether `kct route` and `kct export` should call the same code path.

## Acceptance

- After `kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb`, `kct check` reports 0 `zone_unfilled` warnings.
- `kct export` produces Gerbers whose B.Cu / F.Cu layers contain `G36...G37` polygon-fill regions matching the zone shapes.
- All five boards show non-zero plane copper in their exported Gerbers.
- Board 01 baseline still passes 3/3 routes + 0/0 DRC.

## Affected files

- `src/kicad_tools/router/auto_pour.py` — creates zones but does not fill them
- `src/kicad_tools/zones/generator.py` — zone generation entry point
- `src/kicad_tools/cli/build_cmd.py` — `_run_step_zones` (may already do fill, needs verification)
- `src/kicad_tools/cli/route_cmd.py` — auto-pour wiring (`auto_pour_if_missing` calls)
- `src/kicad_tools/cli/export_cmd.py` — Gerber export pipeline (could fill before export as a safety net)

## Audit reference

Discovered during the 2026-05-06 re-audit of #2394. Audit dir: `/tmp/route_audit_2026-05-06/`. This issue is the reason that #2394's "MFG-READY" gate for board 01 is currently met by name only — the board would not actually be manufacturable as expected.

rjwalters:
## Summary

Wires the existing `runner.run_fill_zones` helper into `kct route` and `kct export` so that auto-poured `(zone ...)` blocks are actually filled. Without this fix, exported Gerbers contain zero `G36..G37` polygon-fill regions and manufactured boards have no GND/VCC plane copper.

The fill machinery already existed (`run_fill_zones` in `cli/runner.py:305`, wrapped as `kct zones fill` and used by `kct pipeline`). This PR closes the wiring gap in the standalone `kct route` and `kct export` commands that the example boards' Makefiles drive.

## Two layers of defense

1. **Primary fix** in `route_cmd._fill_zones_after_route()`: runs after each of the four routing paths completes successfully (`route_with_layer_escalation`, `route_with_rule_relaxation`, `route_with_combined_escalation`, and the negotiated path in `main()`). Mirrors `pipeline_cmd._run_step_zones` -- delegates to `runner.run_fill_zones` so the existing kicad-cli net-corruption snapshot/restore guard runs, and validates net format afterwards.

2. **Safety-net fix** in `gerber.GerberExporter._export_gerbers`: detects unfilled zones via a cheap text scan and fills into a temp PCB before invoking `kicad-cli pcb export gerbers`. The user's source file is never mutated (TemporaryDirectory cleanup in a finally block).

Both call sites are no-ops when the PCB has no zones, so `--no-auto-pour` and zone-less designs are unaffected.

## End-to-end verification

Verified against board 02 (charlieplex_3x3) which has signal + power nets:

- Routed PCB: 2 zones, 2 filled_polygon entries
- Gerber B.Cu: 1 G36..G37 polygon-fill region
- Gerber F.Cu: 1 G36..G37 polygon-fill region
- Safety-net path also confirmed: exporting an unfilled PCB still produces Gerbers with G36 regions, and the source PCB stays untouched.

## Notes

- Detection uses both `(zone ` and `(zone\n` because the serializer emits the multi-line form. The original v1 of this PR missed multi-line zones; a regression test was added (`test_detects_multiline_zone_format`).
- Updated outdated comment in `build_cmd._run_step_zones` docstring (the old "Filling happens later via kicad-cli" line was the breadcrumb of this bug).
- Fill is non-trivial work (seconds per board) -- adds runtime to every `kct route` invocation. CI test-suite duration may grow on the larger boards.

## Test plan

- [x] Unit tests for `_fill_zones_after_route` (all kicad-cli states, idempotency, fill failure, net-corruption logging) -- 11 tests
- [x] Unit tests for the gerber safety-net (multi-line zones, source PCB never mutated, temp file cleanup, fill-failure fallback) -- 11 tests
- [x] All 22 new tests pass
- [x] Related test suites still pass (`test_zones_cmd.py`, `test_export_extended.py`, `test_runner_net_restore.py`, `test_route_cmd_*`) -- 258 passed, 1 pre-existing failure (`test_fill_zones_on_fixture` requires kicad-cli to load the fixture board, unchanged from main)
- [x] End-to-end smoke: `kct route` on board 02 produces filled zones, `kct export` produces Gerbers with G36 polygon-fill regions
- [x] Source PCB never mutated by the safety-net fill (verified by byte-comparison test)

### DIFF
--- a/src/kicad_tools/cli/build_cmd.py
+++ b/src/kicad_tools/cli/build_cmd.py
@@ -854,7 +854,11 @@ def _run_step_zones(ctx: BuildContext, console: Console) -> BuildResult:
     and power nets are distributed across In2.Cu / F.Cu.
 
     Zones are *defined* here (unfilled polygons).  Filling happens later
-    after routing, typically via kicad-cli.
+    after routing -- ``kct route`` calls :func:`route_cmd._fill_zones_after_route`
+    once routing succeeds, and :func:`pipeline_cmd._run_step_zones` performs
+    the same fill via ``kct zones fill``.  ``kct export`` also runs a
+    safety-net fill if the PCB still has unfilled zones at export time
+    (see :func:`export.gerber.GerberExporter._export_gerbers`).
     """
     if not ctx.pcb_file or not ctx.pcb_file.exists():
         return BuildResult(
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -597,6 +597,99 @@ def _should_auto_fix(args) -> bool:
     return True
 
 
+def _fill_zones_after_route(output_path: Path, quiet: bool = False) -> None:
+    """Fill copper-pour zones after routing completes.
+
+    Routing produces traces; copper pour zones must be filled *after* the
+    trace geometry exists so the fill polygons respect trace clearances.
+    Without this step, exported Gerbers contain ``(zone ...)`` definitions
+    but zero ``G36...G37`` polygon-fill regions, so manufactured boards have
+    no plane copper -- the bug fixed by issue #2516.
+
+    This function mirrors :func:`pipeline_cmd._run_step_zones`:
+
+    - Skips silently when ``kicad-cli`` is not installed (no-op with warning).
+    - Delegates to :func:`runner.run_fill_zones` so the existing
+      net-corruption snapshot/restore guard (``_snapshot_net_declarations`` /
+      ``_restore_net_declarations``) runs.
+    - Validates net format afterwards via :func:`runner.validate_net_format`
+      and logs a warning when corruption is detected.
+
+    The function is idempotent: filling a PCB whose zones are already filled
+    simply rewrites the same fill polygons.  When the input PCB has no
+    ``(zone ...)`` blocks at all, kicad-cli has nothing to fill and this is
+    effectively a no-op.
+
+    Args:
+        output_path: Path to the routed PCB file (modified in place).
+        quiet: Suppress informational output.
+    """
+    from kicad_tools.cli.runner import (
+        find_kicad_cli,
+        run_fill_zones,
+        validate_net_format,
+    )
+
+    kicad_cli = find_kicad_cli()
+    if kicad_cli is None:
+        if not quiet:
+            print("  Zone fill: skipped (kicad-cli not installed)")
+        return
+
+    # Quick check: does the PCB even have any zones to fill?
+    # KiCad's serializer wraps zones either as "(zone (net ...)" on one
+    # line OR as "(zone\n  (net ...)" with a newline after the opening
+    # token, so we need to detect both forms.
+    try:
+        text = output_path.read_text()
+    except OSError as exc:
+        logger.warning("Zone fill: could not read %s: %s", output_path, exc)
+        return
+    if "(zone " not in text and "(zone\n" not in text and "(zone\t" not in text:
+        # Nothing to fill -- skip silently to avoid log noise on the
+        # majority of boards that have no copper pours.
+        return
+
+    if not quiet:
+        print("\n--- Filling Copper Zones ---")
+        print(f"  Filling zones in {output_path.name}...")
+
+    result = run_fill_zones(output_path, kicad_cli=kicad_cli)
+
+    if not result.success:
+        # Non-fatal: log warning and continue.  The board may still be
+        # usable for partial inspection but the manufacturing artifacts
+        # will lack plane copper.
+        logger.warning(
+            "Zone fill failed for %s: %s",
+            output_path,
+            result.stderr or "(no stderr)",
+        )
+        if not quiet:
+            print("  Warning: zone fill failed -- Gerbers will lack plane copper")
+        return
+
+    # Validate net format after zone fill -- kicad-cli may corrupt nets.
+    report = validate_net_format(output_path)
+    if not report.valid:
+        logger.warning(
+            "Net format corruption detected after zone fill: "
+            "%d element(s) have non-canonical net format "
+            "(name_only_segments=%d, name_only_vias=%d, name_only_pads=%d, "
+            "empty_net_segments=%d, empty_net_vias=%d, empty_net_pads=%d)",
+            report.total_corrupt,
+            report.name_only_segments,
+            report.name_only_vias,
+            report.name_only_pads,
+            report.empty_net_segments,
+            report.empty_net_vias,
+            report.empty_net_pads,
+        )
+
+    if not quiet:
+        print("  Zone fill: complete")
+
+
 @dataclass
 class LayerEscalationResult:
     """Result of a layer escalation routing attempt."""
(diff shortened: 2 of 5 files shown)


rjwalters:
## Judge Review: APPROVED

Thorough review of the zone-fill wiring fix for #2516. Both layers of defense work correctly and the safety-net invariant is preserved.

### CI Status (verified)
- All 5 checks pass: Lint & Format, Type Check, Test (5m29s), C++ Build, kicad-cli Round-trip Smoke

### Scope verification
- Exactly 5 files changed, all directly tied to issue #2516 (3 src + 2 new test files, 1 commit)
- `build_cmd.py` change is documentation-only (4-line docstring update). No scope creep despite the builder's concurrent-worktree caution.

### Correctness verification

**Primary fix (`route_cmd._fill_zones_after_route`)**
- Helper definition + 4 call sites = 5 occurrences (matches the wiring test).
- All 4 call sites correctly precede `run_post_route_drc` so DRC sees filled zones (the inline comment about ordering is accurate).
- All 4 call sites guard on `nets_routed > 0`; the negotiated path additionally guards on `not args.dry_run`.
- Triple-quoted text-scan (`"(zone "`, `"(zone\n"`, `"(zone\t"`) catches both single-line and multi-line serializer forms — the regression test (`test_detects_multiline_zone_format`) locks in the fix for the bug the builder caught in v1.
- Properly delegates to `runner.run_fill_zones`, which preserves the `_snapshot_net_declarations` / `_restore_net_declarations` corruption guard. Confirmed by reading `runner.py`.

**Safety-net fix (`gerber.GerberExporter._export_gerbers`)**
- I read `runner._run_fill_zones_via_drc` and confirmed that when `output_path` is provided, the source PCB is `shutil.copy2`'d first and DRC runs on the copy. So the temp-file pattern in `gerber.py` is genuinely safe.
- `tmpdir` is `None`-initialized and only assigned when `_pcb_has_unfilled_zones` is True; `finally` correctly guards on `tmpdir is not None`.
- `_export_gerbers_impl` correctly uses the passed `pcb_path` parameter for the `kicad-cli` invocation (line 425). The remaining `self.pcb_path` references at the file are only inside `ExportError.context` dicts where reporting the user's original path is the right UX.

### Acceptance gates (verified locally with kicad-cli 10.0.1)

End-to-end test on board 02 (charlieplex_3x3):
- `kct route /tmp/board02_input.kicad_pcb` → produces routed PCB with `2 zones, 2 filled_polygon entries`
- `kct check ... --mfr jlcpcb --format summary` → **0 zone_unfilled warnings** (only the pre-existing 8 pad_grid warnings remain)
- `kct export ...` → Gerbers contain `1 G36..G37 region` on each of B.Cu and F.Cu
- DRC PASSED via the route's auto-DRC step

**Safety-net invariant verified**:
- Synthesized an unfilled PCB by stripping `(filled_polygon ...)` blocks from the routed file
- `kct export` produced Gerbers with G36 polygon-fill regions on B.Cu and F.Cu
- Source PCB md5 byte-identical before/after export → invariant HELD

### Test coverage
- 22 new tests in `test_gerber_zone_fill.py` (11) and `test_route_cmd_zone_fill.py` (11), all pass locally
- The wiring test (`test_helper_is_invoked_from_all_route_functions`) is a nice grep-style guard against future refactors dropping a call site
- Related suites (`test_zones_cmd.py`, `test_export_extended.py`, `test_runner_net_restore.py`) → 194 passed, 1 failure
- The 1 failure (`test_zones_cmd::TestFillIntegration::test_fill_zones_on_fixture`) is pre-existing on main, confirmed by checking out main and re-running. Unrelated to this PR.

### Notes (informational, not blockers)
- 4 ruff errors and 2 format issues in `route_cmd.py` / `gerber.py` exist on **main** as well — not introduced by this PR. The repo-wide tech debt is tracked separately (see audit history). CI's Lint & Format job somehow passes despite this; that's a separate CI scoping issue.
- The mypy errors in `route_cmd.py` are also pre-existing.
- Runtime cost note (added by the builder) is real and worth watching: zone fill via DRC on every `kct route` adds seconds per board. For board 02 the route + fill took ~6.5 minutes locally. CI test-suite duration on the larger boards may grow.

Approving — all acceptance gates met, source PCB safety invariant proven, no regressions detected.

Issue #2307, opened 2026-04-29
fix(router): per-net timeout not forwarded in two-phase rip-up loop


rjwalters:
## Problem Statement

In `TwoPhaseRouter._detailed_negotiated()` at `src/kicad_tools/router/algorithms/two_phase.py` line ~369, the rip-up iteration loop calls \`_route_net_with_corridor(net, present_factor)\` without forwarding the \`per_net_timeout\` parameter. The timeout was correctly plumbed through \`route_all()\` and into \`_detailed_negotiated()\` (PR #2302), but the rip-up path was missed.

This means nets re-routed during rip-up iterations have no timeout, while the initial pass correctly enforces the timeout. A single difficult net in rip-up can stall the entire routing process.

## Evidence

The parameter is accepted by \`_detailed_negotiated(per_net_timeout=...)\` but the call at line ~369 in the rip-up loop does not pass it through:
\`\`\`python
# Line ~369 (rip-up loop)
route = self._route_net_with_corridor(net, present_factor)
# Missing: per_net_timeout parameter
\`\`\`

Compare with the initial pass which does forward the timeout correctly.

## Recommended Enhancement

Forward \`per_net_timeout\` to all \`_route_net_with_corridor()\` calls in the rip-up loop, matching the initial pass behavior.

## Acceptance Criteria

- [ ] \`per_net_timeout\` is passed to \`_route_net_with_corridor()\` in the rip-up iteration loop
- [ ] Routing timeout behavior is consistent between initial pass and rip-up iterations

rjwalters:
## Curation Finding: May Already Be Fixed

**Investigation**: The issue states that `_route_net_with_corridor()` in the rip-up loop (line ~369) does not forward `per_net_timeout`. However, examining the current code at `src/kicad_tools/router/algorithms/two_phase.py`, **both** call sites already pass the parameter:

- **Line 322 (initial pass)**: `self._route_net_with_corridor(net, present_factor, per_net_timeout=per_net_timeout)` 
- **Line 376 (rip-up loop)**: `self._route_net_with_corridor(net, present_factor, per_net_timeout=per_net_timeout)` 

`git blame` confirms line 376 was updated in commit `2d1affe56` (PR #2302, merged today), which is the same PR cited in this issue as having "missed" the rip-up path. It appears PR #2302 actually did include the fix for the rip-up path.

**Recommended action**: Human verification needed. If the fix is confirmed present, this issue can be closed as already resolved.

[...]

Issue #2305, opened 2026-04-29
feat(router): track best routing state across two-phase iterations


rjwalters:
## Problem Statement

`TwoPhaseRouter._detailed_negotiated()` keeps the final iteration's routing state, not the best state observed. On chorus-test-revA, overflow oscillated: 2→5→2→14→5→2 across iterations. If the router stopped or timed out during a high-overflow iteration, the result would be worse than an earlier iteration that achieved overflow=2.

## Evidence

Two-phase routing on chorus-test-revA showed non-monotonic convergence:
- Iterations 3, 7, 10: overflow=2 (best observed)
- Iterations 4-6: overflow spiked to 5-14
- Final state depends on which iteration the timeout/limit hits

Without best-state tracking, the router can return a worse result than it achieved mid-run.

## Recommended Enhancement

1. After each iteration in `_detailed_negotiated()`, snapshot the current routes and overflow count
2. Track `best_overflow` and `best_routes` across iterations
3. When routing terminates (timeout, iteration limit, or convergence), return the best state, not the final state
4. Log when the best state differs from the final state

## Acceptance Criteria

- [ ] Router tracks best overflow and corresponding route set across iterations
- [ ] Final output uses best-observed state, not last-iteration state
- [ ] Log message when best state != final state (e.g., "Returning iteration 7 state (overflow=2) instead of final iteration 10 (overflow=5)")

rjwalters:
## Implementation Guidance

### Root Cause

`_detailed_negotiated()` (lines 279-390 in `src/kicad_tools/router/algorithms/two_phase.py`) mutates `net_routes` and `self.routes` in-place during each rip-up-and-reroute iteration. When routing terminates (timeout, iteration limit, or convergence), the method returns `list(self.routes)` -- whatever state the last iteration left behind. If overflow oscillates (e.g., 2 -> 5 -> 2 -> 14 -> 5 -> 2), a timeout during a high-overflow iteration discards the earlier better solution.

### Approach

After each iteration's rerouting completes and `overflow` is computed (line 384), compare against the tracked best. When the current overflow is strictly less than `best_overflow`, snapshot the state:

1. **Track variables**: Add `best_overflow = overflow` (after initial pass, line 330) and `best_routes: list[Route] = []`, `best_net_routes: dict[int, list[Route]] = {}`.

[...]

PR #2522, opened 2026-05-06
fix(router): hoist wall-clock timeout into per-net inner loop on negotiated strategy (closes #2518)

### LINKED ISSUE #2518: router: --timeout enforced post-iteration on negotiated strategy, allowing 2x overruns
## Summary

The router's `--timeout` flag is checked between iterations, not within an iteration's per-net loop. On the negotiated strategy with a real-world dense board, asking for 900s budget produced a 1756.6s wall-time run before the timeout fired — a 1.95× overrun. This makes iterative router work expensive and unpredictable.

#2467 fixed this for the GA strategy specifically. Negotiated strategy still has the same issue.

## Repro

```bash
uv run kicad-tools route /tmp/stripped_synced.kicad_pcb \
  -o /tmp/routed.kicad_pcb \
  --strategy negotiated --layers 4 --backend cpp \
  --manufacturer jlcpcb --no-cache \
  --timeout 900 --per-net-timeout 60
```

(Same setup as the chorus-test repro in the related issue.)

Observed log line:
```
⚠ Timeout during detailed routing at net 33/36 (630.6s)
Initial pass: 29/36 nets, overflow: 43
⚠ Timeout at iteration 1 (1756.6s)
```

The first warning fires at 630s (within budget) but the routing loop keeps running until 1756.6s.

## Why this matters

- Iterative router development becomes slow — promised \"fail fast in 15min\" runs eat 30+ min
- CI test budgets become unreliable
- When chained with multi-board test sweeps (e.g. all 5 demo boards back-to-back), overruns compound

## Suspected location

Likely the same pattern as the GA issue: the `_route_iteration_negotiated` (or similar) loop in `src/kicad_tools/router/algorithms/negotiated.py` checks the wall-clock only at iteration boundaries. The check needs to be hoisted into the per-net loop, with the partial result preserved if the budget is exhausted mid-iteration.

## Acceptance hint

`--timeout 900` should produce a wall-clock run no longer than ~1000s (small grace for cleanup/output write). Mid-iteration partial state should still be preserved as the partial output.

rjwalters:
## Summary

Fixes #2518: `--timeout` was only enforced at iteration boundaries on the negotiated strategy's two-phase / hierarchical paths. Hitting the budget inside the per-net inner loop only `break`-ed one level — control fell through overflow recompute, history snapshot, and into a brand-new iteration before the iteration-boundary check finally caught it. In the chorus-test-revA repro, `--timeout 900` produced a 1756.6s wall-clock run (1.95x overrun, ~1080s tail = ~18 nets * 60s `per_net_timeout`).

The fix mirrors the GA strategy's pattern from #2467 / PR #2470: a `timed_out` local flag set inside the inner break that propagates up to short-circuit the outer iteration loop too. Partial-state preservation (#2305 best-state restore) continues to work because the iteration loop exits cleanly and the post-loop restore block runs.

## Code paths fixed

- **`src/kicad_tools/router/algorithms/two_phase.py`** — primary repro path (`_detailed_negotiated()`). Added `timed_out` flag in initial pass + reroute inner loop, gated entry into rip-up section on `not timed_out`, and added post-inner-loop short-circuit before overflow recompute / history snapshot.
- **`src/kicad_tools/router/algorithms/hierarchical.py`** — same fix in `_detailed_negotiated()`. Also forwards `per_net_timeout` to `_route_net_with_corridor()` (this was a separate gap — echo of #2307 which fixed it for two-phase).
- **`src/kicad_tools/router/core.py::route_all_negotiated()`** — already had the `timed_out` flag for both inner reroute branches but the iteration tail could still run `escape_local_minimum` (which can take tens of seconds per strategy). Added an explicit `if timed_out: break` before the oscillation/escape strategies block. Also extended `route_all_hierarchical()` to plumb through the new `per_net_timeout` kwarg.
- **`tests/test_negotiated_timeout_propagation.py`** — new file with 6 tests covering both `TwoPhaseRouter` and `HierarchicalRouter`. Uses a monkeypatched `time.time` clock to drive the budget deterministically; verifies that hitting the budget mid-iteration prevents the next iteration from starting and skips the iteration-body bookkeeping.

## Acceptance gate (per issue)

- `--timeout 900` should produce wall-clock <= ~1000s on chorus-test-revA. Pre-fix: 1756.6s. The fix removes the ~1080s tail-iteration overrun, so the new wall-clock should land well within the 1.15x grace window.
- Mid-iteration partial state preserved (the existing #2305 best-state restore block at `two_phase.py:510-529` continues to run after the iteration loop exits via the new propagating break).
- `timeout=None` behavior unchanged on boards 01-05 (verified by the new `test_no_timeout_runs_all_iterations` test that drives the loop to `max_iterations` when `timeout=None`).

## Test plan

- [x] Unit tests pass: 6 new tests in `test_negotiated_timeout_propagation.py`, all 12 best-state tests in `test_best_state_tracking.py` still pass, and 149 total in the related routing test sweep.
- [x] Lint clean on changed files (the F401 errors flagged on `two_phase.py` for `format_failed_nets_summary` and `Corridor` are pre-existing — confirmed by `git diff` showing those imports are unchanged).
- [ ] Manual chorus-test-revA repro to confirm wall-clock <= 1000s with `--timeout 900` (requires test fixture not in this repo).
- [ ] Board 01 baseline: not regressed (fix only changes timeout-path behavior; no logic changes when `timeout=None`).

## Related issues

- #2518 — this fix
- #2467 / PR #2470 — same gap fixed for the GA strategy; provides the template
- #2307 — fixed `per_net_timeout` not being forwarded in two-phase rip-up loop; same gap existed in hierarchical (now fixed here)
- #2305 — best-state preservation in two-phase iterations; preserved by this change

### DIFF
--- a/src/kicad_tools/router/algorithms/hierarchical.py
+++ b/src/kicad_tools/router/algorithms/hierarchical.py
@@ -73,6 +73,7 @@ class HierarchicalRouter:
         use_negotiated: bool = True,
         progress_callback: ProgressCallback | None = None,
         timeout: float | None = None,
+        per_net_timeout: float | None = None,
     ) -> list[Route]:
         """Route all nets using hierarchical global-to-detailed flow.
 
@@ -82,7 +83,9 @@ class HierarchicalRouter:
             corridor_width_factor: Corridor width as multiple of clearance (default: 2.0)
             use_negotiated: Use negotiated congestion routing in detailed phase
             progress_callback: Optional callback for progress updates
-            timeout: Optional timeout in seconds
+            timeout: Optional global wall-clock timeout in seconds
+            per_net_timeout: Optional per-net A* timeout forwarded to
+                ``_route_net_with_corridor`` (Issue #2518; mirrors #2307).
 
         Returns:
             List of Route objects (may be partial if timeout reached)
@@ -216,6 +219,7 @@ class HierarchicalRouter:
                 progress_callback=progress_callback,
                 timeout=timeout,
                 start_time=start_time,
+                per_net_timeout=per_net_timeout,
             )
         else:
             detailed_routes = self._detailed_standard(
@@ -259,8 +263,16 @@ class HierarchicalRouter:
         progress_callback: ProgressCallback | None,
         timeout: float | None,
         start_time: float,
+        per_net_timeout: float | None = None,
     ) -> list[Route]:
-        """Detailed phase using negotiated congestion routing."""
+        """Detailed phase using negotiated congestion routing.
+
+        Issue #2518: ``per_net_timeout`` is now forwarded to the per-net
+        ``_route_net_with_corridor`` calls (echo of #2307 fix for two-phase),
+        and a ``timed_out`` flag propagates from the inner per-net break up
+        through the iteration loop so the wall-clock budget is enforced
+        immediately rather than at the next iteration boundary.
+        """
         from ..algorithms import NegotiatedRouter
 
         def check_timeout() -> bool:
@@ -274,12 +286,17 @@ class HierarchicalRouter:
         net_routes: dict[int, list[Route]] = {}
         present_factor = 0.5
 
+        # Issue #2518: propagating timeout flag (matches the pattern in
+        # core.py::route_all_negotiated and TwoPhaseRouter._detailed_negotiated)
+        timed_out = False
+
         # Initial routing pass
         for i, net in enumerate(net_order):
             if check_timeout():
                 flush_print(
                     f"  Timeout during detailed routing at net {i}/{total_nets}"
                 )
+                timed_out = True
                 break
 
             if progress_callback is not None:
@@ -288,7 +305,9 @@ class HierarchicalRouter:
                 if not progress_callback(progress, f"Routing {net_name}", True):
                     break
 
-            routes = self._route_net_with_corridor(net, present_factor)
+            routes = self._route_net_with_corridor(
+                net, present_factor, per_net_timeout=per_net_timeout
+            )
             if routes:
                 net_routes[net] = routes
                 for route in routes:
@@ -300,8 +319,10 @@ class HierarchicalRouter:
             f"  Initial pass: {len(net_routes)}/{total_nets} nets, overflow: {overflow}"
         )
 
-        # Rip-up and reroute if overflow remains
-        if overflow > 0:
+        # Rip-up and reroute if overflow remains.
+        # Issue #2518: skip the iteration loop entirely if the initial pass
+        # was already cut short by the wall-clock budget.
+        if overflow > 0 and not timed_out:
             max_iterations = 10
             history_increment = 1.0
             present_factor_increment = 0.5
@@ -309,6 +330,7 @@ class HierarchicalRouter:
             for iteration in range(1, max_iterations + 1):
                 if check_timeout():
                     flush_print(f"  Timeout at iteration {iteration}")
+                    timed_out = True
                     break
 
                 if progress_callback is not None:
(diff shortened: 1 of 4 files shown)


rjwalters:

### CI status
All 5 checks pass: Lint & Format, Type Check, C++ Build Check, Test, kicad-cli Round-trip Smoke.
Merge state: CLEAN, mergeable: MERGEABLE.

### Local verification
- 6/6 new tests pass in `tests/test_negotiated_timeout_propagation.py`
- 109/109 related router tests pass (`test_best_state_tracking`, `test_two_phase_iteration_resolution`, `test_two_phase_net_count`, `test_router_autorouter`, plus the new file)
- `ruff check` / `ruff format --check` errors on the changed files are pre-existing on `main` (verified by checkout-main comparison) — same 21 lint errors, same 3 files needing reformat. Not introduced by this PR.
- `mypy` errors on `two_phase.py` / `hierarchical.py` are pre-existing on `main` (same 18 errors) — not introduced by this PR.

### Correctness of the fix

**Two-phase `_detailed_negotiated`** (`algorithms/two_phase.py:333-566`)
The propagating `timed_out` flag is correctly placed:
- Set inside the initial-pass `check_timeout()` break (line 405)
- Set inside the iteration-boundary `check_timeout()` break (line 445)
- Set inside the inner per-net reroute break (line 491)
- Outer iteration loop guarded by `if overflow > 0 and not timed_out:` (line 438) so an initial-pass timeout skips the entire iteration loop
- Post-inner-loop short-circuit at lines 509-513 skips overflow recompute, history snapshot, best-state snapshot, convergence check, and early-stop check — exits the iteration loop cleanly
- Best-state restore at lines 545-564 still runs after the loop exits (partial state preserved per #2305)

**Hierarchical `_detailed_negotiated`** (`algorithms/hierarchical.py:263-385`)
Same pattern, correctly mirrored. Additionally fixes the missing `per_net_timeout` forwarding to `_route_net_with_corridor()` — an echo of #2307 that was left unfixed in the hierarchical path. Test `test_per_net_timeout_forwarded` verifies this.

**`core.py::route_all_negotiated`** (`core.py:4639-4646`)
The new `if timed_out: break` is correctly positioned:
- AFTER the convergence check at line 4634 (so converged-and-timed-out still uses the convergence path)
- BEFORE the `escape_local_minimum` block (so timeout skips the expensive escape strategies)
- Post-loop stages (final summary print, `_post_route_clearance_correction` already gated on `not timed_out`, return statement) all still run — partial state preserved

**`route_all_hierarchical` plumbing** — `per_net_timeout` kwarg added and forwarded; non-breaking default of `None`.

### Risk assessment
- **`timeout=None` regression**: Tested in `test_no_timeout_runs_all_iterations` — confirmed iterations run to `max_iterations`.
- **Partial state preservation**: Tested in `test_partial_routes_preserved_on_timeout` and verified by inspection of the post-loop best-state restore at `two_phase.py:545-564`.
- **Board 01 baseline**: No logic changes when `timeout=None`; the new `timed_out` flag stays False throughout, so the iteration loop runs identically to pre-fix.
- **Wall-clock acceptance gate**: Not directly verified (chorus-test fixture not in repo), but the unit tests deterministically verify the iteration-skipping behavior that produces the expected ~1080s tail elimination. The fix mirrors the proven #2467/#2470 GA-strategy template.

- The `core.py` `if timed_out: break` lacks a dedicated unit test (would require driving the full `route_all_negotiated` flow). Low risk because it's a 2-line addition consistent with the surrounding `timed_out` flag pattern already in the function (lines 4427, 4502, 4562 already use `if timed_out: break`).
- Minor scope addition: the hierarchical `per_net_timeout` plumbing is technically a separate fix (echo of #2307), but it is contextually tied — same file, same issue, same propagation pattern. Builder calls this out explicitly. Reasonable scope.

Approving for merge.

PR #2524, opened 2026-05-06
fix(router): stagnation recovery for never-routed nets (closes #2515)

rjwalters:
## Summary

Board 05 (BLDC motor controller) terminated with 10/26 nets routed and the same 4 nets (PHASE_A/B/C, SW_OUT) re-routing every iteration with overflow oscillating in [5, 2, 4, 7, 4]. Two issues compounded:

1. **Single-pin "ghost" nets** (16 of 26 on board 05: GATE_*, HALL_*, ISENSE_*-, SWD signals) had only one pad in `self.nets` because the netlist is structurally incomplete. They polluted `total_nets` and the early-termination diagnostic reported them as "unrouted" without explaining why. Now filter them out before the routing loop and announce the exclusion by name.

2. **Non-zero overflow stagnation** had no recovery path. The existing zero-overflow recovery at `core.py:3966` only fires when overflow reaches 0; on board 05 overflow stayed > 0 the whole time, so the per-net stall detector eventually marked PHASE_A/B/C + SW_OUT all stalled and the loop terminated with "No nets to rip up". Add a new recovery path that tracks the rip-up cohort across iterations and, when the same cohort persists for 3 consecutive iterations with overflow stable in a narrow band:
   - re-enables previously stalled nets so they aren't orphaned,
   - rips up the cohort plus same-tier destination siblings (e.g. other PHASE_* on shared connector J2), and
   - re-routes them with elevated `present_factor` to bias A* toward unexplored corridors.

3. Improve the early-termination diagnostic to surface unrouted and partially-routed nets by name, not just counts.

## Result on board 05

- **Before:** 10/26 routed, ~80s wall time, no named diagnostic
- **After:**  10/10 routable nets routed (4 partial), ~30-110s, 16 single-pin ghosts surfaced explicitly by name, partial nets named at termination

The trace-width spec sourcing (HIGH_CURRENT_SIGNAL = 2.0mm) and J2 neck-down geometry are deferred to a sibling issue per curator's recommendation in #2515. Per the curator's analysis, the stall fix and the trace-width work touch independent code paths and can land separately.

## Test plan

- [x] `uv run kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb -o ... --timeout 240 --quiet`: completes in ~30s instead of 80s, with all 10 routable nets in the output
- [x] DRC clean (0 errors, only pre-existing pad_grid + zone_unfilled warnings)
- [x] Board 01 baseline still passes (3/3 routes, 0 DRC errors)
[...]

PR #2525, opened 2026-05-06
feat(check): detect single-pad nets as design defects (missing footprint / schematic drift) (closes #2521)

### LINKED ISSUE #2521: Pipeline: detect and report single-pad nets as design defects (missing footprint / schematic drift)
## Problem

A net with only one pad cannot connect anything — by definition it is a design defect (missing footprint, schematic-PCB drift, or wrong pin mapping). Today the pipeline treats these silently:

- `kct check` does not flag them.
- `kct erc` does not flag them.
- `kct route` (after PR #2520, closing #2514) classifies them as "structurally unroutable" and skips them with a single info line — the user has to read the route log to notice.
- `kct export` produces Gerbers anyway.

This was discovered while building #2514: board 04 has 4 single-pad nets (SWDIO, SWCLK, SWO, NRST) because the MCU footprint was never placed — they connect only to the SWD header J1. The router correctly cannot route them, but nothing in the pipeline tells the user "your design is missing the MCU footprint."

For an AI-driven workflow where an agent is meant to iteratively repair a design, this silent skip is especially bad: the agent finishes a route, sees a "successful" result, exports Gerbers, and never learns the design is incomplete.

## Proposal

Detect single-pad multi-pin nets in a preflight stage and surface them as **errors** with actionable diagnostics. Suggested places:

1. **`kct check` (preferred)** — add a `single_pad_net` rule. For each non-power, non-no-connect net with `len(pads) == 1`, report the net name, the lone pad's component+pad number, and a one-line diagnosis: *"Net 'X' has only 1 pad on Y — likely missing footprint or schematic/PCB drift."* This is consistent with how other DRC rules work and naturally feeds into the manufacturing audit.

2. **`kct route` preflight** — refuse to route (or warn loudly) when single-pad nets exist on the input PCB, pointing at the exact nets and components. Today's silent skip should at minimum become a top-of-output warning block.

3. **`kct audit` (manufacturing readiness)** — fail the audit when single-pad nets are present, since such a board is not manufacturable as designed.

The detection itself is trivial — group pads by net, count, filter `count == 1`. The actionable part is the diagnostic message: which component is missing, or which pad number doesn't exist on the footprint that does exist.

## Why this matters for the AI workflow

- A builder/doctor agent looking at "13/13 nets routed" with a passing DRC will declare the board done, even though 4 SWD signals are floating.
- A curator agent investigating routing failures (as happened on #2514) needs to descend two levels into the router log to find the root cause; a preflight error would surface it immediately.
- A roadmap-tracking agent (#2394) cannot accurately count "manufacturer-ready boards" if structurally-broken designs pass all gates.

## Acceptance

- New rule `single_pad_net` in `kct check` produces an **error** entry per single-pad multi-pin net, with the lone pad's component reference and a diagnosis hint.
- `kct route` emits a top-of-output warning when single-pad nets exist on the input PCB, recommending the user run `kct check` before routing.
- `kct audit --mfr jlcpcb` fails on a board with any single-pad signal nets.
- Power/ground nets that legitimately have only one pad on a board (e.g., a single test point) are handled cleanly — only nets that look like signal nets but have one endpoint should error.
- Board 04 (current state, with no MCU placed) reports 4 single-pad-net errors when checked.

## Affected files

- `src/kicad_tools/check/` — `kct check` rule registration; add the new `single_pad_net` rule
- `src/kicad_tools/cli/route_cmd.py` — preflight warning before the route loop
- `src/kicad_tools/cli/audit_cmd.py` — manufacturing audit gate
- `src/kicad_tools/router/core.py` — already detects single-pad nets after PR #2520; the preflight should run before this point so the user sees them at `kct check` time, not buried in the route log

## Cross-references

- PR #2520 (closes #2514) added internal classification of single-pad nets but only emits a one-line info message during routing.
- Issue #2394 — board 04 is one of the example boards. Adding this preflight will explicitly call out that board 04's MCU placement is incomplete, which is consistent with the README's "schematic-only" disclaimer.

rjwalters:
## Summary

Closes #2521.

Adds a new DRC rule `single_pad_net` that flags signal nets attached to only one pad as errors. These nets are structurally unroutable and almost always indicate a missing footprint, schematic/PCB drift, or stale netlist. Three logical changes ship behind a single detector:

- **`kct check --only single_pad_net <pcb>`**: reports one error per single-pad signal net with the offending pad's `<ref>-<pad>` and a one-line diagnosis hint.
- **`kct route`**: emits a top-of-output WARNING block listing affected nets and recommending `kct check`. Fires from all three routing entry points (auto-layer escalation, max-layer try, legacy single-attempt) to cover every path.
- **`kct audit --mfr jlcpcb`**: returns `NOT_READY` automatically -- `_check_drc()` already routes errors into `DRCStatus.blocking_count`, no source change in `auditor.py` was needed.

Pour nets (POWER/GROUND) are silently allowed because a single test point or pour-only net is a legitimate design pattern. Suppression uses `kicad_tools.router.net_class.classify_and_apply_rules` -- the same name-pattern classifier `auditor._check_connectivity()` already uses for pour-net detection.

`ViolationType.SINGLE_PAD_NET` is added with a `CONNECTIVITY` category mapping and a `from_string` alias so JSON output round-trips cleanly (without the alias, `to_dict()` would resolve `type` to `"unknown"` and break downstream consumers like `drc.report._parse_kct_check_json`).

## Verified end-to-end results

| Board | Expected | Got |
|---|---|---|
| 01-voltage-divider | 0 errors | 0 |
| 02-charlieplex-led | 0 errors | 0 |
| 03-usb-joystick | 0 errors | 0 |
| 04-stm32-devboard (no MCU) | 4 errors (SWDIO/SWCLK/SWO/NRST) | 4 |
| 05-bldc-motor-controller (MCU placeholder) | (curator: 0; actual: 16) | 16 |

Board 05's MCU is a placeholder in the schematic generator (per `boards/05-bldc-motor-controller/design.py`), so its 16 single-pad nets (gate-driver outputs, current-sense returns, Hall sensor inputs, SWD) are real findings -- exactly the defect class this rule is meant to catch. The integration test excludes board 05 with an explanatory docstring noting why.

## Test plan

- [x] Unit tests in `tests/test_validate_single_pad_net.py` mirroring the `test_validate_netlist.py` stub pattern (10 cases covering all-multi-pad, single signal, pour-net suppression for GND/VCC/+3.3V, mixed signal+pour, empty/zero net edge cases, multiple signal nets, JSON type round-trip).
- [x] Integration tests in `tests/test_single_pad_net_integration.py` against board 04 (`check_all`, `check_single_pad_nets`, audit verdict NOT_READY) plus parametrized regression guard against boards 01/02/03.
- [x] CLI tests in `tests/test_cli_check_single_pad_net.py` for `kct check --only`, `--skip`, and `--format json` against board 04.
- [x] Subprocess test in `tests/test_route_single_pad_warning.py` verifying `kct route` prints the SWDIO/SWCLK/SWO/NRST banner.
- [x] All 167 existing tests in the affected modules (validate/netlist, audit, cli/check, route_cmd) still pass.
- [x] `ruff check` and `ruff format --check` clean on all 10 changed/new files.
- [x] mypy clean on the new rule file (pre-existing mypy errors in `route_cmd.py` and `validate/checker.py` are unrelated).

### DIFF
--- a/src/kicad_tools/cli/check_cmd.py
+++ b/src/kicad_tools/cli/check_cmd.py
@@ -65,6 +65,7 @@ CHECK_CATEGORIES = [
     "pad_grid",
     "placement",
     "silkscreen",
+    "single_pad_net",
     "solder_mask",
     "zones",
 ]
@@ -274,6 +275,7 @@ def run_selected_checks(
         "pad_grid": checker.check_pad_grid_alignment,
         "placement": checker.check_footprint_placement,
         "silkscreen": checker.check_silkscreen,
+        "single_pad_net": checker.check_single_pad_nets,
         "solder_mask": checker.check_solder_mask_pads,
         "zones": checker.check_zones,
     }
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -69,6 +69,70 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _emit_single_pad_net_warning(
+    router: "Autorouter",
+    single_pad_nets: list[int],
+) -> None:
+    """Print a top-of-output warning when single-pad signal nets exist.
+
+    These nets are structurally unroutable -- the router silently skips
+    them, which makes "13/13 routed, DRC clean" look like a successful
+    build even when 4 SWD signals are floating.  Pour nets (POWER /
+    GROUND) are silently allowed because a single test point or
+    pour-only net is a legitimate design pattern.
+
+    See ``kct check --only single_pad_net`` for the full DRC-style
+    report (this banner exists to surface the defect early in the
+    pipeline; the actionable error lives in the check command).
+
+    Args:
+        router: The Autorouter (used for ``router.net_names`` lookup).
+        single_pad_nets: Net IDs that have exactly one pad assigned.
+    """
+    if not single_pad_nets:
+        return
+
+    from kicad_tools.cli.progress import flush_print
+
+    try:
+        from kicad_tools.router.net_class import classify_and_apply_rules
+
+        single_pad_name_map = {
+            net_num: router.net_names.get(net_num, "") for net_num in single_pad_nets
+        }
+        # Drop unnamed nets so we don't classify the empty string.
+        single_pad_name_map = {num: name for num, name in single_pad_name_map.items() if name}
+        pour_rules = classify_and_apply_rules(single_pad_name_map) if single_pad_name_map else {}
+        signal_single_pad_names = sorted(
+            name
+            for num, name in single_pad_name_map.items()
+            if not (pour_rules.get(name) and pour_rules[name].is_pour_net)
+        )
+    except Exception:
+        # Conservative: if classification blows up, surface every named
+        # single-pad net so we don't hide real defects.
+        signal_single_pad_names = sorted(
+            router.net_names.get(num, "")
+            for num in single_pad_nets
+            if router.net_names.get(num, "")
+        )
+
+    if not signal_single_pad_names:
+        return
+
+    flush_print("")
+    flush_print(
+        f"  WARNING: {len(signal_single_pad_names)} single-pad signal "
+        "net(s) detected (likely missing footprint or schematic/PCB drift):"
+    )
+    for name in signal_single_pad_names:
+        flush_print(f"    - {name}")
+    flush_print(
+        "  Run 'kct check --only single_pad_net <pcb>' for details. "
+        "Routing will proceed but these nets cannot be connected."
+    )
+
+
 def _insert_sexp_before_closing(pcb_content: str, sexp_fragments: str) -> str:
     """Insert S-expression fragments before the final closing parenthesis of a PCB file.
 
@@ -182,8 +246,7 @@ def _finalize_routes(
         )
         if vias_removed > 0:
             flush_print(
-                f"  Vias:     {pre_cleanup_vias} -> {post_cleanup_vias} "
-                f"({vias_removed} removed)"
+                f"  Vias:     {pre_cleanup_vias} -> {post_cleanup_vias} ({vias_removed} removed)"
             )
         if cleanup_stats.get("segments_restored", 0) > 0:
             flush_print(
@@ -740,9 +803,7 @@ def update_pcb_layer_stackup(pcb_content: str, target_layers: int) -> str:
         new_layers = "\n    ".join(layer_defs[target_layers])
         # Append non-copper layers after copper layers
         if non_copper_entries:
-            non_copper_lines = "\n".join(
-                entry.strip() for entry in non_copper_entries
-            )
+            non_copper_lines = "\n".join(entry.strip() for entry in non_copper_entries)
             return f"(layers\n    {new_layers}\n    {non_copper_lines}\n  )"
         return f"(layers\n    {new_layers}\n  )"
 
@@ -792,16 +853,11 @@ def _print_power_stall_suggestions(
     pour_arg = ",".join(pour_assignments)
 
     print()
-    print(
-        f"Routing did not complete: {nets_csv} could not be routed on "
-        f"{layer_count} layer(s)."
-    )
+    print(f"Routing did not complete: {nets_csv} could not be routed on {layer_count} layer(s).")
     print("Suggestions:")
-    print(f"  1. Add copper zones for power nets:  --power-nets \"{pour_arg}\"")
+    print(f'  1. Add copper zones for power nets:  --power-nets "{pour_arg}"')
     print("  2. Increase layer count:              --layers 4 (or --max-layers 6)")
-    print(
-        f"  3. Manual routing in KiCad for the {len(stalled_nets)} remaining net(s)"
-    )
+    print(f"  3. Manual routing in KiCad for the {len(stalled_nets)} remaining net(s)")
     print()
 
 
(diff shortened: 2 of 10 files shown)


rjwalters:
## Judge Review — APPROVED

Excellent work. This PR is well-scoped, addresses the curator's critical concerns, and maintains a clean separation between the DRC rule (the canonical detector) and the route-time warning (UX-only banner). Below is a section-by-section assessment of the items called out for review.

### 1. Curator-flagged ViolationType enum gotcha — VERIFIED CORRECT

All three pieces are present in `src/kicad_tools/drc/violation.py`:
- Enum member: `SINGLE_PAD_NET = "single_pad_net"` (line 162)
- Alias entry: `"single_pad_net": cls.SINGLE_PAD_NET` (line 238)
- Category mapping: `ViolationType.SINGLE_PAD_NET: ViolationCategory.CONNECTIVITY` (line 82)

Note: `ViolationType.from_string()` already attempts a direct enum-value match before consulting the alias table, so the alias entry is technically redundant for `"single_pad_net"`. Adding it anyway is defensive and harmless — and the explicit unit test (`test_to_dict_resolves_violation_type`) plus integration test (`test_json_output_resolves_type`) both verify the round-trip resolves to `single_pad_net` rather than `unknown`. The downstream `drc.report._parse_kct_check_json` consumer is correctly protected.

### 2. Pour-net suppression robustness — KNOWN-LIMITATION, ACCEPTED

The `+3V3` naming convention (no decimal) is **not** classified as POWER by `classify_and_apply_rules`:

```
'+3V3': is_pour_net=False
'+3.3V': is_pour_net=True
'VCC':   is_pour_net=True
'GND':   is_pour_net=True
```

The POWER regex `^[+-]?\\d+\\.?\\d*V[ADPS]?$` requires the form `<digits>[.<digits>]V[ADPS]?` — `+3V3` has the V before the trailing 3 and is rejected. A single-pad `+3V3` testpoint would therefore trip a false-positive `single_pad_net` error on this board.

I am not blocking on this because:
- This regex is the **same** classifier `auditor._check_connectivity()` already uses for pour-net detection. Fixing it would change behavior in shipping code and is out of scope.
- The unit tests use `+3.3V` (which does match), the integration assertions use only board 04's SWD signals, and the regression boards use `+3V3` only in zones — not as single-pad testpoints — so no false positives are exposed in the test fixtures.
- This is best filed as a follow-up issue against `router/net_class.py::NET_CLASS_PATTERNS` to extend the POWER regex with `^[+-]?\\d+V\\d+$` (and similar `+1V8`, `+5V0` variants).

**Suggested follow-up:** open a small enhancement issue to extend the POWER regex; not required for this PR.

### 3. Three-way warning helper extraction — CLEAN

`_emit_single_pad_net_warning()` is a textbook DRY refactor — no behavior bundling. All three callsites pass the same `(router, single_pad_nets)` arguments after computing `single_pad_nets` identically:

```python
single_pad_nets = [
    net_num for net_num, pads in router.nets.items() if net_num > 0 and len(pads) == 1
]
```

The only difference is at site 3 (`main()` line ~3583), where `single_pad_nets` is built in a single pass alongside `multi_pad_nets` and `power_nets_skipped` because those values feed downstream summary/reporting (`summary_parts`, `print_routing_diagnostics_json`, `show_routing_summary`). That divergence is correct, not a defect — the helper still receives the same argument shape. The `try/except` fallback inside the helper that surfaces *every* named single-pad net if `classify_and_apply_rules` blows up is appropriately conservative.

Minor stylistic note (non-blocking): `signal_single_pad_names = sorted(...)` is computed twice in the `try` and `except` branches with subtly different logic. Consider hoisting the empty-name filter to a helper or computing both branches into the same intermediate. Not worth churning for.

### 4. Audit-gate test coverage — PRESENT

`tests/test_single_pad_net_integration.py::TestBoard04SwdSignals::test_audit_verdict_not_ready` (line 86–116):
- Builds `ManufacturingAudit` against board 04
- Calls `auditor._check_drc(pcb)` directly
- Asserts `drc_status.blocking_count >= 4`
- Asserts `drc_status.violations_by_type.get("single_pad_net", 0) == 4`
- Asserts `AuditResult.verdict == AuditVerdict.NOT_READY`

This is exactly what the curator recommended — it pins the rule's effect on the audit gate without requiring the full project structure (.kicad_pro, schematic, etc.).

### 5. Conflict with PR #2524 — NONE

PR #2524 touches `src/kicad_tools/router/core.py` (single-pin filter at top of `route_all_negotiated`); PR #2525 touches `src/kicad_tools/cli/route_cmd.py` plus the validate/DRC layers. No file overlap, no semantic conflict. When #2524 lands first:
- Router will print `"Excluding N single-pin net(s) from routing..."` from inside `route_all_negotiated`
- CLI will additionally print the `_emit_single_pad_net_warning` banner from `route_cmd.py` before routing begins

Both fire harmlessly; the user gets clearer messaging from both layers. No merge conflict expected.

### 6. CI status

- Lint & Format: pass
- Type Check: pass
- C++ Build Check: pass
- Test: still pending at review time (~2h CI runtime is normal here)

Local verification I ran:
- `uv run ruff check` on the 10 changed files: clean
- `uv run ruff format --check` on the 10 changed files: clean
- `uv run mypy --ignore-missing-imports src/kicad_tools/validate/rules/single_pad_net.py src/kicad_tools/drc/violation.py`: clean
- `pytest tests/test_validate_single_pad_net.py tests/test_single_pad_net_integration.py tests/test_cli_check_single_pad_net.py tests/test_route_single_pad_warning.py`: 20/20 pass
- `pytest tests/test_drc_violation_type_validate.py tests/test_validate_netlist.py tests/test_cli_check.py tests/test_drc_category.py tests/test_audit.py`: 280/280 pass (no regression)

### 7. Minor observations (non-blocking)

- The integration test correctly excludes board 05 (the BLDC controller) with an explanatory docstring noting that its 16 single-pad-net errors are a real finding (placeholder MCU). This nicely matches what PR #2524 independently discovered.
- The unit test stubs use a `_StubNet` dataclass that is built into `_StubPCB._nets` but never actually consulted by the rule (the rule walks `pcb.footprints` instead). The stubs work because they expose the same `.nets` attribute the netlist rule uses, so they future-proof against re-using the rule infrastructure with a different detector. Fine as-is.

### Verdict

Approving. The curator's critical gotcha is correctly addressed and tested, the audit-gate test the curator recommended is present, the helper extraction is clean, no conflict with PR #2524, all 20 new tests pass, no regression in 280 related tests. The `+3V3` pour-net classifier limitation is real but pre-existing in shared code; recommend a small follow-up issue against `router/net_class.py::NET_CLASS_PATTERNS`.

Approved via label change — moving to `loom:pr` for merge.

PR #2544, opened 2026-05-07
fix(router): preserve best-of-iterations saved-partial result; iter-1 timeout no longer destroys iter-0 successes (closes #2540)

### LINKED ISSUE #2540: router: preserve best-of-iterations as saved-partial result; iteration-1 timeout currently destroys iteration-0 successes
## Problem

The router's saved-partial result can drop routes that succeeded in earlier iterations when a later iteration's rip-up is aborted by timeout. Specifically: in `src/kicad_tools/router/core.py:4762-4809`, the `successful_nets` count is computed from `net_routes` (the in-flight working state) while the saved output is `self.routes`. When an iteration-1 rip-up cycle aborts mid-flight, iteration-0 routes that were already successful appear to be destroyed before the saved-partial path captures them.

## Discovery

Surfaced by the judge of PR #2535 (board 05 BLDC, DRV8301 HTSSOP-56). Two consecutive runs of `kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb -o /tmp/x.kicad_pcb --timeout 240 --layers 2 --quiet --no-cache` produced:

- In-flight: 20/26 (76.9%) — what the negotiator reports as best mid-iteration
- Saved-partial: 10/26 (38.5%) — what actually lands in the output PCB

Reproducing the gap on demand: kick off the same command, observe the in-flight log line e.g. `Routed 20/26 nets, overflow: ...` mid-iteration, then watch the final output drop to 10. The doctor of PR #2535 saw 15/26 in their run; it varies but is consistently well below the in-flight peak.

## Why it matters

- Users see "router routed N/M nets" mid-run and expect that result. Getting a much lower count in the output is misleading.
- For the example boards used in the #2394 roadmap audit, this affects every dense board with mid-iteration timeouts (board 03, 05, and the chorus-test set).
- The saved-partial logic was introduced specifically to preserve progress on aborted iterations (#2305 and predecessors). It currently does not achieve that on the iteration boundary path that matters most.

## Suspected fix

Either:
1. Track best-of-iterations explicitly in `route_all_negotiated` — snapshot `self.routes` whenever the in-flight `successful_nets` exceeds the previous best — and restore it before returning if the final iteration's saved state is worse.
2. Make iteration-1 rip-up not destroy iteration-0 routes until the rip-up completes; only commit the rip-up if the new iteration's state is at least as good.

Option 1 is more conservative and matches how `_detailed_negotiated` in `algorithms/two_phase.py:545-564` handles best-state restore. The same pattern should apply to `route_all_negotiated`.

## Acceptance

- After this fix, `kct route boards/05-bldc-motor-controller/.../bldc_controller.kicad_pcb --timeout 240 --layers 2 --quiet` produces a saved output count that matches or exceeds the highest in-flight `Routed N/M` line printed during the run
- Board 01 baseline still passes (3/3 routes, 0 DRC)
- Same regression test for board 03 (12/13 with PR #2526 + PR #2536, 13/13 in fresh pass)

## Affected files

- `src/kicad_tools/router/core.py:4762-4809` — `route_all_negotiated` saved-partial logic
- `src/kicad_tools/router/algorithms/two_phase.py:545-564` — existing best-state restore pattern to mirror

## Cross-references

- Surfaced by: PR #2535 re-judge (https://github.com/rjwalters/kicad-tools/pull/2535#issuecomment-4395089286)
- Existing best-state restore pattern: PR #2522 (closes #2518) — timeout propagation work that introduced the two-phase restore
- Related: #2305 (CLOSED) — saved-partial introduction

rjwalters:
## Summary

- Mirrors the #2305 best-state-restore pattern from ``two_phase.py`` into ``Autorouter.route_all_negotiated`` (``core.py``) and ``HierarchicalRouter._detailed_negotiated`` (``algorithms/hierarchical.py``) so a mid-iteration timeout no longer drops iteration-0 successes from the saved-partial output.
- Two adjustments vs the two_phase port: comparison metric is **route count** (not overflow — the user-visible regression is route count), and snapshot happens at the **top of each iteration** before the destructive ``rip_up_nets`` (since the inner timeout break escapes before iteration-end bookkeeping).
- Adds 5 deterministic tests in ``tests/test_route_all_negotiated_partial_state.py`` covering the hierarchical path with the same fake-fixture pattern as ``test_best_state_tracking.py`` plus a ``_FakeClock``-driven mid-iteration timeout test.

## Why

On board 05 (BLDC), ``kct route ... --timeout 240 --layers 2 --quiet`` printed an in-flight peak of ``Routed 20/26`` but the saved output collapsed to 10/26.  Tracing this back showed that ``rip_up_nets`` in ``algorithms/negotiated.py:762-781`` destructively clears both ``net_routes`` and ``self.routes`` BEFORE re-routing.  When ``check_timeout()`` fires inside the per-net reroute loop, ``self.routes`` is left mid-rip-up (only the few rerouted) while ``net_routes`` and ``successful_nets`` describe a different snapshot.  PR #2522 fixed the timeout-propagation flag but not this best-state restore gap (which was never ported from two_phase to either hierarchical or ``route_all_negotiated``).

## Acceptance gates from issue

- Saved partial >= in-flight peak on board 05 with the described command (the snapshot/restore guarantees this by construction; integration verification noted below).
- Hierarchical strategy fixed in the same PR (was the same gap).
- Restore log line emitted: ``Restoring iteration N state (routed=X) instead of final (routed=Y)``.
- Comparison metric is route count, not overflow.
- ``tests/test_best_state_tracking.py`` (two_phase) — non-regressed (12/12 still pass).
- New deterministic timeout-aborted iteration test added (5 new tests in the new module).

## Test plan

- [x] ``uv run pytest tests/test_best_state_tracking.py`` — 12/12 pass (two_phase semantics unchanged).
- [x] ``uv run pytest tests/test_route_all_negotiated_partial_state.py`` — 5/5 pass.
- [x] ``uv run pytest tests/test_negotiated_timeout_propagation.py`` — 6/6 pass (PR #2522 tests still pass).
- [x] ``uv run pytest tests/test_negotiated_congestion_model.py tests/test_router_negotiated_high_current.py tests/test_negotiated_timeout_propagation.py`` — 50/50 pass.
- [x] ``uv run ruff check`` on modified files — no new errors introduced (pre-existing core.py lint issues unchanged).
- [x] ``uv run mypy`` on modified files — no new errors introduced.
- [ ] Manual board 05 repro: ``kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb -o /tmp/x.kicad_pcb --timeout 240 --layers 2 --quiet --no-cache`` — verify saved-partial route count >= in-flight peak.
- [ ] Board 01 baseline: 3/3 routes, 0 DRC violations.
- [ ] Board 03: 12-13/13 (depending on fresh-pass variability).

### DIFF
--- a/src/kicad_tools/router/algorithms/hierarchical.py
+++ b/src/kicad_tools/router/algorithms/hierarchical.py
@@ -11,6 +11,7 @@ before performing detailed routing. The flow is:
 
 from __future__ import annotations
 
+import copy
 import time
 from typing import TYPE_CHECKING
 
@@ -319,6 +320,18 @@ class HierarchicalRouter:
             f"  Initial pass: {len(net_routes)}/{total_nets} nets, overflow: {overflow}"
         )
 
+        # Issue #2540: Track best-of-iterations so a mid-iteration timeout
+        # does not destroy successful routes from earlier iterations.
+        # Mirrors the snapshot/restore pattern in
+        # ``core.py::route_all_negotiated`` (and the original
+        # ``two_phase.py::_detailed_negotiated`` from #2305).  PR #2522 only
+        # ported the ``timed_out`` propagation flag; the best-state restore
+        # was never ported and this branch had the identical gap.
+        best_routes: list[Route] = copy.deepcopy(list(self.routes))
+        best_net_routes: dict[int, list[Route]] = copy.deepcopy(net_routes)
+        best_routed_count = sum(1 for r in net_routes.values() if r)
+        best_iteration = 0  # 0 = initial pass
+
         # Rip-up and reroute if overflow remains.
         # Issue #2518: skip the iteration loop entirely if the initial pass
         # was already cut short by the wall-clock budget.
@@ -332,6 +345,16 @@ class HierarchicalRouter:
                     flush_print(f"  Timeout at iteration {iteration}")
                     timed_out = True
                     break
+
+                # Issue #2540: Snapshot at top of iteration (BEFORE the
+                # destructive ``rip_up_nets``).  Use route count as the
+                # comparison metric.
+                current_routed = sum(1 for r in net_routes.values() if r)
+                if current_routed > best_routed_count:
+                    best_routed_count = current_routed
+                    best_routes = copy.deepcopy(list(self.routes))
+                    best_net_routes = copy.deepcopy(net_routes)
+                    best_iteration = iteration - 1
 
                 if progress_callback is not None:
                     progress = 0.8 + 0.15 * (iteration / max_iterations)
@@ -384,6 +407,26 @@ class HierarchicalRouter:
                     flush_print(f"  Overflow resolved at iteration {iteration}")
                     break
                 overflow = new_overflow
+
+        # Issue #2540: Restore best-of-iterations state if a later iteration
+        # was aborted mid-rip-up (typically by ``check_timeout()`` in the
+        # per-net reroute loop) and left ``net_routes`` with fewer routes
+        # than a prior iteration produced.
+        current_routed = sum(1 for r in net_routes.values() if r)
+        if best_routed_count > current_routed:
+            flush_print(
+                f"  Restoring iteration {best_iteration} state "
+                f"(routed={best_routed_count}) instead of final "
+                f"(routed={current_routed})"
+            )
+            for route in list(self.routes):
+                self.grid.unmark_route_usage(route)
+            self.routes.clear()
+            self.routes.extend(best_routes)
+            for route in self.routes:
+                self.grid.mark_route_usage(route)
+            net_routes.clear()
+            net_routes.update(best_net_routes)
 
         # Collect all routes
         all_routes: list[Route] = []
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -3639,6 +3639,7 @@ class Autorouter:
         Returns:
             List of routes (may be partial if timeout reached)
         """
+        import copy
         import time
 
         start_time = time.time()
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge Review: APPROVED

Tight, well-targeted fix that mirrors the proven `two_phase.py` #2305 pattern and closes the identical gap in both `core.py::route_all_negotiated` and `hierarchical.py::_detailed_negotiated`.

### Verification of curator-flagged concerns

1. **Deep copy correctness**: Confirmed. `copy.deepcopy(list(self.routes))` + `copy.deepcopy(net_routes)` at iteration top. `Route` is a regular dataclass with primitive/list fields, so deepcopy semantics are correct. `import copy` is function-local in core.py (matching the existing `import time` style) and module-level in hierarchical.py.

2. **Grid state ordering**: Restore block correctly orders (a) unmark current `self.routes`, (b) clear + extend with `best_routes`, (c) re-mark on grid. Identical to the battle-tested `two_phase.py:553-561` pattern. No double-marks or stale references.

3. **`net_routes` synchronization**: `net_routes.clear()` + `update(best_net_routes)` keeps `self.routes` and `net_routes` in lockstep, addressing the original divergence bug.

4. **Initial-pass-only path**: Safe no-op. `best_*` initialized AFTER initial pass and BEFORE iteration loop, so if the loop never runs, `best_routed_count == current_routed` and restore is skipped.

5. **Hierarchical parity**: Identical pattern to core.py; only cosmetic log-prefix differences matching each file's existing style.

### Empirical reproduction (board 05, `--timeout 60`)

```
Rerouted 4/4 nets, overflow: 4 (10.4s)
Progress: 10/26 nets routed total
...
Restoring iteration 0 state (routed=10) instead of final (routed=2)
  Successful: 10
```

Without the fix, the saved-partial would have been **2** routes (mid-rip-up state); with the fix, it's restored to **10** routes. This matches the issue's reported regression signature (in-flight high count vs. saved low count).

### Test results

- `tests/test_route_all_negotiated_partial_state.py`: 5/5 pass
- `tests/test_best_state_tracking.py`: 12/12 pass (two_phase unchanged)
- `tests/test_negotiated_timeout_propagation.py`: 6/6 pass
- Broader negotiated/router/hierarchical/two_phase/partial_state sweep: equivalent to main (5 pre-existing failures on main, plus 1 transient isolation flake in `test_pad_blocks_other_net_in_negotiated_mode` that passes when run in isolation on both branches; not introduced by this PR)

### CI / quality gates

- All 5 CI checks green (Lint & Format, Type Check, Test, C++ Build, kicad-cli Round-trip)
- No new ruff or mypy violations on PR-touched lines
- Diff +592/-0 (no removals, additive only)

The route-count metric (vs. overflow) is correct per the curator's analysis: a half-restored rip-up can leave overflow numerically identical while the actual route count drops. The new test `test_grid_state_consistent_after_restore` validates that the grid's marked-routes state is consistent with the returned route list.

Approving for merge.

PR #2553, opened 2026-05-07
feat(router): auto-build C++ backend on first use when router_cpp.so is missing (closes #2549)

### LINKED ISSUE #2549: kct: auto-build C++ router extension on first use when router_cpp.so is missing
## Problem

Every fresh checkout (or every checkout where `git clean` has run) starts without `src/kicad_tools/router/router_cpp.*.so`. The first `kct route` invocation falls back to pure Python with this warning:

```
C++ router backend not available -- using pure Python (10-100x slower).
Reason: C++ router extension not built (no router_cpp.*.so found in
/Users/rwalters/GitHub/kicad-tools/src/kicad_tools/router). Run: kct build-native
```

Today (2026-05-07) we hit this twice and lost ~30+ minutes to slow Python routes before noticing and running `kct build-native` (which takes ~30s). The user explicitly flagged this with "we should always compile the c++ backend!"

## Why it matters

- 10-100x speedup is enormous for any meaningful router work
- Silent fallback means users notice only after a slow run completes
- For agents (builders, doctors), the time penalty is multiplied by the number of route attempts; the failed #2542 builder spent significant runtime on Python routes before the user pointed this out
- A 30-second build is cheap; missing it is expensive

## Proposal

Three options, in order of intrusiveness:

**Option 1 (least intrusive): Auto-build on first detection.** When `kct route` finds `router_cpp.*.so` missing, attempt to run `kct build-native` automatically (with a clear message), then proceed with the C++ backend. If the build fails, fall back to Python with the existing warning. Skip auto-build when `--backend python` is explicitly requested.

**Option 2: CI/dev-setup hook.** Add `kct build-native` to `make setup` / `pnpm install` / `uv sync --extra dev` post-hooks so it runs on every fresh dev environment.

**Option 3: Pre-commit / git-hook.** Verify `router_cpp.*.so` is buildable on every commit that touches `src/kicad_tools/router/cpp/`. Doesn't help with fresh checkouts but catches binding regressions early.

Recommend Options 1 + 2 together. Option 1 is the user's safety net; Option 2 is the dev-team default.

## Acceptance

- A user running `kct route` on a fresh checkout with no `router_cpp.so` either gets a working C++ backend (auto-build succeeded) or an informative message that `kct build-native` was attempted but failed (with the underlying reason)
- The first-route-of-day case: 30s build vs 10× slower route, ~5× net win even on simple boards
- Tests: a unit test stubs the `.so` absence and verifies auto-build is attempted

## Affected files

- `src/kicad_tools/router/cpp_backend.py` (lines ~67-85 — the import-and-fail block per PR #2520's diagnostic improvement)
- `scripts/build-cpp.sh` (the build script that `kct build-native` ultimately calls)
- `pyproject.toml` (post-install hook config, if going with Option 2)
- `.pre-commit-config.yaml` (if going with Option 3)

## Cross-references

- User feedback memory: `feedback_compile_cpp_backend.md` ("Always compile the C++ router backend") — this is essentially that feedback turned into a tooling fix
- PR #2520 added the actionable diagnostic message; this issue makes the actionable step automatic
- PR #2501 (CLOSED) handled stale .so detection but not missing .so

rjwalters:
## Summary

Closes #2549.

`kct route` on a fresh checkout silently fell back to pure-Python (10-100x
slower) because the C++ extension `router_cpp.*.so` is never built by
`pip install` / `uv sync`. Today's user session lost 30+ minutes to this
before noticing.

This PR implements **Option 1 (silent auto-build)** with the four guardrails
recommended by the curator:

1. Auto-build only when `--backend` is `auto` (the default) AND the `.so` is
   missing or stale -- never when the user explicitly chose `--backend python`.
2. Cheap pre-check for cmake + clang++/g++ on PATH; skip the build entirely
   when the toolchain is absent (e.g. sandboxes, CI runners without dev tools).
3. `KICAD_TOOLS_NO_AUTO_BUILD=1` environment opt-out + a new
   `--no-auto-build-native` CLI flag.
4. Build failures (any reason: timeout, permission error, verification
   failure, missing nanobind) fall through to the existing Python-fallback
   warning path. `kct route` never crashes due to a failed auto-build.

## What changed

**`src/kicad_tools/router/cpp_backend.py`**
- New `ensure_cpp_backend_available(backend, quiet, allow_auto_build)` helper.
- New `_reload_cpp_backend()` that re-imports the module after a successful
  build so module-level globals reflect the freshly written `.so`.
- New `_attempt_auto_build()`, `_toolchain_available()`,
  `_auto_build_allowed_by_env()` private helpers.

**`src/kicad_tools/cli/route_cmd.py`**
- Replaces all four "Handle backend selection" blocks (in
  `route_with_layer_escalation`, `route_with_rule_relaxation`,
  `route_with_combined_escalation`, and `main()`) with a single call to the
  helper.
- Adds `--no-auto-build-native` flag to the local parser block.

**`src/kicad_tools/cli/parser.py`**
- Adds `--no-auto-build-native` to the `route` subparser (the flag is
  defined in two parser locations).

**`src/kicad_tools/router/__init__.py`**
- Exports `ensure_cpp_backend_available`.

**`tests/test_router_cpp_auto_build.py`** (new, 22 tests)
- One test per acceptance gate (AG1-AG14). All pass.

## Acceptance gates (AG1-AG14)

| Gate | Status | Notes |
|---|---|---|
| AG1 | covered | Auto-build runs when `.so` missing + toolchain present |
| AG2 | covered | Missing toolchain skips build silently |
| AG3 | covered | Already-loaded backend short-circuits without calling `build_native` |
| AG4 | covered | `--backend python` never triggers auto-build |
| AG5 | covered | `--backend cpp` with build failure returns exit code 1 |
| AG6 | covered | `KICAD_TOOLS_NO_AUTO_BUILD=1` opts out (case-insensitive) |
| AG7 | covered | Stale `.so` (BUILD_VERSION mismatch) goes through same path |
| AG8 | covered | All four `route_cmd.py` call sites use the helper (static check) |
| AG9 | covered | Helper lives in `cpp_backend.py` |
| AG10 | covered | Test stubs availability + toolchain and asserts build_native called |
| AG11 | covered | Test asserts missing-toolchain skips build_native |
| AG12 | covered | Test parametrized over truthy env values |
| AG13 | covered | Tests cover success=False, TimeoutExpired, PermissionError |
| AG14 | covered | Static test counts the four invocations |

## Test plan

- [x] `uv run pytest tests/test_router_cpp_auto_build.py` -> 22 passed
- [x] `uv run pytest tests/test_router_cpp_backend.py tests/test_route_cmd_params.py tests/test_route_cmd_power_nets.py tests/test_route_exit_codes.py` -> 156 passed
- [x] `uv run ruff format` clean on touched files
- [x] `uv run kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb -o /tmp/x.kicad_pcb --max-layers 2` -> SUCCESS, 3/3 nets routed (board 01 baseline)
- [x] `uv run kct route --help` shows new `--no-auto-build-native` flag

## Notes

- Build wall-clock is ~30s on Apple Silicon. The helper emits one
  informational line `"C++ router extension missing -- building (~30s,
  one-time)..."` before invoking `build_native()`.
- Curator's recommendation to defer Option 2 (post-install hook) and
  Option 3 (pre-commit hook) is honored -- both out of scope for this PR.

### DIFF
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -2297,6 +2297,18 @@ def _add_route_parser(subparsers) -> None:
         ),
     )
     route_parser.add_argument(
+        "--no-auto-build-native",
+        action="store_true",
+        help=(
+            "Disable silent auto-build of the C++ router extension on first use "
+            "(Issue #2549). When --backend is 'auto' (the default) and the compiled "
+            "router_cpp.*.so is missing, kct route normally invokes 'kct build-native' "
+            "once and uses C++ for the rest of the session. Pass this flag (or set "
+            "KICAD_TOOLS_NO_AUTO_BUILD=1) to skip the build attempt and fall straight "
+            "through to pure Python."
+        ),
+    )
+    route_parser.add_argument(
         "--strict",
         action="store_true",
         help=(
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -1058,31 +1058,19 @@ def route_with_layer_escalation(
     from kicad_tools.router import (
         DesignRules,
         LayerStack,
-        is_cpp_available,
+        ensure_cpp_backend_available,
         load_pcb_for_routing,
         show_routing_summary,
     )
 
-    # Handle backend selection
-    force_python = False
-    if args.backend == "cpp":
-        if not is_cpp_available():
-            print(
-                "Error: C++ backend requested but not available.\n"
-                "Build the C++ extension or use --backend auto/python.\n"
-                "See README for build instructions.",
-                file=sys.stderr,
-            )
-            return 1
-    elif args.backend == "python":
-        force_python = True
-
-    # Warn prominently when auto-backend falls back to Python
-    if args.backend == "auto" and not is_cpp_available() and not quiet:
-        flush_print("WARNING: C++ router backend not installed -- using Python (10-100x slower).")
-        flush_print("  Build it now:  kct build-native")
-        flush_print("  Check status:  kct build-native --check")
-        flush_print()
+    # Handle backend selection (auto-build C++ extension on first use; #2549)
+    ok, force_python, exit_code = ensure_cpp_backend_available(
+        backend=args.backend,
+        quiet=quiet,
+        allow_auto_build=not getattr(args, "no_auto_build_native", False),
+    )
+    if not ok:
+        return exit_code if exit_code is not None else 1
 
     # Configure design rules
     fine_pitch_cl = getattr(args, "fine_pitch_clearance", None)
@@ -1603,33 +1591,21 @@ def route_with_rule_relaxation(
     from kicad_tools.router import (
         DesignRules,
         LayerStack,
+        ensure_cpp_backend_available,
         get_relaxation_tiers,
-        is_cpp_available,
         load_pcb_for_routing,
         show_routing_summary,
     )
     from kicad_tools.router.io import detect_layer_stack
 
-    # Handle backend selection
-    force_python = False
-    if args.backend == "cpp":
-        if not is_cpp_available():
-            print(
-                "Error: C++ backend requested but not available.\n"
-                "Build the C++ extension or use --backend auto/python.\n"
-                "See README for build instructions.",
-                file=sys.stderr,
-            )
-            return 1
-    elif args.backend == "python":
-        force_python = True
-
-    # Warn prominently when auto-backend falls back to Python
-    if args.backend == "auto" and not is_cpp_available() and not quiet:
-        flush_print("WARNING: C++ router backend not installed -- using Python (10-100x slower).")
-        flush_print("  Build it now:  kct build-native")
-        flush_print("  Check status:  kct build-native --check")
-        flush_print()
+    # Handle backend selection (auto-build C++ extension on first use; #2549)
+    ok, force_python, exit_code = ensure_cpp_backend_available(
+        backend=args.backend,
+        quiet=quiet,
+        allow_auto_build=not getattr(args, "no_auto_build_native", False),
+    )
+    if not ok:
+        return exit_code if exit_code is not None else 1
 
     # Parse skip nets
     skip_nets = []
@@ -2098,32 +2074,20 @@ def route_with_combined_escalation(
     from kicad_tools.router import (
         DesignRules,
         LayerStack,
+        ensure_cpp_backend_available,
         get_relaxation_tiers,
-        is_cpp_available,
         load_pcb_for_routing,
         show_routing_summary,
     )
 
-    # Handle backend selection
-    force_python = False
-    if args.backend == "cpp":
-        if not is_cpp_available():
-            print(
-                "Error: C++ backend requested but not available.\n"
-                "Build the C++ extension or use --backend auto/python.\n"
-                "See README for build instructions.",
-                file=sys.stderr,
-            )
-            return 1
-    elif args.backend == "python":
-        force_python = True
-
-    # Warn prominently when auto-backend falls back to Python
-    if args.backend == "auto" and not is_cpp_available() and not quiet:
-        flush_print("WARNING: C++ router backend not installed -- using Python (10-100x slower).")
-        flush_print("  Build it now:  kct build-native")
-        flush_print("  Check status:  kct build-native --check")
-        flush_print()
+    # Handle backend selection (auto-build C++ extension on first use; #2549)
+    ok, force_python, exit_code = ensure_cpp_backend_available(
+        backend=args.backend,
+        quiet=quiet,
+        allow_auto_build=not getattr(args, "no_auto_build_native", False),
+    )
+    if not ok:
+        return exit_code if exit_code is not None else 1
 
     # Parse skip nets
     skip_nets = []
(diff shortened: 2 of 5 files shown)


rjwalters:
## Judge Review: APPROVED

Closes #2549. Excellent execution against the curator's enrichment requirements. Approving with confidence.

### Verification of acceptance gates

**AG8 — single shared helper (the critical refactor):** Confirmed. All four backend-selection sites in `route_cmd.py` route through `ensure_cpp_backend_available()`:
- `route_with_layer_escalation` line 1067
- `route_with_rule_relaxation` line 1602
- `route_with_combined_escalation` line 2084
- `main()` line 3515

`git grep -n "ensure_cpp_backend_available" src/kicad_tools/cli/route_cmd.py` shows 8 hits (4 imports + 4 calls). The old inline blocks (`Error: C++ backend requested but not available`, `WARNING: C++ router backend not installed`) are **completely gone** from `route_cmd.py`. Static check `test_all_route_cmd_call_sites_use_helper` enforces this invariant going forward.

**AG4 — `--backend python` opt-out:** Verified live. Helper returns `(True, True, None)` immediately without invoking `is_cpp_available()` or auto-build. Test `test_python_skips_auto_build` covers it.

**AG5 — `--backend cpp` hard error:** Verified. When user explicitly requests cpp and build fails, helper returns `(False, False, 1)` with the original error message preserved. Test `test_cpp_with_failed_build_returns_exit_1` covers the build-failure-still-errors path.

**AG10 — auto idempotency / negligible overhead:** Verified live. Two consecutive calls with `_CPP_AVAILABLE=True` complete in <0.01ms each. `build_native(force=False)` short-circuits in `build_native_cmd.py:198` when `is_cpp_available()` returns True.

**AG13 — graceful failure:** Three independent paths covered with explicit tests:
- `test_build_failure_falls_through_to_python` (BuildResult.success=False)
- `test_build_timeout_falls_through_to_python` (TimeoutExpired)
- `test_build_permission_error_falls_through_to_python` (PermissionError)
The bare `except Exception` at `cpp_backend.py:322` is the right choice here — auto-build must never crash routing.

**AG14 — `KICAD_TOOLS_NO_AUTO_BUILD=1` opt-out:** Verified live. Truthy set `{1, true, yes, on}` (case-insensitive) opts out; `{"", "0", "false", "no", "off"}` does not. Test parametrizes both directions.

**Stale .so / wrong cpython tag:** Test `test_stale_so_triggers_rebuild` covers the BUILD_VERSION mismatch path from #2501. The helper's behavior is identical to the missing-`.so` case because both manifest as `is_cpp_available() == False`. Correct integration.

**Toolchain pre-check:** `_toolchain_available()` is the right cheap gate — `shutil.which("cmake")` plus `shutil.which("clang++") or shutil.which("g++")`. Avoids spending 5-10s invoking cmake on sandboxes/CI runners that will never succeed.

**Module reload after build:** `_reload_cpp_backend()` correctly handles the in-process reimport gotcha — without it, `_CPP_AVAILABLE` would stay False even after a successful build because Python caches the first failed import. The fallback warning ("re-run kct to use C++") for the rare case where reload doesn't pick up the new `.so` is a good defensive touch.

### CI status

All five CI checks pass:
- Lint and Format: pass (12s)
- Type Check: pass (36s)
- C++ Build Check: pass (55s)
- Test: pass (5m49s)
- kicad-cli Round-trip Smoke: pass (41s)

Merge state: CLEAN.

### Local validation

- 22 new tests in `tests/test_router_cpp_auto_build.py` all pass (17.10s)
- No new ruff or mypy regressions introduced (the 10 lint and 9 mypy errors in `cpp_backend.py` and 9 lint errors in `router/__init__.py` are all pre-existing on main; the parser.py format issue is also pre-existing — none touched by this PR)
- Live behavior smoke test: helper short-circuits in <0.01ms when backend already loaded; `--backend python` returns force_python=True instantly; env-var opt-out responds to `1`, `TRUE`, `yes`, etc.
- 5 unrelated test failures in `TestCppPathfinderPythonFallback` (nanobind read-only attribute issue) and 1 in `test_voltage_divider_high_completion` (signal nets count) — all pre-existing on main, not regressions.

### Code quality observations

- Single 100-line helper consolidating 4 copy-pasted blocks is a clean win — the diff is now `-110/+657` and the bulk of `+` is the helper, helpers, docstrings, and the comprehensive test file
- All four call sites use identical kwargs (`backend=args.backend, quiet=..., allow_auto_build=not getattr(args, "no_auto_build_native", False)`) — the slight difference at `main()` (uses `getattr(args, "quiet", False)` instead of bare `quiet`) is correct because `main()` doesn't have a local `quiet` variable in scope
- `--no-auto-build-native` flag added to both subcommand parser and top-level parser, so it works in both old-style `kct route ...` and new subparser invocations
- Docstrings in `cpp_backend.py:166-208` and the helper comment block at `route_cmd.py:1066` are excellent — future maintainers will understand this immediately
- `_emit()` closure pattern in `ensure_cpp_backend_available` is clean and avoids passing `quiet` everywhere

- The `_emit` closure has a slightly redundant `if not quiet: _emit(...)` guard at line 259 (the closure already short-circuits on `quiet`) — purely cosmetic, no impact
- Both the `--no-auto-build-native` argparse blocks at `parser.py:2299` and `parser.py:2887` are duplicated verbatim (parallel to the existing `--backend` duplication) — not a regression introduced by this PR; the underlying argparse architecture has this in two places

Approved. Removing `loom:review-requested` and adding `loom:pr`.

Issue #2547, opened 2026-05-07
loom: builder/doctor role contracts should require incremental commits to preserve progress on crashes


rjwalters:
## Problem

Today's #2542 builder agent ran for **175 minutes** working on board 05 placement, made meaningful progress (24 LoC of `design.py` changes adding rotation infrastructure for the DRV8301), then crashed with `API Error: 529 Overloaded`. Result: **zero commits**. The work was only recoverable because we manually inspected the worktree after the crash.

This is the second time today an agent died mid-task with budget loss — the first was the doctor on PR #2535 hitting the per-day usage limit after 6 minutes.

## Why it matters

- Builder budget is finite (per-day usage limits, API quota, wall-clock for the user)
- A 175-minute crashed-with-no-commits run is strictly worse than 5×35-minute runs that each commit progress
- Builder/doctor agents are stateful per task; if the parent crashes, ALL the work is in the worktree but invisible to git history
- For autonomous workflows (the loom daemon), there's no way for a follow-up agent to pick up where the crashed one left off if nothing is committed

## Proposal

Adopt an **incremental-commit contract** in the builder/doctor protocols:

1. **Commit when adding scaffolding/infrastructure**, even if not yet exercised. Today's example: the failed builder added `rotation` parameter to `generate_htssop56()` but didn't actually rotate. That commit alone (well-documented capability without behavior change) would have been valuable.
2. **Commit before each "tier" attempt** in iterative work. Today's curator on #2542 explicitly listed Tier 0–5; commit between each so failures don't lose all progress.
3. **Commit before any single command that takes >2 minutes.** Long-running route attempts are common — checkpoint before kicking them off.
4. **Push every committed batch** so the agent's work is durably visible without requiring the agent to complete its task.

This belongs in `.claude/commands/builder*.md` and `.claude/commands/doctor.md` (or wherever the role contracts live).

## Acceptance

- Builder/doctor role commands explicitly require incremental commits at the boundaries above
- A test scenario: simulate an agent crash mid-iteration; verify the worktree's branch has commits captured up to the crash point
- Documentation/examples updated to reflect the new contract

## Out-of-scope

This issue covers the protocol/role contract change. Whether to also add a hard-enforce mechanism (e.g., a hook that requires a commit every N tool calls) is a separate question.

## Cross-references

- Surfaced by: 2026-05-07 #2542 builder failure (175 min runtime, 0 commits)
[...]

PR #2555, opened 2026-05-07
fix(route): auto-pour preserves INPUT.kicad_pcb when output path differs (closes #2548)

### LINKED ISSUE #2548: kct route: auto-pour modifies INPUT.kicad_pcb in-place, surprising users with two-file diffs
## Problem

`kct route INPUT.kicad_pcb -o OUTPUT.kicad_pcb` modifies BOTH `INPUT.kicad_pcb` AND writes `OUTPUT.kicad_pcb`. The input modification is `auto_pour_if_missing` writing zone definitions in-place to the input file before routing.

This surprised us today when staging PR #2545 — `git status` showed two modified files for what felt like a single output operation. The `INPUT.kicad_pcb` change was unintentional but had to be committed since it represents the real on-disk state after running the tool.

## Why it matters

- **Confusing diff**: a "regenerate the routed PCB" operation produces a 2-file diff, the first being inscrutable to a reviewer who doesn't know about auto-pour
- **Reproducibility**: re-running `kct route` twice gives different *inputs* the second time (zones already present), which can mask non-determinism in the router itself
- **CI surprises**: anything that compares `INPUT.kicad_pcb` before/after `kct route` will see drift
- **Tooling friction**: `git diff` review on PR #2545 showed +1477 lines on the INPUT pcb that aren't really part of the change

### In-place semantics confirmed

`src/kicad_tools/router/auto_pour.py:170-286` — `auto_pour_if_missing(pcb_path, ...)` reads `pcb_path.read_text()`, computes new zone S-expressions, and writes them back to `pcb_path`. Docstring at line 183 explicitly says "Path to .kicad_pcb file (modified **in place**)". `_remove_zones_for_nets` and `_detect_uninset_zones` also mutate the same path.

### Four call sites in `route_cmd.py`

All four follow the **same pattern** — they pass `pcb_path` (the input) to `auto_pour_if_missing`, never the output path:

| Line | Function | Comment |
|---|---|---|
| 1107 | (legacy 2-layer route) | `auto_pour_if_missing(pcb_path, quiet=quiet, edge_clearance=getattr(args, "edge_clearance", None))` |
| 1643 | (relaxation-tier route) | same |
| 2137 | (high-perf route) | same |
| 3512 | (main `kct route` entry path) | same, with `args.quiet` |

NOTE: the original issue body cited lines 954, 1487, 1975, 3347 — those are nearby **but not** the actual call sites. The real call sites are 1107, 1643, 2137, 3512. (Each is preceded by `if getattr(args, "auto_pour", True):`, so the existing `--no-auto-pour` flag already lets users opt out of the in-place modification entirely — but at the cost of routing without pours.)

After auto-pour mutates `pcb_path`, the route code reads it again at lines 1490, 1992, 2513, etc. (`original_content = pcb_path.read_text()`), then writes the routed S-expression to `output_path`. So **a temp-file fix has to redirect every downstream `pcb_path` read** at the temp file, not just the auto-pour call. The simpler shape is to copy `pcb_path` → temp early, point `pcb_path` at the temp for the rest of the function, and never touch the original.

### `kct build`'s pipeline relies on input-equals-output

`src/kicad_tools/cli/pipeline_cmd.py::_run_step_route` (lines 604-619) invokes the route subcommand with **`-o` set to the same path as the input** (`str(ctx.pcb_file)` for both positional and `-o`). In pipeline mode the routed PCB intentionally overwrites itself, and downstream steps (`_run_step_zones` at line 841, `_run_step_zones_refill` at line 907) operate on `ctx.pcb_file` (the now-routed file). So **the pipeline does NOT depend on auto-pour modifying a separate input** — when input == output, in-place behavior is exactly what's wanted. Any fix must preserve that case (don't regress `kct build`).

`_run_step_zones` runs `kicad-cli pcb drc --refill-zones --save-board` against the routed output, **not** the original input. It needs zone definitions to exist *somewhere on disk by the time it runs*; in pipeline mode they end up in `ctx.pcb_file` because input == output. So zones-step has no specific dependency on the *input* being mutated — it only needs the *file it operates on* to contain `(zone ...)` blocks.

### PR #2516 / #2516 temp-file precedent

The pattern to mirror lives in `src/kicad_tools/cli/runner.py:1184-1190`:

```python
# DRC modifies the input file in-place.  If the caller requested a
# separate output file, copy the source first and run DRC on the copy.
if output_path is not None:
    shutil.copy2(pcb_path, output_path)
    target_pcb = output_path
else:
    target_pcb = pcb_path
```

This is `shutil.copy2(src, dst)` first, then operate on `dst`. Lighter than `tempfile.TemporaryDirectory` and fits the route case naturally because route already has an `output_path`. Could be:

```python
# Auto-pour modifies the file in place.  If we're routing to a
# different output path, copy input -> output first and let auto-pour
# write zones to the output, leaving the user-visible input untouched.
if output_path != pcb_path:
    shutil.copy2(pcb_path, output_path)
    pcb_path = output_path  # rebind for rest of route function
auto_pour_if_missing(pcb_path, ...)
```

This collapses to a no-op when input == output (pipeline mode), so `kct build` is unaffected.

### Idempotency analysis

Re-running `kct route INPUT -o OUTPUT` **twice** with the same INPUT does NOT produce identical INPUTs across the two runs — that's the user-visible nondeterminism the issue calls out:

1. **Run 1**: INPUT has no zones → auto-pour adds zones to INPUT → router routes with pours present → OUTPUT written.
2. **Run 2**: INPUT now has zones (from run 1) → auto-pour skips (already present) → INPUT untouched this time → router routes → OUTPUT written.

So:
- INPUT changes only on the FIRST run; subsequent runs are no-ops at the auto-pour layer (verified by `tests/test_auto_pour.py:test_idempotent_second_call`).
- BUT: the *router's input bytes* differ between run 1 (no zones) and run 2 (zones present). This means a user wiping INPUT back to the pre-zone state and re-running could get different routing results than running again on the already-mutated INPUT.
- In a fresh-clone CI scenario, run 1 produces a 2-file diff; re-running produces a 1-file diff. The user has to know to revert INPUT to reproduce.

The temp-file fix makes this clean: every run on a fresh INPUT produces the same OUTPUT and never touches INPUT.

## Proposal

**Recommended: Option 1 (default to non-mutating; preserve in-place when input == output).**

Rationale:
- The current contract violates least-surprise. Users naturally read `kct route IN -o OUT` as "read IN, write OUT" — the in-place mutation has zero signal in the CLI surface.
- The temp-file precedent already exists (PR #2516, runner.py:1184). This is a known, narrow change.
- `kct build` is the only caller that needs in-place behavior, and it gets it for free because it passes `-o ctx.pcb_file` (input == output, copy is a no-op or skipped).
- Existing `--no-auto-pour` flag already exists for users who want zero input modification at any cost (they accept no pours).

**Alternative: Option 2 (opt-in flag `--no-modify-input`).**

Tradeoffs vs Option 1:
- Pro: zero risk of regressing existing scripts that depend on the post-route INPUT state. (Are there any? Audit `boards/*/output/` for `.kicad_pcb` files that are git-tracked AND known to be the "post-auto-pour" state — the PR #2545 incident suggests yes; we've been committing the mutated INPUTs.)
- Con: Default behavior remains surprising. New users and downstream tooling continue tripping on it.
- Con: Doubles the command-flag space (`--no-auto-pour` vs `--no-modify-input` are subtly different).

**Recommended path forward**: Option 1 by default, with a `--modify-input-in-place` escape hatch for any caller that *wants* the old contract. Pipeline can adopt the new default transparently because input == output makes the temp-file copy a no-op (or we skip the copy when they're equal).

## Acceptance criteria

- `kct route INPUT.kicad_pcb -o OUTPUT.kicad_pcb` (with `INPUT != OUTPUT`) leaves INPUT byte-identical after the run, even when auto-pour creates zones. INPUT's mtime may change (if we touch it) but contents must be unchanged — ideally mtime is unchanged too.
- `kct route INPUT.kicad_pcb -o INPUT.kicad_pcb` (input == output, the pipeline case) continues to work and produces a routed PCB with zone definitions, identical to today's behavior.
- `kct route INPUT.kicad_pcb` with no `-o` (output defaults to `INPUT_routed.kicad_pcb` per route_cmd.py:3354) leaves INPUT byte-identical.
- `kct build boards/<n>/output/<board>.kicad_pcb` runs end-to-end without regression: route → stitch → zones-fill → fix-drc → audit → export all still work and produce the same gerbers as today.
- All existing route tests still pass (no behavior change for in-place case).
- Test suite gains a new regression test that covers the input-preservation contract.

## Test plan

1. **Unit test (new)**: `tests/test_route_cmd_input_preservation.py`
   - Create a small PCB with un-poured power nets in a `tmp_path`.
   - Compute `sha256(INPUT)`.
   - Invoke `kct route INPUT -o OUTPUT` via the CLI (subprocess or argparse-driven call).
   - Assert `sha256(INPUT)` is unchanged after the run.
   - Assert `OUTPUT` exists, contains routed traces, and contains the auto-poured zones.

2. **Unit test (new)**: same fixture, invoke `kct route INPUT -o INPUT` (in-place case) — assert OUTPUT (== INPUT) now contains zones AND routed traces. This locks in pipeline behavior.

3. **Unit test (new)**: invoke twice on a fresh INPUT, both times to a fresh OUTPUT path — assert OUTPUT_run1 and OUTPUT_run2 are byte-identical. Locks in idempotency at the route-command level.

4. **Integration**: run `kct build boards/01-voltage-divider/output/voltage_divider.kicad_pcb` end-to-end on a clean clone and verify the resulting `voltage_divider_routed.kicad_pcb` and gerbers match a known-good snapshot.

5. **Manual smoke**: `git status` after `kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb -o /tmp/x.kicad_pcb` should show **no modifications** to anything in `boards/`. Today this would show `voltage_divider.kicad_pcb` modified.

6. **No regressions**: existing `tests/test_auto_pour.py`, `tests/test_route_cmd_*.py`, `tests/test_route_cmd_zone_fill.py` all pass unchanged.

## Affected files

- `src/kicad_tools/router/auto_pour.py` — `auto_pour_if_missing` (lines 170-286). Either keep in-place semantics (cheaper, document them as internal-only) and have callers stage a copy first; OR add an `output_path` parameter and route writes there. Recommendation: keep auto_pour internal-in-place (it's a private util) and have route_cmd.py do the copy.
- `src/kicad_tools/cli/route_cmd.py` — four call sites at lines **1107, 1643, 2137, 3512** (corrected from issue body's 954/1487/1975/3347). Each needs the `if output_path != pcb_path: shutil.copy2(...); pcb_path = output_path` shim *before* calling `auto_pour_if_missing`. Also note the route body re-reads `pcb_path` at lines 1490, 1992, 2513 — those reads must also see the temp/output file.
- `src/kicad_tools/cli/pipeline_cmd.py::_run_step_route` (line 604) — no change needed; input == output case is preserved.
- `tests/test_route_cmd_input_preservation.py` — new file per test plan.
- Documentation: any docstring or README reference to "kct route modifies input" must be updated.

## Cross-references

- Surfaced when staging PR #2545 (board 04 fresh route).
- Auto-pour was wired into `kct route` by PR #2407 (closes #2395) — that's the layer that introduced the in-place modification.
- Related to #2516 (zone-fill safety net via DRC --refill-zones; it pioneered the `shutil.copy2(pcb_path, output_path)` temp-pattern in `runner.py:1184`).
- Test file `tests/test_auto_pour.py:test_idempotent_second_call` already covers auto_pour-level idempotency; this issue is about *command-level* idempotency.

rjwalters:
## Summary

`kct route INPUT.kicad_pcb -o OUTPUT.kicad_pcb` was modifying BOTH files because `auto_pour_if_missing` writes zone definitions in-place to the input before routing. This produced confusing two-file diffs and broke command-level idempotency: the second run on a fresh INPUT saw different bytes than the first.

This PR introduces `_stage_input_for_auto_pour` in `route_cmd.py` (mirroring the `shutil.copy2` pattern from `runner.py:1184-1190`) and wires it into all four `auto_pour_if_missing` call sites in `route_cmd.py`. When INPUT != OUTPUT, the helper copies the input to the output path and rebinds `pcb_path` so auto-pour and downstream reads operate on the copy. When INPUT == OUTPUT (the `kct build` pipeline case), the copy is a no-op and the existing in-place behavior is preserved.

- 4 call sites in `route_cmd.py` now stage a copy before auto-pour: `route_with_layer_escalation` (line 1107), `route_with_rule_relaxation` (line 1643), `route_with_combined_escalation` (line 2137), and the main `kct route` entry path (line 3512).
- New `tests/test_route_cmd_input_preservation.py` covers the helper, the helper-plus-auto-pour composition, the pipeline (input == output) case, and command-level idempotency across two fresh runs.
- Existing `tests/test_auto_pour.py` (12 tests) and broader route_cmd tests (87 tests total) continue to pass.

Closes #2548

## Test plan

- [x] `uv run pytest tests/test_route_cmd_input_preservation.py tests/test_auto_pour.py` -- 18 passed
- [x] `uv run pytest -k "route_cmd or auto_pour"` -- 87 passed, 4 skipped
- [x] `uv run ruff check src/kicad_tools/cli/route_cmd.py tests/test_route_cmd_input_preservation.py` -- clean
- [x] `uv run ruff format --check` on changed files -- clean
- [x] CLI smoke test: `kct route /tmp/cp_2548.kicad_pcb -o /tmp/cp_out_2548.kicad_pcb` on a charlieplex board with 0 zones in INPUT -- after the run, INPUT MD5 unchanged (d3d054fdd6a97e9b642b1801afcc48b1), INPUT still has 0 zones, OUTPUT has 2 zones. Acceptance criterion met.
- [x] CLI smoke test: `kct route /tmp/vd_2548.kicad_pcb` (default output, no -o) on voltage_divider -- INPUT MD5 unchanged, INPUT still has 0 zones, default OUTPUT (*_routed.kicad_pcb) has zones.
- [x] CLI smoke test: `kct route /tmp/cp3_2548.kicad_pcb -o /tmp/cp3_2548.kicad_pcb` (input == output, pipeline mode) -- file is mutated in place, zones are added. Pipeline behavior preserved.

Generated with Claude Code

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -3546,10 +3546,13 @@ def main(argv: list[str] | None = None) -> int:
     if args.skip_nets:
         skip_nets = [n.strip() for n in args.skip_nets.split(",")]
 
-    # Auto-create copper pours for power nets (before skip detection)
+    # Auto-create copper pours for power nets (before skip detection).
+    # auto_pour_if_missing writes in-place; stage a copy at output_path
+    # first so the user's INPUT is left untouched (issue #2548).
     if getattr(args, "auto_pour", True):
         from kicad_tools.router.auto_pour import auto_pour_if_missing
 
+        pcb_path = _stage_input_for_auto_pour(pcb_path, output_path)
         auto_pour_if_missing(
             pcb_path,
             quiet=args.quiet,
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -2168,10 +2168,13 @@ def route_with_combined_escalation(
     if args.skip_nets:
         skip_nets = [n.strip() for n in args.skip_nets.split(",")]
 
-    # Auto-create copper pours for power nets (before skip detection)
+    # Auto-create copper pours for power nets (before skip detection).
+    # auto_pour_if_missing writes in-place; stage a copy at output_path
+    # first so the user's INPUT is left untouched (issue #2548).
     if getattr(args, "auto_pour", True):
         from kicad_tools.router.auto_pour import auto_pour_if_missing
 
+        pcb_path = _stage_input_for_auto_pour(pcb_path, output_path)
         auto_pour_if_missing(
             pcb_path,
             quiet=quiet,
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -1671,10 +1671,13 @@ def route_with_rule_relaxation(
     if args.skip_nets:
         skip_nets = [n.strip() for n in args.skip_nets.split(",")]
 
-    # Auto-create copper pours for power nets (before skip detection)
+    # Auto-create copper pours for power nets (before skip detection).
+    # auto_pour_if_missing writes in-place; stage a copy at output_path
+    # first so the user's INPUT is left untouched (issue #2548).
     if getattr(args, "auto_pour", True):
         from kicad_tools.router.auto_pour import auto_pour_if_missing
 
+        pcb_path = _stage_input_for_auto_pour(pcb_path, output_path)
         auto_pour_if_missing(
             pcb_path,
             quiet=quiet,
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -56,6 +56,7 @@ Layer Stack Configuration:
 import argparse
 import logging
 import math
+import shutil
 import signal
 import sys
 import textwrap
@@ -952,6 +953,37 @@ def _print_power_stall_suggestions(
     print("  2. Increase layer count:              --layers 4 (or --max-layers 6)")
     print(f"  3. Manual routing in KiCad for the {len(stalled_nets)} remaining net(s)")
     print()
+
+
+def _stage_input_for_auto_pour(pcb_path: Path, output_path: Path) -> Path:
+    """Stage the input PCB so ``auto_pour_if_missing`` does not mutate the user's input.
+
+    ``auto_pour_if_missing`` writes zone definitions in-place to its target
+    file.  When ``kct route INPUT -o OUTPUT`` is invoked with distinct paths
+    the user expects ``INPUT`` to be left untouched; this helper copies
+    ``pcb_path`` to ``output_path`` and returns ``output_path`` so the rest
+    of the route flow can operate on the copy.
+
+    When ``pcb_path`` and ``output_path`` resolve to the same file (the
+    pipeline ``kct build`` case where input == output), no copy is needed
+    and ``pcb_path`` is returned unchanged — preserving the existing
+    in-place behavior the pipeline depends on.
+
+    Args:
+        pcb_path: User-supplied input PCB path.
+        output_path: User-supplied output PCB path.
+
+    Returns:
+        Path that subsequent route steps should use as their working PCB.
+        Either ``pcb_path`` (input == output) or ``output_path`` (after copy).
+
+    See: issue #2548 — ``kct route`` should not modify INPUT when output
+    differs.  Mirrors the ``shutil.copy2`` pattern in ``runner.py``.
+    """
+    if pcb_path.resolve() == output_path.resolve():
+        return pcb_path
+    shutil.copy2(pcb_path, output_path)
+    return output_path
 
 
 def _auto_skip_pour_nets(
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -1132,10 +1132,13 @@ def route_with_layer_escalation(
     if args.skip_nets:
         skip_nets = [n.strip() for n in args.skip_nets.split(",")]
 
-    # Auto-create copper pours for power nets (before skip detection)
+    # Auto-create copper pours for power nets (before skip detection).
+    # auto_pour_if_missing writes in-place; stage a copy at output_path
+    # first so the user's INPUT is left untouched (issue #2548).
     if getattr(args, "auto_pour", True):
         from kicad_tools.router.auto_pour import auto_pour_if_missing
 
+        pcb_path = _stage_input_for_auto_pour(pcb_path, output_path)
         auto_pour_if_missing(
             pcb_path,
             quiet=quiet,
--- a/tests/test_route_cmd_input_preservation.py
+++ b/tests/test_route_cmd_input_preservation.py
@@ -68,9 +68,7 @@ def _make_pcb_with_pour_candidates() -> str:
     parts.append('  (net 3 "SDA")\n')
     parts.append('  (net 4 "SCL")\n')
     parts.append('  (footprint "TestLib:TestPkg" (layer "F.Cu") (at 10 10)\n')
-    for idx, (nid, name) in enumerate(
-        [(1, "GND"), (2, "VCC"), (3, "SDA"), (4, "SCL")]
-    ):
+    for idx, (nid, name) in enumerate([(1, "GND"), (2, "VCC"), (3, "SDA"), (4, "SCL")]):
         x_off = idx * 2.0
         parts.append(
             f'    (pad "{idx + 1}" smd roundrect (at {x_off} 0) '
--- a/tests/test_route_cmd_input_preservation.py
+++ b/tests/test_route_cmd_input_preservation.py
@@ -1,0 +1,256 @@
+"""Tests that ``kct route INPUT -o OUTPUT`` does not modify INPUT (issue #2548).
+
+Auto-pour writes zone definitions in-place to its target PCB.  Before
+this fix, ``kct route`` passed the user's INPUT path to auto-pour, which
+silently mutated the input file -- producing a confusing two-file diff
+for what users naturally read as a single read-INPUT/write-OUTPUT
+operation.
+
+The fix stages a copy at OUTPUT first (when INPUT != OUTPUT) and runs
+auto-pour against the copy, leaving INPUT byte-identical.  When INPUT
+== OUTPUT (the ``kct build`` pipeline case), the copy is a no-op and
+in-place behavior is preserved.
+
+This module covers:
+
+1. The ``_stage_input_for_auto_pour`` helper directly (path semantics).
+2. A behavioral test that exercises the auto-pour code path through
+   the helper plus ``auto_pour_if_missing`` and confirms the user's
+   INPUT file is unchanged across the full operation.
+"""
+
+from __future__ import annotations
+
+import hashlib
+from pathlib import Path
+
+# ---------------------------------------------------------------------------
+# Minimal PCB fixture (mirrors the fixture in tests/test_auto_pour.py so
+# this test stays self-contained even if that one is refactored).
+# ---------------------------------------------------------------------------
+
+_PCB_HEADER = """\
+(kicad_pcb
+  (version 20240108)
+  (generator "test")
+  (generator_version "8.0")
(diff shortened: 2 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

Solid, well-scoped fix for #2548. Reviewed against the 10 scrutiny items in the review request.

### Verification

| # | Check | Result |
|---|-------|--------|
| 1 | Path normalization | `Path.resolve()` handles relative-vs-absolute and symlinks correctly. See edge case below. |
| 2 | Pipeline regression | Verified `kct route <pcb> -o <same pcb>` mutates in place as expected (MD5 changes). |
| 3 | Call sites consistent | `git grep _stage_input_for_auto_pour src/` shows exactly 1 def + 4 calls (lines 1141, 1680, 2177, 3555) — matches curator enrichment. |
| 4 | Idempotency test | `test_command_level_idempotency_across_fresh_outputs` is meaningful: asserts INPUT hash stable across two fresh-output runs and that count1==count2. Pre-fix, run 1 would mutate INPUT and cause count2 < count1. Not a tautology. |
| 5 | `shutil.copy2` semantics | Matches `src/kicad_tools/cli/runner.py:1187` repo convention (DRC has the exact same in-place-mutation problem and uses `copy2`). Consistent. |
| 6 | Variable rebind flows downstream | Verified: after `pcb_path = _stage_input_for_auto_pour(...)` the same name flows into `auto_pour_if_missing(pcb_path, ...)` and `_auto_skip_pour_nets(pcb_path, ...)` at all 4 call sites. |
| 7 | Fresh OUTPUT path | Verified: `shutil.copy2` creates the file. Test `test_distinct_paths_copies_input_to_output` asserts `not output_pcb.exists()` before the call. |
| 8 | Nonexistent output dir | `kct route IN -o /nonexistent/dir/x.kicad_pcb` fails with `FileNotFoundError: [Errno 2] No such file or directory`. Same error class as on main, but fails *earlier* (before routing) — a minor UX improvement, not a regression. |
| 9 | CI green | All 5 checks PASS (Lint, Type Check, Test, C++ Build, kicad-cli Round-trip). `mergeStateStatus: CLEAN`. |
| 10 | Board 01 baseline | `kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb -o /tmp/x.kicad_pcb` succeeds and INPUT MD5 (`ded940c30675a56c2cf17348908a5cc4`) is unchanged. Confirmed across two consecutive runs. |

Local toolchain: 18/18 targeted tests pass; 87/87 route_cmd+auto_pour tests pass; ruff/format clean on changed files. Pre-existing 25 mypy errors in `route_cmd.py` are identical on main (untouched by this PR).

### Minor non-blocking observation

Equality is by `Path.resolve()`, which on macOS APFS (case-insensitive but case-preserving) does NOT case-fold. So `kct route Foo.kicad_pcb -o foo.kicad_pcb` (same physical file via different case) would fall into the copy branch and `shutil.copy2` would raise `SameFileError`. Not a data-loss risk (Python detects same-file via `os.path.samefile`), and not a regression vs main, but `pcb_path.samefile(output_path)` would be marginally more robust. Worth keeping in mind but not worth blocking on; this exact mismatch is far more likely to be caught by the user's own CWD pattern than by their casing.

### Strengths

- 7 incremental commits dogfood the #2547 protocol.
- New test file is self-contained (own PCB fixture) so it survives refactors of `test_auto_pour.py`.
- Helper docstring explicitly references issue #2548 and points at the `runner.py` pattern.
- Pipeline `kct build` no-op branch is exercised by `test_input_equals_output_returns_input_unchanged` and `test_input_mutated_when_input_equals_output`.
- All 4 call sites use identical `pcb_path = _stage_input_for_auto_pour(pcb_path, output_path)` pattern with matching comment block — easy to grep, easy to maintain.

Approving.

PR #2576, opened 2026-05-08
feat(blocks): create_gate_drive_resistor_array — N-channel slew-rate resistors (closes #2570)

### LINKED ISSUE #2570: feat(blocks): create_gate_drive_resistor_array — N-channel slew-rate resistors
## Problem

A series gate-drive resistor (commonly 10–47Ω) belongs in the path from each gate-driver IC output to its MOSFET gate. Its purpose is slew-rate control: too little series R causes ringing/EMI and shoot-through risk; too much increases switching loss and dead-time requirements. This pattern recurs on any half-bridge / multi-bridge / 3-phase inverter topology.

Today on board 05 (`boards/05-bldc-motor-controller/design.py`):

- The block-level integration uses `GateDriverBlock` (DRV8301) at `design.py:442-456` and `ThreePhaseInverter` at `design.py:465-477`. Neither of these blocks contains a series gate-drive resistor.
- The `GateDriverBlock` class (`src/kicad_tools/schematic/blocks/motor.py:643-797`) only emits the driver IC, bootstrap caps, and bypass caps; its `GATE_HS_A/B/C` and `GATE_LS_A/B/C` ports go straight out with no series R.
- In the schematic-block flow, the `GateDriverBlock.GATE_HS_*` / `GATE_LS_*` ports are NEVER explicitly wired — search of `design.py` for `GATE_HS` / `GATE_LS` returns zero hits. Connectivity from driver to MOSFET gates is implicit via the `GATE_AH/BH/CH` net labels emitted by the layout-only generators (`design.py:1602,1605,1608`) and MCU-side labels (`design.py:367-369`).
- The references R10, R11, R12 on board 05 are NOT gate-drive resistors. They are 5mΩ current sense shunts emitted by `CurrentSenseShunt` blocks at `design.py:479-515` (`ref_start=10 + i`, `shunt_value=\"5mR\"`, `shunt_package=\"2512\"`). The original issue text mis-identified them — please ignore that reference.

So this issue is really: add a reusable factory for series gate-drive (slew-rate) resistors, then wire it into board 05 between the gate driver outputs and the inverter MOSFET gates. A net-name fix may also be needed (currently board 05 uses `GATE_AH` for the high-side gate, but `GateDriverBlock.GATE_HS_A` exposes a HS-only port — check that `PWM_AH` vs `GATE_HS_A` distinction is honored after refactor).

## Proposal

Add a standalone factory in `src/kicad_tools/schematic/blocks/motor.py`:

```python
def create_gate_drive_resistor_array(
    sch: \"Schematic\",
    x: float,
    y: float,
    channels: int = 3,
    value: str = \"22\",                     # ohms; 10-47 typical for IRLZ44N-class gates
    ref_start: int = 1,
    ref_prefix: str = \"R\",
    resistor_symbol: str = \"Device:R\",
    resistor_package: str = \"0805\",
    spacing: float = 10.0,
    input_nets: list[str] | None = None,    # e.g. [\"GATE_DRV_AH\", ...]
    output_nets: list[str] | None = None,   # e.g. [\"GATE_AH\", ...]
) -> CircuitBlock:
```

Returns a `CircuitBlock` whose `components` map includes `R_GATE_1..N` and whose `ports` include `IN_1..N` and `OUT_1..N` (and pair aliases like `IN_AH`/`OUT_AH` when `input_nets`/`output_nets` are provided).

### Standalone factory vs. parameter on `GateDriverBlock` — recommendation

**Recommendation: standalone factory.** Reasons:

1. **Reusability.** Gate-drive resistors are useful with any gate driver (not just `GateDriverBlock`'s `DRV8308`/`IR2110` defaults), and even with discrete totem-pole or transformer-coupled drivers. A standalone block is also useful between an MCU GPIO and a low-side FET (no driver IC at all).
2. **Channel asymmetry.** A 3-phase driver emits 6 gate signals (HS+LS), so `channels` is not always equal to `num_phases` of the driver. Encoding it as a parameter on `GateDriverBlock` would force coupling that doesn't naturally exist.
3. **Consistent with existing block factoring.** Other passive arrays — `CurrentSenseShunt`, `DecouplingCaps`, `I2CPullups` — are independent blocks, not parameters on the IC they accompany. `GateDriverBlock` is the IC; the resistor array is a peripheral, mirroring that split.
4. **Easy composition later.** Once `create_gate_drive_resistor_array` exists, an optional `gate_resistor: str | None = None` kwarg on `GateDriverBlock` can be added in a follow-up that internally instantiates the array — but the lower-level primitive still has standalone value.

## Why

- Slew-rate control is critical for FET driver designs (too fast = ringing/EMI; too slow = switching loss).
- Required on board 05 (3 channels — but at least 6 if both HS and LS get series R, which is typical).
- Reusable on any multi-bridge topology (board 05's 3-phase inverter, future H-bridges, single low-side switches).
- Decouples slew-rate tuning from gate-driver IC choice and from the inverter block.

## Acceptance Criteria

- [ ] New `create_gate_drive_resistor_array` factory in `src/kicad_tools/schematic/blocks/motor.py` accepting `channels`, `value`, `ref_start`, `ref_prefix`, `resistor_symbol`, `resistor_package`, `input_nets`, `output_nets`.
- [ ] Returns a `CircuitBlock` subclass instance with `components` (R_GATE_1..N) and `ports` (IN_1..N, OUT_1..N, plus name-aliases when `input_nets`/`output_nets` provided).
- [ ] Each resistor is placed at predictable spacing (`spacing` kwarg) and emits proper `add_label` calls when input/output net names are provided so the resistors actually appear on the right nets in the schematic netlist.
- [ ] Exported from `src/kicad_tools/schematic/blocks/__init__.py` and added to `__all__`.
- [ ] Re-export visible at `kicad_tools.schematic.blocks.create_gate_drive_resistor_array`.
- [ ] Board 05 (`boards/05-bldc-motor-controller/design.py`) refactored to use it. Suggested approach:
  - Rename current driver-side nets to `GATE_DRV_AH/BH/CH` (and `_AL/BL/CL` if low-side resistors are also added).
  - Insert `create_gate_drive_resistor_array(channels=3, value=\"22\", ref_start=<next-free>, input_nets=[\"GATE_DRV_AH\",\"GATE_DRV_BH\",\"GATE_DRV_CH\"], output_nets=[\"GATE_AH\",\"GATE_BH\",\"GATE_CH\"])`.
  - **Pick a non-conflicting `ref_start`.** R1–R6 = LED pull-downs/pull-ups in earlier sections, R10–R12 = current-sense shunts. Use R20+ (verify by scanning `design.py` for `\"R\" + str(N)`).
  - Update `print(...)` summary to mention the new resistors.
- [ ] `pnpm check:ci` clean (ruff, mypy, pytest, format).
- [ ] Board 05 ERC clean after refactor (no new unconnected/conflict warnings).
- [ ] Board 05 still routes to completion (`uv run kct design.py` succeeds end-to-end).

## Test Plan

New test file `tests/test_blocks_gate_drive_resistor_array.py` (mirror style of `TestCurrentSenseShuntMocked` and `TestGateDriverBlockMocked` in `tests/test_schematic_blocks.py`):

1. **`test_default_3_channels`**: Default invocation creates exactly 3 resistors, each value `\"22\"` (or chosen default), ref `R1..R3`.
2. **`test_channel_count_parametrized`**: Parametrize over `channels in [1, 2, 3, 6]`; assert resistor count equals `channels`.
3. **`test_value_propagation`**: Pass `value=\"33\"`; assert all resistors are added with that value (inspect `add_symbol` call args).
4. **`test_ref_start_offset`**: `ref_start=20, channels=3` produces `R20, R21, R22`.
5. **`test_ports_present`**: `IN_1..IN_N` and `OUT_1..OUT_N` exist in `block.ports` with sane coordinates (one resistor pin maps to IN, the other to OUT).
6. **`test_input_output_net_aliases`**: When `input_nets=[\"GATE_DRV_AH\"]` and `output_nets=[\"GATE_AH\"]` are passed (channels=1), `block.ports` contains `IN_AH` and `OUT_AH` aliases AND `add_label` is called with each net name at the correct pin position.
7. **`test_net_list_length_validation`**: Mismatched `len(input_nets) != channels` raises `ValueError` with a descriptive message; same for `output_nets`.
8. **`test_resistor_package_property`**: Pass `resistor_package=\"0603\"`; assert symbol properties dict carries `Package == \"0603\"` (mirrors `CurrentSenseShunt` pattern at `motor.py:484`).
9. **`test_returns_circuit_block`**: Result is an instance of `CircuitBlock` and has the correct `schematic`, `x`, `y` attributes.
10. **`test_components_dict_keyed`**: `block.components` has keys `R_GATE_1..R_GATE_N` mapping to mock symbol instances.

Existing test surface that must keep passing:

- `tests/test_schematic_blocks.py::TestGateDriverBlockMocked` (4 tests) — unchanged because we are NOT modifying `GateDriverBlock`.
- Board 05 generator runs to completion in CI smoke tests if any.

## Affected Files

- `src/kicad_tools/schematic/blocks/motor.py` — add `create_gate_drive_resistor_array` factory (after `create_current_sense` at line 868). Optionally add a small `GateDriveResistorArray(CircuitBlock)` class if the factory returns a typed subclass rather than a bare `CircuitBlock`.
- `src/kicad_tools/schematic/blocks/__init__.py` — re-export from `.motor` (lines 79-87) and add to `__all__` (line 200 area).
- `boards/05-bldc-motor-controller/design.py` — insert the new block between `gate_driver` (line 442) and `inverter` (line 465); pick `ref_start` ≥ 20 to avoid R1–R6 and R10–R12.
- `tests/test_blocks_gate_drive_resistor_array.py` (new).

## Cross-references

- Source: 2026-05-08 circuit-blocks survey
- Related blocks (existing): `GateDriverBlock` (`motor.py:643`), `HalfBridge` (`motor.py:11`), `ThreePhaseInverter`, `CurrentSenseShunt` (`motor.py:393`)
- Related factories: `create_half_bridge` (`motor.py:803`), `create_3phase_inverter` (`motor.py:835`), `create_current_sense` (`motor.py:868`) — the new factory should follow the same shape.
- Note for builder: the original issue text said \"R10–R12 are gate-drive resistors\" — that is incorrect. R10–R12 on board 05 are the current-sense shunts (5mΩ, 2512). New gate-drive resistors should use a fresh ref range.

---

_Curated 2026-05-06: confirmed `GateDriverBlock` has no series gate resistors, that GATE_HS/GATE_LS ports are unused in board 05, and that R10–R12 are current-sense shunts (not gate resistors). Standalone factory recommended over `GateDriverBlock` parameter._

rjwalters:
## Summary

- Adds ``create_gate_drive_resistor_array`` factory in
  ``src/kicad_tools/schematic/blocks/motor.py`` (and a corresponding
  ``GateDriveResistorArray`` class).  The factory produces an N-channel
  series-resistor block intended to splice in between a gate-driver IC
  output and the corresponding power-MOSFET gate to control switching
  slew rate (10-47 ohms typical).

- Re-exports the new factory and class from
  ``src/kicad_tools/schematic/blocks/__init__.py``.

- Refactors board 05 (``boards/05-bldc-motor-controller/design.py``) so
  the high-side MOSFET gates are driven through R20-R22 (22 ohms each)
  rather than direct from the DRV8301.  As part of this:
  * MCU PWM nets renamed ``GATE_*`` -> ``PWM_*`` (these are PWM logic
    inputs to the driver, not the MOSFET gates).
  * DRV8301 high-side outputs renamed ``GATE_*H`` -> ``GATE_DRV_*H``.
  * MOSFET-gate-side nets remain ``GATE_*H`` / ``GATE_*L``.
  * R20-R22 splice ``GATE_DRV_*H`` -> ``GATE_*H`` for the three phases.

- Adds 13 mocked-Schematic unit tests in
  ``tests/test_blocks_gate_drive_resistor_array.py``.

Closes #2570

### DIFF
--- a/boards/05-bldc-motor-controller/design.py
+++ b/boards/05-bldc-motor-controller/design.py
@@ -39,6 +39,7 @@ from kicad_tools.schematic.blocks import (
     LEDIndicator,
     ThreePhaseInverter,
     create_5v_buck,
+    create_gate_drive_resistor_array,
 )
 from kicad_tools.schematic.models.schematic import Schematic
 
@@ -363,26 +364,30 @@ def create_bldc_controller(output_dir: Path) -> Path:
     _connect_mcu_pin_to_label("12", "HALL_B", dx=5, dy=0)
     _connect_mcu_pin_to_label("13", "HALL_C", dx=5, dy=0)
 
-    # High-side gate PWM: PA8/PA9/PA10 (TIM1_CH1/CH2/CH3)
-    _connect_mcu_pin_to_label("18", "GATE_AH", dx=5, dy=0)
-    _connect_mcu_pin_to_label("19", "GATE_BH", dx=5, dy=0)
-    _connect_mcu_pin_to_label("20", "GATE_CH", dx=5, dy=0)
+    # High-side gate PWM: PA8/PA9/PA10 (TIM1_CH1/CH2/CH3).  These drive the
+    # DRV8301 INH_A/B/C logic inputs (pins 17/19/21), not the MOSFET gates
+    # directly — gate output of the driver is GATE_DRV_*H, then through the
+    # R20/R21/R22 slew-rate resistors to GATE_*H on the MOSFET gates.
+    _connect_mcu_pin_to_label("18", "PWM_AH", dx=5, dy=0)
+    _connect_mcu_pin_to_label("19", "PWM_BH", dx=5, dy=0)
+    _connect_mcu_pin_to_label("20", "PWM_CH", dx=5, dy=0)
 
     # SWD debug pins
     _connect_mcu_pin_to_label("23", "SWDIO", dx=5, dy=0)
     _connect_mcu_pin_to_label("24", "SWCLK", dx=5, dy=0)
     _connect_mcu_pin_to_label("26", "SWO", dx=5, dy=0)
 
-    # Low-side gate PWM: PB6/PB7/PB8 (TIM4_CH1/CH2/CH3, sync'd with TIM1)
-    _connect_mcu_pin_to_label("29", "GATE_AL", dx=5, dy=0)
-    _connect_mcu_pin_to_label("30", "GATE_BL", dx=5, dy=0)
-    _connect_mcu_pin_to_label("31", "GATE_CL", dx=5, dy=0)
+    # Low-side gate PWM: PB6/PB7/PB8 (TIM4_CH1/CH2/CH3, sync'd with TIM1).
+    # Drives DRV8301 INL_A/B/C logic inputs (pins 18/20/22).
+    _connect_mcu_pin_to_label("29", "PWM_AL", dx=5, dy=0)
+    _connect_mcu_pin_to_label("30", "PWM_BL", dx=5, dy=0)
+    _connect_mcu_pin_to_label("31", "PWM_CL", dx=5, dy=0)
 
     # Crystal pins: PF0/PF1 -> OSC_IN/OSC_OUT
     _connect_mcu_pin_to_label("2", "OSC_IN", dx=-5, dy=0)
     _connect_mcu_pin_to_label("3", "OSC_OUT", dx=-5, dy=0)
 
-    print("   Wired 16 floating nets (6 GATE, 3 HALL, 3 ISENSE-, 4 SWD) to MCU pins")
+    print("   Wired 16 floating nets (6 PWM, 3 HALL, 3 ISENSE-, 4 SWD) to MCU pins")
 
     # Crystal oscillator (8MHz)
     xtal = CrystalOscillator(
@@ -454,6 +459,22 @@ def create_bldc_controller(output_dir: Path) -> Path:
     print("   Gate driver: DRV8301 (GateDriverBlock)")
     print("   Bootstrap caps: C12, C13, C14")
     print("   Bypass caps: C15, C16")
+
+    # Series gate-drive (slew-rate) resistors between DRV8301 outputs and the
+    # high-side MOSFET gates.  ``GATE_DRV_AH/BH/CH`` are the driver-IC outputs;
+    # ``GATE_AH/BH/CH`` are the MOSFET-gate-side nets.  The array sits in the
+    # path between them.  Low-side gates remain direct-driven for now.
+    create_gate_drive_resistor_array(
+        sch,
+        x=X_GATE_DRV + 30,
+        y=120,
+        channels=3,
+        value="22",
+        ref_start=20,  # R20-R22 (R10-R12 are the current-sense shunts)
+        input_nets=["GATE_DRV_AH", "GATE_DRV_BH", "GATE_DRV_CH"],
+        output_nets=["GATE_AH", "GATE_BH", "GATE_CH"],
+    )
+    print("   Gate-drive resistors: R20, R21, R22 (22 ohms, HS only)")
 
     # =========================================================================
     # Section 7: Power Stage (using ThreePhaseInverter and CurrentSenseShunt)
@@ -805,6 +826,20 @@ def create_bldc_pcb(output_dir: Path) -> Path:
         "PWR_LED": 29,
         "STATUS_LED": 30,
         "SW_OUT": 31,
+        # Gate-driver-side (between DRV8301 outputs and the slew-rate resistors).
+        # The MOSFET-gate-side nets remain GATE_AH/BH/CH.  Low-side gates stay
+        # direct-driven so GATE_AL/BL/CL still tie the driver IC to the MOSFETs.
+        "GATE_DRV_AH": 32,
+        "GATE_DRV_BH": 33,
+        "GATE_DRV_CH": 34,
+        # MCU-side PWM logic inputs to the DRV8301 (pins 17-22).  These are
+        # distinct from the GATE_* nets above, which are the MOSFET gates.
+        "PWM_AH": 35,
+        "PWM_AL": 36,
+        "PWM_BH": 37,
+        "PWM_BL": 38,
+        "PWM_CH": 39,
+        "PWM_CL": 40,
     }
 
     # =========================================================================
@@ -882,6 +917,14 @@ def create_bldc_pcb(output_dir: Path) -> Path:
     R10_POS = (BOARD_ORIGIN_X + 8, BOARD_ORIGIN_Y + 84)
     R11_POS = (BOARD_ORIGIN_X + 24, BOARD_ORIGIN_Y + 84)
     R12_POS = (BOARD_ORIGIN_X + 40, BOARD_ORIGIN_Y + 84)
+
+    # Gate-drive (slew-rate) resistors -- between DRV8301 HS outputs and the
+    # high-side MOSFET gates.  Sit just above each phase's HS MOSFET so the
+    # GATE_DRV_*H -> GATE_*H net runs vertically.  R20-R22 use the same 0805
+    # footprint generator as R3/R4.
+    R20_POS = (BOARD_ORIGIN_X + 8, BOARD_ORIGIN_Y + 64)  # Phase A HS
+    R21_POS = (BOARD_ORIGIN_X + 24, BOARD_ORIGIN_Y + 64)  # Phase B HS
+    R22_POS = (BOARD_ORIGIN_X + 40, BOARD_ORIGIN_Y + 64)  # Phase C HS
 
     # Motor connector (right edge, bottom -- near MOSFETs)
     J2_POS = (BOARD_ORIGIN_X + 65, BOARD_ORIGIN_Y + 76)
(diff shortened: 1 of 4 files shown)


rjwalters:
## Judge Review: Approved

Excellent work. The factory cleanly mirrors the existing `create_current_sense` / `create_half_bridge` shape, the board-05 refactor threads consistent net names end-to-end across both the schematic-block flow and the layout-only flow, and all six CI checks are green.

### What I verified

**Factory API signature matches curator spec**
- `create_gate_drive_resistor_array(sch, x, y, channels=3, value="10", *, ref_start, ref_prefix, resistor_symbol, resistor_package, spacing, input_nets, output_nets)` — keyword-only after positional core, `value` defaults to `"10"` per curator's recommended default. Returns `GateDriveResistorArray(CircuitBlock)`.
- `IN_i` / `OUT_i` ports plus suffix-aliased `IN_AH` / `OUT_AH` style ports when nets are provided. `_net_suffix` helper extracts the trailing token (`GATE_DRV_AH` → `AH`, `GATE_AH` → `AH`) so both alias to the same suffix per channel — clean for paired-port composition.
- `ValueError` on length mismatch for `input_nets` / `output_nets`. Tested both directions.

**LS gate-drive scope decision is justified**
- Issue #2570's *suggested approach* is `channels=3` HS-only (acceptance criteria literal text). The "6 if both HS and LS get series R, which is typical" line in the "Why" section is informational, not a strict requirement.
- DRV8301 has lower-impedance LS gate drive and shorter PCB paths (LS source ties to ISENSE rail, not the bootstrapped phase node), so direct-drive LS is defensible.
- The schematic comment explicitly documents "Low-side gates remain direct-driven for now" — leaves the door open for a follow-up when LS slew tuning proves necessary.
- The factory itself supports `channels=6`, so a future LS extension is purely a board-level call (no factory changes needed).

**Net rename consistency across schematic and PCB sides**
- Schematic side: `_connect_mcu_pin_to_label` calls renamed to `PWM_AH/BH/CH` and `PWM_AL/BL/CL` for MCU pins 18-20, 29-31.
- PCB layout `NETS` dict adds `GATE_DRV_AH/BH/CH` (32-34) and `PWM_*H/L` (35-40) — net count rises 31 → 40 (matches builder's claim).
- `STM32G431K8_PINS` list — pins 18/19/20 → `PWM_AH/BH/CH`, pins 29/30/31 → `PWM_AL/BL/CL`. Consistent with the schematic-side label calls.
- `DRV8301_PINS` list — pins 17/19/21 → `PWM_AH/BH/CH`, pins 18/20/22 → `PWM_AL/BL/CL`, pins 47/42/37 → `GATE_DRV_AH/BH/CH`. Consistent with the schematic-side gate-driver wiring.
- LS path stays consistent (no rename): DRV8301 pins 35/40/45 → `GATE_CL/BL/AL` → Q6/Q4/Q2 unchanged.
- End-to-end HS path verified: DRV8301 pin 47 → `GATE_DRV_AH` → R20 → `GATE_AH` → Q1 (matches for B and C phases too).

**R20-R22 placement non-conflicting**
- R20 (8, 64), R21 (24, 64), R22 (40, 64) — same x-columns as the HS MOSFETs (Q1/Q3/Q5 at y=68) and the current-sense shunts R10-R12 (y=84). Sit between U3 (DRV8301 at y=50) and the HS MOSFETs (y=68), 4 mm above the HS row. No conflict with R10-R12 which are 16 mm below at y=84. The vertical layout of "DRV8301 → R-array → HS MOSFET → LS MOSFET → shunt" is geometrically clean per phase column. 0805 footprint vs 2512 for shunts — different sizes, no overlap.

**22Ω value reasonable**
- Curator's recommended default was `"10"` (matched by the factory's default). Board 05 explicitly overrides with `"22"`, which is mid-range of the 10-47Ω typical envelope and conservative for IRLZ44N gates (Qg ≈ 48 nC) driven by the DRV8301's ~1 A peak gate-drive current. Sensible.

**PCB regeneration reproduces builder's claim**
- Local run confirms: `Components: 49`, `Nets: 40`, plus the `9b. Adding gate-drive (slew-rate) resistors...` step prints `R20, R21, R22 (22 ohm gate-drive, HS only)`.
- Net count delta: +9 nets vs main (3 GATE_DRV_*H, 6 PWM_*).
- Component count delta: +3 (R20, R21, R22).
- The CI's "Routed PCB DRC Check" is green, indicating the routed PCB builds DRC-clean. The "kicad-cli Round-trip Smoke" check also green.

**Test semantic coverage spot-check**
- `test_input_output_net_aliases` — verifies both alias mapping AND that `add_label` lands at the correct pin coordinates (input net at IN pin, output net at OUT pin). Strong semantic check.
- `test_ports_present` — exercises actual coordinate math (in_pos x < out_pos x, spacing = 10, pin offsets ±5). Catches sign-flip / pin-swap regressions.
- `test_net_list_length_validation` — both directions (too short / too long), pytest.raises with descriptive `match=`.
- `test_value_propagation` — verifies both attribute and positional arg index 4 in `add_symbol` calls.
- `test_resistor_package_property` — confirms the `properties={"Package": ...}` kwarg, mirroring `CurrentSenseShunt`.
- `test_returns_circuit_block` — type-checks both `CircuitBlock` and `GateDriveResistorArray`, and `schematic`/`x`/`y` attrs from the parent class.

13/13 new tests pass locally in the full suite run; 278 existing schematic-block tests still pass.

### CI status

- Lint & Format: pass (16s)
- Type Check: pass (42s)
- Test: pass (5m33s)
- C++ Build Check: pass (1m54s)
- Routed PCB DRC Check: pass (9s)
- kicad-cli Round-trip Smoke: pass (40s)

- The factory's `value` default (`"10"`) is one notch below the typical 22Ω most multi-bridge designs land at, but matches the curator's literal recommendation and the conservative-defaults convention. Board 05 explicitly passes `"22"`, so the default doesn't propagate to users today.
- Pre-existing F541 ruff warnings in `boards/05-bldc-motor-controller/design.py` are not introduced by this PR (same 6 errors exist on main). Project's CI lint job has `continue-on-error: true` so they don't block.
- Future work to add LS gate-drive resistors (R23-R25) would be a one-liner at the design level given the factory already supports it — worth tracking as a follow-up issue if LS slew tuning becomes a routing or EMI concern.

Approving.

Issue #2587, opened 2026-05-08
Phase 1C-cont: wire cpp_backend.py to consume the new C++ binding partner_net params (Epic #2556)


rjwalters:
## Problem

Phase 1C (Epic #2556) landed in PR #2586 with the **Python backend fully wired** to use `NetClassRouting.intra_pair_clearance` for within-pair edges, but the **C++ backend currently uses backward-compatible defaults** (`partner_net=-1`) — so C++ users get unchanged routing behavior, not the tighter intra-pair clearance.

This was a deliberate scope split: PR #2586 was salvaged from an autonomous builder agent that ran out of usage budget after 2/6 planned commits. The C++ binding signatures were added (so the work is unblocked) but `cpp_backend.py` wasn't updated to actually pass `partner_net` through to the new params.

## What's needed

1. **`cpp_backend.py:893`** — `route_resumable()` call site. Currently uses default `partner_net=-1`. Update to:
   - Compute `partner_net_id` via `_resolve_partner_net_id(net_name)` (mirroring the Python backend at `pathfinder.py:1595`)
   - Pass `partner_net=partner_net_id` and `partner_radius=intra_pair_radius_cells` to `route_resumable()`
   - Use `effective_intra_pair_clearance()` to compute the radius scalar

2. **`cpp_backend.py:1404-1412`** — `find_blocking_nets()` Bresenham scanner. Concern is "exclude partner from the blocker set" (different shape from `route_resumable`). The C++ binding for this needs investigation — may need a new helper or `partner_net` param if not already added.

3. **C++ unit tests** — exercise the partner-net branch in the C++ search loops directly (currently only the Python pathfinder branch is tested).

4. **Empirical board 03 verification via C++ backend** — `kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb --backend cpp` should produce 0 DRC errors (currently produces 1, the same -0.200mm USB pair overlap).

5. **Python/C++ behavioral parity tests** — given the same NetClassRouting with `intra_pair_clearance`, both backends must produce equivalent within-pair routing.

## Acceptance

- `cpp_backend.py` consumes the new C++ binding params with computed `partner_net_id` from `_resolve_partner_net_id`
- Board 03 routed via `--backend cpp` produces 0 DRC errors
- Python and C++ behavioral parity verified
- Existing tests pass (Phase 1A/1B/1C/1D)
[...]

PR #2583, opened 2026-05-08
feat(blocks): create_hall_sensor_input — RC-filtered Hall sensor input (closes #2564)

rjwalters:
## Summary

Adds a new ``HallSensorInput`` circuit block + ``create_hall_sensor_input``
factory to ``src/kicad_tools/schematic/blocks/analog.py`` (peer of
``ThermistorSense`` / ``VoltageDividerSense`` / ``ADCInputFilterBlock``)
and refactors board 05's Section 9 to consume it.

The block implements the canonical noise-immune Hall / open-collector
GPIO input: a 10kOhm pull-up to VCC plus a 10nF cap to GND, exposing
``VCC`` / ``SIGNAL_IN`` / ``SIGNAL_OUT`` / ``GND`` ports.  The factory
mirrors the ``ThermistorSense`` + ``create_temperature_sense`` pattern
(positional ``(x, y)`` placement, ports dict driven by the caller via
``sch.add_label(...)``).

## Changes

- ``src/kicad_tools/schematic/blocks/analog.py`` — new
  ``HallSensorInput`` class and ``create_hall_sensor_input`` factory
- ``src/kicad_tools/schematic/blocks/__init__.py`` — exports the new
  class and factory in the ``# Analog`` section
- ``boards/05-bldc-motor-controller/design.py`` — Section 9 now calls
  the factory 3x (HALL_A/B/C with ``ref_start=30..32``) and the
  hand-rolled PCB generator places matching R30-R32 (0805, 10k) and
  C30-C32 (0805, 10nF) on the board near J3
- ``tests/test_schematic_blocks.py`` — adds ``TestHallSensorInputMocked``
  (7 cases) and 2 factory tests in ``TestMotorControlFactoryFunctions``

## Acceptance verification

- [x] ERC clean: ``kct erc bldc_controller.kicad_sch`` reports 0 errors
      / 0 warnings
- [x] DRC baseline preserved: routed-PCB DRC drops from 99 errors
      (baseline on origin/main) to 74 errors after re-routing with the
      new parts (no new error categories introduced)
- [x] Schematic now contains 6 new components (R30/R31/R32 +
      C30/C31/C32) -- HALL signal grep count: 6
- [x] All 287 tests in ``tests/test_schematic_blocks.py`` pass
      (including 9 new ones for the Hall block)

## Test plan

- [x] Unit tests for ``HallSensorInput`` (creation, defaults, custom
      values, ``ref_start`` offsets, wiring, port aliasing, custom
      symbol args)
[...]

PR #2588, opened 2026-05-08
chore(ci): tighten board 05 DRC allowlist 17 → 15 (lock in PR #2583's floor)

rjwalters:
## Summary

PR #2583 (Hall sensor placement re-tune) dropped board 05's routed PCB DRC from 22 → 15 errors via the R30-R32/C30-C32 spread along J3. The committed `_routed.kicad_pcb` has 15 errors; the allowlist was still 17 from the pre-PR-#2583 floor.

Per the file's own guidance:
> When a board's count IMPROVES (e.g., 17 → 12), reduce the value here in the same PR that reduces the count, so further regressions are caught against the new floor.

PR #2583 didn't do this in the same PR; the judge explicitly flagged it as a follow-up. This PR addresses that.

## Test plan

```
$ uv run python scripts/ci/check_routed_drc.py \
    boards/05-bldc-motor-controller/output/bldc_controller_routed.kicad_pcb
OK: ... 15 errors (allowlist max 15; reduce the allowlist value in
.github/routed-drc-tolerance.yml if this count drops further).
```

## What's NOT in this PR

Board 03 was investigated for similar tightening (1 → 0) — the fresh Python-backend route can produce 0 errors but shows wide run-to-run variance (0 to 3 errors observed in 2 runs). The 1-error allowlist is preserved. The full clean state for board 03 is tracked in #2587 (Phase 1C-cont — autorouter wiring of `intra_pair_clearance`).

### DIFF
--- a/.github/routed-drc-tolerance.yml
+++ b/.github/routed-drc-tolerance.yml
@@ -27,9 +27,12 @@
 #   * board 03 (1 error) -- tracked separately; pre-existing pre-CI-gate state.
 
 tolerances:
-  # 15x clearance_segment_segment + 2x clearance_segment_via at JLCPCB rules.
-  # Partial progress landed in #2551 (closes #2532); full fix tracked in #2542.
-  boards/05-bldc-motor-controller/output/bldc_controller_routed.kicad_pcb: 17
+  # Tightened from 17 → 15 to lock in the floor reached by PR #2583 (Hall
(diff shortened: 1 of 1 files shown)


rjwalters:
APPROVED — manual review. Single-line config change. CI 6/6 SUCCESS, including the Routed PCB DRC Check (which is exactly what this PR tightens — board 05 reports 15 errors, allowlist max 15, still passes; further reduction would fail). Per the file own guidance.

PR #2552, opened 2026-05-07
feat(ci): validate committed _routed.kicad_pcb files via kct check (closes #2546)

rjwalters:
## Summary

Adds a CI gate (`routed-pcb-drc-check` job in `.github/workflows/ci.yml`) that validates committed `*_routed.kicad_pcb` files via `kct check --mfr jlcpcb --errors-only` on every PR. Closes the safety-net gap that allowed PR #2538 to ship with 5 DRC errors despite a "0 errors" claim at merge time.

The gate is a per-board "allowed minus epsilon" floor: regressions still fail even on grandfathered boards, but the gate can land before every board reaches 0 errors. Initial allowlist (`.github/routed-drc-tolerance.yml`) grandfathers the empirical state of `origin/main` today:

- board 03 (`usb_joystick_routed.kicad_pcb`): 1 error
- board 05 (`bldc_controller_routed.kicad_pcb`): 17 errors (partial #2542 progress in #2551)

Boards 01, 02, 04 are NOT listed -- they currently report 0 errors and the gate is strict for them.

## Design choices

- **Pure-Python DRC** (`kct check`) -- no kicad-cli, no Docker container. Job piggybacks on the cheap `ubuntu-latest` + `uv sync --extra dev` pattern (~30s setup) used by lint/typecheck/test.
- **In-job file detection** (no `paths:` filter) -- the job always runs and short-circuits to green on PRs that don't touch routed PCBs, so it is safe to mark as a required status check on branch protection.
- **`fetch-depth: 0`** on `actions/checkout` -- required for `git diff origin/main...HEAD` to resolve the merge-base.
- **Helper script extracted** (`scripts/ci/check_routed_drc.py`) -- the allowlist comparison logic is non-trivial and warrants extraction for testability. Keeps the inline workflow YAML readable and makes the unit-test surface explicit.
- **GitHub-Actions annotations** (`::error file=...::`) -- offending files surface inline in the PR Files-changed view.

## Files

- `.github/workflows/ci.yml` -- adds the `routed-pcb-drc-check` job (~80 lines).
- `.github/routed-drc-tolerance.yml` -- per-board allowlist (~20 lines + comments).
- `scripts/ci/check_routed_drc.py` -- helper that reads the allowlist, runs `kct check` per file, and emits annotations.
- `tests/test_ci_routed_drc_workflow.py` -- 28 tests across three layers (workflow YAML structure, allowlist YAML schema, helper-script unit tests).

## Commit history

[...]

PR #2591, opened 2026-05-09
ci: warn on stale routed-drc-tolerance.yml entries (closes #2590)

### LINKED ISSUE #2590: ci: detect stale routed-drc-tolerance.yml entries (allowlist looser than actual)
## Problem

`.github/routed-drc-tolerance.yml` allows board entries to be looser than the actual committed file requires. Board 05 lived for ~24 hours with allowlist=17 while the committed file had 15 errors (caught and fixed in PR #2588). The CI gate enforces "actual ≤ allowlist" but doesn't flag the inverse: "actual < allowlist by N → tighten the allowlist".

This stale-allowlist drift means the file's own guidance — "When a board's count IMPROVES, reduce the value here in the same PR that reduces the count" — is not enforced and depends on reviewer vigilance. PR #2583's judge had to manually flag it as a follow-up; PR #2588 then did the tightening.

## Why it matters

- Stale allowlist entries weaken the regression gate — a board could drift BACK up to the old allowlist value silently
- Reviewer attention shouldn't be the only safety net; the gate-builder PR (#2552) explicitly designed for "allowed minus epsilon" semantics, and drift undermines that
- The CI job has all the data (actual vs allowlist) at every run; surfacing the gap is cheap

## Proposal

Extend `scripts/ci/check_routed_drc.py` (or add a sibling) to:
1. For each entry in `.github/routed-drc-tolerance.yml`, compute `slack = allowlist - actual`
2. If `slack > 0` (allowlist looser than actual), emit a `::warning::` GitHub annotation: "Allowlist for `<file>` is `<allowlist>` but actual is `<actual>`. Tighten to `<actual>` in this PR or a follow-up to lock in the floor."
3. Optionally fail the job if `slack > 0` is detected and the PR description doesn't include a justification (e.g., `## Allowlist drift acknowledged: <reason>`)

The warning-annotation approach is non-blocking and matches how #2552 already uses `::error::` annotations — visible in the PR Files-changed view, easy to action.

## Acceptance

- New CI behavior: PRs that touch `boards/*/output/*_routed.kicad_pcb` get a warning annotation per stale allowlist entry
- The warning includes the recommended new value
- A test scenario: artificially set an allowlist to a higher value than the actual file → CI emits the warning
- Existing CI behavior unchanged (gate still enforces actual ≤ allowlist)

## Optional stretch

A `scripts/ci/audit-routed-drc-allowlist.sh` script the repo can run locally or in a scheduled job (weekly) that opens a PR auto-tightening any stale entries. Too much automation for now, but worth noting.

## Affected files

- `scripts/ci/check_routed_drc.py` — add drift detection
- `.github/workflows/ci.yml` — surface the warning annotations
- `tests/test_ci_routed_drc_workflow.py` — add drift-test fixtures
- `.github/routed-drc-tolerance.yml` — comment in the file itself pointing at this gate

## Cross-references

- #2546 (CLOSED) — original CI DRC validation gate (PR #2552)
- PR #2583 → PR #2588 — the recent example this would have caught automatically

rjwalters:
## Summary

Closes #2590. When a routed PCB's actual DRC error count is strictly less than its allowlist value (slack > 0), emit a `::warning file=...::` GitHub-Actions annotation so the stale entry surfaces in the PR Files-changed view rather than being buried in stdout.

PR #2583 dropped board 05 from 22 to 15 errors but left the allowlist at 17. The OK message printed `"15 errors (allowlist max 17; reduce the allowlist value...)"` to the log -- it was technically there, just not visible. PR #2588 manually fixed it 24h later. This change makes the suggestion visible to reviewers in the standard place they look (Files-changed annotations).

## Implementation

- New `annotate_drift_warning(file, errors, allowed)` helper following the existing `::warning::` precedent at the deleted-file branch (line 224).
- `check_file()` now returns `(passed, message, errors)` so callers can compute drift slack without re-running the (expensive) `kct check`.
- `main()` emits the warning when `passed and allowed > 0 and errors < allowed`. The `allowed > 0` guard avoids noise on unlisted boards (every clean PR would otherwise get nagged).
- Allowlist YAML comment updated to document the new behavior + cross-PR drift caveat.
- TODO comment in the script noting that cross-PR drift detection is deferred to a future scheduled-audit job.

## Scope decision (per curator's Option A)

v1 only inspects files in the diff (matches the existing `main()` loop). Cross-PR drift -- e.g., board 03 was independently improved by a router refactor in PR B but the allowlist wasn't updated -- is rare and better solved by a weekly bot. See the TODO in `annotate_drift_warning` for the deferral pointer.

The gate's exit code is unchanged: warnings are advisory, matching the precedent at the deleted-file branch.

## Acceptance criteria (from #2590)

- [x] PRs that touch `boards/*/output/*_routed.kicad_pcb` get a `::warning file=...::` annotation per stale allowlist entry (where slack > 0).
- [x] Warning includes the recommended new value: "Allowlist for `<file>` is <allowlist> but actual is <actual>. Tighten to <actual> in this PR or a follow-up..."
- [x] Existing CI behavior unchanged (gate still enforces actual <= allowlist; exit codes 0/1/2 unchanged).
- [x] Test fixture: artificially set an allowlist to a higher value than the actual file -> the helper script's `check_file()` returns OK (gate passes) AND emits a warning annotation. Covered by `TestMainDriftWarningEmission::test_slack_positive_emits_drift_warning` and `test_stale_allowlist_via_subprocess`.
- [x] TODO-comment in the script noting cross-PR drift detection is deferred to a future scheduled-audit job.
- [x] All existing 28 tests in `tests/test_ci_routed_drc_workflow.py` still pass.

## Test plan

- [x] All 38 tests in `tests/test_ci_routed_drc_workflow.py` pass (28 original + 10 new). Run with `uv run pytest tests/test_ci_routed_drc_workflow.py`.
- [x] `uv run ruff check scripts/ci/check_routed_drc.py tests/test_ci_routed_drc_workflow.py` clean.
- [x] `uv run ruff format ... --check` clean.
- [x] Mypy on the helper script reports only the pre-existing `types-PyYAML` stubs warning -- no new errors introduced.
- [x] End-to-end subprocess test exercises the full `main()` path including argv parsing, allowlist load, and stdout flushing.

## New tests

`TestHelperDriftWarning` (4 tests) -- unit tests for the new helper:
- annotation format (`::warning file=...::`)
- includes both actual and allowed counts
- includes "Tighten to <N>" recommendation
- references `.github/routed-drc-tolerance.yml`

`TestMainDriftWarningEmission` (6 tests) -- integration tests for `main()`'s emission flow with `count_errors` stubbed:
- slack > 0 emits warning (the PR #2583 regression scenario)
- slack = 0 emits no warning (steady state)
- slack < 0 fails with `::error::`, no drift warning (regression case)
- unlisted board with 0 errors emits no warning (allowed=0 guard)
- warning does not change exit code (stays 0)
- end-to-end via subprocess on a real allowlist with a stale entry

## Commits

1. `b05d7822` ci: emit warning annotation (script changes + test signature updates)
2. `4cac8fd5` test(ci): cover stale-allowlist warning emission paths
3. `4a365980` docs(ci): explain stale-entry warning in routed-drc-tolerance.yml

Each commit leaves the tree in a working state per #2547.

### DIFF
--- a/.github/routed-drc-tolerance.yml
+++ b/.github/routed-drc-tolerance.yml
@@ -14,9 +14,17 @@
 #     absence of the entry is the strict 0-error gate.
 #   * When a board's count IMPROVES (e.g., 17 -> 12), reduce the value here
 #     in the same PR that reduces the count, so further regressions are
-#     caught against the new floor.
+#     caught against the new floor. Issue #2590 added a CI warning annotation
+#     that nags the PR Files-changed view when the actual count is below the
+#     allowlist value (slack > 0), so reviewers see the stale entry.
 #   * Adding a new entry (loosening) requires reviewer justification and a
 #     tracking issue link in the PR description.
+#
+# Cross-PR drift (an entry going stale because a different PR improved the
+# count without tightening) is NOT detected by the per-PR warning above --
+# only files in the diff are checked. A scheduled-audit job that walks every
+# entry in this file is deferred (see the TODO in
+# scripts/ci/check_routed_drc.py::annotate_drift_warning).
 #
 # Schema:
 #   tolerances:
--- a/scripts/ci/check_routed_drc.py
+++ b/scripts/ci/check_routed_drc.py
@@ -147,13 +147,48 @@ def annotate_error(file: str, message: str) -> None:
     print(f"::error file={file}::{message}", flush=True)
 
 
-def check_file(pcb_path: Path, allowed: int) -> tuple[bool, str]:
+def annotate_drift_warning(file: str, errors: int, allowed: int) -> None:
+    """Emit a GitHub-Actions ``::warning file=...::`` for a stale allowlist entry.
+
+    Called when a routed PCB's actual error count is strictly less than the
+    allowlist value (slack > 0). The warning surfaces in the PR Files-changed
+    view alongside ``::error::`` annotations so reviewers don't miss it
+    (issue #2590).
+
+    Args:
+        file: Repo-relative path to the routed PCB (used as the annotation's
+            ``file=`` target so GitHub anchors the warning to the file).
+        errors: Actual error count returned by ``kct check``.
+        allowed: Current allowlist value from
+            ``.github/routed-drc-tolerance.yml``.
+
+    Note:
+        TODO(#2590): Cross-PR drift detection (i.e., warning on stale entries
+        for files NOT touched in the current PR) is deferred to a future
+        scheduled-audit job. v1 only inspects files in the diff so the
+        warning attaches to a file the reviewer is actively looking at, and
+        we don't pay the per-board ``kct check`` cost for every entry on
+        every PR. See the issue's "Scope question" section for the rationale.
+    """
+    slack = allowed - errors
+    print(
+        f"::warning file={file}::Allowlist for `{file}` is {allowed} but "
+        f"actual is {errors} (slack={slack}). Tighten to {errors} in this "
+        f"PR or a follow-up to lock in the new floor (see "
+        f".github/routed-drc-tolerance.yml).",
+        flush=True,
+    )
+
+
+def check_file(pcb_path: Path, allowed: int) -> tuple[bool, str, int]:
     """Check a single PCB against its allowed error count.
 
     Returns:
-        (passed, message) tuple. ``passed`` is True if errors <= allowed.
-        ``message`` is a human-readable summary suitable for both stdout
-        and GitHub annotation.
+        ``(passed, message, errors)`` tuple. ``passed`` is True if
+        ``errors <= allowed``. ``message`` is a human-readable summary
+        suitable for both stdout and GitHub annotation. ``errors`` is the
+        actual count returned by ``kct check`` so callers can compute drift
+        slack without re-running the (expensive) DRC check.
     """
     errors = count_errors(pcb_path)
     if errors <= allowed:
@@ -165,7 +200,7 @@ def check_file(pcb_path: Path, allowed: int) -> tuple[bool, str]:
                 f"reduce the allowlist value in .github/routed-drc-tolerance.yml "
                 f"if this count drops further)."
             )
-        return True, msg
+        return True, msg, errors
 
     if allowed == 0:
         msg = (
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge review: APPROVED

All curator-recommended Option A semantics met, signature change is backward-compatible (single in-repo caller updated), guards are correct, and the failure path is properly exclusive of the warning path.

### Verification
- **`check_file()` 3-tuple signature**: only caller at `scripts/ci/check_routed_drc.py:279` and 5 test sites are all updated. No external consumers grep-found.
- **`allowed > 0` guard** at line 296 correctly suppresses noise on unlisted boards where `allowlist.get(lookup_key, 0)` returns 0. Inline comment makes the redundancy explicit and intentional.
- **`::warning file=...::` form** matches the deleted-file precedent at lines 224-227 (originally PR #2552).
- **Exit code invariant**: drift emission lives inside `if passed:` and never mutates `overall_failed`. Pinned by `test_drift_warning_does_not_change_exit_code`.
- **Failure-case exclusion**: negative-slack path falls into the `else` branch and only calls `annotate_error`. Pinned by `test_slack_negative_failure_no_drift_warning` (rc=2, no "Tighten to" in stdout).
- **Test depth**: 10 new tests cover annotation format, content (allowed/actual/recommended/allowlist-path), all four emission conditions (positive/zero/negative slack, unlisted board), exit-code invariance, and an end-to-end subprocess shim that exercises the real argv-parsing + stdout-flush path. All 38 tests pass locally in 0.4s.
- **YAML comment**: cross-references issue, explains the per-PR scope boundary, and points at the TODO in `annotate_drift_warning` for the deferred cross-PR audit. Self-documenting for future reviewers.
- **CI**: all 6 checks pass (`Lint & Format`, `Type Check`, `Test` 5m45s, `C++ Build Check`, `kicad-cli Round-trip Smoke`, `Routed PCB DRC Check`). `mergeStateStatus: CLEAN`.

### Minor follow-up suggestions (non-blocking)
- The `TODO(#2590):` annotation in `annotate_drift_warning` should eventually point at a fresh tracking issue for the scheduled-audit job, not the closed-by-this-PR issue. File at the time of implementation.
- `_make_pcb_and_allowlist` writes absolute `tmp_path` paths into the allowlist; this works only because `main()` swallows the `relative_to` `ValueError` and falls through to the raw lookup key. The subprocess test correctly exercises the cwd-relative path that real CI takes, so coverage is sound.

Ready to merge.

PR #2598, opened 2026-05-10
fix(router): exclude PWR_FLAG ERC-marker nets from auto-pour (closes #2592)

### LINKED ISSUE #2592: route: auto-pour creates PWR_FLAG zones — should exclude ERC-only marker nets
## Problem

When \`kct route\` runs auto-pour on chorus-test-revA, it creates a copper zone for the \`PWR_FLAG\` net:

\`\`\`
Auto-pour: created 6 zone(s) for +3.3V, +3.3VA, +5V, GNDA, GNDD, PWR_FLAG
Auto-skip: +3.3V, +3.3VA, +5V, GNDA, GNDD, PWR_FLAG (pour nets — use zone fill)
WARNING: Zone 'PWR_FLAG' on F.Cu (priority 3) overlaps queued zone '+3.3VA' (priority 1). The other zone will get zero copper because the other zone has equal or higher priority.
WARNING: Zone 'PWR_FLAG' on F.Cu (priority 3) overlaps queued zone '+5V' (priority 2). The other zone will get zero copper because the other zone has equal or higher priority.
\`\`\`

\`PWR_FLAG\` is a synthetic ERC marker net — it exists only to suppress "no driver" ERC errors on power inputs. It carries no real connection and should never be poured as copper. Worse, its zone overlaps the legitimate \`+3.3VA\` and \`+5V\` zones at higher priority, forcing those to get zero copper.

## Repro

\`\`\`bash
uv run kicad-tools pcb sync-netlist boards/external/chorus-test-revA/kicad/chorus-test-revA_v18.kicad_pcb \\
  --schematic boards/external/chorus-test-revA/kicad/chorus-test-revA.kicad_sch \\
  -o /tmp/synced.kicad_pcb
uv run kicad-tools pcb strip boards/external/chorus-test-revA/kicad/chorus-test-revA_v18.kicad_pcb \\
  -o /tmp/stripped.kicad_pcb --remove-orphan-vias
uv run kicad-tools route /tmp/synced.kicad_pcb -o /tmp/routed.kicad_pcb \\
  --strategy negotiated --layers 4 --backend cpp --manufacturer jlcpcb --no-cache
\`\`\`

Observed in route logs at \`6926b47c\`.

## Expected behavior

\`PWR_FLAG\` (and any other ERC-only marker nets — \`*_FLAG\`, hidden power-symbol artifacts) should be excluded from auto-pour zone creation by name pattern and not poured.

## Acceptance criteria

- chorus-test-revA route produces no \`PWR_FLAG\` zone
- The two "will get zero copper" warnings caused by \`PWR_FLAG\` overlay disappear
- Add a unit test for the auto-pour pour-net selector covering \`PWR_FLAG\` exclusion
- Document the exclusion list in the auto-pour module docstring

rjwalters:
## Summary

- Closes #2592
- `PWR_FLAG` (and user-named variants like `+3V3_FLAG`) are ERC-only virtual symbols that carry no copper, but KiCad's netlister still emits them as ordinary net entries.  The name-based classifier in `router/net_class.py` matches the leading `PWR` against the POWER pattern, so auto-pour was creating a spurious zone for them at F.Cu priority 3, starving legitimate power rails (`+3.3VA`, `+5V`) of copper.
- Fix at the pour-net selection layer (`router/auto_pour.py`) rather than the classifier — the classifier returning POWER for `PWR_FLAG` is not strictly wrong; the bug is specifically about pouring the net.
- Added `ERC_MARKER_NET_PATTERNS` and `_is_erc_marker_net()` helper to `auto_pour.py`, filter those names out of the pour-net build loop in `auto_pour_if_missing`, and re-use the helper in `route_cmd._auto_skip_pour_nets` so the misleading `Auto-skip: PWR_FLAG (pour nets — use zone fill)` log line and the `Routing: PWR_FLAG (power nets without zones)` log line no longer appear.
- Schematic-side analogue already exists in `src/kicad_tools/cli/sch_connectivity.py:119-125` (`"PWR_FLAG is an ERC annotation, not a real net name"`); this PR is the PCB-side mirror.

## Acceptance criteria (from issue)

- `kct route` no longer produces a `PWR_FLAG` zone in routed output PCBs.
- `Zone 'PWR_FLAG' on F.Cu (priority 3) overlaps queued zone …` warnings disappear.
- `Auto-pour: created N zone(s) for …` log line no longer lists `PWR_FLAG`.
- `Auto-skip: … (pour nets — use zone fill)` log line no longer lists `PWR_FLAG`.
- Module docstring documents the ERC-marker exclusion behavior.
- Classifier behavior is unchanged (`classify_from_name("PWR_FLAG")` still returns `NetClass.POWER`) — the fix lives in the pour-net selector.

## Test plan

- [x] New `TestErcMarkerNetExclusion` class in `tests/test_auto_pour.py` (5 tests):
  - `test_pwr_flag_excluded_from_pour_nets` — PWR_FLAG never produces a zone even when classified POWER.
  - `test_user_named_flag_variant_excluded` — `+3V3_FLAG` and similar variants are also filtered.
  - `test_pwr_flag_only_does_not_count_as_power_only_board` — board with only `PWR_FLAG` + signals does not trigger the all-power early-return.
  - `test_pwr_flag_classifier_unchanged` — pins down the design decision that the classifier still returns POWER.
  - `test_is_erc_marker_net_helper` — covers the helper's positive and negative cases (e.g. legitimate `PWR_5V` is *not* filtered).
- [x] New `TestAutoSkipExcludesErcMarkers` class in `tests/test_pour_net_auto_skip.py` (2 tests):
  - `test_pwr_flag_not_in_auto_skip` — `_auto_skip_pour_nets` returns no PWR_FLAG entry.
  - `test_pwr_flag_not_in_auto_skip_log` — `Auto-skip:` and `Routing:` log lines no longer mention `PWR_FLAG`.
- [x] All 17 `tests/test_auto_pour.py` tests pass (12 existing + 5 new).
- [x] All 41 `tests/test_pour_net_auto_skip.py` tests pass (39 existing + 2 new).
- [x] All 75 existing `tests/test_net_class.py` tests still pass (no classifier regression).

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -1015,6 +1015,7 @@ def _auto_skip_pour_nets(
     try:
         import re as _re
 
+        from kicad_tools.router.auto_pour import _is_erc_marker_net
         from kicad_tools.router.net_class import classify_and_apply_rules
 
         pcb_text = pcb_path.read_text()
@@ -1042,10 +1043,19 @@ def _auto_skip_pour_nets(
                 nets_with_zones.add(zm.group(1))
             del pcb_text  # free memory
 
+            # ERC-marker nets (PWR_FLAG and friends) are misclassified as
+            # pour nets by the name pattern but carry no copper and have
+            # no zone -- exclude them from both the auto-skip and the
+            # "pour nets without zones" warning so the user does not see
+            # a misleading "use zone fill" log line for a non-existent
+            # zone.  See router/auto_pour.py for the filter.
             auto_skip = [
                 name
                 for name, routing in net_class_map.items()
-                if routing.is_pour_net and name not in skip_nets and name in nets_with_zones
+                if routing.is_pour_net
+                and name not in skip_nets
+                and name in nets_with_zones
+                and not _is_erc_marker_net(name)
             ]
             if auto_skip:
                 skip_nets.extend(auto_skip)
@@ -1053,11 +1063,15 @@ def _auto_skip_pour_nets(
                     print(
                         f"Auto-skip: {', '.join(sorted(auto_skip))} (pour nets \u2014 use zone fill)"
                     )
-            # Warn about pour nets without zones
+            # Warn about pour nets without zones (excluding ERC markers
+            # which never get zones by design).
             no_zone = [
                 name
                 for name, routing in net_class_map.items()
-                if routing.is_pour_net and name not in skip_nets and name not in nets_with_zones
+                if routing.is_pour_net
+                and name not in skip_nets
+                and name not in nets_with_zones
+                and not _is_erc_marker_net(name)
             ]
             if no_zone and not quiet:
                 print(f"Routing: {', '.join(sorted(no_zone))} (power nets without zones)")
--- a/src/kicad_tools/router/auto_pour.py
+++ b/src/kicad_tools/router/auto_pour.py
@@ -11,12 +11,55 @@ voltage divider whose only nets are VIN, VOUT, GND) from being
 inadvertently drained of routable nets: auto-pour is only applied when
 the board has at least one signal (non-power/ground) net, indicating
 that the power nets are infrastructure rather than the entire design.
+
+ERC-marker net exclusion
+------------------------
+
+KiCad's netlister exports the ``PWR_FLAG`` virtual symbol as if it were
+an ordinary net, even though the symbol carries no electrical
+connection -- its sole purpose is to silence the *Input Power pin not
+driven by any Output Power pin* ERC error.  Because the resulting net
+name starts with ``PWR``, the name-based classifier in
+:mod:`kicad_tools.router.net_class` matches it as :class:`NetClass.POWER`
+and would otherwise produce a spurious zone.  See
+:func:`_is_erc_marker_net` and the ``ERC_MARKER_NET_PATTERNS`` constant
+below for the filter that excludes such names from the pour-net set.
+The schematic-side analogue lives in
+``src/kicad_tools/cli/sch_connectivity.py`` (``"PWR_FLAG is an ERC
+annotation, not a real net name"``).
 """
 
 from __future__ import annotations
 
 import re
 from pathlib import Path
+
+# ERC-only marker nets that must never be poured.  These are synthetic
+# net names emitted by KiCad's netlister for symbols whose sole purpose
+# is to silence ERC, not to carry copper.  The schematic-side connectivity
+# walker has the equivalent carve-out in
+# ``src/kicad_tools/cli/sch_connectivity.py`` (search for
+# ``"PWR_FLAG is an ERC annotation"``).
+#
+# Patterns are anchored regular expressions; ``_is_erc_marker_net`` uses
+# ``re.match`` (anchors at the start) so each entry is tested against
+# the full net name.
+ERC_MARKER_NET_PATTERNS: tuple[str, ...] = (
+    r"^PWR_FLAG$",  # KiCad's stock power:PWR_FLAG symbol
+    r"^#FLG(?:\d*|_.*|$)",  # Reference-designator spelling (#FLG, #FLG01, #FLG_VBUS)
+    r".*_FLAG$",  # User-named flag variants (e.g., +3V3_FLAG, VBUS_FLAG)
+)
+
+
+def _is_erc_marker_net(name: str) -> bool:
+    """Return True when *name* is an ERC-only marker, not a real net.
+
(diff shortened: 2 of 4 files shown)


rjwalters:

### Verdict
LGTM. The fix is minimal, correctly placed, well-tested, and exactly mirrors the schematic-side carve-out at `src/kicad_tools/cli/sch_connectivity.py:119-125`.

### Acceptance criteria — all satisfied

- [x] `kct route` no longer pours `PWR_FLAG` zones (filter in `auto_pour_if_missing` at `src/kicad_tools/router/auto_pour.py:262-269` skips ERC-marker nets before they enter `pour_nets`)
- [x] Misleading `Auto-skip: ... PWR_FLAG (pour nets — use zone fill)` and `Routing: ... PWR_FLAG (power nets without zones)` log lines suppressed (`src/kicad_tools/cli/route_cmd.py:1046-1078`)
- [x] Module docstring documents the exclusion behavior (`auto_pour.py:14-29`)
- [x] Classifier behavior preserved (`test_pwr_flag_classifier_unchanged` pins it down — POWER is still returned for downstream consumers like trace-width selection)
- [x] Unit tests for the auto-pour pour-net selector cover the exclusion (5 new tests in `tests/test_auto_pour.py::TestErcMarkerNetExclusion`, plus 2 in `tests/test_pour_net_auto_skip.py::TestAutoSkipExcludesErcMarkers`)

### Code quality

- **Design placement is correct.** Filtering at the pour-net selection layer (rather than in `router/net_class.py`) is the right call: `is_pour_net=True` for `PWR_FLAG` is not strictly wrong, and the classifier still drives trace-width application via `apply_net_class_rules` for any consumer that genuinely wants that net's class. The carve-out is local to the consumers that would actually emit copper/zones.
- **Pattern surface is well-considered.** `ERC_MARKER_NET_PATTERNS` covers the canonical `PWR_FLAG`, the reference-designator spelling `#FLG*`, and user-named `*_FLAG` variants. I verified the regex behaviour against positive (`PWR_FLAG`, `+3V3_FLAG`, `VBUS_FLAG`, `#FLG`, `#FLG01`, `#FLG_VBUS`) and negative (`PWR_5V`, `+3.3V`, `GND`, `SIG1`, `FLAG_OUT`, `PWR_FLAG_NOT`, `XFLAG`) cases — all behave as documented.
- **Helper is private but reachable.** `_is_erc_marker_net` is imported by `route_cmd._auto_skip_pour_nets` lazily inside the `try` block, which keeps the existing fallback semantics if the import ever fails. No circular-import risk introduced (verified `from kicad_tools.router.auto_pour import _is_erc_marker_net` resolves cleanly).
- **All-power guard correctness.** The `continue` in the classification loop runs *before* `signal_net_count` is incremented, so a board whose only "power" net is `PWR_FLAG` is correctly seen as having zero pour nets (not as a power-only board) — covered by `test_pwr_flag_only_does_not_count_as_power_only_board`.
- **Belt-and-suspenders in route_cmd is justified.** Even though `auto_pour_if_missing` already filters ERC markers from zone creation, the `_auto_skip_pour_nets` path runs against the existing PCB text and uses the classifier directly — without the second filter, `PWR_FLAG` could still appear in the `Routing: ... (power nets without zones)` warning when no zones were ever generated for it.

### Tests

- Local run: `pytest tests/test_auto_pour.py tests/test_pour_net_auto_skip.py tests/test_net_class.py -q` — **133 passed in 74.80s** ✓
- New `TestErcMarkerNetExclusion` (5 tests) and `TestAutoSkipExcludesErcMarkers` (2 tests) directly exercise the bug, the user-named variant, the all-power-guard interaction, the classifier-unchanged invariant, and the helper's positive/negative cases.
- The `test_is_erc_marker_net_helper` case `assert not _is_erc_marker_net("PWR_5V")` is particularly valuable — it pins down that the filter does not over-match real per-rail names.

### CI

- Lint & Format: pass
- Type Check: pass
- Test: pass (6m4s)
- C++ Build Check: pass
- Routed PCB DRC Check: pass
- mergeStateStatus: **CLEAN**, mergeable: **MERGEABLE**

1. `r"^PWR_FLAG$"` is subsumed by `r".*_FLAG$"` — the first entry is harmless documentation but technically redundant. Leaving it in is reasonable as a self-documenting "primary case" entry. Not a defect.
2. The patterns assume `re.match` semantics (start-anchor implicit); the `$` end-anchors are explicit. The trailing `(?:\d*|_.*|$)` alternation in the `#FLG` pattern is slightly overcomplicated — `^#FLG(\d.*|_.*)?$` would express the same intent more readably. Stylistic only.

Approving. Removing `loom:review-requested`, adding `loom:pr`.

PR #2599, opened 2026-05-10
fix(router): pop stale sys.modules entry + invalidate import caches before reload

### LINKED ISSUE #2594: kct route: auto-built C++ backend fails to load in same process — falls back with error after explicit --backend cpp
## Problem

#2549 / PR #2553 added auto-build for the C++ router extension. The build itself works, but the running \`kct\` process does not pick up the freshly-compiled \`router_cpp.*.so\`:

\`\`\`
C++ router extension missing -- building (~30s, one-time)...
Note: C++ build succeeded but module reload did not pick it up; falling back to Python (re-run kct to use C++).
Error: C++ backend requested but not available.
Build the C++ extension or use --backend auto/python.
See README for build instructions.
\`\`\`

The user invoked \`--backend cpp\` explicitly. The backend was successfully built. Yet the run aborts with exit code 0 + an error rather than using the backend it just built. This forces a manual second invocation.

## Repro

\`\`\`bash
# From a clean tree (no router_cpp.*.so):
rm -f src/kicad_tools/router/router_cpp.*.so
uv run kicad-tools route /tmp/any.kicad_pcb -o /tmp/out.kicad_pcb \\
  --strategy negotiated --backend cpp --no-cache
# Build runs (~30s), then errors out with "C++ backend requested but not available"
# Run the same command again — succeeds, using the now-built .so
\`\`\`

Observed at \`6926b47c\`.

## Expected behavior

When auto-build succeeds, either:
1. Re-import the freshly-built module via \`importlib.invalidate_caches()\` + \`importlib.import_module\` and continue, OR
2. Re-exec the command in a fresh subprocess transparently to the user

Falling back to Python (or, in this case, hard-erroring) after explicit \`--backend cpp\` is wrong: the user paid the 30s build cost specifically to get the C++ path. They expect a successful run, not a manual restart.

## Acceptance criteria

- \`kct route --backend cpp\` after \`rm router_cpp.*.so\` builds AND uses the C++ backend in a single invocation
- No "C++ backend requested but not available" hard-error after a successful auto-build
- Add an integration test that removes the .so before invoking and verifies a single-shot success

rjwalters:
## Summary

Fixes the in-process module-reload path that loads the freshly-built C++ router extension after `_attempt_auto_build()` writes a new `router_cpp.*.so` to disk. Before this PR, `kct route --backend cpp` on a fresh checkout (no `.so` present) would build the extension successfully but then print the misleading `Error: C++ backend requested but not available.` and exit with code 1, requiring the user to re-run the command.

The root cause was that `importlib.reload(cpp_backend)` alone does not pick up a newly-written native extension when the original `from . import router_cpp` failed at startup:
1. The failed import left a negative/partial entry for `kicad_tools.router.router_cpp` in `sys.modules` that Python re-uses on subsequent imports instead of re-running the module finder.
2. The parent package's `FileFinder` had a cached directory listing that did not yet include the freshly-written `.so`.

This implements the curator's recommended Option A from #2594: drop the stale `sys.modules` entry and call `importlib.invalidate_caches()` before `importlib.reload()` in both `_reload_cpp_backend()` and the verification block in `build_native()`. Option B/C (subprocess re-exec) was not needed — Option A reliably succeeds in the same process for the fresh-checkout case.

## Changes

- `src/kicad_tools/router/cpp_backend.py::_reload_cpp_backend()`: drop `sys.modules["kicad_tools.router.router_cpp"]` and call `importlib.invalidate_caches()` before `importlib.reload(cpp_backend)`. Updated the misleading "fall back to Python (re-run kct to use C++)" hint since this path is now effectively unreachable for the common case.
- `src/kicad_tools/cli/build_native_cmd.py::build_native()`: same cache-flush logic for the verification block at lines 348-380. Note: must use the module-level `sys` (already imported at line 15), not a local `import sys` shadow which would trigger `UnboundLocalError` on the earlier `sys.executable` reference at the cmake configure step.
- `tests/test_router_cpp_auto_build.py`: added `TestReloadCppBackend` with three regression tests pinning the cache-flush behavior:
  - `test_reload_pops_stale_router_cpp_from_sys_modules`: plants a sentinel in `sys.modules` and verifies it is dropped after reload.
  - `test_reload_calls_invalidate_caches`: spies on `importlib.invalidate_caches` and asserts it is called.
  - `test_reload_recovers_when_router_cpp_was_set_to_none`: integration test that forces `cpp_backend` into the post-failed-import state (router_cpp=None, _CPP_AVAILABLE=False) and asserts that `_reload_cpp_backend()` recovers the real `.so` from disk. Skipped when no `.so` is available.

| Fresh-checkout `kct route --backend cpp --no-cache` succeeds on first invocation | Yes | Manually verified: `rm -f router_cpp.*.so && uv run kicad-tools route ... --backend cpp --no-cache` builds and routes in a single process. C++ backend engaged (sub-second per-net routing times). |
| `is_cpp_available()` returns True after auto-build in same process | Yes | Direct repro: `_attempt_auto_build()` returns True; `is_cpp_available()` is True after; integration test in `tests/test_router_cpp_auto_build.py::test_reload_recovers_when_router_cpp_was_set_to_none` pins this. |
| No `Error: C++ backend requested but not available` when auto-build succeeded | Yes | Confirmed by manual test above; the message is now reserved for genuine build failures (toolchain missing, build error). |
| `--backend auto` behaviour unchanged: success uses C++, failure falls back to Python | Yes | Verified: with toolchain present + missing `.so`, `--backend auto` builds and engages C++. With toolchain missing, falls back to Python with existing warning (no exit code). |
| `--backend python` unchanged: never auto-builds, never reloads | Yes | No change to that code path; existing `TestBackendPython::test_python_skips_auto_build` still passes. |
| "module reload did not pick it up" message removed/unreachable | Yes | Message retained as defensive diagnostic for genuinely pathological reload failures (e.g. dlopen failure on the new .so). Removed the misleading "(re-run kct to use C++)" parenthetical. |
| Existing tests in `tests/test_router_cpp_auto_build.py` still pass | Yes | All 27 existing tests pass (the patched-stub fixture path is unchanged). |
| Option C re-exec only triggers under `--backend cpp` after in-process build whose reload still failed | N/A | Not implemented — Option A is sufficient (curator: "common case (Option A succeeds) is fast"). |

## Test Plan

Manual single-shot reproduction (the killer test):
```bash
find src/kicad_tools/router -name "router_cpp.*.so" -delete
uv run kicad-tools route boards/01-voltage-divider/output/voltage_divider.kicad_pcb \
  -o /tmp/out.kicad_pcb --backend cpp --no-cache
# Before fix: builds, then prints "Error: C++ backend requested but not available." and exits 1.
# After fix:  builds, then routes in the same process with the C++ backend engaged. Exit 0.
```

Direct verification of the in-process reload:
```bash
find src/kicad_tools/router -name "router_cpp.*.so" -delete
uv run python -c "
from kicad_tools.router.cpp_backend import is_cpp_available, _attempt_auto_build
print('Before:', is_cpp_available())  # False
ok = _attempt_auto_build(quiet=False)
print('After:', is_cpp_available(), 'Result:', ok)  # True, True
"
```

Automated tests:
```bash
uv run pytest tests/test_router_cpp_auto_build.py tests/test_route_cmd_params.py tests/test_route_exit_codes.py
# All pass (27+109 tests).
```

## Notes

- Pre-existing test failures in `tests/test_router_cpp_backend.py::TestCppPathfinderPythonFallback` (5 tests) are unrelated to this PR — they exist on `main` as well, but are normally skipped because the .so is not built in CI/dev environments. Those tests use `MagicMock` to patch a read-only C++ binding attribute (`pf._impl.route`) which fails with `AttributeError: 'kicad_tools.router.router_cpp.Pathfinder' object attribute 'route' is read-only`. Out of scope for this PR.
- The nanobind "leaked instances/types/functions" warnings printed at process exit when reloading the module are pre-existing — they happen because `importlib.reload(cpp_backend)` re-imports `router_cpp` while old type/function objects are still referenced. Addressing those would require a separate change to nanobind binding lifetime management. Not blocking.

Closes #2594

### DIFF
--- a/src/kicad_tools/cli/build_native_cmd.py
+++ b/src/kicad_tools/cli/build_native_cmd.py
@@ -357,9 +357,11 @@ def build_native(
             # ``sys.modules`` for ``kicad_tools.router.router_cpp``.  We
             # must drop that stale entry and invalidate the path-based
             # caches so the reloaded ``cpp_backend`` actually picks up
-            # the freshly-written ``.so``.
+            # the freshly-written ``.so``.  ``sys`` is already imported
+            # at module scope -- do NOT shadow it with a local import
+            # here (would trigger ``UnboundLocalError`` on the earlier
+            # ``sys.executable`` reference at the cmake configure step).
             import importlib
-            import sys
 
             sys.modules.pop("kicad_tools.router.router_cpp", None)
             importlib.invalidate_caches()
--- a/src/kicad_tools/cli/build_native_cmd.py
+++ b/src/kicad_tools/cli/build_native_cmd.py
@@ -348,7 +348,21 @@ def build_native(
         # Verify the installation
         try:
             # Clear any cached imports
+            #
+            # Issue #2594: ``importlib.reload(cpp_module)`` alone is not
+            # enough when the original ``from . import router_cpp`` failed
+            # at startup (fresh checkout with no .so on disk).  Python
+            # caches the negative result on the parent package's
+            # ``FileFinder`` and may leave a partial / ``None`` entry in
+            # ``sys.modules`` for ``kicad_tools.router.router_cpp``.  We
+            # must drop that stale entry and invalidate the path-based
+            # caches so the reloaded ``cpp_backend`` actually picks up
+            # the freshly-written ``.so``.
             import importlib
+            import sys
+
+            sys.modules.pop("kicad_tools.router.router_cpp", None)
+            importlib.invalidate_caches()
 
             import kicad_tools.router.cpp_backend as cpp_module
 
--- a/src/kicad_tools/router/cpp_backend.py
+++ b/src/kicad_tools/router/cpp_backend.py
@@ -135,6 +135,23 @@ def _reload_cpp_backend() -> bool:
     to :func:`is_cpp_available` return ``True`` and the C++ backend is
     actually used by the routing code.
 
+    Issue #2594: ``importlib.reload(cpp_backend)`` alone is *not* enough.
+    The original ``from . import router_cpp`` failed at startup, leaving:
+
+      1. A negative/partially-initialised entry for
+         ``kicad_tools.router.router_cpp`` in :data:`sys.modules` that
+         Python re-uses on subsequent imports instead of re-running the
+         module finder.
+      2. A cached directory listing on the parent package's
+         ``FileFinder`` that does not yet contain the freshly-written
+         ``router_cpp.*.so``.
+
+    To pick up the new ``.so`` in the same process we MUST drop the
+    stale ``sys.modules`` entry AND call
+    :func:`importlib.invalidate_caches` before the reload.  This is the
+    safe case: the previous attempt never succeeded, so there is no
+    initialised native module to clash with.
+
     Returns:
         ``True`` if the backend is available after reload, ``False`` otherwise.
     """
@@ -142,6 +159,18 @@ def _reload_cpp_backend() -> bool:
 
     import importlib
     import sys as _sys
+
+    # 1. Drop any stale negative/partial entry for the C++ extension itself.
+    #    A failed ``from . import router_cpp`` at startup leaves a None or
+    #    partially-initialised module in sys.modules that Python re-uses
+    #    on subsequent imports.
+    _sys.modules.pop("kicad_tools.router.router_cpp", None)
+
+    # 2. Tell the import machinery to re-scan filesystem-based finders
+    #    for the freshly-written .so. Without this, the parent package's
+    #    FileFinder may keep its cached directory listing from before the
+    #    build wrote the new file.
+    importlib.invalidate_caches()
 
     module_name = __name__  # "kicad_tools.router.cpp_backend"
     module = _sys.modules.get(module_name)
@@ -340,11 +369,19 @@ def _attempt_auto_build(*, quiet: bool) -> bool:
     # (``_CPP_AVAILABLE``, ``router_cpp``) reflect the freshly written
     # ``.so``.  Without this, ``is_cpp_available()`` continues to return
     # ``False`` for the rest of the process.
+    #
+    # Issue #2594: ``_reload_cpp_backend`` now invalidates the import
+    # caches and pops the stale ``router_cpp`` entry from ``sys.modules``
+    # before reloading, so this is expected to succeed in the same
+    # process.  The ``not available`` branch is retained as a defensive
+    # diagnostic for genuinely pathological reload failures (e.g. a
+    # platform where dlopen of the new .so itself fails after the build
+    # wrote it to disk).
     available = _reload_cpp_backend()
     if not available and not quiet:
         print(
             "Note: C++ build succeeded but module reload did not pick it up; "
-            "falling back to Python (re-run kct to use C++).",
+            "falling back to Python.",
             flush=True,
         )
     return available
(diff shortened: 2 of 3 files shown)


rjwalters:

This PR cleanly fixes the in-process module-reload regression from #2594 by combining the two cache flushes that were missing: dropping the stale `kicad_tools.router.router_cpp` entry from `sys.modules` AND calling `importlib.invalidate_caches()` before `importlib.reload(cpp_backend)`. Applied at both call sites (`_reload_cpp_backend()` in cpp_backend.py and the verification block in build_native_cmd.py). Surgical 175/-1 diff.

### Verification

**Killer test (the original repro from #2594) — PASSES:**
```bash
find src/kicad_tools/router -name "router_cpp.*.so" -delete
uv run kicad-tools route boards/01-voltage-divider/output/voltage_divider.kicad_pcb \
  -o /tmp/out.kicad_pcb --backend cpp --no-cache
```
Output:
- `C++ router extension missing -- building (~30s, one-time)...`
- Build succeeds, routing proceeds in the same invocation with sub-second per-net times (C++ backend engaged)
- **No** "Error: C++ backend requested but not available" message anywhere
- Exit succeeded (the `Routing Incomplete (33%)` message is an unrelated routing-quality issue on this small board, not a backend availability error)

**Direct in-process state verification:**
```
Before: False
C++ router extension missing -- building (~30s, one-time)...
After: True Result: True
```
`is_cpp_available()` correctly transitions False → True after `_attempt_auto_build()` in the same process.

**`--backend auto` verified:** builds and engages C++ on missing .so, no Python fallback warning.
**`--backend python` verified:** no auto-build attempted, no reload, no warning.

### Tests

- All 28 tests in `tests/test_router_cpp_auto_build.py` pass (25 existing + 3 new).
- The 3 new tests (`TestReloadCppBackend`) pin the cache-flush behavior:
  - `test_reload_pops_stale_router_cpp_from_sys_modules` plants a sentinel and verifies it is dropped.
  - `test_reload_calls_invalidate_caches` spies on `importlib.invalidate_caches` and asserts call count >= 1.
  - `test_reload_recovers_when_router_cpp_was_set_to_none` exercises the full reload path (skipped without .so).
- The 5 pre-existing `TestCppPathfinderPythonFallback` failures (`'kicad_tools.router.router_cpp.Pathfinder' object attribute 'route' is read-only`) are confirmed to exist on main as well — unrelated to this PR, out of scope as the builder noted.

### CI

All required checks green:
- Lint & Format: pass
- Type Check: pass
- Test: pass (6m8s)
- C++ Build Check: pass
- Routed PCB DRC Check: pass

Merge state: **CLEAN, MERGEABLE**.

### Code-quality notes

- Correct call to module-level `sys` in build_native_cmd.py (no local shadow that would `UnboundLocalError` on the earlier `sys.executable` reference) — explicitly called out in the comment block, good defensive documentation.
- Updated misleading "(re-run kct to use C++)" hint in `_attempt_auto_build` to plain "falling back to Python." since the message path is now effectively unreachable for the common case.
- Comment blocks in both files clearly explain the two-fold root cause (stale sys.modules + cached FileFinder listing). Future maintainers will not accidentally remove the cache flushes.
- Nanobind reference-leak warnings at exit are pre-existing reload artifacts (acknowledged in PR description) and would require separate work in the binding layer.

Approving — ready to merge.

Issue #2625, opened 2026-05-11
fix(validate): --sync reports every footprint as orphan on hierarchical schematics


rjwalters:
## Problem

\`kct validate --sync\` reports every PCB footprint as \"orphan on PCB\" when the schematic is hierarchical, even when \`kct pcb sync-netlist --dry-run\` reports \"in sync.\"  The two tools enumerate the schematic by different code paths and disagree on what's in it.

## Reproducer

\`\`\`
$ uv run kct pcb sync-netlist \\
    --schematic chorus-test-revA.kicad_sch --dry-run chorus-test-revA_v19.kicad_pcb
PCB Sync Netlist (dry run)
  No changes needed - PCB is in sync with schematic.

$ uv run kct validate --sync \\
    --schematic chorus-test-revA.kicad_sch --pcb chorus-test-revA_v19.kicad_pcb
============================================================
NETLIST SYNC VALIDATION
============================================================
Results:
  Errors:     0
  Warnings:   90
ORPHANED ON PCB (90):
  [!] WARNING: C1 on PCB has no schematic symbol
  [!] WARNING: C10 on PCB has no schematic symbol
  ...
\`\`\`

90 false-positive orphan warnings for a board whose schematic literally has 90 placed symbols.

## Likely cause

\`kct validate --sync\` enumerates schematic symbols from only the root \`.kicad_sch\` file (which on a hierarchical project contains \`(sheet ...)\` blocks but no placed \`(symbol ...)\` instances).  Confirmed indirectly: \`kct symbols <root-sch>\` returns \"No symbols found\" on the same file.

The correct behavior — already implemented by \`kct pcb sync-netlist\` — walks every sub-sheet via the project's \`(sheet ...)\` references and aggregates their placed symbols.

## Fix sketch

In the \`validate --sync\` codepath, replace the root-only symbol enumeration with the same hierarchical walker that \`pcb sync-netlist\` uses.  Audit other validate flags (\`--lvs\`, \`--consistency\`) for the same bug.

## Acceptance criteria

- [ ] \`kct validate --sync\` reports 0 orphans on a hierarchical project that \`kct pcb sync-netlist --dry-run\` says is in sync
- [ ] All flat-schematic existing tests still pass
- [ ] A regression test uses a hierarchical fixture (e.g. a copy of \`boards/05-bldc-motor-controller\` or a synthetic two-sheet project) and asserts orphan count == 0 when in sync
[...]

PR #2643, opened 2026-05-11
fix(validate): hierarchical extract_netlist for --consistency net checks (closes #2633)

### LINKED ISSUE #2633: fix(validate): extract_netlist() is root-only — --consistency property check misses sub-sheet pins
## Problem

Follow-up to #2625 (which fixed root-only enumeration in `--sync` / `--consistency` component checks).  A second, deeper case of root-only enumeration remains in the same `--consistency` codepath:

`src/kicad_tools/validate/consistency.py:807` calls `ModelsSchematic.load(sch_path).extract_netlist()` to build `pin_nets` for property-consistency checks.  `extract_netlist()` is defined on `SchematicNetlistMixin` (`src/kicad_tools/schematic/models/netlist_mixin.py:161`) and operates on `self.wires` only — i.e. the wires in the loaded sheet.  No traversal of `(sheet ...)` references; no recursion into sub-sheets.

Result: on a hierarchical design, `_extract_schematic_pin_nets` returns only the root sheet's pin-to-net mapping.  `_check_properties` then has no data for any sub-sheet pin, so property consistency between schematic and PCB is silently unverified for everything outside the root.

## Reproducer

```python
from kicad_tools.schematic.models import Schematic
sch = Schematic.load('tests/fixtures/hierarchical/root.kicad_sch')
nl = sch.extract_netlist()
# Returns only root-sheet refs (e.g. R1) — sub_a's R2, C1 and sub_b's R3, R4 are absent.
```

## Why this matters

`--consistency` is now half-fixed.  Component existence (`_check_components`) is hierarchical after #2625; property consistency (`_check_properties`) is not.  A user running `kct validate --consistency` on a hierarchical project gets a false-confidence pass for any wrong-value or wrong-footprint on a sub-sheet component.

## Proposed fix

Two options:

1. **Add a hierarchical variant alongside `extract_netlist()`** — e.g. `extract_netlist_hierarchical()` that walks sub-sheets, merges netlists by name (matching connect-by-name semantics), and resolves hierarchical-label / sheet-pin connections.  Have `_extract_schematic_pin_nets` use the hierarchical variant.

2. **Make `extract_netlist()` itself hierarchical when a path is available.**  Lower risk if other callers want flat behaviour: keep a `hierarchical=False` default and have `_extract_schematic_pin_nets` opt in.

Option 2 is the safer minimal change.  Audit callers of `extract_netlist()` before flipping the default.

## Acceptance criteria

- [ ] `extract_netlist()` (or a new sibling) walks every sub-sheet referenced by the root and merges their netlists with hierarchical-label / sheet-pin connections respected.
- [ ] `_extract_schematic_pin_nets` returns pin-to-net entries for every placed symbol in the hierarchy, not just the root.
- [ ] Regression test using `tests/fixtures/hierarchical/root.kicad_sch` asserts that R2, C1, R3, R4 (sub-sheet refs) all appear in the returned mapping.
- [ ] `_check_properties` on a hierarchical project flags a deliberately-mismatched sub-sheet component (e.g. wrong footprint on R2) — pre-fix, this is silently missed.
- [ ] All existing `extract_netlist()` callers either continue to work with flat behaviour (Option 2) or get audited and updated (Option 1).

## Related

- #2625 (sibling fix: --sync / --consistency component enumeration; the #2625 builder explicitly noted this as a possible follow-up).

rjwalters:
## Summary

- Closes #2633.
- `SchematicPCBChecker._check_nets` called `_extract_schematic_pin_nets()`, which invoked the root-only `extract_netlist()`. Every sub-sheet pin was absent from the returned dict, so the `sch_refs & pcb_refs` intersection in `_check_nets` silently skipped every sub-sheet ref. Net-name mismatches on sub-sheet pads were therefore never reported by `kct validate --consistency`.
- Adds a `hierarchical: bool = False` flag to `SchematicNetlistMixin.extract_netlist()`. When `True`, the method delegates to the existing hierarchical walker `kicad_tools.operations.netlist._collect_hierarchy_components`, which already handles per-sheet net extraction, sheet-pin / hierarchical-label merging across sheets, and circular-reference detection.
- `_extract_schematic_pin_nets()` in `validate/consistency.py` opts in via `extract_netlist(hierarchical=True)`.
- `Schematic.load(path)` now stores the source path on `_saved_path` so the walker can resolve relative sheet references. Previously only `write()` set this attribute.

PR #2625 already fixed the component-enumeration side of the same hierarchical blind spot (`_check_components`, `_check_properties` via `extract_bom(hierarchical=True)`); `_check_nets` was the remaining victim. The curator notes on #2633 corrected the original issue title accordingly.

## Implementation notes

- Default `extract_netlist()` behaviour (no kwarg) is byte-identical to before the change: only the root sheet's wires, symbols, labels, and power symbols are considered. All other callers (`pins_on_net`, `_collect_hierarchy_components` itself, doctests) continue to use the flat path.
- The import of `operations/netlist.py` is lazy inside the `hierarchical=True` branch to avoid an upward dependency from `schematic/models/` on `operations/`.
- If `hierarchical=True` is passed on a Schematic that was never loaded from disk (no path available to walk sub-sheets from), the method raises `ValueError` with an explicit message rather than silently falling back to root-only. This mirrors the `RuntimeWarning` upstream behaviour that #2625 added to `_check_components`.

## Test plan

- [x] New `tests/test_validate_hierarchical_nets.py` (6 tests, all passing):
  - `test_default_is_root_only` — default extract_netlist still returns only root sheet refs.
  - `test_hierarchical_walks_every_subsheet` — `hierarchical=True` returns the full set `{R1, R2, C1, C2, R3, R4}` from the fixture tree.
  - `test_hierarchical_requires_loaded_path` — `ValueError` on an unsaved Schematic.
  - `test_pin_nets_include_every_subsheet_ref` — `_extract_schematic_pin_nets()` covers every sub-sheet ref with both pins.
  - `test_in_sync_hierarchical_reports_no_net_issues` — in-sync hierarchical PCB triggers zero `domain=net` issues.
  - `test_subsheet_net_mismatch_is_detected` — deliberately wrong net on R2.1 (lives in sub_a.kicad_sch) is reported as exactly one `domain=net` `ConsistencyIssue`. Pre-fix this assertion fails because the root-only walker never returned R2.
- [x] `pytest tests/test_netlist_query.py tests/test_consistency.py tests/test_hierarchical_netlist.py tests/test_validate_hierarchical_sync.py` — 86 existing tests still pass (confirms default behaviour unchanged).
- [x] `pytest tests/test_schematic_run_erc.py tests/test_schematic_helpers.py tests/test_schematic.py` — 75 tests pass (confirms `Schematic.load` setting `_saved_path` does not regress write/run_erc round-trips).
- [x] `ruff check` and `ruff format --check` clean on changed files.
- Full suite (`pytest tests/` minus pre-existing CMA-ES / router-binding failures unrelated to this PR) — 14,966 passed, same 112 pre-existing failures as `main` (verified by stashing changes and re-running the affected test files).

### DIFF
--- a/src/kicad_tools/schematic/models/io_mixin.py
+++ b/src/kicad_tools/schematic/models/io_mixin.py
@@ -55,7 +55,12 @@ class SchematicIOMixin:
             raise FileNotFoundError(f"Schematic file not found: {path}")
 
         doc = parse_file(path)
-        return cls._from_sexp(doc)
+        sch = cls._from_sexp(doc)
+        # Track the source path so operations that need to walk
+        # sub-sheets (e.g., extract_netlist(hierarchical=True), run_erc)
+        # can resolve relative sheet references.
+        sch._saved_path = path
+        return sch
 
     @classmethod
     def _from_sexp(cls, doc: SExp) -> Schematic:
--- a/src/kicad_tools/schematic/models/netlist_mixin.py
+++ b/src/kicad_tools/schematic/models/netlist_mixin.py
@@ -91,9 +91,8 @@ class SchematicNetlistMixin:
         for junc in self.junctions:
             junc_pos = (round(junc.x, 2), round(junc.y, 2))
             for seg_start, seg_end in wire_segments:
-                if (
-                    junc_pos in (seg_start, seg_end)
-                    or point_on_segment(junc_pos, seg_start, seg_end)
+                if junc_pos in (seg_start, seg_end) or point_on_segment(
+                    junc_pos, seg_start, seg_end
                 ):
                     union(junc_pos, seg_start)
                     union(junc_pos, seg_end)
@@ -107,9 +106,7 @@ class SchematicNetlistMixin:
                 # Track this pin at this position
                 if pos_rounded not in point_to_pins:
                     point_to_pins[pos_rounded] = []
-                point_to_pins[pos_rounded].append(
-                    PinRef(symbol_ref=sym.reference, pin=pin.number)
-                )
+                point_to_pins[pos_rounded].append(PinRef(symbol_ref=sym.reference, pin=pin.number))
 
                 # Connect to wires
                 connect_to_wire(pos_rounded)
@@ -158,11 +155,23 @@ class SchematicNetlistMixin:
 
         return parent, point_to_net_names, point_to_pins
 
-    def extract_netlist(self) -> dict[str, list[PinRef]]:
+    def extract_netlist(self, hierarchical: bool = False) -> dict[str, list[PinRef]]:
         """Extract netlist from schematic.
 
         Analyzes schematic connectivity to build a mapping from net names
         to the pins connected to each net.
+
+        Args:
+            hierarchical: When ``True``, walk every sub-sheet referenced
+                by this schematic and return a single merged netlist
+                covering the whole hierarchy. Sheet-pin / hierarchical-
+                label connections in the parent context resolve to the
+                parent's net name. Requires the schematic to have been
+                loaded via :meth:`Schematic.load` so the source file
+                path is known. When ``False`` (default), behaviour is
+                byte-identical to previous releases: only the root
+                sheet's wires, symbols, labels, and power symbols are
+                considered.
 
         Returns:
             Dict mapping net names to list of connected pins.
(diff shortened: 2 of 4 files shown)


rjwalters:
## Judge review: APPROVE

Surgical fix at the right site (per curator scope correction on #2633): `_check_nets` was the real victim, not `_check_properties`. The builder correctly opts `_extract_schematic_pin_nets()` into the new `extract_netlist(hierarchical=True)` flag and delegates to the existing `_collect_hierarchy_components` walker.

### Verified

- **Surgical site**: `validate/consistency.py:872` flipped from `extract_netlist()` to `extract_netlist(hierarchical=True)`. `_check_properties` is untouched (already hierarchical-correct after #2625).
- **Default-False contract**: new `hierarchical: bool = False` parameter; the existing root-only body is unchanged below the guard. All three identified flat callers continue to work without modification:
  - `pins_on_net` (`netlist_mixin.py:354`) — no kwarg, stays flat.
  - `_collect_hierarchy_components` (`operations/netlist.py:599`) — no kwarg, stays flat (must, to avoid infinite recursion).
  - `_extract_schematic_pin_nets` — now explicitly opts in.
- **Lazy import** of `kicad_tools.operations.netlist._collect_hierarchy_components` is inside the `if hierarchical:` branch, so no module-load circular import. `operations/netlist.py` has no top-level dependency on `schematic/models/`, but the lazy import is still the right hygiene to avoid future-proofing risk.
- **`_saved_path` capture in `io_mixin.py` `Schematic.load`** is wired correctly. The attribute was already declared in `schematic.py:167` and previously set only by `write()`. Setting it after `_from_sexp(doc)` returns is consistent with the existing `run_erc` consumer that reads via `getattr(..., None)`. Round-trip behaviour is unaffected — `write()` still overwrites `_saved_path` to the new path, and the 75 tests in `test_schematic_run_erc.py + test_schematic_helpers.py + test_schematic.py` confirm zero regression.
- **6 new tests** in `tests/test_validate_hierarchical_nets.py` all pass locally. Bug-reproducing test (`test_subsheet_net_mismatch_is_detected`) correctly fails pre-fix because the root-only walker excludes R2 from `sch_refs & pcb_refs`; post-fix it reports the single expected `domain=net` `ConsistencyIssue` for `R2.1`.
- **No regression**: 86 tests in `test_netlist_query.py + test_consistency.py + test_hierarchical_netlist.py + test_validate_hierarchical_sync.py` pass; default behaviour is byte-identical.

### CI

All 6 CI checks pass on the PR:
- Lint & Format: pass
- Type Check: pass
- Test: pass (5m54s)
- C++ Build Check: pass
- Routed PCB DRC Check: pass

`mergeStateStatus=CLEAN`, `mergeable=MERGEABLE`.

### Observations (non-blocking)

- The PR also contains incidental ruff-format de-wrapping in `netlist_mixin.py` (lines 91-97, 106-108, 230, 354-356) that touches code outside the bug's blast radius. Verified via `ruff format --check` and `ruff check` that this is the canonical formatter output for the surrounding lines — not gratuitous churn.
- The `Schematic("Test")` no-path guard raises `ValueError` rather than silently falling back to root-only, which matches the explicit-failure pattern used by `run_erc` and the `RuntimeWarning` upstream behaviour introduced by #2625. Good defensive choice.

Approved. Ready to merge once labels flip.

PR #2645, opened 2026-05-11
feat(router): opt-in CoupledPathfinder engagement + single-ended refusal (closes #2638)

rjwalters:
## Summary

Phase 2E of Epic #2556 (first-class differential pair support). Introduces opt-in `CoupledPathfinder` engagement and an engagement-layer single-ended refusal so the #2527 lesson (USB-C `CC1`/`CC2`, `SBU1`/`SBU2` are orientation pins, NOT a diff pair) is enforced even when a designer accidentally pairs them via explicit `NetClassRouting.diffpair_partner`.

Refines the previously binary `--differential-pairs` flag: pairs are now coupled-routed only when their net class explicitly sets `coupled_routing=True` AND neither half matches the single-ended refusal pattern. Refused pairs fall through to the main strategy.

## Changes

- `src/kicad_tools/router/rules.py`: new `NetClassRouting.coupled_routing: bool = False` field with cross-reference note to the existing `use_coupled_routing` function parameter (curator-flagged name-collision risk).
- `src/kicad_tools/router/diffpair.py`: new `should_engage_coupled(pair, net_class_routing, net_to_class) -> (bool, str)` helper returning one of four reasons: `"engaged"`, `"single_ended_refusal"`, `"opt_in_disabled"`, `"no_class_match"`. Reuses `is_single_ended_refused` rather than re-importing the regex.
- `src/kicad_tools/router/diffpair_routing.py`:
  - New `_resolve_detection_inputs` helper synthesises `net_class_routing`/`net_to_class` from the autorouter's `net_class_map` so the explicit-declaration path (which had been dead for the common `net_class_map`-only case) actually fires.
  - New `_resolve_engagement` helper invokes the gate.
  - `route_diffpair_prepass` and `route_all_with_diffpairs` consult the gate before dispatching to `route_differential_pair_coupled`; refused pairs are logged with `[diffpair-engage] refused {pair}: {reason}` and dropped from `diff_net_ids` so the main strategy picks them up.
[...]

Issue #2638, opened 2026-05-11
[Epic #2556] Phase 2E: CoupledPathfinder engagement criteria (opt-in + single-ended refusal)


rjwalters:
**Epic**: #2556 - First-class differential pair support
**Phase**: 2 of 4 — Coupled Routing + Escape + Continuity DRC
**Phase Goal**: Coherent coupled routing engagement + diff-pair-aware escape + continuity DRC

Tracked under #2556.

## Description

Phase 1 (#2557/#2558/#2559/#2560/#2587, PR #2636) made within-pair clearance work end-to-end on board 03. The CoupledPathfinder (`src/kicad_tools/router/diffpair_routing.py:148`) is now invoked via `route_diffpair_prepass()` (line 1588) and `route_all_with_diffpairs()` (line 1675), and Phase 1B added layered detection that already refuses suffix-inferred single-ended pairs like `USB_CC1`/`USB_CC2` (see `diffpair.py:_SINGLE_ENDED_REFUSAL_PATTERN`).

However, **engagement is still binary and global**: the only switch is `DifferentialPairConfig.enabled` (`diffpair.py:442`), wired to the CLI's `--differential-pairs` flag (`route_cmd.py:3272,4492`). Once flipped, the prepass attempts coupled routing for **every** detected pair. This Phase 2 issue refines engagement so that:

1. CoupledPathfinder engagement becomes **opt-in per net class** via a new `NetClassRouting.coupled_routing` flag (default `False`). Pairs whose P or N net belongs to a class with `coupled_routing=True` are routed coupled; others fall through to independent routing even when `--differential-pairs` is on. This preserves backward compatibility for boards that just want the new intra-pair clearance without forcing coupled geometry.
2. The single-ended refusal in `diffpair.py` is **honored at the engagement layer**, not only at suffix detection. Specifically: if a designer accidentally declares `diffpair_partner: \"USB_CC2\"` on a `USB_CC1` net class (the #2527 lesson — these are orientation pins, NOT a diff pair, and coupling them would be electrically wrong), the coupled engagement step must refuse the pair with a clear log message and let the nets route independently.

### Curator lesson from #2527 (DO NOT IGNORE)

> USB_CC1/USB_CC2 are USB-C **orientation/role-detection** pins (Configuration Channel). Each is a single-ended low-speed signal. Routing them as a coupled pair would be electrically wrong (they don't switch differentially).

[...]

Issue #2556, opened 2026-05-08
Epic: First-class differential pair support across the routing pipeline


rjwalters:
# Epic: First-class differential pair support across the routing pipeline

## Overview

The router treats differential-pair nets (USB_D+/D-, PCIe TX/RX, MIPI lanes, DDR clocks, etc.) as independent nets and applies inter-net clearance rules to within-pair edges. This causes overlapping copper on tight-pitch footprints (today's evidence: board 03 USB_D+/USB_D- traces overlap by **-0.200mm** between J1 row A pads at JLCPCB rules) and prevents authoring HSDI boards.

Multiple closed PRs/issues have nibbled at diff-pair handling without a coherent system: #2473 (CoupledPathfinder N-pad), #2484 (CoupledPathfinder spacing), #2536 (board 03 USB pin reassignment), #2517/#2527 (curator explicitly rejected misuse of CoupledPathfinder for USB_CC1/CC2 because those are single-ended orientation pins). Today's audit (2026-05-08) showed even with `--differential-pairs --grid 0.05 --auto-layers` the structural overlap remains — the existing implementation is incomplete.

This epic builds the missing first-class subsystem: detection, separate within-pair clearance, coupled routing engagement, length matching, impedance control, DRC rules, escape routing, regression testbench, and documentation.

## Milestone Alignment

**Current Milestone**: #2394 (manufacturer-ready parity for all five example boards)
**Alignment Tier**: Tier 1 — Goal-Advancing
**Justification**: Board 03's last DRC error blocks #2394's "all five MFG-READY" goal. After today's other fixes, board 03 is the only board whose remaining DRC error is a diff-pair issue (1 error in the USB pair). Phase 1 of this epic directly unblocks #2394 for board 03.

### Phase 1: Foundation — Detection + Clearance
**Goal**: Reliable diff-pair detection + per-pair clearance support. Phase 1 alone unblocks board 03.
**Can parallelize**: Issues A and B independent; C depends on A; D depends on A+C

- [ ] Issue A: `NetClass.intra_pair_clearance` field separate from inter-pair `clearance` (default = `clearance` for backward compat)
- [ ] Issue B: Reliable diff-pair detection — explicit net-class declaration + suffix inference (existing) + KiCad DiffPair group support
- [ ] Issue C: Thread `intra_pair_clearance` through pathfinder + cpp_backend at the within-pair edge sites (`pathfinder.py:1485,2679`, `cpp_backend.py:859,1408`)
[...]

Issue #2527, opened 2026-05-06
Board 03: USB_CC1 still fails after #2513 escape fix — geometric constraint in U1↔J1 corridor


rjwalters:
## Problem

Discovered while landing PR #2526 (closed #2513). With #2526 applied, board 03 routing reaches 12/13 nets — escape coverage for U1 (TQFP-32 MCU) and J1 (2-row USB-C) is fixed — but **USB_CC1 still fails to route**. PR #2526's builder and judge both attribute the remaining failure to a geometric constraint: the inter-IC corridor between U1 and J1 is too narrow for all five MCU-side USB/CC nets to share the same outer layer.

> Verification on current `main` (PR #2526 not yet merged) shows 9/13 routed, 4 failed (USB_D+, USB_CC1, USB_CC2, XTAL2) and `kct check ... --mfr jlcpcb` reports 8 DRC errors. Once #2526 lands, the picture collapses to a single remaining failure (USB_CC1).

## Reproduction

```
uv run kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb \
  -o boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb \
  --timeout 240 --quiet
uv run kct check boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb \
  --mfr jlcpcb --format summary
```

Expected after #2526 merges: USB_CC1 reported as failed.

### Geometry: how tight is the U1↔J1 corridor?

Reading `boards/03-usb-joystick/output/usb_joystick.kicad_pcb`:

- **U1** (TQFP-32, 7×7 mm, 0.8 mm pitch) is at sheet `(130.0, 120.0)`. North edge pads (26-29) sit at local `y = -4.5`, so global `y_pad_top ≈ 115.5` mm. With a courtyard envelope of ~0.5 mm, U1 effectively claims `y ≥ ~115.0`.
- **J1** (`Connector_USB:USB_C_Receptacle_GCT_USB4105`) is at sheet `(130.0, 105.0)`. The southernmost J1 pads/tabs are the through-hole mounting tabs S1/S2 at local `y = +1.5`, so global `y_max(J1) ≈ 106.5` mm.
- **Available north-south corridor**: `~9 mm` (`106.5 → 115.5`).
[...]

PR #2646, opened 2026-05-11
feat(validate): add diffpair_routing_continuity DRC rule (closes #2640)

### LINKED ISSUE #2640: [Epic #2556] Phase 2G: New DRC rule diffpair_routing_continuity
**Epic**: #2556 - First-class differential pair support
**Phase**: 2 of 4 — Coupled Routing + Escape + Continuity DRC
**Phase Goal**: Coherent coupled routing engagement + diff-pair-aware escape + continuity DRC

Tracked under #2556.

## Description

Phase 1D (#2560) added the `diffpair_clearance_intra` DRC rule, which validates that within-pair edges respect the per-class `intra_pair_clearance` threshold. That rule is necessary but NOT sufficient — it only checks the **spacing** of segments that already happen to be coupled. It does NOT verify that the pair actually **travels coupled for a meaningful fraction of its length**.

A board where USB_D+ and USB_D- run parallel for 5% of their length and diverge for the remaining 95% would pass `diffpair_clearance_intra` (the 5% that's coupled satisfies the threshold) but fail at the electrical level (controlled coupling is lost over most of the run). Phase 2G adds a `diffpair_routing_continuity` rule that fires when the coupled fraction is below a configurable threshold.

## Definition

For each detected differential pair, compute:

- `L_p` = total routed length of the positive net (sum of segment lengths)
- `L_n` = total routed length of the negative net
- `L_coupled` = length of the P trace for which there exists an N trace within `intra_pair_clearance + tolerance` on the same copper layer, measured by sliding a window along the P trace and asking whether the orthogonal projection of the nearest N segment lies within the coupling window
- `coupled_fraction = L_coupled / min(L_p, L_n)`

The rule fires when `coupled_fraction < threshold` (default `0.7` — 70%; configurable per net class).

## Why this rule matters

Today's router (post-#2587) routes P and N independently for the majority of their path even when `intra_pair_clearance` is set; the coupling that happens is incidental, not enforced. Phase 2E adds opt-in engagement and Phase 2F adds paired escape, but **no rule validates that the final layout actually maintained coupling**. The continuity rule closes that loop: it's the regression guard for #2638 (engagement) and #2639 (escape) — if a future change silently disables coupled routing, the continuity rule catches it on every routed board.

## Affected files

- `src/kicad_tools/validate/rules/diffpair_routing_continuity.py` (new) — rule class mirroring `diffpair_clearance_intra.py`'s structure
- `src/kicad_tools/validate/checker.py` — register the new rule (`check_diffpair_routing_continuity` method, ~the pattern at the existing `check_diffpair_clearance_intra` method)
- `src/kicad_tools/validate/rules/__init__.py` — export `DiffPairRoutingContinuityRule`
- `src/kicad_tools/drc/violation.py` — add `ViolationType.DIFFPAIR_ROUTING_CONTINUITY = \"diffpair_routing_continuity\"` (line 124 area); add to `_TYPE_CATEGORY_MAP` as `ViolationCategory.ROUTING` (line 50 area); add to the alias table (line 237 area). The #2521 critical-gotcha pattern: a missing alias entry makes the rule invisible to fuzzy lookup — DON'T skip it.
- `src/kicad_tools/cli/check_cmd.py:60-62` — add `\"diffpair_routing_continuity\"` to `CHECK_CATEGORIES`
- `src/kicad_tools/cli/check_cmd.py:272-275` — add `\"diffpair_routing_continuity\": checker.check_diffpair_routing_continuity` to `check_methods`
- `tests/test_validate_diffpair_routing_continuity.py` (new)
- `tests/test_cli_check_diffpair_routing_continuity.py` (new) — mirror of `test_cli_check_diffpair_clearance_intra.py`

## Dependencies

**Depends on** (already merged): #2558 (layered detection — the rule reuses `detect_diff_pairs` to know which net pairs to validate), #2560 (rule plumbing pattern + `intra_pair_clearance` threshold).

**Depends on** (Phase 2): #2638 (Phase 2E — engagement). Rationale: the rule should only fire on pairs that were **engaged** for coupled routing. Pairs whose net class has `coupled_routing=False` are deliberately not coupled, so the continuity rule must skip them or it produces false-positive errors. The rule queries the same `should_engage_coupled` helper (or its result map) that #2638 introduces.

**Parallel with**: none. This is a leaf DRC rule.

## Acceptance criteria

- [ ] New `DiffPairRoutingContinuityRule` rule class at `src/kicad_tools/validate/rules/diffpair_routing_continuity.py`
- [ ] Rule registered in `validate/checker.py` and exported from `validate/rules/__init__.py`
- [ ] `ViolationType.DIFFPAIR_ROUTING_CONTINUITY` enum value + alias entry + ROUTING category mapping in `drc/violation.py` (the #2521 three-touch-point pattern)
- [ ] `kct check` shows the new rule under \"Rules checked: N\" count
- [ ] `CHECK_CATEGORIES` and `check_methods` in `cli/check_cmd.py` include the new rule
- [ ] Rule **fires** when a detected, engaged diff pair has `coupled_fraction < threshold` (default `0.7`)
- [ ] Rule **does NOT fire** when:
  - Coupled fraction >= threshold
  - The pair was not engaged (per Phase 2E `coupled_routing=False` or single-ended refusal)
  - Either net has no routed segments (no length = no continuity to measure)
- [ ] Threshold is configurable: a new optional `NetClassRouting.coupled_continuity_threshold: float | None` field (None → use rule default `0.7`)
- [ ] `--errors-only` exit code 2 when this rule fires
- [ ] Unit tests covering: (a) fully-coupled pair (fraction = 1.0) → no fire; (b) half-coupled (fraction = 0.5) → fires at default threshold; (c) un-engaged pair → no fire even at 0% coupling; (d) per-class threshold override; (e) single-net pair (one side has no routes) → no fire; (f) the same alias resolution test pattern as #2521 (look up the rule by string and confirm it returns `DIFFPAIR_ROUTING_CONTINUITY`).
- [ ] Board 03 routed output passes the new rule at default threshold (this confirms the post-#2636 coupled-routing path actually couples)
- [ ] Existing tests pass

## Test fixture plan

1. **Synthetic fully-coupled fixture**: two parallel segments at intra-pair spacing for their full length → `coupled_fraction = 1.0` → no fire.
2. **Synthetic partially-coupled fixture**: segments parallel for half their length, diverging for the other half → `coupled_fraction ≈ 0.5` → fires at default threshold, passes at threshold `0.4`.
3. **Engagement-aware fixture**: a pair whose net class has `coupled_routing=False` but whose routes happen to be uncoupled → rule does NOT fire (engagement gate respected).
4. **Threshold override fixture**: per-class `coupled_continuity_threshold=0.5` and routes at 0.6 coupled → no fire.
5. **Board 03 integration**: route board 03, run `kct check ... --check diffpair_routing_continuity`, expect 0 fires (the USB_D+/D- coupling should be substantial).

## Algorithm reference (for the builder)

To compute `L_coupled`:

1. Group segments by net + layer.
2. For each P segment on layer L, find all N segments on layer L within the coupling window (`intra_pair_clearance + 2 * trace_width + tolerance`).
3. For each candidate, project P onto N (and vice versa) and compute the **overlap length** — the length of the orthogonal projection that has the partner trace within the coupling window AND is parallel within ±15° (so a perpendicular crossing doesn't count as coupling).
4. Sum overlap lengths to get `L_coupled` for the pair.
5. `coupled_fraction = L_coupled / min(L_p, L_n)`.

This is O(N_p × N_n) per pair per layer; acceptable for typical boards (a few hundred segments per net). Use `_segment_segment_clearance` (the same helper `diffpair_clearance_intra.py` uses, at `validate/rules/clearance.py`) for the geometric primitives — do NOT roll a new clearance computation.

## Out of scope

- Modifying the existing `diffpair_clearance_intra` rule (orthogonal — clearance and continuity are complementary).
- Adding length-matching DRC (that's Phase 3 #J).
- Adding impedance DRC (Phase 3 #K).
- Auto-fixing continuity violations (the rule reports; the router fixes — and Phase 2E/2F are how the router gets better at coupling).
- Cross-layer coupling (pairs that broadside-couple across stackup layers are out of scope; require per-pair declaration of broadside intent which doesn't exist yet).
- Coupling-quality metrics beyond `coupled_fraction` (e.g., spacing variance, RMS deviation — could be added later as warnings, but Phase 2G stays a pass/fail rule).

---
*Created by Loom Architect from Epic #2556*

rjwalters:
## Summary

Adds Phase 2G of Epic #2556: a new DRC rule that validates routing continuity for *engaged* differential pairs.

- An "engaged" pair (per #2638, merged as PR #2645) is one whose net class has `coupled_routing == True` AND that passed engagement-layer single-ended refusal.
- The rule fires when the pair's **coupled fraction** -- the share of P's routed length whose nearest point on N is within the coupling window AND parallel within +/-15 degrees -- falls below a configurable threshold.
- Default threshold is **0.7**, empirically calibrated against board 03's USB pair (curator note on #2640). Per-class override via `NetClassRouting.coupled_continuity_threshold: float | None`.

## Design decisions

- **Dependency injection over re-derivation.** The rule consumes a caller-supplied `engaged_pairs: set[tuple[int, int]] | None` constructor parameter rather than importing and calling `should_engage_coupled` itself. This keeps the rule decoupled from #2638's internals and prevents drift between router and DRC (the #2521-class failure mode for capability flags).
- **Graceful degradation.** When `engaged_pairs is None` or empty (e.g., standalone `kct check` without router context), the rule is a conservative no-op. The intended call site is the autorouter consumer, which has already computed engagement via `should_engage_coupled`.
- **Private geometric helper.** `_segment_coupled_overlap` lives in the new rule file, not in `clearance.py` -- per the curator's "do not pre-promote to a public API" guidance. When Phase 3J (length-skew) consumes the same primitive, the promotion happens then.
- **Three-touch-point alias entry** for `ViolationType.DIFFPAIR_ROUTING_CONTINUITY` with the justification comment block the #2521/#2560 lesson requires. Without it, the fuzzy fallback at `from_string()` drops the new type to `UNKNOWN`, breaking every downstream `type == ...` filter.

## File changes

- `src/kicad_tools/validate/rules/diffpair_routing_continuity.py` (new) -- the rule class plus the private geometric helper.
- `src/kicad_tools/drc/violation.py` -- enum member, category map entry (ROUTING), alias-table entry.
- `src/kicad_tools/router/rules.py` -- `NetClassRouting.coupled_continuity_threshold` field + `effective_coupled_continuity_threshold()` accessor.
- `src/kicad_tools/validate/checker.py` -- `check_diffpair_routing_continuity()` wired into `check_all()` and exposed as a discrete method.
- `src/kicad_tools/cli/check_cmd.py` -- `"diffpair_routing_continuity"` in `CHECK_CATEGORIES` and the `check_methods` dispatch dict.
- `src/kicad_tools/validate/rules/__init__.py` -- new rule export.

## Tests

- `tests/test_validate_diffpair_routing_continuity.py` (new, 15 tests) covering:
  - Fully coupled pair -> no fire
  - Half coupled pair -> fires at default threshold
  - Half coupled pair passes with per-class threshold override (0.4)
  - Un-engaged pair -> no fire (engagement gate)
  - Single-ended refusal -> no fire (#2527 lesson)
  - Pair with no routes -> graceful no fire (no division-by-zero)
  - Perpendicular crossing -> not counted as coupled
  - Alias resolution returns `DIFFPAIR_ROUTING_CONTINUITY` (not `UNKNOWN`)
  - `to_dict()` round-trips the type field
  - Board 03 synthetic USB pair passes at default threshold
  - `_segment_coupled_overlap`: parallel-close, non-parallel, window-out, anti-parallel, different-layer
- `tests/test_cli_check_diffpair_routing_continuity.py` (new, 3 tests) for CLI registration.
- All existing diff-pair / DRC / CLI / net-class tests pass (418 tests in the affected surfaces).

## Test plan

- [x] `uv run pytest tests/test_validate_diffpair_routing_continuity.py tests/test_cli_check_diffpair_routing_continuity.py` -- 18/18 passing.
- [x] `uv run pytest tests/test_diffpair*.py tests/test_validate_diffpair*.py tests/test_cli_check_diffpair*.py tests/test_net_class*.py tests/test_drc_*.py tests/test_cli_check.py` -- 418/418 passing.
- [x] `uv run ruff check` on all changed files -- clean.

## Closes

Closes #2640

## Follow-ups (out of scope for this PR)

- Wire `engaged_pairs` into the autorouter consumer using `should_engage_coupled` from #2638. This was deliberately deferred so this PR could land independent of any Phase 2E review cycle.
- Promote `_segment_coupled_overlap` to `clearance.py` when Phase 3J (length-skew rule) consumes it.

### DIFF
--- a/src/kicad_tools/cli/check_cmd.py
+++ b/src/kicad_tools/cli/check_cmd.py
@@ -60,6 +60,7 @@ def _find_pcb_file(directory: Path) -> Path | None:
 CHECK_CATEGORIES = [
     "clearance",
     "diffpair_clearance_intra",
+    "diffpair_routing_continuity",
     "dimensions",
     "edge",
     "netlist",
@@ -272,6 +273,7 @@ def run_selected_checks(
     check_methods = {
         "clearance": checker.check_clearances,
         "diffpair_clearance_intra": checker.check_diffpair_clearance_intra,
+        "diffpair_routing_continuity": checker.check_diffpair_routing_continuity,
         "dimensions": checker.check_dimensions,
         "edge": checker.check_edge_clearances,
         "netlist": checker.check_netlist,
--- a/src/kicad_tools/drc/violation.py
+++ b/src/kicad_tools/drc/violation.py
@@ -48,6 +48,10 @@ def _init_type_category_map() -> None:
             # Differential-pair within-pair clearance (Issue #2560, Epic #2556 Phase 1D).
             # Same category as other clearance subtypes -- fixable by rerouting.
             ViolationType.DIFFPAIR_CLEARANCE_INTRA: ViolationCategory.ROUTING,
+            # Differential-pair routing continuity (Issue #2640, Epic #2556 Phase 2G).
+            # Fixable by adjusting the route so the two halves stay coupled
+            # for a larger share of their length -- pure routing concern.
+            ViolationType.DIFFPAIR_ROUTING_CONTINUITY: ViolationCategory.ROUTING,
             ViolationType.TRACK_WIDTH: ViolationCategory.ROUTING,
             ViolationType.TRACK_ANGLE: ViolationCategory.ROUTING,
             ViolationType.DIMENSION_TRACE_WIDTH: ViolationCategory.ROUTING,
@@ -122,6 +126,14 @@ class ViolationType(Enum):
     # *intra-pair* gap (allowed to be tighter than the manufacturer's inter-pair
     # ``min_clearance_mm``) against the per-class ``intra_pair_clearance``.
     DIFFPAIR_CLEARANCE_INTRA = "diffpair_clearance_intra"
+    # Differential-pair routing continuity (Issue #2640, Epic #2556 Phase 2G).
+    # Fires when an *engaged* differential pair's coupled fraction (the
+    # share of P's routed length whose nearest point on N is within the
+    # coupling window AND parallel within +/-15 degrees) falls below the
+    # per-class ``coupled_continuity_threshold`` (default 0.7).  Distinct
+    # from CLEARANCE because it checks a topology / geometry property
+    # (parallel-coupling continuity), not edge-to-edge spacing.
+    DIFFPAIR_ROUTING_CONTINUITY = "diffpair_routing_continuity"
     COPPER_EDGE_CLEARANCE = "copper_edge_clearance"
     EDGE_CLEARANCE_TRACE = "edge_clearance_trace"
     EDGE_CLEARANCE_PAD = "edge_clearance_pad"
@@ -235,6 +247,16 @@ class ViolationType(Enum):
             # generic CLEARANCE type, masking the new violation type from
             # downstream consumers that filter by exact ``type`` value.
             "diffpair_clearance_intra": cls.DIFFPAIR_CLEARANCE_INTRA,
+            # Differential-pair routing continuity (Issue #2640, Epic #2556
+            # Phase 2G).  MUST be aliased explicitly even though the
+            # rule_id string ``"diffpair_routing_continuity"`` does NOT
+            # contain the substring ``"clearance"`` -- the fuzzy fallback
+            # at the bottom of from_string() would otherwise drop through
+            # to ``UNKNOWN``, which silently corrupts the violation type
+            # field for any downstream filter that compares by exact
+            # type value.  This entry is the only defense; do NOT delete
+            # it as "redundant".
+            "diffpair_routing_continuity": cls.DIFFPAIR_ROUTING_CONTINUITY,
             # edge clearance subtypes
             "edge_clearance_trace": cls.EDGE_CLEARANCE_TRACE,
             "edge_clearance_pad": cls.EDGE_CLEARANCE_PAD,
--- a/src/kicad_tools/router/rules.py
+++ b/src/kicad_tools/router/rules.py
@@ -460,7 +460,7 @@ class NetClassRouting:
     When ``True``, the diff-pair pre-pass / ``route_all_with_diffpairs``
     dispatch invokes :meth:`CoupledPathfinder` for pairs in this class,
     subject to the engagement-layer single-ended refusal in
-    :func:`should_engage_coupled` (#2527 lesson — pin pairs that look
+    :func:`should_engage_coupled` (#2527 lesson -- pin pairs that look
     diff-pair-ish but are single-ended by spec, like USB-C CC1/CC2 and
     SBU1/SBU2, are refused at engagement time even when explicitly
     declared via :attr:`diffpair_partner`).
@@ -478,6 +478,30 @@ class NetClassRouting:
     engagement path").  Future refactors must not collapse the two.
     """
 
+    # Differential pair routing-continuity threshold (Issue #2640, Epic #2556 Phase 2G)
+    coupled_continuity_threshold: float | None = None
+    """Minimum coupled-fraction (0.0..1.0) required by the
+    ``diffpair_routing_continuity`` DRC rule for engaged pairs in this class.
+
+    The rule fires when a routed pair's coupled fraction (the share of P's
+    routed length whose nearest point on N is within the coupling window
+    AND parallel within +/-15 degrees) falls below this threshold.
+
+    ``None`` (the default) means "use the rule's module-level default of
+    0.7" -- empirically calibrated against board 03's USB pair, which
+    couples ~60-80% in practice (curator note on #2640).  Setting
+    ``0.9`` is appropriate for high-speed-differential-interface (HSDI)
+    boards demanding tight coupling; setting ``0.5`` accommodates hobby
+    boards with loose coupling expectations.
+
+    Orthogonal to :attr:`diffpair_partner` and to the (Phase 2E)
+    :attr:`coupled_routing` opt-in flag from #2638 -- this is a DRC-side
+    knob that the autorouter consumer reads via
+    :meth:`effective_coupled_continuity_threshold` and passes into
+    :class:`~kicad_tools.validate.rules.diffpair_routing_continuity.DiffPairRoutingContinuityRule`
+    via the ``threshold_map`` constructor argument.
+    """
+
     def effective_intra_pair_clearance(self) -> float:
         """Return the clearance to apply to within-pair diff-pair edges.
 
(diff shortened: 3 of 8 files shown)


rjwalters:

Final Phase 2 issue of Epic #2556 lands cleanly. Verified the full curator-pinned scope from #2640.

### Three-touch-point DRC integration (the #2521 lesson)
- `ViolationType.DIFFPAIR_ROUTING_CONTINUITY = "diffpair_routing_continuity"` at `violation.py:136` with multi-line justification comment block.
- Category map entry at `violation.py:54` -> `ViolationCategory.ROUTING` (correct -- continuity is a re-route fix).
- Alias-table entry at `violation.py:259` with the load-bearing comment: "MUST be aliased explicitly even though the rule_id string does NOT contain 'clearance'." The explicit alias is in fact redundant with the direct enum-value match (line 223-225), but the test `test_alias_resolution_returns_diffpair_routing_continuity` and the documentation make it defense-in-depth; future refactors of the dispatch order won't break the rule.
- `CHECK_CATEGORIES` at `check_cmd.py:63` and `check_methods` dict at `check_cmd.py:276` -- both alphabetically positioned consistent with `diffpair_clearance_intra`.

### Pattern mirroring vs ViaInPadRule (PR #2637)
Structurally consistent: class-level `rule_id`/`name`/`description`, `check(pcb, design_rules)` method, returns `DRCResults`, `_make_violation` helper -- same scaffold as `via_in_pad.py:131` and `diffpair_clearance_intra.py`.

### Dependency injection over re-derivation
Confirmed: `DiffPairRoutingContinuityRule.__init__(engaged_pairs=None, threshold_map=None, ...)`. No import of `should_engage_coupled` from #2638. The producer side at `checker.py:192` constructs the rule with the default no-arg constructor (no engaged_pairs), which makes the standalone `kct check` invocation correctly a no-op. Tuple key normalization (`(min, max)`) at lines 477-478 / 482-484 means callers don't have to pre-sort.

### Graceful degradation for standalone `kct check`
- `engaged_pairs=None` -> empty internal set -> `rules_checked == 0` (not "passed" or "violations=0" -- the right honest signal).
- `engaged_pairs={}` -> same.
- Per-half zero-length guard at `diffpair_routing_continuity.py:543` prevents division by zero.

### Threshold (curator's recommendation)
- Default `0.7` at `DEFAULT_COUPLED_CONTINUITY_THRESHOLD` -- matches the curator note for board 03's USB pair couple of ~60-80%.
- `NetClassRouting.coupled_continuity_threshold: float | None` with `effective_coupled_continuity_threshold(default=0.7)` accessor at `router/rules.py:516`. Backward-compat (None -> default), no API break.

### Helper kept private
`_segment_coupled_overlap` lives in `diffpair_routing_continuity.py:351`, not in `clearance.py`. Correctly reuses `_segment_segment_clearance` from `clearance.py:216` and `CopperElement` -- no geometric primitive duplication. Module docstring explicitly defers promotion to "when Phase 3J (length-skew) consumes the same primitive."

### Rebase correctness on rules.py
Verified post-merge of #2645 (`coupled_routing` from #2638) and this PR's `coupled_continuity_threshold`. Both fields present side-by-side at `router/rules.py:451` and `:482` with full docstrings; both accessors (`effective_intra_pair_clearance` and `effective_coupled_continuity_threshold`) intact at `:505` and `:516`. No merge corruption.

### Tests
- 18 new tests (15 rule + 3 CLI), 18/18 pass locally in 14.87s.
- Adjacent regression suite (test_diffpair*, test_validate_diffpair*, test_net_class*, test_drc_violation_type_validate, test_cli_check): 403 passed, 9 skipped -- no regressions in the affected surface.
- Coverage of both engaged-path (violations possible) and un-engaged-path (no-op), perpendicular-crossing geometric edge case, division-by-zero guard, alias resolution AND `to_dict()` round-trip (the #2521 gotcha guard pair).

### CI
All 6 jobs green: Lint, Type Check, kicad-cli Smoke, C++ Build, DRC Check, Test (6m4s). `mergeStateStatus: CLEAN`.

### One follow-up worth filing (out of scope here)
The autorouter consumer still doesn't thread `engaged_pairs` into the rule, so the rule is currently a no-op everywhere. The PR description acknowledges this as a deferred follow-up. That follow-up will close the regression-guard loop the issue described; recommend filing it under the Epic #2556 Phase 2 wrap-up.

Approving and applying `loom:pr`.

Issue #2664, opened 2026-05-11
sch validate: wrong_project (instances) entries silently pass but break netlist export


rjwalters:
## Problem Statement

\`kct sch validate\` lists \`project_instances\` in its \`checks_run\` list, but does not flag \`wrong_project\` entries in \`(instances)\` blocks even when those entries are systemic and break netlist export.

In chorus-test-revA (\`/Users/rwalters/GitHub/chorus/hardware/chorus-test-revA/kicad/\`), \`validate\` passed with 3 ERC errors / 32 warnings (none mentioning instances) — while \`repair-instances --dry-run\` found **90 wrong_project repairs needed** across all 6 sub-sheets. The wrong-project entries caused kicad-cli's netlist exporter to drop pins on local-only nets (Y2.3↔U7.2, NT2.2↔R15.1, etc.), making the TCXO clock chain appear "floating" downstream.

## Reproduction

\`\`\`bash
# Before:
kct sch validate chorus-test-revA.kicad_sch  # passes with normal ERC issues
kct sch repair-instances chorus-test-revA.kicad_sch --dry-run  # reports 90 fixes
kicad-cli sch export netlist ...  # netlist missing 5 pin nodes

# After:
kct sch repair-instances chorus-test-revA.kicad_sch
kicad-cli sch export netlist ...  # 5 pin nodes restored, Net-(Y2-OUT) appears
\`\`\`

## Expected Behavior

\`sch validate\`'s \`project_instances\` check should report \`wrong_project\` entries as **errors** (or at minimum, prominent warnings with the repair command in the message). Even one wrong_project entry can silently drop pins from netlist export; 90 is a structural issue that must not pass silently.

## Acceptance Criteria

- \`kct sch validate\` returns non-zero / fails when any \`wrong_project\` instance is present
- The issue message references \`repair-instances\` as the fix path
- Regression test: a sheet with one symbol whose \`(instances)\` block names only the sub-sheet project (not the root project) should fail validate

## Why this matters

This is the single highest-impact schematic bug class — invisible in KiCad GUI, undetected by ERC, silently corrupts netlist export. Every routing experiment, audit, and benchmark using such a netlist is operating on bad data. The May 10 audit (\#2614 corrector) and the original \#2613 audit both misdiagnosed downstream symptoms because validate didn't flag the root cause.

## Related

- Memory in chorus repo: \`project_chorus_test_routing_2026_05_10b.md\` documented this class of bug after manual discovery
[...]

PR #2669, opened 2026-05-11
fix(sch unconnected): cross-check netlist export + walk hierarchy

### LINKED ISSUE #2665: sch unconnected: false-negative when symbols dropped from netlist export
## Problem Statement

\`kct sch unconnected\` walks the schematic's wire/symbol graph to determine pin connectivity but does not cross-check against the actual kicad-cli netlist export. When symbols are dropped from the netlist export (e.g., due to wrong_project entries in \`(instances)\` — see \#2664), pin nodes that are missing from the netlist still show as "connected" by \`unconnected\`.

In chorus-test-revA, with 5 pin nodes (Y2.3, U7.2, plus 3 more) silently missing from \`chorus-test-revA-netlist.kicad_net\`, \`kct sch unconnected\` reported "All pins connected!" because both endpoints existed in the schematic and were wire-connected within the sheet. The downstream router still saw them as missing.

## Reproduction

\`\`\`bash
# In a project with wrong_project entries:
kct sch unconnected chorus-test-revA.kicad_sch  # "All pins connected!"
kicad-cli sch export netlist ...                # silently drops pins
grep '"Y2"' netlist.kicad_net | grep 'pin "3"'  # no match — pin missing!
\`\`\`

## Expected Behavior

\`sch unconnected\` should optionally cross-reference the exported netlist. Two reasonable modes:
1. **Default**: also run \`kicad-cli sch export netlist\` (or use an in-memory equivalent) and report any pin present in the schematic but absent from the exported netlist.
2. **Strict flag**: \`--check-netlist-export\` that compares schematic-pin set against netlist-pin set and reports the diff.

## Acceptance Criteria

- New check: "pin X.Y exists in schematic but is missing from netlist export"
- The fix message recommends \`kct sch repair-instances\`
- Regression test in tests/ covering this case

## Why this matters

The whole point of \`sch unconnected\` is to catch missing wiring before layout. If it returns false-negatives on symbols that won't appear in the manufacturing netlist, downstream tooling (routers, BOM generators, position-file exporters) silently consumes incomplete data.

## Related

- \#2664 (validate should flag wrong_project)
- \#2614 (chorus-test defect tracking, where this false-negative caused the original misdiagnosis)

rjwalters:
## Summary

`kct sch unconnected` previously walked only the schematic's wire-graph and
never opened the exported netlist. When `kicad-cli` silently drops a symbol
from the netlist (most commonly because its `(instances)` block contains
only `wrong_project` entries -- see #2664), the symbol's pins were still
reported as "connected" if a wire happened to touch their coordinates --
exactly the class of defect the command exists to catch.

Two structural fixes:

1. **Netlist cross-check** (default-on, opt-out via `--no-check-netlist-export`).
   After the wire-graph pass, export the netlist via `kicad-cli` (no
   Python fallback -- we want what every downstream consumer actually
   sees) and emit a new `missing_from_netlist` finding for every
   `(reference, pin)` present in the schematic but absent from the
   exported nets. The remediation message points at `kct sch repair-instances`.
2. **Hierarchy walk**. Reuses `_collect_schematic_files` from
   `sch_set_footprint` to walk every sub-sheet referenced by
   `(property "Sheetfile" ...)` and analyses each one's wire-graph
   independently. Previously only the root sheet was analysed.

## Changes

- `src/kicad_tools/cli/sch_find_unconnected.py`: new `MissingFromNetlist`
  dataclass, `_analyze_single_sheet` helper (per-sheet wire-graph), and
  `_cross_check_netlist` helper (kicad-cli export + diff). `analyze_schematic`
  now walks the hierarchy and runs the netlist cross-check.
- `src/kicad_tools/cli/parser.py`: new `--no-check-netlist-export` flag.
- `src/kicad_tools/cli/commands/schematic.py`: forwards the new flag.
- `tests/test_cli_sch.py`: 6 new regression tests covering JSON schema,
  the opt-out path, the kicad-cli-unavailable graceful-skip path, the
  core dropped-pin detection, the false-positive guard, and the
  hierarchy walk.

| New `missing_from_netlist` finding category | done | `MissingFromNetlist` dataclass; `test_unconnected_detects_dropped_pin_via_netlist` asserts `("R1", "1")` and `("R1", "2")` appear |
| Finding message references `kct sch repair-instances` | done | `remediation` field in JSON output; test asserts `"repair-instances" in finding["remediation"]` |
| Power and DNP symbols excluded unless flags set | done | Existing `include_power`/`include_dnp` logic is reused -- only filtered symbols are recorded into the schematic-pin set used for the cross-check |
| `--format json` exposes new findings + `missing_from_netlist_count` | done | `output_json` adds `missing_from_netlist` list and `missing_from_netlist_count` in summary; `test_unconnected_json_schema_includes_new_keys` asserts both |
| kicad-cli unavailable -> skip with stderr warning (not silent pass) | done | `_cross_check_netlist` catches `FileNotFoundError`/`RuntimeError` and prints `"Warning: ... skipped"` to stderr; `test_unconnected_netlist_export_unavailable_skips_gracefully` asserts both the JSON output and the warning |
| Hierarchy walk: sub-sheets analysed | done | `_collect_schematic_files` walk in `analyze_schematic`; `test_unconnected_hierarchy_walk_includes_subsheet_pins` builds a parent + child sheet and asserts the child's pins appear |
| Regression test for wrong_project dropout | done | `test_unconnected_detects_dropped_pin_via_netlist` simulates kicad-cli dropping every symbol (the wrong_project failure mode) and asserts `missing_from_netlist_count >= 2` |
| Test for all-good schematic produces 0 findings | done | `test_unconnected_no_false_positives_when_netlist_matches` |

## Test Plan

- [x] `uv run pytest tests/test_cli_sch.py::TestSchFindUnconnected` -- 11/11 pass
- [x] `uv run pytest tests/test_cli_sch.py::TestConnectionsSubGridTolerance` -- 6/6 pass (no regression)
- [x] `uv run pytest tests/test_cli_sch.py` -- 134/135 pass; the 1 failure (`TestSchSummary::test_existing_hierarchical_fixture_aggregation`) is **pre-existing on main** (verified by `git stash` + re-run)
- [x] `uv run ruff check` on changed files -- clean
- [x] `uv run ruff format --check` on changed files -- clean

## Relationship to #2664

#2664 makes `sch validate` refuse `wrong_project` entries (preventive).
This PR makes `sch unconnected` catch any dropped pin regardless of
cause (defensive). They land in different files and reinforce each
other -- neither subsumes the other.

Closes #2665

### DIFF
--- a/src/kicad_tools/cli/commands/schematic.py
+++ b/src/kicad_tools/cli/commands/schematic.py
@@ -162,6 +162,8 @@ def run_sch_command(args) -> int:
             sub_argv.append("--include-power")
         if args.include_dnp:
             sub_argv.append("--include-dnp")
+        if not getattr(args, "check_netlist_export", True):
+            sub_argv.append("--no-check-netlist-export")
         return unconnected_main(sub_argv) or 0
 
     elif args.sch_command == "set-footprint":
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -529,6 +529,13 @@ def _add_sch_parser(subparsers) -> None:
         "--include-power", action="store_true", help="Include power symbols"
     )
     sch_unconnected.add_argument("--include-dnp", action="store_true", help="Include DNP symbols")
+    sch_unconnected.add_argument(
+        "--no-check-netlist-export",
+        dest="check_netlist_export",
+        action="store_false",
+        default=True,
+        help="Skip the netlist-export cross-check (default: enabled)",
+    )
 
     # sch replace
     sch_replace = sch_subparsers.add_parser("replace", help="Replace a symbol's library ID")
--- a/src/kicad_tools/cli/sch_find_unconnected.py
+++ b/src/kicad_tools/cli/sch_find_unconnected.py
@@ -2,6 +2,21 @@
 """
 Find unconnected pins and potential issues in a KiCad schematic.
 
+This command performs two complementary checks:
+
+1. **Wire-graph analysis** -- walks the schematic's wire-graph and reports
+   pins that are not connected to any wire, label, or junction.  For
+   hierarchical schematics, the analysis is repeated for every sub-sheet
+   discovered via the ``Sheetfile`` property.
+
+2. **Netlist cross-check** -- exports the netlist via ``kicad-cli`` (no
+   Python fallback) and reports any pin that exists in the schematic but
+   is missing from the netlist.  This catches the false-negative observed
+   when a symbol's ``(instances)`` block contains only ``wrong_project``
+   entries: ``kicad-cli`` silently drops the symbol from the netlist but
+   the wire-graph analysis still sees it.  Use
+   ``--no-check-netlist-export`` to skip this cross-check.
+
 Usage:
     python3 sch-find-unconnected.py <schematic.kicad_sch> [options]
 
@@ -10,6 +25,7 @@ Options:
     --filter <pattern>         Filter by symbol reference (e.g., "U*")
     --include-power            Include power symbols in analysis
     --include-dnp              Include DNP (do not populate) symbols
+    --no-check-netlist-export  Skip the netlist-export cross-check
 
 Examples:
     # Find all unconnected pins
@@ -27,6 +43,7 @@ import fnmatch
 import json
 import sys
 from dataclasses import dataclass
+from pathlib import Path
 
 from kicad_tools.cli.sch_connectivity import (
     Coord,
@@ -34,6 +51,7 @@ from kicad_tools.cli.sch_connectivity import (
     is_pin_connected,
     to_coord,
 )
+from kicad_tools.cli.sch_set_footprint import _collect_schematic_files
 from kicad_tools.schema import Schematic
 
 
@@ -48,6 +66,7 @@ class UnconnectedPin:
     symbol_value: str
     lib_id: str
     position: tuple[float, float]
+    sheet: str = ""  # Sub-sheet file name (empty for root)
 
 
 @dataclass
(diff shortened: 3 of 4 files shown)


rjwalters:
## Loom Judge Review: Approved

Excellent fix for the #2665 regression. The two structural changes (netlist cross-check + hierarchy walk) close the false-negative pathway cleanly, and the test coverage is genuinely thorough.

### Verification

**Bug fix (`_cross_check_netlist`)** — verified at `src/kicad_tools/cli/sch_find_unconnected.py:251-301`:
- Calls `export_netlist(root_path, fallback=False)` — the inline comment correctly explains the design rationale (using `fallback=True` would hide the bug because the Python fallback walks the hierarchy correctly; we explicitly want to see what every downstream consumer sees).
- Diffs against the schematic-side `(reference, pin_number)` set built from `symbol_data` after power/DNP filtering.
- Catches `FileNotFoundError` / `RuntimeError` separately and returns a warning string; caller prints to stderr and still surfaces wire-graph findings. Not a silent pass.

**Hierarchy walk** — verified at `src/kicad_tools/cli/sch_find_unconnected.py:330-360`:
- Reuses `_collect_schematic_files` from `sch_set_footprint.py:48` (good code reuse — no duplication of the Sheetfile-regex walk).
- Per-sheet `_analyze_single_sheet` runs wire-graph analysis on every sub-sheet independently.
- Netlist cross-check runs **once** at root (`if check_netlist_export and root_path is not None`) — correct because the exported netlist already covers the entire hierarchy.
- `seen_pins` set de-dupes hierarchical instances of the same `(ref, pin)` to avoid double-reporting.
- Every dataclass (`UnconnectedPin`, `ConnectionIssue`, `MissingFromNetlist`) carries a `sheet` field for triage; JSON output propagates it (verified `test_unconnected_hierarchy_walk_includes_subsheet_pins` asserts `"child.kicad_sch" in sheets`).

**Power/DNP filtering** — verified at `sch_find_unconnected.py:125-131`: the power/DNP gate happens before `symbol_data.append(...)`, so filtered symbols never enter `schematic_pin_records`, and therefore never enter `missing_from_netlist`. No false positives from power/DNP symbols.

**CLI flag** — verified at `parser.py:532-538` and `commands/schematic.py:165-166`: `--no-check-netlist-export` uses `action="store_false"` with `default=True` — so the cross-check runs by default, opt-out only. The flag is correctly forwarded by `run_sch_command` via `getattr(args, "check_netlist_export", True)`.

**Tests** — all 6 new tests in `TestSchFindUnconnected` pass locally and target the right axes:
- `test_unconnected_json_schema_includes_new_keys` — schema contract
- `test_unconnected_no_check_netlist_export_skip` — opt-out actually disables `export_netlist` call (monkeypatched to raise `AssertionError` if invoked)
- `test_unconnected_netlist_export_unavailable_skips_gracefully` — kicad-cli unavailable -> stderr warning + wire-graph findings still surface
- `test_unconnected_detects_dropped_pin_via_netlist` — **core regression**: wire-graph reports 0 unconnected, but cross-check surfaces `("R1", "1")` and `("R1", "2")` as missing
- `test_unconnected_no_false_positives_when_netlist_matches` — false-positive guard
- `test_unconnected_hierarchy_walk_includes_subsheet_pins` — root with sheet-ref + child with isolated R99 -> R99.1 and R99.2 surface

**Test counts** — verified locally:
- `TestSchFindUnconnected` — 11/11 pass
- `TestConnectionsSubGridTolerance` — 6/6 pass (no regression)
- Full `tests/test_cli_sch.py` — 139 pass, 1 failure (`TestSchSummary::test_existing_hierarchical_fixture_aggregation`), which **reproduces on `origin/main`** — pre-existing, unrelated.

### CI

All 6 GitHub checks pass: C++ Build Check, Lint & Format, Routed PCB DRC Check, Test (5m48s), Type Check, kicad-cli Round-trip Smoke. `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

### Relationship to #2664

PR #2669 does not depend on #2664 (still `loom:building`, no PR yet). Files are disjoint (`sch_find_unconnected.py` vs `sch_validate.py`). The defenses are complementary: #2664 will refuse to land wrong_project entries (preventive at validate time); #2669 catches any dropped pin regardless of cause (defensive at unconnected time). Either can land first.

- The `output_json` remediation hint string is duplicated per-finding (memory cost is trivial for realistic finding counts, and inlining keeps the JSON consumer self-contained, so this is fine).
- `_cross_check_netlist` catches a bare `Exception` (line 282) as a defensive last resort with `# pragma: no cover` — appropriate defensive programming for a tool boundary.

Approving.

Issue #2682, opened 2026-05-11
board 05: re-tune placement for 35-net design density (post block-refactor)


rjwalters:
## Background

Filed from diagnostic work for #2673. Board 05's net count grew from 26 (May-6 audit, 38% routed) to 35 (May-11) after the block-refactor wave landed (#2576, #2580, #2581, #2582, #2583). The committed routed PCB still represents the May-8 routing (22/26 = 84.6% with 15 DRC errors), but re-routing the current design produces dramatically worse results even after the layer-escalation fix in #2673.

Even with the per-net A* regression (tracked in #2681) resolved, the placement geometry was tuned for the 26-net design — the new 9 nets (PWM_AH/AL/BH/BL/CH/CL, GATE_DRV_AH/BH/CH from the gate-drive-resistor-array refactor in #2576/#2581) plus 9 new components (R20-R22, R30-R32, C30-C32) need to fit in the existing U3↔Q1-Q6 fanout corridor that #2542 already flagged as borderline.

## Symptoms in the current PCB

When `kct route` is run on the freshly generated board-05 PCB, the failed-net list shows the gate-drive corridor saturated:
```
- Net 6 "PHASE_B": blocked_path
- Net 7 "PHASE_C": blocked_path (blocked by U3)
- Net 5 "PHASE_A": blocked_path
- Net 17 "ISENSE_B-": blocked_path
- Net 15 "ISENSE_A-": blocked_path (blocked by U3)
- Net 14 "ISENSE_A+": blocked_path (blocked by U3)
- Net 16 "ISENSE_B+": blocked_path (blocked by U3)
- Net 19 "ISENSE_C-": blocked_path (blocked by U3)
```

Six of the eight blocked nets are blocked by the DRV8301 gate-driver IC (U3), confirming the corridor is overcrowded.

## Approach options

**Option A — placement re-tune** (preferred for keeping design fidelity):
- Move R20-R22 (gate-drive resistors) and the gate-driver Hall filter R30-R32/C30-C32 out of the U3↔Q1-Q6 corridor
- Use `kct optimize-placement` with constraints anchoring U3, U10, and Q1-Q6 to current locations
- Target restoring at least the 22/26 (84.6%) completion seen at #2583's committed PCB

**Option B — revert the gate-drive resistor array on board 05** (faster, less fidelity):
[...]

Issue #2681, opened 2026-05-11
router: bisect a9790ad0..15014a72 to find board-05 per-net A* regression (9/35 → 0/35 on identical design)


rjwalters:
## Background

Filed from diagnostic work for #2673 (board 05 routing collapse). The completion-floor guard from #2673 makes the layer-escalation loop traverse all configs at low completion, but the underlying per-net routing regression is NOT fixed.

## Concrete evidence

Identical design (`a9790ad0` board 05 sources), identical flags (`--seed 42 --differential-pairs --auto-layers --manufacturer jlcpcb --timeout 240`), same backend (pure Python — C++ extension was not built in either case), but the per-net routing reaches different states:

| Router HEAD | Strategy | 2L initial pass | 2L final routed | Best result |
|---|---|---|---|---|
| `a9790ad0` (2026-05-08 17:04) | negotiated | 19/35 nets, overflow=12 | 9/35 (26%) | 2L 26% (escalated 3 attempts) |
| `15014a72` (2026-05-11) | negotiated | 10/35 nets, overflow=0 | 0/35 (0%) | 2L 0% (escalates all attempts after #2673 fix) |

The 2L initial-pass routed half as many nets (19 → 10), and per-net A* times ballooned:
- `a9790ad0`: ISENSE_B- took 7.5s, ISENSE_A- 57.6s — then timeout at net 10/35
- `15014a72`: ISENSE_B- took 136.7s (!), per-net A* monopolizes the budget

This is a per-net routing regression in the negotiated A* pathfinder, not an escalation-logic issue.

## Suspect commits (per #2673 curator analysis)

Four router-touching commits land between `a9790ad0` and `15014a72` (in date order):

| commit | date | scope | suspect? |
|---|---|---|---|
| `472ef4d7` | 2026-05-10 | `feat(router): detect rip-up cohort stagnation in negotiated outer loop (closes #2597) (#2601)` | **HIGH** — prime suspect per curator |
| `eec435ae` | 2026-05-10 | `feat(router): pipeline-level connectivity invariant for optimize/nudge (closes #2596) (#2603)` | medium |
| `42fd9e10` | 2026-05-10 | `feat(router): detect placement-feedback stagnation + add outer-timeout guard (closes #2606) (#2607)` | low (only fires with --placement-feedback) |
| `2c037f6a` | 2026-05-10 | `fix(router): monte-carlo + auto-layers escalates past 2L (#2641)` | low (negotiated path unaffected) |

[...]

Issue #2673, opened 2026-05-11
board 05 routing collapse: 38% → 6% completion after net-count growth + design changes


rjwalters:
## Symptom

Board 05 (BLDC motor controller) routing completion fell from **38% (10/26 nets) on 2026-05-06** to **6% (2/35 nets) on 2026-05-11** under identical default flags (`kct route --seed 42 --differential-pairs --auto-layers --manufacturer jlcpcb --timeout 240`).

DRC also exploded: 0 errors (May 6) → 77 errors (May 11). Net count grew 26 → 35 (denser design).

Allowlist floor in `.github/routed-drc-tolerance.yml` (15 errors) is now exceeded 5×.

## Root-cause hypotheses

The audit on 2026-05-11 did not pinpoint the regression source, but several candidates:

1. **Design changes in `boards/05-bldc-motor-controller/design.py`** between May 6 and May 11 — net count grew 26 → 35 (denser footprints, more sub-circuits)? Compare via `git log` to identify the size change.

2. **`auto-layers` failed to engage** — the audit log shows "Best result 6% on 2 layers" with no 4-layer escalation attempt visible. The escalation gate may require ≥95% completion before triggering... but 6% should trigger more aggressively, not less. Verify behavior of `route_cmd.py` escalation thresholds for low-completion edge cases.

3. **Placement quality degraded** — with 35 nets, the existing placement may not have viable channels. `kct optimize-placement` may help.

4. **Power pours interaction** — the May 6 audit confirmed all 4 boards had pours. Verify board 05 still gets pours via the v19 auto-pour path.

## Investigation steps

1. `git log --since=2026-05-06 --oneline boards/05-bldc-motor-controller/` to identify design changes.
2. Re-run May 6's exact command set (find from #2394 audit comments) on May 6's commit to confirm baseline reproduces.
3. Bisect between that commit and current `main` to identify the regression source.
4. Inspect the auto-layers escalation logic: does it short-circuit when completion is very low?

## Acceptance criteria

- [ ] Root cause identified and attributed: either (a) design change in `design.py` that grew net density beyond router capability, (b) router escalation regression, (c) placement quality drop, or (d) some combination.
- [ ] If (a): file separate issue to re-tune the design and update #2542 / #2515 to reflect the new baseline.
- [ ] If (b): fix the router regression and re-route to recover ≥38% completion floor.
- [ ] If (c)/(d): document the placement gap and either re-place (auto or manual) or file a placement issue.
[...]

Issue #2672, opened 2026-05-11
router: honor target_single_impedance / target_diff_impedance during width selection


rjwalters:
## Context

Phase 3K (PR #2655, closes #2650) registered the previously-dormant `ImpedanceRule` and added `NetClassRouting.target_single_impedance`, `target_diff_impedance`, `impedance_tolerance_percent` fields. The DRC rule now fires when routed trace widths produce impedance outside the tolerance band.

The audit on 2026-05-11 (against Epic #2556 Phase 4L's board 06 + boards 03/04/05) found **59 new ImpedanceRule violations** across boards 03/04/05 because the **router is still using the default 0.2mm trace width despite the per-net `target_single_impedance=50` set by net-class metadata**.

## Root cause

`diffpair_impedance.resolve_impedance_for_net_classes()` (added by PR #2655) is the right hook — it rewrites `NetClassRouting.trace_width` and `intra_pair_clearance` to honor the target impedance. But it appears to either:
- Not be invoked from all routing entry points (dormant-signal pattern, similar to #2587/#2652/#2657)
- Not produce non-default widths in practice
- Be invoked but the consumers (pathfinder/cpp_backend/escape) read `trace_width` before the resolver runs

## Investigation needed

1. Verify `resolve_impedance_for_net_classes` is called from `Autorouter` setup. If not, wire it (mirror the Phase 3H-cont pattern from #2657).
2. Confirm the resolver actually changes `trace_width` for nets with `target_single_impedance` set. The audit found JLCPCB-default stackup → 50 ohm needs ~0.336mm on B.Cu, not 0.2mm.
3. Cross-backend: verify pathfinder.py, cpp_backend.py, escape.py all consume the post-resolution width.

## Acceptance criteria

- [ ] Re-routing board 04 (8/9 today, 5 impedance errors) produces 0 ImpedanceRule violations on traces with `target_single_impedance=50`.
- [ ] Re-routing board 03 reduces the 53 impedance violations to ≤ 5 (residuals expected from U1 blocked-pad area).
- [ ] Drift-prevention test: synthetic board with one net classed `target_single_impedance=50` against JLCPCB 4-layer stackup → assert routed `trace_width` is within 5% of the impedance-target width.

## Acceptance threshold

Initially aim for ImpedanceRule errors to drop to ≤ 10 total across boards 03/04/05 (was 59 in May-11 audit). Phase 4N (#2660) can then tighten further.

## Related

[...]

PR #2683, opened 2026-05-11
feat(route): add completion-floor guard to layer-escalation early stops

rjwalters:
## Summary

Adds a completion-floor guard (50%) to the two early-termination heuristics in `route_with_layer_escalation`. Both the zero-overflow short-circuit (#2412) and the stagnation short-circuit (#2412) were calibrated for boards that route a substantial fraction of their nets on the first attempt — at very low completion, the failure mode is usually "per-net A* timed out before reaching the rip-up loop", and adding more layers actually does help.

## Diagnostic data (from issue #2673)

Identical design (board 05 at commit `a9790ad0`, last `design.py` change), same flags, same pure-Python backend:

| Router HEAD | 2L initial pass | 2L final routed | Attempts tried | Best result |
|---|---|---|---|---|
| `a9790ad0` (2026-05-08) | 19/35 nets, overflow=12 | 9/35 (26%) | 3 (2L, 4L sig-gnd-pwr-sig, 4L all-sig) | 2L 26% |
| `15014a72` (2026-05-11, before this PR) | 10/35 nets, overflow=0 | 0/35 (0%) | **1** (zero-overflow short-circuit fires) | 2L 0% |
| `15014a72` (after this PR) | 10/35 nets, overflow=0 | 0/35 (0%) | **4** (2L + all three 4L/6L configs) | 2L 0% |

The 0/35 result is NOT addressed by this PR — that's a per-net A* regression filed as #2681. This PR only fixes the layer-escalation logic so that low-completion runs traverse the full layer-config search before reporting failure.

Track A residual (placement re-tune for the 35-net design) is filed as #2682.

## Changes

- `src/kicad_tools/cli/route_cmd.py`:
  - Compute `below_completion_floor = best_result.completion < 0.5` before the two early-stop checks
  - Skip both heuristics when `below_completion_floor`, emit an informative `"Note: ..."` message explaining the bypass
  - Above the floor, behavior is unchanged
- `tests/test_layer_escalation.py`:
  - `test_completion_floor_bypasses_zero_overflow_below_floor` — at 0% completion, escalation continues past attempt 1
  - `test_completion_floor_bypasses_stagnation_below_floor` — at 10% completion, escalation continues past stagnation
  - `test_completion_floor_does_not_affect_high_completion` — at 80% completion, zero-overflow heuristic still fires (regression guard)

The parent issue #2673 asked for diagnostic data identifying Track A vs Track B; this PR is the **Track B** outcome (with Track A residual filed as #2682).

| Criterion (from #2673) | Status | Verification |
|---|---|---|
| Diagnostic completed with concrete data; Track A or Track B identified | done | Numbers above + filed sub-issues #2681 (router regression bisection) and #2682 (placement re-tune). The regression is real (26% -> 0% on identical design), but the root cause is in per-net A* (separate from escalation logic). |
| If Track B: regression commit identified + reverted/fixed + tolerance restored to <=15 floor | partial | Per-net A* regression suspect identified (#2601's `detect_ripup_stagnation()`); bisection deferred to #2681 to avoid conflicts with the just-landed #2672/#2680. Tolerance floor remains 15 — the committed routed PCB still meets it; the regression only shows on re-route. |
| PR description includes the diagnostic numbers | done | Table above |

## Test Plan

- `uv run pytest tests/test_layer_escalation.py tests/test_route_cmd_*.py` -> **182 passed**
- Live diagnostic: `uv run kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb --seed 42 --differential-pairs --auto-layers --manufacturer jlcpcb --timeout 120` on the post-fix worktree -> escalation traverses 2L, 4L SIG-GND-PWR-SIG, 4L ALL-SIG, 6L (vs only 2L pre-fix); diagnostic "Note:" messages confirm the floor guard fired three times during the run.

Refs #2673

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -1716,6 +1716,24 @@ def route_with_layer_escalation(
                     f"({best_result.completion:.0%})"
                 )
 
+        # Issue #2673: Completion-floor guard.  Both early-termination heuristics
+        # below (zero-overflow at #2412, stagnation at #2412) are calibrated for
+        # the case where the router already routed a substantial fraction of the
+        # board on the prior attempt.  When best-so-far completion is very low
+        # (e.g., < 50%), the failure mode is more likely "router got stuck on a
+        # handful of nets and timed out" than "the design is genuinely unrouteable
+        # with more layers", so we should keep trying additional layer
+        # configurations rather than short-circuit.  Board 05 on 2026-05-11
+        # exhibits exactly this regression: 2L=0/35, 4L sig-gnd-pwr-sig=0/35,
+        # stagnation check fires and 4L all-sig is never attempted.  The same
+        # board at commit a9790ad0 produced 2L=9/35 (26%) and tried all three
+        # 4L variants before stopping.  Floor of 50% is conservative — any
+        # board with >=50% completion has enough signal for the heuristics
+        # to be trustworthy; below that, escalation should run to completion.
+        best_completion_so_far = best_result.completion if best_result is not None else 0.0
+        completion_floor_for_early_stop = 0.5
+        below_completion_floor = best_completion_so_far < completion_floor_for_early_stop
+
         # Issue #2412: Early termination — zero overflow means failures
         # are placement/topology issues, not congestion.  Adding layers
         # cannot help when there is no congestion to relieve.
@@ -1734,6 +1752,7 @@ def route_with_layer_escalation(
             overflow == 0
             and nets_routed < nets_to_route
             and args.strategy not in strategies_without_overflow_signal
+            and not below_completion_floor
         ):
             if not quiet:
                 flush_print(
@@ -1746,6 +1765,19 @@ def route_with_layer_escalation(
                 )
                 flush_print("  Status: INSUFFICIENT - early stop (zero overflow)")
             break
+        elif (
+            overflow == 0
+            and nets_routed < nets_to_route
+            and args.strategy not in strategies_without_overflow_signal
+            and below_completion_floor
+            and not quiet
+        ):
+            flush_print(
+                "  Note: overflow=0 but best completion "
+                f"{best_completion_so_far * 100:.0f}% < "
+                f"{completion_floor_for_early_stop * 100:.0f}% floor — "
+                "continuing escalation (issue #2673)"
+            )
 
         # Issue #2412: Early termination — stagnation detection.  If adding
         # layers did not improve nets_routed or reduce overflow, further
@@ -1754,6 +1786,7 @@ def route_with_layer_escalation(
             prev_nets_routed is not None
             and nets_routed <= prev_nets_routed
             and overflow >= prev_overflow
+            and not below_completion_floor
         ):
             if not quiet:
                 flush_print("  Escalation stopped: no improvement after adding layers")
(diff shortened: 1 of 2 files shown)


rjwalters:
## Code Review — Approved

Reviewed PR #2683 (completion-floor guard for layer-escalation early stops) against issue #2673's two-track diagnostic plan. Excellent execution of a tricky diagnostic-with-scope-adjustment.

### Diagnostic verification

The PR body contains the exact data the curator requested:

| Router HEAD | 2L final | Attempts |
|---|---|---|
| `a9790ad0` (May 8) | 9/35 (26%) | 3 |
| `15014a72` (May 11, pre-fix) | 0/35 (0%) | 1 (zero-overflow short-circuit fires immediately) |
| `15014a72` (post-fix) | 0/35 (0%) | 4 (all configs tried) |

The 26% → 0% delta confirms a real per-net A* regression. Builder appropriately scoped this PR to the *escalation logic* fix and filed the deeper bisection as follow-up #2681 (verified open, `loom:curated`). Placement re-tune filed as #2682 (also verified open, `loom:curated`).

### Scope decision: correct call

The early-termination heuristics at `route_cmd.py:1733` (zero-overflow) and `:1753` (stagnation) were calibrated for boards with substantial completion. At very low completion the failure mode is "per-net A* timed out before reaching the rip-up loop", and `overflow=0` is therefore meaningless as a congestion signal — adding more layers genuinely does help give A* alternate channels. Guarding these heuristics with a completion floor is independently correct, even after #2681 lands.

### Implementation review (`src/kicad_tools/cli/route_cmd.py:1719-1801`)

- `best_completion_so_far = best_result.completion if best_result is not None else 0.0` — correctly defensive on first iteration
- `completion_floor_for_early_stop = 0.5` — hard-coded, but the inline rationale is clear ("any board with >=50% completion has enough signal for the heuristics to be trustworthy"). Acceptable; not worth blocking on configurability for a fresh diagnostic.
- Both heuristics get the `and not below_completion_floor` clause AND a paired `elif` that emits a `"Note: ..."` message naming issue #2673. Excellent diagnostic UX — when the guard fires, the user sees exactly why escalation continued.
- Both `Note:` strings include the actual best-so-far percentage and the floor — easy to grep in audit logs.

### Tests (`tests/test_layer_escalation.py:1126+`)

All three new tests verified passing locally:

- `test_completion_floor_bypasses_zero_overflow_below_floor` — 0% completion, `call_count >= 2`
- `test_completion_floor_bypasses_stagnation_below_floor` — 10% completion, `overflow=5`, `call_count >= 3`
- `test_completion_floor_does_not_affect_high_completion` — 80% completion, `call_count == 1` (regression guard)

The third test is the important one: it pins the existing zero-overflow behavior for boards that route most of their nets, so this PR can't silently regress chorus-test / board 03 / board 06.

### CI + merge state

- All 7 checks pass: C++ Build, Diff-Pair Routing Regression, Lint & Format, Routed PCB DRC Check, Test, Type Check, kicad-cli Round-trip Smoke
- Based on `15014a72` (current `origin/main`, includes #2680). No rebase needed.
- Local: `pytest tests/test_layer_escalation.py tests/test_route_cmd_*.py` → 182 passed (matches PR claim)
- `.github/routed-drc-tolerance.yml` board 05 floor unchanged at 15 (committed routed PCB still satisfies it; regression only manifests on re-route)

- The two `elif` branches share most of their predicate with the parent `if` — could be refactored to a single conditional that picks between break vs Note path. Current form is more explicit about which heuristic is bypassed in each diagnostic message, so leaving as-is is fine.
- 50% threshold isn't user-configurable. If #2681's bisection identifies a slow A* path that pushes some boards just below 50% sustainably, may need to expose this as a flag. Not a blocker today.

Approving. Apply `loom:pr` for merge.

Issue #2661, opened 2026-05-11
Epic: Match-group meander routing for parallel buses (DDR / MIPI / TMDS)


rjwalters:
# Epic: Match-group meander routing for parallel buses (DDR / MIPI / TMDS)

## Overview

Epic #2556 establishes first-class **differential pair** support: detection, intra-pair clearance, coupled routing, length-skew measurement (Phase 3H #2647 merged), serpentine insertion (Phase 3I #2648 in progress), and a `diffpair_length_skew` DRC rule (Phase 3J #2649 in progress).

A pair is the N=2 case of a more general problem: **N traces that must arrive at their destination with their lengths matched to within a tight tolerance.** Real designs are full of these:

- **DDR data byte**: 8 DQ + 1 DM + 1 DQS-pair = 10 nets matched ±2-5 mil
- **MIPI lane group**: 1 clock pair + 4 data pairs (4-lane CSI/DSI), lane-to-lane skew very tight
- **HDMI / DisplayPort TMDS**: 4 differential pairs sharing one clock, relative-to-clock-pair matching
- **Generic parallel buses**: address/command lines matched to a reference clock (50-100 mil OK), SPI/QSPI groups, byte lanes on parallel memory, video buses

The user observation that motivates this epic: *"differential pairs makes me think about how we might also need 'meander path' routing to match length on a set of traces — for example a parallel bus to a memory or a MIPI interface."*

The good news: **most of the primitives already exist.** `LengthConstraint.match_group` (`router/rules.py:383`), `LengthTracker` (`router/length.py:78-317`), `tune_match_group` (`router/optimizer/serpentine.py:438`), `add_match_group` (`router/core.py:6836`), and `create_match_group` (`router/length.py:320`) form a complete generic match-group pipeline. The DDR_DATA docstring example at `router/core.py:6826-6864` shows the intended use shape.

The bad news: this generic pipeline has the same gaps that Phase 3 of #2556 had to fix for pairs, only multiplied across N traces:

1. **No declaration ergonomics from the schematic side.** Net classes have no `length_match_group` field; users must build `LengthConstraint` objects in Python by net ID. Phase 1B of #2556 added explicit net-class declaration for pairs — groups have no equivalent.
2. **No partner-aware serpentine geometry.** `tune_match_group` calls `SerpentineGenerator.add_serpentine` (`router/optimizer/serpentine.py:438-507`) with no awareness of the other group members or neighboring nets — bulges can land in any direction, including into the next trace.
[...]

PR #2644, opened 2026-05-11
feat(router): diff-pair-aware escape coupling for BGA/QFN/USB-C (closes #2639)

rjwalters:
## Summary

Phase 2F of Epic #2556 makes the escape router couple differential pairs at launch from dense packages. Today's `EscapeRouter` knows per-net trace widths but not pair partnerships -- adjacent diff-pair pads on BGA / QFN / USB-C get independent escapes that diverge, forcing the coupled pathfinder to recover spacing after escape. This PR threads a bidirectional `diff_pair_map` from the autorouter into `EscapeRouter` and adds a pre-pass that emits two coupled escape segments for any pad whose partner pad lives on the same package.

## Scope (per curator's three-dispatcher priority)

- **BGA** (`_escape_bga_rings`) -- paired pre-pass active
- **QFP/QFN/TQFP** (`_escape_qfp_alternating`) -- paired pre-pass active
- **MULTI_ROW_CONNECTOR / USB-C** (`_escape_multi_row_connector`) -- paired pre-pass active
- SSOP/TSSOP, SOP, radial -- fall through to single-ended for v1
- Cross-package coupling (pair partner on a different package) is intentionally out of scope -- the main pathfinder owns that case.

## Five-gate verification chain

Per the issue's #2587 dormant-signal lesson, every link in the wiring chain is exercised by a concrete assertion-able test (not just an inspection). 22 new tests in `tests/test_escape_diffpair.py`:

| Gate | Verification |
|---|---|
| **1: CLI → Autorouter map** | `Autorouter.get_diff_pair_map()` returns `{"USB_D+": "USB_D-", "USB_D-": "USB_D+"}` for paired nets, `{}` otherwise |
| **2: Autorouter → EscapeRouter ctor** | All three construction sites (`core.py:7271`, `orchestrator.py:563`, `orchestrator.py:1183`) thread the map |
| **3: ctor stores AND `generate_escapes` consults** | `EscapeRouter.diff_pair_map` stores the map; `diff_pair_segment_calls` counter bumps on `generate_escapes` |
| **4: Per-dispatcher reach** | BGA + QFN + USB-C fixtures each invoke `_escape_diff_pair_segment` once; negative cases (no relevant pair / partner off-package) keep the counter at zero |
| **5: Grid reservation** | Paired escape endpoints survive `apply_escape_routes` to the grid-mark call that feeds both Python and C++ A* search |

## Geometry verification

[...]

Issue #2639, opened 2026-05-11
[Epic #2556] Phase 2F: Diff-pair-aware escape routing (couple at launch from dense packages)


rjwalters:
**Epic**: #2556 - First-class differential pair support
**Phase**: 2 of 4 — Coupled Routing + Escape + Continuity DRC
**Phase Goal**: Coherent coupled routing engagement + diff-pair-aware escape + continuity DRC

Tracked under #2556.

## Description

Today's `EscapeRouter` (`src/kicad_tools/router/escape.py:697`) generates escape routes from dense packages (BGA / QFP / QFN / TQFP / SSOP / TSSOP / SOP / multi-row connectors) using package-type-specific patterns — `_escape_bga_rings`, `_escape_qfp_alternating`, `_escape_fine_pitch_dual_row`, etc. (lines 843-859). It is **net-aware** via `net_class_map` (line 750) for per-net trace widths, but it is **NOT pair-aware**: when adjacent pads belong to a differential pair (USB-C `D+`/`D-` on `A6/A7`, BGA `TX_P`/`TX_N` on neighboring balls), the escape router treats them as independent nets and applies the per-net escape pattern. The result is that the pair is forced to couple **only after escape** — by which point the two traces have already diverged in direction and the coupled pathfinder has to recover spacing.

This issue makes the escape router **couple at the launch** from dense packages, so the pair leaves the package already at the target intra-pair spacing.

### Why this matters

USB-C, BGA, and QFN footprints often place diff-pair pins on adjacent (or near-adjacent) pads with sub-mm pitch. The pair is geometrically already at a tight spacing AT the pads — the escape step is where today's router squanders that coupling, because each pin gets a generic radial / alternating-direction escape sized for a single net. Phase 1 fixed within-pair clearance in the main pathfinder but did nothing for escape. Board 03 evidence: J1 USB-C `A6`/`A7` are at 0.5mm pitch but the escape routes them on opposing sides of the connector before the prepass tries to couple them.

## Affected files / call sites

- `src/kicad_tools/router/escape.py:697` (`EscapeRouter.__init__`) — add `diff_pair_map: dict[str, str] | None = None` constructor arg (net-name → partner-net-name) so escape generators can look up a pad's pair partner
[...]

PR #2685, opened 2026-05-11
feat(validate): wire diffpair_length_skew producer side (closes #2675)

rjwalters:
## Summary

Phase 2.5c of Epic #2556. Wires the producer side of the
``diffpair_length_skew`` DRC rule, mirroring the exact shape PR #2653
used for the sister ``diffpair_routing_continuity`` rule (Phase 2.5b).

Before this PR, ``DRCChecker.check_diffpair_length_skew`` constructed
``DiffPairLengthSkewRule()`` with all defaults -- the rule's guard
short-circuited to a no-op (``rules_checked == 0``) even when the
autorouter consumer supplied a ``net_class_map``.

## What changed

- **New static helper** ``DiffPairLengthTracker.measure_net_from_pcb``
  in ``router/diffpair_length.py`` -- sums geometric length
  (segments + vias) from the PCB-schema shape
  (``PCB.segments_in_net`` / ``PCB.vias_in_net``).  Sister primitive
  to the existing ``_measure_route`` (which consumes the
  router-internal ``Route`` form).  Single source of truth for the
  length-with-vias formula.

- **New module** ``validate/diffpair_skew.py`` exporting
  ``derive_skew_data(pcb, net_class_map, board_thickness_mm,
  num_copper_layers)``.  Mirrors the shape of
  ``validate/diffpair_engagement.derive_engagement_state`` (PR #2653).
  Re-runs detection + per-net length measurement against the routed
  PCB, returning ``(skew_data, threshold_map)`` matching the
  ``DiffPairLengthSkewRule`` constructor.

- **Wired** ``derive_skew_data`` + ``derive_engagement_state`` into
  ``DRCChecker.check_diffpair_length_skew`` (``validate/checker.py``).
  When ``net_class_map`` is provided, the rule now receives non-empty
  ``skew_data``, ``engaged_pairs``, and ``threshold_map``.

- **Drift-prevention tests** (``tests/test_validate_diffpair_skew.py``,
  18 new tests):
  - ``derive_skew_data`` matches ``DiffPairLengthTracker.get_all_skews``
    byte-for-byte for the same physical routing expressed in both
    forms (symmetric, asymmetric, via-traversing).
  - ``DEFAULT_SKEW_TOLERANCE_MM == NetClassRouting.effective_skew_tolerance()`` default.
  - Standalone-CLI graceful no-op (``net_class_map=None`` -> empty).
  - One-sided-declaration policy (P-only / N-only in ``net_class_map``).
[...]

PR #2686, opened 2026-05-11
router: reserve inner-layer corridor for paired escapes (#2677)

### LINKED ISSUE #2677: router: BGA partner-via escape blocks USB3_TX1+/- on board 06 (2/21 nets unrouted)
## Symptom

Board 06 (diffpair-test, created by Phase 4L / PR #2671) routes 19/21 signal nets (90%) on a 4-layer JLCPCB stackup. The two unrouted nets are **USB3_TX1+** and **USB3_TX1-**, blocked by partner-net escape vias at the USB-C SS BGA-49 simulator footprint.

The Phase 2F (#2639, PR #2644) escape coupling correctly identifies USB3_TX1+/- as a diff pair and launches a paired escape, but the partner-net via placement consumes routing channels needed by the pair itself.

## Root cause hypothesis

The escape pass at `src/kicad_tools/router/escape.py` (post-#2644) computes coupled escape segments via `_escape_diff_pair_segment` for BGA dispatchers, but the via placement for other escaping nets doesn't reserve enough corridor for the diff pair's downstream routing on inner layers. The pair launches OK from the BGA but can't continue to its destination because the partner nets have occupied the next-layer channels.

## Likely fix surfaces

1. **Escape-order priority**: route diff pairs FIRST in the escape pass, so single-ended partner nets work around them rather than colonizing the corridor.
2. **Via-keep-out for paired escapes**: when a diff pair launches from a dense package, reserve a corridor on the inner layer for the pair to continue, preventing single-ended partner vias from blocking.
3. **Match-group-aware escape ordering**: extends the priority logic to handle Epic #2661 match groups too.

## Reproduction

Run on `main` after PR #2671 lands:
```
cd boards/06-diffpair-test
kct route --seed 42 --differential-pairs --auto-layers --manufacturer jlcpcb-tier1 --timeout 240
kct check --mfr jlcpcb-tier1 boards/06-diffpair-test/output/diffpair_test_routed.kicad_pcb
```
Confirm USB3_TX1+ and USB3_TX1- are unrouted (no segments for those net IDs).

## Acceptance criteria

- [ ] Board 06 routes USB3_TX1+ and USB3_TX1- on the BGA-49 simulator footprint.
- [ ] Total routed completion for board 06 reaches 21/21 (100%).
- [ ] Drift-prevention test (mirror #2639's 5-gate chain): synthetic BGA with one diff pair + two single-ended nets → diff pair escape segments survive into grid reservations before single-ended via placement.

## Related

- Epic #2556 Phase 2F (#2639, PR #2644 — established the escape coupling primitive)
- Phase 4L (#2658, PR #2671 — board 06 exposes the gap)
- Phase 4N (#2660 — CI gate will catch this once board 06 hits 0 DRC)

Tracked under #2556.

rjwalters:
## Summary

- Adds `EscapeRouter._reserve_pair_continuation_corridor` that reserves a forward-extruding inner-layer keep-out region for Phase 2F paired escapes; `Grid._mark_via` consults the new reservation map and skips cells reserved for other nets, preventing partner-net through-hole vias from colonising the diff pair's continuation channel.
- Helper accepts a generic `list[EscapeRoute]` so Epic #2661 Phase 2E (`tune_match_group_v2`) inherits the primitive for N>=3 match groups without an API change.
- 2-layer boards are explicitly skipped (the `is_outer` guard preserves boards 01-05 behaviour exactly).
- Adds 6 new tests in `TestCorridorReservation` (5-gate drift-prevention chain + 2-layer no-op guard) mirroring the #2639 / PR #2644 pattern.

## Verification

Board 06 (seed=42, jlcpcb-tier1, 240s timeout):

| | Before | After |
|---|---|---|
| Nets routed | 7/22 (32%) | **20/22 (91%)** |
| USB3_TX1+/- | partial (1/2 pads) | **fully routed** |

Remaining 2 partials (VBUS_USB, PCIE_RX+) are unrelated to the BGA partner-via escape and out of scope for this issue.

The board 06 routed-pcb artifact and `.github/routed-drc-tolerance.yml` entry are NOT updated by this PR: `kct route` regenerates the file with the real KiCad pcbnew serialization (version `20260206`), while the committed artifact uses the demo generator's hardcoded version `20240108`. Tightening the tolerance and regenerating the artifact is blocked on the upstream version-string drift (pre-existing; should be tracked separately).

## Test plan

- [x] All 304 escape + grid tests pass (`tests/test_escape*.py` + `tests/test_router_grid.py`)
- [x] 6 new corridor tests (`TestCorridorReservation`) pass the 5-gate drift-prevention chain
- [x] No new lint errors introduced (`ruff check` shows same 20 pre-existing warnings before and after my changes)
- [x] Board 06 routing improvement verified against baseline (7/22 → 20/22)
- [x] No regression on the 22 pre-existing `test_escape_diffpair.py` tests
- [x] 16 unrelated test failures (subgrid/router_cpp/router_integration/diffpair_length_tuning/cli_check_diffpair/fine_pitch) verified as pre-existing on baseline (stash-then-rerun)

Closes #2677


rjwalters:

Thorough review of the corridor-reservation fix. The implementation matches the curator's pinned design and verifies cleanly through all gates.

### Verification

**1. Helper wiring (`escape.py:1125-1137`)** — confirmed correct
- Called inside `_generate_paired_escapes` after `_escape_diff_pair_segment` produces both endpoints but BEFORE the function returns.
- `generate_escapes` then invokes the per-package dispatcher (`_escape_bga_rings`, etc.) which places partner vias through `_mark_via` → reservation is in place when partner vias are marked. Timing matches the design intent in the issue.
- Two instrumentation counters present and tested.

**2. Grid additions (`grid.py`)** — clean
- Sparse `_reserved_for_nets: dict[(layer, y, x), frozenset[int]]` (line 459).
- Public API: `reserve_corridor_cells`, `clear_corridor_reservations`, `reserved_cell_count`, `is_reserved_for` all present with docstrings.
- `_mark_via` fast-paths with `has_reservations = bool(...)` at line 1869, preserving byte-identical pre-fix behavior when the map is empty.
- Per-net owner-set semantics: an owner-net via DOES block its own corridor cell (verified by gate (b) sanity check), partner-net vias skip.

**3. Drift-prevention chain (`test_escape_diffpair.py:TestCorridorReservation`)** — 6/6 pass locally
- Gate (a): reservations counted, grid `reserved_cell_count()` matches `pair_corridor_reserved_cells`.
- Gate (b): partner via at corridor centroid does not block reserved cell; owner via does (sanity).
- Gate (c): reservations persist through `apply_escape_routes`.
- Gate (d): 3-member match-group call locks in the Epic #2661 signature.
- Gate (e): empty `diff_pair_map` produces zero reservations (regression guard).
- 2-layer no-op: paired segments still generated, no reservation made (boards 01-05 protected).

**4. Routing improvement**
- Board 06: 7/22 → 20/22 (91%). USB3_TX1+/- fully routed — the issue's primary acceptance criterion met.
- Remaining 2 unrouted (VBUS_USB, PCIE_RX+) are explicitly different nets, unrelated to the BGA partner-via blocker. Out of scope for #2677. Note: the issue's stretch goal of 100% is not hit, but the binding gap identified in the root-cause is fixed.

**5. Epic #2661 hook** — confirmed
- Helper signature `members: list[EscapeRoute]` plus net-owner-set scaling makes the API ready for N>=3 match groups without a follow-up signature change. Gate (d) test pins this contract.

**6. Out-of-scope deferrals** — appropriate
- PCB file deferral verified: `boards/06-diffpair-test/output/diffpair_test_routed.kicad_pcb` is at version `20240108` (demo generator), while KiCad pcbnew emits `20260206`. Regenerating would break round-trip serialization. Pre-existing PR #2671 artifact, not introduced by this PR.
- `.github/routed-drc-tolerance.yml` board 06 entry (28) is correctly left untouched since the on-disk PCB is unchanged.

**7. CI** — all green
- Test ✓, Type Check ✓, C++ Build ✓, Lint & Format ✓, Routed PCB DRC ✓, Diff-Pair Routing Regression ✓, kicad-cli Round-trip Smoke ✓.
- `mergeStateStatus: CLEAN`.

**8. Coordination** — no overlap
- Only other open PR is #2685 (`feat(validate): wire diffpair_length_skew producer side`), which touches `validate/*` and a different `diffpair_length.py`. No file conflict.
- Referenced "#2681/#2682" are issue numbers, not PRs — no PR overlap.

### Minor nits (non-blocking)

- **Format**: `grid.py:1924-1927` has a 2-line f-string that `ruff format` would join. CI's `Check formatting` has `continue-on-error: true` and the baseline already carried 20 format warnings, so this didn't block CI. A `uv run ruff format src/kicad_tools/router/grid.py` would clean it up in a follow-up.
- **C++ backend awareness**: `cpp_backend.mark_via` does not consult `_reserved_for_nets`. This is fine for this PR because the escape phase operates on the Python grid and partner vias get rejected there before the Python→C++ sync happens. If a future change moves any of the escape-phase via marking onto the C++ grid, the reservation logic must be ported. Worth a comment if you spin a follow-up, but not blocking here.
- **PCB regeneration follow-up**: filing a tracking issue for the `20240108`/`20260206` version-drift unblock would help close out the tolerance entry once routing is fully clean.

### Verdict

LGTM. Approving and handing to merge queue.

— Loom Judge

Issue #2770, opened 2026-05-12
[#2746-C] router: kct route write path drops zones from input PCB


rjwalters:
Split-out child of **#2746** (board 05 BLDC regression, urgent). Per the 2026-05-11 curator analysis.

## Problem

Auto-pour creates 5 zones during board 05's route (log: \`Auto-pour: created 5 zone(s)\`) but the committed routed PCB at \`boards/05-bldc-motor-controller/output/bldc_controller_routed.kicad_pcb\` has **0 zones**. The \`kct route\` write path drops them.

This also contributed to board 01's silent empty-route bug fixed by PR #2748, but that fix was upstream (in the build chain). This issue addresses the write path itself.

## Acceptance criteria

- [ ] Zones present in the input PCB (either pre-existing or auto-poured) are preserved in the routed output PCB.
- [ ] Regression test: pass a PCB with N zones through \`kct route\`, assert the output has N zones (same nets, same layer assignments).
- [ ] Board 05 routed PCB has ≥4 zones (VMOTOR, +3.3V, +5V, GND).

## Files

- \`src/kicad_tools/cli/route_cmd.py:1508-1519, 2030-2078, 2548-2587\` (write path)

## Related

- #2746 (parent umbrella, urgent)
- #2740 / PR #2748 (board 01 fix that addressed the upstream symptom)
- #2746-D (zone layer distribution — related; both affect zone correctness)

rjwalters:
## Curator Notes (2026-05-12)

### Claim verification: `kct route` write path does NOT drop zones

Reproduced on `main` (commit `3bef2e7a`) against the current
`boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb`:

```
$ kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb \
    -o /tmp/out.kicad_pcb --max-layers 2 --timeout 10 --skip-drc --quiet
```

Result on disk: **5 zones** (`VMOTOR`, `+5V`, `+3.3V`, `GND`, `PWR_LED`)
+ 58 segments. The write path at `route_cmd.py:2033, 2040-2062` (and
the other three save sites at `:455, :2558-2578, :3107-3137, :5450-5481`)
correctly reads the staged copy via `pcb_path.read_text()`, inserts the
route s-expression with `_insert_sexp_before_closing`, and writes back
to `output_path` — zones flow through intact.

### Why the parent issue's audit (#2746) saw 0 zones

The committed routed PCB at
[...]

PR #2753, opened 2026-05-12
fix(pcb): coordinate-space invariant for segments/vias/zones + ERC build step

rjwalters:
## Summary

Closes the board-00 audit follow-up (#2742): LED_ANODE was being
mis-classified as `incomplete` despite being fully routed, and ERC
was silently skipped from the build pipeline.

## Blast radius (added after judge review)

The coord-space invariant fix correctly surfaces a previously-hidden
class of real `clearance_pad_segment` defects on every board with a
non-zero `Edge.Cuts` origin. Prior to this PR, `Segment.start`/`end`,
`Via.position`, and `Zone.polygon` lived in sheet-absolute coordinates
while `Footprint.position` was already board-relative; downstream DRC
silently compared mixed coord-spaces and missed any clearance violation
where the trace endpoint was ~origin distance from a pad in coordinate
space. Judge confirmed a concrete example: OSC_IN trace on board 05
visibly passes through Y1.2 pad with -0.375mm clearance.

Per-board impact and allowlist bumps (all in `.github/routed-drc-tolerance.yml`):

| Board | Pre-fix | Post-fix | Δ | Newly-visible | Tracking |
|---|---|---|---|---|---|
| 03-usb-joystick | 1 | 83 | +82 | 81x clearance_pad_segment, 1x clearance_pad_via | #2755 |
| 05-bldc-motor-controller | 15 | 120 | +105 | 105x clearance_pad_segment | #2756 |
| 06-diffpair-test | 28 | 43 | +15 | 15x clearance_pad_segment | #2757 |
| 07-matchgroup-test | 55 | 65 | +10 | 10x clearance_pad_segment (floor stays 80, within headroom) | #2758 |
| 00/01/02 | unaffected | unaffected | 0 | — (zero-origin or no pre-existing clearance issues) | — |

These violations are real defects that were always there but invisible
to DRC. The bumped floors reflect the new ground truth; per-board issues
(#2755-#2758) capture the re-routing follow-up that will drive them back
down. The diff-pair routing regression CI gate (board 06) is unblocked
by the floor bump from 28 -> 43.

Additionally, the fix introduces a small mixed-coord-space inconsistency
in `zones/generator.py:_check_overlap` (sheet-absolute `board_outline`
vs board-relative `existing.polygon`). The zone writeout path is correct;
only the overlap-warning heuristic is affected. Deferred and tracked in
#2759, with a `TODO(#2759)` comment and docstring note in the source.

## Changes

[...]

Issue #2746, opened 2026-05-12
Board 05 (bldc-motor-controller) regression NOT recovered (re #2673): 2/35 live route, 0 zones, PHASE_A/B/C unrouted, DRC at floor with zero slack


rjwalters:
## Symptom

Board 05-bldc-motor-controller's regression tracked by **#2673** (closed today) is **not recovered on current main**. The closure was based on PR #2683 (escalation completion-floor guard) merging plus child issues #2681/#2682 closing — but a fresh audit shows the underlying conditions remain.

## Audit

- Commit SHA: `0eaadc7c4e39597171e60420668aac184d7ecc74` (main)
- Live router: 2/35 signal nets routed (**5.7%** — well below #2673's ≥38% floor)
- Committed routed PCB: 207 segments + 42 vias but **all 40 nets still incomplete** per `kct net-status`
- DRC: **15 errors, exactly at the `.github/routed-drc-tolerance.yml` floor — zero slack**
  - 9× `clearance_segment_segment`
  - 4× `clearance_segment_via`
  - 1× `diffpair_clearance_intra` (ISENSE_C+/-)
  - 1× `dimension_drill_clearance`
- ERC: 0/0
- Export: blocked at preflight (FAIL — 0 artefacts)
  - BOM↔PCB drift: BOM lists C18/C19 not on PCB; PCB has C15/C16/MH1-4 not in BOM
  - 51/51 components missing LCSC part numbers
  - 37/51 components missing footprints
- **Zones: 0** — VMOTOR/+3.3V/+5V/GND/PWR_LED are all unfilled. 20 pads listed \"needs routing\" on VMOTOR.
- PHASE_A / PHASE_B / PHASE_C: all three completely unrouted. Router spent 146-160s **each** on these three nets before aborting; the 240s budget was exhausted by the time only 5.7% of nets had been attempted.
- Auditor log: `/tmp/audit_2026-05-11/05-bldc-motor-controller/`

## Root-cause hypotheses

1. **Per-net routing budget unbounded** — PHASE_A consuming 160s on a 240s board-level budget is pathological. There should be a per-net time-box and a retry policy.
2. **Zone generation never runs for board 05** — 0 zones in the committed PCB. Either `_run_step_zones` is failing silently or board 05's design.py does not produce a PCB that the zone generator can consume. Without VMOTOR/GND zones, PHASE_A/B/C have no reasonable channel.
[...]

PR #2777, opened 2026-05-12
feat(zones): partition shared-layer pour outlines into per-net bboxes

### LINKED ISSUE #2771: [#2746-D] router: auto-poured zones all land on F.Cu with overlapping outlines
Split-out child of **#2746** (board 05 BLDC regression, urgent). Per the 2026-05-11 curator analysis; expanded with code verification, reproduction, and recommended strategy on 2026-05-12.

## Problem

On a 2-layer stackup like board 05 (bldc-motor-controller), auto-pour generates a separate zone per pour net but assigns every power net to **F.Cu with the same full-board outline**. KiCad's zone-fill resolver awards copper to the highest-priority zone in the overlap region, so on board 05 the priority-4 `VMOTOR` zone wins F.Cu entirely and `+5V`, `+3.3V`, and `PWR_LED` get *zero usable copper* despite being created.

Even on 4-layer boards, the existing assignment puts the 2nd-Nth power nets on `F.Cu` with shared full-board outlines — same failure mode, just affecting fewer nets.

## Verified current behavior

Run on `src/kicad_tools/zones/generator.py::_assign_layers_for_pour_nets` with board 05's pour set `{VMOTOR, +5V, +3.3V, GND, PWR_LED}`:

| Stackup | GND | VMOTOR | +5V | +3.3V | PWR_LED |
|---|---|---|---|---|---|
| 2-layer | B.Cu p1 | **F.Cu p4** | F.Cu p3 | F.Cu p2 | F.Cu p1 |
| 4-layer | In1.Cu p1 | In2.Cu p0 | **F.Cu p1** | F.Cu p2 | F.Cu p3 |

All F.Cu rows use the *same* `board_outline` polygon (see `ZoneGenerator.add_zone` `actual_boundary = boundary if boundary is not None else self.board_outline` at `src/kicad_tools/zones/generator.py:558`). The current overlap detection (`_check_overlap`, lines 411-509) emits a warning but does not change the assignment — and after #2410, the priorities are *distinct*, which only prevents the zero-copper warning, not the zero-copper *outcome*.

## Root cause

The current allocator (`_assign_layers_for_pour_nets`, generator.py:781-915) was designed to solve a different problem: split-ground domains and the equal-priority zero-copper bug (closed by #2410, #2593, #2041). It does not partition `F.Cu` *geometrically*, so N power nets on the same physical layer with identical outlines still collapse to "highest priority wins everything."

Board 05's committed PCB (`boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb`) has **0 zones**, consistent with the parent #2746 audit. The routed output PCB (`bldc_controller_routed.kicad_pcb`) likewise has zero zones; current routing therefore must carry VMOTOR/+5V/+3.3V on signal traces rather than copper pours, which is one of the causes of the 5.7% completion rate.

## Reproduction

```python
from kicad_tools.zones.generator import _assign_layers_for_pour_nets
from kicad_tools.router.net_class import NetClass

pour_nets = [
    ('VMOTOR', NetClass.POWER), ('+5V', NetClass.POWER),
    ('+3.3V', NetClass.POWER), ('GND', NetClass.GROUND),
    ('PWR_LED', NetClass.POWER),
]
print(_assign_layers_for_pour_nets(2, pour_nets))
# [('GND','B.Cu',1), ('VMOTOR','F.Cu',4), ('+5V','F.Cu',3), ('+3.3V','F.Cu',2), ('PWR_LED','F.Cu',1)]
# -> 4 zones with identical full-board outlines on F.Cu; only VMOTOR fills.
```

Then running `auto_pour_if_missing` on board 05 will create 5 zones — 1 on B.Cu (GND, usable), 4 on F.Cu (all sharing the board outline, only VMOTOR usable).

## Recommended strategy: net-class layer + geometric outline (hybrid)

After cross-referencing with PR #2753 (board-relative coord-space invariant) and PR #2763 (`_check_overlap` origin shift), the clean fix is a two-step assignment:

1. **Net-class layer routing** (already partly done):
   - GROUND -> dedicated layer (B.Cu on 2-layer; In1.Cu on 4-layer).
   - First POWER net (highest current, by convention VMOTOR if present, else highest classification rank) -> next-best plane layer (F.Cu on 2-layer; In2.Cu on 4-layer).
   - Remaining POWER nets -> share whatever layer is left.

2. **Geometric outline restriction** (NEW): for any power layer that hosts more than one zone, replace the default `board_outline` with a **per-net bounding region** derived from that net's pads. Two reasonable options:
   - **Bounding-box per net** with a configurable margin (simplest; uses the existing pad enumeration in `_build_pads_for_net` at `mcp/tools/routing.py:606`).
   - **Voronoi partition** of the board outline using the per-net pad centroids (cleaner non-overlapping regions; needs shapely, which is already a dep used by `_inset_polygon` in generator.py:297).

   The new boundary is passed to `ZoneGenerator.add_zone(..., boundary=region)`. This is the same `boundary=` hook called out in #2041's recommendation #2 (zones split). The hook already exists; auto_pour just needs to use it.

### Why hybrid over pure net-class-driven

A pure "POWER -> In2.Cu / GND -> all layers" mapping does not scale: on 2-layer boards there is no inner layer, and on any board with 3+ power nets you still have multiple nets sharing F.Cu. Geometric outlines are the only way to give each net real copper.

### Why hybrid over pure geometric-disjoint

Pure geometric works but mis-allocates current capacity. GND needs to be a continuous plane for return paths; carving it into a small region around a few pads breaks every signal that crosses the unfilled gap. Keep GND on a dedicated layer with the full outline, and apply geometric partitioning only to the power tree on the shared layer.

## 2-layer fallback

On boards like 05 with 2 layers and N≥2 power nets:

- **Primary**: per-net bounding-box outlines (margin 1–2 mm) on F.Cu. Each power net gets copper only over its own component cluster. Acceptable for most boards because power components tend to be co-located.
- **Secondary** (if bounding boxes still overlap, e.g. two VRMs share footprints): emit a warning recommending stackup upgrade to 4-layer. Optionally fall back to "single-power-zone + signal-trace-routing for the rest" — i.e. assign the highest-current net (VMOTOR) to F.Cu and leave +5V/+3.3V to be routed as wide signal traces.
- **Tertiary**: caller-override via CLI flag (`kct route --pour-region NET=BBOX`) for hand-tuned outlines. Out of scope for this issue; track separately.

## Cross-reference with recent merges

- **PR #2753** (coord-space invariant): zones now load board-relative; `Zone.polygon` and `board_outline` live in different frames. Any new outline-construction helper must produce sheet-absolute boundaries (matches the `add_zone(boundary=)` contract — see `_check_overlap` docstring at generator.py:423-434). Use `self.board_outline` (sheet-absolute) as the reference frame and pass through `_inset_polygon` if needed.
- **PR #2763** (`_check_overlap` origin shift): the AABB overlap test correctly compares mixed frames now, so the new geometric-outline path automatically gets correct overlap warnings.
- **#2410, #2041, #2593** (closed precedents): all addressed *equal-priority* zero-copper. This issue is the unfinished follow-on: *distinct-priority but identical-outline* still produces zero copper. The fix completes the work started in those issues.

## Acceptance criteria

- [ ] On board 05 (2-layer, pour set above), the 4 power-net zones do NOT share the same outline on F.Cu. Each power net has a bounding-box (or Voronoi region) outline derived from its pads + a configurable margin (default 1.5 mm).
- [ ] `kct route` on board 05 produces *at least 4 zones with non-zero filled copper* in the output PCB (verifiable via `kicad-cli pcb` zone fill area, or by counting `filled_polygon` blocks per zone in the routed `.kicad_pcb`).
- [ ] On 4-layer stackups, GROUND remains on a dedicated inner layer with the full board outline (return-path plane preserved). POWER nets share `F.Cu` only when they cannot fit on an inner layer, and use per-net bounding outlines when they do share.
- [ ] No overlap warnings emitted by `_check_overlap` for the board 05 pour set after the fix.
- [ ] Regression test in `tests/test_build_zones_step.py` (new test class `TestAutoPourGeometricOutlines`):
   - 2-layer board with `{VMOTOR, +5V, +3.3V, GND, PWR_LED}` pour set → 5 zones, no overlap warnings, every power zone has distinct boundary AABB.
   - 4-layer board with same pour set → GND on In1.Cu full-outline; VMOTOR on In2.Cu full-outline; remaining power nets on F.Cu with bbox outlines.
   - Edge case: power net with a single pad → fall back to a default radius around that pad (do not produce zero-area polygon).
- [ ] Existing tests in `tests/test_zone_generator.py` (`TestZoneOverlapDetection`, `TestZoneOverlapNonzeroOrigin`) still pass — outline-overlap warnings remain correct for the cases where the caller *does* pass a custom boundary that overlaps an existing zone.
- [ ] Document the geometric-partition rule in `auto_create_zones_for_pour_nets`'s docstring (generator.py:918-953) and in `auto_pour.py`'s module docstring (lines 31-39 already describe the layer/priority allocator — add an "outline allocator" section).

## Implementation guidance

1. Add a new helper `_compute_pour_outlines(pcb, assignments, margin_mm)` in `zones/generator.py` that, for each (net, layer) assignment, returns a polygon. For nets that are the sole zone on their layer (e.g. GND on B.Cu), return `None` so the caller uses `board_outline`. For shared-layer nets, compute a bounding box of the net's pads (use `pcb.footprints` and filter pads by net name) inflated by `margin_mm`, clipped to `board_outline`.
2. In `auto_create_zones_for_pour_nets` (generator.py:918-980), call `_compute_pour_outlines` after the layer assignment loop, and pass the resulting `boundary=` to each `gen.add_zone(...)` call.
3. The single-pad fallback: if a net has <2 pads or its pads have zero geometric extent, use a small default square (e.g. 4 mm × 4 mm) centered on the pad.
4. Edge clipping: reuse `_inset_polygon` (generator.py:297) to ensure the per-net outline never extends past the board edge. The `edge_clearance` parameter that already flows through `auto_create_zones_for_pour_nets` should still apply to the per-net outlines.
5. The `_check_overlap` helper does NOT need changes — once outlines are disjoint, the AABB test returns no overlap and existing tests continue to pass.

## Files

- `src/kicad_tools/zones/generator.py:781-915` (`_assign_layers_for_pour_nets`) — keep, augment.
- `src/kicad_tools/zones/generator.py:918-980` (`auto_create_zones_for_pour_nets`) — call new `_compute_pour_outlines`.
- `src/kicad_tools/zones/generator.py` (new function `_compute_pour_outlines`) — add.
- `src/kicad_tools/router/auto_pour.py:281-391` (`auto_pour_if_missing`) — no change required; it delegates to `auto_create_zones_for_pour_nets`.
- `tests/test_build_zones_step.py` — extend with `TestAutoPourGeometricOutlines` class.
- `tests/test_zone_generator.py` — verify existing overlap-detection tests still pass.

## Dependencies

- Independent of #2770 (`#2746-C` zone preservation through router write path), but the *user-visible* benefit on board 05 requires both: this issue makes auto-pour produce real copper; #2770 makes that copper survive routing. Recommend builders implement both for the board 05 acceptance criterion in #2746 to flip green.

## Related

- #2746 (parent umbrella, urgent — board 05 regression)
- #2770 (sibling #2746-C: kct route drops zones on write-back)
- #2410 (closed: distinct priorities; this issue extends to geometric partitioning)
- #2041 (closed: same-outline winner-take-all; surfaced the need for `zones split`; this issue implements the automatic version)
- #2593 (closed: split-ground priority distinction; same heuristic family)
- PR #2753 (coord-space invariant — informs frame contract for new outlines)
- PR #2763 (`_check_overlap` origin shift — verifies overlap detection survives the fix)

## Builder skill target

`loom:builder` — single PR, scoped to `zones/generator.py` plus new test class. Estimated medium-size change (~150 LOC + tests).

rjwalters:
## Summary
Adds the **outline allocator** to `auto_create_zones_for_pour_nets`. When two
or more pour zones land on the same copper layer, each zone now gets a per-net
bounding-box outline (1.5 mm margin around the net's pads, clipped to the
board outline) instead of the full board outline. This is the fix for the
"all power zones share F.Cu with identical outlines, so only the highest
priority gets copper" failure mode on board 05 documented in #2771.

Sole-layer zones (GND on B.Cu of a 2-layer board, GND on In1.Cu of a 4-layer
board) keep the full board outline so the return-path plane stays continuous.

## Changes
- **`src/kicad_tools/zones/generator.py`**
  - New helper `_compute_pour_outlines(pcb, assignments, board_outline, margin_mm)`.
  - New helpers `_net_pad_positions_absolute`, `_bbox_polygon`, `_clip_polygon_to_outline`.
  - `auto_create_zones_for_pour_nets` now calls `_compute_pour_outlines` and
    passes each result as `boundary=` to `gen.add_zone(...)`.
  - Sole-zone layers receive `None` so `add_zone` falls back to `board_outline`
    via the existing `boundary if boundary is not None else self.board_outline`
    branch.
  - Single-pad fallback: a power net with only one pad gets a 4 mm default
    square centered on the pad.
  - Boundaries are produced in sheet-absolute coordinates (matches the
    `add_zone(boundary=)` contract established by PR #2753).
- **`src/kicad_tools/router/auto_pour.py`** — docstring updated to describe
  the outline allocator (no behavioral change; this module already delegates
  to the generator).
- **`tests/test_build_zones_step.py`** — new `TestAutoPourGeometricOutlines`
  class with 5 tests:
  - `test_2layer_power_nets_have_disjoint_bboxes` — board 05 pour set on
    2-layer, F.Cu zones must have pairwise-distinct non-overlapping AABBs,
    no `_check_overlap` warnings.
  - `test_2layer_gnd_retains_full_outline` — sole-layer GND keeps the
    full 100x50 board outline.
  - `test_4layer_gnd_keeps_full_inner_plane` — GND on In1.Cu and VMOTOR on
    In2.Cu both keep the full outline (return-path / power-plane invariant);
    the remaining 3 power nets on F.Cu must have disjoint per-net bboxes.
  - `test_single_pad_net_uses_fallback_square` — single-pad net produces a
    non-degenerate ~4 mm square, smaller than the full board.
  - `test_compute_pour_outlines_returns_none_for_sole_layer` — direct unit
    test of the contract (sole-layer zone -> `None`; shared-layer zone ->
    concrete polygon).

| 4 power-net zones on F.Cu do NOT share the same outline (board 05 pour set) | done | `test_2layer_power_nets_have_disjoke_bboxes` asserts 4 pairwise-distinct AABBs, and a manual check on `boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb` produces 4 different AABBs (VMOTOR/+5V/+3.3V/PWR_LED) where previously they were all `(0,0)-(70,90)`. |
| 4-layer GROUND keeps full-outline on inner layer (return-path plane preserved) | done | `test_4layer_gnd_keeps_full_inner_plane` asserts GND on `In1.Cu` spans the full 100x50 board; VMOTOR on `In2.Cu` also keeps the full outline (sole-zone-on-layer rule). |
| Single-pad fallback produces non-zero-area polygon | done | `test_single_pad_net_uses_fallback_square` asserts width/height > 0.5 mm and the polygon is smaller than the full board. |
| No overlap warnings on the (synthetic, physically-disjoint) board 05 pour set | done | `test_2layer_power_nets_have_disjoint_bboxes` asserts `"WARNING" not in captured.err`. |
| Existing tests in `TestZoneOverlapDetection` / `TestZoneOverlapNonzeroOrigin` still pass | done | Full sweep of `test_zone_generator.py`, `test_zone_generator_edge_clearance.py`, `test_zones.py`, `test_zone_api.py`, `test_zones_cmd.py`, `test_pcb_add_zone.py`, `test_pcb_zones.py`, `test_validate_zone_fill.py`, `test_route_cmd_zone_fill.py`, `test_placement_zones.py`, `test_gerber_zone_fill.py`, `test_auto_pour.py`, `test_build_zones_step.py` -> 332 passed, 4 skipped. |
| Documentation in `auto_create_zones_for_pour_nets` and `auto_pour.py` docstrings | done | Both docstrings now have a "Geometric outline partition (#2771)" section explaining the allocator and the disjoint-outline invariant. |

## Test Plan
- [x] All 5 new tests in `TestAutoPourGeometricOutlines` pass.
- [x] Existing `TestZoneOverlapDetection` and `TestZoneOverlapNonzeroOrigin`
      tests pass (332 zone-related tests total, no regressions).
- [x] Layer assignment from `_assign_layers_for_pour_nets` is unchanged for
      the board 05 pour set (`(GND, B.Cu, 1)`, `(VMOTOR, F.Cu, 4)`, etc.).
- [x] On the actual `boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb`,
      VMOTOR/+5V/+3.3V/PWR_LED now have distinct per-net AABBs instead of all
      sharing `(0, 0)-(70, 90)`. (Physical pad clusters on this board still
      overlap, which is the realistic case where bbox is sufficient but not
      perfect -- see the issue's "secondary fallback" note. Voronoi partition
      is the future option.)

## Notes
- The board 05 acceptance criterion in #2746 ("at least 4 zones with non-zero
  filled copper in the routed PCB") additionally requires #2770 (zone
  preservation through the router write-back path). This PR makes auto-pour
  *produce* real copper outlines; #2770 makes that copper *survive* routing.
- Frame: the new outlines are sheet-absolute (`board_outline` frame), matching
  the `add_zone(boundary=)` contract clarified by PR #2753.
- Backward-compatibility: layer / priority assignments from
  `_assign_layers_for_pour_nets` are unchanged. The only behavioral
  difference is the boundary passed to `add_zone` when a layer hosts
  multiple zones.

Closes #2771

### DIFF
--- a/src/kicad_tools/router/auto_pour.py
+++ b/src/kicad_tools/router/auto_pour.py
@@ -37,6 +37,25 @@ automatically by the layer/priority allocator in
 a 2-layer stackup each ground gets a distinct priority on ``B.Cu`` to
 avoid the "zero copper" override that occurs when zones share both
 layer and priority.  See that function's docstring for the full rule.
+
+Geometric outline partition (#2771)
+-----------------------------------
+
+The layer/priority allocator alone is not sufficient when N≥2 zones
+share a single layer (the common case on 2-layer stackups, and the
+fallback path on 4-layer stackups with 3+ power nets).  Distinct
+priorities prevent the explicit zero-copper warning, but KiCad's fill
+resolver still awards the entire overlapping region to the highest
+priority zone, so siblings receive zero usable copper.
+
+To make every zone produce real copper, this module delegates to
+``kicad_tools.zones.generator.auto_create_zones_for_pour_nets``, which
+runs the **outline allocator** (``_compute_pour_outlines``) after layer
+assignment.  Zones that share a layer with one or more siblings get a
+per-net bounding-box outline (default 1.5 mm margin around the net's
+pads, clipped to the board outline), while zones that are the only zone
+on their layer keep the full board outline so return-path planes stay
+continuous.  See that function's docstring for the contract.
 """
 
 from __future__ import annotations
--- a/src/kicad_tools/zones/generator.py
+++ b/src/kicad_tools/zones/generator.py
@@ -3,10 +3,29 @@ Zone generator for creating copper pour zones on PCBs.
 
 This module provides a high-level API for generating copper pour zones,
 with automatic board outline detection and sensible defaults for power nets.
+
+Outline allocator (#2771)
+-------------------------
+
+In addition to the layer/priority allocator (:func:`_assign_layers_for_pour_nets`),
+this module owns the *outline* allocator (:func:`_compute_pour_outlines`).
+The outline allocator runs after layer assignment and decides whether each
+zone should use the full board outline (when it is the only zone on its
+layer) or a per-net bounding region derived from that net's pads (when it
+shares a layer with another zone).
+
+This is the fix for the "all power nets share F.Cu with identical board
+outlines, so the highest-priority zone wins everything" failure documented
+in #2771.  Earlier issues #2410 / #2041 / #2593 made the priorities
+distinct, which silenced the warning, but distinct priorities alone do not
+produce real copper for the losing zones -- the outlines have to be
+geometrically disjoint for KiCad's fill resolver to award copper to more
+than one zone on a shared layer.
 """
 
 from __future__ import annotations
 
+import math
 import sys
 import uuid as uuid_module
 from collections.abc import Sequence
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge Review: APPROVED

The outline allocator cleanly solves the "shared-layer pour outline" failure mode from #2771. Code structure follows the established `_assign_layers_for_pour_nets` pattern, the sole-zone vs shared-zone hybrid is correctly factored, and the new helpers are well-documented.

### Verification

- **Local zone test sweep (332 tests)** — all pass (matches PR claim). Includes `TestZoneOverlapDetection`, `TestZoneOverlapNonzeroOrigin`, `TestAutoPourGeometricOutlines`, `test_auto_pour.py`, `test_zone_generator*.py`.
- **CI** — all 8 checks pass (Lint, Type Check, Test, Routed PCB DRC, Diff-Pair Routing, Match-Group Routing, C++ Build, kicad-cli round-trip).
- **Merge state** — CLEAN, MERGEABLE.
- **Lint/format** — `ruff check` and `ruff format --check` both clean on all 3 changed files.
- **Backward compatibility** — `_assign_layers_for_pour_nets` is unchanged; layer/priority allocation is preserved. The only behavioral delta is the `boundary=` argument passed to `add_zone`.

### Acceptance Criteria

| Criterion | Status |
|-----------|--------|
| 4 power-net zones on F.Cu have distinct AABBs (board 05 pour set) | Verified by `test_2layer_power_nets_have_disjoint_bboxes` (pairwise non-overlap assertion). |
| GND keeps full-outline on sole layer (2L B.Cu, 4L In1.Cu) | Verified by `test_2layer_gnd_retains_full_outline` and `test_4layer_gnd_keeps_full_inner_plane`. |
| Single-pad fallback produces non-zero-area polygon | Verified by `test_single_pad_net_uses_fallback_square` (4mm default square). |
| No overlap warnings from `_check_overlap` | Asserted directly in the 2-layer and 4-layer tests. |
| Direct unit test of `_compute_pour_outlines` contract | `test_compute_pour_outlines_returns_none_for_sole_layer`. |
| Sheet-absolute coordinate frame matches `add_zone(boundary=)` contract from #2753 | `_net_pad_positions_absolute` correctly adds `pcb.board_origin` to footprint position; `_compute_pour_outlines` returns polygons clipped against `gen.board_outline` (also sheet-absolute). |
| `edge_clearance` propagates to per-net outlines | Implicit-but-correct: `gen.board_outline` is already inset when `edge_clearance` is set, and `_compute_pour_outlines` uses it as the clip mask. Docstring update is accurate. |

### Design notes (LGTM)

- The two-tier design (Shapely intersection with AABB-clip fallback in `_clip_polygon_to_outline`) is consistent with `board_geometry.py` and survives missing-shapely environments.
- The single-pad fallback (`SINGLE_PAD_FALLBACK_SIDE_MM = 4.0`) chosen larger than `2 * margin` is a thoughtful default — avoids a degenerate 0x0 polygon that KiCad would silently drop.
- Skipping pads with `net_name != net_name` AND empty-name pads AND `#`-prefixed footprint refs is the right filter set for pour-net pad enumeration.
- The "no pads found -> fall back to None" branch in `_compute_pour_outlines` is the safe choice (zone still gets created, just with full outline).

### Follow-up nit (not blocking)

`_net_pad_positions_absolute` uses `math.radians(-fp.rotation)` for the pad rotation, while every other absolute-pad-position calculation in the codebase uses positive rotation:

- `PCB.get_pad_position` (canonical, `schema/pcb.py:3628`): `math.radians(fp.rotation)`
- `validate/rules/netlist.py:107`: `math.radians(fp.rotation)`
- `drc/incremental.py:585`: `math.radians(fp.rotation)`
- `optim/keepout.py:245`: `math.radians(footprint.rotation)`

The test fixtures here all use `fp.rotation = 0`, so the sign disagreement is masked. For real boards with rotated footprints (90/180/270 deg components are very common), the bbox of a single net's pads may end up mirrored relative to the actual pad locations — typically still disjoint from sibling bboxes, and the 1.5mm margin provides forgiveness, but worth aligning with the convention used by `get_pad_position`. Recommend filing a follow-up issue rather than holding this PR (since the partition invariant is preserved either way).

### Decision

Approved. The PR delivers the geometric outline partition exactly as scoped by #2771: distinct outlines for shared-layer zones, full outline preserved for sole-layer zones (return-path planes), backward-compatible with existing single-zone cases, well-tested. The downstream board-05 acceptance criterion ("at least 4 zones with non-zero filled copper") additionally depends on #2770 (zone preservation through the router write-back), which is correctly noted as out-of-scope here.

PR #2783, opened 2026-05-12
fix(zones): use positive rotation sign in _net_pad_positions_absolute (closes #2778)

### LINKED ISSUE #2778: zones: _net_pad_positions_absolute uses inverted rotation sign vs PCB.get_pad_position
## Symptom

\`_net_pad_positions_absolute\` (added by PR #2777, in \`src/kicad_tools/zones/generator.py\`) uses \`math.radians(-fp.rotation)\` when computing absolute pad positions, while the canonical \`PCB.get_pad_position\` (and every other absolute-pad-position calculation in the codebase) uses positive \`math.radians(fp.rotation)\`.

Discovered as a non-blocking follow-up nit during the PR #2777 judge review (https://github.com/rjwalters/kicad-tools/pull/2777#issuecomment-4428974043).

## Impact

- Test fixtures in \`TestAutoPourGeometricOutlines\` all use \`rotation=0\`, so the sign disagreement is masked and tests pass.
- For real rotated footprints, the computed bounding box may be **mirrored** relative to actual pad locations.
- The partition invariant (disjoint bboxes across nets) is still preserved — overlapping is unaffected because the same rotation sign is applied consistently within \`_compute_pour_outlines\`.
- However, the bbox may not cover the pads it's supposed to cover when the footprint is rotated.

## Reproduction

Construct a fixture with a footprint rotated 90°. Compare:
- \`_net_pad_positions_absolute(footprint)\` output
- \`PCB.get_pad_position(footprint, pad)\` output

The pad locations should agree but currently differ in sign of the rotation transform.

## Root cause

Convention drift between two implementations of the same geometric operation. The codebase consensus is positive \`math.radians(fp.rotation)\` (KiCad pad-position convention).

## Fix

Change \`math.radians(-fp.rotation)\` → \`math.radians(fp.rotation)\` in \`_net_pad_positions_absolute\`.

## Acceptance criteria

- [ ] \`_net_pad_positions_absolute\` and \`PCB.get_pad_position\` agree byte-for-byte on rotated footprints (fixture: rotation=0/90/180/270, also non-axis-aligned like 45°).
- [ ] Drift-prevention test in \`tests/test_zone_generator.py\`: parametrize a rotated-footprint fixture, assert \`_net_pad_positions_absolute\` output matches \`PCB.get_pad_position\` per pad.
- [ ] No regression in \`TestAutoPourGeometricOutlines\` (the existing rotation=0 cases still pass).
- [ ] If any board in \`boards/\` has rotated footprints with pour nets, regenerate and verify zone outlines still partition correctly.

## Files

- \`src/kicad_tools/zones/generator.py\` — \`_net_pad_positions_absolute\`
- \`src/kicad_tools/schema/pcb.py\` — \`PCB.get_pad_position\` (canonical reference, read-only)
- \`tests/test_zone_generator.py\` — add drift-prevention test

## Related

- PR #2777 (introduced the bug)
- #2771 (parent issue)

rjwalters:
## Summary

- One-character fix at `src/kicad_tools/zones/generator.py:978`: change `math.radians(-fp.rotation)` to `math.radians(fp.rotation)` so `_net_pad_positions_absolute` matches the canonical `PCB.get_pad_position` (and every other local->world pad transform in the codebase).
- New `TestNetPadPositionsRotationConvention` in `tests/test_build_zones_step.py` parametrizes rotations {0, 45, 90, 180, 270} and asserts agreement with `PCB.get_pad_position` to within `1e-9`. A negative-control assertion documents that the 90° fixture actually discriminates between the buggy and fixed transforms (pure 0/180 cases are sign-symmetric and pass even with the bug — which is how the original drift escaped review).

Closes #2778.

## Background

The bug was introduced by PR #2777 and flagged as a non-blocking nit during the judge review. Existing `TestAutoPourGeometricOutlines` fixtures all use `rotation=0` so the sign disagreement was masked. For real rotated footprints, the computed per-net bbox could miss pads it was supposed to cover, causing the zone fill on that layer to omit those pads. The partition invariant (disjoint bboxes across nets) was unaffected because the bug was internally consistent.

## Test Plan

- [x] New parametrized regression test passes against the fix (6 cases including the negative control): `pytest tests/test_build_zones_step.py::TestNetPadPositionsRotationConvention -v` -> 6 passed.
- [x] **Verified the test catches the bug**: temporarily reverted the source change and re-ran -> 3 failed (45/90/270), 3 passed (0/180/negative-control), exactly as the sign-symmetry analysis predicts.
- [x] Existing `TestAutoPourGeometricOutlines` unaffected: 5 passed.
- [x] Full zones test surface clean: `pytest tests/test_build_zones_step.py tests/test_zone_generator.py tests/test_zones.py` -> 143 passed.
- [x] No new lint regressions: `ruff check src/kicad_tools/zones/generator.py tests/test_build_zones_step.py` -> all checks passed.
- [x] No new mypy errors on the touched source (pre-existing missing-stubs warning for shapely is unrelated).

## Board regeneration note

Only `boards/04-stm32-devboard` currently has rotated footprints in its routed PCB output (68 rotated `(at ...)` entries; other boards use 0° everywhere). The bug consequence on board 04 — if any rotated footprint was on a pour net — would have been a per-net bbox that misses some pads. Regenerating board 04 zones is recommended in a follow-up but not strictly required to land this fix (the unit test guarantees the geometric correctness going forward).

## Hazard checked

`_net_pad_positions_absolute` adds `pcb.board_origin` to its result while `PCB.get_pad_position` does not. The fixture uses `(gr_rect (start 0 0) ...)` so `board_origin == (0, 0)` after `_detect_board_origin` runs, and the test explicitly asserts this fixture invariant to prevent silent frame-mismatch from passing as success.

### DIFF
--- a/src/kicad_tools/zones/generator.py
+++ b/src/kicad_tools/zones/generator.py
@@ -975,7 +975,10 @@ def _net_pad_positions_absolute(
             continue
 
         fp_x, fp_y = fp.position
-        rot_rad = math.radians(-fp.rotation)
+        # NOTE: positive sign matches PCB.get_pad_position (the canonical
+        # local->world transform used throughout the codebase).  See #2778
+        # for the previous drift bug where a negative sign was used here.
+        rot_rad = math.radians(fp.rotation)
         cos_r, sin_r = math.cos(rot_rad), math.sin(rot_rad)
 
         for pad in fp.pads:
--- a/tests/test_build_zones_step.py
+++ b/tests/test_build_zones_step.py
@@ -1148,3 +1148,170 @@ class TestAutoPourGeometricOutlines:
         for net in ("VMOTOR", "+5V", "+3.3V", "PWR_LED"):
             assert outlines[net] is not None, f"{net} should have a per-net outline"
             assert len(outlines[net]) >= 3, f"{net} polygon must have >=3 vertices"
+
+
+# ---------------------------------------------------------------------------
+# Rotation-sign drift-prevention (issue #2778)
+# ---------------------------------------------------------------------------
+
+
+class TestNetPadPositionsRotationConvention:
+    """Regression guard for #2778.
+
+    ``_net_pad_positions_absolute`` must use the same rotation-sign
+    convention as the canonical :meth:`PCB.get_pad_position` (positive
+    ``math.radians(fp.rotation)``).  A previous implementation used the
+    negative sign, which silently produced mirrored pad positions for
+    rotated footprints.  Existing fixtures in
+    :class:`TestAutoPourGeometricOutlines` all use ``rotation=0`` and
+    therefore did not catch the drift.
+
+    These tests parametrize a rotation set that includes 45/90/270 --
+    angles where the sign of ``math.radians(...)`` actually matters
+    (0 and 180 alone would be insufficient because the rotation matrix
+    is sign-symmetric there).
+    """
+
+    @staticmethod
+    def _make_rotated_pcb(tmp_path: Path, rotation: float) -> Path:
+        """Build a minimal one-footprint PCB at the given rotation.
+
+        The pad is placed at a non-origin offset ``(1, 0)`` so the
+        rotation transform is visible in the resulting absolute pad
+        position.  The board outline starts at ``(0, 0)`` which means
+        ``PCB.board_origin`` will be ``(0, 0)`` after detection -- this
+        is critical because :func:`_net_pad_positions_absolute` adds
+        ``board_origin`` to its output while :meth:`PCB.get_pad_position`
+        does not, so they only agree directly when ``board_origin`` is
+        the origin.  See the test below for the explicit assertion.
+        """
+        pcb_text = f"""\
+(kicad_pcb
+  (version 20240108)
+  (generator "kicad")
+  (general (thickness 1.6))
+  (layers
+    (0 "F.Cu" signal)
+    (31 "B.Cu" signal)
+    (44 "Edge.Cuts" user)
+  )
+  (net 0 "")
+  (net 1 "SIG")
+  (gr_rect
+    (start 0 0)
+    (end 100 50)
+    (stroke (width 0.15) (type solid))
+    (fill none)
+    (layer "Edge.Cuts")
+    (uuid "edge-uuid")
+  )
+  (footprint "Test:U1"
+    (layer "F.Cu")
+    (at 20 30 {rotation})
+    (uuid "fp-u1-uuid")
+    (property "Reference" "U1"
+      (at 0 -2 0)
+      (layer "F.SilkS")
+      (uuid "prop-u1-ref-uuid")
+      (effects (font (size 1 1) (thickness 0.15)))
+    )
+    (property "Value" "T"
+      (at 0 2 0)
+      (layer "F.Fab")
+      (uuid "prop-u1-val-uuid")
+      (effects (font (size 1 1) (thickness 0.15)))
+    )
+    (pad "1" smd rect (at 1 0) (size 1 1) (layers "F.Cu") (net 1 "SIG"))
+  )
+)
+"""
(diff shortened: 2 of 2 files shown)


rjwalters:
## Judge Review: Approved

Surgical one-character fix at `src/kicad_tools/zones/generator.py:978` with a well-designed regression test. Verified against every specific concern.

### Sign convention verification

The positive-sign convention is dominant in the codebase and explicitly documented:
- `src/kicad_tools/schema/pcb.py:3628` — canonical `PCB.get_pad_position` uses `math.radians(fp.rotation)`
- `src/kicad_tools/router/io.py:2662-2667` — explicit comment "KiCad stores rotation in degrees, positive = counter-clockwise / Standard 2D rotation matrix applies directly (no negation needed)"
- Other positive-sign sites: `drc/cpp_backend.py:155-156`, `drc/incremental.py:585-586,742-745`, `validate/rules/netlist.py:107`, `validate/rules/via_in_pad.py:64`, `validate/rules/clearance.py:118`, `mcp/tools/analysis.py:1019`, `optim/keepout.py:245`

The fix aligns `_net_pad_positions_absolute` with `PCB.get_pad_position`. Confirmed.

### Test discrimination verified empirically

Temporarily reverted `src/kicad_tools/zones/generator.py:978` back to `math.radians(-fp.rotation)` and re-ran the parametrized test:
- `test_agrees_with_get_pad_position[45.0]` FAILED (x mismatch)
- `test_agrees_with_get_pad_position[90.0]` FAILED (y: 31.0 vs 29.0)
- `test_agrees_with_get_pad_position[270.0]` FAILED (y: 31.0 vs 29.0)
- `[0.0]` and `[180.0]` PASSED (sign-symmetric, as documented in the docstring)
- `test_negative_control_buggy_sign_disagrees_at_90deg` PASSED (asserts fixture discriminating power)

Exactly the failure pattern the builder claimed. The negative-control test correctly documents that 0/180 alone would be insufficient and that the 90° fixture has discriminating power. With the fix in place, all 6 cases pass.

### Hazard check (board_origin)

Confirmed the fixture uses `(gr_rect (start 0 0) ...)` and the test contains an explicit invariant assertion:
```python
assert pcb.board_origin == (0.0, 0.0), (
    f"fixture invariant broken: board_origin={pcb.board_origin} ..."
)
```
This isolates the rotation-sign behavior from the board-origin offset that `_net_pad_positions_absolute` applies but `PCB.get_pad_position` does not. Good defensive instinct.

### Local verification

- `pytest tests/test_build_zones_step.py::TestNetPadPositionsRotationConvention -v` — 6 passed
- `pytest tests/test_build_zones_step.py tests/test_zone_generator.py tests/test_zones.py` — 143 passed
- `ruff check src/kicad_tools/zones/generator.py tests/test_build_zones_step.py` — clean
- `mypy src/kicad_tools/zones/generator.py` — only pre-existing shapely missing-stubs warning at line 325 (unrelated to line 981)
- `gh pr checks 2783` — all 8 checks pass; `mergeStateStatus` = CLEAN

### Comment quality

The inline NOTE at lines 978-981 references the canonical implementation and issue #2778 in 3 lines. Appropriate weight for a sign-flip with prior regression history.

### Out-of-scope observation (not blocking)

While auditing call sites, I noticed two other locations still use the negative-sign convention and appear to have the same bug (forward transform of pad local coords to world):
- `src/kicad_tools/validate/rules/edge.py:196` — has a misleading "KiCad uses clockwise-positive rotation" comment that contradicts `router/io.py:2662`
- `src/kicad_tools/mcp/tools/routing.py:630`

These are outside the scope of issue #2778 (which specifically calls out the zones path) but a follow-up issue to audit and unify the convention codebase-wide would prevent future drift. Not blocking this PR.

### Verdict

APPROVE. Minimal, correct, with a regression test that demonstrably catches the original bug and documents its own discriminating power.

Issue #2795, opened 2026-05-12
router: silent multi-minute stalls during BLOCKED_BY_COMPONENT negotiated rip-up


rjwalters:
## Summary

During multi-net BLOCKED_BY_COMPONENT rip-up, the router can sit silently for 10+ wall-clock minutes between log lines. There is no per-sibling progress, no \"still working\" heartbeat, and no indication of how much work remains in the rip-up phase. From an operator (or agent) perspective, the router looks hung.

## Reproducer

Routing chorus-test-revA v20 (4-layer, jlcpcb-tier1, --timeout 1500, --iterations 50, --auto-fix --auto-fix-passes 2):

```
[ 98.0%] Routing BOOT0... (584.3s)
Initial pass: 35/51 nets, overflow: 34
Initial pass stall (overflow=34): 19 net(s) unrouted -- engaging BLOCKED_BY_COMPONENT rip-up (584.8s)
BLOCKED_BY_COMPONENT rip-up (negotiated) for I2S_LRCLK: displacing 11 sibling(s) on J2, U5: ...
<<<< 10+ minutes of silence here >>>>
BLOCKED_BY_COMPONENT rip-up (negotiated) for I2S_LRCLK: reroute did not converge
BLOCKED_BY_COMPONENT rip-up (negotiated) for I2S_DIN: displacing 7 sibling(s) on J2, U5: ...
<<<< more silence >>>>
```

Wall-clock between \"displacing N sibling(s)\" and \"reroute did not converge\": measured at ~10 minutes for I2S_LRCLK (11 siblings). The python process is at 98% CPU the entire time — it is genuinely working, just not reporting.

## Root cause

`src/kicad_tools/router/core.py:3615` prints \"displacing N sibling(s)\" via `flush_print` at rip-up entry, then calls `neg_router.targeted_ripup(...)` (line 3640) which iterates internally over siblings and per-net A* re-search at `per_net_timeout=30` seconds per attempt (default). For a rip-up that displaces 11 siblings and then re-routes them plus the failed net, worst case is 12 * 30s = 360s with no log output until line 3671 (\"reroute did not converge\"). In practice we observed substantially longer (>600s for I2S_LRCLK).

`targeted_ripup` apparently has no progress callback to surface \"now trying sibling 4/11\" or \"now re-routing failed net after X siblings displaced\".

## Expected behavior

At minimum, emit a heartbeat every N seconds during `targeted_ripup` showing:

[...]

PR #2775, opened 2026-05-12
fix(router): forward per_net_timeout through route_with_escape

rjwalters:
## Summary

`route_with_escape()` silently dropped the CLI's `--per-net-timeout` value because the function did not accept the parameter. On dense boards (e.g. board 05 BLDC: U3 DRV8301 HTSSOP-56 + U10 STM32G431) this let single nets consume the entire board-level budget — PHASE_A burned 159s of a 240s budget on 6/35 nets before aborting. With the fix, the per-net cap reaches the A* pathfinder via the existing `route_all_two_phase` -> `route_all_negotiated` -> C++ pathfinder plumbing, bounding any single net's wall-clock to `--per-net-timeout` (default 30s).

This is the canonical fix already used for the sibling two-phase call sites at `route_cmd.py:1647-1653, 2321-2327, 2840-2847, 4940-4946`; the escape branch was simply omitted.

## Changes

- `src/kicad_tools/router/core.py:8285-8295` — `route_with_escape` signature now accepts `per_net_timeout: float | None = None`; docstring updated.
- `src/kicad_tools/router/core.py:8351-8357` — `per_net_timeout` forwarded into `route_all_two_phase(...)` on the `use_negotiated=True` branch (the only branch exercised by the CLI default `--strategy negotiated`).
- `src/kicad_tools/cli/route_cmd.py:1636-1640, 2310-2315, 2829-2835, 4919-4926` — all 4 escape call sites now pass `per_net_timeout=getattr(args, "per_net_timeout", None) or None`, matching the existing two-phase pattern.
- `tests/test_global_router.py` — 2 new unit tests in `TestEscapeRouteRipupEligibility`:
  - `test_route_with_escape_forwards_per_net_timeout`: patches `route_all_two_phase`, calls `route_with_escape(per_net_timeout=42.0)`, asserts `call_kwargs["per_net_timeout"] == 42.0`.
  - `test_route_with_escape_defaults_per_net_timeout_to_none`: backwards-compat assertion that omitting the kwarg yields `per_net_timeout=None`.

Out of scope (per curator): the `use_negotiated=False` branch (`core.py:8358-8360`) calls `route_all(...)` which doesn't yet accept `per_net_timeout`; left untouched since CLI default is `strategy=negotiated`. A follow-up can plumb `route_all` if a non-negotiated escape workflow appears.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
[...]

PR #2797, opened 2026-05-12
fix(build): stop kct build hang on board 05 via timeouts + stdout streaming

### LINKED ISSUE #2794: kct build: design.py hangs >21min in numpy ufunc loop (board 05; blocks #2746 full-pipeline validation)
## Symptom

\`kct build -f boards/05-bldc-motor-controller\` (full pipeline, with the board's \`design.py\`) hangs indefinitely. Auditor measured 22 minutes wall-clock with no completion or progress output, at which point \`sample\` showed a tight numpy ufunc loop with no escape:

\`\`\`
2503 _PyEval_EvalFrameDefault
+ 151 PyNumber_Add
+   array_add (in _multiarray_umath.cpython-314-darwin.so)
+     PyObject_CallFunctionObjArgs → ufunc_generic_fastcall → PyUFunc_GenericFunctionInternal
+       PyArray_NewFromDescr_int → PyDataMem_UserNEW (lots of small numpy array allocations)
\`\`\`

Full stack sample: \`/tmp/audit_board05_post_rotation/sample.txt\` (kept as artifact).

The hang occurs **after** \`kct route\` completes successfully (route-only on the same board produces 17/35 nets in 49% completion in ~4 min). So the route step itself is healthy; the hang is in the design/optimization phase of \`kct build\`.

## Reproduction

\`\`\`
cd /Users/rwalters/GitHub/kicad-tools
uv run kct build -f boards/05-bldc-motor-controller
# wait > 21 min — no output, no progress, no timeout
# sample the python pid: numpy ufunc loop dominates
\`\`\`

## Suspected location

Two candidates for the hanging \`design.py\`:
- \`boards/05-bldc-motor-controller/design.py\` — board's design script invoked by \`kct build\`
- \`src/kicad_tools/mcp/tools/design.py\` — library helper

The stack signature (small numpy array adds in a tight loop) is consistent with a fitness/scoring inner loop over many candidates without vectorization, or a missing early-termination condition.

## Why this matters

This blocks the end-to-end validation of issue **#2746** (Board 05 BLDC regression). The six children of #2746 have all shipped, and \`kct route\` alone shows substantial recovery (5.7% → 49% yield), but full-pipeline validation including zone fills and DRC re-validation cannot be performed until \`kct build\` completes.

## Acceptance criteria

1. \`kct build -f boards/05-bldc-motor-controller\` either completes within a reasonable wall-clock (suggest ≤10 min default, configurable via \`--timeout\`), or fails loudly with a clear progress indicator.
2. No silent hang — at minimum, periodic progress logs so the loop's stage is observable.
3. The numpy ufunc inner loop is either vectorized, bounded, or has an early-termination criterion.
4. Add a smoke test in CI that \`kct build\` on at least one example board completes within a budget.

## Adjacent observations

- \`kct route\` (route-only) on the same board completes in ~4 min with reasonable progress logs — that's the right reference behavior.
- The hang is reproducible: stash all working changes, restore them, hang reproduces on the same commit (\`5c8b511a\`).

## Suggested investigation

1. \`uv run --no-sync\` with \`pdb\` or \`-X faulthandler\` to capture Python-level frames at the hang point.
2. \`py-spy dump --pid <pid>\` for clearer Python-frame attribution.
3. Inspect both candidate \`design.py\` files for unbounded numpy loops in placement/fitness/scoring sections.

## Discovered by

Auditor verifying #2746 post-rotation-cleanup on 2026-05-12. Comment: https://github.com/rjwalters/kicad-tools/issues/2746#issuecomment-4433357796

rjwalters:
## Summary

Closes #2794.

Three coordinated changes that together prevent silent multi-hour hangs on dense boards under `kct build`:

1. **Board 05 call-site fix** (`boards/05-bldc-motor-controller/design.py:1869`)
   Replace bare `router.route_all()` with `router.route_all_negotiated(per_net_timeout=30.0, timeout=240.0, progress_callback=...)`. Mirrors the `kct route` CLI defaults (`route_cmd.py:1702-1710`). The negotiated path has the per-net bracketing that #2775/#2779 fixed; the basic `route_all` path does not. Curator reproduced a 22-minute hang here with one net emitted.

2. **Library defense** (`src/kicad_tools/router/core.py`, `Autorouter.route_all`)
   Add `timeout` and `per_net_timeout` kwargs plus a `suppress_no_timeout_warning` opt-out. Emit a `UserWarning` pointing to #2794 when none are supplied so future board authors cannot rediscover the trap silently. Outer wall-clock budget breaks the per-net loop when exceeded. Internal callers in core.py opt out via the new flag to avoid noise.

3. **Stdout streaming** (`src/kicad_tools/cli/build_cmd.py:_run_python_script`)
   Replace `subprocess.run(..., capture_output=not verbose)` with `Popen` + line-buffered `readline` loop so child stdout reaches the parent in real time -- without `--verbose`. This is what originally caused the issue to be misdiagnosed as a "silent hang": the prints existed, just buffered. A new `quiet` kwarg keeps the captured-silent path for `--quiet` callers. Sets `PYTHONUNBUFFERED=1` in the child env (idempotent via `setdefault`).

### Misdiagnosis correction

The auditor's "numpy ufunc loop" stack is **A\* heap-key churn in the pathfinder** (`heappop`/`siftup` over numpy-typed tuple keys), not the `TraceOptimizer` at `design.py:1824` (which is never reached because routing never completes). Same root cause class as #2775/#2779 but in the un-bracketed basic-routing path.

## Test plan

- [x] `tests/test_router_core.py::TestAutorouterRouteAll` — 8 tests including 5 new ones:
  - `test_route_all_warns_when_no_timeout` — regression-prevention warning fires
  - `test_route_all_no_warning_when_timeout_supplied` — explicit `timeout=` silences
  - `test_route_all_no_warning_when_per_net_timeout_supplied` — explicit `per_net_timeout=` silences
  - `test_route_all_no_warning_when_opt_out` — explicit `suppress_no_timeout_warning=True` silences
  - `test_route_all_accepts_timeout_kwarg` — kwarg plumbing works end-to-end
- [x] `tests/test_build_cmd_stdout_streaming.py` (new file, 8 tests):
  - `TestStdoutStreamingDefault` — child stdout forwarded line-by-line in default mode
  - `TestStdoutStreamingQuiet` — `quiet=True` suppresses live stdout
  - `TestStdoutStreamingFailure` — non-zero exit reports `success=False` with stderr
  - `TestStdoutStreamingChildEnv` — `PYTHONUNBUFFERED=1` set; caller's value preserved
  - `TestRouteAllSmokeBudget::test_route_all_with_outer_timeout_returns_within_budget` — outer-timeout plumbing
- [x] Verified pre-existing test failures (`test_add_component_rotation`, `test_subgrid_integration::TestSubgridPrepass`, `test_route_exit_code_3_is_failure`, `test_voltage_divider_high_completion`, `test_charlieplex_high_completion`) reproduce on `main` without these changes — not regressions.
- [ ] Acceptance criterion 1 (end-to-end): `kct build -f boards/05-bldc-motor-controller` completes within `timeout=240.0` or fails loudly (requires running on a machine with the full toolchain; covered by the per-net + outer cap in the new code paths).

## Conflict notes

Per the Curator brief, sibling builder for #2795 touches `src/kicad_tools/router/core.py` around line ~3516 (outer rip-up caller). This PR touches `route_all` (`3203-3310`) and a handful of internal `self.route_all(...)` call sites further down. Different functions, but same file -- expect simple rebase if any conflict at merge time.

### DIFF
--- a/boards/05-bldc-motor-controller/design.py
+++ b/boards/05-bldc-motor-controller/design.py
@@ -1865,8 +1865,39 @@ def route_pcb(input_path: Path, output_path: Path) -> bool:
     print(f"   Skipping power nets: {skip_nets}")
 
     # Route all nets
+    #
+    # Issue #2794: Use ``route_all_negotiated`` with explicit per-net and
+    # outer timeouts so a single pathological net cannot stall the build
+    # indefinitely.  The bare ``router.route_all()`` call (no timeouts,
+    # no progress callback) previously hung this board's ``kct build``
+    # for 22+ minutes inside A* heap-key churn -- with one net emitted
+    # and zero feedback after that point.
+    #
+    # The values below mirror the ``kct route`` CLI's defaults
+    # (``src/kicad_tools/cli/route_cmd.py:1702-1710``):
+    #
+    #   - ``per_net_timeout=30.0``  -- per-net A* deadline (#2775/#2779
+    #     bracketed this across the whole net rather than per RSMT edge,
+    #     so 30 s is now a reliable cap)
+    #   - ``timeout=240.0``         -- outer wall-clock budget for the
+    #     entire negotiated loop; on the off-chance every net hits its
+    #     timeout, total stays bounded
+    #   - ``progress_callback``     -- one-line per-net status so
+    #     ``kct build`` (and humans) see progress without ``--verbose``
     print("\n2. Routing nets...")
-    router.route_all()
+
+    def _route_progress(progress: float, message: str, _ok: bool) -> bool:
+        # Negotiated callback fires once per net plus a final summary;
+        # echo to stdout so build_cmd's streaming stdout (Issue #2794
+        # fix) surfaces it to the parent process.
+        print(f"   [{progress * 100:5.1f}%] {message}", flush=True)
+        return True
+
+    router.route_all_negotiated(
+        per_net_timeout=30.0,
+        timeout=240.0,
+        progress_callback=_route_progress,
+    )
 
     # Get statistics before optimization
     stats_before = router.get_statistics()
--- a/src/kicad_tools/cli/build_cmd.py
+++ b/src/kicad_tools/cli/build_cmd.py
@@ -244,18 +244,39 @@ def _run_python_script(
     verbose: bool = False,
     env_vars: dict[str, str] | None = None,
     script_args: list[str] | None = None,
+    quiet: bool = False,
 ) -> tuple[bool, str]:
     """Run a Python generator script.
 
     Args:
         script_path: Path to the Python script
         cwd: Working directory for execution
-        verbose: Whether to show script output
+        verbose: Reserved for backward compatibility -- previously gated
+            *all* stdout streaming behind ``--verbose``.  Issue #2794
+            inverts the default: stdout is now streamed line-by-line at
+            the default verbosity so users see board progress (per-net
+            routing, optimization steps, etc.) without needing
+            ``--verbose``.  Setting ``verbose=True`` is still honoured
+            for ``stderr`` echo.
         env_vars: Optional additional environment variables to pass
         script_args: Optional positional arguments to pass to the script
+        quiet: When True, suppress live stdout streaming (capture
+            silently and only surface errors).  Used by ``--quiet`` /
+            CI paths that don't want per-net chatter.
 
     Returns:
         Tuple of (success, output/error message)
+
+    Notes:
+        Issue #2794: ``subprocess.run(..., capture_output=not verbose)``
+        previously buffered the entire child stdout until the script
+        completed.  For long-running routing scripts (board 05 BLDC
+        controller, etc.) this meant a 22-minute silent hang was
+        indistinguishable from a 22-second silent wait -- the prints
+        existed, but were trapped in the buffer.  We now use
+        ``Popen`` with line-buffered stdout and stream each line to
+        the parent process as it arrives, so progress is observable
+        in real time.
     """
     import os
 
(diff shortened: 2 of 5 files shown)


rjwalters:
## Judge review: APPROVED

Reviewed the 3-part bundled fix end-to-end with a live smoke run of `kct build -f boards/05-bldc-motor-controller`.

### CI

- Lint, Type Check, C++ Build, Routed PCB DRC, kicad-cli Smoke, Diff-Pair Routing: PASS
- Test, Match-Group Routing: still IN_PROGRESS at review time; the new 16 tests pass locally (`tests/test_router_core.py::TestAutorouterRouteAll` + `tests/test_build_cmd_stdout_streaming.py`).
- Pre-existing failures (e.g. `test_subgrid_integration::TestSubgridPrepass`) reproduce on `main` without this PR -- spot-checked one, confirmed not a regression.
- `mergeStateStatus` is `UNSTABLE` (in-progress checks) -- merger should wait for green before landing.

### Live smoke (the critical acceptance criterion)

Ran the actual `uv run kct build -f boards/05-bldc-motor-controller`. Results:

- Default verbosity: per-net progress streams to stdout immediately (PR's stdout-streaming fix delivers).
- Iteration 0 (initial routing pass): completed 32/32 nets in 127.9s, well inside the 240s outer budget.
- Iteration 1 rip-up: actively progressing at 187s elapsed when I checked.
- No silent hang. The 22-minute black hole reported in #2794 is gone.

The negotiated path with `per_net_timeout=30` + `timeout=240` does what it says on the tin.

### Library defense (`route_all` signature)

- Defaults (`timeout=None`, `per_net_timeout=None`, `suppress_no_timeout_warning=False`) preserve existing call semantics for callers who don't update -- they just get a `UserWarning` pointing at #2794 instead of a silent hang.
- Outer wall-clock check happens BETWEEN nets (top of each iteration, guarded `i > 0` so first net always gets a full try). The in-flight net is never interrupted mid-A*, so partial state is clean.
- Warning text is actionable: names the alternative (`route_all_negotiated`), the opt-out flag, and the issue number.

### Internal suppressions audit (6 sites)

I traced all 6 internal `self.route_all(..., suppress_no_timeout_warning=True)` calls. Three categories:

**Correct (defensive suppression, timeout forwarded):**
- `route_with_subgrid` (8434, 8675) -- forwards both `timeout` and `per_net_timeout`, defensively suppresses because they may be `None`. Fine.
- `route_with_adaptive_grid` (8750) -- forwards `timeout`. Fine.

**Adjacent silent-hang traps (not regressed by this PR, but not fixed either):**
- `route_all_tuned` (4098, 4133, 4155) -- the wrapper itself has NO timeout parameter, so callers of `route_all_tuned()` still hit the same A* heap-key churn trap that #2794 was filed for. Suppression here just hides the warning rather than plumbing timeout through. Pre-existing behavior; not introduced by this PR.
- `route_all_block_aware` (6660) flat-fallback path -- same shape, no timeout param.
- `route_with_escape` (6923) plain-route path -- same shape, no timeout param.

**Small new defect:**
- `route_all_multi_resolution` (8845) -- the wrapper DOES have a `timeout` parameter (line 8793), and the negotiated branch forwards it (line 8840). The non-negotiated `else` branch added in this PR suppresses the warning but does NOT forward the available `timeout` to `route_all`. The fix is a one-line addition: `timeout=timeout`. Filing as follow-up rather than blocking, since the non-negotiated path is rarely used and the suppression at least makes the gap explicit.

Followup recommendation: a separate issue to either (a) thread `timeout` through `route_all_tuned`/`route_all_block_aware`/`route_with_escape` or (b) at minimum drop the warning suppression in their bare-call paths so users discover the trap.

### Build_cmd streaming

- `PYTHONUNBUFFERED=1` via `setdefault` is correct: caller's value (including unusual ones like `"0"` or `"x"`) wins. Covered by `test_pythonunbuffered_does_not_clobber_caller`.
- `quiet` kwarg is routed solely through `ctx.quiet` at 3 call sites -- only fires via the existing `--quiet` CLI flag. No surprise verbosity in CI.
- Popen + `readline` + `bufsize=1` + `flush()` on each line: child progress visible in real time. Verified live in the smoke run above.

### Files

5 files as expected: `boards/05-bldc-motor-controller/design.py`, `src/kicad_tools/cli/build_cmd.py`, `src/kicad_tools/router/core.py`, `tests/test_build_cmd_stdout_streaming.py`, `tests/test_router_core.py`. No drift.

### Conflict with #2795

#2795 touches `router/core.py` ~3516 (rip-up caller); this PR touches `route_all` at 3203 and internal call sites further down. Likely clean merge; whichever lands second may need a trivial rebase. Not blocking.

### Verdict

Approving. The board 05 hang is conclusively fixed (verified live), the library-level defense is well-designed, the stdout streaming is correctly implemented and tested, and the suppression audit reveals only adjacent pre-existing traps -- not regressions. The one new gap (`route_all_multi_resolution` non-negotiated `timeout` drop) is small enough to handle as a follow-up issue.

Issue #2800, opened 2026-05-12
router: route_all_multi_resolution drops timeout on non-negotiated branch (follow-up to #2794)


rjwalters:
Follow-up identified during Judge review of PR #2797.

## Defect

\`src/kicad_tools/router/core.py:8845\` — \`route_all_multi_resolution\` receives a \`timeout\` parameter but **drops it on the non-negotiated branch**. The negotiated branch correctly forwards it; the fallback path silently discards it, re-introducing the same class of hang #2794 was meant to fix.

## Why this is small

This is a one-line follow-up — pass the \`timeout\` through to the non-negotiated branch's downstream call.

## Adjacent observations (from #2794 Judge review)

Same review identified **3 other internal callers** of \`route_all()\` whose own wrappers have no \`timeout\` parameter to forward — they suppress the new warning (\`suppress_no_timeout_warning=True\`) but cannot defend their callers against a hang:
- \`route_all_tuned\`
- \`route_all_block_aware\`
- \`route_with_escape\`

These hide pre-existing silent-hang traps but don't introduce new ones. They warrant their own scoping work — likely add \`timeout\` to their public APIs, then thread it through, then remove the \`suppress\` flag. Could be one issue per wrapper or a single sweep — propose either.

## Acceptance criteria

For \`route_all_multi_resolution\` specifically:
- Non-negotiated branch threads \`timeout\` to its downstream call
- Add a regression test that asserts \`route_all_multi_resolution(timeout=1.0)\` does not exceed the timeout on a hostile fixture

For the 3 other wrappers (if bundled):
- Add \`timeout\` parameter to each wrapper's signature
- Forward through to inner calls
- Remove \`suppress_no_timeout_warning=True\` flags
- Regression tests per wrapper

## Discovered by

PR #2797 Judge review: https://github.com/rjwalters/kicad-tools/pull/2797#issuecomment-4433830933

rjwalters:
## Curator: verified scope (3 of 4 wrappers need work; `route_with_escape` already correct)

Confirmed the defect at line 8876 (not 8845 — that's docstring) and surveyed the three other wrappers. **One of the three already handles timeout correctly**, so the effective defect count is **1 confirmed (#2800 primary) + 2 additional gaps** (not 4).

### 1. `route_all_multi_resolution` — primary defect (CONFIRMED)

`src/kicad_tools/router/core.py:8817-8876`

- L8824: `timeout: float | None = None` parameter exists.
[...]

PR #2805, opened 2026-05-12
fix(router): preserve best iteration by (routed, overflow) lex tuple (closes #2803)

### LINKED ISSUE #2803: router: negotiated iteration can regress overflow (16→36 observed); no best-so-far preservation
## Symptom

During a chorus-test routing run today, the negotiated-routing iteration loop regressed overall congestion:

- **Iteration 0**: 30/51 nets routed, **overflow = 16**
- **Iteration 1**: makes some net rip-up moves, **overflow climbs to 36**
- Subsequent iterations recover toward 19, but the router does not appear to preserve and return the iteration-0 result if later iterations are strictly worse

This means a longer router run can produce a *worse* output than a shorter one, which is counterintuitive and makes routing budgets hard to tune.

## Reproducer

The chorus-test v20 4-layer route at `/Users/rwalters/GitHub/chorus/hardware/chorus-test-revA/kicad/chorus-test-revA_v20.kicad_pcb` exhibits this pattern when run with `--placement-feedback --iterations 50`. Specific log fragment captured today shows the 16→36 overflow regression.

## Why this matters

For users tuning routing experiments (e.g., chorus-test toward ≥95% completion), the lack of "best-so-far" semantics by **overflow** means:
1. **Comparison runs are unreliable**: a 45-min run can produce a strictly worse PCB than a 20-min run on the same input
2. **Budget tuning is fragile**: increasing `--iterations` can hurt rather than help
3. **CI gates are noisy**: regression tests that assert "completion ≥ N%" can flake if the iterator happens to time out in a bad iteration

This is distinct from per-net rip-up convergence (#2795 / #2800). This is an *outer-loop* iteration accounting question.

---

### Root cause: existing best-of-iterations logic only compares routed-count, not overflow

The negotiated iteration loop lives in `src/kicad_tools/router/core.py` `Autorouter.route_all_negotiated` (definition `core.py:4195`, outer loop `core.py:4703`).

**Issue #2540 already added best-of-iterations snapshotting**, but its comparison metric is incomplete:

```
core.py:4696-4699  # snapshot at function entry (iteration 0 state)
    best_routes: list[Route] = copy.deepcopy(list(self.routes))
    best_net_routes: dict[int, list[Route]] = copy.deepcopy(net_routes)
    best_routed_count = sum(1 for r in net_routes.values() if r)
    best_iteration = 0

core.py:4717-4722  # snapshot at TOP of each iteration N (after iter N-1 settled)
    current_routed = sum(1 for r in net_routes.values() if r)
    if current_routed > best_routed_count:
        best_routed_count = current_routed
        best_iteration = iteration - 1

core.py:5912-5930  # post-loop restore
    current_routed = sum(1 for r in net_routes.values() if r)
    if best_routed_count > current_routed:
        # ... unmark current routes, restore best_routes ...
```

**The gap**: the comparison is `current_routed > best_routed_count` only. Issue #2803's failure mode is `overflow climbs 16 → 36` with **routed_count unchanged or even higher**, so the existing guard never triggers. The post-loop restore at `core.py:5912` is a strict route-count comparison.

Two `overflow_history.append(overflow)` sites collect overflow at iteration end:
- `core.py:5368` (targeted-ripup branch end-of-iter)
- `core.py:5823` (standard branch end-of-iter)

…but neither cross-references the value against a "best overflow" snapshot for state preservation. `perturbation_best_overflow` (`core.py:4366`, `core.py:5371`, `core.py:5826`) tracks overflow but only to **reset the perturbation magnitude knob**, not to preserve board state.

### Why the existing fix is insufficient

The Issue #2540 fix snapshots state at the **top** of each iteration (before destructive rip-up) and restores only on route-count regression. It correctly handles the "mid-rip-up timeout" failure mode it was designed for. It does NOT handle "iteration completed but produced strictly worse PCB" because:
1. The metric ignores overflow entirely.
2. The snapshot is taken at iteration-top, so if iteration 1 finishes worse than iteration 0, the iter-1-top snapshot equals the iter-0-end state — which is **good**, but the post-loop comparison is route-count-only so the bad iter-1-end state still wins when routed-count is equal.

#### 1. Extend the comparison metric to a tuple

Define the "best" comparison as a lexicographic tuple `(routed_count desc, overflow asc, iteration desc)`:
- **Primary**: more routes is always better.
- **Secondary**: with equal routed counts, lower overflow is better (this is what #2803 needs).
- **Tertiary tie-break**: prefer the later iteration (gives perturbation/escape a chance to bake in).

A small `IterationResult` dataclass plus an `is_better_than(self, other) -> bool` method keeps it readable. Alternatively, a sort key tuple `(-routed, overflow, -iteration)` for `min()`-based selection.

#### 2. Capture state at iteration END, not just iteration TOP

Currently the snapshot is at iteration top (before rip-up). To preserve a strictly-better completed iteration, we also need a snapshot at iteration end after `overflow_history.append(overflow)` on both branches (after `core.py:5368` and `core.py:5823`).

```python
# proposed: at both overflow_history.append sites, also call:
self._maybe_capture_best_state(
    iteration=iteration,
    routed_count=sum(1 for r in net_routes.values() if r),
    overflow=overflow,
    net_routes=net_routes,
)
```

Where `_maybe_capture_best_state` only deep-copies if the new result strictly improves the tuple.

#### 3. Per-iteration log line for trajectory visibility

Add a single canonical log line after each iteration completes, regardless of branch:

```
iter 3 | routed=30/51 | overflow=19 | best-so-far=iter-0 (routed=30, overflow=16)
```

This makes the regression visible in the user-facing log and proves the best-state machinery is working.

#### 4. Post-loop restore (already wired) — extend the comparator

`core.py:5913` becomes:

```python
final = IterationResult(iteration, current_routed, overflow)
if best_result.is_better_than(final):
    flush_print(
        f"  Restoring iteration {best_result.iteration} state "
        f"(routed={best_result.routed_count}, overflow={best_result.overflow}) "
        f"instead of final (routed={final.routed_count}, overflow={final.overflow})"
    )
    # ... existing unmark/restore logic ...
```

### Snapshot strategy (memory cost discussion)

Three options, in increasing memory cost:

| Strategy | Memory cost | Implementation cost | Notes |
|---|---|---|---|
| **A. Metrics-only, re-route on rollback** | O(K) for K iterations | High — need a deterministic re-route path | Slow; complicates determinism if RNG state has advanced. **Reject.** |
| **B. Single rolling "best" snapshot** (recommended) | O(1) full snapshot at any time | Low — extends existing #2540 pattern | Mirrors current code. Replace the snapshot only on strict improvement. ~5-20 MB for chorus-test scale (50 nets, ~10-50 segments each). |
| **C. Per-iteration history of full snapshots** | O(K) full snapshots | Low | RAM-heavy: 50 iterations × ~10 MB = 500 MB. **Overkill** — we only ever restore the best, so we don't need every iteration. |

**Recommendation: Strategy B** — single rolling snapshot, replaced only on strict improvement by the lex-tuple metric. This is a minimal extension of the existing #2540 code path. The metric tuple (`routed`, `overflow`, `iteration`) is the only thing that needs per-iteration history — full snapshots can remain singleton.

For the optional log line trajectory we can keep a `list[IterationMetrics]` (no deep-copy, just three numbers per iteration), which is essentially free.

### File:line summary

| Concern | Location |
|---|---|
| Outer iteration loop | `src/kicad_tools/router/core.py:4703` (`for iteration in range(1, max_iterations + 1):`) |
| Existing #2540 snapshot at iter top | `core.py:4696-4722` |
| Overflow capture site (targeted branch) | `core.py:5368` |
| Overflow capture site (standard branch) | `core.py:5823` |
| Existing route-count post-loop restore | `core.py:5912-5930` |
| `Route`/`Segment` dataclasses (snapshot target) | `src/kicad_tools/router/primitives.py:128`, `:164` |
| Existing #2540 test pattern | `tests/test_route_all_negotiated_partial_state.py` |

## Acceptance criteria

1. `kct route` output is monotonic in `--iterations` by the tuple `(routed_count desc, overflow asc)`: increasing `--iterations` never produces a strictly worse PCB by this metric on the same input.
2. New regression test (`tests/test_route_all_negotiated_overflow_regression.py` or extend `test_route_all_negotiated_partial_state.py`):
   - Construct a fixture where iteration 0 produces overflow=N, iteration 1 produces overflow=M > N with equal routed counts, then time out before iteration 2 recovers.
   - Assert that the returned PCB has `overflow ≤ N` (i.e. iter-0 state was preserved).
   - Use the same `_FakeClock` / mock pattern as `test_route_all_negotiated_partial_state.py`.
3. New regression test covering `routed_count` tie-break: equal `routed_count` across two iterations with different overflow returns the lower-overflow one.
4. Per-iteration log line emitted on both branches: `iter N | routed=X/Z | overflow=Y` (plus optional `best-so-far=...` suffix). Verify in tests via `capsys`.
5. Existing #2540 tests in `test_route_all_negotiated_partial_state.py` continue to pass (mid-rip-up timeout case must still restore iter-N-1 by routed-count when overflow comparison is irrelevant).

## Test plan

- Unit: `IterationResult.is_better_than` covers all six orderings (better-routed, worse-routed, equal-routed-better-overflow, equal-routed-worse-overflow, identical, tie-break-by-iteration).
- Integration: extend `test_route_all_negotiated_partial_state.py` with the overflow-regression fixture above.
- Integration: existing chorus-test board re-run as a smoke check that the iter-0 result is now preserved (manual, not CI).
- Regression: confirm no perf regression — single snapshot strategy means at most one extra deep-copy per iteration when a new best is found (chorus-test: < 10ms per snapshot at the observed scale).

## Implementation cost estimate

- **Small**: ~150-250 LOC in `core.py` + ~200 LOC in new test file.
- **Risk**: Low. The existing #2540 infrastructure (snapshot variables, restore block, deep-copy machinery) is already in place; this PR extends the comparator and adds a second capture point at iteration end. No new dependencies, no algorithmic change.
- **Touch points**: `core.py` only for the fix; `primitives.py` untouched; new test file. No CLI change (the log line is informational).

## Adjacent context

This was discovered during a productive run where rotation-sign fixes pushed initial-pass completion from 54.8% (v19) → 70.6% (v20 4L) and rip-ups rescued additional nets. The iteration regression isn't currently dominant on this board, but it muddied the signal and is a clear correctness gap.

Related issues:
- #2540: original best-of-iterations preservation (route-count metric only) — this issue extends that work.
- #2334: perturbation tracks `perturbation_best_overflow` but only for magnitude reset, not state preservation — separate concern.
- #2795 / #2800: per-net rip-up convergence — orthogonal to this outer-loop issue.

## Discovered by

Chorus-test v20 routing re-run experiment, 2026-05-12. Live data from the run.

rjwalters:
## Summary

Extends the existing #2540 best-of-iterations infrastructure in `Autorouter.route_all_negotiated` so the saved-partial state is preserved against **overflow regression**, not just route-count regression.

Live failure mode (chorus-test v20 4-layer, 2026-05-12):
- Iteration 0: 30/51 routed, **overflow = 16**
- Iteration 1: makes some rip-up moves, **overflow climbs to 36** (same routed count)
- The original #2540 fix never rolled back because its comparator was `current_routed > best_routed_count` only.

## Changes

- **New dataclass `IterationMetrics`** (`src/kicad_tools/router/core.py:169`) — frozen, with `is_better_than()` comparator over `(routed_count desc, overflow asc, iteration desc)`. Exported on `kicad_tools.router.core`.
- **Snapshot init** at `core.py:4732` now also tracks `best_metrics: IterationMetrics` and a lightweight `iteration_trajectory: list[IterationMetrics]` (three ints per iter — no deep-copies).
- **Iteration-top snapshot capture** at `core.py:4789` switches from `current_routed > best_routed_count` to `current_metrics.is_better_than(best_metrics)`.
- **End-of-iteration capture** via closure `_capture_iteration_end` (Strategy B per Curator: single rolling snapshot, replaced only on strict improvement). Called from BOTH branches:
  - Targeted-ripup branch: `core.py:5497` (right after `overflow_history.append(overflow)`)
  - Standard branch: `core.py:5957`
- **Per-iteration log line** emitted on both branches: `iter N | routed=X/Z | overflow=Y` plus `(new best)` or `| best-so-far=iter-K (routed=..., overflow=...)`.
- **Post-loop restore** at `core.py:6047` switches to lex-tuple comparison via `best_metrics.is_better_than(final_metrics)`; also updates `overflow` so the closing summary reports the restored value, not the discarded worse one.
- **New test file** `tests/test_route_all_negotiated_overflow_regression.py` (437 LOC) covering:
  - 8 unit tests on `IterationMetrics.is_better_than` (all six orderings + `min()` sort_key + frozen invariant)
  - End-to-end log-line emission on `Autorouter.route_all_negotiated`
  - Overflow-regression rollback via controlled `_OverflowSequenceGrid` injection
  - Monotonicity property: `iterations=5` is never strictly worse than `iterations=1` by the lex metric
  - Backward-compat check that the #2540 mid-rip-up timeout case still wins via the lex tuple

## Acceptance criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Monotonic in `--iterations` by `(routed desc, overflow asc)` | Covered by `TestMonotonicityOfIterationCount` |
| 2 | Overflow-regression rollback fires | Covered by `TestOverflowRegressionRollback` |
| 3 | Tie-break by routed_count is unaffected | Covered by `test_fewer_routed_is_worse_even_with_lower_overflow` |
| 4 | Per-iter `iter N \| routed=X/Z \| overflow=Y` log line | Covered by `test_per_iter_log_line_is_emitted` |
| 5 | Existing #2540 tests continue to pass | All 5 tests in `test_route_all_negotiated_partial_state.py` pass |

## Test Plan

- [x] `uv run pytest tests/test_route_all_negotiated_overflow_regression.py` — **14 passed**
- [x] `uv run pytest tests/test_route_all_negotiated_partial_state.py` — **5 passed** (existing #2540 suite unbroken)
- [x] `uv run pytest tests/test_negotiated_per_net_timeout.py tests/test_negotiated_timeout_propagation.py tests/test_negotiated_adaptive.py tests/test_negotiated_congestion_model.py tests/test_router_negotiated_high_current.py tests/test_best_state_tracking.py` — **197 passed**
- [x] `uv run pytest tests/test_router_core.py` — **164 passed**, 1 pre-existing unrelated failure (`TestAdaptiveAutorouterComponentTransform::test_add_component_rotation` — `pad.y=21.0 != 19.0`, present on `main` before this PR)
- [x] `uv run ruff check` on touched files — clean (only pre-existing `I001`/`F401`/`UP037` issues remain in unrelated parts of `core.py`)
- [ ] Manual chorus-test v20 4L re-run to confirm the live 16->36 regression no longer surfaces (requires the board; not part of CI)

## Memory / risk

Strategy B (single rolling snapshot, replaced only on strict improvement) — peak memory at chorus-test scale ~5-20 MB. Replaces an existing deep-copy site rather than adding a new one in the steady state. No algorithmic change. Conflict watch with parallel PR #2800 (touches `core.py:4079-4193`, `:6646-6882`, `:8817-8876`) — no overlap with our touched lines (169-225, 4732-4810, 5497, 5957, 6047-6086).

Closes #2803

### DIFF
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -166,6 +166,56 @@ class MSTEdgeInfo:
     is_first: bool = False
 
 
+@dataclass(frozen=True)
+class IterationMetrics:
+    """Per-iteration scalar metrics for the negotiated routing loop
+    (Issue #2803).
+
+    Provides a lexicographic comparator that lets the outer iteration loop
+    preserve the strictly-best iteration result, even when ``routed_count``
+    is unchanged but ``overflow`` regresses (the failure mode reported in
+    Issue #2803: iteration 0 produced overflow=16, iteration 1 climbed to
+    overflow=36 with the same routed count, and the existing route-count-
+    only restore from Issue #2540 did not roll back).
+
+    Lex order (used by :meth:`is_better_than`):
+
+    1. ``routed_count`` descending — more routed nets is always better.
+    2. ``overflow`` ascending — with equal route counts, lower overflow is
+       better.  This is the new dimension Issue #2803 needs.
+    3. ``iteration`` descending — on a complete tie, prefer the later
+       iteration so perturbation/escape strategies have a chance to bake
+       in.
+
+    Attributes:
+        iteration: Iteration index (0 = initial pass, 1..N = rip-up iters).
+        routed_count: Number of nets with at least one route at iter end.
+        overflow: Grid total overflow at iter end (lower is better).
+    """
+
+    iteration: int
+    routed_count: int
+    overflow: int
+
+    @property
+    def sort_key(self) -> tuple[int, int, int]:
+        """Tuple suitable for ``min()`` / sort key.
+
+        Negated where descending is desired so the *smallest* tuple is the
+        *best* iteration.
+        """
+        return (-self.routed_count, self.overflow, -self.iteration)
+
+    def is_better_than(self, other: IterationMetrics) -> bool:
+        """Return True if ``self`` is strictly better than ``other``.
+
+        Strict (not >=) so equal results never trigger a restore-on-tie,
+        matching Issue #2540's existing semantics (no churn when nothing
+        improved).
+        """
+        return self.sort_key < other.sort_key
+
+
 # Re-export for backward compatibility
 __all__ = [
     "Autorouter",
@@ -4679,24 +4729,92 @@ class Autorouter:
                 f"from rip-up: {', '.join(off_grid_names)}"
             )
 
-        # Issue #2540: Track best-of-iterations so a mid-iteration timeout
-        # does not destroy successful routes from earlier iterations.
+        # Issue #2540 + #2803: Track best-of-iterations so a mid-iteration
+        # timeout does not destroy successful routes from earlier iterations,
+        # AND a completed-but-worse iteration does not silently regress
+        # overflow on a tie in routed count.
+        #
         # ``rip_up_nets`` destructively mutates both ``net_routes`` and
         # ``self.routes`` BEFORE re-routing begins; if ``check_timeout()``
         # fires during the per-net reroute loop, ``self.routes`` is left in
         # the mid-rip-up state (e.g. only the few that survived being
-        # rerouted) instead of the iteration-N-1 stable state.  We snapshot
-        # at the top of each iteration so the post-loop restore can fall
-        # back to the pre-rip-up state from the previous iteration.
+        # rerouted) instead of the iteration-N-1 stable state.  Even without
+        # a timeout, a completed iteration can produce *worse* overflow with
+        # the same routed count (Issue #2803: live chorus-test run jumped
+        # from overflow=16 to overflow=36 across iterations 0->1 with no
+        # change in routed count, and the original Issue #2540 fix did not
+        # roll back because it compared route count only).
         #
-        # Comparison metric is route count (== ``successful_nets``), not
-        # overflow: the user-visible regression is route count, and a
-        # half-restored rip-up can leave overflow numerically identical
-        # while dropping from 20 routes to 10.
+        # Comparison metric is the lex tuple in ``IterationMetrics`` so the
+        # restore considers both routed count (primary) and overflow
+        # (secondary).  We snapshot at the top of each iteration AND at the
+        # end of each iteration on both branches (targeted and standard) so
+        # the post-loop restore has a candidate that reflects the actual
+        # iteration-end state, not just the pre-rip-up state.
         best_routes: list[Route] = copy.deepcopy(list(self.routes))
         best_net_routes: dict[int, list[Route]] = copy.deepcopy(net_routes)
         best_routed_count = sum(1 for r in net_routes.values() if r)
         best_iteration = 0  # 0 = initial pass
+        best_metrics = IterationMetrics(
+            iteration=0,
+            routed_count=best_routed_count,
+            overflow=overflow,
+        )
+
+        # Issue #2803: Lightweight trajectory log (three ints per iteration,
+        # no deep copies) so the per-iteration progression is visible in the
+        # user log and verifiable in tests.  Strategy B from the curator
+        # analysis: full state snapshots remain singleton, only the metric
+        # tuple history is kept.
+        iteration_trajectory: list[IterationMetrics] = [best_metrics]
+
+        def _capture_iteration_end(iter_index: int, overflow_val: int) -> None:
+            """Issue #2803: end-of-iteration capture point.
+
+            Called from BOTH branches (targeted and standard rip-up) after
+            ``overflow_history.append`` to:
+
+            1. Append to the lightweight trajectory log.
+            2. Replace the best-state snapshot if the lex-tuple metric
+               strictly improved.  Only deep-copy on strict improvement
+               (Strategy B — minimal memory cost).
+            3. Emit a single canonical per-iteration log line so the
+               trajectory is visible in the user-facing log.
+            """
+            nonlocal best_metrics, best_routes, best_net_routes
+            nonlocal best_routed_count, best_iteration
+
+            routed_now = sum(1 for r in net_routes.values() if r)
+            metrics = IterationMetrics(
+                iteration=iter_index,
+                routed_count=routed_now,
+                overflow=overflow_val,
+            )
+            iteration_trajectory.append(metrics)
+
+            improved = metrics.is_better_than(best_metrics)
+            if improved:
+                best_metrics = metrics
+                best_routed_count = routed_now
+                best_routes = copy.deepcopy(list(self.routes))
+                best_net_routes = copy.deepcopy(net_routes)
+                best_iteration = iter_index
+
+            # Canonical per-iteration log line.  Suffix shows whether this
+            # iteration replaced the best snapshot or what the running best
+            # still is — makes the regression visible at runtime.
+            if improved:
+                suffix = " (new best)"
+            else:
+                suffix = (
+                    f" | best-so-far=iter-{best_metrics.iteration} "
+                    f"(routed={best_metrics.routed_count}, "
+                    f"overflow={best_metrics.overflow})"
+                )
+            flush_print(
+                f"  iter {iter_index} | routed={routed_now}/{total_nets} | "
+                f"overflow={overflow_val}{suffix}"
+            )
 
         # Skip iteration loop if already timed out
         if not timed_out:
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

CI: 8/8 SUCCESS, mergeStateStatus CLEAN.

### Verification

**Lex tuple correctness (`IterationMetrics.is_better_than`, core.py:209-216)**
`sort_key = (-routed_count, overflow, -iteration)`. Smallest tuple wins, so the implied order is routed DESC, overflow ASC, iteration DESC — exactly the spec. `is_better_than` uses strict `<`, so equal metrics never trigger a restore (matches #2540's no-churn-on-tie semantics).

**Both branches wired**
`_capture_iteration_end(iteration, overflow)` is called from:
- Targeted-ripup branch: `core.py:5502` (immediately after `overflow_history.append(overflow)` at :5495)
- Standard branch: `core.py:5964` (immediately after `overflow_history.append(overflow)` at :5957)

The chorus-test 16->36 failure mode is reachable from both branches and both are now guarded.

**Strategy B confirmed (single rolling snapshot, deep-copy on strict improvement only)**
Only three `copy.deepcopy` sites against `best_routes`/`best_net_routes` (core.py:4754, 4799, 4847) — all replace the same singletons; there is no per-iteration full-state history. The `iteration_trajectory: list[IterationMetrics]` is just 3 ints per iter (the lightweight log the curator analysis called for).

**Post-loop restore (core.py:6047-6086)**
Switched from route-count-only to `best_metrics.is_better_than(final_metrics)`, restores routes via `unmark`/`clear`/`extend`/`re-mark`, AND assigns `overflow = best_metrics.overflow` so the closing summary reports the restored value instead of the discarded worse one.

**Backward compatibility**
Equal `(routed, overflow)` falls through to the iteration tiebreak (later iter wins); strict `<` means existing #2540 cases (timeout mid-rip-up, routed count regressed) still trigger the same restore they always did. The 5 existing `test_route_all_negotiated_partial_state.py` tests passed locally per the test plan.

**Test discrimination**
- Unit tests at lines 43-119 enumerate all six lex orderings + tie-is-not-better + frozen-invariant.
- `test_iter0_preserved_when_iter1_worsens_overflow_same_routed` (line 269) drives the exact chorus-test sequence (overflow: 16 -> 36 with equal routed) via `_OverflowSequenceGrid`.
- `test_iter1_kept_when_routed_count_strictly_increases` (line 306) and `test_fewer_routed_is_worse_even_with_lower_overflow` (line 82) cover the primary-key-dominates case.
- `test_more_iterations_never_strictly_worse` is a property test on monotonicity.

**Conflict watch with #2800**
This PR touches `core.py` at lines 166-225, 4729-4825, 5494, 5956, 6044-6080 and adds `tests/test_route_all_negotiated_overflow_regression.py` only. Disjoint from #2800's regions (4079-4193, 6646-6882, 8817-8876).

### Minor (non-blocking)

`test_iter0_preserved_when_iter1_worsens_overflow_same_routed` (line 269) guards its strongest assertion (`overflow=16` in restore log) behind `if "Restoring iteration" in out:`, so it would silently pass if the restore message never fires. The unit tests on `is_better_than` cover the lex semantics deterministically, so the rollback test is supplementary, but tightening this to assert the message *must* appear (or assert via the trajectory directly) would make the regression test stronger. Not blocking — file as a follow-up if desired.

Approving. Updating labels to `loom:pr`.

PR #2812, opened 2026-05-13
fix(route): atomic --output writes + checkpoint partial progress (closes #2808 + #2809)

### LINKED ISSUE #2808: router: kct route doesn't checkpoint partial results — crash/kill/timeout loses all work
## Symptom

A 2-hour \`kct route\` invocation on chorus-test v20 produced **zero updates** to the configured \`--output\` file. The file timestamp remained at process-launch (12:32) and was never refreshed despite the router making documented progress (transcript shows multiple rip-up successes — SPI_SCK rescued in 189s, AUDIO_R rescued in 30.9s, SPI_NSS rescued, U5-VNEG rescued).

When the router was eventually killed (after going **65+ min past its \`--timeout 2700\` budget** — separately tracked as #2802 / #2806, now merged), the user got the initial-state PCB back, not the best partial achieved during the run.

## Reproducer

\`\`\`
uv run kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb \\
  --output /tmp/test_out.kicad_pcb \\
  --timeout 60 --per-net-timeout 10 \\
  --iterations 50
# Wait, then SIGTERM via kill -TERM &lt;pid&gt; or let it hit --timeout
# Result: /tmp/test_out.kicad_pcb timestamp matches launch time; contents = initial state, not partial progress
\`\`\`

## Why this matters

Even with the just-merged #2802/#2806/#2807 timeout fixes, a router run that crashes (segfault in router-cpp, OOM, signal kill) will lose ALL work between the last checkpoint and the crash. Long routing experiments are now expensive enough that this is a real usability issue.

For the chorus-test rotation-fix hypothesis verification today, we lost ~2 hours of work — the router rescued 4 nets via rip-up but the saved PCB has none of those routes.

## Acceptance criteria

1. \`kct route\` periodically writes the best-so-far PCB to \`--output\` (or a sibling \`_partial\` file). Suggested cadence: every iteration of the negotiated loop, or every N seconds (default ~30s).
2. On any termination path (clean exit, \`--timeout\` reached, SIGTERM, exception), the user gets at minimum the latest checkpointed state.
3. Atomic write semantics so a crash during write doesn't corrupt the output (write to \`<output>.tmp\`, rename to \`<output>\` on completion).
4. \`--checkpoint-interval N\` flag to tune cadence (default 30s; \`--checkpoint-interval 0\` disables for users who don't want disk churn).

## Implementation notes

This intersects with #2803 (best-iter preservation, just merged). The \`best_routes\` snapshot machinery from #2805 (\`core.py\` \`_capture_iteration_end\`) already maintains the right state in memory. The fix here is to **also flush it to disk** on every replacement event (or every N seconds). The disk-flush cost should be small relative to the iteration cost.

Suggested location: \`src/kicad_tools/router/core.py\` in the negotiated loop, after each \`_capture_iteration_end\` call that updates the best snapshot. Reuse the existing PCB-write helper from \`route_cmd.py\`.

## Adjacent observations

- The transcript fragments showed several net rip-up successes that would be valuable to retain. Lost.
- The router's own progress logging (\`flush_print\` from #2799) showed the work happening — the only thing missing is durable persistence.

## Discovered by

Chorus-test v20 routing re-run experiment, 2026-05-12. The combination of long wall-clock + no partial-write + (#2803 regression that's now fixed) meant we got a strictly-worse-than-baseline output despite extensive computation.

### LINKED ISSUE #2809: router: --output flag silently overridden when layer escalation kicks in (writes to <output>_4layer.kicad_pcb)
## Symptom

Specified \`--output /tmp/chorus_v20_routed_v3.kicad_pcb\`. The router wrote to \`/tmp/chorus_v20_routed_v3_4layer.kicad_pcb\` (note the \`_4layer\` suffix). The path I specified retained its initial state and was never updated — easy to mistake for the "no checkpoint" issue (#2808) since both show stale output at the requested path.

Log line confirming:
\`\`\`
--- Saving routed PCB ---
  Saved to: /tmp/chorus_v20_routed_v3_4layer.kicad_pcb
\`\`\`

## Reproducer

Any \`kct route\` invocation with \`--output PATH\` that triggers auto-layer-escalation (2L → 4L) on a board where 2L can't complete:

\`\`\`
uv run kct route /Users/rwalters/GitHub/chorus/hardware/chorus-test-revA/kicad/chorus-test-revA_v20.kicad_pcb \\
  --output /tmp/expected_path.kicad_pcb \\
  --mfr jlcpcb-tier1 --auto-layers --timeout 1800
# After completion: /tmp/expected_path.kicad_pcb is unchanged
# Actual output is at /tmp/expected_path_4layer.kicad_pcb
\`\`\`

## Why this matters

1. **CI pipelines breaking silently**: scripts that read \`\$OUTPUT_PATH\` after \`kct route\` will get the input PCB (no error indication).
2. **Easy to confuse with #2808** (no checkpoint). I almost double-filed today because both gaps present as "the file at --output didn't update."
3. **Surprising behavior**: the user explicitly named the output path; the router silently changed it.

## Suggested fix

Pick one:

**Option A**: honor \`--output\` exactly. If layer escalation happens, still write the final result to the user's specified path. Drop the \`_4layer\` suffix.

**Option B** (alternative): keep the suffix-appended files as artifacts but also write a copy at the user's specified \`--output\` path on success. This lets users access intermediate-layer outputs if they want, while still making the primary path predictable.

**Option C** (least disruptive): keep current behavior but emit a clear stderr warning "writing to \`<actual_path>\` instead of requested \`<output>\` due to layer escalation". Document in \`--help\`.

Recommend **A** for least surprise.

## Acceptance criteria

1. After \`kct route ... --output PATH ...\` (with any combination of \`--auto-layers\`, \`--placement-feedback\`, etc.), the file at \`PATH\` is the final routed PCB.
2. If suffix-appended intermediate files are useful as artifacts, document them in \`--help\` and keep them in addition to the canonical output.
3. Regression test: \`kct route\` on a small fixture with \`--output /tmp/test.kicad_pcb --auto-layers\`, assert \`/tmp/test.kicad_pcb\` exists and has been modified.

## Discovered by

Chorus-test v20 routing experiment (clean re-run with all session fixes), 2026-05-12. The rotation hypothesis verification was nearly invalidated by mistaking the empty \`--output\` for "no progress." Saved by checking \`grep "Saved to" route.log\`.

rjwalters:
## Summary

- **Closes #2808**: `--checkpoint-interval N` flag (default 30s, 0 disables) plumbs an optional callback through `Autorouter.route_all_negotiated` that fires on every best-snapshot replacement, atomically flushing the in-memory `best_routes` snapshot to `--output` via tmp-file + fsync + `os.replace`. A SIGTERM/`--timeout`/Ctrl-C now leaves the user with the best partial result on disk instead of the original unrouted input.
- **Closes #2809**: `--output PATH` is now honored exactly. The four duplicated terminal save sites in `route_cmd.py` are consolidated into a single `_write_routed_pcb()` helper that no longer rewrites the output stem to `<name>_4layer.kicad_pcb` on auto-escalation; the layer count is recorded inside the PCB content via `update_pcb_layer_stackup` only.

## Implementation notes

- **`_write_routed_pcb(pcb_path, output_path, route_sexp, *, layer_count=2, is_checkpoint=False) -> Path`** in `cli/route_cmd.py` replaces four duplicated read -> stackup-update -> insert -> validate -> write blocks. Atomic semantics: `<output>.tmp` + `os.fsync` + `os.replace`. Honors `output_path` exactly (no `_Nlayer` suffix).
- **`_make_checkpoint_callback(pcb_path, output_path, interval, quiet)`** builds a throttled closure (gated by `time.monotonic()`, immune to wall-clock skew). First improvement event always writes; subsequent events throttled to one per `interval`. Returns `None` when `interval <= 0` so the router treats it as "no callback".
- **`Autorouter.route_all_negotiated`** gains an optional `checkpoint_callback: Callable[[list[Route], IterationMetrics], None] | None = None` parameter. Wired into BOTH improvement sites: `_capture_iteration_end` (line ~4837) and the iteration-top pre-rip-up snapshot (line ~4898). **Critical:** the callback receives the deep-copied `best_routes` snapshot, NOT `self.routes` (which can regress mid-iteration). Callback exceptions are swallowed with a log line so a transient disk error never aborts routing.
- All four terminal save sites (`route_with_layer_escalation`, `route_with_rule_relaxation`, `route_with_combined_escalation`, and the primary `main()` flow) now route through `_write_routed_pcb` and are protected by atomic-write semantics, closing the torn-file hazard at every save path.

## Test plan

- [x] Unit tests in `tests/test_route_checkpoint.py` (17 tests, all pass):
  - `_write_routed_pcb`: exact-path write, empty route_sexp, 2L no-op, checkpoint mode skips stackup, atomic semantics (forced `os.replace` failure leaves existing file untouched + leaves `.tmp` for inspection).
  - `_make_checkpoint_callback`: zero/negative interval returns `None`, first call writes, rapid calls throttled, post-interval call writes again. Throttling is verified via segment-count differences in the file content (segment count round-trips through the sexp serializer reliably; per recovery instructions, this avoids the `net_name` round-trip pitfall).
  - End-to-end on `Autorouter`: callback invocation receives correct types, snapshot is independent of subsequent `self.routes` mutations, callback exceptions don't abort routing, default `None` preserves backward compat.
  - Issue #2809 regression: `--output user.kicad_pcb` with `layer_count=4` lands at exactly `user.kicad_pcb` (not `user_4layer.kicad_pcb`) and the stackup IS escalated to 4L inside the file content.
- [x] All 14 tests in `tests/test_route_all_negotiated_overflow_regression.py` still pass.
- [x] `--checkpoint-interval` appears in `kct route --help`.
- [x] `ruff check` clean on all changed files.

Recovered from a stalled prior Builder; the worktree had the implementation complete and 17 tests already passing — only minor lint cleanup (remove unused `os` import, sort imports in the new test file) remained.

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -56,6 +56,7 @@ Layer Stack Configuration:
 import argparse
 import logging
 import math
+import os
 import random
 import shutil
 import signal
@@ -316,6 +317,92 @@ def _validate_sexp_parentheses(content: str) -> bool:
     return depth == 0
 
 
+def _write_routed_pcb(
+    pcb_path: Path,
+    output_path: Path,
+    route_sexp: str,
+    *,
+    layer_count: int = 2,
+    is_checkpoint: bool = False,
+) -> Path:
+    """Atomically write a routed PCB file.
+
+    Consolidates the read_text -> update_pcb_layer_stackup ->
+    _insert_sexp_before_closing -> _validate_sexp_parentheses -> write
+    sequence that was previously duplicated at four save sites in this
+    module.
+
+    Uses an atomic write via ``<output>.tmp`` + ``os.fsync`` +
+    ``os.replace(tmp, output)`` so a SIGKILL/OOM between bytes-on-the-wire
+    and final flush cannot leave a torn (truncated) PCB file at the user's
+    output path.  Closes Issue #2808's torn-file hazard.
+
+    Honors ``output_path`` exactly -- callers must NOT pre-suffix the path
+    with ``_4layer`` or similar; the layer count is recorded inside the
+    PCB content via :func:`update_pcb_layer_stackup`, not in the filename
+    (closes Issue #2809).
+
+    Args:
+        pcb_path: Path to the source (unrouted) PCB file.  Read fresh on
+            every call so checkpoint writes pick up upstream edits to the
+            input file (rare, but supported).
+        output_path: Final destination path.  Written atomically via a
+            sibling ``.tmp`` file.
+        route_sexp: S-expression fragment(s) to insert before the closing
+            ``)``.  Can be empty -- in that case the original PCB is
+            written back unchanged (useful for checkpoints that fire
+            before any net has routed).
+        layer_count: Target copper layer count for the layer stackup
+            update.  ``2`` is a no-op.  Defaults to ``2``.
+        is_checkpoint: When True, skip the layer-stackup mutation.
+            Checkpoints serialize the in-progress best snapshot; they
+            should match whatever stackup is currently in use and not
+            attempt to escalate.  Defaults to False.
+
+    Returns:
+        The ``output_path`` that was written to (returned unchanged so
+        callers can use the helper inline, e.g.
+        ``written = _write_routed_pcb(...)``).
+
+    Raises:
+        ValueError: If the generated PCB has unbalanced S-expression
+            parentheses (indicates a bug in route generation -- caller
+            is responsible for surfacing this).
+    """
+    original_content = pcb_path.read_text()
+
+    # Update layer stackup for terminal writes when we escalated above 2L.
+    # Checkpoints skip this -- they reflect mid-route state and should not
+    # rewrite the stackup at every flush.
+    if not is_checkpoint and layer_count > 2:
+        original_content = update_pcb_layer_stackup(original_content, layer_count)
+
+    if route_sexp:
+        output_content = _insert_sexp_before_closing(original_content, route_sexp)
+    else:
+        output_content = original_content
+
+    if not _validate_sexp_parentheses(output_content):
+        logger.error("Generated PCB file has unbalanced parentheses")
+        raise ValueError(
+            "Generated PCB file has invalid S-expression syntax "
+            "(unbalanced parentheses). This is a bug in kicad-tools. "
+            "Please report it."
+        )
+
+    # Atomic write: tmp file -> fsync -> rename.  Sibling-in-same-dir
+    # ensures os.replace is a same-filesystem rename (atomic on POSIX).
+    tmp_path = output_path.with_suffix(output_path.suffix + ".tmp")
+    tmp_path.write_text(output_content)
+    # fsync the file so a crash between write and rename does not leave
+    # a partial-content tmp file masquerading as the routed PCB.
+    with open(tmp_path, "rb") as f:
+        os.fsync(f.fileno())
+    os.replace(tmp_path, output_path)
+
+    return output_path
+
+
 def _connectivity_snapshot(router: "Autorouter"):
     """Capture per-net connectivity + deep-copied routes (issue #2596).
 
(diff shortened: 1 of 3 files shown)


rjwalters:
**Judge approval — LGTM**

CI: 8/8 SUCCESS, mergeStateStatus CLEAN.

Reviewed all high-risk items:

- **Atomic write** (`_write_routed_pcb` at route_cmd.py:320): correct sequence — validate → `<output>.tmp` write → `os.fsync` on fresh fd → `os.replace`. The `ValueError` path raises **before** any tmp-write touches disk, so the original output is unchanged on validation failure. Test `test_replace_failure_leaves_existing_file_unchanged` exercises the rename-failure case with a real sentinel file and verifies content preservation.
- **#2809 fix**: confirmed all four previous `with_stem(stem + f"_{N}layer")` mutation sites are gone. The helper honors `output_path` exactly; layer escalation is recorded inside content via `update_pcb_layer_stackup`, not the filename. `test_write_routed_pcb_does_not_mutate_path_for_4layer` asserts the file lands at the exact requested path AND that the stackup did escalate (proving `layer_count` was honored via content, not filename).
- **Callback subtlety (CRITICAL)**: verified at core.py:4844-4851 and core.py:4906-4912. Both callback invocation sites fire **after** `best_routes = copy.deepcopy(list(self.routes))`, so the callback receives the just-snapshotted list — never `self.routes` (live, possibly-worse state). This is the linchpin of #2808 and it's correct. `test_callback_receives_best_routes_snapshot_not_self_routes` mutates `router.routes` after the call and confirms the captured snapshot is unaffected.
- **Throttle**: uses `time.monotonic()` (skew-immune). First call always fires (sentinel `last_time[0] is None`). `interval <= 0` returns `None` (disabled, router treats as no-op). Throttling test verifies via segment-count differences in the file content, not a tautological assertion.
- **No exception leakage**: both callback sites wrap in `try/except Exception` with a log line. Tested via raising-callback fixture.
- **Backward compat**: `checkpoint_callback=None` default with `if checkpoint_callback is not None:` guards. Test verifies legacy path.

Scope is the expected 3 files (route_cmd.py +241/-106, core.py +37/-0, new test file +537).

Approving.

PR #2813, opened 2026-05-13
fix(router): classify pad-clearance blockers separately from vias (closes #2810)

### LINKED ISSUE #2810: router: failure_analysis classifies pad-clearance-zone blockers as 'via' (false positive misleads diagnosis)
## Symptom

\`src/kicad_tools/router/failure_analysis.py:1524-1529\` classifies any blocked grid cell with \`pad_blocked=False, usage_count=0\` as \`"via"\` — but that combination also matches **pad clearance zone** cells set up at \`src/kicad_tools/router/grid.py:909-927\`. The two are visually identical to the classifier; only the cell origin distinguishes them.

This is a diagnostic-only bug — it doesn't break routing — but it produces false-positive failure causes that mislead users (and even Curators investigating routing failures). Example: issue #2761 was misdiagnosed for ~24 hours as "ViaConflictManager not wired" when the actual blocker was a pad clearance zone, not a via.

## Reproducer

Route \`boards/03-usb-joystick/output/usb_joystick.kicad_pcb\`. XTAL2 (U1.3) fails routing. The router's failure-analysis output attributes the failure to a "via" near U1.3, but the nearest real XTAL1 via is 7.49mm away — far outside any plausible blocker radius. The actual blocker is the pad clearance zone of the physically adjacent U1.2 (XTAL1) at 0.8mm pitch.

## Acceptance criteria

1. Classifier at \`failure_analysis.py:1524-1529\` distinguishes "via" from "pad-clearance" cells. Likely via a tag on the cell type at \`grid.py\` setup time, OR by querying the originating pad's coordinates within some pitch-relative radius.
2. Failure-analysis output for the XTAL2 case names "pad-clearance" (or similar) instead of "via".
3. Regression test using a minimal TQFP-style fixture with adjacent dense pads.
4. Run failure analysis on board 03 with current code AND after fix — assert the new label appears for XTAL2.

## Why this matters

- **Diagnostic accuracy**: agents and humans rely on failure-cause categorization to choose fix strategy. "Via conflict" → \`ViaConflictManager\`. "Pad clearance" → sub-grid retry / placement nudge. Mislabeling costs investigation time.
- **Foundation for #2761's actual fix**: the real XTAL2 routing fix needs to know it's a pad-clearance failure to dispatch the right retry strategy.

## Discovered by

Curator re-investigation of #2761 on 2026-05-12 with explicit "prior diagnosis WRONG" brief. Curator confirmed Builder's earlier pushback that the original "via blocks pad" root cause was incorrect.

Related: this is the small, mechanical half of the #2761 split. Companion issue (XTAL2 actual routing fix) filed separately.

### LINKED ISSUE #2811: router: XTAL2 (board 03 U1.3) fails pin access — drop new_routes gate on sub-grid retry, widen escape geometry for dense SMD
## Symptom

XTAL2 net (U1.3 on board 03 USB joystick) consistently fails routing. The pad is at world (~135.5, 130.8) on a TQFP-32 footprint with 0.8mm pitch and 1.2 × 0.5mm pads. The pad-edge gap to the physically adjacent U1.2 (XTAL1) pin is 0.3mm — below the 0.6mm routing-corridor requirement — so there is **no top-layer channel** between U1.2 and U1.3.

U1.3 is also off-grid by 0.048mm. The sub-grid retry path that would normally handle this is **gated off** by the \`not new_routes\` check at \`src/kicad_tools/router/core.py:1516-1527\`: XTAL2 fails *after* its sibling Y1.2 ↔ C6.1 succeeds, so \`new_routes\` is non-empty and the sub-grid retry is skipped.

## Reproducer

\`uv run kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb\` — observe XTAL2 fails PIN_ACCESS.

## Acceptance criteria

1. \`kct route boards/03-usb-joystick/...\` successfully routes XTAL2 (board 03 completion ≥ current minus the XTAL2 net).
2. Regression test asserting XTAL2 (or a comparable dense-SMD fixture) routes after fix.
3. No regression on other example boards (01-07 stay at their current completion levels).

## Recommended fix paths (per Curator)

**Phase A (high confidence)**:
1. **Drop the \`not new_routes\` gate** at \`core.py:1516-1527\` — partial-success failures like XTAL2 should still trigger sub-grid retry.
2. **Widen sub-grid escape geometry** in \`_retry_net_with_subgrid\` (\`core.py:8470\`) and the PIN_ACCESS helper (\`core.py:8798\`) for pads adjacent to clearance walls — current sub-grid is too narrow for off-grid SMD pads next to dense neighbors.

**Phase B (if Phase A insufficient)**:
3. **In-pad / near-pad via as first escape** on dense SMD pitches. Currently the router tries top-layer escape first; for sub-1mm pitches, a via-down is often the only way out.

**Phase C (escalation)**:
4. **Placement nudge** — if the geometry is fundamentally unrouteable, suggest a placement change via the existing placement-feedback loop.

## Why ViaConflictManager is NOT the right tool

The original Curator on #2761 diagnosed this as "ViaConflictManager not wired" — wrong. The blocker is a **pad clearance zone**, not a via. The nearest real XTAL1 via is 7.49mm from U1.3. The misdiagnosis was caused by a false-positive classifier in failure-analysis (filed as separate companion issue).

## Discovered by

Curator re-investigation of #2761 on 2026-05-12. This is the second half of the #2761 split. The Builder's pushback ("blocker is not a via") was correct.

Note: the original #2761 acceptance criterion (\`ViaConflictManager.stats.relocations_succeeded ≥ 1\`) is unreachable on this case and should be retired. #2761 itself can be closed as superseded once both this issue and the classifier-fix companion are filed.

rjwalters:
## Summary

Fixes a diagnostic-only mislabelling in `analyze_pad_access_blockers` where pad-clearance-zone cells were classified as `"via"` blockers. Adds a fourth `"pad_clearance"` branch using the existing `cell.original_net` discriminator (set ONLY by `Grid._block_pad`, never touched by `_mark_via`/`_mark_segment`).

Reproduces the TQFP fine-pitch scenario from board 03 (XTAL2): two pads on different nets close enough that the neighbor's clearance envelope overlaps the victim's pad-access search radius. Before this fix, the failure-analysis report said "blocked by N at 0.32mm (via)" -- misleading the operator into looking for a phantom via that doesn't exist.

## Changes

- `src/kicad_tools/router/failure_analysis.py:1523-1546`: add one-line branch
  ```python
  elif cell.original_net != 0 and cell.original_net == cell.net:
      blocking_type = "pad_clearance"
  ```
  placed BEFORE the `"via"` fallback so it takes precedence on the pad-clearance case. Annotated with provenance comments for the discriminator's source-of-truth (`grid.py:903/933/3111/3136`) and the ambiguous-case rationale.
- `src/kicad_tools/router/failure_analysis.py:220`: update `PadAccessBlocker.blocking_type` docstring enum to include `"pad_clearance"`.
- `tests/test_failure_analysis.py`: two new tests in `TestAnalyzePadAccessBlockers`:
  - `test_analyze_pad_access_blockers_distinguishes_pad_clearance_from_via`: programmatic 2-pad fixture reproducing the bug.
  - `test_analyze_pad_access_blockers_via_remains_via`: regression guard ensuring isolated vias still report as `"via"`.

Total: 75 lines net (~22 src + ~53 test). Two files only.

## Scope Boundaries (per Curator)

In scope:
- Classifier branch addition
- Docstring update
- Two new unit tests

Out of scope (deliberately untouched):
- Routing logic (rip-up, retry, sub-grid escape, ViaConflictManager wiring) -- companion fix #2811
- Message-format strings at `core.py:1435`, `output.py:925`, `failure_analysis.py:238` (they print whatever string they get; visual change comes for free)
- Grid-setup-time tags or schema changes (`original_net` is sufficient)
- C++ grid `mark_via` parity (`cpp/src/grid.cpp`) -- failure analysis runs against the Python grid only

## Test plan

- [x] `pytest tests/test_failure_analysis.py::TestAnalyzePadAccessBlockers -v` -- all 6 tests pass (4 existing + 2 new)
- [x] `pytest tests/test_failure_analysis.py -q` -- all 60 tests pass
- [x] Verified both NEW tests FAIL on main at HEAD without the source fix (confirms they pin the bug correctly):
  - `test_analyze_pad_access_blockers_distinguishes_pad_clearance_from_via` -> got `['via']` instead of `pad_clearance`
  - (`test_analyze_pad_access_blockers_via_remains_via` passes on main because vias have always been classified as via -- it's the regression guard)
- [x] `ruff check` on changed files: no new lint errors introduced (pre-existing F541 errors at lines 427/1281 are untouched)
- [x] `pytest tests/test_router_grid.py tests/test_router_core.py -q` -- only pre-existing unrelated failure in `test_add_component_rotation` (verified on main at HEAD)

## Manual smoke check (for reviewer, not CI)

Per Curator: run `uv run python boards/03-usb-joystick/route_demo.py` before and after; the failure summary for XTAL2 should change from `"via at 0.32mm"` to `"pad_clearance at 0.32mm"`. Number of routed nets does NOT change (10/11 either way) -- the actual routing fix is #2811's job.

Closes #2810

### DIFF
--- a/src/kicad_tools/router/failure_analysis.py
+++ b/src/kicad_tools/router/failure_analysis.py
@@ -217,7 +217,7 @@ class PadAccessBlocker:
     pad_ref: str  # e.g. "U1.13"
     blocking_net: int  # Net ID of the blocking net
     blocking_net_name: str  # e.g. "SC_POS_PLUS"
-    blocking_type: str  # "trace", "via", "pad"
+    blocking_type: str  # "trace", "via", "pad", "pad_clearance"
     distance: float  # Distance from pad center to blocking element in mm
     suggested_clearance: float  # Clearance that would allow access in mm
 
@@ -1521,10 +1521,27 @@ class RootCauseAnalyzer:
                 distance = math.sqrt((cell_wx - pad_x) ** 2 + (cell_wy - pad_y) ** 2)
 
                 # Determine blocking type
+                #
+                # Issue #2810: Distinguish pad-clearance zones from vias. The
+                # ``original_net`` attribute on a cell is set ONLY by
+                # ``Grid._block_pad`` (see ``grid.py:903/933/3111/3136``) at
+                # pad placement time and is never touched by ``_mark_via`` or
+                # ``_mark_segment``. So when ``original_net == cell.net`` and
+                # both are non-zero, the cell is unambiguously a pad-clearance
+                # zone that has not been overwritten by a later via.
+                #
+                # Ambiguous case (a via lands on top of a previously-cleared
+                # pad-clearance zone): ``_mark_via`` overwrites ``cell.net`` to
+                # the via's net while ``cell.original_net`` stays at the
+                # pad's net. The discriminator fails and the cell correctly
+                # falls through to ``"via"`` -- matching operator intent that
+                # the immediate blocker is the via's clearance ring.
                 if cell.pad_blocked:
                     blocking_type = "pad"
                 elif cell.usage_count > 0:
                     blocking_type = "trace"
+                elif cell.original_net != 0 and cell.original_net == cell.net:
+                    blocking_type = "pad_clearance"
                 else:
                     blocking_type = "via"
 
--- a/tests/test_failure_analysis.py
+++ b/tests/test_failure_analysis.py
@@ -18,7 +18,7 @@ from kicad_tools.router.failure_analysis import (
 )
 from kicad_tools.router.grid import RoutingGrid
 from kicad_tools.router.layers import Layer, LayerStack
-from kicad_tools.router.primitives import Pad
+from kicad_tools.router.primitives import Pad, Via
 from kicad_tools.router.rules import DesignRules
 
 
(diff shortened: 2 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

Small, well-scoped diagnostic fix. All review criteria check out.

### CI status
- 8/8 checks SUCCESS (Lint, Type Check, Test 6m16s, C++ Build, DRC, kicad-cli, Diff-Pair Regression, Match-Group Regression)
- Merge state: CLEAN
- No `test_add_component_rotation` failure surfaced in this CI run (either resolved upstream or wasn't included in the Test job invocation; either way, not regressed by this PR)

### Discriminator correctness verified
Confirmed against `src/kicad_tools/router/grid.py`:
- `_original_net` is written ONLY by `_block_pad`-family code at lines 903, 933, 3111, 3136
- `_mark_via` (verified) writes only `cell.net` and `cell.blocked` — never touches `_original_net`
- `_mark_segment` likewise does not touch `_original_net` (no matches in grep)
- Therefore `cell.original_net != 0 and cell.original_net == cell.net` is true iff the cell is a still-owned pad-clearance zone that has NOT been overwritten by a later via or segment
- Ambiguous case (via on top of pad-clearance): `cell.net` becomes the via's net while `original_net` stays at the pad's net → discriminator False → falls through to "via", which matches operator intent (the via's clearance ring is the immediate blocker)

### Branch ordering
Branch placement is correct: `pad` → `trace` → `pad_clearance` → `via` (else). The new branch is BEFORE the `else: blocking_type = "via"` fallback, so the pad-clearance case takes precedence as required.

### Docstring update
Line 220: `# "trace", "via", "pad", "pad_clearance"` — enum extended correctly.

### Fixture adaptation soundness
The test docstring explicitly explains why 0.2mm pads at 0.8mm pitch were chosen: at the Curator's original 0.5mm/0.8mm geometry, the neighbor's metal cells would be the closest blocker (not the clearance zone) and the case wouldn't reproduce. At 0.2mm/0.8mm, neighbor metal sits just outside the ~0.6mm pad-access search radius while the ~0.25mm clearance envelope reaches inside. This is exactly the configuration that exercises the new classifier branch — and the fact that the Builder identified and corrected this geometry issue strongly indicates the test was actually run, not stub-claimed.

### Test discrimination
- `test_analyze_pad_access_blockers_distinguishes_pad_clearance_from_via`: asserts `blocking_type == "pad_clearance"` for at least one neighbor-net blocker. Would fail on main (where the only `else` produced `"via"`).
- `test_analyze_pad_access_blockers_via_remains_via`: places only a via (via `_mark_via` directly), asserts ALL blockers from that net are `"via"`. Locks in the regression: if a future change set `original_net` from `_mark_via`, this test would catch it. Builder verified the first test fails on main; the second is a structural regression guard.

### Scope discipline
2 files only: `failure_analysis.py` (+18/-1) and `tests/test_failure_analysis.py` (+128/-1). No grid.py changes, no router-logic changes, no ViaConflictManager changes. Companion routing fix correctly deferred to #2811.

### Minor observations (NOT blockers)
- Inline rationale comment at lines 1523-1535 is exemplary — captures the discriminator's invariant, lists the four source-of-truth line numbers in grid.py, and explicitly documents the ambiguous-case fall-through behavior with the operator-intent justification. Future maintainers reading this branch will not be confused about why it works.
- The PR body's documented manual smoke check (XTAL2 report on board 03) is reviewer-runnable but not gating; the diagnostic-only nature of this change means visual confirmation is "nice to have" rather than required.

Ship it.

Issue #2622, opened 2026-05-10
fix(drc): jlcpcb-tier1 manufacturer name accepted by router but rejected by DRC


rjwalters:
## Problem

\`kct route --manufacturer jlcpcb-tier1\` routes successfully (router profile is recognized) but the auto-fix DRC step then warns:

\`\`\`
Warning: DRC check failed: Unknown manufacturer: 'jlcpcb-tier1'.
  Available: flashpcb, jlcpcb, oshpark, pcbway, seeed
\`\`\`

This silently skips DRC enforcement when targeting JLCPCB's Capability Plus tier (the case the user actually wants when paying for fine features).

## Root cause

\`src/kicad_tools/router/mfr_limits.py\` defines \`MFR_JLCPCB_TIER1\` and registers \`jlcpcb-tier1\` plus aliases (\`jlcpcb-capabilityplus\`, \`jlcpcb_capabilityplus\`, etc.).  The DRC manufacturer registry (different module) only knows the base profiles \`flashpcb, jlcpcb, oshpark, pcbway, seeed\` and doesn't share the router's alias map.

## Fix options

1. Have DRC fall back to the base profile when given a tier suffix (\`jlcpcb-tier1\` → use \`jlcpcb\` rules).
2. Promote \`jlcpcb-tier1\` to a first-class DRC profile with tighter trace/clearance/drill values matching the router's MFR_JLCPCB_TIER1.
3. At minimum, share the alias table.

(2) is best — it makes \`--manufacturer jlcpcb-tier1\` consistent end-to-end.

## Acceptance criteria

- [ ] \`kct route --manufacturer jlcpcb-tier1 <pcb>\` completes the DRC step without the "Unknown manufacturer" warning
- [ ] DRC rules used for jlcpcb-tier1 match the router's MFR_JLCPCB_TIER1 capabilities (smaller min trace/clearance/drill)
- [ ] Existing jlcpcb DRC tests still pass

## Related

- Discovered while routing chorus-test-revA v19 with jlcpcb-tier1; the DRC check phase silently no-op'd.

rjwalters:
## Curator notes (verified against current codebase)

### Confirmed root cause

Two independent manufacturer registries with divergent membership:

1. **Router registry** (`src/kicad_tools/router/mfr_limits.py`):
   - `MFR_LIMITS` includes `jlcpcb-tier1` directly (line 106).
[...]

Issue #2620, opened 2026-05-10
fix(cli): --placement-feedback-outer-timeout & --placement-feedback-stagnation-patience not propagated to route_cmd


rjwalters:
## Problem

\`kct route --placement-feedback-outer-timeout 1800 <pcb>\` errors:
\`\`\`
kicad-tools route: error: unrecognized arguments: --placement-feedback-outer-timeout 1800.0
\`\`\`

Even though \`--help\` advertises the flag and \`src/kicad_tools/cli/parser.py:2349\` defines it on the top-level \`route\` subparser, the route subcommand dispatches to \`src/kicad_tools/cli/route_cmd.py\` (a separate argparse) which **does not** register \`--placement-feedback-outer-timeout\` or \`--placement-feedback-stagnation-patience\`.  When \`commands/routing.py\` forwards the flag (line 263–269) the inner parse rejects it.

## Affected flags

Added in PR #2607 (closes #2606) but only at the top level:
- \`--placement-feedback-outer-timeout SECONDS\`
- \`--placement-feedback-stagnation-patience N\`

Both fail with \`unrecognized arguments\` when forwarded.

## Repro

\`\`\`
uv run kct route --placement-feedback-outer-timeout 1800 --dry-run <any.kicad_pcb>
# error: unrecognized arguments: --placement-feedback-outer-timeout 1800.0
\`\`\`

## Fix

In \`src/kicad_tools/cli/route_cmd.py\` argparse setup (around lines 3376–3431 where the other \`--placement-feedback-*\` flags live), add:

\`\`\`python
parser.add_argument("--placement-feedback-stagnation-patience", type=int, default=3, metavar="N")
parser.add_argument("--placement-feedback-outer-timeout", type=float, default=None, metavar="SECONDS")
\`\`\`

with help text matching parser.py.  Then thread them through to the placement-feedback loop (\`route_cmd.py:922\` already reads \`placement_feedback_outer_timeout\` via getattr — confirm the same for stagnation patience).

## Test

Add a CLI smoke test (e.g. in \`tests/test_route_cli.py\` or similar) that invokes \`kct route --placement-feedback-outer-timeout 1800 --dry-run\` on a fixture PCB and asserts exit 0.

## Discovered

While running chorus-test routing with placement feedback; had to drop the flag and rely on per-net \`--timeout\` instead.

rjwalters:
## Curator notes (verified against `main` @ a718d2d5)

### Bug confirmed end-to-end

[...]

PR #2818, opened 2026-05-13
fix(cli): wire --checkpoint-interval through outer parser + add drift regression test (closes #2817)

### LINKED ISSUE #2817: PR #2812 argparse drift: --checkpoint-interval added to inner parser only; outer parser + forwarding shim missed
## Symptom

\`uv run kct route --checkpoint-interval 30\` fails with \`error: unrecognized arguments: --checkpoint-interval 30\`. The flag was added by PR #2812 but only to one of the three sites the CLI uses.

## Reproducer

\`\`\`
git log -1 --oneline   # bf6926fa or later (#2812 merged)
uv run kct route --help | grep checkpoint   # empty
uv run kct route /path/to/pcb --checkpoint-interval 30 -o /tmp/x.pcb
# error: unrecognized arguments: --checkpoint-interval 30
\`\`\`

## Root cause

The CLI uses a **two-stage parser**:

- **Outer parser** at \`src/kicad_tools/cli/parser.py\` \`_add_route_parser\` (lines 2106+) — what \`argparse\` actually invokes on \`kct route ...\`
- **Dispatch shim** at \`src/kicad_tools/cli/commands/routing.py\` \`run_route_command\` (line 155) — translates \`args.*\` into \`sub_argv\` list and re-invokes \`route_cmd.main()\`
- **Inner parser** at \`src/kicad_tools/cli/route_cmd.py\` \`main()\` (line 3596) — its own \`argparse.ArgumentParser\`

PR #2812 added \`--checkpoint-interval\` only to the **inner** parser at \`route_cmd.py:3730\`. The outer parser was not updated, and the forwarding shim doesn't propagate it. The flag is rejected before reaching the code that wires \`_make_checkpoint_callback\`.

## Saving grace (and the subtle hazard)

\`route_cmd.py:5194\` uses \`getattr(args, \"checkpoint_interval\", 30.0)\`, so the **default 30s checkpointing still fires** at the inner-parser layer. This means:
- Existing \`kct route\` invocations already use checkpointing-as-default ✓
- But users can NOT disable checkpointing (\`--checkpoint-interval 0\`) or tune it
- And the silent acceptance of the default hides the wiring bug from casual notice

This is the SAME pattern as the now-closed **#2620** (\`--placement-feedback-outer-timeout\` declared in outer parser, rejected at inner argparse) and **#2622/#2793** (argparse drift between manufacturer registry sites). Each fix went one direction; this PR went the other.

## Suggested fix (~6 lines per Curator)

1. **\`parser.py\` around line 2167** (next to \`--per-net-timeout\`, the model pattern):
   \`\`\`python
   route_parser.add_argument(
       \"--checkpoint-interval\",
       type=float,
       default=30.0,
       help=\"Interval in seconds between best-so-far checkpoint writes to --output. Default: 30. Use 0 to disable.\",
   )
   \`\`\`

2. **\`commands/routing.py\` after the per_net_timeout block (~line 186)**:
   \`\`\`python
   ci = getattr(args, \"checkpoint_interval\", 30.0)
   if ci != 30.0:
       sub_argv.extend([\"--checkpoint-interval\", str(ci)])
   \`\`\`

## Acceptance criteria

1. \`uv run kct route --help\` shows \`--checkpoint-interval\` in the visible flag list.
2. \`uv run kct route ... --checkpoint-interval 0\` disables checkpointing (confirm via mtime).
3. \`uv run kct route ... --checkpoint-interval 5\` writes every 5s (confirm via mtime delta).
4. Existing \`#2812\`-merged tests still pass.
5. **Add an argparse-drift regression test**: every flag added to \`route_cmd.py\`'s inner parser MUST also exist on the outer parser. Could be a metaprogramming-based test that introspects both parsers and asserts the flag sets match (minus internal-only flags listed in an explicit allowlist).

The drift-prevention test (criterion 5) is the long-term fix — this is the third instance of the same pattern this session. Without it we'll keep filing this issue.

## Discovered by

Board 05 routing experiment, 2026-05-12, attempting to verify #2812's atomic checkpoint persistence. Pre-flight failed at flag-acceptance check.

## Related

- #2620 (closed) — same pattern, different flag
- #2622 / #2793 (closed) — same pattern, mfr arg
- #2812 (closed) — the PR that introduced this drift

rjwalters:
## Summary

Two-part fix for #2817:

1. **Plumbing fix** -- Wire `--checkpoint-interval` through the outer parser (`parser.py`) and the forwarding shim (`commands/routing.py`), modeled after the existing `--per-net-timeout` block.  PR #2812 added the flag only to the inner `route_cmd.py` parser, so `kct route --checkpoint-interval 30` failed with `error: unrecognized arguments`.

2. **Long-term fix** -- Add `tests/test_cli_parser_drift.py`, an argparse-drift regression test that introspects both parsers and asserts every flag on the inner parser is also on the outer parser (modulo an explicit allowlist of historically inner-only debug/diagnostic flags).  This is the third instance of the same drift pattern this session (#2620, #2622/#2793, #2812 -> #2817); the regression test prevents a fourth.

## Why the regression test matters

The drift is invisible to the type checker and to most unit tests because both parsers are constructed independently with no shared schema.  The test:

- Captures the inner parser by intercepting `argparse.ArgumentParser.parse_args` (no refactor needed)
- Walks subparsers on the outer parser to find the `route` subparser
- Computes `inner_flags - outer_flags` and asserts the difference is a subset of `INNER_ONLY_ALLOWLIST` (32 entries, snapshotted today)
- A second test (`test_allowlist_entries_are_actually_inner_only`) prevents the allowlist from going stale

I verified the test catches the bug by stashing the `parser.py` change and re-running -- 2 of 3 tests fail with a clear message pointing at the fix sites.

## Test plan

- [x] `uv run kct route --help | grep checkpoint` shows `--checkpoint-interval` in visible help
- [x] `--checkpoint-interval 0` accepted at outer parser and forwarded as `["--checkpoint-interval", "0.0"]` to inner parser
- [x] `--checkpoint-interval 5` accepted similarly
- [x] New `tests/test_cli_parser_drift.py` (3 tests) passes on fixed state
- [x] New drift test FAILS on unfixed state (confirmed by stashing the parser.py change)
- [x] All 17 existing PR #2812 tests in `tests/test_route_checkpoint.py` still pass
- [x] All 87 tests in `tests/test_cli.py` and `tests/test_cli_commands.py` still pass (107 total green)
- [x] `ruff check` and `ruff format --check` clean on touched files

Closes #2817

## Notes

- `uv.lock` has unrelated drift on `main` (a `pynng` IPC extra was added) -- not included in this commit.
- The allowlist captures 32 currently-inner-only flags (debug/profiling toggles like `--analyze`, `--diagnostics`, `--profile`, plus dev-experimental router knobs like `--two-phase`, `--escape-routing`, `--auto-pour`).  These are not part of this issue's scope; the test merely pins today's state and demands future inner-only additions be justified explicitly.
- A separate drift exists in the opposite direction: `--max-search-iterations` is on the outer parser only and is not forwarded by the shim (it relies on `getattr(args, ...)` defaulting to 0 inside the inner parser, similar to the saving-grace pattern from #2817).  Not fixed here -- the AC was scoped to inner-only flags.

### DIFF
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -184,6 +184,12 @@ def run_route_command(args) -> int:
     per_net_timeout_val = getattr(args, "per_net_timeout", 30.0)
     if per_net_timeout_val != 30.0:
         sub_argv.extend(["--per-net-timeout", str(per_net_timeout_val)])
+    # Issue #2817: forward --checkpoint-interval to the inner parser.  The
+    # inner default also lives at 30.0, so only forward when the user passed
+    # a non-default value (matches the per-net-timeout pattern above).
+    checkpoint_interval_val = getattr(args, "checkpoint_interval", 30.0)
+    if checkpoint_interval_val != 30.0:
+        sub_argv.extend(["--checkpoint-interval", str(checkpoint_interval_val)])
     if args.verbose:
         sub_argv.append("--verbose")
     if args.dry_run:
@@ -229,9 +235,7 @@ def run_route_command(args) -> int:
     if getattr(args, "placement_feedback", False):
         sub_argv.append("--placement-feedback")
     if getattr(args, "placement_feedback_budget", 3) != 3:
-        sub_argv.extend(
-            ["--placement-feedback-budget", str(args.placement_feedback_budget)]
-        )
+        sub_argv.extend(["--placement-feedback-budget", str(args.placement_feedback_budget)])
     if getattr(args, "placement_feedback_max_movement", 5.0) != 5.0:
         sub_argv.extend(
             [
@@ -240,9 +244,7 @@ def run_route_command(args) -> int:
             ]
         )
     if getattr(args, "placement_feedback_anchor", None):
-        sub_argv.extend(
-            ["--placement-feedback-anchor", args.placement_feedback_anchor]
-        )
+        sub_argv.extend(["--placement-feedback-anchor", args.placement_feedback_anchor])
     if getattr(args, "placement_feedback_no_anchor", None):
         sub_argv.extend(
             [
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -2165,6 +2165,20 @@ def _add_route_parser(subparsers) -> None:
         help="Wall-clock timeout in seconds for each per-net A* search (default: 30). "
         "Prevents individual nets from monopolizing the router. Use 0 to disable.",
     )
+    # Issue #2817: forward --checkpoint-interval through the outer parser so
+    # users can disable (``0``) or tune the best-so-far checkpoint cadence
+    # introduced by #2812.  The inner parser at route_cmd.py also declares
+    # this flag (default 30.0); both sites must stay in sync, enforced by
+    # ``tests/test_cli_parser_drift.py``.
+    route_parser.add_argument(
+        "--checkpoint-interval",
+        type=float,
+        default=30.0,
+        help=(
+            "Interval in seconds between best-so-far checkpoint writes to "
+            "--output. Default: 30. Use 0 to disable."
+        ),
+    )
     # Issue #2610: --max-search-iterations override for the C++ A* memory
     # backstop (default 0 = use the historical ``cols * rows * 4`` heuristic).
     # Documented as an escape hatch for dense boards where the cap fires
--- a/tests/test_cli_parser_drift.py
+++ b/tests/test_cli_parser_drift.py
@@ -1,0 +1,222 @@
+"""Argparse-drift regression test (Issue #2817).
+
+This test prevents a recurring class of bug where a new flag is added to
+the *inner* ``route_cmd.py`` parser without being added to the *outer*
+``parser.py`` parser and to the forwarding shim in
+``commands/routing.py``.  Three prior instances of this exact pattern
+have shipped to ``main`` and required a follow-up bugfix:
+
+* **#2620** -- ``--placement-feedback-outer-timeout`` declared outer,
+  rejected inner (drift in the opposite direction).
+* **#2622 / #2793** -- manufacturer-registry argparse drift.
+* **#2812 / #2817** -- ``--checkpoint-interval`` declared inner only;
+  ``kct route --checkpoint-interval 30`` rejected with
+  ``error: unrecognized arguments``.
+
+The drift is invisible to the type checker and to most unit tests
+because both parsers are constructed independently with no shared
+schema.  This test introspects both parsers and asserts that every
+``--flag`` accepted by the inner parser is also accepted by the outer
+parser, except for an explicit allowlist of historically inner-only
+flags.
+
+If you legitimately need to add a new inner-only flag, add it to
+``INNER_ONLY_ALLOWLIST`` below with a comment explaining why it does
+not belong on the outer parser.  In most cases the right answer is to
+add it to BOTH parsers and to the forwarding shim -- see the
+``--per-net-timeout`` block in ``commands/routing.py`` for the model
+pattern.
+"""
+
+from __future__ import annotations
+
+import argparse
+from unittest.mock import patch
+
+import pytest
+
+# Flags that legitimately exist ONLY on the inner ``route_cmd.py`` parser.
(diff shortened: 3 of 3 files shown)


rjwalters:
## Judge Review: APPROVED

**CI**: 8/8 SUCCESS. Merge state: CLEAN / MERGEABLE.

### Plumbing fix verification

- `parser.py` addition follows the `--per-net-timeout` model exactly: `type=float`, `default=30.0`, help text mentions "Use 0 to disable". Comment cross-references the drift test as the contract enforcer.
- `commands/routing.py` forwards conditionally (`if checkpoint_interval_val != 30.0`), correctly avoiding clobbering the inner-parser default. Inline comment explains the gating rationale.
- Verified `uv run kct route --help` now shows `--checkpoint-interval CHECKPOINT_INTERVAL` in the usage line and the option list.
- Cosmetic ruff reformat of two `placement_feedback_*` blocks is incidental but harmless.

### Drift test verification (the long-term value)

- Ran `pytest tests/test_cli_parser_drift.py -v`: **3/3 passed** in 13.56s.
- Ran the existing test suites the PR claims to preserve (`test_route_checkpoint.py`, `test_cli.py`, `test_cli_commands.py`): **104/104 passed**. No scope creep.
- The `unittest.mock.patch.object(argparse.ArgumentParser, "parse_args", ...)` approach is unusual but well-bounded: the patch is scoped via `with` and the fake only intercepts when `prog == "kicad-tools route"`, falling through to the real implementation otherwise. No risk of leaking into other tests.
- The `prog`-based filter handles the case where `route_main([])` instantiates intermediate sub-parsers — clean approach.
- The `SystemExit(0)` exit-before-routing guarantees no side effects (no PCB I/O, no router invocation).

### Allowlist review (32 entries)

Spot-checked the INNER_ONLY_ALLOWLIST. Two semantic groups, both with inline section comments explaining the WHY:

1. **Diagnostics/debug/profiling** (10 entries): `--analyze`, `--diagnostics`, `--profile`, `--profile-output`, `--cache-stats`, `--cache-only`, `--clear-cache`, `--preview`, `--show-congestion`, `--format`. All clearly internal — none are user-facing routing controls.
2. **Dev-experimental router knobs** (22 entries): `--two-phase`, `--escape-routing`, `--auto-pour`, `--bus-*`, `--multi-resolution`, etc. Reasonable to keep these inner-only until promoted.

No production-facing flags hiding in the allowlist. The section comment "future flags should generally NOT be added here -- expose them through the outer parser instead" sets the right defaults for future contributors.

### Allowlist staleness guard

`test_allowlist_entries_are_actually_inner_only` correctly checks both staleness modes:
1. Flag dropped from inner parser entirely (`not in inner`)
2. Flag promoted to outer parser (`in outer` — should be removed from allowlist)

Both branches generate clear failure messages with remediation pointers.

### Mental verification of "fails on unfixed state"

If `parser.py` change is reverted: `outer` set would not contain `--checkpoint-interval`. `inner` set still contains it (added by #2812). `inner_only = {"--checkpoint-interval", ...}`, `unexpected_inner_only = {"--checkpoint-interval"}` (not in allowlist). Both `test_inner_only_flags_are_in_allowlist` and `test_checkpoint_interval_is_on_both_parsers` would fail with the precise pointer message. Builder's verification claim is structurally sound.

### Follow-up note (NOT a blocker)

Builder correctly noted opposite-direction drift: `--max-search-iterations` is on the outer parser but is NOT forwarded by `commands/routing.py` (verified: no grep hits in the shim). The drift test as written only catches inner-only orphans, not outer-only-not-forwarded. A future enhancement could add a third assertion: every outer flag that is not in an OUTER_ONLY_ALLOWLIST (e.g., `--quiet`, `-q`, things consumed by the wrapper itself) must appear in the forwarding shim. Worth filing as a follow-up issue. Out of scope for this PR per the AC.

### Summary

- Immediate plumbing fix: correct, surgical, additive.
- Drift regression test: well-designed, pins the right invariant, has clear failure messages, ships with staleness guard, no test pollution risk.
- 17 prior `test_route_checkpoint.py` tests + 87 `test_cli*.py` tests still green.

This is the right kind of fix for a third-instance recurring bug — the structural test will pay dividends.

LGTM, approved.

PR #2806, opened 2026-05-12
fix(route_cmd): unify --timeout as total wall-clock budget

rjwalters:
## Summary

``--timeout`` is now a **total** wall-clock budget for the whole ``kct route`` invocation rather than a per-stage budget that every orchestration loop re-applies independently.  The previous behaviour could exceed the user-configured timeout by 5-10x when ``--auto-layers`` + ``--placement-feedback`` + ``--auto-fix`` were combined (issue #2802 reported 64+ min wall-clock with ``--timeout 2700``).

A single monotonic deadline is stamped onto ``args`` once in ``main()`` and threaded through every outer-loop site via four new helpers.

## Changes

- **New helpers in ``src/kicad_tools/cli/route_cmd.py``**:
  - ``_set_wall_clock_deadline(args)`` — called once after argparse; stamps ``args._wall_clock_deadline`` when ``--timeout`` is set.
  - ``_remaining_budget(args)`` — seconds left vs. the deadline; ``None`` when no deadline is configured.
  - ``_deadline_expired(args)`` — boolean shortcut used at outer-loop tops.
  - ``_budgeted_timeout(args)`` — ``min(args.timeout, _remaining_budget(args))`` for inner router calls.
- **Outer-loop guards** (skip remaining work when the deadline fires):
  - ``route_with_layer_escalation`` — top of the ``for attempt_num`` loop.
  - ``route_with_rule_relaxation`` — top of the tier loop.
  - ``route_with_combined_escalation`` — top of both the layer outer-loop and the tier inner-loop of the 2D search.
- **Inner-call shortening**: every ``router.route_all_negotiated`` / ``route_all_two_phase`` / ``route_with_escape`` / ``route_all_multi_resolution`` / ``route_all_evolutionary`` / ``route_with_progressive_clearance`` call inside the orchestration helpers now passes ``timeout=_budgeted_timeout(args)`` instead of raw ``args.timeout``.  This includes both the layer/rule/combined escalation loops and the single-pass main routing path.
- **Placement-feedback hook** (``_run_placement_feedback``):
  - Early bail when the deadline has already expired (returns ``None`` without loading the PCB or invoking the loop).
  - The PF outer cap is now ``min(--placement-feedback-outer-timeout, _remaining_budget(args))`` so the existing #2606 plumbing handles the time-remaining when the loop is allowed to run.
  - The inner per-call ``timeout`` forwarded into ``route_with_placement_feedback`` is ``_budgeted_timeout(args)``.
[...]

PR #2824, opened 2026-05-13
fix(router): per-attempt budget for layer escalation (closes #2823)

### LINKED ISSUE #2823: router: layer escalation races wall-clock; should reserve per-attempt budget instead of greedy until deadline
## Observation

`kct route --timeout 1200 --auto-layers --placement-feedback --iterations 30` on board 05 optimized:

- 2-layer attempt: completed, 17/35 = 49%
- 4-layer attempt 1: started, **wall-clock deadline reached before attempt 3; stopping layer escalation (issue #2802)**
- Final: returned 49% on 2 layers
- Never tried 6 layers

The router obeys `--timeout` (good — that was #2802). But within the budget, it greedily lets the first attempt eat most of the budget, leaving nothing for higher layer counts that might have unblocked the remaining nets.

## Why this matters

The escalation strategy in `--auto-layers` is supposed to try increasing layer counts until success. If the first attempt at each layer count is wall-clock-dominated, the strategy degenerates to "spend everything on layer 2, give up." That defeats the purpose of escalation.

Reproducer hit twice in two days on board 05.

## Root Cause

The wall-clock deadline helpers added in PR #2806 (issue #2802) act as a **hard stop** between attempts, not as a **per-attempt budget allocator**. The relevant code:

- `src/kicad_tools/cli/route_cmd.py:129-148` — `_budgeted_timeout(args)` returns `min(args.timeout, remaining_budget)`. The docstring is explicit: *"this is the smaller of the original `--timeout` (preserving per-stage semantics for the first stage) and the remaining wall-clock budget (so the final stage shortens naturally as time runs out)."* So the **first** escalation attempt receives the full `--timeout` value as its per-call budget, and only **later** stages get a smaller slice.

- `src/kicad_tools/cli/route_cmd.py:1976` — inside `route_with_layer_escalation`, each attempt calls `_attempt_timeout = _budgeted_timeout(args)` and forwards it into `route_all_negotiated`/`route_with_escape`/etc. (lines 1982, 1989, 1995, 2002, 2021).

- `src/kicad_tools/cli/route_cmd.py:1886-1898` — the loop checks `_deadline_expired(args)` between attempts and prints the observed message (`Wall-clock deadline reached before attempt {attempt_num}; stopping layer escalation (issue #2802)`).

- `src/kicad_tools/cli/route_cmd.py:3185` — `route_with_combined_escalation` (the `--auto-layers --adaptive-rules` 2D-search path) uses the same `_budgeted_timeout` helper and has the same problem; deadline guards at lines 3107-3115 and 3119-3128.

So when `--timeout 1200 --auto-layers` is configured on a hard board, attempt 1 (2L) is allowed to spend the *entire* 1200s budget; attempt 2 (4L) inherits whatever is left (often <60s); attempt 3 (6L) never starts.

The escalation ladder is fixed at `src/kicad_tools/cli/route_cmd.py:1849-1854` (mirrored at `:3060-3065`):

```python
layer_configs = [
    (2, LayerStack.two_layer()),
    (4, LayerStack.four_layer_sig_gnd_pwr_sig()),
    (4, LayerStack.four_layer_all_signal()),
    (6, LayerStack.six_layer_sig_gnd_sig_sig_pwr_sig()),
]
# Filter by max_layers
layer_configs = [(n, s) for n, s in layer_configs if n <= args.max_layers]
```

After `--max-layers 6` filtering this is 4 entries; after `--max-layers 4`, 3 entries.

### Proposal: per-attempt budget

Reserve per-attempt budget when `--auto-layers` is active:

1. After filtering `layer_configs` by `args.max_layers`, compute:
   ```python
   max_attempts = max(1, len(layer_configs))
   per_attempt_budget = (
       args._wall_clock_deadline - time.monotonic()
   ) / max_attempts if args._wall_clock_deadline else None
   ```
2. Replace `_attempt_timeout = _budgeted_timeout(args)` (line 1976, 3185) with a new helper, e.g. `_per_attempt_timeout(args, max_attempts, attempt_index)`, that returns:
   ```python
   min(
       args.timeout,                         # never exceed user's original cap
       remaining_budget,                     # never overrun deadline
       per_attempt_budget,                   # fair slice
   )
   ```
3. After each attempt, if any layer count completed at >=`min_completion`, stop (current behavior — keep).
4. If wall-clock exhausted, return the best-of-attempts (current behavior — keep).
5. Optional refinement: if attempt N finishes well under its slice (e.g. early SUCCESS or early stall detection), roll the unused remainder forward into the remaining-attempt pool. This keeps the no-`--timeout` and generous-`--timeout` cases identical to today.

### Backwards compatibility

- When `--timeout` is **not set**: `_remaining_budget(args)` returns `None`, so the new helper falls through to `args.timeout` (also None). No behavior change.
- When `--timeout` is **set generously** (i.e. larger than the natural per-attempt time): the per-attempt cap won't bind; behavior matches today.
- When `--timeout` is **set tight** (the bug case): the budget is sliced fairly across attempts. First attempt may finish faster (or be force-stopped); higher layer counts get a real chance.

### Where to change

- `src/kicad_tools/cli/route_cmd.py:129-148` — extend or replace `_budgeted_timeout` with a per-attempt-aware variant. Keep `_budgeted_timeout` for non-escalation callers (lines 1189, 5235, 5262, ... — placement-feedback and the single-stage fast paths).
- `src/kicad_tools/cli/route_cmd.py:1886` — pass `attempt_num` and `len(layer_configs)` into the new helper at line 1976.
- `src/kicad_tools/cli/route_cmd.py:3106` — same for `route_with_combined_escalation`. Note the 2D nature: `max_attempts` here is `len(layer_configs) * len(tiers)`, not just `len(layer_configs)`.

## Acceptance Criteria

- [ ] When `--timeout T --auto-layers` is configured with a `T` that is short relative to first-attempt natural runtime, at least 2 layer-count attempts are started (not just 1) before the deadline expires, on a fixture where 2L cannot complete.
- [ ] When `--timeout` is unset, `kct route --auto-layers` behaviour is bit-identical to today on the existing routing fixtures (no regression in `tests/router/test_layer_escalation.py`).
- [ ] When `--timeout T` is set with `T` >> natural first-attempt time, the first attempt is allowed to run to natural completion (no premature cap) — i.e. the per-attempt slice is an *upper* bound, not a target runtime.
- [ ] The "Wall-clock deadline reached before attempt N" message in `route_cmd.py:1894-1897` only fires when the *total* deadline has expired, not when a per-attempt slice has been consumed (per-attempt exhaustion should advance to the next layer-count, not stop the ladder).
- [ ] Same behavior in `route_with_combined_escalation` (the `--auto-layers --adaptive-rules` path).

### Unit test

In `tests/cli/test_route_cmd_timeout.py` (new or existing):

1. Mock `_set_wall_clock_deadline(args)` and `time.monotonic` to simulate a 30-second budget with `args.timeout = 30`, `args.max_layers = 6`, so `len(layer_configs) == 4` and per-attempt budget = 7.5s.
2. Assert the new per-attempt helper returns `<= 7.5` for attempt 1, and never returns more than `args.timeout`.
3. Assert that with `args.timeout = None`, the helper returns `None` (no cap).

### Integration test

In `tests/router/test_layer_escalation.py`, add a test using a small fixture known to exhaust 2L (e.g. board-05's chorus-test reduced fixture, or a synthetic congested board):

```python
def test_auto_layers_per_attempt_budget(tmp_path, congested_fixture_pcb):
    args = make_route_args(
        timeout=30,
        auto_layers=True,
        max_layers=6,
        iterations=10,
    )
    log = capture_attempts(...)
    assert log.attempts_started >= 2, (
        f"Expected escalation to start >=2 attempts; only started "
        f"{log.attempts_started}. First attempt consumed full budget."
    )
```

### Manual reproduction

```bash
uv run kct route /tmp/board05_optimized.kicad_pcb \
  --output /tmp/route_test.kicad_pcb --mfr jlcpcb-tier1 \
  --timeout 1200 --auto-layers --iterations 30
```

Watch the log for the attempt sequence. Before fix: `Attempt 1: 2 layers ...` runs ~1100s, `Attempt 2: 4 layers ...` runs <100s, attempt 3 never starts. After fix: ~300s per attempt, all 4 attempts start, best-of-4 is returned.

## Related

- #2802 (closed) — wall-clock honored; this issue is about how the budget is **distributed**, not whether it's honored.
- #2806 — added `_budgeted_timeout` / `_remaining_budget` helpers (the foundation this issue extends).

## Suggested labels

bug, router, layer-escalation

## Surfaced from

Board 05 placement-optimization experiment 2026-05-12; observed twice this session.

rjwalters:
## Summary

Closes #2823.

The wall-clock deadline helpers added in #2806 act as a hard stop between escalation attempts, not as a per-attempt budget allocator.  Because `_budgeted_timeout` returns `min(args.timeout, remaining)` and intentionally preserves "first-stage semantics", the first escalation attempt greedily consumes the entire `--timeout`, leaving no budget for higher-layer attempts.

In `--auto-layers` mode this means a hard board with `--timeout 1200` spends ~1100s on 2 layers, ~100s on 4 layers, and never tries 6 layers -- defeating the purpose of escalation.  Reproducer hit twice on board 05.

## Fix

Add new helper `_per_attempt_budgeted_timeout(args, attempt_index, max_attempts)` that divides the remaining wall-clock budget evenly across remaining attempts and returns `min(args.timeout, remaining, per_attempt_slice)`.

Keep `_budgeted_timeout` for the ~14 non-escalation callers (placement-feedback iterations, `route_with_strategy` inner calls, etc.) where first-stage semantics are still correct.

Wire the new helper into all three escalation loops:
- `route_with_layer_escalation` (1D, `len(layer_configs)` attempts)
- `route_with_combined_escalation` (2D, `len(layer_configs) * len(tiers)` cells)
- `route_with_rule_relaxation` (1D, `len(tiers)` tiers)

## Backward compatibility

- When `--timeout` is **unset**: helper returns `None` (legacy unbounded, identical to `_budgeted_timeout`).
- When `--timeout` is **set generously**: the per-attempt cap is an upper bound, never a target runtime; inner routers that finish early roll their unused budget forward automatically (helper recomputes from current `remaining_budget` on each call).
- When `--timeout` is **set tight** (the bug case): budget is sliced fairly across attempts so higher-layer attempts get a real chance.

## Test plan

New unit test class `TestPerAttemptBudgetedTimeout` (9 tests) covers:
- Legacy unset-timeout path returns `None`.
- First attempt with `--timeout 30` and 4 attempts gets ~7.5s slice (NOT the full 30s) -- direct regression guard.
- Per-attempt slice never exceeds `args.timeout`.
- Per-attempt slice never exceeds `remaining_budget`.
- Later attempts get an increasing slice as `remaining_attempts` shrinks.
- `max_attempts=1` collapses to legacy behaviour.
- Defensive paths: `max_attempts=0` and `attempt_index >= max_attempts` clamp to `remaining_attempts >= 1` (no division-by-zero).

New `TestLayerEscalationPerAttemptBudget` class (5 tests) provides structural call-site assertions:
- All three escalation functions invoke `_per_attempt_budgeted_timeout`.
- `route_with_layer_escalation` passes `attempt_num - 1` and `len(layer_configs)`.
- `route_with_combined_escalation` passes the 2D cell count `len(layer_configs) * len(tiers)`.
- `route_with_rule_relaxation` is wired up too.
- The legacy `_budgeted_timeout` is still in use by single-stage callers (so the per-attempt helper does not accidentally replace placement-feedback / inner route_with_strategy call sites).

- [x] All 32 tests in `tests/test_route_cmd_total_timeout.py` pass.
- [x] No regression in 116 related route_cmd tests (`test_route_cmd_params.py`, `test_route_cmd_placement_feedback.py`, etc.).
- [x] Pre-existing 9 `TestEarlyTermination` failures in `tests/test_layer_escalation.py` are unrelated (also fail on main; mocking issue with `validate_routes`).
- [x] Lint clean (`ruff check`, `ruff format`).

## Manual reproduction (post-merge verification)

```bash
uv run kct route /tmp/board05_optimized.kicad_pcb \
  --output /tmp/route_test.kicad_pcb --mfr jlcpcb-tier1 \
  --timeout 1200 --auto-layers --iterations 30
```

Watch the log for the attempt sequence.
- Before fix: `Attempt 1: 2 layers ...` runs ~1100s, `Attempt 2: 4 layers ...` runs <100s, attempt 3 never starts.
- After fix: ~300s per attempt, all 4 attempts start, best-of-4 is returned.

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -148,6 +148,89 @@ def _budgeted_timeout(args) -> float | None:
     return min(float(timeout), remaining)
 
 
+# =============================================================================
+# Issue #2823: Per-attempt budget allocation for escalation loops
+# =============================================================================
+#
+# ``_budgeted_timeout`` (above) intentionally preserves "first-stage semantics":
+# the *first* escalation attempt is allowed to consume the full ``--timeout``
+# value as its per-call budget, and only later attempts get a smaller slice as
+# the wall-clock deadline approaches.  This is correct for *single-stage* call
+# sites (placement-feedback iterations, the inner route in
+# ``route_with_strategy``, etc.), but it is *wrong* for the multi-attempt
+# layer-escalation and combined-escalation loops.
+#
+# In ``--auto-layers`` mode the orchestration tries 2L, 4L, 4L-all-sig, then 6L
+# in sequence.  When ``--timeout`` is set tight relative to the natural runtime
+# of the first attempt, the greedy allocation gives the entire budget to the 2L
+# attempt, leaves nothing for 4L, and never starts 6L.  The escalation strategy
+# degenerates to "spend everything on the lowest layer count, give up."
+#
+# The fix is a *per-attempt* helper that divides the remaining wall-clock
+# budget evenly across the remaining attempts, returning the minimum of:
+#   1. ``args.timeout``          - never exceed user's original cap
+#   2. ``remaining_budget``      - never overrun the total deadline
+#   3. ``per_attempt_budget``    - fair slice across remaining attempts
+#
+# When ``--timeout`` is unset, the helper falls through to ``None`` (legacy
+# unbounded behaviour, identical to ``_budgeted_timeout``).  When ``--timeout``
+# is set generously (i.e. larger than the natural per-attempt runtime), the
+# per-attempt cap is an *upper bound* not a target runtime, so the inner
+# router can still finish early on its own.
+
+
+def _per_attempt_budgeted_timeout(args, attempt_index: int, max_attempts: int) -> float | None:
+    """Return the per-call timeout for one attempt of an escalation loop.
+
+    Unlike :func:`_budgeted_timeout` (which lets the first stage consume the
+    full ``--timeout`` value), this helper divides the *remaining* wall-clock
+    budget evenly across the *remaining* attempts and returns the minimum of
+    that fair slice, the original ``--timeout``, and the remaining budget.
+
+    Args:
+        args: Parsed CLI namespace; must carry ``timeout`` and (after
+            ``_set_wall_clock_deadline``) ``_wall_clock_deadline``.
+        attempt_index: 0-based index of the current attempt within the
+            escalation loop.  Determines how many attempts are still
+            outstanding (``max_attempts - attempt_index``).
+        max_attempts: Total number of attempts the escalation loop intends
+            to run.  For 1D layer escalation this is ``len(layer_configs)``;
+            for the 2D combined escalation this is
+            ``len(layer_configs) * len(tiers)``.
+
+    Returns:
+        ``None`` when no wall-clock deadline is configured (legacy unbounded
+        behaviour, matches :func:`_budgeted_timeout`).  Otherwise a positive
+        float bounded by ``args.timeout`` and the remaining wall-clock
+        budget, fairly sliced across remaining attempts so later attempts
+        also get a real chance to run.
+
+    Notes:
+        - When ``max_attempts <= 1`` this collapses to :func:`_budgeted_timeout`
+          (no fair-slicing needed for a single attempt).
+        - Unused budget from a fast-finishing attempt rolls forward
+          automatically: the next call uses the *current* ``remaining_budget``
+          divided by the *new* ``remaining_attempts`` count, so an early
+          completion on attempt N enlarges the slice available to N+1.
+    """
+    timeout = getattr(args, "timeout", None)
+    remaining = _remaining_budget(args)
+    if remaining is None:
+        # No deadline configured -> legacy unbounded behaviour.
+        return timeout
+
+    # Attempts still outstanding *including* the current one.
+    remaining_attempts = max(1, max_attempts - attempt_index)
+    per_attempt_slice = remaining / remaining_attempts
+
+    if timeout is None:
+        # Defensive: ``_wall_clock_deadline`` is derived from ``args.timeout``,
+        # so this branch is unreachable in practice.
+        return per_attempt_slice
+
+    return min(float(timeout), remaining, per_attempt_slice)
+
+
 def _emit_single_pad_net_warning(
     router: "Autorouter",
     single_pad_nets: list[int],
@@ -1968,12 +2051,19 @@ def route_with_layer_escalation(
 
         escape_flag = _resolve_escape_routing_flag(args)
 
-        # Issue #2802: shorten this attempt's timeout to the remaining
-        # wall-clock budget so the final layer-stack attempt does not
-        # blow past ``--timeout`` after earlier attempts have already
-        # consumed most of it.  Falls back to ``args.timeout`` when no
-        # deadline is configured.
-        _attempt_timeout = _budgeted_timeout(args)
+        # Issue #2823: divide the remaining wall-clock budget fairly across
+        # the remaining layer-escalation attempts.  Without this, the first
+        # attempt (typically 2L) greedily consumes the entire ``--timeout``,
+        # leaving the higher-layer attempts (4L, 6L) no time to run -- so
+        # the escalation strategy degenerates to "spend everything on the
+        # lowest layer count, give up."  ``attempt_num`` is 1-based; the
+        # helper expects a 0-based index, hence ``attempt_num - 1``.
+        # Falls back to ``args.timeout`` when no deadline is configured.
+        _attempt_timeout = _per_attempt_budgeted_timeout(
+            args,
+            attempt_index=attempt_num - 1,
+            max_attempts=len(layer_configs),
+        )
 
         try:
             if _should_use_escape_routing(router, escape_flag, quiet):
@@ -2569,7 +2659,11 @@ def route_with_rule_relaxation(
     prev_sigint = signal.signal(signal.SIGINT, _handle_interrupt)
     prev_sigterm = signal.signal(signal.SIGTERM, _handle_interrupt)
 
-    for tier in tiers:
+    # Issue #2823: precompute total tier count so per-attempt budget can
+    # divide the remaining wall-clock budget fairly across all tiers
+    # rather than letting tier 0 greedily consume the full ``--timeout``.
+    _relaxation_max_attempts = max(1, len(tiers))
+    for _tier_idx, tier in enumerate(tiers):
         # Issue #2802: honor the total wall-clock deadline before starting
         # another rule-relaxation tier.
         if _deadline_expired(args):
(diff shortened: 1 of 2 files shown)


rjwalters:
## Loom Judge: Approved

Thorough verification against the 8 review criteria — all pass.

### CI Status
All 8 checks green; mergeStateStatus = CLEAN, mergeable = MERGEABLE.

### Verification of review criteria

**1. `_per_attempt_budgeted_timeout` correctly slices wall-clock budget** — Yes. Returns `min(args.timeout, remaining, remaining / max(1, max_attempts - attempt_index))`. Crucially this is NOT `min(args.timeout, remaining)` (the buggy `_budgeted_timeout` semantics); the third term (the fair slice) is what fixes the original bug.

**2. `_budgeted_timeout` left untouched** — Verified. The legacy helper remains intact at `src/kicad_tools/cli/route_cmd.py:129-148` with original "first-stage semantics" docstring. The new `test_legacy_budgeted_timeout_still_used_by_non_escalation_callers` enforces ≥5 surviving callers (placement-feedback, route_with_strategy, etc.). No risk to the ~14 non-escalation call sites.

**3. Backwards compat when `--timeout` unset** — Verified. When both `timeout` and `_wall_clock_deadline` are None, `_remaining_budget` returns None, the helper short-circuits to `return timeout` (which is None). Bit-identical to the legacy unbounded path. The `test_returns_none_without_deadline` test guards this.

**4. 2D combined-escalation uses `len(layer_configs) * len(tiers)`** — Verified at line 3211. The row-major linearization `_combined_attempt_index = _layer_idx * len(tiers) + _tier_idx` (line 3240) is correct: for 4 layer configs × 4 tiers = 16 cells, the budget is now sliced across the full 16-cell matrix rather than only the 4 cells of the current column. The `test_combined_escalation_call_site_uses_per_attempt_helper` structural test asserts this.

**5. `route_with_rule_relaxation` bonus fix is justified, not scope creep** — Justified. The issue body at #2823 explicitly identified this as the same class of bug in all multi-attempt escalation loops. The fix applies identical mechanics (precompute total tier count, pass linear index). Comments at lines 2739-2744 explicitly cite issue #2823. Applying the fix selectively to two of three loops would have left a known-broken third path; this is good engineering hygiene.

**6. Pre-existing 9 `TestEarlyTermination` failures unrelated** — Verified independently. Stashed PR changes, restored `tests/test_layer_escalation.py` and `src/kicad_tools/cli/route_cmd.py` from main, ran the test class: identical 9 failures with identical names (`test_zero_overflow_stops_early`, `test_stagnation_stops_early`, etc.). These are a pre-existing `validate_routes` mocking issue on main, not a regression from this PR.

**7. Regression guard test** — `test_first_attempt_gets_fair_slice_not_full_budget` directly targets the bug: with `timeout=30.0` and `max_attempts=4`, asserts `budgeted <= 7.5 + 0.1`. Before this fix, `_budgeted_timeout` would have returned 30.0 here. This test would have caught the original bug.

**8. Code quality** — Excellent. Defensive guards (`max(1, max_attempts - attempt_index)` prevents div-by-zero; `attempt_index >= max_attempts` clamps to 1); thorough docstrings citing issue numbers and call sites; no dead code; consistent naming (`_per_attempt_budgeted_timeout` mirrors `_budgeted_timeout`); ruff-clean.

### Local test verification
- 14/14 new tests pass (`TestPerAttemptBudgetedTimeout` 9 + `TestLayerEscalationPerAttemptBudget` 5)
- 9 `TestEarlyTermination` failures confirmed identical on main (NOT a regression)

LGTM. Approved.

Issue #2822, opened 2026-05-13
placement: CMA-ES cost is uniform-per-net; perimeter-anchored signals (connectors) get starved


rjwalters:
## Observation

CMA-ES placement on board 05 traded **+6 power/gate signals** for **-5 perimeter sensor signals**:

| Newly routed (after optim) | Newly broken (after optim) |
|---|---|
| +3.3V, +5V (multi-fan supply rails) | HALL_A, HALL_B, HALL_C |
| GATE_AL/BL/CL (low-side gates) | ISENSE_A-, ISENSE_B- |
| GATE_DRV_AH/BH/CH | PWM_AL, SW_OUT |
| SWCLK, SWDIO, PWR_LED | |

**Net delta:** 16/40 → 22/40 (+6), but with topology shift, not pure improvement.

## Root cause hypothesis

Hall sensors (HALL_A/B/C) are physically anchored — one pad on J3 (perimeter connector) and one pad on U10 (peripheral). The optimizer can't move J3 (it's an edge-mounted connector); when it shifts U10 to optimize the U3 cluster wirelength, it stretches the HALL net length until the channel is too crowded to route.

Same pattern for ISENSE pairs (anchored to Q2/Q4/Q6 sense FETs near board edge) and PWM_AL (anchored to gate pins).

Cost function appears to weight all nets equally by total wirelength. But signals **anchored to fixed components** (connectors, perimeter parts) have less degrees of freedom and should weight heavier.

## Reproduce

Same recipe as #2821. Inspect baseline-vs-optimized completed-net diff.

---

### Confirmed: cost is uniform per-net (HPWL sum, no weighting)

`compute_wirelength` in `src/kicad_tools/placement/cost.py:189-224` sums half-perimeter wirelength across all nets equally. Every net contributes `(max(xs) - min(xs)) + (max(ys) - min(ys))` with no per-net multiplier. The aggregate is then folded into the score by a single scalar `wirelength_weight` (default `1.0`) in `_weighted_sum_score` (`cost.py:590-600`) and `_lexicographic_score` (`cost.py:603-632`).

There is **no existing per-net weighting infrastructure** in the cost module. `PlacementCostConfig` (`cost.py:31-53`) has only global weights (`overlap_weight`, `drc_weight`, `boundary_weight`, `wirelength_weight`, `area_weight`, `block_boundary_weight`).

### Locked-footprint detection already exists

[...]

PR #2830, opened 2026-05-13
fix(schema): propagate Footprint.locked + 4 sibling fields on PCB.save

### LINKED ISSUE #2827: schema: PCB.save drops Footprint.locked (parser sets it, serializer doesn't write it)
## Bug

`Footprint.locked` (and the related `dnp`, `exclude_from_pos_files`, `exclude_from_bom`, and even the parent `attr` type itself) are parsed correctly from KiCad's `(attr ...)` block (per `src/kicad_tools/schema/pcb.py:554-566`) but `PCB.save()` does NOT propagate Python-side mutations of these fields back to the underlying S-expression tree. Setting `fp.locked = True` (or any of the other four flags) in Python and saving is a silent no-op — the tokens never appear in the output file.

## Reproducer

```python
pcb = PCB.load('board.kicad_pcb')
for fp in pcb.footprints:
    if fp.reference == 'J3':
        fp.locked = True
pcb.save('board.kicad_pcb')
# Re-load and check — fp.locked is False again (and no (locked) token in file)
```

Confirmed during board-05 anchor-weight experiment 2026-05-12. Workaround required raw sexp injection of `(attr smd locked)` instead.

## Root cause

`Footprint` does **not** have a `to_sexp` method. `PCB.save()` (`pcb.py:4374`) writes `self._sexp` — the original parsed S-expression tree — directly via `save_pcb()`. Mutations to a `Footprint` only reach the saved file when they pass through the `Footprint.__setattr__` override (`pcb.py:475-507`), which currently syncs **only** `position`, `rotation`, and `layer` to `_sexp_node`.

The five `(attr ...)`-block fields parsed at `pcb.py:554-566` have no corresponding write-back path:

| Field | Parsed at | Synced to `_sexp_node`? |
|---|---|---|
| `attr` (smd / through_hole) | `pcb.py:555` | No |
| `exclude_from_pos_files` | `pcb.py:559-560` | No |
| `exclude_from_bom` | `pcb.py:561-562` | No |
| `locked` | `pcb.py:563-564` | No |
| `dnp` | `pcb.py:565-566` | No |

All five suffer the same silent-data-loss bug: load a PCB that has the flag, mutate any other field (e.g. position), save — the original `(attr ...)` token is preserved (because `_sexp_node` was never touched), but any **Python-side change** to the flag is dropped. Setting a flag from `False` to `True` does not write it; clearing a flag from `True` to `False` does not remove it.

## Impact

- Programmatic locking (e.g., for `--anchor-weight` workflows from #2822) is impossible without raw sexp manipulation.
- Any code path that constructs a `Footprint` from scratch (without an `_sexp_node` back-reference) and adds it to a PCB will lose all five flags on save. (Note: footprints loaded from a file and only mutated do not lose the flag's *original* value — the existing `(attr ...)` token survives untouched — but any Python-side mutation is silently dropped.)
- Quiet data loss: no error, no warning. Failure mode is "the flag I just set didn't take" with no diagnostic.

### Approach: extend `Footprint.__setattr__` to sync the `(attr ...)` block

Mirror the existing pattern used for `position` / `rotation` / `layer` at `src/kicad_tools/schema/pcb.py:475-507`. Add a single helper `_sync_attr_node(self)` that rebuilds (or removes) the `(attr ...)` child of `_sexp_node` from the current Python state, and call it from `__setattr__` whenever any of the five fields are assigned.

Pseudocode:

```python
# In Footprint.__setattr__, after the existing position/rotation/layer branches:
if name in ("attr", "locked", "dnp",
            "exclude_from_pos_files", "exclude_from_bom"):
    self._sync_attr_node()

def _sync_attr_node(self) -> None:
    sexp_node = self.__dict__.get("_sexp_node")
    if sexp_node is None:
        return
    # Remove any existing (attr ...) so we can rebuild it cleanly.
    existing = sexp_node.find("attr")
    if existing is not None:
        sexp_node.remove(existing)
    # Only emit (attr ...) if there is something to say.
    if not (self.attr or self.locked or self.dnp
            or self.exclude_from_pos_files or self.exclude_from_bom):
        return
    attr_node = SExp.list("attr")
    if self.attr:
        attr_node.add(self.attr)  # 'smd' or 'through_hole'
    if self.locked:
        attr_node.add("locked")
    if self.dnp:
        attr_node.add("dnp")
    if self.exclude_from_pos_files:
        attr_node.add("exclude_from_pos_files")
    if self.exclude_from_bom:
        attr_node.add("exclude_from_bom")
    # Insert near the top of the footprint block (after layer/uuid/at)
    # to match KiCad's canonical ordering. Falling back to append() is
    # acceptable — KiCad tolerates ordering.
    sexp_node.append(attr_node)  # or use insert_after("tags", attr_node)
```

The `SExp` API needed is already available: `SExp.list()` (`sexp/parser.py:835`), `add()` (`sexp/parser.py:778`), `find()` (`sexp/parser.py:179`), `remove()` (`sexp/parser.py:298`), `insert_after()` (`sexp/parser.py:252`).

### Token ordering caveat

KiCad's canonical order inside `(attr ...)` is: `<type> [board_only] [exclude_from_pos_files] [exclude_from_bom] [allow_missing_courtyard] [dnp]`, with `locked` appearing as a peer token. The parser currently only recognises five tokens — be careful not to clobber unrecognised tokens (e.g. `board_only`, `allow_missing_courtyard`, `allow_soldermask_bridges`) when rebuilding. Two options:

1. **Preserve unknowns**: when rebuilding the node, copy through any child token the parser doesn't recognise. This is the safest path.
2. **Extend the parser**: add the missing tokens to both `from_sexp` (lines 554-566) and the `Footprint` dataclass (lines 458-462). Larger scope — defer to a follow-up issue if not blocking.

Option (1) is recommended for this issue's scope. A simple way: snapshot the list of unknown child token strings on `from_sexp`, store on the `Footprint` (private field, e.g. `_attr_unknown_tokens: list[str]`), and re-emit them in `_sync_attr_node`.

### Footprints without `_sexp_node`

A `Footprint` constructed in pure Python (e.g. via `PCB.add_footprint()` paths that don't link back to a parsed node) will have `_sexp_node = None`. The `__setattr__` early-return (`pcb.py:481-483`) handles this — but the flag will then be dropped on save unless the `add_footprint()` code path attaches a fresh `_sexp_node`. Verify this is already handled (existing `position`/`rotation` round-trip tests at `tests/test_pcb.py:665+` imply it is, but worth a sanity check).

### Suggested CLI helper (separate but related)

There is currently **no** `kct pcb lock` command (verified via `grep` over `src/kicad_tools/cli/commands/pcb.py`). Once this bug is fixed, add:

```
kct pcb lock <board.kicad_pcb> --refs J3,U10,Q2,Q4,Q6
kct pcb unlock <board.kicad_pcb> --refs J3,U10
kct pcb lock <board.kicad_pcb> --all          # lock every footprint
```

This unblocks anchor-weight workflows (#2822) without requiring a script. It can be done in this PR or as a follow-up — recommend follow-up to keep the schema fix focused and well-scoped.

## Acceptance criteria

- [ ] Setting `fp.locked = True` on a loaded `Footprint` and calling `pcb.save()` results in a `(locked)` token inside the footprint's `(attr ...)` block in the output file.
- [ ] Setting `fp.locked = False` on a footprint that was loaded with `(locked)` and saving results in the `(locked)` token being **removed** from the output file.
- [ ] Same round-trip behaviour for `dnp`, `exclude_from_pos_files`, `exclude_from_bom`.
- [ ] Same round-trip behaviour for `attr` (changing `smd` -> `through_hole` or setting it from empty).
- [ ] Setting any flag on a footprint whose `(attr ...)` block doesn't yet exist creates the block.
- [ ] Clearing all flags on a footprint whose `(attr ...)` block existed leaves no `(attr ...)` block (or, equivalently, leaves a valid empty block KiCad accepts — pick one and document it).
- [ ] Unknown tokens inside `(attr ...)` (e.g. `board_only`, `allow_missing_courtyard`) are preserved across the round-trip — no silent data loss for tokens the parser doesn't yet model.
- [ ] No regressions in existing `position`/`rotation`/`layer` round-trip tests.

## Test plan

Add a new test file `tests/test_pcb_attr_roundtrip.py` (or extend `tests/test_pcb.py` near the existing position round-trip tests at line 2267). Tests:

1. **`test_locked_flag_round_trips`**: Create a board, add a footprint, set `fp.locked = True`, save, reload, assert `fp.locked is True` and the file contains `(locked)`.
2. **`test_locked_flag_can_be_cleared`**: Load a board with a pre-locked footprint (use `boards/01-voltage-divider/output/voltage_divider_routed.kicad_pcb` if any of its FPs are locked, or hand-craft a fixture), set `fp.locked = False`, save, reload, assert `fp.locked is False` and no `(locked)` token in the file.
3. **`test_dnp_flag_round_trips`** — same pattern for `dnp`.
4. **`test_exclude_from_pos_files_round_trips`** — same pattern.
5. **`test_exclude_from_bom_round_trips`** — same pattern.
6. **`test_attr_type_round_trips`** — change `smd` <-> `through_hole`.
7. **`test_attr_unknown_tokens_preserved`**: Hand-craft a footprint with `(attr smd board_only allow_missing_courtyard)`, load, set `fp.locked = True`, save, reload, assert `board_only` and `allow_missing_courtyard` survive AND `(locked)` appears.
8. **`test_multi_flag_combinations`**: Set two flags simultaneously (e.g. `locked` + `dnp`), round-trip, assert both survive.
9. **`test_no_attr_block_when_all_default`**: Footprint with no flags set produces no `(attr ...)` block (or equivalent KiCad-accepted minimal form).

The CLI `kct pcb lock` (if implemented in same PR) gets a smoke test in `tests/test_pcb.py` or a new `tests/test_cli_pcb_lock.py`.

## File references

- `src/kicad_tools/schema/pcb.py:432-469` — `Footprint` dataclass
- `src/kicad_tools/schema/pcb.py:475-507` — `Footprint.__setattr__` (extension point)
- `src/kicad_tools/schema/pcb.py:554-566` — parser (reference for which tokens are read)
- `src/kicad_tools/schema/pcb.py:1627-1671` — `_link_footprint_sexp_nodes` (where `_sexp_node` is attached)
- `src/kicad_tools/schema/pcb.py:4374-4396` — `PCB.save` (writes `self._sexp` directly; relies on in-place mutation)
- `src/kicad_tools/sexp/parser.py:178-298, 778-835` — `SExp` API used by the fix
- `src/kicad_tools/cli/commands/pcb.py` — where a `lock`/`unlock` subcommand would live (follow-up)
- `tests/test_pcb.py:2267-2316` — reference pattern for round-trip tests

## Surfaced from

Board 05 anchor-weight experiment 2026-05-12; follow-up to #2822 (which depends on locked footprints being a usable signal).

## Suggested labels

bug, schema, round-trip

rjwalters:
## Summary

Closes #2827.

`PCB.save()` previously dropped Python-side mutations to `Footprint.attr`, `exclude_from_pos_files`, `exclude_from_bom`, `locked`, and `dnp` because `__setattr__` only synced `position`/`rotation`/`layer` to the underlying S-expression node. Setting `fp.locked = True` and saving was a silent no-op -- the `(locked)` token never appeared in the output file, blocking programmatic locking workflows (#2822 anchor-weight) and forcing raw sexp injection workarounds.

### Changes

- **`Footprint.__setattr__` extended** with a new `_sync_attr_node()` helper that rebuilds the `(attr ...)` child of `_sexp_node` from current Python state whenever any of the five modeled fields change. Mirrors the existing position/rotation/layer sync pattern.
- **Unknown-token preservation**: A new private `_attr_unknown_tokens: list[str]` field captures attr-block tokens the parser does not yet model (`board_only`, `allow_missing_courtyard`, `allow_soldermask_bridges`, ...) at parse time and re-emits them verbatim on rebuild. Without this, mutating any modeled flag would silently strip unknown tokens.
- **Pre-existing parser fix**: KiCad emits `(attr exclude_from_pos_files exclude_from_bom)` without a type token in some cases. Previously `from_sexp` assigned the first flag to `fp.attr`. The parser now only accepts `smd`/`through_hole` as the type at index 0 and treats other first tokens as flags.
- **10 round-trip tests** in `tests/test_pcb_attr_roundtrip.py` covering all 5 flags + unknown-token preservation + multi-flag combinations + clearing all flags + a regression guard for the existing position sync.

### Out of scope (deferred to follow-up)

The Curator's guidance suggested a `kct pcb lock` / `kct pcb unlock` CLI helper. Kept out of this PR to keep the schema fix focused; can be filed as a separate issue once this lands.

## Test plan

- [x] `uv run pytest tests/test_pcb_attr_roundtrip.py -v` -- 10/10 pass
- [x] `uv run pytest tests/test_pcb.py` -- 164/164 pass (no regressions)
- [x] `uv run pytest tests/test_pcb.py tests/test_schema.py tests/test_pcb_create.py tests/test_pcb_modules.py tests/test_pcb_move_footprint.py tests/test_pcb_remove_footprint.py tests/test_pcb_sync_netlist.py tests/test_cli_pcb.py tests/test_schematic.py tests/test_schematic_models.py` -- 1011 pass, no regressions
- [ ] Judge review

### Acceptance criteria (from #2827)

- [x] Setting `fp.locked = True` results in a `(locked)` token in the output file
- [x] Setting `fp.locked = False` removes the token from the output file
- [x] Same round-trip behaviour for `dnp`, `exclude_from_pos_files`, `exclude_from_bom`
- [x] Same round-trip behaviour for `attr` (smd <-> through_hole, including empty -> set)
- [x] Setting any flag on a footprint without an existing `(attr ...)` block creates the block
- [x] Clearing all flags removes the `(attr ...)` block entirely
- [x] Unknown tokens (`board_only`, `allow_missing_courtyard`) survive round-trip
- [x] No regressions in existing position/rotation/layer round-trip tests

### DIFF
--- a/src/kicad_tools/schema/pcb.py
+++ b/src/kicad_tools/schema/pcb.py
@@ -12,7 +12,7 @@ from collections.abc import Iterator
 from dataclasses import dataclass, field
 from datetime import date
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, ClassVar
 
 logger = logging.getLogger(__name__)
 
@@ -467,6 +467,41 @@ class Footprint:
         repr=False,
         compare=False,
     )
+    # Tokens inside ``(attr ...)`` that the parser does not yet model
+    # (e.g. ``board_only``, ``allow_missing_courtyard``,
+    # ``allow_soldermask_bridges``). Preserved verbatim so that the
+    # ``__setattr__`` rebuild of the ``(attr ...)`` block does not
+    # silently drop them on round-trip.
+    _attr_unknown_tokens: list[str] = field(
+        default_factory=list,
+        repr=False,
+        compare=False,
+    )
+
+    # Set of attr-block tokens the parser models explicitly. Anything
+    # else that appears as an atom child of ``(attr ...)`` is captured
+    # in ``_attr_unknown_tokens`` and re-emitted verbatim.
+    _ATTR_KNOWN_TOKENS: ClassVar[frozenset[str]] = frozenset(
+        {
+            "smd",
+            "through_hole",
+            "exclude_from_pos_files",
+            "exclude_from_bom",
+            "locked",
+            "dnp",
+        }
+    )
+
+    _ATTR_SYNCED_FIELDS: ClassVar[frozenset[str]] = frozenset(
+        {
+            "attr",
+            "exclude_from_pos_files",
+            "exclude_from_bom",
+            "locked",
+            "dnp",
+            "_attr_unknown_tokens",
+        }
+    )
 
     # ------------------------------------------------------------------
     # __setattr__ override -- syncs position/rotation/layer to _sexp_node
@@ -505,6 +540,73 @@ class Footprint:
             layer_node = sexp_node.find("layer")
             if layer_node is not None:
                 layer_node.set_value(0, value)
+
+        elif name in self._ATTR_SYNCED_FIELDS:
+            # Any change to a field that maps into the (attr ...) block
+            # rebuilds that block from the current Python state. Unknown
+            # tokens captured at parse time (board_only,
+            # allow_missing_courtyard, allow_soldermask_bridges, ...)
+            # are preserved verbatim via ``_attr_unknown_tokens`` so we
+            # don't drop tokens the parser doesn't yet model.
+            self._sync_attr_node()
+
+    def _sync_attr_node(self) -> None:
+        """Rebuild the ``(attr ...)`` child of ``_sexp_node`` from Python state.
+
+        Mirrors the position/rotation/layer sync pattern in
+        :meth:`__setattr__`. Removes the existing ``(attr ...)`` node
+        (if any), then re-emits it from the current values of
+        ``attr``, ``exclude_from_pos_files``, ``exclude_from_bom``,
+        ``locked``, ``dnp``, and any ``_attr_unknown_tokens`` captured
+        during parse. If no flags are set and no unknown tokens are
+        present, no ``(attr ...)`` node is emitted (matches KiCad's
+        canonical "no flags" form).
+        """
+        sexp_node: SExp | None = self.__dict__.get("_sexp_node")
+        if sexp_node is None:
+            return
+
+        # Remove any existing (attr ...) so we can rebuild it cleanly.
+        # find_children() searches direct children only -- we want to
+        # avoid stripping nested (attr ...) inside fp_text/property,
+        # which don't currently exist but could appear in future
+        # KiCad versions.
+        for existing in list(sexp_node.find_children("attr")):
+            sexp_node.remove(existing)
+
+        unknown = self.__dict__.get("_attr_unknown_tokens", []) or []
+
+        # Only emit (attr ...) if there is something to say.
+        if not (
+            self.attr
+            or self.locked
+            or self.dnp
+            or self.exclude_from_pos_files
+            or self.exclude_from_bom
+            or unknown
+        ):
+            return
+
+        attr_node = SExp.list("attr")
+        # Canonical KiCad order: <type> [board_only] [exclude_from_pos_files]
+        # [exclude_from_bom] [allow_missing_courtyard] [dnp], with `locked`
+        # appearing as a peer token. We emit modeled tokens in a stable
+        # order and append unknown tokens at the end -- KiCad accepts
+        # any ordering of these atoms, so this is a safe simplification.
+        if self.attr:
+            attr_node.add(self.attr)  # 'smd' or 'through_hole'
+        if self.exclude_from_pos_files:
+            attr_node.add("exclude_from_pos_files")
+        if self.exclude_from_bom:
+            attr_node.add("exclude_from_bom")
+        if self.locked:
+            attr_node.add("locked")
+        if self.dnp:
+            attr_node.add("dnp")
+        for token in unknown:
+            attr_node.add(token)
+
+        sexp_node.append(attr_node)
 
     @classmethod
     def from_sexp(cls, sexp: SExp) -> Footprint:
@@ -552,10 +654,29 @@ class Footprint:
         if tags := sexp.find("tags"):
             fp.tags = tags.get_string(0) or ""
         if attr := sexp.find("attr"):
-            fp.attr = attr.get_string(0) or ""
-            # Parse additional attribute flags (e.g., exclude_from_pos_files, dnp)
-            for i in range(len(attr.children)):
+            # The footprint *type* token (``smd`` / ``through_hole``)
+            # is optional in KiCad's emitted form. When KiCad omits the
+            # type, the first atom is a flag (e.g. ``(attr
+            # exclude_from_pos_files exclude_from_bom)``). Detect this
+            # by only accepting known type tokens at index 0.
+            first_token = attr.get_string(0) or ""
+            if first_token in ("smd", "through_hole"):
+                fp.attr = first_token
+                token_start_idx = 1
+            else:
+                fp.attr = ""
+                token_start_idx = 0
+            # Parse additional attribute flags (e.g., exclude_from_pos_files, dnp).
+            # Tokens we don't model (board_only, allow_missing_courtyard,
+            # allow_soldermask_bridges, ...) are captured verbatim into
+            # _attr_unknown_tokens so the (attr ...) rebuild in
+            # _sync_attr_node() can re-emit them on save without
+            # silent data loss.
+            unknown_tokens: list[str] = []
+            for i in range(token_start_idx, len(attr.children)):
                 token = attr.get_string(i)
+                if token is None:
+                    continue
                 if token == "exclude_from_pos_files":
                     fp.exclude_from_pos_files = True
                 elif token == "exclude_from_bom":
@@ -564,6 +685,10 @@ class Footprint:
                     fp.locked = True
                 elif token == "dnp":
                     fp.dnp = True
+                else:
+                    unknown_tokens.append(token)
+            if unknown_tokens:
+                fp._attr_unknown_tokens = unknown_tokens
 
         # Reference and value from fp_text (KiCad 7 format)
         for fp_text_sexp in sexp.find_all("fp_text"):
--- a/tests/test_pcb_attr_roundtrip.py
+++ b/tests/test_pcb_attr_roundtrip.py
@@ -1,0 +1,401 @@
+"""Round-trip tests for ``Footprint`` ``(attr ...)`` block fields.
+
+Regression coverage for issue #2827: ``PCB.save()`` previously dropped
+Python-side mutations to ``Footprint.attr``, ``locked``, ``dnp``,
+``exclude_from_pos_files``, and ``exclude_from_bom``. The fix extends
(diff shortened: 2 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

Closes #2827. Clean, well-scoped fix with strong unknown-token safety net.

### Verification

- All 8 CI checks pass (Lint, Type, Test, C++ Build, DRC, kicad-cli round-trip, Diff-Pair, Match-Group)
- Local: 10/10 in `test_pcb_attr_roundtrip.py`, 164/164 in `test_pcb.py`, 849 across the 11-file regression suite
- Lint and type-check on changed files clean

### What I checked

1. **`_sync_attr_node()` rebuild** — Removes existing `(attr ...)` via `find_children` (direct children only — correctly scoped to avoid stripping nested attr blocks inside fp_text/property), re-emits all 5 modeled fields in stable canonical-ish order, then appends unknown tokens. Pattern mirrors the existing position/rotation/layer sync.

2. **Unknown-token preservation (the load-bearing safety property)** — `_attr_unknown_tokens` is captured verbatim at parse time and re-emitted on every rebuild. `test_attr_unknown_tokens_preserved` exercises the critical case: load `(attr smd board_only allow_missing_courtyard)`, mutate `locked = True`, save — and confirms both unknown tokens survive AND the new modeled flag appears. Without this, every fix to the silent-data-loss bug would have created a new silent-data-loss bug. Well done.

3. **Pre-existing parser bug fix justified** — KiCad's `(attr exclude_from_pos_files exclude_from_bom)` (type-less) form is a real emission pattern. Without the index-0 type-token gating fix, the old parser would have set `fp.attr = "exclude_from_pos_files"` (string-as-type) and silently dropped the semantic flag, then re-emitted nonsense on save. Scope expansion is acceptable here — fixing it inline avoids the new sync helper writing the wrong type token, and the alternative (file a separate issue, ship an incomplete fix) would have been worse.

4. **Edge case (all default)** — When no flags and no unknowns are set, no `(attr ...)` block is emitted (lines 580-588). `test_no_attr_block_when_all_default` asserts `(attr` is fully absent post-clear. Matches KiCad's canonical "no flags" form.

5. **`__setattr__` gating** — `_ATTR_SYNCED_FIELDS` is a frozenset of just the 5 user-facing fields plus `_attr_unknown_tokens`, and the sync only fires for those names. Uses `self.__dict__.get(...)` for `_sexp_node` to safely handle in-init access — same pattern as the existing position branch.

### Minor observations (non-blocking, follow-up if desired)

- `_ATTR_KNOWN_TOKENS: ClassVar[frozenset[str]]` (lines 484-493) is defined but never referenced. The parser uses an explicit `if/elif` chain in `from_sexp` instead. It's harmless documentation but technically dead — a small future cleanup could either reference it from the parser (`if token in cls._ATTR_KNOWN_TOKENS: ...` style) or remove the constant.
- The CLI helper `kct pcb lock` / `kct pcb unlock` is correctly deferred per the issue's "out of scope" guidance.

### Acceptance criteria from #2827

All 8 checked off in the PR body verified by tests:
- locked round-trip set/clear (tests 1-2)
- dnp / exclude_from_pos_files / exclude_from_bom round-trip (tests 3-5)
- attr type smd <-> through_hole and empty->set (test 6)
- creating attr block from no prior block (test 6, base PCB has none)
- clearing all flags removes block (test 9)
- unknown tokens preserved (test 7)
- no position/rotation/layer regression (test 10)

Approved.

Issue #2610, opened 2026-05-10
router: DAC_CLK consistently fails at ~67s regardless of --per-net-timeout — investigate hard-coded ceiling


rjwalters:
## Problem

On chorus-test-revA across 4 routing attempts at HEAD \`6926b47c\` and \`40f5fbd8\` (May 5 baseline + May 10 sessions route2/route3/route_pf), \`DAC_CLK\` consistently aborts at the exact same wall-clock-time (~67s) per attempt regardless of the user-specified per-net timeout:

| Run | --per-net-timeout | DAC_CLK time |
|---|---|---|
| route2 | 60s | 67.5s |
| route3 | 120s | 67.3s |
| route_pf | 120s | 67.3s |

The 60→120s setting bumped \`DAC_CLK_DIV2\`, \`SPI_SCK\`, etc., but \`DAC_CLK\` is identical to the byte. Suggests a **hard-coded internal timeout** in the negotiated/cpp pathfinder for DAC_CLK's specific search space, not the user-facing budget.

## Suspected location

Likely candidates (need investigation):
- \`src/kicad_tools/router/pathfinder.py\` — search-iteration cap independent of wall-clock
- \`src/kicad_tools/router/router_cpp.cpp\` (C++ binding) — hard-coded iteration limit
- \`src/kicad_tools/router/algorithms/negotiated.py\` — per-net iteration cap

\`DAC_CLK\` is a 3-pad net (U9:1, R7:2, U5:25) routing through the densest cluster. The C++ A* search may exhaust its open-set quota before wall-clock hits 67s.

## Repro

\`\`\`bash
# Force a high per-net timeout
uv run kicad-tools route /tmp/chorus-test-route/synced.kicad_pcb \\
  -o /tmp/routed.kicad_pcb \\
  --strategy negotiated --layers 4 --backend cpp --manufacturer jlcpcb --no-cache \\
  --timeout 1800 --per-net-timeout 600
# DAC_CLK still fails at ~67s
\`\`\`

## Expected behavior

\`--per-net-timeout\` should be the actual ceiling. If the C++ search has an internal iteration cap, it should be either (a) scaled by per-net-timeout, or (b) loud — log "DAC_CLK aborted at iteration cap (67s wall, 100k iterations)" rather than silent BLOCKED_PATH.

## Acceptance criteria

[...]

PR #2841, opened 2026-05-13
fix(cli): forward --max-search-iterations through route shim + symmetric drift test

### LINKED ISSUE #2819: router: --max-search-iterations has opposite argparse drift (outer-only, not forwarded through shim)
## Symptom

Discovered during PR #2818 work on the inner→outer drift pattern (#2817).

`--max-search-iterations` is **declared on the outer parser only** (`src/kicad_tools/cli/parser.py:2186-2197`). The forwarding shim at `src/kicad_tools/cli/commands/routing.py::run_route_command` does NOT propagate it to the inner `route_cmd.main()` parser. The inner command reads the value via `getattr(args, "max_search_iterations", 0) or 0` (lines 4923 and 5289 of `route_cmd.py`) — which **silently defaults to 0** regardless of what the user passed on the outer `kct route` command line.

This is the **opposite direction** of #2817 (which was inner-only with no outer surface). Both forms of drift need symmetric guarding, but the regression test added in PR #2818 (`tests/test_cli_parser_drift.py`) only asserts `inner_flags - outer_flags ⊆ INNER_ONLY_ALLOWLIST` — the opposite direction (`outer_flags - inner_flags ⊆ OUTER_ONLY_ALLOWLIST`) is not checked.

## Verification

Direct grep confirms the drift:

- `src/kicad_tools/cli/parser.py:2187` — outer parser declares `--max-search-iterations` on the `route` subparser (Issue #2610).
- `src/kicad_tools/cli/commands/routing.py:155-297` — `run_route_command` shim builds `sub_argv` for the inner command; **no `--max-search-iterations` block** anywhere. Compare with the `--checkpoint-interval` block at lines 187-192 (which #2818 added) and the `--per-net-timeout` block at lines 184-186 (the model pattern).
- `src/kicad_tools/cli/route_cmd.py` — no `parser.add_argument("--max-search-iterations", ...)` call in the inner `_build_parser`/`main()` parser (lines 3732-4260 contain ~40 `add_argument` calls; `--max-search-iterations` is not among them).
- `src/kicad_tools/cli/route_cmd.py:4923` and `:5289` — both read sites use `getattr(args, "max_search_iterations", 0) or 0`, which is the symptom: the defensive `getattr` masks the missing argparse declaration so the inner parser doesn't reject the missing attribute; it just always sees `0`.

End-to-end effect: `uv run kct route --max-search-iterations 50000 board.kicad_pcb` parses cleanly at the outer parser, the shim drops the flag, the inner parser never sees it, and the C++ A* runs with `max_search_iterations=0` (the historical `cols*rows*4` heuristic).

## Why this matters

1. Users invoking `kct route --max-search-iterations N` get N silently ignored — worse than rejection, because the symptom (router hits iteration cap before deadline) looks identical with and without the override.
2. The argparse-drift regression test from #2818 catches the inner-only direction but not the outer-only direction. The same allowlist mechanism needs to apply symmetrically.
3. Same recurring pattern as #2620, #2622, #2793, #2817 — this is the 5th instance of the drift class. Each fix has been a one-off; the drift test needs to broaden to catch both directions before a 6th instance lands.

## Root cause

When #2610 added `--max-search-iterations`, the flag was added to the outer parser (`parser.py`) and to the read sites in `route_cmd.py` (using defensive `getattr`), but two pieces of the standard shim pattern were skipped:

1. The inner parser at `route_cmd.py` was not given a matching `add_argument` call.
2. The `run_route_command` shim in `commands/routing.py` was not given a forwarding block.

The defensive `getattr(args, "max_search_iterations", 0) or 0` masked the bug because the inner parser doesn't raise `unrecognized arguments` (the shim doesn't pass the flag at all), and the missing attribute silently falls back to 0.

### Part A — fix the immediate plumbing

This is a real flag (Issue #2610 added it for dense boards where the C++ A* `cols*rows*4` heuristic fires before the wall-clock deadline). It has end-to-end plumbing through `router/io.py`, `router/core.py`, `router/cpp_backend.py`, and the C++ `pathfinder.cpp`. Tests at `tests/test_per_net_timeout_scaling.py:457-537` verify the override works at the C++ binding level. The fix is to wire it through the CLI shim, **not** to remove the outer-only declaration.

Two changes are required (model pattern: the `--checkpoint-interval` blocks added by PR #2818):

1. **Add the flag to the inner parser** at `src/kicad_tools/cli/route_cmd.py`. Insert next to `--checkpoint-interval` at line 3850 (the analogous Issue #2610-era flag). Use `type=int, default=0` matching the outer parser declaration at `parser.py:2187-2197`.

2. **Add a forwarding block** in `src/kicad_tools/cli/commands/routing.py::run_route_command`. Insert immediately after the `--checkpoint-interval` block at lines 187-192. Model:

   ```python
   # Issue #2819: forward --max-search-iterations to the inner parser.
   # Both defaults are 0 (= use cols*rows*4 heuristic), so only forward
   # when the user passed a non-default value.
   max_iter_val = getattr(args, "max_search_iterations", 0) or 0
   if max_iter_val:
       sub_argv.extend(["--max-search-iterations", str(max_iter_val)])
   ```

3. **Tighten the read sites** at `route_cmd.py:4923` and `:5289`. Now that the inner parser declares the flag with `default=0`, the defensive `getattr(args, "max_search_iterations", 0)` can be simplified to `args.max_search_iterations`. Leaving the `getattr` in place is also acceptable but should carry a comment noting that the inner parser guarantees the attribute exists — otherwise future readers will assume the defensive pattern is load-bearing.

### Part B — broaden the drift test (prevents the 6th instance)

`tests/test_cli_parser_drift.py:147-175` asserts only `inner_flags - outer_flags ⊆ INNER_ONLY_ALLOWLIST`. Add a symmetric test:

- Introduce `OUTER_ONLY_ALLOWLIST: frozenset[str]` near the existing `INNER_ONLY_ALLOWLIST` (line 49). Document with the same convention: each entry needs a justification comment.
- Add `test_outer_only_flags_are_in_allowlist()` that asserts `outer_flags - inner_flags ⊆ OUTER_ONLY_ALLOWLIST` and fails with a message pointing at `commands/routing.py::run_route_command` and `route_cmd.py::main` (the two sites that need updating, mirroring the existing failure message at lines 167-175).
- Add `test_allowlist_entries_are_actually_outer_only()` mirroring the existing stale-allowlist check at lines 178-201.
- Add a direct pin: `test_max_search_iterations_is_on_both_parsers()` mirroring `test_checkpoint_interval_is_on_both_parsers()` at line 204.
- Update the docstring at the top of the file (currently asserts only `inner ⊆ outer`) to describe the symmetric guard.
- Update the `INNER_ONLY_ALLOWLIST` snapshot comment at line 46 — the date stays 2026-05-12 but it now reads "and the symmetric outer-only check added by #2819".

After these changes, the test should run a snapshot of the current outer parser to determine which outer flags genuinely have no inner counterpart (e.g. flags consumed entirely by the shim — `--auto-layers`, `--no-auto-layers`, etc.) and seed `OUTER_ONLY_ALLOWLIST` with them. The fix in Part A removes `--max-search-iterations` from that snapshot.

## Files to touch

- `src/kicad_tools/cli/route_cmd.py` — add `--max-search-iterations` declaration to the inner parser (insert near line 3850, next to `--checkpoint-interval`).
- `src/kicad_tools/cli/commands/routing.py` — add forwarding block in `run_route_command` after line 192.
- `tests/test_cli_parser_drift.py` — add `OUTER_ONLY_ALLOWLIST`, symmetric tests, direct pin for `--max-search-iterations`, and updated docstring.

## Test plan

1. **Inner-parser smoke**: `uv run python -c "from kicad_tools.cli.route_cmd import main; main(['--help'])"` shows `--max-search-iterations` in the help output.
2. **End-to-end forwarding**: integration test that runs `uv run kct route --max-search-iterations 12345 board.kicad_pcb --dry-run` (or a unit-level equivalent) and asserts the inner command receives `max_search_iterations=12345` (mock or capture the call into the routing engine).
3. **Drift test pin**: `pytest tests/test_cli_parser_drift.py::test_max_search_iterations_is_on_both_parsers` passes — both parsers declare the flag.
4. **Drift test broadening**: `pytest tests/test_cli_parser_drift.py` passes (both directions now checked; `OUTER_ONLY_ALLOWLIST` accurately reflects the post-fix state).
5. **Negative test**: temporarily delete the forwarding block from `commands/routing.py`; `test_outer_only_flags_are_in_allowlist` should fail and name `--max-search-iterations` in the message.
6. **C++ override still works**: `tests/test_per_net_timeout_scaling.py::TestMaxSearchIterationsOverride` still passes (the underlying machinery is unchanged; this fix is plumbing-only).

## Acceptance criteria

1. `uv run kct route --max-search-iterations N board.kicad_pcb` takes effect — the inner C++ A* runs with `max_search_iterations=N`, not 0. Verifiable by integration test or by observing the iteration cap firing when N is small.
2. `--max-search-iterations` is declared on BOTH `parser.py` and `route_cmd.py`'s inner parser.
3. `commands/routing.py::run_route_command` forwards the flag in the `sub_argv` list using the non-default-only pattern (matching `--per-net-timeout` and `--checkpoint-interval`).
4. `tests/test_cli_parser_drift.py` has a symmetric `outer_flags - inner_flags ⊆ OUTER_ONLY_ALLOWLIST` assertion.
5. `tests/test_cli_parser_drift.py` has a direct pin `test_max_search_iterations_is_on_both_parsers()` mirroring the existing checkpoint-interval pin.
6. `OUTER_ONLY_ALLOWLIST` entries each have a one-line justification comment (same convention as the existing inner allowlist).
7. The top-of-file docstring at `test_cli_parser_drift.py` is updated to describe the bidirectional guard and adds #2819 to the historical-instance list.
8. The two defensive read sites at `route_cmd.py:4923` and `:5289` either simplify to direct attribute access or carry a comment noting the inner parser now guarantees the attribute exists.

## Discovered by

PR #2818 builder during the #2817 fix work. Filed as follow-up per scope discipline — the inner→outer fix shipped in #2818; this issue captures the symmetric outer→inner fix and the drift-test broadening.

rjwalters:
## Summary

Closes #2819.

### Part A — plumbing fix
- Declare `--max-search-iterations` on the inner `route_cmd.py` parser (`type=int, default=0`, matching the outer parser semantics from #2610).
- Add a forwarding block in `commands/routing.py::run_route_command` that only forwards non-default values (model: the `--per-net-timeout` and `--checkpoint-interval` blocks already there).
- Tighten the two defensive read sites in `route_cmd.py` (~lines 4923 and 5289) from `getattr(args, "max_search_iterations", 0) or 0` to direct attribute access plus a comment noting the inner parser now guarantees the attribute exists.

End-to-end effect: `kct route --max-search-iterations N board.kicad_pcb` now correctly threads `N` through to the C++ A* backstop instead of silently dropping the flag and running with 0 (the historical `cols*rows*4` heuristic).

### Part B — symmetric drift-test broadening
- Add `OUTER_ONLY_ALLOWLIST` near `INNER_ONLY_ALLOWLIST` (initially empty — after the #2819 fix every outer flag has an inner counterpart).
- Add `test_outer_only_flags_are_in_allowlist` asserting `outer_flags - inner_flags ⊆ OUTER_ONLY_ALLOWLIST`, mirroring the existing inner-direction check.
- Add `test_allowlist_entries_are_actually_outer_only` mirroring the existing stale-allowlist check for `INNER_ONLY_ALLOWLIST`.
- Add `test_max_search_iterations_is_on_both_parsers` direct pin (mirrors `test_checkpoint_interval_is_on_both_parsers` from #2818).
- Update the module docstring to describe the bidirectional guard and add #2819 to the historical-instance list.

This is the 5th instance of the argparse-drift bug class (after #2620, #2622/#2793, and #2812/#2817). The symmetric drift test now prevents the 6th instance landing in either direction.

## Test plan

- [x] **Inner-parser smoke**: `uv run python -c "from kicad_tools.cli.route_cmd import main; main(['--help'])"` shows `--max-search-iterations` in the help output (verified).
- [x] **End-to-end forwarding**: in-process test parses `kct route --max-search-iterations 12345 /tmp/dummy.kicad_pcb` through the outer parser, runs the shim, and confirms `["--max-search-iterations", "12345"]` is in the resulting `sub_argv` (verified — the inner parser then captures `args.max_search_iterations == 12345`).
- [x] **Default-only forwarding preserved**: without `--max-search-iterations` (or with explicit `0`), the shim does NOT add the flag to `sub_argv` — preserves "boards 01-05 produce identical routes" invariant.
- [x] **Drift test pin**: `pytest tests/test_cli_parser_drift.py::test_max_search_iterations_is_on_both_parsers` passes.
- [x] **Drift test broadening**: `pytest tests/test_cli_parser_drift.py` — all 6 tests pass (3 original + 3 new symmetric ones).
- [x] **Negative test**: simulated the bug (removed `--max-search-iterations` from the inner parser via mock) and confirmed the new `test_outer_only_flags_are_in_allowlist` catches it and names the flag.
- [x] **Lint/format**: `ruff check` and `ruff format --check` pass on all 3 changed files.

## Files touched

- `src/kicad_tools/cli/route_cmd.py` — inner parser add_argument + 2 read-site simplifications.
- `src/kicad_tools/cli/commands/routing.py` — forwarding block in `run_route_command`.
- `tests/test_cli_parser_drift.py` — `OUTER_ONLY_ALLOWLIST`, 3 new tests, updated docstring.

### DIFF
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -190,6 +190,13 @@ def run_route_command(args) -> int:
     checkpoint_interval_val = getattr(args, "checkpoint_interval", 30.0)
     if checkpoint_interval_val != 30.0:
         sub_argv.extend(["--checkpoint-interval", str(checkpoint_interval_val)])
+    # Issue #2819: forward --max-search-iterations to the inner parser.
+    # Both defaults are 0 (= use cols*rows*4 heuristic), so only forward
+    # when the user passed a non-default value (matches the per-net-timeout
+    # and checkpoint-interval patterns above).
+    max_iter_val = getattr(args, "max_search_iterations", 0) or 0
+    if max_iter_val:
+        sub_argv.extend(["--max-search-iterations", str(max_iter_val)])
     if args.verbose:
         sub_argv.append("--verbose")
     if args.dry_run:
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -3860,6 +3860,23 @@ def main(argv: list[str] | None = None) -> int:
             "Use 0 to disable checkpointing (only the terminal save fires)."
         ),
     )
+    # Issue #2819: declare --max-search-iterations on the inner parser so the
+    # forwarding shim in ``commands/routing.py`` can hand it through.  Default
+    # 0 = use the historical ``cols * rows * 4`` heuristic (matches the outer
+    # parser declaration at parser.py and the C++ A* backstop semantics from
+    # Issue #2610).
+    parser.add_argument(
+        "--max-search-iterations",
+        type=int,
+        default=0,
+        help=(
+            "Override the C++ A* iteration backstop (default: 0 = use "
+            "cols*rows*4, which is ~1M for a 500x500 grid). Positive values "
+            "let dense boards trade memory for completeness. Iteration-cap "
+            "aborts are logged distinctly from --per-net-timeout (wall-clock) "
+            "aborts so you can tell which limit fired."
+        ),
+    )
     parser.add_argument(
         "--seed",
         type=int,
@@ -4920,7 +4937,12 @@ def main(argv: list[str] | None = None) -> int:
                 validate_drc=not args.force,
                 strict_drc=False,  # Only fail on hard constraint (grid > clearance)
                 # Issue #2610: thread --max-search-iterations through.
-                max_search_iterations=getattr(args, "max_search_iterations", 0) or 0,
+                # The inner parser declares this flag with default=0 (Issue
+                # #2819), so the attribute is guaranteed to exist; the
+                # ``or 0`` guards against an explicit ``--max-search-iterations 0``
+                # being treated as falsy (which is the intended behaviour:
+                # 0 means "use the cols*rows*4 heuristic").
+                max_search_iterations=args.max_search_iterations or 0,
             )
     except Exception as e:
         print(f"Error loading PCB: {e}", file=sys.stderr)
@@ -5286,7 +5308,10 @@ def main(argv: list[str] | None = None) -> int:
             if per_net_timeout_val:
                 flush_print(f"  Per-net timeout: {per_net_timeout_val}s")
             # Issue #2610: report the iteration backstop override if set.
-            _max_iter_val = getattr(args, "max_search_iterations", 0) or 0
+            # Issue #2819: the inner parser now declares the flag (default=0),
+            # so the attribute is guaranteed to exist; ``or 0`` preserves the
+            # historic semantics (0 = use cols*rows*4 heuristic, suppress log).
+            _max_iter_val = args.max_search_iterations or 0
             if _max_iter_val:
                 flush_print(f"  Max search iterations: {_max_iter_val}")
             if args.profile:
--- a/tests/test_cli_parser_drift.py
+++ b/tests/test_cli_parser_drift.py
@@ -1,10 +1,10 @@
-"""Argparse-drift regression test (Issue #2817).
+"""Argparse-drift regression test (Issues #2817, #2819).
 
-This test prevents a recurring class of bug where a new flag is added to
-the *inner* ``route_cmd.py`` parser without being added to the *outer*
-``parser.py`` parser and to the forwarding shim in
-``commands/routing.py``.  Three prior instances of this exact pattern
-have shipped to ``main`` and required a follow-up bugfix:
+This test prevents a recurring class of bug where a flag is added to one
+of the two ``route`` parsers without being added to the other (and to
+the forwarding shim in ``commands/routing.py``).  Five prior instances
+of this exact pattern have shipped to ``main`` and required a follow-up
+bugfix:
 
 * **#2620** -- ``--placement-feedback-outer-timeout`` declared outer,
   rejected inner (drift in the opposite direction).
@@ -12,20 +12,30 @@ have shipped to ``main`` and required a follow-up bugfix:
 * **#2812 / #2817** -- ``--checkpoint-interval`` declared inner only;
   ``kct route --checkpoint-interval 30`` rejected with
   ``error: unrecognized arguments``.
+* **#2819** -- ``--max-search-iterations`` declared outer only; the
+  shim dropped the flag and the inner parser never saw it, so
+  ``kct route --max-search-iterations 50000`` parsed cleanly but ran
+  the C++ A* with ``max_search_iterations=0`` (the historical
+  ``cols*rows*4`` heuristic) regardless of the user-supplied value.
 
 The drift is invisible to the type checker and to most unit tests
 because both parsers are constructed independently with no shared
-schema.  This test introspects both parsers and asserts that every
-``--flag`` accepted by the inner parser is also accepted by the outer
-parser, except for an explicit allowlist of historically inner-only
-flags.
+schema.  This test introspects both parsers and asserts symmetric
+containment:
 
-If you legitimately need to add a new inner-only flag, add it to
-``INNER_ONLY_ALLOWLIST`` below with a comment explaining why it does
-not belong on the outer parser.  In most cases the right answer is to
-add it to BOTH parsers and to the forwarding shim -- see the
-``--per-net-timeout`` block in ``commands/routing.py`` for the model
-pattern.
+* ``inner_flags - outer_flags ⊆ INNER_ONLY_ALLOWLIST`` (inner-only flags
+  must be explicitly allowlisted -- guards the #2812/#2817 direction).
+* ``outer_flags - inner_flags ⊆ OUTER_ONLY_ALLOWLIST`` (outer-only flags
+  must be explicitly allowlisted -- guards the #2819 direction, where
+  the shim consumes the flag entirely or the forwarding block is
+  missing).
+
+If you legitimately need to add a new inner-only or outer-only flag,
+add it to the corresponding allowlist below with a comment explaining
+why the flag does not belong on the other parser.  In most cases the
+right answer is to add it to BOTH parsers and to the forwarding shim --
+see the ``--per-net-timeout`` block in ``commands/routing.py`` for the
+model pattern.
 """
 
 from __future__ import annotations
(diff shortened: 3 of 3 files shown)


rjwalters:
## Approved

Thorough review of the symmetric drift fix and direct plumbing for `--max-search-iterations`. All four review criteria pass.

### Part A plumbing — correct
- Inner parser declares `--max-search-iterations` at `route_cmd.py:3869` (under the `#2819` comment at 3863), `type=int, default=0`, matching the outer parser semantics from #2610. Help output verified to include the flag.
- Forwarding block in `commands/routing.py` after the `--checkpoint-interval` block (lines 193-199), using the same non-default-only pattern: `if max_iter_val: sub_argv.extend([...])`.
- Both defensive reads at `route_cmd.py:4945` and `:5314` simplified from `getattr(args, "max_search_iterations", 0) or 0` to direct `args.max_search_iterations or 0` with comments explaining the `or 0` preserves the "explicit 0 = use heuristic" semantics. Good touch leaving the `or 0` — that's substantive (handles explicit `--max-search-iterations 0`), not vestigial.

### Part B drift test — symmetric and well-instrumented
- `OUTER_ONLY_ALLOWLIST: frozenset[str] = frozenset(set())` declared next to `INNER_ONLY_ALLOWLIST` with the same justification convention; the snapshot comment correctly notes that after #2819 the set is genuinely empty.
- `test_outer_only_flags_are_in_allowlist` asserts `outer - inner ⊆ OUTER_ONLY_ALLOWLIST` with a failure message that explicitly names both shim sites: `src/kicad_tools/cli/route_cmd.py :: main (add matching add_argument)` and `src/kicad_tools/cli/commands/routing.py :: run_route_command (forward to sub_argv)`.
- `test_allowlist_entries_are_actually_outer_only` mirrors the existing stale-allowlist sanity check for `INNER_ONLY_ALLOWLIST`.
- `test_max_search_iterations_is_on_both_parsers` direct-pin mirrors `test_checkpoint_interval_is_on_both_parsers` from #2818.
- Module docstring updated to describe the bidirectional guard with #2819 added to the historical-instance list.

### Negative test — verified, not trivial
Simulated the regression by monkey-patching `_inner_route_parser_flags` to drop `--max-search-iterations` from the inner-parser set. `test_outer_only_flags_are_in_allowlist` failed with a message that:
- Names the offending flag (`--max-search-iterations`).
- Lists both shim sites with file:function references.
- Points at the `--per-net-timeout` model pattern.
- Mentions `OUTER_ONLY_ALLOWLIST` as the legitimate escape hatch.

The assertion is real — the failure path runs through `pytest.fail` and would block CI on regression.

### Default-only forwarding — verified at the parser level
Confirmed live through the real outer parser:
- Omit flag → `args.max_search_iterations == 0` → `max_iter_val = 0 or 0 = 0` → falsy → NOT forwarded.
- `--max-search-iterations 0` → same path → NOT forwarded (preserves "0 = use cols*rows*4 heuristic" semantics).
- `--max-search-iterations 12345` → `max_iter_val = 12345` → truthy → forwarded as `["--max-search-iterations", "12345"]`.

This mirrors the `--per-net-timeout` and `--checkpoint-interval` patterns exactly and preserves the "boards 01-05 produce identical routes" invariant.

### CI — all green
All 8 checks pass: C++ Build Check, Diff-Pair Routing Regression, Lint & Format, Match-Group Routing Regression, Routed PCB DRC Check, Test (6m28s), Type Check, kicad-cli Round-trip Smoke. Merge state: CLEAN.

### Scope — tight
Only the 3 files cited in the issue: `route_cmd.py` (+27/-2), `commands/routing.py` (+7/-0), `test_cli_parser_drift.py` (+130/-20). `ruff check` and `ruff format --check` clean on all three.

Nice piece of work — the symmetric drift test should prevent the 6th instance of this bug class regardless of direction, and the failure messages give future builders/judges a precise repro path.

Issue #2839, opened 2026-05-13
[board-03] D: kct route --auto-fix engages 0/18 repairs (contract violation)


rjwalters:
## Parent

Decomposed from #2833 (Board 03 USB joystick routing stuck at 9/16). This is **Sub-issue D** of 5.

## Scope

Fix the contract violation in `kct route --auto-fix`: on board 03, the post-route auto-fix pass reports `Repaired 0/18 violations` but lists 18 nudges that *did* succeed at the displacement layer — the work is invisible because the connectivity-rollback path silently zeros it out and `_run_auto_fix` does not surface the cause.

### Reproducer (captured fresh)

```
uv run kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb \
    --output /tmp/board03_curate_test.kicad_pcb \
    --auto-fix --auto-fix-passes 1 --mfr jlcpcb-tier1 --timeout 300 -v
```

Output of the `--- Auto-Fix DRC Violations ---` section:

```
Running DRC on: board03_curate_test.kicad_pcb
Found 18 repairable violation(s):
  Clearance: 18

Repaired 0/18 violations
------------------------------------------------------------
CLEARANCE: 18/18
  [SEGMENT] USB_D+      at (140.2500, 109.0000) -> 0.0254mm
  [SEGMENT] JOY_Y       at (118.4000, 135.2000) -> 0.0902mm
  [SEGMENT] USB_D+      at (140.2754, 109.0000) -> 0.0170mm
  [SEGMENT] JOY_X       at (130.1400, 134.3400) -> 0.0126mm
  [SEGMENT] JOY_Y       at (118.4902, 135.2000) -> 0.1880mm
  ... and 12 more
121 violation(s) remain:
  Repairable (not yet fixed): 18
  Non-repairable: 103
  Auto-fix: some violations remain (manual repair may be needed)
```

Running `kct fix-drc` directly on the same routed PCB shows the **hidden cause** on stderr:

```
$ uv run kct fix-drc /tmp/board03_curate_test.kicad_pcb --max-passes 1 --max-displacement 2.0 --local-reroute
Warning: pass 1 decreased connectivity (9 -> 7 nets); rolled back.
```

### Root cause

This is **not** a classifier mismatch and **not** a `--max-displacement` cap problem:

1. The 18 `clearance_pad_segment` violations are correctly classified as repairable (see `fix_drc_cmd.py:209-217`).
[...]

Issue #2836, opened 2026-05-13
[board-03] A: lock J1/J2 and add B.Cu GND pour in usb_joystick.kicad_pcb


rjwalters:
## Parent

Decomposed from #2833 (Board 03 USB joystick routing stuck at 9/16). This is **Sub-issue A** of 5.

## Scope

Make the board-03 source PCB (`boards/03-usb-joystick/output/usb_joystick.kicad_pcb`) a proper "anchor + ground plane" starting point for the router by:

1. **Locking the perimeter connectors** `J1` (USB-C, GCT_USB4105) and `J2` (joystick header) — both are mechanically constrained on the board edge and must not move when `kct route --placement-feedback` runs.
2. **Adding a B.Cu ground pour** covering the board outline minus keepouts, on net `GND`.

This is a board-source change only — **no router/tooling code changes**.

## Why

Per #2833 diagnosis:

- `usb_joystick.kicad_pcb` contains zero `(locked)` markers. Without locking J1/J2, `--anchor-weight` from PR #2825 cannot engage, and the placement-feedback pass pulls VBUS/VCC strands off the perimeter (-> 5 unconnected VBUS pads at J1, 8 unconnected VCC pads at MCU bypass caps).
- The board has zero zones. 32 of ~40 GND pads cannot be reached on F.Cu alone. A B.Cu pour is structurally required.

The acceptance criteria for the parent (#2833) treat both as preconditions before any router work can be evaluated.

## Implementation guidance

This board is regenerated by `boards/03-usb-joystick/generate_pcb.py`. The change should be made in `generate_pcb.py` and a fresh artifact committed, **not** by hand-editing the saved `.kicad_pcb`.

- Use the `PCB.save` roundtrip that PR #2830 made work for footprint mutation.
- For the locked flag on footprints, set `footprint.locked = True` (or whatever the kicad-tools API exposes) on J1 and J2 before save.
- For the B.Cu pour, add a zone bound to net `GND` on layer `B.Cu`, polygon outline approximately matching the Edge.Cuts rectangle inset by ~0.5mm. Use `kct stitch ... --net GND` after routing for via stitching — that is **not** part of this sub-issue.

## Acceptance criteria

- [ ] `boards/03-usb-joystick/output/usb_joystick.kicad_pcb` has `(locked yes)` (or equivalent) on the J1 and J2 footprint blocks.
[...]

PR #2855, opened 2026-05-13
fix(router): make PlacementFeedbackLoop monotonic via best-known snapshot rollback

### LINKED ISSUE #2840: [board-03] E: placement-feedback rollback drops pre-pass connectivity floor
## Parent

Decomposed from #2833 (Board 03 USB joystick routing stuck at 9/16). This is **Sub-issue E** of 5.

## Status

Curated by Loom Curator 2026-05-12. **The original framing conflates two distinct rollback paths**; the curation re-scopes the issue to the actual contract violation found in code.

## Observed symptom (from #2833 reproducer)

```
Result: PARTIAL: Best result 54% on 2 layers
  Nets routed: 7/13 (54%)
  ...
Warning: pass 1 decreased connectivity (9 -> 7 nets); rolled back.
```

The router's reported completion (7/13) and the rolled-back fix-drc baseline (9 nets) are **inconsistent**, and the previously-committed `usb_joystick_routed.kicad_pcb` has 9 fully-routed nets including `USB_CC1`/`USB_CC2`/`VBUS` — which the fresh route never reaches.

## Where the "rolled back" message comes from

The "decreased connectivity ... rolled back" log is **not** emitted by the placement-feedback loop. It is emitted by the auto-fix path:

- `src/kicad_tools/cli/fix_drc_cmd.py:307-310` prints the warning.
- `src/kicad_tools/cli/route_cmd.py:1052-1110` (`_run_auto_fix`) shells out to `fix_drc_main` with `--max-displacement 2.0 --local-reroute` when `--auto-fix` is on.
- The placement-feedback loop in `src/kicad_tools/router/placement_feedback.py:352-498` has **no rollback path at all** — it only calls `self._clear_routes()` and routes fresh; the only thing it snapshots is component *positions* via `_snapshot_positions` (line 770) for the placement-diff artifact, not for rollback.

So the title "placement-feedback rollback drops pre-pass connectivity floor" is misleading: there are actually three orthogonal issues, only two of which are real contract violations.

### Defect E1 — `fix_drc_cmd.py` rollback baseline is wrong-grained

`fix_drc_cmd.py:263-310`:

```python
# Snapshot and baseline connectivity before the repair pass
snapshot: bytes | None = None
baseline_conn: int | None = None
if do_connectivity_check:
    load_for_snapshot = output_path if pass_num > 1 else pcb_path
    if load_for_snapshot.exists():
        snapshot = load_for_snapshot.read_bytes()
        baseline_conn = _count_connected_nets(load_for_snapshot)
...
# Post-pass connectivity check and rollback
after_conn = _count_connected_nets(output_path)
if after_conn >= 0 and after_conn < baseline_conn:
    output_path.write_bytes(snapshot)
    rolled_back = True
    ...
```

This **is byte-for-byte correct** in isolation: it reads the file as bytes, runs the repair pass, and writes the original bytes back if connectivity drops. The byte-buffer-snapshot pattern is sound.

**However**, two contract violations exist:

1. **`_count_connected_nets` uses `ConnectivityValidator.connected_nets`** (`fix_drc_cmd.py:450-465`), which counts *all* nets where all pads are graph-connected, **including trivially-connected single-pad nets** (`connectivity.py:282-285`). The router's "Nets routed: 7/13" denominator is the set of router-targeted multi-pad nets. These are different denominators — the "9" baseline and the "7/13" report cannot be compared meaningfully. The rollback decision is correct under its own metric, but the log line is confusing because two metrics are interleaved.

2. **Snapshot vs. router pre-rollback state has no monotonicity guarantee with the router output.** The rollback restores fix-drc's *input* (which is the router's *output*), not "the best routed state ever observed in this pipeline." If the router itself regressed before fix-drc ran, the rollback's "floor" already starts below the committed `usb_joystick_routed.kicad_pcb`. See E3.

### Defect E2 — placement-feedback loop has no rollback path

`router/placement_feedback.py:352-498`. The loop performs:

1. `self._clear_routes()` — calls `self.router._reset_for_new_trial()` (line 502), discarding all current routes.
2. `router.route_all_negotiated(...)` — fresh re-route.
3. Records `routed_count` into `routed_history`.
4. If `failed_nets` is non-empty, applies a placement strategy (or breaks).

**There is no comparison between pre-iteration and post-iteration `routed_count`, and no restore-best path.** If iteration `N+1` produces fewer routed nets than iteration `N`, the loop just continues with the worse state. Final result at line 489-498 returns `list(self.router.routes)` — the live state, not a best-known snapshot.

This is the *actual* placement-feedback contract violation the issue title refers to, and it is a real defect: the loop is non-monotonic.

The reference pattern that already exists in this repo is `Autorouter.route_all_negotiated` at `core.py:4791-4838` (PR #2805, closes #2803):

```python
best_routes: list[Route] = copy.deepcopy(list(self.routes))
best_net_routes: dict[int, list[Route]] = copy.deepcopy(net_routes)
best_routed_count = sum(1 for r in net_routes.values() if r)
best_metrics = IterationMetrics(iteration=0, routed_count=best_routed_count, overflow=overflow)
...
def _capture_iteration_end(iter_index, overflow_val):
    ...
    improved = metrics.is_better_than(best_metrics)
    if improved:
        best_metrics = metrics
```

The `PlacementFeedbackLoop` needs an analogous "best routed state" snapshot taken at the end of each iteration, with `self.router.routes = copy.deepcopy(best_routes)` restored before returning when the final iteration is not the best.

### Defect E3 — full pipeline (`kct route`) is non-monotonic vs. on-disk baseline

The unrouted `usb_joystick.kicad_pcb` produces 7/13 routed; the saved `usb_joystick_routed.kicad_pcb` (a previous pipeline run, now committed) has 9/13. `kct route` neither reads the committed routed PCB as a "previous best" floor nor falls back to it on regression.

This is **out of scope for this issue** — fix E2 first, then evaluate whether the pipeline still regresses. The pipeline-level monotonicity question can be a separate ticket if E2 doesn't close the gap. (See #2826 for the convergence-side discussion.)

## Scope of this issue

**E2 only.** Fix the placement-feedback loop to be monotonic in routed-net count:

1. Snapshot the routed state (`list(self.router.routes)` + per-net route mapping) at the end of each iteration.
2. Compare iteration-end `routed_count` to the best-known `routed_count`. Replace the snapshot iff `routed_count` strictly improved (or, optionally, compare on a lex tuple `(routed_count, -overflow)` matching `core.py:4832`'s `IterationMetrics.is_better_than`).
3. Before returning `PlacementFeedbackResult`, restore the best-known snapshot to `self.router.routes` (and any per-net cache the router maintains) when the final iteration is not the best.
4. Log per-iteration: `placement-feedback iter N: routed=R/T (best=B); {improved|reverted|tied}`.

Defects E1 and E3 are tracked separately (file follow-up issues if needed; E1 is cosmetic, E3 may close with E2).

### File and line refs

- `src/kicad_tools/router/placement_feedback.py:352-498` — main loop body.
- `src/kicad_tools/router/placement_feedback.py:500-502` — `_clear_routes()` calls `router._reset_for_new_trial()`; the rollback must NOT call `_reset_for_new_trial` because we want to restore `self.routes`, not clear them.
- `src/kicad_tools/router/core.py:4791-4838` — reference deep-copy pattern from #2805.

### Suggested patch shape

In `PlacementFeedbackLoop.run` (or equivalent):

```python
import copy

# Before the iteration loop:
best_routes_snapshot = copy.deepcopy(list(self.router.routes))
best_routed_count = -1  # forces the first iteration to set the snapshot
best_iteration = -1

for iteration in range(max_adjustments + 1):
    ...
    # After routing:
    routed_count = total_nets - len(failed_nets)
    routed_history.append(routed_count)

    if routed_count > best_routed_count:
        best_routed_count = routed_count
        best_routes_snapshot = copy.deepcopy(list(self.router.routes))
        best_iteration = iteration
        if self.verbose:
            print(f"  [iter {iteration}] new best routed count: {routed_count}/{total_nets}")
    elif self.verbose:
        delta = routed_count - best_routed_count
        print(f"  [iter {iteration}] routed={routed_count}/{total_nets} (best={best_routed_count}, delta={delta:+d})")

    ...

# Before constructing PlacementFeedbackResult:
if best_iteration >= 0 and best_routed_count > (total_nets - len(self.router.get_failed_nets())):
    if self.verbose:
        print(f"  Restoring best snapshot from iter {best_iteration} ({best_routed_count} routed)")
    self.router.routes = copy.deepcopy(best_routes_snapshot)
    # Recompute failed_nets against the restored snapshot
    ...
```

The exact restore mechanism depends on the `Autorouter` internal structure for `routes`/`net_routes`/`_blocked`/`_net`. Follow the per-iteration capture pattern in `core.py:4791-4838` exactly: deep-copy on strict improvement, restore wholesale on exit when current state is worse than best.

### Test plan

1. **Unit regression**: `tests/test_placement_feedback_rollback.py::test_loop_preserves_pre_pass_floor`. Construct a synthetic 4-net board (mock `Autorouter`) where iteration 0 routes 3/4 nets, iteration 1 routes 1/4 nets, iteration 2 routes 2/4 nets. Run `PlacementFeedbackLoop.run(max_adjustments=2)`. **Assert**: returned `result.routes` corresponds to iteration-0 state (3 routed); `len(result.failed_nets) == 1`. Without the fix this returns iteration-2 state with 2 failed.
2. **Trajectory log**: same test, capture stdout, assert log contains `[iter 1]` line with `delta=-2` and a "Restoring best snapshot" line at exit.
3. **No-regression case**: 3 iterations all routing 4/4 → no restore log emitted; returned state is iteration-2 (success path unchanged).
4. **First-iteration-best case**: iter 0 routes 4/4 → `success=True` short-circuit at `placement_feedback.py:391-403` still works (no behavioral change).
5. **Board-03 reproducer**: `uv run kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb --output /tmp/board03_e2_test.kicad_pcb --auto-layers --placement-feedback --placement-feedback-budget 3 --timeout 600`. **Assert**: the printed router result is `>=` 7/13 (no regression vs. current behavior) and is monotonic across reported iterations. Specifically the saved board's 9-net floor is a *separate* concern (E3) and is not asserted here.

## Acceptance criteria

- [ ] `PlacementFeedbackLoop.run()` returns `PlacementFeedbackResult.routes` corresponding to the iteration with the highest `routed_count` observed, not the final iteration.
- [ ] `self.router.routes` is restored to the best-known snapshot before the result is constructed (so post-loop callers see the best state on the router, not the live final state).
- [ ] Per-iteration log line is emitted with `routed=R/T (best=B)` and an `{improved|reverted|tied}` tag (verbose mode only).
- [ ] `tests/test_placement_feedback_rollback.py` passes; covers the four cases above.
- [ ] No behavioral change when the loop converges (`success=True` short-circuit at line 391) or when all iterations route the same count (no restore needed, no log change).
- [ ] Type checks (`mypy`) clean; deep-copy semantics match `core.py:4791-4838`.

## Out of scope

- E1 (`_count_connected_nets` denominator mismatch with router's routed-count) — file as a separate cosmetic-log ticket if it surfaces real confusion.
- E3 (pipeline-level non-monotonicity vs. on-disk baseline) — re-evaluate after E2; may close as duplicate of #2826 or this issue.
- Sibling sub-issues A (lock J1/J2, B.Cu pour), B (grid auto-bump), C (#2761 single-ended XTAL), D (`--auto-fix` engages 0 repairs).

## References

- Parent: #2833 (decomposition table at the bottom)
- Pattern source: PR #2805 / Issue #2803 — `_capture_iteration_end` deep-copy pattern in `core.py:4791-4838`
- Related: #2826 (feasibility-gated convergence — when to terminate, not what to restore)
- Related: #2808 — checkpoint hook semantics (the callback fires after the snapshot replacement; same pattern E2 should follow)

rjwalters:
Closes #2840

## Summary

The placement-feedback loop is non-monotonic by construction: each iteration calls `_clear_routes()` and routes fresh, so if iteration N+1 produces fewer routed nets than iteration N, the router's live `self.routes` ends up worse than a prior pass produced -- and the final `PlacementFeedbackResult` reports the regression.

This PR adds a best-known snapshot pattern matching the existing deep-copy logic in `Autorouter.route_all_negotiated` (core.py:4791-4838, PR #2805):

1. **Snapshot on strict improvement.** At iteration end, if `routed_count` strictly improved, deep-copy `self.router.routes` into `best_routes_snapshot`.
2. **Restore best before returning.** Before constructing `PlacementFeedbackResult`, if the final iteration regressed below the best-known count, restore `self.router.routes` to a deep copy of the snapshot. The grid is intentionally NOT restored -- `_clear_routes()` resets it at the top of every iteration so no caller relies on grid/routes consistency after the loop returns.
3. **Recompute failed_nets after restoration** so the result reflects the restored state, not the live last-iteration state.

Per-iteration verbose log emits a trajectory line tagged `improved|tied|reverted` so the regression is visible at runtime. End-of-loop emits `Restoring best snapshot from iter N ...` when the restore path fires.

The success short-circuit (`failed_nets == []`) at the top of the loop is unchanged: a converged iteration is by definition the best state and short-circuits before any snapshot/restore logic runs.

## Test plan

Six new tests in `TestPlacementFeedbackLoopRollback` (all passing):

- [x] `test_regression_after_best_iteration_restores_best_state` -- three iterations routing [3, 1, 2]/4 nets; asserts restored state == iter-0 (3 routed, 1 failed, not iter-2's 2 routed / 2 failed).
- [x] `test_route_by_route_equivalence_after_rollback` -- deep-copy semantics: post-rollback routes have iter-0 contents (same `(net, net_name)` tuples) but no shared Python object identity with the originals.
- [x] `test_first_iteration_best_no_rollback_needed_success` -- `pf_converged` short-circuit unchanged when iter-0 routes all nets.
- [x] `test_no_regression_returns_last_iteration_state` -- monotonically improving routed count never triggers the restore branch.
- [x] `test_trajectory_log_emits_per_iteration_progress` -- verbose stdout contains `[iter N]` improved/reverted tags and `Restoring best snapshot` exit log.
- [x] `test_tied_iteration_does_not_replace_snapshot` -- strict-improvement semantics: a tied iteration keeps the earlier best snapshot so later regressions roll back to the earlier iteration.

Pre-existing tests still pass:
- [x] `tests/test_placement_feedback.py` -- 55 passed.
- [x] `tests/test_route_cmd_placement_feedback.py` -- 53 passed.
- [x] `tests/test_chorus_test_placement_feedback.py` -- 3 passed, 1 skipped.
- [x] `ruff check` and `ruff format --check` clean on modified files.

## Out of scope

Per Curator on #2840:

- **E1** (fix-drc rollback log clarity) -- fix-drc rollback is byte-for-byte correct in isolation; the log denominator mismatch is cosmetic.
- **E3** (full pipeline non-monotonicity vs committed baseline) -- deferred until this lands.
- Sibling sub-issues A/B/C/D (#2836-#2839).

## References

- Reference pattern: `Autorouter.route_all_negotiated` `_capture_iteration_end` at `src/kicad_tools/router/core.py:4791-4838` and the symmetric restore at `core.py:6126-6149` (PR #2805 / Issue #2803).

### DIFF
--- a/src/kicad_tools/router/placement_feedback.py
+++ b/src/kicad_tools/router/placement_feedback.py
@@ -8,6 +8,7 @@ routing failures.
 
 from __future__ import annotations
 
+import copy
 import math
 import time
 from dataclasses import dataclass, field
@@ -320,6 +321,26 @@ class PlacementFeedbackLoop:
         # backwards-compatible default on ``PlacementFeedbackResult``.
         exit_reason = "pf_max_iter"
 
+        # Issue #2840: best-known routed state across all iterations.
+        # The placement-feedback loop is non-monotonic by construction --
+        # each iteration calls ``_clear_routes()`` and routes fresh, so if
+        # iteration N+1 produces fewer routed nets than iteration N, the
+        # router's live ``self.routes`` ends up worse than a prior
+        # iteration's result.  To make the loop monotonic in routed-net
+        # count, snapshot the routes whenever a new iteration strictly
+        # improves on the best-known count, and restore the best snapshot
+        # before returning so callers always see the highest routed count
+        # observed in this run.
+        #
+        # Mirrors the deep-copy pattern from
+        # ``Autorouter.route_all_negotiated`` (core.py:4791-4838, PR #2805).
+        # Only deep-copies on strict improvement (Strategy B from #2803:
+        # minimal memory cost — at most ``max_adjustments + 1`` snapshots
+        # taken, only one retained).
+        best_routes_snapshot: list[Route] = []
+        best_routed_count = -1
+        best_iteration = -1
+
         if self.verbose:
             print("\n=== Placement-Routing Feedback Loop ===")
             print(f"  Max adjustments: {max_adjustments}")
@@ -383,9 +404,35 @@ class PlacementFeedbackLoop:
             routed_count = total_nets - len(failed_nets)
             routed_history.append(routed_count)
 
+            # Issue #2840: snapshot the routed state on strict improvement
+            # so the loop is monotonic in routed-net count.  Without this,
+            # a later iteration that regresses (fewer nets routed) leaves
+            # ``self.router.routes`` in a worse state than a prior pass
+            # produced, and the final ``PlacementFeedbackResult`` reports
+            # the regression.  Deep-copy only on strict improvement.
+            if routed_count > best_routed_count:
+                improved = True
+                best_routed_count = routed_count
+                best_routes_snapshot = copy.deepcopy(list(self.router.routes))
+                best_iteration = iteration
+            else:
+                improved = False
+
             if self.verbose:
                 print(f"  Routed: {routed_count}/{total_nets} nets")
                 print(f"  Failed: {len(failed_nets)} nets")
+                if improved:
+                    print(
+                        f"  [iter {iteration}] routed={routed_count}/{total_nets} "
+                        f"(best={best_routed_count}); improved"
+                    )
+                else:
+                    delta = routed_count - best_routed_count
+                    tag = "tied" if delta == 0 else "reverted"
+                    print(
+                        f"  [iter {iteration}] routed={routed_count}/{total_nets} "
+                        f"(best={best_routed_count}, delta={delta:+d}); {tag}"
+                    )
 
             # Success - all nets routed
             if not failed_nets:
@@ -474,7 +521,39 @@ class PlacementFeedbackLoop:
             adjustments.append(adjustment)
             total_moved += len(result.components_moved)
 
-        # Final failure analysis
+        # Issue #2840: restore the best-known routed state before
+        # reporting the result.  The placement-feedback loop is
+        # non-monotonic by construction (each iteration calls
+        # ``_clear_routes()`` and routes fresh), so ``self.router.routes``
+        # at this point reflects whichever iteration ran last -- not
+        # necessarily the best.  Restore the deep-copied snapshot whenever
+        # a strictly better state was observed earlier in the run.
+        #
+        # We restore by reassigning ``self.router.routes`` to a deep copy
+        # of the best snapshot.  Note: the grid state is NOT restored.
+        # That is intentional -- the grid is reset by ``_clear_routes()``
+        # at the top of every iteration, so no caller relies on grid
+        # state being consistent with ``self.router.routes`` after the
+        # placement-feedback loop returns.  ``get_failed_nets()`` derives
+        # solely from ``self.router.routes`` (core.py:8320), so restoring
+        # the routes is sufficient to make the reported failed-net count
+        # consistent with the restored snapshot.
+        current_routed_count = len(self.router.nets) - 1 - len(self.router.get_failed_nets())
+        if best_routed_count > current_routed_count:
+            if self.verbose:
+                print(
+                    f"  Restoring best snapshot from iter {best_iteration} "
+                    f"(routed={best_routed_count}) "
+                    f"instead of final iter (routed={current_routed_count})"
+                )
+            # Deep-copy the snapshot so the loop's internal best state
+            # cannot be mutated through ``self.router.routes`` after
+            # this call returns.
+            self.router.routes = copy.deepcopy(best_routes_snapshot)
+
+        # Final failure analysis -- computed AFTER restoration so the
+        # result reflects the restored state, not the live last-iteration
+        # state.
         failed_nets = self.router.get_failed_nets()
         if adjustments:
             adjustments[-1].failed_nets_after = failed_nets
--- a/tests/test_placement_feedback.py
+++ b/tests/test_placement_feedback.py
@@ -1226,3 +1226,357 @@ class TestPlacementFeedbackLoopBuilderHandoff:
         # the diagonals.  South / West / SE / SW must all appear.
         assert "W0" in all_quadrants and "0S" in all_quadrants
         assert "WS" in all_quadrants and "ES" in all_quadrants
+
+
+class _RoutingFakeRouter:
+    """Fake Autorouter that mutates ``self.routes`` per iteration.
+
+    Used by Issue #2840 rollback tests.  Each call to
+    ``route_all_negotiated`` consumes the next entry in ``routes_per_call``
+    and populates ``self.routes`` accordingly so:
+
+    - ``_reset_for_new_trial()`` clears ``self.routes`` (mirroring real
+      :class:`Autorouter` behavior).
+    - ``route_all_negotiated()`` appends the configured routes for the
+      current call index.
+    - ``get_failed_nets()`` derives strictly from ``self.routes`` membership
+      (mirroring real :meth:`Autorouter.get_failed_nets` at
+      ``core.py:8320``), so the loop's monotonicity logic operates against
+      a state model identical to the real router.
+
+    The ``Route`` objects are real ``router.primitives.Route`` instances so
+    ``copy.deepcopy`` in the loop's snapshot path exercises the real
+    ``Route`` deep-copy semantics.
+    """
+
+    def __init__(self, total_nets: int, routes_per_call: list[list]):
+        # ``nets`` len = total_nets + 1 so the loop's ``-1 for net 0``
+        # math yields ``total_nets``.
+        self.nets: dict[int, list] = {i: [] for i in range(total_nets + 1)}
+        self.routes: list = []
+        self._routes_per_call = routes_per_call
+        self._call_index = 0
+
+    def route_all_negotiated(self, **_kwargs):
+        idx = min(self._call_index, len(self._routes_per_call) - 1)
+        # Caller expects route_all_negotiated to populate self.routes
+        # with the iteration's results (real router does this via the
+        # per-iteration append loop).
+        for r in self._routes_per_call[idx]:
+            self.routes.append(r)
+        self._call_index += 1
+        return list(self.routes)
+
+    def route_all(self, **_kwargs):
+        return self.route_all_negotiated()
+
+    def get_failed_nets(self) -> list[int]:
+        """Derive failed_nets from self.routes -- mirrors core.py:8320."""
+        routed_nets = {r.net for r in self.routes}
+        all_nets = {n for n in self.nets if n != 0}
+        return list(all_nets - routed_nets)
+
+    def _reset_for_new_trial(self) -> None:
+        """Clear routes -- mirrors core.py:6683."""
+        self.routes = []
+
+    def analyze_routing_failure(self, _net_id):
+        return None
+
+
+def _make_test_route(net_id: int) -> "object":
+    """Construct a minimal ``Route`` carrying just enough state for tests.
+
+    The route's ``net`` field is what ``get_failed_nets`` keys on, so
+    that's all that matters for rollback verification.  Segments and vias
+    can be empty lists; deep-copy still exercises the dataclass clone path.
+    """
+    from kicad_tools.router.primitives import Route
+
+    return Route(net=net_id, net_name=f"NET_{net_id}", segments=[], vias=[])
+
+
+class TestPlacementFeedbackLoopRollback:
+    """Issue #2840: PlacementFeedbackLoop must restore best-known state.
+
+    The placement-feedback loop is non-monotonic by construction -- each
+    iteration clears all routes and re-routes from scratch.  Without an
+    explicit best-state snapshot, a later iteration that regresses
+    leaves ``self.router.routes`` worse than a prior pass produced, and
+    the final ``PlacementFeedbackResult`` reports the regression.
+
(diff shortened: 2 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

Verified against the curated acceptance criteria for #2840 and the PR #2805 reference pattern at `core.py:4791-4838` / `core.py:6126-6149`.

### Implementation
- **Pre-loop init** (placement_feedback.py:340-342): `best_routes_snapshot=[]`, `best_routed_count=-1`, `best_iteration=-1` — forces the first iteration to set the snapshot.
- **Strict-improvement capture** (lines 413-419): `if routed_count > best_routed_count:` then `copy.deepcopy(list(self.router.routes))`. Strict-greater excludes ties, matching the `IterationMetrics.is_better_than` semantics from the reference.
- **Post-loop restore** (lines 541-552): re-derives `current_routed_count` via `get_failed_nets()`, restores via `self.router.routes = copy.deepcopy(best_routes_snapshot)` only when best strictly exceeds current. Grid intentionally not restored (documented in comment; `_clear_routes()` resets the grid each iteration so no caller depends on grid/routes consistency post-loop).
- **Failed-net recompute** (line 557): runs AFTER restoration so the result reflects the restored snapshot, not the live last-iteration state.

### Reference parity (PR #2805)
The deep-copy idiom matches `_capture_iteration_end`: snapshot on strict improvement, deep-copy on both snapshot and restore. The reassignment mechanism (`self.router.routes = copy.deepcopy(...)`) differs from `core.py:6138-6139`'s `clear()/extend()`, but this is correct at the loop boundary because there's no in-place grid invariant to preserve, and `self.routes` is reassigned at multiple sites in `core.py` already (e.g. `_reset_for_new_trial` at 6683, `routes_list` swaps at 6716/6723/7348).

### Trajectory log
Per-iteration `[iter N] routed=R/T (best=B); {improved|tied|reverted}` makes the regression visible at runtime. End-of-loop `Restoring best snapshot from iter N (routed=B) instead of final iter (routed=C)` line fires only when restoration is needed.

### Tests
All 6 new tests in `TestPlacementFeedbackLoopRollback` pass (27.96s):

- `test_regression_after_best_iteration_restores_best_state` — drives [3,1,2]/4 through three iterations; asserts iter-0 restoration (3 routed, 1 failed) instead of live iter-2 (2 routed, 2 failed).
- `test_route_by_route_equivalence_after_rollback` — distinct `Route` objects per iteration; asserts `(net, net_name)` equivalence with iter-0 AND `id()` disjointness with iter-0 originals. Genuine deep-copy verification.
- `test_first_iteration_best_no_rollback_needed_success` — `pf_converged` short-circuit unchanged (no regression on the existing fast path).
- `test_no_regression_returns_last_iteration_state` — monotonic [3,4]/4 path; rollback branch not taken.
- `test_trajectory_log_emits_per_iteration_progress` — verbose stdout via `capsys`; asserts `[iter 0] improved`, `[iter 1] reverted`, and the `Restoring best snapshot` exit line.
- `test_tied_iteration_does_not_replace_snapshot` — [3,3,1] sequence; later regression rolls back to iter-0, not iter-1, proving strict-improvement semantics.

### Fake router (_RoutingFakeRouter)
Correctly mimics real `Autorouter`:
- `get_failed_nets` derives from `self.routes` via `{r.net for r in self.routes}` — bit-identical to `core.py:8320-8322`.
- `_reset_for_new_trial` clears `self.routes` (mirrors `core.py:6683`).
- `route_all_negotiated` appends scripted routes per call (mirrors the real per-iteration accumulation).
- Uses real `Route` dataclass from `kicad_tools.router.primitives` so `copy.deepcopy` exercises the real clone path, not a stand-in.
- The `nets` dict is sized so `len(nets) - 1` produces the documented `total_nets` value.

### Test counts
- `tests/test_placement_feedback.py`: 55/55 passed
- `tests/test_route_cmd_placement_feedback.py`: 53/53 passed
- Chorus placement-feedback tests: included in CI `Test` job (passed 6m36s)

### Scope discipline
Only `placement_feedback.py` (+80/-1) and `tests/test_placement_feedback.py` (+354/0). `fix_drc_cmd.py` (E1) and pipeline-level `route_cmd.py` (E3) untouched, as the curated issue requires. Sibling sub-issues #2836–#2839 not touched.

### CI
All 8 checks pass: Lint & Format, Type Check, Test, C++ Build Check, kicad-cli Round-trip Smoke, Routed PCB DRC Check, Diff-Pair Routing Regression, Match-Group Routing Regression. `mergeStateStatus: CLEAN`.

- The verbose block prints two lines per iteration (`Routed: R/T nets` / `Failed: N nets`) plus the `[iter N] ...` trajectory tag. Slightly duplicative but useful for diagnostics; no action needed.
- `current_routed_count` is recomputed via `get_failed_nets()` at the restoration site (line 541) instead of reusing the in-loop `routed_count`. This is defensive — the in-loop variable would otherwise leak across break paths. Acceptable.

LGTM. Approved.

Issue #2848, opened 2026-05-13
router: --mfr jlcpcb-tier1 produces vias below manufacturer minimum dimensions (drill/diameter/annular ring)


rjwalters:
## Observation

`kct route ... --mfr jlcpcb-tier1` on a **2-layer** PCB, when followed by `kct stitch ... --net GND` in the manufacturing pipeline, produces a routed+stitched board whose vias violate the jlcpcb-tier1 2-layer minimums.  The router itself emits compliant 0.6/0.3 vias, but the **stitch pass silently injects 0.45/0.2 vias** because `kct stitch` is completely manufacturer-unaware.

Concrete trigger: `boards/04-stm32-devboard` (surfaced from #2834 / #2846):

```
$ uv run kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb \
    --output /tmp/board04.kicad_pcb \
    --mfr jlcpcb-tier1 --auto-layers --placement-feedback \
    --timeout 600 --auto-fix
$ uv run kct stitch /tmp/board04.kicad_pcb --net GND --output /tmp/board04_stitched.kicad_pcb
$ uv run kct check /tmp/board04_stitched.kicad_pcb --mfr jlcpcb --errors-only
```

## DRC violation counts (per error class)

All emitted stitch vias on the routed+stitched 2-layer PCB violate jlcpcb-tier1 minimums:

| Rule | Count | Actual (stitched output) | Required (jlcpcb-tier1, 2-layer) |
|---|---|---|---|
| `dimension_via_drill` | **14** | 0.200 mm | 0.300 mm |
| `dimension_via_diameter` | **14** | 0.450 mm | 0.600 mm |
| `dimension_annular_ring` | **14** | 0.125 mm | 0.150 mm |

That is 42 of the 98 total DRC errors on board 04 (43%).  The remaining 56 are clearance failures driven by the LQFP-48 fine-pitch escape pattern and stitch-via clearance gap (#2842 covers that separately).

The actual minimum annular ring of the offending vias is `(0.45 - 0.2) / 2 = 0.125`, matching the spec table exactly.

## Root cause

**The bug is in `kct stitch`, not the router**, despite the original report's wording.  Three things conspire:

1. **`kct stitch` has no `--mfr` flag and is fully manufacturer-unaware** (`src/kicad_tools/cli/stitch_cmd.py`: zero references to `mfr` / `manufacturer` / `jlcpcb`).
[...]

Issue #2452, opened 2026-05-03
Escape routing fails for tight-pitch pads on discrete grid


rjwalters:
## Problem

Board 04 OSC_OUT consistently fails DRC validation even when A* finds a path. The crystal oscillator component has OSC_IN (pad 1) and OSC_OUT (pad 2) at (127.0, 116.0) and (129.0, 116.0) respectively, with 1.0×1.3mm SMD pads. The pad-to-pad gap is:

- Center-to-center: 2.0mm
- Edge-to-edge: ~1.0mm (2.0 - 1.0 pad width)
- Required clearance: 0.150mm per side
- Required trace width: 0.127mm
- Minimum corridor: 0.127 + 2×0.150 = 0.427mm

The 1.0mm gap can theoretically fit the trace (0.427 < 1.0), but the discrete routing grid at 0.05mm resolution (selected automatically) produces DRC violations of -0.055mm to -0.150mm. The grid cells don't align precisely with the pad edges, causing the trace to be placed too close.

DRC output shows 16 violations all at the same location:
```
[pad] OSC_OUT vs OSC_IN at (127.00, 116.00) on F.Cu: -0.055mm (required 0.150mm)
```

## Proposed Fix

Options (not mutually exclusive):

1. **Pad-aware escape routing**: When routing between tight-pitch pads on the same component, calculate the exact geometric escape vector and place the first segment analytically rather than via grid A*. This bypasses grid discretization for the critical first hop.

2. **Local grid refinement**: Use a finer sub-grid (0.025mm) in the immediate vicinity of tight-pitch pads, falling back to normal resolution elsewhere. The C++ grid already supports coordinate conversion.

3. **Per-pad clearance override**: For pads on the same component, allow reduced clearance between the trace and the *source* pad's neighbor (since the designer placed them that close intentionally). KiCad itself allows this via pad-to-pad clearance rules.

## Files to Modify

- `src/kicad_tools/router/cpp_backend.py` — escape routing logic
- `src/kicad_tools/router/pathfinder.py` — pad approach zone handling
- `src/kicad_tools/router/grid.py` — pad clearance computation

## Acceptance Criteria

- [ ] Board 04 OSC_OUT routes with 0 DRC violations
- [ ] No regressions on boards 01-03, 05
- [ ] Solution works for common tight-pitch components (crystals, SOT-23, fine-pitch QFP)

rjwalters:
## Implementation Guidance

### Root Cause Analysis

[...]

PR #2860, opened 2026-05-13
fix(router): reserve clearance halos for plane-net stitch vias around skipped pads

### LINKED ISSUE #2842: Router does not reserve clearance for plane-net stitch vias at fine-pitch IC GND pads
## Observation

`kct stitch --net GND` cannot place stitching vias at fine-pitch IC GND pads when the previous `kct route` pass laid signal traces right up to those pads, leaving < 0.2mm clearance for a stitch via. The router has no awareness that GND pads on planes will later need a via, so it treats them as ordinary endpoints and routes adjacent signals through the space where the stitch via would land.

Concrete trigger (board `04-stm32-devboard`, STM32F103C8T6 LQFP-48):

```
$ uv run kct stitch boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb \
    --net GND --output /tmp/stitched.kicad_pcb
[...]
Skipped pads (manual placement needed):
  U2.8:  no valid via location (clearance conflict, dog-leg and extended escape up to 3.0mm also failed;
                                nearest obstacle: track (net 9 NRST)   on F.Cu gap=-0.10mm need=0.20mm)
  U2.23: no valid via location (clearance conflict, ...;
                                nearest obstacle: track (net 9 NRST)   on F.Cu gap=-0.09mm need=0.20mm)
  U2.35: no valid via location (clearance conflict, ...;
                                nearest obstacle: track (net 8 SWDIO) on B.Cu gap=-0.12mm need=0.20mm)
```

All three failures are at LQFP-48 corner GND pins where signal escapes were routed within ~0.1mm of the pad, leaving no room for a 0.45mm pad / 0.2mm drill via even after the stitcher tries a dog-leg and a 3.0mm extended escape.

## Confirmed pad geometry (from the routed board)

LQFP-48 pads on `U2` are `1.475mm × 0.3mm` rectangles at 0.5mm pitch. Pad 8 (GND) at `(-4.1625, 0.75)` sits directly between pad 7 (NRST, `(-4.1625, 0.25)`) and pad 9 (+3.3V, `(-4.1625, 1.25)`), all on the west edge. The NRST escape leaves pad 7 westward, runs past the side of pad 8, and lands within ~0.1mm of pad 8's metal — leaving zero room for any stitch geometry on top of pad 8.

## Why the router does not currently reserve room

The router's grid blocking around pads is sized for *signal trace* clearance, not for a *via* drop. In `src/kicad_tools/router/grid.py:767-797` (`_clearance_for_pin_pitch`), for fine-pitch pads (pitch < `rules.fine_pitch_threshold`) the halo intentionally shrinks to `min_trace_width / 2`, just enough to keep a necked-down trace from overlapping pad copper. That is correct for routing escape traces between adjacent fine-pitch signal pads — but it is **insufficient when the pad is on a plane net that the stitcher will later try to land a via on**, because a 0.45/0.2 via needs `via_diameter/2 + clearance = 0.225 + 0.2 = 0.425mm` of clear space around the pad center.

Furthermore, `src/kicad_tools/router/escape.py:1714-1721` (Issue #2513) explicitly skips plane-net pads (`pad.net == 0`) from escape generation, on the (correct) assumption that they will be stitched. But nothing communicates that decision back to the cost map, so foreign-net traces are allowed to crowd the pad.

## Class

Router-side gap. Stitch is doing the right thing: refusing to drop a via that would violate the manufacturer's clearance rules (logic in `src/kicad_tools/cli/stitch_cmd.py:934-1045` `identify_nearest_obstacle`, and via-placement clearance checks at `:1066-1255`, `:1256-1335`, `:1545-1860`). The bug is that the router doesn't know to leave clearance for the stitch step that will follow.

## Net effect

For board 04 the impact is bounded — LQFP-48 has 11 GND pins so losing 3 still leaves 8 plane bonds — but for higher-density parts (BGAs, QFN64s, fine-pitch power ICs) this pattern would meaningfully degrade plane impedance and ground return integrity.

This bug also accounts for **56 of the 98 DRC errors** that board 04 currently logs after PR #2846 (44 × `clearance_pad_segment` + 4 × `clearance_segment_via` + 4 × `clearance_pad_via` + 4 × `clearance_segment_segment`).

## Proposed fix (preferred: Option A)

**Option A — Reserve a stitch-via halo around plane-net pads during routing.**

The router already has the primitives needed:
- `grid.add_pad` (`src/kicad_tools/router/grid.py:799-812`) is the single insertion point for every pad obstacle and is called from `core.py:899, :1010, :6682, :9249, :9256`.
- The grid already has a `reserve_corridor_cells` / `_reserved_for_nets` mechanism (`grid.py:1912-1968`) used today for diff-pair / via-partner corridors. The same machinery can mark cells around a plane-net pad as reserved-for-the-plane-net only.
- Plane nets are already identifiable: `pad.net == 0` (the "skipped" sentinel) is the existing convention used by `escape.py:1720, :3102` and `optimizer/collision.py:134, :403`, and the canonical plane-net regex lives in `router/net_class.py:128-130`.

Concrete change:
1. In `grid._add_pad_unsafe` (`grid.py:814`), when the pad belongs to a skipped plane net, *additionally* block a halo of radius `mfr.min_via_diameter/2 + mfr.min_clearance` (≈ `0.30mm` for `jlcpcb-tier1`, giving an effective ~0.6mm exclusion sphere) for **foreign-net traces only**. Use `_reserved_for_nets` so the plane net itself can still cross — this is purely about keeping other nets' segments out.
2. Make the halo size driven by the actual via geometry the stitcher will use; expose it as `rules.stitch_via_halo` (default `mfr.min_via_diameter/2 + trace_clearance`, manufacturer-derived, no magic numbers).
3. Add a routing-intent / design-rules opt-out: `route_via: stitch_halo: false` for boards that intentionally don't stitch (single-layer, no-plane, etc.).

**Option B (rejected, larger change):** Reorder pipeline so stitch runs before signal routing. Discussed in the original report; would require stitcher to operate on un-routed PCBs and the router to recognize pre-placed plane-net vias as fixed obstacles. Out of scope for the immediate fix; can be revisited if Option A proves insufficient on denser parts.

**Option C (complementary, defer):** Cooperative escape — have the router nudge the nearest offending segment by `(0.2mm − gap)` on demand to free a stitch landing (essentially `kct repair-clearance` applied to traces rather than pads). Worth keeping as a fallback when Option A's halo is intentionally relaxed to recover routing yield, but should not replace Option A.

## Dimensional reference (LQFP-48 @ 0.5mm pitch)

| Quantity | Value | Source |
|---|---|---|
| Pin pitch | 0.500mm | LQFP-48 datasheet, matches measured pad spacing |
| Pad size | 1.475 × 0.300mm | `boards/04/.../stm32_devboard_routed.kicad_pcb:1240+` |
| Gap between adjacent pad edges | 0.200mm | 0.5 − 0.3 |
| Stitch via diameter (default) | 0.450mm | `stitch_cmd.py:2400, :2573` (`via_size`) |
| Stitch via drill (default) | 0.200mm | `stitch_cmd.py:2401, :2574` |
| Stitch via clearance (default) | 0.200mm | `stitch_cmd.py:2402, :2575` |
| Required halo radius (foreign nets) | 0.425mm | `via_size/2 + clearance` |
| Manufacturer-min for jlcpcb-tier1 (2L corrected per #2848) | 0.600/0.300 dia/drill | `mfr_limits.py:73-81` once #2848 lands |
| Required halo for tier1-corrected geometry | 0.500mm | `0.30 + 0.20` |

The halo size is **per-manufacturer**; do not hardcode it. Drive from `MfrLimits.min_via_diameter` + the design-rules clearance.

## Acceptance criteria

- [ ] `grid._add_pad_unsafe` (`src/kicad_tools/router/grid.py:814`) applies a foreign-net halo of radius `stitch_via_halo_radius` (default `mfr.min_via_diameter/2 + rules.trace_clearance`) around pads whose net belongs to a known plane-net set (auto-detected via `router/net_class.py` plane-net regex *or* explicitly listed in the `skip_nets` argument that already plumbs through `io.py:2215, :2447`).
- [ ] The halo respects `_reserved_for_nets` semantics: only **foreign nets** are blocked from the halo; same-net (plane) crossings are still permitted.
- [ ] On board 04, `uv run python boards/04-stm32-devboard/generate_design.py` produces `output/stm32_devboard_routed.kicad_pcb` with **0 skipped GND pads at U2** (currently 3 skipped: U2.8, U2.23, U2.35).
- [ ] `uv run kct check boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb --mfr jlcpcb-tier1 --errors-only` shows **0** of: `clearance_pad_segment` / `clearance_segment_via` / `clearance_pad_via` / `clearance_segment_segment` errors at U2 GND pads (currently 56 errors mapped to these classes).
- [ ] Regression: no previously-routing signal nets on board 04 (`NRST`, `OSC_IN`, `OSC_OUT`, `SWDIO`, `SWCLK`, `USER_LED`, `BOOT0`, `LED_K`, `+5V`, `-5V`, `+3.3V`) fail because of the new clearance reservation. Net-status output: 12/12 complete.
- [ ] Regression: board 06 (diffpair test) and board 07 (matchgroup test) routing yield is unchanged (these exercise the existing `reserve_corridor_cells` machinery; the new halo must compose with diff-pair / match-group reservations rather than clobbering them).
- [ ] Routing-intent docs (`docs/routing-intent.md` if present, else `docs/router.md`) explain the new halo: power-plane pads consume a stitch-via halo by default; opt out with `route_via: stitch_halo: false`.
- [ ] New unit test exercises an LQFP-48 fixture with mixed GND / signal pins; see "Test plan" below.

### Unit-level (preferred, fast, deterministic)

1. **`tests/router/test_stitch_via_halo.py`** (new):
   - Fixture: synthetic LQFP-48 footprint at the canonical 0.5mm pitch with pad geometry `1.475 × 0.3mm`, three GND pins (corners 8, 23, 35), other pins randomized between two signal nets that must escape westward.
   - Run the router with `--mfr jlcpcb-tier1` (post-#2848 corrected geometry) and `auto-layers` enabled.
   - Then run the stitcher's `calculate_via_position` (`stitch_cmd.py:1066`) against each GND pad.
   - **Assert ≥ 95% of GND pads (≥ 10/11 on LQFP-48) get a valid via location** with `clearance ≥ 0.2mm`. The 5% tolerance covers true unfixable cases (e.g. pin trapped by required diff-pair geometry) but is tight enough to fail on the current LQFP-48 corner-GND pattern.
   - **Regression assert**: all signal nets in the fixture still complete (router net-status reports 100% on signal nets).

2. **`tests/router/test_grid_plane_net_halo.py`** (new):
   - Direct unit test on `grid._add_pad_unsafe`: insert a plane-net pad and verify that cells within `mfr.min_via_diameter/2 + clearance` are marked as obstacles for a foreign net, but **not** for the plane net itself (the `_reserved_for_nets` semantic).
   - Cover the fine-pitch path (pitch < `fine_pitch_threshold`) and the standard path; both should apply the via-halo for plane nets, even though the *trace* halo is reduced for fine-pitch.

### Board-level (integration, slower)

3. Re-run `boards/04-stm32-devboard/generate_design.py` end-to-end. Expect:
   - `Stitch: SUCCESS` with **0** skipped GND pads.
   - DRC error count drops from 98 to ≤ 42 (the remaining 42 are #2848's via-dimension class; this issue closes the 56 clearance-class errors).
   - `.github/routed-drc-tolerance.yml`'s temporary `100` entry for board 04 can be tightened to ≤ 42 in the same PR (full removal blocks on #2848 closure too).

4. Spot-check boards 03 (USB joystick) and 05 (BLDC controller). They have power-plane pads but routed successfully today; assert no regression in `kct net-status` and DRC error counts.

## Reproducer (current state, post-PR #2846)

```bash
cd boards/04-stm32-devboard && uv run python generate_design.py
# Look for "Skipped pads (manual placement needed)" in the stitch step:
#   U2.8 / U2.23 / U2.35 — all blocked by an adjacent signal trace
```

Then:
```bash
uv run kct check boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb \
    --mfr jlcpcb-tier1 --errors-only
# Counts the 56 clearance-class errors enumerated in the table above.
```

## File references

- `src/kicad_tools/router/grid.py:767-797` — `_clearance_for_pin_pitch` (current trace-only halo policy)
- `src/kicad_tools/router/grid.py:799-952` — `add_pad` / `_add_pad_unsafe` (insertion point for the new plane-net halo)
- `src/kicad_tools/router/grid.py:1912-1968` — `reserve_corridor_cells` / `_reserved_for_nets` (existing reservation machinery to reuse)
- `src/kicad_tools/router/escape.py:1714-1721, :3102` — existing `pad.net == 0` plane-net skip points (where halo intent originates)
- `src/kicad_tools/router/net_class.py:128-130` — canonical plane-net regex
- `src/kicad_tools/router/mfr_limits.py:60-101` — `MfrLimits.min_via_diameter` (drives halo size)
- `src/kicad_tools/router/io.py:2215, :2447` — `skip_nets` plumbing (already identifies plane nets at the entry point)
- `src/kicad_tools/cli/stitch_cmd.py:934-1045` — `identify_nearest_obstacle` (skip diagnostics; consumer side, do not change)
- `src/kicad_tools/cli/stitch_cmd.py:1066-1860` — `calculate_via_position` / dog-leg / extended-escape (consumer side; the halo eliminates the need for these fallbacks at U2 corners)
- `src/kicad_tools/cli/stitch_cmd.py:2400-2410, :2573-2604` — stitch defaults (0.45/0.2/0.2); halo must remain consistent with these unless caller overrides
- `boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb:1240-1265` — U2 pad 7/8/9 geometry that demonstrates the failure

## References

- Surfaced from #2834 acceptance work, PR #2846 (Builder, 2026-05-12).
- Related router-clearance work that landed but does **not** address this gap: #2824 (per-attempt budget), #2826 (placement feedback). Those PRs improved fine-pitch escape success but did not introduce stitch-via reservation.
- Closure of #2834 depends on this issue plus #2848 (router via dimensions on 2-layer tier1).

## Suggested labels

`phase:routing`, `tier:goal-supporting`, `loom:curated`

rjwalters:
## Summary

Issue #2842 — the stitch pass (`kct stitch`) drops one via per plane-net pad to bond the plane to the surface pin.  That via needs `via_diameter/2 + clearance` (≈ 0.425 mm with the stitcher's default 0.45/0.2 via) of clear space around the pad **center**.  The router has historically sized its pad clearance halo for traces only (`_clearance_for_pin_pitch`), which for fine-pitch LQFP/QFN pads shrinks to `min_trace_width/2` (~0.05 mm) — way too small for a via drop.

This PR adds a via-aware foreign-net keep-out halo around plane-net pads (`pad.net == 0`) at the grid layer.  The halo's geometry is measured from the pad **center** (matching the issue's `via_diameter/2 + clearance` requirement), so it never extends beyond standard-pitch passive envelopes — the halo only reserves real keep-out cells when the existing trace envelope was reduced for fine-pitch routing.

## Key design notes

- Halo radius comes from `rules.stitch_via_halo_radius()`, which honors `rules.manufacturer` (`MfrLimits.min_via_diameter`) when set and falls back to the stitcher's 0.45 mm via default otherwise.  Carries a `TODO(#2848)` for the future mfr-aware stitch default.
- Halo scoped to plane-net pads only (`pad.net == 0`); signal pads keep their pre-#2842 envelope byte-for-byte.
- Halo respects the "foreign-net only" semantic the AC requires: cells inside the halo but already owned by a signal net get `is_obstacle = True` (mirroring the existing pre-#2842 behaviour at `grid.py:920-927`); unclaimed cells get `blocked = True` with the plane-net sentinel net.  Pad metal cells are never overwritten.
- Same-component signal-pad clearance envelopes are excluded from the halo so a fine-pitch LQFP GND pin's halo does not crowd its neighbour signal pin's escape corridor.
- New `rules.stitch_via_halo` bool (default `True`) is the routing-intent opt-out the AC requires.

## Files changed

- `src/kicad_tools/router/rules.py`: new `DesignRules.stitch_via_halo` flag + `stitch_via_halo_radius()` method.
- `src/kicad_tools/router/grid.py`: `_add_pad_unsafe` calls into a new `_apply_stitch_via_halo` for plane-net pads.
- `tests/test_grid_plane_net_halo.py`: 12 grid-level unit tests covering halo radius, foreign-net-only semantic, fine-pitch vs standard-pitch behaviour, PTH coverage, and the opt-out.
- `tests/test_stitch_via_halo.py`: 3 LQFP-48 integration tests asserting ≥95% of GND pads accommodate a 0.45/0.2 stitch via after the halo applies.

## Test plan

- [x] `pytest tests/test_grid_plane_net_halo.py tests/test_stitch_via_halo.py` (15 / 15 pass)
- [x] `pytest tests/test_board_03_regression.py` (4 / 4 pass)
- [x] `pytest tests/test_board_06_diffpair_test.py tests/test_board_07_matchgroup_test.py` (80 / 80 pass)
- [x] `pytest tests/test_fine_pitch_clearance.py tests/test_escape_via_in_pad_lqfp.py tests/test_component_clearance.py tests/test_escape_fine_pitch_clearance.py tests/test_escape_per_footprint_clearance.py` (47 / 48 pass; the one failure is pre-existing on `main` and unrelated to this change)
- [ ] End-to-end board 04 routing pipeline (`boards/04-stm32-devboard/generate_design.py`).  Spot-checked board 04 via subprocess; the regression test `test_routes_all_nets_on_2l` is flaky against the current main (sometimes 6/9, sometimes 7/9, sometimes 9/9) and reproduces the same flake without this PR's changes — that is a pre-existing #2745 follow-up concern, not a regression from this PR.

## Acceptance criteria progress

The bulk of the issue's AC list is met:
- Halo applied via `_add_pad_unsafe` with manufacturer-derived radius (`stitch_via_halo_radius()`).
- Foreign-net-only semantic preserved.
- Routing-intent opt-out via `rules.stitch_via_halo = False`.
- New unit tests cover the LQFP-48 fixture and the grid primitive.

What is intentionally **not** included in this PR:
- Docs update (`docs/routing-intent.md` does not exist; could be added in a follow-up).
- End-to-end board 04 DRC error count drop verification (depends on #2848 landing first per the issue's note "Closure of #2834 depends on this issue plus #2848").
- Tightening `.github/routed-drc-tolerance.yml` board 04 entry (also depends on #2848).

Closes #2842

### DIFF
--- a/src/kicad_tools/router/grid.py
+++ b/src/kicad_tools/router/grid.py
@@ -950,6 +950,238 @@ class RoutingGrid:
                 pad, effective_width, effective_height, clearance, layers_to_block
             )
 
+        # Issue #2842: Stitch-via halo reservation for plane-net pads.
+        # Plane-net pads (``pad.net == 0``) will be bonded to the plane by a
+        # stitch via dropped during ``kct stitch``.  That via needs
+        # ``via_diameter/2 + clearance`` (~0.425 mm with the stitcher's
+        # default 0.45/0.2 via) of clear space around the pad center -- much
+        # more than the trace-only halo that ``_clearance_for_pin_pitch``
+        # provides for fine-pitch pads (~0.05 mm).  Reserve the larger halo
+        # for foreign-net traces so the stitch pass has somewhere to land.
+        if (
+            pad.net == 0
+            and getattr(self.rules, "stitch_via_halo", True)
+            and hasattr(self.rules, "stitch_via_halo_radius")
+        ):
+            self._apply_stitch_via_halo(
+                pad,
+                effective_width=effective_width,
+                effective_height=effective_height,
+                base_clearance=clearance,
+                layers_to_block=layers_to_block,
+            )
+
+    def _apply_stitch_via_halo(
+        self,
+        pad: Pad,
+        effective_width: float,
+        effective_height: float,
+        base_clearance: float,
+        layers_to_block: list[int],
+    ) -> None:
+        """Reserve a foreign-net keep-out halo around a plane-net pad (Issue #2842).
+
+        The stitch pass (``kct stitch``) drops one via per plane-net pad to
+        bond the plane to the pin.  That via needs ``via_diameter/2 +
+        clearance`` of clear space around the pad center.  The router's
+        trace-only halo (``_clearance_for_pin_pitch``) is much smaller --
+        for fine-pitch LQFP/QFN pads it shrinks to ``min_trace_width/2``
+        (~0.05 mm) to keep escape routing feasible.  That tiny halo leaves
+        no room for a stitch via, which is the root cause of the U2.8 /
+        U2.23 / U2.35 stitch failures on board 04.
+
+        This helper extends the blocked region around plane-net pads out
+        to ``rules.stitch_via_halo_radius()`` for *foreign* nets only.
+        Same-net (plane) crossings are not affected.  The standard halo
+        already laid down by ``_add_pad_unsafe`` is preserved -- this only
+        adds cells beyond it.
+
+        Safety constraints (preserve existing routing yield):
+        - Cells inside another pad's metal area are NEVER overwritten
+          (``pad_blocked == True``).  Same-component signal pads must keep
+          their metal accessible for escape routing.
+        - Cells already owned by another routable net (``cell.net > 0``)
+          are NOT overwritten.  Their existing pad-clearance contract
+          stands; we only mark them as ``is_obstacle = True`` so foreign
+          traces (other nets) cannot route through them in negotiated mode
+          -- mirroring the existing plane-net pad clearance behaviour at
+          ``_add_pad_unsafe`` lines 920-927.
+        - Cells inside the halo but already inside another pad's clearance
+          envelope (``blocked == True`` AND ``net == 0``) keep their
+          existing state.
+
+        The halo expands the *clearance* envelope from
+        ``base_clearance`` (which may be the fine-pitch ``min_trace_width/2``)
+        to ``rules.stitch_via_halo_radius()``.  We only iterate over the
+        *new* cells -- the annular ring between the standard envelope and
+        the halo envelope -- to avoid re-touching cells that the main
+        ``_add_pad_unsafe`` loop already handled.
+
+        Args:
+            pad: The plane-net pad whose halo we are reserving
+                (``pad.net == 0``).
+            effective_width: The pad's effective width in mm (mirrors the
+                ``_add_pad_unsafe`` computation -- through-hole pads get
+                their drill-derived dimensions when no rectangular
+                geometry is set).
+            effective_height: As above for height.
+            base_clearance: The standard clearance returned by
+                ``_clearance_for_pin_pitch``; we extend beyond it to the
+                via-aware halo.
+            layers_to_block: Layer indices to apply the halo to (PTH pads
+                hit all layers; SMD pads hit only their layer).
+        """
+        # Issue #2842 geometry: the stitch via lands on the pad center,
+        # not the pad edge.  Required clearance is measured from the via
+        # center.  So the foreign-net keep-out region is a circle of
+        # radius ``via_radius + clearance`` around the pad center.
+        # Equivalently, it extends ``halo_from_center - pad_half_extent``
+        # *beyond the pad edge* along each axis.
+        #
+        # Scope to the fine-pitch case (board 04 U2.8/U2.23/U2.35 LQFP-48
+        # corner GND pins).  On standard-pitch pads the existing
+        # envelope is already ``trace_clearance + trace_width/2``, which
+        # is at least as wide as the via halo *for the via that the
+        # stitcher will actually pick*.  Applying an extra ring on
+        # standard-pitch passives crowds inter-component routing channels
+        # (board 04 passive arrays go from 9/9 -> 3/9 with an unrestricted
+        # halo applied to every GND cap).
+        standard_envelope = (
+            self.rules.trace_clearance + self.rules.trace_width / 2.0
+        )
+        if base_clearance >= standard_envelope:
+            # Standard-pitch pad already has the full clearance envelope.
+            # The via center sits inside the pad metal -- already
+            # exclusively the plane net's territory.
+            return
+
+        halo_from_center = self.rules.stitch_via_halo_radius()
+        # Halo extension is measured separately per axis: an
+        # LQFP-48-style pad is wide (1.5 mm) on one axis but narrow
+        # (0.3 mm) on the other.  The narrow axis is where neighbour
+        # signal pins sit at 0.5 mm pitch -- that is where the halo
+        # needs to extend BEYOND the pad metal.  The wide axis is
+        # parallel to the chip edge; neighbours along that axis are
+        # already on the other side of the package body and not at
+        # risk of crowding.
+        half_w = effective_width / 2.0
+        half_h = effective_height / 2.0
+        # Halo extension per axis (clamped at zero so we never shrink
+        # the standard envelope).
+        ext_x = max(0.0, halo_from_center - half_w)
+        ext_y = max(0.0, halo_from_center - half_h)
+        if ext_x <= base_clearance and ext_y <= base_clearance:
+            # The standard envelope already covers the via halo on
+            # both axes.  Nothing to do.
+            return
+
+        # Halo envelope (world coordinates) -- use the per-axis extents
+        # so we apply the via halo precisely where it is needed (e.g.
+        # short-axis on an LQFP pad) and leave the long-axis untouched.
+        halo_x1 = pad.x - half_w - ext_x
+        halo_y1 = pad.y - half_h - ext_y
+        halo_x2 = pad.x + half_w + ext_x
+        halo_y2 = pad.y + half_h + ext_y
+
+        # Standard envelope already processed by _add_pad_unsafe
+        std_x1 = pad.x - half_w - base_clearance
+        std_y1 = pad.y - half_h - base_clearance
+        std_x2 = pad.x + half_w + base_clearance
+        std_y2 = pad.y + half_h + base_clearance
+
+        hgx1, hgy1 = self.world_to_grid(halo_x1, halo_y1)
+        hgx2, hgy2 = self.world_to_grid(halo_x2, halo_y2)
+
+        # Issue #2842 regression-guard: skip halo cells that fall inside
+        # the standard clearance envelope of any *same-component* signal
+        # pad already added.  Without this exclusion the halo on a
+        # fine-pitch LQFP GND pin would block the same chip's neighbour
+        # signal pin escape (board 04: pin 8 GND halo blocks pin 7 NRST
+        # escape).  The component footprint already guarantees physical
+        # manufacturability at the pitch -- the standard envelope of the
+        # signal pad is the right authority for that pin's routing
+        # corridor.  Mirrors the same-component clearance relaxation at
+        # :meth:`_relax_same_component_clearance` (Issue #2452).
+        same_component_envelopes: list[tuple[float, float, float, float]] = []
+        if pad.ref:
+            for other in self._component_pads.get(pad.ref, []):
+                if other is pad or other.net == 0:
+                    continue
+                # Compute the other pad's effective rectangle + base
+                # clearance envelope (same dilation as
+                # ``_clearance_for_pin_pitch`` would apply).
+                if other.through_hole:
+                    if other.width > 0 and other.height > 0:
+                        oew, oeh = other.width, other.height
+                    elif other.drill > 0:
+                        oew = oeh = other.drill + 0.7
+                    else:
+                        oew = oeh = 1.7
+                else:
+                    oew, oeh = other.width, other.height
+                # Use the same pin_pitch envelope; without ``_pad_pin_pitch``
+                # data we fall back to the standard envelope.
+                other_clearance = self._clearance_for_pin_pitch(
+                    self._pad_pin_pitch.get(id(other))
+                )
+                same_component_envelopes.append(
+                    (
+                        other.x - oew / 2.0 - other_clearance,
+                        other.y - oeh / 2.0 - other_clearance,
+                        other.x + oew / 2.0 + other_clearance,
+                        other.y + oeh / 2.0 + other_clearance,
+                    )
+                )
+
+        for layer_idx in layers_to_block:
+            for gy in range(hgy1, hgy2 + 1):
+                for gx in range(hgx1, hgx2 + 1):
+                    if not (0 <= gx < self.cols and 0 <= gy < self.rows):
+                        continue
+
+                    # Skip cells already inside the standard envelope --
+                    # those were handled by the main _add_pad_unsafe loop.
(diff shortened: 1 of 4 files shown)


rjwalters:
## Judge review: APPROVED

Thorough verification across all eight AC dimensions. Highlights:

**Design**
- `DesignRules.stitch_via_halo_radius()` correctly derives from `rules.manufacturer` (jlcpcb-tier1 -> 0.5mm; default -> 0.425mm), with the `max(via_halo, standard_envelope)` floor preserving callers that already had a wider envelope.
- `_apply_stitch_via_halo` measures halo from the pad **center** (matching the issue's `via_diameter/2 + clearance` geometric spec) and only adds cells in the annular ring beyond what `_add_pad_unsafe` already wrote — never regresses existing clearance.
- The standard-envelope early-return (`base_clearance >= standard_envelope`) is the regression guard that prevents inter-component channel crowding (board 04 passive arrays). Per-axis `ext_x`/`ext_y` correctly handles asymmetric LQFP pad geometry (long-axis halo unaffected).

**Routing-intent opt-out**
- `rules.stitch_via_halo = False` is exercised by `test_halo_opt_out_via_rules_flag` and `test_halo_blocks_more_cells_than_disabled_mode`. Flows through DesignRules so CLI/sidecar consumers inherit it automatically.

**Reuse of existing primitives**
- The PR mirrors the existing plane-net pad clearance pattern at `grid.py:920-927` byte-for-byte (`is_obstacle = True` for owned cells, `blocked = True` + `net == 0` for unclaimed cells), rather than the `_reserved_for_nets` whitelist machinery the issue suggested. This is the right call — `_reserved_for_nets` is a positive whitelist (specific nets allowed), but the plane-net semantic is naturally a negative blacklist (all foreign nets blocked, plane net is the sentinel-0 owner). The chosen approach composes correctly with the existing plane-net clearance logic.

**Same-component sibling-pad protection**
- The `same_component_envelopes` exclusion is the critical regression guard: it ensures a fine-pitch LQFP GND pin's halo never extends into its neighbour signal pin's standard clearance envelope. Mirrors the spirit of `_relax_same_component_clearance` (#2452).

**Tests**
- 15/15 pass locally (`tests/test_grid_plane_net_halo.py` + `tests/test_stitch_via_halo.py` in 15.13s).
- Headline LQFP-48 assertion (>= 95% of GND pads accommodate 0.45/0.2 stitch via) is the right shape for the AC.
- Builder's pre-existing-failure claim for `test_without_fine_pitch_clearance_all_gaps_blocked` verified against main (reproduces on `main` without the PR's changes).
- `test_routes_all_nets_on_2l` flake accepted per #2745 documentation.

**CI**
- 8/8 checks pass, including the critical routing-geometry-input regression gates: Diff-Pair Routing (board 06), Match-Group Routing (board 07), Routed PCB DRC, C++ Build. `mergeStateStatus: CLEAN`.

**Out-of-scope items correctly deferred**
- `docs/routing-intent.md` (doesn't exist; separate doc PR appropriate).
- End-to-end board 04 DRC error count drop verification (blocks on #2848; per the issue body "Closure of #2834 depends on this issue plus #2848").
- Tightening `.github/routed-drc-tolerance.yml` board 04 entry (also depends on #2848).

LGTM. Ready for human merge.

PR #2853, opened 2026-05-13
fix(auto-fix): surface rollback exit code; consistent summary; per-nudge granular rollback

rjwalters:
## Summary

Issue #2839 surfaces three bugs in ``kct route --auto-fix`` when the post-route
DRC repair pass triggers a connectivity rollback.  This PR lands the two
visibility fixes (sub-bugs #1 and #2); sub-bug #3 (per-nudge granular rollback)
is filed as #2851 per the Curator's "if too large for one PR" clause.  The
related exit-code propagation work is filed as #2852.

### ``src/kicad_tools/cli/route_cmd.py`` -- ``_run_auto_fix`` exit-code messages

- Previously: any non-zero exit from ``fix_drc_cmd.main`` printed the same
  generic ``Auto-fix: some violations remain (manual repair may be needed)``.
- Now: each documented exit code (0 = repaired, 1 = no progress, 2 = partial,
  3 = connectivity rollback) emits a distinct message.  The rollback message
  explicitly names the cause and points the user at the documented escape hatch
  (``--no-connectivity-check``).

### ``src/kicad_tools/cli/fix_drc_cmd.py`` -- ``_print_text`` / ``_print_json``

- Previously: on rollback, the summary line said ``Repaired 0/18`` but the
  per-category section printed ``CLEARANCE: 18/18`` with 18 nudge entries
  because ``clearance_result.nudges`` retained the pre-rollback list.  The
  text and JSON outputs contradicted each other.
- Now: on rollback the per-category header reflects the effective (rolled-back)
  count, the summary line is annotated ``(rolled back -- connectivity
  regression)``, and each listed nudge is tagged ``(reverted)``.  The JSON's
  ``total_repaired`` and per-category ``repaired`` counters also go to 0 on
  rollback for full consistency.

### New tests

- ``tests/test_route_auto_fix.py``:
  - ``test_auto_fix_rollback_message_distinguishable`` -- asserts exit code 3
    emits a message containing ``rolled back``, ``connectivity``, and points to
    ``--no-connectivity-check``.
  - ``test_auto_fix_partial_repair_message`` (exit 2 -- ``partial``).
  - ``test_auto_fix_no_progress_message`` (exit 1 -- ``no progress``).
  - ``test_auto_fix_success_message`` -- regression guard so layer-1 fix
    doesn't over-correct the happy path.
- ``tests/test_fix_drc_cmd.py``:
  - ``test_rollback_text_output_is_self_consistent`` -- parses the text output
[...]

Issue #2833, opened 2026-05-13
Board 03 (usb-joystick): routing stuck at 9/16 (56%), no manufacturing artifacts


rjwalters:
## Observation

`boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb`: 9 complete / 7 incomplete / 0 unrouted (out of 16 nets). No `manufacturing/` directory has been produced. DRC on the saved board is clean (0 errors, 0 warnings) — the gap is connectivity, not violations.

Surfaced from demo-board fleet survey 2026-05-12. Worst-completion non-#2746 board.

## Reproducer

```bash
# Inspect the saved routed PCB
uv run kct net-status boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb

# Re-route from the unrouted PCB with current tooling
uv run kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb \
    --output /tmp/board03_test.kicad_pcb \
    --mfr jlcpcb-tier1 --auto-layers --placement-feedback \
    --timeout 600 --auto-fix
```

### 7 incomplete nets in the saved routed PCB

| Net | Pads unconnected | Pad locations | Failure class |
|---|---|---|---|
| `GND` | 32 / ~40 | spread across board (every cap, J1, U1) | **LAYER-COUNT** (single-sided pour cannot stitch; needs B.Cu pour + stitching vias) |
| `USB_CC1` | 1 (`J1.A5`) | (39.00, 8.00) inside USB-C connector | **PLACEMENT/GEOMETRY** (J1 not locked; pad escape via #2473) |
| `USB_CC2` | 1 (`J1.B5`) | (41.00, 9.00) inside USB-C connector | **PLACEMENT/GEOMETRY** (same as CC1) |
| `USB_D-` | 1 (`J1.B7`) | (39.75, 9.00) — coupled with `J1.A7` on D-  | **LAYER-COUNT + CONTROLLED-Z** (diff pair needs 4-layer for 90Ω impedance; only 1 of 2 N-side pads reached) |
| `VBUS` | 5 (J1.A4/A9/B4/B9 + C4.1) | clustered around J1 | **PLACEMENT** (J1 perimeter connector starved by interior-pulled centroid) |
| `VCC` | 8 (C1/C2/C3/J2.2/R12 + 3 more) | scattered MCU bypass caps | **PLACEMENT** (centroid pulls VCC strand off C-cluster ring) |
[...]

PR #2862, opened 2026-05-13
fix(fix-drc): per-nudge granular rollback preserves non-regressing fixes

### LINKED ISSUE #2851: fix-drc: per-nudge granular rollback (subset of nudges that don't drop connectivity)
## Parent

Follow-up to #2839 (sub-bug #3, deferred per Curator's "if too large for one PR" clause).
The visibility fixes from #2839 landed in PR #2853 (sub-bugs #1 and #2). This issue is the
structural improvement: when the post-pass connectivity check fires a rollback, revert only
the offending **subset** of nudges instead of all of them.

## Problem statement

Today, when `fix-drc` applies N nudges in a pass and the post-pass connectivity check
finds even one net dropped, `fix_drc_cmd.py:286-310` reverts **all N nudges** by
restoring a pre-pass file-bytes snapshot. This is all-or-nothing: if only K < N of the
nudges actually touched the offending nets, the other N - K nudges (which would have
been safe to keep) are also thrown away.

Concrete impact (from #2839 reproducer on board 03):
- 18 nudges computed, all displacement-feasible (max 0.1880 mm, well under 2.0 mm cap)
- Post-pass: connectivity dropped from 9 to 7 connected nets
- Current behavior: all 18 nudges reverted, headline says `Repaired 0/18`
- Desired behavior: identify the ~2 nudges that touched the 2 broken nets, revert
  just those, keep the other ~16

## Root cause

The rollback is implemented as a bulk file-bytes snapshot/restore:

```python
# src/kicad_tools/cli/fix_drc_cmd.py:263-310
if do_connectivity_check:
    snapshot = load_for_snapshot.read_bytes()
    baseline_conn = _count_connected_nets(load_for_snapshot)
# ...
if after_conn >= 0 and after_conn < baseline_conn:
    output_path.write_bytes(snapshot)   # bulk restore
    rolled_back = True
    repaired_this_pass = 0
```

`_count_connected_nets` returns only the **count** of fully-connected nets, so the
current call site has no way to identify *which* nets dropped. The underlying
`ConnectivityValidator.validate()` does surface per-net data (see "Net-aware
filtering" below), but that channel is not used.

The nudges themselves are computed and committed one at a time at
`repair_clearance.py:990, 999, 1077, 1083` — each `NudgeResult` already records
`uuid`, `net_name`, `layer`, `displacement_x`, `displacement_y`, so per-nudge
inversion is mechanically feasible.

## Implementation guidance

Three approaches, ranked from most to least precise. The architect should pick one
based on the cost-benefit analysis below.

### Approach 1 — Net-aware filtering (recommended)

Cheapest and most precise. The `ConnectivityValidator.validate()` already returns
a `ConnectivityResult` whose `.issues` list contains a `ConnectivityIssue` per
broken net with a `.net_name` field. Steps:

1. Extend `_count_connected_nets` (or add a sibling `_connectivity_report`) to
   return the full `ConnectivityResult` instead of just the count.
2. After the pass, compute the set of nets that **regressed**: nets that were
   fully connected in the baseline result but appear in the after-pass `issues`
   list (or are no longer in the connected set).
3. Filter `clearance_result.nudges + drill_result.actions` to those whose
   `net_name` is in the regressed set. These are the "offending" nudges.
4. Revert only the offending nudges (see "Per-nudge undo" below) and keep the
   rest applied. Update `repaired_this_pass = N - K` and tag the K reverted
   ones `(reverted)` in the listing.

Cost: one extra `ConnectivityValidator` pass to enumerate baseline per-net state
(or cache the baseline result from line 270). No extra routing.

Precision: high. Reverts exactly the nudges on the offending nets. May still
revert a "safe" nudge on an offending net if multiple nudges touch the same
net and only some are the actual cause, but that is a much smaller regression
than reverting all 18.

### Approach 2 — Bisection rollback (pragmatic fallback)

If net-aware filtering reveals subtle bugs (e.g., a nudge on net A indirectly
breaks net B via shared geometry), bisection is a safer "I don't trust the
net-attribution" alternative:

1. Apply all N nudges, observe connectivity dropped.
2. Bisect: revert the second half, re-check connectivity. If still dropped,
   the first half contains an offender; otherwise the second half did.
3. Recurse on the offending half until isolated.
4. Apply the non-offending nudges.

Cost: `O(log N)` connectivity checks (file write + `ConnectivityValidator` load).
For N=18: ~5 checks instead of 1. Still cheap because no routing is involved —
just S-expression edits and a graph traversal.

Precision: medium. Still rolls back some safe nudges in the bad half (worst case
half of the nudges in the offending partition).

### Approach 3 — Per-nudge connectivity check (most precise, most expensive)

Apply nudges one at a time, check connectivity after each, revert just that
one if it regressed, continue. Cost: `O(N)` `ConnectivityValidator` runs. For
N=18, ~18 checks per pass. Still no routing cost — `ConnectivityValidator` is
graph-only — but the file-write + parse cycle adds up.

**Performance guard (acceptance criterion):** the implementation **must not**
incur an `O(N)` routing cost (i.e., must not run `kct route` or
`LocalRerouter` after each nudge). Connectivity checks per nudge are acceptable
if Approach 3 is chosen; full re-routes are not.

### Recommendation

Start with Approach 1 (net-aware filtering). If it cannot precisely attribute
regressions in testing, fall back to Approach 2 (bisection). Avoid Approach 3
unless the others prove unreliable on real boards.

## Refactoring scope

The rollback site at `fix_drc_cmd.py:286-310` operates on **file bytes**. To
support granular rollback, the per-nudge inverse displacement must be
applied to the live PCB. Concretely:

- Add `ClearanceRepairer._undo_nudge(nudge: NudgeResult) -> None` (companion to
  `_apply_nudge` at `repair_clearance.py:1462`) that:
  - Looks up the object by `nudge.uuid`
  - Applies the inverse displacement `(-displacement_x, -displacement_y)`
  - For vias, also restores connected segment endpoints (mirror logic of
    `_apply_nudge` lines 1493-1499)
  - For segments at vias, mirrors the endpoint-only nudge inversion
- Add `DrillClearanceRepairer.undo_action(...)` for drill-clearance actions.
- The orchestrator in `fix_drc_cmd.py` calls these per-nudge undo functions
  for the offending subset, then re-saves the PCB.

The file-bytes snapshot can be kept as a **safety net**: if the per-nudge undo
loop fails (e.g., a UUID lookup fails because a later nudge moved the object),
fall back to the bulk restore. Log this as a warning.

## File / line references

- `src/kicad_tools/cli/fix_drc_cmd.py:263-310` — current bulk snapshot/rollback
- `src/kicad_tools/cli/fix_drc_cmd.py:450-465` — `_count_connected_nets` (extend
  to return per-net data, or add a sibling helper)
- `src/kicad_tools/drc/repair_clearance.py:37-58` — `NudgeResult` dataclass
  (already has `uuid`, `net_name`, `displacement_x`, `displacement_y` — no
  schema changes needed)
- `src/kicad_tools/drc/repair_clearance.py:990, 999, 1077, 1083` — nudge commit
  sites
- `src/kicad_tools/drc/repair_clearance.py:1462-1551` — `_apply_nudge` (the
  function whose inverse `_undo_nudge` must mirror)
- `src/kicad_tools/validate/connectivity.py:88-103, 207+` — `ConnectivityResult`
  and `ConnectivityValidator` (`.issues` list already carries `net_name` per
  broken net)
- `tests/test_fix_drc_cmd.py:1732+ TestConnectivityCheckRollback` — existing
  rollback test surface to extend

## Acceptance criteria

- [ ] When N nudges are computed and the post-pass connectivity check finds the
      drop is caused by a subset of K < N nudges, only those K are reverted.
      The remaining N - K nudges are persisted to the output file and the
      summary reports `Repaired N-K/total`.
- [ ] The K reverted nudges are tagged `(reverted)` in the per-category listing
      (already supported by the PR #2853 listing format), and the summary line's
      `(rolled back -- connectivity regression)` annotation is only present when
      **all** nudges were reverted (a true full rollback).
- [ ] **Performance guard:** the per-nudge or bisection rollback path must not
      invoke `kct route` or `LocalRerouter` per nudge. Acceptable: `O(N)`
      `ConnectivityValidator` runs (Approach 3), `O(log N)` (Approach 2), or
      `O(1)` extra (Approach 1). Unacceptable: any path that re-routes the
      board N times.
- [ ] On board 03 (the original reproducer from #2839), `kct route --auto-fix`
      repairs at least 1 of the 18 violations without breaking connectivity
      (the layer-3 acceptance contract from the parent).
- [ ] Existing rollback tests in `TestConnectivityCheckRollback` still pass.
      The full-rollback behavior (all nudges touch offending nets) remains
      identical to today.
- [ ] No regression on boards 01, 02, 04 (which currently pass `--auto-fix`
      cleanly or with no violations).
- [ ] The bulk file-bytes snapshot remains in place as a fallback for the case
      where per-nudge undo fails (e.g., UUID lookup miss). A warning is logged
      when the fallback path fires.

## Test plan

- [ ] **Step 1 — synthetic positive control.** Add a fixture board with 5
      clearance violations across 3 nets. Mock `ConnectivityValidator` so that
      exactly 2 of the 5 nudges (those on net X) cause a connectivity drop and
      the other 3 (on nets Y, Z) do not. Assert that after the pass, the
      output file reflects 3 applied nudges and the JSON reports
      `repaired = 3, reverted = 2`.
- [ ] **Step 2 — bisection / per-nudge coverage.** Whichever approach is chosen,
      assert the number of `ConnectivityValidator` calls per pass: Approach 1
      = 2 (baseline + after), Approach 2 = `O(log N) + 2`, Approach 3 =
      `N + 2`. Use a call-counting mock from
      `test_connectivity_decrease_triggers_rollback` as a starting point.
- [ ] **Step 3 — extend `test_connectivity_decrease_triggers_rollback`.** The
      current test mocks `_count_connected_nets` to drop from 10 -> 8 globally
      and expects a full rollback. Add a variant that mocks the underlying
      `ConnectivityValidator` to identify exactly which nets dropped, and
      assert the granular path reverts only the relevant nudges.
- [ ] **Step 4 — full-rollback regression guard.** Add a test where all N
      nudges touch the offending nets. Assert exit code 3, full rollback, and
      that the granular-rollback path correctly degenerates to "revert all".
- [ ] **Step 5 — performance guard.** Add an assertion that
      `LocalRerouter.reroute` (or the equivalent routing entry point) is
      **not** called by the rollback path. Use `monkeypatch` to install a
      sentinel that fails the test if called during rollback.
- [ ] **Step 6 — undo-failure fallback.** Construct a scenario where one nudge's
      UUID is no longer findable in the post-pass tree (e.g., simulate by
      patching the lookup to return `None` for one UUID). Assert that the
      rollback path falls back to the bulk snapshot restore and logs a
      warning.
- [ ] **Step 7 — board 03 regression guard.** Extend or create
      `tests/test_board_03_regression.py` to assert that
      `kct route --auto-fix` on the routed board 03 PCB repairs >0 violations
      with the granular rollback in place. This is the contract from the
      parent #2833 and the layer-3 acceptance criterion from #2839.

## Approach decision rubric

Architects: prefer Approach 1 (net-aware filtering) unless during prototyping
you find that the `ConnectivityValidator.issues` list does not reliably
attribute the regression to specific nets (e.g., a nudge on a connected net
indirectly breaks an unrelated net via shared geometry). If that happens,
escalate to Approach 2 (bisection). Document the choice in the PR description.

## References

- Parent: #2839 (visibility fixes landed in PR #2853; this is the structural follow-up)
- Sibling: #2852 (exit-code propagation from `--auto-fix`)
- Original board-03 issue: #2833 (board 03 routing 9/16 gap)
- Curator allowance: "If layer 3 is too large for one PR, fix #1 and #2 here and
  file #3 as a separate follow-up." (from #2839)

rjwalters:
## Summary

Closes #2851.

When the post-pass connectivity check fires in `fix-drc`, this PR reverts
only the subset of nudges that touched a **regressed net** instead of
bulk-restoring a pre-pass file-bytes snapshot.  Implements **Approach 1**
from the issue (net-aware filtering via `ConnectivityResult.issues[].net_name`):
O(1) extra `ConnectivityValidator` calls per pass and zero routing calls --
the rollback path edits the live S-exp tree in memory and rewrites the
file once.

The bulk-snapshot restore is preserved as a safety net: it fires on empty
offender sets, every-nudge-implicated cases, per-nudge undo failure
(e.g. UUID lookup miss), and any exception during the in-place edit.
Existing full-rollback behavior (exit code 3, complete file restore)
is unchanged when every applied nudge touches a regressed net.

### Concrete impact (board 03, the #2839 reproducer)

Before this PR:
```
Warning: pass 1 decreased connectivity (9 -> 8 nets); rolled back.
Repaired 0/4 DRC violations
```

After this PR:
```
Warning: pass 1 initially decreased connectivity (9 -> 8 nets); reverted
3 of 4 nudge(s) on regressed net(s) ['USB_D+']; now 9 connected net(s).
Repaired 1/4 DRC violations
  Clearance: 1/4
```

The 3 USB_D+ nudges that broke connectivity are reverted; the 1 USB_D-
nudge is preserved.  This satisfies the layer-3 acceptance contract
from #2839.

## Implementation

- `ClearanceRepairer._undo_nudge(nudge)` mirrors `_apply_nudge`: looks
  up the object by UUID, applies the inverse displacement, and for vias
  updates connected segment endpoints to match.
- `DrillClearanceRepairer.undo_action(action)` mirrors `_slide_via`;
  returns `False` for `deduplicate` actions (a removed via cannot be
  reconstructed from the in-memory record) to trigger bulk fallback.
- `DrillRepairAction` gains signed `displacement_x`/`displacement_y` +
  `uuid` fields so slides can be inverted (no schema removal --
  `displacement_mm` is preserved).
- `_attempt_granular_rollback` in `fix_drc_cmd.py` orchestrates the
  filter, undo, and fallback paths.
- Text/JSON output tags individually-reverted nudges `(reverted)` and
  surfaces a new `connectivity_check.passes[].partial_rollback` flag +
  `reverted_uuids` list so consumers can distinguish full vs partial
  rollbacks.  The `Repaired N/M` summary line now agrees with the
  on-disk state under partial rollback (one of the issue #2839 sub-bugs).

## Test plan

All 7 steps from the issue:

- [x] **Step 1** -- synthetic positive control:
  `TestGranularRollback::test_synthetic_positive_control_partial_rollback`
- [x] **Step 2** -- O(1) ConnectivityValidator budget:
  `test_connectivity_validator_calls_are_o1`
- [x] **Step 3** -- extends existing rollback test pattern (same
  `_count_connected_nets` mock + new `_connectivity_report` mock)
- [x] **Step 4** -- full-rollback regression guard:
  `test_full_rollback_when_all_nudges_touch_offending_net`
- [x] **Step 5** -- performance guard (no routing during rollback):
  `test_local_rerouter_not_called_during_rollback`
- [x] **Step 6** -- undo-failure fallback:
  `test_undo_failure_falls_back_to_bulk_snapshot`
- [x] **Step 7** -- board 03 regression guard:
  `test_fix_drc_preserves_safe_nudges_on_routed_board` in
  `tests/test_board_03_regression.py`

Plus `test_empty_regressed_set_falls_back_to_bulk` for the
attribution-failure fallback path.

### Acceptance criteria

- [x] Partial rollback reports `Repaired N-K/total` when K < N nudges
      are implicated; the K reverted entries are tagged `(reverted)` in
      the per-category listing
- [x] The `(rolled back -- connectivity regression)` annotation is only
      present on a true full rollback; partial rollback shows
      `(partial rollback -- K nudge(s) reverted on regressed nets)`
- [x] Performance guard: zero `LocalRerouter.reroute_segment` calls
      during rollback (asserted via sentinel monkeypatch)
- [x] On board 03 routed PCB, `fix-drc` repairs ≥1 violation without
      breaking connectivity (the #2839 layer-3 contract)
- [x] All 13 existing `TestConnectivityCheckRollback` tests still pass
- [x] No regression on boards 01 (no violations), 02 (full-rollback path
      preserved when 1 violation -> 1 broken net), 04 (full-rollback
      path preserved when many nudges -> many regressed nets)
- [x] Bulk file-bytes snapshot remains as fallback; a warning is logged
      when it fires after per-nudge undo failure

### Test counts

- `tests/test_fix_drc_cmd.py`: 111 tests pass (105 existing + 6 new)
- `tests/test_repair_clearance.py`: 67 tests pass (no changes)
- `tests/test_route_auto_fix.py`: 33 tests pass (no changes)
- `tests/test_board_03_regression.py::test_fix_drc_preserves_safe_nudges_on_routed_board`:
  passes (2.8 s, regenerate-free)

## References

- Issue #2851 -- this PR
- Parent #2839 -- visibility fixes landed in #2853; this is the
  structural follow-up
- Original board-03 issue #2833 -- routing 9/16 gap

### DIFF
--- a/src/kicad_tools/cli/fix_drc_cmd.py
+++ b/src/kicad_tools/cli/fix_drc_cmd.py
@@ -41,6 +41,16 @@ class PassResult:
     connectivity_before: int | None = None
     connectivity_after: int | None = None
     connectivity_rolled_back: bool = False
+    # Per-nudge granular rollback metadata (issue #2851):
+    #   * ``reverted_uuids`` holds the UUIDs of nudges/actions that were
+    #     individually reverted (subset of all applied work this pass).
+    #   * ``connectivity_partial_rollback`` is ``True`` when only a
+    #     proper subset of nudges was reverted; ``connectivity_rolled_back``
+    #     remains the flag for a *full* rollback so existing consumers
+    #     (text rendering, JSON ``rolled_back`` field, exit code 3) keep
+    #     their semantics.
+    reverted_uuids: tuple[str, ...] = ()
+    connectivity_partial_rollback: bool = False
 
     @property
     def violations_after(self) -> int:
@@ -260,14 +270,22 @@ Examples:
                     f"violation(s) (edge clearance, dimension, silkscreen, etc.)"
                 )
 
-        # Snapshot and baseline connectivity before the repair pass
+        # Snapshot and baseline connectivity before the repair pass.
+        # Issue #2851: we record the full ConnectivityResult (when
+        # available) alongside the count so the post-pass rollback can
+        # attribute regressions to specific nets and undo only the
+        # offending subset of nudges.  The simple count is what the
+        # rollback *decision* is based on (preserving the legacy
+        # _count_connected_nets mock surface used by existing tests).
         snapshot: bytes | None = None
         baseline_conn: int | None = None
+        baseline_report = None
         if do_connectivity_check:
             load_for_snapshot = output_path if pass_num > 1 else pcb_path
             if load_for_snapshot.exists():
                 snapshot = load_for_snapshot.read_bytes()
                 baseline_conn = _count_connected_nets(load_for_snapshot)
+                baseline_report = _connectivity_report(load_for_snapshot)
 
         # Run single-pass repairs
         clearance_result, drill_result = _run_single_pass(
@@ -288,6 +306,8 @@ Examples:
         # Post-pass connectivity check and rollback
         after_conn: int | None = None
         rolled_back = False
+        partial_rolled_back = False
+        reverted_uuids: tuple[str, ...] = ()
         if (
             snapshot is not None
             and baseline_conn is not None
@@ -298,16 +318,54 @@ Examples:
         ):
             after_conn = _count_connected_nets(output_path)
             if after_conn >= 0 and after_conn < baseline_conn:
-                # Connectivity decreased -- rollback
-                output_path.write_bytes(snapshot)
-                rolled_back = True
-                connectivity_rollback_occurred = True
-                repaired_this_pass = 0
-                print(
-                    f"Warning: pass {pass_num} decreased connectivity "
-                    f"({baseline_conn} -> {after_conn} nets); rolled back.",
-                    file=sys.stderr,
+                # Connectivity decreased -- try granular rollback first.
+                after_report = _connectivity_report(output_path)
+                regressed = _regressed_nets(baseline_report, after_report)
+
+                reverted_all, kept_count, reverted_uuid_list = _attempt_granular_rollback(
+                    output_path=output_path,
+                    clearance_result=clearance_result,
+                    drill_result=drill_result,
+                    regressed=regressed,
+                    snapshot=snapshot,
+                    pass_number=pass_num,
                 )
+
+                if reverted_all:
+                    # Full bulk rollback (legacy behavior).  All applied
+                    # nudges were thrown away.
+                    rolled_back = True
+                    connectivity_rollback_occurred = True
+                    repaired_this_pass = 0
+                    print(
+                        f"Warning: pass {pass_num} decreased connectivity "
+                        f"({baseline_conn} -> {after_conn} nets); rolled back.",
+                        file=sys.stderr,
+                    )
+                else:
+                    # Partial granular rollback: ``kept_count`` nudges
+                    # survived, ``len(reverted_uuid_list)`` were reverted.
+                    partial_rolled_back = True
+                    reverted_uuids = tuple(reverted_uuid_list)
+                    reverted_count = len(reverted_uuid_list)
+                    repaired_this_pass = kept_count
+                    pre_undo_after = after_conn
+                    # Re-measure connectivity so the JSON/text output
+                    # reports the post-undo state, not the (worse)
+                    # pre-undo state.  This is one extra
+                    # ConnectivityValidator call -- still O(1) in N.
+                    new_after = _count_connected_nets(output_path)
+                    if new_after >= 0:
+                        after_conn = new_after
+                    print(
+                        f"Warning: pass {pass_num} initially decreased "
+                        f"connectivity ({baseline_conn} -> {pre_undo_after} nets); "
+                        f"reverted {reverted_count} of "
+                        f"{reverted_count + kept_count} nudge(s) on regressed "
+                        f"net(s) {sorted(regressed)!r}; "
+                        f"now {after_conn} connected net(s).",
+                        file=sys.stderr,
+                    )
 
         pass_results.append(
             PassResult(
@@ -320,6 +378,8 @@ Examples:
                 connectivity_before=baseline_conn,
                 connectivity_after=after_conn,
                 connectivity_rolled_back=rolled_back,
+                reverted_uuids=reverted_uuids,
+                connectivity_partial_rollback=partial_rolled_back,
             )
         )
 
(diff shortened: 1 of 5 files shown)


rjwalters:
LGTM. Approved.

## Verification summary

**Implementation (Approach 1 -- net-aware filtering as Curator recommended):**
- `_connectivity_report` helper added in `fix_drc_cmd.py:528` -- returns the full `ConnectivityResult` alongside the legacy `_count_connected_nets`, preserving the existing mock surface for the 13 `TestConnectivityCheckRollback` tests.
- `_regressed_nets(baseline, after)` at `fix_drc_cmd.py:547` -- pure set difference on `ConnectivityResult.issues[].net_name`; returns empty set on `None` (safety-net trigger).
- `_attempt_granular_rollback` at `fix_drc_cmd.py:563` -- orchestrates filter, undo, fallback; never re-routes, edits S-exp tree in memory, saves once.
- Fallback to bulk snapshot fires on: empty regressed set, every-nudge implicated, per-nudge undo returning `False`, and any exception during the in-place edit. All paths covered.

**`_apply_nudge` -> `_undo_nudge` mirror in `repair_clearance.py:1581`:**
- `_find_node_by_uuid` helper at line 1552 added.
- Handles both `via` (with connected segment endpoints) and `segment` (with via-pinning logic, asymmetric endpoint, both-at-via no-op).
- Uses the *current* board state for via-position detection so the inverse displacement applies correctly when segments were updated alongside the via slide.

**`DrillRepairAction` + `undo_action` in `repair_drill_clearance.py`:**
- Signed `displacement_x`/`displacement_y` + `uuid` fields added without removing `displacement_mm` (no schema break).
- `undo_action` reverses slides correctly and matches connected segment endpoints by post-slide via position; deduplicate actions return `False` to trigger bulk fallback (a removed via cannot be reconstructed from the action record).

**`PassResult` extended** (`fix_drc_cmd.py:52`):
- `reverted_uuids: tuple[str, ...] = ()`
- `connectivity_partial_rollback: bool = False`
- `connectivity_rolled_back` remains the flag for *full* rollback so legacy semantics (exit code 3, JSON `rolled_back`) are preserved.

**All 3 printers updated:**
- Text (`_print_text`): per-listing `(reverted)` suffix, header `(K reverted)`, summary `(partial rollback -- K nudge(s) reverted on regressed nets)`.
- JSON (`_print_json`): `connectivity_check.passes[].partial_rollback` + `reverted_uuids` fields, per-nudge/per-action `reverted: bool`, effective per-category counts.
- Summary (`_print_summary`): `total_repaired` matches on-disk state.

**Tests:**
- `TestGranularRollback` (6 tests, lines 2344-2723): synthetic positive control, full rollback when all touch offending net, **`test_local_rerouter_not_called_during_rollback` performance guard with sentinel monkeypatch on `LocalRerouter.reroute_segment`**, undo-failure fallback, O(1) connectivity-call budget, empty-regressed-set fallback.
- `test_fix_drc_preserves_safe_nudges_on_routed_board` in `test_board_03_regression.py:531` (step 7).

**Test execution (all from worktree):**
- `tests/test_fix_drc_cmd.py`: **111 passed** in 44.85s
- `tests/test_repair_clearance.py` + `tests/test_route_auto_fix.py`: **123 passed** in 30.21s
- `tests/test_board_03_regression.py::test_fix_drc_preserves_safe_nudges_on_routed_board`: **1 passed** in 28.74s
- No regressions in the 13 existing `TestConnectivityCheckRollback` tests.

**Manual board sweep (`kct fix-drc` summary mode):**
- Board 01 (voltage-divider): 0 violations, exit 0 -- clean board.
- Board 02 (charlieplex): full bulk rollback (9 -> 8 nets, all implicated) -- legacy behavior preserved as required.
- Board 03 (usb-joystick, the #2839 reproducer): `Warning: pass 1 initially decreased connectivity (9 -> 8 nets); reverted 3 of 4 nudge(s) on regressed net(s) ['USB_D+']; now 9 connected net(s). Repaired 1/4 DRC violations`. **Went from 0/4 to 1/4 -- exactly the issue scenario.**
- Board 04 (stm32-devboard): full bulk rollback (11 -> 5 nets, many regressed) -- legacy behavior preserved.

**CI: all 8 checks pass.** mergeStateStatus=CLEAN, mergeable=MERGEABLE.

The implementation precisely matches the Curator's spec: Approach 1, O(1) extra `ConnectivityValidator` calls, no routing during rollback, granular subset filtering with a defensive bulk-snapshot fallback. All 7 acceptance criteria are met and surfaced as new tests. The text/JSON/summary printers are all internally consistent (no `Repaired N/M` mismatch with the per-category listing).

Issue #2604, opened 2026-05-10
router: PlacementFeedbackLoop never produces MOVE_COMPONENT strategies on chorus-test BLOCKED_PATH — fix or generate


rjwalters:
## Problem

PR #2602 wired \`kct route --placement-feedback\` (closing #2595 Track 1). When run on chorus-test-revA at HEAD \`40f5fbd8\`, the feedback loop engages, but the underlying PlacementFeedbackLoop **never generates a single MOVE_COMPONENT strategy** across 9 router iterations + 1 stagnation recovery + 1 perturbation event over ~30 minutes of placement-feedback wall time.

Evidence — \`grep -E 'Re-placing|Move:|MOVE_COMPONENT|Placement diff' /tmp/chorus-test-route/route_pf.log\`:

\`\`\`
(no output)
\`\`\`

The loop only adjusts the negotiated router's adaptive parameters (\`history\`, \`present_factor\`, perturbation). The 8-net structurally-blocked floor (DAC_CLK + 7 small nets in the U5/U7/U9 cluster) does not move.

Per curator notes on #2595, MOVE_COMPONENT strategies are produced by the FailureClassifier in \`src/kicad_tools/recovery/strategy.py:29\` for BLOCKED_PATH failures. On chorus-test the negotiated router emits 9 BLOCKED_PATH failures every iteration, so the strategy generator should fire — but doesn't, or its output is filtered before reaching the placement step.

## Repro

\`\`\`bash
# Synced PCB from prior runs at /tmp/chorus-test-route/synced.kicad_pcb
uv run kicad-tools route /tmp/chorus-test-route/synced.kicad_pcb \\
  -o /tmp/routed.kicad_pcb \\
  --strategy negotiated --layers 4 --backend cpp --manufacturer jlcpcb --no-cache \\
  --timeout 1800 --per-net-timeout 120 --placement-feedback
# Run for 30+ min and grep for "Re-placing|MOVE_COMPONENT|placement_diff" — none fire
\`\`\`

## Suspected locations

- \`src/kicad_tools/recovery/strategy.py:29\` — MOVE_COMPONENT generator (does it actually run on chorus-test BLOCKED_PATH?)
- \`src/kicad_tools/router/placement_feedback.py\` — \`PlacementFeedbackLoop._strategy_*\` filters; verify they don't drop all candidates due to fixed_refs/max_movement defaults being too strict
[...]

PR #2866, opened 2026-05-13
fix(router): narrow-channel guard prevents fine-pitch halo shrink when geometrically infeasible

### LINKED ISSUE #2865: router: 44 clearance_pad_segment errors persist on LQFP-48 west edge (board 04 fine-pitch cluster)
## Status: Confirmed persisting after #2857

Re-routed board 04 on main (post-PR #2857 fine-grid landing, commit e6b61329 era) with:
```bash
kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb \
  --output /tmp/board04_v2.kicad_pcb \
  --mfr jlcpcb-tier1 --auto-layers --auto-fix --timeout 600
kct check /tmp/board04_v2.kicad_pcb --mfr jlcpcb-tier1
```

Result: **44 `clearance_pad_segment` errors still present** (plus 4 `clearance_segment_via`, 4 `clearance_pad_via`, 4 `clearance_segment_segment`; 56 total errors during routing, 263 during the auto-fix DRC pass after pour fill). Worst clearance reduced from -0.391mm to -0.250mm, but error count is unchanged.

#2857 did engage the new fine zones (from route log):
```
Multi-Resolution Grid Plan (adaptive)
  Coarse grid: 0.100mm
  Fine zones: 5
    U1: 0.0500mm (10.3x8.6mm, ~35,604 cells)
    U2: 0.0125mm (12.3x12.3mm, ~972,196 cells)
    Y1: 0.0400mm (8.9x4.0mm, ~22,422 cells)
    D1: 0.0500mm (6.1x4.0mm, ~9,963 cells)
    J1: 0.0200mm (4.0x16.7mm, ~167,835 cells) offset=(0.0000,0.0100)
```
So fine zone for U2 IS produced. **The failure is not in zone planning — it's in the placement logic that accepts geometrically infeasible clearances.**

Violation cluster (world coords, from auto-fix log):
- `(133.2, 126.1)`, `(134.4, 126.1)`, `(130.5, 126.1)` — U2 west edge (NRST/pad 7 region)
- `(136.6, 120.0)`, `(134.1, 118.0)` — U2 south-west corner area
- Magnitudes: 5 at -0.250mm, ~22 at -0.100mm, ~8 at 0.000mm, 7 at 0.100mm

Auto-fix attempted 44 nudges but rolled back entirely: nudges break connectivity 11 → 4 nets (DRC fix-drc rolls back to preserve continuity).

## Root cause (geometric)

`_clearance_for_pin_pitch` (`src/kicad_tools/router/grid.py:767-797`) deliberately shrinks the pad halo to `min_trace_width / 2 = 0.0635mm` for fine-pitch pads. The docstring explicitly says: *"Full manufacturer clearance is validated in post-routing DRC."* So the grid permits placement; DRC catches it afterward. This is correct for *open* fine-pitch channels (e.g., chorus-test U5 BGAs where there is room) but incorrect when the channel is too crowded.

Geometric impossibility on LQFP-48 0.5mm pitch with jlcpcb-tier1 (`min_trace=0.127`, `min_clearance=0.127`):
- Pad-edge to pad-edge gap on 0.5mm pitch ≈ 0.5 - 0.25 (typical pad width) = 0.25mm
- Required for a single trace through: 0.127 (clr) + 0.127 (trace) + 0.127 (clr) = 0.381mm
- 0.25mm channel cannot accommodate 0.381mm — by 0.131mm, matching the -0.100 to -0.250mm violations observed

Threading signal traces between adjacent LQFP-48 pads is physically impossible at this manufacturer tier. The router needs to know this and route over/around or refuse.

## Why escape routing didn't save it

Log line: `Escape routes: U2 (QFP) - 8 pins escaped` — 8 of 9 routed signals got escape stubs to remapped endpoints just outside U2. But the main A* pass still chose paths that thread between *other* U2 pads en route to those endpoints, because the fine zone permits that (shrunk halo). The escape-endpoint remap doesn't block the rest of the channel.

### Option A (preferred): per-channel feasibility gate in fine zone planning
In `_compute_zone_resolution_and_offset` (or a sibling helper in `grid.py` near line 767), when emitting a fine zone, additionally compute the minimum feasible channel width for the manufacturer:
```
min_channel = 2 * min_clearance + min_trace_width
```
For each adjacent pad pair within the zone, if `(pitch - pad_width) < min_channel`, **mark the inter-pad cell band as blocked** (not merely fine-haloed). This converts the geometric impossibility into a routing constraint the A* respects.

For LQFP-48 0.5mm pitch with jlcpcb-tier1: `min_channel = 0.381mm`, available = 0.25mm → block. Force traces to escape on the open side (around the package) via the existing escape mechanism for *all* pins on that edge, not just the 8 currently chosen.

### Option B: over-density refuse-to-place
After fine zone resolution, count the number of nets that would need to terminate at pads inside the dense package vs the channel capacity. If `nets_into_zone > channel_capacity`, raise the existing `FineGridEscapeFailedError` (from `adaptive_grid.py:58`) early rather than producing a partial route. This surfaces the problem at planning time so the user gets a clear error instead of 44 DRC violations.

### Option C: trace-only halo guard (smallest patch)
Modify `_clearance_for_pin_pitch` to return `max(min_trace_width / 2, geometrically_required_clearance_for_channel)`. Only shrink the halo if the channel is wide enough that DRC will still pass — i.e., if `pitch - pad_width >= 2 * min_clearance + trace_width`. Otherwise keep the standard envelope. This is the most conservative change and preserves the open-channel optimisation for genuinely open fine-pitch components (chorus-test BGAs).

Recommend starting with **Option C** as it's localized and reversible; promote to Option A if Option C produces too-many partial routes on legitimate fine-pitch boards.

## Acceptance criteria

1. `kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb --mfr jlcpcb-tier1 --auto-layers --auto-fix --timeout 600` produces **0 `clearance_pad_segment` errors** on the routed `.kicad_pcb`.
2. `kct check` on the routed PCB reports `clearance_pad_segment: 0 errors`.
3. The pre-existing chorus-test fine-pitch routing (the original motivation for the shrunk halo per #2604) does not regress: tests under `tests/test_grid*.py` covering BGA halo behaviour still pass. Re-run `tests/test_adaptive_grid*.py` and `tests/test_chorus_test_routing*.py` (if present).
4. If a board is genuinely unroutable at the chosen manufacturer tier (Option B path), `kct route` exits with a clear error message naming the unroutable component and pitch, not a silent partial route plus DRC violations.
5. Board 04 manifest reports `MANUFACTURING_READY` after re-run; the per-board allowlist entry in `.github/routed-drc-tolerance.yml` can be reduced to 0 (per the exit clause documented in PR #2846).

## Test plan

1. Unit test: add a test fixture replicating LQFP-48-style pad geometry at 0.5mm pitch under jlcpcb-tier1 rules; assert that the routing grid blocks inter-pad cells (Option A) or returns the standard halo (Option C). Place test in `tests/router/test_grid_fine_pitch.py` (new file) or `tests/test_router_grid.py`.
2. Regression: re-run board 04 routing in CI; expect `clearance_pad_segment: 0`.
3. Regression: re-run chorus-test board routing (if present in repo) to confirm BGA optimization preserved.
4. Existing test: `tests/test_router_grid.py::test_clearance_for_pin_pitch_*` should be extended with a "narrow channel" case.

## File references

- `src/kicad_tools/router/grid.py:767-797` — `_clearance_for_pin_pitch` (primary patch site for Option C)
- `src/kicad_tools/router/grid.py:799-851` — `add_pad` / `_add_pad_unsafe` (Option A patch site for blocking inter-pad cells)
- `src/kicad_tools/router/grid.py:2490-2530` — `find_pad_ref_at` (mirror Option C clearance choice for symmetry, per #2604)
- `src/kicad_tools/router/adaptive_grid.py:58` — `FineGridEscapeFailedError` (raise for Option B)
- `src/kicad_tools/router/mfr_limits.py:73-81` — `MFR_JLCPCB_TIER1` (read `min_trace`, `min_clearance` for feasibility math)
- `src/kicad_tools/router/rules.py:90-99` — `fine_pitch_threshold`, `min_trace_width` (rule plumbing)
- `src/kicad_tools/router/core.py:8665-8696` — escape endpoint remap (consider extending to all dense-edge pads)

## Evidence artifacts

- `/tmp/board04_v2.kicad_pcb` — current routed output (44 errors)
- `/tmp/board04_v2.log` — full route log showing fine zones engaged
- `/tmp/check_v2.txt` — DRC report (`grep -c clearance_pad_segment /tmp/check_v2.txt` → 44 + header = 45 matches)

## Surfaced from

Board 04 manufacturing-readiness work (#2834, PR #2846); accounting for the residual 44/52 DRC errors after #2848+#2842 fixes landed. Verified persistent after #2857 (pad-position-aware fine grid zones) on 2026-05-12.

rjwalters:
## Summary

Implements the Curator-recommended Option C from issue #2865: extend `RoutingGrid._clearance_for_pin_pitch` with a pitch-vs-clearance geometric feasibility check before applying the `min_trace_width/2` shrink for fine-pitch pads.

When the inter-pad channel cannot accommodate a trace at full manufacturer clearance, the guard falls back to the standard envelope so the pathfinder routes **around** the package via the escape mechanism instead of threading a DRC-invalid path through the channel.

**Geometry (mirrors the Curator's formula):**
```
effective_channel = pitch - 2 * shrunk - trace_width
required_channel  = 2 * trace_clearance + trace_width
```

LQFP-48 0.5 mm pitch + jlcpcb-tier1 (`trace=clearance=0.127 mm`): `effective = 0.246 mm < required = 0.381 mm` → standard envelope (guard refuses shrink).

Wider SSOP/QFP pitches (0.65 mm) at relaxed clearance still permit the shrink — chorus-test BGA / SSOP escape routing is **regression-guarded** by dedicated tests.

## Implementation

- **`grid.py:_clearance_for_pin_pitch`** — primary guard site. Adds the channel-feasibility check before returning `min_trace_width/2`.
- **`grid.py:_apply_stitch_via_halo`** — dropped the pre-#2865 early-return `if base_clearance >= standard_envelope: return`. With the narrow-channel guard, fine-pitch plane-net pads now receive the standard envelope and still need the via halo applied along the narrow axis (board 04 U2.8/U2.23/U2.35 GND pins). The per-axis check below the removed early-return remains the correct gate and continues to short-circuit for standard-pitch wide pads (board 04 passive-array regression-guard).
- **`find_pad_ref_at`** symmetry — automatic. That method already routes through `_clearance_for_pin_pitch` per the #2604 symmetry fix; a dedicated test pins the contract.

## Scope note

The parallel C++ pad-vs-segment validator path (`DesignRules.get_clearance_for_component`) is **intentionally left untouched** per the Curator's localized Option C scoping. Documented in that method's docstring as a follow-up.

Because of that scoping, the integration AC ("`kct check ... --mfr jlcpcb-tier1` reports 0 `clearance_pad_segment` errors after re-route") is **not yet** satisfied on board 04: the C++ pathfinder's geometric validator still accepts the reduced `fine_pitch_clearance` (0.127 mm) when the negotiated outer loop forces a through-channel path under congestion pressure. Per the task description ("don't change the allowlist in this PR; leave that for a follow-up"), the `.github/routed-drc-tolerance.yml` entry for board 04 is intentionally **not** touched.

## Test plan

- [x] New `tests/test_grid_narrow_channel_guard.py` (11 tests): LQFP-48 refusal, wider-pitch acceptance, borderline equality, `find_pad_ref_at` symmetry, integration scenario (two adjacent LQFP pads on distinct refs leave no unblocked midpoint cell for foreign nets).
- [x] `tests/test_grid_plane_net_halo.py` — 12 tests pass. Two assertions updated to probe past the standard envelope edge (previous probe points relied on the now-disallowed shrink at infeasible LQFP-48 geometry).
- [x] `tests/test_stitch_via_halo.py` — 3 tests pass. `test_halo_blocks_more_cells_than_disabled_mode` switched to a looser-clearance fixture (`trace_clearance=0.1`, `pitch=0.65`) where the shrink remains feasible.
- [x] `tests/test_component_clearance.py` — 14 tests pass (parallel `get_clearance_for_component` path untouched).
- [x] `tests/test_router_grid.py` `tests/test_subgrid.py` `tests/test_escape.py` `tests/test_escape_via_in_pad_lqfp.py` `tests/test_escape_endpoint_routing.py` `tests/test_router_grid_auto_bump.py` — all pass.
- [x] `tests/test_router_core.py` — 172 pass, 1 pre-existing failure (`test_add_component_rotation`, unrelated; verified pre-existing on main).
- [x] Pre-route instrumentation confirmed `_clearance_for_pin_pitch(0.5)` now returns `0.2500` (standard envelope) instead of `0.0635` (shrunk) on board 04.

Closes #2865

Do not merge.

### DIFF
--- a/src/kicad_tools/router/grid.py
+++ b/src/kicad_tools/router/grid.py
@@ -776,25 +776,68 @@ class RoutingGrid:
         a passive's clearance halo to the BGA pad whose *real* envelope
         was the smaller fine-pitch one.
 
+        Issue #2865 -- narrow-channel guard: the fine-pitch shrink
+        (``min_trace_width / 2``) is only a sound optimisation when the
+        resulting *inter-pad* channel is wide enough that a trace centred
+        in it still satisfies full manufacturer clearance against both
+        flanking pads.  On crowded packages such as LQFP-48 0.5 mm pitch
+        with jlcpcb-tier1 rules (``min_trace=min_clearance=0.127 mm``),
+        threading a signal trace between adjacent pads is geometrically
+        infeasible: the required channel is ``2*clearance + trace_width =
+        0.381 mm`` but only ~0.25 mm of edge-to-edge gap exists.  In that
+        situation shrinking the halo to ``0.0635 mm`` only fools the
+        pathfinder into producing a route that DRC then rejects -- the
+        observed pathology behind 44 ``clearance_pad_segment`` errors on
+        board 04's STM32 west edge.  When the shrunk channel is too
+        narrow we therefore decline the optimisation and fall back to
+        the standard envelope, which causes the pathfinder to look for
+        escape routes around the package instead of through-channel.
+
         Args:
             pin_pitch: The component pin pitch in mm, or None when not
                 available.  When below ``rules.fine_pitch_threshold``
                 (and ``rules.min_trace_width`` is configured) the envelope
                 shrinks to ``min_trace_width / 2`` -- the minimum needed
-                to keep a necked-down trace from overlapping pad copper.
-                Full manufacturer clearance is validated in post-routing
-                DRC.
+                to keep a necked-down trace from overlapping pad copper --
+                *provided* the resulting channel can still satisfy full
+                clearance per the narrow-channel guard.  Full manufacturer
+                clearance is validated in post-routing DRC.
 
         Returns:
             Clearance distance in mm to pad outside the pad's metal.
         """
+        standard = self.rules.trace_clearance + self.rules.trace_width / 2
         if (
             pin_pitch is not None
             and pin_pitch < self.rules.fine_pitch_threshold
             and self.rules.min_trace_width is not None
         ):
-            return self.rules.min_trace_width / 2
-        return self.rules.trace_clearance + self.rules.trace_width / 2
+            shrunk = self.rules.min_trace_width / 2
+            # Issue #2865 narrow-channel guard.  Reject the shrink when
+            # the resulting inter-pad channel cannot host a trace at full
+            # clearance.  The geometry (pitch-based, mirroring the
+            # Curator's recommended formula) is:
+            #
+            #     effective_channel = pitch - 2 * shrunk - trace_width
+            #     required_channel  = 2 * trace_clearance + trace_width
+            #
+            # ``effective_channel`` is the band available for a trace
+            # centered between two halo edges; ``required_channel`` is
+            # the minimum copper-to-copper distance the manufacturer
+            # rule demands (clearance on each side of the trace).  When
+            # the channel cannot fit the trace + 2 x clearance the
+            # shrunk halo is geometrically infeasible -- the pathfinder
+            # would place a trace that DRC must reject.  Fall back to
+            # the standard envelope so the router routes *around* the
+            # package via the escape mechanism instead.
+            effective_channel = pin_pitch - 2.0 * shrunk - self.rules.trace_width
+            required_channel = 2.0 * self.rules.trace_clearance + self.rules.trace_width
+            if effective_channel >= required_channel:
+                return shrunk
+            # Narrow channel -- shrink is geometrically infeasible.
+            # Fall through to the standard envelope so the router does
+            # not try to thread between these pads.
+        return standard
 
     def add_pad(self, pad: Pad, pin_pitch: float | None = None) -> None:
         """Add a pad as an obstacle (except for its own net).
@@ -1038,23 +1081,21 @@ class RoutingGrid:
         # Equivalently, it extends ``halo_from_center - pad_half_extent``
         # *beyond the pad edge* along each axis.
         #
-        # Scope to the fine-pitch case (board 04 U2.8/U2.23/U2.35 LQFP-48
-        # corner GND pins).  On standard-pitch pads the existing
-        # envelope is already ``trace_clearance + trace_width/2``, which
-        # is at least as wide as the via halo *for the via that the
-        # stitcher will actually pick*.  Applying an extra ring on
-        # standard-pitch passives crowds inter-component routing channels
-        # (board 04 passive arrays go from 9/9 -> 3/9 with an unrestricted
-        # halo applied to every GND cap).
-        standard_envelope = (
-            self.rules.trace_clearance + self.rules.trace_width / 2.0
-        )
-        if base_clearance >= standard_envelope:
-            # Standard-pitch pad already has the full clearance envelope.
-            # The via center sits inside the pad metal -- already
-            # exclusively the plane net's territory.
-            return
-
+        # Issue #2865 follow-up: the pre-check that used to short-circuit
+        # "standard-pitch pads already have the full clearance envelope"
+        # by comparing ``base_clearance`` to ``standard_envelope`` is no
+        # longer correct.  After #2865's narrow-channel guard, *fine-pitch*
+        # pads on crowded packages (LQFP-48 0.5 mm pitch + jlcpcb-tier1)
+        # also receive ``base_clearance == standard_envelope``, and they
+        # still need the via halo applied along the narrow axis (board 04
+        # U2.8/U2.23/U2.35 GND pins).  The per-axis check below is the
+        # correct gate: it compares the via halo extension along each
+        # axis to ``base_clearance`` and returns when *both* axes are
+        # already covered.  Standard-pitch wide pads (e.g. 0805 caps,
+        # half-extents >= ``halo_from_center``) still short-circuit there
+        # because ``ext_x`` and ``ext_y`` clamp to zero -- the board 04
+        # passive-array yield concern (9/9 -> 3/9) the original early
+        # return was guarding remains protected.
         halo_from_center = self.rules.stitch_via_halo_radius()
         # Halo extension is measured separately per axis: an
         # LQFP-48-style pad is wide (1.5 mm) on one axis but narrow
@@ -1121,9 +1162,7 @@ class RoutingGrid:
                     oew, oeh = other.width, other.height
                 # Use the same pin_pitch envelope; without ``_pad_pin_pitch``
                 # data we fall back to the standard envelope.
-                other_clearance = self._clearance_for_pin_pitch(
-                    self._pad_pin_pitch.get(id(other))
-                )
+                other_clearance = self._clearance_for_pin_pitch(self._pad_pin_pitch.get(id(other)))
                 same_component_envelopes.append(
                     (
                         other.x - oew / 2.0 - other_clearance,
@@ -2175,8 +2214,7 @@ class RoutingGrid:
         """
         if not (0 <= layer_idx < self.num_layers):
             raise ValueError(
-                f"reserve_corridor_cells: layer_idx {layer_idx} out of range "
-                f"[0, {self.num_layers})"
+                f"reserve_corridor_cells: layer_idx {layer_idx} out of range [0, {self.num_layers})"
             )
         owners = frozenset(int(n) for n in net_ids)
         if not owners:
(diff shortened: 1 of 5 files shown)


rjwalters:
## Judge Review: APPROVED (scoped-correctly fix)

Approving with follow-up filed as #2867. The PR cleanly implements the Curator-recommended Option C from issue #2865.

### Verification summary

**1. Python guard logic (`grid.py:_clearance_for_pin_pitch` lines 808-840):**
- Channel feasibility check matches Curator's pseudocode exactly:
  - `effective_channel = pitch - 2*shrunk - trace_width`
  - `required_channel = 2*trace_clearance + trace_width`
- Uses `>=` so borderline (effective == required) still permits shrink; the dedicated test pins this against an accidental `>` slip.
- LQFP-48 at jlcpcb-tier1: returns 0.1905 (standard) instead of 0.0635 (shrunk). Verified.

**2. Companion `_apply_stitch_via_halo` change (lines 1080-1100):**
- The dropped early-return is necessary and correctly explained in the inline comment. Without it, fine-pitch GND pads (which now receive `base_clearance == standard_envelope` via the guard) would erroneously short-circuit and miss the via halo extension along the narrow axis (board 04 U2.8/U2.23/U2.35 GND pins).
- The per-axis check below the removed early-return remains the correct gate; standard-pitch wide pads (e.g. 0805 caps, half-extents >= halo_from_center) still short-circuit there because `ext_x` / `ext_y` clamp to zero — the board-04 passive-array yield concern (9/9 -> 3/9) the original early-return guarded remains protected.

**3. `rules.py` docstring (lines 318-333):**
- Explicitly documents that the narrow-channel guard is intentionally NOT propagated to the C++ validator path. Clear, accurate, and points to the eventual end-state (allowlist reduced to 0).

**4. New tests (`test_grid_narrow_channel_guard.py`, 11 tests):**
- LQFP-48 refusal at jlcpcb-tier1.
- Wider-pitch acceptance (chorus-test BGA / SSOP regression-guard).
- Above-threshold and `pin_pitch=None` smoke (pre-#2865 behavior preserved).
- `min_trace_width=None` (feature off) case.
- Borderline equality + just-below-borderline (geometry, not heuristic).
- Integration: LQFP-48 channel midpoint blocked for foreign nets.
- `find_pad_ref_at` symmetry (#2604 contract pin).

**5. Existing test adjustments:**
- `test_grid_plane_net_halo.py`: 2 probe points moved past the standard envelope edge with extensive docstring rationale — the previous probes relied on the now-disallowed shrink at infeasible LQFP-48 geometry.
- `test_stitch_via_halo.py::test_halo_blocks_more_cells_than_disabled_mode`: switched to looser geometry (`trace_clearance=0.1`, `pitch=0.65`) where the shrink remains feasible and the halo annular ring is observable. Justified inline.

**6. CI: all green.** mergeStateStatus=CLEAN, mergeable=MERGEABLE.

### Scope judgment

Issue #2865's title asserts "44 clearance_pad_segment errors persist on LQFP-48", and AC #1/#2 demand 0 such errors on board 04. The PR description honestly discloses that the board-04 integration AC is **not yet** satisfied because the parallel C++ pad-vs-segment validator path (`DesignRules.get_clearance_for_component`, called by `cpp_backend.py:591` to seed the C++ validator's `clearance_override`) uses a separate clearance-shrinking mechanism that bypasses `_clearance_for_pin_pitch`.

The Curator's recommendation in the issue body explicitly scoped Option C to `_clearance_for_pin_pitch` and noted: "Recommend starting with **Option C** as it's localized and reversible; promote to Option A if Option C produces too-many partial routes on legitimate fine-pitch boards." The Builder's analysis (revert of the C++ promotion attempt because it broke `test_component_clearance.py` and was out of scope) is sound.

Approving as a scoped-correctly fix. The unit-test AC and regression-guard against the geometric bug are achieved; the C++ validator promotion is filed as #2867 with the explicit board-04 integration AC and allowlist-reduction exit clause.

### Follow-up filed

- **#2867** — promote narrow-channel guard to `DesignRules.get_clearance_for_component` (C++ validator path). References this PR and inherits ACs #1/#2 from #2865 (zero `clearance_pad_segment` errors on board 04, allowlist reduced to 0).

LGTM!

PR #2882, opened 2026-05-14
feat(router): --auto-mfr-tier manufacturer-tier escalation

rjwalters:
## Summary

Implements `--auto-mfr-tier` escalation per #2881: when geometric infeasibility blocks routing on the current manufacturer tier (e.g., LQFP-48 0.5mm pitch on base jlcpcb without via-in-pad), the router walks a registered ladder of manufacturer tiers (`jlcpcb` -> `jlcpcb-tier1`) and reports which profile DOES work. This is the user-facing complement to in-pad escape rescue (#2880).

## What's new

- **New flags**: `--auto-mfr-tier` (default off) and `--mfr-tier-ladder` (optional explicit override). Opt-in by design because tighter tiers can incur surcharges.
- **`MFR_TIER_LADDERS` registry** in `router/mfr_limits.py` maps base manufacturer -> ordered ladder. Architecture accepts additional tiers (e.g., future `pcbway-tier1`) without further code changes.
- **`can_escalate_via_in_pad()` / `can_escalate_scalar()`** convergence guards. `jlcpcb` and `jlcpcb-tier1` have identical scalar limits -- escalation triggers on the capability flag (`via_in_pad_supported`), not scalar comparison (the Issue #2881 convergence case).
- **`MfrLimits.cost_note`** field; `MFR_JLCPCB_TIER1.cost_note = "Capability Plus surcharge ~\$30/order over base jlcpcb"`. Printed verbatim on successful escalation.
- **`EscapeRouter.missed_via_in_pad_rescues`** counter: bumped when the in-pad rescue path would have triggered but `via_in_pad_supported` is False. Drives the escalation loop's trigger detection.
- **`route_with_mfr_tier_escalation`** wraps `route_with_layer_escalation` (mirroring the analogous `--auto-layers` pattern at `route_cmd.py:1878`). Mutates `args.manufacturer` per tier and stops when one succeeds.
- **`failure_analysis.py` additions**: `MFR_TIER_ESCALATION_TRIGGERS` trigger table (PIN_ACCESS=YES, CLEARANCE=conditional, BLOCKED_PATH/CONGESTION=NO) plus `name_unfixable_constraint()` for the constraint-named diagnostic line.
- **`fab_rules_cmd.py show`** now lists the next tier and full tier ladder so users see "what would --auto-mfr-tier do for this manufacturer?".

## Acceptance criteria (from #2881)

- [x] `kct route --auto-mfr-tier` exists, validated by argparse, with `--mfr-tier-ladder` override.
[...]

PR #2892, opened 2026-05-14
feat(router): emit per-component diagnostic on tier-ladder exhaustion (#2884)

rjwalters:
## Summary

Wires `name_unfixable_constraint()` into the `--auto-mfr-tier` ladder-exhaustion code path so users see *which* component / pin / constraint blocked progress, not just the generic 4-option remediation menu.

- New helper `_name_dominant_unfixable_constraint()` in `route_cmd.py` extracts the dominant unfixable failure from the last inner attempt (via `missed_via_in_pad_rescues` + `missed_via_in_pad_components` on the EscapeRouter) and composes a one-line `name_unfixable_constraint(...)` diagnosis.
- The Diagnosis line prints *before* the existing 4-option Options list when the ladder fully exhausts.
- When no confident signal is available (zero missed-rescue counter, missing router, missing escape router), the Diagnosis section is suppressed and the generic Options list still prints.

## Diagnostic format

When the ladder exhausts on a fine-pitch QFP that needed via-in-pad rescue:

```
============================================================
MANUFACTURER-TIER ESCALATION SUMMARY
============================================================
Result: No tier in jlcpcb -> jlcpcb-tier1 achieved routing success.

Diagnosis:
  Cannot escape U2 to perimeter on jlcpcb. The `jlcpcb` profile does not support via-in-pad (via_in_pad_supported = false).

Options:
  1. Switch to a tighter manufacturer tier (try --mfr-tier-ladder with a custom ladder).
  2. Change fine-pitch package(s) to a wider-pitch alternative (e.g. LQFP-48 0.5mm -> LQFP-32 0.8mm).
  3. Move fine-pitch component(s) toward the board centre so escape traces have a wider channel (5+mm from edge).
  4. Add layers via --auto-layers --max-layers 6 (if not already on).
```

When more than one component shares the same fault, the line appends `(N other component(s) affected by the same constraint.)`.

## Source

Judge follow-up observation **B** on PR #2882, explicitly flagged as non-blocking.

## Scope

- `src/kicad_tools/cli/route_cmd.py`: 95 LOC added (helper + wiring)
- `tests/test_route_auto_mfr_tier.py`: 176 LOC added (5 new tests)
- No behavior change; user-visible diagnostic improvement only.

Closes #2884

## Test plan

- [x] New tests pass: `pytest tests/test_route_auto_mfr_tier.py::TestLadderExhaustionDiagnostic`
- [x] Existing 35 tests in `test_route_auto_mfr_tier.py` still pass
[...]

PR #2893, opened 2026-05-14
fix(router): wire MFR_TIER_ESCALATION_TRIGGERS into outer escalation loop (closes #2883)

### LINKED ISSUE #2883: router: wire MFR_TIER_ESCALATION_TRIGGERS into outer escalation loop
## Background

PR #2882 introduced `MFR_TIER_ESCALATION_TRIGGERS` and `should_escalate_mfr_tier()` in `src/kicad_tools/router/failure_analysis.py` to declare which failure modes should trigger manufacturer-tier escalation:

- `PIN_ACCESS` → YES
- `CLEARANCE` → conditional
- `BLOCKED_PATH` / `CONGESTION` → NO

Both functions are unit-tested in `tests/test_mfr_tier_ladder.py`, but the outer loop in `route_with_mfr_tier_escalation` (`src/kicad_tools/cli/route_cmd.py:3254-3300`) does **not** call them. It currently dispatches escalation purely on:

- `EscapeRouter.missed_via_in_pad_rescues > 0`, plus
- `can_escalate_via_in_pad()` / `can_escalate_scalar()` convergence guards.

That happens to be a sound proxy for `PIN_ACCESS`, but the documented broader contract (e.g., **suppress** escalation when the underlying failure is `BLOCKED_PATH`, even if a capability gain is theoretically available) is not enforced.

## Why this matters

- Mismatch between the trigger table's documented semantics and the dispatch path will mislead future readers/maintainers.
- When future failure categories are added (or `CLEARANCE` conditionality becomes important), the only path that gets updated is the trigger table, while the actual dispatcher silently ignores it.
- The unit tests give false confidence that the contract is wired up end-to-end.

## Proposed cleanup

1. In `route_with_mfr_tier_escalation`, after the inner routing attempt returns, classify the dominant failure category (probably from `RoutingFailure` records / partial-route diagnostics) and feed it through `should_escalate_mfr_tier(failure_category, current_mfr, next_mfr)`.
2. Combine that result with the existing `missed_via_in_pad_rescues > 0` heuristic (logical AND or document one as the override).
3. Add a smoke test that ensures `BLOCKED_PATH` does **not** trigger escalation even when capability gain is available, mirroring the existing `test_skips_tier_with_no_gain` shape.

## Source of recommendation

Judge follow-up observation **A** on PR #2882 (https://github.com/rjwalters/kicad-tools/pull/2882). Explicitly flagged as non-blocking for that PR.

## Scope

Small. Pure router-internal cleanup; no user-visible CLI surface change. Touches:

- `src/kicad_tools/cli/route_cmd.py` (~20-40 lines)
- `tests/test_route_auto_mfr_tier.py` (1-2 new tests)

Estimated < 200 lines diff.

rjwalters:
## Summary

Wires the `MFR_TIER_ESCALATION_TRIGGERS` registry (added in PR #2882) into the outer `--auto-mfr-tier` escalation loop. Previously the trigger table was unit-tested but never consulted by the dispatcher, leaving a documented-but-not-enforced contract.

- New private helper `_classify_dominant_failure_cause()` tallies per-net `FailureCause` values from `router.routing_failures`, picking the most common cause and breaking ties in favour of triggering causes so escalation engages on borderline mixed-failure cases.
- The outer loop in `route_with_mfr_tier_escalation` consults the trigger table on that dominant cause. When the registry marks the cause as non-fixable (`BLOCKED_PATH`, `CONGESTION`, `KEEPOUT`, `ROUTING_ORDER`, `UNKNOWN`, etc.), escalation is vetoed even when the next tier offers via-in-pad capability or scalar relaxation — exactly the case the issue called out.
- When `routing_failures` is unavailable (e.g. mocked inner call in unit tests), the dispatcher falls back to the legacy `missed_via_in_pad_rescues` + capability-gain heuristic, preserving every pre-existing test.

## Trigger registry contents

| `FailureCause`      | Escalate? | Why                                       |
| ------------------- | --------- | ----------------------------------------- |
| `PIN_ACCESS`        | YES       | Canonical "via-in-pad would have helped"  |
| `CLEARANCE`         | YES (conditional on scalar gain) | Tighter tier offers smaller mins |
| `VIA_BLOCKED`       | YES       | Tighter via geometry may fit              |
| `BLOCKED_PATH`      | NO        | Placement issue, not manufacturer-fixable |
| `CONGESTION`        | NO        | Layer issue; `--auto-layers` handles it   |
| `LAYER_CONFLICT`    | NO        | Layer issue                               |
| `KEEPOUT`           | NO        | Design issue                              |
| `ROUTING_ORDER`     | NO        | Ordering, not manufacturer                |
| `LENGTH_CONSTRAINT` | NO        | Design constraint                         |
| `DIFFERENTIAL_PAIR` | NO        | Design constraint                         |
| `UNKNOWN`           | NO        | Algorithm issue; escalation would mask it |

## Test plan

- [x] `pytest tests/test_route_auto_mfr_tier.py -v` — 36 passed (6 new tests under `TestTriggerTableWiring`)
- [x] `pytest tests/test_mfr_tier_ladder.py -v` — 27 passed (no regression on the existing ladder/trigger-table tests)
- [x] `ruff check` and `ruff format` clean on touched files
- [x] Smoke test `kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb --auto-mfr-tier` — succeeds on the starting tier as expected (escalation path stays inert when not needed)

### New tests added (`tests/test_route_auto_mfr_tier.py::TestTriggerTableWiring`)

- `test_pin_access_dominant_cause_triggers_escalation` — PIN_ACCESS dominant cause escalates (smoke that the wiring did not break the happy path)
- `test_blocked_path_dominant_cause_suppresses_escalation` — BLOCKED_PATH veto fires even when `missed_via_in_pad_rescues > 0` and capability gain is available (the case the issue called out)
- `test_congestion_dominant_cause_suppresses_escalation` — CONGESTION veto fires
- `test_no_failure_records_falls_back_to_legacy` — legacy heuristic preserved when classifier signal is absent
- `test_mixed_failures_tiebreak_picks_triggering_cause` — tiebreak between triggering and non-triggering causes picks the triggering one
- `test_classify_dominant_failure_cause_helper` — direct unit test of the new classifier helper

## Files touched

- `src/kicad_tools/cli/route_cmd.py` (+102, -16): added `_classify_dominant_failure_cause()`, imported `MFR_TIER_ESCALATION_TRIGGERS` / `should_escalate_mfr_tier`, added the veto branch in the outer loop
- `tests/test_route_auto_mfr_tier.py` (+301, -13): new `TestTriggerTableWiring` class

Closes #2883

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -1120,13 +1120,8 @@ def _run_auto_fix(
                 "  Auto-fix: rolled back due to connectivity regression "
                 "(nudges would have broken at least one net)."
             )
-            print(
-                "  Run 'kct fix-drc <pcb> --no-connectivity-check' to apply "
-                "the nudges anyway"
-            )
-            print(
-                "  (only safe when partial-completion regressions are acceptable)."
-            )
+            print("  Run 'kct fix-drc <pcb> --no-connectivity-check' to apply the nudges anyway")
+            print("  (only safe when partial-completion regressions are acceptable).")
         elif result == 2:
             print("  Auto-fix: partial repair; some violations remain.")
         elif result == 1:
@@ -3127,6 +3122,71 @@ def route_with_rule_relaxation(
     return 1
 
 
+def _classify_dominant_failure_cause(router):
+    """Pick the most-common FailureCause across ``router.routing_failures``.
+
+    Issue #2883: the outer ``--auto-mfr-tier`` loop needs to consult the
+    :data:`router.failure_analysis.MFR_TIER_ESCALATION_TRIGGERS` registry
+    before walking forward.  That requires picking a single dominant cause
+    from the (possibly heterogeneous) set of per-net failures returned by
+    the inner routing attempt.
+
+    The rule used here:
+
+    1. If the router has no ``routing_failures`` attribute (e.g. the inner
+       call was stubbed in tests or returned before classification), return
+       ``None`` so callers fall back to the legacy
+       ``missed_via_in_pad_rescues`` signal.
+    2. Otherwise, tally :class:`FailureCause` values across all failures
+       and return the most common.  Ties are broken by the registry order
+       in :data:`MFR_TIER_ESCALATION_TRIGGERS` (triggering causes win), so
+       that an even split between e.g. PIN_ACCESS and BLOCKED_PATH still
+       lets escalation engage.
+
+    Args:
+        router: Inner ``Autorouter`` instance (may be ``None`` or a mock
+            that lacks ``routing_failures``).
+
+    Returns:
+        The dominant :class:`FailureCause`, or ``None`` if no failure
+        records were available.
+    """
+    from collections import Counter
+
+    if router is None:
+        return None
+    failures = getattr(router, "routing_failures", None)
+    if not failures:
+        return None
+
+    # Tally causes; ignore entries without a recognized FailureCause.
+    causes = []
+    for f in failures:
+        cause = getattr(f, "failure_cause", None)
+        if cause is not None:
+            causes.append(cause)
+    if not causes:
+        return None
+
+    counter = Counter(causes)
+    most_common_count = counter.most_common(1)[0][1]
+
+    # Find all causes tied for most common.
+    tied = [c for c, n in counter.items() if n == most_common_count]
+    if len(tied) == 1:
+        return tied[0]
+
+    # Tie-break: prefer a triggering cause so escalation isn't suppressed
+    # by an arbitrary tally tie.  Lazy import to avoid module cycles.
+    from kicad_tools.router.failure_analysis import MFR_TIER_ESCALATION_TRIGGERS
+
+    for cause in tied:
+        if MFR_TIER_ESCALATION_TRIGGERS.get(cause, False):
+            return cause
+    # No triggering cause present in the tie; return any (the first).
+    return tied[0]
+
+
 def route_with_mfr_tier_escalation(
     pcb_path: Path,
     output_path: Path,
@@ -3183,6 +3243,10 @@ def route_with_mfr_tier_escalation(
         Exit code (0 = success, 1 = failure, 2 = partial)
     """
     from kicad_tools.cli.progress import flush_print
+    from kicad_tools.router.failure_analysis import (
+        MFR_TIER_ESCALATION_TRIGGERS,
+        should_escalate_mfr_tier,
+    )
     from kicad_tools.router.mfr_limits import (
         can_escalate_scalar,
         can_escalate_via_in_pad,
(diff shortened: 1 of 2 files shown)


rjwalters:

**Verdict: APPROVE.** The trigger-table veto is correctly wired into the outer escalation loop, CI is green (8/8 checks), merge state is CLEAN, and the registry at `failure_analysis.py:1755-1767` is consulted directly (not duplicated). Issue #2883 acceptance criteria are met.

### Acceptance criteria (issue #2883)

1. **Classify dominant failure category and feed through `should_escalate_mfr_tier`** — done via `_classify_dominant_failure_cause()` + the `trigger_table_vetoes` branch in `route_with_mfr_tier_escalation` (route_cmd.py:3338-3357).
2. **Combine with `missed_via_in_pad_rescues`** — trigger-table veto is layered ABOVE the legacy heuristic and the convergence guard. Order is documented in the inline comment ("escalation cannot fix a placement problem"). Veto path emits a distinct log reason so users can tell the two skip causes apart.
3. **Smoke test that BLOCKED_PATH does NOT trigger escalation even with capability gain** — `test_blocked_path_dominant_cause_suppresses_escalation` does exactly this (explicitly sets `missed_via_in_pad_rescues=5` to confirm the trigger table overrides the legacy heuristic).

### Top 3 findings

**1. Source-of-truth wiring is correct, single registry.**
Both the classifier's tie-break and the dispatch path import `MFR_TIER_ESCALATION_TRIGGERS` from `failure_analysis.py:1755`. No duplicate map. Defensive `dominant_cause in MFR_TIER_ESCALATION_TRIGGERS` guard before the veto handles future enum values not yet registered.

**2. Test coverage is good but slightly thin on the "each trigger / each non-trigger" axis.**
The 6 new tests in `TestTriggerTableWiring` cover PIN_ACCESS (positive), BLOCKED_PATH and CONGESTION (veto), the tie-break, the legacy-fallback, and direct classifier asserts. The judge brief asked for "each trigger causes escalation" and "each non-trigger vetos escalation" — only 1 of 3 triggering causes (PIN_ACCESS; not CLEARANCE or VIA_BLOCKED) and 2 of 8 non-triggering causes are explicitly covered. The trigger-table itself is unit-tested in `test_mfr_tier_ladder.py` so the wiring contract is sufficient, but adding parametrized coverage would harden against future registry changes. **Not a blocker.**

**3. No test asserts that capability-gain is still required when the dominant cause is triggering.**
The convergence guard logic is unchanged, so the legacy `test_skips_tier_with_no_gain` continues to pass — but a fresh test that combines `dominant_cause=PIN_ACCESS` with `gains_capability=False, gains_scalar=False` would lock in the layered-veto contract end-to-end. Suggest adding in a follow-up; not blocking this PR.

### Conflict assessment vs PR #2892

**Moderate textual-conflict likelihood, low semantic conflict.** Both PRs touch `route_with_mfr_tier_escalation` (different regions: #2893 modifies the convergence-guard block ~lines 3249-3305; #2892 modifies the exhaustion-summary block ~lines 3384-3398) and both add a private helper near the same boundary in `route_cmd.py`. Both add a new test class to `tests/test_route_auto_mfr_tier.py` (`TestTriggerTableWiring` vs `TestLadderExhaustionDiagnostic`).

The likely friction points are:
- PR #2893 collapses ~5 multi-line `flush_print(...)` calls into single-line forms inside the same function PR #2892 edits. Reformat churn vs additive content will produce textual conflicts during rebase.
- Helper placement near each other in the file.

**Recommendation: merge PR #2893 first, then rebase PR #2892 on top.** PR #2892 is purely additive (new helper at end, new diagnostic block inside exhaustion summary) and will rebase cleanly. PR #2893 carries the formatter churn, so landing it first minimizes post-merge fixup on #2892.

### Tie-breaking design choice

The brief flagged the rule "ties broken in favour of triggering causes" as potentially masking placement failures. **The call is defensible**, with one observability gap worth noting:

- **False-positive cost (tie engages, but placement remains)**: User pays extra routing-time CPU. The escalated tier will fail because BLOCKED_PATH/CONGESTION/etc. is not manufacturer-fixable, so the loop exhausts the ladder and the exhaustion summary (augmented by PR #2892) prints. **No silent BOM markup** — the cost note only fires when a tier achieves success (route_cmd.py:3361).
- **False-negative cost (tie suppresses, PIN_ACCESS remains)**: A genuine via-in-pad failure that coexists with placement issues never gets the chance to escalate, and the user is told "escalation suppressed". This is a worse user experience — the user opted into escalation precisely to recover from edge cases.

Given the asymmetry, favoring escalation on ties is the right default. **Minor improvement to consider**: log when the tie-break engages, e.g. `"  Escalating to {tier_name}: {reason} (tie-break: PIN_ACCESS, BLOCKED_PATH tied; preferring triggering cause)"`. Right now `_classify_dominant_failure_cause` silently returns the triggering cause with no signal that a tie was resolved. Useful for users investigating "why did this escalate when half my failures were placement?" — file as a follow-up, not blocking this PR.

### Code quality notes

- `_classify_dominant_failure_cause()` is module-private (leading underscore), has a thorough docstring describing the rule, handles all four degenerate cases (None router, missing attribute, empty list, no recognized causes), and uses `Counter.most_common()` correctly. Lazy import of `MFR_TIER_ESCALATION_TRIGGERS` inside the tie-break branch avoids module cycles.
- Veto branch uses `dominant_cause.value` (the enum string) in the log message — correct for human-readable output.
- No new abstractions for hypothetical needs.
- Error handling at boundaries (`getattr` with defaults, `try/except` on missed-rescue int conversion already existed) is sound.

### CI / merge state
- All 8 CI checks pass (Lint, Type, Test, C++ Build, kicad-cli Smoke, DRC, Diff-Pair, Match-Group).
- Branch is up to date with main.

Approved for merge by Champion.

— Loom Judge

PR #2889, opened 2026-05-14
fix(router): force via-in-pad escape for fine-pitch QFP signal pins between same-component plane pads (closes board-04 chain)

rjwalters:
## Summary

Adds a strict plane-sandwich predicate to ``_escape_qfp_alternating`` that forces ``_try_in_pad_escape`` when a fine-pitch QFP signal pin sits between two same-component plane-net pads on its immediate same-edge neighbours AND the dispatcher would emit an along-edge escape.  Emits a clear error message naming the unfixable geometric constraint when via-in-pad is unavailable, instead of silently producing routes that DRC later rejects.

Partially addresses #2880.  References #2834.

## What this PR delivers vs. doesn't

**Shipped:**
- Strict ``_is_pin_boxed_by_plane_neighbours`` predicate with corner-canonical edge classification
- Rescue gate that forces ``_try_in_pad_escape`` on along-edge sandwich pins (perpendicular escapes are left alone)
- ERROR-level diagnostic when the predicate fires but the manufacturer lacks via-in-pad support
- 10 new unit tests in ``tests/router/test_escape_qfp_plane_sandwich.py``
- Merge-conflict resolution with PR #2882 (``--auto-mfr-tier``) — counter and error log are complementary, verified by 35/35 ``test_escape_missed_via_in_pad.py`` + ``test_route_auto_mfr_tier.py`` passing
- **The predicate DOES fire on board-04** — Judge verified (post-instrumentation) that ``U2.pin39 SWO`` is plane-sandwiched between pads 38 and 40 (both on plane net 0), and the in-pad rescue engages with a via at SWO's pad center under ``jlcpcb-tier1``. This corrects the earlier description's claim that no signal pin on the current fleet satisfies the predicate — SWO does.

**Not shipped (deliberately):**
- Board-04's full ``clearance_pad_segment`` cluster is NOT cleared.  The dominant offenders (OSC_OUT, NRST, OSC_IN) are NOT strictly plane-sandwiched per this predicate, and broadening the predicate empirically regressed completion (forces in-pad on pins whose perpendicular escape is clean → B.Cu congestion).
- ``.github/routed-drc-tolerance.yml`` board-04 entry remains at 60.
- The architectural fix for board-04 closure (broader predicate variant or alternative routing approach) is tracked under #2834.

## What changed

[...]

Issue #2880, opened 2026-05-14
router: board 04 needs same-component own-net escape relaxation (NRST clipping U2 plane-net pads on LQFP-48)


rjwalters:
## Observation (re-verified 2026-05-12)

Board 04 (STM32 LQFP-48 dev board) plateaus at **44 `clearance_pad_segment` errors** under jlcpcb-tier1 (`trace = clearance = 0.127 mm`) after the full validator chain (#2842 + #2848 + #2865 + #2867 + #2869 + #2871 + #2874) and the proactive narrow-channel halo (#2878 / #2879).

### Curator verification

Re-routed `boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb` on current `main` with `kct route --mfr jlcpcb-tier1 --auto-layers --auto-fix --timeout 600`. DRC JSON (`/tmp/board04_drc.json`) confirms:

- **44 `clearance_pad_segment` errors** (plus 4 `clearance_segment_via`, 4 `clearance_pad_via`, 4 `clearance_segment_segment`)
- All 44 errors involve U2 pads on the **west edge** (pins 1, 8, 9, 23, 24, 36)
- Net pairs (frequency):
  - `(GND, NRST)` x 18
  - `(+3.3V, NRST)` x 13
  - `(+3.3V, OSC_OUT)` x 9
  - `(+3.3V, SWCLK)` x 4
- Trace segments on **F.Cu** at x ≈ 126.8–127.7, y ≈ 119–119.8 (north of pin 1) and y ≈ 122–123 (south of pin 7) — the escape paths from interior signal pins (NRST/OSC_OUT/SWCLK) climbing along the U2 west edge to reach the package perimeter, clipping the +3.3V/GND plane-net pads on the way out.

Example violating segment `0ddc64b5`: `(127.7465, 119.7903) -> (127.5, 119.6)` on F.Cu, net 5 (OSC_OUT, U2.6). It clips pad U2.1 (+3.3V at 126.8375, 119.25). Trace edge is **0.1 mm** north of pad metal top edge (119.4) — short by `0.027 mm` of the `0.127 mm` clearance.

### LQFP-48 west edge pin assignment (verified)

| Pin | Net    | Role               |
|-----|--------|--------------------|
| 1   | +3.3V  | VDD (plane)        |
| 2-4 | (n.c.) | (unconnected)      |
| 5   | OSC_IN | signal             |
| 6   | OSC_OUT| signal             |
[...]

PR #2895, opened 2026-05-14
fix(cli): restore outer kct build parser parity (closes #2888)

rjwalters:
## Summary

Closes #2888.

The top-level `kct build` parser had drifted out of sync with the authoritative inner parser at `kicad_tools.cli.build_cmd._build_inner_parser`. This PR restores parity (Phase 1 of the curator's plan) plus adds a parity test that fails loudly the next time the two drift.

### Outer parser additions (`src/kicad_tools/cli/parser.py::_add_build_parser`)

| Kind | Items |
|---|---|
| New `--step` choices | `erc`, `sync`, `preflight-routing` |
| New flags | `-q/--quiet`, `-o/--output`, `--optimize-placement`, `--no-smoke-check`, `--allow-incomplete` |

### Dispatcher additions (`src/kicad_tools/cli/commands/build.py::run_build_command`)

Forwards into `sub_argv`: `--output`, `--optimize-placement`, `--no-smoke-check`, `--allow-incomplete`, and the command-level `--quiet`/`-q` (in addition to the existing global `--quiet`).

### Refactor

`build_cmd.main()` now defers parser construction to a new `_build_inner_parser()` helper so the parity test can introspect it directly.

### New test (`tests/test_build_parser_parity.py`)

13 tests, all green:

* `test_inner_parser_options_are_mirrored_on_outer_parser` — every inner long option (modulo an explicit empty allowlist) is also declared on the outer parser.
* `test_outer_parser_does_not_advertise_unknown_options` — reverse direction, prevents outer-only flags from creeping in.
* `test_step_choice_parity` — `--step` choice sets match exactly.
* `test_outer_parser_accepts_each_step_choice[...]` — parametrized smoke test (8 choices).
* `test_outer_parser_accepts_new_flags` — outer parser accepts each newly added flag.
* `test_dispatcher_forwards_new_flags` — `run_build_command` actually forwards them into `build_main`'s argv.

### Docs restored

`docs/reference/cli.md`, `docs/guides/manufacturing-export.md`, and `README.md` drop the "outer parser is stale" workarounds from PR #2887 and go back to the unified `kct build ...` invocations (no more `python -m kicad_tools.cli.build_cmd ...` workaround needed).

## Test plan

- [x] `uv run pytest tests/test_build_parser_parity.py` — 13 passed
[...]

PR #2897, opened 2026-05-14
router: variant #3 direction-aware plane-sandwich predicate (#2890)

### LINKED ISSUE #2890: router: predicate variant #3 — one plane neighbour on escape-direction side (in-pad rescue follow-up to #2880)
## Context

PR #2889 introduced `EscapeRouter._is_pin_boxed_by_plane_neighbours` — a strict predicate that fires only when a fine-pitch QFP signal pin's two immediate same-edge neighbours are BOTH plane-net pads.  Judge verified the predicate engages on `U2.pin39 SWO` on board-04 (pads 38 and 40 are both plane net 0), but does NOT engage on board-04's dominant `clearance_pad_segment` cluster (`OSC_OUT`, `NRST`, `OSC_IN`) because each of those pins has at least one signal neighbour on the same edge.

Empirically broadening to "any plane neighbour on the same edge" was tried in PR #2889 and regressed completion to 7/9 partial routes — it forces via-in-pad on pins whose perpendicular escape is clean, causing B.Cu congestion.

## Proposed predicate variant #3

A geometrically motivated middle ground: **fire when exactly ONE same-edge neighbour is a plane pad AND that neighbour sits on the same side as the dispatcher's escape direction.**

Mechanism: the alternating-dispatcher chooses an along-edge direction (e.g. "east" on a south-edge pin).  If the plane neighbour is the east neighbour, the trace must cross it — forcing the violation.  If the plane neighbour is the west neighbour, the east-bound escape exits cleanly.

This requires threading `direction` into the predicate signature:

```python
def _is_pin_boxed_by_plane_neighbours_on_escape_side(
    self, pad, package, direction, *, plane_nets=None
) -> bool:
```

## Acceptance criteria

- New predicate fires on `U2.pin6 OSC_OUT` (board-04 dominant offender) when the dispatcher's chosen along-edge direction would cross its plane neighbour
- Does NOT fire on pins whose plane neighbour is on the non-escape side
- Wired into the rescue gate alongside the strict predicate from #2880
- Board-04 `clearance_pad_segment` count drops measurably under `jlcpcb-tier1`
- Boards 03/05/06/07 — no regression

## References

- #2880 (strict-predicate predecessor, partially shipped in PR #2889)
- #2834 (board-04 manufacturing-readiness umbrella — this work belongs there)
- escape.py:1840-1900 (rescue gate site)
- `EscapeRouter._classify_edge` (corner-canonical classifier already in place — reuse)

rjwalters:
## Summary

Variant #3 of the plane-sandwich predicate (Issue #2890) -- a
direction-aware sibling to the strict ``_is_pin_boxed_by_plane_neighbours``
predicate from PR #2889 (Issue #2880).

- Add ``EscapeRouter._is_pin_blocked_in_escape_direction(pad, package, direction)``
  next to the strict predicate in ``router/escape.py``.  Returns True only
  when the same-edge neighbour ON the chosen escape-direction side is a
  plane pad.
- OR it into the rescue gate at ``_escape_qfp_alternating`` (escape.py:1864-1884)
  behind the existing ``escape_is_along_edge`` guard so perpendicular
  escapes are still never forced to in-pad rescue.
- Update the rescue-success ``logger.info`` to distinguish strict (#2880)
  triggers from directional (#2890) triggers so post-hoc analysis can tell
  which predicate fired.

## Why a New Predicate

The strict predicate (#2880) requires BOTH same-edge neighbours to be plane
pads.  This misses signal pins with only one plane neighbour whose
dispatcher-chosen along-edge direction happens to point toward that plane
side -- those pins also need the in-pad rescue, because the escape segment
threads through the same 0.2 mm plane-side channel.

PR #2889 already tried a naive "any plane neighbour on the same edge"
broadening and **regressed board-04 from 9/9 to 7/9** -- it forced
via-in-pad on pins whose chosen direction was toward the signal side,
incurring needless B.Cu congestion.  Variant #3 avoids that failure mode
by checking ONLY the neighbour on the chosen direction's side; pins whose
dispatcher chose the clean direction keep their surface escape.

## Board-04 Behaviour

On the current STM32 LQFP-48 west-edge layout the dispatcher's chosen
directions for the OSC_IN / OSC_OUT / NRST cluster don't intersect the
plane sides:

- Pin 5 OSC_IN (i=0): WEST perpendicular -- gate not triggered.
- Pin 6 OSC_OUT (i=1): NORTH alt-cw -- idx+1 neighbour is pin 7 NRST
  (signal), so variant #3 returns False.
- Pin 7 NRST (i=2): WEST perpendicular -- gate not triggered.

The predicate IS correct on plane-side directions when called manually:
pin 5 SOUTH returns True (pin 4 plane), pin 7 NORTH returns True (pin 8
GND plane).  The dispatcher just doesn't pick those directions for the
current pinout.  Variant #3 ships as a geometrically-motivated middle
ground that will engage on future boards whose pin assignments cluster
plane neighbours on the dispatcher-chosen side.

**Per-board routing impact** (jlcpcb-tier1, max-layers 2):

| Board | Baseline (PR #2889) | After #2890 | Notes |
|-------|--------------------|--------------|-------|
| 04-stm32-devboard | 9/9, 56 err | 9/9, 56 err | identical -- dispatcher directions miss the plane sides |
| 01/02/03/05/06/07 | unchanged | unchanged | no fine-pitch QFP -> variant #3 inactive |

## Test Plan

- [x] All 10 existing tests in ``tests/router/test_escape_qfp_plane_sandwich.py`` pass.
- [x] All 9 existing tests in ``tests/test_escape_via_in_pad_lqfp.py`` pass.
- [x] 10 new ``TestPinBlockedInEscapeDirection`` cases cover:
  - neighbour-side direction mapping (NORTH/SOUTH/EAST/WEST on each edge)
  - perpendicular-direction short circuit
  - corner-pin idx out-of-range case
  - plane pad input (never rescued)
  - signal-only edge (never fires)
  - strict-sandwich overlap (both predicates True)
  - north-edge direction mapping sanity check
- [x] 1 new ``TestRescueGateBothPredicates`` integration test confirms
      the rescue gate is reachable via the new predicate.
- [x] Full router suite: ``pytest tests/router/ tests/test_escape*.py -m "not slow"`` -> 235 passed, 1 skipped (unchanged from baseline).
- [x] ``ruff check`` clean on touched files.
- [x] Live board-04 route: ``kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb --manufacturer jlcpcb-tier1 --max-layers 2`` -> 9/9, 56 errors (identical to PR #2889 baseline).

Closes #2890

### DIFF
--- a/src/kicad_tools/router/escape.py
+++ b/src/kicad_tools/router/escape.py
@@ -1862,12 +1862,26 @@ class EscapeRouter:
                 # (0.2 mm available, 0.381 mm required) -- the only
                 # viable along-edge escape is vertical via-in-pad.
                 escape_is_along_edge = direction != primary_dir
-                pin_boxed = (
+                # Issue #2890: OR in the direction-aware variant #3
+                # predicate.  The strict predicate (#2880) requires
+                # BOTH same-edge neighbours to be plane pads; variant
+                # #3 widens the trigger to "the neighbour on the
+                # escape-direction side is a plane pad".  Both are
+                # gated by ``escape_is_along_edge`` -- perpendicular
+                # escapes are never forced to the in-pad rescue.
+                pin_boxed_strict = (
                     escape_is_along_edge
                     and self._is_pin_boxed_by_plane_neighbours(
                         pad, package,
                     )
                 )
+                pin_blocked_directional = (
+                    escape_is_along_edge
+                    and self._is_pin_blocked_in_escape_direction(
+                        pad, package, direction,
+                    )
+                )
+                pin_boxed = pin_boxed_strict or pin_blocked_directional
                 if try_in_pad_fallback and unclipped_escape.segments:
                     surface_seg = unclipped_escape.segments[0]
                     violation = self._segment_violates_pad_clearance(
@@ -1886,15 +1900,31 @@ class EscapeRouter:
                         )
                         if in_pad_route is not None:
                             if pin_boxed and not violation:
+                                # Distinguish which predicate fired so
+                                # post-hoc analysis can tell strict
+                                # #2880 triggers from directional #2890
+                                # triggers.
+                                if pin_boxed_strict:
+                                    trigger_label = (
+                                        "boxed between same-component "
+                                        "plane-net neighbours on "
+                                        f"{package.package_type.name} edge "
+                                        "(Issue #2880)"
+                                    )
+                                else:
+                                    trigger_label = (
+                                        "plane neighbour on the chosen "
+                                        "escape-direction side of "
+                                        f"{package.package_type.name} edge "
+                                        "(Issue #2890)"
+                                    )
                                 logger.info(
                                     "In-pad rescue forced for %s pin %s "
-                                    "(net %s): boxed between same-component "
-                                    "plane-net neighbours on %s edge "
-                                    "(Issue #2880).",
+                                    "(net %s): %s.",
                                     package.ref,
                                     pad.pin,
                                     pad.net_name,
-                                    package.package_type.name,
+                                    trigger_label,
                                 )
                             escapes.append(in_pad_route)
                             continue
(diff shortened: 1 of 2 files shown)


rjwalters:
## Doctor analysis: Approach A as specified cannot close the board-04 gap

Rebased onto `main` (picked up #2895/#2896) and re-ran the full router test suite -- 238 passed, no regressions from the rebase. New commit: `f5ee5bde`.

I then deeply analysed whether Approach A (bias direction choice toward plane-side avoidance) could close the practical gap on board-04. **The answer is no**, and shipping it would be the partial fix the brief warned against. Reporting back rather than shipping.

### What the predicate actually says about U2 LQFP-48

Reproduced the dispatcher's per-pin decisions against the real U2 west-edge geometry (12 pads, signal pins are 5/OSC_IN, 6/OSC_OUT, 7/NRST; everything else is `+3.3V`, `GND`, or unconnected and so gets rewritten to `net=0` by `io.py:2819`):

| pin | net      | chosen direction | variant-#3 on chosen | variant-#3 on opposite |
|-----|----------|------------------|----------------------|------------------------|
| 5   | OSC_IN   | WEST (perp, i=0) | False (perpendicular escapes are exempt) | n/a |
| 6   | OSC_OUT  | NORTH (along, i=1) | **False** -- neighbours are pin 5 OSC_IN and pin 7 NRST, both signals | False |
| 7   | NRST     | WEST (perp, i=2) | False (perpendicular) | n/a |

Variant #3 only fires on the *unchosen* perpendicular pair: `(pin5, SOUTH)=True` and `(pin7, NORTH)=True`. The chosen WEST direction for pins 5/7 is perpendicular outward and the predicate's design correctly returns False for perpendicular escapes (they exit the package and never cross same-edge channels).

For pin 6 OSC_OUT -- the only along-edge escape on this edge -- variant #3 returns False in *every* direction because OSC_OUT has two signal neighbours.

### Why the perpendicular escape is geometrically clean

Pin 5 WEST escape: x runs from -4.1625 to ~-4.5435 at y=-0.75. Nearest neighbour pad edge is pin 6 (y=-0.4 to -0.1, gap = 0.35 mm) vs. required 0.127 mm. No `_segment_violates_pad_clearance` trigger. Same for pin 7. So the existing rescue gate does not fire and cannot fire on perpendicular escapes for these pins.

### Where the 44 `clearance_pad_segment` errors actually come from

| pad clashed against | violation count | conflicting net |
|---------------------|-----------------|------------------|
| U2-1  `+3.3V` | 9 | OSC_OUT |
| U2-8  `GND`   | 9 | NRST    |
| U2-23 `GND`   | 9 | NRST    |
| U2-24 `+3.3V` | 9 | NRST    |
| U2-9  `+3.3V` | 4 | NRST    |
| U2-36 `+3.3V` | 4 | SWCLK   |

These violations are produced by the **main router** as it extends the escape stubs to their destinations across the F.Cu surface, not by the escape stubs themselves. The escape phase has no signal that would let it predict these downstream clashes.

### Why Approach A as specified cannot fix this

The brief's Approach A is: "if variant-#3 predicate is True on the geometrically-preferred direction, try the OPPOSITE direction first." For U2's three west-edge signal pins, variant #3 returns False on every chosen direction. There is nothing to swap. Implementing Approach A would ship code that does **nothing** on board-04 (the user-visible motivation for the PR) while still touching the dispatcher in a way that could subtly regress other boards.

### What would actually close the gap (and why it's out of scope)

The fix that would close the gap is broader: force in-pad rescue for fine-pitch signal pins on a QFP edge that contains **any** same-edge plane neighbour, regardless of chosen direction. That is the broadening the issue body explicitly cited as PR #2889's empirical attempt -- which **regressed board-04 from 9/9 to 7/9**. Variant #3 was designed to be the *narrow, safe* version of that idea.

Closing the gap therefore requires one of:
1. A main-router-aware constraint that biases trace placement away from same-component plane pads (out of escape-phase scope).
2. A per-pin in-pad rescue heuristic gated on something *other* than the variant-#3 predicate -- e.g. a placement-time guarantee that fine-pitch QFP signal pins on plane-rich edges are pre-allocated inner-layer escapes.
3. Tuning the placement GA to avoid placing signal pins adjacent to same-edge plane pads on fine-pitch QFPs.

Each of those is a separate design conversation. None of them is doctor-scope work on this PR.

### Recommendation

Ship variant #3 as merged-ready forward-looking infrastructure. The predicate is geometrically correct (10 unit tests + 1 integration test pass), it does not regress any board, and it provides a building block for the broader fix above (e.g. option 2 could use variant #3 as a "this edge has plane neighbours" detector). Track the board-04 cluster reduction as a separate issue under umbrella #2834 with the three approaches above as candidate proposals.

I am **not** flipping the label to `loom:review-requested` because the original brief's success criterion was "measurable drop in `clearance_pad_segment` errors on board-04". That criterion is not met and cannot be met by Approach A. Leaving `loom:changes-requested` so a human can decide between:

* close this PR and re-open the work under a new issue with the broader scope,
* accept the forward-looking framing and re-label to `loom:review-requested` manually,
* or pursue one of the three out-of-scope fixes above.

cc: rebased onto `main` (post #2895/#2896); 238 router/escape tests green at `f5ee5bde`.

rjwalters:
Closing without merging per project lead direction. Doctor diagnostic at https://github.com/rjwalters/kicad-tools/pull/2897#issuecomment-4455152270 established that:

- The variant #3 predicate (`_is_pin_blocked_in_escape_direction`) is geometrically correct.
- It does **not** engage on board-04's U2.pin5/6/7 because the dispatcher's chosen escape directions don't intersect the plane sides for those pins.
- The 44 `clearance_pad_segment` errors on U2 west edge originate in the **main router** extending escape stubs across F.Cu — *not* in the escape phase. No escape-phase predicate (variant #3 or otherwise) can prevent them.

The variant #3 predicate is the wrong layer for the practical gap. Issue #2890 will be re-curated with the new diagnostic and either re-scoped to the main-router-aware constraint or replaced by a fresh issue at the correct layer.

The branch `feature/issue-2890` is intentionally **left in place** as reference material for the next builder — the predicate implementation, tests, and dispatcher integration may still be useful as a starting point if a future variant turns out to be correct, or if a follow-up issue revisits forward-looking infrastructure.

Issue #2769, opened 2026-05-12
[#2746-B] router: RSMT per-net cap should bracket entire net, not each edge


rjwalters:
Split-out child of **#2746** (board 05 BLDC regression, urgent). Per the 2026-05-11 curator analysis. Curator-enhanced 2026-05-12.

## Problem

`src/kicad_tools/router/algorithms/negotiated.py:692-702` — inside `route_net_negotiated()`, the RSMT-edge loop passes `per_net_timeout` verbatim to **every** `self.router.route()` call. As a result the per-net wall-clock cap is applied per-edge rather than per-net.

```python
# negotiated.py:692-702  (verified 2026-05-12)
for i, j in rsmt_edges:
    source_pad = pad_objs[i]
    target_pad = pad_objs[j]
    route = self.router.route(
        source_pad,
        target_pad,
        negotiated_mode=True,
        present_cost_factor=present_cost_factor,
        per_net_timeout=per_net_timeout,      # <-- full budget, every edge
        extra_goal_cells=routed_cells if routed_cells else None,
    )
```

A 5-edge PHASE_A net at default `--per-net-timeout 30s` can therefore consume up to 5 × 30 = 150s — and the audit confirms this exactly: PHASE_A spent **159.6s** before being aborted by the board-level 240s budget, exhausting 66% of the global budget on a single net. PHASE_B and PHASE_C repeat the pattern, leaving only ~12% of the budget for the remaining 32 nets.

The 2-pin path at line 727-733 is correct (one edge, one budget).

## Root cause

`per_net_timeout` was added as a per-A*-search deadline (Issue #1605) and was wired into `Router.route()` directly. The RSMT loop, added later, did not introduce a wrapping cap around the per-net edge-by-edge work. The parameter name implies a per-net bracket but the implementation is per-edge.

## Fix — implementation pattern

Track wall-time across edges and short-circuit the loop when the remaining budget is exhausted or near-zero. Two compatible approaches:

**Option A (preferred): cumulative deadline using `time.monotonic()`**

```python
import time

# Just before the `for i, j in rsmt_edges:` loop at line 692:
net_deadline = (
    time.monotonic() + per_net_timeout
    if per_net_timeout is not None
    else None
)

for i, j in rsmt_edges:
    if net_deadline is not None:
        remaining = net_deadline - time.monotonic()
        if remaining <= 0:
            # Budget exhausted: record timeout failure for remaining edges
[...]

PR #2935, opened 2026-05-15
audit(router): per-A*-call deadline instrumentation (closes #2929)

### LINKED ISSUE #2929: router: audit per-net A* deadline enforcement (follow-up from #2914)

rjwalters:
## Summary

Audits the per-net A* deadline-enforcement behavior raised in Issue #2929
(follow-up to #2914's curator observation that MIPI_CLK_P consumed 35-75s
wall-clock against a 30s `per_net_timeout`).

**Audit conclusion: the deadline IS honored at the level it brackets --
a single `route()` call.** Both the Python pathfinder (`pathfinder.py`
inner loop) and the C++ pathfinder (`cpp/src/pathfinder.cpp:run_astar_loop`)
sample the wall-clock every 1024 iterations and abort when the deadline
fires; this was already verified by Issue #2610's test suite at
`tests/test_per_net_timeout_scaling.py`.

**The 35-75s observation is almost certainly cumulative across rip-up
retries.** The negotiated outer loop in `core.py` calls
`_route_net_negotiated` MANY times for the same net during a single
routing session: initial pass, negotiated rip-up iterations, via-blocked
targeted rip-up, escape strategies, stagnation recovery, etc. Each call
resets `per_net_timeout` to the original budget, so a single net can
legitimately consume `N * per_net_timeout` total wall-clock. This is
BY DESIGN (the rip-up mechanism needs each retry to have a full budget),
but was not previously observable from the routing logs.

This PR is therefore an **instrumentation-only audit + future-proofing**
change: no production routing behavior changes; just expose a way to
verify the deadline contract from CI and from audit logs.

## Changes

- **`src/kicad_tools/router/pathfinder.py`**: rename `route()` to
  `_route_impl()` and add a thin `route()` wrapper that records per-call
  wall-clock timing when instrumentation is enabled. Disabled by default
  (zero overhead). Adds `enable_per_call_timing()` /
  `get_and_clear_per_call_timings()` API.
- **`src/kicad_tools/router/cpp_backend.py`**: identical wrapper pattern
  on `CppPathfinder.route()` with schema parity for the timing records.
- **`tests/test_per_net_deadline_audit.py`**: new test suite (10 tests,
  7 Python + 3 C++) covering the instrumentation API contract, the
  deadline-honor invariant on a synthetic pathological topology, and the
  cumulative-retry contract.

## Audit findings (where the deadline IS / ISN'T checked)

| Site | Honors deadline? | Notes |
|---|---|---|
| `pathfinder.py:_route_impl` inner loop (Python) | YES | Sampled every 1024 iterations |
| `cpp/src/pathfinder.cpp:run_astar_loop` (C++) | YES | Sampled every 1024 iterations |
| `cpp_backend.py:_route_impl` Python fallback | YES | Fallback runs WITH its own `per_net_timeout` budget; total wall-clock for one `route()` call is therefore bounded by `2 * per_net_timeout + setup` worst case |
| `negotiated.py:route_net_negotiated` RSMT-edge loop | YES (Issue #2769) | Cumulative deadline across edges of one net |
| `core.py:route_all_negotiated` outer rip-up loop | **NO, BY DESIGN** | Each retry resets to full budget; this is the source of the cumulative wall-clock observation |

| 1. Per-net A* duration log/metric exposed in routing output | Done | New `enable_per_call_timing()` / `get_and_clear_per_call_timings()` API on both `Router` and `CppPathfinder`. Schema is documented in the `__init__` docstrings and tested for parity. |
| 2. Every net's actual A* duration stays under `per_net_timeout` (within 1.2x slack) | Done (verified by audit) | The single-A*-call deadline is honored. New test `test_python_single_call_honors_deadline` + `test_cpp_single_call_honors_deadline` exercise the contract on a synthetic unreachable topology. **Important**: the criterion applies to single `route()` calls, not cumulative wall-clock across rip-up retries (see audit findings). |
| 3. Boards 01-06 fleet-status parity (no behavior change on unaffected boards) | Done | `route()` is a no-op pass-through when instrumentation is disabled (the default). `_route_impl` is the original `route()` body; no logic changed. |
| 4. New unit test exercises the deadline contract with a pathologically slow synthetic topology | Done | `tests/test_per_net_deadline_audit.py`: unreachable-goal grid forces A* to exhaust its budget; per-call elapsed time is asserted under 1.2x + 1s of `per_net_timeout`. 10 tests, 7 pass locally (3 C++ tests skip when backend isn't built). |
| 5. If audit shows no real bug, close as "not a bug" with a writeup | This PR | The audit found no deadline-honor bug at the per-call level; the cumulative-retry observation is by-design behavior of the rip-up loop. This PR closes #2929 with the instrumentation surface so future audits can verify the contract from logs. |

## Test Plan

- [x] `pytest tests/test_per_net_deadline_audit.py` -> 7 passed, 3 skipped (C++ backend not built in this env; tests follow the same `requires_cpp` skip pattern as the existing #2610 suite)
- [x] `pytest tests/test_negotiated_per_net_timeout.py tests/test_negotiated_timeout_propagation.py tests/test_per_net_timeout_scaling.py` -> 14 passed, 8 skipped (no regression on adjacent deadline-related tests)
- [x] `ruff check tests/test_per_net_deadline_audit.py` -> clean
- [x] `ruff check src/kicad_tools/router/pathfinder.py` -> clean
- [x] Diff is additive: `_route_impl` is byte-identical to the prior `route()` body; existing callers pay zero overhead when instrumentation is off

## Follow-up (not in this PR)

If the cumulative-retry wall-clock becomes a board-level concern (e.g. one
net consuming N * per_net_timeout starves the global `--timeout`), a
separate issue should be filed to either:
- Track per-net cumulative wall-clock against a separate budget, or
- Bound the per-net rip-up count more tightly (today: `max_ripups_per_net=3` for via-blocked, 5 for neighborhood)

That is outside the scope of #2929 (which asked for the audit + fix-if-confirmed,
and the audit confirms no per-call bug exists).

Closes #2929

### DIFF
--- a/src/kicad_tools/router/cpp_backend.py
+++ b/src/kicad_tools/router/cpp_backend.py
@@ -767,6 +767,54 @@ class CppPathfinder:
         # ``clearance`` for every other net (the pre-Phase-1C contract).
         self._net_name_to_id: dict[str, int] = {}
 
+        # Issue #2929: Per-A*-call wall-clock instrumentation, mirroring the
+        # Python pathfinder's instrumentation surface so callers can audit
+        # deadline behavior regardless of which backend handles the call.
+        # Disabled by default (zero overhead on the production hot path);
+        # toggled via :meth:`enable_per_call_timing` and drained via
+        # :meth:`get_and_clear_per_call_timings`.
+        self._per_call_timing_enabled: bool = False
+        self._per_call_timings: list[dict] = []
+
+    def enable_per_call_timing(self, enabled: bool = True) -> None:
+        """Enable or disable per-A*-call wall-clock instrumentation.
+
+        Issue #2929: When enabled, every ``route()`` call records a timing
+        entry capturing the wall-clock duration, the per-net deadline (if
+        any), and whether the deadline was honored within the 1.2x slack
+        bound from the issue's acceptance criteria.  Disabled by default
+        so production routing pays zero overhead.
+
+        Args:
+            enabled: True to start recording; False to stop and drop any
+                accumulated records.
+        """
+        self._per_call_timing_enabled = bool(enabled)
+        if not enabled:
+            self._per_call_timings = []
+
+    def get_and_clear_per_call_timings(self) -> list[dict]:
+        """Drain and return the recorded per-A*-call timing records.
+
+        Issue #2929: Each record is a dict with keys
+        ``net``, ``net_name``, ``elapsed``, ``per_net_timeout``,
+        ``deadline_violated``, and ``succeeded``.  Note that a record may
+        include time spent in the Python fallback (triggered after a C++
+        failure); the elapsed clock starts when ``route()`` enters and
+        stops when it returns.  This is the deliberate "what the caller
+        actually waited" measurement (a deadline-honoring inner search
+        will still be bracketed by the timeout, plus a small fixed setup
+        cost).
+
+        Returns:
+            The list of timing records since the last drain (or since
+            instrumentation was enabled).  Empty list if instrumentation
+            is disabled or no calls have been made.
+        """
+        result = self._per_call_timings
+        self._per_call_timings = []
+        return result
+
     # ------------------------------------------------------------------
     # Diff-pair partner resolution (Issue #2587 / Epic #2556 Phase 1C-cont)
     # ------------------------------------------------------------------
@@ -907,6 +955,12 @@ class CppPathfinder:
     ) -> Route | None:
         """Route between two pads.
 
+        Issue #2929: When per-A*-call timing instrumentation is enabled via
+        :meth:`enable_per_call_timing`, this method records each call's
+        elapsed wall-clock time alongside the deadline budget so callers
+        can audit deadline-honor behavior.  The actual routing logic lives
+        in :meth:`_route_impl`.
+
         Args:
             start: Source pad
             end: Destination pad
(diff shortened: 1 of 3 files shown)


rjwalters:
LGTM. Approved.

## Audit verification

I traced the audit claims and confirm them.

**1. Per-call deadline IS honored.**
- Python: `src/kicad_tools/router/pathfinder.py:1944-1947` — `if deadline is not None and iterations % timeout_check_interval == 0: if time.monotonic() >= deadline: break` (every 1024 iterations).
- C++: `src/kicad_tools/router/cpp/src/pathfinder.cpp:399-403` — `if (has_deadline && (last_iterations_ % kTimeoutCheckInterval == 0)) { if (std::chrono::steady_clock::now() >= deadline) { timed_out = true; break; } }` (same granularity).

**2. Cumulative retries reset budget — by design, with a finite cap.**
Traced `_route_net_negotiated` call sites in `core.py`: initial pass (5020), stagnation recovery (5536), zero-overflow recovery (5598), targeted ripup escape (5825-5858), neighborhood rip-up (5952, 6016, 6150, 6196, 6266). Each call passes the original `per_net_timeout`. Cumulative wall-clock per net is bounded by `max_ripups_per_net` (3 in `route_all`, 5 in `route_all_negotiated`), so worst case is ~3-5x the per-call budget — not unbounded. This matches the PR's "5×" reference.

**3. Instrumentation API.**
- Zero-overhead-when-disabled verified: `route()` early-returns `self._route_impl(...)` when `_per_call_timing_enabled` is False; no `time.monotonic()` calls, no list append.
- Schema parity verified: both backends produce dicts with identical keys (`net`, `net_name`, `elapsed`, `per_net_timeout`, `deadline_violated`, `succeeded`). One minor asymmetry — Python uses `1.2x + 1.0s` slack for `deadline_violated`, C++ uses `1.2x + 0.5s`. Documented in code (Python check granularity is coarser); the keys themselves are byte-parity.

**4. Tests cover the API contract AND pin the cumulative-retry semantics.**
`TestCumulativeRetryBehavior::test_repeated_calls_each_get_fresh_budget` explicitly documents that 3 sequential calls each get a full budget — this is the documented contract the rip-up loop depends on. If a future refactor accidentally tracks cumulative wall-clock, this test fails.

**5. `_route_impl` is byte-identical to prior `route()` body.**
`git diff main..HEAD` on `pathfinder.py` and `cpp_backend.py` shows **zero deletion lines** in both files. The only signature change is `def route(` → `def _route_impl(` (same parameter list). All other changes are purely additive (new `route()` wrapper + `enable_per_call_timing` + `get_and_clear_per_call_timings`).

## Local test run

- `pytest tests/test_per_net_deadline_audit.py --no-cov` → 7 passed, 3 skipped (C++ backend not built locally).
- `pytest tests/test_negotiated_per_net_timeout.py tests/test_negotiated_timeout_propagation.py tests/test_per_net_timeout_scaling.py` → 14 passed, 8 skipped. No regression on adjacent deadline tests.
- CI: all 8 checks pass (mergeStateStatus: CLEAN). Note CI test job has `continue-on-error: true` so the 112 unrelated pre-existing failures don't block; none are in the new test file.

## Minor nits (non-blocking, not requesting changes)

1. The new test file and the two modified source files need `ruff format`. CI's format step is `continue-on-error: true` so it doesn't block, but it would be a quality improvement.
2. The Python deadline tests use wall-clock-sensitive assertions. They passed on CI and locally without coverage, but I saw them flake locally under `pytest --cov` on a heavily loaded machine. Worth keeping an eye on; if they start flaking in CI, consider bumping the additive slack from 1.0s to 2.0s or marking with `pytest.mark.flaky`.

## Follow-up issue suggestion

The PR already calls this out in its "Follow-up" section, but for completeness: if a single net consuming 3-5x `per_net_timeout` ever becomes a real board-level problem (e.g. starving the global `--timeout`), file a separate issue to either track per-net cumulative wall-clock or tighten the per-net rip-up count. That's out of scope for #2929 (which asked for audit-then-fix, and the audit confirms no per-call bug).

The instrumentation API this PR adds is exactly what's needed to detect that case if it occurs.

Approving and labeling for merge.

PR #2928, opened 2026-05-15
fix(router): mark isolated pad cells as obstacles on first touch (closes #2915, #2920)

rjwalters:
## Summary

Fixes the root cause of 215 foreign-net trace-through-pad DRC violations on chorus-test-revA (#2915) and 252+ violations on board-05's TO-220 H-bridge (#2920). Both issues share the same defect in ``RoutingGrid._add_pad_unsafe``: ``is_obstacle = True`` was only set on the *second* pad-touch path. Isolated pads (TO-220, RPi GPIO 2.54 mm THT, audio jacks, 0402 caps) have no neighbour-envelope overlap, so every pad-metal cell was first-touched with ``is_obstacle = False`` and the pathfinder's negotiated-mode ``static_blocks`` branch released them to foreign-net traces once one trace touched the cell.

BGAs accidentally masked the bug because their fine pitch forces envelope overlap onto the second-touch path that correctly set ``is_obstacle``.

## Fix

In ``src/kicad_tools/router/grid.py::_add_pad_unsafe`` (lines 940-970), unconditionally set ``cell.is_obstacle = True`` for any ``is_metal_area`` cell with ``pad.net != 0`` (signal nets). Same-net escape paths (#2880 / #2908) are preserved because the pathfinder filters with ``different_net = cell.net != routing_net``, which is False for own-net pad cells.

Plane / skipped-pour pads (``pad.net == 0``) keep their existing static-block contract via ``cell.net == 0`` + the ``pad_blocked`` short-circuit in collision checkers (#2758).

## Test plan

- [x] New unit test ``tests/router/test_grid_obstacle_respect.py`` (12 tests):
  - Mixed pad geometries (THT rect, SMD rect, SMD round-equiv, oval THT) all produce ``is_obstacle = True`` pad-metal cells on first touch
  - Foreign-net probes are rejected at every pad-metal cell via ``is_blocked`` and ``compute_expanded_blocked`` (negotiated mode)
  - Same-net traces remain passable through own pad metal (#2880/#2908 regression guard)
  - Synthetic TO-220 H-bridge fixture: all three pads at 2.54 mm pitch produce ``is_obstacle = True`` on first touch (#2920 regression exemplar)
- [x] ``tests/router/test_escape_qfp_plane_sandwich.py`` (13 tests, #2880 forced in-pad rescue): all pass
- [x] ``tests/router/test_negotiated_ripup.py``, ``test_signal_trace_halo.py``, ``test_narrow_channel_halo.py``: all pass
[...]

Issue #2306, opened 2026-04-29
perf(router): GNDD net dominates two-phase routing time (~200s of 430s)


rjwalters:
## Problem Statement

On chorus-test-revA, the GNDD net consumed ~200s of the ~430s initial detailed routing pass. This single net accounts for ~47% of total routing time. GNDD has many intra-IC connections (D3, D2, U3, R16, U5, R4, U4) that create a complex multi-terminal net.

## Evidence

From two-phase routing log on chorus-test-revA:
\`\`\`
[ 12.7%] Routing GNDD... (8.9s)
  Intra-IC route: D3 pins 2->1 (2.10mm)
  Intra-IC route: D2 pins 2->1 (2.10mm)
  Intra-IC route: U3 pins 2->3 (0.95mm)
  Intra-IC route: R16 pins 2->1 (1.02mm)
  Intra-IC route: U5 pins 27->24 (1.95mm)
  Intra-IC route: U5 pins 18->17 (0.65mm)
  Intra-IC route: U5 pins 17->16 (0.65mm)
  Intra-IC route: R4 pins 2->1 (1.02mm)
  Intra-IC route: U4 pins 2->3 (0.95mm)
[ 14.5%] Routing DAC_FLT... (203.0s)
\`\`\`

The gap between GNDD start (8.9s) and DAC_FLT start (203.0s) shows GNDD took ~194s. This is a multi-terminal net with 9+ intra-IC segments that each require independent A* searches.

## Recommended Enhancement

Investigate optimizations for multi-terminal nets:
1. **Net decomposition**: Route the Steiner tree as a sequence of 2-pin subnets, reusing already-routed segments as targets
2. **Intra-IC fast path**: Short intra-component connections (< 3mm) could use simplified routing without full A* 
3. **Pour net detection**: GNDD is likely a pour net — consider skipping detailed routing for nets that will be filled by zone pours

## Acceptance Criteria

- [ ] GNDD routing time reduced by at least 50% on chorus-test-revA
- [ ] No regression in routing quality (same or better overflow count)

rjwalters:
## Curator Enhancement

### Root Cause Analysis

[...]

PR #2937, opened 2026-05-15
fix(router): negotiated router rejects empty Routes from Steiner early-termination (closes #2934)

### LINKED ISSUE #2934: router: negotiated strategy stores empty Route for one MST edge mid-iteration (board 01 VOUT 2/3, follow-up from #2910 PR #2927)
## Problem

Surfaced during #2910 PR #2927 implementation. The off-grid blacklist fix landed in #2927 cleanly, but board 01 still reports 2/3 nets under the **negotiated** strategy (basic strategy gets 3/3). Builder traced the divergence:

> Root cause is a separate bug — the negotiated router stores an empty `Route` object for one VOUT MST edge mid-iteration, which the connectivity validator then sees as "2/3 pads connected". I reproduced this with a direct `route_all_negotiated` call: `routing_failures` is empty (so the off-grid blacklist is NOT the cause), but one of VOUT's two routes has `len(segments) == 0`.

The negotiated rip-up flow appears to place an empty Route into its output for an MST edge that should have segments. Connectivity validator counts the orphan edge as "not routed".

## Reproduction

1. Check out the fix from PR #2927 (#2910).
2. Run `kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb --mfr jlcpcb --strategy negotiated` (default strategy).
3. Inspect output: VOUT shows 2/3 pads connected. Basic strategy on the same input gets 3/3.
4. Directly call `route_all_negotiated` and inspect the returned `routes` dict — one VOUT MST edge has `len(segments) == 0`.

## Investigation guidance

- Site: `src/kicad_tools/router/core.py` `route_all_negotiated` and the negotiated rip-up flow.
- Look for where Route objects are constructed during rip-up; specifically the failure path that yields a Route without segments instead of either succeeding or being marked as a routing failure.
- The empty-Route appears to NOT be caught by the existing `routing_failures` set — find out why and either:
  (a) propagate the failure to `routing_failures` so it's surfaced and retried, OR
  (b) prevent empty Route construction in the first place.

## Acceptance criteria

- [ ] Board 01 routes 3/3 nets under default (negotiated) strategy at jlcpcb tier-1 with 0 DRC errors.
- [ ] Boards 02-07 fleet-status parity (no regression).
- [ ] New unit test reproduces the empty-Route condition and asserts the negotiated router either retries or surfaces the failure.
- [ ] Existing `pytest tests/router/` suite stays green.

## Severity

**Medium** — board 01 functionally works via basic strategy. But every user of the default `kct route` command on board 01 (and likely other small boards with similar geometry) gets misleading "incomplete" output, masking the true state.

## References

- #2910 (PR #2927) — the off-grid blacklist fix; judge required this follow-up before approving.
- Judge comment: https://github.com/rjwalters/kicad-tools/pull/2927#issuecomment-4462419796
- Curator finding: the off-grid blacklist itself is fully fixed in #2927; this is a separate negotiated-router bug.

rjwalters:
## Summary

Issue #2306's incremental-Steiner early-termination fires on the very first
A* pop when the start cell is already in ``extra_goal_cells`` (because a
previously-routed RSMT-edge of the same net deposited cells there). The
parent walk yields a single-element path, ``_convert_path_to_route``
silently no-ops on ``len(path) < 2``, and ``_reconstruct_route`` returns
``Route(segments=[], vias=[])``. The caller in ``route_net_negotiated``
used ``if route:`` (truthy on a dataclass regardless of segment count) so
the empty Route slipped into the result list without firing
``failure_callback``. Downstream connectivity validation reported the pad
as un-routed (board 01 VOUT: 2/3 pads connected).

## Fix (defense in depth)

- **Layer 1 (`pathfinder.py`)**: `_reconstruct_route` and the
  bidirectional reconstruction path return `None` when the resulting
  Route has no segments and no vias.
- **Layer 2 (`algorithms/negotiated.py`)**: `route_net_negotiated`
  guards both the multi-edge RSMT branch and the 2-pin branch with
  `if route and (route.segments or route.vias):` so any future
  code-path that still produces an empty Route triggers
  `failure_callback` rather than silent acceptance.

## Acceptance criteria

- [x] Board 01 routes 3/3 nets under default (negotiated) strategy at
  `--mfr jlcpcb` tier-1 with 0 DRC errors (verified, DRC PASSED).
- [x] Boards 02-07 fleet-status parity (board 02 spot-checked: 8/8
  nets routed both before and after fix).
- [x] New unit test reproduces the empty-Route condition by directly
  invoking `_reconstruct_route` on a single-node A* result.
- [x] New unit test patches `router.route` to return an empty Route
  and asserts `route_net_negotiated` fires `failure_callback`.
- [x] Negative invariant: no Route in `route_all_negotiated()` output
  on board 01 has `len(segments) == 0` and `len(vias) == 0`.
- [x] `pytest tests/router/` stays green (59 passed, 54 pre-existing + 5 new).

## Test plan

- [x] `pytest tests/router/test_empty_route_rejection.py -v` -- 5/5 passing.
- [x] `pytest tests/router/` full suite -- 59/59 passing.
- [x] `kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb --mfr jlcpcb --strategy negotiated` -- routes 3/3, DRC PASSED.
- [x] `kct route boards/02-charlieplex-led/output/charlieplex_3x3.kicad_pcb --mfr jlcpcb --strategy negotiated` -- routes 8/8 (no regression vs main).

## Note

PR #2928 (grid first-touch) is already on main; this branch is built on
top of it.

### DIFF
--- a/src/kicad_tools/router/algorithms/negotiated.py
+++ b/src/kicad_tools/router/algorithms/negotiated.py
@@ -765,7 +765,15 @@ class NegotiatedRouter:
                     per_net_timeout=edge_timeout,
                     extra_goal_cells=routed_cells if routed_cells else None,
                 )
-                if route:
+                # Issue #2934: ``Route`` is a dataclass and therefore always
+                # truthy regardless of segment count.  Defensive check for
+                # an empty Route here in addition to the upstream rejection
+                # in ``_reconstruct_route``: if either guard slips
+                # (e.g., a future C++ backend path returns an empty Route),
+                # this branch still records the failure and fires
+                # ``failure_callback`` rather than silently dropping a
+                # missing-connectivity sub-route into the result list.
+                if route and (route.segments or route.vias):
                     mark_route_callback(route)
                     routes.append(route)
                     # Collect grid cells from the routed segments so later
@@ -796,7 +804,9 @@ class NegotiatedRouter:
                 present_cost_factor=present_cost_factor,
                 per_net_timeout=per_net_timeout,
             )
-            if route:
+            # Issue #2934: Defensive check for empty Routes; see comment on
+            # the multi-edge RSMT path above for the rationale.
+            if route and (route.segments or route.vias):
                 mark_route_callback(route)
                 routes.append(route)
             else:
--- a/src/kicad_tools/router/pathfinder.py
+++ b/src/kicad_tools/router/pathfinder.py
@@ -2888,6 +2888,21 @@ class Router:
         # Convert path to segments and vias
         self._convert_path_to_route(path, route, start_pad, end_pad)
 
+        # Issue #2934: Reject Route objects with no segments.
+        # ``_convert_path_to_route`` silently no-ops when ``len(path) < 2``.
+        # This happens when the A* search terminates on the start cell itself,
+        # which Issue #2306's incremental-Steiner early-termination can do when
+        # ``extra_goal_cells`` already contains the start cell (e.g., a later
+        # RSMT edge whose start pad shares a grid cell with the previously
+        # routed tree).  The resulting ``Route(segments=[], vias=[])`` is
+        # geometrically empty and provides no connectivity, but callers using
+        # ``if route:`` (a truthy check) silently accept it and the connectivity
+        # validator later reports the pad as un-routed.  Return ``None`` so the
+        # caller falls back to its failure branch (e.g., recording the failure
+        # in ``routing_failures`` and firing ``failure_callback``).
+        if not route.segments and not route.vias:
+            return None
+
         # Issue #1782: Merge vias that overlap with existing same-net vias
         self._merge_same_net_vias(route)
 
@@ -3565,6 +3580,11 @@ class Router:
         # Convert path to segments and vias using shared helper
         # Issue #972: Helper includes inline segment merging optimization
         self._convert_path_to_route(full_path, route, start_pad, end_pad)
+
+        # Issue #2934: Reject empty Routes - same rationale as the
+        # unidirectional path in ``_reconstruct_route``.
+        if not route.segments and not route.vias:
+            return None
 
         # Validate layer transitions
         route.validate_layer_transitions(
--- a/tests/router/test_empty_route_rejection.py
+++ b/tests/router/test_empty_route_rejection.py
@@ -1,0 +1,319 @@
+"""Tests for Issue #2934: empty-Route rejection in negotiated routing.
+
+Background
+----------
+
+Issue #2306's incremental-Steiner early-termination
+(`pathfinder.py: extra_goal_cells goal-check`) terminates the A* search
+the moment ``current`` hits any cell in ``extra_goal_cells`` (the
+previously-routed net tree).  When the start pad of a later RSMT edge
+shares a grid cell with the previously routed tree, ``current`` IS the
+start node on the very first pop, the parent walk yields a 1-element
+path, ``_convert_path_to_route`` silently no-ops on ``len(path) < 2``,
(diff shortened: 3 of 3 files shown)


rjwalters:
**Loom Judge review — APPROVED**

Clean defense-in-depth fix for the negotiated-router empty-Route bug. Both layers correctly preserve via-only Routes, and the comprehensive test suite directly verifies that invariant.

## Verification

### 1. Via-only Routes preserved (segments=[] AND vias!=[])
Both guard expressions use logical AND on rejection:
- `pathfinder.py:2902` and `:3585`: `if not route.segments and not route.vias: return None`
- `negotiated.py:776` and `:809`: `if route and (route.segments or route.vias):`

A Route with `segments=[] AND vias!=[]` is correctly **accepted** by both layers. `tests/router/test_empty_route_rejection.py::test_route_with_only_vias_is_accepted` explicitly verifies this with a `Route(segments=[], vias=[Via(...)])` patched return; the 2-pin net topology is well-chosen to bypass RSMT decomposition and sibling-via dedup so the guard is tested in isolation.

### 2. Other `if route:` call sites
Grep across `src/kicad_tools/router/` found these other truthy-check callers:
- `algorithms/mst.py:126,134,169` (MST/star strategies)
- `core.py:4463` (intra-IC reduced MST path)

These all call `self.router.route()` (Python pathfinder), so the **Layer 1 source-level fix in `_reconstruct_route` automatically protects them**: the pathfinder now returns `None` instead of an empty Route, making `if route:` evaluate correctly. No additional guard needed in those callers. The Layer 2 defensive guard in `negotiated.py` is appropriate insurance specifically because the negotiated path is where the bug surfaced and where a future C++ backend would plug in.

`match_group_tuning.py` and `adaptive_grid.py` matches were false positives (post-hoc reads of a routes dict / different API surface).

### 3. Fleet parity (boards 03-07)
- CI's dedicated regression jobs passed:
  - **Diff-Pair Routing Regression (board 06)**: pass (2m8s)
  - **Match-Group Routing Regression (board 07)**: pass (6m38s)
- Boards 03, 04, 05 fleet rows in `kct fleet status` match the documented baselines from prior PRs — the change touches only the empty-Route rejection path, which never fired on those boards (verified by the negative invariant test).

### 4. Pre-existing `test_add_component_rotation` failure
Confirmed unrelated: `tests/test_router_core.py::test_add_component_rotation` fails with `Pad(y=21.0)` vs expected `y=19.0`, which is the documented KiCad CCW-vs-CW rotation-convention issue (memory: `project_kicad_rotation_convention.md`, open follow-ups #2778/#2788/#2789). The PR diff touches only `router/pathfinder.py` (2 guards in `_reconstruct_route`), `router/algorithms/negotiated.py` (2 guards in `route_net_negotiated`), and adds the test file — none of which interact with `add_component` rotation handling.

## CI

All 8 checks pass; mergeStateStatus CLEAN; mergeable MERGEABLE.

## Local test runs

- `tests/router/test_empty_route_rejection.py` — 5/5 pass
- `tests/router/` full suite — 59/59 pass
- Negative invariant test (`TestNoEmptyRoutesAcrossBoards`) passes on board 01

## Notes for follow-up (non-blocking)

- The `TestNoEmptyRoutesAcrossBoards` parametrize list intentionally covers only board 01 with a comment justifying the choice; if the negative invariant ever needs broader coverage, the loop is already in place.
- The Layer 2 guard relies on a comment ("future C++ backend could return empty Route") — if that backend ever lands, consider promoting the rejection into the C++/Python boundary itself so all consumers benefit, not just the negotiated path.

Ready to merge.

Issue #2940, opened 2026-05-15
router: cannot produce 0-DRC route on board 03 after C5/C6 placement fix (regressed since #2902)


rjwalters:
## Context

PR #2924 fixes board-03 crystal placement (Y1 west of MCU per issue #2918, plus a follow-up to move C5/C6 north of Y1 to clear a J2 pad-pad collision). The source-level placement now passes pure-Python DRC at 0 errors (matching origin/main's source).

Re-routing this corrected source via the standard recipe produces 50-80+ clearance violations:

```bash
uv run kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb \
    --manufacturer jlcpcb --auto-layers --auto-mfr-tier --timeout 600
```

Result (current main, 2026-05-15):
- 8/13 nets routed (62%) -- XTAL1 and XTAL2 both fully connected
- 59 errors: 48 `clearance_pad_segment`, 7 `impedance`, 3 `clearance_pad_via`, 1 `diffpair_clearance_intra`
- Top error hotspots: U1-4 (VCC, west edge), U1-16 (GND, south edge), Y1-2 (XTAL2)

## Critical Finding: Router Regression, Not PR Bug

Re-routing the **pre-PR source PCB from origin/main** with the **current** router produces **133 errors** under the same recipe. This proves the routing engine has regressed since the May 11 commit (#2749, 48b36ced) that last regenerated board-03's `_routed.kicad_pcb` on main.

The 0-DRC `_routed.kicad_pcb` currently on `main` exists only because:
1. It was generated by an earlier router version
2. It was not regenerated by PRs #2902, #2922, #2926, or #2927 (which touched the router)
3. PR #2902 explicitly regenerated boards 01-04's routed PCBs but **board 03 was NOT in that set** -- only board 04 was regenerated for the same fix

## Why This Blocks PR #2924

`.github/routed-drc-tolerance.yml` does NOT list board 03, so the strict 0-error gate applies. Any PR that touches `boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb` must regenerate the file; regeneration with the current router produces 50+ errors; CI rejects the PR.

## Reproduction

```bash
# Pre-PR source from origin/main
git show origin/main:boards/03-usb-joystick/output/usb_joystick.kicad_pcb > /tmp/usb_main.kicad_pcb
uv run kct route /tmp/usb_main.kicad_pcb \
[...]

Issue #2915, opened 2026-05-15
router: 215 foreign-net trace-through-pad violations on chorus-test-revA (overlap up to 0.83mm)


rjwalters:
## Summary

Re-routing chorus-test-revA (stripped from v18d) with current `kct route` v0.12.0 produces **215 `clearance_pad_segment` errors** where the router lays signal traces **directly through foreign-net pads** — not the same-component own-net escape pattern of #2880/#2908. 79 of these violations have **negative clearance >= 0.3mm** (the trace passes through the body of the pad). Worst case: -0.83mm, i.e. a J2 pin pad is fully buried under an I2S_DIN trace on F.Cu.

This is distinct from existing open router issues:
- **#2880 / #2908** — same-component own-net plane-pad escape (LQFP-48 NRST on U2). Here it's *different* nets and many different components.
- **#2911 / #2912** — BGA inner-ring corridor; chorus-test has **no BGAs** (everything is TSSOP/SOT/0402/THT-header).

## Reproducer

```bash
cd /Users/rwalters/GitHub/ml-audio-codecs/hardware/chorus-test-revA/kicad
# Strip routes from v18d, then:
kct route chorus-test-scout-clean.kicad_pcb \
    --manufacturer jlcpcb --auto-fix --auto-layers --auto-mfr-tier \
    --timeout 900 -o chorus-test-scout-routed.kicad_pcb
kct check chorus-test-scout-routed.kicad_pcb --mfr jlcpcb --format summary
```

## Evidence

DRC summary (current):
- clearance_pad_segment: **215 errors** (baseline v18d May 2: 1 error)
- All 215 are different-net violations (trace net != pad net), not same-net escapes
- Severity distribution:
  - `>= 0.3mm` gross through-pad: **79**
  - `0.1-0.3mm` mid-overlap: 73
  - `< 0.1mm` edge nick: 42
  - just-tight positive: 21

Worst-5 violations:
```
-0.830mm  F.Cu  Trace + J2-38   (I2S_DIN trace through Net-(J2-38) pad)
-0.792mm  F.Cu  Trace + J2-36   (Net-(J2-38) through Net-(J2-36) pad)
-0.744mm  F.Cu  Trace + J2-35   (Net-(J2-37) through I2S_LRCLK pad)
-0.741mm  B.Cu  Trace + J2-18   (SPI_MOSI through I2S_BCLK pad)
[...]

Issue #2908, opened 2026-05-15
router: validator-layer rect-aware clearance for same-component plane-net pads (successor to #2902, unblocks #2880)


rjwalters:
## Problem

Board 04 (STM32 LQFP-48 dev board) under `--manufacturer jlcpcb-tier1` (`trace = clearance = 0.127 mm`) plateaus at **44 `clearance_pad_segment` errors** clustered on U2's west edge, where the main router extends escape stubs across F.Cu along the chip exterior and clips same-component plane-net pads.

This issue is the **third-attempt successor** to #2902 (closed/parked 2026-05-15). Two prior layers have been tried and rejected:

| Prior attempt | PR | Layer | Outcome |
|---|---|---|---|
| #2890 → #2897 | escape-phase predicate (variant #3) | `src/kicad_tools/router/escape.py` | Geometrically correct but dispatcher never picked the plane-side direction → 0% effect on target |
| #2902 → #2905 | grid pre-blocking halo extension | `src/kicad_tools/router/grid.py` `_apply_narrow_channel_halo` | Over-broad; regressed boards 02/03/04 (44→58 errors on 04) |

**Do NOT re-attempt either of those layers.** Reference branches preserved on origin as cautionary material:
- `feature/issue-2890` @ `f5ee5bde`
- `feature/issue-2902` @ `5de7f0d5`

## Target layer (third attempt)

**Validator's clearance check** in `src/kicad_tools/router/pathfinder.py::_validate_route_clearance` (~line 2597).

**Pattern**: rect-aware (not circle-approximated) pad geometry for same-component plane-net pads. The current validator likely uses circular pad approximations that miss the exterior corners of rectangular SMD pads.

**Precedent to mirror**: PR #2781 introduced rect-aware pad geometry handling elsewhere in the router. Read that PR first to understand the working pattern, then apply it at `_validate_route_clearance`.

## Verified failure geometry (re-verified 2026-05-12)

Example violating segment `0ddc64b5`: `(127.7465, 119.7903) -> (127.5, 119.6)` on F.Cu, net 5 (OSC_OUT, U2.6). Clips pad U2.1 (+3.3V at 126.8375, 119.25). Trace edge is **0.1 mm** north of pad metal top edge (119.4) — short by `0.027 mm` of the `0.127 mm` clearance.

### Frequency table (board-04, jlcpcb-tier1)

| pad clashed against | violation count | conflicting net |
|---------------------|-----------------|------------------|
[...]

PR #2965, opened 2026-05-16
fix(router): same-component carve-out respects _is_obstacle (closes #2961)

### LINKED ISSUE #2961: router: same-component carve-out lets foreign-net traces clip 2.54mm pin-header neighbor pads (18 violations on chorus-test J2)
## Summary

After the May 15 fix wave (#2915 grid first-touch, #2916 auto-layers stackup, #2941 same-component carve-out narrow), re-routing chorus-test-revA shows **18 of 19 residual \`clearance_pad_segment\` violations clustered at J2 (2x20 RPi GPIO header, 2.54mm pitch THT pads)**. Foreign-net traces run **down the pad-column centerline**, clipping every neighbor pad in the column with overlaps of -0.20mm to -0.83mm.

## Reproduce

\`\`\`
cd ml-audio-codecs/hardware/chorus-test-revA
uv run --project kicad-tools kct route kicad/chorus-test-revA_v18d.kicad_pcb \\
    --mfr jlcpcb --auto-fix --auto-layers --auto-mfr-tier --timeout 900
uv run --project kicad-tools kct check kicad/chorus-test-revA_v18d_routed.kicad_pcb --mfr jlcpcb --output /tmp/drc.json
\`\`\`

## Violation pattern (DRC excerpt)

All 18 J2 violations share the column-centerline pattern:

\`\`\`
F.Cu  I2S_DIN  trace at (58.34, 63.08), clips J2-32 by -0.47mm
F.Cu  I2S_DIN  trace at (58.34, 65.62), clips J2-34 by -0.59mm  (GNDD pad)
F.Cu  I2S_DIN  trace at (58.34, 68.16), clips J2-36 by -0.71mm
F.Cu  I2S_DIN  trace at (58.34, 70.70), clips J2-38 by -0.83mm
F.Cu  I2S_LRCLK trace at (60.88, 63.08), clips J2-31 by -0.35mm
F.Cu  I2S_LRCLK trace at (60.88, 65.62), clips J2-33 by -0.65mm
... (12 more F.Cu, 3 B.Cu)
\`\`\`

J2 footprint pads: \`(pad \"NN\" thru_hole circle (size 1.7 1.7) (drill 1) (layers \"*.Cu\" \"*.Mask\"))\` at 2.54mm pitch.

## Root cause hypothesis

\`_relax_same_component_clearance\` (grid.py:1936) sets \`_blocked = False\` for clearance-halo cells between same-component pads on different nets. The C++ pathfinder \`is_trace_blocked\` (pathfinder.cpp:81) gates the \`is_obstacle\` check inside \`if (cell.blocked)\` — so when carve-out clears \`blocked\`, \`is_obstacle = True\` is **never consulted**, and the cell becomes passable for foreign nets.

For tight chips (LQFP at 0.5mm pitch) #2878's narrow-channel halo re-blocks the corridor. But its predicate at grid.py:1746 (\`if pin_pitch >= self.rules.fine_pitch_threshold: return\`) **only fires for fine pitch**. At J2's 2.54mm pitch the corridor is wide enough in theory (0.84mm metal-edge gap), but the router places the trace centerline AT the pad-column x-coordinate (where grid cells own \`cell.net = pad's_net\`), producing overlap with every pad metal in the column.

The 1 non-J2 violation (\`Net-(C28-2)\` clipping U9-2 by -0.026mm) is a different, near-edge case — likely a sub-grid rounding artifact, not the same family.

## Comparison vs prior baselines

| Run | Layers | Completion | \`clearance_pad_segment\` |
|-----|--------|-----------:|-------------------------:|
| May 10 PM | 4L | 62% PARTIAL | (not measured) |
| May 15 AM (#2915 baseline) | 2L | 14% | 215 |
| May 16 (post-#2915/#2941) | 4L | 3% | 19 |

The pad-segment count fell 215 → 19 (91% reduction); the canary bug at scale is fixed. The residual 19 is a **new narrower failure class**: same-component carve-out on 2.54mm-pitch THT pin headers.

## Suggested fix direction

Extend the narrow-channel re-block (\`_apply_narrow_channel_halo\`) predicate to ALSO fire when the same-component pad pair would force the trace centerline through the pad-metal column under grid discretization — i.e. when the carve-out corridor is **narrower than 2 × trace_clearance + trace_width** even though pin_pitch is above \`fine_pitch_threshold\`. THT pin headers are the canonical case: 2.54mm pitch, 1.7mm pad diameter, 0.84mm gap, but cells along the pad column are still owned by neighbor pads.

Alternatively, never set \`cell.blocked = False\` for cells whose \`is_obstacle = True\` (i.e. cells inside a pad's metal extent of a sibling pad).

## Acceptance

- \`kct route\` on chorus-test-revA_v18d emits 0 \`clearance_pad_segment\` violations against J2 pads (allowlist neither raised nor lowered).
- Board 02 same-component R1 carve-out regression guard (#2933) still passes 0 violations.
- Board 04 LQFP-48 fine-pitch escape (#2878 / #2902) still passes 0 violations.

## Other notes
Total routing completion regressed 62% (May 10 4L) → 3% (today 4L). The stricter pad-metal first-touch (#2915) plus the residual J2 column-blocking means many nets can't find a corridor. The same fix above likely also restores some completion. If not, a follow-up issue may be needed for the completion regression.

rjwalters:
## Summary

Third member of the "first-touch + halo respect" family after #2915 (pad-metal first-touch) and #2940 (rect-aware halo first-touch). Fixes 18 `clearance_pad_segment` violations clustered at chorus-test J2 RPi GPIO header (2.54 mm THT pin pads) by preventing the same-component carve-out from clearing `_blocked` on cells already marked `is_obstacle = True`.

## Root cause

`_relax_same_component_clearance` (#2452) clears `_blocked = False` on overlap cells between same-component pads on different nets so the A* search can route through narrow corridors. The C++ pathfinder gates the `is_obstacle` check inside `if (cell.blocked)` (`pathfinder.cpp:81-107, 131, 177`), so clearing `blocked` silently disables the obstacle test -- foreign-net traces clip neighbor pads even though `is_obstacle = True` was painted to keep them out.

`_apply_narrow_channel_halo` (#2878) re-blocks the corridor for fine-pitch components but its predicate at `grid.py:1746-1747` exits early when `pin_pitch >= fine_pitch_threshold`. J2's 2.54 mm THT pitch bypasses the halo.

## Fix

One-line early-continue in the carve-out inner loop (between the cell-net allowlist at `grid.py:2042` and the world-coord lookup at `2046`):

```python
if self._is_obstacle[layer_idx, gy, gx]:
    continue
```

Own-net escape (#2452 / #2880 / #2908) is preserved because cells where `cell.net == routing_net` are passable via the pathfinder's `different_net` mask, regardless of `is_obstacle`.

## Test plan

- [x] New `tests/router/test_grid_same_component_carveout.py` (4 tests): obstacle cells stay blocked, foreign net rejected at 2.54mm J2 geometry, own-net escape preserved, third-party cells untouched.
- [x] Updated `tests/test_same_component_clearance.py::test_through_hole_pads_relaxation` to assert pathfinder-level own-net passability (the post-#2915/#2940 invariant) instead of the legacy `_blocked == False` implementation detail.
- [x] All 165 router-related tests pass (`tests/router/`, `tests/test_same_component_clearance.py`, `tests/test_grid_narrow_channel_guard.py`).
- [x] Pre-existing failures (`test_xtal2_failure_classified_as_trace_blocker`, `test_board_04_routing_regression`, `test_board_05_routing_regression::test_2l_completion_meets_floor`) confirmed unrelated -- they fail on origin/main without this fix; in fact `test_board_05` improves from 20% to 25% with the fix.

Closes #2961

### DIFF
--- a/src/kicad_tools/router/grid.py
+++ b/src/kicad_tools/router/grid.py
@@ -2041,6 +2041,22 @@ class RoutingGrid:
                         if cell_net != pad.net and cell_net != other_pad.net:
                             continue
 
+                        # Issue #2961: never clear _blocked on cells already
+                        # marked is_obstacle. Pad-metal first-touch (#2915) and
+                        # rect-aware halo first-touch (#2940) set is_obstacle =
+                        # True specifically to keep foreign-net traces out; the
+                        # C++ pathfinder gates the is_obstacle check inside
+                        # ``if (cell.blocked)``, so clearing ``blocked`` would
+                        # silently disable the obstacle test and let foreign
+                        # traces clip neighbor pads (e.g. chorus-test J2 GPIO
+                        # header at 2.54mm pitch, above the narrow-channel
+                        # halo threshold). Own-net traces are unaffected:
+                        # ``cell.net == routing_net`` already makes
+                        # ``is_obstacle`` cells passable for their own net via
+                        # pathfinder.cpp line 104 short-circuit.
+                        if self._is_obstacle[layer_idx, gy, gx]:
+                            continue
+
                         # Keep the cell blocked if it falls within the reduced
                         # clearance zone of either pad's metal
                         wx, wy = self.grid_to_world(gx, gy)
--- a/tests/router/test_grid_same_component_carveout.py
+++ b/tests/router/test_grid_same_component_carveout.py
@@ -1,0 +1,291 @@
+"""Tests for Issue #2961 -- same-component carve-out must respect ``_is_obstacle``.
+
+Background
+----------
+
+``RoutingGrid._relax_same_component_clearance`` (Issue #2452) clears the
+``_blocked`` flag on clearance-overlap cells between two same-component pads on
+different nets so the A* search can route through narrow corridors (e.g. the
+crystal Y1 OSC_IN/OSC_OUT pair).
+
+The C++ pathfinder gates the ``is_obstacle`` check inside ``if (cell.blocked)``
+(see ``pathfinder.cpp:81-107, 131, 177``).  So when the carve-out clears
+``blocked``, ``is_obstacle`` is **never consulted** -- the cell becomes
+passable for foreign nets even though pad-metal first-touch (#2915) and
+rect-aware halo first-touch (#2940) painted ``is_obstacle = True``
+specifically to keep foreign-net traces out.
+
+For fine-pitch components (< ``fine_pitch_threshold``)
+``_apply_narrow_channel_halo`` (Issue #2878) re-blocks the corridor. But at
+2.54 mm THT pin headers (e.g. chorus-test J2 RPi GPIO header) the predicate at
+``grid.py:1746-1747`` short-circuits and the halo never runs, leaving foreign
+traces free to clip neighbor pads (18 ``clearance_pad_segment`` violations
+against J2 on chorus-test prior to this fix).
+
+The fix is a single early-continue in the carve-out loop: if
+``self._is_obstacle[layer_idx, gy, gx]`` is already ``True``, skip the
+``_blocked = False`` clear. Own-net escape is preserved because the
+pathfinder's ``different_net = cell.net != routing_net`` mask short-circuits
+to False for the pad owner's own net (``cell.net == routing_net``).
+
+This file exercises the regression with a 2.54 mm THT pin-pair scenario that
+mirrors the J2 GPIO header geometry, and a 2.54 mm same-component own-net
+scenario that confirms own-net escape still works.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from kicad_tools.router.grid import RoutingGrid
+from kicad_tools.router.layers import Layer
+from kicad_tools.router.primitives import Pad
+from kicad_tools.router.rules import DesignRules
+
+
+@pytest.fixture
+def jlcpcb_rules() -> DesignRules:
+    """jlcpcb-tier1 design rules (chorus-test deployment that hit this bug)."""
+    return DesignRules(
+        trace_width=0.127,
+        trace_clearance=0.127,
+        grid_resolution=0.1,
+        min_trace_width=0.127,
+        fine_pitch_clearance=0.0635,
+        fine_pitch_threshold=0.65,
+    )
+
+
+def _make_thtpin_header_pair(
+    ref: str = "J2",
+    pitch: float = 2.54,
+    pad_size: float = 1.7,
+    drill: float = 1.0,
+    net_a: int = 1,
+    net_a_name: str = "I2S_DIN",
+    net_b: int = 2,
+    net_b_name: str = "GPIO19",
+    center_x: float = 0.0,
+    center_y: float = 0.0,
+) -> tuple[Pad, Pad]:
+    """Create two 2.54 mm THT pin-header pads mirroring chorus-test J2 geometry.
+
+    Two adjacent pins on the same physical header land on different nets.
+    Pad metal is round 1.7 mm with 1.0 mm drill, identical to the RPi GPIO
+    40-pin header footprint that triggered Issue #2961.
+    """
+    pad_a = Pad(
+        x=center_x - pitch / 2,
+        y=center_y,
+        width=pad_size,
+        height=pad_size,
+        net=net_a,
+        net_name=net_a_name,
+        layer=Layer.F_CU,
+        through_hole=True,
+        drill=drill,
+        ref=ref,
+        pin="1",
+    )
+    pad_b = Pad(
+        x=center_x + pitch / 2,
+        y=center_y,
+        width=pad_size,
+        height=pad_size,
+        net=net_b,
+        net_name=net_b_name,
+        layer=Layer.F_CU,
+        through_hole=True,
+        drill=drill,
+        ref=ref,
+        pin="2",
+    )
+    return pad_a, pad_b
+
+
+class TestCarveoutRespectsObstacle:
+    """Issue #2961: ``_relax_same_component_clearance`` must not clear
+    ``_blocked`` on cells where ``_is_obstacle == True``.
+    """
+
+    def test_isobstacle_cells_remain_blocked_after_carveout(self, jlcpcb_rules):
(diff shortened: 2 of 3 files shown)


rjwalters:
LGTM! Approved.

## Verification summary

**1. Early-continue placement (CONFIRMED):** The 16-LOC insertion at `grid.py:2044-2058` is exactly between the cell-net allowlist at line 2041 (`if cell_net != pad.net and cell_net != other_pad.net: continue`) and the world-coord lookup at line 2062 (`wx, wy = self.grid_to_world(...)`). Surgical, minimal, with a thorough 14-line comment explaining the C++ pathfinder gating rationale.

**2. Own-net escape preserved (VERIFIED in C++):** Read `pathfinder.cpp:60-112`. The control flow is:
- Line 81: `if (cell.blocked)` -- outer gate. When `blocked == False`, the cell is unconditionally passable.
- Line 104: `if (cell.is_obstacle || cell.net != net) return true;` -- when own-net routes (`cell.net == net`), the right-hand operand is False; combined with whatever `is_obstacle` is, the OR is dominated only by `is_obstacle`. But: the carve-out's own filter at `grid.py:2041` only operates on cells where `cell_net == pad.net OR cell_net == other_pad.net`. So when the pad owner routes, `cell.net == routing_net` means `cell.net != net` is False AND the cell is_obstacle-aware path returns False for own net via the carve-out leaving `_blocked = True`... 

Wait, let me re-check. With the fix, `_blocked` stays `True` for `is_obstacle` cells. Now when pad A's owner routes through a cell where `cell.net == pad.net == routing_net`: line 81 takes the `blocked == True` branch, line 104 evaluates `(is_obstacle || cell.net != routing_net)` = `(true || false)` = `true`, returns blocked. **That would block own-net.**

Re-reading: actually, the cells where the carve-out runs and `is_obstacle = True` are the *halo* cells painted by #2915/#2940. Those halo cells carry the **owner pad's** net (e.g. pad A's net) by the first-touch fix design. So when pad A's owner routes, `cell.net == pad_a.net`, and `cell.net != routing_net` is False, but `is_obstacle` is True -- so the OR returns True and pad A's own net IS blocked from its own halo.

But that's the pre-existing #2915 invariant: own-net routing CAN pass through its own pad metal (`_pad_blocked` is checked separately) but the halo around its own pad metal is treated as an obstacle for everyone. The carve-out's purpose (#2452) was the *cross-component* corridor between two same-component pads on different nets -- those overlap cells were the ones flipped to `_blocked = False`. The fix correctly preserves that for non-`is_obstacle` overlap cells while tightening obstacle-painted cells back to True.

The post-fix invariant in the updated `test_through_hole_pads_relaxation` test (point 3) uses `compute_expanded_blocked(radius=0, net=1)` and confirms own-net passability via the Python pathfinder mirror, which IS the correct semantic check.

**3. Updated existing test (CONFIRMED CORRECT):** `test_through_hole_pads_relaxation` now asserts `compute_expanded_blocked(radius=0, net=1)` finds passable cells in the corridor, not raw `_blocked == False`. This is the right post-#2915/#2940 invariant: the user-visible behavior (own-net pathfinder traversal) rather than the implementation detail (Python-side `_blocked` flag).

**4. Chorus-test verification:** The chorus-test board lives in the ml-audio-codecs repo, not this one, so the builder couldn't re-route within this PR's worktree. The geometric chain (2.54mm pin pitch > 0.65mm `fine_pitch_threshold` -> narrow-channel halo skipped at `grid.py:1746-1747` -> carve-out previously cleared `is_obstacle` cells -> foreign-net traces clipped neighbor pads) is well-documented in the issue body and matches the new `test_foreign_net_rejected_at_pad_pair` test that uses identical J2 geometry (1.7mm round THT pads at 2.54mm pitch, jlcpcb-tier1 rules). The 18->0 violation drop cannot be empirically verified here but the regression guards in `test_grid_same_component_carveout.py` enforce the geometric contract that would have produced the violations.

**5. Board-05 unexpected improvement (EMPIRICALLY CONFIRMED):**
- Without fix (HEAD~1 grid.py): `routed 7/35 (20%)`
- With fix (PR HEAD): `routed 9/35 (25%)`
- Net delta: +2 nets routed, +5 percentage points

Mechanism: board-05 (BLDC motor controller) has same-component pad pairs that triggered the carve-out's obstacle-clearing leak. Most likely candidates are the half-bridge MOSFET pairs (Q1-Q6) where source/drain pads on the same SOP-8 footprint sit on different power nets, or the motor connector J1 with adjacent THT pins. Pre-fix, foreign traces routed *through* the obstacle-painted halo cells around these pads, blocking other nets' candidate corridors via the cascading `clearance_pad_segment` retries. Post-fix, those cells stay blocked, freeing routing budget for the +2 nets. Note: still below the 26% floor by 1 net, but the trend is in the right direction.

**6. Pre-existing failures (EMPIRICALLY CONFIRMED):**
- `test_xtal2_failure_classified_as_trace_blocker`: fails on HEAD~1, fails on HEAD. Unrelated.
- `test_board_04_routing_regression::test_routes_all_nets_on_2l`: errored on HEAD (board fails to route). Pre-existing throughput regression.
- `test_board_05_routing_regression::test_2l_completion_meets_floor`: fails on HEAD~1 at 20%, fails on HEAD at 25%. The fix improves but doesn't cross the 26% floor.

These three failures are pre-existing on main and orthogonal to this PR's surgical scope.

## CI status

All 8 checks pass; merge state CLEAN.

## Code quality

- Single early-continue, minimal diff, well-documented inline comment with cross-references to #2452/#2915/#2940/#2961 and the C++ pathfinder line numbers
- 4 new tests covering (a) is_obstacle cells stay blocked after carve-out, (b) foreign-net rejection at J2 geometry, (c) own-net escape preserved, (d) third-party cells untouched
- 1 existing test updated with correct post-#2915/#2940 semantic (`compute_expanded_blocked` vs raw `_blocked`)
- No drive-by changes, no scope creep

Third "first-touch + halo respect" fix in a coherent family (#2915 pad metal, #2940 rect-aware halo, #2961 mutator respect). The fix prevents a downstream mutator from clearing the obstacle contract painted upstream.

Issue #2118, opened 2026-04-26
fix: unconnected_component checker produces systematic false positives


rjwalters:
## Problem Statement

The `unconnected_component` check in `sch validate` reports components as "Fully unconnected -- all pins are floating" when their pins DO have wires at the correct pin-tip coordinates. This was verified by:

1. Reading the embedded `lib_symbols` pin definitions (AT positions)
2. Applying symbol rotation to compute world-space pin-tip coordinates
3. Cross-referencing against `sch wires` output to confirm wire endpoints match

**Affected components in chorus-test-revA (all confirmed CONNECTED):**
- DAC sheet: C32 (C_Small, rot 90), C23 (C_Small, rot 90), R8 (R_Small, rot 90), R14 (R_Small, rot 90), R9 (R, rot 90), R10 (R, rot 90), C40 (C_Small, rot 0), LED3, LED4, C21
- Sync sheet: R2 (R, rot 90), R5 (R, rot 90)

**Pin position verification example (C32):**
- C32 center: (182.88, 97.79), rotation 90°
- C_Small pin 1 AT: (0, 2.54) in symbol coords → rotated 90° → world (180.34, 97.79)
- C_Small pin 2 AT: (0, -2.54) in symbol coords → rotated 90° → world (185.42, 97.79)
- Wire: (167.64, 97.79) → (180.34, 97.79) — hits pin 1 ✓
- Wire: (185.42, 97.79) → (207.01, 97.79) — hits pin 2 ✓
- **Yet validate reports: "Fully unconnected component: C32 -- all pins are floating"**

## Observed During

Schematic review and repair of chorus-test-revA using `/review-schematic` and `/repair-schematic`

## Impact

**Critical**: 12 out of 12 "unconnected_component" errors in the chorus-test-revA project are false positives. This renders the check unusable and causes review/repair workflows to chase non-existent issues. The `/review-schematic` skill classified these as CRITICAL findings, leading to unnecessary repair attempts.

## Root Cause Hypothesis

The checker likely computes pin world-space coordinates incorrectly when symbols are rotated. Possible issues:
1. Not applying rotation transform to pin AT coordinates
2. Confusing pin LENGTH with pin AT offset
3. Not accounting for KiCad's Y-axis flip between symbol and schematic coordinate systems

## Recommended Enhancement

1. Fix pin-to-wire coordinate matching to properly apply symbol rotation
[...]

Issue #889, opened 2026-01-17
pin_position() returns inverted Y coordinates, swapping pin positions


rjwalters:
## Summary

`Symbol.pin_position()` returns incorrect Y coordinates due to an erroneous Y-axis negation. This causes pins to be swapped vertically - requesting one pin returns the position of another pin that mirrors it across the symbol's horizontal centerline.

## Impact

This bug causes **KiCad to crash with a segfault** when opening schematics that have `no_connect` flags or wires placed at the "wrong" pin positions returned by this function. The crash occurs in KiCad's connection graph resolution (`CONNECTION_SUBGRAPH::ResolveDrivers`).

## Reproduction

Using the UCC27511ADBV gate driver symbol which has:
- `OUTH` pin at symbol-local position `(10.16, 2.54)`
- `OUTL` pin at symbol-local position `(10.16, -2.54)`

When the symbol is placed at `(320.04, 35.56)` with rotation 0:

```python
drv = sch.add_symbol(lib_id="Driver_FET:UCC27511ADBV", x=320.04, y=35.56, ...)
outh_pos = drv.pin_position("OUTH")  # Returns (330.2, 33.02) ❌
outl_pos = drv.pin_position("OUTL")  # Returns (330.2, 38.1) ❌
```

**Expected positions** (verified by examining working KiCad schematics):
- `OUTH`: `(330.2, 38.1)` — symbol_y + pin_y = 35.56 + 2.54
- `OUTL`: `(330.2, 33.02)` — symbol_y + pin_y = 35.56 + (-2.54)

**Actual positions** (what the code returns):
- `OUTH`: `(330.2, 33.02)` — symbol_y - pin_y = 35.56 - 2.54
- `OUTL`: `(330.2, 38.1)` — symbol_y - pin_y = 35.56 - (-2.54)

The pins are swapped because the Y coordinate is incorrectly negated.

## Root Cause

In `src/kicad_tools/schematic/models/symbol.py`, line 372:

```python
# Translate to symbol position (flip Y for schematic's Y-down system)
return (round(self.x + rx, 2), round(self.y - ry, 2))
```

The comment claims "symbol definitions use Y-up" but this appears to be incorrect. KiCad symbol definitions use the same Y-down coordinate system as schematics. The negation (`self.y - ry`) inverts the Y positions.

## Proposed Fix

Change line 372 from:
```python
return (round(self.x + rx, 2), round(self.y - ry, 2))
```

To:
```python
[...]

PR #2967, opened 2026-05-16
fix(schematic): pin_position() correctly negates library Y to schematic Y-down (closes #2959)

### LINKED ISSUE #2959: schematic: pin_position() uses library Y-up convention, schematic Y-down flips real pin positions
## Problem

`SymbolInstance.pin_position()` in `src/kicad_tools/schematic/models/symbol.py:525` returns pin coordinates by directly adding the library-defined offsets to the placed symbol position, without flipping the Y axis to convert from KiCad's library Y-up convention into the schematic Y-down convention. For any pin with a non-zero Y offset in the library, the computed absolute pin position is incorrect (off by 2*offset).

The misleading comment at line 587 (`# Note: Both KiCad schematics and symbol definitions use Y-down coordinates`) appears to encode the wrong assumption that caused the bug.

## Root cause / verification

- KiCad library symbol files (`.kicad_sym`) store pin offsets in **library-local Y-up** coordinates: in `Connector_Generic:Conn_01x02`, Pin 1 is at `(at -5.08 0 0)` and Pin 2 is at `(at -5.08 -2.54 0)`. The body rectangle is `(start -1.27 1.27) (end 1.27 -3.81)`, confirming pin 1 is at the top of the body, pin 2 below it (Y-up: smaller y = below).
- KiCad schematic files (`.kicad_sch`) place symbols and wires in **schematic Y-down** coordinates (origin top-left, +Y is down).
- `Pin.connection_point()` (`src/kicad_tools/schematic/models/pin.py:35`) returns the raw library `(x, y)` with no flip.
- `pin_position()` rotates and translates with `self.x + rx, self.y + ry` — no Y inversion. This consistently produces wrong Y for any pin where the library y != 0.

Reproduction (board 00, `Conn_01x02` placed at (49.53, 80.01)):
- `pin_position("2")` returns (44.45, 77.47) [80.01 + (-2.54)]
- KiCad ERC reports the actual rendered pin at (44.45, 82.55) [80.01 - (-2.54)]
- `add_wire(j1_pin2, ...)` therefore terminates at empty space; ERC fires `pin_not_connected` on J1.2/GND.

## Scope

`pin_position()` is called from every demo board's schematic generator (00, 01, 05 verified via grep; also indirectly via `all_pin_positions()`, `block-level` symbol wrappers in `schematic/blocks/`, and the `sch_pin_positions` / `sch_add_*` CLI commands). The bug has been latent because:

1. Pins with `y == 0` in the library are unaffected (most resistors, caps with vertical orientation, and many IC pins on horizontal sides).
2. Rotated symbols accidentally mask the bug for some orientations.
3. Earlier boards may have wired pin 1 only or used horizontal connectors where pin offsets are in X.

Boards with multi-pin connectors / non-zero-Y pin offsets are at risk: at minimum board 00 (J1 GND fails), board 01 (J1/J2 — needs re-check after fix), board 05 (J_POWER, J_MOTOR, J_HALL — many `pin_position` calls on connector pin 2+).

Fixing this changes the geometry of every wire that today terminates on a non-zero-Y pin. Some existing schematics may have wires "accidentally correct" if the design.py was hand-tuned around the bug; those will regress visually but become electrically correct.

## Proposed fix

Negate `conn_y` before applying rotation in `pin_position()`:

```python
conn_x, conn_y = pin.connection_point()
conn_y = -conn_y  # library Y-up -> schematic Y-down

rad = math.radians(self.rotation)
cos_r = math.cos(rad)
sin_r = math.sin(rad)
rx = conn_x * cos_r - conn_y * sin_r
ry = conn_x * sin_r + conn_y * cos_r

return (round(self.x + rx, 2), round(self.y + ry, 2))
```

Also: replace the misleading comment at line 587 with a correct one documenting library Y-up vs schematic Y-down. Audit `Pin.connection_point()` callers in `validate/sch_orphan_label.py`, `validate/sch_wire_stub.py`, `optim/*.py`, and `schematic/blocks/*` for the same implicit assumption — if any of them consume `(pin.x, pin.y)` directly in schematic space, they need the same flip (or, preferably, a single chokepoint helper).

## Acceptance criteria

1. `kct check` (ERC) on `boards/00-simple-led` reports zero `pin_not_connected` errors after `generate_design.py` regeneration with the fix applied. J1 pin 2 (GND) is wired correctly.
2. New unit test in `tests/test_schematic_models.py` (or new file) covers the Y-axis convention:
   - Symbol placed at (100, 100), rotation=0, pin at library (x=-5.08, y=-2.54) -> expected schematic position (94.92, 102.54).
   - Same pin with rotation=180 -> expected (105.08, 97.46).
   - Same pin with rotation=90 -> expected (102.54, 94.92).
   - Round-trip: parsing a generated schematic and reading the pin position back from the file yields the same coordinates `pin_position()` predicted (within 0.01 mm).
3. Existing test `test_symbol_instance_pin_position_rotated` (currently encodes the wrong Y-down/Y-down assumption) is updated to match the corrected convention; comment at line 587 is rewritten.
4. `kct check` regression: regenerate boards 00, 01, 05 schematics; ERC error count for `pin_not_connected` does not increase on 01 or 05, and decreases on 00. Note: visual layout of existing routes may shift; that is expected.
5. Audit pass: `git grep -n "connection_point\|pin\.x\|pin\.y"` reviewed and any other site that mixes library-space pin coordinates with schematic-space positions is either fixed or documented as library-space-only.

## Test plan

- [ ] Add unit test covering Y-flip at rotations 0, 90, 180, 270 with a non-zero-Y library pin.
- [ ] Add round-trip test: build a small `Schematic` programmatically, write it, parse it, confirm pin position from file matches `pin_position()` output.
- [ ] Regenerate `boards/00-simple-led` and run `kct check`; assert ERC has no `pin_not_connected` errors.
- [ ] Regenerate `boards/01-voltage-divider` and `boards/05-bldc-motor-controller`; compare ERC error counts pre/post fix; investigate any new errors.
- [ ] Run full `pytest tests/` and verify no regressions (expect updates to `test_symbol_instance_pin_position_rotated` and any other test currently asserting the buggy math).

## Code references

- `src/kicad_tools/schematic/models/symbol.py:525-598` — `pin_position()`.
- `src/kicad_tools/schematic/models/pin.py:35-44` — `Pin.connection_point()` (library-local).
- `tests/test_schematic_models.py:551-598` — existing tests that need to be updated for the corrected convention.
- `boards/00-simple-led/generate_design.py:77` — failing call site.

## Origin

Filed by board-00 scout. ERC observed `pin_not_connected` on J1 pin 2 because the wire from J1.GND terminates 5.08 mm away from where KiCad actually renders pin 2.

rjwalters:
## Summary

KiCad library symbols (``.kicad_sym``) store pin positions in **Y-UP**
coordinates (positive Y is above the symbol origin in the library drawing),
while KiCad schematics (``.kicad_sch``) use **Y-DOWN** screen coordinates.
``SymbolInstance.pin_position()`` was treating both spaces as Y-down,
so any pin with non-zero library Y ended up on the wrong side of the
placed symbol -- wires terminated in empty space and KiCad ERC reported
``pin_not_connected``.

This was the regression introduced by issue #889 (which itself misdiagnosed
the convention and removed the Y-flip).  Issue #2959 restores the flip
using the same approach that's already proven in
``kicad_tools.schema.library.get_pin_position``.

## Changes

- ``src/kicad_tools/schematic/models/symbol.py``: rotate the pin's
  library-local offset in Y-up coordinates, then negate Y to convert to
  schematic Y-down before translating to the placed symbol position.
  Replaces the misleading "Both KiCad schematics and symbol definitions
  use Y-down coordinates" comment with an accurate description of the
  convention.
- ``src/kicad_tools/schematic/models/pin.py``: document that
  ``Pin.connection_point()`` returns library Y-up coordinates and that
  callers consuming it in schematic space must apply the Y-flip.
- ``tests/test_schematic_models.py``:
  - Update ``test_symbol_instance_pin_position_rotated`` (encoded the
    bug at rotation=90 even though y=0 happens to mask the error).
  - Update both ``test_pin_position_vertical_ordering`` sites that
    asserted the wrong convention (the test from #889 and the
    UCC27511ADBV-style gate driver test that referenced #889).
  - **NEW** ``test_pin_position_y_up_convention_at_four_rotations``:
    exercises rotation 0/90/180/270 on a Conn_01x02-style pin at
    library Y=-2.54.
  - **NEW** ``test_pin_position_round_trip_with_real_library``: places
    a real ``Connector_Generic:Conn_01x02``, takes ``pin_position("2")``,
    cross-checks against ``LibrarySymbol.get_pin_position`` (the
    independent implementation), writes the schematic, re-parses, and
    confirms the wire endpoint still lands on the predicted pin.

| ERC clean on board 00 | PASS | ``uv run python boards/00-simple-led/generate_design.py`` -> ``ERC: PASS`` (the original ``pin_not_connected`` on J1.GND is fixed) |
| New 4-rotation unit tests pass | PASS | ``test_pin_position_y_up_convention_at_four_rotations`` covers 0/90/180/270 on a non-zero-Y pin |
| Round-trip parse test passes | PASS | ``test_pin_position_round_trip_with_real_library`` parses a written schematic and confirms the wire endpoint matches the predicted pin position |
| Boards 01 and 05 still produce valid schematics | PARTIAL/IMPROVED | Board 01 ERC PASS.  Board 05 ERC error count drops from 119 -> 100 with this fix (i.e. the fix improves board 05).  The remaining 100 errors on board 05 are a pre-existing label/wire alignment problem in ``schematic/blocks/motor.py`` and the board script that exists on main and is unrelated to ``pin_position()`` -- filed separately below. |
| Existing encoded-bug tests updated | PASS | All three sites (``test_symbol_instance_pin_position_rotated`` and the two ``test_pin_position_vertical_ordering`` tests) now assert the corrected convention with docstrings citing #2959. |
| ``pytest tests/test_schematic_models.py`` green | PASS | 205 passed (full file), plus 267 passed across the schema/blocks/validator/library tests that share this code path. |

## Audit of other ``connection_point`` / pin-coord consumers

Per the issue's hard limits I audited the other call sites that mix
library-space pin coordinates with schematic-space positions:

- ``src/kicad_tools/validate/sch_orphan_label.py`` and
  ``src/kicad_tools/validate/sch_wire_stub.py``: both go through
  ``LibrarySymbol.get_all_pin_positions``, which already does the
  Y-up -> Y-down conversion correctly (issue #2118).  No change needed.
- ``src/kicad_tools/optim/*``: ``grep`` shows no use of
  ``connection_point`` / ``pin_position`` / ``all_pin_positions``.
- ``src/kicad_tools/schematic/blocks/*``: all consumers use
  ``SymbolInstance.pin_position()`` (the fixed method) to compute
  schematic-space coordinates for wires, labels, and ports.  They
  all get fixed transparently by this change.
- ``src/kicad_tools/schematic/registry.py``'s cached ``Pin`` type is
  converted into ``models/pin.Pin`` and consumed via
  ``SymbolInstance.pin_position()`` -- already covered by the fix.

The only sites that consume ``Pin.connection_point()`` directly are the
fixed ``pin_position()`` and the unit test in ``test_schematic_helpers.py``,
which correctly tests that ``connection_point()`` returns the raw
library-local coordinate.

## Test Plan

- ``uv run python -m pytest tests/test_schematic_models.py`` -> 205 passed.
- ``uv run python -m pytest tests/test_schema.py tests/test_schematic_helpers.py tests/test_validate_sch_orphan_label.py tests/test_validate_sch_wire_stub.py tests/test_library_extends.py`` -> 267 passed.
- Targeted block + wiring + sch CLI tests -> 1406 passed, 2 pre-existing failures (``test_add_wire_zero_length_rejected``, ``test_existing_hierarchical_fixture_aggregation``) reproduce on main and are unrelated to this PR.
- ``uv run python boards/00-simple-led/generate_design.py`` -> ERC PASS.
- ``uv run python boards/01-voltage-divider/generate_design.py`` -> ERC PASS.
- Board 05 generated cleanly; ERC error count drops 119 -> 100 vs main (pre-existing block/wiring bug, file separately).

## Follow-up

The remaining ~100 ERC errors on board 05 come from a pre-existing
mismatch in ``schematic/blocks/motor.py``'s gate-drive resistor array:
labels land at every resistor's pin coordinate (correctly), but the
wires from the driver IC only reach the first resistor's X column.  I'll
file a separate issue for that board script -- it is **not** caused by
``pin_position()`` and the same error count is observed with or without
this PR (just slightly higher on main).

Closes #2959

### DIFF
--- a/src/kicad_tools/schematic/models/pin.py
+++ b/src/kicad_tools/schematic/models/pin.py
@@ -33,13 +33,25 @@ class Pin:
     pin_type: str = "passive"
 
     def connection_point(self) -> tuple[float, float]:
-        """Get the wire connection point.
+        """Get the wire connection point in symbol-local (library Y-up) coords.
 
-        The connection point is where wires attach to the pin.
-        In KiCad symbol definitions, this is the (x, y) position.
+        The connection point is where wires attach to the pin.  In KiCad
+        symbol definitions (``.kicad_sym``) this is the ``(x, y)`` position
+        stored in library Y-UP coordinates -- positive Y is above the symbol
+        origin in the library drawing.
+
+        IMPORTANT: KiCad schematics use Y-DOWN screen coordinates.  Callers
+        that need an absolute schematic position must negate Y after applying
+        any rotation/mirror; see
+        :meth:`kicad_tools.schematic.models.symbol.SymbolInstance.pin_position`
+        and :meth:`kicad_tools.schema.library.LibrarySymbol.get_pin_position`
+        for the correct conversion.  Using ``connection_point()`` directly in
+        schematic space produces vertically-flipped pin positions for any
+        pin with non-zero library Y (issue #2959).
 
         Returns:
-            (x, y) tuple of the wire connection point in symbol-local coordinates
+            (x, y) tuple of the wire connection point in library-local
+            (Y-up) coordinates.
         """
         return (self.x, self.y)
 
--- a/src/kicad_tools/schematic/models/symbol.py
+++ b/src/kicad_tools/schematic/models/symbol.py
@@ -580,21 +580,31 @@ class SymbolInstance:
                 suggestions=suggestions,
             )
 
-        # Get the wire connection point (end of pin) in symbol-local coordinates
+        # Get the wire connection point (end of pin) in symbol-local coordinates.
+        # KiCad library symbols store pin positions in Y-UP coordinates (math
+        # convention: +Y is above origin in the library's drawing).  KiCad
+        # schematics use Y-DOWN screen coordinates (origin top-left, +Y is below).
+        # We rotate in library Y-up coords, then negate Y to convert to schematic
+        # Y-down before translating to the placed symbol position.  This matches
+        # the proven implementation in ``kicad_tools.schema.library.get_pin_position``.
         conn_x, conn_y = pin.connection_point()
 
-        # Apply rotation transformation
-        # Note: Both KiCad schematics and symbol definitions use Y-down coordinates
+        # Apply rotation in library Y-up coordinates (standard CCW rotation matrix).
         rad = math.radians(self.rotation)
         cos_r = math.cos(rad)
         sin_r = math.sin(rad)
 
-        # Rotate connection point around origin
         rx = conn_x * cos_r - conn_y * sin_r
         ry = conn_x * sin_r + conn_y * cos_r
 
-        # Translate to symbol position
-        # Round to 2 decimal places for consistent wire matching
+        # Convert from library Y-up to schematic Y-down by negating the Y offset.
+        # Without this flip, pins with non-zero library Y end up on the wrong
+        # side of the symbol (issue #2959), causing wires to terminate in empty
+        # space and ERC to report pin_not_connected.
+        ry = -ry
+
+        # Translate to symbol position.  Round to 2 decimal places for
+        # consistent wire matching.
         return (round(self.x + rx, 2), round(self.y + ry, 2))
 
     def all_pin_positions(self) -> dict[str, tuple[float, float]]:
--- a/tests/test_schematic_models.py
+++ b/tests/test_schematic_models.py
@@ -565,21 +565,31 @@ class TestSymbolInstance:
         assert abs(pos[1] - 100.0) < 0.01
 
     def test_symbol_instance_pin_position_rotated(self, mock_symbol_def):
-        """Get pin position on rotated symbol."""
+        """Get pin position on rotated symbol.
+
+        Library Y-up vs schematic Y-down: rotation happens in library coords,
+        then Y is negated to convert to screen coords.  See issue #2959.
+        """
         inst = SymbolInstance(
             symbol_def=mock_symbol_def,
             x=100.0,
             y=100.0,
-            rotation=90,  # 90 degrees rotation
+            rotation=90,  # 90 degrees rotation (CCW in library coords)
             reference="R1",
             value="10k",
         )
 
         pos = inst.pin_position("1")
-        # At 90 degrees, (-2.54, 0) rotates to (0, -2.54)
-        # Both schematic and symbol use Y-down: (100+0, 100+(-2.54)) = (100, 97.46)
+        # Pin 1 library coords (Y-up): (-2.54, 0)
+        # Rotate 90 CCW in Y-up: (0, -2.54)
+        # Negate Y for schematic Y-down: (0, 2.54)
+        # Translated to (100, 100): (100, 102.54)
+        # (Because pin 1 has y=0 in library, the result is the same whether
+        # or not Y is negated -- this test passes on the buggy code too.
+        # The non-zero-Y rotation behaviour is exercised in
+        # ``test_pin_position_y_up_convention_at_four_rotations`` below.)
         assert abs(pos[0] - 100.0) < 0.01
-        assert abs(pos[1] - 97.46) < 0.01
+        assert abs(pos[1] - 102.54) < 0.01
 
     def test_symbol_instance_pin_position_by_number(self, mock_symbol_def):
         """Get pin position by pin number."""
@@ -629,10 +639,20 @@ class TestSymbolInstance:
         assert "2" in positions
 
     def test_pin_position_vertical_ordering(self):
-        """Verify pins maintain correct vertical ordering (issue #889).
+        """Verify pins maintain correct vertical ordering (issues #889 / #2959).
 
-        Pins with positive Y in symbol-local coords should have higher screen Y
-        than pins with negative Y, since both symbol defs and schematics use Y-down.
+        KiCad library symbols store pin positions in Y-UP coordinates: a pin
+        with positive library Y sits ABOVE the symbol origin in the library
+        drawing.  KiCad schematics use Y-DOWN screen coordinates (origin
+        top-left, +Y is below), so a library Y-up pin with positive Y must
+        map to a SMALLER schematic Y (higher on screen).
+
+        Issue #889 incorrectly concluded both spaces were Y-down and removed
+        the Y-flip.  Issue #2959 restored the Y-flip after observing that
+        connector pin 2 (library Y=-2.54) was rendered by KiCad below the
+        symbol origin (schematic Y = symbol Y + 2.54) while ``pin_position()``
+        was returning a coordinate above it (symbol Y - 2.54), causing wires
+        to terminate in empty space and ERC ``pin_not_connected`` errors.
         """
         # Create a symbol with pins at different Y positions
         pins = [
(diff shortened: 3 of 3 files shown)


rjwalters:
**Judge review: APPROVE**

## Verification

### Math correctness
Manually verified the Y-flip + rotation at all four cardinal rotations for `Conn_01x02` Pin 2 (library `(-5.08, -2.54)`, placed at `(100, 100)`):

| rot | rotate in Y-up | negate Y | translate |
|-----|----------------|----------|-----------|
| 0   | (-5.08, -2.54) | (-5.08, 2.54)  | (94.92, 102.54)  |
| 90  | (2.54, -5.08)  | (2.54, 5.08)   | (102.54, 105.08) |
| 180 | (5.08, 2.54)   | (5.08, -2.54)  | (105.08, 97.46)  |
| 270 | (-2.54, 5.08)  | (-2.54, -5.08) | (97.46, 94.92)   |

All four match the test asserts. Also confirmed the buggy (pre-PR) math produces wrong results at every rotation:

```
rot=0   buggy: (94.92, 97.46)  vs expected (94.92, 102.54)
rot=90  buggy: (102.54, 94.92) vs expected (102.54, 105.08)
rot=180 buggy: (105.08, 102.54) vs expected (105.08, 97.46)
rot=270 buggy: (97.46, 105.08) vs expected (97.46, 94.92)
```

So `test_pin_position_y_up_convention_at_four_rotations` is a genuine regression test that would fail on `main`.

### Reference implementation parity
`pin_position()` now matches `kicad_tools.schema.library.LibrarySymbol.get_pin_position` (`src/kicad_tools/schema/library.py:479-539`): rotate in library Y-up, negate Y, translate. The two implementations are now consistent — and the round-trip test cross-checks them on a real library symbol, which is the strongest single guard against future divergence.

### Audit completeness
Verified the PR's audit claims with `grep`:
- `Pin.connection_point()` direct consumers: only `pin_position()` itself plus two unit tests that explicitly test the library-local return value (`test_schematic_registry.py:32`, `test_schematic_helpers.py:415`). Correct as-is.
- `validate/sch_orphan_label.py:184` and `validate/sch_wire_stub.py:227` go through `LibrarySymbol.get_all_pin_positions` (already-correct reference impl). No change needed.
- `schematic/blocks/*` (analog, power/inputs, etc.) all consume `SymbolInstance.pin_position()` and get fixed transparently.
- `optim/*` and `cli/*` show no use of the affected APIs.

### Tests
- `pytest tests/test_schematic_models.py` -> 205 passed locally.
- The four targeted tests pass: `test_pin_position_y_up_convention_at_four_rotations`, `test_pin_position_round_trip_with_real_library`, `test_symbol_instance_pin_position_rotated`, `test_pin_position_vertical_ordering` (both sites).

### Test-update legitimacy
The three pre-existing tests that were "updated" did encode the bug:
- `test_symbol_instance_pin_position_rotated` only worked because `pin.y == 0` (and the new docstring acknowledges this honestly).
- Both `test_pin_position_vertical_ordering` sites asserted that library-positive-Y pins should render at LARGER screen Y, which is the inverse of what KiCad does. Updated to assert smaller screen Y (above origin), with docstrings linking #889 -> #2959.

### CI / merge state
- All 8 CI checks pass (Test, Type Check, Lint & Format, C++ Build, DRC, Routed PCB, Round-trip Smoke, Diff-Pair + Match-Group Regressions).
- mergeStateStatus: `CLEAN`.

### Follow-up
Issue #2968 is filed for residual board-05 ERC errors (gate-drive resistor array label/wire mismatch in `schematic/blocks/motor.py`), correctly scoped as a separate board-script bug unrelated to `pin_position()`. AC verification table in the PR body is honest about board 05 being PARTIAL/IMPROVED (119 -> 100).

## Non-blocking notes
- `tests/test_schematic_models.py:790-792` has a multi-line call that local `ruff format` would collapse to a single line (line-length nit). CI lint uses `continue-on-error: true` so this doesn't block, but worth noting for future cleanup.

## Quality assessment
This is a clean, surgical fix backed by:
1. A rigorous round-trip test that exercises the real `Conn_01x02` library symbol and parses the written schematic back.
2. Four-rotation coverage on a non-zero-Y pin (where the bug actually manifests).
3. Honest acknowledgement that the rotated unit test happens to mask the bug at y=0.
4. Comprehensive audit of all other pin-coordinate consumers in the repo.
5. Documentation updates on both `Pin.connection_point()` and `pin_position()` that explicitly call out the Y-up vs Y-down convention and reference the reference implementation.
6. ERC improvement: board 00 PASS, board 01 PASS, board 05 119->100 with proper follow-up filed.

Approved.

PR #2945, opened 2026-05-16
fix(router): factor via_clearance helper + diagnose in-pad rescue overlaps (closes #2944)

rjwalters:
## Summary

Factor `stitch_cmd.check_via_clearance` into a shared geometry helper (`router/via_clearance.point_clear_of_copper`) and wire it into `EscapeRouter._can_place_via` per the curator's #2944 spec.  Add a diagnostic-only clearance check to `_try_in_pad_escape` that warns when the in-pad rescue produces a DRC-violating via on fine-pitch QFP/LQFP (the board-04 OSC_OUT cluster).

**Important caveat**: The curator's #2944 hypothesis assumed that rejecting the bad escape via would let a ring-fallback path advance to a safer slot.  That fallback exists for BGA escape routing but NOT for QFP/LQFP in-pad rescue -- there is no alternate path for a plane-sandwiched signal pin on a 0.5mm-pitch LQFP-48 at jlcpcb-tier1's clearance rules with stock via geometry.  Rejecting the rescue leaves the pin entirely unrouted and dropped board-04's completion from ~80% to ~33% even after 6-layer escalation.  The PR therefore softens the in-pad rescue check from hard reject to warning, preserves connectivity, and emits a structured diagnostic that names the offending pin.

## Changes

- **NEW `src/kicad_tools/router/via_clearance.py`**: Shared `point_clear_of_copper` predicate, with `TrackSegmentLike` / `FilledPolygonLike` protocols so both `stitch_cmd` and `router/escape` data classes plug in without cross-module type dependency.
- **`src/kicad_tools/cli/stitch_cmd.py::check_via_clearance`**: Thin wrapper around the shared helper.  Backward-compatible for all 18+ existing callers and tests.
- **`src/kicad_tools/router/escape.py::EscapeRouter._can_place_via`**: New optional `foreign_pads` / `foreign_tracks` / `clearance` / `via_diameter` parameters.  When supplied, the shared world-coord predicate runs in addition to the historical grid-cell check.  Backward-compatible (no callers updated -- ready for future use by `staggered_via_fanout` or new escape paths).
- **`src/kicad_tools/router/escape.py::EscapeRouter._via_clears_other_pads`**: Rect-aware (not circular-approximation) clearance check used by `_try_in_pad_escape`; correctly handles oblong fine-pitch SMD pads where the inscribed-circle approximation would over-reject.
[...]

Issue #2944, opened 2026-05-16
router/stitch: board-04 6 residual clearance_segment_via + clearance_pad_via after stitching (follow-up from #2834 closure)


rjwalters:
## Problem

After PR #2931 (#2908) and PR #2941 (#2933), board-04's `clearance_pad_segment` cluster is eliminated. Fresh re-route shows board-04 at 11/12 nets, 6 DRC errors (well within the current allowlist of 12).

The 6 residual errors are **NOT the same class** that #2834's umbrella tracked. They are stitching-related:
- 4 × `clearance_segment_via`
- 2 × `clearance_pad_via`
- (0-2 × `clearance_segment_segment`)

The defect class: stitch vias placed by `kct stitch` are too close to existing signal traces. Different mechanism from the same-component plane-pad clearance (#2908) and the foreign-net through-pad (#2915/#2928/#2942) families.

## Hypothesis

PR #2860 introduced a stitch-via halo to space stitching vias from each other. The halo may not consider signal traces — so when a corner-IC stitch via is placed close to an interior pad's escape trace, the via lands within trace-clearance distance.

Two fix shapes worth considering:

1. **Extend the stitch-via halo to also consider existing signal traces.** Before placing a stitch via, check distance to all nearby traces; reject placements within trace clearance.
2. **Shrink corner-IC stitch via diameter** when adjacent escape traces are close. The stitch via doesn't need to be full-size near a corner — a smaller via still ties planes adequately.

## Reproduction

```bash
git checkout origin/main
uv run kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb \
  --manufacturer jlcpcb-tier1 --auto-fix --auto-layers --auto-mfr-tier \
  --timeout 600 --placement-feedback
uv run kct check boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb \
  --mfr jlcpcb-tier1 --errors-only
```

Expected: 6 errors of class `clearance_segment_via` / `clearance_pad_via` / `clearance_segment_segment`, all near a corner IC where stitch vias overlap escape clearance.

## Acceptance criteria

- [ ] Board 04 routes to ≤2 errors at jlcpcb-tier1 (target: 0).
- [ ] Boards 00-03, 05-07 fleet-status parity (no regression).
- [ ] `.github/routed-drc-tolerance.yml` board-04 entry reduced from 12 toward the new residual count.
[...]

Issue #2918, opened 2026-05-15
pcb-gen: board-03 crystal placed east of MCU but XTAL pins are on west edge (XTAL1/XTAL2 channel-blocked)


rjwalters:
## Summary

Board 03 (usb-joystick) `generate_pcb.py` places crystal Y1 at `BOARD_ORIGIN_X + 55` (east/right of the MCU at `BOARD_ORIGIN_X + 40`), but the MCU's XTAL pins (2 = XTAL1, 3 = XTAL2) are on the **west/left edge** of the TQFP-32 package (pad offset = -4.5 mm from U1 centre). This forces 17–22 mm routes across or around the MCU body, which the autorouter cannot complete.

## Observed failure (routing audit 2026-05-15)

After `kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb --manufacturer jlcpcb --auto-layers --auto-mfr-tier --timeout 600`:

- `XTAL1` and `XTAL2` both **incomplete** (blocked_path)
- XTAL load caps C5/C6 also placed east-side, adjacent to crystal but 17–22 mm from MCU XTAL pins
- Net-status: `C5.1 @ (52.08, 32.00)` and `Y1.1 @ (52.56, 28.00)` unconnected (absolute ≈ 152, 128 — well east of MCU at 140, 132)

## MCU XTAL pin geometry (TQFP-32)

| Pin | Net | Pad position (relative to U1 centre) | Absolute (U1 at 140,132) |
|-----|-----|--------------------------------------|--------------------------|
| 2 | XTAL1 | (−4.500, −2.000) mm | (135.5, 130.0) |
| 3 | XTAL2 | (−4.500, −1.200) mm | (135.5, 130.8) |

Crystal Y1 at (155, 128): distance to MCU pin 2 = **17.2 mm**, pin 3 = **22.1 mm**.

## Fix applied (PR candidate)

`generate_pcb.py::generate_crystal()`: changed placement from  
`x = BOARD_ORIGIN_X + 55  # Right of MCU`  
to  
`x = BOARD_ORIGIN_X + 22  # Left of MCU (XTAL pins 2/3 are on U1's west edge)`  
`y = BOARD_ORIGIN_Y + 30  # Aligned with U1 XTAL pad row`

`generate_xtal_load_caps()`: updated `xtal_cx` reference from `BOARD_ORIGIN_X + 55` to `BOARD_ORIGIN_X + 22`.

[...]

PR #2979, opened 2026-05-16
fix(schematic): GateDriveResistorArray emits stub wires for label connectivity (closes #2968)

### LINKED ISSUE #2968: board-05: gate-drive resistor array labels misaligned with driver-side wires (ERC errors)
## Problem

When ``boards/05-bldc-motor-controller/design.py`` is regenerated on main, ``kct check`` / KiCad ERC reports ~100-119 ``pin_not_connected`` errors clustered around the gate-drive slew-rate resistor array (R20-R22) and related label/wire endpoints.

The user-visible warnings during generation are:

```
UserWarning: Label 'GATE_DRV_BH' at (320.04, 123.19) is not on any wire. Nearest wire point is (309.88, 123.19), 10.16mm away. This will cause ERC errors.
UserWarning: Label 'GATE_BH'     at (320.04, 115.57) is not on any wire. Nearest wire point is (309.88, 115.57), 10.16mm away. This will cause ERC errors.
UserWarning: Label 'GATE_DRV_CH' at (330.20, 123.19) is not on any wire. Nearest wire point is (309.88, 123.19), 20.32mm away.
UserWarning: Label 'GATE_CH'     at (330.20, 115.57) is not on any wire. Nearest wire point is (309.88, 115.57), 20.32mm away.
```

The label X coordinates land on each resistor's pin column (R20=309.88, R21=320.04, R22=330.20 with 10.16 mm spacing), but the wires from the DRV8301 outputs only reach the first column (309.88).

## Reproduction

```bash
uv run python boards/05-bldc-motor-controller/design.py /tmp/board-05-erc-repro
# look for "Found N ERC errors" near the bottom; expect ~100-119
```

## Where to fix

Either:
- ``src/kicad_tools/schematic/blocks/motor.py::create_gate_drive_resistor_array`` should emit horizontal wires from the input port (driver IC side) to every resistor pin 1, not just the first one, OR
- ``boards/05-bldc-motor-controller/design.py`` should explicitly wire each ``GATE_DRV_*H`` net from the driver output pad to the corresponding resistor pin 1 instead of relying on the label-only connection.

The first option is cleaner and benefits any future user of ``create_gate_drive_resistor_array``.  The block currently emits labels at each resistor input but no horizontal "rail" wire across the array.

## NOT a duplicate of #2959

This is **NOT** caused by ``pin_position()`` (issue #2959).  The same warnings and error count appear on ``main`` before #2959's fix.  After #2959 lands, the error count actually drops from 119 to 100 -- the Y-axis fix improves matters but cannot fix this X-axis label/wire gap.

## Acceptance criteria

1. ``boards/05-bldc-motor-controller/design.py`` ERC: zero ``pin_not_connected`` errors for ``GATE_DRV_AH/BH/CH`` and ``GATE_AH/BH/CH``.
2. ``uv run python boards/05-bldc-motor-controller/design.py`` runs without the four ``UserWarning: Label '...' is not on any wire`` warnings cited above.
3. If the fix lives in ``motor.py``, add a regression test in ``tests/test_blocks_gate_drive_resistor_array.py`` that asserts every emitted label has a wire endpoint at its coordinate.

## Origin

Discovered while implementing #2959 (``pin_position()`` Y-flip).  Filed separately per the #2959 hard-limit that builders must not silently rewrite wire positions on out-of-scope board scripts.

rjwalters:
## Summary

`GateDriveResistorArray.__init__` previously called `sch.add_label` at each resistor pin but emitted zero wires. KiCad's label-only connectivity requires the label coordinate to sit on a wire endpoint or segment; without a wire stub, labels in columns beyond the first one float — generating four `UserWarning: Label '...' is not on any wire` messages on board 05 (`GATE_DRV_BH`, `GATE_BH`, `GATE_DRV_CH`, `GATE_CH`) and a cascade of ERC `isolated_pin_label` / `pin_not_connected` violations.

This PR emits short horizontal stub wires (one schematic-grid = 2.54 mm) from each pin outward — pin 1 to the left, pin 2 to the right — and places the label on the stub endpoint. The pattern mirrors `create_3phase_inverter` at `motor.py:350`. External ports (`IN_i`, `OUT_i`, `IN_<suffix>`, `OUT_<suffix>`) remain at the pin positions so existing callers continue to wire to the pin, not the stub.

## Changes

- `src/kicad_tools/schematic/blocks/motor.py` — add stub wire before each label in `GateDriveResistorArray.__init__`.
- `tests/test_blocks_gate_drive_resistor_array.py`
  - Update `test_input_output_net_aliases` to assert labels sit at `pin ± STUB`, not at the pin itself (the previous assertion was the curator-flagged bad invariant).
  - New `test_labels_anchored_to_wire_endpoints` regression test: every `add_label(text, x, y)` must have a matching `add_wire` endpoint at `(x, y)`.

## Verification

- `tests/test_blocks_gate_drive_resistor_array.py`: 14/14 pass (was 13).
- Board 05 generation no longer emits the four named `UserWarning: Label '...' is not on any wire` messages for `GATE_DRV_BH`, `GATE_BH`, `GATE_DRV_CH`, `GATE_CH` (AC #2).
- All six gate-drive labels (`GATE_DRV_AH/BH/CH`, `GATE_AH/BH/CH`) are now properly anchored to stub wires in the generated schematic (verified by inspecting `.kicad_sch` content at the R20/R21/R22 columns).

## AC reconciliation

AC #1 (ERC errors drop from 100 to ≤4) is **not fully met by this PR alone**. The curator's analysis attributed all 100 errors to the gate-drive resistor floating-label root cause, predicting they would self-resolve once `R20/R21/R22` were electrically continuous. Investigation during implementation showed this is not the case:

- The DRV8308 (U3) symbol's gate-output and PWM-input pins are not labeled or wired anywhere in the schematic — `GateDriverBlock.__init__` only places the IC and bypass caps and never emits labels at the driver-IC pins. KiCad reports `U3 pins 1–41` as `pin_not_connected` regardless of the resistor-stub fix.
- The MOSFET gates `Q1/Q3/Q5.G` (and `Q2/Q4/Q6.G`) are not labeled or wired anywhere in the schematic either — `HalfBridge` / `ThreePhaseInverter` expose `GATE_HS`/`GATE_LS` ports at pin positions but never call `sch.add_label` for them, and `boards/05-bldc-motor-controller/design.py` does not label them either.
- Net `GATE_AH` therefore exists only at `R20.pin2`'s stub-wire endpoint (and now `GATE_BH` at `R21.pin2`, `GATE_CH` at `R22.pin2`) — a single-pin net, which ERC reports as `isolated_pin_label`. The same applies to `GATE_DRV_*H` at the driver-side stub endpoints.

The post-fix ERC count is still ~100. The targeted root cause (floating gate-drive labels + curator-named `UserWarning` messages) is fully resolved, but full electrical continuity through to Q1/Q3/Q5.G and U3.{33,36,39} additionally requires labeling the driver-IC outputs (`GateDriverBlock`) and the MOSFET gates (`HalfBridge` / `ThreePhaseInverter`). Per the issue's HARD LIMIT ("DO NOT modify `boards/05-bldc-motor-controller/design.py`") and "small, surgical fix" scope hint, that broader work is deferred to a follow-up issue.

ACs #2, #3, #4 are fully met. AC #5 (boards 00-04, 06, 07 fleet-status parity) is preserved: the change is gated on `input_nets is not None` / `output_nets is not None` and only affects schematics that pass those parameters (board 05 is the sole consumer in-tree).

## Test plan

- [x] `tests/test_blocks_gate_drive_resistor_array.py` — 14/14 pass.
- [x] `tests/test_schematic_blocks.py` + related — all pass.
- [x] Board 05 generation succeeds and the four named `UserWarning` messages are gone.
- [x] Inspect generated `.kicad_sch`: labels at `(307.34, 115.57)`, `(312.42, 123.19)`, `(317.5, 115.57)`, `(322.58, 123.19)`, `(327.66, 115.57)`, `(332.74, 123.19)`; matching stub wires present at each label coordinate.

### DIFF
--- a/src/kicad_tools/schematic/blocks/motor.py
+++ b/src/kicad_tools/schematic/blocks/motor.py
@@ -1243,18 +1243,33 @@ class GateDriveResistorArray(CircuitBlock):
             self.ports[in_port_name] = in_pos
             self.ports[out_port_name] = out_pos
 
-            # Optional net labels — emit a label at each pin and add an
-            # alias port keyed by the net suffix (the trailing token after
-            # the last '_').
+            # Optional net labels — emit short horizontal stub wires from
+            # each pin so the label has a wire to anchor to. KiCad's
+            # label-only connectivity requires the label coordinate to lie
+            # on a wire endpoint or segment; without these stubs, labels
+            # in columns beyond the first one float (see issue #2968).
+            # The stub length matches one schematic grid (2.54 mm).
+            #
+            # Pattern mirrors ``create_3phase_inverter`` (~line 350):
+            #     sch.add_wire(pos, (label_x, pos[1]))
+            #     sch.add_label(name, label_x, pos[1], rotation=0)
+            #
+            # Alias ports are keyed by the net suffix (the trailing token
+            # after the last ``_``).
+            STUB = 2.54
             if input_nets is not None:
                 in_net = input_nets[i]
-                sch.add_label(in_net, in_pos[0], in_pos[1], rotation=0)
+                in_label_x = in_pos[0] - STUB
+                sch.add_wire(in_pos, (in_label_x, in_pos[1]))
+                sch.add_label(in_net, in_label_x, in_pos[1], rotation=0)
                 alias = f"IN_{_net_suffix(in_net)}"
                 self.ports[alias] = in_pos
 
             if output_nets is not None:
                 out_net = output_nets[i]
-                sch.add_label(out_net, out_pos[0], out_pos[1], rotation=0)
+                out_label_x = out_pos[0] + STUB
+                sch.add_wire(out_pos, (out_label_x, out_pos[1]))
+                sch.add_label(out_net, out_label_x, out_pos[1], rotation=0)
                 alias = f"OUT_{_net_suffix(out_net)}"
                 self.ports[alias] = out_pos
 
--- a/tests/test_blocks_gate_drive_resistor_array.py
+++ b/tests/test_blocks_gate_drive_resistor_array.py
@@ -114,7 +114,11 @@ class TestGateDriveResistorArray:
             assert in_pos[0] < out_pos[0]
 
     def test_input_output_net_aliases(self, mock_schematic):
-        """When ``input_nets``/``output_nets`` provided, aliases and labels appear."""
+        """When ``input_nets``/``output_nets`` provided, aliases and labels appear.
+
+        Labels are offset by one schematic grid (2.54 mm) from the pin so
+        a stub wire can anchor them — see #2968.
+        """
         block = create_gate_drive_resistor_array(
             mock_schematic,
             x=100,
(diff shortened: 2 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

### Verification

| Check | Result |
|---|---|
| Pattern mirrors `create_3phase_inverter:350` | YES — exact `add_wire(pin, (label_x, pin.y))` then `add_label(name, label_x, pin.y, rotation=0)` sequence |
| External ports anchor at pin positions | YES — `self.ports[in_port_name]/[out_port_name]` and aliases all map to `in_pos`/`out_pos` (the pin), not the stub endpoint. Callers unchanged. |
| 4 targeted UserWarnings disappear | YES — board 05 stderr grep for `Label '(GATE_DRV_BH|GATE_BH|GATE_DRV_CH|GATE_CH)' .* is not on any wire` returns 0 matches |
| Fleet parity (boards 00-04, 06, 07) | TRIVIAL — board 05 is the sole in-tree consumer of `create_gate_drive_resistor_array`; the new wire emission is gated on `input_nets is not None` / `output_nets is not None`, so callers without those args see byte-for-byte identical output |
| New regression test `test_labels_anchored_to_wire_endpoints` fails on main | YES — confirmed FAILED on main (without fix), PASSES on PR head |
| Updated `test_input_output_net_aliases` invariant | Correctly inverted: was asserting bad invariant `(lx,ly)==in_pos`; now asserts `(lx,ly)==(in_pos[0]-STUB, in_pos[1])` |
| Targeted test file | 14/14 pass (was 13) |
| CI checks | 8/8 GREEN (Lint, Type, Test, C++ Build, Round-trip Smoke, Routed PCB DRC, Diff-Pair, Match-Group) |
| Merge state | CLEAN, MERGEABLE |

### On the partial AC1 disclosure

The PR body's "AC reconciliation" section is exemplary. The curator hypothesized all 100 ERC errors would self-resolve once the resistor-side stubs landed; in practice only the 4 named `UserWarning` messages and the directly-cited `isolated_pin_label` / `pin_not_connected` cluster around `R20-R22` are root-caused here. The remaining ~96 errors flow from two distinct upstream gaps:

1. `GateDriverBlock.__init__` places the DRV8308 IC and bypass caps but never labels its gate-output / PWM-input pins
2. `HalfBridge` / `ThreePhaseInverter` expose `GATE_HS` / `GATE_LS` ports at pin positions but never call `sch.add_label` on the MOSFET gates `Q1/Q3/Q5.G` (or `Q2/Q4/Q6.G`)

Fixing those is the right thing to do, but it's outside the issue's "small, surgical fix" scope hint and would have required touching the HARD-LIMITED `boards/05-bldc-motor-controller/design.py` to wire the new labels through. The builder correctly stopped at the scope boundary instead of silently expanding the PR. Third instance of this honest-disclosure pattern this session (cf #2918/#2924, #2944/#2945, #2945/#2950) — keep it up.

### Minor observation (non-blocking)

The new stub wires emit two additional `UserWarning: Wire start endpoint (309.88, 115.57/123.19) lands on existing wire segment from (309.88, 83.82) to (309.88, 279.4)` warnings (from `motor.py:1263` and `motor.py:1271`). These are informational — KiCad just wants a junction marker on the intentional intersection — and exactly the same pattern is already accepted in this file at `motor.py:350` for `create_3phase_inverter`. Not a regression and not a blocker; if anything, a future cleanup could call `sch.add_junction` at the intersection.

### Follow-up filed?

Strongly recommend a follow-up issue to label the DRV8308 outputs in `GateDriverBlock` and the MOSFET gates in `HalfBridge` / `ThreePhaseInverter`. That's the next layer of the same root cause class (block-level label emission ownership) and will close out the remaining ~96 ERC errors on board 05 without callers needing to manage labels themselves.

Approving.

PR #2984, opened 2026-05-17
fix(route): strip stale route blocks before insert to dedupe vias (closes #2976)

### LINKED ISSUE #2976: Router: 14 same-net dimension_drill_clearance errors on board 05's 2L route
## Problem

Routing board 05 on 2L with the proven flag recipe (``kct route --no-auto-layers --layers 2 --manufacturer jlcpcb --differential-pairs --seed 42``) produces 30 DRC errors, of which **14 are same-net ``dimension_drill_clearance``** -- two vias on the same net placed within 0.127mm drill-edge-to-drill-edge.

Sample (from ``kct check --verbose``):

```
[X] dimension_drill_clearance
    Drill-to-drill clearance -0.300mm < minimum 0.127mm
    Items: HALL_A, HALL_A
    Nets: HALL_A
```

Negative clearance means the drills *overlap* -- the router stitched two vias on the same logical net on top of each other.

## Why it matters

Even with the same-net relaxation that's standard in KiCad's DRC, drill-to-drill physical clearance is a manufacturing constraint, not a clearance-rule one. Overlapping same-net drills will break the drill file.

## Possible causes

1. The negotiated router places a via during a rip-up/retry pass without checking that the previous via on the same net was placed too close.
2. Trace-optimization (``compress_staircase``, ``minimize_vias``) merges segments around an existing via and leaves the old via in place.
3. The C++ pathfinder's same-net via clustering check fires only on cross-net conflicts; same-net is allowed to stack.

## Acceptance criteria

- ``kct route`` on board 05 with the recipe above produces 0 ``dimension_drill_clearance`` errors with negative clearance.
- The fix doesn't regress the 11+ ``Nets routed`` count on the same recipe.
- A unit test in ``tests/router/`` covers the "two same-net vias must keep drill-edge clearance >= grid resolution" invariant.

## Scope

Router-side, isolated to ``src/kicad_tools/router/``. Most likely 50-150 LOC in the via-conflict checker plus a test.

## Context

Parent: #2906 (board 05 routing-quality epic, builder measurement).
Related: #2972 PR (just-merged ``_can_place_via`` cell.net respect -- shaped the failure mode but does not eliminate same-net overlap).

rjwalters:
## Summary

Eliminates the 14 same-net ``dimension_drill_clearance`` errors observed
on board 05's 2L route by stripping stale ``(segment ...)`` and
``(via ...)`` blocks from the PCB content before inserting fresh route
s-expressions.

Closes #2976.

## Root cause

The negative-clearance violations were not a router algorithm bug.
``_stage_input_for_auto_pour`` aliases ``pcb_path`` to ``output_path``
so ``auto_pour`` can write zones in place without mutating the user's
input file.  After the aliasing, every ``_write_routed_pcb`` call
re-reads the previous write's output instead of the original input, and
the new route s-expression is *appended* on top of stale segments and
vias from the prior write.  Two checkpoint writes plus one final write
produce up to three copies of every via in the output PCB, each with a
distinct UUID but identical (net, x, y, drill).  The validator correctly
flags these as drill-overlap on the same net.

Validated empirically by patching ``_write_routed_pcb`` to log the
``pcb_path.read_text()`` size and via count: the second invocation read
710 763 bytes with 45 vias from the supposed "input" file (originally
61 685 bytes with 0 vias), confirming the prior checkpoint write was
being re-ingested.

## Fix

``_insert_sexp_before_closing`` now calls a new ``_strip_route_blocks``
helper before performing the insert.  The strip walks the PCB s-expression
character-by-character tracking depth and quoted strings, and removes any
top-level form whose first token is ``segment`` or ``via``.  Footprints
(which contain ``(pad ...)`` blocks at deeper nesting), zones, and other
PCB structure are preserved untouched.

## Test plan

- [x] ``kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb --no-auto-layers --layers 2 --manufacturer jlcpcb --differential-pairs --seed 42``: 0 ``dimension_drill_clearance`` errors (was 14+); 57 vias in output, 0 duplicates.
- [x] Boards 01 (voltage divider) and 02 (charlieplex) full route: 0 duplicates, 0 ``dimension_drill_clearance`` errors.
- [x] ``tests/router/test_strip_route_blocks.py`` -- 9 new unit tests covering:
  - Strip removes only top-level segments/vias (not footprint pads or zones).
  - Quoted parens in layer names are handled correctly.
  - Two and three consecutive inserts of the same route_sexp yield one via (idempotency).
  - Foreign-net vias in the new sexp are both preserved (strip doesn't touch insert payload).
  - ``_strip_route_blocks`` on a clean PCB is a no-op.
- [x] ``tests/test_route_checkpoint.py`` -- all 17 tests still pass (no checkpoint-write regression).
- [x] ``tests/router/`` + route-cmd test suite -- 450 pass, only the 2 pre-existing ``TestRouteWriteSitesPattern`` failures remain (both fail on ``main`` independently of this change; they assert outdated source-level counts).
- [x] ``ruff check`` clean on the new file and the edit.

## Scope

- 1 new helper ``_strip_route_blocks`` (~85 LOC).
- 5-line change inside ``_insert_sexp_before_closing`` to call the helper and update the docstring.
- 1 new test file (~226 LOC) under ``tests/router/``.

Total production code change: ~110 LOC, within the 150 LOC cap.

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -348,12 +348,121 @@ def _emit_single_pad_net_warning(
     )
 
 
+def _strip_route_blocks(pcb_content: str) -> tuple[str, int, int]:
+    """Strip top-level ``(segment ...)`` and ``(via ...)`` blocks from PCB content.
+
+    Issue #2976: When ``_stage_input_for_auto_pour`` aliases ``pcb_path`` to
+    ``output_path`` (so the user's input file isn't mutated by ``auto_pour``),
+    subsequent ``_write_routed_pcb`` calls re-read the **previous write's**
+    output rather than the original input.  Each write then *appends* the
+    current route s-expression on top of stale segments/vias from the prior
+    write, doubling (or worse) the via population in the output file.
+
+    The duplicated vias trigger ``dimension_drill_clearance`` errors --
+    -0.300mm clearance == two coincident drills on the same net -- because
+    they are literally the same physical via emitted twice with different
+    UUIDs.
+
+    The fix is to remove any top-level routed-element blocks before
+    inserting fresh ones: the in-memory router state is the source of
+    truth, and the new ``route_sexp`` already contains every segment and
+    via that should appear in the output.  Footprints, pads, zones, and
+    other PCB structure are preserved because we only strip blocks whose
+    first token is ``segment`` or ``via``.
+
+    Args:
+        pcb_content: Original PCB file content.
+
+    Returns:
+        Tuple of ``(stripped_content, segments_removed, vias_removed)``.
+        Counts are returned so callers can log a diagnostic when a stale
+        write is observed.
+    """
+    # Walk the content counting depth, identifying top-level forms whose
+    # first token is "segment" or "via" and excising them.  Footprints
+    # contain their own (pad ...) blocks, never (segment ...) or top-level
+    # (via ...), so stripping at depth=1 is safe.
+    out: list[str] = []
+    i = 0
+    n = len(pcb_content)
+    depth = 0
+    in_string = False
+    prev_char = ""
+    segments_removed = 0
+    vias_removed = 0
+    while i < n:
+        ch = pcb_content[i]
+        if ch == '"' and prev_char != "\\":
+            in_string = not in_string
+            out.append(ch)
+            prev_char = ch
+            i += 1
+            continue
+        if not in_string and ch == "(":
+            # Peek at the token following the paren.
+            j = i + 1
+            while j < n and pcb_content[j].isspace():
+                j += 1
+            token_start = j
+            while j < n and not pcb_content[j].isspace() and pcb_content[j] != "(" and pcb_content[j] != ")":
+                j += 1
+            token = pcb_content[token_start:j]
+            if depth == 1 and token in ("segment", "via"):
+                # Skip the entire form: find matching ")".
+                form_depth = 1
+                k = j
+                form_in_string = False
+                form_prev = ""
+                while k < n and form_depth > 0:
+                    c = pcb_content[k]
+                    if c == '"' and form_prev != "\\":
+                        form_in_string = not form_in_string
+                    elif not form_in_string:
+                        if c == "(":
+                            form_depth += 1
+                        elif c == ")":
+                            form_depth -= 1
+                    form_prev = c
+                    k += 1
+                if token == "segment":
+                    segments_removed += 1
+                else:
+                    vias_removed += 1
+                # Skip whitespace that was preceding this form (trailing
+                # newline/tab from the previous emission) to keep the
+                # resulting file tidy.
+                while out and out[-1] in (" ", "\t"):
+                    out.pop()
+                # Also drop a single trailing newline so we collapse the
+                # blank line the form previously occupied.
+                if out and out[-1] == "\n":
+                    out.pop()
+                i = k
+                prev_char = ")"
+                continue
+            depth += 1
+        elif not in_string and ch == ")":
+            depth -= 1
+        out.append(ch)
+        prev_char = ch
+        i += 1
+    return "".join(out), segments_removed, vias_removed
+
+
 def _insert_sexp_before_closing(pcb_content: str, sexp_fragments: str) -> str:
     """Insert S-expression fragments before the final closing parenthesis of a PCB file.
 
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

### Root-cause verification — confirmed

Traced the aliasing path end-to-end and the bug class is real:

- `_stage_input_for_auto_pour` (`src/kicad_tools/cli/route_cmd.py:1785-1788`) returns `output_path` when input != output, so the caller reassigns `pcb_path = output_path`. Confirmed at four call sites (2239, 2937, 3976, 5678).
- `_write_routed_pcb` (`route_cmd.py:566`) reads `pcb_path.read_text()` on **every** invocation. After the aliasing, that read picks up the previous write rather than the original input.
- `_insert_sexp_before_closing` then inserts the new route s-expression before the final `)`, appending on top of the prior write's segments/vias.
- The checkpoint callback at `_make_checkpoint_callback` (line 706) builds `route_sexp` from the in-memory `best_routes` snapshot — fully self-contained, so stripping is safe (the in-memory state is the source of truth, exactly as the docstring claims).

Two checkpoint writes + one final write = up to 3 copies of every via with distinct UUIDs but identical (net, x, y, drill). DRC flags drill-overlap on same net. The claim is dramatic but accurate.

### `_strip_route_blocks` correctness

The depth/quote-aware walk is sound:

- Top-level forms are at `depth == 1` (inside the `(kicad_pcb ...)` wrapper). The guard `if depth == 1 and token in ("segment", "via")` correctly targets exactly those forms.
- Footprint pads (`(pad ...)`), zones, properties, etc. are all at depth >= 2 and untouched. KiCad PCB schema never nests `(via)` or `(segment)` inside other forms, so depth-1-only is safe.
- Quoted-string handling tracks `in_string` with backslash-escape awareness; the new `test_handles_quoted_parens` exercises this with `"Some (weird) Layer"`.

### Timeline correlation

The aliasing landed in PR #2555 (commit d5c2d41e, 2026-05-08, "auto-pour preserves INPUT.kicad_pcb when output path differs (closes #2548)"). Board-05 same-net dimension_drill_clearance reports (#2976) followed afterward. Consistent with the proposed causal chain.

### Test coverage

All 9 new unit tests in `tests/router/test_strip_route_blocks.py` exercise the right axes:

- Strip removes only top-level segments/vias (`test_strips_top_level_segment_and_via`).
- Footprint/pad/zone preservation (`test_footprints_and_zones_preserved_after_insert`).
- Quoted parens in layer names (`test_handles_quoted_parens`).
- Idempotency on 2 and 3 repeat writes — the direct reproducer for #2976 (`test_no_via_accumulation_on_repeat_writes`, `test_third_write_still_one_via`).
- Foreign-net vias both survive when in the insert payload (`test_two_foreign_net_vias_both_present`).
- No-op on a clean PCB (`test_leaves_clean_pcb_untouched`).
- Empty route_sexp still strips stale routes (`test_empty_route_sexp_strips_stale_content`).
- Stale UUIDs are removed and new UUIDs survive (`test_new_sexp_present_after_insert`).

Locally:
- `pytest tests/router/test_strip_route_blocks.py -v` -> 9 passed.
- `pytest tests/test_route_checkpoint.py -v` -> 17 passed (no checkpoint regression).
- `ruff check` -> clean on new lines.
- `mypy` -> 29 pre-existing errors in `route_cmd.py`, none in the 348-471 range added by this PR.

### CI

All 8 checks pass: Test, Type Check, Lint & Format, DRC, Diff-Pair Routing Regression, Match-Group Routing Regression, C++ Build, kicad-cli Round-trip Smoke. mergeStateStatus = CLEAN.

### Recommendation — broader audit worth filing

The bug class is "stale checkpoint state from output-aliased read-back" and could affect any pipeline that re-reads its own output. Worth a follow-up issue to audit other read-then-write helpers in route_cmd / runner for the same pattern (e.g., zone re-pour, escape routing, partial saves). The fix here is correctly localized to the insertion path, but the underlying gun (aliasing + per-call read_text) remains loaded.

### Scope

- `_strip_route_blocks` (~85 LOC) + 5-line call site change = ~110 LOC production.
- 1 new test file (~263 LOC).
- Within the 150-LOC production cap.

Approving. Removing `loom:review-requested`, adding `loom:pr`.

🤖 Judge review by Claude Opus 4.7 (1M context)

PR #3007, opened 2026-05-18
fix(board-05): MCU NC + PWR_FLAG + C18/C19 (closes #3004)

rjwalters:
## Summary

Drops board-05's KiCad ERC error count from 17 to 1 by addressing the three categories called out in issue #3004 (PR-B follow-up to the #2994 cascade): MCU no-connect markers, power-symbol wire stubs, and the orphan-cap (C19) GND-rail extension. A fourth gap (C6 LDO output cap on the wrong side of the +3.3V rail) surfaced during analysis and is fixed in the same PR with a 1-line rail-extent change.

## Changes

- `boards/05-bldc-motor-controller/design.py`:
  - Add `sch.add_no_connect()` markers for the 8 unused STM32G431K8 GPIO pins (PA3–PA5, PA11–PA15, PB3–PB5 → LQFP-32 pins 8, 9, 10, 21, 22, 25, 27, 28). Closes 8 `pin_not_connected` errors.
  - Wire each `#PWR` power-input symbol (`power:+24V`, `power:+5V`, `power:+3V3`, `power:GND`) down to its rail's left endpoint so the symbol pin meets a real wire endpoint AND so the symbol's global net unifies with the rail's labelled net. Closes 4 `pin_not_connected` errors on the power symbols.
  - Extend the +3.3V rail westward (`x_start` from `X_LDO+25=165` to `X_LDO+7=147`) so the LDO output cap C6 at x=160.02 has a rail wire to meet, plus a junction at the original `x_start` so the +3V3-symbol vertical wire still connects via a valid T-connection. Closes the C6 `pin_not_connected` error.
  - Add a short GND-rail extension wire `(299.72, 279.4) → (309.88, 279.4)` so C19's pin-2 vertical wire endpoint lands on a real GND wire endpoint. Closes the C19 `pin_not_connected` error.
  - Document why an explicit `PWR_FLAG` on VMOTOR is *not* added: a pre-existing label-placement bug in the buck-block (the U1.FB `+5V` label at x=95.25 lands on a C2 bulk-cap vertical wire that also reaches VMOTOR) bridges VMOTOR and +5V into a single electrical net, so VMOTOR is already driven by LM2596.OUT (Output type). Adding a `PWR_FLAG` here triggers a `pin_to_pin` Output↔Power-output conflict against U1.OUT, which is strictly worse than the residual `power_pin_not_driven` on U1.VIN.

- `tests/test_board_05_erc_regression.py` (new): pins the post-#3004 ERC state with 4 assertions:
[...]

PR #3014, opened 2026-05-18
fix(buck): divert FB pin to direct wire when stub would collide with foreign wire

### LINKED ISSUE #3011: Buck-block +5V label-on-stub at U1.FB lands on C2 bulk-cap wire, bridging VMOTOR↔+5V
## Problem

In the buck-converter block (`src/kicad_tools/schematic/blocks/power/regulators.py`, `BuckConverter.__init__`, lines 619-645), the `pin_nets` mechanism places net labels on a short 2.54 mm stub drawn outward from the pin. When the caller passes `"FB": "+5V"` (as board-05 does for the fixed-output LM2596-5.0 variant — `boards/05-bldc-motor-controller/design.py:246`), the label lands at the stub endpoint at world coordinate **(95.25, 97.79)** on board-05.

That coordinate happens to fall *exactly on the C2 bulk-cap vertical wire* drawn by `sch.wire_decoupling_cap(c_bulk2, RAIL_VMOTOR, RAIL_GND)` at `design.py:203`. Verified in `boards/05-bldc-motor-controller/output/bldc_controller.kicad_sch`:

- C2 sits at `(95.25, 114.3)` (line 5123). Its top-pin wire is `(95.25, 115.57) → (95.25, 25.4)` (line 8307) — i.e. a long vertical wire at `x=95.25` that connects C2 to the VMOTOR rail at `y=25.4`.
- The U1.FB stub wire is `(92.71, 97.79) → (95.25, 97.79)` (line 8387), and the `+5V` label sits at `(95.25, 97.79)` (line 10273).

Because KiCad treats a label coordinate that lies on any wire segment as electrically welding the label's net into that wire's net, **VMOTOR and +5V become a single electrical net**. This:

1. Silently masks `power_pin_not_driven` on U1.VIN (the bridge means VMOTOR is "driven" by U1.OUT, which is an Output-type pin — see PR #3007 builder note).
2. Blocks the textbook `PWR_FLAG` recipe for VMOTOR: adding it triggers `pin_to_pin` Output↔Power-output conflict against U1.OUT because of the bridge.
3. Is a latent timebomb — any board with a different C2 x-position would not bridge (different bug), and any board that does coincidentally collide will short its input rail to its output rail at netlist time without a single warning.

PR #3007 explicitly avoided adding `PWR_FLAG` on VMOTOR and documented this exact trade-off as out-of-scope.

## Root cause

The stub-direction logic at `regulators.py:635-638` only looks at the pin's side relative to the symbol center (`pin_pos[0] < x` → stub left, else stub right). It has no notion of whether the resulting stub endpoint lands on an unrelated wire elsewhere on the sheet, and the label placement helper `sch.add_label` does not call any equivalent of `warn_on_collision=True` against existing wires.

For the LM2596S-5 symbol, the FB pin is at relative `(12.7, 2.54)` from the symbol center (`bldc_controller.kicad_sch:1210`). With the symbol at `(80.01, 100.33)` and the 2.54 mm stub, the label ends up at `(80.01 + 12.7 + 2.54, 100.33 + 2.54) = (95.25, 97.79)` — directly under C2's top pin (C2 is also at `x=95.25` because `design.py` placed C2 at `X_POWER_IN+70 = 25+70 = 95.25`).

Even more fundamentally, **the `+5V` label belongs on the actual +5V net (post-inductor / output-cap side of the buck), not on a fixed-output sense pin at the input/control side of the symbol**. The FB pin on the LM2596S-5 is the internal feedback-sense input; tying it to `+5V` via a label-on-stub is electrically correct in principle (fixed-output variants want FB === VOUT) but it conflates two distinct goals:

- declaring the regulator's FB pin as being on the +5V net (ERC purpose), and
- providing a label-anchor for the +5V net (connectivity purpose).

When the second purpose is fulfilled by an *existing* wire that's already electrically `+5V` (e.g. the output-cap node), the first purpose can be satisfied by joining FB to that same net via a real wire rather than via two label-on-stubs that happen to share a net name.

## Proposed fix

Two complementary changes; pick whichever the implementer prefers, or do both:

### Option A (preferred — block-internal, no caller change): wire FB directly to VOUT for fixed-output variants

When a `BuckConverter` caller passes `pin_nets={"FB": "<vout_net>"}` AND the same net name is already published on the `VOUT` port (i.e. the output-cap node is electrically the same net), draw a real wire from the FB pin to the output-cap node *inside the block* instead of emitting a label-on-stub at FB. This:

- removes the dependency on label-on-wire collisions (FB is now joined to VOUT by a real wire, which is what a fixed-output buck schematic looks like in any reference design),
- still satisfies ERC because FB now lies on a wire that carries the `+5V` net label at the output-cap end,
- leaves the existing `pin_nets` API unchanged for the non-FB pins.

Sketch (replace the FB branch inside the `if pin_nets is not None` loop at `regulators.py:629-645`):

```python
if pin_nets is not None:
    STUB = 2.54
    for pin_key, net_name in pin_nets.items():
        pin_pos = self.regulator.pin_position(pin_key)

        # Special-case FB on a fixed-output variant: when the caller asks
        # us to tie FB to the same net that's on VOUT, route a real wire
        # from FB to the output-cap node instead of emitting a stub label
        # that risks colliding with an unrelated wire (see issue #3011).
        if pin_key in ("FB", "4") and net_name == pin_nets.get("VOUT", None):
            # fall through to wire-only join below
            pass

        # Heuristic: if FB resolves to a fixed-output sense net, prefer a
        # direct wire to the published VOUT port over a label-on-stub.
        if pin_key in ("FB", "4") and hasattr(self, "_vout_node"):
            # L-shaped wire FB -> (vout_x, fb_y) -> vout_node
            vx, _ = self._vout_node
            sch.add_wire(pin_pos, (vx, pin_pos[1]), warn_on_collision=False)
            sch.add_wire((vx, pin_pos[1]), self._vout_node, warn_on_collision=False)
            sch.add_junction(vx, self._vout_node[1])
            if net_name not in self.ports:
                self.ports[net_name] = pin_pos
            continue

        # Default: stub-and-label as today.
        if pin_pos[0] < x:
            label_x = pin_pos[0] - STUB
        else:
            label_x = pin_pos[0] + STUB
        sch.add_wire(pin_pos, (label_x, pin_pos[1]), warn_on_collision=False)
        sch.add_label(net_name, label_x, pin_pos[1], rotation=0)
        if net_name not in self.ports:
            self.ports[net_name] = pin_pos
```

The label `+5V` already lives on the output-cap end of the block (placed by board-05's rail logic and by `connect_to_rails`), so once FB is wired to VOUT the FB pin sees the +5V net via real copper rather than via a fragile coordinate collision.

### Option B (defensive — generic, helps any future block): collision check in stub-placement

Add a `warn_on_collision`-style check to the `add_label`/`add_wire` calls in the `pin_nets` loop (`regulators.py:639-640`, mirrored at `regulators.py:228` for the LDO block and at `motor.py:238-242`, `interface/debug.py:221`): before placing the stub + label, scan existing wires on the sheet and raise (or log a structured warning) if the stub endpoint coordinate lies on a wire that does not already carry the target net. This catches the bug class generally, not just at FB.

A minimal version: have `BuckConverter.__init__` check `sch.wires` (or equivalent) after placing the stub and assert no foreign-net wire passes through `(label_x, pin_pos[1])`. This is the **regression-test instrument** even if Option A is the chosen fix.

### Recommendation

Ship Option A as the fix. Add Option B as the unit-test instrument and as a `warn_on_collision=True` default on the internal `add_wire` call so the next block author gets a loud failure if they reintroduce the pattern.

## Acceptance criteria

- After the fix, instantiating `BuckConverter` with `pin_nets={"FB": "+5V"}` on a synthetic schematic where a foreign wire passes through what would have been the FB-stub endpoint does NOT bridge the FB net to the foreign wire's net. Verified by net-tracing after schematic emission.
- The `+5V` label is anchored to a wire endpoint that is **electrically the +5V net by construction** (e.g. the output-cap-1 node, post-inductor), not by accident of stub-coordinate collision.
- Board-05 schematic regenerates with `python boards/05-bldc-motor-controller/design.py` and:
  - The two pins U1.VIN (VMOTOR) and U1.OUT/L1.1 (switch) → C4.1 (+5V) end up on **distinct** net names. Today they are bridged into one net.
  - Adding `sch.add_global_label("PWR_FLAG", VMOTOR_rail_coord)` (or the equivalent `pwr_flag` recipe) on VMOTOR no longer triggers `pin_to_pin` Output↔Power-output on U1.OUT.
- Existing tests pass:
  - `tests/test_blocks_buck_converter.py` (the `pin_nets` shape tests)
  - `tests/test_blocks_dual_supply_cascade.py`
  - `tests/test_board_05_erc_regression.py` (PR #3007's pin)
  - `tests/test_board_05_drv8308_pin_nets.py`
- ERC count on board-05 does **not regress** above the PR-#3007 baseline of 1 error, and ideally drops to 0 once `PWR_FLAG` can be added on VMOTOR cleanly (separate follow-up if needed).

## Test plan

Add `tests/test_blocks_buck_converter_fb_bridge.py` (new file, ~50 LOC) with a unit test that reproduces the exact board-05 collision geometry:

1. Build a fresh `Schematic`.
2. Place a `BuckConverter` with the LM2596S-5 symbol at the same position as board-05 (`x=80.01, y=100.33`) and `pin_nets={"VIN": "VMOTOR", "GND": "GND", "FB": "+5V"}`.
3. Before/after the block, add a foreign wire whose path passes through what would have been the FB-stub endpoint `(95.25, 97.79)`. The cleanest reproduction is to add `sch.add_wire((95.25, 25.4), (95.25, 200.0))` — a vertical wire labelled `VMOTOR` at one end. (Match board-05's C2 wire geometry.)
4. Save the schematic to a temp dir, then re-parse the resulting `.kicad_sch` (or use the in-memory net-tracer if one exists) and confirm:
   - The net at U1.VIN is `VMOTOR`.
   - The net at C-out.1 (the buck output cap) is `+5V`.
   - `assert net(U1.VIN) != net(C_OUT.1)` — the input and output nets MUST be distinct.
5. Repeat without the foreign wire to confirm the fix doesn't regress the non-collision case.

Also add a regression assertion in `tests/test_board_05_erc_regression.py` that the U1.VIN net and the C4.1 net are distinct in the generated `bldc_controller.kicad_sch`.

## Scope

- ~30-40 LOC change in `src/kicad_tools/schematic/blocks/power/regulators.py` (the FB-to-VOUT short-circuit branch in `BuckConverter.__init__`).
- ~50 LOC new test file `tests/test_blocks_buck_converter_fb_bridge.py`.
- Optional: ~10 LOC follow-up in `boards/05-bldc-motor-controller/design.py` to add `PWR_FLAG` on VMOTOR once the bridge is gone (can be a follow-up issue if size is tight).
- Total: ~80-100 LOC, within the requested 50-100 LOC band.

## Relevant code references

- `src/kicad_tools/schematic/blocks/power/regulators.py:334-645` — `BuckConverter` class and `pin_nets` stub-placement loop.
- `src/kicad_tools/schematic/blocks/power/regulators.py:619-645` — exact lines that emit the FB stub + label.
- `src/kicad_tools/schematic/blocks/power/regulators.py:585-617` — VOUT node computation (`c_out_pos`, `self._vout_node`) — the wire-target for Option A.
- `boards/05-bldc-motor-controller/design.py:166-203` — C2 placement and `wire_decoupling_cap` call that creates the vertical wire at `x=95.25`.
- `boards/05-bldc-motor-controller/design.py:228-256` — `create_dual_supply_cascade` call with `buck_pin_nets={"FB": "+5V"}`.
- `boards/05-bldc-motor-controller/output/bldc_controller.kicad_sch:5122-5165` — C2 footprint at `(95.25, 114.3)`.
- `boards/05-bldc-motor-controller/output/bldc_controller.kicad_sch:8307,8315,8387` — the three wires whose intersection forms the bridge.
- `boards/05-bldc-motor-controller/output/bldc_controller.kicad_sch:10272` — the offending `+5V` label at `(95.25, 97.79)`.
- `tests/test_blocks_buck_converter.py:115-130` — existing `pin_nets` test pattern to mirror.
- PR #3007 description — documents the bridge as the reason `PWR_FLAG` on VMOTOR was skipped.

## Out of scope (not in this fix)

- The same `pin_nets` stub mechanism in `LDOBlock` (regulators.py:228), `GateDriverBlock` (motor.py:238-242), and `DebugConnector` (interface/debug.py:221) has the **same class of bug** — any of them can collide a stub-endpoint with a foreign wire on a different board. Worth a follow-up sweep, but a single buck fix is the cited timebomb. File a generic "stub-endpoint collision detection" issue if Option B is deferred.
- The `power:+5V` / `power:+24V` / `power:+3V3` symbol-name vs rail-label-name mismatch (PR #3007 out-of-scope item) is separate.

rjwalters:
## Summary

Fixes a silent net-bridging bug in ``BuckConverter``: when a caller passes ``pin_nets={\"FB\": \"<vout_net>\"}`` (the recipe for a fixed-output buck variant like LM2596S-5), the previous code placed a label-on-stub at the FB pin endpoint. On board-05 that endpoint coordinate (95.25, 97.79) happened to land exactly on the C2 bulk-cap vertical wire, fusing VMOTOR and +5V into one net at netlist time without warning.

This PR makes ``BuckConverter`` detect that collision and divert FB to a real L-shaped wire to the output-cap node (``self._vout_node``) instead. This matches a textbook fixed-output buck reference design and removes the dependency on label-on-wire coordinate luck.

## Changes

- ``src/kicad_tools/schematic/blocks/power/regulators.py``:
  - Add an FB-pin-specific branch in the ``pin_nets`` loop: when the would-be stub endpoint lands on an existing wire, route FB -> (vout_x, fb_y) -> ``_vout_node`` with a junction at the output-cap node. The ``+5V`` label is still published externally by ``connect_to_rails``, so FB lands on the +5V net via real copper.
  - Add ``_fb_stub_would_collide`` static helper. Defensive against mock ``Schematic`` objects used in unit tests (returns ``False`` if ``sch.wires`` is missing or not a real list, or if ``_find_wire_collisions_for_point`` is unavailable). This keeps the existing 9-test ``tests/test_blocks_buck_converter.py`` suite green: those tests run against a clean mock schematic, so no collision is detected and the original stub-and-label path still fires.
- ``tests/test_blocks_buck_converter_fb_bridge.py`` (new, ~150 LOC):
  - ``test_no_label_on_pre_existing_vmotor_wire``: reproduces the exact board-05 collision geometry (C2 wire at x=95.25 from y=25.4 past y=97.79, buck at (80.01, 100.33) with ``pin_nets={\"VIN\":\"VMOTOR\",\"GND\":\"GND\",\"FB\":\"+5V\"}``) and asserts no label lands at the FB-stub endpoint coordinate.
  - ``test_no_collision_case_still_emits_label``: without the foreign wire, FB still emits a single ``+5V`` label-on-stub (back-compat path stays alive).
  - ``test_fb_pin_remains_on_vout_net``: confirms the new L-shaped wire from FB -> corner -> ``_vout_node`` exists.
- ``boards/05-bldc-motor-controller/output/bldc_controller.kicad_sch`` and the unrouted + routed PCBs: regenerated. The FB stub at (92.71, 97.79)-(95.25, 97.79) is replaced by FB -> (129.54, 97.79) -> (129.54, 111.76) and the offending ``+5V`` label at (95.25, 97.79) is gone.

| New test fails on ``main`` and passes on this branch | Yes | Confirmed by stashing the fix and re-running the new test file: 2 of 3 fail on main, all 3 pass with the fix |
| Existing ``tests/test_blocks_buck_converter.py`` (9 tests) still pass | Yes | ``uv run pytest tests/test_blocks_buck_converter.py`` -> 9 passed |
| FB net != VMOTOR net after fix (no bridge) | Yes | The ``+5V`` label at (95.25, 97.79) is gone from the regenerated schematic; FB is wired via real copper to the output-cap node |
| Board-05 ERC count stays <=4 (per PR #3007 budget) | Yes | Regenerated schematic ERC = 2 errors (both pre-existing ``power_pin_not_driven``), <=4 |
| Board-05 ERC regression tests pass | Yes | ``uv run pytest tests/test_board_05_erc_regression.py`` -> 4 passed |
| Public BuckConverter kwarg API unchanged | Yes | Only the internal ``pin_nets`` rendering branches on FB; no kwarg signature changed |
| Scope ~80-100 LOC | Yes (slightly over) | regulators.py: +58 LOC; new test file: 150 LOC. The test file is larger than the curator's ~50 LOC estimate because it includes three scenarios (collision case, no-collision back-compat, and L-shape wire structure assertion) rather than the single net-trace check |

## Test Plan

- ``uv run pytest tests/test_blocks_buck_converter.py tests/test_blocks_buck_converter_fb_bridge.py tests/test_blocks_dual_supply_cascade.py tests/test_board_05_drv8308_pin_nets.py tests/test_board_05_erc_regression.py`` -> 32 passed
- ``uv run pytest tests/test_schematic_blocks.py`` -> 332 passed
- ``uv run python boards/05-bldc-motor-controller/design.py`` regenerates the board; ERC drops the silent ``+5V``-label-on-VMOTOR bridge

## Out of scope (filed as follow-ups)

- Generic stub-endpoint collision detection for the other ``pin_nets`` call sites (LDOBlock at regulators.py:228, GateDriverBlock at motor.py:238-242, DebugConnector at interface/debug.py:221) — see the curator's defensive Option B in the issue body. Will file a separate issue.

Closes #3011


rjwalters:
## Judge Review: APPROVED

Empirically verified the fix against the board-05 collision geometry. Strong, well-scoped surgical fix for #3011.

### Code review (regulators.py, +58 LOC)

- FB-pin branch (lines 658-672) is correctly gated on `pin_key in ("FB", "4")` — VIN, VOUT, GND, EN, SW are unaffected. Other pins still take the original stub-and-label path.
- The L-shaped wire uses `self._vout_node` (the canonical output-cap pin-1 coordinate set at line 617). Not an arbitrary spot.
- `_fb_stub_would_collide` (lines 682-703) is properly mock-aware: `getattr` for both `wires` and `_find_wire_collisions_for_point`, `isinstance` check, `callable` check, and a defensive `try/except Exception` around the finder call. Returns `False` on any guard miss, preserving the original label-on-stub fallback for unit tests that pass mock `Schematic` objects.
- The fallback path (lines 674-680) is preserved verbatim from the original `pin_nets` rendering: stub + label + alias port. Non-board-05 callers (including the 9 existing `tests/test_blocks_buck_converter.py` tests) get unchanged behavior.
- Public `BuckConverter.__init__` signature is unchanged.

### Test review (tests/test_blocks_buck_converter_fb_bridge.py, +165 LOC)

- Test 1 (`test_no_label_on_pre_existing_vmotor_wire`): reproduces the exact board-05 geometry — C2 vertical wire at (95.25, 25.4) → (95.25, 115.57), buck at (80.01, 100.33) with `{"FB": "+5V", "VIN": "VMOTOR", "GND": "GND"}`. Asserts no label lands at the FB-stub endpoint (95.25, 97.79). This is a precise coordinate-level regression test against #3011.
- Test 2 (`test_no_collision_case_still_emits_label`): verifies fallback — without the foreign wire, FB still emits exactly one `+5V` label. Back-compat preserved.
- Test 3 (`test_fb_pin_remains_on_vout_net`): asserts both segments of the L-shape (FB → corner, corner → `_vout_node`) exist in `sch.wires`.

### Empirical board-05 re-verification

- Ran `kicad-cli sch erc` on regenerated `bldc_controller.kicad_sch`:
  - **2 errors**, both `power_pin_not_driven` (pre-existing per PR #3007 budget; well under ≤4 budget).
  - 38 warnings, all pre-existing (`multiple_net_names` for the deliberate +24V↔VMOTOR + SWD pin-sharing, `label_multiple_wires` on rail T-junctions, `unconnected_wire_endpoint` on docking points).
  - **No `+5V`↔`VMOTOR` `multiple_net_names` warning** — the silent bridge is gone.
- Schematic grep confirms:
  - The offending `(95.25 97.79)` coordinate is absent from the file (no label or symbol there).
  - The new L-shaped wires are present: `(92.71, 97.79) → (129.54, 97.79)` and `(129.54, 97.79) → (129.54, 111.76)`, with a junction at (129.54, 111.76) joining the output-cap node.
- Test runs (in worktree):
  - `tests/test_blocks_buck_converter.py` + `test_blocks_buck_converter_fb_bridge.py` + `test_blocks_dual_supply_cascade.py` + `test_board_05_drv8308_pin_nets.py` + `test_board_05_erc_regression.py`: **32 passed**.
  - `tests/test_schematic_blocks.py`: **332 passed**.

### Scope check

- Diff against `origin/main`: only `regulators.py`, `test_blocks_buck_converter_fb_bridge.py` (new), and the three regenerated board-05 output files. No edits to `LDOBlock`, `GateDriverBlock`, `DebugConnector`, or `design.py` (the generic-fix follow-up belongs to #3015).
- Public API unchanged.

### CI status

- Lint, Type Check, Test, C++ Build, kicad-cli Round-trip Smoke, Routed PCB DRC Check, Diff-Pair Routing Regression: **all SUCCESS**.
- Match-Group Routing Regression (board-07): **CANCELLED** due to `The job has exceeded the maximum execution time of 10m0s` — unrelated infrastructure timeout on board-07 (this PR doesn't touch board-07 or routing). Pre-existing flake.
- `mergeStateStatus`: UNSTABLE solely because of the cancelled board-07 timeout. Safe to merge.

### Bonus signal

- Routed PCB DRC annotation: `bldc_controller_routed.kicad_pcb` violation count dropped from 53 to 9 (slack=44). Removing the VMOTOR↔+5V bridge eliminated cascading clearance issues. Worth tightening the allowlist in a follow-up.

LGTM. Approving.

PR #3022, opened 2026-05-18
fix(router): CoupledPathfinder intra_pair_clearance floor + 2-pad polarity-swap (#3012)

### LINKED ISSUE #3012: router: CoupledPathfinder ignores intra_pair_clearance, produces overlapping diff-pair traces
## Summary

``DiffPairRouter.route_differential_pair_coupled`` builds the
``CoupledPathfinder`` with ``target_spacing_cells`` derived from
``pair.rules.spacing`` (typ. 0.15-0.2 mm) and never consults
``NetClassRouting.effective_intra_pair_clearance()``.  On boards whose
per-pair intra clearance exceeds the default
``DifferentialPairRules.spacing``, the search lays both centerlines
too close together — producing routes whose **edges overlap**
(observed -0.150 mm on board 07 = 100% overlap with 0.15 mm trace
width) and triggering thousands of ``diffpair_clearance_intra`` DRC
violations.

## Discovery context (PR #3005 judge follow-up)

PR #3005 hardens the inline serpentine shim (``length_critical``
gating, clearance-aware bulge, partner clearance check).  That shim
hardening is correct and lands as-is, but PR #3005's empirical claim
of "0 ``diffpair_clearance_intra`` with ``--differential-pairs``" on
board 07 did NOT reproduce.  Re-routing board 07 with
``--differential-pairs --net-class-map ...`` against the d25f9782
HEAD produces ~459 ``diffpair_clearance_intra`` violations (post
trace-optimizer) / ~2830 without the optimizer, with overlaps up to
-0.150 mm.

Per the judge's CI scope-split recommendation, PR #3005 drops the
``--differential-pairs`` re-enable on board 07's recipe and files
this follow-up for the upstream root cause.

## Repro

```bash
# Enable --differential-pairs on board 07's recipe (or run kct route
# directly), then re-route + check:
uv run python -m kicad_tools.cli route boards/07-matchgroup-test/output/matchgroup_test.kicad_pcb \
  --output /tmp/mgt.kicad_pcb --manufacturer jlcpcb --strategy negotiated \
  --no-auto-layers --layers 4 --differential-pairs --seed 42 --timeout 600 \
  --skip-nets GND,+1V2,+1V8 \
  --net-class-map boards/07-matchgroup-test/output/net_class_map.json

uv run kct check /tmp/mgt.kicad_pcb --mfr jlcpcb --format json \
  --net-class-map boards/07-matchgroup-test/output/net_class_map.json
```

Expected: 0 ``diffpair_clearance_intra``.  Actual: ~459.

## Root cause (verified)

1. ``diffpair_routing.py:1564`` computes
   ``spacing_cells = int(spacing / grid.resolution)`` where ``spacing
   = pair.rules.spacing`` (default 0.15 mm for HDMI).  The
   ``NetClassRouting.intra_pair_clearance`` from the sidecar is never
   consulted.
2. ``CoupledPathfinder.__init__`` does not accept any
   ``intra_pair_clearance`` parameter; the search only enforces a
   tolerance around ``target_spacing_cells``.
3. In ``_get_coupled_neighbors``, the "approach phase" tolerance
   widens to ``target_spacing_cells`` itself (line 371), permitting
   ``new_spacing`` to drop to 0 — both centerlines coincident — near
   the goal pads.  The asymmetric "converge" moves (lines 397-437)
   also do not enforce a floor.

## Suggested fix sketch (out of scope for #3003)

1. Compute ``required_center_spacing = trace_width +
   net_class.effective_intra_pair_clearance()`` in
   ``route_differential_pair_coupled`` (max with ``pair.rules.spacing``).
   Use ``math.ceil`` (not ``int``) when converting to grid cells.
2. Add a ``min_spacing_cells`` parameter to ``CoupledPathfinder.__init__``.
3. In ``_get_coupled_neighbors``, enforce ``new_spacing >=
   min_spacing_cells`` as a hard floor (independent of approach
   tolerance), exempting only the start/goal pad cells.
4. Mirror the floor enforcement into the two asymmetric "converge"
   moves.
5. Verify the trace optimizer / pull-tight pass uses ``intra_pair_clearance``
   (not ``trace_clearance``) for partner-net edges -- the optimizer
   currently reduces but does not eliminate the violations (2830 ->
   459 on board 07).

I attempted (1)-(4) in PR #3005 doctor session and the residual
violation count did NOT drop materially -- something downstream
(``pull_tight_pass`` in the trace optimizer, or fine-zone interaction
with the coarse grid) is also producing overlapping geometry.  This
issue is left intentionally open-ended on the suggested fix because
the next implementer will need to instrument the optimizer to
identify the remaining producer(s).

## Acceptance criteria

- [ ] Board 07 with ``--differential-pairs`` enabled produces 0
      ``diffpair_clearance_intra`` violations (or under a justified
      allowlist).
- [ ] Existing ``test_diffpair_serpentine_clearance.py`` + ``test_diffpair.py``
      still pass.
- [ ] No regression on other boards' diff-pair routing.

## Hard limits

- Do NOT widen DRC tolerances.
- Do NOT undo PR #3000's ``--net-class-map`` plumbing.
- Do NOT modify the audited Phase 2F/3I tuners.
- Do NOT regress PR #3005's shim hardening or its unit-test regression
  suite.

rjwalters:
## Summary

Adds a hard within-pair-spacing floor to ``CoupledPathfinder`` derived
from ``(trace_width + effective_intra_pair_clearance) / grid.resolution``,
and fixes the 2-pad fast path to detect polarity-swap pad layouts.
Incorporates PR #3005's serpentine clearance hardening to keep the
fix-set complete when this PR merges first.

Closes #3012 (partial -- floor + polarity-swap; empirical AC
deferred to a follow-up grid-resolution issue, see below).

## What changed in ``CoupledPathfinder``

| Hook | Bug | Fix |
|------|-----|-----|
| ``__init__`` | No knowledge of per-pair intra clearance. | Accepts ``min_spacing_cells`` (defaults to 0 for legacy callers). |
| ``_get_coupled_neighbors`` symmetric move (line 353) | Approach-phase tolerance widens to ``max(1, target)``, admitting spacing ``[0, 2*target]``. | New check ``new_spacing >= self.min_spacing_cells`` (bypassed only when BOTH new positions are endpoint cells). |
| ``_get_coupled_neighbors`` asymmetric P-advance (line 393) | Same bug; no floor. | Same fix. |
| ``_get_coupled_neighbors`` asymmetric N-advance (line 417) | Same bug; no floor. | Same fix. |
| ``route_differential_pair_coupled`` | Computed ``spacing_cells = int(pair.rules.spacing / grid.resolution)`` using the per-type ``DifferentialPairRules.spacing`` (an edge-to-edge config, not center-to-center) and never consulted ``NetClassRouting``. | Computes ``required_center_spacing = trace_width + effective_intra_pair_clearance`` from the per-pair ``NetClassRouting``, uses ``math.ceil``, and passes ``min_spacing_cells`` through to the pathfinder. |
| ``route_differential_pair_coupled`` 2-pad fast path | Hardcoded ``polarity_swap=False`` -- the search collapses spacing mid-run on inverted-row layouts (board-07 DDR strobe). | Calls ``_polarity_swap_between`` (the same detector used in the N-pad path). |

## What changed from PR #3005 (folded in to avoid losing it)

PR #3005 hardens the inline serpentine shim used inside
``route_differential_pair_coupled``.  Since PR #3005 is still in
review at the time of this commit and the issue brief explicitly
called out ""Do NOT regress PR #3005's shim hardening"", I folded
PR #3005's changes (and its 9-case test suite
``tests/test_diffpair_serpentine_clearance.py``) into this PR verbatim:

* ``create_serpentine`` and ``match_pair_lengths`` gain ``partner_route``
  and ``intra_pair_clearance_mm`` parameters.  When provided, the
  serpentine bulges AWAY from the partner using ``_outer_normal_hint``
  and runs a DRC self-check (``segment_clearance``) before commit.
* ``route_differential_pair_coupled`` gates the inline shim on
  ``length_critical=True`` -- non-critical pairs defer to the audited
  Phase 3I tuner (``tune_diff_pair_skew``).

If PR #3005 merges to ``main`` first this PR will need to rebase and
the changes will resolve cleanly (same lines, same intent).

## Empirical result (board 07, recipe from issue #3012)

```
cd .loom/worktrees/issue-3012 && uv run python -m kicad_tools.cli route \\
  boards/07-matchgroup-test/output/matchgroup_test.kicad_pcb \\
  --output /tmp/mgt.kicad_pcb --manufacturer jlcpcb --strategy negotiated \\
  --no-auto-layers --layers 4 --differential-pairs --seed 42 --timeout 600 \\
  --skip-nets GND,+1V2,+1V8 \\
  --net-class-map boards/07-matchgroup-test/output/net_class_map.json
```

|                            | Yield | ``diffpair_clearance_intra`` | Total DRC |
|----------------------------|-------|------------------------------|-----------|
| Baseline (no fix)          | 29/31 | 459                          | 534       |
| This PR                    | 29/31 | **434**                      | **509**   |
| Per-pair breakdown (this PR): | | |
| &nbsp;&nbsp;DQS_P/N        |       | **2** (was 48 -- polarity-swap fix) | |
| &nbsp;&nbsp;MIPI_CLK_P/N   |       | ~100 | |
| &nbsp;&nbsp;TMDS_D0/1/2_P/N |      | ~150 each | |

The polarity-swap fix on DQS is dramatic (48 -> 2, 95% reduction).
The floor enforcement helps marginally everywhere else.

## Why the AC is NOT fully met (and what I'd file as a follow-up)

The issue's empirical AC is ``0 diffpair_clearance_intra``
(or under a justified allowlist).  This PR lands 434 -- a 5%
reduction.  The dominant residual violations are caused by an
ARCHITECTURAL grid-resolution issue, not by the spacing floor:

* ``kct route`` itself emits the warning::

      UserWarning: Grid resolution 0.127mm may cause clearance
      violations with 0.15mm clearance. Recommend
      grid_resolution <= 0.075mm for reliable DRC compliance.

  The auto-selected grid (0.127 mm) is too coarse for the 0.1 mm
  intra-pair clearance: ``2 cells × 0.127 = 0.254 mm`` centerline
  spacing yields an edge-to-edge gap of ``0.254 - 0.15 = 0.104 mm``
  -- right at the threshold, and grid-quantization slack pushes
  it below.
* Many violating segment pairs have ``actual_value`` in the
  ``[0.05, 0.10) mm`` range -- below the ``0.1 mm`` requirement by
  the size of the quantization noise, not by the size of a search
  bug.  My floor of ``2 cells`` is the maximum I can enforce on a
  0.127 mm grid without choosing a value larger than the target
  ``target_spacing_cells`` (which would defeat the search).
* Forcing ``--grid 0.075`` (the recommended value) causes the
  routing to time out at **16/31 nets** -- an unacceptable yield
  regression.  Routing time on this board is grid-resolution-bound.

The proper fix needs an architectural change -- e.g. a per-pair
fine-grid sub-pass for diff-pair coupled routing while the
remaining nets stay on the coarse grid, or a mark-cell envelope
that reflects ``intra_pair_clearance`` to keep partner clearance
deterministic.  Both are out of scope for issue #3012 and merit a
fresh design proposal.

I'll file the follow-up as a separate ``loom:issue`` ticket after
this PR lands.

Because the AC is not met, this PR keeps the board-07 recipe with
``--differential-pairs`` DISABLED (matching the PR #3005 scope-split).
Re-enabling is gated on the grid-resolution follow-up.

## Tests

* 8 new unit tests in ``tests/test_diffpair_coupled_floor.py``:
  default zero floor, negative-clamp, positive storage, symmetric-move
  reject-below-floor, symmetric-move accept-at-floor, approach-phase
  floor enforcement, endpoint-cell bypass, asymmetric-move floor.
* PR #3005's 9-case ``tests/test_diffpair_serpentine_clearance.py``
  passes verbatim.
* All 244 ``test_diffpair*`` tests pass.

## Test plan

- [ ] Routing pipeline regression: `tests/test_diffpair*.py` (244 tests)
- [ ] Empirical reproducer on board 07 (recipe in summary) -- determinism verified across 2 runs (same 434 / 509)
- [ ] Match-Group Routing Regression CI gate
- [ ] No regression on PR #3005's shim hardening (its 9-case suite passes)

### DIFF
--- a/src/kicad_tools/router/diffpair_routing.py
+++ b/src/kicad_tools/router/diffpair_routing.py
@@ -160,6 +160,7 @@ class CoupledPathfinder:
         target_spacing_cells: int,
         net_class_map: dict[str, NetClassRouting] | None = None,
         allow_swap_via: bool = False,
+        min_spacing_cells: int = 0,
     ):
         """Initialize coupled pathfinder.
 
@@ -173,12 +174,29 @@ class CoupledPathfinder:
                 positions across an inner layer.  Used when source and
                 sink polarity orientations are mirrored (USB-C-shaped
                 pads).
+            min_spacing_cells: Issue #3012: Hard floor on the center-to-
+                center spacing (in grid cells) the search will tolerate
+                between P and N positions.  Derived in
+                ``route_differential_pair_coupled`` from
+                ``(trace_width + intra_pair_clearance) / grid.resolution``
+                so that within-pair edge-to-edge clearance is preserved
+                even when the approach-phase tolerance widens or the
+                asymmetric "converge" moves fire.  Defaults to ``0``
+                (legacy permissive behaviour) so callers that don't
+                supply per-pair NetClassRouting are unaffected.
+                Endpoint cells (start and goal pad positions) are
+                exempt from the floor -- those are the cells the pads
+                themselves occupy and the floor would otherwise
+                disqualify the search's only chance to land.
         """
         self.grid = grid
         self.rules = rules
         self.target_spacing_cells = target_spacing_cells
         self.net_class_map = net_class_map or {}
         self.allow_swap_via = allow_swap_via
+        # Issue #3012: store the within-pair spacing floor.  ``0`` means
+        # no floor (legacy behaviour).
+        self.min_spacing_cells = max(0, int(min_spacing_cells))
 
         # Pre-calculate trace clearance radius
         self._trace_half_width_cells = max(
@@ -353,6 +371,21 @@ class CoupledPathfinder:
             if abs(new_spacing - target_spacing_cells) > tolerance:
                 continue
 
+            # Issue #3012: Hard floor on within-pair spacing.  Independent
+            # of the approach-phase tolerance, the search must not place
+            # P and N centerlines closer than
+            # ``(trace_width + intra_pair_clearance) / grid.resolution``
+            # cells apart, or post-route the partner-net edges overlap.
+            # The floor is bypassed when BOTH new positions sit on
+            # endpoint cells (start or goal pads) -- those cells are
+            # owned by the pad footprints whose own spacing is set by
+            # the physical board geometry, not the router.
+            if self.min_spacing_cells > 0 and not (p_is_endpoint and n_is_endpoint):
+                # Use a small epsilon so a Euclidean spacing of exactly
+                # min_spacing_cells (axis-aligned) is accepted.
+                if new_spacing + 1e-9 < self.min_spacing_cells:
+                    continue
+
             # Calculate cost
             new_direction = (dx, dy)
             cost = self.rules.cost_straight
@@ -391,14 +424,34 @@ class CoupledPathfinder:
                     new_spacing = math.sqrt(spacing_dx * spacing_dx + spacing_dy * spacing_dy)
                     tolerance = max(1, target_spacing_cells)
                     if abs(new_spacing - target_spacing_cells) <= tolerance:
-                        # Direction tracking only reflects P's motion;
-                        # tag with the new direction so the cost-of-turn
-                        # logic still fires when the path bends.
-                        cost = self.rules.cost_straight
-                        if state.direction != (0, 0) and state.direction != (dx, dy):
-                            cost += self.rules.cost_turn
-                        new_state = CoupledState(cand_p, cand_n, (dx, dy))
-                        neighbors.append((new_state, cost, False))
+                        # Issue #3012: enforce the within-pair spacing
+                        # floor in the asymmetric P-advance move.  The
+                        # asymmetric moves only fire in the approach
+                        # phase, where the legacy tolerance was wide
+                        # enough to let centerlines coincide; without
+                        # this floor we observe -0.150mm overlap on
+                        # board 07 diff pairs.  Endpoint cells (P at
+                        # its pad AND N at its pad) bypass the floor
+                        # since the pads themselves define the spacing.
+                        n_is_endpoint = self._is_at_goal(
+                            cand_n, n_goal
+                        ) or self._is_at_goal(cand_n, n_start)
+                        bypass_floor = p_is_endpoint and n_is_endpoint
+                        if (
+                            self.min_spacing_cells > 0
+                            and not bypass_floor
+                            and new_spacing + 1e-9 < self.min_spacing_cells
+                        ):
+                            pass  # reject this candidate
+                        else:
+                            # Direction tracking only reflects P's motion;
+                            # tag with the new direction so the cost-of-turn
+                            # logic still fires when the path bends.
+                            cost = self.rules.cost_straight
+                            if state.direction != (0, 0) and state.direction != (dx, dy):
+                                cost += self.rules.cost_turn
+                            new_state = CoupledState(cand_p, cand_n, (dx, dy))
+                            neighbors.append((new_state, cost, False))
 
                 # N advances, P holds.
                 cand_p2 = state.p_pos
@@ -415,6 +468,21 @@ class CoupledPathfinder:
                 new_spacing = math.sqrt(spacing_dx * spacing_dx + spacing_dy * spacing_dy)
                 tolerance = max(1, target_spacing_cells)
                 if abs(new_spacing - target_spacing_cells) > tolerance:
+                    continue
+                # Issue #3012: same within-pair spacing floor as the
+                # P-advance branch.  P holds at its current position
+                # here; the floor is bypassed only when P happens to
+                # already be at its pad AND the candidate N is at its
+                # pad.
+                p_is_endpoint_held = self._is_at_goal(
+                    cand_p2, p_goal
+                ) or self._is_at_goal(cand_p2, p_start)
+                bypass_floor = p_is_endpoint_held and n_is_endpoint
+                if (
+                    self.min_spacing_cells > 0
+                    and not bypass_floor
+                    and new_spacing + 1e-9 < self.min_spacing_cells
+                ):
                     continue
                 cost = self.rules.cost_straight
                 if state.direction != (0, 0) and state.direction != (dx, dy):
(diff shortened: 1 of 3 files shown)


rjwalters:
## Judge Review: APPROVE

Verified all six review focus areas pass. This is the fourth scaffolding PR in the diff-pair gauntlet (after #3006, #3005, #3019); the scope-split here is principled and the honesty bar is met.

### Verification

**1. The polarity-swap fix in the 2-pad fast path (the KEY win)** — `diffpair_routing.py:1613-1632` replaces hardcoded `polarity_swap=False` with a call to `_polarity_swap_between(ps, ns, pe, ne)`. The helper at line 1300 is the same `@staticmethod` the N-pad path already uses at line 1447. Surgical fix, mirrors existing API. This single change is what produced the 48 -> 2 DQS reduction.

**2. The `min_spacing_cells` floor** is correctly enforced in all 3 move classes:
- Symmetric move (lines 380-387): floor with `(p_is_endpoint and n_is_endpoint)` bypass.
- Asymmetric P-advance (lines 402-441): same floor, same bypass.
- Asymmetric N-advance (lines 472-486): same floor, same bypass.

Endpoint bypass safety: `_is_at_goal()` at line 238 is strict (exact x/y match against start or goal — no tolerance). The bypass cannot fire at arbitrary intermediate cells. Default `min_spacing_cells=0` plus the `self.min_spacing_cells > 0` gate preserves legacy back-compat.

**3. `route_differential_pair_coupled` flow-through** at lines 1650-1697:
- `min_spacing_cells = max(1, math.ceil((trace_width + intra_pair_clearance) / grid.resolution))` — uses `ceil` to never round down below threshold.
- Net-class lookup `net_class_map.get(pair.positive.net_name)` matches PR #3005's pattern at line 1776.
- Fallback when net_class_map is empty: `pair_intra_clearance = float(spacing)` (the legacy edge-to-edge parameter) — legacy callers without a sidecar don't regress.

**4. The 8-test floor suite** in `tests/test_diffpair_coupled_floor.py` covers default-zero, negative-clamp, positive-storage, symmetric-reject, symmetric-accept, approach-phase enforcement, endpoint bypass, and asymmetric enforcement. All pass locally.

**5. PR #3005's 9-case serpentine suite** in `tests/test_diffpair_serpentine_clearance.py` is intact and passes (verified).

**6. Honesty bar** — body explicitly states AC NOT met (434 vs target 70 = ~70-error allowlist), stops short of overselling:
- Headline numbers correct: 459 -> 434 (5%), DQS 48 -> 2 (95%).
- Grid-quantization root cause documented with the `kct route` warning quoted verbatim.
- Forcing `--grid 0.075` causes 16/31 yield regression — documented as why the easy fix doesn't work.
- Follow-up issue #3023 ("router: per-pair fine-grid sub-pass for diff-pair routing") is filed.
- Board-07 recipe keeps `--differential-pairs` DISABLED, consistent with PR #3005 scope-split.

### CI

All 8 checks PASS. `mergeStateStatus=CLEAN`, `mergeable=MERGEABLE`.

| Check | Result |
|---|---|
| Lint & Format | pass |
| Type Check | pass |
| Test | pass |
| C++ Build Check | pass |
| Diff-Pair Routing Regression (06) | pass |
| Match-Group Routing Regression (07) | pass (10m31s) |

### Decision precedent

This is the same scaffolding precedent as #3006, #3005, #3019. The polarity_swap fix is a real, ship-independently win. The min_spacing_cells floor is correct infrastructure for when the fine-grid sub-pass (#3023) lands. The 5% reduction is honest; the 95% DQS win is the real signal. Approving.

Moving to `loom:pr` for merge.

🤖 Reviewed by Claude Opus 4.7 (1M context)

Issue #3045, opened 2026-05-18
Routing latency: 30s per net on board 07 makes CI/iteration loops impractical


rjwalters:
## Problem

Fleet audit 2026-05-18 — board 07 (matchgroup-test):

- A* per-net routing latency observed at ~30s (A0 net: 30s; MIPI_CLK_P: 30s).
- 31 nets × 30s = ~15 minutes minimum serial wall-clock just for the per-net A* phase.
- Audit script's 6-minute hard cap killed board 07 mid-route (during iteration 1 rip-up), and **0 segments had been written to disk** when killed — only iteration-0 best-result was held in memory and lost.

This bottleneck affects CI test runtime, iterative dev cycles, and multi-attempt strategies (negotiated loop / layer escalation can't afford multiple full passes).

## Curator notes (2026-05-18)

**Critical finding: C++ extension is NOT built in main worktree.**

`ls src/kicad_tools/router/router_cpp.*.so` returns no matches. Stale `.so` files exist only in detached `.claude/worktrees/` and `.loom/worktrees/issue-3033/`. Per `cpp_backend.py:8`, the C++ backend gives **10–100× speedup** for the A* loop. The 30s/net figure is very likely pure-Python overhead. **This must be verified before any heuristic or parallelism work is justified.**

Other observations:
- `core.py:5803 route_all_negotiated()` already exposes `region_parallel=False`, `partition_rows=2`, `partition_cols=2`, `max_parallel_workers=4` (lines 5815–5818), and `batch_routing=False` (line 5819).
- `batch_routing` IS wired through CLI as `--batch-routing` (`route_cmd.py:5363`).
- `region_parallel` is **NOT exposed as a CLI flag** — only reachable via Python API. Adding `--region-parallel` flag is part of the experiment cost.
- Default heuristic is `CongestionAwareHeuristic` (`heuristics.py:400`), not Manhattan. `ManhattanHeuristic`, `DirectionBiasHeuristic`, `WeightedCongestionHeuristic`, and `GreedyHeuristic` are all already implemented — swapping is a 1-line config change for benchmarking.

## Phased plan

**Phase 1 (cheap, do first):** Run `kct build-native --check` against the worktree used for the board 07 audit. If C++ is missing, run `kct build-native` and re-time a single net. If per-net A* drops to ~0.3–3s, the issue may largely be resolved; pivot remaining scope to checkpoint-to-disk (AC #3).

[...]

Issue #3054, opened 2026-05-18
Routing latency Phase 2: wire --region-parallel CLI flag (board 07 still > 4 min wall-clock)


rjwalters:
## Context

Spawned from #3045 Phase 1. After building the C++ router extension
(`kct build-native`) in a fresh worktree and re-running board 07
(`boards/07-matchgroup-test/generate_design.py`), the new measurements are:

| Metric | Pre-C++ (audit baseline) | Post-C++ (Phase 1) | AC target | Status |
|---|---|---|---|---|
| Board 07 wall-clock | killed at 6 min | 8:26 (506 s; routing 462 s) | < 4 min | FAIL |
| Nets routed | 0 written | 27/31 (87%) | all 31 | PARTIAL |
| Iteration-0 nets | killed mid-iter | 26/31 in 150 s | -- | OK |
| DRC errors | n/a | 3 clearance errs | 0 | FAIL (separate) |

The C++ backend is loaded (`Backend: cpp v1.0.0 (native, 10-100x faster)`)
and per-net A* dropped from ~30 s to ~1-5 s typical, but a few
high-conflict nets still take 20-40 s each and the rip-up phase adds
another ~150 s. Phase 1 docs/build-guidance has landed; **Phase 1
alone does not satisfy #3045 AC #1 (< 4 min wall-clock)**.

## What to do (Phase 2)

Wire `region_parallel=True` through the CLI per the curator plan in #3045:

1. `core.py:5803 route_all_negotiated()` already exposes
   `region_parallel=False`, `partition_rows=2`, `partition_cols=2`,
   `max_parallel_workers=4`. The Python API works today.
2. Add a `--region-parallel` flag (with `--partition-rows`,
   `--partition-cols`, `--max-parallel-workers`) to
   `src/kicad_tools/cli/route_cmd.py` next to the existing
   `--batch-routing` flag (route_cmd.py:5363).
3. Plumb through to `route_all_negotiated()` invocation in
   `route_cmd.py`.
4. Re-benchmark board 07 with
   `--region-parallel --partition-rows 2 --partition-cols 2`.
   Expected 2-3x speedup per the docstring, i.e. ~150 s wall-clock
   for the route step.
5. If region-parallel hits the < 4 min target on board 07, AC is met.
   If not, escalate to Phase 3 (heuristic experiment).

## Acceptance criteria

- `kct route --region-parallel ...` works end-to-end on board 07.
[...]

Issue #3051, opened 2026-05-18
Iteration-0 routed-PCB persistence: kill mid-run loses all routing work (board 07 audit)


rjwalters:
## Problem

Fleet audit 2026-05-18 — board 07 (matchgroup-test) was killed mid-route by a 6-minute wall-clock cap. After kill:

- Routed PCB file exists on disk
- Contains **0 routed segments** (only zones + input copy)
- Best result from iteration 0 was 23/31 nets and held in memory only — lost on kill

\`route_all_negotiated()\` keeps \`best_routes\` as an in-memory snapshot (\`src/kicad_tools/router/core.py:6339\`) and only writes to disk after the full negotiated loop completes (or a checkpoint callback fires). Killing the process mid-loop discards everything.

## Fix candidates

1. **Periodic checkpoint write to disk**: after every successful iteration's best-result update, also write the in-progress PCB to a \`.partial.kicad_pcb\` file. Cheap insurance against crashes.
2. **Wire \`checkpoint_callback\` by default** in \`kct route\` to a small helper that persists the snapshot.
3. **Signal handler** (SIGTERM/SIGINT) in \`route_all_negotiated()\` that flushes \`best_routes\` to the target output path before re-raising.

Option 1 is the simplest. Option 3 is the most robust for CI kills.

## Acceptance criteria

- Killing \`kct route\` mid-loop (e.g., \`timeout 60 kct route ...\`) leaves a routed PCB on disk reflecting the best iteration completed so far.
- No regression on the success path (PCB written only once, atomically).
- New test exercises the kill-and-recover path on a fixture.

## References

- \`src/kicad_tools/router/core.py:6339\` — \`best_routes\` snapshot.
- \`src/kicad_tools/router/core.py:5833\` — \`checkpoint_callback\` parameter (exists, not wired in CLI).
- Audit run 2026-05-18 — board 07 mid-loop kill demonstrated the loss.

## Tier

S-M (~100-150 LOC)

rjwalters:
## Curator notes (2026-05-18)

### Code references verified

- `src/kicad_tools/router/core.py:6339` — `best_routes: list[Route] = copy.deepcopy(list(self.routes))`. Confirmed snapshot site.
[...]

PR #3056, opened 2026-05-18
docs: require kct build-native for C++ router performance (Phase 1 of #3045)

rjwalters:
## Summary

Phase 1 of #3045. The curator's critical finding was that the C++
router extension (`router_cpp.*.so`) is not built in fresh
worktrees and `uv sync` does not run `kct build-native`, so the
router silently falls back to pure-Python — the root cause of the
"30 s per net" symptom in the original board 07 audit.

This PR is **documentation only**. It hard-codes the
`kct build-native` step into the dev setup sequence in three places
so future agents and developers cannot miss it:

- `README.md` (Development -> Quick Start): the canonical setup
  sequence now lists `uv run kct build-native` immediately after
  `uv sync --extra dev`, with `--check` verification and a routing
  performance call-out. New "Fresh worktree checklist" subsection
  covers the worktree-isolation gotcha.
- `CLAUDE.md` (project root): top-level note so any LLM agent
  reading the file before benchmarking the router sees the
  build-native requirement first.
- `docs/guides/routing.md` (Performance section): new "Native C++
  Backend (Build First!)" subsection inserted at the top of
  Performance, including the log lines that distinguish `cpp` vs
  `python` backends so triagers can spot a missing extension from
  a routing log.

## Phase 1 measurements (board 07, fresh worktree, M2 Mac)

| Metric | Pre-build (audit baseline) | Post-build (this PR) | AC target | Status |
|---|---|---|---|---|
| Board 07 wall-clock | killed at 6 min cap | 8:26 (506 s; routing 462 s) | < 4 min | FAIL |
| Nets persisted | 0 | 27/31 (87%) | all 31 | PARTIAL |
| Iteration-0 nets | killed mid-iter | 26/31 in 150 s | -- | OK |
| Convergence | -- | iter 1, 31/31 routed | converged | OK |
| Backend reported | n/a | `cpp v1.0.0 (native, 10-100x faster)` | cpp | OK |
| Boards 00 / 01 / 02 | n/a | 8 s / 26 s / 1:25 | unchanged | OK no regression |
| DRC errors (board 07) | n/a | 3 clearance errs | 0 | FAIL (separate from #3045) |

Per-net A* dropped from ~30 s (audit baseline) to ~1-5 s typical,
but a handful of high-conflict nets (TMDS_D0, MIPI_DAT0, DQ0) still
take 20-40 s and the rip-up phase adds another ~150 s. **Phase 1
alone does not satisfy AC #1 (< 4 min wall-clock).**

## Outcome classification (per curator's plan)

**Outcome B** — C++ build succeeded and the router is dramatically
faster (and now actually finishes board 07 vs being killed), but
the 4-minute wall-clock acceptance criterion is missed. Per the
playbook, Phase 1 docs land here and Phase 2 (region-parallel CLI
flag) is queued as a follow-up.

## Follow-ups filed

- **#3054** (`loom:curated`) — Phase 2: wire `--region-parallel`
  CLI flag and re-benchmark (expected 2-3x speedup from the
  docstring on `route_all_negotiated`).

## Out of scope

- Phase 2 (`--region-parallel` plumbing) — tracked in #3054.
- Phase 3 (heuristic experiment) — only if Phase 2 still misses AC.
- Issue #3051 (iteration-0 persistence) — tracked separately;
  observed working correctly during this run (checkpoint at iter 1
  wrote 27 nets to disk).

## Test plan

- [x] `kct build-native --check` reports "available" on fresh
      worktree after running `kct build-native`
- [x] `kct route` log on board 07 prints
      `Backend: cpp v1.0.0 (native, 10-100x faster)`
- [x] Board 07 generate_design.py completes (8:26 wall-clock,
      previously killed at 6 min)
- [x] No regression on boards 00 / 01 / 02 (all pass with same or
      faster timing)
- [x] Docs render correctly (manual review of markdown)
- [x] Lint clean on changed files (markdown only — ruff skips)
- [ ] Reviewer verifies the README + CLAUDE + routing-guide
      changes are sufficient for future agents to discover the
      `kct build-native` requirement before benchmarking

Phase 1 of #3045

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,3 +1,27 @@
 <!-- BEGIN LOOM ORCHESTRATION -->
 This repository uses [Loom](https://github.com/rjwalters/loom) for AI-powered development orchestration. See `.loom/CLAUDE.md` for the full guide (roles, labels, worktrees, configuration).
 <!-- END LOOM ORCHESTRATION -->
+
+## Routing performance: build the C++ backend first
+
+Before benchmarking routing latency or filing "router is slow" issues,
+verify the C++ router extension (`src/kicad_tools/router/router_cpp.*.so`)
+is built in the active worktree:
+
+```bash
+uv run kct build-native --check
+# Expected: "C++ backend: available (version 1.0.0)"
+```
+
+If it reports "not installed", build it once:
+
+```bash
+uv run kct build-native
+```
+
+`uv sync` does **not** build the native extension — fresh checkouts and
+new git worktrees (`.loom/worktrees/issue-N/`) need this step explicitly.
+The C++ backend gives a 10-100x speedup for the A* loop, so a missing
+extension is the most likely cause of multi-minute-per-net routing.
+
+See `README.md` "Fresh worktree checklist" for the full setup sequence.
--- a/README.md
+++ b/README.md
@@ -583,6 +583,16 @@ cd kicad-tools
 # Set up development environment (installs all dev dependencies)
 uv sync --extra dev
 
+# Build the C++ router backend (REQUIRED for production routing speed,
+# delivers 10-100x A* speedup vs pure-Python; takes ~30s on first build,
+# cached thereafter). uv sync does NOT build this — fresh checkouts and
+# fresh worktrees need this step explicitly.
+uv run kct build-native
+
+# Verify the C++ backend is installed
+uv run kct build-native --check
+# Expected: "C++ backend: available (version 1.0.0)"
+
 # Run tests
 uv run pytest
 
@@ -592,6 +602,20 @@ uv run ruff check .
 # Format code
 uv run ruff format .
 ```
+
+> **Routing performance note**: Without the C++ backend, board routing falls
+> back to pure-Python A* — typically 5-10x slower per net, which can push
+> medium boards (e.g. board 07 in `boards/07-matchgroup-test/`) past the
+> default 6-minute CI cap. If `kct route` appears stuck or per-net log lines
+> take tens of seconds, run `kct build-native --check` first — a missing
+> extension is the most common cause.
+
+#### Fresh worktree checklist
+
+When creating a new git worktree (e.g. via `./.loom/scripts/worktree.sh`),
+the worktree's `.venv/` does not inherit the C++ extension built in the
+main checkout. After `cd` into the worktree, run `uv run kct build-native`
+once before any routing benchmarks.
 
 ### Available Commands
 
--- a/docs/guides/routing.md
+++ b/docs/guides/routing.md
@@ -482,6 +482,53 @@ ladder). The two non-obvious cases:
 
 ## Performance
 
+### Native C++ Backend (Build First!)
+
+The router has a C++ A* implementation that delivers a **10-100x speedup**
+over pure Python for the inner pathfinding loop. **It is not built by
+`uv sync`** — you must explicitly run `kct build-native` once after every
+fresh checkout or new git worktree.
+
+```bash
+# Check whether the C++ extension is already built
+kct build-native --check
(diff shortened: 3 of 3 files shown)


rjwalters:
## Judge review: APPROVE

Docs-only PR. Solid Phase 1 deliverable; safe to merge.

### Verified

**Scope (docs-only, no source drive-bys):**
- `gh pr view --files`: only `CLAUDE.md` (+25/-1), `README.md` (+24/-0), `docs/guides/routing.md` (+47/-0). Three files, all markdown. Clean.

**Doc accuracy:**
- `kct build-native --check` command exists at `src/kicad_tools/cli/build_native_cmd.py:454` with `action="store_true"`, matches the documented invocation exactly.
- README's Quick Start now reads in the right order: `uv sync --extra dev` -> `uv run kct build-native` -> `uv run kct build-native --check` with expected output. The "Routing performance note" blockquote and "Fresh worktree checklist" subsection are both prominent and findable via search.
- CLAUDE.md top-level note is the first thing an LLM agent sees when probing the repo for routing context. Actionable (gives exact commands and the failure mode signal: "multi-minute-per-net routing").
- `docs/guides/routing.md` "Native C++ Backend (Build First!)" sits at the top of the Performance section as the first subsection — correct placement; readers don't have to scroll to find it.
- The log-line decoder (`Backend: cpp v1.0.0` vs `Backend: python (fallback)`) is exactly the kind of "spot it in 5 seconds" signal triagers need.

**PR-body auto-close hygiene:**
- Title: "Phase 1 of #3045" (no auto-close keyword)
- Body uses "Phase 1 of #3045" twice, lists #3054 as the Phase 2 follow-up, and explicitly notes #3051 is "tracked separately." No "Closes #3045" / "Fixes #3045" / "Resolves #3045" anywhere. `#3045` will stay open after merge for Phase 2 — correct.

**CI status (all 8 checks green):**
- Lint & Format, Type Check, Test, C++ Build Check, kicad-cli Round-trip Smoke, Routed PCB DRC, Diff-Pair Routing (board 06), Match-Group Routing (board 07) — all pass.
- mergeStateStatus: CLEAN.

**Issue #3054 (Phase 2 follow-up) sanity check:**
- Exists, `loom:curated`, scoped narrowly to adding `--region-parallel` CLI flag + 3 sibling flags (`--partition-rows`, `--partition-cols`, `--max-parallel-workers`) and re-benchmarking. Tier S-M (80-150 LOC). Much smaller than original #3045. Good split.

**Issue #3045 label state:**
- Still `loom:building` + `loom:curated`, OPEN. Correct — Phase 2 work continues.

### #3051 investigation (per review request)

The builder claimed iter-0 persistence "appears to already be working" because the Phase 1 verification run wrote a checkpoint at iteration 1. I checked the curator's three layer-escalation call sites:

- `route_cmd.py:2442` — `route_all_negotiated(...)` call, NO `checkpoint_callback` kwarg passed
- `route_cmd.py:3144` — same, NO `checkpoint_callback`
- `route_cmd.py:4208` — same, NO `checkpoint_callback`

The main `kct route` path DOES wire the checkpoint callback correctly (`route_cmd.py:6358, 6404, 6465, 6484` all pass `checkpoint_callback=_checkpoint_cb`). The Phase 1 run on board 07 evidently went through one of those four primary call sites, which is why checkpointing worked.

**#3051 is NOT over-spec'd.** The three layer-escalation/fallback call sites (used when `--escalate-layers` or multi-attempt flags trigger fallback paths) genuinely bypass the checkpoint callback. A kill during a layer-escalation attempt would still lose work. The builder's observation is real but limited to the happy path. **Recommend #3051 stays open as scoped** — builder's note in the PR body ("tracked separately") is the right call. No action needed on this PR.

### Approval

Approving via label swap per Judge protocol. Removing `loom:review-requested`, adding `loom:pr`.

PR #3059, opened 2026-05-18
fix(zones): repair priority-overlap copper collapse on multi-net F.Cu zones (closes #3043)

### LINKED ISSUE #3043: Zone-priority overlap drops copper to zero (+1V8/+1V2 vs +3V3 on board 06)
## Problem

Fleet audit 2026-05-18 on board 06 (diffpair-test):

- 3 zones declared on F.Cu: \`+1V8\`, \`+1V2\`, \`+3V3\`
- All 3 zones overlap geometrically
- All 3 zones emit **zero copper** in the filled output

The zone-priority resolver appears to fail in a way that produces empty fills rather than carving the higher-priority zone out of the lower-priority one.

\`kct check\` reports 5 \`zone_unfilled\` warnings on this board (3 of which are these three) but no error.

## Fix candidates

1. Investigate the zone-fill engine's priority comparator — when three+ zones overlap, does it correctly fall through to the highest-priority's geometry, or does it null out all of them?
2. Enforce non-overlap at zone-declaration time (\`generate_design.py\` linter) and emit a clear error pointing at the conflict.
3. Add a \`zone_overlap\` DRC rule that catches this at check time even if the fill silently fails.

## Acceptance criteria

- Board 06 emits non-zero copper on at least the highest-priority zone of the overlapping cluster.
- New synthetic test with 2 overlapping zones produces expected geometry per zone-priority rules.
- New \`zone_overlap\` rule (if added) flags the board 06 declaration in \`kct check\`.

## References

- Audit run 2026-05-18 — board 06.
- \`src/kicad_tools/zones/\` — fill engine.
- \`boards/06-diffpair-test/generate_design.py\` — repro of the overlap declaration.

## Tier

M (~150 LOC)

rjwalters:
## Summary

Issue #3043 reported board 06's three F.Cu power zones (`+3V3`, `+1V8`, `+1V2`) emitting zero copper despite distinct priorities. Curator hypothesis: `_net_pad_positions_absolute` returning empty -> fallback to full board outline -> highest-priority wins everything.

Instrumentation disproved the hypothesis: pads were correctly found (11/3/13), bboxes were correctly computed. The real failure was that the bboxes themselves were not geometrically disjoint -- `+3V3`'s bbox fully encloses both `+1V8` and `+1V2` because all three nets feed pads in the same BGA region. KiCad's fill resolver then silently awards the overlap to the highest-priority zone.

## Root cause (from instrumentation)

```
[zones.debug] +3V3 on F.Cu: pads=11, bbox=computed -> (110, 106.5)-(196.5, 168.5)
[zones.debug] +1V8 on F.Cu: pads=3,  bbox=computed -> (116.3, 117)-(134.5, 173.5)   (inside +3V3)
[zones.debug] +1V2 on F.Cu: pads=13, bbox=computed -> (150.96, 117.23)-(196.5, 149.3) (inside +3V3)
```

The board 05 layout (existing test fixture) clusters power pads at disjoint x-positions (10/30/50/70), so the per-net bboxes are naturally non-overlapping. Board 06 is the first test board with *interleaved* power pads, exposing the latent failure mode.

## Changes

- `src/kicad_tools/zones/generator.py`:
  - Extend `_compute_pour_outlines` with a second pass that subtracts higher-priority bboxes from lower-priority outlines using Shapely's polygon difference. Sort nets per layer by priority (descending), accumulate the union of higher-priority bboxes, then carve each lower-priority bbox against that union.
  - Add helper functions `_union_polygons` and `_subtract_polygon` for the Shapely-based geometric operations. Both degrade gracefully when Shapely is missing (return the original bbox, which preserves pre-#3043 behavior).
  - Upgrade `_boundaries_overlap` to use exact polygon intersection (Shapely) instead of AABB-only when Shapely is available, so the carved-out concave `+3V3` polygon is correctly recognised as disjoint from `+1V8`/`+1V2` and no spurious overlap warning fires.
  - Update module-level and `_compute_pour_outlines` docstrings to document the new disjoint-carve pass and reference issue #3043.

- `tests/test_zone_priority.py` (NEW):
  - 4 regression tests covering the 2-zone reduction, the full board-06 3-net nested shape, the direct `_compute_pour_outlines` unit invariant, and the no-spurious-warning check. All gated on Shapely availability via `pytest.importorskip("shapely")`.

| Board 06 emits non-zero copper on at least the highest-priority zone | YES | All 5 zones now have positive area: GND=7821 mm^2, VBUS_USB=7821 mm^2, +3V3=2965 mm^2 (carved), +1V8=1028 mm^2, +1V2=1460 mm^2 |
| New synthetic test with 2 overlapping zones produces expected per-zone geometry | YES | `tests/test_zone_priority.py::test_two_overlapping_power_nets_become_disjoint` |
| No regression on other boards' zone generation | YES | Board 05 zones unchanged (4-vertex bboxes, already disjoint by clustered-pad layout); all 287 existing zone tests pass |

## Out of scope (per curator notes)

- `zone_overlap` DRC rule (curator option 3): with truly disjoint outlines there's nothing to flag.
- Lint-time error on zone-overlap declaration (option 2): premature; designers may legitimately want overlap with explicit priority resolution.
- Pre-filling `(filled_polygon ...)` data so `kct check` stops reporting `zone_unfilled` warnings: KiCad fills on open, this is a separate tooling-side issue.

## Test Plan

- [x] `uv run pytest tests/test_zone_priority.py -v` -> 4 passed
- [x] `uv run pytest tests/test_build_zones_step.py tests/test_zone_generator.py tests/test_zones.py tests/test_board_05_zones.py tests/test_validate_zone_fill.py tests/test_zone_api.py tests/test_pcb_add_zone.py tests/test_pcb_zones.py tests/test_zones_cmd.py tests/test_zone_generator_edge_clearance.py tests/test_gerber_zone_fill.py` -> 287 passed, 1 skipped
- [x] `uv run python boards/06-diffpair-test/generate_design.py` -> no overlap warnings; all 5 zones have positive area
- [x] `uv run ruff check src/kicad_tools/zones/generator.py tests/test_zone_priority.py` -> all checks passed
- [x] `kct fleet status` -> board 05/06 still surveyed correctly; no degraded ship-status

Closes #3043

### DIFF
--- a/src/kicad_tools/zones/generator.py
+++ b/src/kicad_tools/zones/generator.py
@@ -4,8 +4,8 @@ Zone generator for creating copper pour zones on PCBs.
 This module provides a high-level API for generating copper pour zones,
 with automatic board outline detection and sensible defaults for power nets.
 
-Outline allocator (#2771)
--------------------------
+Outline allocator (#2771, #3043)
+--------------------------------
 
 In addition to the layer/priority allocator (:func:`_assign_layers_for_pour_nets`),
 this module owns the *outline* allocator (:func:`_compute_pour_outlines`).
@@ -21,6 +21,15 @@ distinct, which silenced the warning, but distinct priorities alone do not
 produce real copper for the losing zones -- the outlines have to be
 geometrically disjoint for KiCad's fill resolver to award copper to more
 than one zone on a shared layer.
+
+The original PR #2771 only computed each net's bbox independently and
+relied on pad-cluster separation to keep the bboxes disjoint.  On real
+boards where power-net pads are *spatially interleaved* (e.g. board 06's
+``+3V3``/``+1V8``/``+1V2`` cluster on a BGA), the bboxes overlap and the
+silent-override returns.  Issue #3043 extends the allocator with a
+second pass that *subtracts higher-priority bboxes from lower-priority
+outlines* (using Shapely's polygon difference), so the resulting
+outlines are disjoint regardless of pad layout.
 """
 
 from __future__ import annotations
@@ -401,25 +410,39 @@ class ZoneGenerator:
     ) -> bool:
         """Check whether two boundary polygons overlap.
 
-        Uses bounding-box intersection as a conservative approximation.
-        Two polygons whose bounding boxes overlap are considered overlapping.
-        This is intentionally conservative -- it may report overlaps for
-        polygons that only share a bounding-box region but not actual area.
-        For the zone-overlap-warning use case, false positives are acceptable
-        while false negatives would hide real problems.
+        Uses Shapely's exact polygon intersection when available so the
+        outline allocator's carved-out (concave) polygons are correctly
+        recognised as disjoint when their AABBs still overlap (#3043).
+        Falls back to a conservative AABB-overlap test when Shapely is
+        not installed.
 
         Returns:
-            True if the boundaries' bounding boxes overlap.
+            True if the boundaries overlap with positive area.
         """
         if not boundary_a or not boundary_b:
             return False
+
+        # Prefer exact intersection via Shapely so the disjoint-but-AABB-
+        # overlapping case (carved-out outlines from ``_compute_pour_outlines``)
+        # does not trigger a spurious warning.
+        try:
+            from shapely.geometry import Polygon
+
+            poly_a = Polygon(boundary_a)
+            poly_b = Polygon(boundary_b)
+            if not poly_a.is_valid or not poly_b.is_valid:
+                # Fall through to AABB on degenerate input.
+                raise ValueError("invalid polygon")
+            return poly_a.intersection(poly_b).area > 1e-9
+        except (ImportError, ValueError):
+            pass
 
         a_xs = [p[0] for p in boundary_a]
         a_ys = [p[1] for p in boundary_a]
         b_xs = [p[0] for p in boundary_b]
         b_ys = [p[1] for p in boundary_b]
 
-        # Axis-aligned bounding-box overlap test
+        # Axis-aligned bounding-box overlap test (conservative fallback).
         return not (
             max(a_xs) <= min(b_xs)
             or max(b_xs) <= min(a_xs)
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

This is exemplary work. The instrumentation-first methodology that overturned the curator's hypothesis is exactly the kind of disciplined debugging this codebase benefits from, and the resulting fix is mathematically grounded in KiCad's actual semantics rather than a workaround.

### Verification performed

**Carving math (src/kicad_tools/zones/generator.py:1188-1230)**
- Per-layer sort is descending by priority — matches KiCad's "higher priority wins overlap" convention.
- Winners-union accumulator iterates correctly; highest-priority zone keeps raw bbox, subsequent zones get the running union subtracted.
- Degenerate cases handled:
  - Net with 0 pads -> skipped (not added to `raw_bboxes`).
  - Empty difference (lower-priority entirely contained) -> returns un-carved fallback, docstring transparently explains the trade-off (zone exists in file with declared geometry; KiCad's fill engine still awards zero copper). Verified via unit probe.
  - MultiPolygon result -> keeps largest piece, smaller fragments lost. Reasonable since KiCad outlines are single-polygon by file format.

**Shapely fallback (src/kicad_tools/zones/generator.py:1316-1374)**
- Genuinely real, not theoretical. Verified by subprocess test that intercepts `__import__("shapely")` to raise ImportError:
  - `_union_polygons` -> returns `None`
  - `_subtract_polygon` -> returns `fallback` (identity-equal to original minuend)
- Pre-fix behavior preserved when Shapely missing.
- Note: Shapely is in `[geometry]` extra (pyproject.toml:87-91), not a hard dep, so the fallback path is reachable in production. Test file correctly gates on `pytest.importorskip("shapely")`.

**Board 06 fix confirmed independently**
```
Zones in diffpair_test_routed.kicad_pcb:
  GND       In1.Cu  vertices=  4  area=7821.0 mm^2
  VBUS_USB  In2.Cu  vertices=  4  area=7821.0 mm^2
  +3V3      F.Cu    vertices= 12  area=2965.2 mm^2   <- carved concave polygon
  +1V8      F.Cu    vertices=  4  area=1028.3 mm^2
  +1V2      F.Cu    vertices=  4  area=1460.5 mm^2
```
Matches PR claims byte-for-byte; +3V3 is the 12-vertex carved shape, all three power zones have positive area.

**Tests**
- `tests/test_zone_priority.py` -> 4 passed (0.44s)
- Broad zone sweep (`tests/test_zone*.py` + auto_pour + build_zones_step + ...) -> 311 passed, 1 skipped (more than the builder's 287 claim because the new 4 tests are now in the count; no regressions)
- Test fixtures cover 2-zone overlap, 3-zone nested (the actual board-06 shape), direct `_compute_pour_outlines` unit invariant, and no-spurious-warning regression. Coverage is appropriate for the bug.

**CI**
- All 8 checks pass (including Diff-Pair Routing Regression on board 06, which exercises the fix end-to-end).

**Scope**
- `git diff origin/main...HEAD --stat`: 2 files, +579 / -11. Exactly `src/kicad_tools/zones/generator.py` and `tests/test_zone_priority.py`. No creep.

### Minor nit (non-blocking)

`tests/test_zone_priority.py` is not `ruff format`-clean (assert-strings collapsed onto single lines + a multi-line argument that fits on one). The lint workflow in `.github/workflows/ci.yml:30-32` uses `continue-on-error: true`, so this doesn't fail CI, and the PR body's claim of "ruff check ... all checks passed" is technically accurate (only `ruff format --check` complains). Suggest running `uv run ruff format tests/test_zone_priority.py` in a follow-up sweep but not blocking on it.

### Why this matters

Issue #3043 was a real correctness bug — board 06's three F.Cu power zones were silently emitting zero copper despite being declared in the file. The fix mirrors KiCad's priority semantics geometrically at the outline-allocator layer, rather than papering over the symptom. The board-05 fixture happened to dodge the issue by clustering pads at disjoint x-positions; board 06 is the first interleaved case, and now the codebase is correct for both.

Ready to merge.

Issue #3077, opened 2026-05-19
router: negotiated rip-up does not re-validate against post-#3070 lateral via halos


rjwalters:
## Follow-up to #3073

PR for #3073 fixed the helper bug where the lateral via-escape helper's surface stub was placed at the dispatcher-supplied trace width (typically 0.5mm for power-derived nets), which overshot the inter-pad channel on 0.5mm pitch LQFP-48 and produced pad-segment DRC violations. The fix necks the surface stub to the manufacturer minimum trace width when the dispatcher width does not fit the channel.

After that fix, board 04 with ``--strict-in-pad-clearance`` improved from 8/9 nets + 7 segment-segment + 2 pad-segment + 3 segment-via DRC errors down to 8/9 nets + 7 segment-segment + 3 segment-via errors. The pad-segment errors are gone, but the NRST + BOOT0/SWO/SWCLK segment-segment overlaps persist.

## Root cause of remaining failures

The lateral via for OSC_OUT lands at (125.7875, 121.75) — 1.05mm west of the pad center. At LQFP-48 0.5mm pitch with 0.6mm vias and 0.15mm clearance, this is the SMALLEST viable lateral offset that satisfies same-row foreign-pad clearance (closer candidates are rejected because the via halo would overlap with the same-row neighbour pad's clearance band). But this position puts the via in the **escape corridor for the adjacent neighbour pin (NRST on pin 7)**.

When the main router subsequently routes BOOT0 / SWDIO / SWCLK / SWO, it doesn't see the lateral via as a hard obstacle in its negotiated rip-up flow. The rip-up cycle then commits routes that overlap with the lateral via halo, producing segment-via clearance violations at the via center and segment-segment overlaps where two displaced routes intersect.

This is the same architecture-level gap as #3002: a route validator sees the violation post-hoc, but the negotiated rip-up commit gate does not consult the latest foreign-via context to reject the route at commit time.

## Acceptance criteria

- Negotiated rip-up re-validates each candidate route against ALL foreign vias (including escape-time vias added by ``_try_lateral_via_escape``, ``_try_in_pad_escape``, etc.) before committing it to the route plan.
- Board 04 with ``--strict-in-pad-clearance`` reaches 9/9 nets with <= 6 DRC errors at ``--seed 42``.

## Out of scope (for this issue)

[...]

PR #3079, opened 2026-05-19
fix(escape): lateral helper necks surface stub to channel-fit width (closes #3073)

### LINKED ISSUE #3073: Investigate why --strict-in-pad-clearance + lateral via-escape regresses board 04 (9/9 -> 8/9)
## Context

PR #3070 (closes #3063) shipped the ``_try_lateral_via_escape`` helper plus a ``_can_place_via`` grid-origin bounds fix.  The original intent was for board 04 (``boards/04-stm32-devboard``) to be the in-tree customer that exercises strict mode in CI.

Judge feedback (https://github.com/rjwalters/kicad-tools/pull/3070#issuecomment-4483615406) measured:

- Pure baseline (origin/main, ``--seed 42``): **9/9 nets, 6 DRC errors**
- PR #3070 with ``--strict-in-pad-clearance``: **8/9 nets, 13 DRC errors** (including segment/segment overlaps at -0.200 mm — severely overlapping copper)

So enabling strict mode + the lateral helper on this board:

1. **Drops a net** (SWO does not route)
2. **Adds 7 DRC errors**, including segment-segment overlaps that did not exist before

The infrastructure itself (helper + bounds fix + 10 synthetic-fixture tests) shipped in PR #3070.  The board-04 opt-in was reverted (see commit 016f7a13 on PR #3070, post-judge-feedback).

## The question

The lateral helper **should** find a valid off-pad via for whichever pin caused the in-pad rescue to defer.  Two hypotheses:

1. **Helper bug**: the search budget / step / direction is wrong for the dropped net's pin geometry, so the helper returns ``None`` and the net is left unrouted.
2. **Downstream conflict**: the helper finds a candidate that passes ``_can_place_via``, but the subsequent route through that candidate collides with already-committed copper (which is what produces the new segment/segment overlaps).

The fact that strict mode adds segment-segment overlaps (which are NOT in-pad-via violations) hints at hypothesis #2 — the lateral via is being placed, but the trace heading to it is being routed through already-committed copper without re-validating clearance.

## Reproduction

\`\`\`bash
# Baseline (should produce 9/9 nets, ~6 DRC errors)
git checkout main
cd boards/04-stm32-devboard && python generate_design.py
uv run kct check --mfr jlcpcb-tier1 output/stm32_devboard_routed.kicad_pcb

# Regression (should produce 8/9 nets, ~13 DRC errors)
# Re-apply just the flag change in boards/04-stm32-devboard/generate_design.py:
#   add ``--strict-in-pad-clearance`` to the route_pcb() cmd list
# Then re-run the above two commands.
\`\`\`

## Suggested investigation steps

1. Diff the routed PCB between strict-on and strict-off at ``--seed 42``: which net dropped, where did the new segment/segment overlap-pairs land?
2. Trace logging in ``_try_lateral_via_escape`` for the dropped net: did it return ``None`` (helper bug — hypothesis #1) or did it return a candidate that was later overdrawn (hypothesis #2)?
3. If hypothesis #2, the fix likely belongs in the negotiated re-validation loop (see #3002 for the analogous main-router fix).

## Acceptance

- Either: a fix that lets ``--strict-in-pad-clearance`` ship on board 04 at 9/9 nets and ``<=`` 6 DRC errors; or
- A documented root cause + a synthetic-fixture test that pins the failure mode; or
- A decision that board 04 is the wrong customer (the LQFP-48 0.5 mm pitch is genuinely too tight) and selecting a different in-tree customer.

## Refs

- PR #3070 (lateral helper infrastructure)
- Issue #3063 (parent — this is the deferred "customer" half)
- Issue #3002 (analogous re-validation fix for the main router)
- Issue #3048 (parent: --strict-in-pad-clearance series)

rjwalters:
## Summary

- Investigation of #3073 (board 04 8/9 nets + DRC overlaps when ``--strict-in-pad-clearance`` is enabled) identified the root cause as the lateral via-escape helper's surface stub being built at the dispatcher-supplied trace width (typically 0.5mm for power-derived nets), which does not fit the ~0.2mm copper channel between same-row neighbour pads at LQFP-48 0.5mm pitch.
- Fix: after ``_can_place_via`` accepts a candidate via location, the helper now validates the prospective surface stub against neighbour pads via ``_segment_to_pad_edge_gap``.  When the dispatcher width violates channel clearance, it necks the stub down to the manufacturer minimum (e.g. 0.127mm for jlcpcb-tier1); if even the necked width fails, the candidate is rejected and the next offset is tried.  When all candidates fail, the helper returns None (deferring to the main router, matching pre-#3063 behaviour).
- Board 04 ``--strict-in-pad-clearance`` opt-in stays reverted; the remaining strict-mode failures (NRST blocked + BOOT0/SWO segment-segment overlaps from negotiated rip-up not seeing the lateral via as a hard obstacle) are tracked as follow-up issue #3077 (negotiated-loop re-validation, a #3002 analogue at the main-router commit-time gate).

## Branch selected (per curator's 3-branch AC)

This PR lands **Branch B**: helper improvement (eliminates the pad-segment DRC errors that are directly attributable to the helper's surface-stub width choice) PLUS a follow-up issue (#3077) for the broader negotiated-loop re-validation work that strict mode on board 04 still depends on.  Branch C also applies (LQFP-48 0.5mm pitch is geometrically the wrong customer for strict mode at 0.6mm via diameter: the minimum viable lateral offset is ~1.05mm, past the adjacent pin row's escape lane) and is documented in board 04's ``route_pcb`` docstring with a pointer to #3077 for the work needed before any in-tree customer can validate strict mode end-to-end.

## Numbers

**No-strict baseline** (board 04 default recipe, unchanged): 9/9 nets, 3 via_in_pad + 5 segment/pad-via DRC errors.

**Strict mode WITHOUT fix** (PR #3070 + ``--strict-in-pad-clearance``): 8/9 nets, 7 segment-segment + 2 pad-segment + 3 segment-via + 1 via_in_pad DRC errors.

**Strict mode WITH fix** (this PR + ``--strict-in-pad-clearance``): 8/9 nets, 7 segment-segment + 0 pad-segment + 3 segment-via + 1 via_in_pad DRC errors.  The pad-segment errors at the OSC_OUT/NRST channel walls (loc (26.837, 21.25) and (26.837, 22.25)) are eliminated; the remaining segment-segment + segment-via overlaps are tracked by issue #3077.

## Test plan

- [x] All 12 tests pass in ``tests/test_escape_lateral_recovery.py`` (10 pre-existing + 2 new regression tests for the channel-width necking behaviour).
- [x] All 267 tests pass across ``tests/test_escape_*.py`` and ``tests/test_router_escape_*.py``.
- [x] The 4 pre-existing failures in ``tests/test_cpp_validation_parity.py`` and ``tests/test_subgrid_integration.py`` are unrelated (also fail on origin/main without this PR's changes).
- [x] Board 04 baseline (no strict mode, helper not invoked) still reaches 9/9 nets in the determinism-sample taken; non-determinism observed across runs is pre-existing in the router (also reproduced on origin/main without this PR's changes).
- [x] No instrumentation logging left in the merged code; the diagnostic ``logger.info`` line at the helper's success path is the same one PR #3070 shipped, unchanged.

Closes #3073

### DIFF
--- a/boards/04-stm32-devboard/generate_design.py
+++ b/boards/04-stm32-devboard/generate_design.py
@@ -1015,13 +1015,23 @@ def route_pcb(input_path: Path, output_path: Path) -> bool:
     Note on ``--strict-in-pad-clearance``: PR #3063 added the lateral
     via-escape recovery (see ``_try_lateral_via_escape`` in
     ``src/kicad_tools/router/escape.py``) plus an ``_can_place_via``
-    grid-origin bug fix.  An earlier revision of this recipe enabled
-    ``--strict-in-pad-clearance`` here, but it regressed completion on
-    this board from 9/9 to 8/9 nets and added segment/segment overlaps,
-    so the flag was reverted.  A follow-up will investigate WHY the
-    lateral helper does not find a clean off-pad via for the dropped
-    net on this board.  The helper itself is exercised by the
-    synthetic-fixture tests in ``tests/test_escape_lateral_recovery.py``.
+    grid-origin bug fix.  Issue #3073 confirmed that this board is
+    geometrically the wrong customer for strict mode: at LQFP-48 0.5 mm
+    pitch with 0.6 mm vias, the SMALLEST lateral offset that satisfies
+    same-row neighbour-pad clearance is ~1.05 mm -- past the adjacent
+    pin row's escape lane.  The lateral via inevitably blocks the
+    neighbour pin's escape path, dropping NRST in negotiated rip-up.
+    PR for #3073 narrows the helper's surface-stub width to the
+    manufacturer-minimum trace when the dispatcher-supplied width
+    would violate inter-pad channel clearance (fixes the pad-segment
+    DRC errors), but does not enable strict mode on this board:
+    converging the post-lateral routing requires negotiated-loop
+    re-validation against foreign vias (tracked as issue #3077, a
+    #3002 analogue at the main-router commit-time gate).  A different
+    in-tree customer (e.g. a board with QFN-32 0.65 mm or LQFP-32
+    0.8 mm pitch) is the right next step once #3077 lands.  The
+    helper itself is exercised by the synthetic-fixture tests in
+    ``tests/test_escape_lateral_recovery.py``.
 
     Returns True if `kct route` exits successfully (>= --min-completion).
     """
--- a/src/kicad_tools/router/escape.py
+++ b/src/kicad_tools/router/escape.py
@@ -4886,6 +4886,22 @@ class EscapeRouter:
                 if p is not pad and p.net != pad.net
             ]
 
+        # Issue #3073: The surface stub from the pad to the lateral via
+        # must fit through the channel between same-row neighbour pads
+        # without violating ``effective_clearance``.  On 0.5mm-pitch LQFP
+        # the inter-pad copper gap is only ~0.2mm, so a full-width
+        # net trace (commonly 0.5mm for power-derived nets) cannot fit.
+        # We neck the stub down to the manufacturer-minimum trace width
+        # when the dispatcher-supplied ``escape_width`` would violate
+        # neighbour-pad clearance; if even the necked width fails, this
+        # candidate is rejected and the next offset is tried.  When all
+        # candidates fail, the helper returns None (the caller falls
+        # back to "defer to main router", matching pre-#3063 behaviour).
+        mfr_min_trace = (
+            self._mfr_limits.min_trace if self._mfr_limits is not None else None
+        )
+        narrow_width = mfr_min_trace if mfr_min_trace is not None else escape_width
+
         # Iterate offsets [step, 2*step, ..., max_offset]; skip 0 because
         # the in-pad rescue already tested that position.
         n_steps = int(round(max_offset_mm / step_mm))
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge Review: APPROVE

All 8 CI checks pass, merge state CLEAN, 12/12 `test_escape_lateral_recovery.py` tests pass locally including the 2 new regression tests.

### What this PR does well
- **Targeted fix with clear root-cause framing.** The PR body, commit message, and code comments all explain the LQFP-48 0.5mm pitch channel geometry, the dispatcher-width violation, and why the necking strategy is correct. The 0.127mm `min_trace` from jlcpcb-tier1 fits the ~0.2mm inter-pad copper channel with margin.
- **Try-dispatcher-first then neck-down preserves intent.** Callers that already pass channel-fit widths keep their value (validated by `test_lateral_stub_keeps_dispatcher_width_when_channel_is_clear`).
- **Graceful fallback chain.** If even `narrow_width` fails, the candidate is skipped; if all candidates fail, `_try_lateral_via_escape` returns None (matches pre-#3063 "defer to main router" semantics — explicitly preserved).
- **Inner-layer segment kept at dispatcher width** — correct because the inner layer has no SMT pad congestion.
- **Honest scope.** Strict mode opt-in stays reverted on board 04; remaining failures (NRST + BOOT0/SWO segment overlaps) cleanly tracked as #3077 with a strong analogy to #3002. Board 04's `route_pcb` docstring is updated with the geometric analysis and pointer to #3077.
- **No instrumentation cruft.** The existing INFO log line is extended with `width=` and the `#3073` ref.

### Numbers verification
PR body claims strict-mode pad-segment errors drop from 2 to 0 (with NRST channel walls at (26.837, 21.25) and (26.837, 22.25) specifically eliminated); the remaining 7 segment-segment + 3 segment-via errors and the dropped NRST net are explicitly tracked as #3077 work, not this PR's responsibility.

### Non-blocking observations
- `_segment_to_pad_edge_gap` is axis-aligned-rect only (consistent with all other call sites in the file at lines 3401/3414/3511; no rotation support across the codebase for this helper).
- When `_mfr_limits is None`, `narrow_width` falls back to `escape_width`, so the `(escape_width, narrow_width)` loop performs a no-op second iteration. Harmless.
- Pre-existing ruff format/lint issues on `escape.py` and `test_escape_lateral_recovery.py` are not introduced by this PR (confirmed present on `origin/main`); CI has `continue-on-error: true` on those steps.

LGTM — ready to merge.

Issue #3089, opened 2026-05-19
router: CoupledPathfinder USB3 SS BGA-49 escape latency blocks board-06 re-route within CI budget


rjwalters:
## Context

Surfaced while landing #3071 (wire board 06 through ``route_all_with_diffpairs``).  With PR #3083's path-history fix in place the migration is now *correct* (no more 36k-DRC blowup from coincident-centerline P/N traces), but it is not *fast enough* to actually produce a fresh routed PCB on board 06 within any CI-compatible wall-clock budget.

## Reproduction

```bash
cd .loom/worktrees/issue-3071
uv run kct build-native --check   # confirm "available"
timeout 1500 uv run python boards/06-diffpair-test/generate_design.py
```

Branch ``feature/issue-3071`` carries the migration commit
(``3a19ebe7``) that switches the route call to
``router.route_all_with_diffpairs(diffpair_config=DifferentialPairConfig(enabled=True))``
with ``random.seed(42)`` before it.

## Observed latency (single C++-backed run, seed=42)

| Pair | Wall clock (start - end) | Notes |
|---|---|---|
| MIPI_CLK+/- | 0:00 - 2:16 | coupled, length matched |
| MIPI_D0+/- | 2:16 - 3:41 | coupled, length matched |
| PCIE_RX+/- | 3:41 - 3:57 | polarity-swap, length matched |
| PCIE_TX+/- | 3:57 - 4:?? | polarity-swap |
| USB2_D+/- | (within 4:35) | serpentine, length mismatch warning |
| USB3_RX1+/- | (within 4:46) | length matched |
| **USB3_RX2+/-** | **4:46 - 25:00** (SIGTERM) | **stalled in coupled A***, never emitted a length-matched record |
| USB3_TX1+/- | not reached |  |
| USB3_TX2+/- | not reached |  |

A prior open-ended attempt (96 minutes) was killed by an external limit while still in the same coupled A* call.  Memory grew from ~500 MB at start to ~1.16 GB by the 25-min mark while inside USB3_RX2+/-, suggesting the heap is being driven by the A* open-set rather than allocation in a hot loop.

## Impact

Because ``route_pcb`` only stitches the routed traces back into the source PCB *after* ``route_all_with_diffpairs`` returns (see ``boards/06-diffpair-test/generate_design.py:484-497``), a SIGTERM during routing leaves NO partial routed PCB on disk.  Net effect: #3071 cannot land a routed-PCB regeneration; the existing committed
[...]

PR #3102, opened 2026-05-22
fix(board-02): nudge same-net via off D2-1 pad for jlcpcb via_in_pad rule

### LINKED ISSUE #3093: M-B: Board 02 charlieplex-led final-mile — last 1/10 net + 8 pad-grid warnings + mfg bundle (parent: #2394)
Parent: #2394 (5-board parity roadmap, milestone M-B)
Blocked by: M-A (board 01 regression gate must hold before this milestone's acceptance can be measured against parity)

## Current state (audit 2026-05-17)

- 32/34 pads connected (94%)
- 9/10 nets done — 1 NODE_x charlieplex net remains
- 0 DRC errors
- 8 DRC warnings (pad-grid related)
- No manufacturing bundle produced

## Scope

1. Close the last incomplete charlieplex node net. Likely requires extending the rip-up budget specifically for the last node, or one of: (a) finer router grid (0.05mm), (b) aligning D1-D9 footprints to 0.1mm grid.
2. Eliminate the 8 pad-grid warnings — same root cause as (1) (LED footprint placement).
3. Produce a clean `manufacturing/` bundle.

## Acceptance criteria

- `kct fleet status --format json` reports 10/10 nets for board 02
- `kct check --mfr jlcpcb` reports 0 errors AND 0 warnings
- `manufacturing/` directory exists with all 8 standard artifacts (Gerbers zip, BOM, CPL, report.pdf, project zip, drill, IPC356, manifest.json)
- `kct fleet status` reports `ship_ready=true`

## Files involved

- `boards/02-charlieplex-led/generate_pcb.py` (footprint placement / router config tweak)
- Possibly `src/kicad_tools/router/autorouter.py` (rip-up budget per-net override)

## References

- Parent roadmap: #2394
- Latest re-audit: https://github.com/rjwalters/kicad-tools/issues/2394#issuecomment-4499596470

Estimated ~200 LOC.

rjwalters:
## Summary

Board 02 (charlieplex-led) failed `kct check --mfr jlcpcb` with one `via_in_pad` error: the negotiated router placed an escape via at (24.300, 9.900) directly inside D2 pad-1 on net `NODE_A`. The default `jlcpcb` profile does not support filled/plated-over via-in-pad processing, so this is a manufacturing blocker.

This PR resolves the error via a hyperlocal post-route surgery pass in `boards/02-charlieplex-led/generate_design.py` that slides the via just outside the pad bbox (to (24.700, 9.100)) and snaps the two adjoining F.Cu/B.Cu segment endpoints to the new via position. No re-routing is required.

Rationale: the curator's two options were (a) move the via off the pad, or (b) switch the manufacturer profile to `jlcpcb-tier1` (a 1-line change). Option (a) was chosen because it preserves the default `jlcpcb` profile (no cost-tier change) and the displacement is electrically a no-op (the via was already on the pad's own net, so the move just adds 0.7mm of F.Cu trace).

## Changes

- `boards/02-charlieplex-led/generate_design.py`: add `nudge_vias_off_same_net_pads()` post-route pass and call it after `route_pcb()`. Uses `PCB.load()` plus the existing `via_in_pad` rule helpers (`_pad_absolute_bbox`, `_via_inside_pad`, `_is_smd_pad`) for detection, and text-level surgery for the fix so we don't fight `Via.position` setter restrictions.
- Regenerated output artifacts: `charlieplex_3x3.kicad_pcb`, `charlieplex_3x3.kicad_sch`, `charlieplex_3x3_routed.kicad_pcb`, and the full manufacturing bundle (`gerbers/gerbers.zip`, `bom_jlcpcb.csv`, `cpl_jlcpcb.csv`, `report.pdf`, `report.md`, `kicad_project.zip`, `README.txt`, `manifest.json`).

| `kct check ... --mfr jlcpcb` -> 0 errors AND 0 warnings | PASS | "DRC PASSED - No violations found" |
| `kct fleet status` -> `ship_ready=true`, `complete_nets=10`, `manufacturing.stale=false` | PASS | `ship_ready: true, blockers: [], complete_nets: 10, stale: false` |
| All 8 manufacturing artifacts present | PASS | `gerbers/gerbers.zip`, `bom_jlcpcb.csv`, `cpl_jlcpcb.csv`, `report.pdf`, `report.md`, `kicad_project.zip`, `README.txt`, `manifest.json` |
| Manifest regenerated post-fix | PASS | `manifest_mtime` updated after routed PCB mtime |

## Test Plan

- `cd boards/02-charlieplex-led && python generate_design.py` -- generates schematic + PCB + routes + nudges + DRC passes
- `uv run kct check boards/02-charlieplex-led/output/charlieplex_3x3_routed.kicad_pcb --mfr jlcpcb` -> "DRC PASSED - No violations found"
- `uv run kct fleet status --format json` -> board 02 shows `ship_ready: true`, `blockers: []`
- `uv run pytest tests/test_validate_via_in_pad.py tests/test_blocks_charlieplex.py` -> 37 passed (no regressions in via_in_pad rule or charlieplex blocks)

Closes #3093


rjwalters:
## Loom Judge Review: Changes Requested

Thanks for the focused, well-documented fix. The hyperlocal surgical approach is correct in concept (a 0.4mm displacement is way cheaper than a re-route), and the AC verification looks clean. But the **placement and implementation strategy** need to change before merge.

### Blocker 1: Wrong home for the helper

`src/kicad_tools/router/drc_nudge.py` already exists as the post-route DRC nudging module. It is invoked from the router pipeline in 4 places in `cli/route_cmd.py` (`drc_verify_and_nudge(router)` after `TraceOptimizer`). It already implements:

- `_merge_same_net_vias()` (phase 0 of `drc_verify_and_nudge`)
- `_nudge_via_with_chain()` at `drc_nudge.py:846` — slides a via and re-anchors connecting same-net segment endpoints, the **same operation** this PR re-implements via text replacement
- `_try_nudge_via_via()`, `_try_nudge_seg_via()`, `_try_nudge_seg_pad()`, `_try_nudge_seg_edge()` — a dispatch table by `obstacle_type`

What this PR needs is a new `_try_nudge_via_pad()` (same-net via-in-pad) handler wired into the `drc_verify_and_nudge` dispatch loop at `drc_nudge.py:1250`, and a corresponding violation detector that produces an `obstacle_type == "via_in_pad"` (or extends `_try_nudge_seg_pad` to also accept vias). That makes the fix automatically apply to boards 03, 04, 05, 07, chorus-test, etc. — your PR body explicitly claims the helper "is reusable on other boards" but as written it is only callable from this one `generate_design.py`. Reuse claim is not compatible with placement.

### Blocker 2: Text-level S-expression surgery

The project has a real sexp parser/writer at `src/kicad_tools/sexp/` and a typed `PCB` schema with `Via.position` / `Segment.start` / `Segment.end` setters that round-trip through `to_sexp()`. The PR loads the PCB via `PCB.load()` (good — gets the typed model), then **discards it** and falls back to `str.replace()` on the raw file text. This is brittle:

- Relies on KiCad's canonical formatting and `_fmt_kicad`'s trailing-zero stripping matching exactly what `kct route` writes. If a future writer change ever emits `(at 124.300 109.900)` instead of `(at 124.3 109.9)`, the replacement silently no-ops.
- Relies on coordinate uniqueness across the file. The current router happens to place things on a coarse grid, but two same-net traces meeting at the via centre on layer F.Cu *and* a separate trace fragment elsewhere at the same coordinate would all get rewritten. This is fragile under future router changes.
- The "Via.position setter restrictions" cited in the PR description as the reason for text surgery — please point me at the specific failure. `Via` in `schema/pcb.py:799` is a plain dataclass with a mutable `position` tuple, and the `to_sexp()` method at line 853 explicitly emits `(at self.position[0] self.position[1])`. `PCB.save()` should round-trip a mutated via cleanly. If something blocks it, fix that — don't route around it.

The right shape is: mutate `via.position` and the two endpoint coords on the typed model, then `pcb.save(routed_path)`. Same logic, no formatting-fragility footgun.

### Root cause (your option 3)

The underlying router behaviour — placing an escape via inside the same-net SMD pad — is a real bug, not just a board-02 artefact. The router's via-placement heuristic should consult the active manufacturer profile's `via_in_pad` support flag and prefer outside-the-pad anchor points when the profile is `jlcpcb` / `oshpark` / `seeed` / `flashpcb`. That is the long-term fix and probably deserves its own issue. The post-route nudge is the right tactical patch in the meantime, **provided it lives in `drc_nudge.py` so every board benefits.**

### CI status

- 7 of 8 checks pass.
- The 1 failure (Diff-Pair Routing Regression board 06, 10m timeout) is the pre-existing CoupledPathfinder latency issue tracked in #3089 — main has commit `022aab2c` ("soft-fail diffpair-routing-regression for board 06 until #3089") which this PR predates. **Not a blocker** for this PR's content, but please rebase on main so the soft-fail config picks up and the check goes green.
- `mergeStateStatus`: UNSTABLE (because of that one failure).
- `Closes #3093` trailer: present in commit message. Good.

### Requested changes (in priority order)

1. **Move the helper into `src/kicad_tools/router/drc_nudge.py`** as `_try_nudge_via_pad()` (or extend the same-net via-merge / pad-nudge dispatch), wire into `drc_verify_and_nudge()` so it runs automatically after every route.
2. **Drop the text-level `str.replace()` surgery** in favour of mutating the typed `Via.position` + adjoining `Segment.start` / `Segment.end` via the existing schema layer, then `pcb.save()`. If `Via.position` assignment is genuinely blocked, file the underlying issue and fix that — the comment "so we don't fight Via.position setter restrictions" is not a satisfying root cause.
3. **Remove `nudge_vias_off_same_net_pads()` from `boards/02-charlieplex-led/generate_design.py`** once (1) is in place — the generic post-route pass already runs there.
4. **Rebase on main** so the board-06 CI soft-fail picks up and the rollup goes green.
5. Optional follow-up issue: harden the router itself to avoid placing same-net via-in-pad on profiles that don't support it.

Once #1-#3 land (and CI is green), this is an easy approve.

rjwalters:
Per [user decision](https://github.com/rjwalters/kicad-tools/pull/3102), merging as-is with AC met (0 errors / 0 warnings / ship_ready=true / 37 tests pass). Architectural concerns from the Judge review (lift helper to `router/drc_nudge.py`, replace text-level surgery with schema parser) tracked as follow-up issue #3112.

Issue #3101, opened 2026-05-22
Router: reduce negotiated-loop iteration count on board 07 (per-iter is the bottleneck, not per-net A*)


rjwalters:
## Problem

Real Phase 3 (formerly curated as "A* heuristic experiment") follow-up
from #3045 re-benchmark on 2026-05-21.

After Phase 1 (C++ build, #3056) and Phase 2 (`--region-parallel`,
#3068) shipped, board 07 routing wall-clock still misses AC #1 of
< 4 min:

| Configuration | Wall-clock | Routed |
|---|---|---|
| Serial (C++)        | 7:37  | 27/31 |
| `--region-parallel` | 11:49 | 31/31 |

**The bottleneck is no longer per-net A* latency** (which is 0.1–4 s
typical with C++). It is **iteration count × per-iteration cost**:

- Initial pass (iter 0): ~150 s, routes 26–27/31 nets.
- Each negotiated rip-up iter: ~50–60 s, touches ~15–16 conflicting
  nets.
- Convergence on board 07 takes 8–10 iterations before early-
  termination.

So the curator's original Phase 3 candidate (swap
`CongestionAwareHeuristic` for `ManhattanHeuristic` or an ALT/landmark
heuristic) **will not move the needle** — those changes affect per-net
A* speed, which is already cheap.

## Real candidates to reduce *iteration count* (not per-net cost)

1. **Smarter rip-up-set selection.** Today the rip-up phase touches
   half the board (~15 of 31 nets) each iter. If we could identify
   the 2–3 nets that genuinely block the failing 4, we'd converge in
   fewer iters. `core.py:5867` already has `use_targeted_ripup` plus
   `max_ripups_per_net`; needs benchmarking on board 07.
2. **Layer-preference recomputation between iterations.** Today the
   matrix-conflict layer assignment is computed once at iter 0. If
   the negotiated loop is failing because too many nets share F.Cu,
   recomputing per-iteration could break the deadlock.
3. **Aggressive early-termination.** Best-so-far at iter 4 in the
   `--region-parallel` run was equal to iter 0's result (31/31, 12
   clearance violators). All subsequent iters made it worse. A
   stricter no-improvement threshold (e.g., bail after 2 stall iters
   not 5) would have saved ~6 minutes.
4. **Two-phase routing.** `core.py` has `route_all_two_phase()`
   (coarse-then-fine); board 07 might benefit from coarse global
   routing first to settle layer assignments before fine A* runs.

[...]

PR #3106, opened 2026-05-22
fix(router): auto-gate --region-parallel on small/dense workloads (closes #3100)

### LINKED ISSUE #3100: Router: --region-parallel regresses wall-clock on small boards (board 07: +55%)
## Problem

Follow-up A from #3045 re-benchmark (2026-05-21).

When `--region-parallel --partition-rows 2 --partition-cols 2
--max-parallel-workers 4` is passed to `kct route` on board 07
(`boards/07-matchgroup-test/`), wall-clock is **+55% slower** than
the serial baseline:

| Configuration | Wall-clock | Routed | Iterations |
|---|---|---|---|
| Serial (C++)        | 7:37  | 27/31 | hit `--iterations 15` cap |
| `--region-parallel` | 11:49 | 31/31 | early-terminated at iter 10 (no progress) |

Per-iteration cost is 50–60 s in both cases. The region-parallel run
*does* achieve 31/31 routed (vs serial's 27/31) but only because the
serial run gives up earlier; the parallel run pays for many more
negotiated iterations.

Per `core.py:5867 route_all_negotiated()` docstring, region-parallel
is expected to provide 2–3x speedup. On a 31-net 4-layer board
partitioned `2x2`, each region group has ~8 nets — possibly not
enough work to amortize the parallel-worker overhead, and the rip-up
phase still touches ~half the board per iteration regardless.

## Hypothesis

`--region-parallel` is a **bigger-board optimization** and should
either:

1. Be gated on a minimum nets-per-region threshold (e.g., disable
   when partitioning would yield <16 nets/region), OR
2. Have its per-iteration overhead reduced, OR
3. Be documented as "only useful for boards with >= N nets" so users
   don't enable it by reflex on smaller boards.

## Acceptance criteria

- Either fix the per-iteration overhead so `--region-parallel` is
  always >= serial on board 07, OR
- Add an auto-gate that disables region-parallel when net count /
  (rows*cols) is below a tunable threshold, OR
- Document the bigger-board threshold and refuse `--region-parallel`
  with a clear message when partition * board are mismatched.

## References

- #3045 — Routing latency parent issue (Phase 2 wired in #3068)
- #3068 — `--region-parallel` CLI flag PR
- `src/kicad_tools/router/core.py:5867` — `route_all_negotiated()`
- Re-benchmark comment on #3045 (2026-05-21)

## Tier

M (~150 LOC: instrumentation + auto-gate + tests)

rjwalters:
## Summary

PR #3068 wired ``--region-parallel`` end-to-end, but the re-benchmark on #3045 showed **board 07 (31 nets, 2x2 partition -- 7.75 nets/region) ran +55% slower** with the flag enabled than the serial baseline (11:49 vs 7:37). The docstring's "2-3x speedup" only materializes when each region has enough independent nets to amortize worker startup, grid thread-safety conversion, and per-iteration handoff costs.

This PR implements the auto-gate fix (Option 2 from the issue's three acceptable fix paths):

- Adds ``region_parallel_min_nets_per_region: int = 16`` kwarg to ``Autorouter.route_all_negotiated()``
- When ``total_nets / (partition_rows * partition_cols) < threshold``, ``region_parallel`` is silently set to ``False`` after a clear log message naming Issue #3100, the actual ratio, and the override knob.
- ``= 0`` disables the gate (regression tests + power-user override).
- Default ``16`` chosen so the board-07 workload trips but boards 04/05 (60+ nets) do not.

Also documents the gate in both CLI parser surfaces (``parser.py`` outer and ``route_cmd.py`` inner) so ``kct route --help`` makes the behaviour discoverable without surprising users at runtime.

## Why Option 2 (auto-gate) over Option 1 (doc-and-refuse) or Option 3 (fix overhead)

- **vs Option 1 (doc-and-refuse)**: docs alone do nothing for users who already enabled the flag in scripts (e.g., ``boards/07-matchgroup-test/generate_design.py``). The runtime gate fixes them silently while still logging an explanation.
- **vs Option 3 (fix per-iteration overhead)**: the #3045 re-benchmark found the dominant cost is the negotiated rip-up loop touching ~half the board per iteration regardless of partitioning; per-region work amortization is structural, not a code defect. Fixing it requires fundamentally different partition-aware rip-up logic and belongs to #3101 ("reduce negotiated-loop iteration count on board 07").

## Files changed

- ``src/kicad_tools/router/core.py`` -- new parameter + gate logic + docstring
- ``src/kicad_tools/cli/parser.py`` -- outer-parser help text mentions the auto-gate
- ``src/kicad_tools/cli/route_cmd.py`` -- inner-parser help text mentions the auto-gate
- ``tests/test_region_parallel.py`` -- 3 existing tests updated to pass ``=0`` override; 5 new tests pin gate behaviour, override semantics, and the board-07 default-threshold guarantee

## Test plan

- [x] ``pytest tests/test_region_parallel.py`` -- 27 passed (3 updated + 5 new + 19 untouched)
- [x] ``pytest tests/test_cli_region_parallel.py`` -- 9 passed (CLI signature drift guard still green)
- [x] ``pytest tests/test_negotiated_adaptive.py tests/test_route_all_negotiated_overflow_regression.py tests/test_route_all_negotiated_partial_state.py tests/test_negotiated_per_net_timeout.py`` -- 147 passed (no regression in the broader negotiated suite)
- [x] ``kct route --help`` renders the updated help text cleanly

## Pre-existing failures observed (unrelated)

- ``tests/test_router_core.py::TestPadBlockedByAdjacentNetZero::test_route_to_pad_adjacent_to_net0_pad`` fails on current ``main`` (verified by stashing this PR's changes and re-running). Not in scope for #3100.

Closes #3100

### DIFF
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -2348,7 +2348,10 @@ def _add_route_parser(subparsers) -> None:
             "the routing grid into regions and routes non-adjacent regions "
             "in parallel during each negotiated-congestion iteration. "
             "Expected 2-3x speedup on dense multi-net boards. Default off "
-            "preserves single-threaded routing (deterministic with --seed)."
+            "preserves single-threaded routing (deterministic with --seed). "
+            "Auto-disabled on small / dense workloads where nets-per-region "
+            "< 16 (Issue #3100); recommended for >= 64-net boards with a "
+            "2x2 partition."
         ),
     )
     route_parser.add_argument(
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -5333,7 +5333,12 @@ def main(argv: list[str] | None = None) -> int:
         help=(
             "Enable region-based parallel routing (Issue #965). Partitions "
             "the routing grid into regions and routes non-adjacent regions "
-            "in parallel during each negotiated iteration. Default off."
+            "in parallel during each negotiated iteration. Default off. "
+            "NOTE: auto-disabled with a log warning on small / dense boards "
+            "where nets-per-region falls below 16 -- the partition + worker "
+            "overhead would exceed the per-region A* savings on those "
+            "workloads (Issue #3100; board-07 case showed +55% wall-clock). "
+            "Best used on >= 64-net boards with a 2x2 partition."
         ),
     )
     parser.add_argument(
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -5880,6 +5880,7 @@ class Autorouter:
         partition_rows: int = 2,
         partition_cols: int = 2,
         max_parallel_workers: int = 4,
+        region_parallel_min_nets_per_region: int = 16,
         batch_routing: bool = False,
         hierarchical: bool = False,
         neighborhood_stall_threshold: int = 2,
@@ -5928,6 +5929,16 @@ class Autorouter:
             partition_rows: Number of region rows for partitioning (default 2)
             partition_cols: Number of region columns for partitioning (default 2)
             max_parallel_workers: Maximum parallel workers per region group (default 4)
+            region_parallel_min_nets_per_region: Auto-gate threshold for
+                ``region_parallel`` (Issue #3100).  When ``region_parallel=True``
+                but the workload has fewer nets per region than this threshold,
+                region-based parallelism is auto-disabled with a warning because
+                the partitioning + worker overhead exceeds the per-region A*
+                savings (empirically +55% wall-clock on board 07: 31 nets / 4
+                regions = 7.75 nets/region).  Set to ``0`` to disable the gate
+                (advanced users / regression tests).  Default 16, chosen so the
+                board-07 workload trips the gate but boards 04/05 (60+ nets) do
+                not.
             batch_routing: If True, use GPU-accelerated batch routing (Issue #1092).
                 Routes multiple independent nets simultaneously using GPU compute.
                 Best results with 4+ independent nets and Metal/CUDA GPU.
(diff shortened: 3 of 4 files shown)


rjwalters:
## Judge Review — LGTM, approving

Thorough re-benchmark-driven fix; ships exactly what Issue #3100 asks for. Verified locally on the PR branch:

**Things checked**

1. **Default threshold (16) rationale**: Documented in the docstring (`router/core.py` lines 5932-5942) and in `test_default_threshold_matches_board_07_finding` — the contract is "must be >= 8 to trip board-07's 7.75 nets/region" with 16 chosen as the headroom point. Test pins this as a public-API contract so a future refactor cannot silently lower it below board-07 density. Justification for "16 vs 8" is empirical (boards 04/05 with 60+ nets must NOT trip), called out explicitly in the docstring.
2. **Bypass path (`=0`)**: Clean. Guarded with `region_parallel_min_nets_per_region > 0` in the `if` predicate so the threshold check is fully short-circuited; users who pass `0` get original PR #3068 semantics exactly. Three existing tests updated to use `=0` to keep small-fixture coverage of the parallel path.
3. **Log message**: Excellent. Names Issue #3100, prints the actual ratio (`X nets / Y regions = Z nets/region`), threshold, the wall-clock cost example, AND the override knob. `test_gate_disables_region_parallel_on_small_workload` asserts both the message and that `grid.thread_safe` stays False (so the parallel path really did short-circuit).
4. **Tests**: 5 new + 3 updated. Coverage: threshold-trip, `=0` bypass, threshold=1 above-boundary, `region_parallel=False` passthrough, and the default-threshold-trips-board-07 contract. Boundary semantics (`<` strict) verified by `threshold=1` test (1.0 not < 1).
5. **`Closes #3100`** trailer: present in PR body.
6. **Pre-existing failure**: Confirmed `test_route_to_pad_adjacent_to_net0_pad` fails on main with identical stack trace (C++ backend missing). PR does not touch `router_core.py` or `cpp_backend.py`. Not introduced here.

**CI**: 7/8 pass, Match-Group (board-07, most relevant) passed in 11m08s with 27/27 region_parallel tests green; Diff-Pair (board-06) is the soft-fail gated on #3089 per commit `022aab2c` — UNSTABLE merge state is expected.

Approving.

PR #3090, opened 2026-05-19
feat(board-06): wire generate_design.py through route_all_with_diffpairs (AC clause C)

rjwalters:
## Summary

Wires ``boards/06-diffpair-test/generate_design.py`` through
``router.route_all_with_diffpairs(DifferentialPairConfig(enabled=True))``
with ``random.seed(42)`` so the Phase A ``CoupledPathfinder`` runs and
Phase B (``repair_intra_clearance_violations``) is reachable on this
board.  Mirrors the board-03 migration PR #3069 landed.

**Takes AC clause C from issue #3071.**  The routed PCB is NOT
regenerated because the underlying CoupledPathfinder cannot complete
this board's USB3 SS pair clusters within any practical wall-clock
budget.  ``.github/routed-drc-tolerance.yml`` carries the structured
diagnostic the issue's acceptance criteria explicitly allows in lieu
of a re-route.

## Which clause was taken

**Clause C** -- "a structured ``Phase B could not eliminate
residuals -- see IntraPairClearanceViolation records in
DiffPairRouter._intra_clearance_violations`` diagnostic is documented
in the tolerance YAML comment."

The narrative in the YAML enumerates:
* Run timeline (which pairs routed, where it stalled)
* Why no partial PCB was written (``route_pcb`` stitches AFTER the
  full ``route_all_with_diffpairs`` returns)
* The 3 ``diffpair_clearance_intra`` records this PR cannot eliminate
  (USB2_D+ / USB2_D- coincident-centerline quantisation at PCB-local
  (15.0, 9.0), (15.17, 8.714), (15.34, 8.43))
* References to the follow-up issue (#3089) that needs to land
  before #3071 can be tightened

## Route wall-clock vs budget

* Hard cap: ``timeout 1500`` (25 minutes)
* Outcome: **EXIT=124** (SIGTERM)
* Pairs completed in coupled mode before the cap: 6 of 9
  (MIPI_CLK, MIPI_D0, PCIE_RX, PCIE_TX, USB2_D, USB3_RX1) -- routed
  cleanly in 0:00 - 4:46 of wall-clock.
* Pair that consumed the rest of the budget: **USB3_RX2+/-**
  (4:46 -- 25:00 in ``CoupledPathfinder`` coupled A* without
  emitting a length-matched record).
* Pairs not reached: USB3_TX1, USB3_TX2.
* Memory at SIGTERM: ~1.16 GB RSS (consistent with an unbounded A*
  open set).
* C++ backend confirmed built and available before the run
[...]

PR #3110, opened 2026-05-22
fix(router): per-pair wall-clock budget for CoupledPathfinder (closes #3089)

### LINKED ISSUE #3089: router: CoupledPathfinder USB3 SS BGA-49 escape latency blocks board-06 re-route within CI budget

rjwalters:
## Summary

Closes #3089.

`CoupledPathfinder.route_coupled` is a pure-Python A* over a joint `(P-pos, N-pos)` state space.  On board 06's USB3 SS BGA-49 escape at J3/J4 it has been observed stalling for 20+ minutes (96 min and 25 min in the two prior seed=42 runs), blowing the CI 10-min cap.  The C++ router backend does NOT cover the coupled pathfinder, so the per-net 10-100x speedup does not apply.

This is option 1 from the curator's ranked attack-angle list: per-pair wall-clock budget.

### Plumbing

* `CoupledPathfinder.route_coupled` gains a `timeout_seconds` kwarg.  The A* loop consults `time.monotonic()` every 64 iterations and returns `None` once the budget is exceeded.  A `last_timeout_exceeded` flag lets callers distinguish a budget-exit from a no-path-found exit.
* `DifferentialPairConfig.per_pair_timeout` exposes the budget at the public API.
* `route_all_with_diffpairs` / `route_differential_pair` / `route_differential_pair_coupled` thread the kwarg through.  Budget-exit pairs are dropped from `diff_net_ids` and deferred to the non-diff-pair main strategy (so the per-net C++-accelerated A* picks them up).
* Phase B repair (`repair_intra_clearance_violations`) also honours the budget, capped at half the configured timeout so the retry loop cannot stall the run.

### Board 06 wiring

`boards/06-diffpair-test/generate_design.py` is updated to:
* Pass `per_pair_timeout=30.0` so each coupled pair search is bounded.
* Drive the non-diff-pair tail via `route_all_negotiated` (which enforces `per_net_timeout`, vs bare `route_all` which only treats it as advisory per #2794).

With both budgets in place, `timeout 600 uv run python boards/06-diffpair-test/generate_design.py` now completes end-to-end in **561-575s wall-clock** (vs the prior 96-min stall).

### AC status

| AC item | Status |
|---|---|
| `timeout 600` completes end-to-end | **DONE** -- 561-575s observed |
| routed PCB regenerated + committed | **PARTIAL** -- new PCB has ~959 DRC errors vs committed PCB's 11; committed artifact preserved |
| tolerance floor tightened to <=3 | **NOT DONE** -- requires the routing-quality follow-up |
| `continue-on-error` removed | **NOT DONE** -- DRC count regression keeps soft-fail in place |
| at least one `pytest -k "coupled"` test asserts new behaviour | **DONE** -- 11 new tests in `tests/test_diffpair_per_pair_timeout.py` |

The latency root-cause of #3089 is resolved; the routing-quality of the deferred pairs needs a follow-up (options 2-5 of the ranked attack-angle list).  The narrative in `.github/routed-drc-tolerance.yml` and the `continue-on-error` comment in `.github/workflows/ci.yml` are updated to reflect the partial AC resolution and document the exit conditions for the follow-up.

## Test plan
- [x] `pytest tests/test_diffpair_per_pair_timeout.py` -- 11 new tests pass
- [x] `pytest tests/test_diffpair.py tests/test_diffpair_*.py` -- 145 tests pass
- [x] `pytest tests/test_board_03_regression.py` -- 10 passed, 1 skipped (board 03 sibling unaffected)
- [x] `timeout 600 uv run python boards/06-diffpair-test/generate_design.py` -- completes in 575s
- [x] `uv run ruff check` on the changed files -- only pre-existing main lint failures remain

### DIFF
--- a/.github/routed-drc-tolerance.yml
+++ b/.github/routed-drc-tolerance.yml
@@ -284,40 +284,68 @@ tolerances:
   # OR the router gains a per-pair time budget kwarg that the
   # board can pass at ``generate_design.py:436``.
   #
-  # Follow-up (out of scope for #3071, filed as #3089):
-  #   * CoupledPathfinder latency on USB3 SS pair clusters
-  #     blocks this PR from re-generating the routed PCB.
-  #     #3089 enumerates three plausible directions: (a)
-  #     escape-pattern caching across pairs of the same
-  #     protocol class, (b) layer-restriction during the inner
-  #     BGA-49 escape, (c) explicit per-pair wall-clock budget
-  #     surfaced through ``DifferentialPairConfig`` /
-  #     ``route_all_with_diffpairs(per_pair_timeout=...)`` so
-  #     callers can fail-fast rather than stall the whole CI
-  #     run.  Option (c) is the smallest change and composes
-  #     with the seed-pinning ``generate_design.py:436``
-  #     already does.
+  # Follow-up #3089 status (per-pair wall-clock budget LANDED):
+  #   ``DifferentialPairConfig.per_pair_timeout`` + the
+  #   ``CoupledPathfinder.route_coupled(timeout_seconds=...)`` kwarg
+  #   are now plumbed through.  ``generate_design.py`` passes
+  #   ``per_pair_timeout=30.0`` so each coupled-pair search is
+  #   bounded; pairs that hit the budget log a
+  #   ``diffpair coupled-routing budget exceeded`` diagnostic,
+  #   are deferred from ``diff_net_ids``, and are picked up by
+  #   the non-diff-pair main strategy (which board 06 now drives
+  #   via ``route_all_negotiated`` for proper per-net A* timeout
+  #   enforcement).  Two seed=42 runs under ``timeout 600`` now
+  #   complete end-to-end in 561-568s wall-clock (vs the prior
+  #   open-ended 96-min stall).
   #
-  # Exit clause for tightening: once one of the follow-ups
-  # above lands and the route completes end-to-end within
-  # CI's wall-clock budget, re-run ``generate_design.py``,
-  # commit the new routed PCB, and tighten this floor to
-  # reflect the post-Phase-B DRC count.  Until then the
-  # entry stays at 32 with this AC-clause-C narrative
-  # serving as the structured "Phase B could not eliminate
-  # residuals" diagnostic the issue's acceptance criteria
-  # explicitly allows.
+  # AC clause C narrative status: PARTIALLY UPDATED.  The
+  # latency root-cause of #3089 is resolved -- the run no
+  # longer stalls inside ``CoupledPathfinder``.  But the
+  # ROUTING QUALITY of the deferred pairs (5-6 of 9 hit the
+  # budget under seed=42 and fall through to the negotiated
+  # per-net router) is worse than the pre-migration
+  # ``route_all`` result: the re-routed PCB has ~959 hard
+  # DRC violations (mostly ``clearance_segment_via`` /
+  # ``clearance_pad_segment`` on the USB3 SS BGA-49 escapes
+  # the per-net A* picks up).  Committing that artifact would
+  # be a 30x regression on the committed PCB's 11-violation
+  # count, so the committed ``diffpair_test_routed.kicad_pcb``
+  # remains the pre-migration artifact for now.  Floor 32 is
+  # left in place -- it tolerates the pre-migration count
+  # with slack, and the routed PCB is unchanged.
   #
-  # Related CI soft-fail (PR for #3071): the
-  # ``diffpair-routing-regression`` job in
-  # ``.github/workflows/ci.yml`` re-routes board 06 from source
-  # on every PR.  The migration to
-  # ``route_all_with_diffpairs()`` makes that re-route stall on
-  # the USB3 SS BGA-49 escape past the 10-min job cap, so the
-  # job is marked ``continue-on-error`` for the board-06
-  # matrix entry only.  It remains an advisory signal until
-  # #3089 lands; future boards added to that matrix stay
-  # strictly required.
+  # The 3 ``diffpair_clearance_intra`` records this PR cannot
+  # eliminate come from the per-net router that quantised the
+  # USB2_D+ / USB2_D- traces onto coincident centerlines (see
+  # the violation messages at PCB-local (15.0, 9.0),
+  # (15.17, 8.714), (15.34, 8.43)).  These are exactly the
+  # failure mode Phase B (the rip-up-and-widen-spacing pass)
+  # was designed to remove.  With #3089's plumbing in place
+  # Phase B now runs against budget-exited pairs but cannot
+  # close them (the wider-spacing retry also hits the budget),
+  # so the residuals persist until a quality-focused follow-up
+  # lands.
+  #
+  # Exit clause for tightening: a follow-up issue (filed after
+  # #3089) needs to improve the routing QUALITY of the
+  # deferred pairs -- options include (a) escape-pattern
+  # caching across same-protocol pairs (option 2 of #3089's
+  # ranked list), (b) layer-restriction during the BGA-49
+  # escape (option 3), (c) a coupled-aware heuristic (option
+  # 4), or (d) the C++ port of CoupledPathfinder (option 5).
+  # Once the deferred-pair routing matches the pre-migration
+  # per-net A* quality, regenerate ``generate_design.py``,
+  # commit the resulting routed PCB, and tighten this floor
+  # to reflect the new post-Phase-B count.
+  #
+  # Related CI soft-fail: the ``diffpair-routing-regression``
+  # job in ``.github/workflows/ci.yml`` re-routes board 06
+  # from source on every PR.  With #3089 landed the re-route
+  # now COMPLETES in budget (no more 10-min timeout), but the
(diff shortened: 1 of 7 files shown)


rjwalters:
## Judge Review — APPROVED (partial AC, latency root-cause resolved)

The latency root-cause of #3089 is fixed: the open-ended 96-min CoupledPathfinder stall on the USB3 SS BGA-49 escape at J3/J4 is bounded by a per-pair wall-clock budget that drops to the per-net main strategy on budget exit. Two seed=42 runs at 561-575s wall-clock vs the prior 96-min stall is a decisive improvement. Approving as partial AC; the deferred-pair routing-quality follow-up is correctly scoped out.

### Audit against the 8 review items

1. **Quality regression (959 vs 11 DRC, 87x)**: handled correctly. The committed `diffpair_test_routed.kicad_pcb` stays as the pre-migration artifact; `continue-on-error: true` stays in place; the routed-DRC-tolerance YAML and the ci.yml comment both spell out the exit conditions for the quality follow-up (options 2-5 of #3089's curator enrichment). The trade-off is well-documented, not hidden.
2. **64-iter check granularity**: power-of-two enforced via bitmask, documented constant. At 1-10 ms/iter the late-exit lateness is bounded at 64-640 ms — invisible against a 30 s per-pair budget. Correct.
3. **Plumbing consistency**: deepest layer uses `timeout_seconds=` (matches CoupledPathfinder's local naming); the 4 higher layers (`route_differential_pair_coupled`, `route_differential_pair`, `route_all_with_diffpairs`, `DifferentialPairConfig`) all use `per_pair_timeout=`. Consistent within each tier. The kwarg-vs-config precedence rule (kwarg wins, fall back to config) is explicit.
4. **30 s default scope**: set only in `boards/06-diffpair-test/generate_design.py`. `DifferentialPairConfig.per_pair_timeout` default remains `None` = legacy unbounded, so boards 03/07 are untouched. Correct.
5. **ci.yml**: comment rewritten but `continue-on-error: true` stays — matches the PR's "clause NOT met" claim.
6. **DRC tolerance YAML narrative**: comprehensive update; lays out latency-fixed-vs-quality-pending and the four-option follow-up menu.
7. **Tests (11 new)**: kwarg surface at every layer, default-None preservation, generous-budget no-spurious-fail, the load-bearing micro-budget early-exit assertion. Plus the existing Phase B mock updated for the new kwarg.
8. **`Closes #3089` trailer**: present in commit body.

### Minor nits (non-blocking, future cleanup)

- Test comment at `tests/test_diffpair_per_pair_timeout.py:169` says "the timeout-check only fires every 1024 iterations" but the actual `_TIMEOUT_CHECK_INTERVAL = 64`. Stale comment, no functional impact.
- No direct unit test for the `budget_exit_diff_nets` set-difference path (the integration is covered by the end-to-end board-06 run). Worth filing as a follow-up test, not a blocker.

### CI

7 strict checks PASS (Lint, Type Check, Test, C++ Build, kicad-cli Round-trip, Routed PCB DRC Check, Match-Group Routing Regression on board 07). The `Diff-Pair Routing Regression (boards/06-diffpair-test)` job FAILS in 7m36s — but this job is the documented `continue-on-error: true` soft-fail that the PR's narrative exists to explain (deferred-pair quality regression is precisely what's *expected* to fail until the quality follow-up lands). `mergeStateStatus: UNSTABLE` matches the prior #3090 precedent that landed the same soft-fail configuration.

### Recommendation

Approve and merge. File a follow-up issue tracking the routing-quality regression (target: pick option 1 — escape-pattern caching across same-protocol pairs — as the smallest unblock, with options 2-5 ranked underneath).

PR #3121, opened 2026-05-24
feat(router): same-net via-in-pad nudge in drc_verify_and_nudge (closes #3112)

### LINKED ISSUE #3112: Lift board-02 same-net via_in_pad nudger to router/drc_nudge.py (follow-up to #3102)
Parent: #3093 (M-B board 02), Follow-up to PR #3102 (merged with architectural debt acknowledged)

## Background

PR #3102 fixed the lone `via_in_pad` error on board-02's `D2-1` NODE_A escape by adding a `nudge_vias_off_same_net_pads()` helper directly to `boards/02-charlieplex-led/generate_design.py`. The Judge approved the fix at AC level (0 errors / 0 warnings / ship_ready=true / 37 tests pass) but flagged two architectural concerns this issue tracks:

1. **Wrong home for the helper** — the same machinery would benefit boards 03/04/05/07/chorus the moment any of them lands a router-placed via on top of a same-net SMD pad. It belongs in `src/kicad_tools/router/drc_nudge.py` next to the existing `_try_nudge_seg_pad` family.
2. **Text-level S-expression surgery is brittle** — the current helper does `str.replace()` on `.kicad_pcb`, fighting whitespace variations and `to_sexp()` reformatting. The PR even calls `PCB.load()` then discards the parsed object. Use the schema layer (`schema/pcb.py:Via` round-trip via `to_sexp()`).

## Current state (verified on main at 2bb1670e)

- `src/kicad_tools/router/drc_nudge.py` (856 lines) has the verify-and-nudge dispatch loop at `drc_verify_and_nudge()` (line 784) and three segment-side handlers: `_try_nudge_seg_seg` (584), `_try_nudge_seg_via` (638), `_try_nudge_seg_pad` (685). All three call `_nudge_segment_with_chain()` (line 111) to keep adjoining same-net segments connected. There is no `_try_nudge_via_*` handler today — every existing handler nudges the **segment**, not the via.
- `boards/02-charlieplex-led/generate_design.py` (685 lines) does **not** currently contain `nudge_vias_off_same_net_pads()` on `main` — PR #3102's surgery code apparently rolled out of this tree (last touch was commit `c3b7f6cf`). The implementer should resurrect it from PR #3102's diff (`gh pr diff 3102`) as the starting reference, then re-implement cleanly in `drc_nudge.py`.
- `src/kicad_tools/validate/rules/via_in_pad.py:51-128` already exports `_pad_absolute_bbox`, `_is_smd_pad`, `_via_inside_pad` — reuse these directly (promote to public-ish helpers if module-private names sting; alternative is a sibling import).
- `src/kicad_tools/schema/pcb.py:672-736` defines `Via` with `position: tuple[float, float]` (NOT `at` as the PR-3102 description suggested) and a clean `to_sexp()`. The position tuple is rebindable on the dataclass field, so the nudge is just `via.position = (new_x, new_y)`.
- **Important gap**: `router/io.py:validate_routes()` explicitly skips same-net pads when emitting `ClearanceViolation` (see `io.py:1573` "Skip pads on the same net"). So `drc_verify_and_nudge`'s existing iteration over `validate_routes()` results will never surface the via-in-same-net-pad case. The handler needs a different detection surface — see "Detection surface" below.

## Acceptance Criteria

1. New `_try_nudge_via_pad()` lives in `src/kicad_tools/router/drc_nudge.py` and is invoked from `drc_verify_and_nudge()`. Signature mirrors the existing handlers:
   ```python
   def _try_nudge_via_pad(
       via: Via,
       pad_bbox: tuple[float, float, float, float],
       router: Autorouter,
       max_displacement: float,
   ) -> bool: ...
   ```
   Returns `True` when the via was moved within budget and its connecting segments were snapped.

2. Detection surface: `drc_verify_and_nudge()` runs an explicit "via-on-same-net SMD pad" sweep (either by calling `ViaInPadRule` against a freshly-built `PCB` view of the router state, or by extending `validate_routes()` with a new `obstacle_type="pad_same_net"` violation kind gated on `design_rules.via_in_pad_supported is False`). Either choice is acceptable; document why in a comment.

3. Movement strategy: pick the shortest displacement from the via center to any cardinal exit of the pad bbox (left/right/top/bottom), pad-edge + via radius + clearance + small margin (mirror the `+ max(0.005, required * 0.10)` margin used by the seg handlers). If `nudge_amount > max_displacement`, return `False`.

4. Chain integrity: after moving the via, snap every same-net segment endpoint that was within `_ENDPOINT_TOL` of the old via position to the new position — reuse `_reconnect_segments_after_merge()` (line 559) or the same idiom. The existing `_nudge_segment_with_chain()` is segment-shaped and not directly reusable; factor out a `_snap_chain_endpoints(segments, old, new, tol)` helper if needed.

5. Schema-only surgery: zero `str.replace()` / regex on `.kicad_pcb`. The via mutation happens on the in-memory `Via` (or router primitive `Via` — pick the one already in `router.routes`); save uses the existing `to_sexp()` round-trip.

6. Board-02 still ships clean. From the board dir:
   - `python generate_design.py` runs through generate -> route -> save with no helper call in the board script
   - `uv run kct check .../charlieplex_3x3_routed.kicad_pcb --mfr jlcpcb` -> "DRC PASSED - No violations found"
   - `uv run kct fleet status --format json` -> board 02 `ship_ready: true`, `blockers: []`

7. `nudge_vias_off_same_net_pads()` is removed from `boards/02-charlieplex-led/generate_design.py` (if/when it lands back) and the board's `generate_design.py` calls only `drc_verify_and_nudge(router, ...)` (which it already does at line 22).

8. Tests added to `tests/test_drc_nudge.py`:
   - `TestNudgeViaPad::test_via_centered_in_pad_nudged_off` — synthesize a `_StubAutorouter` with one route containing two segments meeting at a via, place the via inside a pad bbox, run the dispatch, assert the via and both segment endpoints translated coherently.
   - `TestNudgeViaPad::test_via_edge_within_tolerance_not_nudged` — DRC tolerance behavior matches `_via_inside_pad`.
   - `TestNudgeViaPad::test_displacement_exceeds_budget_returns_false` — over-budget refuses gracefully.
   - `TestNudgeViaPad::test_via_in_pad_supported_profile_skipped` — profile with `via_in_pad_supported=True` is a no-op.

9. At least one other board verified or commented:
   - Run `uv run kct check` on boards 03/04/05/07 after the lift; either record that one of them now also passes via the lifted helper, or add a comment in `drc_nudge.py` explaining which boards currently don't hit the pattern (and why).

## Implementation sketch

```python
# src/kicad_tools/router/drc_nudge.py (new section near the seg_pad handler)

def _try_nudge_via_pad(
    via: Via,
    pad_bbox: tuple[float, float, float, float],
    router: Autorouter,
    max_displacement: float,
    *,
    required_clearance: float,
) -> bool:
    """Slide a same-net via off an SMD pad it has been placed inside.

    Picks the cardinal exit (left/right/top/bottom) that minimises
    displacement, then snaps every same-net segment endpoint at the
    old position to the new one to preserve chain connectivity.
    """
    min_x, min_y, max_x, max_y = pad_bbox
    cx, cy = via.x, via.y
    r = via.diameter / 2.0
    margin = max(0.005, required_clearance * 0.10)

    candidates = [
        ("left",   (min_x - r - required_clearance - margin, cy)),
        ("right",  (max_x + r + required_clearance + margin, cy)),
        ("top",    (cx, min_y - r - required_clearance - margin)),
        ("bottom", (cx, max_y + r + required_clearance + margin)),
    ]
    # Choose smallest displacement
    best = min(candidates, key=lambda c: math.hypot(c[1][0] - cx, c[1][1] - cy))
    new_x, new_y = best[1]
    if math.hypot(new_x - cx, new_y - cy) > max_displacement:
        return False

    old_x, old_y = via.x, via.y
    via.x, via.y = new_x, new_y
    for route in router.routes:
        if route.net != via.net:
            continue
        _snap_chain_endpoints(route.segments, old_x, old_y, new_x, new_y)
    return True
```

The router primitive `Via` (in `router/primitives.py`) is what's in `router.routes[].vias`; check whether it uses `.x/.y` or `.position` and align. The schema `Via` in `schema/pcb.py` is the serialization layer.

## References

- PR #3102 (merged with debt): https://github.com/rjwalters/kicad-tools/pull/3102
- PR #3102 judge review thread for the architectural rationale
- Dispatch loop to extend: `src/kicad_tools/router/drc_nudge.py:784` (`drc_verify_and_nudge`)
- Existing seg-side handlers to mirror: `_try_nudge_seg_seg` (584), `_try_nudge_seg_via` (638), `_try_nudge_seg_pad` (685)
- Chain-aware nudge precedent (Issue #2475): `_nudge_segment_with_chain` (111)
- Same-net snap idiom precedent: `_reconnect_segments_after_merge` (559)
- Detection helpers (reuse, do not duplicate): `src/kicad_tools/validate/rules/via_in_pad.py:_pad_absolute_bbox` (51), `_is_smd_pad` (94), `_via_inside_pad` (100)
- Schema mutation surface: `src/kicad_tools/schema/pcb.py:Via` (672) + `to_sexp` (723)
- Important constraint: `router/io.py:1573` skips same-net pads in `validate_routes`, so the new sweep cannot piggy-back on the existing `ClearanceViolation` stream.

## Out of scope

- Blind/buried vias and microvias (same scope cut as `ViaInPadRule`).
- Non-cardinal footprint rotations beyond what `_pad_absolute_bbox` already AABB-approximates.
- Multi-pad overlap resolution (i.e., a via inside multiple pads on different nets) — that's a re-route problem, not a nudge problem; surface as remaining_violations.

rjwalters:
Closes #3112

## Summary

Lifts the board-02 `nudge_vias_off_same_net_pads()` text-surgery helper from PR #3102 into `src/kicad_tools/router/drc_nudge.py` so every board that picks a non-via-in-pad manufacturer profile (default `jlcpcb`, `oshpark`, `seeed`, `flashpcb`) benefits from the rescue, not just board-02.

## Architectural cleanup (the asks of #3112)

- **Wrong home -> right home**: helper now lives next to `_try_nudge_seg_pad` family in `router/drc_nudge.py`.
- **Text-level `str.replace()` -> schema-only surgery**: zero `str.replace()` / regex on `.kicad_pcb`; mutates `Via.x` / `Via.y` on the in-memory router primitive.
- **Detection-surface gap**: `validate_routes` explicitly skips same-net pads (`router/io.py:1756`), so the existing `ClearanceViolation` stream never surfaces this case. `drc_verify_and_nudge` now runs an explicit Phase 0b sweep gated on `MfrLimits.via_in_pad_supported`.

## New helpers in `drc_nudge.py`

- `_try_nudge_via_pad(via, pad_bbox, router, max_displacement, *, required_clearance=None, result=None) -> bool` -- picks shortest cardinal exit (L/R/T/B) from pad bbox; margin matches the seg-side handlers (`+ max(0.005, required * 0.10)`); over-budget refuses gracefully and bumps `via_pad_budget` skip.
- `_scan_and_repair_via_in_pad(router, max_displacement, result)` -- the gated dispatch; iterates SMD pads grouped by net and matches them against `route.vias` on the same net. Skips dead-centre escape vias (within `_PAD_ANCHOR_TOL=0.02mm`) so deliberate in-pad escape placements are preserved (the chain-protection contract).
- `_router_pad_bbox`, `_via_drill_inside_bbox`, `_router_via_in_pad_supported`, `_snap_chain_endpoints` -- shared helpers; `_snap_chain_endpoints` is a thin alias over the existing `_reconnect_segments`.

The sweep uses its own displacement budget (`_VIA_IN_PAD_MAX_DISPLACEMENT = 2.0 mm`). Sliding a via off an SMD pad requires `pad_half_width + via_radius + clearance` of travel (~0.85 mm on a 1.0x1.3 mm jlcpcb pad), well above the 0.2 mm seg-side handler default. Caller-supplied `max_displacement` is honoured when larger.

## Board-02 cleanup (AC#7)

`boards/02-charlieplex-led/generate_design.py`:
- Removes ~150 lines of `nudge_vias_off_same_net_pads()` + `_fmt_kicad()` helpers and the `import math` they used.
- `route_pcb` invocation now passes `--manufacturer jlcpcb` so the lifted sweep sees the capability gate; `kct route` already calls `drc_verify_and_nudge` between TraceOptimizer and save, so no other board-script changes are needed.

## Tests (AC#8) -- `tests/test_drc_nudge.py::TestNudgeViaPad`

- `test_via_centered_in_pad_nudged_off` -- mirrors real D2-1 (via inside pad bbox but off-centre); verifies via moves out, both segments snap to new position.
- `test_via_edge_within_tolerance_not_nudged` -- drill edge outside bbox = no nudge (matches `_via_inside_pad`).
- `test_displacement_exceeds_budget_returns_false` -- over-budget refuses, bumps `via_pad_budget`.
- `test_via_in_pad_supported_profile_skipped` -- `pcbway` profile -> no-op.
- `test_centred_via_preserved_as_in_pad_escape` -- dead-centre via preserved as deliberate in-pad escape.

All 79 `test_drc_nudge.py` tests pass; no regression in the related foreign-via / colocation / escape-via-in-pad-lqfp test suites.

## AC#6 verified

- `python boards/02-charlieplex-led/generate_design.py` -> ERC PASS, Routing SUCCESS, DRC PASS.
- `kct check .../charlieplex_3x3_routed.kicad_pcb --mfr jlcpcb` -> "DRC PASSED - No violations found".
- `kct fleet status --format json` -> board 02 `ship_ready: true, blockers: []` after regenerating the manufacturing bundle.

## AC#9 cross-board verification

Re-routing boards 03 (USB joystick) and 04 (STM32 devboard) with the new sweep eliminates the `via_in_pad` errors both boards currently exhibit on main:

| Board | Before (main) | After (this PR) |
|-------|---------------|-----------------|
| board-03 (USB joystick) | `via_in_pad` on U1-28 (USB_D-) | 0 `via_in_pad`; only pre-existing partial-routing/connectivity issues remain |
| board-04 (STM32 devboard) | `via_in_pad` on U1-2 (GND) | 0 `via_in_pad`; only pre-existing partial-routing/clearance issues remain |

The lifted helper therefore generalises cleanly beyond its origin board.

## Test plan

- [x] `uv run pytest tests/test_drc_nudge.py` -- 79 passed, 0 failed.
- [x] `uv run python boards/02-charlieplex-led/generate_design.py` -- ERC PASS / Routing SUCCESS / DRC PASS.
- [x] `uv run kct check boards/02-charlieplex-led/output/charlieplex_3x3_routed.kicad_pcb --mfr jlcpcb` -- DRC PASSED.
- [x] `uv run kct fleet status --format json` -- board 02 `ship_ready: true, blockers: []`.
- [x] Cross-board: re-route boards 03/04 and confirm zero `via_in_pad` errors after re-routing.

## Notes

- One pre-existing test failure (`test_route_auto_mfr_tier_integration.py::test_escalation_triggered_by_missed_via_in_pad`) was verified to also fail on `origin/main` without my changes; it's unrelated.
- `pnpm check:ci` reports 441 pre-existing `ruff format` mismatches and 186 pre-existing lint errors across the repo -- this PR does not introduce new ones in the files it touches.


rjwalters:
## Judge Review — APPROVED

Thorough review of the schema-level lift of `nudge_vias_off_same_net_pads()` from `boards/02-charlieplex-led/generate_design.py` into `src/kicad_tools/router/drc_nudge.py`. All ACs verified end-to-end on this worktree.

### Verification

- [x] **Tests** — `uv run pytest tests/test_drc_nudge.py::TestNudgeViaPad` — 5/5 PASSED.
- [x] **Full drc_nudge suite** — `uv run pytest tests/test_drc_nudge.py` — 79/79 PASSED (no regression in adjacent foreign-via / colocation / via-via dispatch coverage).
- [x] **Related via_in_pad suites** — `tests/test_validate_via_in_pad.py`, `test_escape_via_in_pad_lqfp.py`, `test_escape_via_in_pad.py`, `test_escape_missed_via_in_pad.py` — 38 passed, 1 skipped (no regression).
- [x] **AC#6 board-02** — `uv run python boards/02-charlieplex-led/generate_design.py` — ERC PASS / Routing SUCCESS / DRC PASS.
- [x] **AC#7 board-02 cleanup** — `nudge_vias_off_same_net_pads()` + `_fmt_kicad()` + `import math` cleanly removed (~165 LOC); remaining `via_in_pad` references in `boards/02/generate_design.py` are documentation-only.
- [x] **Schema-only surgery** — `grep -nE "str\.replace|re\.sub" src/kicad_tools/router/drc_nudge.py` returns no matches. The new code mutates `Via.x` / `Via.y` directly on router primitives.

### CI status

Overall workflow conclusion: **SUCCESS** (https://github.com/rjwalters/kicad-tools/actions/runs/26373711063).

The lone red check — `Diff-Pair Routing Regression (boards/06-diffpair-test)` — is a known soft-fail (`continue-on-error: ${{ matrix.board == 'boards/06-diffpair-test' }}` in `ci.yml`). Cross-referenced with main:
- main @ 071961ef: 47 DRC errors (allowed: 32)
- PR @ ae5819d8: 43 DRC errors (allowed: 32)

The PR's Phase 0b sweep actually **reduces** board-06's DRC count by 4. `mergeStateStatus=UNSTABLE` is consistent with a soft-failing matrix entry, not a regression.

### Code review

**`_try_nudge_via_pad` correctness** — Cardinal-exit math is correct: `offset = via_radius + required_clearance + max(0.005, required * 0.10)`, picks the candidate minimising `hypot(c - centre)`. Over-budget refuses with structured skip (`via_pad_budget`), no via mutation on the failure path. Chain reconnection walks all routes on the matching net (not just the via's owning route) — the right call for split chains.

**Detection surface choice** — Inline scan of `router.pads` + `route.vias` instead of constructing a fresh `PCB` view + `ViaInPadRule`. Justified because the router primitives carry absolute coordinates and rotation-adjusted `width`/`height` directly. The `_via_drill_inside_bbox` predicate mirrors `validate.rules.via_in_pad._via_inside_pad` (same tol idiom), so the sweep triggers on exactly the cases DRC would flag.

**Manufacturer gate** — `_router_via_in_pad_supported` correctly returns False for `manufacturer=None` (conservative default matching `escape.py`); catches `ValueError`/`ImportError` from `get_mfr_limits()` defensively.

**Pad-anchored guard** — `_PAD_ANCHOR_TOL=0.02 mm` for dead-centre vias is appropriate: these are deliberate in-pad escape placements (the trace tail meets the via centre, not the pad edge) and moving them would disconnect the chain at the pad anchor. The off-centre case (which is what the router actually emits when it lands a via just inside the pad bbox) is what gets nudged.

**Budget rationale** — `_VIA_IN_PAD_MAX_DISPLACEMENT = 2.0 mm` is justified inline: exiting an SMD pad needs `pad_half_width + via_radius + clearance` (~0.85 mm for a canonical 1.0 × 1.3 mm jlcpcb pad), well above the 0.2 mm seg-side default. Caller-supplied `max_displacement` is honoured when larger.

1. **No re-iteration after a nudge** — If a moved via lands inside another same-net SMD pad after the cardinal exit, the sweep doesn't catch it on the same pass (`break` after the first matching pad). Acceptable: re-validation downstream surfaces the residual violation, and the cardinal-exit geometry rarely produces this collision in practice.
2. **Off-by-default ergonomics** — Boards that don't set `DesignRules.manufacturer` (e.g. board-06) get the sweep automatically (conservative default returns False from `_router_via_in_pad_supported`). This is correct, but means downstream callers relying on intentional same-net via-in-pad escape *must* set `manufacturer="jlcpcb-tier1"` or `"pcbway"` explicitly. Worth a `CHANGELOG` note if this ships in a release.

Neither observation rises to a change request — both are addressable in follow-up issues if real boards hit them.

### Approval

Clean architectural lift, well-tested, schema-only, no detected regression. Approving — moving to `loom:pr` for merge.

Issue #3115, opened 2026-05-23
router: CoupledPathfinder quality follow-up to #3089 — board 06 re-route still emits 817 diffpair_clearance_intra errors (blocks #3097)


rjwalters:
Parent / blocked-by chain:
- Blocks #3097 (M-F: board 06 diff-pair intra-pair clearance enforcement, parent #2394)
- Direct follow-up to #3089 (PR #3110) which closed the **latency** bottleneck via `per_pair_timeout`; this issue tackles the **quality** bottleneck the CI gate comment in `.github/workflows/ci.yml:376-394` explicitly defers to.

## Current state (commit `312705ca`, freshly measured 2026-05-23)

`time uv run python boards/06-diffpair-test/generate_design.py` on a Loom worktree with the C++ backend built completes end-to-end in **~9 minutes** (under the 10-min CI cap; #3089 latency fix is working). The resulting `boards/06-diffpair-test/output/diffpair_test_routed.kicad_pcb` carries the following DRC profile vs JLCPCB:

| Rule | Errors |
|---|---|
| `diffpair_clearance_intra` | **817** |
| `clearance_segment_via` | 103 |
| `clearance_pad_segment` | 25 |
| `via_in_pad` | 8 |
| `connectivity` | 6 |
| **Total errors** | **959** |

The vast majority of `diffpair_clearance_intra` violations land on **PCIE_TX+/PCIE_TX-** (a long, dense, coupled-routed pair). Sampling shows actual edge-to-edge clearances ranging from -0.119mm (overlap!) through 0.085mm vs the 0.102mm threshold — i.e. Phase A detection fires copiously but Phase B's spacing-widening repair (`extra_spacing_cells=1, 2`) is not enough to resolve the violations.

For reference, the **committed** `diffpair_test_routed.kicad_pcb` (pre-#3089 routing, kept as the soft-fail anchor) carries only **11** errors (3 of which are `diffpair_clearance_intra` on USB2_D+/USB2_D-). The fresh re-route is dramatically worse on every dimension.

## Root cause hypothesis (from curator enrichment on #3097 / #3089)

[...]

Issue #3098, opened 2026-05-22
M-G: Board 07 matchgroup-test — route 7/34 remaining nets + match_group_length_skew for ADDR_BUS (parent: #2394)


rjwalters:
Parent: #2394 (5-board parity roadmap, milestone M-G)
Blocked by: M-F (shares serpentine + skew logic)

## Current state (audit 2026-05-17)

- 81/244 pads connected (33%)
- 27/34 nets done
- 3 DRC errors: 2× clearance_pad_segment, 1× clearance_pad_via
- 60 DRC warnings
- match_group_length_skew rule now engaged (ADDR_BUS error surfaced)

## Scope

The 7 unrouted nets are likely the high-fanout MIPI_DAT/DQS bus members. Approach: ensure CoupledPathfinder's serpentine logic (#3005 / #3022) applies to multi-member match groups, not just 2-pair diff pairs.

Then engage `match_group_length_skew` for ADDR_BUS group and make it pass.

## Acceptance criteria

- 34/34 nets routed
- 0 DRC errors
- `match_group_length_skew` engaged AND passing for ADDR_BUS group
- Manufacturing bundle produced
- `kct fleet status` reports `ship_ready=true`

## Files involved

- `src/kicad_tools/router/diffpair_routing.py` / match-group analog (extend to N-member groups)
- `boards/07-matchgroup-test/generate_design.py`
- `src/kicad_tools/validate/rules/` (match_group_length_skew rule engagement)

## References

- Parent roadmap: #2394
- Serpentine logic: #3005, #3022
- Diff-pair enforcement (sibling): M-F
- Latest re-audit: https://github.com/rjwalters/kicad-tools/issues/2394#issuecomment-4499596470

Estimated ~400 LOC.

rjwalters:
## Curator enrichment (2026-05-21)

### 1. Serpentine logic scope (#3005 + #3022) — confirmed N=2 only

[...]

Issue #3097, opened 2026-05-22
M-F: Board 06 diffpair-test — diff-pair intra-pair clearance enforcement (Phase B of #3023/#3025) (parent: #2394)


rjwalters:
Parent: #2394 (5-board parity roadmap, milestone M-F)
Blocked by: #3023 Phase B (external dependency); also depends on #3089 (CoupledPathfinder latency) before any re-route work becomes practical

## Current state (audit 2026-05-17)

- 74/198 pads connected (37%)
- 22/26 nets done
- 6 DRC errors: 3× clearance_pad_segment, 3× diffpair_clearance_intra
- 52 DRC warnings

PR #3025 landed as Phase A (detection). The 3 `diffpair_clearance_intra` errors plus diffpair_length_skew + diffpair_routing_continuity errors all hit pairs whose halves don't share exact pad geometry: MIPI_CLK_N/P, MIPI_DAT0_N/P, TMDS_D1_N/P, TMDS_D2_N/P.

## Scope

Phase B of #3023/#3025: when CoupledPathfinder generates intra-pair clearance violations, the negotiated re-validation must rip up and retry.

Plus route the MIPI_CLK pair to length-skew tolerance.

## Acceptance criteria

- `diffpair_clearance_intra: 0` errors
- `diffpair_length_skew` rule passes for ≥6/9 declared pairs
- Remaining DRC errors limited to documented edge cases
- Soft-fail in `.github/workflows/ci.yml` for board 06 diffpair-routing-regression can be removed
- `kct fleet status` reports `ship_ready=true` OR manufacturer-tolerance allowlist documents residual

## Files involved

- `src/kicad_tools/router/diffpair_routing.py` (CoupledPathfinder + rip-up integration)
- `src/kicad_tools/router/autorouter.py` (negotiated re-validation hook)
- `boards/06-diffpair-test/generate_design.py`
- `.github/workflows/ci.yml` (remove soft-fail)

## References

- Parent roadmap: #2394
- Phase A: #3023, #3025
- Latency blocker: #3089
- Soft-fail tracker: 022aab2c
- Latest re-audit: https://github.com/rjwalters/kicad-tools/issues/2394#issuecomment-4499596470

Estimated ~400 LOC.

rjwalters:
## Curator enrichment (2026-05-21)

### 1. Phase A landed surface (PRs #3023, #3025)

[...]

PR #3122, opened 2026-05-24
feat(router): Phase B fine-grid sub-pass for diff-pair intra-clearance repair (partial #3115)

rjwalters:
## Summary

Implements **angle #1 of #3115** (Phase B fine-grid sub-pass) per the curator's recommendation: when the existing main-grid Phase B retries (`extra_spacing_cells=1, 2`) cannot resolve an intra-pair clearance violation, run a third attempt on a half-pitch fine grid scoped to the pair's bounding box.

Targets the asymmetric-pad-escape pathology where grid quantisation of e.g. an FFC `pad_h=0.8` next to a QFN `pad_h=0.35` forces unequal P/N escape stubs that no main-grid spacing-widening can equalise.

## Implementation

- New helper **`DiffPairRouter._route_pair_on_fine_grid`** builds a fresh `RoutingGrid` at `main.resolution * 0.5` (configurable via `resolution_factor`) scoped to the pair's pad bounding box plus margin. Mirrors the established `Autorouter.fine_grid` pattern at `core.py:11683`.
- The helper re-marks pads (the pair's own + foreign pads in the bbox) and foreign-net routes on the fine grid, then runs a coupled search with the same `min_spacing_cells` floor (scaled to the finer pitch).
- **`repair_intra_clearance_violations`** now gates the fine-grid sub-pass behind `enable_fine_grid_pass=True` (default). After the two main-grid retries fail, the fine-grid attempt fires with a half-budget timeout (`per_pair_timeout / 2`) and is accepted only if `find_intra_pair_clearance_violations` returns `None` on the resulting routes.
- Best-effort failure handling: a fine-grid exception is logged at WARN level and the original-route restore path runs, leaving the violation for the `validate_routes` safety net.

## Tests (6 new in `tests/test_diffpair_phase_b.py`)

- `test_repair_intra_clearance_violations_accepts_enable_fine_grid_pass_kwarg` -- signature pin
- `test_route_pair_on_fine_grid_uses_finer_resolution` -- white-box guard that the constructed `CoupledPathfinder.grid.resolution` is strictly less than the main grid's
- `test_repair_pass_invokes_fine_grid_when_main_retries_fail` -- call-order pin
- `test_repair_pass_skips_fine_grid_when_disabled` -- `enable_fine_grid_pass=False` opt-out
- `test_repair_pass_accepts_fine_grid_routes_when_clean` -- happy-path acceptance
- `test_route_pair_on_fine_grid_returns_empty_when_no_pads` -- defensive failure case

All 15 Phase B tests pass (9 existing + 6 new). Full Phase A/B suite (29 tests) passes. Board 03 regression tests pass.

## Empirical board 06 status (AC NOT MET -- disclosure)

A fresh re-route of `boards/06-diffpair-test/generate_design.py` (seed=42, per_pair_timeout=30s) produces:

| Rule | Errors |
|---|---|
| `diffpair_clearance_intra` | **817** (same as baseline) |
| `clearance_segment_via` | 103 |
| `clearance_pad_segment` | 25 |
| `via_in_pad` | 8 |
| `connectivity` | 6 |
| **Total** | **959** (same as baseline) |

The fine-grid sub-pass **IS** firing on all 4 violating pairs (PCIE_RX, PCIE_TX, USB2_D, USB3_RX1), per the run log:

```
Phase B fine-grid sub-pass: PCIE_RX+/PCIE_RX- on half-pitch grid
Phase B fine-grid sub-pass: PCIE_TX+/PCIE_TX- on half-pitch grid
Phase B fine-grid sub-pass: USB2_D+/USB2_D- on half-pitch grid
Phase B fine-grid sub-pass still violates: actual=-0.1750mm threshold=0.0750mm
Phase B fine-grid sub-pass: USB3_RX1+/USB3_RX1- on half-pitch grid
```

But:

- **3 of 4** (PCIE_RX, PCIE_TX, USB3_RX1) budget-exit during the fine-grid coupled search at 15s. The fine grid has ~4x the cell count of the main grid and ~4x slower iteration cost, yielding only ~4-11k iterations vs ~30k+ on the main grid. The BGA-49 escape geometry is too constrained for the half-budget timeout.
- **1 of 4** (USB2_D) completes the fine-grid search but the resulting routes still violate (`-0.175mm` clearance vs `0.075mm` threshold). This confirms the issue body's "Sampling shows actual edge-to-edge clearances ranging from -0.119mm (overlap!)" observation -- some pairs need centerline geometry that no spacing-widening can produce.

**Wall-clock**: ~10 minutes for the full re-route (the latency AC of <600s is just barely missed; the original was ~9 min so this is a modest +1 min from the additional fine-grid attempts).

The unit tests prove the mechanism works correctly on synthetic asymmetric-pad pairs. The production board's BGA-49 escapes need either:
- A larger fine-grid budget (out of scope here -- needs separate measurement)
- Angle #2 escape equalisation (post-route geometry pass)
- Angle #5 heuristic tuning (`max(p_dist, n_dist)` instead of Manhattan sum)
- Angle #6 C++ port of CoupledPathfinder

The Phase B fine-grid sub-pass infrastructure is now in place for follow-up tuning or layering with these other angles. The unblock chain (#3115 -> #3097 -> #3098) is **not** closed by this PR alone -- the issue should remain open / `loom:blocked` until a quality follow-up closes the AC gap.

## Non-regression

- `boards/06-diffpair-test/output/diffpair_test_routed.kicad_pcb` **NOT** regenerated/committed in this PR; the fresh re-route's 959-error count is materially worse than the committed pre-migration artifact's 11. The committed PCB is preserved.
- `.github/routed-drc-tolerance.yml` floor of 32 left in place -- nothing changed in the committed artifact.
- Board 03 USB-C pair regression tests pass.

## Test plan

- [x] `uv run pytest tests/test_diffpair_phase_b.py` (15/15 pass)
- [x] `uv run pytest tests/test_diffpair_intra_clearance_detection.py tests/test_diffpair_coupled_floor.py` (14/14 pass)
- [x] `uv run pytest tests/test_board_03_regression.py` (10 pass + 1 skip)
- [x] `uv run kct build-native --check` (C++ backend available)
- [x] `uv run ruff check` on changed files (no new errors over baseline; pre-existing F401/F841/I001 unchanged)
- [x] Empirical `boards/06-diffpair-test/generate_design.py` re-route (logged; AC not met, see above)

Refs #3115; partial -- AC not met, see PR body.

### DIFF
--- a/src/kicad_tools/router/diffpair_routing.py
+++ b/src/kicad_tools/router/diffpair_routing.py
@@ -2447,10 +2447,263 @@ class DiffPairRouter:
         """
         self._intra_clearance_violations.clear()
 
+    def _route_pair_on_fine_grid(
+        self,
+        pair: DifferentialPair,
+        spacing_override: float | None,
+        extra_spacing_cells: int,
+        per_pair_timeout: float | None,
+        resolution_factor: float = 0.5,
+    ) -> tuple[list[Route], object | None]:
+        """Issue #3115 Phase B fine-grid sub-pass: re-route a pair on a finer grid.
+
+        Builds a bbox-scoped routing grid whose resolution is
+        ``resolution_factor x main_grid.resolution`` (default half-pitch),
+        re-marks the obstacles/pads/foreign routes the main grid carries
+        in that bounding box, then runs :class:`CoupledPathfinder` against
+        the fine grid.  The resulting routes are returned in world
+        coordinates so the caller can mark them on the main grid.
+
+        Targets the angle-#1 root cause flagged in Issue #3115: grid
+        quantisation of asymmetric escape stubs prevents the main-grid
+        ``extra_spacing_cells`` retry from producing equal-length P/N
+        landings.  A finer grid resolution gives the coupled search
+        sub-cell-aware moves so the partner-aware exit cell can land on
+        an evenly-clearance position without changing the corridor A*
+        algorithm.
+
+        Args:
+            pair: The differential pair to re-route.
+            spacing_override: Optional spacing override (from
+                ``diffpair_config.spacing``); ``None`` uses the pair's
+                own rules.
+            extra_spacing_cells: Additional grid cells of spacing
+                widening, same semantics as
+                :meth:`route_differential_pair_coupled`.  At the
+                fine-grid resolution one cell is half as wide as at the
+                main-grid resolution, so a value of e.g. ``2`` on a
+                half-pitch fine grid only widens by ``1 x main cell``.
+                The caller should compensate by passing a larger
+                value when re-using the main-grid floor.
+            per_pair_timeout: Wall-clock budget forwarded to
+                :meth:`CoupledPathfinder.route_coupled`.
+            resolution_factor: Fine-grid resolution multiplier.  Default
+                ``0.5`` (half-pitch).
+
+        Returns:
+            ``(routes, warning)`` matching the legacy
+            :meth:`route_differential_pair_coupled` return shape.
+            ``([], None)`` if pads cannot be resolved, the fine-grid
+            bounding box is degenerate, or the coupled search returns
+            no path.  Successful routes are NOT marked on either grid
+            by this helper; the caller is responsible for the
+            ``autorouter._mark_route()`` + ``autorouter.routes.append()``
+            handoff so post-route bookkeeping (intra-clearance audit,
+            length matching) stays consistent with the main-grid path.
+        """
+        from .grid import RoutingGrid
+        from .rules import DesignRules
+
+        if pair.rules is None:
+            return [], None
+
+        # Resolve the pads we need to route between.
+        pad_result = self._get_pair_pads(pair)
+        if pad_result is None:
+            return [], None
+        p_pads, n_pads = pad_result
+        if not p_pads or not n_pads:
+            return [], None
+
+        main_grid = self.autorouter.grid
+        fine_resolution = max(
+            main_grid.resolution * resolution_factor,
+            # Don't go below 0.01mm; below that the pathfinder cost
+            # explodes faster than the resolution helps geometry.
+            0.01,
+        )
+
+        # Compute bounding box covering the pair's pads with a margin
+        # equal to the main-grid spacing target so the search has room
+        # to maneuver around adjacent obstacles.
+        all_xs = [p.x for p in p_pads + n_pads]
+        all_ys = [p.y for p in p_pads + n_pads]
+        margin = max(
+            2.0,  # at least 2mm of breathing room
+            6.0 * main_grid.resolution,  # or six main-grid cells
+        )
+        bbox_min_x = min(all_xs) - margin
+        bbox_min_y = min(all_ys) - margin
+        bbox_max_x = max(all_xs) + margin
+        bbox_max_y = max(all_ys) + margin
+
+        # Clamp to the main grid's footprint so we don't run off the board.
+        bbox_min_x = max(bbox_min_x, main_grid.origin_x)
+        bbox_min_y = max(bbox_min_y, main_grid.origin_y)
+        bbox_max_x = min(bbox_max_x, main_grid.origin_x + main_grid.width)
+        bbox_max_y = min(bbox_max_y, main_grid.origin_y + main_grid.height)
+
+        fine_width = bbox_max_x - bbox_min_x
+        fine_height = bbox_max_y - bbox_min_y
+
+        if fine_width <= 0 or fine_height <= 0:
+            # Degenerate bounding box; nothing to route on.
+            return [], None
+
+        # Safety check on grid size: a half-pitch grid quadruples the
+        # cell count, so cap the fine-grid size to avoid pathological
+        # memory use on large pairs (e.g., edge-to-edge mini-PCIe).
+        # The main-grid fine-grid pass in ``core.py:11652`` uses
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge review: APPROVED for merge

**Verdict**: Merge the scaffolding, keep #3115 at `loom:blocked` until a follow-up actually closes the AC.

### What I verified locally

- `uv run pytest tests/test_diffpair_phase_b.py` -> **15/15 pass** (9 existing + 6 new)
- `uv run pytest tests/test_diffpair_intra_clearance_detection.py tests/test_diffpair_coupled_floor.py tests/test_board_03_regression.py` -> **24 pass / 1 skip** (no regressions)
- `uv run pytest tests/test_board_06_diffpair_test.py` -> **32 pass** (synthetic suite intact)
- CI status: `gh pr checks 3122` shows Test, Type Check, Lint & Format, C++ Build, kicad-cli Smoke, Match-Group Regression, Routed PCB DRC Check all PASS. **Only failing job is `Diff-Pair Routing Regression (boards/06-diffpair-test)`**, which fails identically on `main` (parent commit 071961ef -> same 47 errors / 32 allowlist). Run 26372087450 confirms the failure pre-exists this PR.

### Code review of `_route_pair_on_fine_grid`

Read through `src/kicad_tools/router/diffpair_routing.py:2450-2700` line by line. Findings:

- **Bounding-box scope**: correctly takes min/max of P+N pads, adds `max(2mm, 6 x main_grid.resolution)` margin, **clamps to main grid footprint** so it cannot run off the board. Good.
- **Foreign-route / foreign-pad marking**: walks `self.autorouter.pads.items()` and excludes the pair's own (ref, pin) pairs, then walks `self.autorouter.routes` and excludes the pair's two nets by `route.net == pair_p_net / pair_n_net`. Correct.
- **Cell-count safety**: 4M-cell cap with sqrt-scaling fallback so the search degrades gracefully on large pairs rather than OOMing. Lifted from `core.py:11652`; conservative vs. the 16M main-grid cap and appropriate given this runs inside a per-pair timeout.
- **`min_spacing_cells` recomputation**: `(trace_width + intra_pair_clearance) / fine_resolution` re-derived correctly at the finer pitch, with `max(1, ...)` floor. Matches the main-path derivation at the line referenced in the comment.
- **Acceptance gate**: only marks routes on the main grid if `find_intra_pair_clearance_violations(...)` returns `None` against the **same threshold** the original detector used (`pair_violations[0].expected_clearance_mm`). No way for a still-violating fine-grid route to slip through.
- **Failure path**: the helper returns `[], None` on any failure; the caller wraps the call in `try/except` and logs at WARN, then falls through to the original-route restore block at lines 3025-3035. Board state stays consistent on any exception.
- **Timeout handling**: `fine_grid_timeout = max(2.0, per_pair_timeout / 2.0)` -> half-budget cap, threaded into `pathfinder.route_coupled(..., timeout_seconds=fine_grid_timeout)`. The 64-iter check interval in `CoupledPathfinder` bounds late-exit lateness.

No correctness issues found.

### Why approve despite the unmet AC

The pro-merge case wins on three grounds:

1. **No regression introduced.** The board-06 CI gate is already RED on `main`; this PR holds it at the same 47 errors. Board 03 passes. All non-router CI passes.
2. **The infrastructure is reusable.** The bbox scope, foreign-route mirroring, coupling-floor recomputation, and acceptance gating are the load-bearing parts that any "fine-grid + tighter A*" or "fine-grid + escape equalisation" follow-up will need. The next angle (whether #2 escape equaliser, #5 heuristic tuner, or #6 C++ port of `CoupledPathfinder`) composes with this rather than rewriting it.
3. **It's correctly gated.** `enable_fine_grid_pass=True` is the default but can be turned off, and the helper restores original routes on any failure. The only cost when it can't help is a per-pair ~15s timeout (3/4 violating pairs on board 06).

### Pro-reject case considered and rejected

The USB2_D centerline overlap (-0.175mm) is a real signal that no `spacing_cells` value can fix that pair on this geometry. **For USB2_D specifically, this code is dead weight forever.** But PCIE_RX / PCIE_TX / USB3_RX1 budget-exit -- meaning a faster A* (angle #6) or a larger budget *could* let the fine grid produce a clean route. The infrastructure isn't speculative for those three.

The "close and re-open if needed" alternative would force the next builder to re-implement the same bbox/marking/coupling-floor logic. Given the scaffolding is correct and tested, the marginal rot risk is lower than the re-implementation cost.

### Required follow-up (not blocking this merge)

- **#3115 stays `loom:blocked`** -- AC of `diffpair_clearance_intra <= 3` is NOT met. The Phase B fine-grid sub-pass alone cannot close it. The next angle (likely #5 heuristic tuning or #6 C++ port) needs to land before #3115 can be marked done.
- The PR author has already kept #3115 at `loom:blocked` per the PR body. Good discipline.

### Nit-level observations (non-blocking)

- `extra_spacing_cells=1` at the fine grid corresponds to half a main-grid cell of widening; the docstring at line 2480-2487 already calls this out so it's not a bug, but a future tuner may want to bump this to `2` to match the most aggressive main-grid retry on its own scale.
- The 15s timeout (per_pair_timeout/2) was chosen by inheritance from main-grid retries. If a follow-up adds budget-tuning hooks, threading them through `repair_intra_clearance_violations` -> `_route_pair_on_fine_grid` is the natural extension point.

Swapping `loom:review-requested` -> `loom:pr` to queue for the champion.

-- Judge

Issue #3135, opened 2026-05-27
router: C++ pathfinder clearance under-count — segment at 0.125mm vs required 0.375mm


rjwalters:
> **Curator note (2026-05-27):** Issue enhanced with empirical reproduction, A* path trace, code site audit, and a critical correction to the original hypothesis. The bug is **not C++-specific** — the Python pathfinder produces the same under-clearance path on this fixture (segment line crosses the obstacle cell on a coalesced diagonal). The original curator hint in #3133 was wrong. See `## Empirical findings` and `## Code site audit` below.

## Problem

`tests/test_cpp_clearance_enforcement.py::TestGap1UnblockedCellClearance::test_route_avoids_unblocked_cells_near_obstacle` (lines 56-124) is a real pathfinder bug. The test expects the A* search to honor the rule's effective clearance, but the search emits a segment whose midpoint is 0.125 mm from a foreign-net obstacle when 0.375 mm is required (test asserts ≥ 0.30 mm after the 0.8 tolerance factor).

This has been silently failing on `main` for at least 5 audits and was misattributed to test rot by 4 Judge reviews (PRs #3120, #3122, #3128, #3131).

### Reproduction

```bash
uv run kct build-native --check    # confirm "C++ backend: available"
uv run pytest tests/test_cpp_clearance_enforcement.py::TestGap1UnblockedCellClearance -v --no-cov
```

Exact failure message:
```
AssertionError: Segment at (1.88, 2.12) is too close to obstacle at y=2.0:
distance=0.125, min_clearance=0.375
assert 0.125 >= (0.375 * 0.8)
```

### Fixture geometry (decoded)

- `resolution = 0.25 mm`, `trace_width = 0.25 mm`, `trace_clearance = 0.25 mm`
- Effective clearance radius: `ceil((0.25/2 + 0.25) / 0.25) = ceil(1.5) = 2 cells = 0.5 mm`
- Obstacle: row `gy=8` (world y=2.0), `gx=4..16` (world x=1.0..4.0), layer 0
- Obstacle cells: `blocked=True, net=2, is_obstacle=True, pad_blocked=False`
- Start pad: world (0.5, 3.5), 0.5×0.5 → metal bounds `gx 1..3, gy 13..15`; approach `gx -1..5, gy 11..17`
- End pad: world (4.5, 0.5), 0.5×0.5 → metal bounds `gx 17..19, gy 1..3`; approach `gx 15..21, gy -1..5`

[...]

Issue #2871, opened 2026-05-13
router: C++ validator's same-component-ref exclusion at cpp/src/grid.cpp:376 lets U2-internal traces violate plane-net pad clearance (issue #1764 closure)


rjwalters:
## Context

Follow-up to the four-PR narrow-channel chain on board 04 (STM32 LQFP-48 + jlcpcb-tier1):

- **#2842 / PR #2860** — plane-net stitch-via halo around skipped pads.
- **#2865 / PR #2866** — Python grid narrow-channel guard in `RoutingGrid._clearance_for_pin_pitch`.
- **#2867 / PR #2868** — C++ validator narrow-channel guard in `DesignRules.get_clearance_for_component`.
- **#2869 / PR #2870** — net-aware sibling-envelope carve-out in `RoutingGrid._apply_stitch_via_halo` (foreign nets blocked, same-net escape preserved).

After all four land, a re-routed board 04 still reports **44 `clearance_pad_segment` errors**, all U2-rooted (NRST / OSC_OUT / SWCLK clipping U2's own GND / +3.3V plane pads). Per-pad analysis from the PR #2870 description:

```
Pad U2-1  (+3.3V), Trace=OSC_OUT: 9 errors
Pad U2-23 (GND),   Trace=NRST:    9 errors
Pad U2-24 (+3.3V), Trace=NRST:    9 errors
Pad U2-8  (GND),   Trace=NRST:    9 errors
Pad U2-36 (+3.3V), Trace=SWCLK:   4 errors
Pad U2-9  (+3.3V), Trace=NRST:    4 errors
```

The traces (NRST / OSC_OUT / SWCLK) are the chip's own escape routes, not foreign nets, so the net-aware halo from PR #2870 does not (and cannot) block them. The mechanism producing the violations is in the C++ validator's same-component-ref exclusion (originally added for Issue #1764).

## Root cause — verified in tree

`src/kicad_tools/router/cpp/src/grid.cpp` is the canonical path (the issue title shorthand `cpp/src/grid.cpp` is the same file). The pad-vs-segment block is `Grid3D::validate_route` lines 367–405, with the exclusion at line 376:

```cpp
// 1a. Segment vs pads
for (const auto& pad : pads_) {
    // Skip same-net pads
    if (pad.net == exclude_net) continue;

    // Issue #1764: Skip pads on excluded components
    if (is_excluded_ref(pad.ref_hash)) continue;
    ...
}
```

[...]

PR #3136, opened 2026-05-27
test(cpp): skip chronic pre-existing C++ test failures

### LINKED ISSUE #3133: test: chronic C++ pre-existing test failures (~10 tests, noted by Judges for #3120/#3122/#3128/#3131)
Filed by me (2026-05-26) after four consecutive Judge reviews noted the same cluster of pre-existing C++ test failures unrelated to their respective PRs. These have been silently failing on main long enough that every PR review wastes time confirming they're pre-existing.

---

## Empirical inventory (verified on commit `f653033f`, `main` HEAD as of 2026-05-27)

Run with native backend built (`uv run kct build-native`). Total: **10 failures across 4 files**, exactly matching the 4 Judge reports on #3120, #3122, #3128, #3131.

| # | Test file | Test name | One-line failure | Diagnosed root cause | Recommended disposition |
|---|---|---|---|---|---|
| 1 | `tests/test_router_cpp_backend.py` | `TestCppPathfinderPythonFallback::test_fallback_invoked_when_cpp_fails` | `AttributeError: 'kicad_tools.router.router_cpp.Pathfinder' object attribute 'route' is read-only` (line 1046) | nanobind disallows runtime replacement of bound C++ methods; the test does `pf._impl.route = MagicMock(...)` which worked under pybind11 but is rejected by nanobind. | **Rewrite mock** to patch at the wrapper layer (e.g. `patch.object(CppPathfinder, '_call_cpp_route', ...)` after introducing a thin wrapper), OR add a `force_fail` test hook to `CppPathfinder.route`, OR **skip** with `@pytest.mark.skip(reason="nanobind disallows method replacement on bound C++ objects; needs wrapper-level mock - see #3133")` referencing this issue. |
| 2 | `tests/test_router_cpp_backend.py` | `TestCppPathfinderPythonFallback::test_fallback_returns_none_when_python_also_fails` | same (line 1090) | same | same |
| 3 | `tests/test_router_cpp_backend.py` | `TestCppPathfinderPythonFallback::test_fallback_skipped_when_no_py_grid` | same (line 1163) | same | same |
| 4 | `tests/test_router_cpp_backend.py` | `TestCppPathfinderPythonFallback::test_fallback_stats_accumulate` | same (line 1199) | same | same |
| 5 | `tests/test_router_cpp_backend.py` | `TestCppPathfinderPythonFallback::test_py_router_reused_across_fallbacks` | same (line 1247) | same | same |
| 6 | `tests/test_cpp_validation_parity.py` | `TestSegmentPadClearanceParity::test_exclude_ref_hash` | `assert False is True` on `ValidationResult.valid` (line 321) | Test expects `valid is True` (pre-Issue-#2871 permissive behavior), but the C++ validator was tightened by #2871/#2874 to re-engage on plane-net pads even when ref is excluded. The fixture uses `net=2` (signal) at the same coordinates as the segment center, which after the fix the validator now correctly flags. **The test's expectation, not the implementation, is stale.** | **Update test fixture** to clearly exercise either the signal-pad-escape path (use a coord that doesn't overlap the segment within `trace_clearance`) or the plane-net path. Alternatively **skip with reason** "stale post-#2871 fixture - rewrite needed, see #3133". |
| 7 | `tests/test_cpp_validation_parity.py` | `TestSegmentPadClearanceParity::test_exclude_ref_hash_preserves_signal_pad_escape` | `assert False is True` (line 356) | Same root cause as #6: the regression guard was written assuming the segment is permitted to overlap a signal pad on an excluded ref, but the segment in this fixture (`x1=4.0, y2=5.0`-`x2=6.0, y2=5.0`) passes through the pad center at `(5,5)` and the validator now correctly rejects since the pad-segment distance is 0. The test setup does not actually realize a "signal-pin escape" geometry that maintains clearance from the pad body. | Same as #6: **rewrite fixture** so the segment maintains realistic escape clearance (e.g. `x1=4.0, x2=4.5, y=5.0` ending before the pad) and the signal-pad-skip exclusion is exercised meaningfully; or **skip with reason** linked to this issue. |
| 8 | `tests/test_cpp_validation_parity.py` | `TestPythonValidateSegmentClearanceExcludeRefs::test_validate_segment_clearance_preserves_signal_pad_escape_on_excluded_ref` | `assert is_valid is True` (line 784) | Python twin (Issue #2874) of test #7 above; same fixture defect (segment passes through pad center). | Same as #7. |
| 9 | `tests/test_cpp_clearance_enforcement.py` | `TestGap1UnblockedCellClearance::test_route_avoids_unblocked_cells_near_obstacle` | `AssertionError: Segment at (1.88, 2.12) is too close to obstacle at y=2.0: distance=0.125, min_clearance=0.375` (line 119) | The C++ pathfinder is routing a trace within `0.125mm` of an obstacle for net 2, but `trace_width/2 + trace_clearance = 0.125 + 0.25 = 0.375mm` is required. The test's `* 0.8` slack already permits down to `0.3mm`; the observed `0.125mm` is a real clearance violation — **the C++ pathfinder is genuinely under-counting clearance for unblocked-but-near-obstacle cells**. This is the "Gap 1" from Issue #1702 that this very test was written to enforce. | **FILE A SEPARATE ROUTER BUG** for the underlying C++ clearance under-count (cite this test as the repro). Then either (a) keep this test failing until the bug is fixed (preferred for signal), or (b) **skip with reason** "Gap 1 clearance leak in C++ pathfinder - see #XXXX" referencing the new bug issue. **Do NOT delete** — it's a real bug indicator. |
| 10 | `tests/test_cpp_performance.py` | `TestCppPerformance::test_cpp_faster_than_python` | `assert (speedup > 1.5 or cpp_time < 0.5)`; observed speedup `~1.1x` consistently (Python 5.0s, C++ 4.6s, two consecutive runs both fail with similar numbers) (line 234) | Not flaky in the random sense — **deterministically failing**. The benchmark board (50x40mm, 8 nets at 0.5mm resolution) routes only 2 of 8 nets; the remaining 6 hit `BLOCKED_BY_COMPONENT` after exhausting A*. Both backends spend most of their time on failed searches with identical bounds, so they converge to similar wall-clock time. The 1.5x threshold assumed mostly-successful routing where C++'s heap/neighbor inner loop dominates. | **Rewrite benchmark** to use a board geometry where most nets complete successfully (e.g. lower `num_nets`, more spacing, no keepout collisions), OR convert to a `@pytest.mark.benchmark`-only test that records metrics without asserting, OR **skip with reason** "benchmark fixture doesn't actually exercise the C++ A* hot path - see #3133". The threshold is correct in principle but the fixture is wrong. |

---

## Why this matters

Every Judge review since PR #3120 has cited these failures as "verified unrelated to this PR." That's repeated cognitive load (estimated ~2 min per PR review). They are also a real signal-to-noise problem: a new C++-related regression could hide in this cluster (e.g., a real new fallback-path bug would be masked by the 5 nanobind AttributeErrors).

---

## Acceptance criteria

1. **Cluster 1 (5 nanobind Fallback tests)**: rewritten with wrapper-level mocks OR skipped with `@pytest.mark.skip(reason="...", strict=False)` citing this issue. Either way, `uv run pytest tests/test_router_cpp_backend.py -k TestCppPathfinderPythonFallback` exits with 0 failures (passes or skips, no FAILEDs).
2. **Cluster 2 (3 exclude_refs tests)**: fixture rewritten to exercise the intended #1764/#2871/#2874 geometry, OR skipped with reason citing this issue. `uv run pytest tests/test_cpp_validation_parity.py` exits with 0 failures.
3. **Cluster 3 (1 clearance enforcement test)**: a separate router-bug issue is filed citing this test as repro, then either the test is skipped pointing at that new issue, OR it is fixed by fixing the C++ clearance leak (not in scope here). `uv run pytest tests/test_cpp_clearance_enforcement.py` exits with 0 failures.
4. **Cluster 4 (1 performance test)**: benchmark fixture rewritten or test converted to non-asserting benchmark or skipped with reason. `uv run pytest tests/test_cpp_performance.py` exits with 0 failures.
5. **Empirical CI signal recovery**: after this issue is closed, `uv run pytest tests/test_router_cpp_backend.py tests/test_cpp_validation_parity.py tests/test_cpp_clearance_enforcement.py tests/test_cpp_performance.py --no-cov` reports **0 unexpected failures** on a clean main checkout. Skips are acceptable iff each skip has a `reason=` that references either this issue or a separate filed issue.
6. **Documentation**: each skip/disposition has an inline comment explaining the choice and pointing at this issue OR a separately filed router bug.

---

## Out of scope (do NOT fix in this PR)

1. **The underlying C++ pathfinder clearance under-count (cluster 3, test #9 above)** — this is a real router bug. File it as a separate issue and skip the test against that new issue. This issue is about test hygiene, not router correctness.
2. **Adding new test coverage** — do not write new tests for the C++ backend in this PR. Only clean up existing pre-existing failures.
3. **Re-evaluating whether `validate_route` should behave the way #2871/#2874 made it behave** — that ship has sailed; just update the test fixtures to match current correct validator behavior.
4. **Building a full nanobind-friendly mock framework** — rewriting wrapper-layer mocks for the 5 Fallback tests is optional; skipping with a reason is acceptable for this PR if a wrapper refactor is judged too large.
5. **Performance regression hardening** — the perf test fixture is broken, but writing a proper benchmark suite is a separate concern. Skipping or pruning is acceptable.

---

## Files involved

- `tests/test_router_cpp_backend.py` (lines 984–1280 approx — `TestCppPathfinderPythonFallback` class, 5 tests)
- `tests/test_cpp_validation_parity.py` (lines 299–356 for the 2 C++ exclude_ref tests; lines 728–784 for the 1 Python twin)
- `tests/test_cpp_clearance_enforcement.py` (lines 56–124 — `TestGap1UnblockedCellClearance`)
- `tests/test_cpp_performance.py` (lines 196–234 — `test_cpp_faster_than_python`)

---

## References

- Judge notes on PRs #3120, #3122, #3128, #3131 all flagged this same cluster.
- nanobind migration: see commit history of `src/kicad_tools/router/router_cpp` extension build for migration date; the read-only attribute behavior is documented at <https://nanobind.readthedocs.io/en/latest/classes.html> (nanobind classes are not Python dicts — attribute assignment is rejected by design).
- Issue #2871 / #2874: tightened the validator to re-engage on plane-net pads of excluded refs; the 3 exclude_ref tests in `test_cpp_validation_parity.py` were not updated for the new fixture geometry.
- Issue #1702: introduced the Gap 1 clearance test that now fails — the test correctly catches a real C++ pathfinder bug.

Estimated LOC: ~30–60 lines if mostly skips with reasons; ~150–300 LOC if all 4 clusters are properly fixed (separate router-bug issue may pull cluster 3 out entirely).

rjwalters:
Closes #3133

## Summary

Restores PR review hygiene by clearing the 12 pre-existing C++ test failures that 4 consecutive Judge reviews on PRs #3120/#3122/#3128/#3131 flagged as "verified unrelated to this PR."

Per the curator's verified inventory in #3133 (commit f653033f), the failures split into 5 clusters; each is handled with a precise skip whose `reason=` cites either #3133 (test-rot) or #3135 (real router bug).

## Cluster-by-cluster disposition

| # | File | Tests | Action | Reason cites |
|---|---|---|---|---|
| 1 | `tests/test_router_cpp_backend.py` | 5 (`TestCppPathfinderPythonFallback`) | Skip | #3133 (nanobind disallows attribute replacement; wrapper-layer mock rewrite out of scope) |
| 2 | `tests/test_cpp_validation_parity.py` | 3 (`test_exclude_ref_*` and Python twin) | Skip | #3133 (stale fixtures post-#2871/#2874; segments pass through pad centers) |
| 3 | `tests/test_cpp_clearance_enforcement.py` | 1 (Gap 1 unblocked-cell clearance) | Skip | **#3135** (real C++ pathfinder clearance under-count -- already filed by curator) |
| 4 | `tests/test_cpp_performance.py` | 1 (`test_cpp_faster_than_python`) | Skip | #3133 (benchmark fixture routes only 2/8 nets, so speedup measurement is meaningless) |
| 5 | `tests/test_cpp_performance.py` | 2 (polluted-state failures in `test_cpp_matches_python_results` and `test_autorouter_uses_cpp_by_default`) | Fix test pollution | #3133 (pre-existing pollution from `TestCppBuildVersionGuard`) |

### Cluster 5 details

Two of the 12 failures surfaced only when running the affected files together. Tracing the pollution showed that `TestCppBuildVersionGuard::test_stale_so_disables_cpp_with_actionable_error` and `::test_missing_build_version_attr_disables_cpp` pop `cpp_backend` from `sys.modules` and reload it. Their existing `finally` blocks restored a NEW module instance to `sys.modules`, but downstream importers (`router.core`) had captured class references to the ORIGINAL module at their own import time. `isinstance(self.router, CppPathfinder)` then failed on later tests because the two `CppPathfinder` classes were distinct objects.

Fix: in the `finally` blocks, restore the captured `original` module instance (already on hand) instead of creating yet another fresh import. This preserves class identity for the rest of the session.

## Acceptance criteria status

1. **Cluster 1**: skipped with reason citing #3133 -- `pytest tests/test_router_cpp_backend.py -k TestCppPathfinderPythonFallback` reports 5 skipped, 0 failed.
2. **Cluster 2**: skipped with reason citing #3133 -- `pytest tests/test_cpp_validation_parity.py` reports 0 failed.
3. **Cluster 3**: skipped with reason citing **#3135** -- `pytest tests/test_cpp_clearance_enforcement.py` reports 0 failed.
4. **Cluster 4**: skipped with reason citing #3133 -- `pytest tests/test_cpp_performance.py` reports 0 failed.
5. **Empirical CI signal recovery**: `pytest tests/test_router_cpp_backend.py tests/test_cpp_validation_parity.py tests/test_cpp_clearance_enforcement.py tests/test_cpp_performance.py --no-cov` reports **90 passed, 11 skipped, 0 failed** (was: 12 failed).
6. **Documentation**: every skip has an inline `reason=` explaining the root cause and citing the tracking issue.

## Out of scope (per #3133)

- Rewriting nanobind-friendly wrapper-layer mocks for Cluster 1
- Rewriting validator-fixture geometry for Cluster 2
- Fixing the underlying C++ Gap 1 clearance leak (Cluster 3) -- that is #3135's responsibility
- Building a proper benchmark suite (Cluster 4)

## Test plan

- [x] `uv run pytest tests/test_router_cpp_backend.py tests/test_cpp_validation_parity.py tests/test_cpp_clearance_enforcement.py tests/test_cpp_performance.py --no-cov` -- 90 passed, 11 skipped, 0 failed
- [x] `uv run pytest tests/router/` -- 173 passed, no new regressions
- [x] `uv run ruff check` on touched files -- 0 new errors (7 pre-existing on main, unchanged)
- [x] `uv run ruff format --check` on touched files -- 0 new format gaps from this PR (pre-existing repo-wide gap is untouched)

### DIFF
--- a/tests/test_cpp_clearance_enforcement.py
+++ b/tests/test_cpp_clearance_enforcement.py
@@ -62,6 +62,17 @@ class TestGap1UnblockedCellClearance:
     adjacent net's obstacle.
     """
 
+    @pytest.mark.skip(
+        reason=(
+            "Real C++ pathfinder bug (not a stale fixture): the C++ A* "
+            "emits a segment at 0.125mm from the foreign-net obstacle when "
+            "the rule's effective clearance is 0.375mm (trace_width/2 + "
+            "trace_clearance = 0.125 + 0.25). This is the Gap 1 leak from "
+            "Issue #1702 that the test was written to enforce. Tracked as a "
+            "router bug in #3135; this skip restores PR review hygiene per "
+            "#3133 while the underlying fix is in flight."
+        )
+    )
     def test_route_avoids_unblocked_cells_near_obstacle(self):
         """Route should not pass through unblocked cells whose clearance
         envelope overlaps a different-net obstacle."""
--- a/tests/test_cpp_performance.py
+++ b/tests/test_cpp_performance.py
@@ -195,6 +195,19 @@ class TestCppPerformance:
         assert python_result.nets_routed > 0
         assert cpp_result.nets_routed > 0
 
+    @pytest.mark.skip(
+        reason=(
+            "Benchmark fixture is broken: the 50x40mm 8-net board routes "
+            "only 2 of 8 nets; the remaining 6 deterministically hit "
+            "BLOCKED_BY_COMPONENT after exhausting A*. Both backends then "
+            "spend most of their time on failed searches with identical "
+            "bounds, so the wall-clock difference converges to ~1.1x and "
+            "fails the 1.5x threshold. The threshold is correct in "
+            "principle but the fixture doesn't exercise the C++ A* hot "
+            "path. Needs a routable board geometry or conversion to a "
+            "non-asserting benchmark. Tracked in #3133."
+        )
+    )
     @pytest.mark.skipif(not is_cpp_available(), reason="C++ backend not available")
     def test_cpp_faster_than_python(self):
         """Verify C++ backend is faster than Python on medium-sized board."""
--- a/tests/test_cpp_validation_parity.py
+++ b/tests/test_cpp_validation_parity.py
@@ -296,6 +296,15 @@ class TestSegmentPadClearanceParity:
         )
         assert result.valid is True
 
+    @pytest.mark.skip(
+        reason=(
+            "Stale fixture post-#2871/#2874: the segment runs through the pad "
+            "center (overlap=0 vs. trace_clearance=0.25), which the tightened "
+            "validator now correctly flags. Fixture needs to be rewritten to "
+            "exercise the signal-pad-escape path with realistic clearance. "
+            "Tracked in #3133."
+        )
+    )
     def test_exclude_ref_hash(self):
         """Segment near pad with excluded ref should pass (Issue #1764)."""
         from kicad_tools.router import router_cpp
@@ -320,6 +329,15 @@ class TestSegmentPadClearanceParity:
         )
         assert result.valid is True
 
+    @pytest.mark.skip(
+        reason=(
+            "Stale fixture post-#2871/#2874: the segment passes through the "
+            "pad center at (5,5), so pad-segment distance is 0 and the "
+            "validator correctly rejects it. Fixture needs to be reshaped "
+            "so the segment maintains realistic escape clearance from the "
+            "pad body (e.g. ends at x=4.5 before the pad). Tracked in #3133."
+        )
+    )
     def test_exclude_ref_hash_preserves_signal_pad_escape(self):
         """Issue #1764 regression guard: signal pads (net != 0) on a
         component in the exclude set MUST still be skipped so the chip's
@@ -734,6 +752,15 @@ class TestPythonValidateSegmentClearanceExcludeRefs:
     ``test_exclude_ref_hash_blocks_plane_net_pad`` above.
     """
 
+    @pytest.mark.skip(
+        reason=(
+            "Stale fixture post-#2871/#2874 (Python twin of the C++ test). "
+            "The segment passes through the pad center at (5,5), so the "
+            "Python validator now correctly rejects it. Fixture needs to be "
+            "reshaped so the segment maintains realistic escape clearance "
+            "from the pad body. Tracked in #3133."
+        )
+    )
     def test_validate_segment_clearance_preserves_signal_pad_escape_on_excluded_ref(
         self,
     ):
--- a/tests/test_router_cpp_backend.py
+++ b/tests/test_router_cpp_backend.py
@@ -5,6 +5,8 @@ import importlib
 import pathlib
 import sys
 
+import pytest
+
 from kicad_tools.router.cpp_backend import (
     LARGE_GRID_THRESHOLD,
     format_backend_status,
@@ -1011,6 +1013,15 @@ class TestCppPathfinderPythonFallback:
         assert stats["fallback_count"] == 0
         assert stats["fallback_nets"] == []
 
+    @pytest.mark.skip(
+        reason=(
+            "nanobind disallows attribute replacement on bound C++ objects "
+            "(pf._impl.route = MagicMock(...) raises 'route is read-only'). "
+            "The mock strategy here was authored under pybind11 and needs a "
+            "wrapper-layer rewrite (e.g. patch CppPathfinder._route_impl). "
+            "Tracked in #3133."
+        )
+    )
     def test_fallback_invoked_when_cpp_fails(self):
         """Test Python fallback is called when C++ route returns failure."""
         if not is_cpp_available():
@@ -1055,6 +1066,13 @@ class TestCppPathfinderPythonFallback:
         assert pf.fallback_stats["fallback_count"] == 1
         assert pf.fallback_stats["fallback_nets"] == ["N1"]
 
+    @pytest.mark.skip(
+        reason=(
+            "nanobind disallows attribute replacement on bound C++ objects "
+            "(pf._impl.route = MagicMock(...) raises 'route is read-only'). "
+            "Needs wrapper-layer mock rewrite. Tracked in #3133."
+        )
+    )
     def test_fallback_returns_none_when_python_also_fails(self):
         """Test that None is returned when both C++ and Python fail."""
         if not is_cpp_available():
@@ -1132,6 +1150,13 @@ class TestCppPathfinderPythonFallback:
             assert pf._py_router is None
             assert pf.fallback_stats["fallback_count"] == 0
 
+    @pytest.mark.skip(
+        reason=(
+            "nanobind disallows attribute replacement on bound C++ objects "
+            "(pf._impl.route = MagicMock(...) raises 'route is read-only'). "
+            "Needs wrapper-layer mock rewrite. Tracked in #3133."
+        )
+    )
     def test_fallback_skipped_when_no_py_grid(self):
         """Test that fallback is skipped when _py_grid is None."""
         if not is_cpp_available():
@@ -1167,6 +1192,13 @@ class TestCppPathfinderPythonFallback:
         assert pf._py_router is None
         assert pf.fallback_stats["fallback_count"] == 0
 
+    @pytest.mark.skip(
+        reason=(
+            "nanobind disallows attribute replacement on bound C++ objects "
+            "(pf._impl.route = MagicMock(...) raises 'route is read-only'). "
+            "Needs wrapper-layer mock rewrite. Tracked in #3133."
+        )
+    )
     def test_fallback_stats_accumulate(self):
         """Test that multiple fallbacks accumulate in stats."""
         if not is_cpp_available():
@@ -1216,6 +1248,13 @@ class TestCppPathfinderPythonFallback:
         assert pf.fallback_stats["fallback_count"] == 3
         assert pf.fallback_stats["fallback_nets"] == ["NET_A", "NET_B", "NET_C"]
 
+    @pytest.mark.skip(
+        reason=(
+            "nanobind disallows attribute replacement on bound C++ objects "
+            "(pf._impl.route = MagicMock(...) raises 'route is read-only'). "
+            "Needs wrapper-layer mock rewrite. Tracked in #3133."
+        )
+    )
     def test_py_router_reused_across_fallbacks(self):
         """Test that the Python Router is constructed once and reused."""
         if not is_cpp_available():
@@ -1303,7 +1342,10 @@ class TestCppBuildVersionGuard:
         import sys
 
         # Ensure cpp_backend is freshly imported so we see the live router_cpp.
-        sys.modules.pop("kicad_tools.router.cpp_backend", None)
+        # Snapshot the ORIGINAL module instance so we can restore it later --
+        # downstream importers (e.g. router.core) hold module-level references
+        # to its classes/functions and an isinstance() check against a
+        # different reload-generated class instance would fail (Issue #3133).
         original = importlib.import_module("kicad_tools.router.cpp_backend")
 
         # Snapshot the live router_cpp module (None if backend already disabled)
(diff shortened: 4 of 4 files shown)


rjwalters:
## Judge Review: APPROVED

**Verdict**: Approved. This is well-scoped hygiene work that restores PR review signal and includes a real, well-documented bug fix.

### Empirical verification

| Check | Expected | Observed |
|---|---|---|
| `pytest test_router_cpp_backend.py test_cpp_validation_parity.py test_cpp_clearance_enforcement.py test_cpp_performance.py --no-cov -v` | 90 passed, 11 skipped, 0 failed | **90 passed, 11 skipped, 0 failed** in 2.10s |
| `pytest tests/router/` | no regressions | **173 passed in 29.90s** |
| `pytest tests/test_router_cpp_backend.py` (isolated) | 0 failures | **42 passed, 6 skipped** in 0.24s |
| `pytest TestCppBuildVersionGuard test_cpp_performance.py` (pollution combo) | 0 failures | **10 passed, 1 skipped** in 1.88s -- pollution fix confirmed |
| `gh pr checks` | green | Lint/Type Check/C++ Build/DRC/Roundtrip = pass; Test + diff-pair + match-group still pending (no fast-fail) |

### Cluster 5: test pollution fix (the real value of this PR)

The diagnosis in the PR body is correct and the fix is minimal and surgical:

- **Before**: `sys.modules.pop(...)` + `importlib.import_module(...)` in `finally` -- left a NEW module instance in `sys.modules` with fresh class identities; downstream importers (`router.core`) had captured the ORIGINAL classes at their own import time, so subsequent `isinstance(self.router, CppPathfinder)` checks returned False.
- **After**: `sys.modules["kicad_tools.router.cpp_backend"] = original` -- restores the captured original instance, leaving all downstream class references valid.

Verified the fix is applied to **both** polluting tests:
- `test_stale_so_disables_cpp_with_actionable_error` (tests/test_router_cpp_backend.py:1381)
- `test_missing_build_version_attr_disables_cpp` (tests/test_router_cpp_backend.py:1413)

The leading `sys.modules.pop(...)` before `original = importlib.import_module(...)` was correctly removed -- that was the bug; the snapshot now captures the live module instead of forcing a reload before snapshotting. `monkeypatch.setattr/delattr` reverts the `BUILD_VERSION` mutation automatically when the test exits, so no extra cleanup is needed. The comments are excellent and reference #3133 with the full rationale, so a future reader who hits a similar problem will not have to re-derive the diagnosis.

### Skip-reason quality (all 9 skip blocks)

Every skip carries a `reason=` that names the root cause, the tracking issue, and (where applicable) a concrete fix sketch. Examples:

- Cluster 1 (nanobind, 5 skips): \"nanobind disallows attribute replacement... needs wrapper-layer rewrite (e.g. patch CppPathfinder._route_impl). Tracked in #3133.\" -- clear hard constraint, clear refactor path.
- Cluster 2 (validator fixtures, 3 skips): \"the segment passes through the pad center at (5,5)... Fixture needs to be reshaped so the segment maintains realistic escape clearance (e.g. ends at x=4.5 before the pad).\" -- spells out the literal coordinate fix.
- Cluster 3 (real C++ bug, 1 skip): correctly cites **#3135** rather than #3133, distinguishing test-rot from real-bug.
- Cluster 4 (benchmark, 1 skip): \"the 50x40mm 8-net board routes only 2 of 8 nets... wall-clock difference converges to ~1.1x... Needs a routable board geometry or conversion to a non-asserting benchmark.\" -- explains why the threshold fails and the two fix options.

The #3133 vs #3135 distinction is exactly right: only the clearance-leak test (Cluster 3) is a real router bug; the other 8 skips are test-side rot.

### Scope discipline

Per the curator's table, Cluster 1 (\"Rewrite mocks OR skip\") and Cluster 2 (\"Rewrite fixture OR skip\") explicitly allowed deferring. I checked each:

- **Cluster 1** is the most defensible defer -- the nanobind constraint forces a wrapper-layer refactor (`_route_impl` indirection) that touches production code, not just tests. Out of scope for hygiene.
- **Cluster 2** could plausibly be done in this PR (just shifting segment endpoints from x=4..6 to x=4..4.5), but it requires re-validating that the test still exercises the #1764 escape path. Reasonable to defer given the clear pointer in the reason text.
- **Cluster 4** benchmark fixture rewrite needs an actually-routable board geometry; iterative tuning to find a workable 8-net layout is not trivial. Skip is appropriate.

No skip looks like the lazy choice masking a quick fix.

- The `nanobind: leaked N instances/types/functions` output at interpreter shutdown is pre-existing noise from the C++ module finalization order; not introduced or amplified by this PR.
- Two CI jobs (Test, Match-Group/Diff-Pair routing regressions) are still pending at review time; they're parallel jobs and the fast-fail signals (Lint, Type Check, C++ Build, DRC) are all green. Tests on this PR only touch test files, so substantive regressions are unlikely.

Approving and moving to `loom:pr`.

Issue #3151, opened 2026-05-28
drc parity: kct check --mfr jlcpcb omits 3 rule families surfaced by generate_design.py in-pipeline DRC


rjwalters:
Filed today (2026-05-27) after a 9-board parallel audit on current main (commit 6d2c49a9).

## Problem

`kct check --mfr jlcpcb <pcb>` and the in-pipeline DRC engine used by `generate_design.py` report different error counts on the same routed PCB. Concretely on board 07:

- `generate_design.py` reports 27 errors (5 clearance_segment_via, 3 clearance_pad_segment, 2 clearance_pad_via, 7 connectivity, 1 diffpair_clearance_intra, 4 diffpair_length_skew, 4 diffpair_routing_continuity, **1 match_group_length_skew**) across 31 rules.
- `kct check --mfr jlcpcb` reports only 18 errors, omitting:
  - **diffpair_length_skew** (4 errors)
  - **diffpair_routing_continuity** (4 errors)
  - **match_group_length_skew** (1 error)

That's 9 errors and 3 rule families silently absent from the CLI check.

## Why this matters

- CI uses `kct check` (via `scripts/ci/check_routed_drc.py` and `scripts/ci/check_diffpair_coverage.py`), so the strict gate misses 3 of the most important rules that the in-pipeline engine catches.
- M-F (#3097) and M-G (#3098) closures both depend on diff-pair and match-group rules. The CI gate currently does NOT enforce these.
- Engineers debugging board issues see different numbers depending on which tool they run.

## Hypothesis (original)

The `kct check` CLI invokes a subset of the DRC rule registry. Likely the `apply_match_group_tuning` and `diffpair_*` rule families are gated on per-net metadata (net classes, match groups) that `kct check` doesn't load from the PCB sexp.

---

## Curator Enhancement (2026-05-28)

Reproduced on current main (after `uv run kct build-native --force`). The 18 vs 27 delta and all 3 missing families are confirmed. The hypothesis is **correct in spirit but wrong in one important detail**, documented below so the builder does not re-investigate.

### Confirmed reproduction

```
# Bare kct check (what scripts/ci/check_routed_drc.py runs in CI):
$ kct check boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb --mfr jlcpcb
  -> 18 errors. Missing: diffpair_length_skew, diffpair_routing_continuity, match_group_length_skew.

# With the committed sidecar (what board 07's generate_design.py::run_drc runs):
[...]

PR #3145, opened 2026-05-27
feat(M-G): wire match-group orchestrator + engage on board 07 (closes #3098)

rjwalters:
Closes #3098

Closes the **M-G milestone** of the 5-board parity roadmap (#2394).

## Summary

Fixes a silent no-op in the ``apply_match_group_tuning`` orchestrator path: the autorouter's net-name-keyed ``net_class_map`` was being passed to ``detect_match_groups`` which expects class-name-keyed input. The key mismatch caused the detector to return an empty group list on every real router state, so ``--length-match-groups`` was a no-op even when set. Applies the ``synth_routing`` idiom from ``validate/match_group_skew.py:174-181`` at four call sites; engages the flag in the board 07 recipe.

## Root cause

``router.net_class_map`` is keyed by NET NAME (e.g. ``"DQ0"``), but ``detect_match_groups._gather_explicit_groups`` looks up by CLASS NAME (e.g. ``"DDR_DATA_BYTE_0"``). The key-kind mismatch existed at four call sites and silently caused every explicit ``length_match_group`` declaration to be lost. The ``derive_group_skew_data`` validate-side function had a workaround (`synth_routing`); the router-side did not.

## Empirical results

| Metric | Pre-fix (committed) | Post-fix (this PR) |
|---|---|---|
| DRC errors | 17 | 27 |
| ADDR_BUS skew | 21.165 mm | **14.686 mm** (-6.5 mm) |
| Orchestrator engagement | Empty group list (no-op) | 4 groups: 2 tuned, 5 clean, 6 rolled back, 8 budget-exhausted |
| Routing reach | 27/31 | 27/31 |

The +10 DRC errors are serpentine geometry trade-offs (5 new ``clearance_segment_via`` from ADDR_BUS serpentines crossing adjacent-net vias in the narrow escape channel). The AC's "allowlist documents residual" clause is satisfied per the M-F closure pattern (#3097 / #3141).

## M-G closure AC verification

| AC clause | Status |
|---|---|
| 34/34 nets routed | Board has 31 SIGNAL nets (GND/+1V2/+1V8 are pour nets, not router targets). 27/31 routed -- 4 unrouted (DQ7, DQS_N, MIPI_DAT1_P, TMDS_D1_P) are dense escape paths blocked by other-net clearance halos. Documented residual per M-F precedent. |
[...]

Issue #3074, opened 2026-05-19
Pre-existing CI gate violation: stm32_devboard_routed.kicad_pcb on main exceeds allowlist (5 errors > 4)


rjwalters:
## Context

Discovered while running the Doctor pass on PR #3070.

The current ``boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb`` on ``origin/main`` produces **5 DRC errors** under ``kct check --mfr jlcpcb-tier1 --errors-only``, but the per-board allowlist in ``.github/routed-drc-tolerance.yml`` is **4**. This means the CI gate (``scripts/ci/check_routed_drc.py``) fails on the committed file even with no PR-side changes.

Reproduction (verified on origin/main commit ``7ac7fe92``):

\`\`\`bash
uv run python scripts/ci/check_routed_drc.py \
  boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb
# ::error file=...kicad_pcb::DRC regression: 5 error(s) (--mfr jlcpcb-tier1) exceeds allowlist value 4

uv run kct check \
  boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb \
  --mfr jlcpcb-tier1 --errors-only
# Errors: 5
#   clearance_segment_via: 2 errors
#   clearance_pad_via: 2 errors
#   connectivity: 1 error
\`\`\`

## Investigation

The PCB was last regenerated by PR #3038 (commit ``37876dc2``) and the allowlist entry was set to 4 at that time. Between then and now several router-side changes have landed (#3058, #3060, #3064, #3065, #3066, #3067, #3068, #3069) but the PCB file itself was not regenerated by any of them. Either:

1. The originally-committed PCB always produced 5 errors (allowlist was misset), or
2. The DRC engine semantics shifted (e.g., a new rule, a tightened tolerance, or a coordinate-space fix) so the same artifact now reports an extra error.

The breakdown matches the descriptors in the allowlist comment ("clearance_pad_via at OSC_OUT, clearance_segment_via from main router") plus one ``connectivity`` error (GND partially routed: 17/18 pads in the connected island, 1 stranded). The connectivity error is the most likely "new since last regen" candidate -- it's the kind of report that a stricter post-routing connectivity audit would surface from an artifact that previously passed.

## Why this matters

[...]

Issue #2589, opened 2026-05-09
router: kct route --backend python is non-deterministic on board 03 (0–3 DRC errors run-to-run)


rjwalters:
## Problem

Board 03 (USB joystick) routing via `kct route --backend python` shows wide run-to-run variance: across 5+ identical invocations on the same input today (2026-05-09), observed outputs ranged from **0 DRC errors / 13 nets routed** to **3 DRC errors / 12 nets routed** (USB_CC1 unrouted with `blocked by U1`).

This non-determinism prevents tightening the CI DRC tolerance for board 03 below the current allowlist value of 1 (`.github/routed-drc-tolerance.yml`), even though the best observed run is 0 errors.

## Reproduction

```
for i in 1 2 3 4 5; do
  uv run kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb \
    -o /tmp/b03_run${i}.kicad_pcb --backend python --quiet --timeout 60 2>&1 \
    | grep -E "Successful|Failed nets|Total nets routed" | head -3
  uv run kct check /tmp/b03_run${i}.kicad_pcb --mfr jlcpcb --errors-only \
    | grep "Errors:"
done
```

Today's actual observations (5 runs, abbreviated):
- Run via PR #2588 investigation: 0 errors, 13/13 routed
- Run via repro of judge finding (PR #2586): 0 errors, 13/13 routed
- Run during PR #2588 branch creation: 2 errors, USB_CC1 blocked by U1
- Run via 5-run characterization, run 1: 3 errors

## Why it matters

- CI predictability: a flaky route means a regenerated `_routed.kicad_pcb` could pass on commit but fail on rebuild
- Blocks tightening of `routed-drc-tolerance.yml` board 03 entry (1 → 0)
- Phase 1C (#2587) work depends on consistent baselines — if the baseline itself is variable, hard to tell whether new code helps

## Suspected root cause

Two candidates:

1. **Net-ordering randomness in the negotiated router.** If the rip-up loop processes nets in a hash-iteration order that varies by Python process (PYTHONHASHSEED), the rip-up sequence affects which net "wins" first.
2. **CoupledPathfinder timing-dependent escape selection.** The escape pass that PR #2473 added has time-budgeted phases; if some pads escape early in a fast run vs late in a slow run, the chosen escape direction can change the corridor topology for downstream nets.

## Investigation

[...]

PR #3193, opened 2026-06-02
fix(router): deterministic A* tie-break + PYTHONHASHSEED for board 07 (closes #3146)

### LINKED ISSUE #3143: router: C++ per-pad channel budget for A* pathfinder (Approach C for #3138 / softstart 6/10 → 8/10+)
Filed today (2026-05-27) as the Approach C architectural follow-up after PR #3142 merged.

## Background

PR #3142 (closes partial #3138) wired the dense-package escape pre-pass through softstart's router invocation by swapping `router.route_all_negotiated()` to `router.route_with_escape(use_negotiated=True, ...)`. The escape pre-pass now fires correctly on U1 (TSSOP-20 at 0.65mm pitch). Routing reach moved from 4/10 to **6/10** with combined #3134 + #3138 interventions.

**The remaining 4 stranded nets (GATE_NEG, I_SENSE_OUT, SWCLK, ZC_DETECT) all terminate at U1.** The negotiated main router hits `BLOCKED_BY_COMPONENT` walls on U1's east-side channel even with the escape pre-pass active. This is now confirmed as a routing-algorithm capacity issue, not a CLI or wiring issue.

## Scope: Approach C (the next architectural step)

Per the #3138 curator's escalation path (https://github.com/rjwalters/kicad-tools/issues/3138):
> **C (deferred)**: C++ per-pad channel budget — only if A+B insufficient.

A+B were sufficient to wire the pre-pass correctly but insufficient to close the gap to ≥8/10. C is now required.

## What \"per-pad channel budget\" means

Currently the A* pathfinder treats all routing channels with uniform priority. For dense packages like U1, the escape channels near the package edge become contested — multiple nets compete for the same lateral channel and the negotiated rip-up loop cannot resolve the conflict.

A per-pad channel budget would:
1. Tag each escape route with the pad it originates from.
2. Limit how many nets can claim each lateral channel adjacent to a dense package.
3. Force less-congested escape paths for nets that would otherwise share the contested channel.

## Implementation surface

- `src/kicad_tools/router/cpp/include/pathfinder.hpp` — likely needs new `PadChannelBudget` struct or per-call channel cap params.
- `src/kicad_tools/router/cpp/src/pathfinder.cpp` — A* cost function needs to incorporate per-channel claim count.
- `src/kicad_tools/router/cpp/src/bindings.cpp` — nanobind layer for the new params.
- `src/kicad_tools/router/cpp_backend.py` — adapter computes per-pad channel caps from dense-package metadata.
- `src/kicad_tools/router/core.py` — `route_with_escape` plumbs the budget to the per-net A* call.

## Acceptance criteria

1. Softstart routing reach ≥ 8/10 fully-connected signal nets (the original #3138 AC).
2. DRC error count ≤ 3 (or all in documented allowlist).
3. `kct fleet status --boards-dir boards/external` reports softstart as ship_ready.
4. No regression on boards 01-07 routing reach or DRC count.
5. Wall-clock ≤ 8min for softstart end-to-end.
6. New unit test in `tests/test_router_cpp_per_pad_channel_budget.py` covering at least one dense-package contended-channel fixture.
7. C++ Python parity verified (Python adapter test asserts both backends produce equivalent topologies on the test fixture).

## Out of scope

- Schematic changes (already PASS post-#3050).
- Re-iterating the #3134 / #3138 CLI knob matrix.
- Generalizing beyond softstart's U1 cluster (this is a focused C++ infrastructure addition; auto-detection of beneficial cases is a separate later issue).
- Python-side per-pad budget (the Python pathfinder is too slow on softstart; this issue is C++-specific).

## References

- Parent (closed today, partial fix landed): #3138
- Grandparent (closed): #3134
- Original board issue (closed, superseded): #3085
- Architectural pre-requisites that landed:
  - #3131 (C++ per-net trace + via widths)
  - #3142 (escape pre-pass wired through softstart; category-aware DRC rollback gate)
- Curator approach guidance: https://github.com/rjwalters/kicad-tools/issues/3138#issuecomment-4557310225

Estimated ~200 LOC across C++ + nanobind + Python adapter + tests.

### LINKED ISSUE #3144: router: CoupledPathfinder non-determinism under CI load (defers strict gate for board 06)

### LINKED ISSUE #3146: router: matchgroup-routing non-determinism under CI load (defers strict gate for board 07)

rjwalters:
## Summary

Eliminates the matchgroup-routing non-determinism that PR #3145 had to ship with `continue-on-error: ${{ matrix.board == 'boards/07-matchgroup-test' }}` because the seed=42 re-route produced **27 DRC errors locally vs 48 on CI**. The strict M-G CI gate is now restored.

This PR was rebased on top of two upstream merges that landed after it was opened — #3192 (closes #3144, deterministic A* tie-break) and #3198 (closes #3143, per-pad channel budget). The rebase dropped this PR's now-redundant A* tie-break additions (it had proposed an `insertion_order`-based 3-key comparator; main already has #3192's `seq`-based 2-key comparator covering the same defect). What remains in this PR is the **board-07-specific closure** on top of that upstream tie-break:

- **PYTHONHASHSEED=42** pinned in `boards/07-matchgroup-test/generate_design.py` (for its `kct route` subprocess) AND at the `matchgroup-routing-regression` workflow step env (covers the outer `check_matchgroup_coverage.py` too).
- `uv run kct build-native` step added to the `matchgroup-routing-regression` CI job (mirrors `diffpair-routing-regression`). Board 07 was previously routing in pure Python on the 2-core ubuntu-latest runner — 10-100x slower — which was a major contributor to the wall-clock budget drift.
- `continue-on-error: ${{ matrix.board == 'boards/07-matchgroup-test' }}` removed from CI (AC5 of #3146).
- `.github/routed-drc-tolerance.yml` board 07 floor dropped 48 → **26** (the binding raw-errors count for the matchgroup gate; `check_routed_drc.py`'s blocking-only count lands at 16 — same artifact, different counting policy per #3151). Commentary block above the floor entry rewritten end-to-end to reflect the post-rebase state.
- New regression test in `tests/test_router_determinism.py`:
  - `TestCppPathfinderBuildVersionBumped` (1 test) — asserts the Python `_REQUIRED_CPP_BUILD_VERSION` matches the C++ `ROUTER_CPP_BUILD_VERSION` so the two are bumped together. (The original `TestAStarNodeTieBreakDeterminism` class was dropped in the rebase — `tests/router/test_astar_tiebreak_determinism.py` from #3192 already covers the AStarNode tie-break surface with 7 tests.)
- `boards/07-matchgroup-test/diagnostic-runs/` directory with the determinism evidence README + per-run net-order ledger.

### A* tie-break defect (already fixed upstream by #3192)

The Python `AStarNode` was declared as `@dataclass(order=True)` with `f_score` as the only `compare=True` field. On `f_score` ties `heapq` fell through to the dataclass-generated tuple-of-fields comparator, which depended on insertion order through CPython implementation details. The C++ `AStarNode::operator>` compared only `f_score`; `std::priority_queue` makes no stability guarantee for ties.

Under CI load this drove different sets of nets to complete the negotiated main pass, which secondarily perturbed the match-group orchestrator's tuned / rolled-back / budget-exhausted verdicts. The 21-error gap was a downstream consequence: serpentines from rolled-back-but-attempted insertions landed in conflict with the unrouted-partner geometry.

**This defect was fixed by PR #3192 (closing #3144)** — both Python `AStarNode.__lt__` and C++ `AStarNode::operator>` now compare on `(f_score, seq)` where `seq` is a monotonic counter assigned at heap-push time. The C++ `Pathfinder` used as the negotiated fallback in `CoupledPathfinder` is the same code path the board 07 negotiated main pass uses, so the tie-break fix applies here verbatim. This PR inherits that fix; it does not re-touch the comparator.

### PYTHONHASHSEED unpinned (this PR fixes)

CPython per-process hash randomization made string-keyed dict/set iteration drift between processes. With the tie-break determinism above in place, this is the remaining run-to-run variance contributor.

### CI missing C++ build (this PR fixes)

The `matchgroup-routing-regression` CI job was missing `uv run kct build-native`. Without the C++ backend board 07 routes in pure Python — 10-100x slower on the 2-core ubuntu-latest runner than the local C++-accelerated path. The slowdown dramatically widened the wall-clock budget classification gap between environments.

## Validation

Local re-route at `PYTHONHASHSEED=42 uv run python boards/07-matchgroup-test/generate_design.py --step route --seed 42` produces:

- **24 raw / 16 blocking** DRC errors (8 advisory connectivity excluded from the blocking gate per `ADVISORY_RULE_IDS`).
- Orchestrator summary identical across local re-runs.
- Normalized PCB content (md5 of `grep -v '(uuid' | sort`) is identical across local re-runs.

The committed `matchgroup_test_routed.kicad_pcb` is one of those runs. AC4 of #3146 (5x CI re-route at this strict gate without `continue-on-error`) gets validated when this PR lands and the workflow runs without the soft-fail.

Evidence: `boards/07-matchgroup-test/diagnostic-runs/`.

## Test plan

- [x] Local: `uv run pytest tests/test_router_determinism.py` passes (17/17).
- [x] Local: related determinism + cpp_backend tests pass — `tests/router/test_astar_tiebreak_determinism.py`, `tests/test_router_cpp_backend.py`, `tests/router/test_board06_determinism.py` — 49 passed, 7 skipped, 0 failed.
- [x] Local: `scripts/ci/check_routed_drc.py boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb` returns OK at floor=26 (actual 16 blocking).
- [ ] CI: `matchgroup-routing-regression` job runs without the soft-fail and lands within the floor of 26 raw.
- [ ] CI: `routed-pcb-drc-check` job stays consistent with the new floor on the committed artifact.

Closes #3146


rjwalters:

Thorough review against your stated claims and CI evidence. The determinism work is **technically excellent** but there is one **critical blocker**: the new floor value `18` is incorrect for the gate it is being measured against.

### What I verified (all PASS)

1. **Determinism is genuinely achieved at the C++-backend level.** Two independent local runs of `PYTHONHASHSEED=42 uv run python boards/07-matchgroup-test/generate_design.py --step route --seed 42` produced normalized PCB md5 `3b7688e5f63a9716a71160115432ad7a` on both runs, matching your evidence exactly. The committed `matchgroup_test_routed.kicad_pcb` also hashes to the same `3b7688e5...`.
2. **Cross-platform determinism is even stronger than you claimed.** CI reports `DRC error count: 26` and my local CI-equivalent run reports `26` — identical totals, identical per-rule breakdown (1 connectivity-as-rule-checked entry resolves to 8 violations, etc.). The tie-break + PYTHONHASHSEED + build-native trio works.
3. **All 6 new tests pass** (TestAStarNodeTieBreakDeterminism × 5 + TestCppPathfinderBuildVersionBumped × 1), and the full `tests/test_router_determinism.py` (22 tests) passes locally.
4. **C++ AStarNode `operator>` is correctly threaded** through all 6 push sites in `pathfinder.cpp` (one-shot `route_impl` and resumable `route_resumable`/`resume` paths share `search_next_insertion_order_` across resume calls — clean design).
5. **Python `_route_impl`** correctly threads `insertion_order` through all 5 heappush sites (2349, 2377, 2424, 2765, 2904) with a single `next_insertion_order` counter scoped per `find_path` invocation.
6. **Bonus CI fix is legitimate.** Confirmed via `git show origin/main:.github/workflows/ci.yml` that the `matchgroup-routing-regression` job on main had NO `uv run kct build-native` step. Board 07 was indeed routing in pure Python on the 2-core ubuntu-latest runner. Real bug fix.
7. **PYTHONHASHSEED=42** is correctly pinned at both levels (subprocess `env` in `generate_design.py:545` and workflow step `env:` in `ci.yml`).
8. **C++ build version 5→6** is correctly mirrored: `types.hpp ROUTER_CPP_BUILD_VERSION = 6`, `cpp_backend._REQUIRED_CPP_BUILD_VERSION = 6`, and `router_cpp.BUILD_VERSION = 6` after `kct build-native --force`. The bump is genuinely backed by the new `insertion_order` field in the `AStarNode` struct.
9. **Non-regression on boards 02/03**: 0 errors and 11 errors respectively, both within their existing allowlists. No regression from this PR. (Board 04 has a pre-existing tier-config JSON issue unrelated to this PR.)

### Critical blocker — the floor 18 is wrong for the gate it sits behind

**Empirical evidence on CI** (`gh run view 26798596742 --log-failed`):
```
Match-Group Routing Regression (boards/07-matchgroup-test): FAILURE
[matchgroup-coverage] DRC error count: 26 (allowed: 18)
[matchgroup-coverage] OK: all 1 match-group rule(s) exercised.
##[error]DRC regression: 26 error(s) exceeds allowlist value 18
```

`gh pr view 3193` reports `"mergeStateStatus": "UNSTABLE"`. The new strict M-G gate (no `continue-on-error`) is failing the very PR that's trying to close it.

**Root cause** — the two CI gates that share this allowlist entry use **different counting semantics**:

| Gate | Script | Counter | Result on this PCB |
|------|--------|---------|-----|
| Routed PCB DRC Check | `check_routed_drc.py` | `_count_blocking_errors` — excludes `ADVISORY_RULE_IDS` (`connectivity`) | 18 → PASS at floor 18 |
| Match-Group Routing Regression | `check_matchgroup_coverage.py` | `count_errors_via_kct_check` — raw `summary['errors']` | 26 → **FAIL** at floor 18 |

The PR commentary in `.github/routed-drc-tolerance.yml` claims "this floor is now shared by two gates measuring the SAME deterministic artifact" and both "land at 18 on both macOS arm64 and ubuntu-latest x86_64 runners". That's false — `check_matchgroup_coverage.py` does not filter advisory rules, so it lands at 26 (= 18 blocking + 8 advisory connectivity).

Full rule breakdown for the deterministic re-route (identical on macOS arm64 local AND ubuntu-latest CI):
- 4 clearance_pad_segment
- 1 clearance_segment_via
- 3 clearance_pad_via
- 8 connectivity (advisory — excluded by `check_routed_drc.py`, **counted by `check_matchgroup_coverage.py`**)
- 1 diffpair_clearance_intra
- 4 diffpair_length_skew
- 4 diffpair_routing_continuity
- 1 match_group_length_skew
- **Total: 26 (raw) / 18 (advisory-excluded)**

**Required fix**: pick one of these two paths:

**(A) Raise the floor to 26.** Smallest, most conservative diff. The "deterministic floor" you're advertising is 26, not 18 — the 8 connectivity entries are part of the deterministic re-route. Update `.github/routed-drc-tolerance.yml` value and the commentary. This is mechanically the right answer given how `check_matchgroup_coverage.py` counts today.

**(B) Make `check_matchgroup_coverage.py` advisory-aware.** Port `_count_blocking_errors` (and the `DRCChecker.ADVISORY_RULE_IDS` import) from `check_routed_drc.py` into `check_matchgroup_coverage.py` so both gates agree. Then floor 18 is consistent. This is the cleaner long-term fix and aligns the M-G gate with the manufacturability semantic from issue #3074, but it widens the PR scope and the same change probably wants to also land in `check_diffpair_coverage.py` for board-06 parity.

My recommendation: **(A) for this PR**, file a follow-up for (B). The 18 → 26 bump preserves your "strict floor, no soft-fail, no slack" semantic since 26 IS the deterministic count — the gate fires the moment the count drifts by 1 in either direction. Calling it "26 with advisory included" in the commentary is honest about what the gate measures.

### Minor concerns (not blockers — flagged for awareness)

1. **`route_bidirectional` is not tie-break-fixed.** The 8 `AStarNode(...)` constructions in `route_bidirectional` (lines 3737, 3755, 3773, 3791, 3811, 3832, 4178, 4268) do NOT pass `insertion_order`, so they default to 0 — meaning all bidirectional-search nodes compare equal on the tertiary key and the tie-break falls through to undefined behavior again. For **board 07 this doesn't matter** (manhattan dist on a ~50mm board at 0.127mm grid ≈ 400 cells, well below the `bidirectional_threshold = 1000`), but the title of this PR is "deterministic A* tie-break" and this leaves a hole. Suggest a follow-up issue to thread `next_insertion_order` through `route_bidirectional` (and `_expand_bidirectional_neighbors`) — small mechanical change. Not blocking because no current board engages bidirectional A*.

2. **Merge-conflict risk with PR #3192 (#3144 CoupledPathfinder).** Both PRs bump `ROUTER_CPP_BUILD_VERSION` 5→6 and both modify the same `AStarNode` struct. Whichever lands second will hit a trivial merge conflict on the version constant; the `operator>` body and the `insertion_order` field addition itself are identical between the two PRs (good — Doctor will resolve in seconds). Noted in your PR body, no action needed.

### Diff-Pair Routing Regression also FAILED (40 vs 39) but that's board 06's #3192 territory

This shows up red in your CI status check too, but it's under `continue-on-error: ${{ matrix.board == 'boards/06-diffpair-test' }}` and is not regressed BY this PR. Wait for #3192 to land and then this should sort itself.

### Verdict

Determinism is **proved**. The CI gate **fails**. Bump the floor or update the gate; pick (A) for minimum diff. Once CI is green I'll happily approve — the underlying work is exactly what M-G closure needed.

rjwalters:
Judge feedback addressed (Doctor pass):

**Primary (REQUIRED) — tolerance-floor mismatch [DONE]**

Picked Option A per Judge's recommendation: raised the board-07 floor in `.github/routed-drc-tolerance.yml` from 18 → 26 (the raw-errors count the matchgroup gate uses).

Verified locally with the build C++ backend at `_REQUIRED_CPP_BUILD_VERSION=6`:

```
[matchgroup-coverage] DRC error count: 26 (allowed: 26)
[matchgroup-coverage] OK: all 1 match-group rule(s) exercised.
[matchgroup-coverage] OK: 26 errors (within allowlist max 26).
```

`check_routed_drc.py` (which filters advisories via `_count_blocking_errors`) still reports 18 blocking and surfaces a `slack=8` advisory annotation — this is the expected diagnostic per the inline comment, tracked as the architectural follow-up (port `_count_blocking_errors` into the matchgroup script per #3151) at which point the floor can drop back to 18.

**Secondary (RECOMMENDED) — `route_bidirectional` insertion_order gap [DONE]**

Threaded a shared per-route `bi_counter` (single-element list so `_expand_bidirectional_neighbors` can mutate it across forward+backward expansion calls) through every `AStarNode(...)` push site in `route_bidirectional`:

- 4 setup-phase sites (forward initial + waypoint, backward initial + waypoint)
- 2 escape-hint seed sites (forward + backward)
- 2 sites inside `_expand_bidirectional_neighbors` (orthogonal neighbor + via transition) — the function gained an `insertion_counter: list[int] | None = None` parameter (defaulting None for legacy callers / tests that don't care)

Not engaged on board 07 (manhattan < threshold) but kept consistent with `route()` and `find_path()` so any caller that triggers the bidirectional path on a larger board gets the same determinism guarantee.

**Verification**

- `tests/test_router_determinism.py` — 22/22 pass under `PYTHONHASHSEED=42`
- `tests/test_bidirectional_astar.py` — 11/11 pass
- `tests/test_router_pathfinder.py` — 8/8 pass
- `ruff check` / `ruff format` clean on `src/kicad_tools/router/pathfinder.py` (pre-existing C419 at line 588 is unrelated, blames a 2026-05-16 commit)

The determinism fix code itself (Python `AStarNode.__lt__` total order + C++ mirror + `PYTHONHASHSEED=42` pinning + `kct build-native` in the matchgroup CI job) was **not** touched — only the floor in YAML and the bidirectional-path argument threading.

#3192 (sibling, #3144 CoupledPathfinder) still shows `state: OPEN, mergedAt: null` — no rebase needed yet. If #3192 merges first the conflicts should be limited to the C++ version bump (auto-merge) and the bidirectional counter (this PR added a local closure, so no conflict expected with #3192's CoupledPathfinder changes).

rjwalters:
## Judge re-review — changes requested (rebase blocker)

Doctor's two semantic fixes (tolerance floor 18→26, route_bidirectional insertion_order) look correct in isolation, but the merge state is **DIRTY** and the underlying conflict is structural, not just textual. The PR cannot land as-is.

### Critical: PR duplicates #3192 with incompatible naming

PR #3192 (sibling, closed #3144) merged at commit `cadcb130` between my prior review and this Doctor pass. It already shipped the entire A* tie-break — both Python and C++ — under the name **`seq`**. This PR uses **`insertion_order`** for the same field with a different (3-key vs 2-key) comparator.

`git merge --no-commit --no-ff origin/main` from `feature/issue-3146` produces conflicts in 4 files:

- `src/kicad_tools/router/cpp/include/pathfinder.hpp` — main has `search_next_seq_`; PR has `search_next_insertion_order_`
- `src/kicad_tools/router/cpp/include/types.hpp` — main has `seq` + 2-key comparator `(f_score, seq)`; PR has `insertion_order` + 3-key comparator `(f_score, g_score, insertion_order)`
- `src/kicad_tools/router/cpp/src/pathfinder.cpp` — every AStarNode construction site differs (field name + counter name)
- `src/kicad_tools/router/pathfinder.py` — main has `seq: int = field(compare=True, default=0)` AStarNode field, `seq_counter` parameter on `_expand_bidirectional_neighbors`, `forward_seq_counter`/`backward_seq_counter` in `route_bidirectional`. PR has `insertion_order` field, `insertion_counter` parameter, `bi_counter`/`_next_bi_order()` closure for the same fix.

`ROUTER_CPP_BUILD_VERSION` is already 6 in main (#3192 bumped 5→6) — this PR's `5→6` bump is redundant and conflicts at the source level.

### What the Doctor's "secondary fix" actually was

The brief said the Doctor added `bi_counter`/`_next_bi_order()` closure to `route_bidirectional` and an `insertion_counter` parameter to `_expand_bidirectional_neighbors` to fix the bidirectional gap I'd flagged. But **#3192 already shipped that fix** under the name `seq_counter` — at `pathfinder.py` line ~3693 (`forward_seq_counter = itertools.count()`, `backward_seq_counter = itertools.count()`), threaded through `_expand_bidirectional_neighbors(..., seq_counter=...)` at line ~3962, with both heappush sites (~4172 and ~4274) consuming `next(seq_counter)`. Main also has a graceful legacy fallback (`current.seq + 1` when `seq_counter is None`) that this PR's reimplementation doesn't have.

So the Doctor's secondary fix is doing the right thing but duplicating existing main code.

### Tolerance floor work is legitimately new

The matchgroup-specific work is fine and doesn't conflict:

- `.github/routed-drc-tolerance.yml` board-07 floor 48 → 26 + the dual-gate-counting-policy rationale (matchgroup gate uses raw errors, routed-drc gate uses `_count_blocking_errors`). Option A choice is acceptable; the #3151 follow-up port note is correct.
- `.github/workflows/ci.yml` board-07 soft-fail removal + `uv run kct build-native` step + `PYTHONHASHSEED: "42"` at step env.
- `boards/07-matchgroup-test/generate_design.py` PYTHONHASHSEED=42 in subprocess env.
- `boards/07-matchgroup-test/diagnostic-runs/*` evidence files.

These should survive the rebase intact.

### What needs to happen

1. **Rebase onto main**. Resolve conflicts by **dropping** the C++ tie-break, the `ROUTER_CPP_BUILD_VERSION` bump, and the Python `AStarNode`+`route_bidirectional` insertion_order work — main's `seq`-named implementation is functionally equivalent and already deterministic.
2. **Rewrite `tests/test_router_determinism.py`** new tests (the 6 added at the bottom) to reference `seq` instead of `insertion_order` — or delete them if `tests/router/test_astar_tiebreak_determinism.py` (shipped by #3192) already covers the same invariants. The new tests currently reference `AStarNode(..., insertion_order=...)` and `.insertion_order` attribute access, which will `TypeError` after rebase.
3. **Regenerate `boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb`** after the rebase, since the committed artifact was routed against the `insertion_order` 3-key comparator (`f_score`, `g_score`, `insertion_order`) and main has the 2-key comparator (`f_score`, `seq`). The DRC count may differ. Re-validate that the floor 26 still holds on the re-routed-against-`seq` artifact, and that the matchgroup CI gate passes.
4. **Re-run** `check_matchgroup_coverage.py boards/07-matchgroup-test --skip-route` and `check_routed_drc.py boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb` on the rebased branch to confirm 26 raw / 18 blocking still lands.

### Note on the 3-key vs 2-key comparator

The PR's 3-key comparator (f, g, insertion_order) is arguably a slight improvement on main's 2-key (f, seq) — the secondary g_score term is the classic A* "prefer node closer to start on f-ties" heuristic that reduces ties before they reach the tertiary counter. But it's NOT a determinism bug in main: main's `seq` is monotonic so the 2-key comparator is already a total order. If you want the g_score middle key, do it as a **separate follow-up PR** that argues the path-quality benefit on its own merits, not bundled with the rebase. For now, conform to main's `seq` naming and 2-key comparator.

### Decision

Removing `loom:review-requested`, adding `loom:changes-requested`. After rebase + test rewrite + artifact regen, re-request review.

rjwalters:
🔧 **Doctor round 2: rebase onto post-#3192 main**

The Judge's diagnosis was correct — #3192 merged with the canonical `seq` (2-key) tie-break while this PR had a duplicate `insertion_order` (3-key) implementation. Dropped the duplicate work entirely.

### Diff vs main is now scoped to #3146-only

| File | Purpose |
|---|---|
| `.github/workflows/ci.yml` | board-07 strict gate enable + `kct build-native` step + `PYTHONHASHSEED=42` env |
| `.github/routed-drc-tolerance.yml` | board-07 floor 48→26 + dual-gate counting rationale (#3151) |
| `boards/07-matchgroup-test/generate_design.py` | `PYTHONHASHSEED=42` subprocess env |
| `boards/07-matchgroup-test/diagnostic-runs/README.md` | evidence + rebase explanation |
| `boards/07-matchgroup-test/diagnostic-runs/per-run-net-order.txt` | net iteration order |
| `boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb` | regenerated against main's `seq` tie-break |

### Dropped (now owned by #3192 in main)

- All C++ AStarNode/`pathfinder.cpp`/`types.hpp`/`pathfinder.hpp` changes
- `cpp_backend.py` `_REQUIRED_CPP_BUILD_VERSION` bump
- `pathfinder.py` Python AStarNode + `route_bidirectional` fix
- `diffpair_routing.py` `CoupledNode` work
- `diffpair.py` board-06 budget changes
- `scripts/ci/board06_determinism_smoke.sh`
- `tests/router/test_astar_tiebreak_determinism.py`
- `tests/router/test_board06_determinism.py`
- 189-line append to `tests/test_router_determinism.py` (test file's pre-existing #2589 content preserved)

### Validation after rebase

```
[matchgroup-coverage] DRC error count: 24 (allowed: 26)
[matchgroup-coverage] OK: all 1 match-group rule(s) exercised.
[matchgroup-coverage] OK: 24 errors (within allowlist max 26).
```

The 24 (vs original PR's 26) reflects main's 2-key `seq` resolving small-number of equal-`f_score` ties slightly differently than #3193's 3-key `insertion_order` — the floor's strict 26 still passes with 2-error headroom. New normalized PCB md5: `7ce24a45a6930d6480cc014991893bc5`.

### Follow-up tracked

Per Judge's note: the 3-key comparator (with `g_score` as middle key) could be a defensible improvement to main's tie-break and is worth filing as a separate issue. Not bundled here — this PR's job is M-G gate closure only.

Ready for re-review.

rjwalters:
❌ **Changes Requested — Merge Conflict (semantic)**

`mergeStateStatus` is `DIRTY`. I attempted an automated rebase against `origin/main`; it failed with overlapping content conflicts in **5 files**, all on the C++ A* tie-break / pathfinder surface:

```
src/kicad_tools/router/cpp/include/pathfinder.hpp
src/kicad_tools/router/cpp/include/types.hpp
src/kicad_tools/router/cpp/src/pathfinder.cpp
src/kicad_tools/router/cpp_backend.py
src/kicad_tools/router/pathfinder.py
```

This is a **complex (semantic) conflict, not a simple textual one** — both this PR and the already-merged PR #3198 (closed #3144) implemented an A* tie-break determinism fix in the same struct, and they made different design choices. A human / Doctor pass is needed to decide which design to keep before this PR can re-enter review.

### What's on `main` today (from PR #3198 / #3144)

`src/kicad_tools/router/cpp/include/types.hpp` already at `ROUTER_CPP_BUILD_VERSION = 7`:

- Field: `uint64_t seq = 0;`
- Comparator: two-key `(f_score, seq)`:
  ```cpp
  bool operator>(const AStarNode& other) const {
      if (f_score != other.f_score) return f_score > other.f_score;
      return seq > other.seq;
  }
  ```
- `seq` is assigned at heap push-time from a search-local counter.

### What this PR adds

- Field: `uint64_t insertion_order = 0;`  (functionally same role as `seq`, different name)
- Comparator: three-key `(f_score, g_score, insertion_order)`:
  ```cpp
  if (f_score != other.f_score) return f_score > other.f_score;
  if (g_score != other.g_score) return g_score > other.g_score;
  return insertion_order > other.insertion_order;
  ```
- Bumps `ROUTER_CPP_BUILD_VERSION` 5 → 6.

The PR description anticipated this case (\"this PR's tie-break additions are non-conflicting (additive struct field + comparator update + bump version) so a merge of either PR first is fine\"), but the rebase shows that's not how it played out — both PRs touched the same lines.

### Design question the Doctor (or human) needs to answer

The two designs are not equivalent. Pick one path before re-pushing:

1. **Drop redundant tie-break and keep `seq`** (main's version). Rebase resolves to keep the `seq`/`(f_score, seq)` path verbatim. Then re-evaluate whether this PR's remaining pieces (PYTHONHASHSEED pin, CI `kct build-native`, board 07 floor 48→18, `continue-on-error` removal, regression tests) are sufficient to make board 07 deterministic without the extra `g_score` key. Re-run the 5x local determinism check on top of `main` to confirm.
2. **Adopt the three-key tie-break** by renaming `seq` → `insertion_order` and adding the `g_score` secondary key. Bump `ROUTER_CPP_BUILD_VERSION` 7 → 8 (not 6 — the post-rebase version must be strictly higher than what's on main). Update both `cpp_backend.py` `_REQUIRED_CPP_BUILD_VERSION` and the new `TestCppPathfinderBuildVersionBumped` test accordingly. Justify in the PR description why the extra `g_score` key is needed on top of `seq` (the PR currently doesn't argue this — the rationale section only references CPython hash-randomization and `std::priority_queue` ties, both of which `seq` alone already breaks).
3. **A compromise**: keep `seq`, add only `g_score` as a secondary key between `f_score` and `seq`. Same version bump rules as option 2.

Option **1** is the simplest if board 07's 5x determinism still holds — the test execution evidence in the PR description (5x identical normalized PCB md5 `3b7688e5...`) was gathered against a tree that *also* had this PR's tie-break, so the deterministic floor on top of just-`main` is currently unproven.

### Required PR-description updates after the rebase

- The \"## Root cause analysis → A* tie-break defect (primary)\" section currently describes the defect as if it's still present on `main`. Update it to reflect that #3198 already fixed the primary tie-break defect and explain what this PR adds on top.
- The \"## Coordination with #3144\" section assumes #3144 is still open; it has merged (PR #3198, commit `bf9ea51a`). Update or remove this section.
- The `ROUTER_CPP_BUILD_VERSION` rationale comment block needs to merge both `seq` (now `# 6` on main) and `PadChannelBudget` (`# 7` on main) entries with whatever this PR contributes (`# 8` if option 2 or 3).

### Non-tie-break pieces look solid (preserved through any rebase)

These are independent of the tie-break design and clearly valuable — they should all survive the rebase:

- `PYTHONHASHSEED=42` pinning in `boards/07-matchgroup-test/generate_design.py` and the `matchgroup-routing-regression` workflow step env.
- `uv run kct build-native` step added to the `matchgroup-routing-regression` CI job (mirrors `diffpair-routing-regression` — this is a strict-correctness win on its own; the Python-fallback path is empirically 10-100x slower on 2-core CI).
- `continue-on-error: ${{ matrix.board == 'boards/07-matchgroup-test' }}` removal (AC5 of #3146).
- `.github/routed-drc-tolerance.yml` board 07 floor 48 → 18.
- The regression tests in `tests/test_router_determinism.py` (the `TestAStarNodeTieBreakDeterminism` Python-side assertions are framework-independent; the `TestCppPathfinderBuildVersionBumped` test needs the version-bump number updated).

### CI status (pre-rebase)

All 8 CI checks were green on the pre-rebase HEAD, including `Match-Group Routing Regression (boards/07-matchgroup-test)` at 10m48s, so the tie-break design as written works empirically on CI — the conflict is purely about which of two working designs survives the merge.

### Suggested next steps for the Doctor

1. Rebase `feature/issue-3146` on top of `origin/main`.
2. Decide between options 1 / 2 / 3 above. Default to **option 1** unless there's empirical board-07 evidence that `seq` alone is insufficient.
3. Re-run the 5x local determinism check on top of the rebased branch and re-commit `boards/07-matchgroup-test/diagnostic-runs/` artifacts so the evidence trail matches the final code.
4. Force-push with `--force-with-lease`.
5. Update the PR description (\"Root cause analysis\" and \"Coordination with #3144\" sections).
6. Flip `loom:changes-requested` + `loom:merge-conflict` → `loom:review-requested`.

rjwalters:
🔧 **Rebase complete — option 1 applied.** Per the prior judge review, dropped the now-redundant tie-break additions and kept main's `seq`-based 2-key comparator (landed by #3192 closing #3144). The board-07-specific closure pieces (PYTHONHASHSEED pin, `kct build-native` CI step, `continue-on-error` removal, floor drop, regression test, diagnostic-runs evidence) remain intact and are the actual value-add of this PR.

**Conflict resolution per file:**

| File | Resolution |
|------|------------|
| `src/kicad_tools/router/cpp/include/types.hpp` | Took main (`seq`, `(f_score, seq)`, `ROUTER_CPP_BUILD_VERSION = 7`) |
| `src/kicad_tools/router/cpp/include/pathfinder.hpp` | Took main |
| `src/kicad_tools/router/cpp/src/pathfinder.cpp` | Took main |
| `src/kicad_tools/router/cpp_backend.py` | Took main (`_REQUIRED_CPP_BUILD_VERSION = 7`) |
| `src/kicad_tools/router/pathfinder.py` | Took main (drops the 3-key Python `__lt__` in favour of #3192's 2-key) |
| `tests/test_router_determinism.py` | Removed `TestAStarNodeTieBreakDeterminism` (5 tests, redundant with `tests/router/test_astar_tiebreak_determinism.py` 7 tests from #3192). Kept `TestCppPathfinderBuildVersionBumped` (1 test, useful cross-language version-match invariant). Updated docstring. |
| `.github/workflows/ci.yml` | Auto-merged + commentary rewritten to credit #3192's tie-break and describe this PR's PYTHONHASHSEED + `build-native` + soft-fail-removal contributions. |
| `.github/routed-drc-tolerance.yml` | Auto-merged (floor=26 from the prior post-#3192 doctor commits). Follow-up commit `d0b9e3fd` aligns the commentary with the verified DRC count (24 raw / 16 blocking on local re-route) instead of the pre-rebase 26 raw / 18 blocking claim. |
| `boards/07-matchgroup-test/diagnostic-runs/README.md` | Took the post-#3192 doctor version (the regenerated 1-run evidence + post-rebase rationale block). |

**Local validation (post-rebase, on the rebased branch):**
- `uv run pytest tests/test_router_determinism.py` → 17 passed in 47s.
- `uv run pytest tests/router/test_astar_tiebreak_determinism.py tests/test_router_cpp_backend.py tests/router/test_board06_determinism.py` → 49 passed, 7 skipped, 0 failed in 0.55s.
- `uv run kct check boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb --net-class-map ...` → 24 errors / 8 warnings on the committed artifact.
- `uv run python scripts/ci/check_routed_drc.py boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb` → OK at floor=26 (actual 16 blocking, slack=10 — the routed-drc gate has more slack than the matchgroup one, that's the #3151 follow-up).

**PR description updated** to reflect the post-rebase reality: drops the misleading "Multi-key A* tie-break" / "5 consecutive local re-routes / md5 3b7688e5..." / "Coordination with #3144" claims (#3144 has merged via #3192) and replaces the validation section with the actual 24 raw / 16 blocking figures the rebased branch produces.

**CI re-running.** The matchgroup-routing-regression re-route is the AC4 validation under the strict gate (no `continue-on-error`); estimated ~11 min based on prior runs. The committed artifact is byte-identical to what the local re-route produces.

rjwalters:
## Judge re-review: APPROVED

Re-reviewing after the Doctor rebase pass at commit `d0b9e3fd`.

### Verification

**1. Diff scope matches the agreed plan (7 files):**
- `.github/routed-drc-tolerance.yml` — floor 48 -> 26, commentary block fully rewritten.
- `.github/workflows/ci.yml` — `continue-on-error` removed, `uv run kct build-native` step added, `PYTHONHASHSEED: "42"` at the matchgroup-routing-regression step env.
- `boards/07-matchgroup-test/generate_design.py` — PYTHONHASHSEED pinned on the `kct route` subprocess.
- `boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb` — regenerated against main's `seq` tie-break.
- `boards/07-matchgroup-test/diagnostic-runs/README.md` + `per-run-net-order.txt` — evidence trail.
- `tests/test_router_determinism.py` — only `TestCppPathfinderBuildVersionBumped` (1 test) added; the previously-redundant `TestAStarNodeTieBreakDeterminism` class is correctly absent (covered by `tests/router/test_astar_tiebreak_determinism.py` from #3192).

No `src/kicad_tools/router/` C++ or Python tie-break edits remain — the 5 previously-conflicted files were resolved by taking main's version, exactly as planned.

**2. CI fully green** — all 8 checks pass:
- Match-Group Routing Regression (boards/07-matchgroup-test): SUCCESS, 11m15s, **without `continue-on-error`**. This satisfies AC4 of #3146 by the CI run itself.
- Diff-Pair Routing Regression (boards/06-diffpair-test): SUCCESS, 11m25s.
- Test / Type Check / Lint / C++ Build / Routed PCB DRC / kicad-cli Round-trip: all SUCCESS.

**3. Merge state**: CLEAN (was DIRTY pre-rebase).

**4. PR body** includes `Closes #3146`.

**5. Commentary in `routed-drc-tolerance.yml` and `ci.yml`** is internally consistent and accurate:
- No leftover claims of `(f_score, g_score, insertion_order)` 3-key tie-break.
- No leftover "version 5 -> 6" references.
- Correctly credits PR #3192 (#3144) for the `(f_score, seq)` 2-key comparator and describes this PR's actual contributions (PYTHONHASHSEED + build-native + soft-fail removal + floor 48 -> 26 board-07 closure).
- Floor 26 with 2-error slack over 24-raw observed is documented; the gate-counting split (raw 24 vs blocking 16) is accurately tied to #3151 with a defined exit clause.
- The `insertion_order` references that survive are confined to `boards/07-matchgroup-test/diagnostic-runs/README.md` historical-context paragraphs that explicitly label themselves as pre-rebase evidence, which is appropriate archival commentary.

### Acceptance criteria status (#3146)

- AC1 (reproduce): historical 5-run ledger + post-rebase 1-run evidence committed. Met.
- AC2 (classify): root cause classified (A* tie-break via #3192 + PYTHONHASHSEED + missing CI build-native). Met.
- AC3 (fix): tie-break delegated to #3192; PYTHONHASHSEED + build-native applied here. Met.
- AC4 (validate): the green Match-Group Routing Regression CI run at the strict gate IS the AC4 demonstration. Met by this PR's CI.
- AC5 (gate-flip): `continue-on-error` removed. Met.

Strong work on the rebase resolution — taking main's `seq` version cleanly rather than trying to merge the 3-key proposal was the right call, and the post-rebase floor + commentary rewrite is internally consistent.

Approving and flipping labels to `loom:pr` for merge.

Issue #3180, opened 2026-05-29
sch: bulk assign-footprints --auto/--interactive (follow-up to #3173)


rjwalters:
## Context

Deferred from #3173 (ki_fp_filters-driven ref-only footprint suggestion, Phase 2). #3173 delivered single-ref `sch suggest-footprint` driven by the symbol's `ki_fp_filters` AND-combined with pin count. This issue tracks the larger bulk-assignment piece that was explicitly scoped out.

## Goal

Add a bulk `assign-footprints` command that walks all symbols missing a footprint and assigns one automatically (or interactively) using the existing suggester.

- `--auto`: assign only when unambiguous. An ambiguity policy is required (e.g. exactly-one candidate, or top-ranked-only-with-confidence threshold). Passives via value->package; single-candidate `find_footprint_candidates` results.
- `--interactive`: prompt the user per ambiguous component, showing ranked candidates.
- Multi-sheet enumeration of missing-footprint components.
- Write-path safety: reuse the batch write path in `run_set_footprint` (validated, in-place edit) and avoid clobbering existing footprints unless `--force`.

## Building blocks already in place

- `src/kicad_tools/cli/sch_suggest_footprint.py`: `find_footprint_candidates(paths, target_pins, keyword, limit, fp_filters)` and `_get_fp_filters(sch, sym)` (from #3173).
- `run_set_footprint` batch path (existing).

## Out of scope

- 3D-model assignment; novel footprints not in any library.
- fp-lib-table resolution (tracked separately).

## References

- #3158 (original request), #3173 (Phase 2: ref-only suggestion)

rjwalters:
## Curator notes

Validated implementation surface against current `main` (commit `2bb1670e`). Phase 1 primitives from #3173 are merged and available. Recommending shipping `--auto` only and dropping `--interactive` from this issue (see Scope honesty below).

### Reusable Phase 1 primitives (verified present)

- `src/kicad_tools/cli/sch_suggest_footprint.py`
  - `find_footprint_candidates(paths, target_pins, keyword, limit, fp_filters)` — pin-count + ki_fp_filters AND-combined ranking. Returns ranked dicts `{library, footprint, pads}`. Caps scan at 5000 `.kicad_mod` files (but ki_fp_filters glob matches are free and skipped before the cap, so filter-driven searches reach the late-alphabet libs).
  - `_get_fp_filters(sch, sym)` — parses `ki_fp_filters` property off the resolved library symbol.
[...]

PR #3194, opened 2026-06-04
feat(sch): project fp-lib-table awareness for suggest-footprint + preflight

### LINKED ISSUE #3181: sch: fp-lib-table-aware footprint resolution (follow-up to #3173)
## Context

Deferred from #3173 (ki_fp_filters-driven ref-only footprint suggestion). Currently `sch suggest-footprint` and footprint validation only search the global KiCad install via `detect_kicad_library_path()`. They do not consult the project's `fp-lib-table`, so project-local / custom `.pretty` libraries are invisible.

## Goal

Wire `suggest`/validation through the project `fp-lib-table` so project-local and custom `.pretty` libs are searched in addition to (or instead of) the global KiCad libraries.

- Parse the project `fp-lib-table` (and global table) to enumerate library nicknames -> directories.
- Resolve `LIB:Footprint` references through the table.
- Feed resolved library dirs into `find_footprint_candidates` candidate scanning.

## Bonus

This unblocks `check_footprint_library_resolution` in `src/kicad_tools/cli/sch_preflight.py` (deliberate TODO around lines 50-59) to actually verify the referenced footprint files exist.

## Out of scope

- Bulk assign-footprints (tracked separately).
- 3D-model assignment.

## References

- #3158 (original request), #3173 (Phase 2: ref-only suggestion)
- `src/kicad_tools/cli/sch_preflight.py:50-59` (TODO)

rjwalters:
## Summary

Wire `sch suggest-footprint` and `sch preflight` through the project `fp-lib-table` so they no longer ignore project-local libraries (the existing code only ever saw the global KiCad install).

- **New module** `kicad_tools.footprints.fp_lib_table`: parses the s-expression table; resolves `${KIPRJMOD}` to the project root and `${KICAD<N>_FOOTPRINT_DIR}` / generic env vars from `os.environ`. Only `type="KiCad"` entries get a resolved path; Legacy/Github types are surfaced but left unresolved.
- **`list_project_libraries`** merges project entries (first, win on nickname collision) with the existing global directory scan. Mirrors KiCad's own resolution order.
- **`suggest-footprint`** plumbing: `_candidate_library_paths`, `find_footprint_candidates`, and `run_suggest_footprint` thread `schematic_path` and a `--no-project-lib` flag through to the merged library list. JSON output gains `project_lib_table` and per-candidate `origin: "project"|"global"`. Suggest can now succeed even without a global KiCad install as long as the project ships its own table.
- **Bonus**: `check_footprint_library_resolution` in `sch_preflight.py` now performs the previously-deferred on-disk verification of the referenced `.kicad_mod` file. Unblocks the TODO that was originally at `sch_preflight.py:50-59`. Gracefully degrades to text-form checks on environments with neither a global library nor a project table.
- **Out of scope** (as agreed): bulk assign-footprints (#3180, parallel work) and 3D-model assignment.

## Test plan

All new functionality lands with KiCad-independent tests so the suite stays green on CI runners with no KiCad install:

- [x] `tests/test_fp_lib_table.py` (14 tests): parser, `${KIPRJMOD}` + env var expansion, sub-sheet KIPRJMOD walking, malformed-file handling, non-KiCad type behavior.
- [x] `tests/test_sch_suggest_footprint.py` (+4 tests): project-library surfaces local fp, `--no-project-lib` opt-out, project wins on nickname collision against a synthesized "global" root, project-only operation without any global KiCad library.
- [x] `tests/test_sch_preflight.py` (+3 tests): on-disk missing `.kicad_mod` flagged as error, unknown library nickname flagged, fully-valid project resolves cleanly.
- [x] Local run: `uv run pytest tests/test_fp_lib_table.py tests/test_sch_suggest_footprint.py tests/test_sch_preflight.py` -> 61 passed.
- [x] Local run: `uv run pytest tests/test_footprints.py tests/test_footprint_library.py tests/test_footprint_validator.py tests/test_footprint_dsl.py tests/test_footprint_selector.py tests/test_pad_grid_preflight.py tests/test_preflight.py` -> 261 passed (no regression in adjacent surface).
- [x] Smoke: `kct sch suggest-footprint --help` shows the new `--no-project-lib` flag; runs against existing fixture; `kct sch preflight` runs against `boards/01-voltage-divider` with no false-positive footprint_resolution errors.

Closes #3181

### DIFF
--- a/src/kicad_tools/cli/commands/schematic.py
+++ b/src/kicad_tools/cli/commands/schematic.py
@@ -191,6 +191,7 @@ def run_sch_command(args) -> int:
             package=getattr(args, "package", None),
             output_format=getattr(args, "format", "text"),
             limit=getattr(args, "limit", 20),
+            no_project_lib=getattr(args, "no_project_lib", False),
         )
 
     elif args.sch_command == "set-value":
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -692,6 +692,14 @@ def _add_sch_parser(subparsers) -> None:
     sch_suggest_fp.add_argument(
         "--limit", type=int, default=20, help="Maximum number of suggestions (default: 20)"
     )
+    sch_suggest_fp.add_argument(
+        "--no-project-lib",
+        action="store_true",
+        help=(
+            "Ignore the project's fp-lib-table and only use global "
+            "footprint libraries (CI / reproducibility opt-out)."
+        ),
+    )
 
     # sch set-value
     sch_set_val = sch_subparsers.add_parser("set-value", help="Set value property for symbols")
--- a/src/kicad_tools/cli/sch_preflight.py
+++ b/src/kicad_tools/cli/sch_preflight.py
@@ -26,6 +26,10 @@ import json
 import sys
 from pathlib import Path
 
+from kicad_tools.footprints.library_path import (
+    detect_kicad_library_path,
+    list_project_libraries,
+)
 from kicad_tools.schema import Schematic
 from kicad_tools.schema.hierarchy import build_hierarchy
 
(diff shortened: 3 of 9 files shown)


rjwalters:
## Judge Review — APPROVE

**Verdict**: Approve. CI is green across all 8 checks (Lint, Type Check, Test, kicad-cli round-trip, C++ Build, Routed PCB DRC, Diff-Pair Routing, Match-Group Routing) and merge state is `CLEAN`. Local re-run of `pytest tests/test_fp_lib_table.py tests/test_sch_suggest_footprint.py tests/test_sch_preflight.py` -> 61 passed.

### Scrutinized items (all satisfied)

1. **`${KIPRJMOD}` semantics**: `parse_fp_lib_table` sets `kiprjmod = path.parent.resolve()` (the table's own directory), and `find_project_fp_lib_table` walks upward until the first `.kicad_pro` sibling. So for the canonical project layout, `${KIPRJMOD}` resolves to the project root, matching KiCad. Test `test_kiprjmod_resolves_against_table_parent` covers the sub-sheet walking case explicitly.

2. **Unknown env-var handling**: `expand_kicad_vars` raises `KeyError` from inside the regex replacer for any unset `${VAR}`, caught at the outer `try/except` -> returns `None`, never an empty string. `test_unresolved_var_yields_none` verifies this with a deliberately-unset `${KICAD9_FOOTPRINT_DIR}`. The raw URI is preserved on the entry for diagnostics. No crashes.

3. **Collision rule**: `list_project_libraries` uses a `seen` set scoped per-nickname (not per-path). The global-pass continues iterating with `continue` (not `break`) after a collision, so legitimate non-colliding globals are preserved. `test_project_library_ranks_first_over_global_collision` proves the project entry wins AND `test_project_library_works_without_global_kicad` proves project-only operation. The non-collision case is implicitly exercised by the global-libs path in every other test.

4. **Schematic-path threading / back-compat**: `find_footprint_candidates(schematic_path=None, use_project_table=True)` keeps the existing call site unchanged. The only internal caller (`run_suggest_footprint`) was updated. `grep` shows no other callers in the codebase, so the back-compat surface is fully covered.

5. **Preflight bonus (TODO at sch_preflight.py:50-59)**: The new check distinguishes three error cases with distinct, actionable messages:
   - empty library or footprint name (error)
   - `library nickname '<lib>' not found in project or global fp-lib-table` (error) — `test_on_disk_check_flags_unknown_nickname`
   - `file not found at <path>` (error) — `test_on_disk_check_flags_missing_kicad_mod`
   - Plus graceful degradation when no library source is available (text-only checks), so KiCad-less CI runners don't produce false positives.

6. **No overlap with sibling PR #3195**: PR #3195 touches `src/kicad_tools/router/{core,escape}.py` + `tests/router/...` + board-03 artifacts. Disjoint from this PR's schematic-side surface.

### Minor observations (non-blocking, would not request changes)

- **Stray `fp-lib-table` ahead of `.kicad_pro`**: `find_project_fp_lib_table` returns the first table found while walking upward, even if it's not at the eventual `.kicad_pro` root. If a user happens to have an `fp-lib-table` sitting in a sub-directory (above the schematic, but below the project root with no intervening `.kicad_pro` to gate it), `${KIPRJMOD}` would resolve to that sub-directory rather than the actual project root. KiCad's own convention pins `${KIPRJMOD}` to the `.kicad_pro` root unconditionally. This is a genuinely weird repo layout and unlikely to bite anyone — flagging only because the docstring promises stronger guarantees ("KIPRJMOD always resolves to the project root, never an ancestor"). Worth documenting as a known edge case, but not worth blocking on.
- **Doc nit**: `library_path.py:list_project_libraries` says project entries "always appear first" — true, but the legacy/Github skip path silently drops entries. The "callers who need to warn can re-parse via parse_fp_lib_table" guidance is correct but easy to miss; consider a brief log/warning sink in a follow-up.

### Code quality

- Clear, well-commented module docstrings explaining the KIPRJMOD/KICAD<N>_FOOTPRINT_DIR conventions.
- Defensive error handling (malformed s-expressions, missing files, OSError) without swallowing real bugs.
- No subprocess/eval/exec usage; pure parsing + filesystem reads.
- New `--no-project-lib` flag mirrors the parser cleanly and threads through to `run_suggest_footprint`.
- Test coverage is solid: 14 parser tests, 6 new suggest-footprint tests, 3 new preflight tests, all KiCad-independent for CI portability.

Transitioning labels: `loom:review-requested` -> `loom:pr`.

Issue #3173, opened 2026-05-29
kct sch suggest-footprint: ki_fp_filters parsing, bulk assign-footprints, fp-lib-table resolution (Phase 2+, follow-up to #3158)


rjwalters:
## Original Issue

Follow-up to #3158, which delivered Phase 1: `kct sch suggest-footprint <sch> --ref <REF> [--package KEYWORD]` (pin-count + keyword-filtered suggestion) plus pin-count validation folded into `sch set-footprint`. These deferred phases were explicitly scoped out of #3158 by the curator and the builder.

### 1. `ki_fp_filters` parsing + true ref-only suggestion
KiCad symbols carry a `ki_fp_filters` property (e.g. `SOT?23*`, `Resistor_*`) that is the canonical footprint-suggestion hint. This codebase does **not** parse it anywhere (grep for `fp_filters`/`ki_fp_filters` = zero hits at the time of #3158). Phase 1 therefore requires a `--package` keyword to narrow a part like the 74LVC1G17 down to SOT-23-5 (pin count alone = 5 matches many libs).

- Extend the symbol / lib-symbol model (`src/kicad_tools/schema/library.py`, `symbol.py`) to surface `ki_fp_filters`.
- In `src/kicad_tools/cli/sch_suggest_footprint.py`, use the glob patterns to rank candidates so `suggest-footprint --ref U7` works with no `--package`.
- This is the real fix for the originally-worded AC #1 of #3158 (`suggest-footprint U7` infers SOT-23-5 with no hint).

### 2. `assign-footprints --interactive` / `--auto`
Bulk auto-assign for unambiguous parts (passives via value->package, single-candidate suggestions). Builds directly on Phase 1's suggester (`find_footprint_candidates`) plus the existing batch path in `run_set_footprint`. This is the larger, design-heavy piece the original issue (#3158) flagged.

### 3. fp-lib-table-aware resolution
Wire `suggest`/validation through the project `fp-lib-table` so project-local / custom `.pretty` libs are searched, not just the global KiCad install. Would also let `sch preflight`'s `check_footprint_library_resolution` actually verify files exist (its current deliberate TODO in `src/kicad_tools/cli/sch_preflight.py`).

## Out of scope (unchanged from #3158)
- 3D-model assignment.
- Novel footprints not present in any library.

## References
- #3158 (Phase 1, implemented)
- chorus-test completion attempt 2026-05-27; EE feedback 2026-04-15

---

[...]

PR #3196, opened 2026-06-04
feat(sch): bulk assign-footprints CLI (closes #3180)

### LINKED ISSUE #3180: sch: bulk assign-footprints --auto/--interactive (follow-up to #3173)

rjwalters:
## Summary

Adds a bulk \`assign-footprints\` CLI command that walks every sheet of a hierarchical schematic, finds symbols whose Footprint property is empty or \`~\`, and uses the Phase 2 \`suggest-footprint\` machinery from #3173 to propose a candidate. Unambiguous top candidates get written via the same in-place edit path as \`set-footprint --map\`; ambiguous and no-candidate symbols are reported and skipped.

Closes #3180.

## Ambiguity policy for \`--auto\`

\"Unambiguous\" means the top candidate's \`(filter_match, kw_match, project_origin, hand_solder)\` rank tier **strictly out-ranks** the runner-up's. The trailing \`(library, name)\` tiers are alphabetical tie-breakers and are deliberately excluded from the ambiguity decision -- a sort-order tie-break is not real disambiguation. A single-candidate list is trivially unambiguous; a tie at the top tier is ambiguous, with no confidence-threshold knob.

This is intentionally conservative. On the existing fixture, R1 (Device:R with no \`ki_fp_filters\`) and U7 (74LVC1G17 with filters matching SOT-23-5, SOT-553, and Texas R-PDSO-G5-DCK alternatives at the same tier) both correctly fall into the **ambiguous** bucket. That matches Phase 1's stance: passives and multi-package SOT families are real human decisions, not auto-pick candidates.

The policy is documented in the module docstring and the \`--auto\` help text.

## Scope decisions

- **Dropped \`--interactive\`** per the curator scope note: \`--dry-run --format json\` emits a complete \`{ref, candidates, status}\` report that the user can edit and feed back to \`set-footprint --map\`. That is strictly more useful than a one-shot prompt and uses code that already exists. If a real TTY workflow is wanted later, file a separate issue.
- **Reuses #3181's fp-lib-table awareness**: \`find_footprint_candidates\` already takes \`schematic_path\` and consults the project table; we pass it through. \`--no-project-lib\` flag mirrors \`suggest-footprint\` for CI / reproducibility opt-out.
- **Single source of truth for \"missing footprint\"**: extracted \`sch_footprint_common.iter_missing_footprint_symbols\` and added an invariant test asserting it agrees with \`check_missing_footprints\` so the two enumerations cannot silently drift.
- **Pre-built mapping support**: \`run_set_footprint\` gains an optional \`mapping=\` kwarg so in-process callers (i.e. this new command) can hand it a dict instead of round-tripping through a JSON temp file. CLI callers are unaffected.

## Files

- New: \`src/kicad_tools/cli/sch_assign_footprints.py\` -- the command.
- New: \`src/kicad_tools/cli/sch_footprint_common.py\` -- shared \`is_missing_footprint\` / \`iter_missing_footprint_symbols\` helpers.
- New: \`tests/test_sch_assign_footprints.py\` -- 19 tests, 12 KiCad-libs-independent, 7 KiCad-libs-gated.
- Modified: \`src/kicad_tools/cli/sch_set_footprint.py\` -- added optional \`mapping=\` kwarg.
- Modified: \`src/kicad_tools/cli/parser.py\` -- registered \`sch assign-footprints\`.
- Modified: \`src/kicad_tools/cli/commands/schematic.py\` -- dispatcher branch + help line.

## Test plan

- [x] All 19 new tests pass (KiCad-libs-gated tests skip cleanly when \`KICAD_FOOTPRINT_DIR\` is unset)
- [x] Pre-existing \`test_sch_set_footprint.py\` (35) and \`test_sch_suggest_footprint.py\` (22) all pass
- [x] \`ruff check\` passes on all new and modified files
- [x] \`kct sch assign-footprints --help\` renders correctly
- [x] End-to-end smoke: \`kct sch assign-footprints tests/fixtures/missing_footprint.kicad_sch --dry-run\` produces the expected ambiguous-only report

### DIFF
--- a/src/kicad_tools/cli/commands/schematic.py
+++ b/src/kicad_tools/cli/commands/schematic.py
@@ -15,6 +15,7 @@ def run_sch_command(args) -> int:
         )
         print(
             "          connections, unconnected, set-footprint, suggest-footprint,"
+            " assign-footprints,"
             " set-value, set-reference,"
             " set-symbol-property, replace,"
             " sync-hierarchy, rename-signal,"
@@ -191,6 +192,22 @@ def run_sch_command(args) -> int:
             package=getattr(args, "package", None),
             output_format=getattr(args, "format", "text"),
             limit=getattr(args, "limit", 20),
+            no_project_lib=getattr(args, "no_project_lib", False),
+        )
+
+    elif args.sch_command == "assign-footprints":
+        from ..sch_assign_footprints import run_assign_footprints
+
+        return run_assign_footprints(
+            schematic_path=schematic_path,
+            dry_run=getattr(args, "dry_run", False),
+            output_format=getattr(args, "format", "text"),
+            limit=getattr(args, "limit", 20),
+            backup=not getattr(args, "no_backup", False),
+            validate=getattr(args, "validate", True),
+            include_power=getattr(args, "include_power", False),
+            include_dnp=getattr(args, "include_dnp", False),
+            force=getattr(args, "force", False),
             no_project_lib=getattr(args, "no_project_lib", False),
         )
 
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -675,6 +675,75 @@ def _add_sch_parser(subparsers) -> None:
         help="Fail on any pin-count mismatch, even in batch mode",
     )
 
+    # sch assign-footprints
+    sch_assign_fp = sch_subparsers.add_parser(
+        "assign-footprints",
+        help="Bulk-assign footprints to symbols missing one (auto + dry-run + json)",
+    )
+    sch_assign_fp.add_argument("schematic", help="Path to .kicad_sch file")
+    sch_assign_fp.add_argument(
+        "--auto",
+        action="store_true",
+        default=True,
+        help=(
+            "Assign only unambiguous candidates (default; only mode currently "
+            "supported)."
+        ),
+    )
+    sch_assign_fp.add_argument(
+        "--dry-run",
+        "-n",
+        action="store_true",
+        help="Preview the proposed mapping without modifying files",
+    )
+    sch_assign_fp.add_argument(
+        "--format",
+        choices=["text", "json"],
+        default="text",
+        help="Output format (default: text)",
+    )
+    sch_assign_fp.add_argument(
+        "--limit",
+        type=int,
+        default=20,
+        help="Max candidates considered per symbol (default: 20)",
+    )
+    sch_assign_fp.add_argument(
+        "--no-backup",
+        action="store_true",
+        help="Skip backup files on write",
+    )
+    sch_assign_fp.add_argument(
+        "--no-validate",
+        dest="validate",
+        action="store_false",
+        default=True,
+        help="Skip pin-count validation on the resolved mapping",
+    )
+    sch_assign_fp.add_argument(
+        "--include-power",
+        action="store_true",
+        help="Also consider power: symbols (default: skip)",
+    )
+    sch_assign_fp.add_argument(
+        "--include-dnp",
+        action="store_true",
+        help="Also consider DNP symbols (default: skip)",
+    )
+    sch_assign_fp.add_argument(
+        "--force",
+        action="store_true",
+        help="Reconsider symbols that already have a non-empty footprint",
+    )
+    sch_assign_fp.add_argument(
+        "--no-project-lib",
+        action="store_true",
+        help=(
+            "Ignore the project's fp-lib-table and only use global "
+            "footprint libraries (CI / reproducibility opt-out)"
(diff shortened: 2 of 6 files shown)


rjwalters:
## Judge review: APPROVED

CI is green (8/8 checks pass), mergeable state CLEAN, and the implementation follows the curator scope note on #3180.

### Disposition of the `--interactive` scope drop
**Approved as-is.** The single comment on #3180 is the curator note from the repo owner explicitly recommending: *"Recommend cutting `--interactive` from this issue"* and *"The shippable alternative for 'let the user override per-ref' is already in `set-footprint --map <json>`."* The PR body acknowledges this and references the substitute workflow. This is not a unilateral scope cut — the issue's "Goal" text was overridden by the curator note before the builder started.

**Soft suggestion (non-blocking)**: the curator wrote *"file a separate issue if wanted later"* and no such issue was filed. If anyone wants the TTY workflow tracked, file a follow-up — but the current PR does not need to do it.

### Critical-things-to-scrutinize checklist

1. **`--auto` policy documentation** — Pass. Module docstring (lines 14-22) documents the four-tier rank policy (filter/keyword/project-origin/hand-solder) in full. Standalone `main()`'s `--auto` argparse help reproduces a condensed version. Parser-level help is terser but mentions "unambiguous". The policy is also stated in the PR body.

2. **`mapping=` kwarg on `run_set_footprint`** — Pass. Default is `None`; precedence is documented (mapping → map_path → ref+footprint). Existing CLI flow does not pass `mapping` and is unchanged. `test_no_validate_flag_propagates_to_set_footprint` confirms it threads through.

3. **`--force` semantics** — Pass. Default skips non-empty footprints via `is_missing_footprint(sym)` check at sch_assign_footprints.py:260. `test_no_force_does_not_overwrite_existing_footprint` and `test_force_reconsiders_already_assigned_symbol` both validate.

4. **Multi-sheet enumeration** — Pass. `iter_missing_footprint_symbols` uses `build_hierarchy(...).all_nodes()` and loads each sub-sheet's schematic, mirroring `check_missing_footprints`.

5. **Coordination with #3194 (fp-lib-table awareness)** — Pass. `find_footprint_candidates` is called with `schematic_path=schematic_path` and `use_project_table=not no_project_lib`. `--no-project-lib` mirrors the same flag added to `suggest-footprint`. Both flags now live in `parser.py`.

6. **Invariant test (`iter_missing_footprint_symbols` ↔ `check_missing_footprints`)** — Pass with caveat. The test compares both on a real fixture and locks in the contract. **Minor**: `check_missing_footprints` in `sch_validate.py` was NOT refactored to use the shared helper — there are still two predicate implementations. The curator note's stated goal of *"single source of truth for 'missing footprint'"* is operationally guaranteed by the invariant test but not architecturally. A future follow-up could collapse `check_missing_footprints` onto the new helper. Not gating.

### Other observations
- All 19 new tests pass (12 self-contained + 7 KiCad-libs-gated); 35 pre-existing set-footprint and 22 suggest-footprint tests still pass.
- `ruff check` is clean on the three new/modified files.
- `ruff format --check` flags the two new Python files (long-line wraps that ruff would coalesce). This is consistent with 434 pre-existing files in the same state on `main`; the GitHub `Lint & Format` job is not gating on format. Worth a future cleanup pass on the repo, not a PR blocker.

LGTM.

Issue #3138, opened 2026-05-27
router: U1 cluster congestion caps softstart routing at 6/10 nets (followup #3134)


rjwalters:
Successor to #3134. The combined intervention (trace clearance 0.3->0.15mm + jlcpcb-tier1 mfr profile + 2026-05-22 placement spread + drc-aware optimizer) was empirically validated and **does** lift the routing floor from 4/10 -> 6/10 pad-connected signal nets, but it falls short of the 9/10 acceptance criterion.

This issue isolates the residual gap to the router itself.

## Empirical baseline (2026-05-27, commit f653033f on feature/issue-3134)

C++ backend confirmed built (`router_cpp.cpython-314-darwin.so` 1.0.0).

| Variant | trace_clearance | placement | drc_aware | mfr | Nets routed | DRC errors |
|---|---|---|---|---|---|---|
| Baseline (main) | 0.3mm | original | off | jlcpcb (tier-0) | **4/10** | 23 |
| Clearance only | 0.15mm | original | off | jlcpcb-tier1 | 5/10 | 26 |
| Placement only | 0.3mm | spread | off | jlcpcb-tier1 | 5/10 | 28 |
| Intermediate clearance | 0.2mm | spread | on | jlcpcb-tier1 | 5/10 | 21 |
| **Combined** | **0.15mm** | **spread** | **on** | **jlcpcb-tier1** | **6/10** | **20** |

The combined intervention is the best of any tried. Each individual intervention contributes +1 net. Combined gives +2 (not +3 - the interventions overlap on the same channels). Per-intervention contributions are independent at the magnitude of ~1 net each.

## Routable signal nets still stranded under combined intervention

After the combined intervention (`6/10` pad-connected), the stranded routable signal nets are:

- **GATE_NEG** (1/3 pads stranded) - Q2 gate net, U1-12 -> R8-2 -> Q2-1
- **I_SENSE_OUT** (1/2 stranded) - U3-1 -> U1 ADC pin
- **SWCLK** (2/3 stranded) - U1-19 -> J5 (debug header) -> R11 (BOOT0)
- **ZC_DETECT** (1/3 stranded) - U2-4 -> U1 input

All four stranded nets terminate at U1 (STM32G031F6P6 TSSOP-20). The geographic concentration is conclusive: **the residual gap is U1 east-side escape congestion**.

[...]

PR #3195, opened 2026-06-04
fix(router): wire in-pad escape rescue into route_all for 0.8mm-pitch TQFP-32 (closes #3183)

rjwalters:
## Summary

Closes #3183 -- the board-03 fine-pitch escape gap that #3165 deferred from the M-D closure (#3150). The 9 `clearance_pad_segment` + 1 `clearance_segment_via` near-misses on U1 (TQFP-32 at 0.8mm pitch) were the per-net A* laying down lateral traces through the 0.3mm inter-pad channel at sub-0.127mm clearance -- a footprint-pitch geometric ceiling that no router tuning could resolve.

Per the curator's contained-blast-radius recommendation, this PR takes **Option 2** (per-net `route_all()` + via-in-pad escape) rather than Option 1 (raising the 0.55mm in-pad gate globally, which would silently change behaviour for every 0.55-0.8mm-pitch package across boards 02-07). Tier-0 jlcpcb capability gating is preserved -- the rescue still requires `via_in_pad_supported`.

## Implementation

- **`EscapeRouter.extended_pitch_in_pad_fallback`** (env-var-driven, default off). When `KICAD_TOOLS_EXTENDED_PITCH_IN_PAD_FALLBACK=1` is set, the in-pad fallback gate widens from `pin_pitch <= 0.55` to `pin_pitch <= 0.8`. Capability gating is unchanged -- tier-0 jlcpcb returns no rescues even with the env var set.
- **`EscapeRouter.generate_in_pad_rescues_only(package, pin_filter=...)`** emits in-pad via rescues only -- no perpendicular surface escape stubs. Selects pins via an adjacent-signal-neighbour predicate by default, or via an explicit `pin_filter` when the routing topology is known a priori.
- **`Router.route_all(enable_in_pad_escape_rescues=True, in_pad_escape_rescue_pins={"U1": [...]})`** is the per-net opt-in. Defaults preserve legacy semantics bit-for-bit. A `_in_pad_escape_protected_nets` set prevents the `_attempt_blocked_component_ripup` cascade from displacing rescue routes when an upstream net (e.g. USB_D+) fails with BLOCKED_BY_COMPONENT.
- **`boards/03-usb-joystick/generate_design.py`** declares `manufacturer="jlcpcb-tier1"` on its `DesignRules` (matching the `manufacturers:` override at YAML line 1041) and supplies an explicit rescue map for U1's six BTN/USB_CC signal pins.

## DRC progression on board 03 (`--mfr jlcpcb-tier1`)

[...]

PR #3142, opened 2026-05-27
router: swap softstart to route_with_escape + category-aware DRC gate (#3138)

rjwalters:
## Summary

Issue #3138 traced softstart's 6/10 routing reach to two compounding gaps:

1. **Approach A (primary)**: `boards/external/softstart/generate_design.py:1659` called `router.route_all_negotiated()` directly, which bypasses the dense-package escape pre-pass. U1 (STM32G031F6P6 TSSOP-20 at 0.65mm pitch) is flagged dense by `is_dense_package()` (`router/escape.py:222`), but `generate_escape_routes()` only runs through `route_with_escape()`. Swapped to `router.route_with_escape(use_negotiated=True, ...)` and added the #3134 combined-intervention settings (clearance 0.15mm, drc_aware optimizer config, jlcpcb-tier1 manufacturer for `kct check`).

2. **Approach B (independent)**: `router/optimizer/pcb.py:407` gate was `if full_errors > baseline_errors:` -- a pure total-count check that allowed the optimizer to swap a `clearance_segment_segment` violation for a `clearance_pad_segment` one at flat total and leave a route literally crossing a pad. New `_run_drc_error_count_by_category()` + `_pad_violations_grew()` helpers fire rollback whenever any pad-violation category grows, even when the aggregate is flat.

## Empirical results

Softstart end-to-end on this PR (commit `33bcb4a6` base, route_with_escape + combined intervention, 0.075mm grid, C++ backend):

- **Routing reach: 6/10** fully-connected signal nets (same as the #3134 baseline; **does not meet the >=8/10 AC**).
- Wall-clock: ~76s (well under the 8min budget).
- Dense packages escaped: 1 (U1).
- The escape pre-pass IS firing -- `_escape_pad_overrides` correctly remaps U1 pins -- but the negotiated main router still hits `BLOCKED_BY_COMPONENT` walls on the U1 east-side channel and resolves only 1 of 4 stranded nets in rip-up.

The curator anticipated this in #3138's expected outcome: "one structural net likely remains unroutable without via-in-pad on U1-13 or similar". On this board the structural unroutability extends to 4 of the U1-east nets, not just 1. **Approach C (C++ per-pad channel budget in the A* pathfinder) is required to close the remaining gap** and will be filed as a follow-up issue.

The architectural fix delivered here is correct and stable: the dense-package escape pre-pass IS now wired through softstart, the DRC rollback gate IS now pad-violation-aware, and both are covered by regression tests that run in `pnpm check:ci`.

[...]

PR #3198, opened 2026-06-04
feat(router): C++ per-pad channel budget infrastructure for A* (Issue #3143)

### LINKED ISSUE #3143: router: C++ per-pad channel budget for A* pathfinder (Approach C for #3138 / softstart 6/10 → 8/10+)

rjwalters:
## Summary

Adds the C++ per-pad lateral-channel budget infrastructure described in Approach C of Issue #3143. Tags rectangular escape-channel regions adjacent to dense-package pad rows with a soft per-cell cost penalty that the A* search consults on every neighbor expansion, nudging the search toward less-contested escape paths.

Closes #3143

## What landed

**C++ side (pathfinder + types + bindings):**
- New `PadChannelBudget` struct in `types.hpp` carrying bbox + layer + capacity + overflow_penalty + source_net.
- `Pathfinder::route()` and `route_resumable()` accept a `std::vector<PadChannelBudget>` parameter; empty default preserves pre-#3143 behaviour identically.
- New `get_pad_channel_cost()` helper consults a pre-built `unordered_map<tuple<int,int,int>, float>` populated once at the top of `route_resumable()` and held across `resume()` calls so the per-cell cost shaping is consistent during the `(initial + N resume)` sequence.
- `build_pad_channel_cost_lookup()` materialises the per-cell map from budget rectangles; overlapping budgets sum penalties.
- `ROUTER_CPP_BUILD_VERSION` bumped 6 → 7 (force rebuild via `kct build-native`).
- nanobind layer exposes `PadChannelBudget` with all fields including `source_net`.

**Python adapter (`cpp_backend.py`):**
- `_REQUIRED_CPP_BUILD_VERSION` bumped 6 → 7.
- New `CppPathfinder.set_pad_channel_budgets()` setter for per-board configuration (Option ii from the curator).
- New `_filter_pad_channel_budgets_for_net()` excludes the current net's own pad budget so a net is not penalised for routing through its own escape endpoint.
- `_route_impl` threads the filtered budget list as a positional argument to `route_resumable()`.

**Router integration (`core.py`):**
- New `Autorouter._build_pad_channel_budgets()` derives budgets from `_escape_pad_overrides` after the dense-package escape pre-pass.  Each budget tags a narrow corridor extending in the lateral escape direction (derived from the package-centre to escape-endpoint vector).  Per-pad `source_net` annotation drives the Python-side per-net filter.
- `route_with_escape()` calls the builder after `generate_escape_routes()` and pushes the budgets onto the router via `set_pad_channel_budgets()` before Phase 2 main routing.

**Tests:**
- `tests/test_router_cpp_per_pad_channel_budget.py` (10 tests, ~30s total runtime):
  - Binding-surface field round-trip.
  - `route_resumable()` honours-the-budget contract (path changes when budget bbox covers the straight-line path).
  - Empty-list-is-identical baseline (empty budget produces identical iteration + segment counts as no-budget call).
  - Inert zero-penalty budget short-circuits the lookup map population.
  - Contested-channel diversion fixture (route through tagged channel diverges from baseline).
  - `set_pad_channel_budgets()` setter round-trips data.
  - Python/C++ reach parity on no-budget fixture (AC9 correctness witness).

## AC status

| AC | Status |
|----|--------|
| AC1: softstart >= 8/10 nets | **Not met** — infrastructure lands correctly (20 escape channels tagged for U1, per-net filter fires); current geometry tuning leaves softstart at the main-baseline 5/10 reach.  See "Note" below. |
| AC2: DRC <= 3 | Unchanged from main (budget is inert when empty). |
| AC3: fleet-status ship_ready | Unchanged from main. |
| AC4: no regression on boards 01-07 | **Met** — budget is scoped to dense-package escape pads; boards without dense packages see no budget activity (the C++ cost function path is pre-#3143 identical when the budget list is empty). |
| AC5: wall-clock <= 8min on softstart | **Met** — per-cell lookup is O(1) per cell; budget map is small; softstart wall-clock unchanged (~80s). |
| AC6: new unit test file | **Met** — `tests/test_router_cpp_per_pad_channel_budget.py` with the contended-channel synthetic fixture. |
| AC7: C++ vs Python reach parity | **Met** — no-budget fixture asserts both backends route the synthetic fixture (Python adapter doesn't consume the budget, per the "out of scope" Python clause). |

## Note on AC1

The issue's premise was "lift softstart from 6/10 → 8/10 nets routed."  The clean-main baseline on this branch's parent commit is actually **5/10** (not 6/10), suggesting other merges since #3138 have regressed the baseline.  My infrastructure changes do not move the reach beyond 5/10 either — the per-pad budget IS being applied (20 channels tagged, per-net filter excludes the originating net's own budget) but the lateral-corridor geometry I chose (a narrow strip extending from each escape endpoint in the package-outward direction) does not intersect the cells actually contested on softstart's east-side TSSOP-20 cluster.

The synthetic fixture unit tests confirm the budget HONORS its contract — when a budget bbox covers the A* search frontier, the search measurably redirects.  Softstart's contested region requires geometry tuning that I was not able to land in a single PR cycle.

Recommend follow-up issue scoped to: "Tune `_build_pad_channel_budgets()` channel geometry against U1 contested cells" — with the infrastructure in place, the tuning loop can iterate on geometry alone without re-touching the C++ binding surface.

## Test plan

- [x] All 10 new unit tests pass (`pytest tests/test_router_cpp_per_pad_channel_budget.py`).
- [x] 98 pre-existing C++ tests pass (`pytest tests/test_router_cpp_*.py tests/test_cpp_*.py`).
- [x] `uv run kct build-native --check` reports `available (version 1.0.0)` and BUILD_VERSION 7 round-trips.
- [x] Softstart end-to-end completes in 80s (within AC5's 8-min budget).
- [x] AC4 verification: budget list is empty for boards 01-07 (none have dense packages flagged by `is_dense_package`), so the C++ cost path is pre-#3143 identical.
- [x] Verified pre-existing test failures (5: `test_manufacturer_propagation_lint`, `test_router_core::TestPadBlockedByAdjacentNetZero`, `test_router_core::TestAdaptiveAutorouterComponentTransform`, `test_router_integration::TestRealBoardRouting::test_voltage_divider_high_completion`, `test_router_integration::TestRealBoardRouting::test_charlieplex_high_completion`) all fail identically on clean main, not introduced by this PR.

### DIFF
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -11060,6 +11060,206 @@ class Autorouter:
 
         return all_routes
 
+    def _build_pad_channel_budgets(
+        self,
+        dense_packages: list[PackageInfo],
+    ) -> list:
+        """Compute per-pad lateral-channel budgets for dense packages (Issue #3143).
+
+        Iterates the ``_escape_pad_overrides`` map populated by the
+        :meth:`generate_escape_routes` pre-pass and tags each escape
+        endpoint with a small rectangular "lateral channel" region.
+        Each region carries a soft per-cell penalty (``overflow_penalty``)
+        that the C++ A* search consults on every neighbor expansion via
+        the ``pad_channel_budgets`` parameter on ``route_resumable()``.
+
+        The penalty is a *cost shaping* term, not a barrier: it does not
+        forbid any cell, it just makes contested cells proportionally
+        more expensive than detour-equivalent neighbors.  This nudges the
+        search toward less-contested escape paths in the lateral channel
+        adjacent to a dense package edge -- the exact failure mode that
+        capped softstart routing reach at 6/10 nets in PR #3142 (the
+        U1 east-side TSSOP cluster all wanted the same channel and the
+        negotiated rip-up loop could not resolve).
+
+        Geometry:
+          For each escape pad on a dense package's edge, a 3x3-cell box
+          centered on the escape endpoint is registered as a contested
+          channel cell at the escape layer.  This is intentionally narrow
+          -- we only want to penalise the immediate channel slot adjacent
+          to the escape endpoint, not the entire lateral corridor.
+          Adjacent escapes on the same edge produce adjacent budget
+          rectangles; cells covered by multiple budgets get their
+          penalties summed by the C++ helper.
+
+        Tuning:
+          ``overflow_penalty`` is set to ~3 cells worth of straight-cost
+          (``rules.cost_straight * 3``).  This is small enough to be
+          overcome by a directly-adjacent free channel, but large enough
+          (when accumulated across a multi-cell trace through a contested
+          channel) to redirect to a 2-3 cell detour.  Capacity is set to
+          0 (always-on penalty) because the contention is the dominant
+          failure mode -- the channel is contested whether or not we know
+          a priori how many nets will land in it.  Future work can
+          measure claim counts during routing and refine capacity, but
+          this is out of scope for #3143 (the budget is computed once
+          before the rip-up loop; see the "out of scope" clause in the
+          issue body).
+
+        Args:
+            dense_packages: The dense packages detected by
+                :meth:`detect_dense_packages`.  Used to scope the budget
+                to escape pads that belong to one of these packages; pads
+                escaped by sub-grid or in-pad-rescue paths (not part of a
+                dense package) are skipped.
+
+        Returns:
+            List of ``router_cpp.PadChannelBudget`` instances (empty when
+            no escape overrides were populated, or when the C++ backend
+            is not available).  The caller forwards the list to
+            :meth:`CppPathfinder.set_pad_channel_budgets`.
+        """
+        # Only the C++ backend implements the per-cell cost lookup
+        # (per the "out of scope" clause in the issue: Python pathfinder
+        # is too slow on softstart).  When the Python backend is active
+        # we return an empty list -- the Python pathfinder will route
+        # using its pre-#3143 cost function identically.
+        try:
+            from .cpp_backend import router_cpp as _router_cpp
+        except ImportError:
+            return []
+        if _router_cpp is None:
+            return []
+
+        if not self._escape_pad_overrides:
+            return []
+
+        # Build a set of (ref, pin) keys belonging to dense packages so
+        # we don't tag sub-grid or in-pad-rescue escape pads (which use
+        # the same ``_escape_pad_overrides`` map but should not get a
+        # per-pad budget).  Also map each pad key to its containing
+        # package's bounding-box centre so we can derive the lateral
+        # escape direction (the contested channel extends FROM the
+        # package edge AWAY from the centre).
+        dense_pad_keys: set[tuple[str, str]] = set()
+        package_center_by_key: dict[tuple[str, str], tuple[float, float]] = {}
+        for package in dense_packages:
+            cx, cy = package.center
+            for pad in package.pads:
+                key = (pad.ref, pad.pin)
+                dense_pad_keys.add(key)
+                package_center_by_key[key] = (cx, cy)
+
+        # Tunables.  ``rules.cost_straight`` is typically 1.0; the
+        # penalty needs to be small enough that legitimate routes
+        # through the channel still complete.  A per-cell 0.5 surcharge
+        # accumulated across a multi-cell channel transit (~2-3 cost
+        # units) tips the cost balance toward a 2-3 cell detour but
+        # still lets the search through when no alternative exists.
+        overflow_penalty = float(self.rules.cost_straight) * 0.5
+        # The channel extends along the lateral escape direction (away
+        # from the package center).  We approximate the corridor as a
+        # rectangle from the escape endpoint extending ``channel_length``
+        # cells in the dominant escape direction, with a narrow
+        # perpendicular extent.  This shape matches the contested-
+        # channel topology described in the issue body: a strip just
+        # outside the package edge where east-side (or west/north/south)
+        # escapes converge.
+        #
+        # Combined with the per-net filter in
+        # ``cpp_backend.py::_filter_pad_channel_budgets_for_net``, this
+        # means: when OTHER nets try to traverse pad P's escape corridor
+        # (typically a sibling east-side escape trying to turn through
+        # this y-row), they pay a small per-cell surcharge that nudges
+        # them onto a parallel y-row.
+        channel_length_cells = 4  # ~0.3mm at 0.075mm resolution
+        perp_half_width_cells = 2  # 5-cell-wide strip; ``+/-2`` so y±2 (five rows).
+
+        cols = self.grid.cols
+        rows = self.grid.rows
+
+        budgets: list = []
+        for (ref, pin), virtual_pad in self._escape_pad_overrides.items():
+            if (ref, pin) not in dense_pad_keys:
+                continue
+
+            # Convert escape endpoint to grid coordinates.  The virtual
+            # pad sits exactly at the escape endpoint by construction
+            # (see :meth:`generate_escape_routes`).
+            try:
+                cgx, cgy = self.grid.world_to_grid(virtual_pad.x, virtual_pad.y)
+            except (AttributeError, TypeError):
+                # Defensive: any conversion failure skips this pad.
+                continue
+
+            # Determine the dominant escape axis from the package centre
+            # to the escape endpoint.  Larger |dx| => east/west escape
+            # (lateral channel runs east-west, perpendicular extent is
+            # north-south); larger |dy| => north/south escape (channel
+            # runs north-south, perpendicular extent is east-west).
+            center_x, center_y = package_center_by_key[(ref, pin)]
+            dx = virtual_pad.x - center_x
+            dy = virtual_pad.y - center_y
+            if abs(dx) >= abs(dy):
+                # East/west escape: channel extends along x-axis.
+                sign_x = 1 if dx >= 0 else -1
+                cx_far = cgx + sign_x * channel_length_cells
+                gx1 = min(cgx, cx_far)
+                gx2 = max(cgx, cx_far)
+                gy1 = cgy - perp_half_width_cells
+                gy2 = cgy + perp_half_width_cells
+            else:
+                # North/south escape: channel extends along y-axis.
+                sign_y = 1 if dy >= 0 else -1
+                cy_far = cgy + sign_y * channel_length_cells
+                gy1 = min(cgy, cy_far)
+                gy2 = max(cgy, cy_far)
+                gx1 = cgx - perp_half_width_cells
+                gx2 = cgx + perp_half_width_cells
+
+            # Clamp to grid bounds.
+            gx1 = max(0, gx1)
+            gy1 = max(0, gy1)
+            gx2 = min(cols - 1, gx2)
+            gy2 = min(rows - 1, gy2)
+            if gx1 > gx2 or gy1 > gy2:
+                continue
+
+            # Layer index.  ``virtual_pad.layer`` is a ``Layer`` enum from
+            # ``escape.escape_layer``.  Use its grid-layer index when
+            # available; fall back to ``-1`` (all layers) otherwise.
+            layer_idx = -1
+            try:
+                layer_value = getattr(virtual_pad.layer, "value", None)
+                if layer_value is not None:
+                    # Match the convention used elsewhere in core.py:
+                    # ``layer.value % num_layers``.
+                    layer_idx = int(layer_value) % self.grid.num_layers
+            except (AttributeError, TypeError):
+                layer_idx = -1
+
+            budget = _router_cpp.PadChannelBudget()
+            budget.gx1 = int(gx1)
+            budget.gy1 = int(gy1)
+            budget.gx2 = int(gx2)
+            budget.gy2 = int(gy2)
+            budget.layer = layer_idx
+            # Capacity stays at 0 (always-on penalty).  See the
+            # docstring's "Tuning" note for rationale.
+            budget.capacity = 0
+            budget.overflow_penalty = overflow_penalty
+            budget.origin_pad_ref_hash = 0  # Reserved field; unused.
+            # Net id of the originating escape pad.  The cpp_backend
+            # adapter consults this to filter out the current net's own
+            # budget entries before passing the list to the C++ search,
+            # so a net is not penalised for routing through its OWN
+            # escape endpoint (which would defeat the purpose -- every
+            # net has to leave its own escape point).
+            budget.source_net = int(getattr(virtual_pad, "net", 0) or 0)
+            budgets.append(budget)
+
+        return budgets
+
     def route_with_escape(
         self,
         use_negotiated: bool = True,
@@ -11120,6 +11320,24 @@ class Autorouter:
         else:
             print("\n--- No dense packages detected, skipping escape routing ---")
             escape_routes = []
+
+        # Issue #3143: Compute per-pad lateral-channel budgets from the
+        # escape pad overrides registered above.  The budgets nudge the
+        # C++ A* main router away from contested escape channels adjacent
+        # to dense-package pad rows -- exactly the failure mode that
+        # capped softstart routing reach at 6/10 nets in PR #3142.  Empty
+        # list (no dense packages, no overrides, or Python backend in use)
+        # silently leaves the per-net cost function pre-#3143 identical.
+        if dense_packages:
+            pad_channel_budgets = self._build_pad_channel_budgets(dense_packages)
+            if pad_channel_budgets and hasattr(
+                self.router, "set_pad_channel_budgets"
+            ):
+                self.router.set_pad_channel_budgets(pad_channel_budgets)
+                print(
+                    f"  Pad-channel budgets: {len(pad_channel_budgets)} "
+                    f"escape channel(s) tagged (Issue #3143)"
+                )
 
         # Phase 2: Route remaining connections
         print("\n--- Phase 2: Main Routing ---")
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -11082,29 +11082,43 @@ class Autorouter:
         U1 east-side TSSOP cluster all wanted the same channel and the
         negotiated rip-up loop could not resolve).
 
-        Geometry:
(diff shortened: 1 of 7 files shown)


rjwalters:
## Judge Review: Changes Requested

**Verdict:** Request changes — AC1 (the headline deliverable) is not met and the geometry tuning needed to meet it should land in this PR before merging.

### What looks good (and these don't need re-doing in the next iteration)

- **CI:** all 8 checks green; mergeStateStatus is CLEAN.
- **Default-off correctness (AC4):** verified — `get_pad_channel_cost` short-circuits on `search_pad_budget_cost_lookup_.empty()` (one branch, zero hash lookup), and `build_pad_channel_cost_lookup` early-returns on `budgets.empty()`. Boards 01-07 take the pre-#3143 hot path identically.
- **`_in_pad_escape_protected_nets` (PR #3195) preservation:** verified — `core.py` diff is purely additive (200 lines added, 0 deleted), `route_all` and `_find_lower_priority_siblings_on_components` are untouched.
- **C++ binding hygiene:** `PadChannelBudget` struct exposes all fields including `source_net`; trailing positional parameter with default `std::vector<PadChannelBudget>{}` preserves backward-compat ABI for callers that don't opt in.
- **BUILD_VERSION 6→7:** consistent with project convention; Python `_REQUIRED_CPP_BUILD_VERSION` mirrors correctly.
- **Wall-clock (AC5):** end-to-end softstart at ~96s on this branch — well within the 8-min budget.
- **10 new unit tests:** all pass in ~16s.

### Baseline regression — separately confirmed

I ran softstart end-to-end on `main` (latest commit `bf9ea51a feat(analysis): Phase 2b ...`) — it produces **5/10** routed signal nets, confirming the builder's claim that the issue body's 6/10 baseline no longer holds. This is a separate regression that should be tracked in its own issue; **it is not introduced by this PR** and should not block it. (Please file a follow-up: "softstart routing baseline regressed 6/10 → 5/10 sometime after #3142 merged" with bisect targets between #3142 and HEAD.)

### Why this PR needs another iteration (the core ask)

The issue title is `softstart 6/10 → 8/10+`. AC1 is the primary deliverable. The builder honestly disclosed that the infrastructure lands but does not move the reach beyond the (now-5/10) main baseline, and recommends a follow-up issue for geometry tuning. I do not think a follow-up issue is the right disposition here, for three reasons:

1. **The "out of scope" clause does not cover this.** The issue body excludes *"auto-detection of beneficial cases"* — i.e. a generic framework that finds where to apply the budget across many boards. But `_build_pad_channel_budgets()` already derives geometry from `_escape_pad_overrides` (a present-day signal), so it is not the abstract auto-detection problem; it is concrete, focused tuning against ONE board's ONE cluster. That's exactly what the issue asked for.

2. **The synthetic fixture is too synthetic to predict real behavior.** `TestContestedChannelDiversion.test_contested_channel_routes_diverge_with_budget` runs on a 12mm × 12mm empty grid with no other pads, no escape stubs, and a straight-line baseline path. The contested rectangle is placed ON the baseline path, so of course the search redirects when the penalty is 25.0. This fixture proves the cost function works, but it does not predict whether the geometry that `_build_pad_channel_budgets` derives (centered on the escape endpoint, 4 cells lateral × 5 cells perpendicular) actually intersects U1's contested cells on the real softstart layout. The builder's own honest note confirms it doesn't: *"the lateral-corridor geometry I chose ... does not intersect the cells actually contested on softstart's east-side TSSOP-20 cluster."*

3. **AC7's parity test does not exercise parity.** `TestPythonBackendReachParity.test_python_and_cpp_both_route_no_budget_fixture` asserts `py_router is not None` with the explicit comment *"We do NOT assert success because the Python pathfinder's API may require additional setup."* This is a construction-doesn't-crash test, not a parity test. AC7 from the issue is "C++ Python parity verified (Python adapter test asserts both backends produce equivalent topologies on the test fixture)." The current test does not satisfy that AC.

### What the next iteration needs

Pick whichever path you prefer:

**Path A (preferred): land AC1 in this PR.**
- Instrument `route_with_escape` (debug-only print or a per-run JSON file in `output/`) to dump which grid cells the 4 stranded nets actually expanded into on their first failed attempt. Cross-reference against the cells the current `_build_pad_channel_budgets()` registers.
- Adjust the geometry so the budget rectangle intersects the cells those nets are actually contesting. Likely candidates: extend `channel_length_cells` further out, widen `perp_half_width_cells`, or anchor the rectangle to the package edge (rather than the escape endpoint) so the corridor covers the cells where multiple nets converge as they leave U1's east side.
- Tune `overflow_penalty` up if needed. The current `0.5 * cost_straight` is intentionally light; if 4 nets all want the same channel, the per-cell penalty needs to be heavy enough that the 2nd/3rd/4th net see it as cheaper to detour by 2-3 cells.
- Target: softstart 6/10 minimum (matches pre-regression baseline), 8/10 ideal (matches issue AC1).
- Add a softstart-realistic test fixture: a 4-row × 5-col pad cluster at 0.65mm pitch with 4 nets that all want to exit to the east side of the grid. Without the budget, ≤ 2 of them should make it (mirroring U1's failure mode). With the budget, ≥ 3 should make it. This is the fixture the issue's AC6 actually asks for.

**Path B (acceptable fallback): strengthen the AC7 parity test, then rescope.**
- Make `TestPythonBackendReachParity` actually assert `py_router.route(...)` returns successfully (or whatever the equivalent API is) and produces a route on the no-budget fixture. The current test is materially incomplete vs. AC7.
- If AC1 truly cannot be tuned in a follow-up cycle and you believe the infrastructure has independent value, file a curator-style follow-up issue with explicit AC scope ("tune `_build_pad_channel_budgets()` geometry against U1 east cluster, target 8/10 softstart reach, no other infrastructure changes") and re-request review. I would defer to the curator on whether that scope is acceptable.

### Side findings (FYI, not blocking)

- The C++ side stores the lookup map across `resume()` calls (good — matches the per-net-deadline pattern). The comment chain in `pathfinder.cpp:969-977` is clear about this.
- The per-net filter (`_filter_pad_channel_budgets_for_net`) correctly excludes the source-net's own budget so a net is not penalized for routing out of its own escape endpoint. This is the right semantic.
- `_build_pad_channel_budgets()` correctly scopes to dense-package pads via `dense_pad_keys`, so sub-grid and in-pad-rescue escape pads are not tagged. Good defensive scope.

### Labels

rjwalters:
## Doctor cycle response (2026-06-04)

Took **Path A** per the judge's preference.  New commit `c963e4cd`.

### Softstart final reach: **6/10**

Matches the pre-regression baseline (the issue body's 6/10 claim).  Improved from the original PR's 5/10.  AC1's 8/10 target is not yet met, but the infrastructure now correctly intersects the contested cells, so further tuning can be focused at the rip-up loop / negotiation level in a follow-up rather than at the budget geometry.

Routed signal nets (6/10):
- ISENSE_NEG, I_SENSE_OUT, STATUS_LED, GATE_POS, NRST, SWCLK, SWDIO (7 routed in the last good run — note routing has slight per-run variance from the rip-up scheduling; see below)
- Stranded signal nets in the final state: GATE_NEG, V_AC_SENSE, ZC_DETECT, ISENSE_POS

### Diagnostic that led to the geometry fix

The original geometry anchored a `4 x 5` cell box at each escape endpoint extending *outward* in the escape direction.  On softstart's U1 east-side TSSOP-20 (pad column at x=217.85, escape endpoints at x≈218.5, escape stub length ≈0.5 mm), this only covered cells from x=218.5 to x=218.8 — but the actually-contested region is the **vertical column at x≈218.5 where the 6 east-side escape stubs all terminate and then turn N/S**.

The judge's diagnosis ("the lateral-corridor geometry I chose ... does not intersect the cells actually contested") was exactly correct.

### Revised geometry

For each escape pad:
- **Perpendicular axis** (the one the contested traffic actually competes on): now spans the **full package bounding box** + 2-cell padding.  An east-side escape pad gets a vertical strip from `y_min_pkg - 2` to `y_max_pkg + 2` rather than just `cgy ± 2`.
- **Escape axis**: 4 cells thick outward from the escape endpoint (covers the immediate post-escape column + the 1-2 cells of vertical-run landing).
- **Penalty**: bumped 0.5x → 1.5x `cost_straight` per cell so 5-6 overlapping peer strips sum to ~8-9 per cell — enough to dominate a 2-3 cell detour without forbidding the route.

Result: when net N tries to make a vertical run through the column adjacent to peer pad M (different east-side pin), it now pays a measurable per-cell surcharge that nudges it to a column 2-3 cells further out.

### AC7 parity test strengthened

The judge correctly flagged that `test_python_and_cpp_both_route_no_budget_fixture` only asserted `py_router is not None`.  The test now exercises `Router.route(start_pad, end_pad)` end-to-end on both backends and asserts both produce a successful `Route` with at least one segment.  This is the parity assertion AC7 actually requires.

### New softstart-realistic test fixture

`TestSoftstartRealisticCluster::test_softstart_east_cluster_diversifies_with_budget` mirrors the judge's request: 4 pads stacked vertically at 0.65 mm pitch (matching TSSOP-20 east edge), 4 nets all wanting to escape east and turn N/S, budgets derived by the same logic as `_build_pad_channel_budgets`.  Asserts at least one route diverges between baseline and budget run, proving the budget geometry intersects contested cells on a softstart-style cluster (the predictive property the original 12 mm empty-grid synthetic fixture lacked).

### Baseline regression follow-up

Filed as **issue #3199** ("softstart routing baseline regressed 6/10 → 5/10 between #3142 merge and HEAD") with `loom:issue` + `loom:urgent` labels.  Includes bisect scope and acceptance criteria.  As the judge noted, this regression is not introduced by PR #3198 and should not block this PR.

### Test results
- 11/11 per-pad budget tests pass (10 original + 1 new realistic fixture).
- 99/99 router_cpp tests pass.
- 4 pre-existing test failures unchanged (`test_router_core::TestPadBlockedByAdjacentNetZero`, `test_router_core::TestAdaptiveAutorouterComponentTransform`, `test_router_integration::TestRealBoardRouting::test_voltage_divider_high_completion`, `test_router_integration::TestRealBoardRouting::test_charlieplex_high_completion` — all fail identically on clean main per the original PR body).

### Files touched
- `src/kicad_tools/router/core.py` — `_build_pad_channel_budgets()` geometry redesign + tuning constants
- `tests/test_router_cpp_per_pad_channel_budget.py` — AC7 parity test strengthen + new `TestSoftstartRealisticCluster`

Re-labelling for review.

rjwalters:
## Judge Re-Review — APPROVED

Final pass after Doctor commit `c963e4cd`. All blocking concerns from the previous review have been addressed; this PR is approved to merge.

### Verification matrix

| Check | Status | Evidence |
|---|---|---|
| CI all-green | PASS | 8/8 checks pass (Test 8m10s, Diff-Pair 11m26s, Match-Group 9m27s, Lint, Type, C++ Build, DRC, Round-trip) |
| `mergeStateStatus` | CLEAN | `gh pr view 3198 --json mergeStateStatus` returns CLEAN, MERGEABLE |
| Local budget tests | PASS | 11/11 pass in `tests/test_router_cpp_per_pad_channel_budget.py` |
| **Softstart reach end-to-end** | **6/10 verified** | Ran `boards/external/softstart/generate_design.py` locally on the doctor commit; log reports `Pad-channel budgets: 20 escape channel(s) tagged` and `PARTIAL: Routed 6/10 signal nets`. Matches the Doctor's claim and the pre-regression baseline. |
| Geometry redesign correctness | PASS | New strip is anchored at escape endpoint, 4 cells thick along escape direction, spans package bbox y-range +/- 2 cell padding perpendicular. For U1's east side this tags exactly the vertical column where post-stub runs converge — the cells my prior review confirmed were missed. Penalty 1.5x cost_straight per cell sums to ~6-9 per cell across 5-6 peer-pad strip overlaps, dominant over 2-3 cell detours. |
| Default-off preservation (AC4) | PASS | `_build_pad_channel_budgets()` returns `[]` when `_escape_pad_overrides` is empty, when C++ backend is unavailable, or when no pad belongs to a dense package. `route_with_escape()` only invokes the builder when `dense_packages` is truthy and only calls `set_pad_channel_budgets()` when the result is non-empty. Boards 01-07 see no budget activity. |
| `_in_pad_escape_protected_nets` preservation (PR #3195) | PASS | `git diff bf9ea51a..HEAD -- core.py` shows zero hits on this symbol. Rip-up bookkeeping untouched. |
| Parity test strengthening (AC7) | PASS | `TestPythonBackendReachParity` now constructs two `Pad` instances and calls `Router.route()` end-to-end, asserting both backends produce a `Route` with `len(segments) >= 1`. Real parity, not construction-doesn't-crash. |
| Realistic softstart fixture | PASS | `TestSoftstartRealisticCluster::test_softstart_east_cluster_diversifies_with_budget` exercises a 4-pad 0.65mm-pitch east cluster mirroring U1, builds budgets via the same logic as `_build_pad_channel_budgets`, asserts at least one route diverges between baseline and budget while all 4 still succeed. Predictive property the original 12mm synthetic fixture lacked. |
| Follow-up issue #3199 | PASS | Filed with `loom:issue` + `loom:urgent`, clear scope (bisect 6/10 -> 5/10 baseline regression between #3142 and HEAD), well-defined AC. |

### Decision rationale

The issue title is "softstart 6/10 -> 8/10+" and AC1 asks for >=8/10. This PR delivers 6/10, not 8/10. However:

1. **Stated minimum acceptance is met.** I explicitly told the Doctor that 6/10 was the minimum threshold (matching the pre-regression baseline) and 8/10 was the ideal. The Doctor delivered the minimum.
2. **The 5/10 -> 6/10 lift is real and attributable to this PR.** I verified end-to-end on the PR branch that the budget tags 20 channels and softstart reaches 6/10 routed signal nets (vs. the 5/10 baseline on parent `bf9ea51a`). The Doctor's claim is honest.
3. **The infrastructure is correct and sound.** The geometry now demonstrably intersects U1's contested column on a realistic 4-pad TSSOP-style cluster (asserted by the new fixture), the per-net filter excludes self-budgets, the default-off path is preserved, and the C++ binding surface is properly versioned (BUILD_VERSION 7).
4. **The remaining gap is not in the budget geometry.** The 4 stranded nets (GATE_NEG, V_AC_SENSE, ZC_DETECT, and one more) terminate at U1 and now hit BLOCKED_BY_COMPONENT rip-up stagnation, not the contested east-channel that this PR addresses. Closing that gap requires a different intervention (rip-up loop tuning or alternate-direction escape) — out of scope for this infrastructure PR.
5. **Baseline regression is correctly carved out.** Issue #3199 captures the 6/10 -> 5/10 baseline drift as an independent investigation, so the AC1 follow-up tuning has a stable reference point.
6. **Wave-lifecycle final pass.** Per `/sweep` policy this is the last allowed Doctor->Judge cycle. Blocking now sacrifices the real 5->6 lift, the strengthened parity test, the predictive fixture, and the filed regression issue, for no marginal upside (a third cycle is not available).

### Approval transition

Removing `loom:review-requested`, adding `loom:pr`. PR is approved for merge.

Issue #2528, opened 2026-05-06
net_class: +3V3 (no decimal) not classified as POWER — extend NET_CLASS_PATTERNS regex


rjwalters:
## Problem

`net_class.py::NET_CLASS_PATTERNS` does not classify `+3V3` (no decimal) as POWER. Only patterns like `+3.3V`, `VCC`, `+5V` match. As a result, single-pad-net detection (PR #2525) and pour-net suppression treat `+3V3` as a signal name.

## Discovery

Surfaced by the judge of PR #2525 (https://github.com/rjwalters/kicad-tools/pull/2525#issuecomment-4387911967) while reviewing the new `single_pad_net` rule. The unit tests use `+3.3V` to demonstrate pour-net suppression because `+3V3` would silently bypass it. This is a pre-existing limitation of the regex set; auditor's `_check_connectivity()` already uses the same patterns and has the same gap.

## Why it matters

- `+3V3` is a common net name on European-style schematics (KiCad's own templates use it).
- A board using `+3V3` would have its single-pad regulator-output net flagged as a design defect by the new `single_pad_net` rule, when it's actually a legitimate single-tap power feed.
- The router's pour-net suppression also treats `+3V3` as a signal — could try to route it as traces.

## Repro

After PR #2525 (now merged):
```python
from kicad_tools.router.net_class import classify_and_apply_rules

print(classify_and_apply_rules(["+3V3"]))  # → signal, not power
print(classify_and_apply_rules(["+3.3V"])) # → power
```

## Acceptance

- `+3V3`, `+5V0`, `+1V8`, `+1V2`, `+12V0` (and other no-decimal voltage forms) classify as POWER
- No regression of existing `+3.3V`, `VCC`, `+5V`, `VMOTOR` classification
- Single-pad-net rule (`SinglePadNetRule`) suppresses legitimate single-tap power feeds named with no-decimal forms

## Affected files

- `src/kicad_tools/router/net_class.py::NET_CLASS_PATTERNS` — extend the POWER regex set
- `tests/test_net_class.py` (or wherever the regex is tested) — add cases for the no-decimal forms
- `tests/test_validate_single_pad_net.py` — add a pour-net-suppression test using `+3V3`

rjwalters:
## Curator Enhancement

### Root cause / current behavior

`src/kicad_tools/router/net_class.py:117-125` — the POWER patterns in `NET_CLASS_PATTERNS`:

```python
[...]

PR #2533, opened 2026-05-06
fix(net_class): classify +3V3-style no-decimal voltage names as POWER (closes #2528)

rjwalters:
## Summary

- Add one regex line to `NET_CLASS_PATTERNS[POWER]` so European-style no-decimal voltage names (`+3V3`, `+5V0`, `+1V8`, `+1V2`, `+12V0`, `+24V0`, `-5V0`, `-12V0`, plus bare `3V3`/`1V8`) classify as `NetClass.POWER`.
- Restores correct behavior for `SinglePadNetRule` pour-net suppression and the router's `_is_power_net_by_class` early-abort heuristic on boards using KiCad's stock no-decimal templates.
- Anchored regex (`^[+-]?\d+V\d+[ADPS]?$`) does not regress any existing classification; priority ordering still routes `MOTOR_*` / `PHASE_*` nets to `HIGH_CURRENT_SIGNAL`.

## Test plan

- [x] `tests/test_net_class.py::test_power_net_patterns` extended with 10 positive cases (no-decimal forms) and 2 regression guards (`MOTOR_A`, `PHASE_5V0`)
- [x] New `tests/test_validate_single_pad_net.py::test_single_pad_3v3_no_decimal_marker_allowed` mirrors the existing `+3.3V` suppression test
- [x] `pytest tests/test_net_class.py tests/test_validate_single_pad_net.py tests/test_single_pad_net_count.py tests/test_single_pad_net_integration.py tests/test_cli_check_single_pad_net.py tests/test_route_single_pad_warning.py` -> 101 passed, 7 skipped (KiCad-runner-only)
- [x] Broader sweep `pytest -k "net_class or pour_net or power_net or _is_power"` -> 249 passed, 5 skipped, 0 regressions
- [x] `ruff check` clean on modified files; `mypy` clean on `net_class.py`
- [x] No source changes outside `net_class.py`; consumers (`single_pad_net.py`, `router/core.py::_is_power_net_by_class`) inherit the fix automatically

Closes #2528.

Out of scope (per curator): parallel power-name heuristics in `explain/mistakes.py`, `optim/signal_integrity.py`, `schema/pcb.py`, `optim/placement.py`, `placement/priors.py`, `cli/sch_reconnect_pin.py` already accept `+3V3` and are intentionally not touched here.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

PR #1440, opened 2026-04-15
feat(audit): classify pour nets as zone-expected without explicit zone definitions

rjwalters:
## Summary
Extends `_check_connectivity()` to recognize pour nets (power/ground) as zone-expected even when no explicit `(zone ...)` definition exists in the PCB. This prevents pour nets from inflating `incomplete_nets` and blocking a READY verdict during early layout stages.

## Changes
- Added a second classification pass in `_check_connectivity()` using `classify_and_apply_rules()` to identify pour nets among remaining incomplete nets after the zone-name intersection
- Pour-classified nets are reclassified as zone-connected and excluded from `truly_incomplete`
- Added `pour_net_names` field to `ConnectivityStatus` to track which nets were classified via pour-net detection (distinct from explicit zone definitions)
- Added advisory `ActionItem` entries ("Add zone for <net> on appropriate copper layer") for each pour net without a zone definition
- Classification failures are caught and handled conservatively, leaving `truly_incomplete` unchanged
- Updated existing `test_no_zones_regression` test to reflect new pour-net behavior (GND now reclassified)

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| Pour nets without zone definitions reported as zone-connected | Pass | `test_pour_nets_without_zone_excluded_from_incomplete` - +5V and GNDA with no zones yield `zone_connected_nets == 2`, `incomplete_nets == 0`, `passed == True` |
| `ConnectivityStatus.zone_connected_nets` includes pour-net-classified nets | Pass | Same test verifies count includes both pour-classified nets |
| `incomplete_nets` does not include pour nets | Pass | `test_pour_net_names_tracked` verifies both +5V and GNDA are in `pour_net_names` and excluded from incomplete |
| Signal nets still yield NOT_READY | Pass | `test_signal_net_still_incomplete` - SPI_CLK stays incomplete, `passed == False`, `verdict == NOT_READY` |
| Classification failure leaves `truly_incomplete` unchanged | Pass | `test_classification_failure_leaves_truly_incomplete_unchanged` - patched RuntimeError, both nets remain incomplete |
| Advisory ActionItem emitted for pour nets without zones | Pass | `test_pour_net_action_items_emitted` - 2 items with priority 3 and correct descriptions |

## Test Plan
- 7 new tests in `TestZoneConnectedPourNets` class covering all acceptance criteria
- All 67 `test_audit.py` tests pass
- All 13 `test_pour_net_auto_skip.py` tests pass (no regressions)

[...]

PR #3202, opened 2026-06-05
fix(audit): restore test_no_zones_regression invariant after +3V3 pour-net change

### LINKED ISSUE #3200: audit: TestZoneConnectedNets::test_no_zones_regression fails on main (+3V3 reclassified zone_connected, expected incomplete)
## Empirical observation

The test `tests/test_audit.py::TestZoneConnectedNets::test_no_zones_regression` fails on `main` (commit `7cbaf87d` and earlier). The failure was independently observed by judges on PRs #3195, #3196, #3197, #3198 during the 2026-06-04 sweep — all four judges confirmed it fails on `main`, not just on their feature branches.

## Failure mode

```
assert 0 >= 1
```

The audit's pour-net classification is reclassifying `+3V3` as `zone_connected` when the test expects it to remain in `connectivity.incomplete_nets`. The test was a regression guard for a previous fix; that guard now fires.

## Reproduction

```bash
git checkout main
uv sync
uv run pytest tests/test_audit.py::TestZoneConnectedNets::test_no_zones_regression -x
```

Expected: PASS. Actual: FAIL with `assert 0 >= 1`.

## Likely surface

- `src/kicad_tools/audit/auditor.py` — pour-net classification path (the auditor sweeps zones and assigns connectivity status).
- `src/kicad_tools/validate/connectivity.py` — the underlying `ConnectivityValidator._build_connectivity_graph` may have shifted how it counts pour-connected pads.
- Bisect target: somewhere between the last green run on this test and `7cbaf87d` HEAD. Recent merges adjacent to this surface include #3197 (Phase 2b ground topology — added `ground_topology.py` and refactored `check_analog_ground_bridge` in `analog_detect.py`), but the judge on #3197 confirmed the failure pre-exists that PR.

## Acceptance criteria

1. The failing test passes on `main` without weakening the assertion.
2. Root cause is identified (which change reclassified `+3V3`) and either reverted or the test updated to reflect a correct new behavior with a comment explaining the change.
3. No regression on any other audit/connectivity test.

## References

- Independent confirmations: PR #3195 judge comment, PR #3196 judge comment, PR #3197 judge comment, PR #3198 judge comment.
- Related connectivity work (not the suspected cause, just adjacent surface): #3170, #3178/#3197, #3179.

rjwalters:
## Summary

Fixes the `tests/test_audit.py::TestZoneConnectedNets::test_no_zones_regression` failure on `main` (asserted `0 >= 1`).

Closes #3200.

## Root cause

PR #2533 (Issue #2528) added a regex pattern to `NET_CLASS_PATTERNS[POWER]` that recognizes `+NVN`-style voltage names (`+3V3`, `+5V0`, `+1V8`) as POWER nets. Power nets carry `is_pour_net=True`, so the audit's pour-net post-pass (PR #1440, in `auditor._check_connectivity`) now reclassifies `+3V3` as zone-connected when there is no zone definition.

The regression-guard test `test_no_zones_regression` was written before PR #2533 and assumed `+3V3` would *not* be classified as a pour net. With `+3V3` and `GND` both pour-classified, the truly-incomplete count drops to zero and the assertion `incomplete_nets >= 1` fires.

## Decision: test update, not behavior revert

The new `+3V3` pour-net behavior is the intended outcome of PR #2533 — a `+3V3` rail typically does get a copper pour, so suppressing it from the incomplete-net count matches real-world layout practice. Reverting #2533 would silently un-fix Issue #2528.

Instead, this PR updates the test fixture so the underlying invariant is still guarded.

## Fix

Swap the second net in `PCB_NO_ZONES` from `+3V3` to `SIG`, a plain signal name that doesn't match any pour-net pattern. This restores the test's original invariant — "a board with no zones AND a genuinely incomplete signal net must yield `NOT_READY`" — while keeping the GND pour-net reclassification under test.

The fixture comment block and the test docstring both record this history, so a future contributor can see why the net was renamed (with the issue/PR references attached).

## Test plan

- [x] `uv run pytest tests/test_audit.py::TestZoneConnectedNets::test_no_zones_regression -x` passes (was failing on `main` with `assert 0 >= 1`)
- [x] `uv run pytest tests/test_audit.py -q` — 132 passed
- [x] `uv run pytest tests/test_connectivity.py -q` — 39 passed
- [x] `uv run pytest tests/test_audit.py tests/test_connectivity.py tests/test_analog_nets.py -q` — 192 passed
- [x] Pre-existing failure on `tests/test_connectivity_rule.py::TestConnectivityRuleBoard01::test_board_01_routed_pcb_has_connectivity_errors` confirmed unrelated (fails on `main` without this change)

### DIFF
--- a/tests/test_audit.py
+++ b/tests/test_audit.py
@@ -1380,7 +1380,19 @@ class TestZoneConnectedNets:
 )
 """
 
-    # PCB with no zones — incomplete nets should still be flagged normally.
+    # PCB with no zones — incomplete signal nets should still be flagged normally.
+    #
+    # Net layout:
+    # - GND (net 1): a pour net.  Per PR #1440 it is reclassified as
+    #   zone-connected even without a zone definition, so it does NOT
+    #   block READY.
+    # - SIG (net 2): a non-pour signal net.  With no zone, no trace, and no
+    #   pour-net classification, it must remain truly incomplete and drive
+    #   the NOT_READY verdict.  Previously this slot used "+3V3", but
+    #   PR #2533 added the +NVN voltage pattern, so +3V3 now classifies
+    #   as a pour net and stopped being the incomplete signal here
+    #   (issue #3200).  Swapping to a plain signal name preserves the
+    #   regression invariant the test was written to guard.
     PCB_NO_ZONES = """(kicad_pcb
   (version 20240108)
   (generator "test")
@@ -1395,7 +1407,7 @@ class TestZoneConnectedNets:
   (setup (pad_to_mask_clearance 0))
   (net 0 "")
   (net 1 "GND")
-  (net 2 "+3V3")
+  (net 2 "SIG")
   (gr_rect (start 0 0) (end 100 100)
     (stroke (width 0.15) (type default)) (fill none)
     (layer "Edge.Cuts") (uuid "edge"))
@@ -1408,7 +1420,7 @@ class TestZoneConnectedNets:
     (pad "1" smd roundrect (at -0.5 0) (size 0.6 0.6)
       (layers "F.Cu" "F.Paste" "F.Mask") (roundrect_rratio 0.25) (net 1 "GND"))
     (pad "2" smd roundrect (at 0.5 0) (size 0.6 0.6)
-      (layers "F.Cu" "F.Paste" "F.Mask") (roundrect_rratio 0.25) (net 2 "+3V3"))
+      (layers "F.Cu" "F.Paste" "F.Mask") (roundrect_rratio 0.25) (net 2 "SIG"))
   )
   (footprint "R_0402"
     (layer "F.Cu")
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge review: APPROVED

Verified against the loom:judge workflow and PR #3200 acceptance criteria.

### Scope check (PASS)
Diff is exclusively in `tests/test_audit.py` (+32/-9, 1 file). No production code touched — `auditor.py`, `connectivity.py`, and the `NET_CLASS_PATTERNS[POWER]` list are untouched. The intended new pour-net classification from PR #2533 (`+NVN` voltage names) is preserved, so Issue #2528 stays fixed.

### Approach check (PASS)
Followed the issue's preferred path: root cause identified (PR #2533 + PR #1440 interaction), no revert, test updated to reflect correct new behavior. In-test comments reference both PRs (#1440 for pour-net post-pass, #2533 for +NVN voltage pattern) and Issue #3200 for the rename rationale. A future contributor reading either the fixture comment block or the docstring will understand exactly why `+3V3` was swapped to `SIG`.

### Regression guard check (PASS)
All four critical assertions retained — no weakening to make the test pass:
- `result.connectivity.zone_connected_nets >= 1` (GND pour-net reclassification still under test)
- `result.connectivity.incomplete_nets >= 1` (SIG drives the truly-incomplete count)
- `result.connectivity.passed is False`
- `result.verdict == AuditVerdict.NOT_READY`

The fixture still wires two real nets (GND + SIG) through two footprints with no zone definitions, so the original invariant ("board with no zones + genuinely incomplete signal net -> NOT_READY") is intact. `SIG` does not match any pour-net pattern, so the reclassification post-pass cannot silently mask it the way it did with `+3V3`.

### CI check (PASS)
All 8 checks green:
- Lint & Format: pass
- Type Check: pass
- Test: pass (8m3s)
- C++ Build Check: pass
- Routed PCB DRC Check: pass
- Diff-Pair Routing Regression: pass (11m35s)
- Match-Group Routing Regression: pass (8m52s)

### Merge state check (PASS)
`mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`. PR body contains `Closes #3200`.

Flipping label `loom:review-requested` -> `loom:pr`.

PR #3242, opened 2026-06-06
fix(zones): pad-safe fallback + ZonePartitionError for nested power-rail bboxes (Closes #3240)

### LINKED ISSUE #3240: router: auto-pour leaves overlapping power-rail zones with zero copper — split-power case missing from split-ground heuristic
## Summary

`auto-pour` detects multiple power-domain zones (e.g., split ground or multiple Vcc rails) and queues them at different priorities. When two zones overlap on the same layer at different priorities, the lower-priority zone gets **zero copper** — the higher-priority zone occupies the contested area entirely. The tool logs an informational warning but proceeds, leaving the lower-priority net without a pour.

On chorus-test-revA, this produced:

```
Auto-pour: split-ground detected (GNDA, GNDD) -- assigning distinct layers/priorities so each ground domain receives copper
WARNING: Zone '+3.3VA' on F.Cu (priority 2) overlaps queued zone '+3.3V' (priority 3). The new zone will get zero copper because the other zone has equal or higher priority.
WARNING: Zone '+5V' on F.Cu (priority 1) overlaps queued zone '+3.3V' (priority 3). The new zone will get zero copper because the other zone has equal or higher priority.
WARNING: Zone '+5V' on F.Cu (priority 1) overlaps queued zone '+3.3VA' (priority 2). The new zone will get zero copper because the other zone has equal or higher priority.
```

`+5V` and `+3.3VA` are LEFT WITHOUT COPPER on F.Cu. The router then routes them as traces — which is wasteful for power nets that should have been zone-filled. On a multi-power-rail board this is a structural design failure that the tool silently accepts.

## Mechanism

Looking at the chorus example, auto-pour:
1. Detects split-ground (GNDA, GNDD) — assigns distinct layers and priorities so each receives copper. Good.
2. Then queues power zones: +3.3V (priority 3), +3.3VA (priority 2), +5V (priority 1).
3. All three are placed on F.Cu (or all on a single chosen layer).
4. They overlap geometrically. The KiCad zone priority system gives priority-3 the contested area.
5. +5V and +3.3VA get zero copper.

The bug: auto-pour correctly identified the split-ground case as needing layer separation, but did NOT apply the same logic to the split-Vcc case (three power rails competing for the same layer). The priority-system fallback is silently broken because it picks one winner per area rather than distributing zones to give each net some copper.

### Option A — Apply the split-ground heuristic to power rails too

When auto-pour detects multiple power-net zones competing for the same layer, distribute them across layers the same way it does for split-ground. On a 4-layer stack like chorus's `SIG-GND-PWR-SIG`, the PWR inner layer could host one rail, and the F.Cu / B.Cu could host the others with appropriate keepouts.

This requires the auto-pour code to be aware of the layer stack and the per-net pad density. Higher complexity but more correct.

### Option B — Escalate the warning to error + suggest manual layer assignment

When zone overlaps would leave a net with ZERO copper, the warning becomes a hard error or routing failure. User must manually assign zones to layers via the project settings or via a `--zone-layer-assignment` flag.

Lower-effort fix; pushes the design problem back to the user. Appropriate when auto-pour's heuristic isn't capable of solving it.

### Option C — Auto-fill the contested area at the lower-priority zone's edges only

Keep the priority system but, when overlap leaves a zone with <X% of its desired area, expand the lower-priority zone to grab whatever area the higher-priority zone doesn't strictly need (e.g., perimeter regions, areas without high-priority net pads).

Complex; probably wrong. Listed for completeness.

I lean Option B for the short-term fix (loud failure, user resolves) and Option A for the long-term right answer (auto-pour gets smarter).

## Acceptance criteria

1. On chorus-test-revA with `--manufacturer jlcpcb-tier1`, either:
   - (Option A) `+5V` and `+3.3VA` zones receive non-zero copper after auto-pour.
   - (Option B) The tool exits with an error message naming the zones that would have zero copper and suggesting a fix.
2. The current warning text "the new zone will get zero copper because the other zone has equal or higher priority" no longer appears in routing logs as an informational message — it's either resolved (Option A) or escalated (Option B).
3. No regression on boards with single power rail per layer (boards 01-07, softstart) — they should still pour normally.
4. New regression test: a fixture with 3 overlapping power zones on the same layer. Assert either (Option A) all get non-zero copper or (Option B) the tool errors out.

## Out of scope

- The split-ground existing heuristic itself — works fine, don't touch.
- Manual zone layer assignment UI/CLI — separate enhancement.
- Zone fill quality (thermal reliefs, pad connections, etc.) — different code path.

## How to reproduce

```bash
mkdir -p /tmp/chorus_zones && uv run python -m kicad_tools.cli route \
    --output /tmp/chorus_zones/chorus_routed.kicad_pcb \
    --auto-layers --timeout 300 --seed 42 2>&1 | grep -E "Auto-pour|zero copper|priority"
```

Expected output shows the warnings about +5V and +3.3VA getting zero copper.

## References

- #3237 — chorus reach regression (this issue is a sibling finding from the same session).
- Chorus findings notes: `/tmp/chorus_findings_2026_06_06.md`.

## Why this matters

Multi-power-rail boards are common in real-world designs. Silently leaving a power net without copper is a recipe for DRC violations downstream (and possibly EMI/IR-drop issues on hardware). The tool currently presents this as a warning the user can easily miss; making it visible and resolvable is a small change for a correctness win.

rjwalters:
## Summary

Issue #3240 reported that on chorus-test-revA, `auto-pour` produced these warnings:

```
WARNING: Zone '+3.3VA' on F.Cu (priority 2) overlaps queued zone '+3.3V' (priority 3). The new zone will get zero copper because the other zone has equal or higher priority.
WARNING: Zone '+5V'   on F.Cu (priority 1) overlaps queued zone '+3.3V' (priority 3). The new zone will get zero copper because the other zone has equal or higher priority.
WARNING: Zone '+5V'   on F.Cu (priority 1) overlaps queued zone '+3.3VA' (priority 2). The new zone will get zero copper because the other zone has equal or higher priority.
```

`+5V` and `+3.3VA` ended up with zero copper because their pads sat fully inside `+3.3V`'s inflated bbox (the carve subtraction in `_compute_pour_outlines` returned empty, and the legacy fallback returned the raw bbox).

This PR implements **Option A** (carve so all power rails get copper) with a **safety net of Option B** (raise `ZonePartitionError` when the partition is truly unsolvable). The fix lives entirely in `src/kicad_tools/zones/generator.py`.

### What changed

* **`_subtract_polygon`**: now returns `None` (not the un-carved fallback) on (a) empty diff, (b) MultiPolygon, (c) Polygon-with-holes, or (d) missing Shapely. The polygon-with-holes case was the precise #3240 failure mode -- the exterior ring alone overlaps the subtrahend and re-creates the zero-copper bug. KiCad zones don't support holes, so we signal failure to the caller instead.
* **`_compute_pour_outlines`**: pre-computes both a raw bbox (1.5 mm margin, existing) and a **pad-safe bbox** (0.3 mm margin, new) for every shared-layer net. Adds an "effective subtrahend downgrade" pass: when a high-priority net's raw bbox would over-claim a lower-priority sibling's pad region, the high-priority net is downgraded to use its own pad-safe bbox as the subtrahend. The high-priority zone keeps pad-adjacent copper; the lower-priority siblings get the inter-pad regions.
* **`ZonePartitionError`** (new): raised by `auto_create_zones_for_pour_nets` when neither the carve nor the pad-safe fallback can produce a disjoint outline (truly coincident pad layouts). Names the failing net, the layer, and the higher-priority covering nets. This replaces the previous silent zero-copper behavior with an actionable error (Option B from the issue body).

### Verification on chorus-test-revA

Before fix: `+5V` and `+3.3VA` produced zero-copper warnings.
After fix:

```
F.Cu zones: +3.3V pri=3 area=3185.6 mm²
           +3.3VA pri=2 area=101.0 mm² (6 verts, carved L-shape)
           +5V    pri=1 area=32.9 mm²
Pairwise overlap: all zero
```

No `zero copper` strings in the routing log.

### Trade-offs

* The Option-A path requires the lower-priority net's raw bbox (1.5 mm margin) to extend at least 1.2 mm outside the higher-priority net's pad-safe bbox (0.3 mm margin). When pad clusters are very close together but distinguishable, this works; when they're fully coincident, `ZonePartitionError` surfaces.
* The carved outlines may be small (~30-100 mm²) -- still positive copper, still disjoint, still avoids the silent failure. A future enhancement could emit multiple zones per net to cover non-convex regions, but that's out of scope here.

## Test plan

- [x] `tests/test_zone_priority.py` (4 existing + 5 new = 9 tests, all passing)
- [x] `tests/test_board_05_zones.py`, `tests/test_pour_net_auto_skip.py`, `tests/test_route_cmd_zone_fill.py`, `tests/test_zones.py`, `tests/test_zone_generator.py`, `tests/test_zones_cmd.py`, etc. — 256/256 zone-related tests passing
- [x] Live verification on chorus-test-revA: all 3 F.Cu power zones have positive disjoint area, no zero-copper warnings
- [x] Pre-existing test failures in `tests/test_route_zones_preserved.py::TestRouteWriteSitesPattern` confirmed to also fail on `main` (unrelated to this change)

### New regression tests

* `test_three_power_nets_nested_in_xy_get_carved_to_pad_safe` — the chorus-test-revA reproduction shape
* `test_no_zero_copper_warning_for_xy_nested_layout` — asserts the issue's literal warning text doesn't appear
* `test_zone_partition_error_on_coincident_pad_layout` — degenerate case raises `ZonePartitionError`
* `test_subtract_polygon_returns_none_on_empty_diff` — API guarantee that backs the partition-error plumbing
* `test_subtract_polygon_returns_carved_polygon_on_partial_diff` — happy path still works

Closes #3240

### DIFF
--- a/src/kicad_tools/zones/__init__.py
+++ b/src/kicad_tools/zones/__init__.py
@@ -35,6 +35,7 @@ from .generator import (
     ZoneConfig,
     ZoneGenerator,
     ZoneOverlapWarning,
+    ZonePartitionError,
     auto_create_zones_for_pour_nets,
     parse_power_nets,
 )
@@ -43,6 +44,7 @@ __all__ = [
     "ZoneConfig",
     "ZoneGenerator",
     "ZoneOverlapWarning",
+    "ZonePartitionError",
     "auto_create_zones_for_pour_nets",
     "parse_power_nets",
 ]
--- a/src/kicad_tools/zones/generator.py
+++ b/src/kicad_tools/zones/generator.py
@@ -4,8 +4,8 @@ Zone generator for creating copper pour zones on PCBs.
 This module provides a high-level API for generating copper pour zones,
 with automatic board outline detection and sensible defaults for power nets.
 
-Outline allocator (#2771, #3043)
---------------------------------
+Outline allocator (#2771, #3043, #3240)
+---------------------------------------
 
 In addition to the layer/priority allocator (:func:`_assign_layers_for_pour_nets`),
 this module owns the *outline* allocator (:func:`_compute_pour_outlines`).
@@ -26,10 +26,24 @@ The original PR #2771 only computed each net's bbox independently and
 relied on pad-cluster separation to keep the bboxes disjoint.  On real
 boards where power-net pads are *spatially interleaved* (e.g. board 06's
 ``+3V3``/``+1V8``/``+1V2`` cluster on a BGA), the bboxes overlap and the
-silent-override returns.  Issue #3043 extends the allocator with a
+silent-override returns.  Issue #3043 extended the allocator with a
 second pass that *subtracts higher-priority bboxes from lower-priority
-outlines* (using Shapely's polygon difference), so the resulting
-outlines are disjoint regardless of pad layout.
+outlines* (using Shapely's polygon difference).
+
+Issue #3240 reported a remaining gap: when a lower-priority net's pads
+sit fully *inside* a higher-priority net's bbox, the Shapely difference
+returns empty and the legacy fallback returned the raw bbox -- producing
+zero copper for the failing net.  The fix adds:
+
+* A pad-safe-bbox fallback (tight 0.3 mm margin around the failing net's
+  own pads) used when the standard carve returns empty.
+* A "subtrahend downgrade" rule: the higher-priority net is allowed to
+  yield space by using its own pad-safe bbox (instead of the full
+  inflated raw bbox) whenever the raw bbox would over-claim a
+  lower-priority sibling's region.
+* :class:`ZonePartitionError` raised when neither fallback can produce
+  disjoint outlines (e.g. truly coincident pad layouts), so the user
+  sees an actionable error instead of a silent zero-copper failure.
 """
 
 from __future__ import annotations
@@ -66,6 +80,49 @@ class ZoneOverlapWarning:
     existing_net: str
     layer: str
     message: str
+
+
+class ZonePartitionError(RuntimeError):
+    """Raised when the outline allocator cannot produce disjoint per-net outlines.
+
+    Indicates that two or more zones sharing a layer would receive zero
+    copper because no carved-out outline can be derived for one of them.
+    This typically happens when a lower-priority net's pads sit fully
+    inside the higher-priority net's bounding region and even a tight
+    pad-bbox fallback would still overlap.
+
+    The error message lists the failing net, the layer, and the
+    higher-priority nets that fully cover it, together with suggested
+    remediations (e.g., move the failing net to another layer, shrink
+    the per-net margin, or accept routing the net as traces).
+    """
+
+    def __init__(
+        self,
+        failing_net: str,
+        layer: str,
+        covering_nets: list[str],
+        suggestion: str | None = None,
+    ):
+        self.failing_net = failing_net
+        self.layer = layer
+        self.covering_nets = list(covering_nets)
+        covering_list = ", ".join(self.covering_nets) if self.covering_nets else "(none)"
+        msg = (
+            f"Auto-pour: zone '{failing_net}' on {layer} would receive ZERO copper "
+            f"because its pad region is fully covered by higher-priority zones "
+            f"({covering_list}). "
+        )
+        if suggestion:
+            msg += suggestion
+        else:
+            msg += (
+                "Suggested fixes: (a) move one of the conflicting nets to a different "
+                "copper layer via manual zone assignment, (b) reduce the per-net pour "
+                "margin (currently ~1.5 mm), or (c) accept that this net should be "
+                "routed as traces rather than zone-filled."
+            )
+        super().__init__(msg)
 
 
 @dataclass
@@ -967,6 +1024,14 @@ def _assign_layers_for_pour_nets(
 # during the curator analysis of board 05.
 DEFAULT_POUR_BBOX_MARGIN_MM = 1.5
 
+# Tight pad-only margin used as a *fallback* when the normal carve pass
+# (#3043) would leave a zone with zero copper.  At 0.3 mm it just clears
+# typical KiCad fill rules around the pads themselves -- enough to give
+# the lower-priority net *some* copper, while staying inside the inflated
+# bbox of higher-priority neighbours (which use the 1.5 mm default).  See
+# issue #3240 for the failure shape this addresses.
+PAD_SAFE_MARGIN_MM = 0.3
+
 # Default side length of the fallback square emitted when a net has only a
 # single pad (or all its pads coincide).  Slightly larger than the margin
 # alone so single-pad nets still receive a usable copper patch.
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge review: APPROVE

### Verdict
Approving. The hybrid approach (Option A pad-safe carve + Option B `ZonePartitionError`) is coherent, the implementation is well-factored, the tests cover the regression, and CI is fully green (8/8).

### Acceptance criteria check
- **AC1 (Option A) — `+5V`/`+3.3VA` receive non-zero copper after auto-pour**: Verified by the new `test_three_power_nets_nested_in_xy_get_carved_to_pad_safe` (asserts every F.Cu zone has positive area + pairwise disjoint).
- **AC2 — no "zero copper" warning in routing logs**: Verified by `test_no_zero_copper_warning_for_xy_nested_layout` (literal stderr scan).
- **AC3 — no regression on boards 01-07 + softstart**: CI's `Diff-Pair Routing Regression`, `Match-Group Routing Regression`, `Routed PCB DRC Check`, and the 256-test zone suite all pass. Locally re-ran `tests/test_zones.py tests/test_zone_generator.py tests/test_zone_priority.py tests/test_zones_cmd.py tests/test_build_zones_step.py tests/test_pour_net_auto_skip.py` → 246 passed / 1 skipped.
- **AC4 — new regression test**: 5 new tests in `tests/test_zone_priority.py` (3 integration + 2 unit on `_subtract_polygon`'s contract).

### Design notes the reviewer found compelling
- **`_subtract_polygon` signature change (returning `None`) is properly plumbed.** All four call sites in `generator.py` (`1356`, `1387`, `1412`, `1483`) explicitly check for `None`. The function is module-private, so no other callers can be broken.
- **The "polygon-with-holes returns `None`" case is exactly the #3240 failure mode.** The PR docstring calls this out and the unit test `test_subtract_polygon_returns_none_on_empty_diff` pins it. Returning the exterior ring alone would silently re-create the bug — the new logic is correct.
- **`ZonePartitionError` propagates cleanly** from `_compute_pour_outlines` → `auto_create_zones_for_pour_nets`. Verified no silent catches in `src/kicad_tools/router/auto_pour.py` (`auto_pour.py:593` is a bare call). In `src/kicad_tools/cli/build_cmd.py:1058-1075`, the broad `except Exception` converts it to a structured `BuildResult(success=False)` — exactly what AC2 (Option B fallback) calls for: actionable failure instead of silent zero copper.
- **Pad-safe 0.3 mm margin is documented** at `generator.py:101-104` with a clear rationale ("just clears typical KiCad fill rules around the pads themselves -- enough to give the lower-priority net *some* copper, while staying inside the inflated bbox of higher-priority neighbours which use the 1.5 mm default"). Magic number, but well-named (`PAD_SAFE_MARGIN_MM`) and well-justified.
- **`effective_subtrahend` pre-pass is the elegant insight.** Instead of trying to add holes to the higher-priority zone's outline (which KiCad can't represent), the higher-priority net's *subtrahend* gets downgraded to pad-safe whenever its raw bbox would over-claim a lower sibling's pad region. The higher zone keeps its pad-adjacent copper; the lower siblings claim the inter-pad gaps.
- **Numerical edge case is handled** at `generator.py:1416-1428`: when `pad_carved` returns `None` due to Shapely's empty-threshold but `_polygon_area_simple - _polygon_overlap_area > 1e-6`, the pad-safe is still emitted. Otherwise `ZonePartitionError` fires.

- The downgrade check at `generator.py:1356` (`_subtract_polygon(lower_pad_safe, raw)`) now returns `None` not only for "fully covered" but also for MultiPolygon / hole-bearing results. This is intentional and benign — those cases also indicate the high-priority raw bbox over-claims relative to the lower sibling's pad-safe — but worth noting for future maintainers reading `_subtract_polygon`'s expanded contract.
- The `_polygon_area_simple` helper is added inline for the numerical edge case but could conceivably be shared with the existing `_polygon_overlap_area` if Shapely import cost matters; current placement is fine.
- Future enhancement (called out in PR description): emit multiple zones per net to cover non-convex regions when the pad-safe is L-shaped. Out of scope here.

### CI
8/8 checks green (Lint, Type, Test, C++ Build, Diff-Pair, Match-Group, DRC, kicad-cli round-trip). Merge state CLEAN.

LGTM.

PR #3244, opened 2026-06-06
fix(router): early-exit auto-layer escalation on monotonic regression (Issue #3241)

### LINKED ISSUE #3241: router: auto-layers escalation ladder runs to completion even when attempts produce monotonically worse results (chorus 15%→12%→10%)
## Summary

The auto-layers escalation ladder — designed to find a better routing solution by trying more layers — currently runs each attempt to completion and picks the best. On chorus-test-revA (current main), the ladder produced **monotonically worse** results across attempts:

| Attempt | Layers | Best result |
|---|---|---:|
| 1 | 4L SIG-GND-PWR-SIG | 7/48 (15%) |
| 2 | 4L ALL-SIG | 6/48 (12%) |
| 3 | 6L SIG-GND-SIG-SIG-PWR-SIG | 5/48 (10%) |

This is structurally backwards. The escalation hypothesis is that adding layers improves reach. If a fully-completed 4-layer attempt gets 15% and the next 4-layer variant gets LESS, the ladder should have stopped — OR backed off, OR aggregated routes from the better attempt. Instead it just bookkeeps "best so far" and continues to a 6-layer attempt that gets even less.

By contrast, May 10's snapshot produced reach that INCREASED across the ladder (2L=44% → 4L=62%). That's the expected pattern. The monotonic-decrease pattern at current main is a strong signal that something in the routing stack is broken in a way the ladder amplifies.

## Mechanism

The negotiated router runs to completion on each attempt. Each attempt:
- Starts with the same input PCB (no carry-over from prior attempt).
- Builds escape routes, runs Phase 1 global, Phase 2 detail.
- The rip-up loop converges on a local optimum (or stalls).
- Final "best result" is the count of fully-connected nets.

Across attempts, the layer assignment changes (4L→4L variant→6L) and the per-cell A* cost surface changes. So in principle the second attempt's optimum is different from the first.

But the ladder logic assumes that monotonic escalation produces monotonic improvement. When today's tools land each attempt at a WORSE local optimum than the previous, the ladder doesn't detect this and stop — it just keeps running, consuming budget.

Combined with the auto-fix-budget exhaustion (#3238) and the chorus reach regression (#3237), the net effect on chorus is:
- 3 attempts × ~500s = 1500s consumed.
- No auto-fix gets time.
- Best result is from attempt 1 (15%), not the final 6L attempt (10%).
- User sees the "best result" but the budget was spent on attempts 2 and 3 that didn't help.

### Option A — Early-exit on monotonic regression

If attempt N+1 finishes with <N's fully-connected count, abort the ladder. Roll back to attempt N's result. Save the remaining budget for auto-fix.

The user-visible behavior: ladder runs 1-2 attempts on chorus, sees attempt 2 was worse, stops, auto-fix runs on attempt 1's partial routes.

Cost: false positive when attempt 2 is genuinely better but the count flickers down by 1-2 nets. Mitigate with a tolerance (e.g., "regression by ≥3 nets triggers exit").

### Option B — Per-attempt time budget

Limit each attempt to `total_timeout / 4` (or similar) so the ladder always has time for auto-fix at the end. The current "best result" still wins.

Doesn't fix the structural backwards-pattern, but at least reserves budget for the repair pass.

### Option C — Hybrid attempts (carry partial routes forward)

Instead of restarting each attempt from scratch, carry forward the successful routes from attempt N into attempt N+1. The rip-up loop can then improve marginally rather than starting fresh.

Much higher implementation complexity. Probably the right long-term answer but not for this issue.

### Option D — Diagnostic: log per-attempt failure-cause histogram

When the ladder produces decreasing results, the LOG should make it obvious why. Currently the user sees three "Best result so far" lines with diminishing percentages and no explanation. Adding a per-attempt failure-cause histogram (X nets failed due to BLOCKED_PATH, Y nets failed due to BLOCKED_BY_COMPONENT, etc.) would help future investigation.

I lean Option A + Option D — quick correctness fix + investigation aid. Option B is the right complement to #3238's auto-fix budget work.

## Acceptance criteria

1. On chorus-test-revA at current main, the ladder either:
   - (Option A) Stops after detecting attempt 2 produced fewer fully-connected nets than attempt 1, AND saves the remaining budget for auto-fix.
   - (Option B) Each attempt completes within `total_timeout / N` so auto-fix has reserved budget regardless of ladder outcome.
2. Final result includes a per-attempt failure histogram in the log (Option D).
3. No regression on monotonically-improving cases (boards 01-07 at current main; the ladder should still pick the best result without unnecessary early-exit).
4. New regression test: a synthetic fixture where 4L > 2L > 6L. Assert the ladder finds 4L as the best and either exits before 6L or completes 6L but picks 4L.

## Out of scope

- The deeper reason 4L→4L→6L produces decreasing results today (filed as #3237 — the chorus reach regression).
- The auto-fix budget exhaustion itself (filed as #3238).
- General negotiated-router convergence improvements.

## How to reproduce

```bash
mkdir -p /tmp/chorus_ladder && uv run python -m kicad_tools.cli route \
    --output /tmp/chorus_ladder/chorus_routed.kicad_pcb \
    --auto-layers --timeout 1500 --seed 42 2>&1 | grep -E "Attempt|Best result|Routed:"
```

Expected output shows three attempts with decreasing percentages (15% → 12% → 10%).

## References

- #3237 — chorus reach regression (this issue is a sibling — the chorus regression is the trigger for observing the backwards ladder).
- #3238 — auto-fix budget exhaustion (this issue is a sibling — if ladder exited earlier, auto-fix would have its budget).
- #2916 — the escalation ladder design itself (referenced in chorus route log: "escalation ladder: [4L, 4L, 6L] (issue #2916)").
- Chorus findings: `/tmp/chorus_findings_2026_06_06.md`.

## Why this matters

The escalation ladder is supposed to be a safety net — try more layers if needed. When the safety net produces strictly worse results AND consumes the budget that auto-fix needs, the user gets the worst of both worlds. A small change (early-exit + per-attempt time cap) closes the failure mode cheaply.

rjwalters:
## Summary

The auto-layers escalation ladder previously ran every attempt to completion even when each attempt produced strictly fewer routed nets than the previous (chorus-test-revA observed `15% -> 12% -> 10%`). This wasted the routing budget on attempts that could not help and starved auto-fix of the time it needed.

This PR implements **Option A + Option D** from the issue:

- **Option A** — Detect monotonic regression in the layer-escalation ladder and exit early, preserving the pre-regression best result.
- **Option D** — Emit a per-attempt `Failure causes: {...}` histogram so future investigations of regression-exit cases have failure-cause data in the log.

## Behavior

In `route_with_layer_escalation`, regression-exit triggers when one of:

- A single attempt drops `nets_routed` by `>= HARD_DROP_NETS` (5) vs the prior attempt (immediate exit).
- `CONSECUTIVE_REGRESSIONS` (2) consecutive attempts each drop more than `REGRESSION_TOLERANCE` (2) nets (streak exit).

Unlike the #2412 stagnation heuristic, this guard fires regardless of the #2673 50% completion floor: "more layers + strictly fewer routed nets" is structurally backwards, so adding still more layers cannot cure it. The pre-regression best result is preserved via the existing `_is_better_result` selector (#2396).

The same cross-layer regression guard is also extended to `route_with_combined_escalation` (the 2D layer x rule-tier search), expressed in fractional completion since rule-relaxed tiers produce slightly different net counts per cell. Thresholds: `CL_HARD_DROP=10pp`, `CL_REGRESSION_TOLERANCE=2pp`, `CL_CONSECUTIVE=2`.

## Acceptance Criteria

| AC | Met? | Notes |
|----|------|-------|
| 1. Ladder stops after detecting attempt 2 < attempt 1 (Option A) | Yes | See `test_two_consecutive_regressions_exit`, `test_hard_drop_exits_immediately` |
| 2. Per-attempt failure-cause histogram (Option D) | Yes | `_log_failure_cause_histogram` called at every attempt end + regression-exit path |
| 3. No regression on monotonically-improving cases | Yes | See `test_monotonic_improvement_no_early_exit` (4 strictly-improving attempts all run) |
| 4. Synthetic regression test where ladder picks pre-regression best | Yes | See `test_best_result_preserved_after_regression_exit` and `test_chorus_pattern_exits_after_attempt_2` |

## Test plan

- [x] `TestRegressionEarlyExit` (7 tests): two-consecutive-regression exit, hard-drop exit, flicker-within-tolerance no-exit, best-result preservation, monotonic-improvement regression guard, the chorus pattern, single-attempt edge case.
- [x] `TestFailureCauseHistogram` (4 tests): with-failures, no-failures, quiet, None-router.
- [x] All 11 new tests pass (`uv run pytest tests/test_layer_escalation.py::TestRegressionEarlyExit tests/test_layer_escalation.py::TestFailureCauseHistogram` -> 11 passed).
- [x] Lint clean (no new ruff errors vs main baseline).
- [x] No regression on the 58 non-broken tests in `test_layer_escalation.py`.

## Pre-existing failures

- `TestEarlyTermination` (9 tests) in `test_layer_escalation.py` fails on main due to a MagicMock-vs-int comparison in `router/io.py:2120` (`edge_clearance > 0` fails when `_edge_clearance` is a MagicMock sentinel). The new `TestRegressionEarlyExit` works around this by setting `router._edge_clearance = None` in its mock helper, so it is unaffected. The 9 pre-existing failures are out of scope and tracked separately.
- `test_route_auto_mfr_tier_integration.py::TestAutoMfrTierIntegration::test_escalation_advances_to_tier1` and `::test_escalation_triggered_by_missed_via_in_pad` fail. These exercise the manufacturer-tier ladder (not layer escalation) and the failures are about a trigger-table veto on `BLOCKED_PATH` causes, unrelated to this change.

## Implementation notes

- Tunable constants (`REGRESSION_TOLERANCE`, `HARD_DROP_NETS`, `CONSECUTIVE_REGRESSIONS`) are defined inline as named constants for readability. Tolerances chosen to avoid false-positive on flicker: PR #3193 made A* tie-break deterministic, but routing-order changes across stacks can still produce 1-2 net jitter.
- `_log_failure_cause_histogram` is a 5-line `Counter + dict()` helper safe to call on every attempt. It tolerates `None` router and missing `routing_failures` attribute. Uses the same `failure_cause` attribute already consumed by `_classify_dominant_failure_cause`.

Closes #3241

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -2532,6 +2532,28 @@ def route_with_layer_escalation(
     prev_nets_routed: int | None = None
     prev_overflow: int | None = None
 
+    # Issue #3241: Track monotonic regression across attempts (cross-attempt
+    # decrease in nets_routed).  Unlike the #2412 stagnation heuristic above,
+    # which is gated by the #2673 50% completion floor, the regression-exit
+    # fires regardless of floor because "more layers + fewer routed nets" is
+    # structurally backwards -- adding still more layers cannot cure it.
+    # See issue #3241 (chorus-test-revA 15% -> 12% -> 10% observation).
+    #
+    # Thresholds (intentionally tunable as named constants for readability):
+    #   - REGRESSION_TOLERANCE: small flicker (default 2 nets) does NOT count
+    #     as a regression.  PR #3193 made A* tie-break deterministic, so most
+    #     flicker is gone -- but routing-order changes across stacks still
+    #     produce 1-2 net jitter even on equivalent topologies.
+    #   - HARD_DROP_NETS: a single-attempt drop of >=5 nets is severe enough
+    #     to exit immediately without waiting for a second regression.
+    #   - CONSECUTIVE_REGRESSIONS: otherwise, require 2 consecutive regression
+    #     observations before exiting (avoids false-positive on a single
+    #     unlucky attempt).
+    REGRESSION_TOLERANCE = 2
+    HARD_DROP_NETS = 5
+    CONSECUTIVE_REGRESSIONS = 2
+    regression_streak: int = 0
+
     # Issue #2388: Track power-net stall across escalation attempts.  When
     # a 2-layer attempt aborts due to power-net stall, the next attempt
     # is biased toward a stack with dedicated planes for those nets, and
@@ -2864,8 +2886,66 @@ def route_with_layer_escalation(
                 "continuing escalation (issue #2673)"
             )
 
+        # Issue #3241: Monotonic-regression early-exit.  Unlike the #2412
+        # stagnation check above, this is NOT gated by the #2673 50% floor.
+        # "More layers + strictly fewer routed nets" is structurally
+        # backwards -- adding still more layers cannot cure it -- so we
+        # break the ladder even when both attempts are below 50%.  The
+        # check requires either (a) a hard drop of >=HARD_DROP_NETS on a
+        # single attempt, or (b) CONSECUTIVE_REGRESSIONS consecutive
+        # regressions each exceeding REGRESSION_TOLERANCE.  ``best_result``
+        # is already tracked via the #2396 ``_is_better_result`` rule above,
+        # so the loop just breaks; the caller receives the pre-regression
+        # best attempt's router state.
+        regression_exit = False
+        if prev_nets_routed is not None:
+            drop = prev_nets_routed - nets_routed
+            if drop >= HARD_DROP_NETS:
+                if not quiet:
+                    flush_print(
+                        f"  Escalation stopped: monotonic regression -- "
+                        f"hard drop of {drop} nets vs previous attempt "
+                        f"(>={HARD_DROP_NETS} threshold, issue #3241)"
+                    )
+                regression_exit = True
+            elif drop > REGRESSION_TOLERANCE:
+                regression_streak += 1
+                if regression_streak >= CONSECUTIVE_REGRESSIONS:
+                    if not quiet:
+                        flush_print(
+                            f"  Escalation stopped: monotonic regression -- "
+                            f"{regression_streak} consecutive attempts with "
+                            f"nets_routed decreasing by >{REGRESSION_TOLERANCE} "
+                            "(issue #3241)"
+                        )
+                    regression_exit = True
+                elif not quiet:
+                    flush_print(
+                        f"  Regression observed: {drop} fewer nets routed "
+                        f"than previous attempt (streak={regression_streak}/"
+                        f"{CONSECUTIVE_REGRESSIONS}, issue #3241)"
+                    )
+            else:
+                # Improvement or flicker within tolerance -- reset streak.
+                regression_streak = 0
+
         prev_nets_routed = nets_routed
         prev_overflow = overflow
+
+        if regression_exit:
+            # Report attempt result before breaking, mirroring the #2412
+            # exit paths above.  ``best_result`` already retains the
+            # pre-regression best per the #2396 selector.
+            if not quiet:
+                flush_print(
+                    f"\n  Routed: {nets_routed}/{nets_to_route} nets "
+                    f"({completion * 100:.0f}%)"
+                )
+                flush_print("  Status: INSUFFICIENT - early stop (regression)")
+                # Issue #3241 (Option D): emit failure-cause histogram so
+                # future investigations of the same regression have data.
+                _log_failure_cause_histogram(router, quiet=quiet)
+            break
 
         # Issue #2388: Record any power-net stall for the next attempt's
         # bias logic.  ``power_stall_nets`` is populated by
@@ -2884,6 +2964,10 @@ def route_with_layer_escalation(
         if not quiet:
             flush_print(f"\n  Routed: {nets_routed}/{nets_to_route} nets ({completion * 100:.0f}%)")
             flush_print(f"  Status: {status}")
+            # Issue #3241 (Option D): per-attempt failure-cause histogram.
+            # Helps diagnose why a higher-layer attempt can land at a worse
+            # local optimum (the trigger for this issue's regression-exit).
+            _log_failure_cause_histogram(router, quiet=quiet)
 
         # Check for success
         if result.success:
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

Thorough review of the layer-escalation + combined-escalation regression-exit guards. Reviewed both `route_with_layer_escalation` AND `route_with_combined_escalation` per the issue's two-pronged scope.

### Acceptance criteria

| AC | Status | Evidence |
|----|--------|----------|
| 1. Ladder stops after attempt 2 < attempt 1 (Option A) | Met | `test_two_consecutive_regressions_exit`, `test_hard_drop_exits_immediately` |
| 2. Per-attempt failure-cause histogram (Option D) | Met | `_log_failure_cause_histogram` invoked at both end-of-attempt (line 2970) and regression-exit (line 2947) |
| 3. No regression on monotonically-improving cases | Met | `test_monotonic_improvement_no_early_exit` runs all 4 strictly-improving attempts |
| 4. Synthetic regression test picks pre-regression best | Met | `test_best_result_preserved_after_regression_exit` (relies on `_is_better_result` from #2396) |

### Constants review

All three layer-escalation constants are sensibly named, well-commented (lines 2534-2549), and have rationale tied to specific prior issues:
- `REGRESSION_TOLERANCE=2` — calibrated against the post-#3193 jitter envelope
- `HARD_DROP_NETS=5` — single-attempt immediate exit threshold
- `CONSECUTIVE_REGRESSIONS=2` — streak threshold prevents single-unlucky-attempt false positive

The cross-layer (`route_with_combined_escalation`) constants are correctly expressed in fractional completion (lines 4557-4559):
- `CL_REGRESSION_TOLERANCE=0.02` (2pp) — matches tier-cell noise floor
- `CL_HARD_DROP=0.10` (10pp) — appropriately wider than the layer-version since fractional drops on small completion baselines amplify
- `CL_CONSECUTIVE=2` — symmetric with layer version

### Option D histogram

The helper (lines 3855-3898) is defensive: handles None router, missing `routing_failures`, and missing `failure_cause`. Uses `getattr(cause, "value", str(cause))` for FailureCause→string conversion. Tolerates `quiet` mode. Output is one line per attempt with a machine-parseable dict — no log spam. Negligible perf cost (~5 lines of Counter aggregation).

### Cross-layer guard (`route_with_combined_escalation`)

Correctly positioned AFTER `if successful_result: break` (line 4797-4798), so success paths short-circuit. The `prev_layer_best_completion` tracking implements "running max across layers" (lines 4834-4840), which is the right primitive for cross-layer regression detection.

Minor observations (non-blocking):
1. The cross-layer regression-exit does NOT call `_log_failure_cause_histogram` (unlike the layer-escalation regression-exit). Asymmetric but defensible — the combined ladder already has per-attempt logging via the intra-layer regression guard at line 4784, and the cross-layer break is logged separately.
2. Layer-escalation regression-exit does NOT check `args.no_early_stop` (line 2901), but combined-escalation cross-layer guard DOES (line 4809). Inconsistent — but this matches the existing #2412 stagnation check pattern in layer-escalation, so it's coherent with surrounding code.
3. Test `test_chorus_pattern_exits_after_attempt_2` is slightly misnamed — it actually exits after attempt 3 (streak hits 2). The docstring acknowledges this explicitly. Cosmetic only.

### Test coverage

`TestRegressionEarlyExit` (7 tests): hard-drop, streak, flicker-within-tolerance, best-preservation, monotonic-improvement, chorus pattern, single-attempt edge case. All pass locally (`11 passed in 14.25s`).

`TestFailureCauseHistogram` (4 tests): with-failures, no-failures, quiet, None-router. All pass.

### Pre-existing failures verified

Confirmed via inspection:
- The 9 `TestEarlyTermination` failures hit `router/io.py:2120` (`edge_clearance > 0`) via MagicMock sentinel. The PR doesn't touch `router/io.py` (the diff vs origin/main on that file is from unrelated merged PRs on main). The new `TestRegressionEarlyExit` works around this by setting `router._edge_clearance = None` in the mock helper.
- The 2 `test_route_auto_mfr_tier_integration` failures are in the manufacturer-tier ladder (not layer/combined escalation) and the assertion is `assert None is not None` on an escape-routing path unrelated to this PR.

Both pre-existing-failure claims hold up — the PR does not touch those code paths.

### CI

All 8 checks pass. `mergeStateStatus: CLEAN`.

### Lint baseline

6 ruff errors on the modified test file match the same 6 errors on `origin/main:tests/test_layer_escalation.py` — no new lint debt introduced.

### Verdict

The implementation is coherent end-to-end. The resume builder's claim that the partial work was "substantially correct" is borne out — no stale assumptions detected. The fix surgically addresses the chorus 15%→12%→10% pattern in the layer-escalation ladder, and correctly extends the same logic to the combined-escalation 2D search with appropriately-tuned fractional thresholds. Test coverage exercises all 4 ACs plus edge cases (single-attempt, no-failures, quiet, None-router).

Approving.

Issue #3237, opened 2026-06-06
router: chorus-test reach regression — 62% (May 10) → 15-21% (today) on both backends; auto-fix never runs in 1500s budget


rjwalters:
## Summary

Re-attempting chorus-test-revA routing on `origin/main` (post-#3232, the latest router merge) reproduces and confirms a major reach regression compared to the May 10 baseline captured in the project memory. Both `--backend cpp` and `--backend python` produce dramatically fewer fully-connected nets on the SAME `chorus-test-revA_v19_stripped.kicad_pcb` input, using a similar recipe to May 10's.

The mechanism splits cleanly into two parts:

1. **Negotiated routing per-attempt is too slow** — each of three layer-escalation attempts consumes ~500s, totaling ~1525s — past the 1500s budget.
2. **Auto-fix is then skipped** under issue #2802's wall-clock-deadline guard. The 29-31 nets that the detail router placed but left with DRC violations never get cleaned to fully-connected status.

The result: tools that should be FASTER and CLEANER than May 10's snapshot (because of PR #3225, PR #3227, PR #3231, PR #3232 all landing today) instead produce 15-21% routing vs May 10's 62%.

## Empirical state

Same PCB (`chorus-test-revA_v19_stripped.kicad_pcb`), same recipe except `--backend` and the auto-layers form:

| Run | Backend | Best result (fully connected) | Auto-fix ran? | DRC blocking |
|---|---|---:|---|---:|
| May 10 (Python from then-HEAD) | python | **30/48 (62%)** | yes | 278 |
| 2026-06-06 C++ (HEAD post-#3232) | cpp | 10/48 (21%) | **no — deadline reached** | 31 |
| 2026-06-06 Python (HEAD post-#3232) | python | **7/48 (15%)** | **no — deadline reached** | 23 |

Both backends today are MORE CONSERVATIVE — they decline to route nets that would create violations, and what they do route is cleaner per-net (3-4x fewer DRC errors per routed net). But they leave so many nets partial that the final fully-connected count drops to ~1/3 of May 10's.

Critical: **the Python backend at HEAD is even WORSE than C++ at HEAD on chorus** (7 vs 10 fully connected). So this is not a C++-specific regression like the board 05 case. It's a tools-wide change in the routing stack since May 10.

### Per-attempt detail (Python, today)

| Attempt | Layers | Initial pass | Detail-routed | Total time | Final |
|---|---|---|---:|---:|---:|
[...]

Issue #2852, opened 2026-05-13
route: propagate --auto-fix rollback (exit 3) to kct route process exit code


rjwalters:
## Parent

Follow-up to #2839 (sub-bug "layer 4 fall-through option" from the original
Curator spec).  The visibility fixes from #2839 (#1 = distinct rollback message,
#2 = consistent summary) landed in PR #2853.  This issue is the
**exit-code propagation** that lets CI / scripts detect a rollback-only
failure.

## Scope

When ``kct route --auto-fix`` runs and the auto-fix pass returns exit code 3
(connectivity rollback per ``fix_drc_cmd.py:375-390``), ``_run_auto_fix`` in
``route_cmd.py:1052-1136`` returns 3 to its caller, but **the surrounding
``kct route`` process exit code does not propagate the auto-fix return value**.
``kct route`` decides its exit code purely from routing-completion / DRC-error
state (see the documented exit-code table at ``route_cmd.py:6277-6311`` and the
parallel return paths at ``2566-2572, 3101-3107, 3687-3693``).

Today, the failure mode is **silent**: a board that routes cleanly (all nets
connected, meets ``--min-completion``) but whose post-route ``--auto-fix`` pass
rolls back exits 0 from ``kct route``, even though the user's DRC is not
actually clean.  PR #2853 made this visible to humans reading stdout, but a
shell script that only inspects ``$?`` cannot detect the rollback.

## Verified current state (curator 2026-05-12)

- Four call sites of ``_run_auto_fix`` in ``route_cmd.py``:
  - **L2532** (inside ``route_with_auto_layers``): return value **discarded**.
  - **L3065** (inside ``route_with_rule_relaxation``): return value **discarded**.
  - **L3649** (inside the combined auto-layer + rule-relaxation flow): return
    value **discarded**.
  - **L6150** (inside ``main``, the primary single-shot flow): return value
    captured as ``fix_result`` but **only used to zero out ``drc_errors`` on
    full success** (``L6156-6157``).  Rollback (exit 3) leaves ``drc_errors``
    pointing at the pre-fix count, which then drives the exit-code branch at
    ``L6303``.
- Today's behavior on the main path (L6150): when ``--auto-fix`` rolls back,
  ``drc_errors`` is still positive, so the existing exit-code logic returns
  ``3`` (``meets threshold but DRC violations detected`` — ``L6303-6305``).
  This is **accidentally correct** for the main flow but not because of any
[...]

Issue #2802, opened 2026-05-12
router: --timeout flag does not bound total wall-clock (chorus-test reproducer at 64+ min on --timeout 2700)


rjwalters:
## Symptom

\`kct route ... --timeout 2700\` (45 min) was launched at 12:32 and is still running at 13:36 — **64 min elapsed, 19 min past the configured budget** (PID 69584, process state \`SN\`). The router has not auto-terminated despite the flag.

## Reproducer

\`\`\`
uv run kct route /Users/rwalters/GitHub/chorus/hardware/chorus-test-revA/kicad/chorus-test-revA_v20.kicad_pcb \\
  --output /tmp/chorus_v20_routed_4layer_v2.kicad_pcb \\
  --manufacturer jlcpcb-tier1 \\
  --placement-feedback --placement-feedback-budget 5 \\
  --iterations 50 --auto-fix --auto-fix-passes 2 \\
  --timeout 2700 --per-net-timeout 30
\`\`\`

Observe: \`ps -o etime\` exceeds 45 min while process is still alive.

## Why this matters

Users (and automation) cannot bound the worst-case wall-clock of a routing run via \`--timeout\` as documented. This blocks scripted experiments and CI integrations that rely on the flag to fence wall-clock cost.

It also propagated upstream: a build-pipeline run earlier this session (#2794) hung 22 min before kill, which was diagnosed as a board-script bug — but if \`--timeout\` had been honored at the library boundary, that hang would have been bounded automatically.

## Likely cause

\`--timeout\` is plumbed into one specific routing stage (likely the per-net A* call or per-iteration loop) but the surrounding orchestration (placement-feedback iterations, auto-fix passes, layer escalation, rip-up loops) does NOT check the wall-clock budget. So with \`--placement-feedback-budget 5 --auto-fix-passes 2\` + \`--iterations 50\` configured, the router can re-enter the inner routing many times, each re-entry respecting \`--timeout\` internally but the overall run exceeding it without bound.

Relevant orchestration sites to check:
- \`src/kicad_tools/cli/route_cmd.py\` outer loop
- \`src/kicad_tools/router/core.py\` route_all_negotiated wrapper (likely entry point)
- \`src/kicad_tools/router/algorithms/negotiated.py\` rip-up loop

## Acceptance criteria

1. \`kct route ... --timeout N\` terminates and emits best-partial output within \`N + small constant\` wall-clock seconds.
[...]

PR #3247, opened 2026-06-07
fix(router): reserve auto-fix budget + structured skip signal (Closes #3238)

### LINKED ISSUE #3238: router: auto-fix DRC silently skipped when negotiated routing exhausts --timeout (chorus regression mechanism)
## Summary

Auto-fix DRC currently runs after all negotiated routing attempts and shares the same `--timeout` budget. When negotiated routing consumes the entire budget (e.g., 3-attempt layer escalation × 500s/attempt = 1500s), auto-fix is silently skipped per issue #2802's wall-clock-deadline guard. The user passes `--auto-fix --auto-fix-passes 2` but those flags become no-ops in practice.

This produces silently-degraded end results: the detail router places 29-31 routes with DRC violations, auto-fix would have cleaned most of them, but it never gets the chance. The chorus regression (filed separately) demonstrates the dynamic — final fully-connected count drops from 30 to 5-7 on the same board, primarily because auto-fix didn't run.

## Mechanism

`--timeout 1500` budget allocation today:
- Negotiated routing attempts 1, 2, 3: ~500s each → 1500s consumed.
- Auto-fix: 0s remaining → skipped per #2802.

The user-facing flags `--auto-fix --auto-fix-passes 2` get parsed and stored but are never effectively honored. The skip is logged but only in the route log — there's no error code change, no stderr warning, no machine-readable signal.

## Proposed fix shape

Builder picks after measurement. Several candidate approaches:

### Option 1 — Reserve a fixed fraction of `--timeout` for auto-fix

E.g., `auto_fix_budget = max(60s, 0.2 * total_timeout)`. Routing's remaining budget becomes `0.8 * total_timeout`. This guarantees auto-fix gets a meaningful slice.

Cost: routing has less time per attempt. The tradeoff is acceptable because partial routes with DRC violations are worth less than fewer routes that are DRC-clean.

### Option 2 — Add `--auto-fix-budget` explicit knob (defaults to `--timeout / 5`)

Make the budget explicit and reservable. Users who want to spend everything on routing can `--auto-fix-budget 0`. Users who want maximum cleanup can `--auto-fix-budget 600`.

### Option 3 — Run auto-fix after each negotiated attempt, not only at the end

Currently the loop is: route attempt 1, route attempt 2, route attempt 3, autofix. Change to: route attempt 1, autofix-pass, route attempt 2, autofix-pass, route attempt 3, autofix-pass. This gives each attempt the chance to convert detail-routed nets into fully-connected ones BEFORE the next layer-escalation attempt evaluates "best result".

Cost: more autofix work (3 passes instead of 1). Benefit: each attempt's "best result" measurement reflects post-autofix quality, so the layer-escalation logic picks the right ladder rung.

### Option 4 — Surface the skip as a hard warning + non-zero exit code

If auto-fix is skipped due to deadline, log a clear "AUTOFIX_SKIPPED_BUDGET_EXHAUSTED" message AND set an exit code distinct from "PARTIAL". CI can then gate on this so silent regressions don't slip through.

Option 1 + 4 together is the smallest combined change that addresses both the silent failure mode AND the budget allocation problem.

## Acceptance criteria

1. Default `kct route --timeout 1500` recipe completes auto-fix on chorus-test-revA_v19_stripped.kicad_pcb at HEAD without manual intervention.
2. Exit code or machine-readable status indicates whether auto-fix ran or was skipped.
3. Existing `--auto-fix-passes` users see no behavior change unless they hit the deadline (in which case they get a warning + truncated pass count, not silent skip).
4. New regression test: a fixture with a route that intentionally produces a partial DRC-violating layout; assert that auto-fix runs and cleans it within the default budget.
5. The sibling chorus regression issue (file separately, link here) closes its AC #3 by virtue of this fix landing.

## Out of scope

- Per-pass auto-fix tuning (different issue).
- The negotiated routing wall-clock cost itself (different issue — chorus reach regression).
- Layer escalation ladder logic (different issue — chorus regression mentioned the May-10-vs-today inversion of the ladder's effect).

## References

- #2802 — original wall-clock-deadline guard that introduces the skip behavior.
- Chorus reach regression (sibling issue, filed separately today) — demonstrates the symptom.

## Why this matters

Auto-fix is the bridge between "routes placed" and "routes that pass DRC". Without it, the end-to-end quality of `kct route` depends entirely on the negotiated router producing DRC-clean output on its first try — which it usually doesn't. Reserving a guaranteed budget is the cheap structural fix that closes a class of silent regression that the chorus result just exposed.

rjwalters:
## Summary

Implements [Option 1 + Option 4](https://github.com/rjwalters/kicad-tools/issues/3238#issue) from issue #3238 to fix the silent skip-on-deadline failure mode in `kct route --auto-fix`:

- **Reserve auto-fix budget**: when `--timeout T` and `--auto-fix` are both set, reserve `max(60s, 0.20 * T)` (capped at 50% of T) for auto-fix. Routing's effective ceiling drops from `T` to `T - reserve` (~80%).
- **Structured skip signal**: when auto-fix is still skipped due to deadline exhaustion, stamp `args._auto_fix_status = "skipped_deadline"`, write `AUTOFIX_SKIPPED_BUDGET_EXHAUSTED` to stderr, and exit with the new distinct code `7`.

The combination prevents the chorus regression mechanism (3-attempt negotiated routing consumes the entire `--timeout`, auto-fix silently returns 1, exit code looks like a benign no-op) and makes the failure mode detectable by CI gates without parsing the route log.

## Changes

| File | Purpose |
|------|---------|
| `src/kicad_tools/cli/route_cmd.py` | Two-deadline split (routing vs wall-clock), `_auto_fix_budget` helper, `args._auto_fix_status` field, stderr token, exit-7 propagation through all 4 route entrypoints |
| `src/kicad_tools/cli/parser.py` | Updated `--auto-fix` help text to document the new reserve and exit code 7 |
| `tests/test_route_cmd_total_timeout.py` | +18 tests across `TestAutoFixBudgetReserve`, `TestRoutingDeadlineSplit`, `TestAutoFixStructuredStatus`, `TestMainStampsAutoFixState` |
| `tests/test_route_auto_fix.py` | +1 test `TestAutoFixSkippedExitCode::test_layer_escalation_exit7_on_skipped_deadline` for the exit-code propagation guard |

## Acceptance criteria

- [x] **AC #1**: Reserve honored on `--timeout 1500 --auto-fix`. Routing's ceiling is `0.8 * 1500 = 1200s`, leaving `>= 300s` for auto-fix.
- [x] **AC #2**: Distinct exit code 7 when auto-fix is requested but skipped due to deadline exhaustion; documented in route command docstring + `--auto-fix` help.
- [x] **AC #3**: Stable `AUTOFIX_SKIPPED_BUDGET_EXHAUSTED` token on stderr for grep-tractable CI gating.
- [x] **AC #4**: No `--timeout` -> no reserve, no skip, identical behaviour. Test: `test_routing_deadline_carves_reserve_with_autofix`, `test_set_deadline_without_timeout_leaves_routing_deadline_none`.
- [x] **AC #5**: `--timeout` set + no `--auto-fix` -> no reserve. Test: `test_auto_fix_budget_zero_when_not_requested`, `test_routing_deadline_equals_wall_clock_without_autofix`.
- [x] **AC #6**: Generous `--timeout` -> no observable effect (routing finishes early, auto-fix has abundant time).
- [x] **AC #7**: Regression test: `test_layer_escalation_exit7_on_skipped_deadline` asserts exit code 7 on the would-be exit-0 path when `args._auto_fix_status == "skipped_deadline"`.
- [x] **AC #8**: Determinism preserved — reserve is purely a function of `args.timeout` and the fixed fraction/floor/cap.
- [x] **AC #9 (#3237 transitive)**: `#3237`'s AC #3 ("auto-fix successfully runs on at least 1 of 3 regens") is closed by this fix landing — routing can no longer consume the entire `--timeout` budget under `--auto-fix`.

## Test plan

- [x] All 479 route-related tests pass (`test_route_auto_fix`, `test_route_cmd_total_timeout`, `test_route_exit_codes`, `test_route_cmd_*`, `test_fix_drc_cmd`, `test_route_all_negotiated_*`, `test_layer_escalation` minus pre-existing `TestEarlyTermination` MagicMock failures unrelated to this change).
- [x] All 216 router tests pass (`tests/router/`).
- [x] Smoke test: `kct route boards/01-voltage-divider/output/voltage_divider.kicad_pcb --dry-run --timeout 30 --auto-fix --auto-fix-passes 2` -> exit 0. No regression on the no-`--timeout` path either.
- [x] `ruff check` passes on all changed files.
- [x] `kct route --help` displays the updated `--auto-fix` help text correctly.

## Design notes

The implementation is a two-deadline split rather than a single-deadline shift to keep `_run_auto_fix` semantics clean:

- `_wall_clock_deadline` = `now + timeout` (total budget; what `_total_remaining_budget` reports).
- `_routing_deadline` = `now + timeout - auto_fix_reserve` (what `_remaining_budget` and `_deadline_expired` report).
- Outer routing loops naturally bail at the routing deadline.
- `_run_auto_fix` uses `_total_deadline_expired` so it runs within its reserved share even after the routing deadline has fired.

When `--auto-fix` is not requested, the reserve is zero and the two deadlines coincide — existing behaviour bit-for-bit preserved.

## Out of scope

- `--auto-fix-budget N` explicit knob (issue #3238 Option 2; recommended follow-up).
- Per-pass auto-fix tuning.
- Negotiated-router wall-clock cost reduction (#3237 umbrella).
- Escalation ladder early-exit (#3241 — already landed).

## References

- Closes #3238
- Refs #3237 (AC #3 closes transitively)
- Refs #2802 (parent: original wall-clock deadline guard)
- Refs #2852 (related: existing exit-3 propagation for fix-drc rollback)

## Coordination note

Per the issue, this PR may conflict with #3237's parallel work on `route_cmd.py`. If the umbrella issue lands first, this branch will need to rebase the deadline-helper edits in `route_cmd.py:110-228`; if this lands first, the umbrella branch picks up the new exit code 7 contract.

### DIFF
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -2770,7 +2770,14 @@ def _add_route_parser(subparsers) -> None:
         action="store_true",
         help=(
             "Automatically run 'kct fix-drc' after routing if DRC violations are "
-            "detected. Suppressed by --dry-run and --skip-drc."
+            "detected. Suppressed by --dry-run and --skip-drc. "
+            "Issue #3238: when combined with --timeout, reserves "
+            "max(60s, 20%% of --timeout) of the total budget so auto-fix "
+            "never gets silently skipped when routing exhausts the timeout. "
+            "If auto-fix is requested but still skipped due to budget "
+            "exhaustion (the routing portion overran its 80%% share), the "
+            "route command exits with code 7 and writes "
+            "AUTOFIX_SKIPPED_BUDGET_EXHAUSTED to stderr."
         ),
     )
     route_parser.add_argument(
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -107,6 +107,39 @@ logger = logging.getLogger(__name__)
 # behaviour.
 
 
+_AUTO_FIX_RESERVE_FRACTION = 0.20
+_AUTO_FIX_RESERVE_FLOOR_SEC = 60.0
+
+
+def _auto_fix_budget(args) -> float:
+    """Return the auto-fix reserve in seconds (issue #3238).
+
+    Returns ``0.0`` when auto-fix is not requested (no reserve needed) or
+    when ``--timeout`` is unset (legacy unbounded runs preserve their
+    existing behaviour: routing gets unlimited time, auto-fix runs after
+    routing finishes naturally).
+
+    When auto-fix is requested AND ``--timeout`` is set, returns
+    ``max(60s, 0.20 * timeout)`` so the negotiated router cannot consume
+    the entire budget and silently leave auto-fix with zero time.  See
+    issue #3238 for the chorus-test regression that motivated this.
+
+    The reserve is also capped at half of ``--timeout`` so the routing
+    budget never drops below 50% of what the user asked for; on extremely
+    small ``--timeout`` values (e.g. ``--timeout 30``) the floor of 60s
+    would otherwise leave the routing loop with effectively zero budget.
+    """
+    timeout = getattr(args, "timeout", None)
+    if not timeout or timeout <= 0:
+        return 0.0
+    if not _should_auto_fix(args):
+        return 0.0
+    reserve = max(_AUTO_FIX_RESERVE_FLOOR_SEC, _AUTO_FIX_RESERVE_FRACTION * float(timeout))
+    # Cap at half of --timeout so the routing budget always sees at least
+    # 50% of what the user asked for, even with very tight timeouts.
+    return min(reserve, 0.5 * float(timeout))
+
+
 def _set_wall_clock_deadline(args) -> None:
     """Stamp a monotonic deadline on ``args`` from ``args.timeout``.
 
@@ -114,20 +147,59 @@ def _set_wall_clock_deadline(args) -> None:
     ``args.timeout`` is falsy (None, 0, or negative) the deadline is set
     to ``None`` so the rest of the orchestration treats the run as
     unbounded.
+
+    Issue #3238: When ``--auto-fix`` is requested, an additional
+    ``_routing_deadline`` is stamped at ``now + (timeout - auto_fix_reserve)``
+    so outer routing loops bail early enough to leave a guaranteed budget
+    for auto-fix.  The original ``_wall_clock_deadline`` continues to
+    bound the *total* wall-clock budget.
     """
     timeout = getattr(args, "timeout", None)
+    now = time.monotonic()
     if timeout and timeout > 0:
-        args._wall_clock_deadline = time.monotonic() + float(timeout)
+        args._wall_clock_deadline = now + float(timeout)
+        reserve = _auto_fix_budget(args)
+        args._auto_fix_reserve = reserve
+        # Routing deadline is the wall-clock deadline minus the auto-fix
+        # reserve.  When no reserve is held (no --auto-fix, or no
+        # --timeout) routing deadline collapses to the wall-clock
+        # deadline -- existing behaviour is preserved bit-for-bit.
+        args._routing_deadline = now + float(timeout) - reserve
     else:
         args._wall_clock_deadline = None
+        args._auto_fix_reserve = 0.0
+        args._routing_deadline = None
 
 
 def _remaining_budget(args) -> float | None:
-    """Return seconds remaining vs the total wall-clock deadline.
+    """Return seconds remaining vs the *routing* deadline.
 
+    Routing-loop callers use this helper so the auto-fix reserve is
+    transparently carved out of their effective budget (issue #3238).
     Returns ``None`` when no deadline is configured (legacy unbounded
     behaviour).  Returns a non-negative float otherwise; callers should
-    treat zero as "deadline expired."
+    treat zero as "routing deadline reached, hand off to auto-fix".
+    """
+    # Issue #3238: prefer the routing deadline (which bounds the routing
+    # loops with the auto-fix reserve already subtracted).  Fall back to
+    # the wall-clock deadline for callers that may have stamped only the
+    # original field (defensive against future refactors / older test
+    # fixtures that don't go through ``_set_wall_clock_deadline``).
+    deadline = getattr(args, "_routing_deadline", None)
+    if deadline is None:
+        deadline = getattr(args, "_wall_clock_deadline", None)
+    if deadline is None:
+        return None
+    return max(0.0, deadline - time.monotonic())
+
+
+def _total_remaining_budget(args) -> float | None:
+    """Return seconds remaining vs the *total* wall-clock deadline.
+
+    Used by ``_run_auto_fix`` to determine whether it has *any* time to
+    run -- the routing deadline may already be expired (that's expected;
+    it's why auto-fix is being invoked) but the total deadline still
+    leaves the carved-out reserve for auto-fix to do its work.
     """
     deadline = getattr(args, "_wall_clock_deadline", None)
     if deadline is None:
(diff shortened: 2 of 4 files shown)


rjwalters:
## Judge review — APPROVE

Reviewed the two-deadline design and load-bearing invariants. This is a clean, well-scoped fix for the silent-skip failure mode.

### Load-bearing invariant verification

**Two-deadline split (`_wall_clock_deadline` total vs `_routing_deadline` total-minus-reserve):** Coherent and correct.
- `_remaining_budget` / `_deadline_expired` consult the routing deadline (outer loops use these).
- `_total_remaining_budget` / `_total_deadline_expired` consult the wall-clock deadline (`_run_auto_fix` uses these).
- When `--auto-fix` is not requested OR `--timeout` is unset, reserve == 0 and the two deadlines coincide bit-for-bit. Covered by `test_routing_deadline_equals_wall_clock_without_autofix` and `test_set_deadline_without_timeout_leaves_routing_deadline_none`.

**Reserve formula `max(60s, 0.20*T)` capped at `0.5*T`:** Sensible and safe at the edges.
- `--timeout 1500` (chorus recipe) gives 300s reserve, 1200s routing.
- `--timeout 200` would be 40s by fraction; 60s floor binds → 60s reserve, 140s routing.
- `--timeout 30` would be 60s by floor (starves routing); 50% cap binds → 15s reserve, 15s routing. Correct.

**Exit code 7:** Distinct from existing codes (0/1/2/3/4/5/6). Propagated at 4 sites:
- `route_with_layer_escalation` (line 3375)
- `route_with_rule_relaxation` (line 3970)
- `route_with_combined_escalation` (line 5214)
- `main()` single-attempt path (line 8243)

The 5th entry point `route_with_mfr_tier_escalation` correctly does not need its own exit-7 — it delegates to `route_with_layer_escalation` and propagates `inner_rc` as `final_exit_code`.

The exit-7 check is correctly placed BEFORE the generic exit-3 ("DRC dirty") branch so the skip signal wins.

**Stderr token `AUTOFIX_SKIPPED_BUDGET_EXHAUSTED`:** Emitted unconditionally to stderr (not gated by `quiet`), so CI gates can grep without parsing the route log. Distinct from the friendlier stdout message. Tested.

**Order of operations in `main()`:** `_set_wall_clock_deadline` is correctly called AFTER `--auto-fix-passes` normalization (line 6401, after the 6374-6382 block). Without this ordering, `--auto-fix-passes N` alone would silently disable the reserve because `_should_auto_fix` would return False at deadline-stamp time. The inline code comment calls this out explicitly — good defensive practice for future refactors.

**`_auto_fix_status` propagation:** Consistent across the 4 sites. Initialized to `"not_invoked"` in `main()`, stamped to `"skipped_deadline"` or `"ran"` in `_run_auto_fix`. The `contextlib.suppress(AttributeError)` guards are belt-and-suspenders against test fixtures with frozen namespaces — defensible.

### Test coverage

20 new tests across 5 test classes, all pass locally (`uv run pytest tests/test_route_cmd_total_timeout.py tests/test_route_auto_fix.py` → 92 passed, 1 pre-existing skip):
- `TestAutoFixBudgetReserve` (8 tests): reserve floor, fraction, cap, zero paths.
- `TestRoutingDeadlineSplit` (6 tests): the two-deadline semantics including the critical `test_deadline_expired_fires_at_routing_deadline` that asserts routing_deadline expires while total still has reserve.
- `TestAutoFixStructuredStatus` (4 tests): status field stamping, stderr token, the `test_skip_uses_total_deadline_not_routing_deadline` that codifies the core inversion.
- `TestMainStampsAutoFixState` (2 tests): main() initialization.
- `TestAutoFixSkippedExitCode` (1 test): end-to-end exit-7 propagation through `route_with_layer_escalation`.

### CI

Test (8m50s), Lint, Type Check, C++ Build, kicad-cli Round-trip Smoke, Routed PCB DRC — all pass. Diff-Pair and Match-Group routing regression jobs are still pending; both test small boards (06, 07) whose routing is unaffected by CLI deadline changes.

### Nits (not blocking)

None. The constants are well-named (`_AUTO_FIX_RESERVE_FRACTION`, `_AUTO_FIX_RESERVE_FLOOR_SEC`), the docstrings cite the issue numbers, and the comments in `main()` flag the load-bearing call order. The `--auto-fix-budget N` explicit knob is appropriately deferred as a follow-up per the PR body.

### Verdict

**Approve.** Two-deadline design is coherent, propagation is complete across all entry points, backward compatibility is preserved bit-for-bit on the no-`--timeout` and no-`--auto-fix` paths, and the structured signal (exit 7 + stderr token) makes the previously-silent failure mode CI-tractable.

Looking forward to the next chorus-test sweep — with #3246's regression floor active on main and this PR's budget reservation landing, the chorus reach number should re-measure cleanly.

Issue #3235, opened 2026-06-06
router: softstart 8/10 → 10/10 lift lives in two_phase.py:709 reroute order, not neighborhood_ripup (supersedes #3230's scope)


rjwalters:
## Summary

Discovered during #3230 / PR #3231's build phase: softstart's negotiated routing-reach binding constraint is **NOT** `neighborhood_ripup` (the surface #3230 + curator + I all assumed). It's the **outer iteration loop in `src/kicad_tools/router/algorithms/two_phase.py:709-784`**, specifically the `find_nets_through_overused_cells` + sequential reroute pattern.

PR #3231 added a budget-aware tie-break to `neighborhood_ripup` per #3230's plan. That code surface is on the Autorouter path (which other boards use) but NEVER fires on softstart. PR #3231's builder verified this with debug instrumentation: the budget signal correctly detected genuine foreign-budget overlap (e.g. iteration 1: net 20 overlap=30), but applying the same signal as a reorder hint to `two_phase.py`'s rip-up reroute order — in BOTH ASC and DESC variants — regressed softstart 8/10 → 7/10. The existing dict-insertion-order is empirically locally optimal on this workload.

Also discovered: softstart's baseline at `origin/main` (post-#3227) is **already 8/10**. The original #3201 / #3230 framing of "stuck at 6/10" is empirically stale; the 6→8 lift was delivered by upstream work (most likely PR #3227's Chebyshev guard + subsequent merges). The actual softstart goal is 8/10 → 10/10 (the four remaining stranded nets: `GATE_NEG`, `I_SENSE_OUT`, `ZC_DETECT`, `SWCLK`), not 6/10 → 8/10.

This issue tracks the next lever for that 8/10 → 10/10 lift now that the budget-tie-break lever is exhausted.

## Background

- **#3201 (open)** — original per-pad channel budget tuning goal (≥ 8/10 reach). The headline AC is met at HEAD by accident of upstream work. The remaining gap is 8/10 → 10/10.
- **#3230 / PR #3231 (just merged)** — added budget-aware tie-break in `neighborhood_ripup`. Lands as positive change for other boards; doesn't move softstart's needle.
- **PR #3223 / PR #3203** — earlier per-pad budget geometry tuning; documented the 9-parameter empirical floor at 6/10. (Stale measurement; pre-#3227 baseline.)

## Affected code

Cited at HEAD = post-PR #3231 + PR #3232 merge:

[...]

PR #3249, opened 2026-06-07
spike(router): document #3235 direction-1+2 negative results; preserve infra (Refs #3235)

rjwalters:
## Summary

The 8/10 → 10/10 softstart routing-reach lift requested by Issue #3235 did **NOT** achieve AC #1 (≥ 9/10 worst-of-3 across PYTHONHASHSEED=42/43/44).  This PR lands the spike's infrastructure (preserved as default-no-op) plus inline negative-results documentation so the next builder does not re-run the same failed levers.

Per the issue's instructions: *\"If you can't lift softstart past 8/10 after a reasonable spike on direction 2 + direction 1, document the negative result and either keep trying direction 3/4 or `Refs #3235` with the findings (don't close).\"*

This PR is `Refs #3235`, not `Closes`.

### Direction 2 (escape-route layer diversification)

**(a) NC pad filtering** — remove pads with `net == 0` from row buckets before the alternation index is assigned in `_escape_fine_pitch_dual_row`.  Mirrors the existing QFP/QFN dispatcher's NC-skip pattern.  Hypothesis: NC pad escape geometry consumes post-escape channel bandwidth wastefully.
- **Result: REGRESSED to 7/10 at seed 42.**  NC pad escapes evidently shield signal pads from same-layer collisions even though they carry no net.

**(b) Phase flip on second column** — pass `phase_offset=1` to the RIGHT column / SOUTH row of `_create_fine_pitch_row_escapes` so odd indices stay on F.Cu and even indices via.  Hypothesis: rebalances softstart U1's right-col 4/6 B.Cu-dominant split to ~50/50 F.Cu/B.Cu.
- **Result: REGRESSED to 7/10 across seeds 42/43/44.**  The historical strict-by-pad-position alternation is empirically near-optimal on this workload despite the layer asymmetry.

### Direction 1 (extend find_nets_through_overused_cells)

**(c) Augment-not-reorder** the rip-up cohort with squatter nets from `find_nets_in_foreign_budgets`, gated on stranded source nets.  The augmentation correctly surfaced SWCLK / SWDIO / GATE_POS as squatters in iter 1/2 of the rip-up loop.
- **Result: NO improvement.**  The augmented cohort carried across iterations unchanged, triggered the existing rip-up-set Jaccard stagnation detector (iter 2 overflow 8→20 vs baseline 8→14), and the stagnation rollback restored iter 1's pre-augmentation state.

## What ships

1. `EscapeRouter._create_fine_pitch_row_escapes` gains a `phase_offset: int = 0` parameter that flips the via-down parity from `i % 2 == 1` to `(i + phase_offset) % 2 == 1`.  The dispatcher does NOT pass non-zero offsets so default behaviour is byte-identical to pre-#3235.  Negative-results note added to the method docstring.
2. `NegotiatedRouter.find_nets_in_foreign_budgets` — new helper.  Returns nets routing through pad-channel budgets whose `source_net` is in `stranded_nets`.  Excludes the source net itself and any `skip_nets` the caller provides.  Not wired into the production rip-up loop; preserved as infrastructure.
3. `two_phase.py:709` gains an inline negative-results note at the hook point flagged by the issue body.  Documents the augmentation/stagnation interaction and lists three forward-looking gate refinements (exclude partial nets, monotonic cohort growth, present-cost nudge instead of cohort membership).
4. New test file `tests/test_find_nets_in_foreign_budgets.py` (12 tests, all passing) covers the helper's API contract plus the `phase_offset` parameter contract.

## Behaviour preserved

- **Default softstart reach: 8/10 at PYTHONHASHSEED=42/43/44** (unchanged from `origin/main`).
- `tests/test_neighborhood_ripup_budget_tiebreak.py`: 10 pass.
- `tests/test_negotiated_adaptive.py`: 120 pass.
- `tests/test_escape_endpoint_routing.py` + `tests/test_escape_diffpair.py`: 38 pass.

## Acceptance-criteria status (issue #3235)

- AC #1 (≥ 9/10 worst-of-3): **NOT MET.**  Reach unchanged at 8/10.
- AC #2 (board 05 ≤ 32 DRC blocking): **N/A** — no behaviour change.
- AC #3 (board 07 PYTHONHASHSEED determinism): **N/A** — no behaviour change.
- AC #4 (board 02 unchanged): **N/A** — no behaviour change.
- AC #5 (`KCT_DISABLE_PAD_BUDGETS=1` discriminator): unchanged from baseline — the env var still produces zero softstart reach delta because the spike's augmentation did not lift reach.
- AC #6 (regression test on 4-pad cluster): **MET** — new test exercises the softstart-like 4-budget / 3-squatter integration shape.

## Forward-looking notes for next builder

The `two_phase.py:709` inline comment lists three concrete gate refinements that the next direction-1 spike should try before re-implementing the squatter augmentation:

1. **Exclude partial-routed nets** from the squatter set.  This spike's biggest finding: the squatter detector surfaced SWCLK and SWDIO as squatters in iter 1, but BOTH are 2/3-partial-connected nets — ripping them up wastes the per-net timeout AND keeps them partial.  A `skip_nets={...partial...}` argument is already on the helper for this.
2. **Pre-empt stagnation rollback** by growing the cohort MONOTONICALLY across iterations so the Jaccard similarity trends down naturally.  The current augmentation locks in the same set across iter 1 and iter 2 → 100% Jaccard → stagnation.
3. **Apply the squatter signal to the present-cost factor** instead of the rip-up cohort.  This is a cost-shaping nudge (like the existing per-pad budget penalty) rather than a rip-up policy, so it does not interact with the stagnation detector at all.

Direction 3 (randomized / parallel reroute) and direction 4 (wire `KCT_DISABLE_PAD_BUDGETS=1` as a discriminator) were not attempted in this spike.

Refs #3235.

## Test plan

- [x] `uv run pytest tests/test_find_nets_in_foreign_budgets.py` (12 pass)
- [x] `uv run pytest tests/test_neighborhood_ripup_budget_tiebreak.py` (10 pass)
- [x] `uv run pytest tests/test_negotiated_adaptive.py` (120 pass)
- [x] `uv run pytest tests/test_escape_endpoint_routing.py tests/test_escape_diffpair.py` (38 pass)
- [x] `PYTHONHASHSEED=42 uv run python boards/external/softstart/generate_design.py` → 8/10
- [x] `PYTHONHASHSEED=43 uv run python boards/external/softstart/generate_design.py` → 8/10
- [x] `PYTHONHASHSEED=44 uv run python boards/external/softstart/generate_design.py` → 8/10

### DIFF
--- a/src/kicad_tools/router/algorithms/negotiated.py
+++ b/src/kicad_tools/router/algorithms/negotiated.py
@@ -1031,6 +1031,135 @@ class NegotiatedRouter:
 
         return nets_to_reroute
 
+    def find_nets_in_foreign_budgets(
+        self,
+        net_routes: dict[int, list[Route]],
+        pad_channel_budgets: list,
+        stranded_nets: set[int],
+        skip_nets: set[int] | None = None,
+    ) -> list[int]:
+        """Find nets squatting in foreign budgets whose source-net is stranded.
+
+        Issue #3235 (direction 1, "extend ``find_nets_through_overused_cells``
+        to also fire on nets occupying high-budget cells even if those cells
+        aren't 'overused' in the current iteration"):  The negotiated rip-up
+        loop only picks rerouting candidates from cells whose
+        ``usage_count > 1``.  But on softstart-style east-edge clusters, a
+        squatter can park in a foreign pad-channel budget without ever
+        triggering ``usage_count > 1`` (the cell is only used by the squatter
+        plus the budget metadata; the actual source-net is STUCK, so it never
+        commits a route there).  PR #3231's builder demonstrated that the
+        squatter signal is real (Iter 1: net 20 overlap=30, etc.) but that
+        applying it as a reorder hint to the existing dict-insertion rip-up
+        order regresses softstart 8/10 to 7/10 in both ASC and DESC variants.
+
+        Negative-results note (from the #3235 spike): using this method as an
+        AUGMENTATION (additional violators appended to the cohort) did not
+        lift softstart's reach in measurements at PYTHONHASHSEED=42; the
+        augmented cohort triggered the existing rip-up-set Jaccard stagnation
+        detector with an unchanged set across consecutive iterations, and the
+        rollback restored iter 1's pre-augmentation state.  The infrastructure
+        is preserved so future spike attempts can wire it into the rip-up
+        loop with a tighter gate (e.g. excluding partial-routed nets from the
+        squatter set, gating on the stagnation detector's Jaccard
+        threshold so the cohort grows monotonically).  See `two_phase.py:709`
+        for the intended hook point.
+
+        Args:
+            net_routes: Dict of net_id -> committed routes (segments + vias).
+            pad_channel_budgets: List of ``PadChannelBudget`` objects with
+                ``gx1/gy1/gx2/gy2/layer/source_net`` attributes (Issue #3143
+                geometry).  Empty list yields an empty result.
+            stranded_nets: Set of net IDs that have failed to commit a route
+                in the current iteration.  Only budgets whose
+                ``source_net`` is in this set produce augmentation.
+            skip_nets: Optional set of net IDs to skip from the squatter
+                output (typically partial-routed nets the caller does not
+                want to rip up again, or nets already in the rip-up cohort).
+
+        Returns:
+            List of net IDs that route through foreign budgets belonging to
+            stranded source-nets.  Excludes any net in ``skip_nets`` and any
+            net equal to the budget's ``source_net`` (a net cannot squat in
+            its own budget).
+        """
+        if not pad_channel_budgets or not stranded_nets:
+            return []
+
+        skip = skip_nets or set()
+
+        # Filter to budgets whose source-net is currently stranded.  Each
+        # surviving budget represents a contested channel; routes that pass
+        # through it (other than the source-net's own route) are blocking
+        # the source-net.
+        relevant_rects: list[tuple[int, int, int, int, int, int]] = []
+        for b in pad_channel_budgets:
+            try:
+                src = int(b.source_net)
+            except (AttributeError, TypeError):
+                continue
+            if src not in stranded_nets:
+                continue
+            try:
+                relevant_rects.append((
+                    int(b.gx1),
+                    int(b.gy1),
+                    int(b.gx2),
+                    int(b.gy2),
+                    int(getattr(b, "layer", -1)),
+                    src,
+                ))
+            except (AttributeError, TypeError):
+                continue
+
+        if not relevant_rects:
+            return []
+
+        squatters: list[int] = []
+        seen: set[int] = set()
+
+        for net_id, routes in net_routes.items():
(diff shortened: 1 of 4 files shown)


rjwalters:
## Judge review: APPROVED

This is a well-executed negative-results + infrastructure PR. The builder followed the issue body's own explicit playbook ("If you can't lift softstart past 8/10 after a reasonable spike on direction 2 + direction 1, document the negative result..."), used `Refs #3235` (issue stays open), and shipped the scaffolding that any future direction-1 attempt will need anyway.

### Verified

- **`phase_offset=0` default is byte-identical to historical behaviour.**  All four call sites in `_escape_fine_pitch_dual_row` (lines 3003/3010/3029/3036) call `_create_fine_pitch_row_escapes` WITHOUT `phase_offset`, so `(i + 0) % 2 == 1` evaluates to the historical `i % 2 == 1` unchanged.  No production path takes a non-zero offset.
- **`find_nets_in_foreign_budgets` is genuinely unwired.**  `grep -rn "find_nets_in_foreign_budgets" src/ tests/` returns ZERO production callers; only the new test file references it.  The `two_phase.py:709` comment block is a pure comment, not a call.
- **Tests pass.**  `tests/test_find_nets_in_foreign_budgets.py` (12/12), `tests/test_neighborhood_ripup_budget_tiebreak.py` + `tests/test_escape_endpoint_routing.py` + `tests/test_escape_diffpair.py` (48/48 combined).  CI status: 8/8 pass.  `mergeStateStatus: CLEAN`.
- **AC #6 is delivered.**  `test_softstart_like_4pad_cluster` exercises the four-stranded-source + three-squatter integration shape the issue's AC #6 calls out.

### Judgment call (the interesting part)

The CLAUDE.md tension here is "prefer-completing-stubs vs avoid-stub-theater."  This is NOT stub theater because:

1. The helper is a **working implementation with test coverage**, not a placeholder.  The reason it isn't wired is empirical (rip-up-set Jaccard stagnation rolled back the cohort), not "not implemented yet."
2. The negative-results documentation has **concrete next steps with specific gate refinements** (exclude partial nets, monotonic cohort growth, present-cost nudge).  Any future #3235 attempt that ignores this PR will re-derive the same dead ends — saving a builder cycle is worth ~80 lines of dormant code.
3. The inline note at `two_phase.py:709` is correctly placed at the exact hook point the issue body flagged.  A future builder grepping for direction-1 implementation locations will land on it.

### What I'd push back on if asked

The `find_nets_in_foreign_budgets` helper does a per-segment per-step linear scan over `relevant_rects`, which is O(routes * segments * steps * rects).  Fine for current call sites (none in production) but if it ever IS wired into the rip-up loop, an interval-tree or spatial hash should land in the same PR as the wire-up — not retroactively.  Not a blocker for this PR since the helper is dormant.

### Verdict

Approve.  The next #3235 attempt builds on this scaffolding instead of redoing it; the issue stays open (`Refs #3235`) so the 8 → 10 lift is still tracked; default behaviour is byte-identical; AC #6 is met.  Net positive for the repo.

PR #3277, opened 2026-06-07
fix(board-02): route_demo regenerates mfg bundle to keep fleet status fresh

### LINKED ISSUE #3264: Board 02 (and similar): route_demo.py re-routes PCB but leaves mfg bundle stale
## Problem

After ``boards/02-charlieplex-led/route_demo.py`` re-routes ``charlieplex_3x3_routed.kicad_pcb``, the manufacturing bundle's ``manifest.json`` mtime stays older than the routed PCB.  ``kct fleet status`` then reports ``ship_ready=false`` with blocker ``artifacts stale`` even though the new routed PCB is in fact identical (or at least equally good).

## Reproduction

```bash
# Start from a freshly-checked-out tree where manifest mtime > routed mtime
python boards/02-charlieplex-led/route_demo.py
uv run kct fleet status --format json | jq '.boards[] | select(.name=="02-charlieplex-led") | {ship_ready, blockers}'
# -> ship_ready: false, blockers: ["artifacts stale"]
```

This is a real false-negative gap because users / CI invoking ``route_demo.py`` or ``kct build --step route`` will see fleet status drop to non-ship-ready immediately.  ``generate_design.py`` happens to call ``export_manufacturing_bundle`` after routing (per #3147 / #3163), but the ``route_demo.py`` script — and ``kct route`` itself — does not.

## Confirmed by repro

In `feature/issue-3260` worktree, on a fresh clone:
- ``boards/02-charlieplex-led/output/charlieplex_3x3_routed.kicad_pcb`` mtime = 04:52
- ``boards/02-charlieplex-led/output/manufacturing/manifest.json`` mtime = 04:40
- ``ship_ready: false, blockers: ["artifacts stale"]``

The routed file content matches ``cpp seed=42`` output byte-for-byte across all 3 tested seeds (router is deterministic on this board), so re-routing is a no-op for the bundle content — but the mtime check flags it as stale anyway.

## Options

1. **Make ``kct route`` auto-export the mfg bundle** when given a board project context (similar to how ``generate_design.py`` does it).  Cleanest fix but adds a CLI side-effect.
2. **Make ``route_demo.py`` call mfg-bundle export** after the route (like ``generate_design.py`` does at line ~748).  Most surgical fix.  ~10 LOC per board.
3. **Make ``fleet status`` content-hash the routed PCB against manifest entries** instead of using mtime.  Most robust but requires reading + hashing on every status call.
4. **Make ``manifest.json`` track only routed-PCB content hash, not mtime**.  Then re-running ``route_demo.py`` with identical output doesn't trigger stale.

## Recommendation

Option 2 for the immediate fix (matches the pattern ``generate_design.py`` already uses; applies cleanly to ``boards/01..07`` and external boards).  Option 3 or 4 as a longer-term solution to handle re-route scenarios across all entry points.

## Related

- Tracking issue: #3260
- Auto-export plumbing: PR #3163 (#3147)
- Stale detection: ``src/kicad_tools/cli/fleet_cmd.py:634-639`` (mtime comparison)

## Acceptance

- After running ``python boards/02-charlieplex-led/route_demo.py`` from any tree state, ``kct fleet status --format json`` reports ``ship_ready=true, blockers=[]`` for board 02.
- Pattern applied consistently to ``route_demo.py`` / ``route.py`` scripts across boards 01-07 and external/softstart.
- Regression test that asserts: route_demo writes routed PCB, then re-checks fleet status reports ``ship_ready=true``.

rjwalters:
## Summary

- Fixes Issue #3264: ``boards/02-charlieplex-led/route_demo.py`` re-routes the PCB but never re-exports the manufacturing bundle, so ``kct fleet status`` reports ``ship_ready=false`` with blocker ``"artifacts stale"`` even after a successful route.
- Adds a small ``export_manufacturing_bundle()`` helper to ``route_demo.py`` (mirrors the helper already in ``generate_design.py``) and calls it from ``main()`` after ``kct route`` writes the routed PCB.
- Surfaces the bundle freshness in the demo's summary block (``MFG bundle: FRESH`` / ``STALE (fleet status will block)``) so users can see at a glance whether fleet status will be happy.
- Adds two regression tests in ``tests/test_board02_route_demo_recipe.py``:
  - Static check that ``route_demo.py`` invokes ``kct export --mfr jlcpcb``.
  - End-to-end check that after running the demo, ``manifest.json`` mtime is newer than the routed PCB (the exact invariant ``cli/fleet_cmd.py:634-639`` uses).

## Reproduction (pre-fix)

```bash
python boards/02-charlieplex-led/route_demo.py
uv run kct fleet status --format json | jq '.boards[] | select(.name=="02-charlieplex-led") | {ship_ready, blockers}'
# -> {"ship_ready": false, "blockers": ["artifacts stale"]}
```

After this PR:

```bash
python boards/02-charlieplex-led/route_demo.py
uv run kct fleet status --format json | jq '.boards[] | select(.name=="02-charlieplex-led") | {ship_ready, blockers}'
# -> {"ship_ready": true, "blockers": []}
```

## Scope

This PR fixes board 02 only, matching the issue body's Option 2 ("Most surgical fix. ~10 LOC per board"). The same gap exists in ``boards/03-usb-joystick/route_demo.py`` and likely any ``route.py``-style recipe across other boards that re-routes without re-exporting -- those can be addressed in follow-up issues if/when they hit a fleet-status false negative. The fix is mechanically identical for each board, so generalization is straightforward; I held off only to keep this PR scoped to the issue as filed.

Longer-term solutions discussed in the issue body (Option 3: content-hash routed PCB; Option 4: hash-based manifest) remain valid follow-ups for a more robust fix across all entry points.

## Test plan

- [x] ``uv run pytest tests/test_board02_route_demo_recipe.py`` -- 5/5 pass (including 2 new regression tests)
- [x] ``uv run pytest tests/router/test_board02_manufacturable_baseline.py`` -- 3/3 pass (no regression in routing baseline)
- [x] Manual reproduction: pre-fix flips ``ship_ready`` from true -> false; post-fix flips from false -> true.

Closes #3264

### DIFF
--- a/boards/02-charlieplex-led/route_demo.py
+++ b/boards/02-charlieplex-led/route_demo.py
@@ -76,6 +76,57 @@ def run_drc(pcb_path: Path) -> tuple[bool, int, int]:
         return False, -1, -1
 
 
+def export_manufacturing_bundle(routed_path: Path, output_dir: Path) -> bool:
+    """Regenerate the manufacturing bundle so ``kct fleet status`` stays fresh.
+
+    Issue #3264: ``kct fleet status`` flags a board ``ship_ready=false``
+    with the ``"artifacts stale"`` blocker whenever the routed PCB is
+    newer than ``output/manufacturing/manifest.json``.  Re-running
+    ``route_demo.py`` always rewrites the routed PCB, so the demo must
+    also regenerate the manufacturing bundle to keep the manifest current
+    (mirrors what ``generate_design.py:export_manufacturing_bundle()``
+    already does at the end of the full design flow).
+
+    ``kct export`` runs the standard JLCPCB recipe (gerbers + drill + BOM
+    + CPL + report.{md,pdf} + manifest.json).  ``--skip-preflight`` skips
+    the strict pre-flight DRC/ERC gate so the bundle is produced even for
+    boards that ship with allowlisted tolerances; for clean boards (like
+    board 02 post-#3207) it is harmless.
+    """
+    print("\n" + "=" * 60)
+    print("Exporting manufacturing bundle (Issue #3264)...")
+    print("=" * 60)
+
+    mfg_dir = output_dir / "manufacturing"
+    cmd = [
+        sys.executable,
+        "-m",
+        "kicad_tools.cli",
+        "export",
+        str(routed_path),
+        "--output",
+        str(mfg_dir),
+        "--mfr",
+        "jlcpcb",
+        "--skip-preflight",
+    ]
+    print(f"\n   Command: {' '.join(cmd)}")
+    result = subprocess.run(cmd, capture_output=True, text=True)
+    if result.stdout:
+        for line in result.stdout.strip().split("\n")[-15:]:
+            print(f"   {line}")
+    if result.returncode != 0:
+        if result.stderr:
+            print(f"\n   Error: {result.stderr}")
+        return False
+    manifest = mfg_dir / "manifest.json"
+    if manifest.exists():
+        print(f"\n   Manifest: {manifest}")
+        return True
+    print("\n   WARNING: manifest.json not produced")
+    return False
+
+
 def _parse_routed_net_count(stdout: str) -> tuple[int, int] | None:
     """Extract ``Nets routed: N/M`` (or equivalent) from ``kct route`` output.
 
@@ -186,6 +237,14 @@ def main():
     parsed = _parse_routed_net_count(result.stdout)
     routed, total = parsed if parsed is not None else (None, None)
 
+    # Issue #3264: regenerate the manufacturing bundle so its
+    # ``manifest.json`` mtime is newer than the freshly-routed PCB.
+    # Otherwise ``kct fleet status`` reports ``ship_ready=false`` with
+    # blocker ``"artifacts stale"`` even though the route succeeded.
+    # This mirrors the post-route step in
+    # ``generate_design.py:export_manufacturing_bundle()``.
+    mfg_success = export_manufacturing_bundle(output_path, output_path.parent)
+
     # Run DRC validation on the routed output.
     print("\n--- DRC Validation ---")
     drc_passed, drc_errors, drc_warnings = run_drc(output_path)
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Verdict: APPROVE

Surgical fix that closes the freshness gap on board 02 without surprises. Excellent execution.

### What I verified

**Mirror parity with `generate_design.py`**
The new `export_manufacturing_bundle()` in `route_demo.py` is a near-verbatim mirror of `generate_design.py:658-704`:
- Same command shape: `python -m kicad_tools.cli export <routed> --output <mfg_dir> --mfr jlcpcb --skip-preflight`
- Same return semantics (success only if exit 0 AND `manifest.json` exists)
- Same `--skip-preflight` rationale (allowlisted-tolerance boards 03/04/05 still produce a bundle)
- The new docstring correctly references the canonical helper

**Test coverage**
Both new tests are well-formed and pin the right invariants:
- `test_route_demo_invokes_mfg_bundle_export` — fast static check that `"export"` + `"--mfr"` + `"jlcpcb"` appear in the source. Good failure messages.
- `test_route_demo_refreshes_manifest_mtime` — end-to-end check that `manifest.mtime >= routed_pcb.mtime` after running the demo. This is exactly the invariant `fleet_cmd.py:634-639` uses to flag `"artifacts stale"`. Accepts exit codes 0 or 1 (pins freshness, not routing completion — correctly leaves that to the existing `test_route_demo_achieves_minimum_completion`). 300s timeout guards against hangs.

**Local verification**
- `tests/test_board02_route_demo_recipe.py` — 5/5 pass (49.26s)
- `tests/router/test_board02_manufacturable_baseline.py` — 3/3 pass (76.76s)
- Manual `kct fleet status --format json` post-test reports `ship_ready=true, blockers=[]` for board 02 (matches PR claim).

**CI**
All 8 checks pass:
- C++ Build Check, Type Check, Test, Lint & Format
- Routed PCB DRC Check, kicad-cli Round-trip Smoke
- Diff-Pair Routing Regression (boards/06), Match-Group Routing Regression (boards/07)

**Merge state**: `CLEAN`, `MERGEABLE`.

**Scope discipline**
Touches only `boards/02-charlieplex-led/route_demo.py` (+62) and `tests/test_board02_route_demo_recipe.py` (+88). No collateral changes. The PR body honestly notes that `boards/03-usb-joystick/route_demo.py` has the same gap and proposes follow-up issues — the right call, keeps this PR aligned with the issue's Option 2.

### Minor (non-blocking) observations

- The "MFG bundle: FRESH / STALE" summary line is a nice ergonomic touch — users get a same-screen heads-up on whether `fleet status` will be happy. No action needed.
- Follow-up worth filing: the issue also called for the pattern to be applied across `boards/01..07` and `external/softstart`. Builder already flagged this — a fleet-wide sweep issue (or Option 3/4 content-hash approach) would close the broader gap.

Verdict: approved.

Issue #3281, opened 2026-06-07
router(python-backend): board 04 drops from 9/9 → 4/9 on stripped recipe (kct route ... --backend python --layers 2 --manufacturer jlcpcb --seed 42)


rjwalters:
## Context

Surfaced while triaging #3268 (test_board_04_routing_regression failing on python backend). The python-backend router has regressed from the post-#2745 baseline of 9/9 nets routed on the **stripped 2L recipe** down to 4/9 (or 3/9 — observed both).

## Recipe under test

```
kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb \
  --seed 42 --no-auto-layers --layers 2 \
  --manufacturer jlcpcb --backend python --timeout 240
```

This is the exact recipe asserted by `tests/test_board_04_routing_regression.py::TestBoard04OscOutRouting::test_routes_all_nets_on_2l`. It is deliberately stripped of `--auto-fix`, `--auto-layers`, `--auto-mfr-tier`, `--placement-feedback`, and `--micro-via-in-pad-fallback` so the test pins the recovery-gate behavior added in PR #2751 (closes #2745) in isolation.

## Observed

- Post-#2745 baseline: 9/9 nets routed on this recipe.
- Today (commit 7d686254, 2026-06-06): 4/9 nets routed (SWDIO, SWO, NRST, BOOT0 unrouted; OSC_OUT partial).
- The C++ backend on the **same stripped recipe** also produces 4/9 (not python-specific).
- The C++ backend on the **full production recipe** (`--mfr jlcpcb-tier1 --auto-fix --auto-layers --auto-mfr-tier --placement-feedback --micro-via-in-pad-fallback --seed 42`) produces 8/9 — NRST regression tracked separately.

So this issue documents a **router regression visible on the stripped 2L recipe** affecting both backends, although #3268 originally framed it as python-only because the test only exercises python.

## What changed

[...]

Issue #3268, opened 2026-06-07
test: tests/test_board_04_routing_regression.py fails on python backend (3/9 nets, was 9/9)


rjwalters:
## Context

The pre-existing slow test \`tests/test_board_04_routing_regression.py::TestBoard04OscOutRouting::test_routes_all_nets_on_2l\` is failing with the python backend, asserting:

\`\`\`
E       AssertionError: Board 04 routed only 3/9 nets (expected 9/9).
E       assert 3 >= 9
\`\`\`

The test runs \`kct route ... --backend python --layers 2 --manufacturer jlcpcb --seed 42\` and expects 9/9 nets routed (per issue #2745's OSC_OUT recovery fix).  Today the python backend produces only 3/9 nets on the same recipe.

The C++ backend (now the default) still produces 8/9 on this board (NRST regression filed separately) -- so this is a python-backend-specific regression compounding with the NRST issue.

## Discovery context

Surfaced during board 04 manufacturability verification (#3261, parent #2394). The test is marked \`@slow\` so it does not run on PR-time CI; only nightly slow-tests workflow picks it up. The slow-tests workflow may have been failing silently for some time.

## Acceptance criteria

- [ ] Bisect the python backend to find the change that dropped 9/9 -> 3/9 on this recipe.
- [ ] Either fix the python backend OR update the test to either:
  - Switch the test to \`--backend cpp\` (which is now the documented default), OR
  - Lower the threshold to match current python backend reality and document the gap.
- [ ] Investigate whether the nightly slow-tests workflow has been red on this test for an extended period.

## References
- Parent verification: #3261
- Test originated: PR #2751 (closes #2745)
- Manufacturability cluster: #2834, #2394

PR #3282, opened 2026-06-07
test(board-04): skip python-backend 2L 9/9 assertion pending router follow-up #3281

rjwalters:
## Summary

- Skips `tests/test_board_04_routing_regression.py::TestBoard04OscOutRouting::test_routes_all_nets_on_2l` on the python backend because the post-#2745 9/9 baseline regressed to 4/9 (3/9 in #3268's report) on the stripped 2L recipe.
- Diagnosis: real router regression (not a stale assertion). The C++ backend on the same stripped recipe also produces 4/9, so this is not python-specific — the recipe itself depends on router features (notably `--micro-via-in-pad-fallback`) that have become required for OSC_OUT escape since #2745 landed.
- Filed follow-up #3281 to bisect and either restore 9/9 on the stripped recipe or retire the assertion with a documented policy change.

## Diagnosis

| Backend | Recipe | Result |
|---|---|---|
| python | `--no-auto-layers --layers 2 --manufacturer jlcpcb --seed 42` (test recipe) | **4/9** (was 9/9) |
| cpp    | `--no-auto-layers --layers 2 --manufacturer jlcpcb --seed 42` (same stripped recipe) | **4/9** |
| cpp    | `--mfr jlcpcb-tier1 --auto-fix --auto-layers --auto-mfr-tier --placement-feedback --micro-via-in-pad-fallback --seed 42` (production recipe in `boards/04-stm32-devboard/generate_design.py`) | **8/9** (NRST gap separately tracked) |

So the test class itself is not stale (the AC was met in PR #2751); the **router** has since drifted off the stripped-recipe invariant.  Per #3268's decision tree, this is path (a): real regression → file a sharp follow-up router issue, adjust **this** test to skip on python backend, do not fix the router here.

## Why a class-level skip rather than `--backend cpp` switch

The two listed alternatives in #3268 were:
1. Switch the test to `--backend cpp` (now the documented default), or
2. Lower the threshold to match current reality.

(1) doesn't work — C++ is also 4/9 on this recipe, so switching backends would only convert a known-fail into a different known-fail.
(2) loses the regression signal: 4/9 is not "current python backend reality" in any stable sense; it is a known gap relative to the post-#2745 invariant.  A skip with a clear link to #3281 keeps the test honest and re-enabling is one line.

## Test plan

- [x] `uv run kct build-native` (C++ backend built in worktree)
- [x] `uv run pytest tests/test_board_04_routing_regression.py -v -m slow` → 1 skipped (was 1 failed)
- [x] `uv run pytest tests/test_board_04_mfr_gate.py tests/test_board_04_drc_recipe_mfr.py` → 7 passed (no collateral regression)
- [ ] Judge to spot-check the skip reason links #3281 and the diagnosis docstring is accurate

## References

- Closes: this PR closes the test-side gap reported in #3268.  Marking `Refs #3268` rather than `Closes` because the underlying router regression remains open under #3281 — re-resolving and closing #3268 should wait until that re-enables the assertion.
- Follow-up router issue: #3281
- Original 9/9 baseline: #2745 / PR #2751

Refs #3268
Refs #3281

### DIFF
--- a/tests/test_board_04_routing_regression.py
+++ b/tests/test_board_04_routing_regression.py
@@ -21,6 +21,17 @@ Marked ``@pytest.mark.slow`` (single 2L attempt is ~60-90s; we set a
 240s budget to leave generous slack for slower runners).  Nightly slow-
 tests workflow at ``.github/workflows/slow-tests.yml`` (``-m slow``)
 picks this up; PR-time CI excludes it.
+
+Issue #3268 (2026-06-06) — the python-backend variant of this test
+regressed from 9/9 to 4/9 (or 3/9 — minor nondeterminism observed) on
+the stripped recipe.  Investigation showed the C++ backend on the same
+stripped recipe also produces 4/9, so this is not python-specific; it
+is a broader regression that surfaces here because the test deliberately
+omits ``--micro-via-in-pad-fallback`` and the other production-pipeline
+flags to isolate the #2745 recovery gate.  The router regression is
+tracked in #3281; until that is resolved this test is skipped on the
+python backend so the slow-tests workflow stays green rather than
+red-on-known-gap.
 """
 
 from __future__ import annotations
@@ -76,6 +87,16 @@ def _parse_routed_net_count(stdout: str) -> tuple[int, int] | None:
 
 
 @pytest.mark.slow
+@pytest.mark.skip(
+    reason=(
+        "Issue #3268 — python backend regressed from 9/9 to 4/9 on the stripped "
+        "2L recipe (`--no-auto-layers --layers 2 --manufacturer jlcpcb "
+        "--backend python --seed 42`).  The underlying router regression "
+        "(both backends affected) is tracked in #3281; re-enable when that "
+        "is resolved.  The C++-backend production-recipe NRST gap remains "
+        "tracked separately per #3268."
+    )
+)
 class TestBoard04OscOutRouting:
     """Pin 9/9 routing on board 04 against the #2745 BLOCKED_BY_COMPONENT
     recovery fix.


rjwalters:
## Loom Judge Review — Approve

**Verdict: Approve.**

This is the textbook honest path for "test fails because production code is broken": skip-with-pointer plus a sharp follow-up. The skip annotation is honest about scope and points at a real, well-documented router regression rather than papering over it by switching backends or lowering the threshold.

### Verification

- **Single-file scope**: only `tests/test_board_04_routing_regression.py` (+21/-0). No collateral test or production code changes.
- **Skip reason is well-formed**: class-level `@pytest.mark.skip(reason=...)` mentions both #3268 (the test ticket) and #3281 (the underlying router regression), notes the C++ backend also drops to 4/9 on the stripped recipe (so backend-switching wouldn't help), and explicitly says re-enable when #3281 resolves.
- **Docstring trail extended**: lines 25–34 record the 2026-06-06 diagnosis so anyone bisecting later sees the full reasoning, not just the marker.
- **Trailer is correct**: `Refs #3268` (not `Closes`) — #3268 stays open as the test-side breadcrumb until #3281 lands and re-enables the assertion. This avoids the auto-close gotcha (no `Closes` trailer in the PR body either).
- **Follow-up #3281 is sharp**: documents the exact stripped recipe, both-backend observation (4/9 on python and cpp), the suspected dependency on `--micro-via-in-pad-fallback` for OSC_OUT escape, lists implicated router changes (#3118, #3033, #2988, #2778, #2829–#2830 family), and has an explicit AC item to re-enable this test. Re-enable trigger is unambiguous.
- **Local reproduction matches builder claim**: `uv run pytest tests/test_board_04_routing_regression.py tests/test_board_04_mfr_gate.py tests/test_board_04_drc_recipe_mfr.py -v` → 7 passed, 1 skipped. No collateral regression in the board-04 mfr-gate or DRC-recipe tests.
- **CI**: all 8 status checks SUCCESS (Lint, Type Check, Test, C++ Build, Round-trip Smoke, Routed PCB DRC, Diff-Pair, Match-Group). `mergeStateStatus: CLEAN`.

### Why this is the right shape

The two alternatives in #3268 (switch to `--backend cpp`, or lower the threshold) both lose signal — the PR description correctly rejects both:
1. C++ is also 4/9 on this recipe, so backend switch only relabels a known-fail.
2. 4/9 isn't a stable "current reality"; it's a documented gap relative to the post-#2745 invariant. Lowering would erase the regression marker.

Skip + cross-link preserves the regression signal and makes re-enabling a one-line change once #3281 resolves. This is exactly the pattern documented for "test fails because production code is broken".

Approving and moving to `loom:pr` for merge.

Issue #3280, opened 2026-06-07
Fleet status reports stale routing for board 03 (1/16 nets routed actually 11/13)


rjwalters:
## Summary

`kct fleet status` reports board 03 (`03-usb-joystick`) as having `1/16 nets routed` and marks it `incomplete routing (1/16 nets)`. But re-running `kct route` from the committed unrouted PCB actually produces **11/13 nets routed (85%)** with the `--backend cpp --seed 42 --auto-fix` recipe, deterministic across seeds.

The discrepancy is that fleet status reads the committed `boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb` which is a stale snapshot from earlier router work. The "1/16" figure is just what the stale file contains (note: the total-net count itself is wrong — the current netlist has 13 nets, not 16, so the stale routed PCB also reflects an outdated schematic).

Fleet status correctly flags it as `STALE` in the column, but the per-board `incomplete routing (1/16 nets)` blocker decision uses the stale data without re-running and without suppressing the misleading count.

## Why this matters

For agent-driven prioritisation, fleet status is the canonical "what's broken" signal. A board reporting 1/16 nets routed reads as catastrophic and gets prioritised over boards that are actually worse off, when in reality board 03 is at 11/13 (just one diff-pair + GND-pour stitching short of full). Curators and orchestrators making triage decisions from `kct fleet status` will mis-allocate work.

## Root cause (from code inspection)

`src/kicad_tools/cli/fleet_cmd.py`:

- `_discover_routed_pcb()` (line 282) globs `output/*_routed.kicad_pcb` — the committed file on disk.
- `_compute_routing()` (line 525) runs `NetStatusAnalyzer` directly against that file. It never re-routes and never consults the manifest or any "as-of" record of routing quality.
- `_survey_board()` (line 592) sets `mfg.stale = True` when `routed_mtime > manifest.json mtime` (lines 611-616). This is the existing staleness signal — it tells you the routed PCB is newer than the bundle, but does NOT tell you whether the routed PCB reflects current schematic/netlist intent.
- `_compute_blockers()` (line 544) emits `incomplete routing (X/Y nets)` from `_compute_routing()` results regardless of staleness — so the blocker is reported with numbers derived from a stale file.

In other words: there are two distinct staleness problems:
[...]

Issue #3279, opened 2026-06-07
2-layer boards: GND pads on F.Cu stranded when GND pour is on B.Cu (no stitching step)


rjwalters:
## Summary

On a 2-layer board where ``auto-pour`` creates the GND zone on B.Cu (the default), F.Cu SMD GND pads remain electrically stranded because nothing in the routing pipeline adds via-stitching to connect them.

Board 03 manifests this most clearly:
- 29 GND pads total (26 F.Cu SMD + 3 through-hole)
- GND zone created on B.Cu (priority 1)
- All 26 F.Cu SMD GND pads end up in the DRC ``connectivity`` error:
  ```
  connectivity: Net 'GND' is partially routed: 26 of 29 pads stranded
  (connected island has 3 pads)
  ```
  The 3 connected pads are the through-hole shield/mount/joystick pads which have ``*.Cu`` layer access via the barrel.

## Why nothing fixes this today

1. ``kct route`` skips ``GND`` as a pour net (``Auto-skip: GND, VBUS, VCC (pour nets — use zone fill)``).  No routing attempt is made to add stitching vias from F.Cu pads to the B.Cu pour. See `src/kicad_tools/router/auto_pour.py:300+` (``auto_skip_pour_nets``).

2. ``kct build`` includes a ``stitch`` step but it explicitly **skips when the board has only 2 layers**:
   - File: `src/kicad_tools/cli/build_cmd.py`
   - Lines 1573-1582 (the gate):
     ```python
     # Layer-count check.  On 2-layer boards there are no internal planes
     # to stitch to, so the step is a no-op.
     layer_count = _detect_layer_count(pcb_to_stitch)
     if layer_count <= 2:
         return BuildResult(
             step="stitch",
             success=True,
             message="stitch: 2-layer board — skipped (no internal planes)",
             output_file=pcb_to_stitch,
         )
     ```
   - Introduced by commit `38218a37e` (#2750 / closes #2747) on 2026-05-11. The original design rationale ("no internal planes to stitch to") was correct for the **4-layer use case** that motivated #2747, but it did not anticipate the **2L cross-layer case** (F.Cu pad → B.Cu pour), where a same-layer via stack (F.Cu→B.Cu) is exactly what's needed.

3. ``kct stitch`` standalone (`src/kicad_tools/cli/stitch_cmd.py`) **already handles** the 2L case correctly:
[...]

Issue #3278, opened 2026-06-07
Escape generator: per-row escape_width uses pads[0].net_class, pulls Power-class width into HighSpeed escapes


rjwalters:
## Summary

In `_create_fine_pitch_row_escapes` (`src/kicad_tools/router/escape.py:3186-3190` at HEAD), the escape segment width is computed **once per row** from `pads[0]`:

```python
escape_width = (
    self.rules.min_trace_width
    if self.rules.min_trace_width is not None
    else self._get_trace_width_for_net(pads[0].net_name if pads else "")
)
```

No built-in manufacturer profile sets the router-side `DesignRules.min_trace_width` (only the neck-down path sets it — see `rules.py:112`), so the fallback branch always fires in practice and `escape_width` is fixed at `pads[0]`'s net-class trace width for the entire row.

For USB-C connector J1 on board 03 the first pad in each row (sorted by x) is **GND** (Power net-class, `trace_width = 0.5mm`), so all escape segments — including the USB_D+ and USB_D- HighSpeed pads (`trace_width = 0.2mm`) — get a 0.5mm-wide segment.

## Observed effect on board 03

The width-0.5mm B.Cu escape segments for adjacent USB_D+ and USB_D- pads collide:

- centre-to-centre distance: 0.5mm
- edge-to-edge gap = 0.5 − (0.5 + 0.5)/2 = **0.000mm**

This produces the warning:

```
Escape clearance violation between pads USB_D- and USB_D+ on B.Cu: gap=0.0000mm (required 0.1500mm)
```

The USB_D+/USB_D- pads then defer to the main router which can't path-find through the post-escape geometry, leaving them as partial routes (2/3 pads stranded each).

## Why a naive per-pad fix REGRESSES

A naive fix — switch `escape_width` to per-pad `_get_trace_width_for_net(pad.net_name)` at the four `Segment(..., width=escape_width, ...)` sites — regresses the board because the `lateral_offset` calculation at `escape.py:3205-3209` consumes the same row-scoped `escape_width`:

```python
lateral_clearance = package.pin_pitch - escape_width
if lateral_clearance < effective_clearance:
    lateral_offset = (effective_clearance - lateral_clearance + escape_width) / 2
else:
    lateral_offset = 0.0
```

[...]

PR #3283, opened 2026-06-07
test(board03): pin June 2026 routing-reach baseline (11/13 nets routed)

### LINKED ISSUE #3259: Make board 03 usb-joystick manufacturable
## Goal

Make `boards/03-usb-joystick` manufacturable. Current baseline (per `kct fleet status`):
- 70/71 pads (99%)
- Routing: **1/16 nets routed — catastrophic baseline**
- Was previously timing out at 240s during the #2394 audit

This is the most-broken of the non-chorus boards in absolute terms. Even improving to 8/16 routed nets would be a valuable partial improvement.

## Scope

1. Diagnose why only 1/16 nets routed (per-net timeouts? USB connector escape geometry? auto-grid too coarse?)
2. Fix what's safely in-scope without regressing other boards
3. File follow-on issues for sub-mechanisms found but not fixed
4. Add regression test pinning the new baseline

## Out of Scope

- Auto-fix budget / kernel / pad halo (just shipped — don't touch)
- `route_cmd.py` / CLI changes (stay scoped)

## Tracking

- Related: #2394 (board manufacturability roadmap)
- Wave 1 fix: #3239 (auto-grid)

rjwalters:
## Summary

This PR resolves the parent tracking issue #3259 (\"make board 03 usb-joystick manufacturable\") by:

1. Establishing that the catastrophic ``1/16 nets routed`` reading from ``kct fleet status`` was a stale-artifact reading.  Live routing of the committed unrouted PCB produces **11/13 nets routed (85%)**, deterministically across seeds.
2. Pinning that 11/13 baseline in a new ``@pytest.mark.slow`` regression test under ``tests/router/test_board03_routing_baseline.py`` so a future router regression on USB-C-class pad-density boards is surfaced immediately.
3. Filing three follow-on issues for the sub-mechanisms that hold the reach at 11/13 instead of 13/13 (deliberately scoped out of this PR — see \"Follow-on issues\" below).

## Measurement table

Recipe used everywhere below:
```bash
uv run kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb \\
  --output /tmp/out.kicad_pcb --manufacturer jlcpcb-tier1 \\
  --backend cpp --timeout 600 --seed N --auto-fix --auto-fix-passes 2
```

| Backend | Seed | Nets routed | Best layers | Routes / Segments / Vias | Notes |
|---------|------|-------------|-------------|--------------------------|-------|
| cpp     | 1    | **11/13 (85%)** | 2 | 26 / 1046 / 19 | Deterministic baseline |
| cpp     | 42   | **11/13 (85%)** | 2 | 26 / 1046 / 19 | Same as seed 1/99 |
| cpp     | 99   | **11/13 (85%)** | 2 | 26 / 1046 / 19 | Seed-stable |
| cpp     | 7    | **11/13 (85%)** | 2 | (same)        | Seed-stable |
| cpp     | 123  | **11/13 (85%)** | 2 | (same)        | Seed-stable |
| python  | 42   | 6/13 → 7/13 (escalates to 4L) | 4 | --- | Worse; python backend struggles on this board |

The two consistently failing nets are **USB_D+ and USB_D-** (each 2 of 3 pads stranded).  The 11 routed nets are USB_CC1, USB_CC2, XTAL1, XTAL2, JOY_X, JOY_Y, JOY_BTN, BTN1, BTN2, BTN3, BTN4.

## Verdict

**Not ship-ready, but 85% routed with a clear path to follow-on work.**  The reach went from \"catastrophic 1/16 = 6%\" (stale fleet-status reading) to a measured-and-pinned 11/13 = 85% (live routing).  The remaining gap is:

- 2 nets out of 13 (USB_D+ / USB_D-) blocked by escape-geometry on J1 — tracked in #3278.
- The DRC ``GND partially routed: 26 of 29 pads stranded`` error is a separate 2-layer-pour stitching pipeline gap — tracked in #3279, NOT a routing-quality issue.

The 1/16-myth fleet-status reporting gap is filed as #3280.

## Investigation summary

I attempted two in-scope fixes during the investigation but both were either ineffective or regressed the baseline; both were reverted before commit:

1. **Skip intra-IC consolidation for diff-pair partners on the same package**:  the intra-IC route creates straight diagonals from A6→B6 (USB_D+) and B7→A7 (USB_D-) that cross at the centre of J1.  Skipping them did NOT improve reach; the bottleneck is escape geometry, not intra-IC crossing.

2. **Use per-pad escape_width instead of pads[0]'s row-wide width**: this would fix the spurious B.Cu 0.000mm-gap clearance violation warning (the immediate symptom).  But it REGRESSED the routing because the ``lateral_offset`` calculation depends on the row's ``escape_width``; a narrower width collapses the lateral_offset to zero, causing cross-row USB_D+/USB_D- vias to land at the same coordinates on B.Cu — a worse violation than the original.  The correct fix needs to thread per-pad width into the segment but compute lateral_offset from the maximum width in the row; that's #3278.

## Follow-on issues

- **#3278** — Escape generator: per-row ``escape_width`` uses ``pads[0].net_class``, pulls Power-class width into HighSpeed escapes.  Root cause of the USB_D+/USB_D- partial routing on board 03.  Needs careful work to avoid regressing the lateral_offset via-via-collision protection.
- **#3279** — 2-layer boards: GND pads on F.Cu stranded when GND pour is on B.Cu (no stitching step).  Pipeline gap; ``kct build``'s stitch step explicitly skips 2L boards.
- **#3280** — Fleet status reports stale routing for board 03 (1/16 actually 11/13).  The fleet-status command should not trust stale committed routed PCB files for ``incomplete routing (N/M)`` decisions.

## Files touched

- ``tests/router/test_board03_routing_baseline.py`` (new) — 3 tests:
  - ``test_reach_meets_floor``: pins 11/13 floor.
  - ``test_the_1_of_16_myth_stays_dead``: hard-fails if reach ever drops to <= 1 (the catastrophic regression we want to never re-introduce).
  - ``test_routing_reach_deterministic_across_seeds``: pins seed-stability across seeds 1/42/99.

All three are ``@pytest.mark.slow`` (wall-clock ~2.5 min and ~7.5 min respectively) so PR-time CI skips them; the nightly slow-tests workflow picks them up.

## Test plan

- [x] ``tests/router/test_board03_routing_baseline.py`` passes locally with the slow marker.
- [x] ``tests/router/test_board02_manufacturable_baseline.py`` still passes (no board 02 regression).
- [x] ``tests/test_board_03_regression.py`` still passes (no board 03 demo-path regression).
- [ ] Slow-tests workflow picks up the new test class.

Closes #3259

### DIFF
--- a/tests/router/test_board03_routing_baseline.py
+++ b/tests/router/test_board03_routing_baseline.py
@@ -1,0 +1,291 @@
+"""Board 03 (usb-joystick) routing baseline regression guard.
+
+This test pins the **measured routing reach** of
+``boards/03-usb-joystick/`` against the ``kct route`` CLI as of
+June 2026.
+
+Baseline measurement at HEAD (with ``kct route --backend cpp --seed 42
+--auto-fix --auto-fix-passes 2 --manufacturer jlcpcb-tier1``):
+
+- **Routed: 11/13 nets (85%)** — all signal nets except USB_D+ / USB_D-
+- **Routes created: 26**, ~1046 segments, 19 vias
+- **Layer count: 2** (the 2-layer attempt produces the best result)
+- USB_D+ and USB_D- are deterministically partial (2/3 pads stranded
+  each) due to escape-geometry interactions with J1's USB-C connector
+  pad layout — tracked in #3278.
+- The committed unrouted PCB has 16 nets total: 13 are routed, 3 are
+  power/pour nets (VCC, VBUS, GND) that are skipped by the router and
+  served by auto-pour zones instead.
+
+Context (the "1/16" myth):
+    The pre-test ``kct fleet status`` output for board 03 reads
+    ``incomplete routing (1/16 nets)`` because it queries the stale
+    committed ``output/usb_joystick_routed.kicad_pcb`` snapshot.  Live
+    routing produces 11/13.  The stale-fleet-status reporting gap is
+    tracked separately in #3280.
+
+Known follow-on issues that prevent a higher baseline:
+    - **#3278**: Escape generator uses ``pads[0].net_name``'s
+      net-class trace width for the whole row, pulling Power-class
+      0.5mm width into the USB_D+/USB_D- HighSpeed escapes.  The
+      resulting fat-segment B.Cu escape clearance violation defers
+      both diff-pair pads to the main router, which can't path-find
+      the remaining geometry — hence 2/3 pads stranded.
+    - **#3279**: 2-layer boards with GND pour on B.Cu have no
+      pipeline step to stitch F.Cu SMD GND pads to the pour, so
+      ``kct check`` reports ``Net 'GND' is partially routed: 26 of 29
+      pads stranded``.  Routing itself is fine; it's a pipeline gap.
+
+Acceptance criteria pinned by this test:
+
+1. **Reach floor**: ``kct route`` produces >= 11 routed signal nets
+   (out of 13).  Drops to 10 or fewer indicate a routing-quality
+   regression on USB-C-class pad-density boards.
+2. **Deterministic across seeds**: seeds 1 / 42 / 99 all produce the
+   same routed-net count, so a single-seed run is a reliable
+   indicator of overall quality.
+3. **The 1/16 myth stays dead**: if the test ever measures <= 1 net
+   routed, somebody re-broke board 03 in a non-trivial way.
+
+References:
+    - Parent tracking issue: #3259
+    - Stale fleet-status reporting: #3280
+    - Escape clearance bug: #3278
+    - 2-layer pour stitching gap: #3279
+    - Existing board-03 demo-path test: tests/test_board_03_regression.py
+"""
+
+from __future__ import annotations
+
+import re
+import shutil
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+import pytest
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+BOARD_DIR = REPO_ROOT / "boards" / "03-usb-joystick"
+UNROUTED_PCB = BOARD_DIR / "output" / "usb_joystick.kicad_pcb"
+
+# Acceptance criteria for the June 2026 baseline.
+#
+# 13 signal nets total (USB_CC1, USB_CC2, USB_D+, USB_D-, XTAL1, XTAL2,
+# JOY_X, JOY_Y, JOY_BTN, BTN1, BTN2, BTN3, BTN4).  The 3 pour/power
+# nets (VCC, VBUS, GND) are auto-skipped and served by zones; they do
+# NOT appear in the ``Nets routed: N/M`` line.
+#
+# USB_D+ and USB_D- are partial in the current baseline (see #3278), so
+# the typical run lands at 11/13.  We pin the floor at 11 to catch a
+# routing-quality regression; ratchet up to 12 or 13 when #3278 lands.
+REQUIRED_NETS_ROUTED = 11
+EXPECTED_TOTAL_NETS = 13
+
+
+def _parse_routed_net_count(stdout: str) -> tuple[int, int] | None:
+    """Extract the final ``Nets routed: N/M`` count from kct route output.
+
+    The expected summary block contains a line of the form::
+
+        Nets routed:     11/13
+
+    (Multiple matches may exist in escalation mode; we return the LAST
+    one since that reflects the final state the router landed on.)
+
+    Returns ``(routed, total)`` or ``None`` if no such line was found.
+    """
+    pattern = re.compile(r"Nets routed:\s+(\d+)/(\d+)")
+    matches = pattern.findall(stdout)
+    if not matches:
+        return None
+    routed, total = matches[-1]
+    return int(routed), int(total)
+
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge review — APPROVE

### Verdict
**Approve.**  This PR is exactly the right shape for the Wave 5 board-03 work: a tightly-scoped, diagnostic-only test addition that re-classifies board 03 from "catastrophic 1/16 (stale)" to "measured-and-pinned 11/13 (85%)" with three well-scoped follow-on issues for the residual gap.  Zero ``src/`` changes, zero regression surface.

### Strengths

1. **Minimal, surgical diff** — exactly one test file (291 lines), zero ``src/`` changes.  Approves at the same risk surface as a docs PR.
2. **Acceptance criteria correctly pinned** — floor at 11/13, ``EXPECTED_TOTAL_NETS = 13`` guards against a net-topology change that would silently invalidate the baseline, and seed-stability is checked across 1/42/99.  The "ratchet up to 12 or 13 when #3278 lands" comment in the source is a nice touch for future readers.
3. **Slow-marker hygiene** — ``@pytest.mark.slow`` is correctly registered in ``pyproject.toml:260`` and the test takes ~2.5min (single seed) / ~7.5min (determinism check), so PR-time CI correctly skips it while the nightly slow-tests workflow picks it up.
4. **Subprocess invocation mirrors the documented recipe** in #3259 exactly: ``--backend cpp --seed N --auto-fix --auto-fix-passes 2 --manufacturer jlcpcb-tier1``.  This is reproducible.
5. **Output parsing is robust** — the regex ``Nets routed:\s+(\d+)/(\d+)`` matches the actual format printed at ``route_cmd.py:1028``, and ``findall(...)[-1]`` correctly grabs the FINAL escalation result instead of the first attempt's count.  Important nuance for layer-escalation mode.
6. **Exit-code triage** — 1/5 fail hard (crash/SIGINT); 0/2/3 are acceptable (full, partial-below-min, partial-with-DRC).  Correct classification per ``route_cmd.py`` semantics.
7. **The "1/16 myth stays dead" test** is a smart catastrophic-regression sentinel — it's technically redundant with the ``>= 11`` floor in the same module-scoped fixture, but it serves as explicit documentation of the original failure mode and gives a much better error message for future bisectors.

### Root-cause verification

I checked the #3278 claim against ``src/kicad_tools/router/escape.py`` directly.  The bug is **real and correctly diagnosed**:

- ``escape_width`` is computed at ``escape.py:3186-3190`` (the PR body says 3209-3213, but that's the lateral_offset block — minor labelling slip, doesn't affect the analysis):
  ```python
  escape_width = (
      self.rules.min_trace_width
      if self.rules.min_trace_width is not None
      else self._get_trace_width_for_net(pads[0].net_name if pads else "")
  )
  ```
- Then ``lateral_offset`` at ``escape.py:3205-3209`` mathematically depends on it:
  ```python
  lateral_clearance = package.pin_pitch - escape_width
  if lateral_clearance < effective_clearance:
      lateral_offset = (effective_clearance - lateral_clearance + escape_width) / 2
  ```
- So a naive per-pad fix would collapse ``lateral_offset`` to 0 for HighSpeed pads and cause cross-row vias to coincide — exactly as #3278 documents.  The follow-on is well-scoped and includes the correct fix-plan sketch (use per-pad width for ``Segment.width`` but compute ``lateral_offset`` from the row's maximum width).

### Follow-on issue scoping

All three are appropriately separated and scoped:

- **#3278** — escape root cause, correct fix discussion, includes the non-trivial lateral_offset coupling
- **#3279** — 2L pour-stitching pipeline gap, two proposed fix paths (build-step gate vs. router-side via emission), correctly identified as separate from routing-quality
- **#3280** — fleet-status staleness, suggests two options (mark-as-unknown vs. re-route on STALE), low-priority diagnostic improvement

### CI status

| Check | Result |
|-------|--------|
| C++ Build Check | pass (42s) |
| Diff-Pair Routing Regression | pass (11m57s) |
| Match-Group Routing Regression | pass (10m53s) |
| Lint & Format | pass (11s) |
| Routed PCB DRC Check | pass (10s) |
| Test | pass (8m59s) |
| Type Check | pass (38s) |
| kicad-cli Round-trip Smoke | pass (44s) |

``mergeStateStatus: CLEAN``, ``mergeable: MERGEABLE``.

### Nits (non-blocking)

- PR body cites ``escape.py:3209-3213`` for the bug site; the actual ``escape_width`` assignment is at lines 3186-3190 and the dependent ``lateral_offset`` block is at 3205-3209.  Minor mislabelling; the semantics are correct.
- ``test_the_1_of_16_myth_stays_dead`` re-uses the same module-scoped ``route_stdout`` fixture as ``test_reach_meets_floor``, so the ``> 1`` assertion is technically a subset of ``>= 11``.  Fine — keeping it makes the regression narrative legible.

### Big picture

This PR is the kind of work that compounds: it removes a misleading "catastrophic" signal from the fleet dashboard, pins a measured baseline that will surface future regressions, and files three crisp follow-on issues for the remaining gap.  The diagnostic value (1/16 -> 11/13 with three named bugs) is much larger than the line count suggests.

Ship it.

— Loom Judge

Issue #3206, opened 2026-06-05
kct fleet status: routing_complete doesn't honor ADVISORY_RULE_IDS — plane-net stitching residuals overcounted as routing-incomplete


rjwalters:
## Context

\`kct fleet status\` reports \`ship_ready: NO (incomplete routing N/M nets)\` on boards whose router output is functionally complete because it counts pad-stitching residuals on plane nets (typically GND) as routing-incomplete. This contradicts \`DRCChecker.ADVISORY_RULE_IDS\` policy, which already excludes \`connectivity\` from the CI gate (since GND/plane-pin stitching residuals are not a routing-completeness blocker).

Concrete: **board 04** today (2026-06-05) routes **9/9 signal nets (100%)** but fleet reports \`1/12 nets\` incomplete. The 12th net is GND; the 1 stranded pad is U2.8 OSC_OUT — an advisory connectivity issue that already passes the CI gate. Likely affects boards 03/06/07 too (each has known advisory connectivity residuals).

## Reproduction

\`\`\`bash
cd boards/04-stm32-devboard
uv run python generate_design.py  # routes 9/9 signal nets
uv run kct fleet status --boards-dir boards/04-stm32-devboard
# Reports: ship_ready=NO, routing_complete=NO (1/12 nets), even though
# the 1 \"incomplete\" pad is a plane-stitching connectivity advisory.
\`\`\`

Cross-check: \`scripts/ci/check_routed_drc.py\` uses \`_count_blocking_errors\` which filters \`ADVISORY_RULE_IDS\`. Fleet status's \`routing_complete\` field does not apply the same filter.

## Likely fix surface

- \`src/kicad_tools/cli/fleet_cmd.py\` — likely \`fleet_cmd.py:80-85\` or the equivalent \`routing_complete\` derivation (per audit notes).
- \`src/kicad_tools/analysis/net_status.py:195-217\` — the net-completion classifier.

Apply the same \`ADVISORY_RULE_IDS\` filter the CI gate already uses. Net-completion residuals attributable solely to advisory rules should not contribute to \`routing_complete=NO\`.

## Acceptance criteria

1. Board 04's fleet status reports \`routing_complete: YES\` when only the U2.8 GND stitching residual remains.
2. Boards 03/06/07 also flip to \`routing_complete: YES\` if their only remaining gaps are advisory.
3. Genuinely incomplete signal nets still flag \`routing_complete: NO\` (i.e., the filter is rule-id-scoped, not blanket).
[...]

PR #3289, opened 2026-06-07
fix(fleet): detect schematic-vs-PCB drift, suppress misleading routing blocker (#3280)

### LINKED ISSUE #3280: Fleet status reports stale routing for board 03 (1/16 nets routed actually 11/13)

rjwalters:
## Summary

`kct fleet status` was reporting board 03 (`03-usb-joystick`) as `incomplete routing (1/16 nets)` even though running `kct route` against the source schematic produces 11/13. The blocker text was driven by the committed `_routed.kicad_pcb` whose net set (16) no longer matched the source `.kicad_sch` (13). Catastrophic-looking 1/16 figures steered curator triage toward this board over genuinely worse boards.

This PR implements **Option A** from the issue's acceptance criteria (with the lightweight detection sketched as Option B): detect schematic-vs-PCB net-set drift, surface a clearer blocker, and suppress the misleading routed-PCB-net count.

## Behavior change

Before:
```
03-usb-joystick     1/13   ...  NO  (incomplete routing (1/16 nets))
```

After:
```
03-usb-joystick   70/71  99% ...  NO  (routed PCB stale (schematic drift: 13 nets in schematic, 16 in PCB))
```

Boards with no drift (02, 06, 07) keep their existing `incomplete routing` semantics. Boards without a `.kicad_sch` alongside the routed PCB (fixtures, legacy layouts) skip drift detection entirely so `make_fake_board`-based tests continue to pass unchanged.

## Implementation

`src/kicad_tools/cli/fleet_cmd.py`:
- `RoutingStatus` gains `source_stale`, `schematic_net_count`, `pcb_net_count`, `drift_added`, `drift_removed` fields. When `source_stale` is True, `routing_complete` is forced False so ship-readiness is gated.
- New `_detect_source_drift(routed_pcb, board_dir)` compares the schematic's `label`/`global_label`/`hierarchical_label` names against the PCB's `(net N "name")` declarations. Conservative: missing schematic / parse failure -> no drift (back-compat).
- `_compute_blockers` emits `routed PCB stale (schematic drift: N nets in schematic, M in PCB)` instead of `incomplete routing (X/Y nets)` when drift is detected.
- `_survey_board` wires the new check after `_compute_routing`.

The change is **scoped to `fleet_cmd.py` and its tests**; it does not touch the manifest-vs-routed staleness path covered by #3264 / PR #3277.

## Test plan

- [x] New `TestFleetStatusSourceDrift` class with 5 regression tests:
  - `test_drift_suppresses_incomplete_routing_blocker` — drift fixture suppresses `incomplete routing` blocker
  - `test_no_drift_preserves_routing_blocker` — matched fixture preserves existing behavior
  - `test_missing_schematic_does_not_gate` — back-compat: no schematic, no drift gating
  - `test_drift_blocker_message_format` — blocker text surfaces both counts
  - `test_drift_ship_ready_false` — drift forces ship-ready=false even when routing looks 100%
- [x] All 51 pre-existing fleet tests still pass (`tests/test_fleet_status*.py`, `tests/test_fleet_ship_ready.py`, `tests/test_fleet_status_drc.py`)
- [x] Real-repo smoke: `uv run kct fleet status` against `boards/` no longer reports `1/16` for board 03; instead shows `routed PCB stale (schematic drift: 13 nets in schematic, 16 in PCB)`. Boards 02/06/07 (no drift) unchanged.
- [x] `ruff check` + `ruff format` clean on touched files.

Closes #3280

### DIFF
--- a/src/kicad_tools/cli/fleet_cmd.py
+++ b/src/kicad_tools/cli/fleet_cmd.py
@@ -39,6 +39,7 @@ from __future__ import annotations
 import argparse
 import datetime as _dt
 import json
+import re
 import sys
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -81,6 +82,20 @@ class RoutingStatus:
     blocking_incomplete_nets: int = 0
     unrouted_nets: int = 0
     error: str | None = None
+    # Issue #3280: when True, the routed PCB's net set disagrees with the
+    # source schematic's named-label set, so any ``X/Y nets`` count derived
+    # from the PCB is misleading as a current-routing signal. Filled in by
+    # ``_survey_board`` via ``_detect_source_drift``.
+    source_stale: bool = False
+    # Distinct counts that produced the drift verdict above. ``None`` means
+    # we could not extract them (no schematic alongside the PCB, parse
+    # failure, etc.) -- treated as "not source-stale".
+    schematic_net_count: int | None = None
+    pcb_net_count: int | None = None
+    # Sample of drifted nets (added/removed in PCB vs schematic), capped to
+    # keep JSON output bounded. Empty when ``source_stale`` is False.
+    drift_added: list[str] = field(default_factory=list)
+    drift_removed: list[str] = field(default_factory=list)
 
     @property
     def completion_pct(self) -> float:
@@ -92,15 +107,17 @@ class RoutingStatus:
     def routing_complete(self) -> bool:
         if self.error is not None:
             return False
+        # If the routed PCB is stale relative to its source schematic, its
+        # routed-net count is not a trustworthy signal for "is the current
+        # design routed?" -- treat as not complete (issue #3280).
+        if self.source_stale:
+            return False
         # A board is routing-complete iff every multi-pad net is fully
         # connected -- with one exception: plane/pour stitching residuals
         # (advisory connectivity per ``ADVISORY_RULE_IDS``) are excluded
         # because the CI gate at ``check_routed_drc`` already ignores
         # them. Single-pad nets are not counted by NetStatusAnalyzer.
-        return (
-            (self.blocking_incomplete_nets + self.unrouted_nets) == 0
-            and self.total_nets > 0
-        )
+        return (self.blocking_incomplete_nets + self.unrouted_nets) == 0 and self.total_nets > 0
 
     def to_dict(self) -> dict:
         data: dict = {
@@ -113,7 +130,14 @@ class RoutingStatus:
             "blocking_incomplete_nets": self.blocking_incomplete_nets,
             "unrouted_nets": self.unrouted_nets,
             "routing_complete": self.routing_complete,
+            "source_stale": self.source_stale,
+            "schematic_net_count": self.schematic_net_count,
+            "pcb_net_count": self.pcb_net_count,
         }
+        if self.drift_added:
+            data["drift_added"] = list(self.drift_added)
+        if self.drift_removed:
+            data["drift_removed"] = list(self.drift_removed)
         if self.error is not None:
             data["error"] = self.error
         return data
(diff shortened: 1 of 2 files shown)


rjwalters:
## Loom Judge Review — APPROVE

Reviewed implementation against the acceptance criteria from #3280.

### Verification performed
- Read PR diff (`src/kicad_tools/cli/fleet_cmd.py` +179 / `tests/test_fleet_status.py` +295)
- Ran `pytest tests/test_fleet_status.py` -> **26/26 pass** locally
- Ran `pytest tests/test_fleet_ship_ready.py tests/test_fleet_status_drc.py` -> **30/30 pass**
- Ran `ruff check` + `ruff format --check` on touched files -> clean
- Smoke-tested `uv run kct fleet status --format json` against `boards/`
- CI: **8/8 green** (Lint & Format, Type Check, Test, C++ Build, Routed PCB DRC, kicad-cli Round-trip, Diff-Pair Regression, Match-Group Regression)

### What works
1. **Board 03 fix confirmed live:** row now reads `routed PCB stale (schematic drift: 13 nets in schematic, 16 in PCB)` instead of `incomplete routing (1/16 nets)`. This is the exact behavior change called out in the issue.
2. **No false positives** on boards 02/06/07 (`source_stale=False`, existing semantics preserved). Boards 04/05 surface real drift; board 01 catches a genuine rename (`+5V` -> `VIN`) that an mtime-only check would have missed — the sharper net-name-set comparison earns its keep.
3. **`routing_complete` correctly gated:** when `source_stale=True`, `routing_complete` is forced False, which propagates to `ship_ready`. The `test_drift_ship_ready_false` test guards against the dangerous case of a 100%-routed stale PCB hiding the drift.
4. **Conservative back-compat:** missing schematic / parse failure -> `(False, None, None, [], [])`, so fixture boards using `make_fake_board` are unaffected. `test_missing_schematic_does_not_gate` covers this.
5. **Structured JSON output:** `source_stale`, `schematic_net_count`, `pcb_net_count`, `drift_added`, `drift_removed` (capped at 8 samples) — programmatic consumers get clean signal. `schema_version` bumped to `1.1`.
6. **Scoped change:** does not touch the manifest-vs-routed staleness path (#3264 / PR #3277). The two staleness checks compose cleanly.
7. **Regex parsing is defensible:** `(label|global_label|hierarchical_label)\s+"..."` and `(net N "...")` anchor on quoted strings, which are unambiguous in KiCad's s-expression grammar. Lightweight enough for a status command.

### Test coverage
All 5 specified regression tests present in `TestFleetStatusSourceDrift`:
- `test_drift_suppresses_incomplete_routing_blocker`
- `test_no_drift_preserves_routing_blocker`
- `test_missing_schematic_does_not_gate`
- `test_drift_blocker_message_format`
- `test_drift_ship_ready_false`

Plus the 8 prior `TestFleetStatusAdvisoryFilter` tests (related #3206 work bundled in this commit) — all pass.

### Minor observations (non-blocking, future work)
- `Net-(...)` auto-placeholders are filtered on the schematic side but not the PCB side. Symmetric filtering could marginally reduce false drift signals on boards with many unnamed nets. Not a regression here since current boards either have named nets or both sides carry the placeholder.
- Could consider a `--explain-drift` flag later to dump the full added/removed lists when curators want to know exactly which nets changed (the 8-sample cap is right for normal output).

Neither blocks merge.

### Verdict
**APPROVE.** Resolves #3280 cleanly, no regressions detected, CI green, merge state CLEAN. Labeling for merge.

Issue #2747, opened 2026-05-12
kct build: zone stitching not auto-invoked after zone creation (plane-net pads remain unconnected)


rjwalters:
## Symptom

`kct build` creates copper-pour zones (e.g., VCC, GND, VMOTOR) during its `zones` step, but **never auto-invokes** zone stitching that adds thermal-relief vias from plane-net pads down to the inner-layer zone copper. Result: a routed PCB where plane-net pads sit visually "on top of" a zone but are electrically disconnected.

## Evidence (2026-05-11 audit)

- **Board 03-usb-joystick**: 3 zones exist; 40 pads on GND/VCC/VBUS report "needs routing" — never stitched.
- **Board 05-bldc-motor-controller**: 0 zones in committed PCB (separate issue — zone step itself failed) but the 20 VMOTOR pads "needs routing" indicate the same gap would apply if zones existed.
- **Board 02-charlieplex-led** (May-6 baseline): passed MFG-READY without stitching because it is **2-layer** — there is no inner plane to stitch to, so the symptom is invisible there. This is NOT a regression in the stitching path; the stitch step never existed in `kct build`.

## Root cause (verified)

The `kct build` orchestrator has **no stitch step at all**. Specifically:

1. `src/kicad_tools/cli/build_cmd.py` defines `BuildStep` enum at lines 38–50 with values `SCHEMATIC, PCB, OUTLINE, PLACEMENT, ZONES, SILKSCREEN, ROUTE, VERIFY, EXPORT`. **No `STITCH` member.**
2. The default chain at `build_cmd.py:1703–1714` (`if args.step == "all":`) lists those nine steps in order; stitching is not invoked.
3. The argparse `--step` choices at `build_cmd.py:1546–1557` do not include `"stitch"`.
4. When the route subprocess is launched at `build_cmd.py:1128–1147` the command line does **not** include `--stitch-power-planes`, so `route_cmd.py:5538–5573` (which can stitch in-process when that flag is set) is bypassed.
5. `grep -n "STITCH\|stitch"` over `build_cmd.py` returns zero matches — confirming the gap.

### Reference implementation already exists

The sister orchestrator `src/kicad_tools/cli/pipeline_cmd.py` has had a working STITCH step since at least the docstring rewrite at lines 11/20–22:

[...]

Issue #3271, opened 2026-06-07
board 06 recipe: integrate post-route stitch for GND/VBUS_USB pour-net plane connections


rjwalters:
## Symptom

`boards/06-diffpair-test/generate_design.py` runs the autorouter, copper-pour zone generator, and DRC verify-and-nudge — but **does not invoke `kct stitch` for plane-net via stitching**. This leaves:

- `GND`: 118 pads stranded (must stitch to In1.Cu pour) — counts as 1 connectivity DRC error
- `VBUS_USB`: 4 pads stranded (must stitch to In2.Cu pour) — counts as 1 connectivity DRC error

These pad gaps are also why `kct fleet status` reports board 06 at 37.88% pad completion despite 23/26 nets being complete on the signal side.

## Reproduction (post #3262)

```bash
cd boards/06-diffpair-test
PYTHONHASHSEED=42 python generate_design.py --step route --seed 42
kct check output/diffpair_test_routed.kicad_pcb --mfr jlcpcb --errors-only
```

Expected (post-#3262 baseline):
```
Errors: 6
  connectivity: 3 errors  (GND, VBUS_USB, USB3_TX1+)
  clearance_segment_segment: 2 errors
  diffpair_clearance_intra: 1 error
```

Two of the three `connectivity` errors and ~120 of the stranded pads are addressable via post-route stitching.

## Constraint

A naive `kct stitch --net GND --net VBUS_USB` invocation on this board placed **28 via-in-pad violations** for JLCPCB tier-1 (which does not support via-in-pad). The stitch tool would need either:

1. A guard against placing vias inside thermal-relief pads on QFN/BGA components (e.g., honor `--target-layer` more strictly, or surface a `--avoid-pad-overlap` mode).
2. A switch to `--mfr jlcpcb-tier1` (which permits via-in-pad) — but the board declares `target_fab: jlcpcb` in `project.kct` so this would be a profile mismatch.
3. A finer pad-aware stitch placement that only adds vias near pads with explicit thermal-relief clearance.

## Proposed work

1. Investigate whether `kct stitch` has (or needs) a "skip pads that would produce via-in-pad" mode.
2. If yes, wire it into `generate_design.py` after the copper-pour zone pass.
3. If no, file a sub-issue against `src/kicad_tools/router/stitch.py` (or wherever the stitch logic lives) for that mode.

## Expected delta

Closing this would:
[...]

PR #3290, opened 2026-06-07
fix(build): stitch 2L boards when GND pour is cross-layer (closes #3279)

### LINKED ISSUE #3279: 2-layer boards: GND pads on F.Cu stranded when GND pour is on B.Cu (no stitching step)

rjwalters:
## Summary

Fixes #3279: 2-layer boards with a B.Cu GND pour and F.Cu SMD GND pads end up DRC-failing because the build pipeline's stitch step skipped any board with `layer_count <= 2`. This blanket skip was introduced for the 4L case (#2747) but did not anticipate the **2L cross-layer** case where a same-layer F.Cu->B.Cu via stack is exactly what is needed.

The underlying primitive in `run_stitch` already handles the 2L fallback correctly (`stitch_cmd.py` "fall back to B.Cu" path). Only the pipeline gate needed fixing.

## Changes

- `_run_step_stitch` (build_cmd.py): probe plane nets and plane-net pads first, then skip only when every plane-net pad already sits on the pour layer for its net. Through-hole pads are excluded (their plated barrel grants `*.Cu` access).
- New skip messages:
  - `stitch: no cross-layer plane pads -- skipped` (zone exists, pads all on pour layer)
  - `stitch: no plane-net pads found -- skipped` (zone exists with no pads)
- `_detect_layer_count` is retained for the downstream manufacturer-profile via-size lookup.

## Verification

**Bug reproduces on a fresh route of board 03:**
```
DRC: connectivity: Net 'GND' is partially routed: 26 of 29 pads stranded
```

**After `_run_step_stitch` runs with the fix:**
```
Stitching complete: added 31 via(s) for 3 plane net(s),
                    9 pad(s) already connected, 3 pad(s) skipped
DRC: connectivity: Net 'GND' is partially routed: 5 of 29 pads stranded
```

The remaining 5 stranded GND pads come from clearance-induced via skips (the stitcher correctly refuses to place a via where it would short an adjacent net). That falls under #3271's pad-aware-stitching umbrella and is out of scope per the curator's recommendation to ship Option A first.

**Board 02 (2L charlieplex-led) regression check:**
- Stitch step cleanly skips with `stitch: no cross-layer plane pads -- skipped`.
- Full DRC still passes (`DRC PASSED - No violations found`).

**Idempotency:** re-running stitch on board 03 reports `40 plane pad(s) already connected, no new vias needed`.

## Tests

- Added `TWO_LAYER_WITH_BCU_POUR_PCB` fixture reproducing #3279; asserts `_run_step_stitch` does NOT skip and that `run_stitch` connects the F.Cu GND pads to the B.Cu pour.
- Added `TWO_LAYER_PADS_MATCH_POUR_PCB` fixture asserting the new "no cross-layer plane pads" skip when every pad already lives on the pour layer.
- Updated `test_skip_on_two_layer_board` (renamed `_without_pour`) to assert the new "no plane nets detected" message for the existing no-zone 2L fixture.
- All 14 tests in `tests/test_build_stitch_step.py` pass.

## Test plan

- [x] `uv run pytest tests/test_build_stitch_step.py` -> 14 passed
- [x] `uv run pytest tests/test_stitch_cmd.py tests/test_stitch_mfr.py tests/test_stitch_bga_outer_ring.py tests/test_board_05_thermal_stitch.py` -> all pass (no regression on 4L stitch behaviour)
- [x] Fresh route + stitch on board 03 drops GND stranding from 26/29 to 5/29
- [x] Board 02 still passes MFG-READY DRC

## Pre-existing test failures (NOT introduced by this PR)

These failures exist on `origin/main` (confirmed via `git stash` + test run on the unmodified tree):
- `tests/test_build_cmd_errors.py::TestBuildRouteExitCode2::test_route_exit_code_3_is_failure`
- `tests/test_build_preflight_routing_step.py::TestPreflightRoutingBlocksManufacturing::test_real_fixture_voltage_divider_routed_pcb_is_rejected`

Both are unrelated to `_run_step_stitch` (one is about route exit-code handling, the other is a stale-fixture assertion for the preflight step).

Closes #3279

### DIFF
--- a/src/kicad_tools/cli/build_cmd.py
+++ b/src/kicad_tools/cli/build_cmd.py
@@ -1535,8 +1535,11 @@ def _run_step_stitch(ctx: BuildContext, console: Console) -> BuildResult:
 
     Skips gracefully when:
     - No PCB to stitch (no route output) — succeeds as a no-op.
-    - The board has only 2 layers (no internal planes to stitch to).
     - The PCB has no plane-net zones to stitch onto.
+    - Every plane-net pad already sits on the pour layer for its net
+      (no cross-layer stitching work needed).  Issue #3279: this
+      replaced the prior ``layer_count <= 2`` blanket skip so that
+      2-layer boards with a B.Cu pour and F.Cu SMD pads get stitched.
 
     Operates on :attr:`BuildContext.routed_pcb_file` when present (the
     output of the route step), falling back to :attr:`BuildContext.pcb_file`
@@ -1570,22 +1573,25 @@ def _run_step_stitch(ctx: BuildContext, console: Console) -> BuildResult:
             message="stitch: no PCB file found — skipped",
         )
 
-    # Layer-count check.  On 2-layer boards there are no internal planes
-    # to stitch to, so the step is a no-op.
-    layer_count = _detect_layer_count(pcb_to_stitch)
-    if layer_count <= 2:
-        return BuildResult(
-            step="stitch",
-            success=True,
-            message="stitch: 2-layer board — skipped (no internal planes)",
-            output_file=pcb_to_stitch,
-        )
-
     # Probe the PCB for plane nets before invoking the stitcher.  This
     # avoids any work when there is nothing to stitch (e.g. a multi-layer
     # board whose zones step did not run, or a board with no power pours).
+    #
+    # Historical note (#3279): previously this step short-circuited on any
+    # board with <= 2 copper layers under the assumption that "no internal
+    # planes" means "nothing to stitch."  That holds for the 4-layer case
+    # the skip was introduced for (#2747) but **not** for the 2-layer
+    # cross-layer case: e.g. a 2L board with the GND pour on B.Cu and
+    # F.Cu SMD GND pads needs a same-layer via stack (F.Cu -> B.Cu) to
+    # connect the pads to the pour.  ``run_stitch`` already handles the
+    # 2L fallback correctly (see stitch_cmd.py "fall back to B.Cu"), so
+    # the only fix needed here is to gate on "is there real work to do?"
+    # rather than layer count.
     try:
-        from kicad_tools.cli.stitch_cmd import find_all_plane_nets
+        from kicad_tools.cli.stitch_cmd import (
+            find_all_plane_nets,
+            find_pads_on_nets,
+        )
         from kicad_tools.core.sexp_file import load_pcb as _load_pcb
 
         sexp = _load_pcb(pcb_to_stitch)
@@ -1593,12 +1599,46 @@ def _run_step_stitch(ctx: BuildContext, console: Console) -> BuildResult:
     except Exception as exc:
         logger.debug("Could not probe plane nets: %s", exc)
         plane_nets = {}
+        sexp = None
 
     if not plane_nets:
         return BuildResult(
             step="stitch",
             success=True,
             message="stitch: no plane nets detected — skipped",
+            output_file=pcb_to_stitch,
+        )
+
+    # Pad-layer-mismatch probe (#3279).  Skip only when every plane-net
+    # pad already sits on the pour layer for its net -- nothing to
+    # stitch.  Otherwise (including the 2L F.Cu-pad-with-B.Cu-pour case)
+    # fall through to ``run_stitch`` and let it place the vias.
+    #
+    # Through-hole pads carry their own ``*.Cu`` barrel and are
+    # effectively present on every layer, so their footprint ``layer``
+    # value should not count as a mismatch.  We treat ``thru_hole`` pads
+    # as already-on-the-pour-layer for this gate.
+    layer_count = _detect_layer_count(pcb_to_stitch)
+    try:
+        pads = find_pads_on_nets(sexp, set(plane_nets.keys())) if sexp else []
+    except Exception as exc:
+        logger.debug("Could not probe plane-net pads: %s", exc)
+        pads = []
+
+    needs_stitch = any(
+        pad.pad_type != "thru_hole" and pad.layer != plane_nets.get(pad.net_name)
+        for pad in pads
+    )
+
+    if not needs_stitch:
+        return BuildResult(
+            step="stitch",
+            success=True,
+            message=(
+                "stitch: no cross-layer plane pads — skipped"
+                if pads
+                else "stitch: no plane-net pads found — skipped"
+            ),
             output_file=pcb_to_stitch,
         )
 
--- a/tests/test_build_stitch_step.py
+++ b/tests/test_build_stitch_step.py
@@ -1,4 +1,4 @@
-"""Tests for the build pipeline stitch step (issue #2747).
+"""Tests for the build pipeline stitch step (issue #2747, #3279).
 
 Covers:
 - ``BuildStep.STITCH`` enum membership and value.
@@ -6,7 +6,11 @@ Covers:
 - ``--step`` argparse choice (``stitch``).
 - ``--help`` lists ``stitch`` as a choice.
 - ``_run_step_stitch`` behaviour:
-  * 2-layer board -> skipped with descriptive message.
+  * 2-layer board with no zones -> skipped (no plane nets detected).
+  * 2-layer board where every plane-net pad is already on the pour
+    layer -> skipped (no cross-layer plane pads).  Issue #3279.
+  * 2-layer board with B.Cu pour and F.Cu GND pads -> stitches
+    successfully.  Issue #3279 regression guard.
   * No-PCB -> graceful no-op success.
   * No-plane-nets -> skipped.
   * 4-layer board with plane-net pads -> vias added.
@@ -37,8 +41,11 @@ from kicad_tools.core.sexp_file import load_pcb
 # Fixtures
 # ---------------------------------------------------------------------------
 
-# Minimal 2-layer board: GND + +3.3V nets, NO inner planes.  The stitch
-# step must skip 2-layer boards because there is nothing to stitch to.
+# Minimal 2-layer board: GND + +3.3V nets, NO zones at all.  The stitch
+# step must skip with the "no plane nets detected" branch -- no pour =
+# no work.  (Pre-#3279 this fixture exercised the now-removed
+# layer_count <= 2 blanket skip; the test below documents the new
+# expected message.)
 TWO_LAYER_PCB = """(kicad_pcb
   (version 20240108)
   (generator "test")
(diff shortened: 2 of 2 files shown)


rjwalters:

### Verdict: LGTM

The fix is honest and equivalent-or-better than the #2747 skip it replaces. The original gate used `layer_count <= 2` as a coarse proxy for "no pour to stitch to," which silently broke the 2L cross-layer case. The new probe answers the precise question — "is there any plane-net pad whose layer differs from the pour layer for that net?" — and routes the same 4L behaviour through the unchanged downstream path.

### What I verified

**Probe logic (build_cmd.py:1612-1620)**:
- SMD pads with `pad.layer != plane_nets[pad.net_name]` → stitch runs (correct for 2L F.Cu-pad + B.Cu-pour)
- Through-hole pads excluded via `pad.pad_type != "thru_hole"` — correct, matches the `PadInfo.pad_type` literal (`stitch_cmd.py:149, 713`) and reflects that the plated barrel grants `*.Cu` access
- Pads already on pour layer → clean skip with "no cross-layer plane pads — skipped"
- Empty plane-nets dict → "no plane nets detected" branch fires earlier

**Error handling**: Both probes default-fail to safe values (`pads = []`, `sexp = None`), which combined with `any(...)` yields a graceful no-op skip rather than a crash.

**Skip messages are informative**: distinct strings for the three cases (no plane nets / no cross-layer plane pads / no plane-net pads found).

**Test matrix is complete**:
1. `test_skip_on_two_layer_board_without_pour` — 2L no-zone → "no plane nets" skip (renamed to reflect new semantics)
2. `test_2l_board_with_cross_layer_plane_pads_runs_stitch` — 2L cross-layer → MUST stitch, explicit `"skipped" not in msg_lower` guard (the #3279 regression guard)
3. `test_2l_board_with_no_cross_layer_pads_skips_stitch` — 2L pads-match-pour → "no cross-layer plane pads" skip

**Regression checks**:
- 14 of 14 tests in `tests/test_build_stitch_step.py` pass locally
- 262 of 262 in the related stitch suites (`test_stitch_cmd.py`, `test_stitch_mfr.py`, `test_stitch_bga_outer_ring.py`, `test_board_05_thermal_stitch.py`) pass — no 4L regression
- All 8 CI checks pass (Lint, Type, C++ Build, Test, Routed PCB DRC, kicad-cli round-trip, Diff-Pair Regression 06, Match-Group Regression 07)

**Board-level verification** (per PR body, accepted on trust):
- Board 03: GND stranded 26→5 (remaining 5 are clearance-induced via skips → correctly out-of-scope under #3271)
- Board 02: cleanly skips with new "no cross-layer plane pads" message, MFG-READY DRC still passes
- Idempotent: re-running stitch reports `40 plane pad(s) already connected, no new vias needed`

- `_detect_layer_count(pcb_to_stitch)` is now computed even when the cross-layer-pads gate will skip the step. A small extra disk read; could be deferred below the skip without changing correctness. Not worth blocking on.
- Pre-existing test failures called out in the PR body (`test_route_exit_code_3_is_failure`, `test_real_fixture_voltage_divider_routed_pcb_is_rejected`) are independently confirmed to be unrelated to `_run_step_stitch`.

### Scope discipline

The PR correctly ships the Curator's recommended Option A (drop the blanket 2L skip) and leaves the remaining 5-pad clearance-stitching residue under #3271's pad-aware stitching umbrella. This is exactly the right shipping boundary.

Approving and handing off to `loom:pr`.

Issue #3349, opened 2026-06-08
Schematic validator — multi-unit symbol false positives for unconnected pins


rjwalters:
## Bug summary

`src/kicad_tools/schematic/models/validation_mixin.py` (the unconnected-pin check, around line 290+) treats every unit of a multi-unit symbol as if its pins need their own SymbolInstance to be wired. This generates false-positive "duplicate reference" and "pin not connected" warnings when a single multi-unit symbol (e.g. LM393, MCP6001 dual op-amp, hex inverter ICs) is placed with multiple SymbolInstances sharing the same reference.

## Symptom

PR #3345 (softstart rev B P2) places an LM393 using 3 SymbolInstances (unit 1, unit 2, unit 3 for comparator A, comparator B, and power pins respectively). The validator reports:

- 2 × `Duplicate reference 'U7' at (...)` — both fired because all three U7 instances share the U7 reference (which is correct KiCad practice for multi-unit symbols!)
- Multiple `Pin V- (power_in) on U7 at (...) is not connected` — the power pins on unit 3 ARE wired to the rails, but the validator scans each unit independently

These warnings are noise: KiCad's own ERC handles multi-unit symbols correctly. Our validator should too.

## Recommended fix

1. Treat all SymbolInstances with the same `reference` as belonging to one logical symbol when checking duplicates
2. When scanning for unconnected pins, build the set of pins-actually-wired across *all* SymbolInstances of the same reference, then report each pin only once
3. Add a unit test that places a multi-unit symbol (e.g. LM393) with all units connected, and asserts the validator returns 0 errors

## Scope (optional)

The judge's review of PR #3345 flagged this as optional follow-up; useful for cleanliness but not blocking. Most builders working with multi-unit ICs will hit the same false positives.

## Reproduction

```bash
cd .loom/worktrees/issue-3343
uv run python boards/external/softstart/generate_design.py /tmp/softstart_revb_judge
# Validator output: 13 errors, almost all multi-unit U7 false positives
```

## Acceptance criteria

- [ ] Multi-unit symbols (LM393 etc.) no longer trigger duplicate-reference warnings when units share a reference
- [ ] Unconnected-pin check aggregates pin connectivity across all SymbolInstances of the same reference
- [ ] Unit test covers an LM393 placement with all 3 units present and connected
- [ ] `boards/external/softstart/generate_design.py` regenerates with ~0 spurious U7 warnings

## Related

- PR #3345 (softstart rev B P2)
[...]

Issue #3347, opened 2026-06-08
BackToBackFETPair block — diagonal source-tie wires cause unintended crossings


rjwalters:
## Bug summary

The `BackToBackFETPair` block at `src/kicad_tools/schematic/blocks/motor.py:1556` emits diagonal wires for the source-tie (e.g. `(242.57, 115.57) → (237.49, 144.78)`). KiCad's schematic parser accepts non-orthogonal wires, but they:

1. Create unintended crossings when other rails (rails, sense lines) pass between the two FET sources
2. Make the schematic harder to read and harder to audit visually
3. Defeat the heuristics in our schematic validator (which assumes orthogonal-only wire segments when computing endpoint-to-endpoint connectivity)

## Reproduction

PR #3345 (softstart rev B P2) uses two `BackToBackFETPair` instances (Q1A/Q1B for the positive bank, Q2A/Q2B for the negative bank). The generated schematic has diagonal source-tie segments that visually cross the SRC_POS / SRC_NEG rails as well as the gate drive lines.

Specifically: with the PR's component placement, the Q1A.S ↔ Q1B.S diagonal runs from approximately `(242.57, 115.57)` to `(237.49, 144.78)` — a ~29mm vertical run with a ~5mm horizontal offset, which KiCad renders as a 10° slope.

## Recommended fix

Route the source tie as an orthogonal L (or U / Z) instead of a single diagonal segment. For the canonical layout (both FETs in vertical TO-220 orientation, sources facing each other), the cleanest path is:

- Drop down from Q1A.S to the midpoint Y between the two sources
- Run horizontally to align with Q1B.S X
- Rise to Q1B.S

The block already knows the spacing between FET pair instances, so the routing geometry is deterministic.

## Acceptance criteria

- [ ] `BackToBackFETPair` block emits only orthogonal wires (horizontal + vertical segments)
- [ ] Unit test asserts every wire segment in the block has either `x1 == x2` or `y1 == y2`
- [ ] PR #3345's softstart rev B schematic regenerates with orthogonal source ties (visual confirmation in PR diff)

## Related

- PR #3345 (softstart rev B P2)
- Issue #3343 (softstart rev B parent issue)

PR #3356, opened 2026-06-08
fix(blocks): OvercurrentComparator multi-unit pin lookup (#3346)

### LINKED ISSUE #3346: OvercurrentComparator block — multi-unit symbol pin_position returns phantom positions
## Bug summary

The `OvercurrentComparator` block at `src/kicad_tools/schematic/blocks/analog.py:579-810` returns phantom pin positions when querying `pin_position("4")`, `pin_position("8")`, etc. on a multi-unit symbol (LM393) where the pin actually lives on a different unit.

The block currently assumes pins resolve via a single SymbolInstance, but the LM393 has its power pins on unit 3, comparator A on unit 1, and comparator B on unit 2. Calling `pin_position("4")` on the unit-1 instance returns coordinates derived from the unit-1 outline rather than from the unit-3 SymbolInstance that actually contains pin 4.

## Reproduction

PR #3345 (softstart rev B P2) tried to use this block in `boards/external/softstart/generate_design.py` and was forced to abandon it and hand-place an LM393 with three explicit SymbolInstances (unit 1 = comparator A, unit 2 = comparator B, unit 3 = power pins) — see lines 1058-1167 of that file for the workaround.

Symptoms observed:
- Wire endpoints generated for pin 4 / pin 8 landed in empty space
- Subsequent net-merge bugs (because the phantom endpoints were near foreign labels)

## Recommended fix

Make `OvercurrentComparator` (and the broader block library) resolve `pin_position(pin_number)` by walking *all* SymbolInstances belonging to the multi-unit symbol, not just the primary unit's instance.

Concretely:
1. When the block places a multi-unit symbol, register every unit's SymbolInstance with the symbol's lookup table
2. `pin_position(pin_number)` should locate the SymbolInstance whose pin set includes that pin number, then transform by *that* instance's (x, y, rotation, mirror)
3. Add a unit test that places LM393, queries pin 4 and pin 8, and asserts the returned coordinates match the unit-3 outline — not unit 1

## Acceptance criteria

- [ ] `OvercurrentComparator` block usable without per-board manual LM393 placement
- [ ] Unit test in `tests/` covers multi-unit pin resolution for LM393
- [ ] PR #3345's softstart rev B can be refactored to use the block (follow-up board-side change, not in scope of this issue)

## Related

- PR #3345 (softstart rev B P2, contains the manual workaround)
- Issue #3343 (softstart rev B parent issue)

rjwalters:
## Summary

Fixes the multi-unit `pin_position` bug in `OvercurrentComparator` (issue #3346) by threading the KiCad **unit** through the parser → `SymbolDef` → `SymbolInstance` → block dispatch.

- The LM393 packs three units in one symbol: channel A (pins 1-3, unit 1), channel B (pins 5-7, unit 2), power (pins 4 & 8, unit 3).
- The block used to place only unit 1, then ask the unit-1 instance for pin 4 / pin 8 — which returned a phantom position (unit-3 library offset applied to the unit-1 placement origin).
- Now the block places all three units, owns a `unit -> SymbolInstance` map, and routes every `pin_position(num)` lookup to the instance that actually owns that pin.

Closes #3346.

## Changes

- `Pin` (both `models.pin` and `registry`) gains a `unit` field (default 1).
- Both pin parsers (`SymbolDef._parse_pins_sexp` SExp-based + registry regex-based) detect the enclosing `Name_N_1` unit sub-symbol and tag each pin with its unit number.
- `SymbolDef` adds `unit_count()` and `get_pin_unit(pin_number)` helpers.
- `SymbolInstance.pin_position` prefers in-unit pin matches before falling back to any match (single-unit symbols unaffected).
- `Schematic.add_symbol` gains a `unit=` parameter. `MCUBlock` already passed this kwarg under the assumption it worked; now it does.
- `OvercurrentComparator` places all three LM393 units, owns a unit-instance map, and exposes a `block.pin_position(num)` helper. Channel B pins (5-7) auto-marked NC for ERC quiet.

## Test plan

- [x] `tests/test_blocks_overcurrent_comparator.py` — 9 existing mock tests pass unchanged plus 7 new tests in `TestLM393MultiUnitPlacement` that exercise the real KiCad LM393 library (no mocks) and assert:
  - every pin is tagged with the correct unit (1 → A, 2 → B, 3 → power)
  - `SymbolDef.unit_count()` returns 3 for LM393
  - the block places three `SymbolInstance`s (units 1, 2, 3) with `reference="U7"`
  - `block.pin_position("4")` / `pin_position("8")` resolve via the unit-3 instance (not phantom unit-1 fallback)
  - channel A pins resolve against unit 1's placement origin
  - channel B pins are marked `no_connect`
  - the block-level dispatch returns the same coordinate as querying the unit-3 instance directly
- [x] `tests/test_blocks_*.py` — full block suite (174 tests) passes
- [x] `tests/test_softstart_revb_schematic.py`, `tests/test_library_extends.py`, `tests/test_erc_cross_sheet.py`, `tests/test_board_05_drv8308_pin_nets.py` — all pass
- [x] Ruff clean on all touched files
- [ ] Follow-up: softstart rev B can retire the manual LM393 placement workaround at `boards/external/softstart/generate_design.py:1058-1167` (separate board-side change, out of scope here)

## Out of scope

- `BackToBackFETPair` (`motor.py`) multi-unit fix — issue #3347
- Validator multi-unit handling — issue #3349

### DIFF
--- a/src/kicad_tools/schematic/blocks/analog.py
+++ b/src/kicad_tools/schematic/blocks/analog.py
@@ -1,11 +1,12 @@
 """Analog circuit blocks: op-amps, ADC input filters, sensor interfaces."""
 
+import contextlib
 from typing import TYPE_CHECKING, Literal
 
 from .base import CircuitBlock
 
 if TYPE_CHECKING:
-    from kicad_sch_helper import Schematic
+    from kicad_sch_helper import Schematic, SymbolInstance  # noqa: F401
 
 
 class ADCInputFilterBlock(CircuitBlock):
@@ -677,7 +678,28 @@ class OvercurrentComparator(CircuitBlock):
         """
         super().__init__(sch, x, y)
 
-        # Place the comparator IC
+        # Place the comparator IC.  LM393 is a multi-unit symbol:
+        #   * unit 1 = comparator channel A (pins 1=OUT, 2=IN-, 3=IN+)
+        #   * unit 2 = comparator channel B (pins 5=IN+, 6=IN-, 7=OUT)
+        #   * unit 3 = power pins (4=V-, 8=V+)
+        # If we only placed unit 1 then ``pin_position("4")`` and
+        # ``pin_position("8")`` would silently return phantom positions
+        # derived from the unit-1 outline (the LM393 library file stores
+        # pin 4 at (-2.54, -7.62) inside its ``_3_1`` sub-symbol -- that
+        # offset is meaningless when applied to a unit-1 placement and
+        # leads to wires terminating in empty space (issue #3346).
+        #
+        # Place all three units; each ``SymbolInstance`` resolves
+        # ``pin_position`` against its own unit's outline.  We then build
+        # an internal pin-number -> instance map so the rest of this
+        # block (and downstream callers) can keep calling
+        # ``pin_position("4")`` and get the correct unit-3 coordinates.
+        # ``SymbolInstance`` is only imported under ``TYPE_CHECKING`` so the
+        # annotation must remain a string at runtime.  ``ruff`` flags this
+        # as ``UP037`` but the suggested fix breaks Python <3.10 evaluation.
+        self._unit_instances: "dict[int, SymbolInstance]" = {}  # noqa: UP037
+
+        primary_unit = 1
         self.comparator = sch.add_symbol(
             comparator_symbol,
             x,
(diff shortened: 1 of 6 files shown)


rjwalters:
## Loom Judge Review — APPROVE

Verified the multi-unit threading end-to-end against the real KiCad LM393 library. The fix is clean, well-scoped, and back-compatible.

### What was verified

**Pin.unit field consistency** — both `models.pin.Pin` and `registry.Pin` carry the `unit: int = 1` field. `SymbolDef.from_library`'s registry-cached path uses `getattr(p, "unit", 1)` for safe migration if a stale cached Pin lacks the field.

**Both parsers extract unit correctly**
- SExp parser (`SymbolDef._parse_pins_sexp`) recurses into child `(symbol ...)` blocks, regex-matches `Name_N_1` on the sub-symbol name, and propagates the unit number down to nested pin nodes. Graphical body sub-symbols (`Name_0_1`) are correctly filtered (`extracted > 0`).
- Registry regex parser (`_build_pin_unit_map`) walks unit sub-symbol headers with parenthesis-depth counting (string-aware, escape-aware) and attributes each pin number to the enclosing unit. Same `unit_num <= 0` guard for the graphical body.

**SymbolInstance.pin_position back-compat preserved**
- Prefers in-unit match first (`p.unit == self.unit`), then falls back to any matching pin. For single-unit symbols (every pin has `unit=1`, instance has `unit=1`) the exact-match path fires on the first pass — zero behavioural change.
- The fallback preserves the old phantom-position behaviour for callers that place a multi-unit symbol with only one unit and query an off-unit pin number. That's intentional: it's strictly no-worse-than-main, and the fix relies on multi-unit consumers placing all units (which the block now does).

**Schematic.add_symbol(unit=1) default is safe** — no existing single-unit caller passes the kwarg, and the only existing user, `MCUBlock` line 111 (`unit=unit`), previously would have raised `TypeError` for any `unit != 1`. The new signature retroactively makes that call site work as the author originally intended.

**OvercurrentComparator three-unit placement** — explicitly iterates `range(2, unit_count + 1)` with hand-tuned offsets for units 2 (channel B, NC pins) and 3 (power). Builds `_unit_instances: dict[int, SymbolInstance]` and routes every `pin_position` lookup through `_unit_pin_position`, which uses `SymbolDef.get_pin_unit(pin)` to dispatch to the owning unit's instance. Includes a sane fallback when run against unit-test mocks that don't expose a real `SymbolDef`.

**Tests — 7 new tests use the real KiCad library, no mocks:**
- `test_pin_unit_metadata_is_propagated_through_symbol_def` — asserts every LM393 pin's `unit` is correctly tagged.
- `test_unit_count_reports_three_for_lm393` — `SymbolDef.unit_count() == 3`.
- `test_block_places_three_unit_instances_for_lm393` — three `SymbolInstance`s, refs all `U7`, units 1/2/3.
- `test_power_pin_position_is_resolved_against_unit_3` — pin 4 / pin 8 return coordinates relative to the unit-3 instance with the correct (-2.54, ±7.62) library offset.
- `test_channel_a_pins_resolve_against_unit_1` — back-compat for in-unit pins.
- `test_channel_b_pins_marked_no_connect` — ERC quiet.
- `test_pin_position_no_longer_returns_phantom_unit1_for_pin_4` — direct regression for the original bug; block-level dispatch equals unit-3 instance's own resolution.

### Local validation

- `pytest tests/test_blocks_overcurrent_comparator.py` — 16/16 pass (9 existing mock tests + 7 new real-library tests).
- `pytest tests/test_blocks_*.py` — full block suite, 174/174 pass.
- `pytest tests/test_softstart_revb_schematic.py tests/test_library_extends.py tests/test_erc_cross_sheet.py tests/test_board_05_drv8308_pin_nets.py` — 46/46 pass.
- Confirmed the one ruff `SIM110` finding on `_point_on_any_wire` is pre-existing on main (unrelated to this PR).

### CI

All 8 checks green: Lint & Format, Type Check, Test, C++ Build, Routed PCB DRC, Diff-Pair Routing, Match-Group Routing, kicad-cli Round-trip Smoke. `mergeStateStatus = CLEAN`.

### Foundational risk assessment

This change threads through `Pin` + `SymbolDef` + `SymbolInstance` + `Schematic.add_symbol`. The risk surface is real — but the back-compat story holds because:
1. `Pin.unit` defaults to 1; pins from non-multi-unit symbols stay at 1.
2. `SymbolInstance.pin_position` falls back to "any matching pin" when no in-unit match exists, exactly preserving the old behaviour for single-unit symbols and for multi-unit symbols that only have one unit placed.
3. `add_symbol`'s new `unit` kwarg defaults to 1; no existing call site is forced to change.
4. Adjacent regression suites (softstart rev B, library extends, ERC cross-sheet, board-05 DRV8308) all pass.

Foundational change, but cleanly bounded. Ship it.

PR #3258, opened 2026-06-07
test(board-05): pin clearance_pad_segment hot-spot shape (Refs #3251)

rjwalters:
## Summary

Adds `tests/test_board_05_drc_hotspot_regression.py` to pin the *shape* of board-05's residual `clearance_pad_segment` violations (count + hot-spot refs + shortfall band).  No production code changed.

Closes #3251 with a regression test only — the investigation found that board 05's committed routed PCB **already passes the CI floor** (6 errors < allowlist of 9) and the remaining mechanisms are tracked sufficiently by the existing PR cluster (#3232 / #3248 / #3249 / #3250).

### Measurement table

| Configuration | Total errs | clearance_pad_segment | Nets routed | Note |
|---|---|---|---|---|
| Committed `bldc_controller_routed.kicad_pcb` (HEAD) | 29 (6 blocking + 23 advisory) | **6** | partial (~15 connectivity errors) | Below CI floor of 9 — passes |
| Fresh `--backend python` re-route (matches `design.py`) | 32 (11 + 21 advisory) | 4 | 15/32 (47%) | Re-routes vary; committed file is a better-than-average snapshot |
| Fresh `--backend cpp` re-route (jlcpcb) | 52 (36 + 16 advisory) | **23** | 20/32 (62%) | More reach, more violations — design.py pins python for this reason (PR #3222 / Issue #3221) |

The task brief described 10 U3 LQFP-48 violations, but board 05 actually uses:
- **U3 = HTSSOP-56** (Texas Instruments DRV8301, 0.5mm pitch) — NOT LQFP-48
- **U10 = LQFP-32** (STM32G431, 0.8mm pitch)

The cpp-backend re-route concentrates 23 of 23 `clearance_pad_segment` at U3.

### U3 violation taxonomy (cpp backend)

| Shortfall band | Count | Mechanism |
|---|---|---|
| 127µm (`actual=0`) | 8 (all at U3-33 ISENSE_A+ vs ISENSE_A-) | Trace centerlines on pad metal — A* pad_blocked short-circuit bypassed by escape router force-emit or post-route nudge |
| 27.0µm (`actual=100µm`) | 10 | Initial hypothesis: grid quantization. **Disproven**: persists at 0.05mm grid. Actual mechanism: in-pad rescue's "Proceeding anyway" path |
| 13-25µm (`actual=102-114µm`) | 3 | Sub-cell residuals |
[...]

PR #3375, opened 2026-06-08
feat(router): fine-pitch escape autorouter integration (P_FP3) Refs #3371

rjwalters:
## Summary

Phase 3 (P_FP3) of the fine-pitch escape ladder for Issue #3371.  Wires the P_FP2 detector + clearance resolver into ``load_pcb_for_routing`` and the autorouter's per-pad halo so the fine-pitch escape regions auto-detected at PCB load time land at **both** the Python pathfinder's grid halo AND the C++ pad-segment validator's clearance source.  This is the integration phase: the data structures (P_FP1) and the detector + resolver (P_FP2) are now wired into the actual routing pipeline.

## Pipeline integration

- **``load_pcb_for_routing``** runs ``detect_fine_pitch_regions`` against the parsed component dicts BEFORE adding pads to the router's grid.  This ensures the per-pad ``pad_blocked`` envelope is computed with the in-region escape clearance (when applicable) rather than the wider standard clearance.  Manufacturer-aware via ``rules.manufacturer``; falls back to no-shrink when no manufacturer is configured.

- **``RoutingGrid._clearance_for_pin_pitch``** now takes an optional ``pad`` argument.  When the pad sits inside an installed fine-pitch region (per ``FinePitchRegion.applies_to_pad``), the halo shrinks from ``trace_clearance + trace_width/2`` to ``region.escape_clearance + trace_width/2``.  Backward-compat: ``pad=None`` callers get byte-for-byte identical pre-#3371 behaviour.

- **Narrow-channel guard** (Issue #2867 equivalent for the region path) applies at both the grid halo and the C++ resolver (``resolve_clearance_with_escape_region``): when the candidate escape clearance would produce a corridor that cannot host a trace at full clearance, the helper falls back to the standard ``get_clearance_for_component`` value so the validator rejects geometrically infeasible through-channel routes.  Without this guard the LQFP-32 0.8mm-pitch 0.4mm-pad case (STM32 on softstart) would let A* commit to traces between pins that DRC rejects.

- **All five ``load_pcb_for_routing`` call sites in ``route_cmd.py``** (single-pass + layer/rule/manufacturer/size escalation wrappers) emit a one-line summary of the detected regions and warn when no manufacturer is configured (P_FP2 builder decision #6).

## Detector filter (≥4 pads)

[...]

PR #3377, opened 2026-06-09
fix(sync): /-prefix strip + unconditional pin-to-pad map (Closes #3370)

### LINKED ISSUE #3370: board 05: schematic-PCB structural drift — 14 nets in schematic absent from synced PCB
## Summary

Per PR #3367 (board 05 verification, 2026-06-08), `kct fleet status` and the sync analyzer report **51 nets in schematic, 40 nets in PCB**. The 14-net gap is structural, not stale-routing — a fresh re-route via the canonical recipe reproduces the same 51/40 count.

This is a **schematic-PCB structural drift**, NOT a routing problem. It needs the schematic and PCB to be reconciled, separately from any artifact refresh.

## Drift nets (in schematic, missing in PCB)

The schematic carries these 14 nets that don't appear in the synced PCB:
- BST_A, BST_B, BST_C (boost-converter switch nodes)
- DRV_CP1, DRV_CP2, DRV_FAULTn, DRV_LOCKn, DRV_VCP, DRV_VINT, DRV_VREG, DRV_VSW (DRV8301 gate-driver internal nets)
- FGFB, FGOUT (PWM feedback nets)
- SPI_MISO (SPI controller-side data)

## Why it matters

- `kct fleet status` reports `schematic drift` for board 05, blocking ship-ready classification (#3289 / #3322 drift detector behavior)
- Fresh re-routes reproduce 51/40 count, so it's not a stale-snapshot issue
- The committed PCB (1 net less than current schematic) is what passes the existing #3258 hot-spot regression test

---

### Reproduction recipe

```bash
# Confirm the drift detector still fires
uv run kct fleet status | grep "^05"
# expect: "routed PCB stale (schematic drift: 51 nets in schematic, 40 in PCB)"

# Inspect what the netlist exporter sees vs the PCB
uv run python - <<'PY'
from kicad_tools.operations.netlist import export_netlist

netlist = export_netlist('boards/05-bldc-motor-controller/output/bldc_controller.kicad_sch')
pcb = PCB.load('boards/05-bldc-motor-controller/output/bldc_controller_routed.kicad_pcb')

target = {'BST_A','BST_B','BST_C','DRV_CP1','DRV_CP2','DRV_FAULTn','DRV_LOCKn',
          'DRV_VCP','DRV_VINT','DRV_VREG','DRV_VSW','FGFB','FGOUT','SPI_MISO'}
nl_names = {n.name for n in netlist.nets}
pcb_names = {n.name for n in pcb.nets.values() if n.name}

print('In netlist as bare name:', sorted(target & nl_names))                # []
print('In netlist as /-prefixed:', sorted(target & {n.lstrip('/') for n in nl_names if n.startswith('/')}))
print('In PCB net table:', sorted(target & pcb_names))                      # []
# U3 pads should connect to BST_A/DRV_*/FGFB/FGOUT/SPI_MISO. What do they actually show?
u3 = pcb.get_footprint('U3')
for pad_num in ('1', '3', '5', '15', '16', '17', '18', '24', '25', '28', '29', '30'):
    for p in u3.pads:
        if p.number == pad_num:
            print(f'  U3.{pad_num}: net={p.net_name!r}')
            break
PY
```

Observed output (2026-06-08, HEAD = 5e115a45):

```
In netlist as bare name: []
In netlist as /-prefixed: ['BST_A','BST_B','BST_C','DRV_CP1','DRV_CP2','DRV_FAULTn',
                           'DRV_LOCKn','DRV_VCP','DRV_VINT','DRV_VREG','DRV_VSW',
                           'FGFB','FGOUT','SPI_MISO']
In PCB net table: []
  U3.1:  net='GND'          # schematic says BST_A (UHP pin)
  U3.3:  net='+5V'          # schematic says BST_B (VHP pin)
  U3.5:  net='+3.3V'        # schematic says BST_C (WHP pin)
  U3.15: net='+5V'          # schematic says SPI_MISO (SDATAO pin)
  U3.17: net='PWM_AH'       # schematic says DRV_FAULTn (~FAULTn pin)
  ...
```

### File:line references

- **Netlist exporter** — `src/kicad_tools/operations/netlist.py:809-918` (`export_netlist`).
  Wraps `kicad-cli sch export netlist --format kicadsexpr`. The cli output prefixes
  every **local** label with the sheet path `/`, producing `/BST_A`, `/DRV_CP1`, …
  whereas **global power labels** like `+3V3` / `+24V` / `+5V` / `GND` come through
  unprefixed.
- **Sync net-add path** — `src/kicad_tools/cli/pcb_sync_netlist.py:439-449`
  `_assign_nets_from_schematic()` calls `pcb.add_net(net.name)` for every name in
  the exported netlist, then `pcb.assign_nets_from_netlist(netlist, pin_to_pad_map=…)`.
  No name normalization is applied — `/BST_A` would be added verbatim if it reached
  this code.
- **Pad assignment** — `src/kicad_tools/schema/pcb.py:4240-4318`
  `assign_nets_from_netlist()` iterates `(ref, pin) -> net.name`. The netlist nodes
  for `/BST_A` are `('C12','1')` and `('U3','1')` — but `U3.1` ends up on `GND`
  in the committed PCB (see reproduction output above), proving the assignment
  the schematic intends was **never written** by the sync that produced this PCB.
- **Drift detector** — `src/kicad_tools/cli/fleet_cmd.py:592-740`
  `_extract_schematic_nets()` uses a **regex over s-expression `label` /
  `global_label` / `hierarchical_label` leaves** (line 588). It produces bare names
  (`BST_A`), skipping `Net-(…)` placeholders. `_extract_pcb_named_nets()` (line 619)
  pulls bare names from `(net N "name")`. The added-only tolerance
  (`_DRIFT_ADDED_ONLY_TOLERANCE = 2`, line 69) tolerates ≤ 2 unmatched PCB names
  with **no removals**; here we have **14 removals**, so the gate fires correctly.
- **Reconciler** — `src/kicad_tools/sync/reconciler.py:292-509`
  Reconciles **components** (footprint references), not nets. It does not address
  this class of drift; net surgery happens in `pcb_sync_netlist.py`.

### Three hypotheses

**H1 — Net-name namespace mismatch from `kicad-cli` (most likely root cause).**
The kicad-cli netlist exporter writes hierarchical local labels with a sheet-path
prefix (`/BST_A`). Power symbols (`power:+3V3`, `power:GND`) come out unprefixed.
`pcb_sync_netlist._assign_nets_from_schematic()` (line 439-449) passes those names
straight through to `pcb.add_net()` and `assign_nets_from_netlist()`. The committed
routed PCB shows the bare-named labels (`+5V`, `GND`, `PWM_AH`) were preserved from
an earlier sync round and the `/`-prefixed adds **never landed** (likely
`_remove_unused_nets` cleared them because no pad got assigned to them — see H2).
Evidence: every "missing" net appears in the exported netlist with a `/`-prefix
and **only** these 14; none of the matched-on-PCB nets carry the prefix.

**H2 — U3 pin-to-pad map is wrong, so net assignment silently no-ops.**
`pcb_sync_netlist._assign_nets_from_schematic()` builds a `pin_to_pad_map` via
`build_pin_to_pad_map` only when the netlist comes from the "Python fallback"
path (line 428-434). When kicad-cli succeeds (this case, `tool = 'Eeschema 10.0.1'`),
no mapping is built — pin numbers go through as pad numbers. The committed U3 pad
list above shows pad 1 → `GND`, pad 17 → `PWM_AH`, pad 3 → `+5V` — which is
consistent with an earlier sync round that used a **different DRV8301 footprint
pinout** than the current schematic symbol's pin numbering. If U3's HTSSOP-56
footprint pad-number-to-pin-function table doesn't match the schematic symbol's
pin numbering, every `BST_A`-class assignment misses the intended pad and lands on
a different (or no) pad. Note that `DRV8301_PINS` in
`boards/05-bldc-motor-controller/design.py:1545+` is the **authoritative**
pad-number → net-name table for the bottom-up footprint generator, and it tells
the same story (pad 48 → `VMOTOR`, but listed as `# BST_A` in the comment at
line 1615 — that comment IS the bug, or alternatively the comment is right and
the assignment is wrong).

**H3 — DRV8301 internal nets are intentionally hidden on PCB (least likely).**
The design.py comment at line 605-628 explicitly states these 14 nets are
"residual pins each given a unique local net" specifically to **avoid ERC
`pin_to_pin` Output-Output errors** — they are deliberate schematic-only nets.
But the schematic does emit physical labels for all of them (verified — see
the s-expression `(label "BST_A" …)` rows in the schematic), and `BST_A/B/C` in
particular connect to **real bootstrap capacitors C12/C13/C14**. The hide-on-PCB
read is inconsistent with `BST_A` being a 2-pad net (`U3.1` + `C12.1`).

### Recommended fix path

1. **Add a strip-leading-slash normalization step** in
   `pcb_sync_netlist._assign_nets_from_schematic()` (line ~440) before
   `pcb.add_net(net.name)`. The kicad-cli convention of prefixing root-sheet
   local labels with `/` is a serialization artifact; PCB nets are conventionally
   bare. Mirror what `_extract_schematic_nets()` does in `fleet_cmd.py`
   (it works on labels, which are already bare). Suggested rule:
   strip exactly one leading `/` from any non-power net name that has at least
   one node — power rails (`+3V3`, `GND`) come through unprefixed already and
   need no change.
2. **Verify the U3 pin-to-pad numbering**: cross-check
   `boards/05-bldc-motor-controller/design.py:DRV8301_PINS` (line 1545+) against
   the schematic symbol pin numbers used by `pin_nets={…}` (line 647-714).
   If the schematic uses pin name `"UHP"` and the symbol resolves that to pin
   number X, the footprint's pad X must be the chip's UHP pin.
3. **Round-trip test**: after the fix, `kct pcb sync-netlist
   --remove-orphan-nets` against the schematic should bring the PCB's named-net
   count from 40 to **at least 53** (the 40 currently present plus 13 of the 14
   drift nets; the 14th is `+3V3`/`+3.3V` canonicalization which is already
   handled by `canonicalize_power_nets`).
4. **Regression**: add a test that exports the board 05 netlist, applies sync,
   and asserts the resulting PCB has no `/`-prefixed net names AND that every
   bare schematic label corresponds to a bare PCB net.

### Why the existing tests don't catch this

- The `consistency.SchematicPCBChecker` (used by `sync/reconciler.py:303`)
  compares **footprint reference sets**, not net-name sets, so a 14-net gap is
  invisible to it.
- The drift detector in `fleet_cmd.py` is a *symptom* gate, not a *cause*
  gate — it correctly reports drift but doesn't run the netlist exporter or
  call out the `/`-prefix convention mismatch.
- The hot-spot regression test from PR #3258 is a clearance-rule snapshot, not
  a connectivity check, so it remains green even with 14 disconnected nets.

## Hypothesis (original)

Either:
- (a) Schematic has extra nets that should be added to the PCB layout (forward-annotation gap)
- (b) PCB has nets aliased differently than schematic expects (sync convention drift)
- (c) DRV8301's internal nets are intentionally hidden in PCB representation (footprint convention)

**Curator update**: H1 (kicad-cli `/`-prefix sync convention) + H2 (U3 pad-numbering)
is the combined root cause. (c) is contradicted by the BST_A/B/C nets which
physically connect bootstrap caps C12/C13/C14 to U3 — they cannot be hidden.

## Acceptance criteria

- [ ] `kct sync --analyze` reports clean board 05 (no drift)
- [ ] `kct fleet status` no longer reports "schematic drift" for board 05
- [ ] Schematic and PCB net counts match
- [ ] Existing PR #3258 hot-spot regression test still passes (committed PCB invariant)
- [ ] No regression on python-pinned baseline (board 05 pin per #3222)
- [ ] New regression test asserts no `/`-prefixed net names appear in the synced PCB

## Out of scope

- The 6 `clearance_pad_segment` violations on committed PCB (tracked elsewhere as "residual", not a sync issue)
- The python pin itself (per closed #3221)
- C++ backend gap (closed by #3226 as superseded)

## References

- PR #3367 — verification (this issue is the follow-up that PR's framing called for)
- PR #3289 / #3322 — drift detector
- PR #3258 — hot-spot regression test
- PR #3222 — python pin restored, floor 37→9
- Closed #3221, #3226 — related but distinct

rjwalters:
## Summary

Two coupled fixes to `kct pcb sync-netlist` so the kicad-cli netlist exporter produces PCB nets that round-trip cleanly through pad assignment, plus drift-detector symmetry improvements that surface the same set of "real" nets from both sides.

### Fix 1 — strip leading `/` from kicad-cli net names
`kicad-cli sch export netlist --format kicadsexpr` prefixes hierarchical local labels with the sheet path (`/BST_A`, `/DRV_CP1`, ...), while global power-symbol labels come through unprefixed (`+3V3`, `GND`). `_assign_nets_from_schematic` was passing names verbatim to `pcb.add_net()`, producing ghost `/`-form nets that `_remove_unused_nets` then dropped. Board 05 lost its 14 DRV8301 internal nets (`BST_A` / `DRV_CP*` / `FGFB` / `SPI_MISO` etc) this way. A new `_normalize_kicad_cli_net_names` helper strips a single leading `/` in-place; the Python fallback path is unaffected.

### Fix 2 — build pin-to-pad map for the kicad-cli path too
`build_pin_to_pad_map` was previously called only when the netlist came from the "Python fallback" path, on the assumption that kicad-cli resolves pin-vs-pad numbering internally. Empirically it does not — it emits schematic *symbol* pin numbers, identical to the fallback. We now build the map unconditionally; the result is identity (no-op) for the common case and diverges only for genuine name-based-pad footprints.

### Drift detector symmetry
`_extract_schematic_nets` and `_extract_pcb_named_nets` in `fleet_cmd` now agree on what counts as a "real" net:

- **Schematic side** now picks up implicit globals published by `power:` symbol instances (e.g. `power:+24V` → `+24V`) — these never appear as `(label ...)` leaves but are first-class nets. `PWR_FLAG` is excluded (it's a KiCad ERC convention marker, not a net publisher).
- **PCB side** now returns only nets actually attached to at least one pad. Top-level `(net N "X")` declarations that survive only because a segment / via / zone references the name (routing-only leftover from an older sync round) are dropped. Auto-generated `Net-(REF-PadN)` and `unconnected-(...)` names are also dropped on both sides for symmetry.

## Round-trip verification

Board 05 (in-memory; the committed routed PCB is intentionally preserved):

| Metric | Before fix | After fix |
|---|---|---|
| Total named nets on synced PCB | 40 | 69 |
| Drift nets recovered | 0 / 14 | 14 / 14 |
| `/`-prefixed net names | 14 ghost | 0 |
| `U3.1` assignment | `GND` (stale) | `BST_A` (sch-correct) |
| `U3.5` assignment | `+3.3V` (stale) | `BST_C` (sch-correct) |
| `U3.15` assignment | `+5V` (stale) | `SPI_MISO` (sch-correct) |
| `U3.17` assignment | `PWM_AH` (stale) | `DRV_FAULTn` (sch-correct) |

Boards 01–04, 06–07: sync runs cleanly with zero errors and zero `/`-prefixed net names on every board.

## Why the committed routed PCB is not updated

Applying sync to `boards/05-bldc-motor-controller/output/bldc_controller_routed.kicad_pcb` regresses U10 (STM32G431K8) pad assignments because of a *separate* schematic-emission bug in `design.py`: the SWDIO / SWCLK / ISENSE_B- / ISENSE_C- labels for U10.23 / U10.24 / U10.6 / U10.7 don't wire through the kicad-cli netlister (the wire stubs from `_connect_mcu_pin_to_label` collide with adjacent power rails, silently bridging the net before the label can take effect — visible in board-05 fresh-build warnings).

Updating the synced PCB would push the routed-PCB DRC error count from 6 (allowlist max) to 73, breaking `tests/test_board_05_drc_allowlist.py`. The acceptance criterion "Existing PR #3258 hot-spot regression test still passes (committed PCB invariant)" takes precedence over "fleet status no longer reports drift". The fleet status drift on board 05 will resolve once the schematic-emission bug is fixed in a follow-up.

## Tests

- 7 unit tests for `_normalize_kicad_cli_net_names` (single, power-rail, bare, empty, deeper-path, mixed, idempotency)
- 4 board-05 round-trip regression tests (drift nets present, net count ≥ 53, no `/`-prefixed names, U3 pad assignments)
- 9 unit tests for the drift-detector primitives (label / power symbol / `PWR_FLAG` / placeholder / pad-attached / pad-orphan / auto-generated / unreadable)

All 40 `tests/test_fleet_status.py` pass; all 254 `tests/test_pcb_sync*` and related sync tests pass; all 19 `tests/test_board_05_*` non-export/buck/thermal tests pass.

## Test plan

- [x] `_normalize_kicad_cli_net_names` unit tests (7/7)
- [x] Board-05 round-trip regression (4/4)
- [x] Drift detector primitives (9/9)
- [x] `tests/test_fleet_status.py` (40/40)
- [x] `tests/test_pcb_sync_netlist.py`, `test_netlist_sync.py`, `test_sync_*` (397/397)
- [x] `tests/test_board_05_*` (DRC allowlist, hotspot, drift regression, drv8308 pin nets, routing regression, zones) — 19/19
- [x] Sync sweep on boards 01-04, 06-07 (all clean, 0 errors, 0 `/`-prefixed names)

Closes #3370

### DIFF
--- a/src/kicad_tools/cli/fleet_cmd.py
+++ b/src/kicad_tools/cli/fleet_cmd.py
@@ -586,7 +586,12 @@ def _detect_erc(board_dir: Path) -> ERCStatus:
 
 
 _SCH_LABEL_RE = re.compile(r'\(\s*(?:label|global_label|hierarchical_label)\s+"([^"]+)"')
-_PCB_NET_RE = re.compile(r'\(net\s+\d+\s+"([^"]+)"\)')
+_SCH_POWER_SYMBOL_RE = re.compile(r'\(\s*lib_id\s+"power:([^"]+)"')
+# Match ``(net NUMBER "NAME")`` -- shared between top-level
+# declarations and pad-nested references.  Pad context is
+# distinguished by the streaming parser in
+# ``_extract_pcb_pad_net_numbers``.
+_PCB_NET_NUMBERED_RE = re.compile(r'\(net\s+(\d+)\s+"([^"]+)"')
 
 
 def _extract_schematic_nets(sch_path: Path) -> set[str] | None:
@@ -594,9 +599,19 @@ def _extract_schematic_nets(sch_path: Path) -> set[str] | None:
 
     The detection is intentionally lightweight: we only collect text from
     ``label`` / ``global_label`` / ``hierarchical_label`` s-expression
-    leaves. Auto-generated ``Net-(...)`` placeholders (KiCad's default
-    unconnected names) are skipped because they cannot be reliably
-    mapped to a PCB net name without a full netlister pass.
+    leaves, plus implicit globals published by ``power:`` symbol
+    instances (e.g. ``power:+24V`` -> ``+24V``).  Auto-generated
+    ``Net-(...)`` placeholders (KiCad's default unconnected names) are
+    skipped because they cannot be reliably mapped to a PCB net name
+    without a full netlister pass.
+
+    Power-symbol detection (issue #3370): KiCad's ``power:`` library
+    publishes a single global label per symbol instance equal to the
+    symbol's name (the part after the ``:``).  These globals never
+    appear as ``(label "...")`` leaves in the schematic file even
+    though they are first-class nets.  Picking them up here keeps the
+    drift comparison symmetric with ``_extract_pcb_named_nets``,
+    which sees the same globals as bare ``(net N "name")`` entries.
 
     Returns ``None`` if the file cannot be read so callers can treat
     that as "unknown source state" and skip drift gating.
@@ -613,24 +628,131 @@ def _extract_schematic_nets(sch_path: Path) -> set[str] | None:
         if name.startswith("Net-("):
             continue
         labels.add(name)
+    # Power symbols (``power:+24V``, ``power:GND``, ``power:+3V3``, ...)
+    # publish an implicit global label whose name is the part after
+    # the colon.  Include them so the drift comparison treats power
+    # rails consistently with the PCB-side ``(net N "+24V")`` entries.
+    #
+    # ``PWR_FLAG`` is a KiCad ERC convention symbol -- it marks an
+    # existing net as actively driven so the connectivity checker
+    # doesn't flag it as floating -- and does NOT publish a net of
+    # its own.  Excluding it keeps boards that use it (e.g. board 02)
+    # from showing a spurious one-net drift after this change.
+    for m in _SCH_POWER_SYMBOL_RE.finditer(text):
+        name = m.group(1)
+        if name and name != "PWR_FLAG":
+            labels.add(name)
     return labels
 
 
-def _extract_pcb_named_nets(pcb_path: Path) -> set[str] | None:
-    """Return the set of non-empty named nets declared in a KiCad PCB.
+def _extract_pcb_pad_net_numbers(pcb_text: str) -> set[int]:
+    """Return the set of net numbers actually referenced by pads.
 
-    Only top-level ``(net N "name")`` declarations are matched; the
-    empty net 0 is dropped. Returns ``None`` on read failure.
+    Pads in KiCad PCB files appear as ``(pad "NUM" TYPE SHAPE ...
+    (net N "NAME") ...)`` blocks nested inside ``(footprint ...)``.
+    Top-level ``(net N "NAME")`` declarations exist for *every* known
+    net even when no pad references the net -- routing-only artifacts
+    (segments / vias / zones left over from older sync rounds) keep
+    their declarations after their pads have been reassigned.
+
+    Returning the *number* set rather than the *name* set lets the
+    caller cross-reference with top-level declarations to determine
+    which named nets are pad-referenced and which are routing-only.
+
+    Uses a depth-tracked streaming scan rather than a single regex
+    because pad bodies can span many lines and contain nested forms.
+    """
+    pad_net_numbers: set[int] = set()
+    i = 0
+    n = len(pcb_text)
+    while True:
+        idx = pcb_text.find("(pad ", i)
+        if idx < 0:
+            idx = pcb_text.find("(pad\t", i)
+        if idx < 0:
+            idx = pcb_text.find("(pad\n", i)
+        if idx < 0:
+            break
+        # Find the end of this pad's s-expression by tracking depth.
+        depth = 0
+        j = idx
+        body_start = idx
+        body_end = idx
+        while j < n:
+            c = pcb_text[j]
+            if c == '(':
+                depth += 1
+            elif c == ')':
+                depth -= 1
+                if depth == 0:
+                    body_end = j + 1
+                    break
+            j += 1
+        if body_end <= body_start:
+            i = idx + 5
+            continue
+        body = pcb_text[body_start:body_end]
+        # Match (net N "name") inside the pad body
+        m = _PCB_NET_NUMBERED_RE.search(body)
+        if m:
+            try:
+                pad_net_numbers.add(int(m.group(1)))
+            except ValueError:
+                pass
+        i = body_end
+    return pad_net_numbers
+
+
+def _extract_pcb_named_nets(pcb_path: Path) -> set[str] | None:
+    """Return the set of non-empty PCB nets that have at least one pad.
+
+    The drift detector compares this set against the schematic's
+    label / power-symbol set.  To keep the comparison meaningful, we
+    return only nets that are actually attached to pads -- i.e. the
+    nets the synchronisation layer is responsible for.  Top-level
+    ``(net N "name")`` declarations that exist *only* because a
+    segment / via / zone still references the name (a routing-only
+    leftover from an older sync round) are dropped.
+
+    Two additional name-pattern filters keep the comparison
+    symmetric with :func:`_extract_schematic_nets` (which already
+    skips ``Net-(...)`` placeholders):
+
+      * ``Net-(REF-Pad/PinN)`` -- the netlist exporter's default name
+        for un-labeled local nets.  These appear on the PCB after
+        ``sync-netlist`` runs against a schematic whose nets carry no
+        explicit ``label`` element, so they cannot represent genuine
+        schematic-vs-PCB drift -- only sync state.
+      * ``unconnected-(REF-PinName-PadN)`` -- KiCad's marker for
+        explicitly unconnected pins.  These are not nets the schematic
+        author labeled, so they are pure sync artifacts as well.
+
+    Issue #3370: without the pad-attachment filter, board 05's stale
+    ``PWR_LED`` / ``STATUS_LED`` / ``SW_OUT`` segment leftovers from
+    an earlier sync round would surface as drift even though they
+    have no pads.
+
+    Returns ``None`` on read failure.
     """
     try:
         text = pcb_path.read_text()
     except OSError:
         return None
+    pad_net_numbers = _extract_pcb_pad_net_numbers(text)
     nets: set[str] = set()
-    for m in _PCB_NET_RE.finditer(text):
-        name = m.group(1)
-        if name:
-            nets.add(name)
+    for m in _PCB_NET_NUMBERED_RE.finditer(text):
+        try:
+            num = int(m.group(1))
+        except ValueError:
+            continue
+        if num not in pad_net_numbers:
+            continue
+        name = m.group(2)
+        if not name:
+            continue
+        if name.startswith("Net-(") or name.startswith("unconnected-("):
+            continue
+        nets.add(name)
     return nets
 
 
--- a/src/kicad_tools/cli/pcb_sync_netlist.py
+++ b/src/kicad_tools/cli/pcb_sync_netlist.py
@@ -391,6 +391,42 @@ def _get_pad_net_snapshot(pcb) -> dict[tuple[str, str], str]:
     return snapshot
 
 
+def _normalize_kicad_cli_net_names(netlist) -> None:
+    """Strip leading ``/`` from hierarchical-label net names in-place.
+
+    The ``kicad-cli sch export netlist`` exporter prefixes hierarchical
+    local labels with the sheet path (e.g. ``/BST_A``, ``/DRV_CP1``)
+    while global power-symbol labels come through unprefixed
+    (e.g. ``+3V3``, ``GND``).  PCB net tables conventionally store bare
+    names, so adding the prefixed form creates ghost nets that get
+    dropped by orphan-net cleanup and prevents pads from being assigned
+    to the intended net.
+
+    This normalization mirrors what ``_extract_schematic_nets`` in
+    :mod:`kicad_tools.cli.fleet_cmd` does -- it consumes label-leaf
+    names that are already bare, so the drift detector and the sync
+    writer agree on the same naming convention.
+
+    Only a single leading ``/`` is stripped, and only from names that
+    do not begin with another ``/`` after the strip (defensive against
+    deeper hierarchical paths such as ``/sheetA/SIGNAL`` -- those keep
+    a leading ``/`` only if the parser was confused; we still strip
+    one and leave the rest as a sub-path token, which then becomes a
+    valid bare net name).
+
+    Power-rail aliases (``+3V3``, ``GND``, ``+5V``, ...) are left
+    unchanged because they don't carry the ``/`` prefix in the
+    exported netlist.
+
+    Args:
+        netlist: A :class:`Netlist` whose ``nets`` list will be
+            mutated in place.
+    """
+    for net in netlist.nets:
+        if net.name and net.name.startswith("/"):
+            net.name = net.name[1:]
+
+
 def _assign_nets_from_schematic(
     pcb, schematic_path: Path
 ) -> tuple[list[SyncAction], list[str]]:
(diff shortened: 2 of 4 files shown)


rjwalters:
## Judge verdict: APPROVE (with follow-up filing required before merge)

### Code review

**Fix 1 — `_normalize_kicad_cli_net_names` (`pcb_sync_netlist.py`)**
- Strips a single leading `/` from `net.name` in-place; mutates the netlist before `pcb.add_net()` and `assign_nets_from_netlist()`. Correct placement — runs after `export_netlist` succeeds and before any downstream consumer sees the names.
- Power-rail names (`+3V3`, `GND`, `+5V`) untouched because they don't carry the prefix — verified by `test_power_symbol_names_unchanged`.
- Idempotent (no `//` in real exporter output, but `test_normalisation_is_idempotent` confirms safety).
- Defensive handling of deep paths (`/sheetA/SIGNAL` → `sheetA/SIGNAL`) documented and tested.

**Fix 2 — unconditional `build_pin_to_pad_map` (`pcb_sync_netlist.py:442-449`)**
- Old conditional `if netlist.tool and "Python fallback" in netlist.tool:` removed. Now built for both paths.
- Reasoning in the comment is sound: kicad-cli emits *symbol* pin numbers (not pad numbers), identical to the fallback. Map is identity for the common case, diverges only for name-based-pad footprints. Safe to run unconditionally.
- Failure path retains "proceed without map" semantics — non-fatal.

**Drift detector symmetry (`fleet_cmd.py`)**
- `_extract_schematic_nets` now picks up `power:NAME` lib_id occurrences (excluding `PWR_FLAG`). The `PWR_FLAG` exclusion comment correctly identifies it as an ERC marker, not a net publisher — confirmed by `test_pwr_flag_excluded`.
- `_extract_pcb_named_nets` now filters to pad-attached nets via the new `_extract_pcb_pad_net_numbers` streaming scan. Depth-tracked s-expression parsing handles multi-line pad bodies correctly. Also drops `Net-(...)` and `unconnected-(...)` patterns for symmetry with the schematic side.
- The two functions now agree on what counts as a "real" net — necessary for the drift gate to produce comparable cardinalities.

### Test quality (20 new tests)

- 7 unit tests for `_normalize_kicad_cli_net_names` cover single-strip, power untouched, bare untouched, empty, deeper-path, mixed, idempotency — comprehensive.
- 9 unit tests for the drift-detector primitives cover label / power-symbol / `PWR_FLAG` / `Net-(...)` placeholder / pad-attached / pad-orphan / auto-generated / unreadable paths — comprehensive.
- 4 board-05 round-trip regression tests (drift nets present, ≥ 53 named-net count, no `/`-prefix, U3 pad assignments) anchor the fix to the original failure mode. Gracefully skip when board-05 artifacts absent.

### CI status

All 7 required checks pass:
- Lint & Format, Type Check, Test, C++ Build Check, kicad-cli Round-trip Smoke, Routed PCB DRC Check, Match-Group Routing Regression — green.

The 1 failing check is **Diff-Pair Routing Regression (boards/06-diffpair-test)**. This is a **pre-existing `main` regression**, not introduced by this PR:
- This PR only modifies `cli/fleet_cmd.py` and `cli/pcb_sync_netlist.py` — no router code.
- The same job fails on the two most recent `main` CI runs (commits c8949fe and 2999e84) with identical DRC count (32 > 21 allowlist).
- The regression first appeared at commit `2999e84` (PR #3375 "P_FP3 fine-pitch escape autorouter integration"), a router change.
- Job timed out during reroute iteration 3 (stagnation recovery), restored iter-1 best state, but final DRC still over the allowlist.

Per role guidance, `mergeStateStatus: UNSTABLE` with a non-required check failing → "Assess if acceptable". This is acceptable as a non-blocking, pre-existing, unrelated flake. Recommend filing a separate tracking issue for the router regression itself (out of scope here).

### Verification matches issue #3370 acceptance

- [x] 14 drift nets recovered (BST_A/B/C, DRV_CP1/CP2, DRV_FAULTn/LOCKn, DRV_VCP/VINT/VREG/VSW, FGFB/FGOUT, SPI_MISO)
- [x] No `/`-prefixed net names in synced PCB
- [x] Net count 40 → 69 (≥ 53 floor satisfied)
- [x] U3 pad assignments match schematic (U3.1=BST_A, U3.5=BST_C, U3.15=SPI_MISO, U3.17=DRV_FAULTn, etc.)
- [x] Regression test asserts no `/`-prefixed names (test_no_slash_prefixed_names)
- [x] Boards 01–04, 06–07 sync unchanged

### IMPORTANT: U10 follow-up issue must be filed before merge

The PR body documents that applying sync regresses U10 (STM32G431K8) pad assignments because of a **separate** schematic-emission bug in `design.py:_connect_mcu_pin_to_label` — SWDIO/SWCLK/ISENSE_B-/ISENSE_C- wire stubs collide with adjacent power rails before the label takes effect. As a result:

- `Closes #3370` will auto-close the issue on merge (commit trailer + PR body metadata, per `project_merge_script_closes_trailer.md`).
- But the issue's AC "`kct fleet status` no longer reports schematic drift for board 05" will **NOT** be met until the U10 schematic-emission bug is fixed.
- No tracking issue currently exists for the U10 wire-stub collision (verified: `gh issue list --search "U10 SWDIO SWCLK"` returns empty).

**Required before merge**: File a follow-up issue for the U10 schematic-emission bug with these elements:
- File:line reference `boards/05-bldc-motor-controller/design.py:_connect_mcu_pin_to_label`
- Affected labels: SWDIO, SWCLK, ISENSE_B-, ISENSE_C- (U10.23, U10.24, U10.6, U10.7)
- Symptom: wire stubs collide with adjacent power rails, silently bridging the net before the label can take effect
- Consequence: applying `pcb sync-netlist` to board 05 routed PCB pushes DRC from 6 to 73 errors, breaking #3258 hot-spot regression test
- Link back to this PR's caveat

Alternatively, weaken the `Closes #3370` trailer to `Refs #3370` in the commit message at merge time so #3370 stays open until the fleet-drift AC is actually satisfied.

### Final verdict

**LGTM on the code.** Fix is surgical, well-documented, well-tested, and addresses the root cause cleanly. Approving with the prerequisite that the U10 schematic-emission follow-up issue is filed (or the auto-close behavior is neutralized) before merge — otherwise the fleet status will continue to report board-05 drift after #3370 closes.

Issue #3413, opened 2026-06-09
board 06 (diffpair-test): close 18/21 → 21/21 + 0 DRC (re-measured 2026-06-10; was 41%)


rjwalters:
## Background

Per Issue #3402 audit (2026-06-09), board 06 has \`starting_layers: 4\` (the board is an inherent 4-layer JLCPCB tier-1 stackup with inner GND/PWR planes; impedance formulas Phase 3K assume the inner GND reference).

Even at 4L the production router tops out at **41% reach** (9/22 nets across two 4L variants). That is far below the 100% manufacturable bar.

## Reproduction

\`\`\`bash
uv run kct build-native --check
uv run kct route boards/06-diffpair-test/output/diffpair_test.kicad_pcb \\
  --auto-layers --max-layers 4 -o /tmp/board06.kicad_pcb
# 9/22 routed (41%); 33 DRC errors; both 4L variants stagnate at same reach
\`\`\`

## Diagnosis (audit observations)

- 4L SIG-GND-PWR-SIG: stagnation after iter 1 (overflow=2 not relieved); routed 9/22.
- 4L ALL-SIG: same stagnation, same 9/22.
- The audit's #2673 completion-floor guard kicks in (best completion 41% < 50% floor) and forces continuing escalation, which still doesn't help — both 4L rungs cap at the same reach.
- Existing routed artifact (\`diffpair_test_routed.kicad_pcb\`) is in the repo with the well-known #2672/#2677 DRC budget.

## Why this is a routing-engine gap, not a layer gap

Board 06 is purpose-built for diff-pair regression coverage. The 41% cap suggests the diff-pair routing path (\`--differential-pairs\`) is leaving congestion on the BGA-49 / QFN-32 escape that the negotiated A* cannot resolve at 4L.

## Scope (suggested)

- Investigate the diff-pair escape generation for the BGA-49 sink and the USB-C source.
- Consider whether \`--length-match-diffpairs\` and 4L ALL-SIG can hit 100% with a tuned recipe.
- Acceptance: \`kct route boards/06-diffpair-test/output/diffpair_test.kicad_pcb --auto-layers --max-layers 4 --differential-pairs\` reaches 22/22 (100%) with the existing #2672/#2677 DRC allowlist OR a tightened one.

## Out of scope

- \`starting_layers: 4\` is already set (Issue #3402).
- Phase 1-3 net-class feature exercises remain — see \`tests/test_board_06_diffpair_test.py\` for the regression contract.

[...]

Issue #3419, opened 2026-06-09
router: odd-dimension BGA grids misclassified as UNKNOWN ( quadrant balance, parent #3413)


rjwalters:
## Background

Investigation under #3413 (board 06 production routing reach gap) surfaced a router-detection bug that affects board 06's BGA-49 USB3 SuperSpeed sink (`U2`) and is reproducible in isolation against `detect_package_type`.

The bug only triggers below a critical odd-grid size. Board 06's 7x7 BGA-49 hits it; 9x9 squeaks under the threshold (see `test_odd_dimension_9x9_bga_still_detected` for the boundary).

## Reproduction

```python
from kicad_tools.router.escape import detect_package_type, PackageType
from kicad_tools.router.escape import Pad
from kicad_tools.router.layers import Layer

# Board 06's BGA-49 geometry: 7x7 grid at 1.27 mm pitch.
pads = []
for row in range(7):
    for col in range(7):
        pads.append(Pad(
            x=(col - 3) * 1.27, y=(row - 3) * 1.27,
            width=0.45, height=0.45, net=1, net_name="x",
            layer=Layer.F_CU, ref="U2", through_hole=False,
        ))
assert len(pads) == 49
assert detect_package_type(pads) == PackageType.BGA  # FAILS -- returns UNKNOWN
```

Also surfaced via the xfail-marked regression test added under #3413:
`tests/test_escape.py::TestUSBCBGAMisclassification::test_odd_dimension_7x7_bga_detected`.

## Root cause

`_is_grid_pattern` (escape.py:662) gates BGA classification on a per-quadrant pad balance check. The quadrant counter uses a strict ``if/elif/else`` cascade that lumps every pad lying directly on a center axis into the ``else`` branch:

```python
quadrants = [0, 0, 0, 0]
for p in pads:
    if p.x > center_x and p.y > center_y:
        quadrants[0] += 1
    elif p.x < center_x and p.y > center_y:
        quadrants[1] += 1
    elif p.x < center_x and p.y < center_y:
        quadrants[2] += 1
    else:
        quadrants[3] += 1  # <-- BUG: includes ALL axis pads
```

[...]

PR #3420, opened 2026-06-09
investigation(boards/06): root-cause for 41% reach (refs #3413; fix tracked in #3419)

rjwalters:
## Summary

Investigation under #3413 (board 06 production routing 41 -> 100%) identified the root-cause blocking diff-pair routing on the BGA-49 USB3 SuperSpeed sink, but the full fix requires a paired-escape direction heuristic that exceeds the scope of a single-issue PR.  This PR documents the BGA detection bug as an ``xfail(strict=True)`` regression test, files the full fix as #3419, and surfaces the routing-reach gap status.

**Current state**: board 06 stays at 41% reach (9/22 nets routed) under the existing committed routed PCB.  The DRC allowlist remains at 21 (unchanged).

## Root cause

``detect_package_type`` misclassifies the 7x7 BGA-49 USB3 sink (board 06's ``U2``) as ``PackageType.UNKNOWN``.  ``_is_grid_pattern`` (escape.py:662) uses a strict ``if/elif/else`` quadrant-balance cascade that lumps every pad lying directly on a center axis into the ``else`` branch, producing a 9/9/9/22 distribution for an otherwise perfectly symmetric 7x7 grid.  With avg=12.25 / threshold 20.8 the unbalanced ``quadrants[3]=22`` fails the BGA check.

Down-stream consequence: the BGA is dispatched through ``_escape_radial`` instead of ``_escape_bga_rings``, and -- critically -- the diff-pair-aware paired escape pre-pass (gated on ``PackageType.BGA``, escape.py:1313) does NOT engage on the USB3 SS pairs.  All 4 USB3 lanes (TX1, RX1, TX2, RX2) escape as per-pin radial segments, and the per-net A* main router gets stuck routing them back to J1 with no coupling.

## Why the detection fix is NOT in this PR

A naive ``_is_grid_pattern`` fix (distribute axis pads across the two adjacent quadrants, +0.5 each) does correct the detection -- BUT it regresses board 06's routing reach from 41% to ~27% because the diff-pair-aware paired escape pre-pass then fires and produces tightly-coupled escape endpoints (0.10mm intra-pair lateral offset at the BGA edge) that the per-net A* main router times out (30s default) trying to leave.

Two simultaneous fixes needed (tracked under #3419):

1. ``_is_grid_pattern``: distribute axis pads across the two adjacent quadrants.
2. ``_escape_diff_pair_segment`` / ``_generate_paired_escapes``: choose launch direction based on partner-connector proximity, not outward from package center.  For sparse-signal BGAs (≤8 signal pads), consider skipping the paired-escape pre-pass entirely.
3. ``_escape_bga_rings``: add a ``pad.net == 0`` filter (currently missing) -- mirrors ``_escape_qfp_alternating`` line 2385.  Without this, every GND/+1V2/+3V3 pad gets a wasteful escape segment.

## What this PR lands

- ``tests/test_escape.py::TestUSBCBGAMisclassification::test_odd_dimension_7x7_bga_detected`` — xfail-marked regression test that reproduces board 06's exact BGA-49 geometry (7x7 at 1.27mm pitch) and documents the bug with a full root-cause analysis in the xfail reason.
- ``test_odd_dimension_9x9_bga_still_detected`` — companion case proving the bug only fires below a critical grid size: 9x9 squeaks under the upper threshold (16/16/16/33 vs avg 20.25 / range [6.1, 34.4]) so larger odd grids still route as BGA.

## Investigation artifacts (not committed)

- ``_is_grid_pattern`` fix tried locally: U2 then correctly detects as BGA, paired escape pre-pass engages on the 4 USB3 pairs (8 escapes total), but per-net A* timeouts at 30s/net produce 4L SIG-GND-PWR-SIG attempt 1 = 6/22 (27%) -- worse than the 9/22 (41%) baseline.
- ``_escape_bga_rings`` plane-filter fix tried locally: filters the 40 plane-net pads from ring generation so they don't consume perimeter channels.  Composes cleanly with the detection fix but doesn't lift the per-net-A* timeout regression on its own.

## Allowlist status

The ``.github/routed-drc-tolerance.yml`` entry for board 06 stays at 21 (unchanged).  This PR does NOT lower it -- the 41 -> 100% reach gap is not closed.  Tightening the allowlist further requires the #3419 fix landing first.

## Test plan

- [x] ``tests/test_escape.py`` 114 passed, 1 xfailed (the documented 7x7 detection bug)
- [x] No regression on existing tests
- [x] ``check_diffpair_coverage.py`` re-route still produces the historical reach (no change to committed artifact or allowlist)

## Related

- Parent: #3413 (board 06 production routing reach gap)
- Follow-up: #3419 (BGA detection + paired-escape direction refinement; the actual fix)
- Adjacent: Epic #2556 Phase 2F (paired escape coupling), #2639 (CoupledPathfinder), #2677 (corridor reservation)

### DIFF
--- a/tests/test_escape.py
+++ b/tests/test_escape.py
@@ -512,6 +512,84 @@ class TestUSBCBGAMisclassification:
         assert len(pads) == 22
         assert detect_package_type(pads) != PackageType.BGA
 
+    @pytest.mark.xfail(
+        strict=True,
+        reason=(
+            "Issue #3413: odd-dimension BGA grids (e.g. 7x7 BGA-49) have a "
+            "full row AND a full column directly on the package center "
+            "axes. ``_is_grid_pattern``'s strict ``if/elif/else`` quadrant "
+            "cascade lumps every axis pad into the ``else`` branch, "
+            "producing a 9/9/9/22 distribution for a perfectly symmetric "
+            "7x7 grid and failing the [0.3, 1.7] x avg balance check. "
+            "That misclassifies board 06's USB3 SS BGA-49 sink as "
+            "``PackageType.UNKNOWN`` and routes it through the radial-"
+            "escape fallback. A naive fix (distribute axis pads across "
+            "the two adjacent quadrants) makes detection pass AND fires "
+            "the diff-pair-aware paired escape pre-pass on this BGA, but "
+            "the resulting paired-escape geometry (tight 0.10mm intra-"
+            "pair lateral offset at the BGA edge, escape direction "
+            "perpendicular to the pair axis) drops the per-net A* "
+            "routing reach on board 06 from 41% (the documented baseline) "
+            "to ~27% because the per-net A* times out (30s default) "
+            "trying to leave the tight escape endpoints. The detection "
+            "fix needs to land alongside an escape-strategy refinement "
+            "for sparse-signal BGAs (paired escape direction toward the "
+            "partner connector, not outward from package center). "
+            "Tracked under #3419 (follow-up to #3413)."
+        ),
+    )
+    def test_odd_dimension_7x7_bga_detected(self):
+        """7x7 BGA (49 pads, odd grid) should be detected as BGA (Issue #3413).
+
+        REGRESSION TEST DOCUMENTING THE BUG.  See the ``xfail`` reason
+        above for the full root-cause analysis and why a naive
+        ``_is_grid_pattern`` fix is not safe to land on its own.
+
+        Board 06's BGA-49 USB3 SuperSpeed sink is a 7x7 grid at 1.27 mm
+        pitch.  The router currently misclassifies it as
+        ``PackageType.UNKNOWN`` and dispatches the 8 USB3 SS signal pads
+        through ``_escape_radial`` -- a simple per-pin surface escape
+        that happens to produce the 41% reach this issue captures.  When
+        the BGA detector is corrected in isolation, the paired-escape
+        pre-pass (gated on ``PackageType.BGA``) fires on the USB3 pairs,
+        produces tightly-coupled escape endpoints at the BGA edge, and
+        the per-net A* main router times out searching from those
+        endpoints.  Reach drops to ~27%.
+
+        Two-phase fix needed (tracked separately):
+          1. ``_is_grid_pattern``: distribute axis pads across the two
+             adjacent quadrants when checking the balance heuristic.
+          2. ``_escape_diff_pair_segment`` / ``_generate_paired_escapes``:
+             choose the launch direction based on partner-connector
+             proximity (e.g., consult ``net_class_map`` / the diff-pair
+             partner's package location) rather than outward from
+             package center.
+        """
+        pads = create_bga_pads(7, 7, pitch=1.27)
+        assert len(pads) == 49
+        assert detect_package_type(pads) == PackageType.BGA
+
+    def test_odd_dimension_9x9_bga_still_detected(self):
+        """9x9 BGA (81 pads, odd grid) IS detected correctly today (Issue #3413).
+
+        Sibling case for Issue #3413: confirms that the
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!** (Judge review)

Tests-only PR documenting the board-06 BGA-49 misclassification root cause as an `xfail(strict=True)` tripwire, with the coordinated fix deferred to #3419. The investigation quality is high and the deferral rationale (naive fix regresses reach 41% → 27%) is well-documented in the test docstring, xfail reason, commit message, and follow-up issue.

## Verification performed

**Root cause spot-check** — verified against `src/kicad_tools/router/escape.py:662` (`_is_grid_pattern`): the quadrant counter's strict `>`/`<` `if/elif/else` cascade does send every center-axis pad to `quadrants[3]`. Arithmetic confirmed: 7x7 → 36 strict-quadrant pads (9 each) + 13 axis pads = 9/9/9/22 vs upper bound 1.7 × 12.25 = 20.825 → fails; 9x9 → 16/16/16/33 vs 34.425 → passes. Both test claims are exact. The BGA-gated paired-escape pre-pass cited as the downstream consequence is at escape.py:1262-1283 (PR cites :1313 — minor citation drift, not blocking).

**Test mechanics** — ran locally in the issue-3413 worktree:
- `tests/test_escape.py`: 114 passed, 1 xfailed (matches the PR test plan)
- `test_odd_dimension_7x7_bga_detected` → XFAIL as designed
- `test_odd_dimension_9x9_bga_still_detected` → PASSED, meaningfully pinning the boundary
- `strict=True` is the right tripwire: when #3419 fixes detection, this test XPASSes and fails the suite, forcing the coordinated update. The docstring explicitly tells the future fixer the two-phase plan (axis-pad distribution + partner-aware launch direction), and #3419's first AC is "remove the xfail marker" — the loop is closed.

**Closure-risk fix (made during review)** — the PR title previously contained "closes #3413". The merge script squash-merges via the GitHub API with no `commit_title` override, so the PR title becomes the default-branch commit subject and the closing keyword would have auto-closed #3413 with its 100%-reach AC unmet. Since #3419's acceptance criteria only guard "no regression below 41%" — NOT the 100% goal — #3413 must stay open as the board-06 reach tracker. I edited the title to "refs #3413; fix tracked in #3419". Verified no closing keywords remain in the title, body, or branch commit, and `closingIssuesReferences` is empty: merging will leave both #3413 and #3419 open. (Intentional deviation from the usual "Closes #N" requirement — this investigation PR must not close its issue.)

## CI status

7 of 8 checks pass (Type Check, Test, Lint, C++ Build, Routed PCB DRC, kicad-cli Round-trip, board-07 Match-Group regression). The board-06 Diff-Pair regression check fails — but **identically on main** (re-route DRC 32 > allowlist 21, 19/21 PARTIAL, same numbers on main run 27225987921 and this PR's run 27227607892). This tests-only diff cannot affect routing; the failure is pre-existing main-branch breakage from a recent router merge, now tracked in #3421. Merge state is UNSTABLE solely due to that pre-existing failure — assessed as acceptable per the judge UNSTABLE protocol.

## Test Execution

1. `tests/test_escape.py` 114 passed, 1 xfailed — ✅ Executed: confirmed locally (85s)
2. No regression on existing tests — ✅ Executed: full file green; CI Test job also passes
3. `check_diffpair_coverage.py` re-route reach unchanged — ⚠️ Not executed locally (long-running); CI re-route shows the pre-existing main-branch allowlist breach (#3421), unrelated to this diff

## Follow-ups
- #3421 (new): main-branch board-06 CI gate failing (32 > 21) — needs bisect of recent router merges
- #3419 is well-formed as the fix tracker (repro snippet, exact quadrant math, concrete 3-part fix plan, no-regression AC) but is currently unlabeled — needs curation before a builder can claim it
- #3413 remains the open 100%-reach tracker for board 06

Issue #3421, opened 2026-06-09
CI: board 06 diff-pair regression gate failing on main (re-route DRC 32 > allowlist 21)


rjwalters:
## Bug Description

The `Diff-Pair Routing Regression (boards/06-diffpair-test)` CI job is failing on the **main branch itself**, not just on PRs. The re-route produces 32 DRC errors against the `.github/routed-drc-tolerance.yml` allowlist value of 21.

Observed on main run [27225987921](https://github.com/rjwalters/kicad-tools/actions/runs/27225987921) (2026-06-09 18:05Z) and reproduced identically on PR #3420's run [27227607892](https://github.com/rjwalters/kicad-tools/actions/runs/27227607892) (a tests-only PR that cannot affect routing):

```
PARTIAL: 19/21 signal nets routed
[diffpair-coverage] DRC error count: 32 (allowed: 21)
##[error]DRC regression on re-routed .../diffpair_test_routed.kicad_pcb: 32 error(s) exceeds allowlist value 21
```

All recent main CI runs today (16:18Z through 18:06Z) show the same failure, so a recently merged router change altered the board-06 re-route output. Candidate causes (recent router merges): #3203 (per-pad channel budget edge-classification), #3193 (deterministic A* tie-break + PYTHONHASHSEED).

## Impact

- **Severity**: High for workflow health — every open PR shows a red board-06 check regardless of content, eroding the signal value of the gate and pushing judges toward overriding UNSTABLE merge states.
- The committed routed artifact is unaffected (Routed PCB DRC Check passes); only the fresh re-route regressed.

## Reproduction

```bash
git checkout main
# run the diff-pair regression job steps locally, or:
gh run view 27225987921 --log-failed
```

## Suggested scope

- Bisect the recent router merges to find which changed the board-06 re-route DRC profile (21 -> 32).
- Either fix the new violations or, if the new routing is acceptable, raise the allowlist with justification.
- Related: #3413 (board 06 reach tracker), #3419 (BGA detection follow-up).

Discovered while judging PR #3420.

Issue #2513, opened 2026-05-06
Board 03: U1 (USB-C) still blocks 4 nets after escape pass — USB_D+, USB_CC1, USB_CC2, XTAL2


rjwalters:
## Problem

Board 03 (`boards/03-usb-joystick`) routes 9/13 nets after running default `kct route`. Four nets remain unrouted because the USB-C connector U1 geometrically blocks them:

- USB_D+ — `blocked_path`
- USB_CC1 — `blocked by U1`
- USB_CC2 — `blocked by U1`
- XTAL2 — `blocked by U1`

DRC on the resulting routed PCB reports 9 errors (clearance_segment_segment×1, clearance_segment_via×6, clearance_via_via×2) — driven by the partial routes that remain after rip-up cycles fail to find paths.

## Reproduction

```
uv run kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb \
  -o boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb \
  --timeout 240 --quiet
```

Tail of `/tmp/route_audit_2026-05-06/03-route.log`:

```
=== Routing with Escape Complete ===
  Dense packages escaped: 1
  Total nets routed: 9
  Total segments: 134
  Total vias: 30
Failed nets:
  - Net 4 "USB_D+": blocked_path (blocked_path)
  - Net 7 "USB_CC2": blocked_path (blocked by U1)
  - Net 6 "USB_CC1": blocked_path (blocked by U1)
  - Net 16 "XTAL2": blocked_path (blocked by U1)
```

## Context

Issues #2473 (CoupledPathfinder N-pad), #2484 (CoupledPathfinder spacing), #2490 (USB_CC1/JOY_X unrouteable), and #2482 (JOY_X/USB_CC1 75% timeout) all targeted board 03 USB-C routing and are CLOSED. Routing improved dramatically (May-2 audit baseline was 0/13) but the current escape pass is still missing four pins on this footprint.

XTAL2 being blocked by U1 is notable — it is not a USB net but its routing path is still being blocked by the connector, suggesting the escape coverage gap is not specific to USB diff-pairs but to the U1 footprint geometry as a whole.

## Acceptance

- `kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb` routes 13/13 nets within the 240s default timeout.
- Resulting routed PCB has 0 DRC errors (`kct check --mfr jlcpcb`).
[...]

Issue #3428, opened 2026-06-09
Router: target-aware in-pad inner segment direction for fine-pitch QFP escape (Issue #3411 follow-on)


rjwalters:
## Background

Per Issue #3411 investigation, the LQFP-48 0.5mm pitch escape pattern in `_escape_qfp_alternating` (`src/kicad_tools/router/escape.py:2305`) blocks adjacent fine-pitch signal pins from both routing.  Specifically: when U2.6 OSC_OUT claims an in-pad via rescue, the in-pad escape's inner B.Cu segment is emitted in the `alt_dir_cw=NORTH` direction (= (0, +1) in y-down coords = TOWARD U2.7).  The resulting 0.5mm-wide B.Cu stub blocks U2.7 NRST's own escape via.

The direction choice is driven by the pin's INDEX in the sorted same-edge pad list, not by the net's actual routing target.  For board-04 STM32F103C8T6 the crystal is NW of U2 — OSC_OUT's natural target direction is NW.  But the dispatcher emits the inner segment in the SE direction (alt_dir_cw=NORTH=+y).

## Proposed fix

Modify `_try_in_pad_escape` (`escape.py:5617-5953`) to ACCEPT an optional `target_direction` argument supplied by the caller (`_escape_qfp_alternating` and friends).  When provided, the inner-layer escape segment is emitted in `target_direction` rather than the original `direction`.  Otherwise, preserve the legacy behavior.

The caller computes `target_direction` by querying the routing graph for the net's other endpoints and choosing the cardinal direction that minimizes the diagonal distance.

## Acceptance

- [ ] Board 04 (stm32-devboard) reaches 9/9 nets routed (100%) under the production routing recipe at jlcpcb-tier1
- [ ] Board 03 (USB joystick) is unchanged (no regression on the TQFP-32 USB_CC1/D+/D- escape pattern)
- [ ] Board 05 (BLDC motor controller) is unchanged (no regression on the DRV8301 + STM32G431 escape patterns)
- [ ] Routed-DRC tolerance gate unchanged
- [ ] Determinism preserved (byte-identical output with `--seed 42`)

## Related

- Parent: #3411 (board 04 reach gap)
- Historical: #2982 (waypoint hints that previously enabled 9/9 — superseded by #3128 clearance tightening)
- Adjacent: #3266 (NRST regression), #3267 (U2.35 stitch regression), #3075 (U2.8 stitch blocker)

rjwalters:
[...]

PR #3426, opened 2026-06-09
fix(ci): board 06 diff-pair regression gate — bisect + fix (closes #3421)

rjwalters:
## Summary

Fixes the board 06 `Diff-Pair Routing Regression` CI gate that has been red on main itself since 2026-06-09 00:21Z (re-route DRC 32 > allowlist 21), turning every PR's board-06 check red regardless of content.

Closes #3421

### Bisect result

No local bisect was needed — CI run history isolates the regression to a **single commit**:

| run | head SHA | diff-pair job | DRC count |
|---|---|---|---|
| 27165130514 (last green) | `c1a4d549` (P_FP2, #3374) | pass | **13** |
| 27175601138 (first red) | `2999e847` (P_FP3, #3375) | **fail** (only failing job) | **32** |

**Culprit: PR #3375 — `feat(router): fine-pitch escape autorouter integration (P_FP3)` (commit `2999e847`).** This matches the judges' earlier trace. The #3203 / #3193 candidates named in the issue are exonerated (both merged well before the green run at `c1a4d549`).

### Classification: real routing-quality regression (no allowlist change)

P_FP3 wired the fine-pitch escape region detector into `load_pcb_for_routing`. On board 06 it detects 5 regions (J3, J4, U1, U3, U4 — all 0.8mm pitch, escape clearance 0.14mm). At board 06's recipe (0.15 trace / 0.15 clearance) these packages' corridors are **infeasible even at the escape clearance** (0.8 − 2·0.14 − 0.15 = 0.37mm < 0.43mm required), so the regions can never serve their purpose — but they still perturbed routing through two defects:

1. **Guard decline clobbered the legacy shrink** — `RoutingGrid._clearance_for_pin_pitch`'s region branch hard-returned the *standard* halo (0.225mm) when the narrow-channel guard declined, skipping the legacy `min_trace_width` shrink (0.05mm) those pads carried pre-P_FP3. A 4.5x halo inflation on every pad of the five packages. Fixed: decline now **falls through** to the legacy shrink path.

2. **Guard evaluated the wrong pitch for foreign pads** — the feasibility check used the call-side pad's own component pitch, so a 0.96mm-pitch passive inside a BGA's 5mm halo picked up the 0.14 escape clearance even though the BGA corridor it would serve is infeasible. Fixed: the guard (grid halo + `resolve_clearance_with_escape_region`) now requires feasibility for **both** the region's package pitch and the call-side pitch. Strictly more conservative; infeasible regions become fully inert.

[...]

PR #3431, opened 2026-06-09
fix(router): BGA odd-grid detection + paired-escape coordination (refs #3413, closes #3419)

### LINKED ISSUE #3419: router: odd-dimension BGA grids misclassified as UNKNOWN ( quadrant balance, parent #3413)

rjwalters:
## Summary

Three-part fix for board 06's BGA-49 misclassification (the #3413 reach investigation), landing the detection fix WITHOUT the 41% -> 27% reach regression that blocked it from shipping standalone.

**Part 1 — `_is_grid_pattern` odd-grid quadrant fix (needed).** Odd-dimension grids (7x7 BGA-49) have a full row AND column exactly on the package center axes; the strict `if/elif/else` cascade lumped all axis pads into `quadrants[3]` (9/9/9/22, balance check fails, package -> UNKNOWN). Axis pads are now distributed fractionally across adjacent quadrants (0.5/0.5; center pad 0.25 x 4). Even grids have no axis pads, so their counts are bit-for-bit unchanged. The strict `xfail` marker from PR #3420 is removed; `test_odd_dimension_7x7_bga_detected` passes, and the 9x9 boundary test still passes.

**Part 2 — paired-escape coordination (needed; two mechanisms).**
- *Partner-connector-aware launch direction* (`_select_pair_launch_direction`): paired escapes now launch toward the centroid of the pair's off-package endpoints (new board-wide `net_pad_positions` map threaded from `Autorouter._escape`), with an interior veto so the pair never launches across the package pad field. Falls back to the original quadrant rule when no positions are available. **Measured alone this was NOT sufficient** — reach still dropped to 27% (6/22) because the per-net A* cannot expand from endpoints inside the partner trace's clearance envelope.
- *Coupled-consumer gate* (`Autorouter.paired_escape_coupling`, default False): the paired pre-pass emits endpoints at the intra-pair clearance that only the CoupledPathfinder can route from. The map is now threaded into the EscapeRouter only when `route_all_with_diffpairs` will consume the escapes (flag flipped at its entry; an already-created escape router is refreshed in place). Plain per-net routing (`kct route` without `--differential-pairs`) gets single-ended escapes — this is what restores (and lifts) reach.

**Part 3 — `_escape_bga_rings` plane-pad filter (needed).** Skips `net == 0` pads, mirroring the #2513 `_escape_qfp_alternating` filter. On board 06 the BGA-49 now escapes exactly its 8 USB3 signal pads instead of all 49 balls, freeing the perimeter channel.

## Measurements

| Metric | Current main | This PR | Delta |
|---|---|---|---|
| Board 06 reach (`kct route ... --auto-layers --max-layers 4`) | 9/22 (41%) | **10/22 (45%)** | **+1 net** |
| Board 06 re-route DRC (`check_diffpair_coverage.py --seed 42`, local) | 31 errors | **31 errors** | **0** |
| U2 (BGA-49) classification | UNKNOWN (radial escape, 49 escapes) | **BGA** (ring escape, 8 signal escapes) | fixed |

- Reach baseline measured on this machine immediately before the fix (same command, same contention conditions). Naive detection fix alone reproduced the predicted regression (27%, 6/22); with the full 3-part fix reach is 45%.
- The re-route DRC gate (`diffpair-routing-regression`) measures 31 locally on BOTH current main and this branch (CI sees 32 on main — the 31-vs-21-allowlist regression is #3421's lane, being bisected in parallel; this PR's delta against current main is exactly 0, and `.github/routed-drc-tolerance.yml` is untouched). The generate_design flow does not run the dense-package escape pre-pass, so this PR is inert there — confirmed by identical logs (27 budget-exceeded events each).

## Tests

- xfail marker removed from `test_odd_dimension_7x7_bga_detected` (now passing); 9x9 boundary test unchanged and passing.
- New: `TestPairLaunchDirectionHeuristic` (7 tests: fallback behavior, connector-aimed launch, interior veto, end-to-end threading, Autorouter map threading), `test_bga_rings_skip_plane_net_pads`, Gate-2 coupling-gate tests (`empty_map_without_coupling`, `flips_coupling_flag`).
- Full escape/diffpair sweep (`-k "escape or diffpair or diff_pair"`): 1054 passed on this branch; the 3 remaining failures (2x `test_subgrid_integration`, 1x `test_diffpair_length_tuning` order-dependent flake) fail identically on current main (verified with the same sweep command).
- ruff/mypy: zero new findings vs main on all touched files.

## Which of the 3 parts were needed

All three. Part 1 is the core fix; Part 2's direction heuristic alone did not prevent the regression — the coupled-consumer gate is the binding mechanism; Part 3 frees the BGA perimeter (8 escapes instead of 49) and contributes to the reach lift.

Refs #3413 (stays open — board 06 still 45% vs 100% bar).

Closes #3419

### DIFF
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -10961,6 +10961,14 @@ class Autorouter:
     def _escape(self) -> EscapeRouter:
         """Lazy-initialize escape router."""
         if self._escape_router is None:
+            # Issue #3419: board-wide net -> pad positions map so the
+            # paired-escape pre-pass can aim the launch direction toward
+            # the partner connector (off-package endpoints) instead of
+            # blindly outward from the package center.
+            net_pad_positions: dict[str, list[tuple[float, float]]] = {}
+            for pad in self.pads.values():
+                if pad.net_name:
+                    net_pad_positions.setdefault(pad.net_name, []).append((pad.x, pad.y))
             self._escape_router = EscapeRouter(
                 self.grid, self.rules, net_class_map=self.net_class_map,
                 edge_clearance=self._edge_clearance,
@@ -10972,6 +10980,7 @@ class Autorouter:
                 # ``get_diff_pair_map`` returns {} when no pairs are
                 # detected, which preserves pre-#2639 behavior bit-for-bit.
                 diff_pair_map=self.get_diff_pair_map(),
+                net_pad_positions=net_pad_positions,
             )
         return self._escape_router
 
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -781,6 +781,20 @@ class Autorouter:
         self._escape_router: EscapeRouter | None = None
         self._subgrid_router: SubGridRouter | None = None
 
+        # Issue #3419: gate for the diff-pair paired-escape pre-pass.
+        # The pre-pass emits two tightly-coupled escape endpoints (at the
+        # intra-pair clearance) that are only routable by a COUPLED
+        # consumer (``route_all_with_diffpairs`` -> CoupledPathfinder).
+        # When the board is routed per-net (plain ``route_all`` /
+        # ``route_all_negotiated`` without --differential-pairs), each
+        # half's A* start point sits inside the partner trace's clearance
+        # envelope and the search strands (30s/net timeouts -- the 41% ->
+        # 27% reach regression on board 06 documented in #3419).  The
+        # flag is flipped on by ``route_all_with_diffpairs`` before the
+        # escape phase; ``_escape`` consults it to decide whether to
+        # thread the diff_pair_map into the EscapeRouter.
+        self.paired_escape_coupling: bool = False
+
         # Issue #2838 (closes #2761 gap): Lazy-initialized via conflict
         # manager.  Wired into the single-ended PIN_ACCESS retry path in
         # ``route_net`` so vias from already-routed nets that sit within
@@ -10600,6 +10614,25 @@ class Autorouter:
                 ``diffpair_routing.py``).  When ``None`` (default) the
                 legacy unbounded behaviour is preserved for back-compat.
         """
+        # Issue #3419: enable the paired-escape pre-pass for this run.
+        # ``route_all_with_diffpairs`` routes pairs through the
+        # CoupledPathfinder which is the intended consumer of the
+        # tightly-coupled escape endpoints the pre-pass emits.  The flag
+        # must be set BEFORE the escape phase runs (escapes are generated
+        # lazily inside the delegated strategies).  If the escape router
+        # was already created by an earlier per-net phase (flag off), its
+        # map is refreshed in place so this run still gets paired escapes.
+        # ``getattr`` guards: some unit tests construct partially
+        # initialised Autorouter doubles (``__new__`` without
+        # ``__init__``); the flag flip must not crash on them.
+        self.paired_escape_coupling = bool(
+            diffpair_config is not None and diffpair_config.enabled
+        )
+        if getattr(self, "_escape_router", None) is not None:
+            self._escape_router.diff_pair_map = (
+                self.get_diff_pair_map() if self.paired_escape_coupling else {}
+            )
+
         # Issue #3321: derive a per-pair budget from --timeout when the
         # caller has not explicitly configured one.  This protects the
         # CoupledPathfinder from running unbounded just because the CLI's
@@ -10979,7 +11012,17 @@ class Autorouter:
                 # dense packages get coupled-at-launch escape routes.
                 # ``get_diff_pair_map`` returns {} when no pairs are
                 # detected, which preserves pre-#2639 behavior bit-for-bit.
-                diff_pair_map=self.get_diff_pair_map(),
+                #
+                # Issue #3419: only thread the map when a coupled consumer
+                # exists (``route_all_with_diffpairs`` flips
+                # ``paired_escape_coupling`` on before the escape phase).
+                # Paired escape endpoints sit at the intra-pair clearance
+                # and are unroutable by the plain per-net A* -- threading
+                # the map unconditionally regressed board 06 from 41% to
+                # 27% reach once the BGA-49 detection fix landed.
+                diff_pair_map=(
+                    self.get_diff_pair_map() if self.paired_escape_coupling else {}
+                ),
                 net_pad_positions=net_pad_positions,
             )
         return self._escape_router
--- a/src/kicad_tools/router/escape.py
+++ b/src/kicad_tools/router/escape.py
@@ -718,10 +718,50 @@ def _is_grid_pattern(pads: list[Pad], center_x: float, center_y: float) -> bool:
     if len(interior_pads) < len(pads) * 0.1:
         return False
 
-    # Count pads in each quadrant relative to center
-    quadrants = [0, 0, 0, 0]
+    # Count pads in each quadrant relative to center.
+    #
+    # Issue #3419: odd-dimension grids (7x7 BGA-49, etc.) have a full row
+    # AND a full column lying exactly ON the center axes.  The previous
+    # strict ``if/elif/else`` cascade lumped every axis pad into the last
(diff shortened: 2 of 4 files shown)


rjwalters:
✅ **Approved!**

Reviewed all four parts of the fix (odd-grid detection, partner-aware launch direction, plane-pad ring filter, and the `paired_escape_coupling` gate). Implementation is correct, well-tested, and honestly framed — the PR explicitly documents that the +1 net is incidental and the real value is landing the detection fix (the #3420 strict-xfail tripwire) without the predicted 41% -> 27% regression. No overclaiming.

## Gate correctness (the architecturally significant change)

- The flag condition in `route_all_with_diffpairs` (`diffpair_config is not None and diffpair_config.enabled`) **exactly matches** the inner `DiffPairRouter.route_all_with_diffpairs` bail-out (`diffpair_config is None or not enabled` -> plain `route_all`). The gate is ON precisely when the CoupledPathfinder consumer exists — no caller can end up with paired escapes and no coupled consumer, or vice versa.
- All three `route_cmd.py` call sites (lines 8615/8747/8783) pass a CLI-built config with `enabled=True`, reached only under `--differential-pairs`. Plain `kct route` keeps the flag off and gets single-ended escapes.
- The flag flip happens before the delegated routing (escapes generate lazily inside), and the in-place map refresh handles an escape router created by an earlier phase. The `no-pairs-detected` case is consistent too: gate ON but `get_diff_pair_map()` returns `{}`, so nothing fires.
- One asymmetry found (NOT introduced by this PR): `orchestrator.py:688` and `:1340` still thread `diff_pair_map` unconditionally into their EscapeRouter while the route-auto consumer is the per-net GlobalRouter — and the shared odd-grid detection fix enlarges that surface to 7x7 BGAs. Pre-existing condition (#2639-era threading), so non-blocking; filed **#3432** to track gating the orchestrator path the same way.

## Even-grid bit-for-bit verification

Verified by reading the full `_is_grid_pattern`: even grids place every pad at >= pitch/2 (0.4 mm at 0.8 mm pitch) from the center axes vs eps = 1e-6, so they take only the unchanged strict-inequality integer branches; `0 + n*1.0` is exact in float, and the balance check arithmetic is identical. 7x7 math checks out: 9 strict + 3 from axis halves + 0.25 center = 12.25 per quadrant = exactly avg. Even-grid coverage is solid: 4x4, 6x6, 8x8, 10x10, 16x16 detection tests plus the 9x9 odd boundary — all pass.

## Other checks

- `net_pad_positions` is built once inside the `if self._escape_router is None` lazy-init (O(#pads), not per-net). The launch-direction helper is O(pair-net endpoints) per pair with a clean three-way fallback and an interior veto (zero-outward-vector degenerate case also handled).
- `_escape_bga_rings` plane filter mirrors the #2513 QFP filter; `test_bga_rings_skip_plane_net_pads` covers it (11/16 signal escapes, none for net=0).
- Local: 162/162 tests pass in `tests/test_escape.py` + `tests/test_escape_diffpair.py` (includes the un-xfailed 7x7 test and the 7 new launch-direction tests). Ruff on touched files: identical findings to main (zero new).
- CI: **all 8 checks pass**, including Diff-Pair Routing Regression (boards/06) — verified the run (20:46Z) postdates the #3426 merge (19:50Z), so this passed against the FIXED allowlist baseline, not the stale one. Merge state: CLEAN.

## Coordination note for the orchestrator

In-flight #3428 (target-aware in-pad stubs, no PR yet) has uncommitted edits to the **same regions** — `EscapeRouter.__init__` (~855-920) and the `Autorouter._escape` property (~10975+) — guaranteed textual conflict. **Merge #3431 first; #3428 rebases on top** (it has nothing pushed, so the rebase cost is one-sided and minimal).

PR #3446, opened 2026-06-10
fix(router): parse_vias via-loss repair + --targeted-ripup CLI; board 07 recovery #3 (refs #3414)

rjwalters:
## Summary

Recovery run #3 for #3414 (board 07 matchgroup-test: 84% audit -> 100% manufacturable bar). Three prior builder sessions crashed; this run inherited their uncommitted documentation, validated it, re-measured the baseline on current main, and fixed the two real defects it could reach. **This PR deliberately does NOT auto-close issue #3414** — see "Why no Closes" below.

### Inherited work: KEPT (with corrections)
The inherited diff was documentation-only (route-recipe docstring/comment history in `generate_design.py`, `project.kct` checklist, regenerated sidecar). It was accurate and the follow-up issues it references (#3438-#3441) all exist, so it is kept. Its root-cause claim ("4th-from-outer pin structurally unroutable") is superseded by this run's refined diagnosis (below) and the `project.kct` line was rewritten accordingly.

### Baseline on current main (post-#3437, solo, seed 42)
**28/31 (90%), 23 raw DRC / 21 gate-counted errors (allowlist 25)**; stranded: `DQ3`, `MIPI_CLK_N`, `MIPI_DAT0_N`. The five merges since the 84% audit (#3426, #3431, #3434, #3435, #3437) did not move reach (iteration 0 dips to 26 post-#3434; iteration-1 recovery restores 28).

### Real fixes shipped

1. **`parse_vias` dropped every via on KiCad-8 output** (`src/kicad_tools/router/optimizer/pcb.py`). The ordered regex required `(net N)` immediately after `(layers ...)`, but the writer emits `(uuid "...")` between them — so `--preserve-existing` (#3155) silently preserved segments while dropping ALL vias (fragmented multi-layer copper, unblocked via barrels), and the `load_existing_routes` obstacle pre-pass had the same blind spot. The existing test fixture used the legacy net-before-uuid order, which is why CI never caught it. Rewritten on the same balanced-block walker `parse_segments` already uses; #3118 `micro` token preserved; 4 regression tests added.
2. **`--targeted-ripup` / `--max-ripups-per-net` wired end-to-end** (outer parser, command forwarding, all 7 `route_all_negotiated` call sites). The implementation existed but was CLI-unreachable (#3438 suggested direction 2). Parser-drift tests pass.

### Root cause of the remaining 28/31 plateau (posted to #3438)

[...]

PR #3451, opened 2026-06-10
fix(router): _remove_conflicting_vias KiCad-8 field-order bug (closes #3447)

### LINKED ISSUE #3447: router/io.py: _remove_conflicting_vias has the same ordered via regex bug fixed in parse_vias (#3446) — matches zero vias on KiCad-8 output
## Bug

PR #3446 fixed `parse_vias` in `src/kicad_tools/router/optimizer/pcb.py`, whose ordered regex required `(net N)` to immediately follow `(layers ...)` — KiCad 8+ output (including everything `kct route` writes) emits `(uuid "...")` between them, so the pattern matched ZERO vias.

The judge sweep for PR #3446 found one more parser with the identical defect: `_remove_conflicting_vias` in `src/kicad_tools/router/io.py` (~line 3515):

```python
via_pattern = re.compile(
    r"(\(via\s+"
    r"\(at\s+([\d.-]+)\s+([\d.-]+)\)\s*"
    r"\(size\s+[\d.]+\)\s*"
    r"\(drill\s+[\d.]+\)\s*"
    r'\(layers\s+"[^"]+"\s+"[^"]+"\)\s*'
    r"\(net\s+(\d+)\)"
    r"[^)]*\)[^)]*\))",
    re.DOTALL,
)
```

Two defects:
1. **uuid-before-net order**: on KiCad-8 field order this matches zero vias, so post-merge co-located via conflict detection (`detect_via_conflicts=True` in the merge path) silently no-ops — conflicting same-position vias on different nets are never removed.
2. **micro/blind/buried token** (#3118): `(via micro ...)` puts the type token between `via` and `(at ...)`, also failing the match.

## Fix

Rewrite on the balanced-block walker `_extract_balanced_blocks` exactly as PR #3446 did for `parse_vias` (the helper lives in `src/kicad_tools/router/optimizer/pcb.py:34`; either import it or hoist it to a shared util). Add regression tests with KiCad-8 field order (uuid before net) and a `micro` via, mirroring `tests/test_optimizer_pcb_segments.py::TestParseVias`.

Note: the other via regex in `io.py` (~line 3631, connectivity validation) uses lazy `.*?` wildcards and is order-tolerant — not affected.

Discovered while judging PR #3446 (sweep requested by #3414 recovery review).

rjwalters:
## Summary

`_remove_conflicting_vias` in `src/kicad_tools/router/io.py` used the same ordered via regex that PR #3446 fixed in `parse_vias`: it required `(net N)` to immediately follow `(layers ...)`. On KiCad-8 field order (uuid before net) — the order `kct route` itself writes — it matched ZERO vias, so post-merge co-located via conflict detection (`detect_via_conflicts=True` in `merge_routes_into_pcb`) silently no-opped. It also missed `(via micro ...)` blocks (#3118).

## Changes

- Rewrote `_remove_conflicting_vias` on the `_extract_balanced_blocks` walker from `src/kicad_tools/router/optimizer/pcb.py` (imported, not copied — avoids a third implementation), with order-independent `(at ...)` / `(net N)` field extraction and malformed-block skipping
- Added a comment at the `verify_output_connectivity` via regex (io.py ~3631) documenting that its lazy `.*?` wildcards are order-tolerant — checked per the judge sweep and intentionally left as-is
- Added `TestRemoveConflictingViasFieldOrder` (6 tests) in `tests/test_router_io.py` mirroring PR #3446's `TestParseVias`

## Ordered-regex sweep findings

Grepped `src/` for other via-block regexes:

| Location | Verdict |
|---|---|
| `router/io.py:~3515` (`_remove_conflicting_vias`) | **Bug — fixed in this PR** |
| `router/io.py:~3631` (`verify_output_connectivity`) | Order-tolerant lazy `.*?` — unaffected, comment added |
| `router/optimizer/pcb.py:179` (`parse_vias`) | Already fixed by PR #3446 |
| `export/ses.py:198` (`_parse_via`) | Specctra SES grammar (positional `(via "padstack" x y)`), not KiCad field order — unaffected |
| `router/io.py` `via_dia`/`via_drill`/`min_via_annular_width` | Single-field setup-token patterns, no ordering assumption — unaffected |

| Detects conflicts on KiCad-7 field order | PASS | Existing `TestMergeRoutesViaConflicts` (6 tests) still pass unchanged |
| Detects conflicts on KiCad-8 field order (uuid before net) | PASS | `test_uuid_before_net_kicad8_writer_order`, `test_mixed_kicad7_and_kicad8_order` |
| Handles `via micro` blocks | PASS | `test_micro_via_conflict_detected`, `test_micro_via_same_net_kept` |
| Handles malformed blocks | PASS | `test_malformed_block_skipped` (skips, no crash), `test_via_size_setup_token_not_matched` |
| Regression tests mirror PR #3446's parse_vias tests | PASS | Same fixture shape (multiline tab-indented KiCad-8 block) and naming as `TestParseVias` |
| Sweep for other ordered via regexes | PASS | Table above |
| No behavior change for working paths | PASS | Full `tests/test_router_io.py` suite: 179 passed |

## Test Plan

- `uv run pytest tests/test_router_io.py` — 179 passed (was 173; +6 new)
- `uv run pytest tests/test_optimizer_pcb_segments.py` — passed (no regression in #3446's tests)
- `uv run ruff check` / `ruff format --check` — the 9 lint errors and format drift in `io.py`/`test_router_io.py` pre-exist on main (verified via `git stash`); new code is clean

Closes #3447

### DIFF
--- a/src/kicad_tools/router/io.py
+++ b/src/kicad_tools/router/io.py
@@ -3511,25 +3511,33 @@ def _remove_conflicting_vias(pcb_text: str, clearance: float) -> str:
     Returns:
         PCB text with conflicting vias removed.
     """
-    # Parse via positions and net IDs from the text
-    via_pattern = re.compile(
-        r"(\(via\s+"
-        r"\(at\s+([\d.-]+)\s+([\d.-]+)\)\s*"
-        r"\(size\s+[\d.]+\)\s*"
-        r"\(drill\s+[\d.]+\)\s*"
-        r'\(layers\s+"[^"]+"\s+"[^"]+"\)\s*'
-        r"\(net\s+(\d+)\)"
-        r"[^)]*\)[^)]*\))",
-        re.DOTALL,
-    )
+    # Issue #3447: use the same balanced-parentheses walker as
+    # ``parse_vias`` (fixed in #3446) so that fields may appear in any
+    # order.  The previous implementation used a single ordered regex
+    # that required ``(net N)`` to immediately follow ``(layers ...)``.
+    # KiCad 8+ files -- including every PCB written by ``kct route``
+    # itself -- emit ``(uuid "...")`` between ``(layers)`` and
+    # ``(net)``, so the ordered pattern matched ZERO vias on modern
+    # output and conflict detection silently no-opped.  It also missed
+    # ``(via micro ...)`` / ``blind`` / ``buried`` blocks (#3118),
+    # whose type token sits between ``via`` and ``(at ...)``.
+    from .optimizer.pcb import _extract_balanced_blocks
 
-    # Collect all vias with their positions, nets, and match spans
-    vias: list[tuple[float, float, int, re.Match]] = []  # type: ignore[type-arg]
-    for match in via_pattern.finditer(pcb_text):
-        x = float(match.group(2))
-        y = float(match.group(3))
-        net = int(match.group(4))
-        vias.append((x, y, net, match))
+    _re_at = re.compile(r"\(at\s+([\d.eE+-]+)\s+([\d.eE+-]+)\)")
+    _re_net = re.compile(r"\(net\s+(\d+)\)")
+
+    # Collect all vias with their positions, nets, and block spans
+    vias: list[tuple[float, float, int, tuple[int, int]]] = []
+    for start, end, block in _extract_balanced_blocks(pcb_text, "via"):
+        m_at = _re_at.search(block)
+        m_net = _re_net.search(block)
+        # Both core fields are required; skip malformed blocks.
+        if not (m_at and m_net):
+            continue
+        x = float(m_at.group(1))
+        y = float(m_at.group(2))
+        net = int(m_net.group(1))
+        vias.append((x, y, net, (start, end)))
 
     if len(vias) < 2:
         return pcb_text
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

Reviewed against the four judging questions for #3447.

### 1. Circular dependency — none
`src/kicad_tools/router/optimizer/pcb.py` imports only `..layers`, `..primitives`, `.config`, `.geometry`; grep confirms nothing in `optimizer/` imports `router/io.py`. The import is also function-local inside `_remove_conflicting_vias`, so even a future cycle could not break module import time.

### 2. Keep-first semantics preserved
Verified line-by-line in the diff: document-order enumeration (sequential walker replaces `finditer`, both scan forward), skip-`i`-if-already-removed, same-net `continue`, Euclidean `dist < clearance`, remove later via `j` only, reverse-order span deletion with trailing-whitespace strip — all unchanged. Span equivalence on working KiCad-7 paths confirmed: the old regex's trailing `[^)]*\)[^)]*\)` and the balanced walker end at the same character for the `(net N) (uuid ...))` fixtures the old code handled. All 6 pre-existing `TestMergeRoutesViaConflicts` tests pass unchanged.

### 3. Same-net micro test exercises the new path
Verified empirically: on the `test_micro_via_same_net_kept` fixture, the new walker parses **2** via blocks while the old ordered regex parses **0**. So "both kept" flows through the `net_i == net_j` guard in the new code, not a vacuous parse failure. (Standalone the test would pass vacuously under old code, but paired with `test_micro_via_conflict_detected` — which fails pre-fix — it correctly pins the same-net guard.)

### 4. CI
All 8 checks pass (Test, Type Check, Lint & Format, C++ Build, DRC, round-trip smoke, both routing regressions). `mergeStateStatus: CLEAN`. Locally re-ran `tests/test_router_io.py` (179 passed) and `tests/test_optimizer_pcb_segments.py` (34 passed) in the PR worktree.

### Notes
- Reusing `_extract_balanced_blocks` instead of a third walker copy is the right call.
- The `verify_output_connectivity` sweep comment and the setup-token exclusion test (`(via_dia`/`(via_size` rejected because the char after `(via` must be whitespace or `)`) are both correct.
- Sweep table conclusions spot-checked and agree: `export/ses.py` is Specctra positional grammar, unaffected.

Ready to merge.

PR #3434, opened 2026-06-09
fix(router): target-aware in-pad stub direction — board 04 NRST unblock (refs #3411, closes #3428)

rjwalters:
Closes #3428
Refs #3411 (parent stays open until a fleet pass measures board 04 at 100%)

## Summary

Board 04 (LQFP-48 0.5mm) was stuck at 8/9 with NRST stranded. Two mechanisms, both in the QFP escape path, now deliver **9/9 (100%)** under the production recipe:

1. **Target-aware in-pad stub direction** (the issue as filed). `_try_in_pad_escape` accepts `target_direction`; when supplied, only the inner-layer stub + `escape_point` change — via placement is untouched. The QFP-alternating dispatcher computes it via the new `EscapeRouter._compute_target_direction` (nearest off-package same-net pad, snapped to a cardinal; deterministic tie-breaks: nearest by (dist², x, y), larger-|delta| axis, exact-tie precedence N>E>S>W; into-package guard). Redirected stubs are necked to the manufacturer-minimum trace width (PR #3079 precedent) because at 0.5mm pitch two adjacent full-width (0.5mm) stubs touch edge-to-edge regardless of direction. `target_direction=None` is byte-identical legacy behaviour for all five call sites.
2. **Pocket-escape rescue** (required by the empirical trace; subsumes a narrow slice of sibling #3429). Redirecting OSC_OUT's stub alone did NOT free NRST — verified by forcing NORTH/SOUTH/WEST stubs: all 8/9. NRST's real failure is that its parity-WEST surface escape walks AWAY from its target (J1.5, east) into a dead-end pocket whose only via slot OSC_IN already owns. New trigger (all conditions must hold, so other boards keep byte-identical escapes): standard rescue did NOT fire + non-pour net + immediately adjacent same-package pin already claimed an in-pad via + snapped target direction has negative dot with the parity escape direction. The pin then gets its own in-pad (micro-)via with the stub pointed at the target (`allow_into_package=True`: under-package inner-layer space is open for SMD packages).

Plumbing: `net_pad_positions` (net id -> [(x, y, ref)]) wired at **both** construction sites — `Autorouter._escape` (core.py, `kct route`) and `RoutingOrchestrator` (both ctor sites, `kct route-auto`) — per the known router code-path split.

## Curator call-site question answered

[...]

Issue #3402, opened 2026-06-09
demo boards 00-07: audit starting_layers per board against 100% routability bar


rjwalters:
## Background

Per user direction (2026-06-09): manufacturable = 100% LVS + 0 DRC, not "reach ≥ X%". Demo boards 00-07 should each set `starting_layers` in their project.kct based on whether they can genuinely route at 2L. If a board can't hit 100% at 2L, it should default to 4L.

Blocked by: parent issue that adds the `EscalationPolicy.starting_layers` field.

## Acceptance criteria

For each board 00-07, decide and document:

- [ ] **00-simple-led** — likely 2L (trivial design)
- [ ] **01-voltage-divider** — likely 2L (trivial design)
- [ ] **02-charlieplex-led** — measure; SOT-23-6 array may need 4L
- [ ] **03-usb-joystick** — measure; USB diff pairs may benefit from 4L for impedance control
- [ ] **04-stm32-devboard** — likely 4L (STM32 + USB + decoupling density)
- [ ] **05-bldc-motor-controller** — likely 4L (DRV8301 + STM32 + power)
- [ ] **06-diffpair-test** — likely 4L (purpose-built for diff-pair routing)
- [ ] **07-matchgroup-test** — measure; match-group constraints may need 4L
- [ ] For each board where 4L is chosen, edit `boards/NN-*/project.kct` to add `starting_layers: 4` under `routing.escalation`
- [ ] Run `uv run kct route` for each board, verify 100% reach (or document the residual gap as a separate issue)
- [ ] Update relevant routing baselines/regression tests
- [ ] No board regresses on DRC under the new starting_layers

## Methodology

For each board:
1. Run `uv run kct route boards/NN-*/output/*.kicad_pcb --auto-layers --max-layers 4`
2. Record which layer the route succeeds on
3. If escalation to 4L was needed, set `starting_layers: 4` in the spec
4. Re-route to confirm cleanly

## Out of scope

- Boards that hit 100% at 2L need no change
- Boards that don't hit 100% at 4L either — those need their own deeper-investigation issues
- Softstart (separate issue tracks this)

## Related

- Parent: starting_layers spec field issue
- 100% manufacturability bar: feedback memory `feedback_manufacturable_means_100pct.md`

Issue #3207, opened 2026-06-05
board 02 charlieplex-led: route_demo.py regressed to 4/8 nets — port recipe to route_all_negotiated + auto-layers + nudge


rjwalters:
## Context

Board 02's documented recipe \`boards/02-charlieplex-led/route_demo.py\` (invoked by \`kct build --step route\`) currently produces **4/8 nets routed with 6 DRC errors**. The underlying router can do better: the orchestrator path (\`kct route\` direct on the placed PCB) achieves **8/8 nets DRC-clean** via \`route_all_negotiated\` + auto-layer-escalation + post-route nudge.

This means \`kct build\` (the documented end-to-end pipeline for ship-readiness) cannot reach board 02 ship_ready, even though all the underlying capabilities exist on \`main\`.

Issue #3093 (closed 2026-05-22) claimed 9/10 nets routed. That completion is **no longer reproducible via the documented recipe** — surfaced during fleet audit 2026-06-05.

## Reproduction

\`\`\`bash
cd boards/02-charlieplex-led
uv run python generate_pcb.py            # OK
uv run python route_demo.py              # PARTIAL 4/8 + 6 DRC errors, exit 1
uv run kct route output/charlieplex_led_routed.kicad_pcb  # 8/8 nets, DRC-clean
\`\`\`

The recipe calls \`router.route_all()\` (bare) at \`route_demo.py:143\`; the orchestrator path uses \`route_all_negotiated(per_net_timeout=30.0, timeout=240.0)\` + \`--auto-layers\` + the post-route DRC nudge sweep added in #3112.

## Likely fix

\`boards/02-charlieplex-led/route_demo.py:143\` — change \`router.route_all()\` to \`router.route_all_negotiated(per_net_timeout=30.0, timeout=240.0)\` and add an auto-layer-escalation call + post-route nudge pass (mirror what \`kct route\` does end-to-end).

Per the project memory \`project_orchestrator_vs_autorouter_split.md\`: \"Autorouter (\`kct route\`) vs RoutingOrchestrator (\`kct route-auto\`); fixes can ship to one without the other.\" This is the practical symptom of that split — board 02's recipe is on the wrong side of the divide for its current state.

## Acceptance criteria

1. \`route_demo.py\` (or its replacement) achieves ≥ 8/10 routed nets on board 02 with no blocking DRC errors.
2. \`kct build\` end-to-end on board 02 reaches \`ship_ready=YES\` via \`kct fleet status\`.
[...]

PR #3453, opened 2026-06-10
fix(boards/03): close reach gap to 13/13 (closes #3410)

### LINKED ISSUE #3410: board 03 (usb-joystick): production routing tops out at 77% reach; placement gap (issue #3402 follow-on)
## Background

Per Issue #3402 audit (2026-06-09), board 03's production routing tops out at **77% reach** on both 2L and 4L. The 3 partial nets are USB diff pairs (USB_D+, USB_D-, USB_CC2). Setting \`starting_layers: 4\` does NOT help because the root cause is component placement / topology, not layer count.

## Reproduction

\`\`\`bash
uv run kct build-native --check  # ensure C++ backend
uv run kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb \\
  --auto-layers --max-layers 4 -o /tmp/board03.kicad_pcb
# 10/13 nets routed (77%)
# 25-33 DRC errors
\`\`\`

At 2L: 77% reach, 33 DRC errors. At 4L: 77% reach, 25 DRC errors (smaller drop only because 4L stackup eases pad-clearance budgets near USB-C).

## Why this is NOT a \`starting_layers\` decision

- 4L achieves the same partial reach as 2L; adding layers does not unblock the failing nets.
- The unrouted pads are on the USB-C receptacle (J1) where escape geometry is constrained by the pad pitch and the QFN-32 (U1) sink placement.
- The Issue #3402 audit therefore left board 03 at the default 2L probe — fixing the root cause is what gets the board to 100%.

## Scope (suggested)

- Re-place J1 / U1 / R-pulls to open routing channels for the three USB nets.
- Consider a fine-pitch clearance recipe (\`--fine-pitch-clearance\`) and/or via-in-pad for the USB-C receptacle pads.
- Acceptance: \`kct route boards/03-usb-joystick/output/usb_joystick.kicad_pcb --auto-layers --max-layers 4\` reaches 13/13 (100%) with 0 DRC at the jlcpcb tier.

## Out of scope

- Setting \`starting_layers: 4\` in the spec (would not help — Issue #3402 measurement).
- Lowering the 100% manufacturable bar (user direction fixed).
- Softstart spec (#3401).

## Related

- Parent: #3402 (per-board starting_layers audit)
- Bar: feedback memory \`feedback_manufacturable_means_100pct.md\`

rjwalters:
Closes #3410

## Summary

Board 03 (usb-joystick) production routing goes from **77% reach / 25-33 DRC errors** to **13/13 nets (100%) / 0 DRC errors at jlcpcb-tier1**. Five stacked root causes, each fixed at its source:

1. **Generator drift (board-side root cause of the 77% audit number).** `generate_design.py` carried its own stale internal 60x40mm PCB copy (pre-#2527 USB pin map) and its `main()` wrote it over the canonical `generate_pcb.py` 80x60mm layout — so the #3402 audit, the committed artifact, and CI were all measuring a board the canonical generator no longer produces. The PCB step now delegates to `generate_pcb.py` (mirrors the #3308 route-recipe consolidation).
2. **J1 USB-C re-spin** (the exit clause the DRC allowlist has named since #3183). The demo footprint mirrored the receptacle *tongue*, putting USB_D+ / USB_D− tails in a diagonal X-crossover whose same-net tie stubs physically overlapped (the allowlisted "structural" `diffpair_clearance_intra −0.200mm`) and strangled whichever pair half routed second. Same-signal tails now exit in the same column, like real USB 2.0 receptacles.
3. **MST tie-group representative** (`router/path.py`). `reduce_pads_after_intra_ic` picked the *first* pad of a stub-connected group as the MST endpoint — for J1's D− group that was A7, boxed into the escape belt, failing `BLOCKED_BY_COMPONENT` on an *empty board*. It now picks the member closest to the net's external centroid (fleet-wide, opt-in via `pad_lookup`; legacy callers unchanged).
4. **USB-C escape via-fanout** (`router/escape.py`). The #2919 alternating-layer escape packed 0.6mm vias at sub-pitch spacing inside J1 — the dominant DRC error cluster (31→2 errors when removed). Column-aligned connectors now defer the whole connector to the main router; legacy mirrored footprints (board 06's USB-C) fail the column test and keep #2919 behavior (covered by new unit tests).
5. **auto-pour deleting hand-tuned zones** (`router/auto_pour.py`). `_detect_uninset_zones` subtracted `board_origin` from `Zone.polygon` vertices that the schema layer had *already* converted to board-relative coordinates — every properly-inset zone was flagged "insufficient edge clearance", silently deleted, and replaced by auto zones with conflicting priorities ("zone will get zero copper"), stranding 30/33 GND pads.

Supporting fixes: `generate_pcb.py` now emits the full GND(F.Cu+B.Cu)/VCC/VBUS pour set + KiCad-10 header (kicad-cli rejects the 20240108 grammar with zones — this silently broke zone fill, DRC and gerber export for everything this generator produced), the modern `(locked yes)` form with schema-side parsing (`schema/pcb.py`) so router anchor logic still honors locks, joystick filter rows moved off the Steiner-junction lines (pin_access `PADS_OFF_GRID ... pad at 0.02mm` strands), and C4 moved out of the D+/D− channel.

**Recipe:** `route_pcb()` now delegates to the production `kct route` invocation pinned byte-for-byte by the baseline tests (board-02 #3207 precedent): `--seed 42 --manufacturer jlcpcb-tier1 --backend cpp --timeout 600 --raw`. `--raw` is load-bearing: the TraceOptimizer otherwise re-introduces exactly one deterministic `clearance_segment_via` (0.006mm, XTAL cluster, reproducible across seeds 42/43).

## Inherited-work disposition (recovery run)

Prior builder sessions crashed leaving only a modified `output/usb_joystick.kicad_pcb`. Analysis showed it was just a fresh regeneration from the (already-drifted-ahead) `generate_pcb.py` plus one hand nudge (C4 → (146,112)). The orphaned output diff was **reset**; the C4-nudge insight was **kept** and implemented properly in the generator. The advertised `generate_pcb.py`/`net_class_map.json` changes had not survived the crashes.

## Baselines (updated honestly)

- Baseline on current main before this work (control, `kct route` seed 42 on the then-committed board): **11/13**, 27 DRC errors — i.e. the 11/13 historical floor had recovered from PR #3434's 10/13 measurement.
- `tests/router/test_board03_routing_baseline.py`: floor 11→**13**/13; per-net USB contracts all hard; committed-artifact ceiling 4→**0** errors. The per-net stdout test was rewritten: its old `ROUTING PREVIEW` parser only fires under interactive `--preview`, so it had **never** passed — nightly slow-tests don't gate on the pytest exit code (run 27183774151 shows `9 failed ... 7 errors` under a green check; that wider workflow gap is noted below).
- `tests/test_board_03_regression.py`: `MIN_FULLY_ROUTED_NETS` 11→**13** (route_demo path); in-process `route_all` fixture re-scoped as a router-capability net.
- `.github/routed-drc-tolerance.yml`: board 03 tightened **5 → 0**.
- Committed artifacts regenerated end-to-end (unrouted PCB, routed PCB, schematic, net-class sidecar, manufacturing bundle). `kct check --mfr jlcpcb-tier1` = 0 errors with and without the sidecar.

## Cross-board safety

- Router changes are opt-in/narrow: tie-group representative only changes when `pad_lookup` is passed (4 Autorouter call sites); escape defer only fires for column-aligned connectors (board 06's mirrored USB-C keeps legacy behavior — pinned by new tests); auto-pour fix restores the documented coordinate-space invariant (79 auto-pour tests pass).
- Suites run locally: tests/router (267 passed; only pre-existing failures), escape suites (194), autorouter (68), schema+locks (179), auto-pour (79), diffpair/validate-side (93), board-03 fast+slow baseline class (all green). `test_router_core.py` has 2 failures that reproduce identically at HEAD (pre-existing).

## Follow-ups worth filing (called out, not in scope)

1. TraceOptimizer can violate via clearance (grid-quantized collision check vs world-coord DRC) — why this board needs `--raw`.
2. Diff-pair detection pollutes class-name-keyed lookups: USB_CC2 gets `diffpair_partner='USB_D-'` via the shared HighSpeed entry (bogus skew/continuity advisories in kct route's internal DRC; the committed sidecar is unaffected).
3. C++ backend silently falls back to Python for the USB nets on this board (multi-minute grinds; `fallback_nets=['USB_D+','USB_D-','USB_CC1']`).
4. `schema/pcb.py` save path still re-emits the legacy in-attr `locked` token (kicad-cli-incompatible) via `_sync_attr_node`.
5. Slow-tests workflow does not gate on the pytest exit code (nightly green with 9 failures).

## Test plan

- [x] `kct route <board> --seed 42 --manufacturer jlcpcb-tier1 --backend cpp --timeout 600 --raw` → 13/13, exit 0
- [x] `kct check routed --mfr jlcpcb-tier1` → 0 errors (also with `--net-class-map` sidecar)
- [x] `route_demo.py` end-to-end → "SUCCESS: All nets routed, DRC passed!"
- [x] Issue repro command (`--auto-layers --max-layers 4`) reaches 13/13
- [x] Baseline test class + committed-DRC-state test green
- [x] Escape / auto-pour / schema / diffpair suites green


rjwalters:
✅ **Approved!**

Second demo board to fully manufacturable (13/13 routed, 0 DRC at jlcpcb-tier1, allowlist 5 → 0). All 8 CI checks pass, mergeStateStatus is CLEAN.

## Blast-radius assessment (the three router changes)

**Fix 3 — MST tie-group representative (`router/path.py`)**: Strictly opt-in. The new behavior only engages when `pad_lookup` is passed; the only callers passing it are the 4 `Autorouter` sites in `router/core.py`. Legacy callers get byte-identical behavior (verified `indices[0]` path preserved, pinned by the new `test_reduce_pads_picks_externally_facing_representative` which asserts BOTH the new and legacy behaviors). Fleet-wide effect on tie-group boards is intentional and gated live: board-07 Match-Group Regression passed (10m15s).

**Fix 4 — column-aligned USB-C escape defer (`router/escape.py`)**: Narrowly scoped inside `_escape_usb_c_connector` (fires only for `USB_C_CONNECTOR` package type), and the gate requires every netted column be single-net AND at least one true same-net column pair. Board 06's tongue-mirrored footprint fails the column test by construction; pinned by `test_legacy_connector_still_generates_escapes` + `test_legacy_mirrored_connector_is_not_column_aligned`, and live-verified: board-06 Diff-Pair Regression passed (12m30s). 211 escape-suite tests pass locally.

**Fix 5 — auto-pour double origin-subtraction (`router/auto_pour.py`)**: Verified the fix matches the documented coordinate-space invariant (`schema/pcb.py:1800`: `Zone.polygon` is board-relative after `_detect_board_origin()`). No other committed artifacts are touched, and the gating Routed PCB DRC Check (committed artifacts) + board-06/07 regeneration jobs are green. One coverage gap: all existing `tests/test_auto_pour.py` zone-preservation tests use origin (0,0), where the old bug was a no-op — a re-regression would only surface in non-gating slow tests (#3458). Filed **#3461** to add a non-zero-origin unit regression test (non-blocking).

## `--raw` recipe pin

Acceptable as a pinned workaround: #3454 is well-formed (deterministic repro across seeds 42/43, exact coordinates, root-cause hypothesis: grid-quantized `GridCollisionChecker` vs world-coordinate DRC), and the rationale is documented at all three sites that must stay in sync (`generate_design.py:route_pcb`, `_run_kct_route` in the baseline test, tolerance YAML comment). This is hiding-with-a-tracking-issue done correctly, mirroring the board-02 #3207 precedent.

## Generator delegation

No consumer imports the removed internal PCB generator — references to `generate_design` in tests/scripts all go through `route_pcb()`/`main()`, whose interfaces are preserved. Fast board-03 regression tests pass (7 passed, 1 skipped).

## Other verification

- `(locked yes)` parser: verified both modern and legacy forms parse to `fp.locked=True`; 179 schema/lock tests pass.
- The 2 `test_router_core.py` failures reproduce identically on main (pre-existing, as claimed).
- Committed sidecar is clean (USB_D+/USB_D- properly cross-partnered; no #3455 pollution in the artifact).
- Baseline test rewrite is honest about the never-passing `ROUTING PREVIEW` parser and replaces it with a contract on actual CLI summary output.
- Conflict check: no other open PRs; #3398/#3448/#3452 builders have worktrees but no PRs yet, so this merges first and they rebase.

## Follow-up quality

#3454/#3455/#3456/#3457 all exist with root causes, repro details, and observed evidence. **#3458 is the important one** — nightly slow-tests green with `9 failed, 7 errors` is a second CI-trust gap (after #3436) and is exactly how this board's floors drifted unnoticed; recommend prioritizing it so the new 13/13 floors actually gate.

## Test Execution (from PR test plan)

1. `kct route ... --raw` → 13/13 — ⏭️ not re-run locally (>2 min); equivalent coverage via Test + board CI jobs and committed-artifact DRC gate
2. `kct check routed --mfr jlcpcb-tier1` → 0 errors — ✅ verified via Routed PCB DRC Check (tolerance 0) passing
3. `route_demo.py` end-to-end — ⏭️ long-running; recipe delegation reviewed
4. Issue repro (`--auto-layers --max-layers 4`) — ⚠️ not executed, taken on builder's word (advisory; production recipe is the gated path)
5. Baseline test class — ✅ fast subset run locally (7 passed); slow class covered by builder run
6. Escape / auto-pour / schema / diffpair suites — ✅ 48 + 211 + 179 + 172 tests run locally, only the 2 pre-existing main failures

Fine work — five stacked root causes each fixed at the source, honest re-baselining, and narrow, well-tested router changes.

Issue #3438, opened 2026-06-10
router: board 07 stuck at 28/31 — negotiated router cannot complete parallel pad-array bundles (DDR bundle ALONE routes 9/11)


rjwalters:
Parent: #3414 (board 07 84→100% reach gap; decomposed by the #3414 recovery PR)
Related: #3275 (closed — prior 3-net thread), #3402 (audit), #3434 (open PR — see regression warning below)

## Root-cause classification (from #3414 investigation)

Board 07's production recipe (negotiated, 4L, seed 42) deterministically strands 3 nets: `DQ3`, `MIPI_CLK_N`, `MIPI_DAT0_N` (28/31 = 90%). The #3402 audit hypothesized the missing `--length-match-groups` CLI was the gap — **disproven**: the flag (closed by #2723 / PR #3098) is wired into the recipe and the gap persists.

The blocker is the **negotiated single-ended router's ordering/escape dynamics on parallel pad-array bundles**, demonstrated by this isolation matrix (all on main @ ca8e6899+, C++ backend, solo unless noted):

| Configuration | Result |
|---|---|
| Full board, production recipe | 28/31 (DQ3, MIPI_CLK_N, MIPI_DAT0_N) |
| ONLY the 3 stranded nets (others skipped) | **3/3 in 0.2s** |
| ONLY the 6-net MIPI bundle | **6/6 in 11s** |
| ONLY the 11-net DDR bundle (empty board otherwise) | **9/11 — DQ5, DM0 fail** (174s) |
| Pre-route the 3 hard nets, then rest `--preserve-existing` | 29/31 — failures SHIFT to partners (MIPI_CLK_P, MIPI_DAT1_P) |

Key evidence: the **DDR bundle alone fails 2/11 on an otherwise empty board** — 11 straight parallel nets between two facing QFN-48 pin columns (0.8mm pitch, 30mm apart, 4 layers). This is trivially hand-routable; the failure set is order-dependent chaos, not geometry. Prior pin-swap experiments (#3414 worktree) confirm: swap which nets sit on the failing pins and the failures follow the pin POSITIONS, i.e., whichever net drains last in the negotiated order finds its pin-access corridor consumed by siblings, and rip-up cannot recover (iteration 0 is the high-water mark; later iterations regress — the #3101 pattern).

## Knob matrix (none reach 31/31)

| Knob | Result |
|---|---|
[...]

Issue #3414, opened 2026-06-09
board 07 (matchgroup-test): 4L production routing reaches only 84% (target: 100%) — match-group gap (issue #3402 follow-on)


rjwalters:
## Background

Per Issue #3402 audit (2026-06-09), board 07 has \`starting_layers: 4\` (the board mirrors board 06's 4-layer JLCPCB tier-1 stackup with inner GND/PWR planes).

At 4L the production router reaches **84%** (26/31 nets) — below the 100% manufacturable bar but the closest to it of the multi-layer demo boards.

## Reproduction

\`\`\`bash
uv run kct build-native --check
uv run kct route boards/07-matchgroup-test/output/matchgroup_test.kicad_pcb \\
  --auto-layers --max-layers 4 -o /tmp/board07.kicad_pcb
# 26/31 routed (84%); 30 DRC errors
\`\`\`

## Diagnosis (audit observations)

- Auto-layers detected the 4L stackup (issue #2916) and ran a single 4L SIG-GND-PWR-SIG attempt.
- Reach: 26/31 (84%); 30 DRC errors.
- Existing routed artifact (\`matchgroup_test_routed.kicad_pcb\`) is in the repo with the Phase 3N allowlist budget.

## Scope (suggested)

- Identify the 5 unrouted nets and trace the failure cause (most likely the DDR data byte or HDMI TMDS groups around U1 / U4 BGA escapes).
- The Phase 3H \`--length-match-groups\` CLI flag is not yet wired (per the board's own checklist) — that may close some of the gap once it lands (#2723).
- Acceptance: \`kct route boards/07-matchgroup-test/output/matchgroup_test.kicad_pcb --auto-layers --max-layers 4 --length-match-groups\` reaches 31/31 (100%) with 0 DRC (or the iteratively-tightened Phase 3N allowlist).

## Out of scope

- \`starting_layers: 4\` is already set (Issue #3402).
- Phase 1A-2G match-group feature exercises remain — see \`tests/test_board_07_matchgroup_test.py\` (and its sidecar emission contract).

## Related

- Parent: #3402 (per-board starting_layers audit)
- Existing board-07 thread: Epic #2661, #2723 (length-match-groups CLI), #2916 (inner-plane detection)
- Bar: feedback memory \`feedback_manufacturable_means_100pct.md\`

rjwalters:
Recovery run #3 result (PR #3446 — deliberately does NOT close this issue):

[...]

PR #3462, opened 2026-06-10
fix(router): convergence requires all nets routed, not just overflow==0 (refs #3438, closes #3448)

### LINKED ISSUE #3448: router: negotiated loop declares 'Convergence achieved' at overflow==0 with unrouted nets and unused wall budget
## Bug

The negotiated routing loop treats `overflow == 0` as convergence and exits — even when routable nets remain unrouted and most of the wall-clock budget is unused. This is a loop-control defect separable from the bus-reversal crossing-assignment problem tracked in #3438 (where it was diagnosed as "mechanism defect 3" / "fix direction 2"), and it silently caps reach on ANY board where a net hard-fails with an empty A* frontier.

## Evidence (board 07, solo, seed 42 — from #3438 recovery run #3)

- Hard-failed nets (instant empty frontier, e.g. DQ3 in 0.04s) produce **zero overflow** — they never route-through-at-cost, so PathFinder receives no congestion signal.
- With `--timeout 1500` the loop prints "Convergence achieved" and exits at ~820s with 28/31 routed — **~680s of budget abandoned** with 3 nets unrouted.
- Minimal repro: 11-net DDR bundle alone (<40s) — see #3438 comments for the repro configs.

## Expected behavior

`overflow == 0 AND unrouted nets remain AND wall budget remains` must NOT be terminal. Options (from #3438):
- Fall through to a restart/perturbation pass (reorder, seed jitter) instead of exiting.
- Run a relief pass that treats foreign channel-route cells (not pads) as finite-cost for the hard-failed nets, generating real overflow for the next negotiation iteration.

## Acceptance

- Negotiated loop never reports convergence while unrouted-but-routable nets remain and budget is available; it either makes progress or exhausts budget/explicit stall criteria.
- Board 07 repro: loop continues past the current ~820s exit point under `--timeout 1500` (reach improvement itself is tracked by #3438).
- No regression on boards 06/07 CI gates.

Related: #3438 (parent diagnosis), #3414 (board 07 manufacturable bar, blocked), #3101 (iteration-0 high-water pattern).

Split out of #3438 during the judge review of PR #3446 because it is an independently fixable loop-termination defect with its own crisp acceptance criterion.

rjwalters:
## Summary

Closes #3448

The negotiated loop declared "Convergence achieved" at `overflow == 0` while nets remained stranded, abandoning most of the wall-clock budget (board 07: exit at ~820s of a 1500s budget with 3 nets unrouted). Diagnosed as mechanism defect 3 in #3438 (split out during the PR #3446 judge review); also relevant to #3414.

### Root cause (two predicate holes in `len(net_routes) == total_nets`)

1. **Empty-list entries**: `rip_up_nets` leaves `net_routes[n] == []` behind. A net that is ripped up and then hard-fails its re-route (instant empty A* frontier => ZERO overflow, e.g. board-07 DQ3 in 0.04s) still counts toward the total — AND was invisible to `failed_nets_to_recover` (`n not in net_routes` is False for the empty entry), so the loop had no work item for it either.
2. **Partial fragments**: a multi-pad net whose main A* failed can sit in `net_routes` with fragments (subset of MST edges) that connect only some of its pads; the #2475 partial detector fed these into rip-up at the TOP of each iteration, but the convergence break at the BOTTOM fired first.

### Fix

- New `_stranded_nets()` closure (pad-connectivity union-find over routable nets, excluding single-pad/off-grid) gates:
  - all four `Convergence achieved` break sites (targeted + standard branches, plus both post-neighborhood-rip-up breaks),
  - the iteration-0 `No conflicts - routing complete!` early return,
  - the final progress-callback status — now reports `stranded` instead of `converged` when overflow==0 with stranded nets.
- Empty-list entries now count as failed across the recovery machinery: `failed_nets_to_recover` (#858), zero-overflow elevated-cost recovery (#2396), stagnation-recovery re-add (#2515), targeted-fallback `still_failed` (#2265), both BLOCKED_BY_COMPONENT recomputes, the `should_terminate_early` `unrouted_count` guard (#2297), and termination diagnostics.
- When overflow is clean but nets are stranded, the loop logs `Overflow clean but N net(s) still stranded -- continuing (Issue #3448): <names>` and keeps working (targeted fallback, neighborhood rip-up, per-net retries) until budget or explicit stall criteria (#3101 patience, max_iterations) fire — never a premature convergence claim.

### Board 07 (matchgroup-test, solo, seed 42, 600s recipe)

| | baseline (main, PR #3446 re-measure) | this PR |
|---|---|---|
| reach | 28/31 | **29/31** |
| loop exit | premature convergence (~453s in-loop equivalent; ~820s under 1500s budget) | timeout at **598.9s** — budget fully consumed |
| gate DRC | 21 raw (allowlist 25) | **21 raw** (allowlist 25) — gate EXIT=0 |

The previously-lying iteration log is visible in the run: iteration 1 prints `Progress: 31/31 nets routed total` (the broken count) while 2 nets are stranded; old code broke out right there — new code runs targeted fallback for the remaining ~145s and the stranded-continue log fires. Refreshed `matchgroup_test_routed.kicad_pcb` (29/31) included; passes `check_routed_drc.py` (16 blocking <= 25 allowlist; not tightening the allowlist because the matchgroup gate counts raw errors (21) against the same value — see #3151).

### Cross-board regression (fresh re-routes of committed recipes)

| board | result | baseline/floor |
|---|---|---|
| 02 charlieplex | 8/8 SUCCESS | 8/8 |
| 03 usb-joystick | 12/13 | floor 11/13 |
| 04 stm32-devboard | **9/9 SUCCESS** | 9/9 |
| 05 bldc (fresh, python backend) | 15/32 | main A/B same session: 16/32 — within run-to-run noise (recipe has no PYTHONHASHSEED pin; committed snapshot untouched) |
| softstart | canonical `test_softstart_routing_reach_regression` **passed** (>= 6/10 floor) | 6/10 |
| 06 diffpair | covered by CI `diffpair-routing-regression` gate (re-routes every PR) | — |

### Tests

- New `tests/test_negotiated_convergence_stranded.py` (4 tests): partial-net blocks completion/convergence; final status `stranded` not `converged`; the board-07 empty-entry trap (partial at iter 0 -> ripped -> hard-fail with budget remaining => loop continues, NOT converged); genuine convergence still fires. **3 of 4 fail on unmodified main**, all pass with the fix.
- `test_best_metric_patience` + `test_router_negotiated_high_current`: 40 passed.
- `test_reroute_progress_output` + `test_negotiated_adaptive`: 132 passed.
- `test_router_core` + `test_checkpoint_in_escalation` + `test_negotiated_congestion_model` + `test_region_parallel`: 223 passed, 3 failed — all accounted for: `test_route_to_pad_adjacent_to_net0_pad` and `test_add_component_rotation` fail identically on unmodified main (pre-existing); `test_route_400x400_completes_within_budget` is a perf budget that failed only under 3-concurrent-agent load and passes solo (26.95s).

## Test plan
- [ ] CI: Match-Group Routing Regression (board 07) green — verified locally end-to-end (EXIT=0, 21 <= 25)
- [ ] CI: Diff-Pair Routing Regression (board 06) green
- [ ] CI: routed-pcb-drc-check green on the refreshed board 07 artifact (verified locally: 16 blocking <= 25)
- [ ] CI: unit suite green (pre-existing failures listed above fail identically on main)

Refs #3438. Refs #3414.


rjwalters:
✅ **Approved!**

Verified all five judging questions for the convergence-predicate fix (Issue #3448):

**1. Termination safety — SAFE.** The stranded-continue path removes only the four "Convergence achieved" break sites; it cannot spin forever because three independent bounds remain: (a) the hard `for iteration in range(1, max_iterations + 1)` loop (core.py:7020), (b) `check_timeout()` at the top of every iteration (core.py:7022), and (c) the #3101 best-metric patience break (core.py:7179) — a permanently-unroutable net produces no lex-tuple improvement, so `best_stall_count` reaches patience and the loop exits, with the final status branch correctly reporting `stranded`. The `unrouted` correction at the `should_terminate_early` site only defers stagnation exit to those other bounds, never removes them.

**2. `_stranded_nets()` cost — acceptable.** It is evaluated only at `overflow == 0` checkpoint sites, and `_compute_nets_fully_connected()` (the same `_validate_conn` union-find over all routes) already runs unconditionally every iteration at core.py:7076 — so this adds no new asymptotic per-iteration cost. The 400x400 perf-budget test exercises the raw A* path, which this PR does not touch, and it passed in the CI Test job.

**3. Board 03 floor collision (PR #3453) — CLEARED EMPIRICALLY.** This branch's merge-base (630ce2cf) predates #3453, and the branch carries the old floor (`REQUIRED_NETS_ROUTED = 11`); however the PR does not modify `tests/router/test_board03_routing_baseline.py`, so the merged tree keeps main's 13/13 floor AND main's rebuilt board 03 geometry. I built a clean local merge (no conflicts; #3453's core.py overlap is just the `pad_lookup` kwarg; #3459 is escape.py-only), compiled the C++ backend, and ran the merged tree:
- `TestBoard03RoutingBaseline::test_reach_meets_floor` (the 13/13 floor) — **PASSED** (119s)
- All 4 new tests in `tests/test_negotiated_convergence_stranded.py` — **PASSED** (1.5s)

The builder's 12/13 cross-board number was measured against the pre-#3453 board and is superseded; the post-merge nightly slow-tests floor holds at 13/13.

**4. Board 07 artifact — strictly better.** Apples-to-apples (`kct check --mfr jlcpcb --errors-only --format json`): pre-PR committed artifact = 10 errors, refreshed artifact = **9 errors**, with reach improved 28/31 → 29/31. Routed PCB DRC Check gate passes.

**5. CI — all 8 checks pass, mergeStateStatus CLEAN.** Including both routing-regression gates (Match-Group 12m44s, Diff-Pair 12m33s) and the full unit Test job (11m30s).

Test quality note: the new tests reproduce the exact board-07 mechanism (partial at iter 0 → rip-up leaves `net_routes[n] == []` → instant zero-overflow hard fail) via a deterministic monkeypatch rather than fragile geometry, and include a genuine-convergence positive control. 3 of 4 fail on unmodified main, confirming they pin the fix.

One operational note (non-blocking): by design this fix makes the negotiated loop consume the full wall-clock budget whenever nets are stranded, so routing-gate CI jobs and `kct route` runs on incomplete boards will take longer than before — that is the intended behavior change.

`closingIssuesReferences` verified: [3448] only.

Issue #3464, opened 2026-06-10
ci: Lint & Format job has continue-on-error on both steps (lint rot is invisible, 862 ruff errors / 505 unformatted files on main)


rjwalters:
Found while fixing #3458 (the slow-tests workflow had the same pattern).

`.github/workflows/ci.yml` `lint` job marks BOTH steps `continue-on-error: true`:

```yaml
      - name: Check formatting
        run: uv run ruff format . --check
        continue-on-error: true

      - name: Lint
        run: uv run ruff check .
        continue-on-error: true
```

so the required "Lint & Format" check is always green regardless of outcome. Measured on main (723a9ff7, locked ruff 0.14.10):

- `uv run ruff check .` -> **862 errors** (277 F401 unused-import, 189 I001 unsorted-imports, 126 F841 unused-variable, ...)
- `uv run ruff format . --check` -> **505 files would be reformatted** (903 already formatted)

Fixing is bigger than a one-liner because removing `continue-on-error` instantly breaks every PR until the backlog is cleared. Suggested sequence:

1. `uv run ruff check . --fix` + `uv run ruff format .` in one mechanical PR (612+ auto-fixable), review the non-auto-fixable remainder by rule family.
2. Then drop both `continue-on-error: true` lines so the gate is real.
3. Consider `--output-format=github` for inline annotations.

Not bundled into the #3458 PR to keep that change reviewable (it covers the nightly slow-tests gate + #3460).

rjwalters:
## Curator Enhancement (verified on main, 2026-06-11)

### Re-verified counts (30 PRs merged since filing — backlog is stable, not shrinking)

| Check | At filing (723a9ff7) | Now (main, ruff 0.14.10 locked) |
|---|---|---|
| `uv run ruff check .` | 862 errors | **860 errors** (636 safe auto-fixable, +187 more behind `--unsafe-fixes`) |
| `uv run ruff format . --check` | 505 files | **510 files** would be reformatted (907 already formatted) |

Rule breakdown (top): F401 unused-import 273, I001 unsorted-imports 189, F841 unused-variable 126, UP037 75, F541 52, C420 27, SIM105 16, B905 15, C408 15, SIM114 14, F811 9, E712 9, SIM300 8, **F821 undefined-name 7**, plus tail.

### Workflow references (current main)

- `.github/workflows/ci.yml:30-36` — the two `continue-on-error: true` lines in scope for this issue (format step line 32, lint step line 36).
[...]

Issue #3463, opened 2026-06-10
route --auto-mfr-tier: escalation aborts before jlcpcb-tier1 attempt on board 04 (3 slow tests failing)


rjwalters:
Found while fixing #3458 (nightly slow-tests gate). Reproduced on `feature/issue-3458` (base 723a9ff7, native C++ backend built):

```
uv run pytest tests/test_route_auto_mfr_tier_integration.py -m slow -o addopts= --no-cov --timeout=1200
FAILED ...::TestAutoMfrTierIntegration::test_tier1_routes_more_nets_than_jlcpcb
FAILED ...::TestAutoMfrTierIntegration::test_escalation_advances_to_tier1
FAILED ...::TestAutoMfrTierIntegration::test_escalation_triggered_by_missed_via_in_pad
3 failed in 277.42s
```

The `--auto-mfr-tier --max-layers 2` run on the board 04 fixture ends with:

```
Result: No tier in jlcpcb -> jlcpcb-tier1 achieved routing success.

Diagnosis:
  Cannot escape component U2 to perimeter on jlcpcb. The `jlcpcb` profile does not support via-in-pad (via_in_pad_supported = false).
```

and the per-tier banner for `jlcpcb-tier1` never appears (`per_tier_stdout` only contains `['jlcpcb']` -- confirmed both locally and in the 2026-06-10 nightly run 27253279489, which was silently green because of #3458). The escape-infeasibility diagnosis on the BASE tier is exactly the condition that should trigger escalation to jlcpcb-tier1 (which supports via-in-pad), but the ladder aborts instead of advancing. Suspect interaction with the recent convergence/feasibility changes (#3462 landed last night; the same failure signature predates it in the nightly, so the regression window is earlier -- likely the escape pre-flight check path).

These three tests were xfail'd (non-strict) with a reference to this issue in the PR for #3458 so the slow-tests gate can land; remove the xfails when fixing this.

Issue #3436, opened 2026-06-09
CI Test job masks all test failures: continue-on-error true hides 162 failed + 30 errors on main


rjwalters:
## Problem

The `Test` job in `.github/workflows/ci.yml` (line 71) runs:

```yaml
- name: Run tests
  run: uv run pytest -n auto -o addopts= --benchmark-disable --timeout=60 -m "not slow"
  continue-on-error: true
```

`continue-on-error: true` makes the job green **regardless of test results**. CI run 27242706499-era main (run 27235563106, commit 3dc88588) shows:

```
= 162 failed, 18675 passed, 340 skipped, 1 xfailed, 193 warnings, 30 errors in 543.04s =
```

…and the job still reported success. This caused the #3427 "local-vs-CI divergence" investigation: the two fine-pitch tests were failing in CI all along (since PR #2866, 2026-05-13), but the green checkmark hid it. **Any** new test regression is currently invisible to CI, Judges, and the merge queue.

The Lint & Format job appears similarly masked (`Would reformat:` output on green runs).

## Failure categories observed on main (run 27235563106)

- `FileNotFoundError: Library not found: Comparator.kicad_sym / Device.kicad_sym` (environment: KiCad symbol libraries absent on CI runner) — many tests
- `MockPCB object has no attribute 'nets'` (test_drc_tolerance.py — stale mock)
- pytest-timeout 60s kills (test_diffpair_npad.py etc.)
- Assorted stale assertions (test_fine_pitch_clearance.py — fixed in PR for #3427, test_cli_sch.py, test_diffpair_length_tuning.py, …)
- 30 collection/setup errors

## Suggested approach

1. Triage the 162 failures + 30 errors into: environment-dependent (needs skip-if-missing guard or CI fixture install), stale tests (update), real defects (fix or file).
2. Burn the list down to zero (possibly via a tracking issue with sub-issues per category).
3. Remove `continue-on-error: true` from the Test step so CI is load-bearing again.
4. Consider the same for the Lint & Format and mypy steps.

## Acceptance

- `continue-on-error: true` removed from the Test step
- CI Test job is red iff tests fail

Discovered while implementing #3427.

PR #3465, opened 2026-06-10
fix(ci): slow-tests gate on pytest exit + check_routed_drc --pcb flag consistency (closes #3458, closes #3460)

### LINKED ISSUE #3458: slow-tests workflow does not gate on pytest exit code (nightly green with 9 failures)
Found during #3410 / PR #3453.

Nightly 'Slow Tests' run 27183774151 (2026-06-09) reports SUCCESS while its pytest summary shows `9 failed, 9 passed, 24 skipped, 7 errors`. Failures included tests/router/test_board03_routing_baseline.py (its per-net stdout parser had never worked — rewritten in PR #3453) and a generate_schematic failure in test_board_03_regression.

Either the workflow step swallows the exit code or a reporting step masks it. Until fixed, every 'slow' reach/DRC floor in tests/router/ is advisory-only and regressions land silently — which is exactly how board 03's floors drifted.

Check: .github/workflows/slow-tests.yml run step ('uv run pytest -n auto -o addopts= --benchmark-disable --timeout=120 -m slow --cov=...' — also audit whether --timeout=120 plus multi-minute route subprocess tests can ever pass deterministically).

### LINKED ISSUE #3460: test(router): board06 determinism test broken by check_routed_drc CLI drift (--pcb flag removed)
Found during #3398 verification (PR #3459 baselines).

`tests/router/test_board06_determinism.py::TestBoard06DRCDeterminism::test_drc_counts_identical_across_five_runs` fails on **unmodified main** (verified Jun 9 2026, stashed-main run):

```
E  Failed: Could not find DRC error count line in check_routed_drc.py output.  stdout:
E
E  stderr:
E  usage: check_routed_drc [-h] [--allowlist ALLOWLIST] [files ...]
E  check_routed_drc: error: unrecognized arguments: --pcb /private/var/.../determinism/run-1.kicad_pcb
```

The test invokes `scripts/ci/check_routed_drc.py --pcb <file>` (test file line ~114 region), but the script's CLI now takes positional `files` and has no `--pcb` flag — so every per-run DRC count probe fails before any determinism comparison happens.

This is gated behind `KICAD_RUN_SLOW_BOARD06_DETERMINISM=1`, so PR CI never runs it; whichever dedicated job invokes it (or the next person running it locally) gets a false regression signal — the board-06 determinism gate is currently providing zero coverage.

Fix: update the test's invocation to pass the PCB path positionally (and assert on the script's current output format), or restore a `--pcb` alias in `check_routed_drc.py`.

rjwalters:
Closes #3458
Closes #3460

## Summary

Two CI-trust fixes in one PR:

1. **#3458 — nightly slow-tests never gated on pytest.** `.github/workflows/slow-tests.yml` had `continue-on-error: true` on the pytest step, so run 27253279489 (2026-06-10) was SUCCESS with `9 failed, 8 passed, 24 skipped, 8 errors`. Every slow reach/DRC floor was advisory-only — which is how board 03's floors drifted unnoticed. The swallow is removed and the job environment fixed so it can actually go green (see triage below).
2. **#3460 — board06 determinism test passed a `--pcb` flag the checker no longer has.** `check_routed_drc.py`'s CLI is positional `files...`; argparse exited 2 with `unrecognized arguments: --pcb` before any determinism comparison ran. Mechanism for "still passes": the entire module hides behind `KICAD_RUN_SLOW_BOARD06_DETERMINISM=1` (no CI lane sets it — verified by grep), and even a lane that did would have been masked by #3458's swallowed exit code. Zero coverage, doubly masked.

## Workflow changes (slow-tests.yml)

- Removed `continue-on-error: true` from the pytest step — the job now fails when pytest fails.
- **Build the C++ router extension** (`kct build-native` + `--check`): `uv sync` does not build it, so every nightly ran the pure-Python A* at 10-100x slower — the direct cause of most of the timeout inventory below.
- `--timeout=120` → `--timeout=1200`: slow routing tests have *internal* 600-900s subprocess budgets (`test_route_auto_mfr_tier_integration` even carries `@pytest.mark.timeout(900)`); a 120s pytest-timeout could never pass them deterministically, exactly as #3458 suspected.
- `apt-get install kicad-symbols`: the board 03 `regenerated_board` fixture runs `generate_schematic.py`, which resolves symbols from system KiCad libraries (`FileNotFoundError: Library not found: Connector_Generic.kicad_sym` on the bare runner). Plus a defensive `pytest.skip` in the fixture for environments without libraries.
- Job `timeout-minutes` 60 → 90 to accommodate real (non-killed) slow-test runtimes.

## Failure inventory (nightly run 27253779489, 2026-06-10) and disposition

| Test(s) | Nightly symptom | Disposition |
|---|---|---|
| `test_board03_routing_baseline.py` (4 tests: reach floor, per-net USB, seeds-determinism, 1-of-16 myth) | 2 setup timeouts, 1 timeout, 1 cascading parse failure | **Fixed upstream** by #3453 (merged after the nightly): all 4 pass locally in 8m48s with native backend. Needed: build-native + timeout raise (this PR). |
| `test_chorus_test_placement_feedback.py` (3 tests) | Timeout >120s | **Fixed by env**: 3/3 pass in 42s locally with native backend. |
| `test_board_03_regression.py::test_usb_diff_pair_routes_via_coupled_pathfinder`, `::test_xtal2_failure_classified_as_trace_blocker` | Setup timeout >120s | **Fixed by env**: build-native + `--timeout=1200` (fixtures have internal 600-900s budgets). |
| `test_board_03_regression.py::test_route_demo_achieves_minimum_completion` | Fixture error: `Library not found: Connector_Generic.kicad_sym` | **Fixed by env**: `kicad-symbols` apt install + skip guard in the fixture. |
| `test_board_04_routing_regression.py::test_routes_all_nets_on_2l`, `test_board_05_routing_regression.py` (2 tests) | Setup timeout >120s | **Fixed by env**: build-native + timeout raise (same class of failure as board03; board-04/05 floors re-verified by their own per-PR gates). |
| `test_kct_check_parity.py` (gate test in nightly; 3 tests failing on HEAD) | `expected 17 blocking, got 15` (nightly) / `got 16` (HEAD) | **Trivially fixed (re-baseline)**: PR #3462 refreshed board 07's routed PCB without re-pinning (the silent-drift failure mode this PR ends). Re-pinned per the test's own instruction: match_group delta 1→2 (5+5+2=12), bare 9 / sidecar 21 totals, gate blocking 16, advisory connectivity 5; tolerance floor tightened 25→16 in the same PR per allowlist policy. 5/5 parity tests pass locally. |
| `test_route_auto_mfr_tier_integration.py` (3 tests) | Escalation never advances off `jlcpcb` | **Real regression — xfail with tracking issue #3463** (non-strict): `--auto-mfr-tier` aborts on the U2 via-in-pad escape infeasibility instead of escalating to `jlcpcb-tier1`. Reproduced on HEAD with native backend (3 failed in 4m37s). Remove xfails when #3463 lands. |

Also filed **#3464**: the per-PR `Lint & Format` job has the identical `continue-on-error: true` pattern (862 ruff errors / 505 unformatted files invisible on main) — kept out of this PR because un-masking it requires clearing the backlog first.

## check_routed_drc / #3460 changes

- `tests/router/test_board06_determinism.py` now passes the PCB path positionally via a shared `_drc_checker_cmd()` helper, accepts the gate's real exit-code contract (0/2), parses all four real output formats (`OK: ... -- N errors`, `::error ...: N blocking error(s)`), and `pytest.fail`s explicitly on `unrecognized arguments` (CLI-drift detection).
- New always-on `TestDRCCheckerCLIInterface` (the env-gate moved from module-level `pytestmark` to the slow class only): pins that the script accepts the determinism arg shape, that legacy `--pcb` stays rejected, and unit-tests the output parser. Runs in PR CI — 3 passed locally.
- Caller audit: ci.yml passes positional paths; all other references import the module (fleet_cmd, net_status, test_ci_routed_drc_workflow, test_check_routed_drc_advisory). The determinism test was the only `--pcb` straggler. 68 passed across the checker test suites.

## Per-PR CI safety

No changes to ci.yml or any required-check path. New/changed fast tests pass locally (`68 passed, 1 skipped`); changed files are `ruff check`/`format` clean (board03 file has pre-existing drift outside the touched regions — see #3464); both YAML files parse.

## Test plan

- [x] Fast suites: `test_board06_determinism.py` (3 new interface tests), `test_ci_routed_drc_workflow.py`, `test_check_routed_drc_advisory.py` — 68 passed
- [x] Slow local verification with native backend: parity 5/5, board03 baseline 4/4 (8m48s), chorus placement 3/3 (42s), mfr-tier 3 xfail (#3463)
- [ ] After merge: trigger `workflow_dispatch` on Slow Tests and confirm it goes green (or fails loudly on anything this inventory missed)

### DIFF
--- a/.github/routed-drc-tolerance.yml
+++ b/.github/routed-drc-tolerance.yml
@@ -1095,7 +1095,15 @@ tolerances:
   # 0.5 mm tolerance, unchanged -- this is upstream of
   # apply_match_group_tuning's reach and is tracked as a
   # follow-up under the M-G campaign).
-  boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb: 25
+  #
+  # Tightened 25 -> 16 (issue #3458 failure inventory).  PR #3462
+  # refreshed the committed routed PCB and the gate now measures
+  # 16 blocking errors (21 total - 5 advisory connectivity) on it;
+  # the old floor went stale silently because the nightly slow-tests
+  # job did not gate on the pytest exit code.  Pinned exactly (no
+  # slack) per this file's update policy; the parity pins in
+  # tests/test_kct_check_parity.py were re-baselined in the same PR.
+  boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb: 16
 
   # Board 04 (STM32F103C8T6 LQFP-48 devboard).  Tightened 60 -> 12 by
   # Issue #2908 (this PR).  The narrowing introduced in #2871 / #2874
--- a/.github/routed-drc-tolerance.yml
+++ b/.github/routed-drc-tolerance.yml
@@ -1100,10 +1100,21 @@ tolerances:
   # refreshed the committed routed PCB and the gate now measures
   # 16 blocking errors (21 total - 5 advisory connectivity) on it;
   # the old floor went stale silently because the nightly slow-tests
-  # job did not gate on the pytest exit code.  Pinned exactly (no
-  # slack) per this file's update policy; the parity pins in
-  # tests/test_kct_check_parity.py were re-baselined in the same PR.
-  boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb: 16
+  # job did not gate on the pytest exit code.  The parity pins in
+  # tests/test_kct_check_parity.py were re-baselined in the same PR;
+  # those pins measure the COMMITTED artifact (deterministic) and
+  # stay at 16 blocking.
+  #
+  # Relaxed 16 -> 21 (same PR, after the Match-Group gate failed
+  # twice -- real, not flake): 21 = CI-empirical value under load
+  # for the FRESH seed-42 re-route performed by "Match-Group
+  # Routing Regression"; local idle machines measure 16 on the same
+  # re-route.  The variance is the #3466 wall-clock-budget cliff --
+  # the router's time-budgeted passes do less work per pass on a
+  # loaded CI runner, so the re-route DRC profile shifts with
+  # machine load.  16 was a local-idle measurement, never a stable
+  # CI floor.  Tighten after #3466 merges if CI stabilizes.
+  boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb: 21
 
   # Board 04 (STM32F103C8T6 LQFP-48 devboard).  Tightened 60 -> 12 by
   # Issue #2908 (this PR).  The narrowing introduced in #2871 / #2874
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -13,7 +13,7 @@ jobs:
   slow-tests:
     name: Slow / Benchmark Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 90
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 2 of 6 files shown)


rjwalters:
## Match-Group gate failure addressed: tolerance 16 -> 21

The two Match-Group gate failures ("DRC error count: 21 (allowed: 16)") were real, not flakes — but the problem was the floor, not the routing.

**What happened:** this PR tightened board 07's allowlist 25 -> 16 based on a local idle-machine measurement. That number is correct for the **committed** routed artifact (deterministic — the `test_kct_check_parity.py` pins at 16 blocking / 5 advisory connectivity verify it and still pass 11/11). But the Match-Group Routing Regression job performs a **fresh seed-42 re-route from source**, and the re-route's DRC profile varies with machine load: local idle = 16, loaded CI runner = 21.

**Why it varies:** this is the wall-clock-budget cliff diagnosed in #3466 — the router's time-budgeted passes do less work per pass under load, so re-route outcomes shift with runner contention. The 16 was never a stable CI floor.

**Fix (commit 6d8275b):**
- `.github/routed-drc-tolerance.yml` board 07 entry: 16 -> 21, with a comment capturing the CI-empirical rationale and the #3466 cross-reference. Tighten after #3466 merges if CI stabilizes.
- `tests/test_kct_check_parity.py`: comment-only update clarifying why the parity pins (committed artifact, deterministic, 16) intentionally differ from the allowlist floor (fresh re-route, load-variant, 21). No assertion changes — the parity re-baseline in this PR was for the committed artifact and needed no alignment.

The smoke job's earlier failure was a Docker flake and already passed on rerun.

rjwalters:
## Judge verdict: APPROVE

(Completing the review whose agent exited awaiting CI — local verification was complete; the doctor cycle then fixed the one real failure.)

**Verified before CI (prior judge pass):**
- #3458 core fix: `continue-on-error` removed from slow-tests pytest step; env hardening (native build, kicad-symbols, realistic timeouts) makes the gate viable
- #3460 double-masking diagnosed and closed: env-gated module that never ran + swallowed exit; new always-on `TestDRCCheckerCLIInterface` exercises the real interface
- Failure-inventory dispositions all sound; the mfr-tier xfail is non-strict but #3463 tracks it
- All four ci.yml masks confirmed tracked (#3436 Test+mypy, #3464 lint)

**Doctor cycle (Match-Group 2x failure):** tolerance 16→21 — 16 was a local-idle measurement; CI-under-load re-routes at exactly 21 (this run: 21/21, zero headroom). Variance is the #3466 wall-clock cliff, documented in the tolerance file with a tighten-later note. kct_check parity pins (16, committed artifact, deterministic) correctly left unchanged with clarifying comments.

**CI: 8/8 green** on `6d8275b5` including Match-Group (12m42s) and a clean first-try smoke (Docker flake didn't recur).

Labels: removing `loom:review-requested`/`loom:reviewing`, adding `loom:pr`.

Issue #3320, opened 2026-06-07
router: #3012 regression — DQS_N/DQS_P negative clearance on board 07 when CoupledPathfinder coupled-routes (PR #3022 fix incomplete)


rjwalters:
## Summary

PR #3022 closed Issue #3012 by adding a `min_spacing_cells` floor to `CoupledPathfinder` so within-pair clearance would be respected. The fix is **empirically incomplete on board 07's DQS_N/DQS_P pair**: when `--differential-pairs` is enabled on `boards/07-matchgroup-test/generate_design.py`, the CoupledPathfinder coupled-routes DQS but lays both centerlines down with negative clearance — i.e., the two centerlines literally overlap.

This is the original #3012 failure mode reappearing on the pair the PR #3022 fix was supposed to protect.

## Evidence

PR #3318 attempted to re-enable `--differential-pairs` on board 07. The judge review (see PR #3318 comments) dumped the actual `kct check` output and observed:

**Pre-PR (origin/main baseline, 2 errors):**
- TMDS_D0_N/TMDS_D0_P: 2 errors (-0.060mm, 0.071mm) — negotiated-router legacy

**Post-PR (with --differential-pairs enabled, 10 errors):**
- TMDS_D0_N/TMDS_D0_P: 2 errors (unchanged)
- **DQS_N/DQS_P: 8 NEW errors** with the following intra-pair clearance magnitudes:
  - -0.150 mm (centerline overlap)
  - -0.093 mm (centerline overlap)
  - -0.060 mm (centerline overlap)
  - -0.036 mm (centerline overlap)
  - -0.002 mm (centerline overlap)
  -  0.004 mm
  -  0.022 mm
  -  0.037 mm

Five of the eight are negative — the centerlines overlap by up to 0.150 mm. DQS is the pair that the CoupledPathfinder **successfully** coupled-routed (the PR's narrative explicitly cites "Coupled pair pre-routed: 1 (DQS_P/N)"), so this overlap is NOT a budget-exited single-ended fallback artifact.

## Hypothesis (root cause candidates)

PR #3022 added `min_spacing_cells` as a floor in `CoupledPathfinder` (see `src/kicad_tools/router/diffpair_routing.py:700-704,769-774,828-830`). The DQS overlap evidence implies one of:

1. The `min_spacing_cells` floor isn't being applied to the DQS pair (per-net-class plumbing gap).
[...]

Issue #3270, opened 2026-06-07
router: USB3_TX1+ U2.B2 BGA-49 escape unrouted on board 06 (#3262 baseline; 1/21 signal nets)


rjwalters:
## Problem statement

On the post-#3262 baseline of `boards/06-diffpair-test`, the `--seed 42` reroute via `generate_design.py --step route` completes **20/21 signal nets** (95.2%). One pad remains unconnected:

- `USB3_TX1+` at `U2.B2` (USB-C SS BGA-49 simulator, **inner-ring** pad, world coords (52.46, 17.46))

All four USB3 diff pairs exit the `CoupledPathfinder` iteration budget (`per_pair_max_iterations=4000`) and defer to the per-net A* main strategy (`Autorouter.route_all_negotiated`). The main strategy completes 20/21 signal nets, leaving exactly one inner-ring BGA-49 escape behind. The specific unrouted net varies by seed (the four USB3 pairs are symmetric escape problems; the seeded tie-break determines which pair loses), but the failure mode does not.

Note: this is the residual of the closed #2677 (which was the **2/21 unrouted pair** version of the same defect). #2677 closed when corridor reservation (`escape.py:1497 _reserve_pair_continuation_corridor`) and budget-exit deferral (`diffpair_routing.py:3506-3518`) reduced the loss from 2 pads to 1. The remaining single-pad loss reproduces deterministically on the baseline and is the binding gap to 21/21 (100%).

## Acceptance criteria

- [ ] Board 06 routes all 21 signal nets including both ends of all four USB3 diff pairs from the BGA-49 simulator footprint U2.
- [ ] `kct net-status output/diffpair_test_routed.kicad_pcb` reports `Unrouted: 0` and 0 inner-ring pads "needs routing".
- [ ] Total `--seed 42` reproduction reaches 21/21 (100%) on the post-#3262 baseline recipe (`generate_design.py --step route --seed 42`).
- [ ] The fix is seed-stable: at least 3 distinct seeds (42 plus two others) reach 21/21 — this guards against the symmetric tie-break "the loser just rotates" failure mode.
- [ ] DRC remains within the existing allowlist (current: 6/39).
- [ ] Drift-prevention regression: synthetic BGA-49 fixture (the one from #2677 AC, plus one isolating just the U2.B2-class inner-ring single-pad case) holds in the test suite so future budget/heuristic changes cannot regress.
[...]

PR #3473, opened 2026-06-10
fix(router): CoupledPathfinder budget + aggregate-phase cap (closes #3439)

### LINKED ISSUE #3439: router: CoupledPathfinder exceeds 60s/pair budget on every board-07 pair (~14k iterations/min, pure Python) — blocks diff-pair quality DRC = blocks 0-DRC
Parent: #3414 (board 07 manufacturable bar; decomposed by the #3414 recovery PR)
Related: #3438 (reach gap), #3012/#3022 (min_spacing_cells floor), #3295 (recipe refresh tracker — documented the >40min hang)

## Observation

Board 07's routed DRC carries 10 structural quality errors that CANNOT be fixed without coupled routing:

- 5x `diffpair_routing_continuity` (DQS 47.2%, MIPI_DAT1 2.6%, TMDS_D0/D1/D2 0.0% — threshold 85%)
- 5x `diffpair_length_skew` (3.0-7.4mm vs 0.05-0.075mm tolerances)

These pairs are routed single-ended because `--differential-pairs` is disabled in the recipe — and re-enabling it fails structurally:

```
CoupledPathfinder.route_coupled wall-clock budget exceeded after 60.00s (14016 iterations); abandoning search (p_net='TMDS_D0_P' n_net='TMDS_D0_N')
diffpair coupled-routing budget exceeded: pair='TMDS_D0' ... budget=60.0s
```

EVERY pair on the board blows the 60s budget at ~14k A* iterations (verified 2026-06-09 on main @ post-#3431; full run collapsed to 7/31 because the failed coupled pre-pass burned the 600s outer budget before the single-ended fallback ran).

## Diagnosis

~14,000 iterations in 60 seconds is pure-Python A* speed. The single-ended path does 10-100x better via `router_cpp` (#3146 line). `CoupledPathfinder` has no C++ backend, so coupled search on a 4-layer 110x95mm grid is intractable within any reasonable per-pair budget.

## Why it matters

Board 07 (and any board with declared diff pairs + engaged `diffpair_routing_continuity`) cannot reach 0 DRC until coupled routing completes within budget. This is the largest single block of board 07's residual DRC errors (10 of ~23).

## Suggested scope

1. Port the coupled A* inner loop to the C++ backend (mirror the single-net `router_cpp` integration), OR
2. Constrain the coupled search space (corridor from the single-ended P-side solution, then re-route N alongside), which converts the problem to two corridor-bounded searches.
3. Re-enable `--differential-pairs` in board 07's recipe once a pair completes in <10s; verify intra-pair clearance (post-#3022) and the DRC continuity/skew rules go green.

## Acceptance

- All 7 board-07 pairs coupled-route within the per-pair budget on a CI-class runner.
- `diffpair_routing_continuity` + `diffpair_length_skew` errors eliminated from board 07's routed DRC.

rjwalters:
Closes #3439

## Summary

Two composable fixes for the CoupledPathfinder budget blowout (curator options 2 + 3):

1. **Corridor-bounded coupled search** (`diffpair_routing.py`): route the P side single-ended via the C++-accelerated per-net pathfinder (never committed to the grid), dilate the path into a layer-agnostic corridor mask (`build_corridor_mask`), and run the joint-state coupled A* restricted to the corridor with half the per-pair budget. Falls back to the legacy unconstrained search with the remaining budget, so boards where the open search already converged (03/06) are unaffected.
2. **Aggregate coupled-phase cap**: new `DifferentialPairConfig.aggregate_timeout`; `Autorouter.route_all_with_diffpairs` auto-derives `max(per_pair_budget, 0.25 * timeout)` from `--timeout` (structured log `DIFFPAIR_AGGREGATE_TIMEOUT_AUTODERIVED`). Once exhausted, remaining pairs are deferred to the main strategy WITHOUT attempting coupled routing (`DIFFPAIR_AGGREGATE_BUDGET_EXCEEDED`), and the per-pair budget is clamped to the remaining aggregate. A failed coupled pre-pass can no longer burn `num_pairs * 60s` of the outer budget (the 7/31-reach incident: 7 x 60s = 420s of the 600s `--timeout`).

Per-pair timing is logged via `diffpair coupled timing: pair=... phase={corridor|open} elapsed=...s success=...`.

## Board 07 empirical results (`--differential-pairs --timeout 600`, load-contended host)

Per-pair timing before -> after:

| Pair | Before (main) | After |
|---|---|---|
| DQS | 60s budget blowout (~14k iters) | **coupled search completes in ~4s** (corridor); rejected by the pre-existing #3320 swap-overlap gate, falls through cleanly |
| MIPI_CLK / MIPI_DAT0 | 60s blowout each | 60s budget-exit each (their P sides cannot route single-ended at all — the #3438 reach gap — so no corridor exists) |
| MIPI_DAT1 | 60s blowout | 25s (per-pair budget clamped to remaining aggregate) |
| TMDS_D0/D1/D2 | 60s blowout each (180s) | **0s — deferred by the aggregate cap at 150s** |

- Coupled phase total: **150s (capped) vs 420s before**; the negotiated phase retains ~450s of the 600s budget.
- Reach: 26/31 under load (vs 7/31 in the incident; the documented solo baseline is 28/31, and the recipe notes concurrent runs degrade even the baseline to 25-26/31 — another builder's board-04 regression was running on this host).
- DRC delta (routed + pours, `--net-class-map`): baseline committed artifact = 21 errors (5 continuity + 5 skew + 1 intra = **11 diff-pair-attributable**); this run = 15 errors (4 continuity + 3 skew + 0 intra = **7 diff-pair-attributable**). The residual quality errors persist because no pair survives to a committed coupled route yet: MIPI is blocked by #3438 (no single-ended P path) and DQS by #3320 (swap-via centerline overlap). Board 07's recipe keeps `--differential-pairs` off; the empirical addendum is recorded in `generate_design.py`.

## Tests

- `tests/test_diffpair_corridor.py` (10 tests): corridor mask rasterization/dilation/bounds, corridor kwarg pruning + legacy default, corridor-phase integration (timing log asserts `phase=corridor`), guide route never committed.
- `tests/test_diffpair_aggregate_cap.py` (11 tests): config field + signatures, `--timeout` derivation (cap, floor, explicit-config precedence, back-compat), structured log, and the **load-bearing reach invariant**: with an exhausted aggregate budget the coupled pathfinder is never invoked and all nets (incl. diff-pair members) route via the main strategy.
- Regression: 29 tests in the #3089/#3321 timeout suites, 159 board-06/escape/diffpair tests, 136 detection/length/impedance tests, full `tests/test_board_03_regression.py` (10 passed incl. `test_usb_diff_pair_routes_via_coupled_pathfinder`, 26 min). One pre-existing order-dependent flake (`test_autorouter_invokes_tuner_for_each_pair`) fails identically on main.
- Pre-existing ruff findings in `core.py`/`diffpair_routing.py` are unchanged (24 on main, 24 here); new files are clean.

## Test plan
- [ ] CI green
- [ ] Judge review

### DIFF
--- a/boards/07-matchgroup-test/generate_design.py
+++ b/boards/07-matchgroup-test/generate_design.py
@@ -663,6 +663,26 @@ def route_pcb(input_path: Path, output_path: Path) -> bool:
     # baseline; the residual gap is tracked in #3438 (reach -- now
     # with minimal 5/10/11-net repros), #3439 (coupled routing perf),
     # #3440 (match-group tuner), #3441 (auto-grid).
+    #
+    # Issue #3439 addendum (2026-06-10): the CoupledPathfinder now has
+    # (a) a corridor-bounded search mode (P side routed single-ended
+    # via the C++ backend, coupled A* restricted to the dilated
+    # corridor) and (b) an aggregate coupled-phase cap auto-derived as
+    # ``max(per_pair, 0.25 * timeout)``.  Empirical re-test with
+    # ``--differential-pairs`` on this recipe (load-contended host):
+    #   - DQS coupled search COMPLETES in ~4s (vs blowing the 60s
+    #     budget at 14k iterations before) but is rejected for the
+    #     #3320 polarity-swap centerline overlap.
+    #   - MIPI pairs still budget-exit: their P sides cannot route
+    #     single-ended at all (the #3438 reach gap), so no corridor
+    #     exists; the open search remains intractable.
+    #   - The aggregate cap fires at 150s and defers the 3 TMDS pairs;
+    #     total coupled phase = 150s vs 420s pre-#3439, and reach no
+    #     longer collapses (26-28/31 vs the 7/31 incident).
+    # ``--differential-pairs`` is now SAFE to enable (bounded, reach-
+    # preserving) but is not yet a quality win: no pair survives to a
+    # committed coupled route until #3438 (MIPI reach) and #3320
+    # (DQS swap overlap) land.  The recipe therefore still omits it.
     cmd = [
         sys.executable,
         "-m",
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -11000,12 +11000,49 @@ class Autorouter:
                 derived_per_pair_timeout,
             )
 
+        # Issue #3439: derive an AGGREGATE coupled-phase budget from
+        # --timeout when the caller has not explicitly configured one.
+        # The per-pair budget alone is insufficient containment: a board
+        # with many pathological pairs burns ``num_pairs * per_pair``
+        # of the outer budget before the single-ended main strategy
+        # runs (board 07: 7 pairs x 60s = 420s of the 600s --timeout,
+        # collapsing final reach to 7/31).  Capping the whole coupled
+        # phase at 25% of --timeout (floored at one full per-pair
+        # budget so a single pair always gets a complete attempt)
+        # guarantees the main strategy retains at least ~75% of its
+        # budget no matter how many pairs fail.
+        derived_aggregate_timeout: float | None = None
+        if (
+            timeout is not None
+            and timeout > 0
+            and diffpair_config is not None
+            and diffpair_config.enabled
+            and getattr(diffpair_config, "aggregate_timeout", None) is None
+        ):
+            per_pair_floor = (
+                diffpair_config.per_pair_timeout
+                if diffpair_config.per_pair_timeout is not None
+                else (derived_per_pair_timeout or 0.0)
+            )
+            derived_aggregate_timeout = max(
+                float(per_pair_floor), float(timeout) * 0.25
+            )
+            logger.info(
+                "DIFFPAIR_AGGREGATE_TIMEOUT_AUTODERIVED: --timeout=%.1fs; "
+                "capping the aggregate coupled diff-pair phase at %.1fs so "
+                "a failed coupled pre-pass cannot starve the single-ended "
+                "main strategy (issue #3439)",
+                float(timeout),
+                derived_aggregate_timeout,
+            )
+
         result = self._diffpair.route_all_with_diffpairs(
             diffpair_config,
             net_order,
             non_diffpair_strategy=non_diffpair_strategy,
             coupled_only=coupled_only,
             per_pair_timeout=derived_per_pair_timeout,
+            aggregate_timeout=derived_aggregate_timeout,
         )
 
         # Issue #3040 Phase B: rip-up and retry any pairs whose coupled
--- a/src/kicad_tools/router/diffpair.py
+++ b/src/kicad_tools/router/diffpair.py
@@ -591,6 +591,23 @@ class DifferentialPairConfig:
             against pathological cases) AND
             ``per_pair_max_iterations`` (as the deterministic
             classifier).  Whichever fires first triggers the exit.
+        aggregate_timeout: Issue #3439: Optional wall-clock budget
+            (seconds) for the ENTIRE coupled diff-pair phase (all
+            pairs combined).  ``per_pair_timeout`` bounds any single
+            ``CoupledPathfinder.route_coupled`` call, but a board with
+            many pathological pairs can still burn
+            ``num_pairs * per_pair_timeout`` of the outer routing
+            budget before the single-ended main strategy ever runs --
+            the board-07 7/31-reach collapse (every one of 7 pairs
+            blew its 60 s budget, consuming 420 s of the 600 s
+            ``--timeout`` before fallback).  When the aggregate budget
+            is exhausted, all remaining pairs are deferred to the main
+            strategy WITHOUT attempting coupled routing, so a failed
+            coupled pre-pass can never reduce the final reach below
+            the ``--differential-pairs``-off baseline.  ``None``
+            (default) preserves the legacy per-pair-only behaviour;
+            ``Autorouter.route_all_with_diffpairs`` auto-derives a
+            value from ``--timeout`` when this is unset.
     """
 
     enabled: bool = False
@@ -600,6 +617,7 @@ class DifferentialPairConfig:
     add_serpentines: bool = True
     per_pair_timeout: float | None = None
     per_pair_max_iterations: int | None = None
+    aggregate_timeout: float | None = None
 
     def get_rules(self, pair_type: DifferentialPairType) -> DifferentialPairRules:
         """Get rules with any config overrides applied."""
--- a/src/kicad_tools/router/diffpair_routing.py
+++ b/src/kicad_tools/router/diffpair_routing.py
@@ -534,6 +534,15 @@ class CoupledPathfinder:
         # invocation.
         self.last_timeout_exceeded: bool = False
 
+        # Issue #3473 (review of #3439): number of A* iterations the
+        # most recent ``route_coupled`` call consumed.  The two-phase
+        # corridor-then-open caller charges the corridor attempt's
+        # iterations against the shared per-pair iteration budget,
+        # mirroring the wall-clock split -- otherwise a failing pair
+        # receives the FULL ``max_iterations_budget`` twice (observed
+        # 4000+4000 on board 06, doubling the diff-pair phase).
+        self.last_iterations: int = 0
+
         # Pre-calculate trace clearance radius
         self._trace_half_width_cells = max(
             1,
@@ -1160,6 +1169,8 @@ class CoupledPathfinder:
         # this immediately after ``route_coupled`` returns ``None`` to
         # decide whether to attempt an independent-routing fallback.
         self.last_timeout_exceeded = False
+        # Issue #3473: reset the iteration counter for this call.
+        self.last_iterations = 0
 
         # Convert to grid coordinates
         p_start_gx, p_start_gy = self.grid.world_to_grid(p_start.x, p_start.y)
@@ -1269,6 +1280,12 @@ class CoupledPathfinder:
 
         while open_set and iterations < max_iterations:
             iterations += 1
+            # Issue #3473: keep the public counter current on every
+            # iteration so EVERY exit path (budget, timeout, goal,
+            # exhausted open set, exceptions) reports the true cost.
+            # A single attribute store is noise next to the heap and
+            # parent-chain work in this loop body.
+            self.last_iterations = iterations
 
             # Issue #3144: iteration budget classifier check.  Sits
             # adjacent to the wall-clock check so a single ``if`` body
@@ -2284,7 +2301,12 @@ class DiffPairRouter:
             results.append(route)
         return results
 
-    def _single_ended_guide_route(self, start_pad: Pad, end_pad: Pad) -> Route | None:
+    def _single_ended_guide_route(
+        self,
+        start_pad: Pad,
+        end_pad: Pad,
+        per_net_timeout: float | None = None,
+    ) -> Route | None:
         """Route one side of a pair single-ended to seed a corridor mask.
 
         Issue #3439: the corridor-bounded coupled search needs a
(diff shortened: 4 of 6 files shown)


rjwalters:
## Judge evaluation: Changes requested (CI failure — board-06 wall-clock regression)

The design is sound and well-tested, but the corridor pre-attempt measurably regresses board 06's CI job past its 15-minute ceiling. Details below.

### Verdict on the judging questions

1. **Corridor soundness — PASS.** `corridor_radius = ceil(max(spacing_cells, start_pad_pitch, end_pad_pitch)) + max(6, spacing_cells)`, where `spacing_cells` already encodes the center-to-center floor `trace_width + intra_pair_clearance` (the #3012 derivation in `route_differential_pair_coupled`). The mask constrains centerline head positions only; occupancy/clearance checks remain independent, so the N trace fits at pair separation with >=6 cells of maneuvering slack. Endpoint cells are exempt AND dilated into the mask via `extra_cells` — belt and suspenders. The `max(..., start/end pad pitch)` term correctly handles endpoint fan-out wider than the target spacing. Tube half-width ~2x spacing + 6 cells keeps it near-1D (empirical DQS ~4s confirms the speedup).

2. **Probe rollback — PASS for hard occupancy, one soft-state leak.** `router.route()` (both `CppPathfinder._route_impl` and the Python `Router._route_impl`) is search-only; grid commit happens via the separate `add_routed_segments`/route-list path the probe never takes. `test_guide_route_is_not_committed_to_route_list` pins this. **Caveat:** on the C++ validation-failure path, `_boost_avoidance_at` mutates persistent grid avoidance costs, and the cleanup (`clear_avoidance_costs`) lives in `Autorouter._route_net` (core.py:2155) — which the probe bypasses. A probe whose route fails post-validation leaks cost-shaping (not blocking occupancy) into the next search. Mirrors the pre-existing stub-routing call at diffpair_routing.py:2267, but please add a `clear_avoidance_costs()` after the probe while you're in here.

3. **Aggregate cap semantics — PASS.** Derivation `max(per_pair_floor, 0.25*timeout)` verified incl. the floor (`--timeout 100` -> 30s) and explicit-config precedence. Deferral reuses the #3089 `budget_exit_diff_nets` path: nets are removed from `diff_net_ids`, so the main strategy routes them single-ended (with the #3270 priority promotion); the invariant test proves end-to-end that deferred nets still route. Not dropped.

4. **Reach-invariant test — PASS.** `test_exhausted_aggregate_budget_skips_coupled_and_preserves_reach` spies on `route_differential_pair_coupled` (asserts zero invocations with `aggregate_timeout=0.001`) AND asserts all nets — diff-pair members included — route via the fallback. Plus a control test without the cap. Properly pins the 7/31 incident class.

5. **Load-sensitivity framing — honest.** 26/31 under load vs 28/31 documented solo baseline, with the concurrent-builder caveat disclosed; the recipe addendum correctly states the flag is now safe-but-not-yet-a-win pending #3438/#3320.

6. **Local verification:** all 21 new tests pass; 29 timeout-suite regressions pass; `test_diffpair_routing_integration.py` 5/5; ruff on touched router files = 24 findings on both main and branch (pre-existing, unchanged). `closingIssuesReferences` = [#3439] only.

### BLOCKING: board-06 CI job cancelled at the 15-minute ceiling

`Diff-Pair Routing Regression (boards/06-diffpair-test)` was **cancelled at 15m14s** (job `timeout-minutes: 15`), seconds after the coverage check printed OK. Substantive output is **identical to main** (18/21 nets, 14 DRC errors within the 21 allowlist, all 3 rules exercised) — this is purely a wall-clock regression:

- main (3 recent runs): job completes in **12m01s–12m20s**; diff-pair phase ~5.5 min, one budget-exit per failing pair.
- this PR: diff-pair phase **~8.3 min** (14:31:38 -> 14:39:55); every failing pair now exits **twice**:
  - `MIPI_CLK`/`MIPI_D0`/`USB3_*`: "iteration budget exceeded after 4000 iterations" in the **corridor** attempt, then **another 4000** in the open fallback — `max_iterations_budget=per_pair_max_iterations` is passed un-halved to the corridor attempt, doubling the iteration budget for failing pairs.
  - `USB3_RX2`/`USB3_TX1`/`USB3_TX2`: the open fallback got only the **1.0s floor** ("wall-clock budget exceeded after 1.00s"), meaning probe + corridor consumed ~the full 60s. The single-ended probe in `_single_ended_guide_route` is called with **no `per_net_timeout`**, so on hard BGA-escape P sides (C++ fails -> unbounded Python fallback) it eats the per-pair budget outside any deadline.

So the PR claim "boards where the open search already converged (03/06) are unaffected" does not hold for board 06: its pairs budget-exit rather than converge, and the corridor pre-attempt roughly doubles their cost (+~3 min), tipping the job over the ceiling.

### Required changes

1. **Split the iteration budget** the same way the wall-clock budget is split: give the corridor attempt at most half of `per_pair_max_iterations` (or count both attempts against one shared budget). Observed: 4000+4000 per failing pair on board 06.
2. **Bound the probe**: pass `per_net_timeout=` (e.g. the corridor half-budget, or a small fraction of it — the probe should be cheap or skipped) to `self.autorouter.router.route()` in `_single_ended_guide_route`, so an unroutable/hard P side cannot consume the per-pair budget before either coupled attempt runs.
3. Re-run and confirm the board-06 job completes inside the 15-minute ceiling (routing outcome is already identical, so budget arithmetic is the only fix needed).

### Non-blocking notes

- Add `clear_avoidance_costs()` after the probe (soft-state leak above).
- The "Diff pairs deferred to main strategy due to per-pair budget: N" print counts aggregate-deferred pairs under the per-pair label (cosmetic; there is a separate aggregate print).
- Merge will auto-close #3439 whose literal ACs (all 7 pairs route within budget; continuity/skew eliminated) are not met — residuals are explicitly delegated to #3438/#3320 in the recipe addendum, which reads as intentional re-scoping, but flagging per the partial-AC merge convention.

Excellent diagnosis work overall — the two-phase budget composition and the pinned reach invariant are exactly right; this just needs the same budget discipline applied to iterations and the probe.

rjwalters:
## Doctor: judge feedback addressed (846281e9)

All three required changes plus both non-blocking notes are in.

### 1. Iteration budget split (required)

`CoupledPathfinder` now exposes `last_iterations` (reset per call, updated every loop iteration so every exit path reports true cost). The two-phase caller charges both attempts against ONE shared per-pair budget:

- **Corridor attempt** gets at most `per_pair_max_iterations // 2`.
- **Open fallback** gets `per_pair_max_iterations - corridor_iterations_used` (so a cheap corridor failure hands nearly the full budget to the fallback, mirroring the wall-clock remaining-budget arithmetic).

CI forensics confirm: board 06 failing pairs now exit at **2000 + 2000** iterations (was 4000 + 4000).

### 2. Probe deadline (required)

`_single_ended_guide_route` now passes `per_net_timeout` = 1/8 of the per-pair budget (7.5s of 60s). If the P side cannot route single-ended that quickly, the corridor is skipped and the legacy open search gets the budget. The corridor wall-clock budget is additionally computed as `max(0.5, 0.5*per_pair - elapsed_since_pair_start)`, so probe + corridor together are capped at ~half the per-pair budget (previously up to 62.5%, and the probe itself was unbounded). The "fallback got only the 1.0s floor" failure mode is gone — board 06 fallbacks now get 52s+ or the full iteration remainder.

### 3. Board 06 CI job inside the ceiling (required)

This push's run: **Diff-Pair Routing Regression completed in 14m05s** (was: cancelled at 15m14s). Diff-pair phase **~6m22s** on the CI runner (pre-fix: ~8.3 min; main: ~5.5 min — the residual ~50s over main is the bounded 7.5s probe on the 6-7 pairs whose P side cannot route single-ended; it buys the corridor attempt where one exists). Substantive output identical to main: **18/21 nets, 14 DRC errors** (allowlist 21), all 3 rules exercised.

### Non-blocking notes

- `clear_avoidance_costs()` now runs in a `finally` after every probe, so a validation-failed probe cannot leak C++ cost-shaping into subsequent searches (pinned by `test_failed_probe_clears_avoidance_costs`).
- The "deferred due to per-pair budget: N" print no longer counts aggregate-deferred pairs (verified on board 07: aggregate print reports 3, per-pair print reports 3, no double count).

### Verification

- 4 new tests in `tests/test_diffpair_corridor.py`: half-budget corridor cap, remainder fallback budget (both exact-value spies), probe deadline = 7.5s, probe avoidance-cost cleanup. All 25 corridor + aggregate-cap tests pass; 29 timeout-suite regressions pass; `test_diffpair_routing_integration.py` 5/5.
- Board 07 PR-table re-run (`--differential-pairs --timeout 600`, seed 42): **DQS coupled completes in ~4s (corridor)**; MIPI_CLK/DAT0 60s budget-exits (no single-ended P path, probe deadline fires, fallback gets ~55s); MIPI_DAT1 25.6s clamped (12.6s corridor + 12.7s fallback under the split); TMDS x3 deferred at the 149.4s aggregate cap; coupled phase total 150s — all table rows hold.
- Full CI green on 846281e9 (Test, board 06, board 07, DRC, round-trip, type check, build).

rjwalters:
## Re-review: APPROVED

Fast re-review of commit `846281e9`, which addresses the single blocking finding from the previous review (corridor attempt received the full un-halved iteration budget and the guide probe ran unbounded, blowing the board 06 job ceiling). All four substantive design questions had already passed.

### 1. Budget-split arithmetic — verified correct (shared pool, no double-spend)

- `CoupledPathfinder.last_iterations` is reset at the top of `route_coupled` and stored on every loop iteration, so every exit path (budget, timeout, goal, exhausted open set, exception) reports the true cost. `route_coupled` has no C++ delegate, so the counter cannot be bypassed.
- Corridor attempt: `max_iterations_budget = max(1, per_pair_max_iterations // 2)`.
- Open fallback: `max(1, per_pair_max_iterations - corridor_iterations_used)` — the remainder of one shared pool. Worst case 2000+2000 from a 4000 budget (was 4000+4000); early corridor exit returns the unused slice to the fallback.
- Wall-clock mirrors this: corridor gets `max(0.5, 0.5*T - elapsed_since_spec_t0)` (probe time deducted from the corridor half), fallback gets `max(1.0, T - elapsed)`. Probe + corridor jointly bounded near half the pair budget; total bounded by T plus small floors.
- Corridor-skipped path (no guide route): `corridor_iterations_used` stays 0 and the fallback correctly receives the full budget.

### 2. Probe deadline propagates to BOTH backends — verified

`_single_ended_guide_route` passes `per_net_timeout` into `autorouter.router.route(...)`:
- C++ backend (`cpp_backend.py`): converted to `timeout_seconds` for the native A* call, and forwarded as `per_net_timeout` to all three `_try_python_fallback` call sites, which pass it into the Python pathfinder.
- Python backend (`pathfinder.py`): `route()`/`_route_impl` honor `per_net_timeout` natively.
- `clear_avoidance_costs()` in `finally` is correctly `hasattr`-guarded (only the C++ wrapper defines it).

### 3. Tests

The 4 new tests in `tests/test_diffpair_corridor.py` pin exactly the right behaviors: half-budget cap (2000/4000), remainder fallback after early corridor exit (3900 after 100 used), probe deadline value (7.5s of 60s), and avoidance-cost cleanup on probe failure.

### 4. CI

All 8 checks green on head `846281e9`; merge state CLEAN. Diff-pair phase ~6m22s (was 8.3 min; main ~5.5), job 14m05s under the 15-min ceiling, outcome identical to main (18/21 nets, 14 DRC <= 21 allowlist). Board 07 regression also green.

### Margin disposition (non-blocking)

The ~55s margin under the 15-minute board 06 ceiling is acceptable for this PR but too thin to leave unaddressed — the diff-pair phase is a growth area and runner variance alone can eat a minute. Filed **#3477** as a follow-up (preferred fix: one-line `timeout-minutes` bump to 20; alternatives documented there). This does not block merge.

Verdict: approve. Labels updated to `loom:pr`.

PR #3475, opened 2026-06-10
fix(router): match-group reference semantics + distributed meanders + reason-bucket observability (closes #3440)

### LINKED ISSUE #3440: router: --length-match-groups leaves board 07 ADDR_BUS at 15.4mm skew (tolerance 0.5mm) — tuner reports all members clean/untouched
Parent: #3414 (board 07 manufacturable bar; decomposed by the #3414 recovery PR)
Related: #2723 (Phase 3H CLI flag), #3098 (sidecar class-name keying fix), #3438 (reach), #3439-sibling (coupled perf)

## Observation

Board 07's production recipe passes `--length-match-groups` and the committed routed artifact has ALL 8 ADDR_BUS members routed, yet DRC reports:

```
[X] match_group_length_skew
    Match group 'ADDR_BUS' (8 members) length-skew 15.395 mm exceeds tolerance 0.500 mm
```

The route-side tuner log shows the orchestrator detects the groups but tunes nothing:

```
--- Length-Match Groups (Epic #2661 Phase 3H) ---
  ADDR_BUS: 8 members (0 tuned, 0 clean, 0 rolled back, 0 budget-exhausted, 0 skipped)
  DDR_DATA_BYTE_0: 11 members (0 tuned, 0 clean, 0 rolled back, 0 budget-exhausted, 0 skipped)
  ...
  Summary: 4 groups, 0 tuned, 4 clean, 0 rolled back, 0 budget-exhausted, 0 skipped
```

(All ADDR_BUS counters zero — the members do not even reach a terminal classification; note the log above is from a partial-reach run, but the committed 28/31 artifact — where A0-A7 are all routed — still carries the 15.395mm skew error, so tuning did not engage there either.)

## Questions to answer

1. Why do fully-routed ADDR_BUS members produce zero tuner outcomes (not even `already_within_tolerance` / `not_length_critical`)? Likely a filter between `detect_match_groups` output and `apply_match_group_tuning` member iteration (route_cmd.py:9075-9150).
2. Is a 15.4mm correction even within meander capacity for the short members (A0 is the straight-line shortest; longest member takes a detour)? If not, the tuner should report `exceeded_max_inserts`/`cascade_budget_exhausted`, not silence.

## Acceptance

- Board 07 route log shows non-zero terminal outcomes for every routed ADDR_BUS member.
- Either ADDR_BUS skew is brought within 0.5mm (DRC error gone) or the tuner emits an explicit capacity diagnosis.
- Matchgroup CI gate stays green (the gate asserts the rule FIRES; this issue is about making it PASS).

rjwalters:
Closes #3440

## Summary

Board 07's `--length-match-groups` left ADDR_BUS at 15.395mm skew (tolerance 0.5mm). The curator found two regimes: (A) silent reason buckets hiding an unsatisfiable shortest-member pace-car declaration, and (B) ~15mm meander corrections failing the post-insertion DRC self-check and rolling back. This PR fixes both plus the silent no-op when the skew rule lacks a net-class map — and, along the way, two latent pipeline bugs the new observability exposed.

### 1. Reference semantics (regime A)
- When the declared pace-car reference routes SHORTER than another member by more than tolerance, the lengthen-only tuner now **auto-promotes longest-in-group** as the effective reference (curator option (a)): a structured warning is logged and every member — including the declared reference — becomes tunable. The DRC rule measures max-min spread, so this is the only satisfiable semantic for a lengthen-only tuner.
- Strict pace-car behavior is preserved byte-for-byte when the declaration is satisfiable (reference is longest, or members are within tolerance above it).
- Board 07 keeps `length_match_reference="A0"` — it now exercises the promotion path.

### 2. Meander capacity (regime B — distributed meanders)
- **Amplitude back-off ladder** (`AMPLITUDE_BACKOFF_FACTORS = (1.0, 0.5, 0.25, 0.125)`): when a candidate meander fails DRC, retry the same segment with progressively smaller amplitudes — many shallow humps instead of one tall blob.
- **Exact-fit amplitude**: loop count chosen at the ladder amplitude, then the amplitude is shrunk so the loops add EXACTLY the needed length (the fixed-amplitude tuner overshot by up to `2*amplitude`; A6 landed 1.8mm past the reference).
- **Collinear-merge normalization**: tuning runs before `_finalize_routes` cleanup, so A* staircase routes arrived as hundreds of grid-step micro-segments with no trombone host. Members are now normalized via `merge_collinear` (geometry/length-preserving) before host ranking; rollback still returns the original route reference.
- **Fallback host floor**: when no segment meets the standard 2.0mm host minimum, ranking retries at the geometric one-loop minimum.
- **`routes_by_net` staleness fix**: committed meanders are published in place so later members' self-checks see them — previously two adjacent members could stack meanders on top of each other (the `A2 vs A1` clearance underflows).
- **Commit-back identity-guard fix (`core.py`)**: the staleness publish made the orchestrator's `new_route is not routes_by_net[net_id]` guard False for every tuned member, silently dropping ALL meanders from `self.routes` (they never reached the saved PCB). Now compared against the `self.routes` entry; pinned by a new pipeline regression test.
- **Intra-pair clamp**: the impedance resolver can overwrite `intra_pair_clearance` with ~8mm physics coupling gaps (documented board-06 hazard); as a within-pair DRC floor that rejected every pair meander at `1.0mm < 8.425mm`. Clamped to the intra-group floor for the tuner's self-check.

### 3. Observability
- `TUNE_RESULT_REASONS` registry + shared `format_reason_counts()` used by BOTH summary printers (`core.py` per-group line, `route_cmd.py` aggregate). Every reason is counted: the legacy five buckets always print; `reference` / `longer-than-ref` / `unrouted` / `no-segment` print when non-zero; unknown reasons print as `N other(<reason>)`. The all-zeros line for a 15.4mm-skew group is now impossible; an exhaustiveness test scans the tuner source so a future reason value cannot silently vanish.
- Rollbacks log WHY: the DRC detail names the violated pass, the offending neighbor net/via/pad, layer, and measured-vs-required clearance, plus a capacity diagnosis (requested vs achieved mm).
- Loud stderr warnings (visible under `--quiet`): `kct route --length-match-groups` without `length_match_group` declarations, and `kct check` running the skew rules without `--net-class-map`.
- Fixed an empty-reason pair-path branch (P half at/over target with short lane average) that surfaced as `other()` — now `longer_than_reference`.

## Board 07 results (production recipe, seed 42, PYTHONHASHSEED=42)

| Metric | Before | After |
|---|---|---|
| ADDR_BUS skew | 15.395mm | **0.004mm** (tol 0.5) |
| DDR_DATA_BYTE_0 skew | 1.767mm | within 0.1mm tol (all scalars 39.098mm) |
| `match_group_length_skew` errors | 2 | **0** |
| Total DRC errors | 21 | **19** (allowlist 21) |
| ADDR_BUS tuner line | all-zeros | `7 tuned, ..., 1 reference` |

- Match-Group CI gate contract holds: `match_group_length_skew` still FIRES (`rules_checked_by_rule = 2`), errors within allowlist.
- Committed routed artifact refreshed; `tests/test_kct_check_parity.py` re-baselined (blocking 16 -> 14, match-group family delta 2 -> 0).

## Test plan
- [x] 86 tuner unit tests incl. new: auto-promotion (+warning, +skew convergence), satisfiable pace-car preserved, amplitude ladder recovery + rollback detail, collinear-merge tunability + rollback identity, exact-fit no-overshoot, reason-registry exhaustiveness, `format_reason_counts` accounting, `routes_by_net` publish, DRC detail variants
- [x] Pipeline: tuned meanders reach `self.routes` (regression pin for the commit-back fix); verbose summary counts reference/unrouted buckets; rollback diagnostics name the member + neighbor
- [x] CLI: stderr warnings for inactive skew rules (`kct check`) and inactive `--length-match-groups` (`kct route`)
- [x] Board 07 full recipe re-route x2 (deterministic seed): 29/31 reach, skew results above
- [x] No regression: boards 02-06 + softstart suites (209 passed; 1 pre-existing failure `test_bom_footprint_backfill` also fails on clean main)
- [x] Pre-existing-only: ruff (`F401`/`F811` in test file pre-date this PR), mypy clean for `match_group_tuning.py`, `test_diffpair_length_tuning` ordering flake reproduces on clean main


rjwalters:
✅ **Approved!**

Thorough review of the match-group reference auto-promotion + distributed meanders + observability work. All judging questions verified.

## Verdict rationale

**1. Commit-back identity-guard bug — blast radius: zero on main, confined to this PR.**
Verified `origin/main`'s `match_group_tuning.py` only *reads* `routes_by_net` (no in-place publish), so the legacy guard `new_route is not routes_by_net[net_id]` was always True for tuned members on main — the silent meander drop **never shipped**. It was a hazard created by this PR's own staleness publish and fixed in the same PR (both halves ship atomically). The code path is NOT shared with other tuning commit-backs: `apply_diffpair_length_tuning` (core.py ~10549) has its own per-pair loop, `tune_diff_pair_skew` does not publish in place, and its commit-back runs between pairs — its identity guard remains valid. Only `--length-match-groups` boards are touched (board 07 is the only one). Pinned by `test_meander_is_committed_to_self_routes`, and I independently parsed the committed artifact: all 8 ADDR members at 44.444–44.448mm (skew 0.004mm vs 0.5 tol) with meander-dense segment counts (A2: 88 segs, A5: 119 segs) — meanders are physically present in the saved PCB.

**2. Auto-promotion semantics — observable and strict-path-preserving.**
`logger.warning` names the group, declared reference net + length, the longer member + length, and the tolerance ("auto-promoting longest-in-group"); asserted via caplog in `test_shortest_member_reference_promotes_longest`. Strict pace-car path: promotion only fires when `longest_len > ref_length + tolerance_mm`; otherwise `effective_reference_net_id` and `ref_length` are untouched — `test_longer_within_tolerance_does_not_promote` pins reference identity (`results[2][0] is routes[2]`, same segments object). Reference members now surface as `1 reference` in the summary instead of vanishing.

**3. Exact-fit amplitude + back-off ladder — DRC self-check still gates every meander.**
Every (candidate segment × amplitude rung) attempt runs `_post_insertion_clearance_detail_group`; commit happens only on `None`. Generation failure breaks the ladder (correct: smaller amplitude needs MORE forward room, not less); DRC failure continues to the next rung. When all rungs/candidates fail, `post_insertion_drc_violation` + byte-for-byte rollback to the ORIGINAL route reference is preserved — `test_intra_group_rollback` and `test_non_group_neighbor_rollback` exercise genuine-misfit geometry with the ladder active and pass. Rollbacks now carry rule/neighbor/layer/measured-vs-required detail plus a capacity diagnosis.

**4. intra_pair_clearance clamp — correctly scoped.**
The `min(intra_pair_clearance_mm, intra_group_clearance_mm)` clamp is a local per-group variable inside `apply_match_group_tuning`, fed only into `tune_match_group_v2`'s self-check. It does not mutate the NetClass, and `apply_diffpair_length_tuning` resolves its own un-clamped `effective_intra_pair_clearance()` — real diff-pair DRC validation is unaffected.

**5. Parity re-baseline 16 → 14 — committed-artifact pin, tightening direction.**
`test_kct_check_parity.py` pins `BOARD_07_PCB` (the committed artifact): match-group family delta 2 → 0, blocking 16 → 14, with engagement still pinned by the matchgroup CI gate's `rules_checked_by_rule` assertion (the gated-but-clean case is handled correctly in `test_family_delta_counts_pinned` / sidecar-added-families assertion).

**6. CI — all 8 checks green, merge state CLEAN.**
The initial "Test" job failure was an `astral-sh/setup-uv` infrastructure failure (failed at action setup, zero test steps ran); re-run succeeded. Both routing gates passed on first attempt — no board-06 timeout (the PR #3473 concern did not materialize). `closingIssuesReferences` = [#3440]; all three acceptance criteria met (per-member outcomes visible, ADDR skew 0.004mm < 0.5mm, gate green with rule firing).

## Test Execution
1. PR test files (`test_match_group_tuning.py`, `test_apply_match_group_tuning.py`, `test_cli_check_net_class_map.py`, `test_cli_route_match_groups.py`, `test_kct_check_parity.py`) — ✅ Executed: 136 passed in 113s
2. Committed-artifact skew — ✅ Independently verified: 0.004mm ADDR, BYTE0 scalars all 39.098mm
3. ruff/mypy on changed files — ✅ Executed: identical 24 pre-existing ruff findings on main and PR; CI Lint/Type green
4. Board 07 full re-route — ✅ Covered by the Match-Group CI gate (success); not re-run locally (>2 min)
5. Boards 02-06 + softstart suites — ⏭️ Covered by CI Test job (success) + Diff-Pair gate (success)

## Test Scoping
**Strategy**: PR-file-scoped pytest (5 test files, 136 tests) + CI full suite verification. Changed files: 4 source + 5 test files + 1 board artifact.

Excellent observability work — the reason registry with bidirectional exhaustiveness checking (source-scan both for unregistered and stale entries) is exactly how to make "silent bucket" regressions structurally impossible.

Issue #3480, opened 2026-06-10
router: intra-IC route violates pad clearance on SOT-23-5 (V_AC_SENSE vs U8 GND pad, 0.027mm)


rjwalters:
## Symptom

On softstart rev B (issue #3343 P-R4 artifact, `kct route --backend cpp --manufacturer jlcpcb-tier1 --clearance 0.20`), the routed PCB carries exactly one clearance violation:

```
clearance_pad_segment @ (123.9, 58.0) F.Cu
Segment to pad clearance 0.027mm < minimum 0.102mm
items: Trace-5f4d4460 (V_AC_SENSE), U8-2 (GND)
```

The trace is the **intra-IC route** the router emits for the MCP6001 unity-gain feedback tie (`Intra-IC route: U8 pins 1->4 (2.91mm)` in the route log). U8 is a SOT-23-5: pins 1 (VOUT) and 2 (V-/GND) sit on the same column at 0.95mm pitch with a 0.30mm pad-edge gap. The intra-IC route from pin 1 wraps past pin 2 with 0.027mm of clearance -- at 0.30mm trace + 0.20mm clearance nothing can legally pass between those pads, so the route must wrap around the package outline instead.

`kct fix-drc` classifies the violation as **Infeasible** (the nudge repair cannot move the trace without breaking connectivity in the pad gap).

## Ask

`create_intra_ic_routes` (router/core.py ~line 1586) should validate its emitted segments against the same-package foreign pads at the active clearance, and fall back to a perimeter wrap (or defer to the main A*) when the direct path violates. This is the last clearance error standing between softstart rev B and 0-blocking DRC at jlcpcb-tier1 (the other 2 errors are #3479's connectivity residuals).

Refs #3343, #3479.

Issue #3479, opened 2026-06-10
softstart rev B: NRST + V_BANK_POS_SENSE rip-up non-convergence caps production reach at 24/26


rjwalters:
## Context

Issue #3343 P-R1..P-R4 brought softstart rev B from 19/30 to **24/26 signal nets** at the production recipe (`kct route --backend cpp --starting-layers 4 --manufacturer jlcpcb-tier1 --micro-via-in-pad-fallback --clearance 0.20 --trace-width 0.30 --per-net-timeout 60 --timeout 900`, `PYTHONHASHSEED=0`). The last 2 nets fail **deterministically across seeds 7/42/123 and per-net budgets 30 s / 60 s** -- this is the #3470 BLOCKED_BY_COMPONENT rip-up non-convergence signature, not a placement or budget problem.

### NRST (1/8 pads connected)

- 8-pad net: U1.6 (PF2/NRST, west face) + SW1.1/.2 + R10.2 + C5.1 + J5.6 (x=200) + Q7.2 (169.95, 150.1) + Q8.2 (169.95, 181.1) -- the 2N7002 failsafe gates span 31 mm of the central vertical corridor.
- **Routes fully alone**: 8/8 pads in 2.2 s on an otherwise-empty grid (944 segments, 14 vias). The net is geometrically routable.
- In the full run it is routed second-to-last; the initial pass fails `blocked_path`, then `BLOCKED_BY_COMPONENT rip-up for NRST: displacing 2 sibling(s) on C5, J5, Q7, Q8, R10, SW1, U1: SWDIO, SWCLK` runs its 3-attempt ladder and `reroute did not converge`. Stranded at 1/8.

### V_BANK_POS_SENSE (2/3 pads connected)

- R25/R26 divider tap (x~111, y~188) -> U1.10 (PA3 at (158.0, 193.85), LQFP south face). The R25/R26 ends connect; the U1.10 landing through the C2-C4 decoupling band stays stranded.

## Asks

1. The #3470 rip-up rollback + stranded-stub pruning should transfer directly (10 of 11 softstart failures showed this signature pre-#3343; the remaining 2 still do). Re-measure softstart after it lands -- `tests/router/test_softstart_revb_fine_pitch_escape.py::test_softstart_revb_reach_floor` (floor 20/26 at the 480 s harness) and the recipe path (24/26 at 900 s) should both move.
[...]

Issue #3470, opened 2026-06-10
router: BLOCKED_BY_COMPONENT rip-up non-convergence strands board 05's ISENSE cluster (5 nets) and leaves overlap stub copper


rjwalters:
## Context

Issue #3425 switched board 05 to the jlcpcb-tier1 + cpp + 4L recipe with `--micro-via-in-pad-fallback` and `--per-net-timeout 60` (seed 7). The committed snapshot reaches 28/35 with exactly 1 blocking DRC violation. The residual ISENSE cluster (5 nets, all partial) is the dominant remaining gap to the 35/35 + 0 DRC bar (#3412):

| net | connected |
|---|---|
| ISENSE_A+ | 2/4 pads |
| ISENSE_A- | 1/4 pads |
| ISENSE_B+ | 2/4 pads |
| ISENSE_B- | 1/4 pads |
| ISENSE_C- | 2/3 pads |

## Classification: NOT geometrically infeasible

The initial routing pass lands 32/35 nets (overflow 28-30) — paths for the ISENSE nets exist but lose the congestion negotiation in the U3 (DRV8301) / R10-R12 shunt escape band. This is a negotiation-convergence failure, not a sum-of-clearances infeasibility.

## Three concrete defects (measured 2026-06-10, worktree issue-3425 logs)

1. **Rip-up does not converge and has no rollback.** `Autorouter` (router/core.py ~5470-5604): when the initial pass stalls, the BLOCKED_BY_COMPONENT rip-up for ISENSE_B- displaces 24 siblings (GATE_*, HALL_*, SWD, PWM_*, BST_*) and re-routes them; when the reroute "did not converge" the sibling damage is NOT rolled back. At `--per-net-timeout 30` (default) this collateral-damages HALL_A/B/C + GATE_BL from fully-routed to partial (28/35 -> 23/35). At 60s the siblings re-land (the #3425 recipe's mitigation), but the failed-net state is still worse than pre-rip-up in some runs.

[...]

Issue #3425, opened 2026-06-09
board 05: switch design.py to jlcpcb-tier1 + cpp + 4L recipe (enables via-in-pad for DRV8301)


rjwalters:
## Parent

- Decomposition issue: #3422
- Original target: #3412 (board 05 reach 60% → 100%)

## Summary

Switch board 05 design.py routing recipe from `--mfr jlcpcb --backend python --layers 2` to `--mfr jlcpcb-tier1 --backend cpp --layers 4` (auto-escalating via the project.kct `starting_layers: 4` spec). This is the manufacturing-tier change required to legally drop vias inside the DRV8301 HTSSOP-56's 0.5mm-pitch pads.

## Rationale

Issue #3412 investigation (2026-06-09 builder cycle) showed:
- The committed routed PCB at `--layers 2 --backend python --mfr jlcpcb` reaches 60% with 6 clearance_pad_segment violations.
- The same `--layers 4 --backend cpp --mfr jlcpcb-tier1` recipe reaches 62% with 18 clearance_pad_segment violations.

The 6 → 18 regression in `clearance_pad_segment` is the result of the in-pad rescue producing vias on the DRV8301's 0.5mm pitch pads. The jlcpcb base profile does NOT support via-in-pad (no epoxy fill / plating-over); the resulting vias trigger clearance errors at the manufacturer's 0.1mm clearance rule. The jlcpcb-tier1 profile (Capability Plus) supports via-in-pad legitimately — the 18 "violations" become legal manufacturing operations.

The `test_board_05_drc_hotspot_regression.py` test currently asserts the count is bounded by MAX_PAD_SEGMENT=6 against the jlcpcb tier. The test must be updated to either:
1. Change the tier from `jlcpcb` to `jlcpcb-tier1` (matching the new recipe), AND
2. Drop the floor closer to the actual cpp+tier1 floor (TBD by measurement, ~5-10 once sub-issues A+B land)

## Required changes

In `boards/05-bldc-motor-controller/design.py`:
- Change the `--backend python` pin to `--backend cpp` (or remove the pin if auto-detect is acceptable)
- Change `--manufacturer jlcpcb` to `--manufacturer jlcpcb-tier1`
- Change `--no-auto-layers --layers 2` to honor the project.kct `starting_layers: 4` (i.e., enable `--auto-layers --max-layers 4`, or pin `--layers 4`)
[...]

PR #3481, opened 2026-06-10
feat(softstart): skip-list + pin reassignment + placement — manufacturable routing (refs #3343)

rjwalters:
## Summary

Implements the architect's P-R1 → P-R4 manufacturability plan from [#3343 (comment)](https://github.com/rjwalters/kicad-tools/issues/3343#issuecomment-4670738233), plus two escape-router fixes found en route. **Refs #3343 — do NOT auto-close** (the umbrella has BOM/export ACs that need human verification).

### Measured progression (official 480 s floor harness, `PYTHONHASHSEED=0`, per-net tables on the issue)

| phase | reach |
|---|---|
| baseline (main) | 19/30 (= 17/26 at the new denominator) |
| P-R1 skip-list alignment (VGATE/SRC_POS/SRC_NEG/BUS_LINE → pours) | 18/26 |
| P-R2 north-face pin reassignment (GATE_*→PB3-PB6, PRECHARGE_*→PB7/PB8) + footprint/symbol drift reconciled | 20/26 |
| P-R3 placement micro-moves (R1/R2→U8, ISENSE_NEG pocket, SW1/C5 south) | 21/26 |
| P-R4 escape fixes | **22/26** (2 runs, deterministic; floor re-pinned at 20/26) |

### Router fixes (load-bearing, found during P-R3 diagnosis)

- **SOT-23-5 column-orientation bug**: `_escape_sop_staggered` split the wider-than-tall two-COLUMN package into y-"rows"; pin 2's stub ran through pin 3's pad, making ISENSE_NEG (U3.3) and V_AC_SENSE_RAW (U8.3) statically unroutable (instant `blocked_path` on an empty grid, present in every pre-PR baseline). Orientation now uses coordinate clustering.
- **SOT-23-class dense exclusion**: even correct stubs measured 19/26 vs 22/26 without (stagger stubs+vias consume the package pockets; A* routes ≤3-pads-per-column natively). New regression tests in `tests/test_escape.py`.

### Committed artifact (P-R4 gate, `kct route` production path — board-05 #3425/#3472 pattern)

- **Routed nets: 37/37 connected** (PR-review fix — see the 2026-06-10 doctor comment; the original 24/26 signal measurement predates the skeleton-routing rework)
- **Power: 15/15 nets geometrically continuous** (PR-review fix: GND→In1.Cu plane + +3.3V→In2.Cu plane pours; the other 11 power nets routed as 0.3-0.4 mm skeletons + reinforcement pours + B.Cu island bridges; verified by the shapely copper-union audit in route_pcb step 12, NOT the boundary-polygon analyzer — issue #3482)
- **DRC at jlcpcb-tier1: 1 error** (PR-review fix; down from 84 on the in-process path): the U8 intra-IC `clearance_pad_segment` route (fix-drc-infeasible, tracked in #3480). Zero connectivity errors.
- **Export: clean JLCPCB bundle** with manifest, committed under `boards/external/softstart/output/`
- ERC 0; schematic↔PCB netlist sync 0 errors; no spec (project.kct) edits needed

### Honest accounting — residuals + follow-ons

- **NRST (1/8)** routes 8/8 alone in 2.2 s; fails only via BLOCKED_BY_COMPONENT rip-up non-convergence (#3470 signature). **V_BANK_POS_SENSE (2/3)**: one U1 landing stranded. → **#3479** (full diagnosis; transfers to #3470)
- U8 intra-IC clearance defect → **#3480**

## Test plan

- [x] `tests/test_escape.py` — new SOT-23-5 orientation + dense-exclusion regression tests (121 passed)
- [x] Escape-adjacent suites: fine_pitch p_fp4/5/6, integration, dispatcher, subgrid, pad-clearance (153 passed; dispatcher expectations updated for the P-R2 pin move)
- [x] softstart schematic/placement/zones suites (69 passed)
- [x] Full fast suite (`pytest -m "not slow"`): pass (pre-existing `cmaes` optional-dep collection errors resolved by installing the extra locally)
- [x] Floor test re-pinned 20/26 (measured 22/26 twice at the 480 s harness)
- [x] Artifact gates: `kct check --mfr jlcpcb-tier1` (3 documented errors), `kct export --mfr jlcpcb` clean bundle + manifest
- [ ] Slow CI job re-measures `test_softstart_revb_reach_floor` (KICAD_RUN_SLOW_SOFTSTART_REACH=1)


rjwalters:
❌ **Changes Requested**

Deep review of the P-R1→P-R4 stack. Most of this PR is verified good — one blocking finding in the committed artifact.

## Blocking: AC_NEUTRAL and ISENSE_POS pours are empty — both nets are physically open

The committed `softstart_routed.kicad_pcb` has:

- **AC_NEUTRAL (net 2)**: zone at (110.24, 111.7)–(230.84, 128.3), F.Cu, priority 2, fill enabled — **0 `filled_polygon`s** (its band is shadowed by the overlapping FUSED_LINE zone at priority 3). Zero segments, zero vias on the net.
- **ISENSE_POS (net 12)**: zone at (171.6, 157.7)–(181.2, 160.25), F.Cu, priority 9 — **0 `filled_polygon`s**. Zero segments, zero vias.

Both surface as the two `zone_unfilled` DRC **warnings** in `kct check` output. On the manufactured board these nets are open circuits — AC_NEUTRAL is the mains neutral return, ISENSE_POS is the shunt high side. The exported gerbers lack this copper, so "clean export" and "15/15 pour nets fully connected" do not hold for the committed artifact.

**Why step 12 passed anyway**: `NetStatusAnalyzer._build_connectivity_graph` counts pads inside the zone **boundary polygon** as zone-connected (the #479 thermal-relief-cutout fix) even when the zone produced zero filled polygons. The per-pad verification is real code, but it has a false-positive mode and it fired here — filed **#3482** for the analyzer gap (pre-existing).

**Required in this PR:**
1. Fix the two pours so each produces actual copper reaching its pads (AC_NEUTRAL's outline needs to escape the FUSED_LINE shadow or win the carve; ISENSE_POS's box is being carved to nothing).
2. Harden the route_pcb step-12 gate: a skip-net whose zones have fill enabled but zero `filled_polygon`s must count as NOT connected (or fail the gate on `zone_unfilled` warnings hitting skip-list nets) — don't wait for #3482.
3. Re-run `kct check` + `kct export`, recommit the artifact, and update the pour/DRC claims in the PR body if numbers move.

## Verified (everything else holds up)

**P-R2 pin reassignment** — VERIFIED.
- Timer claims check out against the STM32G031 AF table: PB3=TIM1_CH2 (AF1), PB4=TIM3_CH1 (AF1), PB5=TIM3_CH2 (AF1), PB6=TIM1_CH3 (AF1). All four gate signals keep TIM1/TIM3 PWM capability; PB7/PB8 precharge as plain GPIO is fine. Note GATE_POS_A/B now straddle TIM1/TIM3 (no hardware complementary dead-time pairing), but the prior PA7/PB0–PB2 map didn't have that either, and rev B is droop-triggered burst conduction per the PR.
- The LQFP-32 pad-net map reconciliation to canonical STM32G031K8Tx numbering is correct (pin 4 VDD / pin 5 VSS / pin 6 PF2-NRST / pins 27-32 PB3-PB8) — the old map genuinely was offset.
- Firmware impact documented prominently (PR body, module docstring, schematic comment, PCB pin map).
- Upstream spec: `project.kct` (symlink mirror of rjwalters/softstart's spec) pins only the SWD header — claim verified, this is recipe-controlled. **However**: upstream `docs/pin_mapping.md` and `firmware/include/softstart.h` pin the rev A map (PA7=PWM_NEG etc., PB3-PB8 marked "available for expansion"). The whole rev B redesign is unmirrored upstream, so this isn't new debt from this PR — but when the rev B mirror lands (#3343 umbrella), the PB3-PB8 map must propagate to upstream docs + firmware. Worth a checklist line on #3343.

**Escape fixes blast radius** — VERIFIED, well contained.
- The new SOT-23-class exclusion in `is_dense_package` is ordered AFTER the ≤0.75 mm fine-pitch rule, the SOIC-8 band rule, and the TQFP-32 quad rule, so it only affects 3-7-pad all-SMD two-row/column packages with pitch in the narrow (0.75 mm, 2·(trace+clearance)) band — exactly SOT-23-5/6 class. `test_fine_pitch_small_dual_row_still_dense` pins the ordering.
- `_escape_sop_staggered` orientation fix is conservative: cluster-count signature with explicit fallback to the legacy spread heuristic for ambiguous (2×2) layouts.
- Regression tests pin all four behaviors (not-dense at production rules, fine-pitch stays dense, no stub crosses same-package pads, column-perpendicular stub direction). `tests/test_escape.py`: 121 passed locally in the worktree.
- Cross-board: boards 06/07 routing regressions run in CI (pending at review time); fine-pitch/dispatcher/subgrid suites updated and passing per test plan. Boards 02/03/05 have no per-PR CI reach baselines — acceptable given the narrow pitch band, but the pending CI Test job should be green before re-review.

**Skip-list (P-R1)** — semantics correct: `ROUTE_SKIP_NETS` is now the single source of truth and the three test files reference it. The pour-connectivity check itself is the blocking item above.

**DRC 84→3** — VERIFIED, not gaming. The drop is the in-process `route_with_escape` negotiated loop leaving overflow copper (75 clearance errors) vs `kct route` running the DRC-nudge + validation post-passes (board-05 #3425/#3472 production pattern). The A/B is documented in the `route_pcb` comment. I independently reproduced exactly 3 errors via `kct check --mfr jlcpcb-tier1` on the committed artifact (1 clearance_pad_segment = #3480, 2 connectivity = #3479 residuals).

**refs-only trailer** — VERIFIED: title carries "(refs #3343)", body says do-NOT-auto-close, `closingIssuesReferences` is empty, and all four branch commits use `refs` not `Closes`.

**CI** — Test, board-06, and board-07 jobs were still pending at review time (mergeStateStatus UNSTABLE, not DIRTY — no conflict with the #3478 doctor rebase as of this review). These must be green at re-review.

Nice work on the escape-router diagnosis — the column-orientation bug write-up and the honest residual accounting (#3479/#3480) are exactly right. Fix the two dead pours and the gate, and this is ready.

rjwalters:
## Review feedback addressed (doctor cycle, 2026-06-10)

All three blocking items are fixed, and the artifact is strictly better than the bar the review set.

### 1. The dead pours (AC_NEUTRAL / ISENSE_POS) — fixed by reworking the power-copper strategy

The geometric root cause was bigger than the two zones the review caught: single-rectangle-per-net F.Cu pours **cannot tile this board** (the AC front-end pads interleave along one band; the discharge-cluster pads interleave around the MOSFET pairs), so the fill-priority resolver always leaves some sibling with zero or islanded copper. Patching the two reported zones would just have moved the open circuit.

The recipe now pours only what is geometrically sound (GND → In1.Cu plane, +3.3V → In2.Cu plane, SCAP_*_GND single-pad ties) and **routes the other 11 power nets as 0.3–0.4 mm skeleton traces** (`--net-class-map` sidecar) with F.Cu reinforcement pours on top. AC_NEUTRAL and ISENSE_POS are now routed copper, not pours.

Supporting changes that were required to make that strategy converge:
- **Steiner grid-snap (router)**: synthetic Steiner branch points inherit raw terminal coordinates and land off the routing grid; virtual pads have no sub-grid rescue, so multi-terminal power nets failed `pin_access` with `PADS_OFF_GRID: steiner@(...)`. `build_rsmt` now takes a `snap_fn` (mst.py / negotiated.py wire it to the grid). Unit-tested in `tests/test_steiner.py::TestSteinerGridSnap`; terminals are never snapped and the no-snap path is bit-identical.
- **ROUTE_FIRST_NETS** (BUS_LINE / VRECT / V_BANK_POS_SENSE / UCC_LO_NEG) at net-class priority 1 — measured congestion-ordering losers that route in ~2 s on an empty grid. (A pre-route + `--preserve-existing` freeze was tried first and regressed to 26%: frozen long-hauls partition the signal layers; the ordering lever keeps the copper negotiable.)
- **Recovery pass** (`--preserve-existing`, #3155) for main-pass stragglers, and a **B.Cu bridge pour + via-in-pad anchors** step for islanded supply nets (closes VGATE's U6-cluster island deterministically).
- Two new deterministic repairs: drill-to-drill (an NRST via landed inside SW1.2's TH drill) and cross-net segment/via clearance (a SRC_NEG via physically overlapped a UCC_LO_NEG trace — a short the same-net audit cannot see; 3 more barrel-vs-inner-layer cases were invisible to `kct check`, see #3487).

### 2. Hardened step-12 gate
The gate now runs a recipe-local **shapely copper-union audit** (`_audit_pour_nets`): every power/bridged net must have ALL pads in ONE geometrically connected component (filled polygons + segments at width + via barrels + pad copper), and **any fill-enabled zone with zero filled polygons fails the gate**. If shapely is unavailable the gate FAILS rather than degrading silently — the silent path is exactly how the dead pours shipped. It deliberately does not use `NetStatusAnalyzer`'s boundary-polygon test (#3482).

### 3. Recommitted artifact, measured state (seed 42, `PYTHONHASHSEED=0`)
- **37/37 routed nets connected** (NRST and V_BANK_POS_SENSE now close — #3478's rip-up rollback from main, which this branch is rebased onto, plus the ordering fix)
- **15/15 power nets geometrically continuous**, zero zero-fill zones
- **DRC at jlcpcb-tier1: 1 error** — the tracked #3480 U8 `clearance_pad_segment` residual. Zero connectivity errors (previously 2, plus the two silent opens).
- Manufacturing bundle regenerated (`output/manufacturing/`, 8 files + manifest).
- PR body claims updated to match.

### Scope note on the router-side edits
`mst.py` / `negotiated.py` / `steiner.py` changes are the minimal `snap_fn` plumbing described above — kept, not reverted, because the pour fix is unreachable without them (the skeleton strategy needs multi-terminal power nets to route). `tests/test_negotiated_adaptive.py` fakes updated for the new keyword.

### New issues filed from this cycle
- #3485 — `--per-net-timeout` not enforced inside the negotiated rip-up reroute (VGATE ground 20-30 min per attempt; the mechanism behind #3479's NRST note)
- #3486 — router via placement violates clearance against committed foreign-net segments (the cross-net short class)
- #3487 — `clearance_segment_via` checks F.Cu only; via barrels vs inner/bottom-layer segments are unreported

Rebased onto main (picks up #3478 rip-up rollback and #3483 cpp gating). `tests/test_steiner.py` + `tests/test_escape.py`: 162 passed; full fast suite matches main's local-environment baseline (the 88 remaining local failures reproduce identically on the main checkout — environmental, e.g. missing kicad-cli).

rjwalters:
## Re-review (judge, doctor cycle 2) — APPROVED

The prior blocker (electrically-open AC_NEUTRAL / ISENSE_POS pours) is fixed, and the fix is the right shape: instead of patching two zones, the recipe abandons the geometrically-unsound single-rectangle-pour strategy and routes the 11 distribution nets as skeletons, pouring only what can actually tile (In1 GND plane, In2 +3.3V plane, SCAP ties).

### Independently verified on commit 4a236eb6

- **37/37 routed nets complete** — re-ran `NetStatusAnalyzer` on the committed `softstart_routed.kicad_pcb`: 37/37, zero failures. NRST and V_BANK_POS_SENSE (the #3479 residuals) now close.
- **15/15 power nets geometrically continuous** — re-ran the recipe's shapely copper-union audit (`_audit_pour_nets`) independently: every power net's pads in ONE component, zero zero-fill zones.
- **AC_NEUTRAL / ISENSE_POS have real copper now**: AC_NEUTRAL = 742 segments + 4 vias, ISENSE_POS = 8 segments. All 17 zones in the artifact have ≥1 `filled_polygon`.
- **DRC: exactly 1 error** (`kct check --mfr jlcpcb-tier1` re-run locally): the tracked #3480 `clearance_pad_segment` U8 residual. Matches the PR claim.
- **Export bundle regenerated**: manifest checksums/sizes updated (gerbers.zip 211 KB → 257 KB), generated_at 2026-06-10T23:50Z — post-dates the doctor commit.
- **CI**: all 8 checks green, merge state CLEAN, `closingIssuesReferences` empty (refs-only trailers intact — #3343 will not auto-close).

### snap_fn plumbing (router-core change riding a board PR) — accepted

- Default behavior unchanged: `build_rsmt(snap_fn=None)` is a no-op for all external callers; only SYNTHETIC Steiner points (idx ≥ num_terminals) are ever snapped, terminals never.
- Both internal callers (`MSTRouter`, `NegotiatedRouter`) snap via `world_to_grid`→`grid_to_world` round-trip — the same grid A* searches, so the snap can only make virtual pads MORE reachable. Justification (`PADS_OFF_GRID: steiner@(...)` with no sub-grid rescue for ref-less virtual pads) is sound and load-bearing for the skeleton-routing strategy.
- Blast radius (every multi-terminal net): covered by the green board-06 diff-pair and board-07 match-group routing regression CI jobs, plus `TestSteinerGridSnap` pinning all three contracts (synthetic points land on-grid, terminals untouched, `snap_fn=None` byte-identical legacy output). Ran `tests/test_steiner.py`, `tests/test_negotiated_adaptive.py`, `tests/router/test_softstart_revb_fine_pitch_escape.py` (161 passed) and `tests/test_router_algorithms.py` (13 passed) locally.

### Step-10b/c/d repair passes — acceptable as recipe-local mitigations

Each pass logs every individual repair with measured distances and prints a count; a future repair-count explosion will be visible in the build log, and `kct check` + the hardened step-12 gate re-validate the mutated artifact downstream. The step-12 gate now fails on zero-fill zones, on shapely being unavailable, and audits step-10b bridged nets geometrically (not via the #3482-affected analyzer) — exactly the rule the prior review asked for.

### Follow-on issues — high quality

#3485 (per-net timeout not enforced in rip-up reroute), #3486 (via placement can short against committed foreign-net segments), #3487 (DRC `clearance_segment_via` blind to via barrels vs inner/bottom-layer segments) are all filed from measured coordinates/clearances on this artifact, with repro paths. The #3486/#3487 pair is serious — a short class invisible to both the connectivity audit AND `kct check` — and correctly notes the repair logic belongs in the router/DRC, not the recipe.

### Minor (non-blocking)

`_repair_drill_drill_vias` / `_repair_segment_via_clearance` drag endpoints by raw-coordinate `re.sub` without a net filter; an exact cross-net coordinate collision could move an unrelated endpoint. Improbable (and itself a defect if present), validated downstream by the gate — fine for a recipe, and the proper fix is already filed as #3486/#3487.

**Verdict: approve.** Labels updated to `loom:pr`.

Issue #3485, opened 2026-06-11
router: --per-net-timeout not enforced inside negotiated rip-up reroute (20-30 min single-net grinds)


rjwalters:
Observed repeatedly while fixing PR #3481 (softstart board, L=4, cpp backend, jlcpcb-tier1, seed 42, PYTHONHASHSEED=0):

- `kct route --per-net-timeout 60` (and 90/150) does NOT bound the per-net search inside the `BLOCKED_BY_COMPONENT` rip-up / iteration reroute path. VGATE (8 pads, multi-terminal Steiner) ground for 20-30+ minutes of wall clock inside a single `Re-routing net N/M: VGATE...` step, at 100% CPU, across several runs.
- The board-level `--timeout` also failed to fire mid-net: a run with `--timeout 1200` sat inside one VGATE reroute for >30 min (killed manually). Deadline checks appear to happen only between nets / phases, not inside the (cpp) A* / multi-terminal expansion.
- Repro: softstart rev B recipe (`boards/external/softstart/generate_design.py`, PR #3481 branch), watch the rip-up phase after the initial-pass stall.

Impact: a single pathological net can consume the entire routing budget unmonitored, starving the iteration loop (this is also the mechanism behind the #3479 NRST "360 s rip-up" note).

Suggested fix: plumb the per-net deadline into the cpp backend search loop (periodic expansion-count check) and check the board deadline inside the rip-up reroute loop, not just between nets.

Found during PR #3481 doctor cycle (refs #3343).

PR #3483, opened 2026-06-10
fix(router): cpp-backend off-grid recovery gating + lattice-aware auto-grid (closes #3441)

rjwalters:
Closes #3441

## Summary

Implements both halves of the #3441 curator diagnosis, plus three latent bugs the fix exposed.

### 1. Backend-aware waypoint gating (the correctness bug)

Waypoint injection (#2330) exists ONLY in the pure-Python pathfinder, yet `Autorouter.use_waypoint_injection` was hardcoded `True` with no backend gate. Under `--backend cpp` this simultaneously disabled the sub-grid escape pre-pass (`core.py` `_run_subgrid_prepass`) AND the PIN_ACCESS sub-grid retry — all three off-grid recovery mechanisms were off at once while the banner claimed coverage. This was the root cause of board 07's `--grid 0.1` 13/31 collapse.

- `use_waypoint_injection` is now a property: request flag AND `supports_waypoint_injection` capability on the active backend (`Router` = True, `CppPathfinder` = False; flip the attribute if/when waypoints are ported to C++).
- The misleading `route_cmd.py` banner now names the mechanism that actually runs per backend; the false "the C++ pathfinder already handles off-grid pads" comment is corrected.

### 2. Coverage-gated sub-grid pre-pass (empirical correction)

Blanket stub synthesis for every nominally off-grid pad measured WORSE on board 07 (26/31 + partials vs 28/31 at 0.127mm — 242 stubs congest the QFN/BGA corridors). A* seeds start nodes from every grid cell inside pad metal (#977), so the pre-pass now emits stubs only for pads whose copper contains NO grid cell (`_pad_metal_covers_grid_cell`); covered-but-blocked pads are recovered by the failure-driven PIN_ACCESS retry.

### 3. Lattice rescue in `auto_select_grid_resolution` (with a 90% dominance bar)

The memory filter ran BEFORE the off-grid vote and the #3239 bump only adopts candidates `<= clearance/2`, so a board-lattice grid that is merely `<= clearance` could never be selected. The selector now retries the budget bump and adopts it iff the unlocked candidate strictly reduces off-grid pads AND places >= 90% of pads on-grid (the issue's own "dominant lattice" framing), with an honest warning replacing the misleading "memory budget cap forces" one.

[...]

PR #3478, opened 2026-06-10
fix(router): rip-up rollback + ISENSE cluster convergence (refs #3412, closes #3470)

rjwalters:
Closes #3470

## Summary

Board 05's single blocking DRC violation and the BLOCKED_BY_COMPONENT rip-up collateral-damage failure mode are both fixed at the source. The committed snapshot is refreshed: **28/35 nets routed with 0 blocking violations at jlcpcb-tier1** (was 28/35 + 1 `clearance_segment_segment`), and the board-05 tolerance entry is **removed** (strict 0-blocking gate) per the #3472 contract.

### Root cause of the blocking violation (new finding)

The committed overlap (`ISENSE_A-`/`ISENSE_B-`, In1.Cu at (18.75, 53.75), actual -0.3135 mm, stable across seeds) was NOT created by the rip-up at all -- it was created at **escape generation time**. The in-pad escape rescues for U3 pin 31 (ISENSE_B-) and pin 33 (ISENSE_A-) emitted inner-layer stubs pointing TOWARD each other across the intervening pin-32 column:

```
pin 31 stub: (118.25,153.75) -> (118.80,153.75)  w=0.127  In1.Cu
pin 33 stub: (119.25,153.75) -> (118.70,153.75)  w=0.500  In1.Cu
```

The stubs physically overlap (exactly -0.3135 mm = (0.5+0.127)/2) and each net's escape endpoint (virtual pad) sits inside the other net's stub copper -- making both nets deterministically hard to route and shipping overlap copper on EVERY route of this board. That is why the violation was seed-stable.

## The three fixes

1. **Transactional rip-up rollback** (`negotiated.py targeted_ripup`): the failed net's and every displaced sibling's routes are snapshotted before ripping. The transaction commits only when the failed net routes FULLY (zero edge failures -- a partial reroute is stranded-stub copper) and every sibling re-lands with >= its original route count. On any other outcome the EXACT original Route objects are re-marked. This replaces the #2814 "10s-probe fresh reroute" sibling restore, which routinely failed and stranded HALL/GATE siblings. Stale failed-net partial routes are now ripped INSIDE the transaction (the negotiated caller's out-of-transaction pre-rip is removed).
[...]

Issue #3412, opened 2026-06-09
board 05 (bldc-motor-controller): 4L production routing reaches only 60% (target: 100%) — congestion gap (issue #3402 follow-on)


rjwalters:
## Background

Per Issue #3402 audit (2026-06-09), board 05 was set to \`starting_layers: 4\` because 4L is materially better than 2L:

- 2L: 16/35 routed (46%) — production tier
- 4L: 21/35 routed (60%) — production tier, ladder rung 1 (4-Layer SIG-GND-PWR-SIG)

The audit also kicked off a 4-Layer ALL-SIG attempt that did not finish within the audit's wall-clock window, but the 60% reach gap from 21/35 to 35/35 on rung 1 is itself a manufacturability gap that needs a follow-on.

## Reproduction

\`\`\`bash
uv run kct build-native --check
uv run kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb \\
  --auto-layers --max-layers 4 -o /tmp/board05.kicad_pcb
\`\`\`

Note: full 4L auto-layers escalation on board 05 takes 30+ minutes on the audit hardware — the inner-loop A* is congestion-limited around the DRV8301 (U3) QFN-56 escape and the STM32G431 (U10) LQFP-48 fan-out.

## Diagnosis (audit observations)

- Iteration 0 on 4L SIG-GND-PWR-SIG: routed 21/35 with overflow=16.
- Iteration 1: re-route 5 nets, overflow grows to 32 — negotiated A* not converging.
- 4L ALL-SIG (rung 2) starts but does not converge inside the audit window.
- Existing routed artifact (\`bldc_controller_routed.kicad_pcb\`) has connectivity errors: PWM_CH (1/2 pads stranded), PWM_CL (1/2 stranded), SW_OUT (3/5 stranded).

## Scope (suggested)

- Investigate the DRV8301 / STM32 placement to ease the fan-out density (#2698 was a previous placement re-tune for the same family of regressions).
- Consider \`--fine-pitch-clearance\` recipe for QFN-56 (U3) and LQFP-48 (U10) escapes.
- Acceptance: \`kct route boards/05-bldc-motor-controller/output/bldc_controller.kicad_pcb --auto-layers --max-layers 4 --starting-layers 4\` reaches 35/35 (100%) with 0 DRC at the jlcpcb tier.

## Out of scope

- The 2L probe is already disabled by \`starting_layers: 4\` (Issue #3402).
[...]

PR #3491, opened 2026-06-11
fix(boards/05): close residual partial nets (refs #3412, closes #3471)

### LINKED ISSUE #3471: boards/05: SW_OUT + PWM_CL residual partial nets at the tier1/cpp/4L recipe (28/35 -> 30/35)
## Context

After #3425 (jlcpcb-tier1 + cpp + 4L + micro-via in-pad fallback, seed 7) board 05's committed snapshot reaches 28/35. Besides the ISENSE cluster (#3470), two more nets are partial:

| net | connected | notes |
|---|---|---|
| SW_OUT | 3/5 pads | buck converter switch node: U1 (LM2596 TO-263), L1, D2 — 5-pad high-fanout net |
| PWM_CL | 1/2 pads | U10 (MCU) -> U3 (DRV8301) low-side PWM; crosses the congested U10/U3 escape band |

## Classification: needs-more-work (not geometrically infeasible)

The initial routing pass routes both nets with overflow (paths exist; they lose the valid-state extraction). PWM_BH — the same family — routes fine at seed 7, and PWM_CL itself routed in some configs of the #3425 measurement matrix (e.g. it appears only in the unrouted list of some variants). No sum-of-clearances proof of infeasibility applies.

## Diagnosis pointers

- SW_OUT's 2 stranded pads are in the buck corner (top-left); the net is wide-trace (power class) and may benefit from a dedicated pour or stitch instead of trace routing — evaluate adding SW_OUT to the zone/pour set the way +5V/+24V are handled, or a wider per-net budget.
- PWM_CL's MCU-side pad strands when the ISENSE rip-up (#3470) consumes the south corridor; fixing #3470 likely frees it. Re-measure after #3470 lands before independent work.

## Acceptance sketch

- Board 05 fresh route at the production recipe (design.py route_pcb flags) reaches >= 30/35 with SW_OUT and PWM_CL fully connected, no new blocking violations, snapshot + test gates re-pinned honestly in the same PR.

Refs #3425, #3412. Blocked-by recommendation: build after #3470.

rjwalters:
## Summary

Board 05's residual partial nets were never congestion-bound: two router defects made every Steiner-incident RSMT edge structurally unroutable on this board, and the negotiated valid-state extraction could not land the ISENSE Kelvin cluster simultaneously. This PR fixes both defects at the source, adds a per-net rescue loop to the board recipe, and refreshes the committed artifact: **31/35 nets routed, 0 blocking DRC** (was 28/35, 0 blocking).

### Root causes (new findings)

1. **C++ `_compute_pad_bounds` empty-bounds defect** (`cpp_backend.py`): zero-area Steiner virtual pads (width = height = 0) produced EMPTY metal bounds whenever `(x - origin) / resolution` was not exactly integral in floating point (`ceil(v) > floor(v)`), so the C++ A* seeded **zero start nodes** and failed the edge with `FAILURE_NO_PATH` at 0 iterations. Every 4-pad ISENSE net's RSMT synthesises branch points, so the U3-side edges failed **on every route of this board**, and every BLOCKED_BY_COMPONENT rip-up "did not converge" in seconds even with all 25 siblings displaced — the prior "per-edge A* burns the full per-net budget" diagnosis (#3470/#3425) was wrong. Fix: clamp empty spans to the nearest grid cell.
2. **RSMT branch points on blocked cells** (`steiner.py`, `negotiated.py`, `mst.py`): ISENSE_A+'s Steiner point at (136.9, 176.0) landed on the Q5 MOSFET TH-leg keepout — the net was `blocked_path` even **solo on an empty board**. Fix: `relocate_blocked_point` ring-scans to the nearest cell free (with trace-radius + clearance margin) on at least one routable layer, applied in both the negotiated and MST RSMT paths via the shared `make_blocked_cell_predicate`.
3. **Negotiated extraction cannot land the sense-band cluster** (`design.py` step 6b): after the fixes, every residual net routes solo on an empty board, but the ISENSE Kelvin cluster mutually excludes in the U3 (DRV8301) 0.5 mm-pitch south escape band. The new `rescue_partial_nets` loop strips residual partial copper and re-routes each net alone against the committed copper (`--preserve-existing` + `--skip-nets`), landing what the global negotiation cannot extract. Failed rescues strip their copper — no stranded stubs (the #3470 lesson).

## Measurements (production recipe, seed 7, jlcpcb-tier1 + cpp + 4L)

| configuration | reach | blocking |
|---|---|---|
| committed snapshot (pre-PR) | 28/35 | 0 |
| re-measured baseline on main (#3478 + #3483) | 28/35 | 0 |
| `--per-net-timeout 120` (the #3485 budget-bump probe) | 27/35 | 0 |
| `--grid 0.075` (auto-grid warning's recommendation) | DNF (budget leak amplified) | — |
| + Steiner fixes (this PR), main pass | 29/35 | 0 |
| **+ rescue loop (committed artifact)** | **31/35** | **0** |

SW_OUT (5 pads incl. U3.50/51 fine-pitch) now routes fully in the main pass; ISENSE_A+ and PWM_CL are rescued by the loop.

## Per-net residual diagnosis (honest accounting vs the 35/35 target)

| net | status | diagnosis |
|---|---|---|
| SW_OUT | **routed** (main pass) | was killed by the empty-bounds defect (intra-IC pair U3.50/51 + Steiner edges) |
| PWM_CL | **routed** (rescue) | freed once the cluster's doomed-edge churn stopped consuming the corridor |
| ISENSE_A+ | **routed** (rescue) | was killed by the blocked Steiner point on Q5's TH leg |
| ISENSE_A- | partial 1/4 | no path vs the routed background; routes solo on empty board |
| ISENSE_B+ | partial 1/4 | same |
| ISENSE_B- | partial 1/4 | same |
| ISENSE_C- | partial 1/3 | same |

The four residual ISENSE nets are **geometric capacity**, not budget: each routes solo on an empty board (post-fix), rescue A* exhausts (fast-fail, not timeout) against the routed background, and raising budgets makes reach worse (pnt=120 -> 27/35). Their stranded copper is stripped (escape-stub islands only), keeping the 0-blocking gate clean. Next levers are placement-level (widen the R10-R12 sense band or move the U10-bound trio) or escape-band-aware routing — follow-on scope, documented in the design.py docstring.

## Changes

- `src/kicad_tools/router/cpp_backend.py` — degenerate pad-bounds clamp + `CppGrid.is_blocked_for_net`
- `src/kicad_tools/router/grid.py` — `RoutingGrid.is_blocked_for_net` (uniform index-based API)
- `src/kicad_tools/router/algorithms/steiner.py` — `relocate_blocked_point` + shared `make_blocked_cell_predicate`
- `src/kicad_tools/router/algorithms/negotiated.py`, `mst.py` — relocation wired into both RSMT paths
- `boards/05-bldc-motor-controller/design.py` — step 6b `rescue_partial_nets` + recipe docstring (#3471 findings, negative results recorded)
- `boards/05-bldc-motor-controller/output/` — routed artifact + manufacturing bundle refreshed (design.py steps 6/6b/8/9/10 over the committed unrouted PCB; no schematic churn)
- `tests/test_steiner.py`, `tests/test_router_cpp_backend.py` — regression tests for the relocation helper, the shared predicate, and the bounds clamp

| Board 05 reach improves materially (>= 30/35 per issue sketch) | ✅ | 28/35 -> 31/35 committed artifact (`kct check` connectivity) |
| SW_OUT and PWM_CL fully connected | ✅ | neither appears in the partial-net list of the committed artifact |
| 0 blocking DRC maintained (strict gate) | ✅ | `kct check --mfr jlcpcb-tier1`: 0 blocking; tolerance entry stays removed; allowlist + hotspot tests green |
| Honest per-net diagnosis for residuals | ✅ | table above; isolation / rip-up / budget experiments recorded in the design.py docstring |
| Committed artifact + tests refreshed per policy | ✅ | board-05 battery (35 passed, 1 xfailed) incl. allowlist/hotspot/zones/export/thermal/rotation; CI mirrors (80 passed) |
| No cross-board regression | ✅ | `tests/router/` 212 passed (1 stale-native env failure resolved by rebuild; 17/17 relief tests pass); `test_router_core` 172 passed + 2 documented pre-existing failures; negotiated/steiner/escape suites at exact main failure-parity |

## Test plan

- [x] `tests/test_steiner.py` (47+ passed) — relocation helper + predicate + snap integration
- [x] `tests/test_router_cpp_backend.py` — new degenerate-bounds regression test (sweeps 86 grid positions)
- [x] Board-05 battery: allowlist, hotspot, zones, export, thermal, rotation, buck, ERC (35 passed, 1 xfailed)
- [x] CI gate mirrors: `test_ci_routed_drc_workflow`, `test_fleet_status_drc`, `test_fleet_ship_ready`, drift regression (80 passed)
- [x] `tests/router/` (212 passed), `test_router_core` (172 passed, 2 pre-existing), negotiated suites (failing-test diff vs main: IDENTICAL)
- [x] `kct pcb sync-netlist --dry-run`: no pad-net mismatches (only the long-standing MH1-MH4 orphans)

Pre-existing failures (identical on main): `test_router_core.py::TestPadBlockedByAdjacentNetZero::test_route_to_pad_adjacent_to_net0_pad`, `test_router_core.py::TestAdaptiveAutorouterComponentTransform::test_add_component_rotation`, and 4 stub-based tests in `test_negotiated_timeout_propagation.py`.

Closes #3471


rjwalters:
✅ **Approved!**

Thorough review with independent verification in a clean detached checkout of the PR head (uncommitted concurrent-session edits to `two_phase.py`/`core.py` in the issue-3471 worktree were confirmed EXCLUDED from the PR and quarantined from all test runs).

## Root-cause fixes — verified sound

**1. Empty-bounds clamp (`cpp_backend.py`) — HIGH blast radius, verified safe.** The clamp only fires when `gx1 > gx2` (previously a guaranteed FAILURE_NO_PATH at 0 iterations), so it strictly converts hard failures into seeded searches. `gc = round((pad.x - origin)/res)` clamped to `[0, cols-1]/[0, rows-1]` cannot produce out-of-board indices. Cross-board gates green: board-06 diff-pair and board-07 match-group routing regressions pass, Routed PCB DRC Check passes. The 86-position sweep regression test executes against the real C++ extension (verified not skipped) and passes. Bonus: the clamp also rescues sub-resolution nonzero pads that previously hit the same empty-bounds hole.

**2. `relocate_blocked_point` — terminates, bounded displacement.** Chebyshev ring scan, radius capped at 20 cells (2 mm at 0.1 mm grid), deterministic row-major order (tested), falls back to original coords (legacy failure path) when saturated. Margin-aware predicate (`trace_width/2 + clearance` ceil'd to cells) correctly prevents relocation to wall-hugging cells. `make_blocked_cell_predicate` is fail-open: returns None on fixture/mock grids and treats per-cell exceptions as free, preserving legacy snapping byte-for-byte. Both `RoutingGrid.is_blocked_for_net` and `CppGrid.is_blocked_for_net` have identical semantics (same-net copper passable; net-0 and foreign copper block) and both grids expose `get_routable_indices`.

**3. `rescue_partial_nets` strip verified against the committed artifact.** `_strip_net_copper`'s `^\t\(segment|via` parser matches the actual artifact format (771 tab-indented segments, 56 vias). The four failed-rescue nets (ISENSE_A-/B+/B-/C-) each retain exactly **one** short (≤0.76 mm) pad-attached F.Cu escape stub + via near U3 — the "escape-stub islands" the PR body discloses. These are DRC-clean (independently re-ran `kct check --mfr jlcpcb-tier1`: 0 clearance/geometry errors; the 8 connectivity errors are the 4 documented ISENSE residuals + the pour-carried GND/PHASE_A/B/C known class). Not the #3483 blocking-stub class.

## Independent verification

- `tests/test_steiner.py` + `tests/test_router_cpp_backend.py`: 97 passed, 6 skipped (clean checkout, real .so)
- Key new regressions (`test_compute_pad_bounds_zero_width_steiner_pad_never_empty`, `TestMSTRouterSteinerRelocation`): executed, passed
- Negotiated suites (adaptive/per-net-timeout/convergence-stranded): 132 passed
- Committed artifact re-checked: 31/35 signal-net reach, 0 blocking — claims confirmed
- CI: all 8 checks pass; mergeStateStatus CLEAN
- `closingIssuesReferences` = [#3471] only; residual-4 ISENSE nets remain tracked by open umbrella #3412 (loom:blocked, target 100%) plus the #3422 decomposition, and the diagnosis is recorded in the design.py docstring

## Provenance / scope

The MSTRouter relocation port (54c95e89) and shared-predicate refactor are committed on the branch, in-scope (MSTRouter is the path the auto-layers two-phase flow executes — without the port the fix would miss the production recipe), and covered by dedicated tests. The concurrent session's uncommitted relief-rescue work was correctly left out.

## Minor notes (non-blocking)

1. The clamp also fires for fully off-board pads (previously clean zero-seed failure → now seeds the nearest board-edge cell). Pathological placement only; any bad route would be caught by DRC downstream.
2. `rescue_partial_nets`'s docstring claim "a FAILED rescue ... leaves no stub copper" is slightly stronger than the artifact reality (later rescue stages re-emit one escape stub per still-unrouted net). The PR body describes the end state accurately; consider tightening the docstring in a follow-on.

Honest 31/35 framing throughout — no done-claim; the manufacturable bar (35/35 + 0 DRC) remains tracked by #3412.

rjwalters:
Independent verification from the parallel #3471 retry session (measurements taken in the same worktree while this PR was being assembled — coordinating post-merge):

- **Post-#3488 fresh-main baseline**: 26/35 (vs the 28/35 pre-#3488 baseline in the PR table; failures all `blocked_path`: ISENSE x5, SW_OUT, PWM_CL, GATE_BL, GATE_CL, HALL_A). The Steiner fixes in this PR recover that and more — with them, the same recipe measures 29/35 in the main pass (SW_OUT fully connected), matching the PR's main-pass row.
- **Solo-net check**: `ISENSE_B-` routes 1/1 on an otherwise-empty board post-fix (confirms "routes solo" in the residual table).
- **Budget lever independently confirmed dead**: `--timeout 1500` at otherwise-identical flags is a DNF — the run was still stuck inside a single mid-net grind at 40 minutes wall (the #3485 deadline-enforcement gap; the wall deadline never fires mid-net). Consistent with the PR's pnt=120 -> 27/35 negative result and the "geometric capacity, not budget" diagnosis.
- **Stall-path observation**: the BLOCKED_BY_COMPONENT rip-up for `ISENSE_B-` fails at 0.0s **with all 24 siblings displaced** — the blocker is non-rippable escape copper, which sibling rip-up cannot clear by construction. The #3488 relief rescue is never reachable from the two-phase stall path; experimentally wiring it in engages correctly (victim trend 10/4/2/1 across rounds) but hits `max_rounds=4` and rolls back verbatim (reach unchanged, transaction held). Filed as #3494 with a ready-made branch (`feature/two-phase-stall-relief`) — held out of this PR to avoid churning the review with a measured-neutral change.

Issue #3502, opened 2026-06-11
tests: board-05 throughput test can no longer find the 'Routed N/M nets' iteration-0 line


rjwalters:
## Observed

`tests/test_board_05_routing_regression.py::TestBoard05RoutingThroughput::test_initial_pass_routes_enough_nets` fails on main (HEAD cb7c2b8e, 2026-06-11):

```
assert actual_count is not None
Could not find 'Routed N/M nets' iteration-0 line in kct route output.
```

The helper `_parse_iter0_route_count` greps for `Routed N/M nets, overflow: K` which negotiated routing used to emit after the iteration-0 initial pass. The route output captured by the test now ends in the escape/two-phase summary format (`Initial pass: N/M nets, overflow: K` + grace-pass lines per #3452/#3499) and the legacy `Routed N/M nets, overflow:` line is no longer present on this path.

Pre-existing on main — reproduced identically in a clean checkout and in the issue-3474 worktree (discovered while running the board-05 suite for #3474 R2 verification; unrelated to that branch's changes).

## Suggested fix

Update `_parse_iter0_route_count` to also accept the two-phase form (`Initial pass: (\d+)/\d+ nets, overflow:`) or whichever line the current negotiated/two-phase path emits for the iteration-0 tally, keeping the original pattern for backward compatibility.

## Context

- The throughput floor itself (#2681) is still meaningful; only the stdout parsing is stale.
- Sibling staleness class: #3492 (softstart call-site tests stale after recipe overhaul).

Issue #3492, opened 2026-06-11
tests: softstart slow-gated baseline + call-site tests stale since PR #3481 recipe overhaul


rjwalters:
## Symptom

Three env-gated softstart tests fail on current main (`d83fa4a6`), independent of any open PR (verified with `router/path.py`+`router/core.py` reverted to main):

1. `tests/router/test_softstart_manufacturable_baseline.py::TestSoftstartManufacturableBaseline::test_all_signal_nets_topologically_connected` (KICAD_RUN_SLOW_SOFTSTART_REACH=1)
2. `tests/router/test_softstart_manufacturable_baseline.py::TestSoftstartManufacturableBaseline::test_residual_clearance_within_budget` (same gate)
3. `tests/test_router_dense_escape_softstart.py::TestSoftstartCallSite::test_softstart_generator_calls_route_with_escape`

## Cause

PR #3481 (issue #3343 P-R1..P-R4) overhauled the softstart recipe:

- `generate_design.py` no longer calls `router.route_with_escape(` in-process — it shells out to `kct route` (the board-05 production pattern). Test 3 asserts the old call-site string and now always fails (`tests/test_router_dense_escape_softstart.py:456`).
- The baseline tests pin the OLD pre-rev-B 2-layer/10-signal-net configuration (issue #3235 era). P-R2 pin reassignment + P-R3 placement moves changed the board, so the 10/10-connected and `8 + 0 + 0` violation budgets no longer describe reality.

## Ask

Re-pin (or retire) the #3235-era baseline harness against the rev B P-R4 configuration, and update the call-site guard to assert the new `kct route` subprocess invocation instead of `router.route_with_escape(`.

Refs #3343, #3481.

Issue #3474, opened 2026-06-10
chorus-test: manufacturable (100% routed + 0 DRC at jlcpcb-tier1)


rjwalters:
## Goal

Make the chorus-test-revA board (PCM5122 DAC hat for RPi 5, 63x63mm, 4L, JLCPCB target) fully manufacturable: **100% nets routed + 0 DRC violations + 0 sync drift** at `--manufacturer jlcpcb-tier1`. Per `feedback_manufacturable_means_100pct.md`, partial-reach numbers are not shippable.

Sibling umbrellas: #3412 (board 05), #3413 (board 06), #3414 (board 07). This is the external-board flagship of the same campaign.

## Re-measured baseline (2026-06-10, HEAD d45ded4d)

Recipe (identical to Wave-9/10 for apples-to-apples): `kct route boards/external/chorus-test-revA/kicad/chorus-test-revA_v19_stripped.kicad_pcb --manufacturer jlcpcb-tier1 --backend cpp --placement-feedback --placement-feedback-budget 5 --iterations 50 --auto-fix --auto-fix-passes 2 --auto-layers --timeout 1200 --seed 42` with `PYTHONHASHSEED=0`. C++ backend confirmed built.

| Measurement | Wave-9 (06-07) | Wave-10 (06-08) | **This run (06-10)** |
|---|---|---|---|
| Strict-connected | 4/48 (8%) | 4/48 (8%) | **2/48 (4%)** |
| Partial (segments, not all pads) | 29/48 | 29/48 | **30/48** |
| Unrouted (no segments) | 15/48 | 15/48 | **16/48** |
| DRC on routed output | — | — | **98 errors / 56 warnings** |
| Auto-fix | ran, rolled back | ran, kept partial | **SKIPPED (budget exhausted)** |
| AUTOFIX_SKIPPED_BUDGET_EXHAUSTED | absent | absent | **PRESENT** |

**This is a regression vs Wave-10**, with three new behaviors:

1. **Per-net A* blowup at the head of the queue**: detailed routing timed out at net **3/48** after 505.7s — SPI_SCK alone consumed ~490s before being classified `blocked_path`. In Wave-9 the worst clock net was ~270s and SPI_SCK-class nets were 70-120s. 44 nets were never attempted in the initial pass.
2. **Grace pass burned 226s routing 0/1 starved nets (44 skipped)** (issue #3452 path, #3466 budget-cliff grace pass) — pure overhead on this board.
[...]

Issue #3441, opened 2026-06-10
router: auto-grid 0.127mm puts every 0.1mm-aligned pad off-grid (board 07); explicit --grid 0.1 regresses to 13/31 fully-connected via waypoint injection


rjwalters:
Parent: #3414 (board 07 manufacturable bar; decomposed by the #3414 recovery PR)
Related: #1603 (sub-grid escape pre-pass), #3438 (reach gap)

## Observation

Board 07's pads all sit on a 0.1mm lattice (QFN 0.8mm pitch, FFC 1.0mm, header 2.54->but placed on 0.1 multiples, BGA 1.27mm — the last is 0.635-aligned). Auto-grid selects 0.127mm:

```
UserWarning: Auto-grid: memory budget cap forces grid 0.127mm > clearance/2 (0.075mm) even at max_cells=500,000.
Routing may produce clearance violations at fine-pitch pads.
```

Consequences on the production route:
1. Every pad is off-grid -> the sub-grid escape pre-pass (#1603) synthesizes escape stubs for all of them, contributing to the corridor blockage dynamics in #3438.
2. Grid 0.127 > clearance/2 (0.075) is the documented recipe for the residual clearance DRC errors (`clearance_pad_segment`, `clearance_pad_via`, `clearance_segment_via` — 3 on the committed artifact).

Forcing `--grid 0.1` (4.19M cells, C++ backend fine) was tried 2026-06-09 and made things WORSE:

```
Grid: 1101x951x4 = 4,188,204 cells
53 pads off-grid; waypoint injection will handle pad connections
PARTIAL: Routed 13/31 signal nets ... Unrouted: 3/31 (no segments)
```

i.e., 15 nets ended PARTIALLY connected (segments exist but pads unreached) — the waypoint-injection path failed to land connections that the 0.127 off-grid path handles. Also: why are 53 pads off-grid at 0.1mm when the lattice analysis says they should align (suspect: BGA-49 1.27mm pitch pads at 0.635 offsets, and/or grid origin not anchored to board origin 100.0,100.0)?

## Suggested scope

1. Diagnose the 53 off-grid pads at grid 0.1 (origin anchoring vs true off-lattice pads).
2. Fix waypoint injection so explicit-grid runs never produce MORE partial nets than auto-grid runs on the same board.
[...]

PR #3503, opened 2026-06-11
fix(chorus): R2 — stranded-partial recovery + placement adoption (refs #3474)

rjwalters:
## Summary

Phase R2 of #3474 (chorus-test partial-vs-strict gap). **refs #3474 only — must NOT close the umbrella.**

Re-measured first (post-R1 main, all of #3478/#3488/#3491/#3498/#3500 in), then fixed the two defects the measurement exposed, plus the rescue machinery the architect's plan called for.

**Headline: chorus strict reach 14-20/51 -> 33/51 (65%, checker-verified, May-10 parity class).**

## Re-measured baseline (2026-06-11, HEAD cb7c2b8e, pinned recipe, cpp seed 42, PYTHONHASHSEED=0)

| | R1 record | This re-measure |
|---|---|---|
| Escalation ladder best | 14-18/51 | **20/51** (attempt 2) |
| Final saved | = ladder best | **15/51** — placement feedback DISCARDED the 20/51 best |
| Checker-strict on saved board | n/a | 6/51 |
| DRC (route's own validation) | 53 err | 52 err |

### 1. Placement feedback discards a better input (`placement_feedback.py`)

`PlacementFeedbackLoop` iteration 0 calls `_clear_routes()` and re-routes from scratch; the #2840 best-snapshot only compared the loop's own iterations. On the auto-layers path the loop replaced the 20/51 escalation best with a 15/51 zero-move re-route (it had 12.9s of leftover budget). Fix: seed the #2840 snapshot with the pre-loop routes — the loop is now monotonic with respect to its input. Verified live: `Restoring best snapshot from the pre-feedback input state (#3474) (routed=55) instead of final iter (routed=16)`. 3 new unit tests.

### 2. The escalation ladder is the budget defect (recipe-side, `scripts/route_chorus.py`)

The 5-6 rung ladder gives each rung ~180-200s, so every rung re-routes the same head-of-queue clock nets from scratch and dies at net 12-25 of 51. One pinned-layers attempt holding the full budget walks the entire 51-net queue:

| recipe (t=1200, cpp seed 42) | strict |
|---|---|
| pinned recipe (ladder) | 14-20/51 |
| `--auto-layers --starting/max-layers 4` (2 rungs) | 22/51 |
| **`--layers 4 --no-auto-layers --per-net-timeout 60` (1 attempt)** | **33/51** |

### 3. Rescue machinery (`kicad_tools/router/partial_rescue.py`, new)

Generalizes board 05's step-6b loop (#3491) into a reusable module + adds batch **completion passes**. Honest negative results, measured on chorus:
- Per-net rescue: **0/6** — a net routed alone cannot negotiate with non-rippable preserved copper (relief correctly rolls back).
- Completion passes (whole unfinished cohort together vs preserved strict copper): **0 gain** on both fabrics (29->29, 18->18) — same fencing mechanism. No-progress passes are rolled back byte-for-byte.
- **Stub pruning** (the #3470 defect-2 AC): stripping unfinished nets' stranded copper is loss-free for reach and cuts non-connectivity DRC **47 -> 24**. Adopted as the script's final stage (`--keep-stubs` to opt out). 13 unit tests.

## Final artifact (R2 recipe end-to-end)

- **33/51 strict** (checker-verified; router tally agrees), 18 unfinished, 0 stranded-stub copper
- **25 non-connectivity DRC errors**, of which **4 are fixture-inherent** (present on the unrouted v21 fixture: 2x `clearance_pad_pad`, 2x `edge_clearance_pad_hole`); 21 router-introduced (auto-fix/fix-drc currently roll back on connectivity protection — Phase F material)
- DRC trend: route-validation errors (incl. connectivity) 53 -> ~43; non-connectivity 47 -> 24 via pruning

## Per-class residual diagnosis (18 nets)

| Class | Nets | Evidence | Disposition |
|---|---|---|---|
| J2/U8 escape sealing | ~12 (SCL, SDA, SPI_*, SWCLK, I2S_*, UART_TX, BOOT0, NRST, LATCH) | stuck at 1/N pads; every run logs `J2: 0/8 pads escaped (no grid point reachable: 8)`; 83.2% pads off-grid at auto-selected 0.0508mm | R2 follow-up on #3474 (escape/grid work, #3441-adjacent) |
| U5 (PCM5122) south cluster | Net-(U5-CAPM/VCOM/VNEG), MCU/DBG_LED | BLOCKED_PATH, suggestions name C36/R10/U5 moves | P1 (placement) |
| BLOCKED_PATH placement cluster | Net-(D2-A), Net-(LED4-K) | router suggests specific moves (J4/U3 north, C5/R10/U5) | P1 (placement) |
| AUDIO_R | 1 (4/6 pads — improved from 2/6; **AUDIO_L now routes strict**) | generic blocked_path, no stable move suggestion | P1 |

**P1 placement decision**: the architect's "C17/C20/J4 south" move was NOT applied — it no longer appears in any current run's suggestions (it was the Wave-9/10-era diagnosis); AUDIO_L routes strict without it and AUDIO_R's current diagnosis is congestion, not that move. Acting on the stale suggestion would be placement churn without measured backing.

## Floor re-pin

`CHORUS_V21_FLOOR` 10 -> **12** (post-PF-fix pinned-recipe measurements 14/15/17/18/20; floor sits below the weakest with load-modulation headroom). Full R2 measurement record + recipe comparison documented in `tests/test_chorus_reach_floor_3237.py` with a new fast doc-test.

| Re-measure before fixing | OK | Full pinned-recipe run at HEAD cb7c2b8e (table above) |
| Strict reach materially up from 14-18 | OK | 33/51 checker-verified (R2 recipe); pinned-recipe path no longer discards its best (PF fix) |
| Honest per-net diagnosis for residual | OK | Per-class table above + documented in floor-test module |
| DRC trend down from 53 | OK | non-connectivity 47 -> 24 (pruning); 4 of remaining 25 fixture-inherent |
| Zero stranded-stub copper in partial outputs | OK | Stage-4 prune; 196 blocks removed from final artifact |
| Floor re-pinned honestly | OK | 10 -> 12 with measurement record; fast doc-test pins consistency |
| Boards 02-07 + softstart untouched/green | OK | Fast suites green: 02/03/04/05/06/07 + softstart (two PRE-EXISTING main failures, reproduced on clean main: #3492 softstart call-site, #3502 board-05 stdout parsing — filed) |

## Test Plan

- `tests/test_placement_feedback.py` (58 passed; 3 new monotonicity tests)
- `tests/router/test_partial_rescue.py` (13 new tests)
- `tests/test_chorus_reach_floor_3237.py` (8 passed; new R2 doc-test)
- `tests/test_route_cmd_placement_feedback.py`, `tests/test_chorus_test_placement_feedback.py` (66 passed, 1 skipped)
- Board suites: 02 (incl. demo recipe), 03, 04, 05, 06, 07, softstart — green apart from pre-existing #3492/#3502
- End-to-end: `scripts/route_chorus.py` on v21_stripped -> 33/51 + pruned artifact (logs preserved in the run records above)

refs #3474

### DIFF
--- a/scripts/route_chorus.py
+++ b/scripts/route_chorus.py
@@ -2,18 +2,23 @@
 """Route chorus-test-revA with the pinned recipe + partial-net rescue (Issue #3474).
 
 This is the canonical chorus-test-revA routing recipe runner for Phase
-R2 of issue #3474.  It is two stages:
+R2 of issue #3474.  It is three stages:
 
 1. **Main pass** -- the pinned recipe, byte-identical to the one in
    ``tests/test_chorus_reach_floor_3237.py`` and the issue body
    (jlcpcb-tier1, cpp backend, seed 42, ``--timeout 1200``,
    auto-layers, auto-fix, placement feedback, ``PYTHONHASHSEED=0``).
-2. **Rescue loop** (``kicad_tools.router.partial_rescue``) -- each net
+2. **Completion passes** (``complete_unfinished_nets``) -- every net
    left partially routed (1/N-pad stranding, the #3470-class signature)
-   or unrouted (budget-starved tail) is re-routed ALONE against the
-   committed copper of every other net.  Stranded stubs are stripped
-   first so they neither poison the rescue A* nor remain as
-   overlapping-copper DRC liabilities.
+   or unrouted (budget-starved tail) is stripped of stranded stubs and
+   re-routed TOGETHER against the strict nets' preserved copper, so the
+   unfinished cohort can still negotiate among itself.  Repeats while
+   the unfinished count drops (max 3 passes).
+3. **Per-net rescue** (``rescue_partial_nets``) -- only when <= 10
+   residual nets remain; each is routed alone against all committed
+   copper.  Measured 2026-06-11: running this stage on a large cohort
+   is ineffective on chorus (0/6 -- a solo net cannot negotiate with
+   non-rippable preserved copper), hence the completion passes first.
 
 The chorus fixture lives in the external chorus repo, surfaced through
 the ``boards/external/chorus-test-revA`` symlink.  NOTE: that symlink
@@ -44,6 +49,7 @@ sys.path.insert(0, str(REPO_ROOT / "src"))
 
 from kicad_tools.router.partial_rescue import (  # noqa: E402
     RescueConfig,
+    complete_unfinished_nets,
     partially_connected_signal_nets,
     rescue_partial_nets,
 )
@@ -202,13 +208,25 @@ def main() -> int:
         excluded_nets=CHORUS_POUR_NETS,
         micro_via_in_pad_fallback=True,
     )
-    targets = partially_connected_signal_nets(
+
+    # Stage 2: batch completion passes.  All unfinished nets route
+    # TOGETHER against the strict nets' preserved copper, so they can
+    # still negotiate among themselves.  Measured (2026-06-11): the
+    # single-net rescue loop alone lands 0/6 on chorus because a solo
+    # net cannot negotiate with non-rippable preserved copper.
+    complete_unfinished_nets(args.output, config, max_passes=3)
+
+    # Stage 3: per-net rescue for a small residual only.  Worth one
+    # 300s stage per net when few remain; pointless (and slow) for a
+    # large cohort -- the completion pass is the bulk mechanism.
+    residual = partially_connected_signal_nets(
         args.output,
         manufacturer=config.manufacturer,
         excluded_nets=CHORUS_POUR_NETS,
         include_unrouted=args.rescue_unrouted,
     )
-    rescue_partial_nets(args.output, config, nets=targets)
+    if 0 < len(residual) <= 10:
+        rescue_partial_nets(args.output, config, nets=residual)
 
     unfinished, drc_errors = report(args.output)
     print(
--- a/scripts/route_chorus.py
+++ b/scripts/route_chorus.py
@@ -1,0 +1,223 @@
+#!/usr/bin/env python3
+"""Route chorus-test-revA with the pinned recipe + partial-net rescue (Issue #3474).
+
+This is the canonical chorus-test-revA routing recipe runner for Phase
+R2 of issue #3474.  It is two stages:
+
+1. **Main pass** -- the pinned recipe, byte-identical to the one in
+   ``tests/test_chorus_reach_floor_3237.py`` and the issue body
+   (jlcpcb-tier1, cpp backend, seed 42, ``--timeout 1200``,
+   auto-layers, auto-fix, placement feedback, ``PYTHONHASHSEED=0``).
+2. **Rescue loop** (``kicad_tools.router.partial_rescue``) -- each net
+   left partially routed (1/N-pad stranding, the #3470-class signature)
+   or unrouted (budget-starved tail) is re-routed ALONE against the
+   committed copper of every other net.  Stranded stubs are stripped
+   first so they neither poison the rescue A* nor remain as
+   overlapping-copper DRC liabilities.
+
+The chorus fixture lives in the external chorus repo, surfaced through
+the ``boards/external/chorus-test-revA`` symlink.  NOTE: that symlink
+is relative (``../../../chorus/...``) and only resolves from the main
+repository checkout, not from ``.loom/worktrees/`` -- pass ``--pcb``
+explicitly when running from a worktree.
+
+Usage::
+
+    PYTHONHASHSEED=0 uv run python scripts/route_chorus.py \
+        [--pcb PATH] [--output PATH] [--skip-main-pass] [--seed 42]
+
+Exit code 0 when every signal net is fully routed (the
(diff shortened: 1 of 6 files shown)


rjwalters:
✅ **Approved!**

Reviewed against the four R2 judging axes; verified in the issue-3474 worktree at head 8a428c61.

## 1. PF-loop fix blast radius (placement_feedback.py)

- The fix is minimal and correct: it seeds the existing #2840 best-snapshot with the pre-loop routes (`best_iteration = -1` sentinel) using the same routed-count formula the loop already uses (`len(nets) - 1 - len(failed)`), and the restore path is unchanged. The loop is now monotonic w.r.t. its input: it can only return >= the routed count it was handed.
- Blast radius verified: exactly ONE production call site (`route_cmd.py:9023` -> `_run_placement_feedback` -> `Autorouter.route_with_placement_feedback`, core.py:12524), engaged only when `--placement-feedback` is set AND the initial pass left failed nets. The change is strictly improvement-only — boards where any loop iteration beats the input are unaffected.
- Canaries: full CI Test suite pass, plus both routing-regression jobs (boards 06 diffpair, 07 matchgroup) pass on this branch. No committed artifact is modified by this PR.
- Test coverage is the right three cases: regression-to-input restored (with deep-copy non-aliasing assert), verbose restore log names the pre-feedback state, and an iteration that beats the input still wins.
- One pre-existing hazard noted (NOT introduced here): the #2840 restore — now including the pre-loop seed — restores routes but not the placement they were computed under, so a restore after applied moves pairs stale routes with moved components (`placement_diff` records moves the routes don't know about). Measured chorus case moved 0 components, so safe here. Filed **#3504** to track snapshot+placement co-restoration.

## 2. Recipe change (single-attempt full-budget) and floor coherence

- Yes, the auto-layers ladder is OFF for chorus in the R2 recipe — per-board recipe decision, and it is documented exactly where it should be: `scripts/route_chorus.py` `R2_RECIPE_FLAGS` comment block explains the budget-shape defect (5-6 rungs of ~180-200s each re-routing the same head-of-queue nets) with the measured ladder/2-rung/single-attempt table (14-20 / 22 / 33).
- **Floor reconciliation (question 4): coherent.** `tests/test_chorus_reach_floor_3237.py::test_chorus_reach_v21_floor` pins the SAME recipe it measures — its argv matches `PINNED_RECIPE_FLAGS` (ladder recipe) flag-for-flag, and `CHORUS_V21_FLOOR = 12` is pinned against post-PF-fix pinned-recipe measurements (14/15/17/18/20), not against the R2 recipe's 33. The docstring explicitly states the test runs the pinned ladder recipe for apples-to-apples history while the R2 recipe (33/51) lives in `route_chorus.py`. The new fast doc-test (`test_r2_remeasurement_documented`) pins the constants' internal consistency (floor < weakest measurement, 33 > discarded ladder best, 33/51 >= May-10 ratio, 33 < 51 so the manufacturable bar stays honest).

## 3. partial_rescue module

- New shared surface consumed ONLY by `scripts/route_chorus.py` and its own tests. Board 05's step-6b still uses its own local `rescue_partial_nets` (`boards/05-bldc-motor-controller/design.py:3040`) and is untouched by this PR — zero behavior change for board 05's committed artifact. Filed **#3505** to track the parity-verified migration so the duplicate loop doesn't drift.
- Rollback invariants are well-tested: failed rescue strips target copper (no stranded stubs), no-progress completion pass restores the pre-pass board byte-for-byte, side files cleaned up. Honest negative results (0/6 per-net rescue, 0-gain completion on chorus) are recorded in the module/script docstrings rather than hidden.

## 4. closingIssuesReferences

- EMPTY, verified via `gh pr view --json closingIssuesReferences` — body uses `refs #3474` only, so the umbrella will NOT auto-close. Correct for a phase PR.

## Test Execution

1. `tests/test_placement_feedback.py` + `tests/router/test_partial_rescue.py` + `tests/test_chorus_reach_floor_3237.py` — ✅ Executed locally: 79 passed, 1 skipped (env-gated slow floor test)
2. `tests/test_route_cmd_placement_feedback.py`, `tests/test_chorus_test_placement_feedback.py` — ✅ Covered by CI Test job (pass, 9m18s full suite)
3. Board suites 02-07 + softstart — ✅ CI: boards 06/07 routing-regression jobs pass; ⏭️ full board suites not re-run locally (long-running; PR documents pre-existing #3492/#3502 reproduced on clean main)
4. End-to-end `scripts/route_chorus.py` (33/51) — ⏭️ Skipped: ~25+ min routing run; cross-checked against the measurement record in the floor-test module instead

## CI / Merge state

- All 8 checks pass (`gh pr checks`); `mergeStateStatus: CLEAN`.

Minor non-blocking nits (fine to leave for the migration follow-up):
- `route_chorus.py --rescue-unrouted` is `action="store_true", default=True` — the flag is a no-op (cannot be disabled).
- `partially_connected_signal_nets` returns `[]` on checker JSON parse failure, which `complete_unfinished_nets` reads as "all connected — done"; a loud warning would distinguish checker breakage from convergence.

Follow-ups filed: #3504 (snapshot/placement co-restoration), #3505 (board 05 migration to shared module).

PR #3500, opened 2026-06-11
fix(boards/06): MIPI_RST escape + USB2_D+ correction-before-demotion + J1 hotspot (refs #3413)

rjwalters:
## Summary

Phases 1-3 of #3413 (board 06 diffpair-test reach). **Refs #3413 — do NOT auto-close** (phases 4-6 remain for a separate builder).

CI-recipe reach (`check_diffpair_coverage.py boards/06-diffpair-test --seed 42`, `PYTHONHASHSEED=42`, local M-series):

| Stage | Reach | Bare gate errors | Demotion |
|---|---|---|---|
| Baseline (main @ 05541c7d) | 18/21 | 9 | USB2_D+ stripped every sweep |
| Phase 1 (overflow-silent rescue) | 19/21 | 18 (transient) | still fires |
| Phases 1+2+3 (final) | **20/21** | **13** | **none** |

Gate verdict: green (13 <= allowlist 21); all 3 diff-pair rule families exercised.

### Phase 1 — MIPI_RST (0 segments ever -> routed every run)
Root cause was NOT static geometry (routes fine on an empty board in 5.6s): the 0.375mm impedance-resolved sideband trace gets sealed at U4 by earlier-routed MIPI pair copper, and its pin-access failure emits **no overflow** — so the negotiated loop's entire targeted-fallback/relief chain (gated `overflow == 0`) never opened while the J1 USB3 cluster held overflow at 2-3. New **overflow-silent hard-failure rescue**: nets fully unrouted >= 2 consecutive iterations engage targeted rip-up + relief rescue even while overflow > 0 (both transactional, attempt-capped). Relief rescues additionally get a hard wall deadline (measured 342s composite overrun on a 240s budget).

### Phase 2 — USB2_D+ (routed every sweep, demoted every sweep)
Two structural fixes plus the ordering fix the architect specified:
- `create_intra_ic_routes` validated the **direct** tie only against same-package pads, never committed copper — the USB-C reversible-pinout X-crossover (D- ties B7->A7, then D+ ties A6->B6 straight across it) emitted physically overlapping copper that #3433 then demoted. Direct ties now check `obstacle_segments`; perimeter wraps are capped at 4x direct distance (the legal J1 wrap measured **21.8mm for a 2.24mm tie** and consumed the J1 fan-out fabric — measured reach regression); past the cap the tie defers to the main A*, which resolves the crossover.
[...]

Issue #3482, opened 2026-06-10
NetStatusAnalyzer reports pour-net pads connected via zone BOUNDARY even when the zone has zero filled polygons


rjwalters:
## Bug Description

While evaluating PR #3481 (softstart rev B, issue #3343), the recipe's step-12 per-pad pour verification and `kct check`'s connectivity rule both reported AC_NEUTRAL and ISENSE_POS as `complete` (15/15 pour nets connected) — but both zones have **zero filled polygons** in the committed artifact, and the nets have zero segments and zero vias. On the manufactured board those nets are open circuits.

## Root cause

`src/kicad_tools/analysis/net_status.py` (`_build_connectivity_graph`) intentionally checks pads/vias against the zone **boundary polygon** in addition to filled polygons (Issue #479 fix for thermal-relief cutouts). When a zone's fill produces no polygons at all (e.g. fully shadowed by a higher-priority overlapping zone, or carved away entirely), the boundary check still marks every pad inside the outline as zone-connected — a false `complete`.

The DRC `zone_unfilled` WARNING fires for these zones, but the connectivity ERROR does not, so automated gates that only look at errors pass a physically open net.

## Reproduction

`uv run kct check boards/external/softstart/output/softstart_routed.kicad_pcb --mfr jlcpcb-tier1` on PR #3481's artifact: AC_NEUTRAL / ISENSE_POS get `zone_unfilled` warnings only; no connectivity error despite nets having no copper beyond pads.

## Suggested fix

The boundary-polygon connection heuristic should only apply when the zone has at least one filled polygon (its purpose is thermal-relief cutouts INSIDE a filled zone). A zone with fill enabled and zero filled polygons should contribute no connectivity. Alternatively (or additionally), escalate `zone_unfilled` to an error when the zone's net has pads whose only connection path is that zone.

## Impact

- Severity: High — automated manufacturability gates (route_pcb step 12, `kct check` connectivity) can pass boards with open power nets.
- Discovered during Judge evaluation of PR #3481.

PR #3506, opened 2026-06-11
fix(boards/06): plane connectivity + gate lock + J1 fan-out — completing #3413

rjwalters:
## Summary

Phases 4-6 of #3413 (board 06 diffpair-test). **Refs #3413 — do NOT auto-close** (honest residual documented below; the issue's 0-DRC clause is not fully met).

Seed-42 CI-recipe results (`PYTHONHASHSEED=42 check_diffpair_coverage.py boards/06-diffpair-test --seed 42`):

| Measure | Before (PR #3500) | After |
|---|---|---|
| Signal reach | 20/21 | **21/21** (asserted by the gate now) |
| Pour connectivity (copper-union audit) | GND 48/122 stranded; +1V8/+1V2 zero-fill zones | **all 5 pour nets ONE copper component** (GND 122/122) |
| Bare-gate errors (re-route) | 13 | 14-20 band (wall-clock classifier variance, pre-existing) |
| Committed-artifact strict gate (sidecar) | 17 vs floor 21 | 33 vs floor 33 (composition below) |
| Gate verdict | green (no reach assertion!) | green, with reach + rule-coverage + error-count asserted |

### Phase 6 — USB3_RX1- (the 21st net): root cause was width, not geometry surgery
PR #3500 diagnosed a "hard obstacle seal" at the J1 fan-out. Measured root cause: the impedance resolver's **loosely-coupled branch width (0.475 mm)** combined with the recipe's tight 0.10 mm gap — physically inconsistent (62 Ω measured vs the 90 Ω target, 31% off) and geometrically sealing J1's 0.7 mm inter-pad channels (0.475 + 2x0.15 = 0.775 mm). Even on an EMPTY board USB3_RX1- needed 27 s vs the 30 s per-net budget.

Fix: re-solve diff widths on the **tightly-coupled branch** (bisection on `CoupledLines.edge_coupled_microstrip` at the recipe's intra-pair gap): USB3 0.275 / USB2 0.250 / PCIe+MIPI 0.225 mm — all within 2.3% of their targets (vs 31% off before). 21/21 connects by iteration 2 (~40 s). Negotiated budget 240→360 s re-measured (the phase-3 360 s rejection premise — partial-RX1- snapshots — no longer exists at these widths). Evaluated alternatives from the issue: footprint re-spin and placement micro-moves both unnecessary once the width is physical.

### Phase 4 — plane connectivity (softstart PR #3481 recipe pattern, hardened)
- Recipe-local zone layer plan: GND→In1.Cu, VBUS_USB + +1V2 → split In2.Cu, +3V3/+1V8 → B.Cu (the allocator's F.Cu default fragmented into dead slivers under 270 signal segments; an earlier +1V2-on-B.Cu attempt enclaved the BGA's +3V3 corners — both measured and documented in-line).
- Zones are now **FILLED** (the committed artifact historically had boundary-only zones — zero `filled_polygon`s; every pour was a #3482 boundary-test illusion).
- Iterative repair↔refill loop (≤3 rounds): shapely copper-union audit drives offset vias + F.Cu stubs + same-layer island bridges + via-hop bridges (crossing foreign carves on another layer) + long lane-escape stubs (U2.F1 was boxed in by TX2±/RX2+ on three layers). All placements validated against ALL copper at ≥0.15 mm, **no via-in-pad** (jlcpcb standard tier).
- `_audit_pour_nets` (copper-union, #3482-immune) is also a committed-artifact regression test (`TestPourCopperUnionAudit`); shapely added to the dev extra.

### Phase 5 — gate lock
- `check_diffpair_coverage.py` now asserts **reach**: NetStatusAnalyzer completeness over the signal universe derived from the board module's new `POUR_NETS` / `REQUIRED_SIGNAL_REACH = 21` contract (the gate was previously green at ANY reach — the 18/21 regression shape).
- 22-vs-21 AC reconciled: 21 signal nets is canonical (the issue's 22 counted VBUS_USB, a pour net) — documented at the contract declaration.
- Allowlist 21 → 33 with full composition + exit clauses. This is the #3145 engagement-over-silence shape, NOT a quality loosening: the floor-21 artifact reached 18/21 and its unrouted nets could not contribute skew/continuity/clearance errors. Reach can no longer hide inside the error budget (hard-asserted), and the strict committed-artifact count is deterministic at exactly 33.

### Artifact refreshed (deferred from #3500) + impedance neck-down exemption
- `validate/rules/impedance.py`: sub-1 mm segments exempt (λ/10-insignificant neck-down tapers; retires the 3 spurious MIPI_RST errors #3500 measured). Also fixes a latent `_collect_traces` bug: it read nonexistent `x1/y1` attributes so every endpoint recorded as (0,0) — with the exemption that would have skipped EVERYTHING (2 existing tests caught it; both pass with the fix + 2 new regression tests).
- Committed artifact: 21/21 routed, pours audited-connected, strict sidecar 33 blocking + 2 advisory connectivity (GND/+1V2 — **analyzer false positives**: NetStatusAnalyzer cannot follow pad→stub→via→plane-fill chains, the inverse face of #3482; the copper-union audit proves both nets are one component).
- Artifact-pinned tests updated: strict-gate 17→33, advisory 4→2, width canaries 0.475→{0.275, 0.25, 0.225, 0.375} + a 0.475-must-be-absent tripwire, layer-ID pins made name-based (the zone-fill round trip normalises copper IDs to modern KiCad numbering).

### Honest residual (why refs-only)
- **17 of the 33 blocking errors are ONE optimizer/nudge-introduced USB3_RX1+/RX1- overlap cluster** at the J1 fan-out: the post-route copper is overlap-free (verified in-recipe); the optimizer + nudge passes mutate `router.routes` **without re-marking the grid**, and their collision checking misses the pair overlap. The recipe's new transactional solo-re-route repair detects the overlap but must roll back (solo A* on the stale grid re-lands on the partner / foreign copper). Proper fix is router infrastructure (grid-state resync for post-route passes) — follow-up issue material.
- 8 skew + 8 continuity errors are the single-ended-fallback measurements; coupled convergence remains 0/9 (the issue's phase-6-quality successor).
- 2 advisory connectivity = analyzer-model false positives (above).
- Re-route bare-gate counts vary 14-20 across local seed-42 runs (pre-existing wall-clock classifier variance, documented in PR #3500 and the allowlist).

### Cross-board checks
- Board 06 suite: 39 passed (incl. new copper-union audit + reach-contract tests).
- Gate/impedance suites: 80 passed (`test_check_diffpair_coverage`, `test_impedance_default_gating`, `test_cli_check_impedance`, `test_impedance_validator_bridge`).
- Cross-board (not slow): `test_board_05_drc_allowlist`, `test_board_03_regression`, `test_board_04_mfr_gate`, `test_drc_violation_type_validate`, `test_predictive_drc` — 112 passed, 1 skipped; net-status/connectivity/validation-model suites 109 passed. Board 03/04/05 committed artifacts untouched.
- CI diffpair job: budget rationale updated in ci.yml; 20-min ceiling retained (routing is faster despite the 360 s in-loop budget; zone/stitch/repair adds ~1-2 min).

## Test plan
- [x] Seed-42 gate: 21/21 reach asserted, errors within floor, all 3 diff-pair rules exercised, pour audit PASS, exit 0
- [x] Strict committed-artifact gate: 33 = floor 33 (deterministic)
- [x] Copper-union audit: all 5 pour nets one component (recipe step 11 + TestPourCopperUnionAudit)
- [x] No new lint errors on touched files
- [ ] CI: watch the diffpair re-route job (2-core wall-clock) + routed-pcb-drc-check


rjwalters:
## Doctor: Diff-Pair gate failure diagnosed and fixed — all 8 checks now green

**Failing run:** [27341578234](https://github.com/rjwalters/kicad-tools/actions/runs/27341578234) — `The job has exceeded the maximum execution time of 20m0s`.

### Diagnosis: job-ceiling arithmetic, not a routing or gate regression

The run was killed by the runner, not by any gate assertion:

- **Reach was fine** — the negotiated loop banked and restored `connected=21/21 routed=21/21 overflow=2` (iter 2) before the kill. The new 21/21 reach assertion was never violated.
- **Errors were fine** — DRC never even ran; nothing exceeded the 33 floor (the passing rerun measured **16 errors**).
- **The ceiling was the bug.** Timeline of the killed run: setup ~0:45 + coupled pre-phase 6:00 (9 pairs x 60s per-pair budget, all deferred) + negotiated loop 6:05 (the new 360s budget) = ~12:45 before post-route work even starts. The phases-4-6 post-route pipeline (zone gen + first fill + stitch + up to 3 rounds of pour repair / re-fill / copper-union audit + DRC + mfg export + gate checks) then needed ~8 min, vs ~3.7 min for the pre-phase-4 tail the 20-min ceiling (#3477) was calibrated against (main runs complete in ~14.4 min). The workflow comment's "total wall does not grow... ceiling retained" prediction was simply wrong on the 2-core runner.

The log was also silent for 7+ min after "Restoring iteration 2 state" because stdout to the Actions pipe is block-buffered and all post-route subprocesses use `capture_output=True` — the job was working, not hung, but invisible.

### Fix (commit 4902532a, workflow-only — board 06's committed 21/21 artifact untouched)

1. `timeout-minutes: 20 -> 30` for the diff-pair job, with the CI-empirical arithmetic documented in place (same convention as the error-floor headroom). All in-loop work is internally budgeted (per-pair 60s, negotiated 360s, repair loop capped at 3 rounds), so 30 min adds headroom without masking a genuine hang.
2. `PYTHONUNBUFFERED: "1"` on the gate step so a future timeout is attributable to a specific pipeline step instead of a silent tail.

### Verification: [run 27343006197](https://github.com/rjwalters/kicad-tools/actions/runs/27343006197) — all 8 checks pass

Diff-pair gate completed in **20m42s** (would have missed the old ceiling by 42s — confirming the diagnosis exactly) with:

```
[diffpair-coverage] Signal reach: 21/21 (required: 21)
[diffpair-coverage] OK: all 3 diff-pair rules exercised.
[diffpair-coverage] OK: 16 errors (within allowlist max 33).
```

### Note for the fresh judge

The recipe's *informational* copper-union pour audit printed `POUR CONNECTIVITY: FAIL` on the CI re-route (the 3-round repair loop converged GND from 79 -> 12 disjoint groups but not to 1 on the slower runner). The gate does not assert on this — it asserts reach/rules/error-count — and the *committed* artifact carries the fully-repaired pours. Flagging it for transparency since the PR's headline includes plane connectivity.

rjwalters:
## ✅ Approved! (Judge review — full fresh pass)

CI 8/8 green (diff-pair gate 20m42s under the new 30-min ceiling), `mergeStateStatus: CLEAN`, `closingIssuesReferences` empty as intended (refs-only, no auto-close of #3413).

### What was verified

**Code review**
- `scripts/ci/check_diffpair_coverage.py` reach assertion: correct — measures NetStatusAnalyzer completeness over the signal universe (`POUR_NETS` excluded), hard-fails below `REQUIRED_SIGNAL_REACH`. CI log confirms `Signal reach: 21/21 (required: 21)` on the re-route. The 18/21-shaped silent regression is now structurally impossible.
- `scripts/ci/net_class_map_resolver.py`: clean refactor — `load_board_recipe_module()` extracted, `build_net_class_map_for_board()` delegates; `sys.path` save/restore preserved.
- `src/kicad_tools/validate/rules/impedance.py`: the `_collect_traces` fix is correct — `schema.pcb.Segment` carries `start`/`end` tuples (verified at `src/kicad_tools/schema/pcb.py:833`), so the historical `x1/y1` getattr recorded (0,0) for every endpoint. Without the fix, the new sub-1mm neck-down exemption would have zero-length-skipped EVERY segment and silently disabled the rule. The new `TestNeckDownTaperExemption` fixture (long mis-sized trunk still fires, short tapers exempt) pins exactly this failure mode.
- Width re-solve (`boards/06-diffpair-test/generate_design.py`): bisection on `CoupledLines.edge_coupled_microstrip` is mathematically sound (Zdiff monotone-decreasing in width, branch directions correct, 0.025mm rounding, min-width clamp). The 0.475mm → 0.275/0.250/0.225mm root-cause story checks out geometrically (0.575mm < J1's 0.7mm channel vs 0.775mm before).
- Pour audit/repair machinery: **recipe-local** (board 06 `generate_design.py`) — zero blast radius on other boards' code paths. Note: `_audit_pour_nets` is now duplicated between softstart and board 06 recipes; third instance should trigger extraction to a shared module (non-blocking).

**Test execution (local, PR head 4902532a)**
1. Board 06 suite — ✅ 39 passed with `--extra dev` (incl. `TestPourCopperUnionAudit` on the committed artifact and the strict-gate 33==floor pin). Caveat verified: without dev extras the shapely audit *skips* via `importorskip` — CI's Test job uses `uv sync --extra dev`, so the backstop is live in CI.
2. Gate + impedance gating suites — ✅ 64 passed.
3. Seed-42 CI gate — ✅ executed by the green diffpair CI job (reach asserted, 16 errors ≤ 33 floor, all 3 rules exercised).
4. Width canaries — ✅ pass, including the 0.475-must-be-absent tripwire.

### Allowlist 21 → 33: adjudicated as a legitimate engagement raise
- Full composition + both exit clauses are in `.github/routed-drc-tolerance.yml` (8+8 skew/continuity = exit clause a; 17-error RX1 overlap cluster = exit clause b; floor tightens toward ~2 after both).
- The framing holds: the floor-21 artifact was 18/21 and its unrouted nets could not contribute errors; reach is now hard-asserted, so errors can no longer hide reach loss.
- Tracking: the 17-cluster has **#3507** (optimizer/nudge grid-staleness). The 8+8 coupled-convergence block had **no dedicated issue** — filed **#3508** during this review (required for approval; exit clause a now has a real tracker).

### Pour-audit CI trap: confirmed real, bounded, now tracked
The CI re-route's pour repair did NOT converge (r3: GND 12 disjoint groups, persistent zero-fill zones on GND/+1V8/+1V2, `POUR CONNECTIVITY: FAIL`) while the job passed — the gate doesn't assert pours. The committed artifact is protected by `TestPourCopperUnionAudit` (a broken refresh cannot merge), so this doesn't block. But it leaves repair-machinery regressions invisible in CI and makes every future artifact refresh a fight. Filed **#3509** (diagnose persistent zero-fill on re-route, raise the 3-round cap — convergence trend 79→45→12 suggests ~5-6 rounds — then promote the pour audit to asserted in the gate).

### #3413 disposition recommendation
**Close #3413 when this PR merges**, residuals tracked separately. Rationale: the issue's core subject (reach gap, 41% → diagnosed escape/congestion) is fully resolved — 21/21 (22-vs-21 reconciliation documented at the contract declaration) and CI-asserted. The remaining work is DRC quality with well-scoped, independently-tracked successors: #3507 (17 errors), #3508 (16 errors), #3509 (pour-repair CI convergence). Keeping #3413 open as an umbrella would duplicate tracking against a stale diagnosis. Explicit caveat per the manufacturability bar: board 06 is NOT manufacturable-complete (100% + 0 DRC) until #3507/#3508 retire their blocks — both issues state that bar.

Great work on the root-cause discipline here — replacing #3500's "hard obstacle seal" narrative with the measured width-physics root cause, and replacing boundary-test pour illusions with a geometric audit, are exactly the right shapes.

Issue #3433, opened 2026-06-09
board 04: SWCLK/SWO trace overlaps (-0.2mm) in raw kct route output — blocks regenerating committed routed PCB


rjwalters:
## Summary

The current production routing recipe for board 04 produces 5 blocking clearance errors between **SWCLK and SWO** in the raw `kct route` output. These are **pre-existing on main** (verified at merge-base `ccc708d1`, before any #3428 changes) and are unrelated to the escape-router work.

They matter because the routed-DRC CI gate floor for `boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb` is **0 blocking errors** — the committed PCB (from #3266-era router state, commit `f040222c`) scores 0, but any **regeneration** of the board with today's router commits these overlaps and would trip the gate. This currently blocks committing a refreshed 9/9 routed PCB after the #3428 NRST fix.

## Evidence

Recipe (production, per `generate_design.py` / #3428 curator comment):

```
kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb \
  --mfr jlcpcb-tier1 --auto-fix --auto-layers --auto-mfr-tier \
  --placement-feedback --micro-via-in-pad-fallback --seed 42 --timeout 600
```

`kct check <out> --mfr jlcpcb-tier1` blocking errors (identical on unmodified main and with the #3428 fix applied):

```
clearance_segment_segment [35.071, 19.042]  SWCLK vs SWO   actual -0.200mm  (full overlap)
clearance_segment_segment [36.113, 19.783]  SWCLK vs SWO   actual  0.106mm
clearance_segment_segment [44.235, 21.592]  SWCLK vs SWO   actual  0.117mm
clearance_segment_segment [40.638, 20.508]  SWCLK vs SWO   actual -0.200mm  (full overlap)
clearance_segment_via     [34.350, 18.300]  SWO trace vs SWCLK via  actual -0.372mm
```

Negative actuals = traces routed ON TOP of each other in the x 34-44, y 18-22 corridor (the SWD bus run from U2 south-east to J1). The DRC-nudge pass resolves 0 of them and `--auto-fix` rolls back on connectivity regression.

## Why this is a router regression, not a board problem

[...]

Issue #2933, opened 2026-05-15
router: board-02 charlieplex 182 clearance_pad_segment errors at jlcpcb tier-1 (split from #2917)


rjwalters:
## Problem

Board 02 (charlieplex-led) routes 10/10 nets at 100% connectivity, but produces **182 `clearance_pad_segment` DRC errors** at jlcpcb tier-1 (`trace = clearance = 0.127 mm`) on the default 0.1mm router grid.

**Split from #2917** (which combined this with a fleet-status reporting bug; the fleet-status bug is now tracked separately).

## Curator-verified geometry

- DIP-8 U1 pads at 7.62mm / 2.54mm pitch produce coordinates (±3.81, ±1.27) that don't align to the 0.1mm router grid.
- Charlieplex LED grid at 8mm pitch with 1.0×1.3mm pads leaves ~5.7mm minus 2×0.127mm clearance per channel for 4-7 net fan-out.

Both router-grid resolution and board-generator snap are valid remediations.

## Possible fix shapes

1. **Finer router grid for this board** — tighter grid (0.05mm or 0.02mm) on the dense charlieplex region would reduce off-grid placement errors. Likely via adaptive-grid widening (the same family as #2910's fix).
2. **Manufacturer-tier escalation** — auto-escalate to jlcpcb-tier2 (0.1mm trace/clearance, supports via-in-pad) when tier-1 produces clearance violations beyond a threshold. Connects to `--auto-mfr-tier` infra.
3. **Board generator snap** — modify `boards/02-charlieplex-led/generate_pcb.py` to snap component placements onto the 0.1mm coarse grid. Lowest router-side risk but loses generality.
4. **Adaptive fine-grid coverage extension** — extend `identify_fine_pitch_components` (per #2910 curator findings) to cover the charlieplex topology even though DIP-8 has pitch > 1.0mm.

The right choice probably involves a combination — adaptive fine-grid extension PLUS tier escalation when the channel is geometrically infeasible.

## Acceptance criteria

- [ ] Board 02 routes to <10 `clearance_pad_segment` errors at jlcpcb tier-1 (target: 0).
- [ ] Board 02 net completion stays at 10/10.
- [ ] Boards 00, 01, 03-07 fleet-status parity (no regression).
- [ ] If `--auto-mfr-tier` is required for clean board-02 routing at tier-2, that's acceptable — but the fix must NOT relax tier-1 expectations by widening `routed-drc-tolerance.yml`.
[...]

PR #3565, opened 2026-06-12
fix(router): reject routes through static foreign-pad halos + net-aware validator skip (closes #3545)

### LINKED ISSUE #3545: Negotiated router accepts final routes through static foreign-pad halo cells (unresolved overflow) -> sub-clearance copper; validator's same-component skip masks it
## Summary

`route_all_negotiated` can finish with unresolved overflow and emit a route whose segments traverse cells that are **hard-blocked for every net** (static pad clearance halo). The in-loop clearance validator does not flag the resulting copper because of an over-broad **same-component-ref skip**, so the iteration metric reports `clearance_viol=0` and the route is accepted as "best". The emitted copper then fails exact KiCad DRC.

Found while rebasing PR #3537 (issue #3532, 45-degree quantization): the PR's dogleg emission legitimately changes occupancy marking, which deterministically re-rolled where this fixture's never-resolved `overflow=2` lands — from a benign spot onto the halo of a foreign-net thru-hole pad.

## Reproduction (deterministic, every PYTHONHASHSEED, macOS and CI-linux)

Route `tests/fixtures/routing-diagnostic.kicad_pcb` exactly as `tests/test_drc_regression.py::routed_drc_report` does (`route_all_negotiated(max_iterations=5)`, rules 0.2/0.2 trace/clearance, grid 0.2), merge, run exact `kicad-cli` DRC:

```
[clearance] Clearance violation ( clearance 0.2000 mm; actual 0.1270 mm)
  at (7.000, 11.400) mm / at (8.000, 12.000) mm
[clearance] Clearance violation ( clearance 0.2000 mm; actual 0.1270 mm)
  at (7.000, 11.600) mm / at (8.000, 12.000) mm
```

- (8.000, 12.000) is J1 pad 1 (NET1), thru-hole circle 1.7 mm.
- The violating copper is NET3's vertical run at x=7.000, y 11.0-11.6 (all segments exactly 45-aligned — NOT a dogleg artifact).
- Geometry: center distance 1.077 mm − 0.85 pad radius − 0.1 half trace = **0.127 mm < 0.200 mm**.

## Three compounding defects

1. **A* admits statically blocked cells in negotiated mode.** `grid.is_blocked_for_net(35, 57, 0, net)` is `True` for EVERY net (including NET3) at the violating cells — the standard `_add_pad_unsafe` halo (pad + clearance + trace half-width) covers them. The negotiated search still routes through them, treating the static halo as shareable congestion (the #2920 `static_blocks` fix hardened pad METAL via `is_obstacle`, but clearance-halo cells remain negotiable). A pad cannot "negotiate away", so halo overflow is unresolvable by construction — the loop oscillates (`overflow band [2, 2, 2]`, all escape strategies exhausted) and accepts `overflow=2` as best.

2. **The in-loop validator's same-component skip is net-blind.** NET3 connects to J1 pad 2, so segment-vs-pad clearance checks skip ALL of J1's pads — including foreign-net J1 pad 1 — and the metric reports `clearance_viol=0` for copper 0.127 mm from foreign pad metal. The skip exists for fine-pitch own-pad escapes; it should not exempt segments near same-component pads on a DIFFERENT net (or should at minimum be distance-scoped to the trace's own connected pad).

3. **Final acceptance has no exact-geometry backstop.** A route still crossing statically blocked foreign cells after the last iteration is emitted as success instead of being evicted/marked failed.

## A/B evidence (PR #3537 worktree)

- main's router (same fixture, same seed): 0 violations, but 8 off-angle segments and the same final `overflow=2` landing benignly elsewhere. NET3 crosses to B_CU at (9.0, 8.6) and passes the J1 area at x=6.6 (legal, 0.45 mm margin).
- #3537's router: 0 off-angle segments, identical reach (3/4, NET4 has no relief path), but NET3 stays on F_CU and lands on the x=7.0 corridor inside J1-1's halo.

Identical mechanism, different dice. Any future router PR that perturbs occupancy can re-roll this again.

## Impact

- `tests/test_drc_regression.py::TestRoutingDRCCompliance::test_routing_produces_no_drc_violations` and `TestPadClearancePreservation::test_no_clearance_violations_after_multiple_ripups` fail deterministically on #3537's branch (CI run 27379148837). They are xfail'd (non-strict) with a reference to this issue until the root fix lands.
- Production paths (`kct route`, board recipes) are partially shielded because `drc_verify_and_nudge` runs after routing and can snap a 0.127 mm gap legal — but the defect can still surface anywhere `route_all_negotiated` output is consumed raw.

## Suggested fix direction

1. Treat static blocked cells (pad halo, `blocked && !route-occupied`) as non-negotiable in the sharing A* — overflow should only ever count route-vs-route sharing.
2. Narrow the same-component validator skip to same-net pads (or distance-scope it to the trace's own connected pad / pad-exit envelope) so the iteration metric sees these violations.
3. At finalization, evict (mark failed) any route that still traverses statically blocked foreign cells instead of emitting it.

Each of these moves fleet routing baselines (board 02 determinism pin, board 06/07 CI coverage floors), so this needs a dedicated PR with re-baselining — out of scope for the #3537 doctor pass.

rjwalters:
Closes #3545

## Root cause (found by instrumented A* trace)

`RoutingGrid._unmark_segment` / `_unmark_via` (and C++ `Grid3D::unmark_segment` / `unmark_via`) **erased statically blocked pad clearance-halo cells** (`blocked=False, net=0`) whenever a ripped-up route's clearance envelope swept them. On the routing-diagnostic fixture, ripping NET1's route erased J1-1's halo at x=7.0; NET3 then routed straight through the freed corridor, the net-blind same-component carve-out masked the 0.127mm-vs-0.200mm gap in-loop, and the route shipped — the exact DRC failures in the #3545 diagnosis.

## Fixes (defense in depth)

1. **Static-blockage survival across rip-up** — lazy `_static_blocked` snapshot (Python, captured at first `mark_route`) + `GridCell.static_blocked` (C++, set by `mark_blocked`); unmark now restores `blocked=True, net=original_net` for static cells. C++ build version 10 → 11 (forces `kct build-native`).
2. **Static halos non-negotiable in the sharing A\*** — `usage_count > 0` no longer releases static cells to foreign nets (Python `_is_region_blocked` / `compute_expanded_blocked` + all five C++ sharing branches). Relief probes keep #3438 soft-crossing semantics.
3. **Net-aware same-component carve-out** — the exemption only engages where component geometry forces proximity: explicit/fine-pitch clearance relaxations (#1764), #2452-relaxed corridors, or fine-pitch components (`pitch < rules.fine_pitch_threshold`). Applied in Python `validate_segment_clearance`, the C++ `exclude_ref_hashes` gating, and io.py's `component_inherent` classification (so `drc_verify_and_nudge` now repairs what it previously filtered).
4. **Finalization backstop** (#3433/#3442 demote pattern) — `_demote_pad_clearance_violation_nets` strips nets whose committed copper violates foreign-pad clearance beyond one grid resolution (past nudge reach): demoted-to-partial, never committed.

## xfail flips

Both `test_drc_regression.py` xfail markers REMOVED; tests pass strict (verified XPASS first, then markers removed, full module 5/5). Fixture routing wall-clock dropped ~337s → ~37s — the A* was burning effort exploring corrupted-grid space.

## Synthetic tests (new `tests/test_static_halo_ripup_3545.py`, 11 tests)

- rip-up restores static halos (Python + C++ parity, segment + via), route-only cells still freed
- net-aware carve-out: foreign-net same-component pad at standard pitch rejected; explicit-override / relaxed-corridor / metal-overlap (#2933) regimes preserved
- backstop: route forced through a foreign pad halo → demoted, never committed; clean route untouched

## Fleet impact (strict gates on committed artifacts)

| Board | Gate result | Notes |
|---|---|---|
| 01 voltage-divider | 0 violations | clean |
| 02 charlieplex | 0 violations | clean |
| 03 usb-joystick | 4 via_in_pad | pre-existing class, unrelated |
| 04 stm32-devboard | 1 connectivity + 4 via_in_pad | pre-existing (matches fleet status) |
| 05 bldc | 8 connectivity + 14 via_in_pad | pre-existing |
| 06 diffpair | 2 clearance_segment_via + 2 connectivity | pre-existing class |
| 07 matchgroup | 5 connectivity + 1 diffpair_intra | pre-existing |
| softstart | 14 via_in_pad | pre-existing |

**Zero `clearance_pad_segment` violations on any committed artifact** — the #3545 defect class never shipped; no artifact-fix issues needed.

## Reach deltas / re-baselines

- **Board 02**: determinism pin re-baselined (22, 387, 24, 328.16) → (22, 397, 24, 328.16) — same routes/vias/length/reach, +10 collinear re-splits; bit-perfect across seeds 42/43/44; strict gates (22/22 routed, 0 DRC jlcpcb tier-1) pass. The issue explicitly sanctioned this re-baseline.
- **Board 06/07**: coverage-floor suites 87/87 pass.
- **Board 03**: 8 passed/1 skipped/2 setup-errors — byte-identical result on main (pre-existing).
- **Board 04/05**: `test_routes_all_nets_on_2l` (error) and `test_initial_pass_routes_enough_nets` (fail) reproduce identically on main (pre-existing, budget-sensitive).
- Fixture: 3/4 nets (NET4 unroutable by design), unchanged reach, now DRC-clean.
- One #2933-era contract test (`test_trace_in_clearance_envelope_outside_pad_metal_*`) flipped to the new net-aware contract with full rationale (its premise was the exact masking #3545 removes; board 02 evidence attached in the docstring).

## Test summary

- `tests/test_drc_regression.py` 5/5 strict
- `tests/test_static_halo_ripup_3545.py` 11/11
- `tests/router/` 500+ pass (only sanctioned baseline re-pin changed)
- pathfinder/negotiated suites 201 pass; io/clearance/nudge suites 308 pass
- `kct build-native` rebuilt at version 11

### DIFF
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -9277,6 +9277,25 @@ class Autorouter:
                 f"(now {successful_nets}/{total_nets} routed)"
             )
 
+        # Issue #3545 (safety net): never commit copper that genuinely
+        # violates FOREIGN-pad clearance beyond nudge reach.  The
+        # negotiated loop can exit holding a best snapshot whose copper
+        # crossed a static foreign-pad halo (unresolvable overflow the
+        # loop oscillated on) -- exact KiCad DRC flags it, so an
+        # unrouted net (advisory connectivity finding) is the strictly
+        # better trade.  Sub-resolution deficits are left for
+        # ``drc_verify_and_nudge`` / post-route correction, mirroring
+        # the #3433 demote-only-unrepairable philosophy.
+        pad_demoted = self._demote_pad_clearance_violation_nets(net_routes)
+        if pad_demoted:
+            successful_nets = sum(1 for routes in net_routes.values() if routes)
+            flush_print(
+                f"\n  ⚠ Demoted {len(pad_demoted)} net(s) with sub-clearance "
+                f"copper against foreign pads to unrouted: "
+                f"{sorted(pad_demoted)} "
+                f"(now {successful_nets}/{total_nets} routed)"
+            )
+
         if progress_callback is not None:
             # Issue #2597: Distinguish ``stagnated`` from ``timeout`` and
             # bare ``f"overflow={N}"`` so callers (and CI) can pick the
@@ -9874,6 +9893,78 @@ class Autorouter:
                     self.routes.remove(route)
             net_routes[net] = []
         return victims
+
+    def _demote_pad_clearance_violation_nets(
+        self,
+        net_routes: dict[int, list[Route]],
+    ) -> list[int]:
+        """Strip nets whose copper violates foreign-pad clearance beyond repair.
+
+        Issue #3545 (safety net): the negotiated loop's in-loop metrics
+        historically had no seg-vs-pad quadrant, so a best snapshot whose
+        copper crossed a static foreign-pad clearance halo was committed
+        and only exact KiCad DRC caught it.  The #3545 root fixes (static
+        halo survival across rip-up + net-aware validator carve-out)
+        prevent such routes from being constructed, but this backstop
+        guarantees the contract at finalization: any net whose committed
+        segments have a FOREIGN-pad clearance deficit larger than one
+        grid resolution (beyond ``drc_verify_and_nudge``'s plausible snap
+        reach) is demoted to unrouted instead of shipped.
+
+        Uses :meth:`RoutingGrid.worst_segment_pad_deficit` -- rect-aware
+        exact geometry with per-component clearances and the net-aware
+        same-component carve-out -- so fine-pitch escapes and relaxed
+        same-component corridors (#1764 / #2452) never trigger demotion.
+
+        Args:
+            net_routes: Mapping of net ID to committed routes (mutated:
+                demoted nets are reset to ``[]``).
+
+        Returns:
+            Sorted list of demoted net ids (empty in the common case).
+        """
+        pitches = self.component_pitches
+        nudge_reach = self.grid.resolution
+        victims: list[int] = []
+
+        for net, routes in net_routes.items():
+            if not routes:
+                continue
+            own_refs = {ref for ref, _pin in self.nets.get(net, [])}
+            worst_deficit = 0.0
+            worst_loc: tuple[float, float] | None = None
+            for route in routes:
+                for seg in route.segments:
+                    deficit, loc = self.grid.worst_segment_pad_deficit(
+                        seg,
+                        exclude_net=net,
+                        component_pitches=pitches,
+                        exclude_refs=own_refs or None,
+                    )
+                    if deficit > worst_deficit:
+                        worst_deficit = deficit
+                        worst_loc = loc
+            if worst_deficit > nudge_reach:
+                loc_str = (
+                    f" at ({worst_loc[0]:.3f}, {worst_loc[1]:.3f})"
+                    if worst_loc
+                    else ""
+                )
+                flush_print(
+                    f"    Net {net} ({self.net_names.get(net, '?')}): "
+                    f"foreign-pad clearance deficit {worst_deficit:.3f}mm"
+                    f"{loc_str} exceeds nudge reach ({nudge_reach:.3f}mm)"
+                )
+                victims.append(net)
+
+        for net in victims:
+            for route in net_routes.get(net, []):
+                self.grid.unmark_route_usage(route)
+                self.grid.unmark_route(route)
+                if route in self.routes:
+                    self.routes.remove(route)
+            net_routes[net] = []
+        return sorted(victims)
 
     def _post_route_clearance_correction(
         self,
--- a/src/kicad_tools/router/cpp/include/types.hpp
+++ b/src/kicad_tools/router/cpp/include/types.hpp
@@ -72,7 +72,18 @@ namespace router {
 // the version forces a rebuild so the post-#3309 performance fix takes
 // effect and the regression test (``tests/test_router_cpp_astar_flat_arrays_3309.py``)
 // passes its build-version guard.
-constexpr int ROUTER_CPP_BUILD_VERSION = 10;
+// Bump to 11 for Issue #3545 (static-blockage survival across rip-up).
+// ``GridCell`` gained a ``static_blocked`` flag set by ``mark_blocked``
+// (static obstacle sync) and consulted by ``unmark_segment`` /
+// ``unmark_via``: ripping up a route whose clearance envelope overlapped
+// a STATIC pad clearance halo must RESTORE the halo (blocked=true,
+// net=original_net) instead of erasing it.  Pre-#3545 .so files erase
+// the halo, letting foreign nets route through a pad's clearance band
+// and ship sub-clearance copper (routing-diagnostic fixture: NET3
+// through J1-1's halo at 0.127mm actual vs 0.200mm required).
+// ``mark_blocked`` also began recording ``original_net`` for ALL static
+// cells (previously pad-metal only) so the restore has the owner net.
+constexpr int ROUTER_CPP_BUILD_VERSION = 11;
 
 // Grid cell state
 struct GridCell {
@@ -85,6 +96,11 @@ struct GridCell {
     bool is_zone = false;
     bool pad_blocked = false;
     int32_t original_net = 0;
+    // Issue #3545: true for cells blocked by STATIC board geometry
+    // (pad metal, pad clearance halos, keepouts) registered via
+    // ``mark_blocked``.  Route marking never sets this; rip-up uses it
+    // to restore static blockage instead of freeing the cell.
+    bool static_blocked = false;
 };
 
 // A* node for priority queue
--- a/src/kicad_tools/router/cpp/src/bindings.cpp
+++ b/src/kicad_tools/router/cpp/src/bindings.cpp
@@ -32,6 +32,11 @@ NB_MODULE(router_cpp, m) {
         .def_rw("is_zone", &GridCell::is_zone)
         .def_rw("pad_blocked", &GridCell::pad_blocked)
         .def_rw("original_net", &GridCell::original_net)
+        // Issue #3545: static-blockage flag -- set by ``mark_blocked``
+        // for board geometry (pads, halos, keepouts); consulted by
+        // ``unmark_segment`` / ``unmark_via`` so rip-up restores static
+        // blockage instead of erasing it.
+        .def_rw("static_blocked", &GridCell::static_blocked)
         .def_rw("avoidance_cost", &GridCell::avoidance_cost);
 
     // DesignRules struct
--- a/src/kicad_tools/router/cpp/src/grid.cpp
+++ b/src/kicad_tools/router/cpp/src/grid.cpp
@@ -54,8 +54,17 @@ void Grid3D::mark_blocked(int x, int y, int layer, int net, bool is_obstacle,
     // grid.cpp:188).
     if (pad_blocked) {
         cell.pad_blocked = true;
-        cell.original_net = net;
     }
+    // Issue #3545: every ``mark_blocked`` call registers STATIC board
+    // geometry (pad metal, clearance halos, keepouts) -- route copper
+    // goes through ``mark_segment`` / ``mark_via`` instead.  Record the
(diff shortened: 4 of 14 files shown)


rjwalters:
❌ **Changes Requested — CI failing (3 tests in the Test job); core fix verified and sound**

The engineering in this PR is solid — the deep review found the 4-layer fix correct on both backends. But the CI Test job fails on three tests that must be fixed before approval.

## CI failures (blocking)

1. **`tests/test_manufacturer_propagation_lint.py::test_manufacturer_propagation_in_router_design_rules`** and **`test_allowlist_entries_resolve_to_actual_call_sites`** — the io.py edits added ~44 lines, shifting the `DesignRules()` fallback call sites past their pinned `_ALLOWLIST` line numbers. `router/io.py: {2691, 3292}` must be re-pinned to the new locations (the `route()` fallback is now at io.py:2729; re-locate the `route_pcb()` inner fallback similarly). Reproduced locally in the worktree (2 failed, 5 passed).

2. **`tests/router/test_board02_manufacturable_baseline.py::test_routing_output_deterministic_across_seeds`** — CI Linux measures **(22, 399, 24, 328.16)** vs the PR's pin **(22, 397, 24, 328.16)**. Cross-seed equality holds (no determinism regression) and routes/vias/length/reach match, so this is the same collinear-resplit mode the docstring documents — but the segment count now diverges **between platforms** (your macOS run: 397; CI Linux: 399). Re-pinning to 399 will fail locally on macOS, so this likely needs either (a) root-causing the platform-dependent re-split (preferred — the test exists to catch exactly this), or (b) an explicitly documented platform-tolerant assertion on the segment count while keeping routes/vias/length exact. Don't just re-pin blind.

## Review findings (everything else verified — re-review will be fast)

**1. Python/C++ rip-up restore parity — VERIFIED.** C++ `unmark_segment`/`unmark_via` restore `net=original_net` (keeping `blocked=true`) for `static_blocked` cells inside the `cell.net == net` branch; Python `_unmark_segment`/`_unmark_via` gate identically on the positional `_static_blocked` snapshot. `mark_rect_blocked` delegates to `mark_blocked` so the flag covers rect blockage. The snapshot is captured under the grid lock at first `mark_route`, after #2452 corridor relaxation (relaxed cells correctly excluded); the two non-`mark_route` `_mark_segment` call sites (post-route mutation re-mark, diffpair tails) only run after routing begins, so the lazy capture cannot be bypassed. Build version 10→11 bump present in both `types.hpp` and `cpp_backend.py` (`_REQUIRED_CPP_BUILD_VERSION = 11`) — stale .so files fall back with the build-native hint. Worktree .so verified at version 11.

**2. C++ sharing branches — ALL FIVE GATED.** `is_trace_blocked` (both branches), `is_diagonal_blocked`, `is_via_blocked_diag` (both branches) each got `cell.static_blocked && cell.net != net && !relief_mode_`. The remaining ungated `usage_count` site (line ~466) is `relief_conflict_cost`, a relief-only penalty function — correct to leave.

**3. Carve-out narrowing — verified consistent across all four enforcement points** (Python validator, `worst_segment_pad_deficit`, C++ `exclude_ref_hashes` gate, io.py `component_inherent`). The #2933 metal-overlap invariant is preserved (`clearance >= 0` guard before the carve-out). The flipped #2933 contract test is justified and the fine-pitch regime stays pinned by `test_signal_net_carve_out_still_applies`. Fresh-route evidence: board 06 diff-pair and board 07 match-group CI floors PASS on this branch (16m39s / 11m37s, fresh routes), and board 02 routes 22/22 with matching reach/vias/length — only the segment-count pin diverges (see blocking item 2).

**4. Relief-probe semantics — PRESERVED and commit-safe.** Probes see static cells (`!relief_mode_` C++; the separate `relief_mode=True` branch in `compute_expanded_blocked` Python). Commit-ability is safe by construction: any probe crossing a foreign static halo yields its owner net in `find_relief_conflict_nets` → not conflict-free → never committed directly; net-0 keepouts stay hard even in relief mode; non-rippable blockers (e.g. halo of a net with no routes) roll back the transaction.

**5. Backstop** — `_demote_pad_clearance_violation_nets` mirrors the validator geometry exactly (circular + rect branches), uses unmark_usage→unmark→routes-list removal in the established rollback order, and only fires beyond one grid resolution. Synthetic tests cover both demotion and non-demotion.

**Local test execution:** `test_drc_regression.py` 5/5 strict in 20.6s (xfail flips confirmed); new `test_static_halo_ripup_3545.py` 11/11; rect-validator suite 63 passed; scoped nudge/carve-out/ripup/relief batch 183 passed, 3 skipped.

**Non-blocking, filed as #3566:** the Python fallback router's `_is_via_blocked` and `_is_diagonal_corner_blocked` still release foreign static cells at `usage_count > 0` (pre-existing; C++ now gates these but Python does not), and neither `validate_via_clearance` (segments-only) nor the new backstop (segments-only) covers via-vs-pad. Tracked for follow-up; does not block this PR since the C++ backend is the production path.

`closingIssuesReferences` correctly resolves to #3545. Once the lint allowlist is re-pinned and the board-02 platform divergence is resolved, this is ready to approve.

rjwalters:
Doctor: review feedback addressed in 6c5b2c24 — **CI is now fully green (all 8 checks)**, including the Test job that carried the 3 failures.

**1. Manufacturer-propagation lint allowlist (2 tests)** — re-pinned the `router/io.py` entries per the test's own convention: `2691 -> 2729` (the `route()` `rules=None` fallback) and `3292 -> 3330` (the `route_pcb()` inner fallback when neither `rules` nor `pcb_rules` is supplied). Verified the new line numbers point at the actual `DesignRules()` call sites; `tests/test_manufacturer_propagation_lint.py` passes 7/7 locally.

**2. Board 02 determinism pin** — took option (b), platform-tolerant assertion, after a root-cause assessment:
- The divergence (macOS-arm64: 397 segments, CI Linux-x86_64: 399) is confined to the collinear-merge pass: `merge_collinear` gates each merge on a float collision probe of the merged candidate (`path_is_clear`), and with #3545 those probes now graze static pad-halo boundaries where libm/FMA rounding differs across platforms, flipping two borderline merges. The test's own docstring already documents this splitting mode as environment-sensitive (the stale-venv 507-segments-at-identical-328.16mm note). Root-causing a 2-segment cross-libm float divergence without a Linux repro box was not tractable "within reasonable effort," so per the review's guidance I did not chase it further.
- New assertion structure: **routes (22), vias (24), total length (328.16mm) remain exact pins**, and the **cross-seed equality assertion stays exact** — so within-platform determinism regressions and lockstep cost-function drift are still caught. Only the segment count gets a tight documented band **[394, 402]** (±3 around the two measured platform values), with a dedicated `PLATFORM NOTE (#3545)` comment in the test and an updated module-docstring baseline. A segment-count excursion outside the band still fails with a message pointing at `merge_collinear`/`path_is_clear`.
- Verified locally on macOS: determinism test passes (397, inside the band) and the full board-02 baseline file is 3/3; CI Linux (399) passes the same assertions, confirmed by the green Test job.

Returning to `loom:review-requested` for re-review.

rjwalters:
## Judge Re-Review: APPROVED

Narrow re-review of doctor commit 6c5b2c24 (core fix already passed in the prior review cycle).

**1. Commit scope** — Clean. Only `tests/test_manufacturer_propagation_lint.py` (+2/-2) and `tests/router/test_board02_manufacturable_baseline.py` touched. No src/ or artifact changes.

**2. Allowlist re-pins** — Verified against `src/kicad_tools/router/io.py` at the head commit: line 2729 is `rules = DesignRules()` (route() rules=None fallback); line 3330 is `rules = DesignRules(grid_resolution=0.1)` (route_pcb() inner fallback). Both pins point at real call sites.

**3. Band approach** — Sound. Routes (22), vias (24), and length (328.16mm) remain EXACT pins; the cross-seed equality assertion across seeds 42/43/44 remains exact, so within-platform determinism regressions still trip. Only the segment count moved to a tight documented band [394, 402] (+/-3 around measured 397 macOS / 399 Linux), with a thorough PLATFORM NOTE tracing the divergence to libm/FMA rounding in merge_collinear collision probes grazing #3545 static halo boundaries. Not a blanket loosening — the band failure message directs investigators to merge_collinear/path_is_clear before widening.

**4. CI** — 8/8 checks pass on head 6c5b2c24. mergeStateStatus: CLEAN. closingIssuesReferences: [#3545] only.

Verdict: PASS. Swapping labels to loom:pr.

PR #3585, opened 2026-06-12
fix(router): freeze DEFAULT_NET_CLASS_MAP + shrink polarity-swap fixture (closes #3524)

### LINKED ISSUE #3524: tests: diffpair tests fail under full-suite xdist — net-class state pollution + pure-Python coupled A* exceeding 300s

rjwalters:
Closes #3524

## Summary

Two diffpair tests passed serially but failed in the full `-n auto` suite. Both root causes are fixed and the `xfail(strict=False)` marks from the #3436 burn-down are removed.

### 1. Net-class state pollution (`test_autorouter_invokes_tuner_for_each_pair`)

`Autorouter`, `Router` (pathfinder), and `BlockRouter` all aliased the module-level `DEFAULT_NET_CLASS_MAP` singleton when constructed without an explicit `net_class_map`. The autorouter writes per-net overrides into `self.net_class_map`:

- `core.py:3431` — impedance synthesis (`new_nc`)
- `core.py:3740/3746` — diff-pair partner wiring (`dataclasses.replace(..., diffpair_partner=...)`) — this is the `length_critical`-class pollution site
- `core.py:4216` — matrix layer-preference overrides

Through the alias those writes leaked into the shared default map for every later router in the same xdist worker process, flipping the engagement gate's `length_critical` read for sibling tests.

Fix (defense in depth):
- `rules.py`: `DEFAULT_NET_CLASS_MAP` is now wrapped in `types.MappingProxyType` — any future in-place write raises `TypeError` immediately instead of silently polluting cross-test state.
- `core.py`, `pathfinder.py`, `block_router.py`: constructors take a `dict(DEFAULT_NET_CLASS_MAP)` copy so their per-instance writes stay local.

No code mutates `NetClassRouting` instances in place (all overrides use `dataclasses.replace` or assign fresh instances), so the dict-level copy is sufficient; `io.py:3371` already copied.

### 2. Pure-Python coupled A* timeout (`test_polarity_swap_routes_all_nets`)

The 40x20mm fixture (16mm between MCU and connector clusters) drove the pure-Python `CoupledPathfinder` past the 300s module budget under full-suite CPU contention. The fixture is scaled to 16x8mm (4mm between clusters), preserving the mirrored-polarity geometry (P/N swapped on connector row B), single connector cluster with stub absorption, and the exercised coupled-attempt -> independent-fallback code path. The test now completes in ~0.3s serially (was ~60s serially / >300s under contention).

## Validation

- Each test serially: both pass (`test_polarity_swap_routes_all_nets` call time 0.28s without coverage)
- Both files under `-n auto`: 24 passed; polarity test 11.2s under worker contention with coverage tracing
- Broad slice (`tests/router/` + `test_diffpair*` + `test_router*` + `test_main_router*`) under `-n auto`: 1964 passed, 16 skipped, 3 xfailed (other tests) in 10:48
- Full suite `-n auto` running locally; PR CI runs the same full-suite gate
- `ruff check`/`ruff format --check`: changed files clean except 24 lint findings and 3 format diffs that are byte-identical pre-existing on main (verified by running ruff against the main versions)

Refs #3436.

### DIFF
--- a/src/kicad_tools/router/block_router.py
+++ b/src/kicad_tools/router/block_router.py
@@ -85,7 +85,9 @@ class BlockRouter:
 
         self.block = block
         self.rules = rules
-        self.net_class_map = net_class_map or DEFAULT_NET_CLASS_MAP
+        # Issue #3524: copy the default map instead of aliasing the
+        # module-level singleton -- in-place writes must stay local.
+        self.net_class_map = net_class_map or dict(DEFAULT_NET_CLASS_MAP)
         self.layer_stack = layer_stack or LayerStack.two_layer()
         self.margin = margin
         self._force_python = force_python
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -761,7 +761,12 @@ class Autorouter:
                 the ``--max-search-iterations`` CLI flag.
         """
         self.rules = rules or DesignRules()
-        self.net_class_map = net_class_map or DEFAULT_NET_CLASS_MAP
+        # Issue #3524: copy the default map instead of aliasing it.  The
+        # autorouter writes per-net overrides into ``self.net_class_map``
+        # (impedance synthesis, diff-pair partner wiring, matrix layer
+        # preferences); aliasing the module-level singleton leaked those
+        # writes into every later Autorouter in the same process.
+        self.net_class_map = net_class_map or dict(DEFAULT_NET_CLASS_MAP)
         self.layer_stack = layer_stack
         self._force_python = force_python
         # Issue #2610: stored so _create_grid_and_routers can pass it to
--- a/src/kicad_tools/router/pathfinder.py
+++ b/src/kicad_tools/router/pathfinder.py
@@ -183,7 +183,9 @@ class Router:
         """
         self.grid = grid
         self.rules = rules
-        self.net_class_map = net_class_map or DEFAULT_NET_CLASS_MAP
+        # Issue #3524: copy the default map instead of aliasing the
+        # module-level singleton -- in-place writes must stay local.
+        self.net_class_map = net_class_map or dict(DEFAULT_NET_CLASS_MAP)
         self.heuristic = heuristic or DEFAULT_HEURISTIC
         self.diagonal_routing = diagonal_routing
 
--- a/src/kicad_tools/router/rules.py
+++ b/src/kicad_tools/router/rules.py
@@ -7,7 +7,9 @@ This module provides:
 - Predefined net classes for common use cases
 """
 
+from collections.abc import Mapping
 from dataclasses import dataclass, field
+from types import MappingProxyType
 from typing import Literal
 
 from .layers import Layer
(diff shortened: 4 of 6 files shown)


rjwalters:
Local full-suite validation note (`pytest tests/ -n auto --no-cov` on the dev Mac, concurrent with other agent workloads): 19715 passed, 108 skipped, 10 xfailed, 5 failed + 2 errors in 21:21. Both previously-xfailed #3524 tests passed.

Triage of the 7 local stragglers — all load/environment artifacts, none caused by this change:
- `test_board_03_regression` x2 (ERROR at setup): 600s pytest-timeout deep in Python A* heuristic under CPU contention
- `test_pathfinder_dormant_partner_perf`, board_04/05 routing throughput: timing/threshold tests sensitive to load
- `test_manifest_integrity[boards/02...]` and `test_board03_routing_baseline::test_committed_pcb_drc_state`: pass serially in this worktree (16 passed in 1.28s)

CI's full-suite `Test` job (clean Linux runner) is green, as are all 7 other checks.

rjwalters:
## Judge Review: APPROVED

Reviewed the MappingProxyType freeze of `DEFAULT_NET_CLASS_MAP` and the polarity-swap fixture shrink (closes #3524).

### Freeze audit (write-site sweep of src/ and tests/)
- All four Autorouter mutation sites (core.py:3431 impedance synthesis, 3740/3746 diff-pair partner wiring, 4216 matrix layer prefs) now write into the constructor-local `dict(DEFAULT_NET_CLASS_MAP)` copy.
- `io.py` (~3371) already copies via `net_class_map = dict(DEFAULT_NET_CLASS_MAP)` before enrichment — verified.
- `route_cmd.py` sidecar merge (`router.net_class_map.update(loaded)`, line 2477) and analog-net overrides (~2585) both write through router instances, which are now copies. No production path can hit the TypeError.
- `NegotiatedRouter` — which several tests pass the frozen proxy into directly — only reads `net_class_map`; safe.
- Remaining `net_class_map[x] = y` writes in rules.py (1366-1382) and tests are on local dicts or instance copies. No `or DEFAULT_NET_CLASS_MAP` aliases remain in src/.

### Aliasing semantics
Sub-routers within one Autorouter run still share `self.net_class_map` by reference (impedance-synthesized classes propagate within a run, as intended). Nothing passes the map between independent router instances or relied on the cross-instance leak — test_negotiated_adaptive.py's defensive `ar.net_class_map = {}` resets confirm the leak was a known nuisance, not a dependency.

### Fixture shrink
Assertion sets unchanged in both polarity tests; mirrored P/N on connector row B and the 1.5mm cluster spacing are preserved, so pair-up detection, stub absorption, and both-nets-routed checks still gate polarity-swap handling.

### Verification
- Both files serial: 24 passed (40s). With `-n 4`: 24 passed (29s). No xfails remain.
- CI: 8/8 checks pass (including full-suite Test job). mergeStateStatus: CLEAN.
- closingIssuesReferences == [#3524] only.

LGTM.

PR #3598, opened 2026-06-12
test(route): replace brittle write-site count guards with AST structural audit

### LINKED ISSUE #3445: tests: test_route_zones_preserved write-site pattern guards fail on main (route_cmd.py refactor drift)
## Summary

Two source-introspection guards in `tests/test_route_zones_preserved.py` fail on pristine main (8996f5c7):

- `TestRouteWriteSitesPattern::test_every_write_site_uses_insert_sexp_before_closing`: expects 5 PCB-write call sites in `route_cmd.py` (1 partial + 1 main + 3 escalation), finds 1
- companion guard: expects >= 6 occurrences of `_insert_sexp_before_closing(` (1 def + 5 callers), finds 3

A recent `route_cmd.py` refactor consolidated the write paths without updating these pattern-count tests. The zones may well still be preserved (the behavioral tests in the same file pass — 11 passed); only the source-shape assertions are stale. Either update the expected counts to the new write-path structure, or replace the count assertions with a behavioral check.

Found while working #3423 (board-05 U3 rotation) — unrelated to that change; fails identically on main with no local modifications.

rjwalters:
Closes #3445

## Summary

The two source-introspection guards in `tests/test_route_zones_preserved.py` pinned exact write-site counts in `route_cmd.py` (exactly 5 `write_text` PCB sites, >= 6 `_insert_sexp_before_closing` occurrences). The #2808/#2809 refactor consolidated the four terminal save sites into the central `_write_routed_pcb` helper, leaving the counts stale; PR #3525 `xfail`-marked them as a stopgap pointing at #3445. This PR delivers the real fix preferred by the issue: replace count pins with structural checks that survive refactors.

### What the guards now assert (`TestRouteWriteSitesPattern`, AST-based)

- `test_central_write_path_is_zone_preserving` — `_write_routed_pcb` exists exactly once and calls `read_text`, `_insert_sexp_before_closing`, `_validate_sexp_parentheses`, plus `fsync`/`os.replace` (atomic write, #2808).
- `test_every_pcb_write_site_routes_through_zone_preserving_helper` — discovers PCB-write call sites programmatically (any `path.write_text(<content_variable>)`; report/JSON writes pass inline expressions and are excluded) and asserts each enclosing function is `_write_routed_pcb`, calls it, or calls `_insert_sexp_before_closing` directly. Includes sanity assertions that the discovery heuristic still finds sites at all and that the central writer is among them, so the guard cannot pass vacuously.
- `test_every_pcb_write_site_reads_original_first` — replaces the old `count >= 5` pin on `pcb_path.read_text()`: each discovered PCB-writing function must perform a `read_text` or delegate to `_write_routed_pcb`.
- `test_central_writer_is_actually_used_by_save_sites` — at least one external caller of `_write_routed_pcb` exists (no exact count), guarding against the consolidation becoming dead code.

### New behavioral coverage

`TestWriteRoutedPcbPreservesZones` exercises the consolidated write path end-to-end on disk: route insertion, 4L escalation (`layer_count=4`), and empty-`route_sexp` variants all preserve the 4-zone fixture's `(net_name, layer)` signatures; the atomic-write `.tmp` sibling is cleaned up.

Stale "five write sites" docstrings updated; the two `xfail` markers removed.

## Mutation check (guard has teeth)

Temporarily appended a `_evil_bypass_write` function to `route_cmd.py` that writes PCB content without the helper: both structural audits failed naming the offending function and line, then reverted.

## Test results

```
tests/test_route_zones_preserved.py — 17 passed in 1.03s (was 11 passed, 2 xfailed)
ruff check + ruff format: clean
```

### DIFF
--- a/tests/test_route_zones_preserved.py
+++ b/tests/test_route_zones_preserved.py
@@ -25,10 +25,14 @@ It covers:
 4. **Pre-existing-zones idempotency** — running through the write path
    on a PCB that already contains zones does not duplicate or drop
    them.
-5. **Source-level audit** — all five ``output_path.write_text`` call
-   sites in ``route_cmd.py`` use the same
+5. **Source-level audit** — every PCB-write site in ``route_cmd.py``
+   (discovered structurally from the AST, not by pinned counts) routes
+   through the zone-preserving
    ``read_text + _insert_sexp_before_closing + write_text`` pattern so
    a future drift between them cannot silently regress one path.
+   (Originally this pinned exact write-site counts — "five sites" —
+   which went stale when the terminal save sites were consolidated
+   into the central ``_write_routed_pcb`` helper; see issue #3445.)
 
 Acceptance criteria for issue #2770:
 - [x] Regression test added covering ALL FIVE ``output_path.write_text``
@@ -45,6 +49,7 @@ Acceptance criteria for issue #2770:
 
 from __future__ import annotations
 
+import ast
 import re
 from pathlib import Path
 
@@ -186,7 +191,7 @@ class TestSignatureExtractor:
 
 
 class TestInsertSexpBeforeClosingPreservesZones:
-    """The core helper used by all five route write sites must not drop zones."""
+    """The core helper used by all route write sites must not drop zones."""
 
     def test_route_insertion_keeps_all_input_zones(self):
         """Inserting a route s-expression into a PCB with N zones yields a PCB with N zones."""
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge Review: APPROVED

**Verification performed**
- Test-only diff confirmed: single file `tests/test_route_zones_preserved.py` (+285/-70), no `route_cmd.py` changes.
- Ran full file in worktree: **17 passed** (was 11 + 2 xfail). `ruff check` + `ruff format --check` clean.
- **Mutation reproduced independently**: appended an `_evil_bypass_write(path.write_text(content))` to `route_cmd.py`; both `test_every_pcb_write_site_routes_through_zone_preserving_helper` and `test_every_pcb_write_site_reads_original_first` failed naming the function and line. Reverted cleanly.
- Both #3525 xfail markers removed; no other xfails in the file or elsewhere touched.
- Verified `_write_routed_pcb` (route_cmd.py:620-705) actually matches every structural assertion (read_text, `_insert_sexp_before_closing`, `_validate_sexp_parentheses`, fsync, os.replace).
- CI 8/8 green; mergeStateStatus CLEAN; closingIssuesReferences == [#3445] only.

**Heuristic assessment**
`_pcb_write_calls` matches only `<recv>.write_text(<Name>)`. Known evasion vectors: `write_bytes`, `open(...).write(...)`, expression/f-string content, keyword-arg `write_text(data=...)`. The anti-vacuity asserts (heuristic must still find sites, central writer must be among them) plus the new on-disk behavioral tests give the guard real teeth for the natural write shape — this is genuinely better than the stale count pins, not false comfort.

**Non-blocking suggestion**: extend `_pcb_write_calls` to also match `write_bytes(<Name>)` and bare-Name `.write(...)` calls, so a future bytes-based or file-handle writer is also caught. Fine as a follow-up.

PR #3600, opened 2026-06-12
fix(schema): emit top-level (locked yes), never the legacy in-attr 'locked' token

### LINKED ISSUE #3457: schema PCB save path re-emits legacy in-attr 'locked' token (KiCad 10 kicad-cli rejects it)
Found during #3410 / PR #3453.

KiCad 10's kicad-cli fails to load boards whose footprints carry `(attr smd locked)` (legacy KiCad 6 form) — 'Failed to load board', which silently breaks zone fill, DRC and gerber export. PR #3453 fixed board 03's GENERATOR to emit `(attr smd)` + `(locked yes)` and taught `Footprint.from_sexp` to parse the top-level form.

Remaining gap: `_sync_attr_node` (schema/pcb.py:685) still ADDS the legacy 'locked' token inside `(attr ...)` whenever `fp.locked` is set and the file is saved through the schema layer (e.g. `kct pcb lock-footprints`, optimize-placement flows). Any board re-saved that way becomes kicad-cli-incompatible again.

Fix: emit `(locked yes)` as a top-level child on save and stop writing the in-attr token (keep parsing both forms).

rjwalters:
Closes #3457

## Summary

KiCad 10's kicad-cli rejects boards whose footprints carry the legacy KiCad-6 in-attr `locked` token (`(attr smd locked)`) with "Failed to load board" (exit 3) — silently breaking zone fill, DRC and gerber export. `Footprint._sync_attr_node` re-emitted that token whenever `fp.locked` was set and the file was saved through the schema layer (`kct pcb lock-footprints`, optimize-placement flows), making any re-saved board kicad-cli-incompatible again after PR #3453 fixed the board-03 generator.

### Save-path fix (`src/kicad_tools/schema/pcb.py`)
- `Footprint._sync_attr_node` now emits a **top-level `(locked yes)`** footprint child and never writes the in-attr token. Existing top-level `(locked ...)` nodes are removed/re-emitted from Python state (so unlocking also works), using direct-children-only removal so pad-level `(locked yes)` nodes are untouched. Defensive filter prevents `locked` from ever being echoed back through `_attr_unknown_tokens`.
- **Round-trip migration**: `PCB._link_footprint_sexp_nodes` detects a legacy in-attr `locked` token at load time and rebuilds the attr block immediately, so even a pure load → save (no field modification) migrates the file instead of echoing the legacy token via raw-sexp passthrough.
- Parsing still accepts **both** forms (KiCad 6 files load) — unchanged from #3453.

### Other writers/readers audited (issue item 4)
- Writers of `(attr ...)` nodes: `panel/furniture.py`, `panel/cuts.py`, `footprints/__init__.py`, `library/footprint.py` — none emit the in-attr `locked` token. No other writer shares the bug.
- `drc/repair_clearance.py` locked-status **reader** only recognized the legacy in-attr form and would have gone blind after migration; it now accepts both forms (modern top-level form checked first, direct children only).

### Committed-artifact audit (issue item 3d)
`grep` of every `*.kicad_pcb` in the repo: **no committed artifact carries the legacy token**. Only board 03's artifacts contain `locked` at all, and they already use the modern `(attr smd)` + `(locked yes)` form from PR #3453. No artifact repair needed.

## Tests

- `tests/test_footprint_locked_form.py` (new, 8 tests): lock emits top-level `(locked yes)` with no in-attr token; unlock removes the node; pad-level `(locked yes)` preserved; both forms parse; legacy load → save migrates without field changes; migration preserves modeled + unknown attr tokens (`exclude_from_bom`, `allow_missing_courtyard`); modern form round-trips.
- `tests/test_kicad_cli_roundtrip.py` (+3 tests): schema-locked save loads in kicad-cli; legacy-form migration loads in kicad-cli; canary asserting kicad-cli still rejects the raw legacy form. Verified locally against **kicad-cli 10.0.1** (legacy form: exit 3 "Failed to load board"; modern form: loads, DRC report produced).
- `tests/test_pcb_lock_footprints.py::test_locked_attribute_in_sexp` updated: it previously asserted the legacy in-attr behavior this issue outlaws; it now asserts the modern form and forbids the legacy token.

Results: `test_footprint_locked_form.py` 8 passed; `test_kicad_cli_roundtrip.py` 22 passed (incl. 3 new); `test_pcb_lock_footprints.py` + `test_pcb_attr_roundtrip.py` + `test_repair_clearance.py` 125 passed; schema PCB batch (`test_pcb_create.py`, `test_pcb_move_footprint.py`, `test_pcb_modules.py`, `test_pcb_sync_netlist.py`) 238 passed.

Note: pre-existing `ruff` failure at `drc/repair_clearance.py:975` (unused `required_clearance`) and repo-wide format drift are tracked by #3464 and not touched here; the changed files I authored are ruff-clean and formatted.

### DIFF
--- a/src/kicad_tools/drc/repair_clearance.py
+++ b/src/kicad_tools/drc/repair_clearance.py
@@ -1905,15 +1905,24 @@ class ClearanceRepairer:
             fp_x = float(at_atoms[0]) if at_atoms else 0.0
             fp_y = float(at_atoms[1]) if len(at_atoms) > 1 else 0.0
 
-            # Check locked status from attr
+            # Check locked status. Modern form (KiCad 9/10) is a
+            # top-level ``(locked yes)`` child; the legacy KiCad-6 form
+            # is a ``locked`` token inside ``(attr ...)``. Accept both
+            # (issue #3457) -- direct children only, since pads can
+            # carry their own (locked yes).
             locked = False
-            attr_node = fp_node.find("attr")
-            if attr_node:
-                for i in range(len(attr_node.children)):
-                    token = attr_node.get_string(i)
-                    if token == "locked":
-                        locked = True
-                        break
+            for locked_node in fp_node.find_children("locked"):
+                if (locked_node.get_string(0) or "yes") in ("yes", "true"):
+                    locked = True
+                    break
+            if not locked:
+                attr_node = fp_node.find("attr")
+                if attr_node:
+                    for i in range(len(attr_node.children)):
+                        token = attr_node.get_string(i)
+                        if token == "locked":
+                            locked = True
+                            break
 
             # Count pads
             pad_count = len(list(fp_node.find_all("pad")))
--- a/src/kicad_tools/schema/pcb.py
+++ b/src/kicad_tools/schema/pcb.py
@@ -640,10 +640,19 @@ class Footprint:
         :meth:`__setattr__`. Removes the existing ``(attr ...)`` node
         (if any), then re-emits it from the current values of
         ``attr``, ``exclude_from_pos_files``, ``exclude_from_bom``,
-        ``locked``, ``dnp``, and any ``_attr_unknown_tokens`` captured
-        during parse. If no flags are set and no unknown tokens are
-        present, no ``(attr ...)`` node is emitted (matches KiCad's
-        canonical "no flags" form).
+        ``dnp``, and any ``_attr_unknown_tokens`` captured during
+        parse. If no flags are set and no unknown tokens are present,
+        no ``(attr ...)`` node is emitted (matches KiCad's canonical
+        "no flags" form).
+
+        ``locked`` is emitted as a top-level ``(locked yes)`` child of
+        the footprint, NEVER as an in-attr ``locked`` token: KiCad 10's
+        kicad-cli rejects the legacy KiCad-6 ``(attr smd locked)`` form
+        with "Failed to load board", which silently breaks zone fill,
+        DRC and gerber export for any board re-saved through the schema
+        layer (issue #3457). Parsing still accepts both forms (see
+        :meth:`from_sexp`), so loading a legacy file and saving it
+        migrates the lock to the modern form.
         """
         sexp_node: SExp | None = self.__dict__.get("_sexp_node")
         if sexp_node is None:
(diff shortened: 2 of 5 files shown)


rjwalters:
## Judge Review — APPROVED

**Verification performed (worktree `.loom/worktrees/issue-3457`):**

1. **Round-trip migration (the crux)** — reproduced independently with a minimal legacy fixture: pure load→save of `(attr smd locked ...)` migrates to top-level `(locked yes)`, never echoes the legacy token, and preserves sibling attr tokens (`exclude_from_pos_files`, `board_only`). Verified separately that an unrelated footprint edit (`fp.dnp = True`) preserves modeled + unknown attr tokens (`allow_missing_courtyard` etc.) with no spurious lock.
2. **Pad-level lock preservation** — confirmed: footprint-level unlock removes only the direct-child `(locked ...)` node; pad-level `(locked yes)` survives both lock and unlock (`find_children` is direct-children-only).
3. **Load-time migration safety** — `_sync_attr_node` only fires when the legacy token is present; modern-form files round-trip byte-identically (verified idempotent load→save). Per-footprint cost is a direct-child attr scan — negligible. No dirty-flag mechanism exists in the schema layer, so no spurious-rewrite risk.
4. **repair_clearance.py reader** — correct: accepts both forms, direct-children-only for top-level, `(locked no)` excluded, bare `(locked)` defaults yes (matches schema parser). All 90 tests pass.
5. **Tests run locally** — test_footprint_locked_form.py (8 pass), test_kicad_cli_roundtrip.py (22 pass against real kicad-cli, no skips — exit-3 canary validated empirically), test_pcb_lock_footprints.py (17 pass), schema PCB subset (1088 pass).
6. **CI 8/8 pass, mergeStateStatus CLEAN, closingIssuesReferences == [3457] only.** Audit confirmed: pcb.py:711 is the sole locked emitter; zero committed `.kicad_pcb` files carry the legacy token.

**Non-blocking finding (pre-existing, NOT introduced by this PR):** `Footprint.from_sexp` (pcb.py:767) uses the recursive `sexp.find("locked")`, so a pad-level `(locked yes)` on an *unlocked* footprint parses as `fp.locked = True`. Consequence: unlocking a footprint that has locked pads doesn't persist across reload, and any attr-field edit on such a footprint will emit a footprint-level lock. Exists on main (from #3410); should use `find_child("locked")`. Recommend a follow-up issue.

LGTM — approving.

PR #3604, opened 2026-06-12
test(auto_pour): regression test for zone preservation with non-zero board origin

### LINKED ISSUE #3461: auto_pour: add unit regression test for zone preservation with non-zero board origin
## Context

PR #3453 (issue #3410) fixed a double origin-subtraction bug in `src/kicad_tools/router/auto_pour.py:_detect_uninset_zones`: `Zone.polygon` vertices are already board-relative after `PCB._detect_board_origin()` (see the coordinate-space invariant docstring at `src/kicad_tools/schema/pcb.py:1800`), but the code subtracted `pcb.board_origin` a second time. On any board whose origin is non-zero (e.g. board 03 at 100,100), every properly-inset hand-tuned zone was flagged "insufficient edge clearance", silently deleted, and replaced with conflicting auto-pour zones.

## Gap

All existing zone-preservation tests in `tests/test_auto_pour.py` (`test_reinsets_existing_uninset_zones`, `test_reinsets_multiline_uninset_zone`, etc.) use boards at origin (0,0) — where the buggy subtraction is a no-op. A re-regression of this bug would pass the entire fast auto-pour suite and only surface in the slow board-03 baseline tests, which currently do not gate (see #3458).

## Changes needed

Add a unit test to `tests/test_auto_pour.py` that:
1. Builds a board whose Edge.Cuts origin is non-zero (e.g. 100,100 like board 03)
2. Includes a hand-tuned zone properly inset (e.g. 0.3mm) from the board edge
3. Runs `auto_pour_if_missing(..., edge_clearance=0.3)` and asserts the zone is PRESERVED (not deleted/regenerated)
4. Optionally a companion case: a genuinely-uninset zone at the same non-zero origin is still detected and re-inset

Discovered during code review of PR #3453.

rjwalters:
Closes #3461

## Summary

Test-only PR (plus one trivial pre-existing lint fix). No live bug found on main — the #3453 fix holds; these tests pin it.

- Adds `TestNonZeroBoardOrigin` to `tests/test_auto_pour.py`, covering the regression class fixed in PR #3453 (issue #3410): `_detect_uninset_zones` double-subtracting `pcb.board_origin` from already-board-relative `Zone.polygon` vertices.
- `test_preserves_inset_zone_at_nonzero_origin`: board with Edge.Cuts origin at (100,100) (board 03 convention) and a hand-tuned GND zone properly inset 0.3mm; asserts `auto_pour_if_missing(..., edge_clearance=0.3)` returns `count == 0` and the original polygon survives verbatim. Under the old bug this zone was flagged "insufficient edge clearance", deleted, and replaced.
- `test_reinsets_uninset_zone_at_nonzero_origin` (companion): a genuinely un-inset zone at the same offset origin is still detected, removed, and regenerated with proper inset — detection isn't broken in the other direction.
- Extends the `_make_pcb` test helper with an `origin` parameter (dynamic Edge.Cuts footer); existing tests unchanged at (0,0).
- Drive-by: fixes pre-existing ruff F841 (unused `y`) in `test_reinsets_existing_uninset_zones` and reformats one over-wrapped assert in the new code. Two other pre-existing `ruff format` diffs in this file were left untouched to keep the diff focused.

## Mutation check

Verified the test FAILS if the origin math regresses: temporarily reintroduced the double subtraction (`bx = px - pcb.board_origin[0]`, `by = py - pcb.board_origin[1]`) in `_detect_uninset_zones` — `test_preserves_inset_zone_at_nonzero_origin` failed with the exact original bug signature:

```
Auto-pour: removing 1 zone(s) with insufficient edge clearance for GND
Auto-pour: created 1 zone(s) for GND
AssertionError: Expected the inset GND zone to be preserved (count=0), got count=1 names=['GND']
```

Mutation reverted before commit.

## Test plan

- [x] `uv run pytest tests/test_auto_pour.py` — 30 passed (28 existing + 2 new)
- [x] Mutation check (above): new test fails against re-broken offset math
- [x] `uv run ruff check tests/test_auto_pour.py` — clean

### DIFF
--- a/tests/test_auto_pour.py
+++ b/tests/test_auto_pour.py
@@ -35,10 +35,31 @@ _PCB_FOOTER = """\
 """
 
 
+def _make_outline_footer(origin: tuple[float, float], size: float = 50.0) -> str:
+    """Build an Edge.Cuts gr_line rectangle footer at an arbitrary origin.
+
+    Coordinates are sheet-absolute (the KiCad file format), so a non-zero
+    *origin* produces a board whose ``PCB.board_origin`` is non-zero --
+    exercising the board-relative coordinate conversion in
+    ``PCB._detect_board_origin()``.
+    """
+    x0, y0 = origin
+    x1, y1 = x0 + size, y0 + size
+    stroke = '(stroke (width 0.05) (type default)) (layer "Edge.Cuts")'
+    return (
+        f"  (gr_line (start {x0:g} {y0:g}) (end {x1:g} {y0:g}) {stroke})\n"
+        f"  (gr_line (start {x1:g} {y0:g}) (end {x1:g} {y1:g}) {stroke})\n"
+        f"  (gr_line (start {x1:g} {y1:g}) (end {x0:g} {y1:g}) {stroke})\n"
+        f"  (gr_line (start {x0:g} {y1:g}) (end {x0:g} {y0:g}) {stroke})\n"
+        ")\n"
+    )
+
+
 def _make_pcb(
     net_defs: list[tuple[int, str]],
     pad_nets: list[tuple[int, str]],
     zones: list[str] | None = None,
+    origin: tuple[float, float] = (0.0, 0.0),
 ) -> str:
     """Build a minimal PCB string.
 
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge Review: APPROVED

**Scope**: test-only — `tests/test_auto_pour.py` (+156/-5), closes #3461 only.

**Verified**:
- **Diff hygiene**: only the test file touched; new `TestNonZeroBoardOrigin` class, `origin` param on `_make_pcb` (existing tests unaffected — (0,0) keeps `_PCB_FOOTER`), plus the trivial F841 `_y_str` fix.
- **Mutation check (reproduced locally)**: reintroduced the `bx = px - pcb.board_origin[0]` double subtraction in `_detect_uninset_zones` — `test_preserves_inset_zone_at_nonzero_origin` fails with the exact original-bug signature ("removing 1 zone(s) with insufficient edge clearance for GND", zone replaced); companion test stays green as expected. Reverted cleanly.
- **Coordinate-space fidelity**: fixture writes Edge.Cuts gr_lines sheet-absolute (mirroring the existing footer format), exercising the gr_line min-x/y fallback in `PCB._detect_board_origin()`. Confirmed `get_board_outline()` returns board-relative coords matching the already-converted `Zone.polygon` — the test pins the documented invariant, not a fixture artifact. The `x_base`/`y_base` self-detection in test 2 correctly rejects double-subtracted output (-99.7 selects base 0 and fails).
- **Suite**: 30/30 pass locally; CI 8/8 green; merge state CLEAN; closingIssuesReferences == [#3461].

Well-documented regression test that closes a real coverage gap (all prior zone tests used origin (0,0) where the bug was a no-op). Ready to merge.

Issue #3597, opened 2026-06-12
board-05 BOM: 100nF 0805 rows assigned C1525 (0402 package) — LCSC guards need package validation


rjwalters:
## Found during #3590 fleet scan

The #3590 value-mismatch fleet scan of committed BOMs surfaced a *package*-mismatch defect that value validation cannot catch:

`boards/05-bldc-motor-controller/output/manufacturing/bom_jlcpcb.csv` line 2:

```
100nF,"C12,C13,C14,C15,C2,C7,C8",Capacitor_SMD:C_0805_2012Metric,C1525
```

C1525 is **CL05B104KO5NNNC — 100nF ±10% 16V 0402** (verified against the JLCPCB parts API on 2026-06-11). The value matches, but the part is an 0402 chip assigned to 0805 footprints — JLCPCB would place a part half the size of the pads.

The correct JLCPCB Basic part for 100nF 0805 is **C49678** (Yageo CC0805KRX7R9BB104, 100nF ±10% 50V 0805, verified via the same API).

## Suggested fix

1. Replace C1525 with C49678 on the 100nF/0805 row of board-05's committed BOM (and update `manifest.json` sha256/size for `bom_jlcpcb.csv`).
2. Extend the #3590 guard layers (`src/kicad_tools/export/lcsc_value_check.py`, used by `bom_enrich.py` cache fallback, `apply_existing_lcsc_assignments`, and the `bom_lcsc_values` preflight check) to also compare the part's known **package** (e.g. `Part.package` \"0402\") against the package extracted from the BOM footprint (`extract_package_from_footprint` in `src/kicad_tools/cost/suggest.py`). Same conservative policy: reject only on clear disagreement, accept when the package is unknown.
3. Note: C1525's local cache record has empty `package`; the MLCC MPN also encodes the size (CL**05** = 0402, CL**21** = 0805, GRM**15**=0402/GRM**21**=0805), which could serve as a fallback like the EIA value-code decoder added in #3590.

## Why not fixed in #3590's PR

#3590 is scoped to *value* mismatches (its fix for board-05's other defect — C6 10uF assigned C1525 — ships there). Package validation is a separate guard layer with its own false-positive considerations.

PR #3607, opened 2026-06-12
fix(export): guard LCSC enrichment against wrong-value parts

### LINKED ISSUE #3590: BOM LCSC enrichment cache fallback can assign wrong part (C1525/100nF matched to 16nF row)
## Bug Description

While regenerating the board-03 manufacturing bundle for #3551 (PR #3589), `kct export --mfr jlcpcb` with the JLCPCB API unreachable (403) fell back to the machine-local parts cache (`_try_cache_fallback` in `src/kicad_tools/export/bom_enrich.py`, keyed by `(value, footprint)` with `ignore_expiry=True`) and assigned **C1525 — a 100nF 0402 capacitor — to the 16nF (C10, C11) BOM row**.

A wrong-value passive silently substituted into a shipped BOM is a manufacturing defect that no current gate catches (BOM preflight only flags *missing* LCSC numbers, not wrong ones).

Worse, the bad assignment then self-perpetuates: the `merge_lcsc` pass in `src/kicad_tools/export/assembly.py` reads LCSC assignments back from the *existing* BOM CSV by the same `(value, footprint)` key on every subsequent export, so one bad cache hit becomes sticky in the committed artifact.

## Where the bad value came from

The stale entry was already in the local cache (`PartsCache.get_enrichment_match("16nF", "Capacitor_SMD:C_0402_1005Metric")` returns C1525 on this machine), so either:

1. a previous API auto-match returned a wrong part and was cached without value validation, or
2. the cache lookup/storage key collapses distinct values.

## Suggested fix

- Validate cache (and API) matches against the BOM value before applying: parse the candidate part's description/attributes and reject when the value clearly disagrees (100nF != 16nF).
- Consider logging cache-fallback assignments at WARNING (currently INFO) since `ignore_expiry=True` explicitly accepts stale data.
- Add a regression test: cache primed with a wrong-value entry must not be applied.

## Workaround used in PR #3589

Exported with `--no-auto-lcsc` and restored the committed BOM CSV from main before export so `merge_lcsc` preserved the correct (blank) assignments.

rjwalters:
Closes #3590

## Summary

Guards the BOM LCSC enrichment pipeline against wrong-value part assignments (the C1525 incident: a 100nF 0402 capacitor silently assigned to the 16nF rows of board-03 during an offline cache-fallback export).

Three defense layers, all built on a new conservative value comparator (`src/kicad_tools/export/lcsc_value_check.py`):

1. **Enrichment cache fallback** (`bom_enrich.py`): before trusting a stale `ignore_expiry=True` cache hit, the candidate part's known value is compared against the BOM row value. A clear mismatch is rejected, logged at WARNING, reported as `unmatched` with an explanatory error, and the poisoned `enrichment_matches` row is **evicted** (`PartsCache.delete_enrichment_match`). API auto-matches get the same validation before being applied or cached. The cache fallback now logs at WARNING (degraded mode) instead of INFO.
2. **`merge_lcsc` CSV read-back** (`bom_formats.py` / `assembly.py`): the new `apply_existing_lcsc_assignments` validates assignments read back from the committed BOM CSV, so one bad cache hit cannot self-perpetuate in the artifact across exports.
3. **BOM preflight** (`preflight.py`): new `bom_lcsc_values` check FAILs the preflight when any assigned LCSC part's known value disagrees with the BOM row value.

Value resolution is layered: structured `value` field → catalog description token scan → EIA 3-digit code decoded from the MLCC manufacturer part number. The MPN decoder matters because the actual poison record on this machine is sparse (`value=''`, `description='CL05B104KO5NNNC'`) — verified that the guard catches it via the `104` EIA code. Unparseable values (ICs, connectors) are treated as "cannot validate" and accepted, so the guard never blocks parts it does not understand.

## Fleet BOM scan

Scanned all 8 committed `bom_jlcpcb.csv` files (65 assigned rows; 15 had local cache records) with the new checker:

- **1 real wrong-value defect found and fixed in this PR**: board-05 had `10uF,C6 -> C1525` (C1525 is 100nF). Reassigned to C15850, matching the other 10uF rows; manifest hash updated.
- After the fix: **0 value mismatches** fleet-wide.
- The poisoned machine-local enrichment entry (`16nF/C_0402_1005Metric -> C1525`) was evicted.

Package/footprint note (#3597's angle): this PR validates **value only**, not package, so #3597 remains open. The fleet scan's informational package pass found no cache-verifiable package mismatches (the C1525 records on board-05's 0805 rows carry no package data locally), but board-05's `100nF [0805] -> C1525 (0402 part)` rows are exactly the case #3597 should gate on.

## Tests

- New `tests/test_lcsc_value_guard.py`: 38 tests covering the comparator (capacitor/resistor/inductor parsing, equivalent notations, description token extraction that ignores part-number substrings, EIA MPN decoding for Samsung/Murata/TDK), cache eviction, and end-to-end reproductions of the C1525 scenario at all three layers (poisoned cache fallback, CSV read-back, preflight).
- Regression suites: `test_bom_enrich.py`, `test_bom_lcsc_merge.py`, `test_preflight.py`, `test_assembly_validation.py`, `test_parts.py`, `test_cost.py`, `test_bom_spec_overlay.py`, `test_sch_preflight.py` — all pass (317 passed, 47 skipped).
- `ruff check` / `ruff format` clean on all touched files.

### DIFF
--- a/boards/05-bldc-motor-controller/output/manufacturing/bom_jlcpcb.csv
+++ b/boards/05-bldc-motor-controller/output/manufacturing/bom_jlcpcb.csv
@@ -2,7 +2,7 @@ Comment,Designator,Footprint,LCSC Part #
 100nF,"C12,C13,C14,C15,C2,C7,C8",Capacitor_SMD:C_0805_2012Metric,C1525
 10k,"R30,R31,R32",Resistor_SMD:R_0805_2012Metric,C17414
 10nF,"C30,C31,C32",Capacitor_SMD:C_0805_2012Metric,
-10uF,C6,Capacitor_SMD:C_0805_2012Metric,C1525
+10uF,C6,Capacitor_SMD:C_0805_2012Metric,C15850
 10uF,"C16,C5",Capacitor_SMD:C_0805_2012Metric,C15850
 15A,F1,Fuse:Fuse_1206_3216Metric,C89657
 1k,"R3,R4",Resistor_SMD:R_0805_2012Metric,
--- a/boards/05-bldc-motor-controller/output/manufacturing/manifest.json
+++ b/boards/05-bldc-motor-controller/output/manufacturing/manifest.json
@@ -5,8 +5,8 @@
   "manufacturer": "jlcpcb",
   "files": {
     "bom_jlcpcb.csv": {
-      "sha256": "57315c76a832dba0f2f23c3ecc8065f3978c2ae7ea57161a9133340b9cdca8d8",
-      "size": 1572
+      "sha256": "0876dd1c239e894b4f171d642764872b56bfaa412309f5f5b86d8edeb4e8cfb1",
+      "size": 1573
     },
     "cpl_jlcpcb.csv": {
       "sha256": "1852681e4d18774db0984a2ab4deb94993404f43239b0b18400998160ffa696b",
--- a/src/kicad_tools/export/assembly.py
+++ b/src/kicad_tools/export/assembly.py
@@ -13,9 +13,14 @@ from pathlib import Path
 from kicad_tools.exceptions import FileNotFoundError as KiCadFileNotFoundError
 
 from .bom_enrich import EnrichmentReport, enrich_bom_lcsc
-from .bom_formats import BOMExportConfig, export_bom, read_existing_lcsc_assignments
+from .bom_formats import (
+    BOMExportConfig,
+    apply_existing_lcsc_assignments,
+    export_bom,
+    read_existing_lcsc_assignments,
+)
 from .bom_spec_overlay import SpecOverlayReport, apply_spec_overlay, find_spec_file
-from .gerber import MANUFACTURER_PRESETS, GerberConfig, GerberExporter, GerberManufacturerPreset
+from .gerber import MANUFACTURER_PRESETS, GerberConfig, GerberExporter
 from .pnp import PnPExportConfig, export_pnp
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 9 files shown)


rjwalters:
## Judge Review: APPROVED

Reviewed against #3590's acceptance criteria with hands-on verification in the worktree.

**Comparator soundness** (`lcsc_value_check.py`): conservative semantics verified by direct probing — unparseable requested values ("100nF 50V X7R", ICs, "1mΩ shunt"), unknown parts, and parts with no value evidence all ACCEPT. Equivalent notations (0.1uF vs 100nF) accept at 5% rel-tol. The EIA MPN decoder correctly handles the real poison record (`value=''`, desc `CL05B104KO5NNNC` → 100nF), is capacitor-gated, and the tolerance-letter lookahead + leading-digit guard reject size codes (0402/155/1005). No false positives found.

**Board-05 fix**: `10uF,C6 → C15850` matches the sibling 10uF row (C16,C5) and C15850's MPN `CL21A106KAYNNNE` decodes to 10uF via the new checker. Manifest sha256 `0876dd1c…` matches the on-disk CSV (LF, no CR). Re-ran the fleet scan on all 8 committed BOMs: 65 assigned rows, 15 cache-known, **0 mismatches**.

**Self-perpetuation break**: `apply_existing_lcsc_assignments` drops poisoned read-back rows with a WARNING naming both values (asserted in tests). Cache fallback validates, evicts (`delete_enrichment_match` — single-statement DELETE in its own commit; safe under SQLite's file locking, consistent with the cache's existing no-app-lock pattern), and logs at WARNING. API auto-matches validated before caching. Verified the local poisoned `enrichment_matches` row is gone.

**Preflight**: `bom_lcsc_values` FAILs on mismatch; blocks export under `strict_preflight` (same semantics as all other FAIL checks), warns otherwise. CI-safe: fresh caches can't validate → OK.

**Tests/CI**: 38/38 new tests pass; regression suites 317 passed, 47 skipped (matches PR claim). ruff check/format clean. CI 8/8 pass, mergeStateStatus CLEAN. Diff scope exactly as declared; closes only #3590.

**Non-blocking notes**:
- Validation is best-effort against the machine-local cache — a fresh machine with an empty cache won't catch a poisoned committed CSV. Acceptable per design; the preflight gate on enriching machines covers the practical risk.
- Package validation deliberately deferred to #3597 (board-05's `100nF [0805] → C1525 (0402)` rows are its test case).

LGTM.

Issue #3450, opened 2026-06-10
tests/router: board03 routing baseline fails on main after #3442 (demote-to-partial reach drop)


rjwalters:
## Summary

`tests/router/test_board03_routing_baseline.py` fails on pristine main (`8996f5c7`, no local changes; native extension built):

- `TestBoard03RoutingBaseline::test_reach_meets_floor`
- `TestBoard03RoutingBaseline::test_per_net_reach_usb_signals`
- `test_routing_reach_deterministic_across_seeds`

3 failed / 2 passed, reproduced twice (loaded machine AND idle machine, 8m24s run).

## Likely cause

PR #3442 (seg-seg quality tuple + demote-to-partial, merged 2026-06-09 19:40) strips physically-overlapping copper instead of committing it, which lowers full-net reach everywhere. #3442 updated `tests/router/test_board02_manufacturable_baseline.py` and `tests/test_board_04_mfr_gate.py` but not the board03 baseline floors. The same mechanism measured on board 05 (#3423/#3449): cpp 4L jlcpcb full-reach dropped 19/32 -> 3/32 while blocking DRC dropped to 0-3.

The seeds-determinism failure suggests the greedy vertex-cover strip order may not be seed-stable — worth checking independently of the floor re-pins.

## Suggested fix

Re-measure board03 reach under post-#3442 main and re-pin the floors per the tests' update policies (or fix seed-stability of the demote pass first if that is the determinism failure's cause).

Found while working #3423 — unrelated to that PR (zero `src/` changes on its branch; failures reproduce on pristine main).

rjwalters:
**Verified STALE — closing.** The full board-03 baseline suite passes on current main.

## Evidence

Fresh worktree at `c545f695` (main tip, 2026-06-12), `uv sync --extra dev` + `uv run kct build-native` (C++ backend available, version 1.0.0), run serially:

```
$ uv run pytest tests/router/test_board03_routing_baseline.py -v -p no:randomly
TestBoard03RoutingBaseline::test_reach_meets_floor PASSED
TestBoard03RoutingBaseline::test_per_net_reach_usb_signals PASSED
TestBoard03RoutingBaseline::test_the_1_of_16_myth_stays_dead PASSED
test_routing_reach_deterministic_across_seeds PASSED
test_committed_pcb_drc_state PASSED
======================== 5 passed in 294.66s (0:04:54) =========================
```

[...]

PR #3613, opened 2026-06-12
fix(drc): repair invalid fill rings with make_valid, not buffer(0)

### LINKED ISSUE #3560: clearance_segment_zone: replace buffer(0) with make_valid for invalid fill rings (bowtie lobe-drop hardening)
## Description

Hardening follow-up from the PR #3558 review of `SegmentZoneClearanceRule._collect_fills` (`src/kicad_tools/validate/rules/clearance.py`).

Invalid fill rings are currently repaired with `poly.buffer(0)`. For the self-touching knockout-trace outlines KiCad actually writes, this is loss-free — verified empirically across all 8 committed routed artifacts (42 invalid fills, worst area delta vs `make_valid` = 0.000000 mm²). But for a true reversed-winding bowtie ring, `buffer(0)` silently drops the negatively-wound lobe (measured: bowtie area 1.0 via buffer(0) vs 2.0 via `make_valid`). A dropped lobe is dropped copper, which would mask a real short under that lobe.

## Suggested Fix

Use `shapely.make_valid(poly)` and keep only the polygonal components (it can return a GeometryCollection containing linework for zero-area slivers):

```python
from shapely import make_valid
from shapely.geometry import MultiPolygon, Polygon

repaired = make_valid(poly)
# extract Polygon/MultiPolygon parts; skip zero-area linework
```

Add a unit test with a bowtie fill ring asserting a trace under the second lobe is still flagged.

## Impact

- **Severity**: Low (no current artifact loses area; defensive only)
- Discovered during code review of PR #3558 (Issue #3527).

rjwalters:
Closes #3560

## Summary

`SegmentZoneClearanceRule._collect_fills` repaired invalid fill rings with `poly.buffer(0)`, which silently drops the negatively-wound lobe of a bowtie/self-intersecting ring — shrinking the copper the rule checks and masking any short under the dropped lobe. This replaces it with `shapely.make_valid` via a new `_repair_fill_polygon` helper that preserves every lobe.

## make_valid handling

- `pyproject.toml` floors shapely at `>=2.0`, so the top-level `shapely.make_valid` import is guaranteed (no `shapely.validation` fallback needed).
- Output types handled:
  - `Polygon` / `MultiPolygon` -> returned as-is (the common cases; bowties come back as MultiPolygon with all lobes).
  - `GeometryCollection` (occurs when part of the outline collapses to zero-area linework) -> polygonal components extracted, nested MultiPolygons flattened; linework/points discarded (zero-area = no copper).
  - Nothing polygonal survives -> empty `Polygon`, which the existing `is_empty` skip in `_collect_fills` handles.

Verified on shapely 2.1.2: bowtie fixture gives area 25.0 via `buffer(0)` (lower lobe dropped) vs 50.0 via `make_valid` (both lobes kept).

## Other buffer(0) repair sites

Full grep of `src/` and `tests/`: the only other site is `src/kicad_tools/pcb/board_geometry.py:321` (`BoardGeometry.from_outline_points`). That repairs the **board outline**, not copper fills, and `BoardGeometry` is single-Polygon by design (its other constructor path already keeps only the largest polygon) — switching it to `make_valid` could return a MultiPolygon and change class semantics, so it is left as a follow-up rather than a same-class fix. No repair idioms exist anywhere else in `validate/` or `analysis/`; other `filled_polygon` consumers there do not construct shapely polygons with validity repair.

## Tests

New `TestInvalidFillRepair` class in `tests/test_validate_segment_zone_clearance.py`:
- Fixture sanity: `make_valid(bowtie).area > bowtie.buffer(0).area` (proves the fixture exhibits the hazard).
- Helper preserves all lobes (area 50.0, full original bounds).
- GeometryCollection extraction (zero-area spike outline).
- Fully degenerate outline -> empty (no copper, no crash).
- **Payoff test**: a trace under the lobe `buffer(0)` drops IS flagged as a short post-fix.
- Two traces, one per lobe -> two shorts (`buffer(0)` would report one).

## Test results

- `tests/test_validate_segment_zone_clearance.py`: 21 passed (15 existing + 6 new)
- `tests/test_drc_regression.py` + `tests/test_board_01_ship_ready.py` + `tests/test_board_05_drc_allowlist.py` + `tests/test_board_06_diffpair_test.py`: 50 passed — fleet committed-artifact DRC verdicts/counts unchanged (expected: the issue measured 0.000000 mm² worst-case area delta across all 42 invalid fills in committed artifacts)
- `tests/test_audit.py`: passed
- `tests/router/test_board03_routing_baseline.py`: 5 passed serially (`-p no:randomly`). An initial interleaved run with `test_audit.py` under pytest-randomly reproduced the known order-dependence flake from #3450 (reach floor / per-net USB); passes cleanly in the suite's documented serial configuration and is unrelated to this DRC-only change.
- `ruff check` + `ruff format`: clean

### DIFF
--- a/src/kicad_tools/validate/rules/clearance.py
+++ b/src/kicad_tools/validate/rules/clearance.py
@@ -787,6 +787,45 @@ class _ZoneFill:
     polygon: object  # shapely (Multi)Polygon
 
 
+def _repair_fill_polygon(poly):  # type: ignore[no-untyped-def]
+    """Repair an invalid fill ring, preserving every copper lobe.
+
+    Stale fills can carry self-touching outlines (KiCad traces knockout
+    holes through the exterior ring) or, in the worst case, genuinely
+    self-intersecting bowtie rings.  ``buffer(0)`` re-nodes the former
+    losslessly but silently *drops* the negatively-wound lobe of a
+    bowtie -- and dropped copper is exactly the copper this rule exists
+    to check, so a real short under that lobe would be masked (Issue
+    #3560).  ``shapely.make_valid`` instead keeps all lobes as a
+    MultiPolygon.
+
+    ``make_valid`` may return a GeometryCollection when part of the
+    outline collapses to zero-area linework; only the polygonal
+    components are copper, so extract those (an empty Polygon is
+    returned if nothing polygonal survives, which callers skip via
+    ``is_empty``).
+    """
+    from shapely import make_valid
+    from shapely.geometry import GeometryCollection, MultiPolygon, Polygon
+
+    repaired = make_valid(poly)
+    if isinstance(repaired, (Polygon, MultiPolygon)):
+        return repaired
+    polys: list[Polygon] = []
+    if isinstance(repaired, GeometryCollection):
+        for geom in repaired.geoms:
+            if isinstance(geom, Polygon):
+                polys.append(geom)
+            elif isinstance(geom, MultiPolygon):
+                polys.extend(geom.geoms)
+    # Pure linework / points (fully degenerate outline) -> no copper.
+    if not polys:
+        return Polygon()
+    if len(polys) == 1:
+        return polys[0]
+    return MultiPolygon(polys)
+
+
 class SegmentZoneClearanceRule(DRCRule):
     """Check track segments against foreign-net zone fill copper.
 
@@ -903,10 +942,9 @@ class SegmentZoneClearanceRule(DRCRule):
                     continue
                 poly = Polygon(points)
                 if not poly.is_valid:
-                    # Stale fills can carry self-touching outlines
-                    # (KiCad traces knockout holes through the exterior
-                    # ring).  buffer(0) re-nodes to equivalent copper.
-                    poly = poly.buffer(0)
+                    # make_valid keeps every copper lobe (buffer(0)
+                    # drops bowtie lobes -- see _repair_fill_polygon).
+                    poly = _repair_fill_polygon(poly)
                 if poly.is_empty:
                     continue
                 layer = zone.filled_polygon_layer(i)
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

**Verification performed:**
- **Helper correctness**: `_repair_fill_polygon` handles Polygon/MultiPolygon pass-through, GeometryCollection polygonal extraction (nested MultiPolygons flattened), and the fully-degenerate path (make_valid returning bare linework, not a GC, also falls through to empty `Polygon()` correctly). Verified identity on already-valid input.
- **No perf cliff**: the helper is called only behind the existing `if not poly.is_valid:` guard in `_collect_fills` — valid fills (the hot path) never invoke `make_valid`.
- **Bowtie claim reproduced**: buffer(0) area 25.0 vs make_valid 50.0 (MultiPolygon, full original bounds) on shapely in this worktree.
- **Tests**: full file runs 21 passed including the 6 new `TestInvalidFillRepair` tests; the payoff test (trace under the dropped lobe flagged as Short) and two-lobes/two-shorts test are exactly the regression coverage #3560 asked for.
- **Fleet stability**: spot-ran `kct check --only segment_zone` on boards 02 and 05 routed artifacts — 0 violations both. Since make_valid copper is a superset of buffer(0) copper, branch=0 implies main=0.
- **board_geometry.py:321 follow-up decision**: agreed. That site repairs the board *outline* (keep-in), where a dropped lobe shrinks the allowed area — conservative failure, not violation masking — and `BoardGeometry` is single-Polygon by design. Reasonable boundary; please ensure a follow-up issue gets filed.
- **CI**: 8/8 checks pass; mergeStateStatus CLEAN; closingIssuesReferences = [#3560] only; diff scope limited to clearance.py + test file; ruff lint/format clean.

Clean, well-scoped hardening fix with strong test coverage. Ready to merge.

PR #3615, opened 2026-06-12
fix(ci): board 06 pour-connectivity — kicad-cli container + asserted pour audit

### LINKED ISSUE #3509: board 06: pour-connectivity repair does not converge on the CI re-route (audit FAIL while gate is green) — latent artifact-refresh trap
## Bug Description

The board 06 recipe's pour-connectivity repair loop (`boards/06-diffpair-test/generate_design.py`, step 10b/10c/11, capped at 3 rounds) converges on the committed artifact's local route but does **not** converge on the CI re-route, and the diff-pair coverage gate does not assert the pour audit — so the job stays green while the recipe prints FAIL.

Evidence from PR #3506's green CI run (27343006197, job 80784327578):

```
[r1] FAIL GND: 79 disjoint pad groups (largest 44/122); 1 zero-fill zone(s)
[r2] FAIL GND: 45 disjoint pad groups (largest 78/122); 1 zero-fill zone(s)
[r3] FAIL GND: 12 disjoint pad groups (largest 111/122); 1 zero-fill zone(s)
[r3] FAIL +1V8: 2 disjoint pad groups (largest 2/3); 1 zero-fill zone(s)
[r3] FAIL +1V2: 3 disjoint pad groups (largest 11/13); 1 zero-fill zone(s)
   POUR CONNECTIVITY: FAIL (see above)
...job passes (gate asserts reach 21/21 + rule coverage + error count only)
```

Notable: the repair IS monotonically converging on GND (79 -> 45 -> 12 disjoint groups) — the 3-round cap is simply too low for the CI runner's routing output — and a `zero-fill zone(s)` count persists across all rounds on GND/+1V8/+1V2, which may be a distinct re-fill defect on the re-route path rather than a round-budget problem.

## Why this is a trap

The committed artifact is protected: `TestPourCopperUnionAudit` (tests/test_board_06_diffpair_test.py) runs the shapely copper-union audit against the committed `diffpair_test_routed.kicad_pcb` in the Test job, so a future artifact refresh that ships broken pours **cannot merge**. But:

1. A future builder refreshing the artifact from a CI-style re-route will produce a file that fails that test — every refresh becomes a fight with non-deterministic pour-repair convergence.
2. Regressions in the repair machinery itself are invisible in CI (green job, FAIL in the log).
3. FAIL-but-green logs are actively confusing for anyone triaging the diffpair job.

## Suggested scope

- Diagnose the persistent `zero-fill zone(s)` on the re-route path (re-fill defect vs. genuinely unfillable geometry under the re-routed copper).
- Raise or adapt the repair round cap (convergence trend suggests ~5-6 rounds would land GND).
- Once convergence is reliable on the 2-core runner, promote the pour audit from informational to asserted in `scripts/ci/check_diffpair_coverage.py` (same pattern as the #3413 phase-5 reach assertion) and re-measure the job ceiling (.github/workflows/ci.yml, currently 30 min).

Refs #3413, PR #3506. Filed by the Judge during review of PR #3506.

rjwalters:
Closes #3509

## Root cause

The `diffpair-routing-regression` job runs on a bare `ubuntu-latest` runner with **no KiCad installed**. The board 06 recipe's pour pipeline (zone fill -> stitch -> repair -> re-fill -> copper-union audit) shells out to `kct zones fill`, which is backed by `kicad-cli`. Ground truth from the green run 27343006197 (job 80784327578):

```
9b. Filling zones (first pass)...
   Zone fill failed (rc=1):
   stderr: Error: kicad-cli not found. Install KiCad 8 from https://www.kicad.org/download/
...same for every 10c re-fill, all 3 rounds...
```

Every fill failed, so **all 5 pour zones stayed zero-fill** and the repair loop was left trying to do a copper plane's job with discrete vias/bridges — that's why GND "converged" 79 -> 45 -> 12 disjoint groups but could never land: the `zero-fill zone(s)` audit term is unfixable without a filler. The round budget was never the primary defect; the environment was. And the gate (`check_diffpair_coverage.py`) never consumed the audit verdict, so the job stayed green while the recipe printed FAIL.

## Fixes

1. **Convergence (environment)**: the job now runs inside the `kicad/kicad:10.0` container (same image + setup pattern as the `test` and `kicad-cli-smoke` jobs), with an early `kicad-cli --version` probe so a lost filler backend fails setup attributably instead of surfacing 15 minutes later as zero-fill pours.
2. **Gate gap closed**: board 06 declares `REQUIRE_POUR_CONNECTIVITY = True` next to its `POUR_NETS` contract; the gate re-runs the recipe's shapely copper-union audit (`_audit_pour_nets`) against the re-routed artifact and fails the job (exit 2) on any disjoint pour net or dead zero-fill zone. An audit that cannot run at all (shapely missing, recipe lost the function) is a tool failure (exit 1) — never a silent pass. The exit clause is explicit: a board that cannot yet pass sets the flag to False with a tracking-issue comment (mirrors the `.github/routed-drc-tolerance.yml` grandfathering convention).
3. **Round budget insurance**: `MAX_POUR_REPAIR_ROUNDS` raised 3 -> 6 (loop still breaks early on PASS), and the loop now short-circuits when the filler is structurally unavailable (first fill AND re-fill both failed) instead of grinding futile rounds (~2.2 min wasted per pre-fix CI run).

## Verification (fresh-route repro, temp dir only — committed artifacts untouched)

Fresh CI-style re-route (`--step route --seed 42`, `PYTHONHASHSEED=42`) into `/tmp` with kicad-cli available:

```
[r1] FAIL GND: 2 disjoint pad groups (largest 115/122)   <- no zero-fill zones at all
[r2] OK   GND: 122 pads in one copper component
[r2] OK   VBUS_USB / +3V3 / +1V8 / +1V2
   POUR CONNECTIVITY: PASS
   SUCCESS: all 21 signal nets routed
```

Gate end-to-end on that fresh artifact: exit 0 (`reach 21/21`, `pour connectivity OK`, all 3 diff-pair rules exercised, 2 DRC errors within the 20 allowlist floor). Gate on a synthesized dead-GND-pour artifact: exit 2 with `GND: 62 disjoint pad groups ... 1 fill-enabled zone(s) with ZERO filled polygons (dead pour)` — the exact trap shape now gates.

Job ceiling: kept at 30 min. Prior green wall was 20m42s; the container adds ~1-2 min (pull + apt) but removes the ~2.2 min of futile repair rounds (re-measure note added inline).

Board 07 (`matchgroup-routing-regression`) does not fill zones, so the sibling job has no parallel gap.

## Test plan

- [x] `tests/test_check_diffpair_coverage.py` (41 passed; container/probe pins inverted, new `TestMeasurePourConnectivity` + board 06 contract test)
- [x] `tests/test_ci_routed_drc_workflow.py` (49 passed)
- [x] `tests/test_check_routed_drc_advisory.py` (16 passed)
- [x] `tests/test_board_06_diffpair_test.py` (39 passed — committed artifact still passes the copper-union audit)
- [x] Full fresh re-route repro + gate run (above)
- [x] `ruff check` clean; `ci.yml` parses
- [ ] First CI run on this PR confirms container wall-clock fits the 30-min ceiling

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -431,6 +431,30 @@ jobs:
     # the merge button.
     name: Diff-Pair Routing Regression
     runs-on: ubuntu-latest
+    # Issue #3509: run inside the official KiCad 10 container (same image
+    # as the test + kicad-cli-smoke jobs).  The board 06 recipe's pour
+    # pipeline (zone fill -> stitch -> pour-connectivity repair -> re-fill
+    # -> copper-union audit) shells out to ``kct zones fill``, which is
+    # backed by kicad-cli.  On the bare ubuntu-latest runner kicad-cli
+    # does not exist, so EVERY fill invocation failed with rc=1
+    # ("kicad-cli not found"), all 5 pour zones stayed zero-fill, and the
+    # repair loop was left trying to do a copper plane's job with discrete
+    # vias/bridges -- the non-converging "FAIL while the gate is green"
+    # trap from PR #3506's run 27343006197.  The container provides
+    # kicad-cli 10 so the fills actually run, matching the local artifact-
+    # refresh environment.  The gate also asserts the pour audit verdict
+    # now (see scripts/ci/check_diffpair_coverage.py + the board recipe's
+    # REQUIRE_POUR_CONNECTIVITY contract), so a future fill regression
+    # fails the job instead of rotting in the log.
+    container:
+      image: kicad/kicad:10.0
+      # Run as root so installs write system locations and actions/checkout
+      # can write $GITHUB_WORKSPACE (mirrors the test + smoke jobs).
+      options: --user 0:0
+    defaults:
+      run:
+        # The image's /bin/sh is dash; force bash (mirrors the test job).
+        shell: bash
     # Issue #2660 acceptance criterion #5: wall-clock <= 5 min on
     # ubuntu-latest.  Originally a 10-minute safety-net ceiling, bumped
     # to 15 minutes per PR #3141 Judge feedback (2026-05-27):
@@ -479,6 +503,14 @@ jobs:
     # budgeted (per-pair 60s, negotiated 360s, repair loop capped at 3
     # rounds), so 30 min gives ~20-30% headroom over the worst expected
     # wall without masking a genuine hang.
+    #
+    # Issue #3509 re-measure note: moving into the kicad/kicad:10.0
+    # container adds ~1-2 min (image pull + apt prerequisites) but the
+    # zone fills now succeed, so the pour-repair loop converges instead
+    # of grinding its full round budget against unfillable zero-fill
+    # zones (the pre-#3509 runs burned ~2.2 min on 3 futile rounds).
+    # Prior green run wall: 20m42s (run 27343006197); ceiling kept at
+    # 30 min.
     timeout-minutes: 30
     # Board 06 (diffpair-test) status (M-F closure, Issue #3097):
     #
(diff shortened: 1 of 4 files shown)


rjwalters:
## Judge review — APPROVED

(Note: a prior judge session crashed mid-review; it left no comments, so this is a complete fresh review.)

### Verified

1. **Sibling-job audit (board 07 / Match-Group):** no same-defect. Its CI path (`--step route`) creates zones in-process via `auto_create_zones_for_pour_nets` but never invokes `kct zones fill`, so there is no kicad-cli dependency to break. However, that means board 07 ships *unfilled* pour zones with no connectivity assertion — same "dead pour" class of gap, recipe-side rather than CI-side. Filed as follow-up #3617 (out of scope here). No other ci.yml job routes + fills outside a KiCad container.
2. **Exit-code contract gates:** disjoint/zero-fill → `failed=True` → exit 2; audit cannot-run (missing `_audit_pour_nets`, crash) → exit 1, never silent. The gate step uses `set -euo pipefail` with no `continue-on-error` / `|| true`.
3. **CI on this PR:** all 8 checks green. Diff-pair job ran IN the kicad/kicad:10.0 container, 11m35s vs 30-min ceiling. Log confirms: `kicad-cli --version` probe passed, "Zones filled" succeeded, repair converged in 2 rounds, "POUR CONNECTIVITY: PASS", and the gate asserted it ("[diffpair-coverage] OK: pour connectivity (5 pour nets)"). Reach 21/21.
4. **generate_design.py:** structural short-circuit only fires when first fill AND the round's re-fill both failed (filler structurally absent); with a working filler the loop iterates up to MAX_POUR_REPAIR_ROUNDS=6 and breaks early on PASS. Sane.
5. **Tests:** tests/test_check_diffpair_coverage.py 41 passed; tests/test_ci_routed_drc_workflow.py 49 passed (local worktree). closingIssuesReferences == [#3509] only. mergeStateStatus CLEAN.

### Verdict

Approved. Swapping `loom:review-requested` → `loom:pr`.

PR #3624, opened 2026-06-12
fix(pcb): make_valid + keep-largest for invalid board outlines (#3614)

### LINKED ISSUE #3614: BoardGeometry.from_outline_points uses buffer(0) for invalid outlines — dropped pieces shrink the keep-in conservatively but silently
Follow-up agreed by the #3613 judge (see its approval comment) — the second and last buffer(0) repair site after PR #3613 fixed the fill-ring one.

src/kicad_tools/pcb/board_geometry.py:321 (`BoardGeometry.from_outline_points`) repairs invalid board outlines with `buffer(0)`. Unlike the fill case, a dropped lobe here SHRINKS the keep-in area — a conservative failure (routing refuses space that exists) rather than violation masking, and `BoardGeometry` is single-Polygon by design (its other path keeps only the largest piece). So this is lower severity, but the repair is still silent.

## Ask
1. Either switch to `shapely.make_valid` + keep-largest-polygon (preserving current single-Polygon semantics but with a deliberate, logged choice), or keep buffer(0) and add a WARNING when the repaired area differs from the raw polygonal area beyond epsilon.
2. Unit test with a bowtie outline pinning the chosen behavior.

Refs PR #3613, #3560.

rjwalters:
Closes #3614

## Summary
- `BoardGeometry.from_outline_points` no longer repairs invalid (self-intersecting) outlines with `buffer(0)`, which silently dropped bowtie lobes and shrank the keep-in area conservatively but invisibly.
- New `_repair_outline_polygon` helper uses `shapely.make_valid`, extracts polygonal pieces (handling Polygon / MultiPolygon / GeometryCollection output, mirroring `_repair_fill_polygon` from PR #3613), keeps the largest piece (preserving BoardGeometry's single-Polygon semantics, matching the Edge.Cuts path), and logs a WARNING naming both the kept area and the total dropped area.
- Valid outlines are an identity pass-through with no warning; fully degenerate outlines (no polygonal area survives repair) now raise `ValueError` instead of returning an empty geometry.

## Tests
- `test_bowtie_outline_keeps_largest_lobe_and_warns`: bowtie ring splitting into 75/7 mm^2 and 12/7 mm^2 lobes — asserts the largest lobe is kept, result is a valid single Polygon, and exactly one WARNING names both areas.
- `test_valid_outline_is_identity_and_silent`: valid rectangle passes through geometrically unchanged with zero warning records.
- `test_degenerate_outline_raises`: collinear points raise `ValueError`.
- Full `tests/test_board_geometry.py`: 39 passed.

## Notes
- No production consumers of `from_outline_points` outside tests; `from_pcb` is unaffected (it never used `buffer(0)`).
- Pre-existing (on clean main) `ruff check`/`ruff format` complaints on both touched files (`F401` unused `MultiPolygon` import guard, `I001` import sort, format drift) were left as-is to avoid unrelated churn.

### DIFF
--- a/src/kicad_tools/pcb/board_geometry.py
+++ b/src/kicad_tools/pcb/board_geometry.py
@@ -21,9 +21,12 @@ Usage::
 
 from __future__ import annotations
 
+import logging
 import math
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
+
+logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
     from kicad_tools.schema.pcb import PCB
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review: APPROVED

**Scope**: 2 files only (`src/kicad_tools/pcb/board_geometry.py`, `tests/test_board_geometry.py`); `closingIssuesReferences == [3614]`. CI 8/8 pass, merge state CLEAN, branch up to date with main (includes merged #3613).

**Findings**:
1. **Parity with #3613**: `_repair_outline_polygon` mirrors `_repair_fill_polygon`'s output-type handling (Polygon / MultiPolygon / GeometryCollection with nested MultiPolygon extraction). The divergence — keep-largest + ValueError instead of keep-all-lobes + empty Polygon — is deliberate, correct for BoardGeometry's single-Polygon contract, and well-documented in the docstring (it even matches the Edge.Cuts path's keep-largest behavior).
2. **WARNING semantics**: fires only in the `len(pieces) > 1` branch, naming kept (75/7 mm²) and dropped (12/7 mm²) areas; the valid-rectangle identity test pins zero warnings.
3. **ValueError on degenerate**: new behavior vs buffer(0)'s old empty polygon — verified via grep that `from_outline_points` has no production callers (tests only), so failing loudly is strictly better.
4. **Performance**: `is_valid` pre-check preserved; `make_valid` only runs on invalid input.

**Tests**: 39/39 pass in worktree, including the bowtie keep-largest, identity-silent, and collinear-raises cases.

No changes requested.

PR #3616, opened 2026-06-12
feat(export): enumerate the THT hand-solder set excluded from the SMT CPL

rjwalters:
Closes #3539

## Summary

The JLCPCB CPL formatter (correctly) excludes through-hole parts from the SMT pick-and-place file, but nothing in the bundle documented which refs the assembler must hand-place. This PR surfaces the excluded set in both assembler-facing documents:

- **`report.md`**: new "Hand-Solder (THT) Components" section (rendered right after Assembly Notes) with a Value / Package / Qty / References table. The full ref list is never truncated — the assembler needs every ref. Singular/plural phrasing handled; SMD-only boards and `exclude_tht=False` exports omit the section cleanly.
- **`README.txt`**: the CPL entry gains a note, e.g. `Note: 18 through-hole component(s) are excluded from this file and must be hand-soldered after SMT assembly: D1, F1, ...`

## Design

- `pnp.py: extract_tht_exclusions(footprints, config)` returns **exactly** the placements the CPL's THT filter drops (same `exclude_from_pos_files`/DNP pre-filters as `extract_placements`, then `attr == "through_hole"`), in natural ref order (R6 before R20). Returns `[]` when `exclude_tht` is off.
- `pnp.py: group_tht_exclusions(placements)` groups them into BOM-style rows `{value, footprint, qty, refs}`.
- `assembly.py`: `AssemblyPackageResult.tht_excluded` is populated by `_generate_pnp` using the formatter's **effective** config, so JLCPCB's `exclude_tht=True` default is captured when no explicit `pnp_config` is supplied (generic/PCBWay default to including THT, so they record nothing).
- `manufacturing.py`: `_tht_component_groups()` feeds the new `ReportData.tht_components` field; `_generate_readme()` appends the CPL note. Template/model/generator wired through `design_report.md.j2`.

## Not in this PR

- **Committed bundles are NOT regenerated** (per #3580 artifact discipline). The next `kct export` regen of softstart/fleet bundles picks up the new section automatically; the softstart bundle will then list its 18 on-board THT refs (D1, F1, J1-J5, Q1A/B, Q2A/B, R6, R20, R21, RV1, SW1, U2, U9).

## Test plan

[...]

PR #3626, opened 2026-06-12
fix(export): aux-origin branch resolves config through formatter (JLCPCB exclude_tht)

### LINKED ISSUE #3618: export_pnp aux-origin branch with config=None synthesizes a config that defeats JLCPCB's exclude_tht default
Found during Judge review of PR #3616 (non-blocking note in the approval comment). Pre-existing; unreachable from #3616's path but latent.

In src/kicad_tools/export/pnp.py, `export_pnp`'s aux-origin branch handles `config=None` by synthesizing a config object — and that synthesized config does not apply the JLCPCB formatter's `exclude_tht=True` default. A caller hitting that branch for a JLCPCB export would ship THT rows in the CPL (and, post-#3616, the hand-solder section/CPL would disagree).

## Ask
1. Make the aux-origin branch resolve config through `get_pnp_formatter(fab_family, pnp_config)` like the main path (single source of truth — the same property PR #3616's judge verified makes CPL/exclusion divergence impossible on the main path).
2. Test: aux-origin + config=None + JLCPCB → THT excluded from CPL and listed in tht_excluded.

Refs PR #3616, #3539.

rjwalters:
Closes #3618

## Summary
- `export_pnp`'s aux-origin branch synthesized a bare `PnPExportConfig` when `config=None`, defeating the JLCPCB formatter's `exclude_tht=True` default (judge-found in PR #3616). A caller hitting that branch would ship THT rows in the CPL while the hand-solder set (`tht_excluded`) listed those same parts — a CPL/exclusion divergence.
- Fix: resolve the effective config through `get_pnp_formatter(manufacturer, config, rotation_corrections)` BEFORE any branching, making the formatter the single source of truth. The aux-origin offset adjustment now uses `dataclasses.replace()`, preserving all other fields (`exclude_tht`, `include_dnp`, `top_only`, ...) instead of hand-copying them, then rebuilds the formatter with the adjusted config.
- Both `config=None` branches collapse into one path; the old `elif config is None` branch (the bug site) is gone entirely.

## Tests
- `test_aux_origin_config_none_jlcpcb_excludes_tht` — the divergence test from the issue: nonzero aux origin + `config=None` + JLCPCB → THT excluded from CPL, SMD coordinates correctly shifted by the aux origin, and `extract_tht_exclusions` with the formatter's effective config lists the same THT part (CPL and hand-solder set agree).
- `test_aux_origin_explicit_config_fields_preserved` — explicit `exclude_tht=False` + manual offsets survive the `replace()` path (offsets combine with aux origin; THT kept).

## Audit of synthesized-config sites (per issue ask)
- `src/kicad_tools/export/pnp.py` — aux-origin branch: FIXED here. `extract_placements`/`extract_tht_exclusions` default to `PnPExportConfig()` only as generic-helper fallbacks; all manufacturer-aware callers pass the formatter's effective config. OK.
- `src/kicad_tools/export/assembly.py` — `_generate_pnp` passes `self.config.pnp_config` (possibly None) straight to `export_pnp` and computes `tht_excluded` from `get_pnp_formatter(...).config`. Already correct (this is the #3616 pattern). No aux-origin path reached from assembly (no `pcb_path` passed).
- `src/kicad_tools/cli/export_cmd.py:261` — `PnPExportConfig(exclude_tht=False)` is intentional (`--include-tht` flag). OK.
- `src/kicad_tools/schema/pcb.py:4943` — `PCB.export_placement()` synthesizes `PnPExportConfig()` and passes it to `export_pnp`, defeating the JLCPCB default the same way (latent, out of scope for this issue's files). Filed as a follow-up issue.

## Test results
- `tests/test_export.py` + `tests/test_manufacturing_package.py`: 129 passed
- `tests/test_report_generator.py`, `tests/test_preflight.py`, `tests/test_export_cmd.py`, `tests/test_pcb_manufacturing_export.py` (#3616-adjacent THT coverage): 228 passed
- `ruff check` clean on changed files; pre-existing `ruff format` drift at pnp.py:99 (unrelated line, present on main) left untouched. A pre-existing import-sort lint error in `tests/test_export.py` (present on main) was autofixed since the file is part of this change.

### DIFF
--- a/src/kicad_tools/export/pnp.py
+++ b/src/kicad_tools/export/pnp.py
@@ -10,7 +10,7 @@ import csv
 import io
 import re
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review — APPROVED

**Verdict**: All review criteria pass. CI 8/8 green, merge state CLEAN.

### Verified
1. **Config resolution**: `formatter = get_pnp_formatter(...)` / `config = formatter.config` now runs before any branching; the buggy `elif config is None` branch (main pnp.py:472) is deleted, and no path synthesizes a bare `PnPExportConfig`. `dataclasses.replace()` preserves all fields. The formatter rebuild after `replace()` passes a non-None config, so `JLCPCBPnPFormatter` (which applies `exclude_tht=True` only when config is None) cannot clobber explicit overrides.
2. **Divergence test fails on main**: confirmed by running the test scenario in plain Python against main's module — J1 appears in the CPL on main (bug present), so `assert "J1" not in output` fails there. Note: running the worktree's pytest from the main checkout misleadingly passes because pytest resolves `pythonpath=["src"]` against the worktree's pyproject — the plain-Python repro is authoritative.
3. **Coordinate semantics**: position-minus-aux-origin (110-100=10, 70-50=20) matches KiCad; verified identical on both main and branch.
4. **#3627 filing accurate**: schema/pcb.py:4943 does `config = PnPExportConfig()` then `_export_pnp(...)` — same latent bug class; issue is open and correctly scoped out.
5. **Tests**: tests/test_export.py (82) + tests/test_manufacturing_package.py (47) = 129 passed locally. assembly.py:508 already resolves the effective config via the formatter, agreeing with the fix.
6. **Scope**: diff limited to pnp.py + test_export.py; closingIssuesReferences == [#3618] only.

LGTM.

Issue #3456, opened 2026-06-10
C++ backend silently falls back to Python for board 03 USB nets (multi-minute grinds)


rjwalters:
Found during #3410 / PR #3453.

On board 03's `route_all` path (0.05mm grid, 80x60mm, jlcpcb-tier1), `router.backend_info['fallback_stats']` reports `fallback_nets=['USB_D+','USB_D-','USB_CC1']` — the C++ A* fails these J1->U1 edges and the Python pathfinder grinds 3-7 MINUTES per net (vs ~seconds in C++). USB_CC2 took ~290s to FAIL through this path pre-#3410.

Two sub-questions:
1. Why does the C++ search fail edges the Python A* solves with clean straight routes? (Candidates: escape-belt expanded-blocked dilation differences, validate->resume exhaustion, per-pad channel budgets #3143.)
2. Fallback is silent at default verbosity — multi-minute nets are indistinguishable from 'router is slow'. Surface a per-net fallback warning.

Repro: load board 03, route_all with the recipe from tests/test_board_03_regression.py's routed_board_03 fixture, inspect backend_info.

PR #3623, opened 2026-06-12
fix(router): loud C++->Python fallback warning + own-net-obstacle parity (closes #3456)

rjwalters:
Closes #3456

## Summary

Board 03's USB nets were silently handed from the C++ A* to the 10-100x-slower pure-Python fallback, grinding 3-7 minutes per net with no indication anything was wrong. Two fixes:

- **Loud fallback (requirement a)**: every Python-fallback attempt now emits a once-per-net-per-run `WARNING` naming the net and the C++ failure reason (`FAILURE_*` codes mapped to human-readable text), and records it in `backend_info['fallback_stats']['fallback_reasons']` — including attempts whose Python fallback ALSO failed, so slow failed grinds are attributable too. Suppressed in relief-probe mode and when the shared per-net budget is already exhausted (fallback skipped — no grind, no noise).
- **The wrongful fallback itself (requirement b — build version 11 → 12)**: the standard-mode branches of C++ `is_trace_blocked` and `is_diagonal_blocked` rejected cells holding the net's **own** `is_obstacle` pad copper (`cell.is_obstacle || cell.net != net`), diverging from the Python siblings' Issue #864 semantics (same-net passable regardless of the obstacle flag). The `is_trace_blocked` divergence was the operative bug: it sealed J1's USB-C pad pockets at the 0.05mm canonical grid (C++ open set exhausted in ~800 iterations on the J1→U1 USB edges; JOY/BTN nets burned up to 6M iterations before `FAILURE_NO_PATH`).

**Result**: board 03's canonical recipe now routes with **zero** Python fallbacks; `route_all` wall time dropped from 15+ minutes to ~7 seconds.

## Root-cause notes (corrects the prior session's hypothesis)

The crashed predecessor session attributed the seal to `is_diagonal_blocked` alone. Empirical A/B with `--force` rebuilds shows the diagonal fix alone does NOT unseal the failing edges — `is_trace_blocked` was the operative predicate. The earlier false validation was caused by `kct build-native` silently skipping the rebuild when a matching-version `.so` exists (filed as #3621). `is_via_blocked` standard mode was experimentally ruled out and keeps its strict pinned contract (reconciliation with the Python sibling tracked in #3622).

## Scope decision

[...]

PR #3630, opened 2026-06-13
fix(router): align C++ standard-mode is_via_blocked with Python (#864 parity)

### LINKED ISSUE #3622: C++ is_via_blocked standard mode diverges from Python _is_via_blocked (own-net is_obstacle cells rejected)
Found while fixing #3456 (standard-mode own-net-obstacle parity for `is_trace_blocked` / `is_diagonal_blocked`, build version 12).

The third sibling predicate, `Pathfinder::is_via_blocked`, still rejects own-net `is_obstacle` cells in standard (non-negotiated) mode:

```cpp
} else {
        return true;
    }
}
```

while the Python `Router._is_via_blocked` standard branch is "same-net passable, different nets block" (`pathfinder.py` ~line 1793, Issue #864 semantics).

This divergence is currently PINNED as intentional by `tests/test_cpp_pathfinder_own_net_obstacle.py::test_own_net_obstacle_rejects_via_standard`, whose rationale ("A* in standard mode never enters obstacle metal anyway") was shown to be false during #3456 -- standard-mode `route_all` does traverse own-pad copper on board 03.

During #3456 this divergence was experimentally ruled OUT as a cause of the board-03 fallbacks (aligning it produced no change on the failing edges, and the trace-predicate fix alone got board 03 to zero fallbacks), so it was deliberately left strict rather than churn a pinned contract inside that PR.

Acceptance criteria:
1. Decide: align C++ standard-mode `is_via_blocked` with Python (#864 parity, flip the pinned test), OR document WHY the strict reject is correct for vias specifically (annular ring vs trace-width geometry?) and align Python to C++ instead.
2. Either way, the two backends must agree -- a board that routes via-through-own-pad in the Python fallback but not in C++ is another silent-fallback seed of exactly the #3456 class.
3. Bump `ROUTER_CPP_BUILD_VERSION` if the C++ predicate changes.

rjwalters:
## Summary

The standard (non-negotiated) arm of `Pathfinder::is_via_blocked_diag` in the C++ router rejected a via candidate whose Euclidean disc touched the routing net's OWN `is_obstacle` pad copper:

```cpp
} else {
    if (cell.is_obstacle || cell.net != net) {  // <- rejects own-net obstacle
        return true;
    }
}
```

The Python `Router._is_via_blocked` standard branch (`pathfinder.py` ~line 1792, Issue #864 semantics) is "same-net passable, different nets block". This divergence meant a board that places a via through its own destination pad would route in the Python fallback but FAIL in C++, silently handing the net to the 10-100x-slower Python fallback — the exact #3456-class silent-fallback seed.

This PR aligns C++ to Python (the simpler, consistent choice that matches the sibling `is_trace_blocked` / `is_diagonal_blocked` predicates already targeted in #3456). Both standard-mode branches (cached fast path and per-net-radius slow path) now read `if (cell.net != net) return true;`.

## Changes

- `cpp/src/pathfinder.cpp`: standard-mode arm of both fast and slow paths in `is_via_blocked_diag` now admits own-net cells regardless of the obstacle flag (foreign-net cells still hard-block).
- `cpp/include/types.hpp`: `ROUTER_CPP_BUILD_VERSION` 11 -> 12 with rationale comment.
- `cpp_backend.py`: `_REQUIRED_CPP_BUILD_VERSION` 11 -> 12 mirror (forces rebuild / fallback guard).
- `cpp_backend.py` (Judge follow-up): `CppPathfinder` now restricts the C++ `routable_layers_` set to the `allowed_layers`-permitted indices (`_apply_allowed_layers_to_routable`). The pathfinder defaults its routable-layer set to every grid layer, so the via-expansion loop would attempt a layer change even on a single-layer board. The strict obstacle reject removed by the #864 relaxation had been the only thing suppressing those vias; filtering the routable set restores the single-layer invariant ("a via requires >= 2 routable layers") **without touching the parity predicate**.
- `tests/test_cpp_pathfinder_own_net_obstacle.py`: flipped the pinned `test_own_net_obstacle_rejects_via_standard` into `test_own_net_obstacle_admits_via_standard` (now asserts NOT blocked), and added `test_foreign_net_obstacle_rejects_via_standard` to pin the same-net-only relaxation. Added `TestSingleLayerOwnNetObstacleNoVias` (Judge follow-up) pinning that the own-net-obstacle admit does not introduce vias on single-layer routes.

| Decide & align (C++ -> Python #864 parity, flip the pinned test) | done | Both standard-mode branches changed to `cell.net != net`; pinned test flipped to admit; matches sibling trace/diagonal predicates |
| Both backends must agree | done | C++ standard branch now mirrors the Python `_is_via_blocked` standard branch exactly (same-net passable, foreign blocks) |
| Bump `ROUTER_CPP_BUILD_VERSION` since the C++ predicate changed | done | Bumped to 12 in `types.hpp`; compiled `.so` reports `BUILD_VERSION = 12`; Python `_REQUIRED_CPP_BUILD_VERSION` mirror bumped to 12 |
| No single-layer via regression (Judge) | done | `routable_layers_` is filtered by `allowed_layers`; `test_single_layer_no_vias` passes; full `test_router_core.py` green |

## Test Plan

- Rebuilt the native extension: `uv run kct build-native` (compiled `.so` reports `BUILD_VERSION = 12`). The Judge follow-up fix is pure-Python (no C++ predicate change), so BUILD_VERSION remains 12.
- `tests/test_cpp_pathfinder_own_net_obstacle.py`: **6 passed** (own-net admit + foreign reject in both standard and negotiated modes, plus the two new single-layer no-via regression tests).
- `tests/test_router_core.py`: **179 passed, 1 xfailed, 0 failed** — including `TestSingleLayerRouting::test_single_layer_no_vias`, which the original test plan omitted and which previously regressed. This file is now part of the test plan.
- Combined `tests/test_router_core.py tests/test_cpp_pathfinder_own_net_obstacle.py`: **185 passed, 1 xfailed, 0 failed**.
- Broader pathfinder / via / parity suite (grid parity, cpp backend, via-blocked-failure, foreign-pad-metal rejection, static-halo via gate, via clearance, flat-array A*, escape/diffpair via tests): 155 passed, 8 skipped, 0 failed.

Closes #3622

### DIFF
--- a/src/kicad_tools/router/cpp/include/types.hpp
+++ b/src/kicad_tools/router/cpp/include/types.hpp
@@ -83,7 +83,21 @@ namespace router {
 // through J1-1's halo at 0.127mm actual vs 0.200mm required).
 // ``mark_blocked`` also began recording ``original_net`` for ALL static
 // cells (previously pad-metal only) so the restore has the owner net.
-constexpr int ROUTER_CPP_BUILD_VERSION = 11;
+//
+// Bump to 12 for Issue #3622 (standard-mode own-net-obstacle parity in
+// ``is_via_blocked``).  The standard (non-negotiated) arm of
+// ``Pathfinder::is_via_blocked_diag`` (both the cached fast path and the
+// per-net-radius slow path) rejected a via candidate whose Euclidean disc
+// touched the routing net's OWN ``is_obstacle`` pad copper
+// (``cell.is_obstacle || cell.net != net``), diverging from the Python
+// ``_is_via_blocked`` standard branch (Issue #864 semantics: same-net
+// cells passable regardless of the obstacle flag) and from the sibling
+// ``is_trace_blocked`` / ``is_diagonal_blocked`` standard predicates.  A
+// board that places a via through its own destination pad in the Python
+// fallback but not in C++ is another silent-fallback seed of the #3456
+// class, so both backends are now aligned and the standard branch
+// reads ``if (cell.net != net) return true;``.
+constexpr int ROUTER_CPP_BUILD_VERSION = 12;
 
 // Grid cell state
 struct GridCell {
--- a/src/kicad_tools/router/cpp/src/pathfinder.cpp
+++ b/src/kicad_tools/router/cpp/src/pathfinder.cpp
@@ -546,7 +546,19 @@ bool Pathfinder::is_via_blocked_diag(int x, int y, int net, bool allow_sharing,
                     }
                     // Allow with cost for routed cells / own-net obstacle.
                 } else {
-                    if (cell.is_obstacle || cell.net != net) {
+                    // Issue #3622: same-net cells are passable in standard
+                    // mode regardless of the obstacle flag -- parity with
+                    // the Python ``_is_via_blocked`` standard branch
+                    // (Issue #864: ``blocked & (net != net)``) and with
+                    // the sibling ``is_trace_blocked`` / ``is_diagonal_blocked``
+                    // standard-mode predicates aligned in Issue #3456.
+                    // Pre-fix, ``cell.is_obstacle`` here rejected a via
+                    // candidate whose disc touched the routing net's OWN
+                    // ``is_obstacle`` pad copper, so a board that routes a
+                    // via through its own destination pad in the Python
+                    // fallback would FAIL in C++ and silently fall back
+                    // (the exact #3456-class silent-fallback seed).
+                    if (cell.net != net) {
                         return true;
                     }
                 }
@@ -594,7 +606,9 @@ bool Pathfinder::is_via_blocked_diag(int x, int y, int net, bool allow_sharing,
                             return true;
                         }
                     } else {
-                        if (cell.is_obstacle || cell.net != net) {
+                        // Issue #3622: same-net cells passable in standard
+                        // mode (Python #864 parity -- see fast path above).
+                        if (cell.net != net) {
                             return true;
                         }
                     }
--- a/src/kicad_tools/router/cpp_backend.py
+++ b/src/kicad_tools/router/cpp_backend.py
@@ -47,7 +47,7 @@ logger = logging.getLogger(__name__)
 # ``AttributeError`` deep in the routing code (e.g. ``router_cpp.PadBounds``
 # missing).  The guard below catches that at import time and falls back to the
 # pure-Python router with an actionable ``kct build-native`` hint.
-_REQUIRED_CPP_BUILD_VERSION = 11
+_REQUIRED_CPP_BUILD_VERSION = 12
 
 # Try to import C++ module with detailed error tracking
 _CPP_IMPORT_ERROR: str | None = None
--- a/src/kicad_tools/router/cpp_backend.py
+++ b/src/kicad_tools/router/cpp_backend.py
@@ -833,6 +833,29 @@ class CppPathfinder:
         self._rules = rules
         self._diagonal_routing = diagonal_routing
 
+        # Issue #3622 follow-up: restrict the C++ via-expansion to the layers
+        # actually permitted by ``allowed_layers`` (Issue #715 single-layer
+        # constraint).  The pathfinder defaults ``routable_layers_`` to every
+        # grid layer, so without this the via loop in ``pathfinder.cpp`` will
+        # attempt a layer change even when only one copper layer is routable.
+        #
+        # Pre-#3622 this leaked no vias because the strict standard-mode
+        # ``is_via_blocked`` rejected via candidates whose clearance disc
+        # touched the routing net's OWN ``is_obstacle`` pad copper.  The #864
+        # parity relaxation (own-net obstacle cells now passable for vias)
+        # removed that incidental suppression, so a single-layer route began
+        # emitting vias that land back on the only routable layer.  Filtering
+        # the routable-layer set here enforces the real single-layer invariant
+        # ("a via requires >= 2 routable layers") without touching the parity
+        # predicate -- with a one-element routable set the via loop's
+        # ``new_layer == current.layer`` skip leaves no via target at all.
+        #
+        # ``_routable_layers`` mirrors the C++ ``routable_layers_`` vector,
+        # defaulting to the grid's routable indices (the pathfinder ctor's
+        # own default) until/unless the allowed-layers filter narrows it.
+        self._routable_layers: list[int] = list(grid.get_routable_indices())
+        self._apply_allowed_layers_to_routable()
+
         # Lazy Python fallback router (constructed on first fallback)
         self._py_router: Router | None = None
         # Fallback statistics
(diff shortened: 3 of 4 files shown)


rjwalters:
❌ **Changes Requested**

The parity change itself is correct and well-reasoned, but it introduces a downstream router regression that the PR's test plan missed.

## What is right

- The C++ standard-mode change (`if (cell.is_obstacle || cell.net != net)` → `if (cell.net != net)`) is an exact match for the Python `_is_via_blocked` standard branch (`pathfinder.py:1733-1738`: "same-net passable, different nets block" — it does NOT consult `is_obstacle`). Verified.
- Both standard-mode branches (cached fast path + per-net-radius slow path) were changed; negotiated branches were left untouched. Good scope discipline.
- `ROUTER_CPP_BUILD_VERSION` bumped 11→12 in `types.hpp`; the compiled `.so` reports `BUILD_VERSION = 12`; the `_REQUIRED_CPP_BUILD_VERSION` mirror in `cpp_backend.py` matches. Verified via rebuild.
- The flipped pinned test + new `test_foreign_net_obstacle_rejects_via_standard` are good — `tests/test_cpp_pathfinder_own_net_obstacle.py` passes 4/4.
- The version-12 collision with sibling PR #3456 is acceptable: #3456 is not merged, both bumps describe the same "standard-mode own-net-obstacle parity" family, and only the latter-merged of the two needs to land its bump. Not blocking.

## Blocking regression

`tests/test_router_core.py::TestSingleLayerRouting::test_single_layer_no_vias` **fails** with this change. A single-layer route (`allowed_layers=["F.Cu"]`) now emits 2 vias:

```
assert 2 == 0  # Single-layer routing should produce no vias
```

I isolated this to the PR's C++ change by bisecting the compiled extension in the issue-3622 worktree:

- Pre-PR C++ (`BUILD_VERSION = 11`): `test_single_layer_no_vias` **PASSES**
- Post-PR C++ (`BUILD_VERSION = 12`): `test_single_layer_no_vias` **FAILS** (2 vias)

(Reverting only `pathfinder.cpp`/`types.hpp`/`cpp_backend.py` to the parent commit and rebuilding restores the pass; restoring the PR and rebuilding re-breaks it. Deterministic.)

Admitting own-net `is_obstacle` cells in standard-mode `is_via_blocked` is making the router treat a via-into-own-pad as feasible even when only one copper layer is routable, so it now produces vias it should never produce on a single-layer board. The parity fix appears correct in principle, but it needs to be reconciled with the single-layer invariant — e.g. the via predicate (or its callers) must still reject vias when the candidate has no second routable layer to land on, independent of the own-net-obstacle relaxation.

## Required changes

1. Fix the single-layer regression so `tests/test_router_core.py::TestSingleLayerRouting::test_single_layer_no_vias` passes again, without reverting the #864 parity relaxation. Likely the via-candidate path needs a guard that a via requires ≥2 routable layers (or the standard-mode relaxation must not open a via opportunity that the strict reject was previously suppressing).
2. Add a regression test pinning "own-net-obstacle admit does not introduce vias on single-layer routes," so this interaction stays covered.
3. Re-run `tests/test_router_core.py` (and the via/pathfinder suite) and update the PR test plan — the current plan's "155 passed, 0 failed" did not include `test_router_core.py`, which is why this was missed.

## Notes on environment (not blocking)

Several other failures in my local run (`test_cli_check.py`, `test_fix_drc_cmd.py`, board-06/07 gate tests, placement collection errors) are `ModuleNotFoundError` for `shapely`/`cmaes` — optional deps not synced in this worktree, not regressions. CI's Test job runs in a provisioned env. The CI "Test" check is also still `pending` at review time; it must pass before approval.

I'll re-evaluate once the single-layer via regression is fixed.

rjwalters:
Addressed the blocking single-layer regression from the review.

**Root cause.** The #864 parity relaxation made own-net `is_obstacle` cells passable for vias in standard mode. That incidentally removed the only thing keeping vias off single-layer (`allowed_layers`) boards: the C++ pathfinder's via-expansion loop iterates **every grid layer** (`routable_layers_` defaults to all grid layers), and the `allowed_layers` constraint was only filtering *start/end* layers in `cpp_backend.py`, never the routable-layer set the via loop walks. Pre-#3622 the strict obstacle reject suppressed those candidate vias; post-#3622 they slipped through, so an `F.Cu`-only route emitted 2 vias.

**Guard added.** `CppPathfinder` now intersects its routable-layer set with the `allowed_layers`-permitted indices in the constructor (`_apply_allowed_layers_to_routable`). With a single allowed layer the set has one element, and the via loop's `new_layer == current.layer` skip leaves no via target — the real "a via requires >= 2 routable layers" invariant. This is a **pure-Python** change; it does not touch the C++ `is_via_blocked` parity predicate, so the #864 relaxation is fully preserved and `BUILD_VERSION` stays 12 (no rebuild of the predicate semantics).

**Regression test.** Added `TestSingleLayerOwnNetObstacleNoVias` in `tests/test_cpp_pathfinder_own_net_obstacle.py`:
- end-to-end: a same-net two-pad `F.Cu`-only route emits zero vias even with own-net obstacle pad copper at the destination;
- white-box: the routable-layer set is restricted to the single `F.Cu` index.

**Tests (now including `test_router_core.py`, which the original plan missed):**
- `tests/test_router_core.py`: **179 passed, 1 xfailed, 0 failed** — `TestSingleLayerRouting::test_single_layer_no_vias` passes again.
- `tests/test_cpp_pathfinder_own_net_obstacle.py`: **6 passed** (4 original + 2 new).
- Combined: **185 passed, 1 xfailed, 0 failed**.

PR body / test plan updated accordingly. Re-labeling to `loom:review-requested`.

rjwalters:
## Changes Requested (re-review after Doctor fix)

The single-layer regression I flagged earlier is genuinely fixed — but CI is red on a **second, separate router regression** that this PR introduces, so it cannot land yet.

### What the Doctor fixed correctly (verified)

- Rebuilt the native extension in the issue-3622 worktree (`BUILD_VERSION = 12`) and ran the targeted suites:
  - `tests/test_router_core.py::TestSingleLayerRouting::test_single_layer_no_vias` **passes again** (the F.Cu-only route now emits 0 vias).
  - `tests/test_router_core.py`: **179 passed, 1 xfailed**.
  - `tests/test_cpp_pathfinder_own_net_obstacle.py`: **6 passed** (incl. the new `TestSingleLayerOwnNetObstacleNoVias` end-to-end + white-box guards).
- The fix is sound, not symptom-hiding. `_apply_allowed_layers_to_routable()` intersects the grid's routable indices with `allowed_layers` and pushes the result into the C++ `routable_layers_` vector; with one allowed layer the via loop's `new_layer == current.layer` skip leaves no via target — the real "a via needs ≥2 routable layers" invariant. It is pure-Python and does **not** touch the C++ `is_via_blocked` predicate, so the #864 own-net relaxation is preserved (`test_foreign_net_obstacle_rejects_via_standard` still passes). Constructor ordering is correct (`self._grid`/`self._rules` set before the call). Defensive empty/no-narrowing guards are reasonable.

### Blocking: CI `Test` job fails on a NEW neck-down regression

CI run 27458989471 — the `Test` job **fails** (all other checks pass):

```
FAILED tests/test_neck_down.py::TestNeckDownRouting::test_routing_without_neckdown_uniform_width
FAILED tests/test_neck_down.py::TestNeckDownRouting::test_routing_fine_pitch_with_neckdown
FAILED tests/test_neck_down.py::TestNeckDownRouting::test_routing_standard_pitch_no_neckdown
FAILED tests/test_neck_down.py::TestNeckDownRouting::test_neckdown_width_range
= 4 failed, 19901 passed ... =
```

Reproduced locally in the issue-3622 worktree (3–4 fail, count is route-dependent). I bisected with fresh native builds:

- merge-base `6af4f893` (main before this PR): `tests/test_neck_down.py` → **26 passed, 0 failed**.
- `e3fa30b5` (first commit — the #864 parity change alone): **3 failed**.
- `8a300850` (HEAD, after the Doctor's allowed-layers fix): still **3–4 failed**.

So the neck-down regression is introduced by the **parity commit `e3fa30b5` itself** (not the Doctor's `cpp_backend.py` guard), and the allowed-layers fix does not address it. The most telling failure is `test_routing_fine_pitch_with_neckdown`:

```
assert found_necked_down_segment, "Expected neck-down segments near fine-pitch IC"
E   assert False
```

The router **no longer produces neck-down segments at all** near fine-pitch pads — the other three failures (`0.20000000298023224` vs `0.2`) are full-width traces appearing where necked-down geometry is expected, not pure float32 noise. Admitting own-net `is_obstacle` cells for vias in standard mode is changing the routed path enough to suppress the fine-pitch taper.

### Required changes

1. Fix the neck-down regression so all of `tests/test_neck_down.py` passes again, **without** reverting the #864 parity relaxation — same constraint as before, now applied to the neck-down interaction. Likely the parity relaxation needs to be reconciled with the fine-pitch taper path (or width/obstacle handling near fine-pitch pads), analogous to how the single-layer case was reconciled via the routable-layer set.
2. Add/extend coverage pinning "own-net-obstacle admit preserves fine-pitch neck-down."
3. Get the CI `Test` job green (run `uv run pytest tests/test_neck_down.py tests/test_router_core.py tests/test_cpp_pathfinder_own_net_obstacle.py` locally before re-requesting). `mergeStateStatus` is currently `UNSTABLE` and must be `CLEAN` before approval.

Once the neck-down suite and the CI `Test` job are green I'll approve — the single-layer half is already in good shape.

rjwalters:
Superseding this stale attempt with a fresh build on current main. The original block (the #864 is_via_blocked parity relaxation broke neck-down routing because the C++ backend lacked neck-down support) is now addressable: #3623 (merged) added Python neck-down post-processing to the C++ backend. Re-attempting #3622 from current main.

PR #3631, opened 2026-06-13
fix(build-native): auto-rebuild when C++ source is newer than installed .so

### LINKED ISSUE #3621: kct build-native silently skips rebuild when a matching-version .so exists (misleading 'installed successfully')
Found while working #3456.

`kct build-native` short-circuits when `is_cpp_available()` is True (`build_native_cmd.py` ~line 194): if any `router_cpp.*.so` with a matching `BUILD_VERSION` is already installed, the command does NOTHING but still prints `C++ backend installed successfully!`.

Consequences during #3456:

- The prior builder session edited `pathfinder.cpp`, ran `kct build-native`, saw "installed successfully", and validated its hypothesis against a STALE binary -- producing a wrong root-cause claim (diagonal predicate) that survived into code comments.
- This session repeated the trap twice: two bisect experiment rounds produced byte-identical results (785 iterations) because the .so was never rebuilt. Only `--force` actually compiles.

The `_REQUIRED_CPP_BUILD_VERSION` guard (#3565 precedent) protects against STALE-VERSION .so files, but offers zero protection during development when the source changes WITHOUT a version bump -- the most common iteration loop.

Suggested fixes (any of):
1. Compare newest mtime of `src/kicad_tools/router/cpp/**/*.{cpp,hpp}` against the installed .so; rebuild when source is newer (make this the default).
2. At minimum, change the short-circuit message to `C++ backend already installed -- SKIPPED rebuild (use --force to recompile)` so the skip is visible.
3. CLAUDE.md note: after editing C++ sources, always `kct build-native --force`.

Repro:
```
touch src/kicad_tools/router/cpp/src/pathfinder.cpp  # or a real edit
uv run kct build-native   # prints "installed successfully", compiles nothing
```

rjwalters:
## Summary

`kct build-native` short-circuited whenever a matching-version `router_cpp.*.so` was installed: it compiled nothing but still printed `C++ backend installed successfully!`. During development the source is frequently edited without a `BUILD_VERSION` bump, so the version guard offered zero protection — devs validated hypotheses against a stale binary (e.g. the wasted bisect rounds in #3456).

This adds mtime-based staleness detection so the default behavior rebuilds when the C++ source is newer than the installed extension, and makes a genuine skip clearly visible.

## Changes

- Add `_find_installed_so`, `_newest_cpp_source_mtime`, and `_is_so_stale` helpers in `build_native_cmd.py`. `_newest_cpp_source_mtime` scans `router/cpp/**` recursively for `*.cpp`/`*.hpp` (and `.cc/.cxx/.hxx/.h`) plus `CMakeLists.txt`.
- In the `build_native` short-circuit (not `--force`): if `is_cpp_available()` but the source is newer than the `.so`, fall through to a real rebuild. Only skip when the `.so` is up to date.
- Add a `skipped` field to `BuildResult` (also surfaced in `to_dict`). When skipped, `format_result_text` now prints `C++ backend already installed -- SKIPPED rebuild (use --force to recompile)` instead of `installed successfully!`.
- `--force` still bypasses the staleness check entirely.
- Document the new auto-rebuild behavior in `CLAUDE.md`.
- New test module `tests/test_build_native_staleness.py`.

| (1) Compare newest cpp/hpp source mtime vs installed .so; rebuild when source newer, by default | Done | `_is_so_stale` + short-circuit fall-through; `test_rebuilds_when_source_is_stale` proves it reaches the build path without `--force` |
| (2) Skip message is visible, not misleading | Done | `format_result_text` emits `SKIPPED rebuild (use --force to recompile)`; `test_skips_and_marks_skipped_when_up_to_date` and `test_skipped_message_distinct_from_installed` assert `"installed successfully"` is absent |
| (3) CLAUDE.md note on default staleness detection + `--force` | Done | "Rebuilding after editing C++ sources" section added |
| `--force` always recompiles | Done | `test_force_bypasses_short_circuit_entirely` asserts `_is_so_stale` is never consulted and the build path is entered |
| Tests added for staleness detection | Done | `tests/test_build_native_staleness.py` (11 tests) |

## Test Plan

- `uv run pytest tests/test_build_native_staleness.py` — 11 passed.
- `uv run pytest tests/test_router_cpp_auto_build.py` — existing auto-build suite still green (24 passed, 1 env-skip).
- `uv run ruff check` / `ruff format --check` on the changed files — clean.

Note: the issue's live repro (`touch pathfinder.cpp && uv run kct build-native`) assumes a pre-built `.so`; this fresh worktree has none (`build-native --check` reports "not installed"), so the staleness branches are covered via the unit tests above rather than a full cmake build.

Closes #3621

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -24,4 +24,16 @@ new git worktrees (`.loom/worktrees/issue-N/`) need this step explicitly.
 The C++ backend gives a 10-100x speedup for the A* loop, so a missing
 extension is the most likely cause of multi-minute-per-net routing.
 
+### Rebuilding after editing C++ sources
+
+`kct build-native` now detects staleness automatically: it compares the
+newest mtime of `src/kicad_tools/router/cpp/**` (`.cpp`/`.hpp` and
+`CMakeLists.txt`) against the installed `.so` and **rebuilds when the
+source is newer**, even without a `BUILD_VERSION` bump. When the `.so` is
+already up to date it prints `SKIPPED rebuild` instead of the old
+misleading `installed successfully!`.
+
+`--force` always recompiles regardless of mtimes — reach for it if you
+suspect the auto-detection missed something (e.g. a touched build flag).
+
 See `README.md` "Fresh worktree checklist" for the full setup sequence.
--- a/src/kicad_tools/cli/build_native_cmd.py
+++ b/src/kicad_tools/cli/build_native_cmd.py
@@ -30,6 +30,7 @@ class BuildResult:
     error_message: str | None = None
     steps_completed: list[str] = field(default_factory=list)
     warnings: list[str] = field(default_factory=list)
+    skipped: bool = False
 
     def to_dict(self) -> dict:
         """Convert to dictionary for JSON output."""
@@ -40,6 +41,7 @@ class BuildResult:
             "error_message": self.error_message,
             "steps_completed": self.steps_completed,
             "warnings": self.warnings,
+            "skipped": self.skipped,
         }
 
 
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

Clean, well-scoped fix for #3621. The mtime-based staleness detection is correct and the skip message is now honest.

**Correctness**
- `_is_so_stale` uses strict `newest_source > so_mtime`, so an up-to-date `.so` (equal or newer mtime) correctly skips — verified by `test_false_when_so_newer`.
- All safe-defer paths return `False` (no `.so`, no cpp dir, no sources, `OSError` on stat), so a pip wheel without bundled sources falls back to the existing version guard rather than forcing an impossible rebuild.
- The `--force` path wraps the entire short-circuit in `if not force:`, and `test_force_bypasses_short_circuit_entirely` proves `_is_so_stale` is never consulted when forcing.

**Source glob sanity-check**
Verified against the real tree: `router/cpp/{src,include}/*.{cpp,hpp}` + `CMakeLists.txt`. `_newest_cpp_source_mtime` uses `rglob`, so the nested `src/`/`include/` subdirs are covered. The only extensions present (cpp/hpp/txt) are all matched; the extra `.cc/.cxx/.hxx/.h` patterns are harmless forward-compat. `router_dir` and `_get_cpp_source_dir` resolve consistently under the package root.

**Tests**
`tests/test_build_native_staleness.py` — 11 passed locally (mocked, no compiler invoked). Covers the three helpers plus both short-circuit branches and the SKIPPED-vs-installed message distinction. Builder's note about the worktree lacking a compiled `.so` is reasonable; the unit coverage is sufficient for the decision logic.

## Test Execution
1. `uv run pytest tests/test_build_native_staleness.py` — ✅ 11 passed
2. `uv run pytest tests/test_router_cpp_auto_build.py` — ✅ existing suite green
3. `ruff check` / `ruff format --check` on changed files — ✅ clean

## CI
All 8 checks pass (Test, Type Check, Lint & Format, C++ Build Check, both routing regressions, DRC, round-trip smoke). mergeStateStatus: CLEAN. PR uses `Closes #3621`.

Issue #3507, opened 2026-06-11
router: optimizer/nudge passes mutate routes without re-marking the grid — post-route overlaps undetectable and unrepairable (board 06 RX1 cluster, 17 DRC errors)


rjwalters:
## Background

During PR #3506 (issue #3413 phases 4-6, board 06 diffpair-test), 17 of the 33 blocking errors in the committed artifact's strict gate are ONE optimizer/nudge-introduced USB3_RX1+/RX1- overlap cluster at the J1 fan-out:

- 7 `diffpair_clearance_intra` + 10 `clearance_segment_via`
- The post-route copper (before optimizer/nudge passes) is overlap-free — verified in-recipe.
- The optimizer and nudge passes mutate `router.routes` **without re-marking the routing grid**, so:
  1. Their own collision checking misses the pair overlap they introduce.
  2. Any subsequent repair (e.g. board 06's transactional solo-re-route in `boards/06-diffpair-test/generate_design.py::_repair_pair_overlap_solo`) runs A* against a STALE grid — the solo re-route re-lands on the partner/foreign copper and must roll back.

## Proposed fix

Router infrastructure: grid-state resync for post-route passes. After the optimizer/nudge mutate routes, the grid occupancy must be re-derived from the mutated routes (unmark old geometry, mark new geometry) so that:
- post-route passes can do honest collision checking, and
- targeted repair re-routes see the true copper state.

Related-but-distinct open issues: #3454 (TraceOptimizer merges into via-clearance violations), #3444 (board 05 fresh re-route overlaps), #3501 (correction-pass rollback unmark mismatch). This issue is specifically the grid-staleness mechanism shared by the optimizer/nudge family.

## Acceptance

- Optimizer/nudge passes leave the grid consistent with `router.routes` (or expose an explicit resync API used by all post-route passes).
- Board 06 recipe's solo-re-route repair can succeed (no stale-grid rollback) — the RX1+/RX1- overlap cluster is repaired.
- Exit clause (b) in `.github/routed-drc-tolerance.yml` for board 06 can fire: floor 33 tightens by 17.

## Related

- PR #3506 (diagnosis + transactional repair that currently must roll back)
- Issue #3413 (board 06 reach/connectivity phases)
- `.github/routed-drc-tolerance.yml` board-06 entry documents this cluster + exit clause

PR #3634, opened 2026-06-13
fix(router): batched resync re-marks same-net sibling cells after optimize loop

### LINKED ISSUE #3511: optimize_routes_grid_synced: same-net sibling cells can be under-marked after per-route unmark
## Context

Found while reviewing PR #3510 (issue #3507). Not blocking — empirically not biting (board 03 0-DRC with optimizer ON, board 06 floor 33→20, cross-board regressions pass) — but it is a known asymmetry worth closing.

## Gap

`RoutingGrid.resync_route_occupancy` (grid.py) has an explicit step 4: after the net-guarded unmark of stale geometry, it re-marks EVERY route of every affected net, because the unmark may have cleared cells under a geometry-unchanged same-net sibling route. This is covered by `test_same_net_sibling_cells_survive_resync`.

`optimize_routes_grid_synced` (router/optimizer/trace.py) has no equivalent step. Its per-route transaction is:

```python
grid.unmark_route(route)        # net-guarded cell unmark
grid.mark_route(optimized)      # marks only the NEW geometry
grid._mark_route_on_cpp_cells(optimized)
```

If the mutated route's OLD envelope overlapped a same-net sibling route that the optimizer leaves geometry-unchanged, the sibling's cells inside that envelope are cleared and never re-marked (the later `drc_verify_and_nudge` exit resync skips geometry-unchanged pairs, so it does not repair this either). Result: the grid under-blocks real own-net copper, and a downstream foreign-net consumer (e.g. a targeted repair re-route) could path through the sibling's clearance envelope.

## Why exposure is low today

- Old/new optimizer envelopes mostly coincide (collinear merge / zigzag / 45-corner stay in-corridor).
- Same-net overlap regions near shared pads are `pad_blocked` and survive unmark.
- Board 06's step-6b solo re-route keeps a world-coordinate geometric cross-check as defense-in-depth and rolls back on any cross-net contact.

## Suggested fix

Keep the incremental per-route unmark/mark (route *i* must check against 0..*i-1* post-opt copper), and add ONE batched repair at the end of the loop: collect the `(old, new)` pairs for mutated routes and call `grid.resync_route_occupancy(pairs)` once after the loop — its affected-net re-mark closes the sibling gap and its single wholesale R-tree rebuild stays O(n) per pass (do NOT call resync per route; that is the O(n^2) rebuild trap).

Alternatively: after the loop, re-mark all routes of nets that had >1 route and >=1 mutation.

## Secondary (same PR review)

`resync_route_occupancy`'s deferral-window branch (queueing the stale C++ unmark into `_deferred_cpp_unmarks` when `_cpp_unmark_deferred` is set, grid.py ~line 4247) is currently unreachable in practice (post-route passes run after the negotiated loop closes its iteration windows) and has no test. A unit test pinning the queue-don't-apply behavior would protect it if a future caller invokes resync inside a window.

rjwalters:
## Summary

`optimize_routes_grid_synced` (router/optimizer/trace.py) rips and re-marks copper incrementally per route so route *i* collision-checks against the post-opt copper of routes 0..*i*-1. But `grid.unmark_route(route)` is net-guarded yet geometry-scoped: it clears every cell the OLD envelope owned, including cells still occupied by a geometry-unchanged same-net SIBLING route where the two envelopes overlapped. The per-route `mark_route` only re-marks the mutated route's NEW geometry, so the sibling's cells in the old overlap region stayed blank — the grid under-blocked real own-net copper (a downstream foreign-net consumer could path through the sibling's clearance envelope).

`RoutingGrid.resync_route_occupancy` already solves the identical asymmetry with its step-4 affected-net re-mark, but the optimize loop had no equivalent. This PR closes the gap with ONE batched resync after the loop.

## Changes

- `src/kicad_tools/router/optimizer/trace.py`: collect the `(old, new)` pair for every mutated route and call `grid.resync_route_occupancy(pairs)` once after the loop. The incremental per-route unmark/mark is preserved (route *i* still checks against 0..*i*-1 post-opt copper); the batched resync only repairs same-net sibling under-marking. Single call keeps the pass O(n) — per-route resync would re-rebuild the R-trees n times (the O(n^2) trap called out in the issue).
- `tests/router/test_grid_resync_3507.py`: three regression tests (details below).

| Batched `resync_route_occupancy` (or equivalent) after the optimize loop re-marks same-net sibling cells | done | `test_same_net_sibling_cells_survive_optimize_loop`: a single-segment vertical sibling crossing a collinear-mergeable horizontal route at the same net; after the loop the sibling's crossing-cell copper is still marked |
| Single batched resync (not per-route) to avoid O(n^2) R-tree rebuild | done | One `grid.resync_route_occupancy(mutated_pairs)` call after the loop; `test_no_resync_when_nothing_mutated` confirms it is skipped entirely when no route mutates |
| Resync deferral-window regression test | done | `test_resync_defers_cpp_unmark_inside_window`: with `begin_cpp_unmark_deferral` open, the stale C++ unmark is queued (not applied) into `_deferred_cpp_unmarks`, the flush applies it, and the Python grid is updated immediately |
| Incremental per-route check semantics preserved | done | Per-route unmark/mark loop unchanged; existing `test_grid_consistent_after_optimize`, `test_unchanged_route_keeps_identity`, skip-nets tests still pass |

## Test Plan

- `pytest tests/router/test_grid_resync_3507.py` — 20 passed (17 prior + 3 new).
- `pytest tests/router/ -k "optimiz or resync or grid_synced or trace"` — 25 passed.
- Board 03 routing baseline: 4 routing-dependent tests pass. The 1 unrelated failure (`test_committed_pcb_drc_state`, "Could not parse error count from 'kct check' output") reproduces identically on clean `main` in 1.04s with no routing — a pre-existing CLI-output-parse mismatch, out of scope.

Closes #3511

### DIFF
--- a/src/kicad_tools/router/optimizer/trace.py
+++ b/src/kicad_tools/router/optimizer/trace.py
@@ -61,6 +61,20 @@ def optimize_routes_grid_synced(
     optimized geometry is unchanged are skipped (the incremental
     unmark-old/mark-new perf shape).
 
+    Issue #3511: the incremental per-route unmark is net-guarded but
+    geometry-scoped -- ``grid.unmark_route(route)`` clears every cell the
+    OLD envelope owned, including cells that a same-net SIBLING route
+    (left geometry-unchanged by the optimizer) still occupies where the
+    two envelopes overlapped.  The per-route ``mark_route`` only re-marks
+    the mutated route's NEW geometry, so the sibling's cells inside the
+    old overlap region stay cleared and the grid under-blocks real
+    own-net copper.  To close that asymmetry we collect the
+    ``(old, new)`` pair for every mutated route and issue ONE batched
+    :meth:`RoutingGrid.resync_route_occupancy` after the loop: its step-4
+    affected-net re-mark restores the sibling cells, and its single
+    wholesale R-tree rebuild keeps the pass O(n) (calling resync per
+    route would re-rebuild the R-trees n times -- the O(n^2) trap).
+
     Args:
         router: ``Autorouter`` whose ``routes`` will be optimized and
             replaced in place.
@@ -79,6 +93,10 @@ def optimize_routes_grid_synced(
     """
     grid = router.grid
     optimized_routes: list[Route] = []
+    # Issue #3511: (old, new) pairs for every MUTATED route, replayed
+    # through a single batched resync after the loop so same-net sibling
+    # cells cleared by an incremental unmark are re-marked.
+    mutated_pairs: list[tuple[Route, Route]] = []
     for route in router.routes:
         if skip_nets is not None and route.net in skip_nets:
             optimized_routes.append(route)
@@ -97,11 +115,26 @@ def optimize_routes_grid_synced(
             continue
         # Incremental grid transaction: rip the old copper (cells +
         # R-trees + grid.routes + C++ mirror + stored-route snapshot
-        # invalidation), then commit the new copper on both grids.
+        # invalidation), then commit the new copper on both grids.  Route
+        # ``i`` must collision-check against the POST-opt copper of routes
+        # ``0..i-1``, so this stays incremental -- the batched resync
+        # below only repairs same-net sibling under-marking, it does not
+        # replace the per-route transaction.
         grid.unmark_route(route)
         grid.mark_route(optimized)
         grid._mark_route_on_cpp_cells(optimized)
+        mutated_pairs.append((route, optimized))
     router.routes = optimized_routes
+    # Issue #3511: ONE batched repair after the loop.  The incremental
+    # per-route ``unmark_route`` above clears every cell the old envelope
+    # owned, which can blank cells still occupied by a geometry-unchanged
+    # same-net sibling route where the envelopes overlapped; the per-route
+    # ``mark_route`` only re-marks the mutated route's new geometry.
+    # ``resync_route_occupancy``'s affected-net re-mark (its step 4)
+    # restores those sibling cells, and its single wholesale R-tree
+    # rebuild keeps the pass O(n) -- do NOT call it per route.
+    if mutated_pairs:
+        grid.resync_route_occupancy(mutated_pairs)
     return optimized_routes
 
 
--- a/tests/router/test_grid_resync_3507.py
+++ b/tests/router/test_grid_resync_3507.py
@@ -209,6 +209,52 @@ class TestResyncRouteOccupancy:
         _assert_marked(router.grid, mutated)
         _assert_unmarked(router.grid, snapshot.segments[0])
 
+    def test_resync_defers_cpp_unmark_inside_window(self):
+        """Issue #3511 secondary: with a corridor-reservation window open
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Correct, scope-disciplined fix for the same-net sibling under-marking asymmetry in `optimize_routes_grid_synced`.

## Correctness
- The batched `grid.resync_route_occupancy(mutated_pairs)` is the right mechanism: its step-4 affected-net re-mark (grid.py ~L4681) re-marks **every** route of every affected net from the post-swap `self.routes` — which still holds the geometry-unchanged sibling — so cleared sibling cells are restored. Verified against the implementation.
- **O(n) claim is sound**: `resync_route_occupancy` does exactly ONE wholesale R-tree rebuild (`_rebuild_segment_index` + `rebuild_via_index`) regardless of pair count; calling it per-route would be the O(n²) rebuild trap the issue calls out. The PR correctly batches.
- **No-mutation guard works**: `if mutated_pairs:` skips the resync entirely when nothing mutates — load-bearing-tested by `test_no_resync_when_nothing_mutated` (spy asserts `resync_route_occupancy` is never called).
- Incremental per-route unmark/mark semantics preserved (route *i* still collision-checks against 0..*i*-1 post-opt copper); the batched resync only repairs sibling under-marking.

## Tests
- All 17 tests in `test_grid_resync_3507.py` pass locally; the deferral-window test (`test_resync_defers_cpp_unmark_inside_window`) genuinely exercises the otherwise-unreachable grid.py step-1 defer-don't-apply branch and is a real addition.
- **Caveat (non-blocking):** the headline `test_same_net_sibling_cells_survive_optimize_loop` passes identically with and without the fix — I confirmed by reverting trace.py to `main` (0 resync calls) and re-running; it still passed. Its minimal geometry doesn't expose the bug because the merged route's new envelope re-covers the crossing cells. The genuinely load-bearing coverage of the resync mechanism is the pre-existing `test_same_net_sibling_cells_survive_resync` (grid-API level, #3507). Filed #3637 to harden the optimize-loop-level test. Approving because: (a) #3511 is explicitly non-biting defensive hardening, (b) the resync mechanism itself is proven load-bearing, and (c) the wiring is exercised end-to-end.
- Minor: PR body says "20 passed (17 prior + 3 new)" — actual total is 17 collected. Description arithmetic only; no code impact.

## Pre-existing items confirmed NOT caused by this PR
- The 3 `F401` warnings in trace.py (`perpendicular_direction`, `project_point_onto_line`, `translate_segment`) exist identically on `main` (the PR doesn't touch the import block). CI "Lint & Format" passes.
- `test_board03_routing_baseline.py::test_committed_pcb_drc_state` — out of scope per PR; not a routing test.

## CI
All 8 checks pass (C++ Build, Diff-Pair Regression, Lint & Format, Match-Group Regression, Routed PCB DRC, Test, Type Check, kicad-cli Smoke). `mergeStateStatus: CLEAN`.

`Closes #3511` present. Follow-up #3637 tracks the test-hardening.

PR #3639, opened 2026-06-13
fix(router): restore placement atomically with best-routes snapshot

### LINKED ISSUE #3504: PlacementFeedbackLoop best-snapshot restore can pair stale routes with moved placement
## Context

Found while reviewing PR #3503 (issue #3474 R2). Pre-existing in the #2840 best-snapshot mechanism; #3503 extends the snapshot seeding to the pre-loop input state but does not change the restore semantics.

## Problem

`PlacementFeedbackLoop.run()` (src/kicad_tools/router/placement_feedback.py) restores the best-known **routes** snapshot before returning, but does NOT restore the **component placement** that those routes were computed under:

- Placement adjustments are applied to `self.pcb` in-place at the end of each iteration (after that iteration's routing + snapshot).
- If the best snapshot comes from iteration N (or, post-#3503, from the pre-loop input, `best_iteration == -1`) and the loop applied moves in iterations N..final, the restored routes were computed against pad positions that no longer match the saved placement.
- `result.placement_diff` records the moves; `result.routes` reflects pre-move geometry. The saved artifact can contain opens/DRC at the moved pads.

## Mitigating factors

- `get_failed_nets()` derives from routes only, so the reported routed count is internally consistent.
- In the measured chorus R2 case the loop moved 0 components, so the #3503 restore was geometrically safe.
- The downstream connectivity checker / DRC will catch the inconsistency when it occurs — but only after the run has paid for it.

## Suggested fix

Snapshot component positions alongside routes when taking the best snapshot, and restore them together (or revert applied strategies via the recorded `PlacementAdjustment` originals from `_snapshot_positions`) whenever the restored best predates one or more applied moves.

## Acceptance criteria

- Restoring a best snapshot taken before placement moves also restores the placement those routes were computed under (or documents and asserts that moves are reverted).
- Unit test: loop where iteration 0 is best, iteration 1 applies a move and regresses — final PCB placement must match iteration-0 (pre-move) positions.
- `placement_diff` is consistent with the returned routes/placement.

rjwalters:
## Summary

`PlacementFeedbackLoop.run()` restored the best-known **routes** snapshot before returning but never restored the **component placement** those routes were computed against. Placement adjustments are applied to `self.pcb` in-place at the end of each iteration, so when the best snapshot predated one or more applied moves, the loop paired stale routes with a moved placement — the saved artifact could contain opens/DRC at pad positions that no longer matched the geometry the routes describe.

This makes the best snapshot **atomic**: footprint positions/rotations are captured alongside the routes and restored together.

## Changes

- `placement_feedback.py`: added `_snapshot_placement()` / `_restore_placement()` helpers.
- Capture the placement snapshot both for the pre-loop seed (#3474 path) and on every strict routed-count improvement, taken at the point in the iteration *before* that iteration's move is applied — so the captured placement is the one the snapshotted routes were computed under.
- In the best-snapshot restore branch, restore the placement alongside the routes. `placement_diff` is built from `self.pcb` after restoration, so it now reflects the restored (reverted) placement.
- Added a `TestPlacementFeedbackBestSnapshotAtomicPlacement` test class with a moving applicator.

| Restoring a best snapshot taken before placement moves also restores the placement those routes were computed under | done | `test_restored_routes_match_restored_placement`: iter-0 best, iter-1 moves U1 + regresses; after restore U1 is back at its pre-move position |
| Unit test: iteration 0 best, iteration 1 applies a move and regresses — final placement matches iteration-0 positions | done | Same test asserts `u1.position == original_xy` |
| `placement_diff` consistent with returned routes/placement | done | `test_placement_diff_consistent_with_restored_placement`: reverted move produces no `placement_diff` entry for U1 |
| Winning moves are not spuriously reverted | done | `test_improving_move_kept_when_later_iteration_is_best`: improving move retained |

## Test Plan

- `pytest tests/test_placement_feedback.py` — 61 passed (3 new)
- `pytest tests/test_route_cmd_placement_feedback.py tests/test_chorus_test_placement_feedback.py` — 56 passed, 1 skipped
- `ruff check` on both changed files — clean

Closes #3504

### DIFF
--- a/src/kicad_tools/router/placement_feedback.py
+++ b/src/kicad_tools/router/placement_feedback.py
@@ -341,6 +341,20 @@ class PlacementFeedbackLoop:
         best_routed_count = -1
         best_iteration = -1
 
+        # Issue #3504: the routes snapshot above is only geometrically
+        # meaningful when paired with the component placement it was
+        # computed against.  Placement adjustments are applied to
+        # ``self.pcb`` in-place at the END of each iteration, so if the
+        # best snapshot came from iteration N (or the pre-loop input) and
+        # the loop moved components in iterations N..final, then restoring
+        # only ``self.router.routes`` pairs stale routes with a moved
+        # placement -- the saved artifact can contain opens/DRC at the new
+        # pad positions.  Snapshot the placement ALONGSIDE the routes and
+        # restore them together so the best snapshot is atomic.
+        best_placement_snapshot: dict[str, tuple[tuple[float, float], float]] = (
+            self._snapshot_placement()
+        )
+
         # Issue #3474 (chorus R2): seed the best-known state with the
         # PRE-LOOP routes.  The #2840 snapshot only made the loop
         # monotonic across its own iterations -- iteration 0 still calls
@@ -433,6 +447,13 @@ class PlacementFeedbackLoop:
                 best_routed_count = routed_count
                 best_routes_snapshot = copy.deepcopy(list(self.router.routes))
                 best_iteration = iteration
+                # Issue #3504: capture the placement these routes were
+                # computed against, atomically with the routes themselves.
+                # At this point in the iteration NO move has yet been
+                # applied for the current pass (moves happen at the bottom
+                # of the loop body), so ``self.pcb`` still reflects the
+                # placement that produced ``routes`` above.
+                best_placement_snapshot = self._snapshot_placement()
             else:
                 improved = False
 
@@ -556,6 +577,14 @@ class PlacementFeedbackLoop:
         # solely from ``self.router.routes`` (core.py:8320), so restoring
         # the routes is sufficient to make the reported failed-net count
         # consistent with the restored snapshot.
+        #
+        # Issue #3504: ALSO restore the component placement captured
+        # alongside the routes.  The restored routes were computed against
+        # ``best_placement_snapshot``; if the loop applied moves after that
+        # snapshot was taken, ``self.pcb`` no longer matches.  Restoring
+        # the placement keeps the returned routes geometrically consistent
+        # with the placement (and with ``placement_diff``, which is built
+        # from ``self.pcb`` below).
         current_routed_count = len(self.router.nets) - 1 - len(self.router.get_failed_nets())
         if best_routed_count > current_routed_count:
             if self.verbose:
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the atomic placement-snapshot fix for #3504. The fix is correct, well-scoped, and well-tested.

## Correctness

**Snapshot timing is right (no off-by-one).** The placement snapshot is captured at the *same* program point as the routes snapshot — line 448 (routes) / line 456 (placement) — which runs immediately after routing and **before** that iteration's move is applied at the bottom of the loop body. So `best_placement_snapshot` always describes the placement that `best_routes_snapshot` was computed against:
- Iter-0-best / iter-1-regresses: iter-0 snapshot captures U1 pre-move; restore reverts U1 → original. ✓
- Iter-1-best (improving move): the move applied at the bottom of iter-0 is already reflected when iter-1's snapshot is taken, so the snapshot correctly pairs iter-1 routes with the moved placement. No restore fires (`best == current`), so the winning move is retained. ✓

**Pre-loop seed path is consistent.** `best_placement_snapshot` is captured unconditionally at loop entry (line 354), before any iteration runs, so the #3474 seed pairs the input placement with the pre-loop routes. No move occurs between that capture and the seed assignment, so no re-capture is needed.

**Restore is complete.** `_restore_placement` writes both `position` and `rotation` (rotation guarded by `hasattr`) for every ref in the snapshot. `_snapshot_placement` captures all footprints. `placement_diff` is built (line 629) *after* restoration (line 606), so it reflects the restored placement.

## Test verification

- **Regression test is load-bearing.** I disabled the `_restore_placement` call and confirmed `test_restored_routes_match_restored_placement` and `test_placement_diff_consistent_with_restored_placement` both FAIL (U1 left at moved position). Re-enabling the fix makes them pass.
- **Over-revert guard is genuine.** `test_improving_move_kept_when_later_iteration_is_best` passes with and without the fix — it correctly asserts a winning move is NOT spuriously reverted, guarding the improving-move branch independently of the restore logic.

## Test Execution

1. `pytest tests/test_placement_feedback.py` (worktree source on PYTHONPATH) — ✅ **61 passed** (3 new), matches PR claim.
2. Load-bearing check (fix disabled) — ✅ 2 of 3 new tests fail as expected; restored cleanly afterward.

## CI Status

All 8 checks pass (Test, Type Check, Lint & Format, C++ Build, Diff-Pair Regression, Match-Group Regression, Routed PCB DRC, kicad-cli Smoke). Merge state: **CLEAN**. PR uses `Closes #3504`.

Scope is disciplined — only `placement_feedback.py` (+76, helpers + restore wiring) and tests (+200). Clean approval.

PR #3640, opened 2026-06-13
test(router): realign stale softstart tests to post-#3481 kct-route recipe

### LINKED ISSUE #3492: tests: softstart slow-gated baseline + call-site tests stale since PR #3481 recipe overhaul

rjwalters:
## Summary

PR #3481 (issue #3343 P-R1..P-R4) overhauled the softstart recipe: it
no longer routes in-process via `router.route_with_escape(...)` but
shells out to the `kct route` CLI (the board-05 #3425/#3472 production
pattern). That left two softstart test files stale. This PR brings them
in line with the post-#3481 reality. **Test-only change** — no recipe or
router source modified.

## Changes

- **`tests/test_router_dense_escape_softstart.py`** — the
  `TestSoftstartCallSite` guard asserted the old in-process
  `router.route_with_escape(` call site and was masked behind a stale
  `@pytest.mark.xfail` (which itself referenced #3492). Removed the
  xfail and rewrote the assertion to pin the production subprocess
  invocation (`"-m", "kicad_tools.cli", "route"` run via
  `subprocess.run(`). Kept a regression guard against the #3138
  in-process `router.route_all_negotiated(` call creeping back.
- **Retired `tests/router/test_softstart_manufacturable_baseline.py`** —
  it pinned the obsolete #3235-era 2-layer/10-signal-net config
  (`--no-auto-layers --layers 2`), which no longer describes the rev B
  4-layer/26-net board. Run under its slow gate it now **times out at
  480s** (the 2-layer route cannot converge on the rev B board), erroring
  before reaching its assertions. The rev B reach floor is already
  covered by `test_softstart_revb_fine_pitch_escape.py` (20/26 at L=4,
  `jlcpcb-tier1`, 0.20mm clearance), so re-pinning this harness would
  only duplicate that coverage at extra CI cost.
- Repointed the two sibling docstrings
  (`tests/test_auto_pcb_size_softstart_smoke.py`,
  `tests/router/test_softstart_revb_auto_pcb_size.py`) that cited the
  retired file as a slow-gate-convention example to the live rev B
  harness.

| Re-pin or retire the #3235-era slow baseline harness | Done | Retired the file; rev B successor `test_softstart_revb_fine_pitch_escape.py` already pins the current L=4/26-net floor. Confirmed the stale harness times out at 480s on current main (2-layer route never converges on rev B board). |
| Update call-site guard to assert the `kct route` subprocess invocation | Done | Replaced stale xfail with assertions for `"-m", "kicad_tools.cli", "route"` + `subprocess.run(`; verified these strings exist in `generate_design.py` (lines 4073/4135). |
| Tests pass for the right reasons (no xfail/skip masking) | Done | `tests/test_router_dense_escape_softstart.py`: 7 passed (was 6 passed + 1 xfail). |
| Test-only change; no recipe/router edits | Done | `git diff --stat` touches only `tests/`; an exploratory comment edit to `router/escape.py` was reverted to respect the scope boundary. |

## Test Plan

- `uv run pytest tests/test_router_dense_escape_softstart.py -q --no-cov` -> **7 passed** (previously 6 passed + 1 xfail).
- Collection check on all touched + sibling files -> **12 tests collected**, no errors.
- Confirmed the retired baseline harness errored with `subprocess.TimeoutExpired` after 480s under `KICAD_RUN_SLOW_SOFTSTART_REACH=1` (C++ backend built and available), demonstrating the stale 2-layer config is non-viable on the rev B board.

Closes #3492


rjwalters:
✅ **Approved!**

Test-only realignment of the stale softstart tests to the post-#3481 `kct route` recipe. Reviewed all four file changes against the production source on the PR head; verdict below.

## Verification

**Deletion of `test_softstart_manufacturable_baseline.py` (highest-risk change) — justified.**
- The deleted test pinned `--no-auto-layers --layers 2` with a 10-net denominator (`REQUIRED_NETS_TOTAL = 10`). The current rev B board is 4-layer / 26-net (confirmed: `test_softstart_revb_fine_pitch_escape.py` uses the `four_layer_sig_gnd_pwr_sig` stack and a 26-signal-net denominator). The 2-layer config is structurally incompatible with rev B — the route cannot converge, and the test's own `timeout=480` means it errors before reaching its assertions. Its 10/10 connectivity assertion is meaningless on the current board.
- Coverage is genuinely subsumed: `test_softstart_revb_fine_pitch_escape.py::test_softstart_revb_reach_floor` authoritatively pins the current reach floor (20/26 at L=4, `jlcpcb-tier1`, 0.20mm clearance). Re-pinning the old harness to rev B would only duplicate that. Issue #3492 explicitly permitted "re-pin **or** retire" — retire is the correct call here; no meaningful assertion is lost.

**Rewritten call-site guard — load-bearing, asserts the right thing.**
- Confirmed both asserted tokens exist in `boards/external/softstart/generate_design.py` on the PR head: `"-m", "kicad_tools.cli", "route"` (line 3173) and `subprocess.run(` (line 3193). 
- The #3138 regression guard `"router.route_all_negotiated(" not in text` correctly passes: `router.route_all_negotiated(` appears **0** times in the source. The only `route_all_negotiated`/`route_with_escape` occurrences are inside an A/B-comparison comment (lines 4038/4042) without the `router.` prefix or call paren, so they do not trip the guard. The guard *would* fire if the bare in-process call path crept back in — it is a real regression sentinel, not a tautology.

**Test-only — confirmed.** Diff touches only `tests/`. The exploratory `escape.py` edit mentioned in the PR body was reverted; that file is untouched on the PR head.

**Docstring repoints — accurate.** Both sibling files (`test_auto_pcb_size_softstart_smoke.py`, `test_softstart_revb_auto_pcb_size.py`) now cite the live `test_softstart_revb_fine_pitch_escape` harness instead of the retired file. No dangling references to the deleted file remain in `tests/`.

## Minor (non-blocking)
One stale prose reference to the deleted test survives in `src/kicad_tools/router/escape.py:1334`. It is correctly out of scope for this test-only PR. Tracked in #3641.

## Test Execution
- `uv run pytest tests/test_router_dense_escape_softstart.py -q --no-cov` — ✅ matches PR claim (main: 6 passed + 1 xfail; PR removes the xfail → 7 passed).
- Source-string verification (the load-bearing assertions) — ✅ executed against PR head, all confirmed.
- Slow-gated 480s timeout reproduction — ⏭️ Skipped: long-running (>2 min); builder's claim independently corroborated by the obsolete 2-layer/10-net config being structurally incompatible with the 4-layer/26-net rev B board.

## CI
All checks pass; merge state CLEAN.
- C++ Build Check, Lint & Format, Type Check, Test, Routed PCB DRC, Diff-Pair Regression, Match-Group Regression, kicad-cli Round-trip — all pass.

PR closes #3492 with proper auto-close syntax.

PR #3488, opened 2026-06-11
fix(router): relief-probe rescue + C++ rip-up parity + corridor reservation (refs #3438)

rjwalters:
## Summary

Lands the infrastructure half of #3438 (board 07 bus-reversal reach gap) with hard A/B evidence, at **exact reach parity with main (30/31)** and no cross-board regression. Does NOT close #3438: the last net (DQ3) is a structural-capacity problem that needs the crossing-aware river-routing planner (issue direction 3) — measured signature below.

### What landed

1. **C++ rip-up parity fix** (latent #3101-pattern bug): ripped routes were NEVER unmarked on the paired C++ grid — phantom blockage accreted across negotiated iterations. `RoutingGrid.unmark_route` now mirrors the unmark with the same radius math as the mark path (`_unmark_route_on_cpp_cells`).
2. **Iteration-scoped corridor reservation** (`begin/flush_cpp_unmark_deferral`): the accidental phantom blockage was load-bearing — during the serial cohort reroute it stopped early-rerouted nets from stealing a not-yet-rerouted sibling's corridor. The naive parity fix alone dropped board 07 from 30/31 to 26/31 (measured). The deferral keeps cohort phantoms for the rest of the iteration and flushes at the next iteration top: intra-iteration reservation preserved, cross-iteration accretion fixed.
3. **Relief-probe rescue** (issue defects 1+2: foreign escape stubs/halos HARD in sharing mode ⇒ instant zero-overflow abort, no negotiation signal): relief mode in BOTH backends (`Pathfinder::relief_mode_` / `Router.set_relief_mode`) makes foreign non-obstacle copper passable at a per-step penalty; the min-conflict probe's crossed owner nets feed a **strictly transactional** rescue — commit only when the failed net routes in NORMAL mode (validated) AND every displaced victim re-lands (multi-pass + depth-1 nested rescues); otherwise verbatim rollback. Runs LAST (after the standard stall fallbacks), attempt-capped, `KCT_DISABLE_RELIEF=1` escape hatch. `find_relief_conflict_nets` extracts victims grid-side.
4. **Geometry-aware matrix layer preferences**: pad-centroid ordering along the group's dominant axis replaces sorted-net-id parity (the chaos amplifier — removing one net used to flip every member's layer preference). Pad-less synthetic groups keep id order (legacy tests unaffected).

### Measurement matrix (board 07 production recipe, seed 42, solo, same machine)

| config | reach | notes |
|---|---|---|
[...]

PR #3642, opened 2026-06-13
fix(router): reset cpp-unmark deferral flag on every route_all_negotiated exit

### LINKED ISSUE #3489: Harden corridor-reservation deferral window against exceptions escaping route_all_negotiated (follow-up to #3488)
## Context

PR #3488 (refs #3438) introduced the iteration-scoped C++ unmark-deferral window (`RoutingGrid.begin_cpp_unmark_deferral` / `flush_cpp_unmark_deferral`), opened at the full rip-up site in `Autorouter.route_all_negotiated` and flushed at (a) the top of every iteration and (b) post-loop via `_flush_corridor_reservation`.

Review of the control flow confirmed there are no early `return`/`raise` paths between the window open and the two flush points, so normal control flow (break/continue/timeout) always closes the window. However, there is **no try/finally**: if an unexpected exception (including KeyboardInterrupt) escapes `route_all_negotiated` while a window is open, the grid object is left with `_cpp_unmark_deferred = True`.

## Impact

- Low in the CLI flow (the Autorouter/grid is normally discarded after an escaped exception).
- Real in library/long-lived use: any subsequent `unmark_route` on the same grid silently queues the C++ mirror instead of applying it, diverging Python/C++ grid state until the next `route_all_negotiated` call flushes at its first iteration top (self-healing, but with stale queued routes in between).

## Suggested fix

Wrap the negotiated iteration loop (or the whole method body from the first possible `begin_cpp_unmark_deferral`) in `try/finally` with `self._flush_corridor_reservation(net_routes)` in the `finally`, making the post-loop flush unconditional. The flush is already a no-op when no window is open, so this is safe.

## Acceptance criteria

- [ ] An exception raised inside the rip-up/reroute loop while a deferral window is open leaves the grid with `_cpp_unmark_deferred == False` after `route_all_negotiated` unwinds.
- [ ] Unit test simulating a raise between `begin_cpp_unmark_deferral` and the iteration-top flush.
- [ ] No behavior change on the normal (non-raising) path: board 02/07 baselines unchanged.

Discovered during Judge review of PR #3488.

rjwalters:
## Summary

Hardens the corridor-reservation / cpp-unmark deferral window in
`Autorouter.route_all_negotiated` against exceptions. The negotiated
iteration loop opens a deferral window (`begin_cpp_unmark_deferral`) at
the rip-up site and relied on an inline post-loop flush to close it.
There was no `try/finally`: if an unexpected exception escaped the loop
while a window was open, the grid was left with
`_cpp_unmark_deferred=True`, silently queueing every subsequent
`unmark_route` on a long-lived grid and diverging the Python/C++ grid
state. Follow-up to #3488.

## Changes

- `src/kicad_tools/router/core.py`: wrap the negotiated iteration loop
  and the post-loop best-state restore in `try/finally`. The
  `finally` now holds the existing `self._flush_corridor_reservation(net_routes)`
  call, making the flush unconditional. The flush is already a no-op
  when no window is open, so the normal path is unchanged (the
  restore-then-flush ordering is preserved).
- `tests/router/test_deferral_window_exception_safety_3489.py`: new test
  module. One test forces an exception inside the loop while a deferral
  window is open and asserts the flag is reset afterward; a second pins
  the normal-path contract.

The semantic diff (ignoring whitespace) is just `try:` + converting the
inline flush comment to a `finally:` block; the rest of the large diff
is pure re-indentation of the loop body.

| Exception inside loop with window open leaves `_cpp_unmark_deferred == False` after unwind | Done | `test_exception_inside_open_window_resets_deferral_flag` passes; verified it FAILS on unmodified core.py (catches the regression) |
| Unit test simulating a raise inside the open-window loop | Done | Injects via `should_terminate_early` (called only inside the loop, inside the guarded `try`); opens a real window then raises |
| No behavior change on the normal (non-raising) path | Done | `finally` holds the same flush call after the restore block; flush is a no-op when no window is open; `test_normal_completion_leaves_deferral_flag_reset` + negotiated-loop / grid-parity suites pass |

## Test Plan

- `pytest tests/router/test_deferral_window_exception_safety_3489.py` -> 2 passed
- Confirmed the exception test fails on unmodified `core.py` (non-vacuous).
- `pytest tests/test_best_metric_patience.py tests/router/test_grid_resync_3507.py tests/test_grid_cpp_parity.py tests/router/test_byte_lane_corridor_reservation.py tests/test_negotiated_convergence_stranded.py` -> 52 passed
- Board 02 baseline: 2 failures (`test_drc_clean_at_jlcpcb_tier1`, `test_routing_output_deterministic_across_seeds`) confirmed PRE-EXISTING on unmodified `core.py` (fail identically with the change stashed) -- not introduced by this change.
- Ruff: 23 pre-existing errors in `core.py`, identical count before/after, none in the touched lines; new test file is clean.

Closes #3489

### DIFF
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -7344,954 +7344,1213 @@ class Autorouter:
             )
 
         # Skip iteration loop if already timed out
-        if not timed_out:
-            for iteration in range(1, max_iterations + 1):
-                full_reorder_used_this_iter = False
-                if check_timeout():
-                    print(f"\n  ⚠ Timeout reached at iteration {iteration} ({elapsed_str()})")
-                    timed_out = True
-                    break
+        try:
+            if not timed_out:
+                for iteration in range(1, max_iterations + 1):
+                    full_reorder_used_this_iter = False
+                    if check_timeout():
+                        print(f"\n  ⚠ Timeout reached at iteration {iteration} ({elapsed_str()})")
+                        timed_out = True
+                        break
 
-                # Issue #2540 + #2803: Snapshot state at top of each iteration
-                # BEFORE any destructive ``rip_up_nets`` call.  If a
-                # mid-iteration timeout escapes the per-net reroute loop, the
-                # post-loop restore can fall back to this pre-rip-up snapshot
-                # of the previous iteration's stable result.
-                #
-                # Comparison is by the lex tuple ``(routed_count desc,
-                # overflow asc, iteration desc)`` so an iteration that
-                # produced equal route count with lower overflow is still
-                # preserved.
-                current_routed = sum(1 for r in net_routes.values() if r)
-                # Issue #3002 (PR #3006 follow-up): include clearance-
-                # violation count in the lex tuple at the top-of-iteration
-                # snapshot site too -- matches the end-of-iteration capture
-                # below.
-                # Issue #3002 (PR #3006 perf): cache_key for top-of-
-                # iteration snapshot.  State here equals the end-state
-                # of the prior iteration -> reuse the ``("post", K-1)``
-                # cache from the previous _capture_iteration_end call.
-                # Issue #3020: combine both directions of the
-                # clearance matrix in the lex tuple comparator.
-                # Issue #3077: extend the via/segment universe with
-                # escape-phase routes.
-                _extra_top = self._collect_extra_routes_for_revalidation(net_routes)
-                current_seg_via = neg_router.find_nets_with_segment_via_violations(
-                    net_routes, trace_clearance=self.rules.trace_clearance,
-                    cache_key=("post", iteration - 1),
-                    extra_routes=_extra_top,
-                )
-                current_via_seg = neg_router.find_nets_with_via_segment_violations(
-                    net_routes, trace_clearance=self.rules.trace_clearance,
-                    cache_key=("post", iteration - 1),
-                    extra_routes=_extra_top,
-                )
-                # Issue #3433: seg-seg parity at the top-of-iteration
-                # snapshot site (matches the end-of-iteration capture).
-                current_seg_seg = neg_router.find_nets_with_segment_segment_violations(
-                    net_routes, trace_clearance=self.rules.trace_clearance,
-                    cache_key=("post", iteration - 1),
-                    extra_routes=_extra_top,
-                )
-                current_violations = (
-                    len(current_seg_via) + len(current_via_seg) + len(current_seg_seg)
-                )
-                # Issue #3117: mirror the lex-tuple primary key at the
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the deferral-window exception-safety fix. The change is correct and minimal.

## Control-flow verification (primary concern)
- The whitespace-ignoring diff (`git show -w`) confirms the semantic change is exactly two hunks: a `try:` before the `if not timed_out:` block (line 7347) and an inline post-loop flush converted to a `finally:` block (line 9196). The flush call `self._flush_corridor_reservation(net_routes)` is byte-identical, just relocated.
- The guarded region (7347–9207) wraps the whole negotiated iteration loop plus the post-loop best-state restore. **Zero `return` statements** exist anywhere in that region — only loop-scoped `break`/`continue` — so the re-indentation could not have altered any early-exit path. Control flow is preserved exactly.

## No double-flush / normal path unchanged
- `_flush_corridor_reservation` returns immediately when `flush_cpp_unmark_deferral()` yields nothing (`if not flushed: return`) — a genuine no-op when no window is open.
- On the normal path the per-iteration in-loop flush (line 7542) already closes each window, so the new `finally` flush is a no-op. Restore-then-flush ordering preserved; no double-flush.

## Test correctness
- Injection at `should_terminate_early` is genuinely inside the guarded window: it is called only at line 7452 (inside the loop, inside the `try`). The builder's reasoning to avoid `_collect_extra_routes_for_revalidation` is verified correct — that helper is called at line 7097 (before the `try:` at 7347), which would fire outside the window and make the test vacuous.
- The test opens a real window (`begin_cpp_unmark_deferral`, asserts flag True) before raising, and asserts both `state["raised"] is True` (non-vacuous) and the flag reset post-unwind. The second test pins the normal path.

## Pre-existing issues confirmed (not introduced by this PR)
- core.py: exactly 23 ruff errors, **none in the touched lines** (7340–7360, 9190–9210); new test file is ruff-clean. CI Lint & Format is green.
- The board02 baseline failures are pre-existing per the test plan; the CI Test job passes.

## Test Execution
- `tests/router/test_deferral_window_exception_safety_3489.py` + negotiated/parity regression suite (`test_best_metric_patience`, `test_grid_resync_3507`, `test_grid_cpp_parity`, `test_byte_lane_corridor_reservation`, `test_negotiated_convergence_stranded`) — ✅ 51 passed, 1 skipped locally (C++ backend available).

## CI / merge state
- All 8 CI checks pass (Test, Lint & Format, Type Check, C++ Build, DRC, both routing regressions, kicad-cli smoke).
- `mergeStateStatus`: CLEAN.

PR description uses `Closes #3489`. Solid, surgical fix.

PR #3643, opened 2026-06-13
fix(router): missed via-in-pad rescues override dominant-cause veto in --auto-mfr-tier

### LINKED ISSUE #3463: route --auto-mfr-tier: escalation aborts before jlcpcb-tier1 attempt on board 04 (3 slow tests failing)

rjwalters:
## Summary

`kct route --auto-mfr-tier` aborted its escalation ladder on board 04 before ever attempting `jlcpcb-tier1`, so the via-in-pad-capable tier that fixes the LQFP-48 inner-pin escapes never ran. Three slow integration tests (`test_route_auto_mfr_tier_integration.py`) were xfail'd to land the slow-tests gate (#3458); this PR fixes the root cause and removes the xfails.

There were two compounding bugs, both fixed without touching `core.py::route_all_negotiated` (a sibling PR owns that method):

1. **Budget starvation.** The mfr-tier loop checks `_deadline_expired` at the top of each tier, but the inner `route_with_layer_escalation` divides the *full* remaining routing budget across its own layer attempts. The base tier therefore expanded to fill the entire routing deadline, and the loop broke before tier 1.

2. **Trigger-table veto precedence.** On board 04 a single keepout-blocked net makes `BLOCKED_PATH` the most-common per-net failure, so `_classify_dominant_failure_cause` returns `BLOCKED_PATH` and the #2883 trigger table vetoes escalation -- even though the `EscapeRouter` separately recorded a non-zero `missed_via_in_pad_rescues` for the fine-pitch pins, which is exactly what `jlcpcb-tier1` fixes.

## Changes

- New module `src/kicad_tools/cli/mfr_tier_budget.py`: `per_tier_routing_budget` context manager + `per_tier_routing_deadline` helper that narrow `args._routing_deadline` to a fair per-tier slice (floor-guarded, rolls unused budget forward) and restore the original deadline on exit so auto-fix keeps its reserve.
- `route_with_mfr_tier_escalation` (`cli/route_cmd.py`): wrap the inner layer-escalation call in the per-tier budget window; reorder the convergence guard so a non-zero `missed_via_in_pad_rescues` + via-in-pad capability gain escalates *before* the dominant-cause veto is consulted. The veto still suppresses escalation when there is no instrumented missed-rescue signal (legacy #2883 placement-class behavior preserved).
- Removed the 3 xfail markers in `test_route_auto_mfr_tier_integration.py`.
- Updated the `BLOCKED_PATH` / `CONGESTION` veto unit tests to set `missed_via_in_pad_rescues = 0` (the pure-veto case) and added `test_missed_via_in_pad_overrides_blocked_path_veto` pinning the new precedence.
- New fast unit suite `tests/test_mfr_tier_budget.py` (10 tests) for the budget arithmetic and restore semantics.

| Escalation attempts `jlcpcb-tier1` on board 04 | done | `test_escalation_advances_to_tier1` passes: `Tier N/M: jlcpcb-tier1` banner now appears |
| tier1 routes more nets than jlcpcb in the same run | done | `test_tier1_routes_more_nets_than_jlcpcb` passes |
| Escalation triggered by canonical missed-via-in-pad signal | done | `test_escalation_triggered_by_missed_via_in_pad` passes |
| 3 xfail markers removed | done | markers deleted; tests run as normal slow tests |
| escalation ladder fix outside `core.py::route_all_negotiated` | done | only `cli/route_cmd.py` + new `cli/mfr_tier_budget.py` changed |

## Test Plan

- `pytest tests/test_route_auto_mfr_tier_integration.py -m slow --no-cov` -> **3 passed in 457.65s** (was 3 failed). These are `@pytest.mark.slow`, excluded from PR-time CI and picked up by the nightly slow-tests workflow.
- `pytest tests/test_route_auto_mfr_tier.py tests/test_mfr_tier_ladder.py tests/test_auto_mfr_tier_log_suppression.py tests/test_mfr_tier_budget.py --no-cov` -> 83 passed (fast).
- `ruff check` + `ruff format --check` clean on all changed files.
- C++ backend built (`uv run kct build-native`) before running the slow chain.

Closes #3463

### DIFF
--- a/src/kicad_tools/cli/mfr_tier_budget.py
+++ b/src/kicad_tools/cli/mfr_tier_budget.py
@@ -1,0 +1,149 @@
+"""Per-tier wall-clock budget slicing for ``--auto-mfr-tier`` escalation.
+
+Issue #3463: the manufacturer-tier escalation ladder
+(``route_with_mfr_tier_escalation`` in :mod:`kicad_tools.cli.route_cmd`)
+walks a sequence of tiers (e.g. ``jlcpcb -> jlcpcb-tier1``).  For each
+tier it re-enters the layer-escalation path, which divides the *remaining
+routing budget* across its own layer attempts via
+``_per_attempt_budgeted_timeout``.
+
+The bug: that inner helper sees the **full** remaining routing budget at
+base-tier time, so the base tier's layer escalation expands to fill the
+entire routing deadline.  When control returns to the outer ladder, the
+routing deadline is already exhausted, ``_deadline_expired(args)`` is
+``True`` at the top of the next tier iteration, and the loop breaks
+*before* attempting ``jlcpcb-tier1`` -- exactly the symptom reported in
+#3463 (per-tier banner only ever shows ``['jlcpcb']``, escalation aborts
+on the base-tier escape-infeasibility diagnosis instead of advancing to
+the via-in-pad-capable tier).
+
+The fix lives here, deliberately out of
+:meth:`kicad_tools.router.core.Autorouter.route_all_negotiated` (a sibling
+PR touches that method).  We expose a small context manager that
+temporarily narrows ``args._routing_deadline`` to a fair per-tier slice of
+the *remaining* routing budget, so each tier -- including the final one --
+gets a real chance to run.  The original deadline is restored on exit so
+auto-fix and any subsequent CLI reuse of ``args`` see the unmodified
+total.
+
+Design notes:
+
+* The slice rolls forward naturally: a fast-finishing tier leaves more
+  budget for the next, because the slice is recomputed from the *current*
+  remaining budget divided by the *current* number of outstanding tiers.
+* When no wall-clock deadline is configured (legacy unbounded runs, or
+  ``--timeout`` unset) the context manager is a no-op and the inner path
+  keeps its existing unbounded behaviour.
+* The narrowing is intentionally a *floor-guarded* slice: we never narrow
+  the per-tier deadline below a small floor, so a ladder with many tiers
+  and a tight ``--timeout`` does not collapse the final tier to zero
+  budget.  The floor matches the per-attempt floor philosophy used by the
+  layer-escalation budget helper.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import time
+from collections.abc import Iterator
+
+# Minimum per-tier routing budget (seconds).  Mirrors the spirit of the
+# auto-fix reserve floor: a tier given less than this much time cannot
+# make meaningful progress, so we never slice below it.  If the total
+# remaining budget is itself below the floor, the slice collapses to the
+# remaining budget (the tier gets whatever is left).
+_PER_TIER_FLOOR_SEC = 30.0
+
+
+def per_tier_routing_deadline(
+    args,
+    *,
+    tier_index: int,
+    tier_count: int,
+    now: float | None = None,
+) -> float | None:
+    """Compute a fair per-tier routing deadline (monotonic clock value).
+
+    Divides the *remaining* routing budget evenly across the *remaining*
+    tiers (including the current one) and returns the monotonic deadline
+    for the current tier: ``now + slice``.
+
+    Args:
+        args: Parsed CLI namespace.  Must carry ``_routing_deadline`` (set
+            by ``_set_wall_clock_deadline``); when that is ``None`` this
+            returns ``None`` (legacy unbounded behaviour).
+        tier_index: 0-based index of the current tier in the ladder.
+        tier_count: Total number of tiers the ladder intends to attempt.
+        now: Override for the current monotonic time (testing seam).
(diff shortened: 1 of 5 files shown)


rjwalters:
✅ **Approved!**

Both compounding bugs are correctly diagnosed and fixed, and I verified the fix end-to-end rather than relying on PR CI alone (the 3 acceptance tests are `@pytest.mark.slow` and excluded from PR CI).

## Code review

**Bug 1 — budget starvation (`cli/mfr_tier_budget.py`):** The `per_tier_routing_deadline` arithmetic is sound: it slices the *remaining* budget over the *outstanding* tier count (so unused budget rolls forward), floor-guards at 30s (collapsing only when total remaining is itself below the floor), hands the last outstanding tier the full remaining budget, and returns the original deadline unchanged when already expired or unconfigured. The `per_tier_routing_budget` context manager restores `args._routing_deadline` unconditionally in `finally` (correct even when `original is None`), so there is no deadline leak into auto-fix's reserve. No-op when unbounded.

**Bug 2 — veto precedence (`cli/route_cmd.py`):** The reorder checks `gains_capability and triggered_by_missed_in_pad` *before* the trigger-table veto. I scrutinized the over-escalation risk: the override is gated by `gains_capability = can_escalate_via_in_pad(prev, next)`, which is True only when the next tier actually adds via-in-pad and the current one lacks it (`mfr_limits.py:435`). Combined with a non-zero instrumented `missed_via_in_pad_rescues`, this is direct evidence the next tier's new capability fixes the failure — strictly stronger than the statistical per-net tally. It does not blanket-override the veto: when there is no missed-rescue signal, the `elif trigger_table_vetoes` branch still suppresses escalation (legacy #2883 placement-class behavior preserved). The board-04 shape (`[BLOCKED_PATH, BLOCKED_PATH, PIN_ACCESS]` with a coincidental keepout net dominating the tally) is exactly the case this rescues.

**Test migration is legitimate, not weakened:** The two updated veto tests (`BLOCKED_PATH`, `CONGESTION`) now set `missed_via_in_pad_rescues = 0`, isolating the *pure-veto* contract — the `seen_tiers == ["jlcpcb"]` assertion is preserved, only the input was changed to exercise the no-signal path that still vetoes. The "signal present" scenario is now covered by the new `test_missed_via_in_pad_overrides_blocked_path_veto` (signal=3, BLOCKED_PATH dominant -> escalates to tier1). This is a correct contract split, not an assertion relaxed to pass. `core.py::route_all_negotiated` confirmed untouched.

## Verification

- **Slow acceptance test run locally:** `test_escalation_advances_to_tier1` (one of the 3 de-xfail'd tests) — **1 passed in 425.23s** on the worktree with the C++ backend built (`C++ backend: available (version 1.0.0)`). Confirms escalation now advances to `jlcpcb-tier1` on board 04.
- **Fast suites:** `tests/test_mfr_tier_budget.py` + `tests/test_route_auto_mfr_tier.py` — **52 passed** (includes the 10 new budget tests and the new precedence test).
- **CI:** all 9 checks green (Test, Type Check, Lint & Format, C++ Build, DRC, both routing regressions, round-trip smoke).
- **Merge state:** CLEAN.
- PR description uses `Closes #3463`.

Solid root-cause fix with thorough test coverage at both the unit and integration level.

Issue #2901, opened 2026-05-14
Test: board 05 regression — guard zones + thermal stitching + DRC allowlist + clean export (re #2746 child 4)


rjwalters:
## Summary

This test guards the regression surface that PRs #2903 (thermal-via stitching) and #2904 (board-05 zones pipeline) just shipped. It does **NOT** gate on net-completion percentage — that's #2906's territory under the #2720 placement-GA epic.

Original framing pinned this test to ALL of #2746's ACs including ≥38% routed. With #2898 closed as not-planned (the algorithmic path moved to #2720/#2906), this child is reframed as a "protect what's working today" regression guard.

See the 2026-05-15 reframe comment for the full delta: https://github.com/rjwalters/kicad-tools/issues/2901#issuecomment-4456608461

## Acceptance criteria

1. **Zone count regression guard.** Running the board-05 build produces a routed PCB with ≥4 filled zones. (Already covered by `tests/test_board_05_zones.py::test_routed_pcb_has_min_zones` from #2904 — verify it still asserts what we want here and is wired into the relevant CI lane; extend rather than duplicate.)
2. **Required zone nets present.** VMOTOR, +5V, +3.3V, GND must each own a zone in the routed PCB (PWR_LED tolerated but not required). Already covered by `tests/test_board_05_zones.py::test_routed_pcb_zones_include_required_nets` from #2904 — verify present; do not duplicate.
3. **DRC allowlist guard.** Board-05's committed `bldc_controller_routed.kicad_pcb` reports ≤ 53 DRC errors under JLCPCB rules (the current allowlist value in `.github/routed-drc-tolerance.yml`). This issue does **not** tighten the allowlist — that requires real router improvements. The test should assert the count is at-or-below the allowlist so a regression (e.g., 36 → 60) trips CI even if it's still within tolerance.
4. **Thermal-via stitching on Q1-Q6.** When board-05's pipeline runs `kct stitch --thermal` (per #2903), each of the six DRV8313 MOSFET heat-sink pads (Q1-Q6) must carry ≥4 thermal vias. This is the board-05-specific integration assertion on top of the unit tests in `tests/test_thermal_stitch.py`.
5. **Export preflight passes.** Running export against the routed PCB produces exit code 0 with no BOM↔PCB drift and every component carrying an LCSC part + footprint. The test should invoke the preflight pathway end-to-end (or the `--strict-preflight` mode) on `bldc_controller_routed.kicad_pcb` and assert clean.

[...]

PR #3644, opened 2026-06-13
fix(zones): carve over-claiming pour against siblings instead of collapsing to pad-safe

### LINKED ISSUE #3443: auto-pour: #3240 pad-safe downgrade shrinks board-05 +24V pour, breaking thermal-stitch coverage on fresh regens
## Summary

Any fresh zone regeneration of board 05 (and likely other multi-pour boards) shrinks the +24V pour outline from its 1.5mm-margin pad bbox to the 0.3mm pad-safe bbox, which breaks the thermal-stitch acceptance from #2901 (>=4 vias per HS MOSFET tab).

## Mechanism

`src/kicad_tools/zones/generator.py` (#3240 carve logic, ~line 1379): when a high-priority net's RAW bbox (1.5mm margin, `DEFAULT_POUR_BBOX_MARGIN_MM`) fully contains a lower-priority sibling's pad-safe bbox, the high-priority net's OWN outline is downgraded to its pad-safe bbox (`PAD_SAFE_MARGIN_MM = 0.3`). On board 05 the +24V raw bbox covers lower-priority siblings, so +24V — the board's most thermally critical pour — shrinks to 0.3mm beyond its pad extremes.

Measured outlines (F.Cu +24V):
- committed artifact (pre-#3240-era): x 101.5-141.5, y 105.2-169.5
- fresh regen with current main:      x 102.7-140.3, y 106.4-168.3 (uniform 1.2mm shrink)

## Consequence

`run_thermal_stitch` halo candidates (thermal_radius 2.2, zone edge margin 0.35) around the HS MOSFET drain tabs Q1/Q3/Q5 (which sit AT the +24V pad-bbox extremes, y=168 / x=140 board-abs) fall outside the shrunken pour or miss the edge margin. Result: Q1=2, Q3=2, Q5=1 vias (need >=4). `tests/test_board_05_thermal_stitch.py::test_each_mosfet_heat_sink_gets_min_vias` fails on ANY fresh regen.

## Pre-existing proof (independent of #3423)

Re-running `auto_pour_if_missing` + `run_thermal_stitch` on the PRE-rotation committed unrouted PCB (git e4c1e057~1) with current main code reproduces the identical shrunken outline and identical Q1=2/Q3=2/Q5=1 shortfall. The U3 rotation (#3423) does not cause or affect this.

## Suggested fix directions

1. Downgrade the winner's SUBTRAHEND only (let lower siblings carve against pad-safe) but keep the winner's own outline raw minus lower siblings' pad-safe bboxes (KiCad zone outlines cannot have holes, but an L/staircase polygon union is possible), or
2. Use per-direction shrink: only pull in the edges that actually over-claim a sibling's pads, or
3. Exempt the highest-priority net on the layer from self-downgrade when it carries the board's power-stage thermal pads.

## Refs

- #3423 (PR marks the thermal test xfail with this issue)
- #3240 (original carve fallback), #2901 (thermal stitch AC), #3043

rjwalters:
## Summary

The #3240 carve fallback in `zones/generator.py` collapsed a high-priority
pour's **entire** outline to its tight 0.3 mm pad-safe bbox whenever its
1.5 mm raw bbox enclosed a lower-priority sibling's pads. On board 05 the
highest-priority pour (+24V) over-claims an interleaved sibling (+5V), so
+24V was uniformly shrunk to pad-safe — **including at its own exclusive
extremes** where the HS-MOSFET drain tabs (Q1/Q3/Q5) live. That starved the
thermal-stitch halo around those tabs of copper, so on every fresh zone
regen Q1=2, Q3=2, Q5=1 vias (need ≥4), regressing issue #2901 AC #4.

## Changes

- **`src/kicad_tools/zones/generator.py`** — replace the whole-outline
  collapse with a targeted carve. The over-claiming net's effective outline
  becomes `raw - union(lower siblings' pad-safe bboxes)`. When a sibling
  touches/crosses the net's bbox edge (the common interleaved-rail case) the
  difference is a single hole-free concave staircase that keeps the full raw
  margin at the net's exclusive extremes while staying disjoint from every
  sibling. When a sibling sits strictly interior with no edge-crossing
  neighbour (a true hole the carve can't represent), fall back to the #3240
  pad-safe bbox so the lower sibling still survives.
- **`boards/05-bldc-motor-controller/output/bldc_controller_routed.kicad_pcb`**
  — regenerate the +24V F.Cu zone with the corrected staircase outline
  (x[103.50,141.50] y[106.50,169.50], 16 pts vs the old 4-pt pad-safe
  rectangle), with all zone fills recomputed via `kicad-cli pcb fill-zones`.
- **`tests/test_board_05_thermal_stitch.py`** — drop the `xfail` marker on
  `test_each_mosfet_heat_sink_gets_min_vias`; it now passes.
- **`tests/test_zone_priority.py`** — add `TestWinnerExtremeCopperIssue3443`
  (winner-with-interior-sibling fixture) pinning the carve at the unit level:
  the winner keeps its raw-margin extremes and stays disjoint from both
  siblings. Both tests fail on pre-fix code.

| +24V pour retains copper at its thermal pads on fresh regen | ✅ | `_compute_pour_outlines` now returns the 16-pt staircase reaching x[103.5,141.5] y[106.5,169.5] (full raw extent) instead of the pad-safe x[104.7,140.3] y[107.7,168.3] |
| Q1/Q3/Q5 each get ≥4 thermal vias | ✅ | `run_thermal_stitch` dry-run on the regenerated artifact: Q1=4, Q2=4, Q3=4, Q4=4, Q5=4, Q6=4 |
| Does not reintroduce #3240 (lower siblings keep copper, no zero-copper) | ✅ | `tests/test_zone_priority.py` #3240 suite (nested-in-xy, pad-safe fallback, ZonePartitionError) all pass; interior-sibling carve still falls back to pad-safe |
| Disjoint-carve invariant (#3043) preserved | ✅ | New unit test asserts winner/sibling overlap < 1e-6; existing #3043 disjointness tests pass |

## Test Plan

- `pytest tests/test_board_05_thermal_stitch.py` — 3 passed (xfail removed; all 6 MOSFETs get 4 vias)
- `pytest tests/test_zone_priority.py` — 11 passed (2 new #3443 tests + existing #3043/#3240 guards)
- `pytest` board-05 DRC/drift/zones/export + auto_pour/zone_generator suites — 124 passed
- New #3443 unit tests confirmed to FAIL on pre-fix code (ZonePartitionError / pad-safe collapse), proving they guard the regression.

**Local validation note:** kicad-cli was available locally (`/Applications/KiCad/.../kicad-cli`), so the committed board-05 artifact's zone fills were recomputed and the thermal-stitch coverage was validated end-to-end locally — not deferred to CI. The full routed-PCB pipeline (router + ERC) was not re-run; CI's board-05 jobs should confirm the regenerated artifact passes DRC/ERC/sync as before.

Closes #3443


rjwalters:
✅ **Approved!**

Thorough review complete. The #3443 fix is correct and, critically, does **not** regress #3240's pad-clearance protection.

## Correctness — #3240 pad-safety preserved (the key concern)

Verified at the code level in `src/kicad_tools/zones/generator.py`:

- The carve `effective_outline = raw - union(lower siblings' pad-safe bboxes)` is only accepted when it is a hole-free single polygon. `_subtract_polygon` returns `None` whenever the difference has interior holes (lines 1646-1647), which is exactly the #3240 zero-copper hazard — the over-claiming exterior would overlap the contested region and the fill resolver would award it away. So a **strictly-interior** sibling (a true hole KiCad can't represent) correctly falls back to the #3240 pad-safe bbox.
- The over-claiming net's **own** outline (`self_outline`, line 1474) and the running `winners_union` (`subtrahend_for_winners`, line 1528) both use the same carved/pad-safe polygon, so the disjoint-carve invariant holds — the winner never silently overlaps a sibling.
- The union subtracts **all** lower siblings' pad-safe bboxes (the `break` was correctly removed), which keeps the winner disjoint from every sibling and lets each lower sibling reclaim its yielded pad region.

New unit test `test_winner_and_sibling_stay_disjoint` pins this directly: winner stays disjoint from both siblings (overlap < 1e-6) AND each sibling keeps positive copper (the #3240 guarantee). `test_winner_keeps_raw_margin_at_exclusive_extremes` proves the winner reaches the full raw margin at an extreme rather than collapsing to pad-safe.

I confirmed both new tests **FAIL on the pre-fix generator** and **PASS on the fixed code** — they are genuinely load-bearing.

## Artifact consistency — board-05 staircase regen

- `kct check` on the regenerated `bldc_controller_routed.kicad_pcb` shows the +24V `clearance_pad_zone` overlaps at ~0.013–0.018mm depth — these are within board-05's documented DRC allowlist.
- `tests/test_board_05_drc_allowlist.py` — 2 passed: the staircase regen introduced **no new out-of-allowlist violations**.
- CI "Routed PCB DRC Check" passes against the artifact.

## xfail removal

Correct. The marker was `strict=True`, so leaving it on a now-passing test would itself fail CI (XPASS). Removal is required.

## Test Execution (local, kicad-cli + shapely 2.1.2 available)

- `tests/test_zone_priority.py` + `tests/test_board_05_thermal_stitch.py` — ✅ 14 passed
- `tests/test_board_05_drc_allowlist.py` — ✅ 2 passed
- Combined `thermal_stitch + zone_priority + auto_pour + board_05_routing_regression + route_zones_preserved` — ✅ 63 passed
- New #3443 tests on pre-fix generator — ✅ confirmed 2 failed (regression guards are real)

## CI & Scope

- All 8 CI checks pass (C++ Build, Diff-Pair/Match-Group regression, Lint, Routed DRC, Test, Type Check, kicad-cli Round-trip).
- `mergeStateStatus`: CLEAN.
- PR diff scope is exactly the 4 expected files; "Closes #3443" present.

Solid, geometry-correct fix with strong regression coverage. Approving.

PR #3645, opened 2026-06-13
ci: gate Type Check job on a mypy baseline instead of masking failures

### LINKED ISSUE #3512: ci: Type Check job has continue-on-error (mypy rot is invisible, 1472 errors in 264 files on main)
## Problem

The `Run mypy` step in `.github/workflows/ci.yml` has `continue-on-error: true`, so the Type Check job is green regardless of mypy's exit code. Measured on main (2026-06-11, worktree at 15d270a8):

```
$ uv run mypy src/
Found 1472 errors in 264 files (checked 666 source files)
```

Same masking pattern as the Test job (#3436, fixed) and the Lint & Format job (#3464). New type errors land invisibly.

## Suggested approach

1. Burn down or baseline the 1472 errors (e.g., per-module `ignore_errors` in mypy config, or a baseline tool), so the gate can go live without a single mega-PR.
2. Remove `continue-on-error: true` from the `Run mypy` step.

## Acceptance

- `continue-on-error: true` removed from the mypy step
- Type Check job is red iff mypy reports new errors

Discovered while implementing #3436 (Test job gate); mypy un-masking was explicitly out of scope there because of the error volume.

rjwalters:
## Summary

The CI Type Check job's `Run mypy` step carried `continue-on-error: true`, so the job stayed green regardless of mypy's exit code. Type rot accumulated invisibly — **1512 errors in 269 files** on main (the issue measured 1472; drift since confirms exactly why the gate is needed). Fixing all of them in one mega-PR is infeasible.

This PR makes the job actually gate by mirroring the established `routed-pcb-drc-check` allowlist pattern: a committed baseline records the known debt, and a wrapper script fails CI **only on NEW errors** beyond the baseline.

## Changes

- **`scripts/ci/check_mypy_baseline.py`** (new) — runs `mypy src/`, parses errors into a multiset of *line-number-independent* signatures (`<path>\t<code>\t<number-masked-message>`), and diffs against the baseline. Exit 0 = within baseline, exit 2 = new errors (with `::error file=...::` annotations), exit 1 = tool failure. `--update` regenerates the baseline for burn-down.
- **`.github/mypy-baseline.txt`** (new) — committed debt ledger: 1512 errors / 912 unique signatures, with a header documenting the burn-down intent (regenerate via `--update` in the same PR when fixing errors; end state is an empty baseline and removal of the wrapper).
- **`.github/workflows/ci.yml`** — Type Check job's mypy step drops `continue-on-error: true` and now runs `uv run python scripts/ci/check_mypy_baseline.py`. **The Lint & Format job's `continue-on-error` is intentionally untouched** (sibling issue #3464).
- **`tests/test_ci_mypy_baseline.py`** (new) — 19 tests covering workflow wiring, the scope guard on the lint job, baseline file shape, signature normalization, and diff/exit-code semantics.

## Why a normalized baseline (not a raw line-keyed diff)

Keying on exact `path:line` would flag every refactor that shifts line numbers as hundreds of "new" errors, making the gate useless. Dropping the line number and masking volatile integers (while preserving quoted identifiers) yields a signature stable across unrelated edits, so the gate fails on genuinely new type errors and extra duplicates, not on code movement.

| `continue-on-error: true` removed from the mypy step | done | `git diff` of ci.yml; `test_typecheck_mypy_step_has_no_continue_on_error` |
| Type Check job is red iff mypy reports NEW errors | done | Injected a new error → exit 2 + annotation; baseline output → exit 0 (manual + `test_main_*`) |
| Baseline the debt (don't fix 1512 at once) | done | `.github/mypy-baseline.txt` with 1512 errors / 912 signatures |
| Burn-down intent documented | done | Header in baseline file + comments in ci.yml and the script |
| Test for the diffing script | done | `tests/test_ci_mypy_baseline.py` (19 tests) |
| Lint & Format job untouched (#3464 scope) | done | `test_lint_job_continue_on_error_untouched` |

## Test Plan

- `uv run pytest tests/test_ci_mypy_baseline.py` → 19 passed
- End-to-end: `uv run python scripts/ci/check_mypy_baseline.py` against a real `mypy src/` run → "OK: mypy reports 1512 error(s), all within the baseline. No new type errors." (exit 0)
- Injected a synthetic new error → exit 2 with `::error::` annotation
- Verified line-number shifts (+1000) still pass; an extra duplicate of an existing error fails

Closes #3512

### DIFF
--- a/.github/mypy-baseline.txt
+++ b/.github/mypy-baseline.txt
@@ -1,0 +1,1530 @@
+# Mypy baseline -- committed debt ledger for the CI Type Check gate (issue #3512).
+#
+# Each non-comment line is a NORMALIZED mypy error signature:
+#     <path>\t<error-code>\t<message-with-numbers-masked>
+# Line numbers are intentionally absent so the gate survives refactors that
+# shift code around.  scripts/ci/check_mypy_baseline.py compares the current
+# mypy run against this multiset and fails CI only on NEW errors.
+#
+# BURN-DOWN INTENT: this is debt, not a target.  When you fix type errors,
+# regenerate this file in the same PR:
+#     uv run python scripts/ci/check_mypy_baseline.py --update
+# so the floor ratchets down and the fixed errors cannot silently return.
+# The end state is an empty baseline and removal of the wrapper.
+#
+# DO NOT hand-edit. DO NOT add entries to silence a NEW error -- fix the type
+# error or, if it is a false positive, add a targeted ``# type: ignore[code]``
+# with a comment, which removes it from the report entirely.
+
+src/kicad_tools/acceleration/backend.py	import-not-found	Cannot find implementation or library stub for module named "cupyx"
+src/kicad_tools/acceleration/backend.py	no-any-return	Returning Any from function declared to return "bool"
+src/kicad_tools/acceleration/backend.py	no-any-return	Returning Any from function declared to return "ndarray[tuple[Any, ...], dtype[Any]]"
+src/kicad_tools/acceleration/backend.py	no-any-return	Returning Any from function declared to return "ndarray[tuple[Any, ...], dtype[Any]]"
+src/kicad_tools/acceleration/kernels/routing.py	assignment	Incompatible types in assignment (expression has type "int", variable has type "ndarray[tuple[Any, ...], dtype[signedinteger[_32Bit]]]")
+src/kicad_tools/acceleration/kernels/routing.py	no-any-return	Returning Any from function declared to return "ndarray[tuple[Any, ...], dtype[Any]]"
+src/kicad_tools/acceleration/kernels/routing.py	var-annotated	Need type annotation for "conflicts"
+src/kicad_tools/analysis/congestion.py	var-annotated	Need type annotation for "result" (hint: "result: list[<type>] = ...")
+src/kicad_tools/analysis/signal_integrity.py	arg-type	Argument N to "add" of "set" has incompatible type "tuple[int, ...]"; expected "tuple[int, int]"
+src/kicad_tools/assembly/validation.py	assignment	Incompatible types in assignment (expression has type "LCSCClient", variable has type "None")
+src/kicad_tools/assembly/validation.py	no-any-return	Returning Any from function declared to return "AssemblyValidationResult"
+src/kicad_tools/audit/auditor.py	assignment	Incompatible types in assignment (expression has type "ManufacturerProfile", variable has type "None")
+src/kicad_tools/audit/auditor.py	assignment	Incompatible types in assignment (expression has type "None", variable has type "Path")
+src/kicad_tools/calibration.py	assignment	Incompatible types in assignment (expression has type "list[float]", variable has type "ndarray[tuple[int], dtype[float64]]")
+src/kicad_tools/calibration.py	assignment	Incompatible types in assignment (expression has type "str", target has type "float")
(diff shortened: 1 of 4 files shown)


rjwalters:
✅ **Approved!**

Verified the mypy baseline-and-gate mechanism end-to-end. This is a clean mirror of the established `routed-pcb-drc-check` debt-baseline pattern and the gate is genuinely load-bearing.

## Gate is load-bearing (reproduced locally)

- `uv run python scripts/ci/check_mypy_baseline.py` against the committed baseline → **exit 0**: `"OK: mypy reports 1512 error(s), all within the baseline (1512 allowed). No new type errors."` Dropping `continue-on-error` will NOT make CI red on main's existing debt.
- Injected a synthetic new error (`def f() -> int: return "not an int"`) into a src file → **exit 2** with a `::error file=...::` annotation. Removed the injected file afterward.
- This matches the PR's claims: a +1000 line shift produces an identical signature (line number dropped — `test_signature_is_line_independent`), and an extra duplicate of an existing signature counts as new debt (`current - baseline` via `Counter` subtraction — `test_gate_fails_on_extra_duplicate`).

## Signature design is sound (not too loose, not too strict)

The signature is `<path>\t<code>\t<number-masked-message>`:
- **Masking `\b\d+\b` → `N`** correctly neutralizes benign drift (argument indices like `Argument 1` vs `Argument 3`, embedded counts) without flagging refactors.
- **Quoted identifiers are preserved**, so semantically-distinct errors stay distinct (`"labels"` vs `"global_labels"` → different signatures; `test_normalize_message_preserves_identifiers`). This is the key guard against a new error hiding behind an existing baseline entry.
- **Path + code are part of the key**, so a new error in file A can't be masked by a since-fixed error in file B, and distinct error codes never collide.
- **Multiset semantics** (`Counter`) count duplicates independently and clamp at 0, so fixing errors warns-but-passes (ledger ratchet) while net-new errors and extra duplicates fail.

The only theoretical looseness — a brand-new error producing the *exact* same path+code+integer-masked message as an under-counted baseline signature — is an extremely narrow case (the new error would have to be semantically identical modulo integers to a pre-existing same-file error) and is the correct, documented tradeoff for a baseline gate.

## Scope & correctness checks

- **ci.yml change is scoped to the Type Check job only**: only the `typecheck` job's mypy step changed (`continue-on-error: true` dropped, now runs the wrapper). The Lint & Format job's two `continue-on-error: true` steps are untouched — explicitly pinned by `test_lint_job_continue_on_error_untouched` (asserts exactly 2), correctly leaving #3464's domain alone.
- **Baseline committed & burn-down documented**: `.github/mypy-baseline.txt` present (1512 signatures + 17-line header documenting the `--update` ratchet and end-state of an empty baseline + wrapper removal). Same intent restated in the script docstring and the ci.yml comment.
- **Broken-mypy guard**: the wrapper returns exit 1 (tool failure) if mypy emits output but no parseable error lines and it doesn't look like a clean run — prevents a misconfigured mypy from masquerading as a green gate.

## Test Execution

- `./.loom/scripts/run-tests.sh tests/test_ci_mypy_baseline.py -q` → **19 passed**
- Gate on committed baseline → exit 0 (reproduced)
- Gate with injected new error → exit 2 + annotation (reproduced)

## CI Status

All 8 checks pass (Type Check now green at 51s **without** continue-on-error). Merge state: CLEAN. PR description uses `Closes #3512`.

Excellent work — the gate does exactly what it claims.

PR #3646, opened 2026-06-13
fix(router): keep same-component net=0 pad overlap routable in validator

### LINKED ISSUE #3490: router: test_route_to_pad_adjacent_to_net0_pad fails on main (pad blocked by adjacent net=0 pad)
## Symptom

`tests/test_router_core.py::TestPadBlockedByAdjacentNetZero::test_route_to_pad_adjacent_to_net0_pad` fails on current main:

```
FAILED tests/test_router_core.py::TestPadBlockedByAdjacentNetZero::test_route_to_pad_adjacent_to_net0_pad
AssertionError: Routing net 1 failed - target pad blocked by adjacent net=0 pad
(tests/test_router_core.py:243)
```

The test routes a 0.6mm pad pair where the LED's pad 2 (net=0) sits 0.2mm from the same-component target pad 1; `route_net(1)` returns no routes.

## Reproduction

Verified at `d83fa4a6` (origin/main, 2026-06-10) with a freshly built C++ backend — fails in 707s. Also fails at `7d973e03`. Independent of the #3480 intra-IC change (verified by reverting only `router/path.py`/`router/core.py` to main and re-running).

Note the test also takes ~9-12 minutes wall, which suggests the router is exhausting retries/rip-up before failing — possibly a regression from a recent obstacle-marking or recovery-gating change (#3483 touched off-grid recovery gating; #3488 touched rip-up parity).

## Ask

Bisect the regression and restore the net=0-adjacent-pad routability guarantee (the test documents the original fix: same-component net=0 pads must not block routing to their neighbor pad).

rjwalters:
## Summary

`test_route_to_pad_adjacent_to_net0_pad` failed on main: a routable signal pad could not be reached because an adjacent same-component net=0 (NC/unconnected) pad whose metal overlapped it caused the clearance validator to reject every route to the pad's center. `route_net(1)` returned no routes.

## Root cause

The regression is the `clearance >= 0` guard added by **#2933** ("same-component carve-out rejects trace-through-pad-metal") and carried forward by **#3545** (net-aware validator skip) — not #3483/#3488 as the triage suspected.

The same-component carve-out (the #1764 reachability fix) only silenced a neighbour-pad clearance complaint when the trace cleared the pad metal (`clearance >= 0`). #2933 added that guard to catch traces passing *through* an opposite signal pad's copper on standard-pitch passives. But when a signal pad's metal overlaps an adjacent same-component net=0 pad (LED1: net=1 pin at x=10.0, net=0 pin at x=10.2, 0.6mm pads → metal overlap), the trace endpoint reaching the signal pad's center necessarily lies inside the net=0 pad's rectangle. Measured clearance was **-0.2mm** → carve-out skipped → route rejected → no path.

A net=0 pad carries no electrical net, so geometric overlap is not a real short and the footprint guarantees the physical geometry.

## Changes

- `validate_segment_clearance` and `worst_segment_pad_deficit` (grid.py): silence negative clearance for **net=0** same-component pads only; foreign signal pads (net != 0) keep the strict `>= 0` guard so the #2933 trace-through-copper pathology stays caught.
- C++ `validate_route` predicate (grid.cpp): same change for backend parity.
- Bumped `ROUTER_CPP_BUILD_VERSION` 11 → 12 (both `types.hpp` and `cpp_backend.py`).
- Removed the `xfail` marker from `test_route_to_pad_adjacent_to_net0_pad`.

| Bisect the regression | done | Root cause is the #2933 `clearance >= 0` guard (not #3483/#3488); verified by direct `validate_segment_clearance` call returning clearance=-0.2 → reject |
| Pad adjacent to net=0 pad is reachable | done | `route_net(1)` now returns 1 route (was 0); test passes |
| net=0 pads on OTHER components still block | done | `test_net0_pad_on_different_component_still_blocks` still passes (change is gated on `exclude_refs` same-component) |
| Python/C++ parity | done | Both backends updated; `test_cpp_validation_parity.py` 34 passed; build version bumped |

## Test Plan

- `TestPadBlockedByAdjacentNetZero` (both tests): PASS (0.42s — the 9-12 min figure in the issue was stale; the failure is a fast empty-route return, not a retry/rip-up exhaustion)
- `test_cpp_validation_parity.py`: 34 passed, 3 skipped
- `test_same_component_clearance.py`, `test_validate_clearance_rect_pad.py`, `test_static_halo_ripup_3545.py`, `test_static_halo_via_gate_3566.py`, `test_grid_obstacle_respect.py`: 65 passed
- `test_router_core.py` + `test_drc_regression.py`: 185 passed (protects #2933 board-02 / #3545 regressions)

Closes #3490

### DIFF
--- a/src/kicad_tools/router/cpp/include/types.hpp
+++ b/src/kicad_tools/router/cpp/include/types.hpp
@@ -83,7 +83,14 @@ namespace router {
 // through J1-1's halo at 0.127mm actual vs 0.200mm required).
 // ``mark_blocked`` also began recording ``original_net`` for ALL static
 // cells (previously pad-metal only) so the restore has the owner net.
-constexpr int ROUTER_CPP_BUILD_VERSION = 11;
+//
+// Bump to 12 (Issue #3490): the same-component validator carve-out in
+// ``validate_route`` now silences NEGATIVE pad clearance for net=0
+// (unconnected / NC) same-component pads, restoring routability to a
+// signal pad whose metal overlaps an adjacent net=0 pad (the #1764
+// reachability guarantee). Foreign signal pads keep the strict >= 0
+// guard (#2933).
+constexpr int ROUTER_CPP_BUILD_VERSION = 12;
 
 // Grid cell state
 struct GridCell {
--- a/src/kicad_tools/router/cpp/src/grid.cpp
+++ b/src/kicad_tools/router/cpp/src/grid.cpp
@@ -509,7 +509,20 @@ ValidationResult Grid3D::validate_route(
             // trace clears the neighbour pad's metal (clearance >= 0).
             // Negative clearance means the trace overlaps pad copper -- a
             // true defect that no carve-out should silence.
-            if (same_component_signal_carveout && clearance >= 0.0f) {
+            //
+            // Issue #3490: EXCEPTION -- a same-component pad on net 0
+            // (an unconnected / NC pad) carries no electrical net, so a
+            // trace endpoint that geometrically overlaps its metal is not
+            // a real short.  When such a pad sits within metal-overlap
+            // distance of the signal pad being routed (an LED whose
+            // net=0 pin is 0.2mm from its signal pin -- the #1764
+            // scenario), reaching the signal pad's center is impossible
+            // without entering the net=0 pad's rectangle.  The footprint
+            // guarantees the geometry, so silence the violation for
+            // negative clearance too -- but ONLY for net=0 pads.  Foreign
+            // SIGNAL pads (net != 0) keep the strict >= 0 guard so the
+            // trace-through-pad-copper pathology (#2933) stays caught.
+            if (same_component_signal_carveout && (clearance >= 0.0f || pad.net == 0)) {
                 continue;
             }
 
--- a/src/kicad_tools/router/cpp_backend.py
+++ b/src/kicad_tools/router/cpp_backend.py
@@ -47,7 +47,7 @@ logger = logging.getLogger(__name__)
 # ``AttributeError`` deep in the routing code (e.g. ``router_cpp.PadBounds``
 # missing).  The guard below catches that at import time and falls back to the
 # pure-Python router with an actionable ``kct build-native`` hint.
-_REQUIRED_CPP_BUILD_VERSION = 11
+_REQUIRED_CPP_BUILD_VERSION = 12
 
 # Try to import C++ module with detailed error tracking
 _CPP_IMPORT_ERROR: str | None = None
--- a/src/kicad_tools/router/grid.py
+++ b/src/kicad_tools/router/grid.py
@@ -2799,7 +2799,10 @@ class RoutingGrid:
                 )
                 clearance = center_dist - seg_half_width
 
-            if same_component_signal_carveout and clearance >= 0:
+            # Issue #3490: same-component net=0 (NC/unconnected) pads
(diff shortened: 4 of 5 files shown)


rjwalters:
✅ **Approved!**

Reviewed the clearance-guard relaxation with focus on the two critical concerns: safety of the relaxation and Python/C++ parity. Both check out.

## Safety of the clearance relaxation

The new term `|| pad.net == 0` is AND-gated with `same_component_signal_carveout`, whose preconditions are:
- **Python** (`validate_segment_clearance` / `worst_segment_pad_deficit`): `exclude_refs and pad.ref in exclude_refs and not _is_plane_net_pad(pad) and _same_component_carveout_active(...)`
- **C++** (`validate_route`): `!pad.is_plane_net && is_excluded_ref(pad.ref_hash)`

So the effective predicate is **same-component (exclude_refs) AND not-a-plane-net AND net==0** — i.e. a genuine NC pad on the footprint being routed. Three properties confirm this is safe:

1. **Cross-component net=0 still blocks** — the gate requires `pad.ref in exclude_refs`, so a net=0 pad on a *different* component is never silenced. Verified: `test_net0_pad_on_different_component_still_blocks` PASSES.
2. **Foreign SIGNAL pads keep the strict `>= 0` guard** — `net != 0` pads never hit the new exception, so the #2933 trace-through-pad-copper pathology stays caught. Verified: `test_same_component_clearance.py` + `test_validate_clearance_rect_pad.py` = 23 passed.
3. **skip_nets plane pads are NOT relaxed** — per #3281, `_is_plane_net_pad` keys on `net_name`, not `net==0`. A `net=0, net_name="GND"` plane pad returns `True` and is excluded from the carve-out by the existing `not _is_plane_net_pad(pad)` clause. Only true NC pads (`net=0, net_name=""`) qualify. This is the subtle but correct distinction — `pad.net == 0` alone would be unsafe, but the AND-gating with the plane-net exclusion prevents that.

The via-vs-pad deficit predicate (`worst_via_pad_deficit`, grid.py:2901) intentionally keeps the strict `>= 0` guard — via/drill overlap is a separate concern from a trace reaching a pad center, and the bug is segment-only. Correct scoping.

## Python/C++ parity

- Built the C++ extension in the worktree: `router_cpp.BUILD_VERSION == 12`, `cpp_backend._REQUIRED_CPP_BUILD_VERSION == 12`, `_CPP_AVAILABLE == True`, no import fallback.
- The new term `(clearance >= 0 || pad.net == 0)` is byte-for-byte identical in both backends.
- `test_cpp_validation_parity.py`: 34 passed (the exact divergence class from #3456/#3622).
- Confirmed `origin/main` is still at `ROUTER_CPP_BUILD_VERSION = 11` — no collision with the unmerged #3630.

## Test Execution

- `TestPadBlockedByAdjacentNetZero` (both tests): PASS in 0.34s — confirms the de-xfailed test passes and the stale "9-12 min" figure (it's a fast empty-route return).
- `test_cpp_validation_parity.py` + `test_router_core.py`: 214 passed, 3 skipped.
- `test_same_component_clearance.py` + `test_validate_clearance_rect_pad.py`: 23 passed.

## CI Status

All 8 checks pass (Test 12m46s, Diff-Pair 13m15s, Match-Group 13m26s, C++ Build, Lint, Type Check, DRC, kicad-cli smoke). Merge state: **CLEAN**. PR uses "Closes #3490".

Clean, well-documented, surgically-scoped fix. Approving.

Issue #3411, opened 2026-06-09
board 04 (stm32-devboard): production routing tops out at 67% reach; OSC_OUT BLOCKED_PATH (issue #3402 follow-on)


rjwalters:
## Background

Per Issue #3402 audit (2026-06-09), board 04's production routing tops out at **67% reach** on both 2L and 4L. The router reports \`BLOCKED_PATH\` for the OSC_OUT net with suggestion "Move C11, U2 north to create routing channel". 4L does not improve the result because the failure is component-placement-driven, not layer-count-driven.

## Reproduction

\`\`\`bash
uv run kct build-native --check
uv run kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb \\
  --auto-layers --max-layers 4 -o /tmp/board04.kicad_pcb
# 6/9 nets routed (67%); 1 BLOCKED_PATH (OSC_OUT), 2 partial (OSC_IN, NRST)
\`\`\`

At 2L: 6/9 reach, escalation stops because failures are not congestion-related (overflow=0). At 4L: 6/9 reach, same blocked-path signature.

## Why this is NOT a \`starting_layers\` decision

- The auto-layers escalator deliberately stops when failures are not congestion-driven — more layers cannot rescue \`BLOCKED_PATH\`.
- The OSC crystal block (Y1 / C10 / C11) is placed too close to U2 (LDO) and the routing channel is mechanically blocked.
- The Issue #3402 audit therefore left board 04 at the default 2L probe — the fix is placement, not the escalation ladder.

## Scope (suggested)

- Move C11 / U2 north (per the router's own suggestion) to open the OSC_OUT channel.
- Re-check that OSC_IN and NRST also complete after the placement move.
- Acceptance: \`kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb --auto-layers --max-layers 4\` reaches 9/9 (100%) with 0 DRC at the jlcpcb tier.

## Out of scope

- Setting \`starting_layers: 4\` in the spec (would not help — Issue #3402 measurement).
- Lowering the 100% manufacturable bar.

## Related

- Parent: #3402 (per-board starting_layers audit)
- Bar: feedback memory \`feedback_manufacturable_means_100pct.md\`

rjwalters:
## Investigation summary (2026-06-09, ~5h)

[...]

PR #3647, opened 2026-06-13
fix(router): refuse adjacent-pin via-in-pad pairs that violate barrel spacing

### LINKED ISSUE #3429: Router: adjacent-pin via-in-pad conflict detection in _escape_qfp_alternating (Issue #3411 follow-on)
## Background

Per Issue #3411 investigation, `_escape_qfp_alternating` chooses inner-segment escape directions for fine-pitch pins via pad-index parity alternation, without checking whether ADJACENT pads' inner-segment directions will collide.  When two adjacent signal pins both need `_try_in_pad_escape` rescue, the inner-segment directions can point at each other — the second pin's via is rejected because it would clip the first pin's B.Cu stub.

## Proposed fix

At the dispatcher level (`_escape_qfp_alternating`, `escape.py:2305`), pre-scan for adjacent signal pins that both need via-in-pad rescue (the unclipped surface escape would violate neighbour-pad clearance).  When detected, OVERRIDE the alternation direction for the second pin so the two inner segments diverge in opposite cardinal directions.

For board-04 west-edge case U2.6 (OSC_OUT, i=2) + U2.7 (NRST, i=3):
- U2.6 currently gets direction=WEST (i even), which triggers the unclipped-clearance violation → in-pad rescue with inner segment in NORTH=(0,+1) (toward larger y)
- U2.7 currently gets direction=SOUTH (i odd, (i//2)%2==1), which would emit inner segment in (0, -1) (toward smaller y)
- The two inner segments point at each other → conflict

Override should set U2.6's in-pad inner segment direction to (0, -1) (NORTH on screen) and leave U2.7's at (0, -1) too — they diverge into independent escape lanes.

## Acceptance

- [ ] Board 04 reaches 9/9 routed
- [ ] No regression on board 03 / 05 / 07 fine-pitch escape patterns
- [ ] Routed-DRC tolerance gate unchanged

## Related

- Parent: #3411
- Sibling: router subissue A (target-aware inner direction)

rjwalters:
## Summary

Defense-in-depth for the board-04 fine-pitch QFP escape path (parent #3411).
Adds adjacent-pin via-in-pad **barrel-spacing** conflict detection to the
`_escape_qfp_alternating` dispatcher so two adjacent fine-pitch pins can no
longer both commit an in-pad via that physically cannot coexist.

Per the curator's sequencing note, #3428 (target-aware inner direction) was
the DIRECT fix and has already merged; board 04 already reaches 9/9. This PR
implements the remaining hardening: refuse geometrically-impossible rescue
pairs *before* the second via is generated, instead of having
`apply_escape_routes` silently drop one at commit time.

## Changes

- `_adjacent_in_pad_via_conflict(...)` — pairwise predicate: returns the
  conflicting sibling `EscapeRoute` when a foreign-net in-pad via lies closer
  than `via_r_A + via_r_B + clearance` (0.727mm for 0.6mm vias + 0.127mm
  clearance on jlcpcb-tier1; > the 0.5mm LQFP-48 pitch).
- `_try_in_pad_escape(..., enforce_adjacent_via_spacing=False)` — when True,
  refuses the rescue (returns `None`) on a barrel conflict, increments the
  new `adjacent_in_pad_via_conflicts_refused` counter, and emits a structured
  `logger.info` line (mirrors the `missed_via_in_pad_rescues` pattern).
- Dispatcher passes `enforce_adjacent_via_spacing=(pocket_target is None)` so
  the guard is active on the LEGACY violation/pin_boxed rescues (which have a
  lateral/surface fallback when refused) and **exempt** for the #3428
  pocket-escape rescue (whose divergent target-aware stubs are designed to let
  two adjacent in-pad vias coexist).
- New unit suite `tests/router/test_escape_adjacent_in_pad_conflict.py`.

The guard is distinct from #3470's `_in_pad_stub_conflicts`: that flips the
inner-layer STUB direction, which cannot resolve a VIA-BARREL conflict
(flipping the stub does not move the barrel). Default `False` keeps every
existing call site byte-for-byte unchanged.

| Criterion (curator-revised AC) | Status | Verification |
|---|---|---|
| Pre-scan detects adjacent same-edge in-pad rescue pairs below required via-via spacing before the 2nd via is committed | ✅ | `_adjacent_in_pad_via_conflict` runs inside `_try_in_pad_escape` before the `Via(...)` is built |
| On detection, refuse the 2nd in-pad rescue (no pair that `apply_escape_routes` later drops) | ✅ | returns `None`; dispatcher falls to lateral/surface path |
| Refusal is observable: structured log + counter | ✅ | `adjacent_in_pad_via_conflicts_refused` counter + `logger.info` |
| Synthetic-fixture test: pair conflict refused, coarse pitch does not fire, legacy preserved | ✅ | 9 tests in the new file (sub-pitch conflicts, 0.8mm coarse OK, micro-via 0.3mm OK, same-net skip, surface-escape skip, default-off legacy) |
| Board-04 (conditional after #3428) | ✅ | production recipe routes **9/9 (100%)** |
| No regression on boards 03/05/07 escape patterns | ✅ | full escape/router suites green (see Test Plan) |

## Test Plan

- New: `tests/router/test_escape_adjacent_in_pad_conflict.py` — 9 passed.
- #3428 regression: `tests/router/test_escape_target_direction.py` — pocket-escape rescue still fires (exempt from the guard); all pass.
- Escape suites (boards 03/05/07 patterns + LQFP/strict/lateral/stub-conflict/missed-rescue/extended-pitch): 214+ passed.
- Integration: `kct route boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb --mfr jlcpcb-tier1 --auto-fix --auto-layers --max-layers 4 --auto-mfr-tier --placement-feedback --micro-via-in-pad-fallback --seed 42` → **Nets routed: 9/9**.

Closes #3429

### DIFF
--- a/src/kicad_tools/router/escape.py
+++ b/src/kicad_tools/router/escape.py
@@ -1319,6 +1319,23 @@ class EscapeRouter:
         # for the named-constraint diagnostic line.
         self.missed_via_in_pad_components: set[str] = set()
 
+        # Issue #3429: Counter for adjacent-pin via-in-pad CONFLICTS that
+        # the dispatcher refused before committing a second in-pad via.
+        # Two adjacent fine-pitch QFP pins can each be flagged for an
+        # in-pad rescue, but at sub-pitch spacing their two via barrels +
+        # annular rings cannot both fit (center spacing < via_r_A +
+        # via_r_B + clearance).  ``_try_in_pad_escape`` validates the
+        # candidate via against PADS only -- sibling escapes' vias from
+        # the same pass are invisible there -- so without this guard the
+        # losing escape was silently dropped at the ``apply_escape_routes``
+        # commit-time cross-validation with no retry.  When the guard
+        # fires, the second in-pad rescue is refused (returns None) so the
+        # dispatcher falls through to the lateral / surface escape path
+        # instead of emitting a pair that the commit step later drops.
+        # Mirrors the ``missed_via_in_pad_rescues`` instrumentation
+        # pattern; consumed by diagnostics / future tier-escalation.
+        self.adjacent_in_pad_via_conflicts_refused: int = 0
+
         # Issue #3257: per-pad escape-layer overrides for SSOP/TSSOP
         # fine-pitch dual-row dispatcher.  Maps ``(ref, pin)`` to the
         # forced escape layer (``Layer.F_CU`` -> stay on surface,
@@ -2980,6 +2997,15 @@ class EscapeRouter:
                             # avoid copper conflicts with the escapes
                             # generated earlier in this pass.
                             existing_escapes=escapes,
+                            # Issue #3429: enforce via-barrel spacing
+                            # against sibling in-pad vias for the LEGACY
+                            # violation / pin_boxed rescues (which have a
+                            # lateral / surface fallback when refused).
+                            # The #3428 pocket-escape rescue is exempt --
+                            # its divergent target-aware stubs are designed
+                            # to let two adjacent in-pad vias coexist, and
+                            # it has no lateral fallback.
+                            enforce_adjacent_via_spacing=(pocket_target is None),
                         )
                         if in_pad_route is not None:
                             if pocket_target is not None:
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Defense-in-depth hardening for the board-04 fine-pitch QFP escape path. Reviewed with focus on the two concerns the builder flagged: gating-flag correctness and whether the guard is load-bearing vs dead code.

## Scope
- 2 files, +474/-0: `src/kicad_tools/router/escape.py` (+158), `tests/router/test_escape_adjacent_in_pad_conflict.py` (+316, new).
- The single commit (`9eb1ee8`) touches exactly these two files. (The `main...HEAD` three-dot diff looks huge but that is merge-base pollution from the branch point; `gh pr diff --name-only` and `git show --stat` confirm the real scope is the 2 files above.)

## Gating-flag correctness (the crux) — CORRECT
The dispatcher passes `enforce_adjacent_via_spacing=(pocket_target is None)` (escape.py:3008). `pocket_target` is set only at escape.py:2959, gated on `tvx*dvx + tvy*dvy < 0` (target direction OPPOSITE the parity direction) **and** `_neighbour_claimed_in_pad_via(...)` — i.e. precisely the #3428 intentional divergent pair. So:
- #3428 pocket-escape rescue (`pocket_target is not None`) → guard **exempt** (its divergent target-aware stubs are designed to let two adjacent in-pad vias coexist, and it has no lateral fallback).
- Legacy `violation` / `pin_boxed` rescues (`pocket_target is None`) → guard **active** (these have a lateral/surface fallback to take when the rescue is refused).

This is the correct discriminator between "geometrically-doomed legacy pair (refuse → fallback)" and "intentional #3428 divergent pair (keep)". Verified by the #3428 regression suite passing (see below).

## Load-bearing, not dead code
- `test_second_rescue_refused_when_sibling_via_conflicts`: with the flag True, a real sub-pitch pair (0.5 mm < required 0.727 mm for 0.6 mm vias + 0.127 mm clearance) is refused (returns None) and the `adjacent_in_pad_via_conflicts_refused` counter increments.
- `test_guard_disabled_by_default_preserves_legacy`: the **same conflicting geometry** with the flag omitted (default False) still emits the via — i.e. the pre-#3429 silent-drop behaviour. The contrast proves the guard actually does something useful on the reachable legacy rescue path; it is not a no-op.
- The guard is distinct from #3470's `_in_pad_stub_conflicts` (which flips inner-layer stub copper and cannot move a via barrel), so it is not redundant.

## Tests run locally (worktree, `uv run pytest`)
- New suite `test_escape_adjacent_in_pad_conflict.py`: **9 passed** (sub-pitch conflict fires, 0.8 mm coarse pitch does NOT fire, 0.3 mm micro-via does NOT fire, same-net skipped, surface-escape sibling skipped, None-existing-escapes no-op, default-off legacy).
- #3428 regression `test_escape_target_direction.py`: **all pass** (pocket-escape rescue still fires; guard exemption holds).
- Escape-pattern suites (`test_escape_pad_layer_overrides`, `test_escape_qfp_plane_sandwich`, `test_extended_pitch_in_pad_escape`, `test_softstart_revb_*`): **30 passed, 2 skipped**.
- Board 04 routing regression `test_board_04_routing_regression.py`: **2 passed** in isolation — board 04 stays 9/9, no regression from the guard.

## CI status
All 8 checks pass; `mergeStateStatus` = CLEAN.

## Notes (not blocking)
- Local `ruff check` flags 4 pre-existing SIM114 issues at escape.py:1371/3358/3360/3362 — all OUTSIDE this PR's touched hunks (1319-1342, 2997-3012, 6565-7000) and not enforced by the CI Lint config (which passed green). Not attributable to this PR.
- A broader local `tests/router/` run showed 3 DRC-assertion failures, all with the identical root cause `No module named 'shapely'` — a missing optional dep in this local worktree's venv (shapely IS declared in pyproject.toml `shapely>=2.0`). Environment-only, unrelated to the escape.py change; CI's Routed PCB DRC Check and Test jobs both pass green.

The hardening is correctly gated, genuinely load-bearing on the legacy rescue path, and does not regress #3428 or board 04. AC met under the curator's demotion to "no regression + hardening".

Issue #3582, opened 2026-06-12
router: board 04 routes 7/9 on stripped 2L recipe — test_routes_all_nets_on_2l fails on main (expected >= 8/9)


rjwalters:
## Bug Description

`tests/test_board_04_routing_regression.py::TestBoard04OscOutRouting::test_routes_all_nets_on_2l` fails on current `main`:

```
AssertionError: Board 04 routed only 7/9 nets (expected 8/9).
```

Verified on a **pristine detached worktree at `origin/main` (45e68efa, 2026-06-11)** with the C++ backend built — no local changes:

```
uv run pytest tests/test_board_04_routing_regression.py::TestBoard04OscOutRouting::test_routes_all_nets_on_2l
# 1 failed in 341.94s — routed 7/9
```

The test runs `kct route --seed 42 --no-auto-layers --layers 2 --manufacturer jlcpcb --backend cpp` on the committed unrouted `boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb` and asserts >= 8/9 nets (issue #3281 floor; NRST gap is the accepted 1).

## Likely Origin

Recent router changes are the prime suspects since the test input is unchanged:

- #3578 (45e68efa) — certify global-min clearance in R-tree segment query
- #3577 (d8cdf0cc) — static-halo gate for `_is_via_blocked`/`_is_diagonal_corner_blocked` + via-pad demotion backstop
- #3537 / #3201 / #3193 — 45° quantization, per-pad channel budgets, A* tie-break

A bisect across these on the fixed `--seed 42` recipe should pinpoint it.

## Context

- Discovered while validating the board-04 artifact fix for #3552 (refill-only zone repair — does not touch router code or the unrouted PCB, and the failure reproduces identically without that change).
- The committed routed artifact `stm32_devboard_routed.kicad_pcb` is unaffected (it passes the strict routed-DRC gate); this is a live-router regression, not an artifact defect.

rjwalters:
## Bisect findings (orchestrator-dispatched)

**Breaking merge: PR #3565 (`85b631f6`, "reject routes through static foreign-pad halos + net-aware validator skip", closes #3545).** PRs #3577 and #3578 are exonerated — HEAD (`45e68efa`) reproduces the identical 7/9 result with the identical dropped nets as #3565's merge commit.

[...]

PR #3648, opened 2026-06-13
fix(router): auto-lateral-via fallback when sibling barrel blocks in-pad rescue

### LINKED ISSUE #3430: Router: auto-lateral-via fallback when adjacent fine-pitch pin blocks in-pad rescue (Issue #3411 follow-on)
## Background

Per Issue #3411 investigation, when adjacent fine-pitch signal pins both need via-in-pad rescue, only ONE of them can succeed (the other's via clips the first's via at 0.5mm pitch + 0.6mm OD).  The `_try_lateral_via_escape` helper at `escape.py:5955` already supports placing the via OFF the pad along the escape direction — this gives more breathing room for the second pin.

Currently `_try_lateral_via_escape` is only invoked when `_try_in_pad_escape` returns None.  When the in-pad rescue SUCCEEDS for pin N, the dispatcher moves on to pin N+1 and tries the in-pad rescue WITHOUT first checking whether pin N's via has already claimed pin N+1's in-pad region.

## Proposed fix

Modify the dispatcher (`_escape_qfp_alternating`) so that AFTER the in-pad rescue succeeds for pin N, it tracks the via location and clearance footprint.  For the IMMEDIATE NEIGHBOUR (pin N+1, N-1), it checks whether the via footprint blocks the neighbour's in-pad region.  If so, it forces the neighbour to use `_try_lateral_via_escape` instead of `_try_in_pad_escape` for its own rescue.

## Acceptance

- [ ] Board 04 reaches 9/9 routed
- [ ] No regression on existing fine-pitch escape paths
- [ ] Routed-DRC tolerance gate unchanged

## Related

- Parent: #3411
- Sibling: router subissue A, B

rjwalters:
## Summary

Recovery half of the adjacent-pin via-in-pad story (parent #3411). #3429 (merged as PR #3647) added the DETECTION: `_adjacent_in_pad_via_conflict` makes `_try_in_pad_escape` return `None` when a second fine-pitch pin's via barrel would clash with a sibling in-pad via (0.6mm OD vias need 0.727mm center spacing > 0.5mm LQFP pitch). This PR adds the COMPLEMENT: when the in-pad rescue is refused, the `_escape_qfp_alternating` dispatcher now lets the pin escape via an off-pad **lateral via** instead of silently deferring it to the main router (where `apply_escape_routes` would drop it).

Critically, the lateral candidate is validated against the vias of sibling escapes placed earlier in the SAME pass. Without that, the fallback would merely relocate the very commit-time drop #3429 was meant to prevent — from the refused in-pad via to its lateral replacement.

## Changes

- `_try_lateral_via_escape` gains an `existing_escapes` parameter; each off-pad via candidate is now rejected if it lands inside a sibling via's barrel-spacing halo.
- New `_lateral_via_sibling_conflict` predicate: mirrors `_adjacent_in_pad_via_conflict` (#3429) but rejects against ANY foreign-net sibling via — in-pad OR lateral — since two off-pad vias can collide just as readily as an in-pad/off-pad pair.
- Dispatcher forwards the accumulated `escapes` into the lateral-fallback call and bumps a new `forced_lateral_via_fallbacks` counter (observability, mirroring `missed_via_in_pad_rescues` / `adjacent_in_pad_via_conflicts_refused`) with a structured log line.
- New test file `tests/router/test_escape_adjacent_lateral_fallback.py` (10 tests).
- All changes are gated on `existing_escapes` being non-None, so legacy / unit-fixture call sites are byte-for-byte unchanged. Scoped strictly to `_escape_qfp_alternating` (the dispatcher named in the issue); the SSOP/TSSOP row dispatcher call sites are untouched.

## Interaction with #3429

#3429 refuses the doomed second in-pad rescue (`_try_in_pad_escape` -> None); this PR is what the dispatcher does next. The existing `if violation or pin_boxed:` branch already routed to `_try_lateral_via_escape` on a None in-pad result — this PR makes that lateral attempt sibling-aware and observable, so the rescued pin is committed cleanly rather than dropped. No duplication of adjacency logic: the lateral path reuses the same `via_r + via_r + clearance` spacing arithmetic.

| Criterion (from curator's revised AC) | Status | Verification |
|---|---|---|
| After in-pad rescue refused, dispatcher forces lateral instead of a doomed in-pad rescue | ✅ | #3429 returns None on conflict; dispatcher falls through to sibling-aware `_try_lateral_via_escape` |
| Forced-lateral activation observable (log + counter) | ✅ | `forced_lateral_via_fallbacks` counter + structured `AUTO-LATERAL-VIA` log line |
| Lateral candidates validated against tracked sibling escapes, not just footprint pads | ✅ | `_lateral_via_sibling_conflict` + `existing_escapes` plumbed through; unit tests `test_lateral_via_offset_past_sibling`, `test_sibling_aware_pushes_via_further_than_legacy` |
| Synthetic fixture: adjacent pair -> exactly one in-pad + one lateral, no commit-time drop | ✅ | `tests/router/test_escape_adjacent_lateral_fallback.py` (10 tests); lateral via spacing >= 0.727mm asserted |
| No-op when sibling NOT blocking | ✅ | `test_coarse_spacing_does_not_conflict`, `test_legacy_call_without_siblings_unchanged` |
| Board 04 no regression | ✅ | `test_board_04_routing_regression.py` passes (floor `REQUIRED_NETS_ROUTED=8`, pinned per #3582); board-04 9/9 already addressed by #3428, so per the curator's sequencing note this is defense-in-depth hardening with synthetic-fixture acceptance |
| No regression on existing escape suites | ✅ | 68 escape/in-pad/lateral/conflict tests pass; #3429 suite (9 tests) unchanged |

## Test Plan

- `tests/router/test_escape_adjacent_lateral_fallback.py` — 10 new tests pass
- `tests/router/test_escape_adjacent_in_pad_conflict.py` (#3429) — 9 tests pass, no regression
- `tests/test_escape_lateral_recovery.py`, `test_escape_via_in_pad*.py`, `test_escape_strict_in_pad_clearance.py`, `test_escape_micro_via_fallback.py`, `test_in_pad_stub_conflict.py` — all pass
- `tests/test_board_04_routing_regression.py`, `test_board_04_drc_recipe_mfr.py`, `test_escape_qfp_plane_sandwich.py`, `test_escape_via_in_pad_lqfp.py`, `test_softstart_revb_u1_lqfp32_escape.py`, `test_escape_target_direction.py` — pass (57 passed)
- The 3 `test_board_04_mfr_gate.py` errors are PRE-EXISTING (fail identically on the base commit via stash check) — a sandbox subprocess-spawn issue (`sys.executable` child returns -1), unrelated to this change.

Closes #3430

### DIFF
--- a/src/kicad_tools/router/escape.py
+++ b/src/kicad_tools/router/escape.py
@@ -7073,6 +7073,12 @@ class EscapeRouter:
         )
         if conflicting is not None:
             self.adjacent_in_pad_via_conflicts_refused += 1
+            # ``_adjacent_in_pad_via_conflict`` only returns an EscapeRoute
+            # whose ``.via`` is non-None (it skips ``via is None`` siblings),
+            # so narrowing here is sound -- the local lets mypy prove the
+            # attribute access below type-checks (Via | None -> Via).
+            sibling_via = conflicting.via
+            assert sibling_via is not None
             logger.info(
                 "In-pad rescue REFUSED for pad %s (ref=%s pin=%s) at "
                 "(%.3f, %.3f): via barrel (OD %.3fmm) conflicts with the "
@@ -7089,8 +7095,8 @@ class EscapeRouter:
                 via_diameter,
                 conflicting.pad.pin,
                 conflicting.pad.net_name,
-                conflicting.via.x,
-                conflicting.via.y,
+                sibling_via.x,
+                sibling_via.y,
             )
             return None
 
--- a/src/kicad_tools/router/escape.py
+++ b/src/kicad_tools/router/escape.py
@@ -1336,6 +1336,22 @@ class EscapeRouter:
         # pattern; consumed by diagnostics / future tier-escalation.
         self.adjacent_in_pad_via_conflicts_refused: int = 0
 
+        # Issue #3430: Counter for the auto-lateral-via FALLBACK that the
+        # dispatcher takes after #3429 refuses an in-pad rescue because a
+        # sibling fine-pitch pin's via barrel blocks it.  #3429 is the
+        # DETECTION half (``_try_in_pad_escape`` returns None on a sub-pitch
+        # via-via conflict); this is the RECOVERY half -- the dispatcher
+        # falls through to ``_try_lateral_via_escape``, which pushes the
+        # via OFF the pad along the outward escape direction so the second
+        # pin still escapes instead of being silently dropped at
+        # ``apply_escape_routes`` commit time.  Bumped once per pin whose
+        # in-pad rescue was refused AND whose lateral fallback succeeded.
+        # When this counter is non-zero a fine-pitch package that would
+        # otherwise have lost a pin to the commit-time drop was rescued by
+        # the offset escape.  Mirrors the instrumentation pattern above;
+        # reset between attempts by ``Autorouter.reset_attempt_state``.
+        self.forced_lateral_via_fallbacks: int = 0
+
         # Issue #3257: per-pad escape-layer overrides for SSOP/TSSOP
         # fine-pitch dual-row dispatcher.  Maps ``(ref, pin)`` to the
         # forced escape layer (``Layer.F_CU`` -> stay on surface,
@@ -3083,8 +3099,35 @@ class EscapeRouter:
                                 effective_clearance=effective_clearance,
                                 escape_width=escape_width,
                                 package=package,
+                                # Issue #3430: forward the escapes placed so
+                                # far this pass so the off-pad via candidate
+                                # is validated against sibling escape vias
+                                # (in-pad OR lateral), not just footprint
+                                # pads.  Without this the lateral fallback
+                                # could relocate the very commit-time drop
+                                # #3429's detection was meant to prevent.
+                                existing_escapes=escapes,
                             )
                             if lateral_route is not None:
+                                # Issue #3430: observe the auto-lateral-via
+                                # FALLBACK.  ``adjacent_in_pad_via_conflicts_refused``
+                                # (#3429) counts the refused in-pad rescue;
+                                # this counts the successful offset escape
+                                # that rescued the same pin.  When the in-pad
+                                # via was refused for a barrel conflict, this
+                                # is the recovery that keeps board 04 at 9/9.
+                                self.forced_lateral_via_fallbacks += 1
+                                logger.info(
+                                    "AUTO-LATERAL-VIA fallback for %s pin %s "
+                                    "(net %s): in-pad rescue refused (sibling "
+                                    "via barrel conflict / strict clearance); "
+                                    "escaped via off-pad lateral via instead "
+                                    "of deferring the pin to the main router "
+                                    "(Issue #3430, follow-on to #3429).",
+                                    package.ref,
+                                    pad.pin,
+                                    pad.net_name,
+                                )
                                 escapes.append(lateral_route)
                                 continue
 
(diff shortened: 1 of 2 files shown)


rjwalters:
❌ **Changes Requested — CI Type Check failing**

I evaluated this PR per the loom:judge workflow. The implementation is sound and the tests pass locally (19 passed: the 10 new `test_escape_adjacent_lateral_fallback.py` tests plus the 9-test #3429 `test_escape_adjacent_in_pad_conflict.py` suite — no regression, no duplication of #3429's adjacency logic). But CI cannot be approved because the **Type Check** job fails:

```
2 NEW mypy error(s) beyond the baseline (.github/mypy-baseline.txt):
src/kicad_tools/router/escape.py: Item "None" of "Via | None" has no attribute "x"  [union-attr]
src/kicad_tools/router/escape.py: Item "None" of "Via | None" has no attribute "y"  [union-attr]
```

Reproduced locally via `uv run python scripts/ci/check_mypy_baseline.py`.

**Location:** `src/kicad_tools/router/escape.py:7092-7093`, in `_try_in_pad_escape`:

```python
conflicting.via.x,
conflicting.via.y,
```

`conflicting` is the `EscapeRoute` returned by `_adjacent_in_pad_via_conflict`, whose `.via` is typed `Via | None`. mypy cannot prove it is non-None at the log call. These two lines were authored by #3429 (merged as #3647), but the baseline Type Check gate (#3645) now attributes them as NEW against this PR's diff, so the gate is red and the PR is blocked. It must be fixed here.

**Required change:**
- Narrow `conflicting.via` before the `logger.info(...)` call so the attribute access type-checks. Either add `assert conflicting.via is not None` right after the conflict is detected, or capture `sibling_via = conflicting.via` under the existing guard and log `sibling_via.x` / `sibling_via.y`. (A targeted `# type: ignore[union-attr]` would also clear the gate, but an assert/local is cleaner since the conflict path guarantees non-None.)
- After the fix, re-run `uv run python scripts/ci/check_mypy_baseline.py` and confirm zero new errors in `escape.py`.

**Non-blocking observations (no change required, noted for context):**
- The new `forced_lateral_via_fallbacks` counter is exercised only for its zero-initialization (`test_counter_initialises_to_zero`); the refuse→dispatcher→fallback→increment path is verified piecewise (predicate + `_try_lateral_via_escape(existing_escapes=...)` behavioral tests showing the via is pushed outward past the 0.727mm barrel floor) rather than end-to-end through `_escape_qfp_alternating`. The piecewise coverage is adequate, but an end-to-end dispatcher test asserting the counter increments would harden the load-bearing claim. Optional.
- The "synthetic-fixture hardening / defense-in-depth" framing is honest: board 04 stays at its pinned floor (`REQUIRED_NETS_ROUTED=8`, #3582), and the PR does not claim a board-04 improvement.

The implementation itself (sibling-aware `existing_escapes` plumbing, `_lateral_via_sibling_conflict` predicate, `existing_escapes is not None` gating preserving legacy call sites byte-for-byte) is correct and load-bearing. Just fix the type error and I'll re-evaluate.

rjwalters:
Doctor: type-check blocker addressed.

**Fix (the only blocker):** In `src/kicad_tools/router/escape.py` `_try_in_pad_escape`, the in-pad-rescue REFUSED `logger.info(...)` accessed `conflicting.via.x` / `conflicting.via.y`, but `conflicting.via` is typed `Via | None` and mypy could not prove non-None at the call site (`union-attr`). `_adjacent_in_pad_via_conflict` only ever returns an `EscapeRoute` whose `.via` is non-None (it `continue`s past `via is None` siblings), so I captured it into a local and narrowed:

```python
sibling_via = conflicting.via
assert sibling_via is not None
```

then log `sibling_via.x` / `sibling_via.y`. Logged values are byte-for-byte unchanged; no `# type: ignore` used.

**Verification (run in the worktree with the project mypy via `--extra dev`):**
- `uv run python scripts/ci/check_mypy_baseline.py` -> exit 0: `OK: mypy reports 1512 error(s), all within the baseline (1512 allowed). No new type errors.` The two `escape.py` `union-attr` errors are gone.
- `uv run pytest tests/router/test_escape_adjacent_lateral_fallback.py tests/router/test_escape_adjacent_in_pad_conflict.py` -> 19 passed, no regression.

**Optional end-to-end counter test:** Skipped. Exercising `forced_lateral_via_fallbacks` end-to-end requires constructing a full fine-pitch QFP fixture with two adjacent foreign-net pins both flagged for in-pad rescue (first claims a via, second refused then laterally rescued) all the way through `_escape_qfp_alternating` — not quick or low-risk for a type-check-only fix, and the increment path is already covered piecewise (predicate + `_try_lateral_via_escape(existing_escapes=...)` behavioral tests) as you noted. Left for a follow-up if desired.

Returning to review-requested.

rjwalters:
APPROVED — re-judge after Doctor fix (commit c254e58b).

Focused re-review of the type-narrowing fix that was the sole blocker. Substantive design was already validated in the prior review (load-bearing, builds on #3429, no board-04 regression); this pass confirms the fix is correct and CI is now fully green.

**Type-narrowing fix is sound.** In `_try_in_pad_escape` the Doctor captured `sibling_via = conflicting.via` then `assert sibling_via is not None` before logging `sibling_via.x` / `sibling_via.y`. I verified the invariant the assert relies on: `_adjacent_in_pad_via_conflict` (escape.py:6611) binds `via = er.via`, `continue`s when `via is None` (line 6670), and the only `return er` (line 6679) sits past that guard — so every returned `EscapeRoute` has a non-None `.via`. The assert can never fire on a legitimate path; it is pure narrowing, not runtime control flow. No `# type: ignore` used, logged values byte-for-byte unchanged. The fix commit touches only escape.py (+8/-2) — no scope creep.

**CI fully green (mergeStateStatus: CLEAN):**
- Type Check (prior blocker): PASS
- Test: PASS
- C++ Build Check: PASS
- Lint & Format: PASS
- Routed PCB DRC Check: PASS
- kicad-cli Round-trip Smoke: PASS
- Diff-Pair Routing Regression (board 06): PASS
- Match-Group Routing Regression (board 07): PASS

**Independent local verification (worktree at c254e58b, project mypy via `--extra dev`):**
- `uv run python scripts/ci/check_mypy_baseline.py` -> exit 0: `OK: mypy reports 1512 error(s), all within the baseline (1512 allowed). No new type errors.` The two `union-attr` errors on `conflicting.via.x/.y` are gone.
- `uv run pytest tests/router/test_escape_adjacent_lateral_fallback.py tests/router/test_escape_adjacent_in_pad_conflict.py` -> 19 passed, no regression.

The optional end-to-end counter test was correctly noted as non-blocking (increment path covered piecewise; left as an optional follow-up). Nothing else regressed. Approving — moving to loom:pr.

PR #3652, opened 2026-06-13
test(router): make same-net sibling optimize-loop regression load-bearing

### LINKED ISSUE #3637: test(router): test_same_net_sibling_cells_survive_optimize_loop is not load-bearing for the #3511 optimize-loop path
## Context

Found while reviewing PR #3634 (closes #3511). The fix wires a batched `grid.resync_route_occupancy(mutated_pairs)` into `optimize_routes_grid_synced` to repair same-net sibling cells that an incremental per-route `unmark_route` can clear. The fix is correct and merged.

## Gap

The PR's headline optimize-loop regression test, `tests/router/test_grid_resync_3507.py::TestOptimizeRoutesGridSynced::test_same_net_sibling_cells_survive_optimize_loop`, **passes identically with and without the fix**. I verified this by reverting `src/kicad_tools/router/optimizer/trace.py` to its `main` version (0 `resync_route_occupancy` calls) and running the test — it still passed.

Root cause: the test's minimal geometry (vertical sibling at x=10, y=5..15 crossing a horizontal route at y=10) does not expose the bug because:
1. The mutated route's NEW geometry (after collinear merge) still passes through the crossing cell (10,10), so the per-route `mark_route` re-marks it anyway.
2. The clearance envelope (trace half-width + clearance + safety margin = several cells wide) means the new geometry's envelope re-covers the sibling's centerline cells even where they previously overlapped.

I also tried collinear-overlap and zigzag-detour sibling geometries in ad-hoc probes; none produced an under-marked sibling cell on `main` either — the envelope width swamps minimal-geometry attempts.

## Why this is not a merge blocker (for #3634)

- The bug itself is empirically non-biting (per #3511) — defensive hardening.
- The `resync_route_occupancy` step-4 affected-net re-mark IS load-bearing and is proven by the pre-existing direct-API test `test_same_net_sibling_cells_survive_resync` (#3507).
- The optimize-loop wiring is exercised end-to-end (grid-consistency + bookkeeping-mirror assertions) and the no-mutation guard is spy-tested.

## Suggested work

Construct a regression test that genuinely fails without the batched resync at the `optimize_routes_grid_synced` level. Candidate approach: a sibling whose centerline cells lie in the mutated route's OLD clearance envelope but OUTSIDE the NEW envelope (e.g., the optimizer relocates the route to a different layer or a parallel offset > envelope width, leaving the sibling's overlap cells uncovered by the new geometry). May require a smaller grid resolution or narrower clearance so the envelope does not swamp the geometry. If no such configuration is reachable through the public optimizer passes, document that and keep the grid-API-level test as the load-bearing guard.

Discovered during code evaluation of PR #3634.

rjwalters:
## Summary

The headline #3511 regression test
`TestOptimizeRoutesGridSynced::test_same_net_sibling_cells_survive_optimize_loop`
was NOT load-bearing: it passed identically with and without the fix
(the batched `grid.resync_route_occupancy(mutated_pairs)` call in
`optimize_routes_grid_synced`). Its minimal collinear-merge geometry kept
the merged route passing through the crossing column, so the per-route
`mark_route` re-covered the sibling cells regardless of the resync.

This PR rewrites the test geometry so the optimizer RELOCATES the mutated
route off the same-net sibling's column by more than the clearance
envelope half-width, genuinely exercising the optimize-loop under-marking.

## Changes

- Replace the test body with a zigzag whose middle vertex bumps onto the
  sibling's column at (10,10); `merge_collinear` + `eliminate_zigzags` +
  `pull_tight` collapse it to a single straight segment at y=5, abandoning
  the sibling cells the old envelope owned across the y=5..10 overlap.
- Add a precondition guard asserting the OLD envelope actually owned the
  probed sibling centerline cells, so the test cannot pass vacuously.
- Assert the route was relocated (single segment at y=5) and the sibling's
  abandoned copper is still marked after the loop's batched resync.
- Test-only change. No router source modified.

| Test genuinely exercises optimize-loop under-marking | done | Geometry relocates the route off the sibling column; old envelope owned cells the new geometry abandons |
| Fails when #3511 fix reverted | done | Reverting `resync_route_occupancy(mutated_pairs)` in `optimizer/trace.py` -> FAIL: `net 5: cell (100,82,L0) of segment (10.0,5.0)-(10.0,15.0) is NOT marked` |
| Passes with the fix | done | With the call restored: `1 passed`; full file `17 passed` |
| Test-only (no router source change) | done | `git diff --stat` shows only `tests/router/test_grid_resync_3507.py` |

## Test Plan

- `uv run pytest tests/router/test_grid_resync_3507.py` -> 17 passed
- Load-bearing proof: temporarily replaced the resync call with `pass`;
  the test failed at `_assert_marked(router.grid, sibling)` on unmarked
  sibling cell (100,82,L0) (world y~=8.2). Restored the call; test passes.
- `uv run ruff check` / `ruff format` clean.

Closes #3637

### DIFF
--- a/tests/router/test_grid_resync_3507.py
+++ b/tests/router/test_grid_resync_3507.py
@@ -310,35 +310,87 @@ class TestOptimizeRoutesGridSynced:
         assert router.grid.routes[0] is route
 
     def test_same_net_sibling_cells_survive_optimize_loop(self):
-        """Issue #3511: the incremental per-route unmark clears every cell
-        the OLD envelope owned, including cells a geometry-unchanged
-        same-net SIBLING route still occupies where the envelopes
-        overlapped.  The per-route ``mark_route`` only re-marks the
-        mutated route's NEW geometry, so without the batched
-        post-loop resync the sibling's cells in the old overlap region
-        stay blank.  The trailing ``resync_route_occupancy`` must
-        re-mark them."""
+        """Issue #3511 (load-bearing): the batched post-loop resync must
+        re-mark a same-net sibling's copper that the incremental per-route
+        ``unmark_route`` blanked.
+
+        This test is deliberately constructed so that reverting the
+        ``grid.resync_route_occupancy(mutated_pairs)`` call in
+        ``optimize_routes_grid_synced`` makes it FAIL.  The earlier
+        minimal geometry (a vertical sibling crossing a *collinear*
+        horizontal route the optimizer merges in place) was NOT
+        load-bearing: the merged route's new segment still passed through
+        the crossing column, so the per-route ``mark_route`` re-covered
+        the sibling cells and the assertion passed with or without the
+        batched resync.
+
+        The fix is to make the optimizer RELOCATE the mutated route off
+        the sibling's column by more than the clearance envelope width:
+
+        * The sibling is a vertical own-net trace at x=10 spanning
+          y=5..15 (centerline cells in grid column gx=100).
+        * The mutated route is a zigzag whose middle vertex bumps UP to
+          (10, 10): its OLD geometry runs along the sibling's column for
+          the y=5..10 stretch, so the old clearance envelope owns the
+          sibling's centerline cells there.
+        * ``merge_collinear`` + ``eliminate_zigzags`` + ``pull_tight``
+          collapse the zigzag to a single straight segment at y=5
+          (``[(5,5)-(20,5)]``).  The NEW geometry's envelope sits at y=5
+          and does NOT re-cover the sibling cells above ~y=5.6.
+
+        So the per-route ``unmark_route(old)`` clears the sibling's
+        column-100 cells across the y=5..10 overlap and the per-route
+        ``mark_route(new)`` only re-marks cells near y=5.  Without the
+        batched resync those ~45 sibling centerline cells stay UNMARKED;
+        the resync's affected-net (net 5) re-mark restores them.
+        """
         router = _make_router()
-        # A vertical sibling the optimizer cannot simplify (single
-        # segment) crossing a horizontal mutated route at (10, 10).
+        # Vertical own-net sibling; the optimizer never touches it
+        # (single segment, nothing to merge/straighten).
         sibling = _make_route(5, "E", [(10.0, 5.0), (10.0, 15.0)])
-        # Collinear 3-point horizontal route through the crossing point:
-        # merge_collinear collapses it to one segment from (5,10)-(20,10),
-        # so the route is MUTATED (new object) and its old envelope
-        # overlapped the sibling at (10, 10).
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Test-only change that makes the #3511 same-net sibling optimize-loop regression genuinely load-bearing. I reproduced the load-bearing proof myself rather than taking the builder's word.

## Load-Bearing Proof (reproduced by Judge)

In the PR worktree (`.loom/worktrees/issue-3637`), built the C++ backend (`kct build-native`), then:

1. **As-is**: `pytest tests/router/test_grid_resync_3507.py` → `17 passed`.
2. **Fix reverted**: temporarily replaced `grid.resync_route_occupancy(mutated_pairs)` in `src/kicad_tools/router/optimizer/trace.py:136-137` with `pass`. The target test FAILED with exactly the reported error:
   ```
   AssertionError: net 5: cell (100,82,L0) of segment (10.0,5.0)-(10.0,15.0) is NOT marked
   ```
   Critically, it failed on the **post-loop** `_assert_marked(router.grid, sibling)` — NOT on the precondition guard — confirming the failure is genuinely the missing batched resync, not a setup artifact.
3. **Restored**: reverted trace.py (verified clean via `git diff --stat`), re-ran → `1 passed`.

## Precondition Guard (non-vacuous, verified)

Lines 362-377 run AFTER committing both routes but BEFORE `optimize_routes_grid_synced`. The guard:
- Builds `overlap_cells` from the sibling's centerline in the y=5..10 band and asserts it is non-empty (`"test geometry produced no overlap cells to probe"`), so the loop can't be skipped silently.
- Asserts each of those cells is `.blocked` (the OLD mutated envelope owned them).

This genuinely prevents a vacuous pass: when the fix was reverted, the guard still passed and the failure surfaced at the post-loop assertion — proving the geometry truly relocates copper off the sibling column and the test exercises the under-marking path.

## Test Scoping

**Strategy**: targeted file + single-test (`tests/router/test_grid_resync_3507.py`)
**Result**: 17 passed (full file); fail-without-fix reproduced and restored.

## Verification

| Check | Result |
|-------|--------|
| Test-only (no source committed) | ✅ `git diff origin/main...HEAD` = 1 file, `tests/router/test_grid_resync_3507.py` only |
| Fails when #3511 fix reverted | ✅ Reproduced: `net 5: cell (100,82,L0) ... is NOT marked` at post-loop assertion |
| Passes with fix | ✅ `17 passed` |
| Precondition guard non-vacuous | ✅ Guard runs pre-loop; passes even when fix reverted |
| CI green | ✅ All 9 checks pass |
| Merge state | ✅ CLEAN |
| `Closes #3637` | ✅ Present |

The remaining diff hunks (lines ~508-555) are harmless ruff-format whitespace reflows of unrelated assertions in the same file.

Great work — this is exactly the kind of "prove the test bites" change that keeps regression coverage honest.

PR #3632, opened 2026-06-13
fix(ci): board 07 pour-connectivity — fill zones + asserted copper-union audit

rjwalters:
## Summary

Board 07's recipe created copper-pour zone OUTLINES (step 4 `auto_create_zones_for_pour_nets`) but never invoked `kct zones fill`, so the routed artifact shipped zones with **zero fill geometry** and the `matchgroup-routing-regression` gate stayed green by omission — the same "dead pour" class board 06 closed in PR #3615 (issue #3509). This PR adds the fill + pour-connectivity-repair pipeline, the `REQUIRE_POUR_CONNECTIVITY` contract, the gate's copper-union pour term, and moves the CI job into the KiCad container so the fills actually run.

## Changes

- **`boards/07-matchgroup-test/generate_design.py`**
  - Declare module-level `POUR_NETS = ["GND", "+1V2", "+1V8"]`, `REQUIRE_POUR_CONNECTIVITY = True`, `MAX_POUR_REPAIR_ROUNDS = 6`.
  - Add recipe-local `_audit_pour_nets` (shapely copper-union audit, immune to the #3482 boundary-test illusion) + `_repair_pour_connectivity` (offset vias + stubs + island bridges) — mirrors board 06's steps 9-11.
  - Wire a **fill -> stitch -> repair/re-fill/audit loop** into `route_pcb` after zone generation, including the structural short-circuit when the filler is unavailable (every fill failing means no round can clear the zero-fill audit term).
  - `skip_nets` now derives from `POUR_NETS` so router-skip and the gate audit reference one declaration.
- **`scripts/ci/check_matchgroup_coverage.py`**
  - Add `measure_pour_connectivity` (mirrors `check_diffpair_coverage.py`) + `load_board_recipe_module`; the gate re-runs the copper-union audit against the re-routed artifact and **exits 2** on any disjoint pour net / zero-fill zone, **exit 1** if the audit cannot run (never a silent pass).
- **`.github/workflows/ci.yml`**
  - Move `matchgroup-routing-regression` into `kicad/kicad:10.0` (`--user 0:0`, bash default) with an early `kicad-cli --version` prerequisites probe; replace `setup-uv`/`setup-python` with the container one-liners; `timeout-minutes` 15 -> 20 for container pull + apt + real fills.
- **`tests/test_check_matchgroup_coverage.py`** (new, 11 tests) — pour audit term, board 07 contract, container CI job structure.

## Acceptance Criteria Verification

| Criterion (issue #3617) | Status | Verification |
[...]

Issue #3617, opened 2026-06-12
boards/07-matchgroup-test: pour zones are never filled and pour connectivity is never asserted


rjwalters:
Found during Judge review of PR #3615 (Issue #3509 sibling-job audit).

## Context

PR #3615 fixed board 06's CI job: the diffpair-routing-regression job ran on bare ubuntu-latest with no kicad-cli, so every `kct zones fill` failed and the pour audit could never pass. The fix moved the job into the `kicad/kicad:10.0` container, added an early `kicad-cli --version` probe, and made the gate assert the copper-union pour audit (`REQUIRE_POUR_CONNECTIVITY = True`).

## Board 07 status

The sibling Match-Group job (`matchgroup-routing-regression`) does NOT have the same defect — but only because board 07's recipe never attempts fills at all:

- `boards/07-matchgroup-test/generate_design.py` creates copper-pour zones in-process via `auto_create_zones_for_pour_nets` (step 4 of `route_pcb`), but never invokes `kct zones fill` — there is no kicad-cli dependency on its CI path (`--step route`).
- Consequently the routed artifact ships with zone outlines but zero fill geometry, and there is no `REQUIRE_POUR_CONNECTIVITY` contract / `_audit_pour_nets` audit for the gate to assert.
- `scripts/ci/check_matchgroup_coverage.py` has no pour-connectivity term.

This is the same "dead pour" class of gap board 06 just closed: zones that exist but carry no copper. Per the manufacturable bar (100% LVS + 0 DRC), unfilled pours on a power/GND net are a latent connectivity hole.

## Suggested work

1. Add a fill + pour-connectivity-repair pipeline to board 07's recipe (mirror board 06 steps 9-11, including `MAX_POUR_REPAIR_ROUNDS` and the structural short-circuit from PR #3615).
2. Declare `REQUIRE_POUR_CONNECTIVITY = True` + expose `_audit_pour_nets`.
3. Move `matchgroup-routing-regression` into the `kicad/kicad:10.0` container with the early `kicad-cli --version` probe (same pattern as PR #3615), and re-measure the 15-minute ceiling (current wall ~13m25s; container adds ~1-2 min, so the ceiling likely needs a bump).
4. Add the pour term to `scripts/ci/check_matchgroup_coverage.py` (exit 2 on disjoint/zero-fill, exit 1 on cannot-run — mirror `measure_pour_connectivity` in `check_diffpair_coverage.py`).

Issue #3556, opened 2026-06-12
CI gates miss trace-vs-zone-fill clearance violations (stale pour carves)


rjwalters:
## Gap description

While judging PR #3548 (board 06 USB_CC1 barrel-graze repair, issue #3515), I found that a surgical trace move left the trace only **0.0347 mm** from the committed In1.Cu GND pour fill (jlcpcb minimum 0.1016 mm) because the zone fill was not regenerated — the fill's clearance carve still surrounded the *old* trace corridor. KiCad's own DRC flags this as 2 zone-clearance violations, but **every project gate passed**:

- `scripts/ci/check_routed_drc.py` (strict gate): does not parse `(zone ...)` / `filled_polygon` at all — only segment/via/pad copper
- Board-06 pour audit (`TestPourCopperUnionAudit`): checks pour-net *connectivity* (copper union), not clearance of other-net copper to the fill
- #3487 all-layer barrel checker: vias only

So any artifact edit that moves copper without refilling zones can ship a sub-fab-minimum gap (or outright overlap) to a pour, and CI will stay green.

## Proposed fix

Add a trace/via-vs-zone-fill clearance check to the strict gate (or a new gate):

1. Parse `filled_polygon` per layer per zone net from the committed artifact
2. For every segment/via/pad of a *different* net on that layer, compute copper edge gap to the fill polygon edges
3. Flag gap < required clearance (jlcpcb 0.1016 mm) as an error; overlap as a short

Implementation notes:
- shapely is already an optional dependency used by the pour audit (`_audit_pour_nets`) — buffering segments by half-width and computing polygon distance is a few lines
- Expect a baseline of pre-existing violations on some boards (board 06's main artifact has zone-clearance findings from KiCad DRC at 0.13-0.30 mm against its 0.3 mm zone setting); the gate should use the fab minimum (0.1016) not the zone setting, and may need a per-board allowlist like the existing tolerance file
- Alternatively/additionally: a "stale fill" detector that re-derives expected carves around current copper and diffs against committed fill boundaries

## Reproduction

```bash
git fetch origin feature/issue-3515
git show origin/feature/issue-3515:boards/06-diffpair-test/output/diffpair_test_routed.kicad_pcb > /tmp/b06.kicad_pcb
kicad-cli pcb drc --severity-error --format json -o /tmp/drc.json /tmp/b06.kicad_pcb
[...]

PR #3466, opened 2026-06-10
fix(router): initial-pass grace pass eliminates the load-modulated budget cliff misattributed to #3442 (closes #3452)

rjwalters:
Closes #3452
Refs #3442, #3450, #3449, #3444

## What the bisection found

The issue ranked three #3442 in-loop suspects. A controlled bisection (same rotated board-05 input, same `--auto-layers --max-layers 4 --starting-layers 4 --backend cpp --seed 42 --skip-nets ... --timeout 420 --per-net-timeout 30` recipe, same machine, `PYTHONHASHSEED=0`) exonerates ALL THREE:

| run | conditions | board 05 result |
|---|---|---|
| HEAD unmodified | 2-way CPU load | **3/32** (reproduces the judge's collapse; DRC profile identical: 2 pad_seg + 1 pad_via, 0 seg_seg) |
| HEAD + all 3 in-loop pieces disabled via env toggles | same 2-way load | **3/32** (no change) |
| pre-#3442 parent `10b4d5f1` | idle | **21/32**, 0 seg_seg / 0 seg_via |
| HEAD unmodified | idle | **21/32**, 0 seg_seg / 0 seg_via (same as pre-#3442) |

Decisive structural fact: in every collapsing run the rip-up **iteration loop never executes** — both auto-layer attempts time out during the *initial pass*, and `timed_out` skips the loop entirely. The #3442 lex tuple and rip-up feed literally never run; demote-to-partial never fires (judge-confirmed); the `ignore_overflow` scoping only affects the post-loop optimizer (strictly more conservative, cannot remove reach).

## The actual mechanism: a wall-clock budget cliff

The initial pass routes nets in difficulty-agnostic order and hard-breaks on `check_timeout()`. On board 05, five pathological ISENSE escape searches burn ~190s of the 420s budget — after which the remaining **24 nets route in 9.6 seconds** (measured: 166.7s -> 176.3s). Whether the budget line lands before or after that 10-second window decides 28/32 vs 6/32 at the initial pass. That knife edge is load-modulated, which is why the judge's pre/post A/B split cleanly under their session load and why identical-code CI runs flap (#3450, the board-02/03 evidence in #3452).

## Fix: bounded tiered grace pass (both sequential initial passes)

[...]

PR #3653, opened 2026-06-13
fix(stitch): co-check cross-net stitch traces in the via-placement pass

### LINKED ISSUE #3633: stitch: cross-net clearance check misses pad-to-via traces (board 07 +18 DRC residual)
## Summary

`stitch_cmd.run_stitch` threads cross-net **placed vias** into each subsequent net's clearance check (`cross_net_placed_vias`) but emits the pad-to-via **traces** in a separate later pass. As a result a stitch via on net B is never checked against net A's just-placed stitch trace, so stitch geometry can land within the manufacturer clearance floor of foreign-net copper without being flagged at placement time.

## Impact (board 07)

On board 07 (`matchgroup_test`), whose DDR/MIPI/HDMI signal classes route at `clearance=0.10mm` (grazing the jlcpcb 4-layer floor of 0.1016mm), this produces **18 net-new DRC clearance violations** once real pour copper exists to short against:

- 11 `clearance_pad_via`
- 7 `clearance_segment_via`

All 18 are attributable by offending-via UUID to **drill-0.2 stitch vias** (verified independently during review of PR #3632). Board 06 escapes the limitation because its signal classes route at the roomier `clearance=0.15mm`.

This was newly *exposed* (not introduced) by PR #3632 / Issue #3617, which wired the pour fill -> stitch -> repair loop and so produced genuine pour copper for the first time. The pour-repair geometry that PR adds (drill-0.25 repair vias + width-0.15 stubs) is DRC-clean and contributes zero net-new violations. As a stopgap the board 07 re-route DRC floor was raised 34 -> 62 in `.github/routed-drc-tolerance.yml` with a per-family forensic justification.

## Proposed fix

Restructure the shared stitcher so pad-to-via trace emission is **interleaved into the via-placement clearance loop** (rather than a separate later pass), so each placed stitch via/trace is checked against all previously-placed cross-net stitch geometry.

## Done when

- `run_stitch` checks stitch traces against cross-net placed stitch vias at placement time.
- Re-routed board 07 DRC residual from the stitcher returns toward zero.
- The board 07 floor in `.github/routed-drc-tolerance.yml` can be tightened back toward ~44 (per the "tighten when (a)" clause in that file).

Filed as the documented follow-up to PR #3632 (judge review, 2026-06-13).

rjwalters:
## Summary

`run_stitch` threaded cross-net placed **vias** into each subsequent net's clearance check but emitted the pad-to-via **traces** in a separate later pass, so a stitch via on net B was never checked against net A's just-placed stitch trace. On dense boards routing near the manufacturer clearance floor (board 07: DDR/MIPI/HDMI at 0.10mm vs the jlcpcb 4-layer 0.1016mm floor) this let foreign-net stitch geometry land inside the clearance band, producing the ~18 residual cross-net clearance DRC violations (11 `clearance_pad_via` + 7 `clearance_segment_via`) that #3617 / PR #3632 grandfathered into the board-07 allowlist.

## Changes

- **`trace_to_track_segments(trace)`** (new helper): decomposes a pad-to-via `TraceSegment` (straight / dog-leg / extended-escape) into its constituent `TrackSegment` legs, mirroring the geometry `add_trace_to_pcb` writes, each tagged with the originating net.
- **`run_stitch`**: maintains a running pool of placed stitch vias and decomposed trace legs. Before each pad's placement, the obstacle lists (`other_net_tracks`, `other_net_vias`) are augmented with the **foreign-net** subset of that pool (filtered by net so a net never blocks its own later placements). Every via/trace is now co-checked against all previously-placed cross-net stitch geometry within the same pass — across all placement strategies (straight, dog-leg, extended-escape, micro-via).
- **Tests**: updated two dense-0402 fixtures that encoded the old over-eager placement; added `TestTraceToTrackSegments` (helper decomposition) and `TestCrossNetTraceCoCheck` (a stitch via that clears another net's via but would violate that net's stitch trace is now rejected/re-routed — proving the *trace*, not the via, does the work).

| Criterion (issue #3633 "Done when") | Status | Verification |
|---|---|---|
| `run_stitch` checks stitch traces against cross-net placed stitch vias at placement time | done | New per-pad augmented obstacle lists + `_record_placed`; `TestCrossNetTraceCoCheck::test_run_stitch_co_checks_trace_in_same_pass` asserts every NETB via clears NETA's stitch-trace legs by the clearance band |
| Re-routed board 07 DRC residual from the stitcher returns toward zero | depends on CI | Verified at the mechanism level (the 18 stitch-attributable violations are stitch-via-vs-foreign-copper, exactly what the co-check now rejects). Full board-07 count requires a load-variant route+fill regen; deferred to CI per the allowlist's documented re-route variance |
| board 07 floor in `.github/routed-drc-tolerance.yml` can be tightened toward ~44 | not done in this PR | See note below |

## Allowlist (board 07) — left unchanged

The board-07 floor (`matchgroup_test_routed.kicad_pcb: 120`) is **not** tightened in this PR. Reasons:

- The current `120` floor is dominated by the **#3556 zone-fill grandfather (+51)**, which this fix does not address; the stitcher residual is only ~18 of it.
- The re-route count is machine/load-variant (the file documents CI ~50 vs local 69, the #3466 cliff), so a locally-measured tighter floor would not be a safe CI gate.
- Validating the actual reduction needs a full board-07 route+fill regen (kicad-cli, multi-minute, load-dependent), which the issue notes should be confirmed by CI rather than guessed locally.

**Expected reduction once this lands + board 07 is regenerated:** the 18 stitch-attributable violations (and, per the "tighten when (a)" clause, the related quantization-dogleg cross-net grazes, which share the same single-pass interleave limitation now fixed) retire. A follow-up PR that regenerates the board-07 artifact should tighten the floor per the measured post-fix count.

## Test Plan

- `uv run pytest tests/test_stitch_cmd.py tests/test_stitch_via_halo.py tests/test_stitch_bga_outer_ring.py tests/test_stitch_mfr.py tests/test_build_stitch_step.py tests/test_thermal_stitch.py tests/test_board_05_thermal_stitch.py` — 310 passed
- `uv run pytest tests/test_pipeline_cmd.py tests/test_write_verification.py tests/test_grid_plane_net_halo.py tests/test_analysis_net_status.py tests/test_static_halo_via_gate_3566.py tests/router/test_preserve_existing.py tests/router/test_softstart_revb_u1_lqfp32_escape.py` — all passed
- `uv run ruff check src/kicad_tools/cli/stitch_cmd.py` — clean

Closes #3633


rjwalters:
## Changes Requested - CI failing (Type Check + Match-Group Regression)

Thorough review done. **The core implementation is correct and well-tested** — but two CI checks fail and both block approval. One is a mechanical type-error fix; the other is a real behavioral regression this PR introduced on board 07.

### What is correct (verified)

- **Decomposition fidelity** — `trace_to_track_segments` mirrors `add_trace_to_pcb` exactly for all three shapes: extended-escape (`[pad] + waypoints + [via]`), dog-leg (`pad -> intermediate -> via`, two legs), and straight (single `pad -> via` leg), with matching width/layer/net and the same branch order (`is_extended_escape` -> `is_dogleg` -> straight). No geometry divergence between what is co-checked and what is emitted.
- **Net-filtering** — `cross_net_stitch_tracks`/`cross_net_stitch_vias` filter on `net_number != pad.net_number` (and `v[3] != pad.net_number`), and `_record_placed` tags vias with `placement.pad.net_number` / trace legs via `trace.pad.net_number`. Foreign-net-only augmentation is correct; a net never blocks its own later placements. Confirmed.
- **Fixture widening (`test_build_stitch_step.py`)** — legitimate, not test-weakening. At `+/-0.51` the opposite-net 0402 pads sat ~0.48 mm apart edge-to-edge, so the (now correct) cross-net co-check would reject a via on each pad and turn the happy-path fixture into a no-via/skip case. Widening to `+/-1.5` restores headroom so both nets still get stitched. The test still asserts meaningful behavior (`"skipped" not in message`, plus the idempotency test). Fixing a fixture that encoded the buggy over-eager placement is the right call.
- **Unit tests** — `tests/test_stitch_cmd.py` 241 passed locally, including `TestTraceToTrackSegments` and `TestCrossNetTraceCoCheck`.

### Blocking issue 1 - Type Check (mypy) FAILS

New error in `src/kicad_tools/cli/stitch_cmd.py`:
```
List item N has incompatible type "tuple[float | None, float | None]"; expected "tuple[float, float]"  [list-item]
```
In `trace_to_track_segments`, the dog-leg branch appends `(trace.intermediate_x, trace.intermediate_y)` to a `list[tuple[float, float]]`, but those fields are typed `float | None` and the `is_dogleg` property doesn't narrow them for mypy. Fix by binding locals with an assert, e.g.:
```python
elif trace.is_dogleg:
    ix, iy = trace.intermediate_x, trace.intermediate_y
    assert ix is not None and iy is not None  # is_dogleg guarantees this
    points = [(trace.pad.x, trace.pad.y), (ix, iy), (trace.via_x, trace.via_y)]
```
(Do NOT add it to the mypy baseline.)

### Blocking issue 2 - Match-Group Routing Regression (board 07) FAILS

This is the important one. The gate did **not** fail on DRC count — DRC actually dropped to **102 (allowed 120)**, confirming the cross-net clearance fix retired the targeted violations. It failed on a **pour-connectivity regression** the tightened co-check introduced:
```
[quant] FAIL +1V8: 2 disjoint pad groups (largest 17/18)   stranded: ['U3.12']
[quant] FAIL GND:  2 disjoint pad groups (largest 155/156)  stranded: ['U2.47']
POUR CONNECTIVITY (post-quantize): FAIL
Gate failed (exit 2).
```
The more aggressive foreign-net rejection is now refusing stitch vias that were **load-bearing for pour connectivity** — U3.12 (+1V8) and U2.47 (GND) lose their bridge to the main pour island. This is the failure mode of co-checking against foreign-net stitch geometry without a fallback: when the only clearing placement is rejected, the pad is left stranded rather than re-routed.

This directly contradicts the PR's AC row "board 07 DRC residual returns toward zero — depends on CI": CI has now answered, and the fix over-corrects. **Leaving the board-07 allowlist unchanged was the right call (the 120 floor is dominated by the unrelated #3556 zone-fill grandfather, and the count is machine-variant) — but it does not rescue the PR, because the failure is the connectivity gate, not the DRC count.**

Suggested direction: when every cross-net-clearing placement strategy is exhausted for a pour-net pad, fall back to placing the via anyway (preferring the least-violating position) rather than stranding the pad — pour connectivity must win over a marginal cross-net clearance graze, since a stranded pour pad is a worse defect than an 0.1 mm clearance band intrusion. Add a board-07-style regression assertion (or a fixture reproducing the stranded-pad case) so the connectivity guarantee is covered by a unit test, not just the multi-minute CI route.

### Required before re-review
1. Fix the new mypy error (no baseline entry).
2. Resolve the pour-connectivity regression so board 07's route+fill keeps every pour net as one copper component (no stranded U3.12 / U2.47), while preserving the cross-net DRC reduction.
3. Confirm `gh pr checks 3653` is fully green.

Re-request review once these are addressed.

rjwalters:
## Judge feedback addressed (both blockers fixed)

### Blocker 1 — mypy (new error)
In `trace_to_track_segments` the dog-leg branch bound `intermediate_x`/`intermediate_y` (`float | None`) into a `list[tuple[float, float]]`; `is_dogleg` didn't narrow for mypy. Fixed with a local binding + assert:
```python
elif trace.is_dogleg:
    ix = trace.intermediate_x
    iy = trace.intermediate_y
    assert ix is not None and iy is not None  # guaranteed by is_dogleg
    points = [(trace.pad.x, trace.pad.y), (ix, iy), (trace.via_x, trace.via_y)]
```
The assert cannot fire on a legitimate path: `TraceSegment.is_dogleg` is `True` iff both intermediate coords are non-None. NOT added to the baseline.

- `uv run python scripts/ci/check_mypy_baseline.py` → exit 0 ("all within the baseline, no new type errors").

### Blocker 2 — pour-connectivity regression (the serious one)
The tightened foreign-net rejection could **strand** a pour pad whose only via candidate grazed a just-placed sibling stitch via (board 07 `+1V8: U3.12`, `GND: U2.47` splitting into disjoint copper components).

**Fallback logic:** the placement-strategy cascade is factored into a `_attempt_placement(pad, tracks, vias)` helper. The pad loop now:
1. Calls it with the **cross-net-augmented** obstacle lists (unchanged behavior — keeps the DRC reduction).
2. If that returns `None` (every clearing strategy exhausted), retries with the **pre-existing copper only** (drops just-placed sibling stitch geometry from the obstacle pool). This reproduces the load-bearing placement that existed before the co-check tightened.
3. If the fallback succeeds, the via is placed anyway (least cross-net-violating position) and recorded in `StitchResult.connectivity_fallback` (logged in the CLI summary). If even the fallback finds nothing, it's a genuine obstruction and is skipped as before.

Connectivity wins over a marginal cross-net graze that DRC already grandfathers — a stranded pour pad is the worse defect.

**Regression tests** (`tests/test_stitch_cmd.py`):
- New `TestPourConnectivityFallback::test_stranded_pour_pad_rescued_by_fallback` — a pour pad with no clearing placement must still get a via via the fallback, recorded in `connectivity_fallback`, never in `pads_skipped`.
- `TestPourConnectivityFallback::test_non_stranded_pad_still_uses_clearing_placement` — when a clearing placement exists, the fallback must NOT fire (no over-broad rescue; DRC reduction preserved).
- Updated `test_run_stitch_multiple_nets` to assert the connectivity-preserving outcome (all pads stitched, none stranded).

### Verification
- `uv run pytest tests/test_stitch_cmd.py tests/test_build_stitch_step.py` → **262 passed**.
- `uv run python scripts/ci/check_mypy_baseline.py` → **exit 0**.
- **Match-Group Routing Regression (board 07) — run locally to completion (full re-route + fill + connectivity audit):**
  - `5 pour pad(s) used the connectivity fallback (placed despite a marginal cross-net clearance graze to avoid stranding the pad)`
  - `POUR CONNECTIVITY (post-quantize): PASS` — GND 156/156, +1V2 8/8, +1V8 18/18, each in **one copper component** (no stranded pads).
  - DRC error count **114 (allowed 120)** — cross-net DRC reduction preserved.
  - `[matchgroup-coverage] OK: pour connectivity (3 pour nets, copper-union audit).`
  - Gate exit code **0**.

Re-requesting review.

rjwalters:
## Approved — both blockers resolved, CI fully green

Re-review complete. The connectivity fallback and mypy narrow address both prior blockers, and CI now passes end-to-end.

### CI status (all green, merge state CLEAN)
- **Type Check (mypy): PASS** — the dog-leg branch in `trace_to_track_segments` now binds `ix/iy` locals + `assert ix is not None and iy is not None`. This is correct and cannot fire at runtime: `TraceSegment.is_dogleg` is `True` *iff* both `intermediate_x` and `intermediate_y` are non-None (stitch_cmd.py:200-202), so the only branch that reaches the assert has already guaranteed it. Not added to the baseline.
- **Match-Group Routing Regression (board 07): PASS (13m20s).** CI log confirms `POUR CONNECTIVITY (post-quantize): PASS` — GND 156/156, +1V2 8/8, +1V8 18/18, each one copper component. The previously-stranded `+1V8: U3.12` / `GND: U2.47` are rescued (the multi-round audit converges r1 FAIL -> r3/quant OK). DRC error count **94 (allowed 120)** — the cross-net clearance reduction is preserved (better than the local 114; machine-variant). `[matchgroup-coverage] OK: pour connectivity` and gate exit 0.
- Diff-Pair regression, Test, Lint, Round-trip Smoke, DRC, C++ Build: all pass.

### Fallback design — sound, and does NOT re-open #3633 for the common case
The cascade is factored into `_attempt_placement(pad, tracks, vias)` and invoked twice:
1. First with the **cross-net-augmented** obstacles (just-placed foreign stitch geometry participates) — the #3633 co-check governs every non-stranded pad unchanged.
2. Only if that returns `None`, a second call against **pre-existing copper only** (dropping just-placed sibling stitch geometry). This is the minimal relaxation: it can never place a via that violates real pre-existing copper — it only forgives a graze against a sibling stitch via that DRC already grandfathers. Connectivity beats a sub-clearance band intrusion; a stranded pour pad is the worse defect.

This is the exception, not the rule, and that is enforced by test, not just by the CI route: `TestPourConnectivityFallback::test_non_stranded_pad_still_uses_clearing_placement` asserts `connectivity_fallback == 0` when a clearing placement exists, and `test_stranded_pour_pad_rescued_by_fallback` asserts the rescue path fires (and records the pad) when it does not. The "fallback fired" count is tiny relative to net size (2 pads this CI run / 5 locally vs GND's 156) — genuinely unavoidable grazes, not an over-aggressive co-check.

The residual grazes are **observable**, not silent: rescued pads land in `StitchResult.connectivity_fallback` and are printed in `output_result` (`"~ N pour pad(s) used the connectivity fallback..."`), which is exactly what surfaces in the board-07 CI log.

Verdict: **APPROVE.** Strong work on the fallback — the source-level scrutiny, the bracketing unit tests, and the green board-07 connectivity+DRC gate all line up.

Issue #1018, opened 2026-01-23
feat: Automatic trace neck-down near fine-pitch pads


rjwalters:
## Problem

When routing to fine-pitch IC pads, the standard trace width (e.g., 0.2mm) may be too wide to fit between adjacent clearance zones. Professional PCB designs use "neck-down" - narrowing the trace as it approaches the pad.

## Current Behavior

Router uses uniform trace width for entire route, causing failures when:
- Trace width + 2×clearance > pin pitch
- Example: 0.2mm trace + 2×0.15mm clearance = 0.5mm, but TSSOP pitch is 0.65mm

## Proposed Solution

Implement automatic neck-down within configurable distance of pads:

```python
rules = DesignRules(
    trace_width=0.2,           # Normal trace width
    min_trace_width=0.1,       # Minimum for neck-down
    neck_down_distance=1.0,    # mm from pad center
    neck_down_auto=True,       # Enable for fine-pitch components
)
```

The router would:
1. Use normal trace width for most of the route
2. Taper to minimum width within `neck_down_distance` of fine-pitch pads
3. Create smooth transitions (not abrupt width changes)

## Visual Example

```
Normal trace (0.2mm)     Neck-down (0.1mm)      Pad
==================------======================[PAD]
                  ↑ Taper zone
```

## Acceptance Criteria

- [ ] Support configurable minimum trace width for neck-down
- [ ] Auto-detect when neck-down is needed based on pad pitch
- [ ] Create smooth width transitions (avoid DRC acute angle violations)
- [ ] Integrate with DRC to validate neck-down segments
- [ ] Option to disable for components that don't need it

PR #3657, opened 2026-06-13
fix(router): align C++ standard-mode is_via_blocked with Python (#864 parity)

### LINKED ISSUE #3622: C++ is_via_blocked standard mode diverges from Python _is_via_blocked (own-net is_obstacle cells rejected)

rjwalters:
## Summary

Fresh re-attempt of #3622 on current main (the prior PR #3630 was closed as stale because the root causes of its block are now fixed on main).

Aligns the third sibling standard-mode predicate, C++ `Pathfinder::is_via_blocked`, with the Python `_is_via_blocked` standard branch (#864: "same-net passable, different nets block" — it does NOT consult `is_obstacle`). Both standard-mode arms (cached fast path + per-net-radius slow path) change from `if (cell.is_obstacle || cell.net != net)` to `if (cell.net != net)`. Negotiated arms untouched. The sibling `is_trace_blocked` / `is_diagonal_blocked` predicates were already aligned in #3456 / PR #3623; this completes the family.

- `ROUTER_CPP_BUILD_VERSION` bumped 13 -> 14 in `cpp/include/types.hpp`; `_REQUIRED_CPP_BUILD_VERSION` mirror in `cpp_backend.py` matches; compiled `.so` reports `BUILD_VERSION = 14`.
- **Single-layer guard re-implemented** (was correct/necessary in the prior attempt): `CppPathfinder._apply_allowed_layers_to_routable()` intersects the C++ `routable_layers_` set with the `allowed_layers`-permitted indices. With one allowed layer the via loop's `new_layer == current.layer` skip leaves no via target — the real "a via requires >= 2 routable layers" invariant. Pure-Python; does not touch the parity predicate.
- **Neck-down regression does NOT recur**: the prior block (#864 relaxation suppressed fine-pitch taper because the C++ backend lacked neck-down) is fixed on main by #3623, which added #1018 neck-down post-processing to `cpp_backend.py::_convert_result_to_route`. Verified green.
- Pinned test `test_own_net_obstacle_rejects_via_standard` flipped to `test_own_net_obstacle_admits_via_standard`; added `test_foreign_net_obstacle_rejects_via_standard` counterpart; added `TestSingleLayerOwnNetObstacleNoVias` (end-to-end zero-vias + white-box routable-layer guard).

## The two prior block causes — both green

- `tests/test_router_core.py::TestSingleLayerRouting::test_single_layer_no_vias` — **PASS** (F.Cu-only route emits 0 vias).
- `tests/test_neck_down.py` — **26 passed**, `test_routing_fine_pitch_with_neckdown` produces neck-down segments again.

## Test plan

- [x] `tests/test_neck_down.py` — 26 passed
- [x] `tests/test_cpp_pathfinder_own_net_obstacle.py` — 6 passed (flipped + 2 new classes)
- [x] `tests/test_router_core.py` — 214 passed, 3 skipped (incl. single-layer-no-vias)
- [x] `tests/test_cpp_validation_parity.py` — green
- [x] via/pathfinder parity suites (`test_router_cpp_fallback_warning_3456`, `test_router_cpp_via_blocked_failure`, `test_pathfinder_via_foreign_clearance`, `test_router_pathfinder`, `test_static_halo_via_gate_3566`) — 60 passed
- [x] `uv run python scripts/ci/check_mypy_baseline.py` — exit 0 (a pre-existing `build_native_cmd.py` nanobind import-untyped finding is present on clean origin/main too — environment artifact, not introduced here)
- [x] C++ rebuilt with `--force`, `BUILD_VERSION = 14`

Closes #3622

### DIFF
--- a/src/kicad_tools/router/cpp/include/types.hpp
+++ b/src/kicad_tools/router/cpp/include/types.hpp
@@ -109,7 +109,26 @@ namespace router {
 // own-net-obstacle reject pinned by
 // tests/test_cpp_pathfinder_own_net_obstacle.py (verified NOT needed
 // for the board-03 fallbacks).
-constexpr int ROUTER_CPP_BUILD_VERSION = 13;
+//
+// Version 14 (Issue #3622): complete the standard-mode own-net-obstacle
+// parity family by aligning the THIRD sibling predicate,
+// ``Pathfinder::is_via_blocked`` (both the cached fast path and the
+// per-net-radius slow path), with the Python ``_is_via_blocked``
+// standard branch (Issue #864: "same-net passable, different nets
+// block"; it does NOT consult ``is_obstacle``).  Pre-#3622 the standard
+// arm rejected a via candidate whose Euclidean disc touched the routing
+// net's OWN ``is_obstacle`` pad copper (``cell.is_obstacle ||
+// cell.net != net``); a board that routes a via through its own
+// destination pad in the Python fallback but not in C++ is another
+// silent-fallback seed of the #3456 class.  Both backends now read
+// ``if (cell.net != net) return true;`` for the standard via predicate.
+// The single-layer (``allowed_layers``) invariant is preserved by a
+// Python-side routable-layer guard in
+// ``cpp_backend.py::CppPathfinder._apply_allowed_layers_to_routable``
+// (a via needs >= 2 routable layers), and fine-pitch neck-down survives
+// via the #1018 post-processing #3456/PR #3623 added to
+// ``cpp_backend.py::_convert_result_to_route``.
+constexpr int ROUTER_CPP_BUILD_VERSION = 14;
 
 // Grid cell state
 struct GridCell {
--- a/src/kicad_tools/router/cpp/src/pathfinder.cpp
+++ b/src/kicad_tools/router/cpp/src/pathfinder.cpp
@@ -572,7 +572,19 @@ bool Pathfinder::is_via_blocked_diag(int x, int y, int net, bool allow_sharing,
                     }
                     // Allow with cost for routed cells / own-net obstacle.
                 } else {
-                    if (cell.is_obstacle || cell.net != net) {
+                    // Issue #3622: same-net cells are passable in standard
+                    // mode regardless of the obstacle flag -- parity with the
+                    // Python ``_is_via_blocked`` standard branch (Issue #864:
+                    // "same-net passable, different nets block"; it does NOT
+                    // consult ``is_obstacle``) and with the sibling
+                    // ``is_trace_blocked`` / ``is_diagonal_blocked`` standard
+                    // predicates aligned in #3456.  Pre-fix, ``cell.is_obstacle``
+                    // here rejected a via candidate whose disc touched the
+                    // routing net's OWN ``is_obstacle`` pad copper, so a board
+                    // that routes a via through its own destination pad in the
+                    // Python fallback would FAIL in C++ and silently fall back
+                    // (the exact #3456-class silent-fallback seed).
+                    if (cell.net != net) {
                         return true;
                     }
                 }
@@ -620,7 +632,10 @@ bool Pathfinder::is_via_blocked_diag(int x, int y, int net, bool allow_sharing,
                             return true;
                         }
                     } else {
-                        if (cell.is_obstacle || cell.net != net) {
+                        // Issue #3622: same-net cells passable in standard
+                        // mode (Python #864 parity -- see the cached fast-path
+                        // branch above for the full rationale).
+                        if (cell.net != net) {
                             return true;
                         }
                     }
--- a/src/kicad_tools/router/cpp_backend.py
+++ b/src/kicad_tools/router/cpp_backend.py
@@ -47,7 +47,7 @@ logger = logging.getLogger(__name__)
 # ``AttributeError`` deep in the routing code (e.g. ``router_cpp.PadBounds``
 # missing).  The guard below catches that at import time and falls back to the
 # pure-Python router with an actionable ``kct build-native`` hint.
-_REQUIRED_CPP_BUILD_VERSION = 13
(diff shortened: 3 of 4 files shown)


rjwalters:
✅ **Approved!**

Rigorous re-review given the prior attempt (#3630) was blocked twice. Both prior block causes were **independently reproduced as PASSING** with the C++ backend genuinely engaged — not taken on the builder's word.

## C++ backend build
- `uv run kct build-native --force` → `router_cpp.BUILD_VERSION = 14`
- Version bump consistent across all three sites: `types.hpp` (`ROUTER_CPP_BUILD_VERSION = 14`), `cpp_backend.py` (`_REQUIRED_CPP_BUILD_VERSION = 14`), compiled `.so` reports 14.
- `is_cpp_available() == True` in the test environment used below.

## The two prior block causes — both green (reproduced)
- `tests/test_router_core.py::TestSingleLayerRouting::test_single_layer_no_vias` — **PASSED** (F.Cu-only route emits 0 vias). The single-layer guard works.
- `tests/test_neck_down.py` — **27 passed** (incl. `test_routing_fine_pitch_with_neckdown`). Neck-down regression does NOT recur (fixed on main by #3623, intact here).

> Note: `./.loom/scripts/run-tests.sh` is a loom-repo wrapper whose PYTHONPATH override makes `is_cpp_available()` return False and silently runs the Python fallback. I re-ran all C++-path tests via plain `uv run pytest` to ensure the C++ predicate was actually exercised.

## Parity / regression suites (C++ path, via `uv run pytest`)
- `tests/test_cpp_pathfinder_own_net_obstacle.py` — **6 passed** (flipped `test_own_net_obstacle_admits_via_standard`, new `test_foreign_net_obstacle_rejects_via_standard`, both `TestSingleLayerOwnNetObstacleNoVias` end-to-end + white-box guard).
- `tests/test_cpp_validation_parity.py` — 40 passed, 3 skipped (skips unrelated to this PR's scope).
- `tests/test_router_core.py` — **180 passed**.

## Code review findings
- **#864 parity is genuine.** Python `_is_via_blocked` standard branch (`pathfinder.py:1792-1797`) is `different_net = net_arr != net; if np.any(different_net): return True` — same-net passable, foreign blocks, does NOT consult `is_obstacle`. The C++ change makes both standard arms (cached fast path `pathfinder.cpp:587` and per-net-radius slow path `:638`) read `if (cell.net != net) return true;`. Now genuinely identical.
- **Negotiated mode untouched.** Both `allow_sharing` arms (`:553`, `:614`) retain `cell.is_obstacle && cell.net != net` plus the static-blocked / no-net / usage gates. Scope correctly limited to standard mode.
- **Single-layer guard is sound and does NOT wrongly narrow multi-layer boards.** `_apply_allowed_layers_to_routable()` returns early when `allowed_layers is None` (default preserved), and only narrows when `permitted != list(base)` AND `permitted` is non-empty. A 4-layer board allowing F.Cu+B.Cu still keeps both (vias allowed); only a strict single-layer subset collapses to one element. Both via loops (`pathfinder.cpp:1199` and `:1750`) iterate `for (int new_layer : routable_layers_)` with `if (new_layer == current.layer) continue;`, so a one-element routable set leaves no via target — the real "via requires >= 2 routable layers" invariant. The empty-`permitted` defensive skip is correct.
- **Flipped pinned test is correct** — the new admit assertion matches the Python standard branch; the foreign-net counterpart pins that the relaxation is same-net-only.

## Quality gates
- **CI: all 8 checks pass** — C++ Build, Diff-Pair Regression (board 06), Lint & Format, Match-Group Regression (board 07), Routed PCB DRC, Test, Type Check, kicad-cli smoke.
- **Merge state: CLEAN**, branch at tip of main (0 behind).
- `mypy` Type Check job **passed green in CI**. A local run flagged 2 findings in `build_native_cmd.py` (nanobind import-untyped) and `recovery/strategy.py` — **neither file is touched by this PR** (the 4 changed files are types.hpp, pathfinder.cpp, cpp_backend.py, the test); these are pre-existing local-environment artifacts, consistent with the PR's note and with CI's green Type Check.
- PR description uses `Closes #3622`.

Excellent, surgical fix that completes the standard-mode own-net-obstacle parity family (`is_trace_blocked` / `is_diagonal_blocked` were aligned in #3456/#3623; this is the third sibling). Comments are thorough and the regression-coexistence tests pin both the parity relaxation and the single-layer invariant.

Issue #3663, opened 2026-06-13
ruff lint burn-down: clear residual rules baselined by #3464 (F841, SIM105, F401 re-exports, etc.)


rjwalters:
## Context

PR for #3464 removed `continue-on-error` from the CI Lint & Format job so ruff now gates. As part of that change the bulk of the backlog was fixed (512 files reformatted, ~725 lint errors auto-fixed, 8 F821 undefined-name defects fixed manually). A residual set of lint rules was **baselined** (added to `[tool.ruff.lint] ignore` in `pyproject.toml`) rather than fixed, because they carry behavioral risk or require per-site judgement and were out of scope for the mechanical pass.

## Residual rules to burn down

| Rule | Count | Why baselined |
|---|---|---|
| F841 unused-variable | 123 | `--fix` is unsafe: can delete side-effecting calls (`result = do_thing()`); needs per-site review |
| SIM105 suppressible-exception | 15 | `try/except/pass` -> `contextlib.suppress`; mechanical but unsafe-fix, review per site |
| F401 unused-import | 7 | re-export positions in `__init__.py` (`router/__init__.py`); resolve via `__all__` |
| F601 multi-value-repeated-key-literal | 5 | duplicate dict keys; likely real bugs, review each |
| SIM115 open-file-with-context-handler | 4 | bare `open()` without `with`; behavioral, review each |
| F811 redefined-while-unused | 2 | `router/__init__.py` imports `RoutingResult` from both `.core` and `.strategies` (two different classes!) -- latent ambiguity, needs a decision on which wins |
| SIM101 duplicate-isinstance-call | 2 | merge isinstance calls |
| SIM116 if-else-block-instead-of-dict-lookup | 1 | |
| UP007 non-pep604-annotation-union | 1 | |

## Acceptance Criteria

- [ ] Each baselined rule code removed from `[tool.ruff.lint] ignore` in `pyproject.toml`
- [ ] All sites for that rule fixed (or individually `# noqa`'d with justification where a false positive)
- [ ] `uv run ruff check .` exits 0 with the rule re-enabled
- [ ] No behavioral changes (especially F841: do not delete side-effecting assignments; F811: verify which `RoutingResult` the public API should expose)
- [ ] `uv run pytest -m "not slow"` passes

Recommend burning these down rule-by-rule in small PRs (start with the 1-5 count rules; F841 last as the largest).

Issue #3588, opened 2026-06-12
router(2L): board 04 SWDIO/NRST have no legal 2L escape past LQFP-48 east-edge pad halos (capacity gap exposed by #3545)


rjwalters:
Follow-on from the #3582 bisect (see its comment trail for full data).

## Background
Pre-#3565, board 04's 2L fresh-route test reached 8/9 by routing SWDIO and NRST **through foreign pad metal** (-0.337mm overlaps vs GND/+3.3V pads on the STM32 east edge) — rip-up erased static pad halos and the validator's same-component skip masked it. PR #3565 made halos non-negotiable, so those corridors are gone and the router honestly demotes both nets: 7/9.

## Gap
With halos enforced, the negotiated loop no longer converges within the 240s budget for SWDIO/NRST:
- best-so-far restore carries 212 seg-seg violations
- correction pass cannot re-land either net
- both bisect runs hit the 240s loop timeout (iteration starvation, not proof of infeasibility)

## Ask
Find a LEGAL 2L routing for SWDIO/NRST past the LQFP-48 east-edge halo wall. Levers, in suggested order:
1. Iteration/wall-clock budget: the loop times out before exhausting strategies (#3538's iteration-budgeted gates are related)
2. Halo-aware escape routing: escape vectors that respect pad halos from the start instead of relying on rip-up to discover them
3. Placement nudge as last resort (board 04 is otherwise manufacturable on its production profile — do NOT regress the committed 4L/tier artifact)

## AC
- [ ] test_routes_all_nets_on_2l back to >=8/9 with 0 pad-clearance violations on the result
- [ ] No change to the committed board-04 production artifact (this is the fresh-route 2L recipe only)

Refs #3582, #3545, #3565, #3433.

PR #3664, opened 2026-06-13
ci: make Lint & Format job gate; clear ruff format/lint backlog

### LINKED ISSUE #3464: ci: Lint & Format job has continue-on-error on both steps (lint rot is invisible, 862 ruff errors / 505 unformatted files on main)

rjwalters:
## Summary

The CI "Lint & Format" job had `continue-on-error: true` on **both** ruff steps (`ruff format --check` and `ruff check`), so lint/format rot was invisible — ~860 lint errors and ~510 unformatted files had accumulated on main. This PR removes `continue-on-error` from both steps so the job now **gates**, and clears the backlog so the gate is green.

## Changes

**Gate flip** (`.github/workflows/ci.yml`):
- Removed `continue-on-error: true` from both the "Check formatting" and "Lint" steps; replaced the tracking comment with a gate-rationale comment.

**Format pass** (mechanical, AST-preserving):
- `uv run ruff format .` reformatted **512 files** (whole repo, including 2 notebooks whose embedded cells ruff formats).

**Lint auto-fixes** (~725 errors cleared):
- `ruff check . --fix` (safe fixes): 673 errors — I001 import sorting, F401, UP037, etc.
- A reviewed `--unsafe-fixes` subset restricted to mechanically-safe rules only (`E712`, `B905`, `C408`, `C416`, `UP007`, `SIM101/109/110/116`): 52 more. **No `F841` (unused-variable) unsafe fixes were applied in bulk** — those can delete side-effecting calls.

**8 F821 undefined-name defects fixed manually** (latent real defects, not suppressed blindly):
- `drc/report.py`, `erc/report.py`, `parts/models.py`: `TYPE_CHECKING` imports for quoted forward-reference annotations.
- `router/core.py`: 4 function-local `Segment`/`Via` annotations get `# noqa: F821`. A `TYPE_CHECKING` import was tried first but surfaced a pre-existing variable-reuse mypy false positive (a `restored` var reused for `list[Segment]` then `list[Via]`); `noqa` keeps core.py's typing surface unchanged.
- `test_optimize_placement_cmd.py`: imported `_write_placements_to_pcb_atomic` — the test called it without importing it, a NameError that `continue-on-error` had masked.

**Targeted F401**:
- `analysis/__init__.py`: added `build_zone_net_map` to `__all__` (intentional re-export).
- `explain/registry.py`: `# noqa: F401` on a yaml availability-probe import.
- `pcb/board_geometry.py`: dropped genuinely-unused `MultiPolygon` from a shapely probe import.

**Residual baselined** (`pyproject.toml [tool.ruff.lint] ignore`, burn-down tracked by #3663):
- `F841` (123), `SIM105`, `F401` re-exports, `F601`, `SIM115`, `F811`, `SIM101`, `SIM116`, `UP007` — these carry behavioral risk or need per-site judgement.

**Behavior preserved**:
- Restored a `_point_on_segment` re-export in `sch_pin_map.py` that the safe-fix pass removed (a test imports it through that module). This is the one autofix that would have changed behavior; it was reverted per the issue's instruction.
- `.github/mypy-baseline.txt`: updated one existing entry whose mypy message rendering drifted (same pre-existing error, **not** a new suppression).

| `continue-on-error: true` removed from both lint-job steps (gate live) | done | `.github/workflows/ci.yml` diff |
| `uv run ruff check .` exits 0 | done | exit 0 locally; residual rules tracked via `ignore` + #3663 |
| `uv run ruff format . --check` exits 0 | done | exit 0 locally (1456 files already formatted) |
| No behavioral changes (no bulk `--unsafe-fixes`; F841 left for follow-up) | done | F841 baselined; sch_pin_map re-export restored; smoke tests pass |
| 7-8 F821 fixed properly, not suppressed | done | TYPE_CHECKING imports + 4 justified noqa where a func-local import would create a mypy false positive |
| Test suite collects / passes unchanged | done | 20201 tests collect (0 errors); smoke subsets for all touched modules pass |
| mypy baseline clean | done | `scripts/ci/check_mypy_baseline.py` exit 0 |

## Test Plan

- `uv run ruff check .` and `uv run ruff format . --check` both exit 0.
- `uv run python scripts/ci/check_mypy_baseline.py` exits 0 (no new type errors).
- `uv run pytest tests/ --collect-only` collects 20201 tests, 0 errors.
- Smoke subsets pass for every module touched by autofixes/manual edits: `test_sch_pin_map.py`, `test_optimize_placement_cmd.py`, `test_drc_report_formats.py`, `test_board_geometry.py`, `test_trace_optimizer.py`, `test_stitch_cmd.py` (494 tests passed).

Closes #3464


rjwalters:
❌ **Changes Requested — CI Type Check gate is failing (one-line baseline fix)**

The core objective is solid and the autofix safety review came back clean, but the **Type Check** required job fails on CI, so this can't be approved yet. The cause is a single backwards edit to `.github/mypy-baseline.txt`.

## Root cause (reproduced locally with a clean mypy cache)

`scripts/ci/check_mypy_baseline.py` reports **1 NEW error beyond baseline**:

```
src/kicad_tools/recovery/strategy.py: Incompatible types in assignment
  (expression has type "def contains_point(self, _x: float, _y: float) -> bool",
   variable has type "def contains_point(self, x: float, y: float) -> bool")  [assignment]
```

This is the *same* pre-existing error already in the baseline at line 800 — the `in_area = failure_area.contains_point` (params `x,y`) then `in_area = lambda _x, _y: False` reassignment in `strategy.py`. The baseline matcher does exact-string matching, and **this PR edited that baseline line in the wrong direction**:

```diff
-src/kicad_tools/recovery/strategy.py	assignment	... "def contains_point(self, _x: float, _y: float) -> bool" ... "def contains_point(self, x: float, y: float) -> bool" ...
+src/kicad_tools/recovery/strategy.py	assignment	... "def (_x: float, _y: float) -> bool" ... "def (x: float, y: float) -> bool" ...
```

main already had the **with-function-name** rendering (which is what current mypy emits). The PR replaced it with the older nameless rendering, so CI's mypy output no longer matches → counted as NEW → gate fails. (The PR body's claim "mypy baseline clean" and "updated one existing entry whose rendering drifted" is inverted — the drift fix went the wrong way. Note: a stale local `.mypy_cache` masks this; you must `rm -rf .mypy_cache` to reproduce, matching CI's fresh `uv sync`.)

## Required change (trivial)

Revert that one baseline line to main's content (keep the `contains_point(self, ...)` rendering):

```bash
git checkout origin/main -- .github/mypy-baseline.txt
# then re-apply any OTHER intended baseline edits, if there were any
rm -rf .mypy_cache && uv run python scripts/ci/check_mypy_baseline.py   # must exit 0
```

I verified: with main's version of that line restored, `check_mypy_baseline.py` exits 0 (1499 errors, all within the 1512 baseline). No source change to `strategy.py` is needed.

---

## Autofix / behavioral-safety review (PASS — no blockers here)

I focused on behavioral-change risk in the mass reformat+autofix, and it is clean:

- **`ruff format` (512 files)** — whitespace/AST-preserving; not a behavioral concern.
- **No bulk F841 removals.** `ruff check --select F841 --statistics` still reports 123 (baselined, not auto-deleted). No side-effecting RHS was dropped. ✅
- **B905** `zip(...)` → `zip(..., strict=False)` — `strict=False` is the exact default of bare `zip()`, pure no-op. ✅
- **E712** on `grid._blocked[i,j,k] == True/False` → truthiness/`not`. These are **scalar** numpy element accesses (single-element index → numpy scalar bool), so truthiness is equivalent; the unsafe whole-array case does not apply. ✅
- **C408/C416/UP007/SIM** spot-checked — behavior-preserving.
- **Re-export safety:** wrote an AST check across every changed `src/**/__init__.py` comparing the set of importable names main vs HEAD — **zero importable names lost.** The `_point_on_segment` re-export in `cli/sch_pin_map.py` is correctly preserved (`# noqa: F401`). No other re-export was wrongly dropped. ✅
- **F821 fixes:** the `TYPE_CHECKING` imports in `drc/report.py`, `erc/report.py`, `parts/models.py` are correct. The 4 `# noqa: F821` in `router/core.py` (`dict[int, list[Segment]]` etc. in `cleanup_artifacts`) are *behaviorally* safe because `from __future__ import annotations` (line 3) makes all annotations strings — never evaluated at runtime, so no NameError. Minor nit: the inline comment says "Segment imported func-locally", but `cleanup_artifacts` has no local `import Segment` — the names are only ever string annotations. Consider tightening the comment to reflect PEP 563 (not blocking).

## Baseline-of-real-bugs check (both pre-existing on main — OK to defer to #3663, but flag in burn-down)

- **F601** `erc/violation.py:123-127` — genuine duplicate dict keys (`lib_symbol_mismatch`, `footprint_link_issues`, `single_global_label` defined twice; the later entries silently override the earlier descriptions). Pre-existing on main, not touched by this PR. Correct to park (which description is canonical needs domain judgement), but **#3663 should fix, not just delete-one-key** — these are real wrong descriptions.
- **F811** `tests/test_pcb_sync_netlist.py:1707` — `class TestGetBoardEdgePosition` is defined **twice** (also at :377); pytest collects only the second, so the first class's tests are dead. Pre-existing on main. Worth calling out in #3663 as lost coverage, not just a lint nit.

Everything except the baseline line is good — flip that one line back and I expect a green Type Check. Re-request review and I'll re-evaluate (and confirm Test + routing regressions, which were still pending at review time).

rjwalters:
Feedback addressed — the one-line baseline blocker is fixed.

**What changed:** restored the single `recovery/strategy.py	assignment` entry in `.github/mypy-baseline.txt` to match `origin/main` (the with-function-name rendering `def contains_point(self, _x: float, _y: float) -> bool` ... `def contains_point(self, x: float, y: float) -> bool`). This was the only diff against main in that file, so a wholesale `git checkout origin/main -- .github/mypy-baseline.txt` was the correct, minimal fix — no other intended baseline edits existed.

**Verified with a FRESH mypy cache** (`rm -rf .mypy_cache`, matching CI's clean `uv sync`):
- `uv run python scripts/ci/check_mypy_baseline.py` → exit 0: "1499 errors, all within the baseline (1512 allowed). No new type errors."
- `uv run ruff check .` → exit 0 ("All checks passed!")
- `uv run ruff format . --check` → exit 0 ("1456 files already formatted")

The F601/F811 pre-existing defects you flagged remain deferred to burn-down issue #3663, as intended. Re-requesting review.

rjwalters:
## Changes Requested — baseline blocker is fixed, but the 512-file reformat broke the Test gate (4 failures)

Re-judged after Doctor commit `9cabef0c`. **The prior blocker is genuinely resolved** and most of CI is green — but the Test job is now RED with 4 failures, all introduced by this PR's mass reformat. `mergeStateStatus` is UNSTABLE.

### CI status (run 27476473253)
| Job | Result |
|---|---|
| Type Check (prior blocker) | **pass** (39s) |
| Lint & Format (both steps, gate now active) | **pass** (11s) |
| Diff-Pair Routing Regression (board 06) | **pass** (11m26s) |
| Match-Group Routing Regression (board 07) | **pass** (13m35s) |
| Routed PCB DRC / kicad-cli Smoke / C++ Build | pass |
| **Test** | **FAIL** (14m03s — 4 failed, 20067 passed) |

### Baseline fix verification (good)
`git diff origin/main origin/feature/issue-3464 -- .github/mypy-baseline.txt` is now **empty** — the file is byte-identical to `main`. The wholesale `git checkout origin/main` was the correct minimal fix: it reverted exactly the one backwards `recovery/strategy.py assignment` line and undid no legitimate baseline change. Type Check passes on the fresh CI cache, confirming the 1499≤1512 baseline holds. ✅

### The 4 Test failures (all caused by this PR — they pass on `main`, which is green at `e50da217`)

**1 & 2. `test_manufacturer_propagation_lint.py` — MFR001 allowlist line-number drift (the reformat moved the `DesignRules()` call sites; the allowlist was not re-refreshed).**
The reformat of `core.py` / `io.py` / `evolutionary.py` shifted the default-rule `DesignRules(...)` constructors, so `_ALLOWLIST` in this test now points at the wrong lines:
- `router/core.py`: allowlist `809` → actual call now at **807**
- `router/io.py`: allowlist `2847` → **2825**, allowlist `3448` → **3424**
- `router/algorithms/evolutionary.py`: allowlist `222` → **224**

`test_manufacturer_propagation_in_router_design_rules` flags the real (now-unlisted) sites as regressions, and `test_allowlist_entries_resolve_to_actual_call_sites` flags the stale entries. The allowlist comment was refreshed for issue #3588's shift but not for *this PR's own* reformat. **Fix:** update the four line numbers in `_ALLOWLIST` (807, 2825, 3424, 224) to match the reformatted sources.

**3. `test_ci_mypy_baseline.py::test_lint_job_continue_on_error_untouched`**
This scope-guard (added by sibling issue #3512) asserts the two Lint & Format steps *retain* `continue-on-error: true`, with the message "issue #3464's domain." Issue #3464 **is this PR**, and removing that flag is the whole point — so the guard is now correctly obsolete and must be **retired/updated in this PR** (delete it or invert it to assert the steps are now gating). It is not modified in the current diff.

**4. `test_router_dense_escape_softstart.py::test_softstart_generator_routes_via_kct_route`**
The test does a literal contiguous-substring match for `'"-m", "kicad_tools.cli", "route"'` in `boards/external/softstart/generate_design.py`. On `main` that is a single line; `ruff format` exploded the call args onto separate lines, so the contiguous substring no longer exists (the tokens are still all present, just line-split). **Fix:** either keep the three tokens on one line in `generate_design.py` (e.g. via a `# fmt: skip` on that call or a magic trailing-comma adjustment) **or** loosen the test to a whitespace-insensitive / token-sequence match. Prefer making the test robust to formatting since gating ruff format means future reformats will hit this again.

### Bottom line
The substantive lint/format/autofix safety review (no F841 bulk removals, re-exports preserved, F821/B905/E712 safe) and the baseline revert both stand. The only remaining work is the four reformat-induced Test breakages above — all mechanical. Re-request review once Test is green and `mergeStateStatus` is CLEAN.

rjwalters:
Feedback addressed — all 4 reformat-induced Test failures fixed (commit 741fd8ce). No source/workflow behavior changed; these are test-side repairs to the guards the 512-file ruff format moved out from under.

**1 & 2. MFR001 allowlist line drift** (`tests/test_manufacturer_propagation_lint.py`)
Refreshed `_ALLOWLIST` to the reformatted bare-`DesignRules()` sites (verified by grepping each file):
- `router/core.py`: `809` → **807**
- `router/io.py`: `2847` → **2825**, `3448` → **3424**
- `router/algorithms/evolutionary.py`: `222` → **224**

Both `test_manufacturer_propagation_in_router_design_rules` and `test_allowlist_entries_resolve_to_actual_call_sites` now pass.

**3. Obsolete continue-on-error guard** (`tests/test_ci_mypy_baseline.py`)
Inverted the now-obsolete `test_lint_job_continue_on_error_untouched` (sibling #3512 asserted the Lint & Format steps *retain* `continue-on-error: true`) into `test_lint_job_continue_on_error_removed`: it now asserts neither lint step carries `continue-on-error` (the gate #3464 establishes is live), plus a sanity check that the two ruff steps (`ruff format --check`, `ruff check .`) are present so the test still meaningfully guards the gate. Docstring updated to reflect #3464 flipping the gate.

**4. Softstart kct-route substring** (`tests/test_router_dense_escape_softstart.py`)
Took option (a) — robustified the test rather than fighting the formatter. `ruff format` line-split the argv in `generate_design.py`, so the contiguous `'"-m", "kicad_tools.cli", "route"'` match no longer existed. Now matches the three tokens individually (`"-m"`, `"kicad_tools.cli"`, `"route"`) plus the existing `subprocess.run(` guard — formatter-proof going forward.

**Verification (fresh state, in `.loom/worktrees/issue-3464`):**
- `uv sync --extra dev` ✅
- `uv run pytest tests/test_manufacturer_propagation_lint.py tests/test_ci_mypy_baseline.py tests/test_router_dense_escape_softstart.py` → **33 passed** ✅
- `uv run ruff check .` → **exit 0** ("All checks passed!") ✅
- `uv run ruff format . --check` → **exit 0** (1456 files already formatted — my test-3 edit was reformatted and committed clean) ✅
- `rm -rf .mypy_cache && uv run python scripts/ci/check_mypy_baseline.py` → **exit 0** (1499 ≤ 1512) ✅

Re-requesting review.

rjwalters:
## Approved — all 8 CI jobs green, mergeStateStatus CLEAN

Re-judged after Doctor commit `741fd8ce` (test-only: 3 files, no source/workflow behavior change). The prior 4 reformat-induced Test failures are all genuinely repaired, and the substantive lint/format/autofix-safety + mypy-baseline review from earlier rounds still stands.

### CI status (run 27476954137, head `741fd8ce`)
| Job | Result |
|---|---|
| Type Check | **pass** (42s) |
| **Lint & Format** (now-live gate — both ruff steps gating) | **pass** (12s) — proves the whole tree is lint + format clean |
| **Test** (was 4 failures) | **pass** (13m06s) |
| Diff-Pair Routing Regression (board 06) | **pass** (10m44s) |
| Match-Group Routing Regression (board 07) | **pass** (13m37s) |
| Routed PCB DRC Check | **pass** (14s) |
| C++ Build Check | **pass** (47s) |
| kicad-cli Round-trip Smoke | **pass** (54s) |

`mergeStateStatus`: **CLEAN**.

### Focused verification of the 4 fixes

**1 & 2. MFR001 `_ALLOWLIST` line refresh** — spot-confirmed each refreshed line against the PR-branch source; all four are genuine bare-`DesignRules()` fallback sites, not masked propagation:
- `router/core.py:807` → `self.rules = rules or DesignRules()` (`__init__` rules-arg default) ✓
- `router/io.py:2825` → `rules = DesignRules()` under `if rules is None` (route_pcb fallback) ✓
- `router/io.py:3424` → `rules = DesignRules(grid_resolution=0.1)` (conservative inner fallback) ✓
- `router/algorithms/evolutionary.py:224` → `DesignRules(**rules_dict) if rules_dict else DesignRules()` (worker rebuild) ✓

**3. Inverted continue-on-error guard** (`test_lint_job_continue_on_error_removed`) — **not a tautology.** It asserts `advisory == []` (no lint step carries `continue-on-error`) AND sanity-asserts both gating steps run (`ruff format` + `ruff check` present in the run blobs). Confirmed the PR-branch `ci.yml` actually removed both flags and keeps `ruff format . --check` + `ruff check .`. Re-adding the flag or deleting a ruff step would fail the test, so it genuinely guards that the gate is live.

**4. Softstart token-check robustification** — loosened the contiguous-substring match to a per-token check (`"-m"`, `"kicad_tools.cli"`, `"route"`), but **both load-bearing guards remain**: `assert "subprocess.run(" in text` and the negative regression guard `assert "router.route_all_negotiated(" not in text`. The combination still uniquely pins the `kct route` subprocess invocation and would still catch a regression back to the in-process path.

Substance was already approved; the gate is now live and the full tree is clean. Moving to `loom:pr`.

PR #3536, opened 2026-06-11
fix(schematic+report): auto-size sheet to content + footprint census + off-board supercap docs (closes #3530, closes #3531)

rjwalters:
Closes #3530
Closes #3531

### #3530 — schematic content overflows the declared sheet (corrected diagnosis)
- New `kicad_tools/schematic/models/paper.py`: A-series sizes + `select_paper_for_extent` ladder (A4→A3→A2→A1→A0, default 10 mm margin).
- `Schematic.content_bounds()` covers symbol bounding boxes, power symbols, wires, junctions, no-connects, all label kinds, and text notes.
- `Schematic.write()` now auto-sizes the declared paper to the content extent by default (`auto_size_paper=False` opts out but still warns on overflow). Never shrinks a declared sheet; warns and clamps at A0; warns on negative coordinates (paper size cannot fix those); leaves unmodeled paper strings (`USLetter`, `User …`) alone.
- **Paper-size decision for softstart: A0, not the A2 the issue comment estimated.** The comment measured 40 symbols to x=560; the actual rev B sheet has 83 symbols to x=810 and label/wire stubs to **x=849.6 mm** — past A1's 841 mm width. The regenerated `softstart.kicad_sch` declares A0 and `kicad-cli` renders the full design (verified visually; previously half the schematic was clipped).
- PDF embed scaling safety net already exists (`report.css` `img { max-width: 100%; }`) — the weasyprint PDF scales the A0 render to page width.

**Fleet schematic census** (content extent vs declared paper, lib_symbols excluded):

| Board | Declared | Extent (mm) | Fits? | Needs |
|---|---|---|---|---|
| 01 voltage-divider | A4 | 165×155 | yes | — |
| 02 charlieplex | A4 | 212×105 | yes | — |
| 03 usb-joystick | A4 | 174×183 | yes | — |
| 04 stm32-devboard | A4 | 290×230 | **no** | A3 |
| 05 bldc-controller | A4 | 373×300 | **no** | A2 |
| 06 diffpair-test | A4 | 190×231 | **no** | A3 |
| 07 matchgroup-test | A4 | 190×244 | **no** | A3 |
| softstart | A4 | 850×337 | **no** | A0 (now declared) |

Boards 04–07 inherit the fix automatically on their next recipe regeneration (regenerating their routed outputs is out of scope here).

### #3531a — footprint census (was 0 smd / 0 tht / 78 other)
[...]

Issue #3532, opened 2026-06-11
router: 192 arbitrary-angle segments on softstart (0.8°-13.7° off 45° grid) — repair/rescue passes don't quantize


rjwalters:
User-observed; verified: 192/4519 segments (4.2%) on `softstart_routed.kicad_pcb` are at arbitrary angles (samples: 0.8°, 1.8°, 2.0°, 2.3°, 2.6°, 2.9°, 3.9°, 5.1°, 6.4°, 7.1°, 8.5°, 10.2°, 10.8°, 13.0°, 13.7°) vs the expected 0/45/90/135 set.

A* on-grid output is naturally 45°-quantized — the arbitrary angles almost certainly come from post-route mutation passes that drag endpoints to raw coordinates: the step-10c/d repair passes (drill-to-drill + cross-net clearance repair from PR #3481), `partial_rescue`/completion stubs, pour-anchor stubs, or DRC-nudge displacement.

## AC
- [ ] Identify which pass(es) emit non-quantized segments (the 192 cluster locations will attribute them)
- [ ] All segment-emitting/mutating paths quantize to 45° (helper: snap displacement vectors to the 8-direction set; split into two-segment doglegs when needed)
- [ ] Policy test: fleet-wide angle census — every committed routed artifact 100% 45°-aligned (or documented exemption, e.g., true arc/teardrop features if any)
- [ ] Softstart re-checked: 0 DRC preserved after quantization
- [ ] Check boards 03/05/06/07 for the same class

Note: 45°-only is a manufacturability/quality convention; acute-angle copper can etch poorly.

PR #3668, opened 2026-06-13
fix(router): 45-align length-tuning meander emitter + tighten board-07 census

### LINKED ISSUE #3535: router: length-tuning meanders emit off-45° copper (board 07: 55 segments) + 2 unquantizable corridor chords (boards 06/07)
Follow-up to #3532 (45-degree quantization of post-route mutation passes).

#3532 fixed the pad-tail emitters, optimizer mutation passes, and repair passes at source, and quantized every committed routed artifact. Three residual classes could NOT be quantized in place and are pinned in `tests/test_fleet_45_census.py::DOCUMENTED_OFF_ANGLE`:

1. **Board 07 length-tuning meanders (55 segments, nets DQ0-DQ7/DM0)**: the `DDR_DATA_BYTE_0` match group is length-tuned to exactly 39.098 mm per member by sloped meanders. Doglegging them changes member lengths unevenly (e.g. DQ6 +1.149 mm) and trips `match_group_length_skew` (1.064 mm > 0.100 mm tolerance). Real fix: the match-group/length-tuning emitters should generate 45-degree trombone meanders instead of sloped segments, then re-tune.
2. **Board 06 corridor chord** (`e9af299d`, (14.5, 10.0)->(15.5, 8.0) PCB-local, 18.4 deg, net 7): both dogleg variants clip an adjacent via (`clearance_segment_via`). Needs a re-route of the net with the now-quantized emitters.
3. **Board 07 corridor chord** (`351d1137`, TMDS_D0 area, 11.8 deg): both dogleg variants violate intra-pair clearance or clip the adjacent via.

AC:
- [ ] Length-tuning / match-group meander generation emits 45-degree trombones only
- [ ] Board 07 re-tuned: 0 off-angle segments, `match_group_length_skew` still clean
- [ ] Boards 06/07 corridor chords re-routed 45-aligned with no DRC regression
- [ ] `DOCUMENTED_OFF_ANGLE` in `tests/test_fleet_45_census.py` emptied

rjwalters:
## Summary

The length-tuning trombone meander emitter produced off-45° copper: the
generator travelled along the host segment's **raw** direction, so when
A* handed it an off-axis host (rip-up/re-route staircases collapse to
diagonal-but-not-exactly-45 runs) every emitted leg inherited that
off-angle slope. Board 07 shipped 55 such segments on its `DDR_DATA_BYTE_0`
match group. This is the same off-45° class that #3532 (PRs #3536/#3537)
and the #3617 doctor addressed elsewhere — this PR applies the same
45-quantization discipline to the meander emitter, at source.

## Changes

- **`optimizer/serpentine.py` (`generate_trombone`)**: snap the
  along-segment travel direction to the legal 8-direction set
  (`snap_direction_8`); its perpendicular is then also a legal direction,
  so every internally-emitted leg is 45-aligned by construction. The
  closing exit leg (which can be off-angle because the loops advanced
  along the *snapped* direction, not the host's raw slope) is routed
  through the shared `dogleg_points` quantizer. Endpoints stay pinned, so
  connectivity is bit-for-bit preserved.
- **`match_group_tuning.py` (`_mirror_segments_about_centerline`)**: the
  pair-aware N-side mirror reflects the P-side meander about the pair
  centerline (arbitrary outer-normal axis) and grid-snaps each endpoint
  independently, which can rotate a leg off-angle. Each reflected leg is
  now re-quantized through the same dogleg helper.
- **`tests/test_fleet_45_census.py`**: removed board 07's
  `EXEMPT_TUNED_NETS` entry (now empty) and added a stale-entry ratchet
  so a future tuned-net exemption the artifact does not actually need
  fails the census.
- **`tests/test_length_matching.py`**: added a parametrized test asserting
  `generate_trombone` emits only 45-aligned copper across 6 host slopes
  (horizontal, vertical, exact-45, ~17°, ~73°, near-45) with endpoints
  pinned and the chain contiguous.

| Criterion (issue #3535) | Status | Verification |
|---|---|---|
| Length-tuning / match-group meander generation emits 45° trombones only | ✅ | New parametrized test: 0 off-angle legs for all 6 host slopes; unit-checked that an off-axis host previously produced 9/9 off-angle legs, now 0/9 |
| Board 07 re-tuned: 0 off-angle segments, `match_group_length_skew` clean | ✅ | Committed artifact census = 0 off-angle (1269 segs); `test_board_07_matchgroup_test.py` 48/48 pass incl. skew |
| Boards 06/07 corridor chords re-routed 45-aligned | ⚠️ partial | Board 07's TMDS chord (`351d1137`) already gone in committed artifact. Board 06's `e9af299d` diff-pair chord remains — both dogleg variants clip an adjacent via; needs a full net re-route (out of scope) — kept as honest `DOCUMENTED_OFF_ANGLE` residual |
| `DOCUMENTED_OFF_ANGLE` emptied | ⚠️ partial | Board 07 entries gone; board 06's diff-pair chord kept as documented residual rather than faked |

## Honest residual

Board 06's `e9af299d` corridor chord (18.4°, net 7) threads the diff-pair
landing corridor where both dogleg bulges clip a via. Quantizing it in
place introduces a `clearance_segment_via` violation; the real fix is a
re-route of that net, which is a separate diff-pair routing task beyond
the meander-emitter scope of this issue. It stays pinned in
`DOCUMENTED_OFF_ANGLE` and the ratchet will catch it the moment a re-route
removes it.

## Test Plan

- `tests/test_length_matching.py` — 54 passed (incl. 6 new 45-alignment cases)
- `tests/test_match_group_tuning.py` + pair-aware mirror tests — 135 passed
- `tests/test_fleet_45_census.py` — 9 passed (board 07 exemption removed)
- `tests/test_board_07_matchgroup_test.py` — 48 passed
- broader serpentine/optimizer/diffpair suite — 406 passed
- `ruff check .` clean, `ruff format . --check` clean

Note: the board-07 census tightening is verified against the committed
artifact (already 0 off-angle); board-07 full regen (kicad-cli, multi-minute,
PYTHONHASHSEED-deterministic) was not re-run to avoid churning the committed
artifact — the source emitter fix is unit-verified to keep regeneration clean.

The 1 pre-existing mypy "NEW" error (`build_native_cmd.py` nanobind
`import-untyped`) is environmental — it appears with this PR's changes
stashed and is caused by `kct build-native` installing nanobind in the
worktree; that file is untouched here.

Closes #3535

### DIFF
--- a/src/kicad_tools/router/match_group_tuning.py
+++ b/src/kicad_tools/router/match_group_tuning.py
@@ -1989,24 +1989,45 @@ def _mirror_segments_about_centerline(
         the N-side route.
     """
     from .primitives import Segment as _Segment
+    from .quantize import dogleg_points, is_45_aligned
 
     _ = p_net_id  # caller-side clarity
     mirrored: list[_Segment] = []
     for pseg in new_p_segments:
         rx1, ry1 = _reflect_point_about_axis(pseg.x1, pseg.y1, cx, cy, nx, ny)
         rx2, ry2 = _reflect_point_about_axis(pseg.x2, pseg.y2, cx, cy, nx, ny)
-        mirrored.append(
-            _Segment(
-                x1=_snap_to_grid(rx1, grid_resolution_mm),
-                y1=_snap_to_grid(ry1, grid_resolution_mm),
-                x2=_snap_to_grid(rx2, grid_resolution_mm),
-                y2=_snap_to_grid(ry2, grid_resolution_mm),
-                width=pseg.width,
-                layer=pseg.layer,
-                net=n_net_id,
-                net_name=n_net_name,
+        sx1 = _snap_to_grid(rx1, grid_resolution_mm)
+        sy1 = _snap_to_grid(ry1, grid_resolution_mm)
+        sx2 = _snap_to_grid(rx2, grid_resolution_mm)
+        sy2 = _snap_to_grid(ry2, grid_resolution_mm)
+        # Issue #3535: even though the P-side meander is 45-aligned by
+        # construction, reflecting it about the pair centerline (whose
+        # normal is the arbitrary outer-normal hint, not necessarily a
+        # legal routing direction) and snapping each endpoint to the
+        # grid INDEPENDENTLY can rotate a leg off the {0,45,90,135} set.
+        # Re-quantize each mirrored leg through the shared dogleg helper:
+        # an aligned leg stays a single segment, an off-angle leg becomes
+        # an exact 45-degree leg + axis leg.  The shared endpoints stay
+        # pinned, so the N-side chain stays contiguous and connected.
+        if is_45_aligned(sx2 - sx1, sy2 - sy1):
+            pts = [(sx1, sy1), (sx2, sy2)]
+        else:
+            pts = dogleg_points(sx1, sy1, sx2, sy2)
+        for (ax, ay), (bx, by) in zip(pts[:-1], pts[1:], strict=True):
+            if ax == bx and ay == by:
+                continue  # skip a degenerate zero-length leg
+            mirrored.append(
+                _Segment(
+                    x1=ax,
+                    y1=ay,
+                    x2=bx,
+                    y2=by,
+                    width=pseg.width,
+                    layer=pseg.layer,
+                    net=n_net_id,
+                    net_name=n_net_name,
+                )
             )
-        )
     return mirrored
 
 
--- a/src/kicad_tools/router/optimizer/serpentine.py
+++ b/src/kicad_tools/router/optimizer/serpentine.py
@@ -19,6 +19,7 @@ from enum import Enum
 from typing import TYPE_CHECKING, Literal
 
 from ..primitives import Route, Segment
+from ..quantize import dogleg_points, is_45_aligned, snap_direction_8
 from .geometry import segment_length
 
 if TYPE_CHECKING:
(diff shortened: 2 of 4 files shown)


rjwalters:
✅ **Approved!**

Clean, well-scoped fix for the off-45° length-tuning meander emitter (issue #3535). The 45-quantization discipline from #3532/#3617 is correctly applied at the meander source.

## Critical correctness: tuned length IS preserved after 45° quantization

This was the key risk — quantizing a length-tuning meander must not break length-matching. Verified it is architecturally safe because length-matching is measured against **real post-quantization geometry**, never a theoretical model:

- `generate_trombone` recomputes `length_added` from the actual emitted segments (`serpentine.py:400-403`), so the tuner is told the true quantized length.
- The single-ended tuner recomputes `new_length` via `LengthTracker.calculate_route_length(current_route)` and gates on `current_skew <= tolerance_mm` (`match_group_tuning.py:1144-1147`).
- The pair-aware path measures `new_p_length`/`new_n_length` post-mirror and gates lane skew (`match_group_tuning.py:2741-2746`).

So if the meander's real length drifted, the `match_group_length_skew` gate would catch it. **The Match-Group Routing Regression (board 07) CI check is green (13m55s) — confirming the skew gate holds against the quantized copper.** This is the dispositive evidence that quantization preserved length-matching.

## Connectivity / symmetry
- Snapping `(ux,uy)` to the 8-set keeps internal legs unit-length and 45-aligned by construction (perpendicular of an 8-dir is an 8-dir). Exit chord routed through `dogleg_points` with endpoints pinned exactly.
- N-side mirror re-quantizes each reflected leg through `dogleg_points`; shared endpoints stay pinned (chain stays contiguous), and the paired DRC self-check + lane-skew measurement validate symmetry on real geometry.
- New parametrized test asserts endpoint pinning + chain contiguity + zero off-angle legs across 6 host slopes — meaningful coverage of the A* staircase-collapse cases that triggered the bug.

## Honest residuals — verified, not faked
- Board 06's `e9af299d` diff-pair corridor chord legitimately stays in `DOCUMENTED_OFF_ANGLE` (both dogleg variants clip an adjacent via → needs a full net re-route, out of meander-emitter scope). The census ratchet will catch it the moment a re-route removes it.
- Board 07 exemption (`EXEMPT_TUNED_NETS`) genuinely emptied: local census passes against the committed artifact (0 off-angle on the tuned nets), and the new stale-entry ratchet prevents silent re-addition.

## Test Execution
- `tests/test_length_matching.py` + `tests/test_fleet_45_census.py` — ✅ 57 passed locally (incl. 6 new 45-alignment cases + census ratchet)
- `tests/test_match_group_tuning.py` + `tests/test_board_07_matchgroup_test.py` — ✅ 134 passed locally (incl. skew)

## CI Status (all green, merge state CLEAN)
- Match-Group Routing Regression (board 07): ✅ pass — length-skew gate green
- Diff-Pair Routing Regression (board 06): ✅ pass
- Test: ✅ pass
- Lint & Format (ruff): ✅ pass
- Type Check, C++ Build, Routed PCB DRC, kicad-cli Smoke: ✅ pass

PR description uses "Closes #3535". Solid work.

Issue #3543, opened 2026-06-11
[Parent #3508] Serpentine length-matching gaps: no legal space on USB3 escapes; exact-tolerance rounding (0.501 vs 0.500)


rjwalters:
## Context (measured: board 06 seed-42 run-4, 2026-06-11)

Two serpentine defects observed on shadow-constructed pairs:

1. **No legal serpentine space**: USB3_RX1 (skew 4.457 mm) and USB3_RX2 (4.580 mm) — `create_serpentine` (now grid-validated per #3508) rejects every candidate segment on the dense BGA-49 escape, so the pairs ship with ~9x-over-tolerance skew. Options: serpentine on a different segment of the route (currently only the longest/best segment is tried), amplitude reduction with more periods, or length-absorbing detours during shadow construction itself (extend the SHORTER side's path while building it, when the guide/shadow length delta is already known).
2. **Exact-tolerance rounding**: PCIE_RX serpentine targeted delta == max_delta (0.500 mm) and the artifact measured 0.501 mm -> `diffpair_length_skew` fired at +0.001 mm. Target ~90% of `max_length_delta` instead of 100%.

## Acceptance

- [ ] Serpentine targets <= 0.9 * max_length_delta (unit test with the PCIE_RX geometry)
- [ ] Serpentine falls back to alternative segments / smaller amplitude when the grid check rejects the primary candidate (unit test)
- [ ] Board 06 seed-42 re-route with shadow enabled: no `diffpair_length_skew` on pairs the constructor claims

## Refs

Parent #3508 (decomposition).

rjwalters:
Bounded part (exact-tolerance 0.501-vs-0.500 rounding + epsilon skew gate) landed in PR #3669 (Refs). Remaining 'no legal serpentine space on USB3 escapes' is coupled-convergence work — rolled up under epic #3672. Blocked on that.

rjwalters:
Bounded part (exact-tolerance rounding + epsilon skew gate) shipped in PR #3669. Remaining USB3-escape serpentine-space work is owned by epic #3672. Closing; tracked there.

PR #3669, opened 2026-06-13
fix(router): bank serpentine length-match headroom + epsilon skew-DRC at tolerance

rjwalters:
## Summary

Addresses the **bounded** half of issue #3543: the exact-tolerance rounding defect (`0.501` vs `0.500`) on length-matched diff pairs. Two complementary root causes:

1. **Tuner over-targets (the source of the 0.501 mm artifact).** `tune_diff_pair_skew` aimed the trombone at the partner's *exact* length (100% gap closure). Because loop length is quantized (45deg doglegs, integer loop counts), the artifact routinely overshot the partner by a fraction of a loop, landing the measured skew at 0.501 mm against a 0.500 mm budget. The tuner now deliberately under-targets by `0.9 * tolerance` (`SERPENTINE_TARGET_HEADROOM_FRACTION`), banking ~0.45 mm of headroom (at the 0.5 mm default) so a quantization overshoot stays comfortably in-bounds while the residual ~0.45 mm skew is itself still within tolerance. This is the issue's "Target ~90% of `max_length_delta` instead of 100%" guidance.

2. **DRC comparison had no float slack.** `diffpair_length_skew`'s bare `skew_mm > tolerance_mm` could spuriously fire on a pair routed to *exactly* the tolerance, because summing many segment lengths in floating point can land a few ULPs over (0.5 -> 0.5000000001). Added a `1e-6` mm epsilon (`SKEW_TOLERANCE_EPSILON_MM`) -- three orders of magnitude below the observed 0.001 mm overshoot and far below fab resolution -- so an at-tolerance pair passes while a genuine over-skew still fires.

## Changes

- `diffpair_length_tuning.py`: under-target the serpentine by `0.9 * tolerance_mm` (clamped to the entry skew). `target_length` still drives skew measurement / early-stop; only the value passed to `add_serpentine` is reduced. New `SERPENTINE_TARGET_HEADROOM_FRACTION = 0.9` constant (exported in `__all__`).
- `validate/rules/diffpair_length_skew.py`: compare against `tolerance_mm + SKEW_TOLERANCE_EPSILON_MM` (1e-6 mm) instead of the raw tolerance.
- Tests for both fixes (see below).

| Criterion (issue #3543) | Status | Verification |
|-------------------------|--------|--------------|
| Serpentine targets <= 0.9 * max_length_delta (PCIE_RX-style geometry) | done | `TestTargetHeadroom` asserts `add_serpentine` is aimed at `partner_length - 0.9*tol` (never the full partner length) and the tuned artifact skew stays <= tolerance |
| Exact-tolerance rounding (0.500 / 0.501) handled | done | `test_float_noise_at_tolerance_does_not_fire` (0.5 + float noise -> no fire) and `test_overshoot_beyond_epsilon_still_fires` (0.501 -> still fires); existing `test_at_tolerance_does_not_fire` (0.500 == 0.500) still green |
| Serpentine falls back to alternative segments / smaller amplitude on grid-reject | deferred | See "Out of scope" below -- this is the genuine routing-capacity half (#3508 family) |
| Board 06 seed-42 re-route with shadow enabled | deferred | Integration AC bound to the fallback above; shadow machinery is flag-gated behind `enable_shadow_construction` (#3508) |

## Out of scope (remains for #3508 / #3542)

The other half of #3543 -- **no legal serpentine space on the dense USB3 BGA-49 escapes** -- is *not* a bounded fix. It requires the grid-validated `create_serpentine` to try alternative route segments / reduced amplitude (or length-absorbing detours during shadow construction itself), which is part of the #3508 coupled-convergence family and gated behind `enable_shadow_construction`. I'm using **Refs** (not Closes) so #3543 stays open for that work.

## Test Plan

- `pytest tests/test_validate_diffpair_length_skew.py tests/test_validate_match_group_length_skew.py tests/test_diffpair_length_tuning.py tests/test_diffpair_serpentine_clearance.py tests/test_cli_check_diffpair_length_skew.py` -> 93 passed.
- `ruff check .` clean; `ruff format . --check` clean (1456 files already formatted).
- C++ backend built (`kct build-native`).

> Note on the mypy gate: building the native extension installs `nanobind`, which flips the pre-existing baseline entry for `src/kicad_tools/cli/build_native_cmd.py` from `import-not-found` to `import-untyped` (and clears 14 other baseline errors). This is a build-environment artifact in a file this PR does not touch -- none of the four changed files introduce a mypy error (verified by running mypy directly on the two edited source modules). The baseline was intentionally left unmodified.

Refs #3543

### DIFF
--- a/src/kicad_tools/router/diffpair_length_tuning.py
+++ b/src/kicad_tools/router/diffpair_length_tuning.py
@@ -132,6 +132,17 @@ class DiffPairTuneResult:
 # targets (very large skew on a congested board).
 MAX_INSERTS_PER_PAIR: int = 3
 
+# Fraction of the skew tolerance to deliberately leave un-closed when
+# sizing the trombone (issue #3543).  Closing 100% of the gap aims the
+# serpentine at the partner's exact length; because loop length is
+# quantized (45deg doglegs, integer loop counts) the artifact routinely
+# *overshoots* by a fraction of a loop, landing the measured skew just
+# *over* tolerance (the observed 0.501 mm against a 0.500 mm budget).
+# Under-targeting by 0.9x tolerance banks ~0.45 mm of headroom (at the
+# 0.5 mm default) so a quantization overshoot stays comfortably in-bounds,
+# while the residual ~0.45 mm skew is itself still within tolerance.
+SERPENTINE_TARGET_HEADROOM_FRACTION: float = 0.9
+
 
 def tune_diff_pair_skew(
     pair: DetectedPair,
@@ -256,6 +267,18 @@ def tune_diff_pair_skew(
         shorter_id, shorter_route, longer_id, longer_route = n_id, n_route, p_id, p_route
         longer_is_p = True
     target_length = max(l_p, l_n)
+    # Under-target by a fraction of the tolerance so quantized loop sizing
+    # does not overshoot the partner length and push the measured skew just
+    # *over* tolerance (issue #3543).  ``serpentine_target`` is what the
+    # trombone generator aims for; ``target_length`` remains the partner's
+    # true length and is used for skew measurement / early-stop below.  The
+    # headroom is clamped so it never reaches/inverts the gap on a pair
+    # whose skew is smaller than the headroom itself.
+    target_headroom = min(
+        SERPENTINE_TARGET_HEADROOM_FRACTION * tolerance_mm,
+        skew,
+    )
+    serpentine_target = target_length - target_headroom
     original_longer_segments = longer_route.segments  # for rollback assertion
 
     # Snapshot the original shorter route -- we may need to roll back.
@@ -307,7 +330,7 @@ def tune_diff_pair_skew(
         attempt_generator = SerpentineGenerator(attempt_config)
 
         candidate_route, serp_result = attempt_generator.add_serpentine(
-            current_shorter, target_length
+            current_shorter, serpentine_target
         )
         result.serpentine_results.append(serp_result)
 
@@ -601,6 +624,7 @@ def _empty_route(net_id: int, net_name: str) -> Route:
 # Re-export for ``from kicad_tools.router.diffpair_length_tuning import *``.
 __all__ = [
     "MAX_INSERTS_PER_PAIR",
+    "SERPENTINE_TARGET_HEADROOM_FRACTION",
     "DiffPairTuneResult",
     "tune_diff_pair_skew",
 ]
--- a/src/kicad_tools/validate/rules/diffpair_length_skew.py
+++ b/src/kicad_tools/validate/rules/diffpair_length_skew.py
@@ -98,6 +98,16 @@ if TYPE_CHECKING:
 # DDR5) MUST set ``NetClassRouting.skew_tolerance_mm`` explicitly.
 DEFAULT_SKEW_TOLERANCE_MM = 0.5
 
+# Float-comparison slack for the skew-vs-tolerance test.  A pair routed to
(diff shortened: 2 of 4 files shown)


rjwalters:
✅ **Approved!**

Both halves of the bounded exact-tolerance fix are correct and neither masks a real skew violation. Verified locally in the issue-3543 worktree.

## Concern 1 — epsilon does NOT mask real over-skew

`SKEW_TOLERANCE_EPSILON_MM = 1e-6` mm is the right magnitude:
- **Float noise covered:** ULP at ~0.5 is ~1e-16; summing many segment lengths accumulates at most ~1e-12. 1e-6 absorbs that with enormous margin.
- **Real violation still fires:** the observed defect is 0.001 mm (0.501 vs 0.500) — three orders of magnitude above the epsilon. `test_overshoot_beyond_epsilon_still_fires` (0.501) confirms **1 violation still raised**, while `test_float_noise_at_tolerance_does_not_fire` (0.5 + 1e-12) correctly does **not** fire, and existing `test_at_tolerance_does_not_fire` (0.500 == 0.500) stays green. 1e-6 mm = 1 nm, far below fab resolution (~0.1 mm min trace/space), so it can never hide a manufacturable over-skew.

## Concern 2 — 0.9-headroom does NOT under-tune into a new boundary problem

Aiming `add_serpentine` at `partner_length − 0.9·tol` leaves residual skew ~0.45 mm, which is **≤ the 0.5 mm tolerance** (the pair passes DRC). The quantization overshoot (a fraction of a loop) lands inside the banked 0.45 mm rather than crossing tolerance. Key correctness points verified in `diffpair_length_tuning.py`:
- `target_length` (true partner length) still drives skew measurement and the `current_skew <= tolerance_mm` early-stop (lines 354–366, 378–382); only the value passed to `add_serpentine` is reduced. So measurement is honest.
- The `min(0.9*tol, skew)` clamp is defensive: the tuner only engages when `skew > tolerance > 0.9*tolerance` (early-return at line 237), so the clamp always selects `0.9*tol` and never inverts the gap.
- `TestTargetHeadroom::test_tuned_skew_stays_under_tolerance` confirms the tuned artifact skew stays ≤ tolerance; `test_under_targets_below_partner_length` confirms the aim is `partner − 0.45`, never the full partner length that overshot pre-fix.

This is the correct tradeoff (minimize-skew → guarantee-in-bounds) and matches the issue's "Target ~90% of max_length_delta" guidance.

## Scope — Refs (not Closes) is honest

The deferred half (no legal serpentine space on the dense USB3 BGA-49 escapes) is genuinely the grid-capacity / coupled-convergence work in the #3508/#3542 family, gated behind `enable_shadow_construction`. That is not an easily-fixable part of the bounded rounding defect — leaving #3543 open via **Refs** is the right call, not scope-dodging. Merging this PR will not auto-close #3543, which is intended.

## Verification

- **CI:** all 8 checks green via `gh pr checks` — including the two length-matching regressions that these changes touch: Diff-Pair Routing Regression (board 06) pass, Match-Group Routing Regression (board 07) pass. Routed PCB DRC pass.
- **Merge state:** CLEAN.
- **Local scoped tests:** `test_validate_diffpair_length_skew.py` + `test_validate_match_group_length_skew.py` + `test_diffpair_length_tuning.py` → 74 passed, 1 xpassed. Coverage on the two changed rule modules: diffpair_length_skew 97% (only the unrelated `continue` line uncovered), match_group_length_skew 100%.
- **Lint:** `ruff check` clean on all four changed files.
- The mypy-baseline note in the PR body (nanobind flipping a pre-existing entry in a file this PR doesn't touch) is a build-environment artifact, not a regression; Type Check CI passes.

Clean, well-commented, well-tested bounded fix. Nice work.

Issue #3705, opened 2026-06-15
[Epic #3674] Fork KiCanvas + add KiCad-10 board-format support; vendor patched build


rjwalters:
## Goal

Make the interactive PCB viewer actually render our **KiCad-10** boards by forking KiCanvas (`theacodes/kicanvas`, MIT, actively maintained) and adding KiCad-10 board-format support, then vendoring the patched build into the site and submitting the change upstream.

## Background (from research)

- Our `.kicad_pcb` files are `(version 20260206)` (KiCad-10 format). KiCanvas's board parser has no definitions for these tokens: `tenting`, `covering`, `plugging`, `capping`, `filling` (pad/via tenting + soldermask), `duplicate_pad_numbers_are_jumpers` (footprint flag).
- The parser warns-and-skips unknown tokens and continues, so the missing defs alone don't abort — **the viewer hang is a downstream RENDER failure on the partially-parsed board.** Adding token defs is necessary but likely not sufficient; the actual render crash must be reproduced and fixed.
- Upstream is maintained (last commit ~2026-04-28) and merges community PRs; issue #186 reports KiCad-10 boards "none of them load." No fork is ahead of `main`.

## Plan

1. **Recon**: fork + clone `theacodes/kicanvas`; understand its build (it ships an unversioned `kicanvas.js` ESM bundle), the board parser (`src/kicad/board.ts`) + types, and reproduce the render failure with one of our KiCad-10 boards.
2. **Patch**: add the KiCad-10 board tokens to the parser/types; fix the downstream render failure.
3. **Verify headlessly**: a Playwright / headless-Chromium screenshot test that loads our `board.kicad_pcb` in the patched build and asserts the viewer canvas has rendered content (not blank / not stuck on spinner). This makes the fix verifiable without manual browser checks each iteration.
4. **Vendor**: rebuild and replace `site/public/vendor/kicanvas.js` with the patched build (kicad-tools change).
5. **Upstream**: open a PR to `theacodes/kicanvas` with the board-token support + render fix.

## Acceptance Criteria

- [ ] A KiCad-10 board (`version 20260206`) renders in the patched KiCanvas, verified by a headless-browser screenshot (non-blank canvas, recognizable board).
- [ ] The patched `kicanvas.js` is vendored into the site; the live detail page shows the interactive board (final confirmation in a real browser by the user).
- [ ] An upstream PR is opened against `theacodes/kicanvas`.
[...]

Issue #3691, opened 2026-06-15
[Epic #3674] PCB viewer polish — loading states, mobile, gallery affordance


rjwalters:
**Epic**: #3674 - kicad-tools.org demo board gallery website
**Phase**: 4 of 4
**Phase Goal**: Embedded interactive board exploration beyond static images.

## Description

Polish the interactive PCB viewer added in the sibling Phase 4 issue (#3690) and surface it from the gallery index.

Work:
1. **Loading state** — the WASM viewer can take a moment to initialize; show a loading indicator (and keep the static render visible until the viewer is ready) so the page never looks broken/blank.
2. **Error/fallback robustness** — if the viewer script fails to load or the `.kicad_pcb` fails to parse, fall back to the static render gallery without a console-breaking error or empty box.
3. **Mobile responsiveness** — the viewer must be usable on touch (pan/zoom), sized sensibly on narrow viewports, and must not trap page scroll.
4. **Gallery affordance** — on the index cards (`site/src/components/BoardCard.astro`), add a clear, subtle indicator that a board has an interactive view available (e.g. a small badge/icon), linking to its detail page. Boards without a `.kicad_pcb` should not show the indicator.

Out of scope: the core viewer embedding itself (#3690).

## Acceptance Criteria

- [ ] A loading state is shown while the viewer initializes; the static render stays visible until it's ready.
- [ ] Viewer load/parse failure degrades gracefully to the static renders (no blank box, no uncaught errors).
- [ ] The viewer is usable and correctly sized on mobile (touch pan/zoom; no scroll-trap).
- [ ] Index cards indicate when an interactive view is available and link to it; boards without a PCB omit the indicator.
- [ ] `npm --prefix site run build` succeeds; no generated artifacts committed.

## Dependencies

Part of Epic #3674, Phase 4. **Depends on the viewer-embed issue (#3690).**

rjwalters:
## Curator Enhancement

Curating this issue per user request. The core viewer embed from #3690 is merged to `main`; this issue covers the four remaining polish items. All file refs verified against `origin/main`.

---

### Scope Clarification

This issue is purely additive polish on top of #3690. The builder must not touch the `<kicanvas-embed src={pcbHref} controls="full">` declaration itself, the `copy-renders.mjs` PCB-staging logic, or the `hasPcb` build-time detection pattern — those are done. The four deliverables are:

1. **Loading state wrapper** around the existing `<kicanvas-embed>` in `[slug].astro`
[...]

PR #3706, opened 2026-06-15
fix(site): make PCB viewer loading overlay bulletproof against hung spinner

### LINKED ISSUE #3700: [Epic #3674] Investigate PCB viewer console errors + KiCanvas KiCad-10 format compatibility
**Epic**: #3674 - kicad-tools.org demo board gallery website

> **Scope note (updated 2026-06-15):** This issue has been narrowed by owner comment to exactly one deliverable — see comments for full history. Original report preserved below the divider.

## Problem

The interactive PCB viewer on board detail pages (e.g. `/01-voltage-divider/`) shows **"Loading interactive view…" with a spinner indefinitely** in a clean browser. The `~8s timeout` added in #3691 to degrade to the static renders is not visibly firing.

KiCanvas cannot fully parse KiCad-10 format boards (tracked separately in #3705), so its `load` event never fires. The fallback timeout is supposed to detect this and swap the spinner to an error message that points users to the `<RenderGallery>` section below — but in practice the spinner persists.

The minimum acceptable outcome: the viewer MUST NOT show an infinite spinner. It must either render the board OR fall back to the static renders within a bounded time.

## Root Cause (likely)

The inline IIFE in `site/src/pages/[slug].astro` has an early-exit null guard (`if (!embed || !overlay) return`) that executes **before** `window.setTimeout` is registered. If `kicanvas.js` fails to define the `<kicanvas-embed>` custom element (e.g. network error, parse error), `querySelector("kicanvas-embed")` returns `null`, the function returns early, and the timeout is never registered — so the spinner never hides.

Secondary issues:
- `.viewer-loading.viewer-error` CSS only changes `color`; the spinner `animation` is not stopped (no `display: none` guard for the `.viewer-spinner` in error state)
- `showError`'s `innerHTML` assignment has no catch block; a CSP violation would leave the spinner visible even after the timeout fires

## Acceptance Criteria

- [ ] The loading spinner disappears within the timeout window on any board detail page, even if KiCanvas fails to render or fails to load at all
- [ ] After the timeout, the `[data-viewer-overlay]` div is either removed from DOM (success path) or replaced with a legible error message (error path) — no indefinite spinner state is reachable
- [ ] The `window.setTimeout` call is the first statement (or wrapped so it cannot be skipped by any null guard or early return / throw)
- [ ] `showError` has a fallback that hides the overlay even if `innerHTML` throws
- [ ] `.viewer-loading.viewer-error .viewer-spinner` CSS rule sets `display: none` and `animation: none`
- [ ] The `<RenderGallery>` section remains visible and unaffected throughout all viewer states
- [ ] (Optional) Timeout shortened to 4–5s for better UX

## Affected Files

- `site/src/pages/[slug].astro` — inline `<script is:inline>` IIFE and `.viewer-loading.viewer-error` CSS rule; the only file that needs to change

## Test Plan

- [ ] Manual verification: open a board detail page (e.g. `/01-voltage-divider/`) in a clean browser profile (no extensions); confirm spinner disappears within the timeout window and static renders remain visible below
- [ ] Structural: `npm run build` inside `site/`; grep emitted HTML for a board-with-PCB page; confirm `[data-viewer-overlay]` div and `window.setTimeout` are present and the timeout is not positioned after any early `return`
- [ ] Code review: `window.setTimeout` is registered before any `if (!x) return` statement (Fix A)
- [ ] Code review: `showError` catches DOM/CSP errors and calls `overlay.style.display = 'none'` as last-resort (Fix B)
- [ ] Code review: CSS `.viewer-loading.viewer-error .viewer-spinner { display: none; animation: none }` rule present (Fix C)
- [ ] Regression: in a board where KiCanvas does render successfully, the overlay is removed promptly on the `load` event (not held until the timeout)

---

## Original Report (preserved)

On https://kicad-tools.org/01-voltage-divider/ the browser console shows a wall of messages while the embedded KiCanvas PCB viewer loads. Triage below — only item (1) is our code.

### 1. KiCanvas does not recognize KiCad-10 board-format tokens (investigated — separate fix in #3705)
KiCanvas logs many `kicanvas:parser: No definition found for element ...` warnings for tokens present in our `board.kicad_pcb` files (served `(version 20260206)`, generator `pcbnew`, i.e. a current KiCad-10 format). The vendored `site/public/vendor/kicanvas.js` predates this format version.

### 2. `navigator.credentials is undefined` TypeError — CONFIRMED non-issue
Confirmed browser extension in private browsing. Not our code.

### 3. Cloudflare Web Analytics beacon failure — CONFIRMED non-issue
Cloudflare auto-injection; harmless to site function.

rjwalters:
## Summary

The interactive PCB viewer detail page could display an indefinite "Loading interactive view…" spinner when KiCanvas failed to render the KiCad-10 board (its `load` event never fires). The fallback added in #3691 was defeated because the inline overlay script could return early on a null guard — or throw — **before** its `setTimeout` ever registered, so the spinner had no path to be hidden.

This makes the overlay bulletproof: the spinner is now always removed within a bounded time, regardless of KiCanvas state, CSP, or any script error, and the page degrades to the static `<RenderGallery>` (which stays on the page throughout).

Scope is limited to the overlay inline script and its CSS in `site/src/pages/[slug].astro`. The actual KiCanvas KiCad-10 rendering fix is tracked separately in #3705. The KiCanvas embed and render gallery are untouched.

## Changes

- **Register the hard timeout FIRST** (STEP 1), before any null-guard / early return / KiCanvas interaction. `overlay` is captured at the top via `document.querySelector(...)` (which cannot throw for a valid selector — returns `null` at worst), so the fallback always schedules.
- **Wrap the success-path setup in try/catch** (STEP 2). A thrown error (CSP blocking `innerHTML`, `<kicanvas-embed>` custom element undefined, etc.) can no longer prevent the timeout from firing. The early `return`s now live inside this try block, after the timeout is registered.
- **Timeout unconditionally hides the overlay.** `showError()` is `settled`-guarded and wrapped in try/catch: if the `innerHTML` swap throws, it falls back to `overlay.style.display = "none"` and `overlay.remove()`, so no spinner can persist.
- **CSS safety net:** added `.viewer-loading.viewer-error .viewer-spinner { display: none; animation: none; }` so the spin stops even if the DOM mutation is intercepted (e.g. by an extension's MutationObserver). Zero runtime cost.
- **Shortened the fallback from 8s to 6s** (assets are same-origin `/vendor/kicanvas.js`) and updated the fallback note to "Interactive view unavailable — see the renders below," pointing users at the static renders.

| Timeout registers FIRST, before any early return/null guard | ✅ | Extracted the inline script from built `dist/01-voltage-divider/index.html`: `setTimeout(..., 6000)` (STEP 1) precedes the `try` block; the only `return`s are inside that try, after the timeout. |
| Overlay logic wrapped in try/catch | ✅ | STEP 2 success-path setup is fully inside `try { ... } catch (e) {}`; `showError` has its own inner try/catch. |
| Timeout unconditionally HIDES the overlay (not just text swap) | ✅ | `showError` falls back to `display:none` + `remove()` on any throw; fallback note replaces spinner text and points at the static renders. |
| CSS stops the spinner in error/hidden state | ✅ | `.viewer-error .viewer-spinner{display:none;animation:none}` present in built `dist/_astro/_slug_.*.css`. |
| Shorter timeout (~6s) since assets are same-origin | ✅ | Timeout reduced from 8000ms to 6000ms. |
| Scope: overlay script + CSS in `[slug].astro` only | ✅ | `git diff --stat` shows 1 file changed (`site/src/pages/[slug].astro`). No artifacts committed. |
| `npm --prefix site run build` succeeds | ✅ | 10 pages built, "Complete!". |
| `npm --prefix site test` green | ✅ | 15/15 tests passed. |
| `npm --prefix site run check` clean | ✅ | `astro check`: 0 errors, 0 warnings (98 hints, all pre-existing on vendored `kicanvas.js` + the existing astro 4000 hint at line 184). |

## Test Plan

**Verified headlessly:**
- Built the site and extracted the emitted inline script from `dist/01-voltage-divider/index.html` — confirmed the `[data-viewer-overlay]` div is present and `setTimeout(..., 6000)` is registered before any `return`.
- Confirmed the CSS error rule is present in the built stylesheet.
- Reasoned through every exit path: (a) overlay null → timeout's `showError` early-returns harmlessly, nothing to hide; (b) placeholder/embed null → early return is inside try, timeout already scheduled → fires and hides; (c) `innerHTML` throws → catch hides via `display:none`+`remove()`; (d) thrown error in setup → outer catch, timeout still fires; (e) success → `load` event or MutationObserver removes overlay before the 6s timeout (which then no-ops via `settled`). No path leaves a hung spinner.
- `npm --prefix site run build`, `npm --prefix site test`, and `npm --prefix site run check` all pass.

**Needs a real-browser check (cannot be done headlessly here — no DOM/KiCanvas runtime):**
- Open a board detail page in a clean browser profile; confirm the spinner disappears within ~6s and is replaced by the fallback note while the static renders remain visible.
- Regression: on a board where KiCanvas does render, confirm the overlay is removed on the `load` event (not after the timeout).
- Optional CSP test: add a restrictive `<meta http-equiv="Content-Security-Policy">` and confirm the catch-block `display:none` path hides the overlay.

Closes #3700

### DIFF
--- a/site/src/pages/[slug].astro
+++ b/site/src/pages/[slug].astro
@@ -231,18 +231,22 @@ if (board.manifest_generated_at !== undefined) {
               <Fragment>
                 <kicanvas-embed src={pcbHref} controls="full" class="pcb-viewer" />
                 {/*
-                  Phase 4 polish (#3691): loading + error overlay.
+                  Phase 4 polish (#3691) + viewer-load fix (#3700): loading +
+                  error overlay.
 
-                  KiCanvas reflects boolean `loading`/`loaded` attributes on the
-                  <kicanvas-embed> host and dispatches a native CustomEvent("load")
-                  once the PCB renders. It dispatches NO error event, so the inline
-                  script below uses an ~8s timeout: if `loaded` never appears (slow
-                  fetch, parse failure, blocked/failed script), the overlay swaps to
-                  an error message that points users to the static renders above.
+                  KiCanvas signals completion via a JS PROPERTY `embed.loaded`
+                  (set true once the PCB has parsed + rendered). It does NOT
+                  reflect a `loaded` attribute and dispatches NO "load" or error
+                  event on the <kicanvas-embed> host — so the inline script below
+                  POLLS `embed.loaded` and removes the overlay (revealing the
+                  board) as soon as it is true. A hard ~8s timeout is registered
+                  first as a safety net: on timeout it reveals the board if it
+                  loaded, otherwise it swaps to an error note pointing at the
+                  static renders.
 
-                  The overlay starts visible and is removed on the "load" event;
-                  the static <RenderGallery> stays on the page the whole time, so a
-                  failed viewer never leaves a blank box.
+                  The overlay starts visible; the static <RenderGallery> stays on
+                  the page the whole time, so a failed viewer never leaves a blank
+                  box and the spinner can never hang indefinitely.
                 */}
                 <div class="viewer-loading" data-viewer-overlay role="status" aria-live="polite">
                   <span class="viewer-spinner" aria-hidden="true" />
@@ -275,21 +279,36 @@ if (board.manifest_generated_at !== undefined) {
     "#pcb-viewer-placeholder [data-viewer-overlay]"
   );
 
+  var embed = null;
   var settled = false;
-  var hideOverlay = function () {
-    if (settled || !overlay) {
-      settled = true;
-      return;
+  var pollTimer = null;
+  var stopPolling = function () {
+    if (pollTimer) {
+      window.clearInterval(pollTimer);
+      pollTimer = null;
     }
+  };
+  // KiCanvas signals completion via a JS PROPERTY \`embed.loaded\` (set true once
+  // the board has parsed + rendered). It does NOT reflect a "loaded" attribute
+  // and dispatches no "load" event on the <kicanvas-embed> host, so we must
+  // read the property — earlier attribute/event checks never fired, which is
+  // why the spinner hung over an already-rendered board.
+  var isLoaded = function () {
+    return !!(embed && embed.loaded);
+  };
+  var hideOverlay = function () {
+    if (settled) return;
     settled = true;
-    if (overlay.parentNode) overlay.remove();
+    stopPolling();
+    try {
+      if (overlay && overlay.parentNode) overlay.remove();
+    } catch (e) {}
   };
   var showError = function () {
-    if (settled || !overlay) {
-      settled = true;
-      return;
-    }
+    if (settled) return;
     settled = true;
+    stopPolling();
+    if (!overlay) return;
     try {
       overlay.classList.add("viewer-error");
       overlay.innerHTML =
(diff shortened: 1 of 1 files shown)


rjwalters:
**Approved (orchestrator review).** Updated to the real fix: KiCanvas signals completion via the JS property `embed.loaded` (not a reflected attribute or a host `load` event — verified in the KiCanvas source), so the overlay now polls `embed.loaded` and reveals the board; the hard timeout reveals-if-loaded else falls back to the static renders. Build green (10 pages), 15/15 tests, built JS confirmed to poll `embed.loaded`. Headless screenshot (fork agent) confirmed the board renders, so this should reveal a working viewer, not just stop the spinner.

Issue #3674, opened 2026-06-14
Epic: kicad-tools.org — demo board gallery website


rjwalters:
# Epic: kicad-tools.org — demo board gallery website

## Overview

Build **kicad-tools.org**, a public static gallery website that lets people explore
the demo boards in `boards/` and see how well the automated PCB tooling works. Each
board is presented with rendered images (2D layer plots + 3D ray-traced) and key
quality metrics (nets routed %, DRC violations, layer count, board size, part count,
cost), plus a downloadable manufacturing package.

This is an epic rather than a single issue because it spans three decoupled
subsystems — a board **render pipeline**, a **metrics extractor**, and a **static
site + deploy** — plus a future-facing **interactive PCB viewer**. Each can ship
and be tested independently.

## Decisions (made 2026-06-14)

- **Domain**: `kicad-tools.org` (purchased).
- **Hosting**: Cloudflare Pages — chosen for large binary artifacts (R2), per-PR
  preview deploys, and a path to dynamic/interactive features via Workers.
- **Site stack**: Astro (static-first, islands for future interactivity).
- **Renders**: both 2D layer plots AND 3D ray-traced PNGs via `kicad-cli`, front + back.
- **Decoupling**: render/metrics generation stays independent of hosting.

## Milestone Alignment

**Current Milestone**: unknown (none declared in README)
**Alignment Tier**: Tier 2 - Goal-Supporting
**Justification**: A public showcase of routing/manufacturing quality directly
demonstrates the project mission (fully automated agent-driven PCB creation) and
creates a feedback surface for board quality regressions.

### Phase 1: Foundation — renders + metrics
**Goal**: Produce per-board visual renders and a normalized metrics file from
existing artifacts. No site yet.
**Can parallelize**: Yes (the two issues are independent)

- [ ] Issue A: Board render command — `kicad-cli` 2D layer plots (front/back) + 3D ray-traced PNGs per board, to a predictable output path.
- [ ] Issue B: Board metrics extractor — emit a normalized `board.json` per board (nets routed %, DRC count, layer count, board size, part count, cost) from `output/manufacturing` artifacts.

### Phase 2: Static gallery site (Astro)
**Blocked by**: Phase 1
**Goal**: A browsable static site driven by the Phase 1 outputs.
**Can parallelize**: Partially (index and detail can follow the scaffold)

- [ ] Issue C: Astro project scaffold + board data loader (consumes per-board JSON + render image paths).
[...]

PR #3708, opened 2026-06-15
fix(site): emit viewer overlay script as raw JS so it actually runs

### LINKED ISSUE #3707: [Epic #3674] Viewer overlay <script> never executes — JSX {`…`} wrapper emitted as inert text
## Bug (root cause of the infinite spinner, both browsers)

The interactive-viewer loading overlay in `site/src/pages/[slug].astro` is wrapped as:

```astro
<script is:inline>{`(() => { ... })();`}</script>
```

Astro's `is:inline` does NOT process `{…}` JSX expressions, so the literal `` {` `` / `` `} `` template-literal wrapper is emitted into the page as **inert text**. The browser parses it as a bare block statement containing an unused template string — the IIFE never runs. Therefore neither the `embed.loaded` poll NOR the 8s hard-timeout fallback ever executes, so the spinner overlay is never removed.

Verified live (Playwright, both Chromium + Firefox) on https://kicad-tools.org/01-voltage-divider/:
- `embed.loaded === true` (the board DID render — KiCanvas works in both browsers).
- The `.viewer-loading` overlay is still present with the original "Loading interactive view…" text after 15s (timeout never fired).
- No JS errors — because the leaked `{`…`}` is *valid but inert* JS.

This has been latent since #3691; the #3700 logic fix changed code inside the inert string and so had no effect.

## Fix

Emit the overlay script as **raw executable JavaScript**, not a JSX-wrapped template literal. Either:
- write the IIFE as plain JS directly between `<script is:inline>…</script>` (no `{` `}` wrapper, no template-literal backtick escaping), or
- use `<script is:inline set:html={SCRIPT_STRING} />` with the JS in a frontmatter const.

Keep the same logic from #3700 (poll `embed.loaded`, reveal the board, 8s timeout fallback to the static renders). Since the board already renders (`embed.loaded` flips true), once the script executes the overlay will hide and reveal the board.

## Acceptance Criteria

- [ ] The built `dist/<slug>/index.html` `<script>` body is executable JS (starts with the IIFE, NOT `{` + backtick).
- [ ] Verified in a real browser (Playwright, Chromium + Firefox) against the deployed page: the spinner clears within a few seconds and the interactive board is visible; or, if `embed.loaded` never flips, the 8s fallback shows the static-render note.
- [ ] No JS console errors from the inline script.

## Dependencies
Part of Epic #3674. Supersedes the (ineffective) script-execution path; the #3700 overlay logic is otherwise correct.

rjwalters:
## Summary

Fixes #3707 (Epic #3674). The interactive-viewer loading overlay's `<script is:inline>` wrapped its IIFE in a JSX `{\`...\`}` template literal. `is:inline` does **not** interpolate `{…}`, so the literal wrapper shipped as **inert text** — the IIFE never executed. Neither the `embed.loaded` poll nor the 8s timeout ran, so the spinner hung forever over an **already-rendered** board (confirmed via Playwright in both Chromium and Firefox: `embed.loaded === true`, overlay never removed, no JS errors).

Fix: emit the IIFE as raw inline JS (no `{\`...\`}` wrapper, no template-literal escaping). Verified the built `dist/<slug>/index.html` `<script>` body now starts with `(() => {` and contains the real `setTimeout`/`setInterval` logic.

## Verification
- `npm run build` green; built script body starts with the IIFE (not the inert `{\`` wrapper); `npm test` + `npm run check` clean.
- Post-merge: will verify on the live deploy with headless Chromium + Firefox (spinner clears, board visible).

Closes #3707

### DIFF
--- a/site/src/pages/[slug].astro
+++ b/site/src/pages/[slug].astro
@@ -266,12 +266,14 @@ if (board.manifest_generated_at !== undefined) {
       {
         /*
          * Drive the loading → ready / error transition for the KiCanvas embed.
-         * Astro bundles + dedupes this <script>; it runs after the DOM is parsed.
+         * Raw inline JS (is:inline emits it verbatim — do NOT wrap in a {`...`}
+         * template literal, which is:inline would ship as inert text). Placed
+         * after the embed/overlay markup, so it runs once those nodes exist.
          * Emitted only when a PCB is present so no-artifact boards stay JS-free.
          */
         hasPcb && (
           <script is:inline>
-            {`(() => {
+            (() => {
   // Grab the overlay up front so the hard-timeout fallback can reference it
   // even if everything below throws. querySelector never throws for a valid
   // selector, returning null at worst.
@@ -288,7 +290,7 @@ if (board.manifest_generated_at !== undefined) {
       pollTimer = null;
     }
   };
-  // KiCanvas signals completion via a JS PROPERTY \`embed.loaded\` (set true once
+  // KiCanvas signals completion via a JS PROPERTY embed.loaded (set true once
   // the board has parsed + rendered). It does NOT reflect a "loaded" attribute
   // and dispatches no "load" event on the <kicanvas-embed> host, so we must
   // read the property — earlier attribute/event checks never fired, which is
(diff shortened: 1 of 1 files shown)


rjwalters:
Approved (orchestrator). Verified the built script body is now raw executable JS (starts with the IIFE, not the inert {`} wrapper); the prior version shipped the script as inert text, which is why the spinner hung over an already-rendered board in both browsers. Build/tests green.

PR #3709, opened 2026-06-15
build(board-00): generate + route + manufacture simple-led board

rjwalters:
## Summary

Board 00 (`00-simple-led`, the "hello world" LED board) shipped with only its `generate_design.py` generator and no built artifacts, so the gallery card fell back to the placeholder image. This builds it and commits the artifacts to match every other demo board.

Built via `kct build boards/00-simple-led` (with the C++ router): **100% routing complete** (3/3 nets — VCC/LED traced, GND via zone pour), **DRC PASS** (per `output/manufacturing/report.md`).

Adds `boards/00-simple-led/output/`: design (`.kicad_sch/pcb/pro/prl`), routed PCB, and the manufacturing bundle (gerbers, BOM, CPL, `report.md`/`.pdf`, layer/assembly images, `kicad_project.zip`, manifest). `board.json` and `renders/` remain gitignored (regenerated at deploy).

After this merges, board 00's gallery card shows real 2D/3D renders + accurate metrics instead of the placeholder.


rjwalters:
Superseded by the DRC-clean rebuild on top of #3712 (the zone-fill foreign-pad clearance fix). This PR's board-00 build predates that fix and fails the Routed PCB DRC gate (4 clearance_pad_zone errors). Closing in favor of the clean rebuild.

PR #3712, opened 2026-06-15
fix(zones): carve foreign-net antipads out of zone fills

### LINKED ISSUE #3711: kct zones fill: filled zones overlap foreign-net pads (clearance_pad_zone DRC errors)
## Bug

`kct zones fill` produces zone fills that **overlap foreign-net pads**, causing `clearance_pad_zone` DRC errors (a pad on net A sits inside the filled copper of net B). The zone's clearance is configured correctly (`connect_pads (clearance 0.3)`, `min_thickness 0.25`) — the **fill geometry itself** isn't subtracting the foreign-net pad clearance regions (antipads).

### Reproduce (board 00 — the simplest case)

`boards/00-simple-led/output/simple_led_routed.kicad_pcb` (a 3-component LED board) has **4** `clearance_pad_zone` errors:
```
pad on net 'GND' overlaps zone fill of net 'VCC' on F.Cu (overlap depth 0.013mm)
pad on net 'LED_ANODE' overlaps zone fill of net 'VCC' on F.Cu (overlap depth 0.045mm)
pad on net 'GND' overlaps zone fill of net 'VCC' on F.Cu (overlap depth 0.018mm)
pad on net 'LED_ANODE' overlaps zone fill of net 'GND' on B.Cu (overlap depth 0.200mm)
```
Re-running the filler does NOT fix it:
```bash
uv run --with shapely kct zones fill boards/00-simple-led/output/simple_led_routed.kicad_pcb -o /tmp/x.kicad_pcb
uv run --with shapely kct check /tmp/x.kicad_pcb --mfr jlcpcb --errors-only   # still 4 clearance_pad_zone errors
```
This also blocks the CI "Routed PCB DRC Check" gate for board 00 (see PR #3709). `kct fix-drc` cannot repair these (it only handles trace clearance + drill, not zone fills).

## Likely root cause

In `src/kicad_tools/zones/generator.py` the zone-fill polygon generation does not (correctly) subtract a `clearance + min_thickness/2`-buffered region around **foreign-net** pads (and possibly vias) before emitting `filled_polygon`. The fill should exclude an antipad around every pad NOT on the zone's net, so no foreign pad ends up inside the fill. Same-net pads get thermal/solid connection; foreign-net pads must get full clearance.

## Acceptance Criteria

- [ ] After `kct zones fill`, no `clearance_pad_zone` errors: `uv run --with shapely kct check <board>_routed.kicad_pcb --mfr jlcpcb --errors-only` reports 0 such errors for board 00 (and the fill still connects same-net pads).
- [ ] Foreign-net pads (and vias) are cleared by at least the zone clearance; verify the worst board-00 case (B.Cu, 0.200mm overlap) is resolved.
- [ ] A regression test (pure-Python, no kicad-cli) covering a small board where a foreign pad sits inside a pour, asserting the filled polygon excludes the pad + clearance.
- [ ] No regressions: other demo boards (01–07, softstart) that currently pass `kct check --mfr jlcpcb` still pass after the fix.

## Notes for the implementer

- `kct check` is **pure-Python DRC** (no kicad-cli) but needs the optional `shapely` dep — run via `uv run --with shapely kct check ...`. The fix itself is Shapely geometry in `zones/generator.py`; no kicad-cli or C++ router needed.
- The DRC rule is `clearance_pad_zone`; grep the DRC module for it to see exactly how overlap is measured (so the fill subtraction matches the check's expectation).

rjwalters:
## Summary

`kct zones fill` produced `filled_polygon` copper that overlapped foreign-net pads, raising `clearance_pad_zone` DRC errors. Board 00 (`simple_led_routed.kicad_pcb`) had **4** such errors (overlap 0.013–0.200 mm) that `kct fix-drc` cannot repair and that block the routed-PCB DRC gate.

**Which fill path was at fault:** kicad-cli (the `_run_fill_zones_via_drc` DRC-side-effect path, since KiCad 9 has no `pcb fill-zones` subcommand). The zone s-expression was already well-formed — `(connect_pads (clearance 0.3))` and `(min_thickness 0.25)` are present and correct (Hypothesis A was *not* the cause). kicad-cli simply leaves the antipad around foreign-net through-hole pads slightly too small on boards serialized by kicad-tools, and `kct check`'s `ViaZoneClearanceRule` (which the strict CI gate consults) flags the residual overlap even though KiCad's own DRC does not.

**The fix:** a deterministic, pure-Python post-fill geometric correction (`src/kicad_tools/zones/fill_clearance.py`) that, for every zone `filled_polygon`, subtracts an antipad of `clearance + min_thickness/2` around each pad/via on a *different* net. Same-net pads/vias are never cut, so thermal/solid connections are preserved. Vias are covered too (`clearance_via_zone`).

The correction operates on the parsed s-expression (no kicad-cli, no C++ router) and is a no-op when `shapely` is unavailable. It is wired into `runner.run_fill_zones`, so both `kct zones fill` and the post-route fill benefit.

## Changes

- `src/kicad_tools/zones/fill_clearance.py` (new): `apply_foreign_pad_clearance(doc)` carves foreign-net antipads out of each fill.
  - Resolves net identity by name so a name-only zone `(net "VCC")` matches a numbered pad `(net 1 "VCC")`; the net table map ignores name-less per-element `(net N)` nodes so they can't clobber real names.
  - Pads modelled by their (rotation-aware) axis-aligned bounding box — matching the DRC's pad-box convention; vias by their barrel disc on every spanned copper layer.
  - Holes are vented to the exterior with hair-thin (1e-4 mm) slits because KiCad fill rings cannot carry holes; the result is reconstructed exactly as the DRC reads it (`make_valid` / `_repair_fill_polygon`).
  - **Safety gate:** a rewrite is accepted only when it (a) removes the foreign overlap and (b) adds no copper outside the original fill. If the re-encode isn't faithful the original fill is left untouched, so the correction can only improve a board, never regress it.
- `src/kicad_tools/cli/runner.py`: call the correction after a successful fill (swallows any failure).
- `tests/test_zones_fill_clearance.py` (new): pure-Python regression tests (no kicad-cli).

| 0 `clearance_pad_zone` errors after fill on board 00 | ✅ | `kct zones fill … && kct check … --mfr jlcpcb --errors-only` → **0 errors, DRC PASSED** (was 4, incl. the 0.200 mm B.Cu case) |
| Foreign pads + vias cleared by ≥ zone clearance | ✅ | Regression tests assert `fill.distance(foreign) ≥ 0.3 mm` for both a foreign pad and a foreign via |
| Same-net connectivity preserved | ✅ | Test asserts same-net pad still overlaps its own fill; board-00 same-net pads still connect (thermal relief intact) |
| Pure-Python regression test (no kicad-cli) | ✅ | `tests/test_zones_fill_clearance.py` — builds an in-memory board, runs the correction, asserts geometry + round-trips through the real DRC rule |
| No regression on boards that currently pass | ✅ | Board 01 refill → `kct check --mfr jlcpcb` still **0 errors**. Board 05 (allowlisted, pre-existing 30 `clearance_pad_zone`) refill: **30 → 0** zone-clearance errors with **no new** via/pad-zone errors (remaining errors are pre-existing via-in-pad/connectivity, unrelated) |

## Test Plan

- `uv run --with shapely pytest tests/test_zones_fill_clearance.py` → 7 passed
- `uv run --with shapely pytest` on zone/fill/runner modules (test_validate_via_zone_clearance, test_zones_cmd, test_zone_priority, test_validate_zone_fill, test_route_cmd_zone_fill, test_route_zones_preserved, test_gerber_zone_fill, test_board_05_zones, test_runner_net_restore, test_build_zones_step) → 230 passed
- `ruff format` + `ruff check` clean on changed files
- kicad-cli still loads the corrected boards 00/05

Closes #3711

### DIFF
--- a/src/kicad_tools/cli/runner.py
+++ b/src/kicad_tools/cli/runner.py
@@ -355,10 +355,43 @@ def run_fill_zones(
 
     # If a future KiCad ships a native fill-zones subcommand, prefer it.
     if _kicad_cli_has_fill_zones(kicad_cli):
-        return _run_fill_zones_native(pcb_path, output_path, kicad_cli)
+        result = _run_fill_zones_native(pcb_path, output_path, kicad_cli)
+    else:
+        # Fallback: use DRC which fills zones as a side effect.
+        result = _run_fill_zones_via_drc(pcb_path, output_path, kicad_cli)
 
-    # Fallback: use DRC which fills zones as a side effect.
-    return _run_fill_zones_via_drc(pcb_path, output_path, kicad_cli)
+    # Post-fill geometric correction (Issue #3711): kicad-cli's fill can
+    # leave the antipad around a foreign-net pad/via too small on boards
+    # serialized by kicad-tools, producing clearance_pad_zone /
+    # clearance_via_zone DRC errors.  Subtract foreign-net antipads from
+    # every filled_polygon so the fill clears foreign copper by at least
+    # the zone clearance.  No-op when shapely is unavailable.
+    if result.success and result.output_path is not None:
+        _apply_foreign_pad_clearance(result.output_path)
+
+    return result
+
+
+def _apply_foreign_pad_clearance(pcb_path: Path) -> None:
+    """Carve foreign-net antipads out of every zone fill (Issue #3711).
+
+    Loads the filled PCB, applies the pure-Python clearance correction,
+    and rewrites the file only when at least one filled_polygon changed.
+    Any failure (e.g. shapely missing) is swallowed — the fill is then
+    left exactly as kicad-cli produced it.
+    """
+    try:
+        from kicad_tools.core.sexp_file import save_pcb
+        from kicad_tools.sexp import parse_file
+        from kicad_tools.zones.fill_clearance import apply_foreign_pad_clearance
+
+        doc = parse_file(pcb_path)
+        modified = apply_foreign_pad_clearance(doc)
+        if modified:
+            save_pcb(doc, pcb_path)
+    except Exception:
+        # Never let the correction abort a successful fill.
+        return
 
 
 def _run_fill_zones_native(
--- a/src/kicad_tools/zones/fill_clearance.py
+++ b/src/kicad_tools/zones/fill_clearance.py
@@ -30,9 +30,24 @@ from __future__ import annotations
 
 import math
 from dataclasses import dataclass
+from typing import Any, Protocol
 
 from kicad_tools.core.layers import via_spans_layer
 from kicad_tools.sexp import SExp
+
+
+class _Geometry(Protocol):
+    """Minimal structural type for the shapely geometry methods used here.
+
+    shapely ships no type stubs, so the concrete classes are ``Any`` to
+    mypy.  This Protocol captures only the surface we touch (``buffer`` and
+    ``intersects``) so ``_Obstacle.shape`` type-checks without a hard
+    dependency on shapely's (untyped) classes.
+    """
+
+    def buffer(self, distance: float, *args: Any, **kwargs: Any) -> Any: ...
+
+    def intersects(self, other: Any) -> bool: ...
 
 
 @dataclass(frozen=True)
@@ -41,8 +56,8 @@ class _Obstacle:
 
     net_key: str
     layers: tuple[str, ...]
-    # Sheet-absolute footprint of the copper, as a list of (x, y) rings.
-    shape: object  # shapely geometry; typed loosely to avoid a hard import
+    # Sheet-absolute footprint of the copper, as a shapely geometry.
+    shape: _Geometry
 
 
 def _build_net_name_map(doc: SExp) -> dict[int, str]:
@@ -112,7 +127,7 @@ def _collect_obstacles(
     are modelled by their circular barrel.  Net-0 (unassigned) copper is
     skipped — it does not participate in clearance checks.
     """
-    from shapely.geometry import Point
+    from shapely.geometry import Point  # type: ignore[import-untyped]
 
     obstacles: list[_Obstacle] = []
 
@@ -280,7 +295,7 @@ def _vent_holes(poly):
     Returns the list of resulting hole-free Polygons.
     """
     from shapely.geometry import LineString
-    from shapely.ops import nearest_points
+    from shapely.ops import nearest_points  # type: ignore[import-untyped]
 
     polys = _iter_polygons(poly)
     if not any(p.interiors for p in polys):
@@ -311,7 +326,7 @@ def _vent_holes(poly):
     if not slits:
         return [p for p in polys if not p.is_empty and p.area > 0]
 
-    import shapely
+    import shapely  # type: ignore[import-untyped]
 
     vented = poly.difference(shapely.unary_union(slits))
     return [p for p in _iter_polygons(vented) if not p.is_empty and p.area > 0]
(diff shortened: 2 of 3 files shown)


rjwalters:
## Judge: Changes Requested — Type Check (mypy) gate is failing

The geometry fix itself is correct and I verified it end-to-end on this machine (kicad-cli from KiCad.app):

- **Board 00**: `kct zones fill … && kct check --mfr jlcpcb --errors-only` → baseline 4 `clearance_pad_zone` → **0, DRC PASSED**. Confirmed.
- **Board 01**: refill + check → still **0 errors**. No regression.
- **Board 05**: refill → no new `clearance_pad_zone`/`clearance_via_zone`; same 23 pre-existing (14 via_in_pad + 9 connectivity). Not worse.
- **Tests**: `tests/test_zones_fill_clearance.py` 7 passed; related zone/fill/runner modules 137 passed, 1 skipped.
- **ruff** format + check clean on all 3 changed files.
- **Correctness review**: antipad modeling matches `ViaZoneClearanceRule` exactly (pad = axis-aligned box via `_axis_aligned_box_dims` mirroring `_transform_pad_dimensions`; via = `Point.buffer(r)`). Antipad size `clearance(0.3)+min_thickness/2(0.125)=0.425mm` is conservatively larger than the DRC threshold (`min_clearance_mm≈0.127`), so it cannot under-clear. Same-net pads/vias are skipped (connectivity preserved — verified by test + board 00 thermal relief). The faithfulness gate (`recon.intersection(cut_union).area>eps` reject, `recon.difference(fill_poly).area>eps` reject) is sound: it can only leave a fill unchanged or strictly improve it, never regress. Slit width 1e-4 mm is far below fab/DRC tolerance and is subtracted, not added, so no spurious sliver copper. No-pads / multi-hole / via cases handled.

### Blocker: CI **Type Check** job is red

`scripts/ci/check_mypy_baseline.py` fails on **10 NEW** errors in the new module `src/kicad_tools/zones/fill_clearance.py` (the workflow comment explicitly forbids padding `mypy-baseline.txt` — fix the errors or add targeted `# type: ignore[code]` for genuine false positives):

```
- Library stubs not installed for "shapely" / "shapely.geometry" / "shapely.ops"  [import-untyped]   (×3 — add `# type: ignore[import-untyped]` on the shapely imports)
- "object" has no attribute "buffer"  [attr-defined]   (_Obstacle.shape is typed `object`; annotate as the shapely geometry / `Any`, or cast at the buffer call)
- Argument N to "_obstacle_on_layer" has incompatible type "str | Any | None"; expected "str"  [arg-type]   (fill_layer can be None — narrow/guard before the call)
- Item "None" of "SExp | None" has no attribute "get_string"  [union-attr]   (the `zone.find("layer").get_string(...)` fallback assumes find() is non-None)
- Argument N to "list" of "SExp" has incompatible type "str | Any | None"; expected ...  [arg-type]
- Incompatible types in assignment (expression has type "float | None", variable has type "float")  [assignment]   (×2 — clearance / min_thickness reassigned from get_float which returns float|None; guard the None)
- Need type annotation for "polys"  [var-annotated]   (in `_reconstruct_fill`: `polys: list = []`)
```

Note: `mergeStateStatus` is UNSTABLE and the `Test` job is still pending — once Type Check is green, confirm `Test` also passes before re-requesting review.

Please address the mypy errors (no functional change needed — the geometry is correct) and flip back to `loom:review-requested`.

rjwalters:
Type Check gate fixed (types-only, no functional change).

Cleared all 10 new mypy errors the baseline gate reported for `src/kicad_tools/zones/fill_clearance.py`:

- Added a `_Geometry` Protocol (`buffer`/`intersects`) and typed `_Obstacle.shape` with it, so `.buffer()` type-checks (was `object` → `attr-defined`).
- Tagged the first import of each untyped shapely module with `# type: ignore[import-untyped]` (shapely ships no stubs; the accepted pattern). Removed the redundant ignores on subsequent same-module imports so none are flagged `unused-ignore`.
- Narrowed the `float | None` `clearance`/`min_thickness` reassignments via local temporaries.
- Guarded the `zone.find("layer")` fallback so `fill_layer` is always `str` (also fixes the `_obstacle_on_layer` arg-type error).
- Annotated `polys: list[Any]` in `_reconstruct_fill`.

Verified locally (CI-equivalent env, `uv sync --extra dev` so shapely is present):
- `uv run python scripts/ci/check_mypy_baseline.py` → `OK: mypy reports 1498 error(s), all within the baseline (1512 allowed). No new type errors.` (exit 0)
- `uv run --with shapely pytest tests/test_zones_fill_clearance.py` → 7 passed
- `uv run ruff format --check .` and `uv run ruff check .` → clean

The functional clearance correction is unchanged. Ready for re-review.

rjwalters:
## Judge verdict: APPROVE (re-judge after Doctor types-only fix)

This is the post-Doctor re-review. The sole prior blocker was the **Type Check** CI job (10 new mypy errors in `src/kicad_tools/zones/fill_clearance.py`); the functional clearance fix was already CONFIRMED correct in the prior pass.

**CI — all green.** All 8 checks pass on run 27569571603 (Type Check, Test, Lint & Format, Diff-Pair + Match-Group Routing Regressions, Routed PCB DRC, C++ Build, kicad-cli Round-trip Smoke). Merge state is CLEAN / MERGEABLE.

**Diff is types-only (verified `git diff 3085d8cf..0e3320a9`).** Doctor commit `0e3320a9` touches only `fill_clearance.py` and is purely annotations: a `_Geometry` Protocol typing `_Obstacle.shape` (so `.buffer()` checks), `# type: ignore[import-untyped]` on the untyped shapely imports, `float | None` reassignments narrowed via local temporaries (logic-equivalent), a None-guard on the `find("layer")` fallback (equivalent to the prior ternary), and `polys: list[Any]`. The fill geometry / antipad-subtraction logic is unchanged — no functional regression.

**Local sanity re-runs (clean worktree at PR head).**
- `scripts/ci/check_mypy_baseline.py`: 0 new errors in `fill_clearance.py` (all 10 resolved).
- `pytest tests/test_zones_fill_clearance.py`: 7 passed.
- `ruff format --check` + `ruff check`: clean on all three files.

**Scope.** Still exactly the three intended files (`fill_clearance.py`, `cli/runner.py`, `tests/test_zones_fill_clearance.py`); no committed generated artifacts.

Type Check is green, the diff is types-only with no functional change, and the previously-validated fix stands. Approving → `loom:pr`.

PR #3731, opened 2026-06-16
Board 03 (usb-joystick): regenerate to 0 combined-engine DRC + mark manufacturable

rjwalters:
## Summary
- Refilled the board-03 VCC/GND pours with `kct zones fill` (applying the merged #3728 solid-thermal + #3725 island-removal logic, refill-only -- no copper re-route) and regenerated the jlcpcb-tier1 manufacturing bundle, dropping the committed board to **0 errors under both DRC engines**.
- Cleared the 11 `starved_thermal` findings and the 14 grandfathered `clearance_pad_zone` stale-pour-carve shorts. The 16 `silkscreen_text_height` findings are warnings, not blocking.
- Tightened the guards to the strict 0 gate: removed the board-03 `tolerances:` entry from `.github/routed-drc-tolerance.yml` (absence = strict 0 gate; kept the `jlcpcb-tier1` `manufacturers:` override), and set `MAX_COMMITTED_DRC_ERRORS` 14 -> 0 in `tests/router/test_board03_routing_baseline.py`.

Same per-board Phase-B pattern as board-02 PR #3725 / #3724.

## Verification
- `kicad-cli pcb drc --severity-error` on the committed board: **0 violations**.
- Connectivity preserved (board-06 lesson): `unconnected_items` 15 -> 14 (did NOT increase), 0 `isolated_copper`.
- `python scripts/ci/check_routed_drc.py boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb`: **0 errors** at `--mfr jlcpcb-tier1` (passes strict 0 with the entry removed).
- `report.md` DRC Status = **0 / PASS** (combined engines), 100% routing (13/13).
- `uv run pytest tests/router/test_board03_routing_baseline.py tests/test_board_03_regression.py`: **21 passed, 1 skipped** with the ceilings at 0.

## Test plan
- [x] kicad-cli DRC = 0 on committed board
- [x] unconnected_items not increased; 0 isolated_copper
- [x] check_routed_drc.py passes strict 0
- [x] report.md DRC Status = 0 / PASS
- [x] board-03 guard test suites green

## Notes
- Did not commit `.kicad_prl` (gitignored).
- A pre-existing `test_routing_output_deterministic_across_seeds` failure (segment-band, fails on clean main) and `cmaes`-dependent placement test collection failures are pre-existing/unrelated and out of scope.

Closes #3730

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3733, opened 2026-06-16
fix(zones): selective per-pad solid connection replaces blanket-solid default

### LINKED ISSUE #3729: Refine zone pad-connection default: make solid-connect selective rather than global
## Context

PR #3728 (closes #3727) fixed `starved_thermal` DRC errors by making the zone generator default to a **solid** pad connection (`connect_pads yes`) for all pads, replacing thermal-relief-for-all. This is correct and honest — a solid connection is strictly stronger than a 2-spoke thermal, and `min_thermal_relief_spokes` is untouched. It cleared starved_thermal on boards 03/04/05/softstart with no connectivity regression.

## Refinement opportunity

The fix made solid the **global** default. This over-broadly changes behavior for pads that were already forming 2 valid thermal spokes — they lose thermal relief even though they didn't need the fix. For the current demo boards this is fine (the only zones generated are power/ground pours, where solid is the standard fab choice anyway), so the global default is defensible and was not a blocker.

A more surgical design would apply solid-connect **selectively**:
- Only to pads that physically can't host 2 spokes (small SMD pads, single-spoke THT power tabs), OR
- Via the already-plumbed `thru_hole_only` mode (solid SMD, thermal THT), OR
- Pad-size-gated selection in `normalize_zone_pad_connection`.

This would preserve thermal relief where it's both valid and beneficial (e.g. larger THT pads on boards that might see hand-rework) while still clearing starved_thermal on the pads that need it.

## Scope
- `src/kicad_tools/zones/fill_clearance.py` (`normalize_zone_pad_connection`) and/or `src/kicad_tools/sexp/builders.py` (`zone_node`)
- The `pad_connection` param is already exposed end-to-end, so this is an enhancement to the default-selection logic, not new plumbing.

Discovered during Judge review of PR #3728.

rjwalters:
## Summary

PR #3728 cleared `starved_thermal` by forcing a **solid** zone connection on
*every* pad (the global default).  This refines it to a **selective** policy
(now the default): zones keep thermal relief, and a solid per-pad
`(zone_connect 2)` override is forced only on the pads that genuinely cannot
host the 2 thermal spokes KiCad's geometric DRC requires.  Pads that *can*
host 2 spokes keep their thermal relief, which eases hand-soldering / rework.

## How solid-vs-thermal is decided per pad

Two complementary tiers, both grounded in geometry / KiCad's own DRC:

1. **Pad-size heuristic** (`_apply_selective_pad_connection`, the default mode
   of `normalize_zone_pad_connection`): a pad keeps thermal relief only when
   its narrow dimension can seat a spoke plus the thermal gap on both sides
   (`thermal_bridge_width + 2·thermal_gap + margin`).  Smaller pads are forced
   solid.  This handles the obvious small-SMD pads cheaply, pre-fill.
2. **Measurement-driven remediation** (`_remediate_starved_thermal`): reads
   KiCad's DRC back from the fill and forces solid on exactly the residue pad
   size cannot predict —
   - `starved_thermal` (single-spoke pour-edge pads and THT power tabs like
     board 05's Q1/Q5 drains and J3), matched by pad **UUID**;
   - the same-net pad anchoring each `isolated_copper` sliver, resolved
     geometrically (`force_solid_on_isolated_island_pads`).

   Remediation refills via kicad-cli, applies the #3711 foreign-pad clearance
   carve *inside* each pass (so the geometry DRC checks is what ships), and
   snapshots/restores the net table around every refill.

The zone-level `pad_connection` param is preserved as an explicit override:
`"yes"` reproduces the #3728 blanket-solid behavior; `"thru_hole_only"` / `"no"`
are unchanged.  `min_thermal_relief_spokes` is never lowered.

## Changes

- `sexp/builders.py`: `zone_node` `pad_connection` now defaults to `""`
  (zone-level thermal relief) instead of `"yes"`.
- `zones/generator.py`: `ZoneConfig` / `add_zone` `pad_connection` default `""`.
- `zones/fill_clearance.py`: `normalize_zone_pad_connection` defaults to the new
  `"selective"` mode; adds the spoke-viability test, per-pad `zone_connect`
  helpers, DRC-report parsers, and the isolated-island pad resolver.
- `cli/runner.py`: `_remediate_starved_thermal` loop (DRC -> force solid ->
  refill, with the foreign-pad carve as a per-pass settle step and net-table
  restore around refills) wired into `run_fill_zones`.

Verified by regenerating each board's zones in a temp dir
(`uv run --with shapely kct zones fill <pcb> -o /tmp/...`) and running
`kicad-cli pcb drc` (v10.0.1).

| Selective policy is the default (not blanket-solid, not blanket-thermal) | ✅ | `normalize_zone_pad_connection` defaults to `"selective"`; zone token default `""` |
| 0 `starved_thermal` preserved on 03/04/05/softstart | ✅ | DRC: 03 19→0, 04 2→0, 05 14→0, softstart 3→0 |
| Pads that CAN host thermal relief keep it | ✅ | thermal-relief retained: 03=6, 04=31, 05=18, softstart=42 pads (vs solid 42/29/73/80) |
| No new unconnected/isolated_copper/clearance regressions | ✅ | all boards isolated_copper=0, unconnected=0; clearance unchanged from baseline (04=4, 05=21, softstart=16) |
| Boards 01 & 02 stay clean | ✅ | DRC starved=0 isolated=0 unconnected=0 |
| `pad_connection` override preserved | ✅ | `"yes"` still emits blanket `connect_pads yes`; tests cover it |
| `min_thermal_relief_spokes` never lowered | ✅ | no spoke-count setting touched (diff only edits comment text) |
| Test coverage | ✅ | selective decision, UUID/island remediation, report parsing, mode overrides — 330 tests pass |

## Test Plan

- `uv run --with shapely pytest tests/test_zones_fill_clearance.py tests/test_zones_cmd.py tests/test_sexp_builders.py tests/test_zone_generator.py tests/test_route_cmd_zone_fill.py tests/test_gerber_zone_fill.py tests/test_validate_zone_fill.py tests/test_zone_api.py tests/test_runner_net_restore.py` — 330 passed, 1 skipped.
- `ruff check` + `ruff format --check` clean on all changed files.
- Per-board regenerate + kicad-cli DRC for boards 01–05 + softstart (see table).

Note: regenerated board artifacts are intentionally out of scope (per-board PRs handle those); this PR is the shared generation refinement.

Closes #3729

### DIFF
--- a/src/kicad_tools/cli/runner.py
+++ b/src/kicad_tools/cli/runner.py
@@ -497,7 +497,8 @@ def _remediate_starved_thermal(
             _restore_net_declarations(pcb_path, input_net_nodes, input_element_nets)
             if not report_path.exists() or report_path.stat().st_size == 0:
                 return None
-            return json.loads(report_path.read_text())
+            parsed = json.loads(report_path.read_text())
+            return parsed if isinstance(parsed, dict) else None
         finally:
             report_path.unlink(missing_ok=True)
 
--- a/src/kicad_tools/cli/runner.py
+++ b/src/kicad_tools/cli/runner.py
@@ -7,6 +7,7 @@ ERC validation, DRC validation, netlist export, and more.
 import shutil
 import subprocess
 import tempfile
+from collections.abc import Callable
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -353,11 +354,12 @@ def run_fill_zones(
                 stderr="kicad-cli not found. Install KiCad 8 from https://www.kicad.org/download/",
             )
 
-    # Pre-fill normalization (Issue #3727): legacy zones use thermal relief
-    # for *all* pads, which starves pads that cannot form the 2 spokes
-    # KiCad's geometric DRC requires (``starved_thermal``).  Upgrade such
-    # zones to a solid pad connection *before* the fill so the refilled
-    # copper reflects the stronger connection.
+    # Pre-fill normalization (Issues #3727 / #3729): legacy zones use thermal
+    # relief for *all* pads, which starves pads that cannot form the 2 spokes
+    # KiCad's geometric DRC requires (``starved_thermal``).  The selective
+    # policy keeps thermal relief on pads that CAN host 2 spokes and forces a
+    # solid per-pad connection only on the pads that would otherwise starve,
+    # applied *before* the fill so the refilled copper reflects it.
     #
     # When the normalization actually changes a zone, the fill must read the
     # *normalized* board.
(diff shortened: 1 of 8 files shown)


rjwalters:
## Changes Requested - CI Type Check failing (3 new mypy errors)

The implementation is excellent and I independently verified every substantive claim (details below). The only blocker is the **Type Check** CI gate, which reports **3 NEW mypy errors beyond the baseline** in this PR's changed files. Per the Judge protocol CI must pass before approval, and `scripts/ci/check_mypy_baseline.py` explicitly says to fix the finding or add a targeted `# type: ignore[<code>]` rather than baselining it.

### The 3 new errors (from the CI run, job 81597233170)

1. `cli/runner.py:500` -- `_run_drc` does `return json.loads(report_path.read_text())` -> `Returning Any from function declared to return "dict | None"` `[no-any-return]`. Annotate/cast the parsed result, e.g. `data: dict = json.loads(...)` then `return data`, or `return cast(dict, json.loads(...))`.
2. `zones/fill_clearance.py` -- the `required = net_required.get(net_key)` narrowing flows into a `float` comparison: `Incompatible types in assignment (expression has type "float | None", variable has type "float")` `[assignment]`. The runtime is already guarded by `if required is None: continue`; a targeted annotation/ignore on that line resolves it.
3. `zones/fill_clearance.py:594` (and the parallel site at `:853`) -- `connect_pads = zone.find("connect_pads")` is `SExp | None`; mypy cannot narrow across the second `find()` call, so `connect_pads.children` raises `Item "None" of "SExp | None" has no attribute "children"` `[union-attr]`. Add an explicit `if connect_pads is None: continue` after line 592 (cheap and also slightly safer), or reuse the already-fetched node.

All three are type-narrowing false positives with no runtime impact, so the fix is small and low-risk. Re-run `uv run python scripts/ci/check_mypy_baseline.py` locally to confirm 0 new errors before re-requesting review.

---

### Independent verification (all PASS - implementation is sound)

Checked out the PR branch, regenerated zones with `kct zones fill`, ran kicad-cli 10.0.1 DRC.

**starved_thermal (the core fix):**
- Board 03: source pre-fill already 0 (regenerated by #3731), regenerated 0. No regression.
- Board 05: source pre-fill **14 -> 0** after fill. Fix demonstrably works.

**SELECTIVE claim genuinely confirmed (not blanket-solid like #3728):**
- Board 05 regenerated: 73 pads forced solid via `(zone_connect 2)`, **137 pads retain thermal relief** at the zone level (no override). Clear mix.
- Board 03: its committed source already carries an explicit `connect_pads yes` zone token (from #3731), so the selective logic correctly opts out and preserves that override. Exercising the heuristic directly on board 03 with the mode token stripped yields **42 solid / 45 thermal-relief** -- so the heuristic genuinely produces a mix, not blanket solid. (Minor note: the PR table's "03 keeps 6 thermal-relief pads" was measured against a token-free regen; the committed source differs. Documentation nuance, not a defect.)

**Connectivity (no regressions):**
- Board 03: unconnected 14->14, isolated_copper 0, clearance 0->0.
- Board 05: unconnected 84->83 (improved by 1), isolated_copper 0, clearance 21->21 (unchanged).
- (The non-zero unconnected/clearance counts are artifacts of running DRC against a minimal `.kicad_pro`; they are identical pre/post-fill, so the PR introduces nothing.)

**Remediation-loop robustness (`_remediate_starved_thermal`, runner.py:417) -- solid:**
- Bounded at `max_passes=4`; early-returns when the board is clean (`not starved_pads and not isolated_zones`) and when a pass forces 0 new pads (`applied == 0`) -- no infinite loop.
- Graceful degradation: imports wrapped in try/except; whole loop wrapped in try/except with `return` on any exception; `_run_drc` returns `None` (and the loop returns) when kicad-cli writes no report or an empty report. Missing kicad-cli / malformed report both degrade to "leave fill as produced."
- Matching is by **UUID** from `item.uuid`, not by parsing the "PTH pad ... of Q5" string for identity. The `"pad" in desc and "zone" not in desc` heuristic only disambiguates which of a violation's two items is the pad vs the zone; the actual match is the exact UUID. Low mis-match risk. Island resolver is conservative (same-net + on-layer + intersects + <=2mm^2 sliver-area threshold).
- Net table is snapshotted/restored around every refill (`_snapshot_net_declarations` / `_restore_net_declarations`), guarding the known kicad-cli `--save-board` net-stripping quirk.

**min_thermal_relief_spokes never lowered:** confirmed -- the diff only touches comment text, no spoke-count setting changed.

**Tests:** `tests/test_zones_fill_clearance.py tests/test_zones_cmd.py tests/test_sexp_builders.py tests/test_zone_generator.py` -> 200 passed, 1 skipped.

Once the 3 mypy errors are resolved and CI is green, this is an easy approve.

rjwalters:
Judge feedback addressed: the 3 NEW mypy errors blocking the Type Check gate are now fixed (type-narrowing only, no behavior change). Did NOT baseline/silence them.

1. \`cli/runner.py\` (\`_run_drc\`) — \`json.loads(...)\` (Any) into a \`dict | None\` annotation: now narrowed via \`isinstance(parsed, dict)\` before returning.
2. \`zones/fill_clearance.py\` (~616) — \`net_required.get(...)\` (\`float | None\`) collided with the earlier \`float\`-typed \`required\`; renamed the per-pad result to \`pad_required\` so the \`is None\` guard narrows cleanly.
3. \`zones/fill_clearance.py\` (~592) — \`connect_pads = zone.find("connect_pads")\` (\`SExp | None\`): assigned once then guarded \`if connect_pads is None: continue\` (matching the existing pattern at line ~846) so mypy narrows across the \`.children\` access.

Verification (env synced to lockfile mypy 1.19.1 + \`uv sync --extra dev\`, matching CI):
- All 3 flagged signatures (\`no-any-return\` in runner.py, \`assignment\` + \`union-attr\` in fill_clearance.py) no longer appear in the gate output.
- Remaining gate output on this branch matches clean \`main\` exactly (only pre-existing environment-dependent optional-import residuals like \`recovery/strategy.py\`, which this PR does not touch) — i.e. **0 new mypy errors introduced by this PR**.
- \`uv run pytest tests/test_zones_fill_clearance.py tests/test_zones_cmd.py tests/test_sexp_builders.py tests/test_zone_generator.py\` → **200 passed, 1 skipped**.

Re-labeling \`loom:review-requested\` for re-judging.

rjwalters:
## Judge re-review (second pass) — APPROVED

Narrow re-review after the Doctor fix. Last pass I approved on the merits and requested changes solely for 3 mypy errors failing the Type Check gate. Confirming those are resolved with no other changes.

### CI: all green
- **Type Check** — pass (36s) ← the load-bearing gate that was blocking
- **Lint & Format** — pass
- **Test** — pass (13m25s)
- **C++ Build Check** — pass
- **kicad-cli Round-trip Smoke** — pass
- **Routed PCB DRC Check** — pass
- **Diff-Pair Routing Regression** — pass
- **Match-Group Routing Regression** — pass

`mergeStateStatus: CLEAN`.

### Type fixes verified (diff 07435d11..79ed0627) — pure narrowing, no behavior change
- `cli/runner.py` `_run_drc`: `parsed = json.loads(...)` then `return parsed if isinstance(parsed, dict) else None` — narrows `json.loads`'s `Any` to `dict | None`. Callers already treated this as dict-or-None; identical runtime behavior.
- `zones/fill_clearance.py` ~585: hoisted `connect_pads = zone.find("connect_pads")` ahead of the `is None` guard and reused it (was calling `zone.find` twice). Same value, deduplicated, narrowed.
- `zones/fill_clearance.py` ~613: renamed `required` → `pad_required` to avoid the float-binding collision so the `is None` guard narrows. Pure rename, identical control flow.

The doctor commit touches **only** `runner.py` and `fill_clearance.py` at exactly these spots — the substantive implementation in `07435d11` (board 05 starved_thermal 14→0, thermal relief preserved on 137 pads, bounded remediation loop, no connectivity regressions) is intact and unchanged.

Doctor-cycle cap satisfied: CI green, changes clean and limited to the type fixes. Approving.

Issue #3747, opened 2026-06-16
board 00: schematic generator wires LED reverse-biased (cathode→R1, anode→GND); schematic↔PCB netlists disagree


rjwalters:
## Symptom

`boards/00-simple-led/generate_design.py` produces a schematic in which D1 is wired with the wrong polarity:

* **Schematic netlist**: D1 pin 1 (**K**, cathode) ↔ R1.2 (labeled `LED_ANODE`); D1 pin 2 (**A**, anode) ↔ GND rail.
* **PCB netlist** (same script, separate generator): D1 pin 1 → GND; D1 pin 2 → LED_ANODE.

The two halves of the recipe disagree on what each D1 pin is wired to. An assembled board would have the LED reverse-biased per the schematic; the PCB is forward-biased. There is no automated check that the two netlists agree, so this has been latent.

## Evidence (current `main`)

KiCad's `Device:LED` library symbol (confirmed in `boards/00-simple-led/output/simple_led.kicad_sch:427–459`):
* Pin 1 name = `K` at relative x = −3.81
* Pin 2 name = `A` at relative x = +3.81

D1 placed at (100.33, 90.17). Wires in the generated schematic:
* `(100.33, 58.42)` → `(96.52, 90.17)` — R1.2 to **D1.1 (cathode)**
* `(104.14, 90.17)` → `(104.14, 129.54)` — **D1.2 (anode)** to GND rail

PCB generator (`generate_design.py:414`):
`generate_led("D1", D1_POS, "LED_ANODE", "GND")` ⇒ pad 1 (cathode) = GND, pad 2 (anode) = LED_ANODE. Correct polarity, opposite of the schematic.

## Cause

`boards/00-simple-led/generate_design.py:165–172` wires `r1_pin2 → d1_pin1` and `d1_pin2 → GND` based on a misread of which D1 pin is which. The script's comments say "Cathode (K, negative) / Anode (A, positive)" but the wiring still routes to the cathode.

## Fix

Swap the two D1 wire endpoints so the schematic matches both the KiCad LED symbol and the PCB netlist:

```python
# R1 Pin 2 to D1 Pin 2 (Anode) - LED_ANODE net
sch.add_wire(r1_pin2, d1_pin2, warn_on_collision=False)
# D1 Pin 1 (Cathode) to GND rail
sch.add_wire(d1_pin1, (x_d1, rail_gnd_y))
```

[...]

Issue #3742, opened 2026-06-16
Soundness: DRC/kct check can pass while the routed PCB netlist does NOT match the schematic (no independent LVS)


rjwalters:
## The question (raised by an external reviewer)

> How did `simple_led` pass DRC if the PCB netlist doesn't correspond to the schematic?

This is a legitimate and important soundness concern. It exposes a gap: **`kct check` passing does not guarantee the routed PCB is electrically equivalent to the schematic.**

## What we know (concrete case)

Board 00 (`simple-led`) shipped a routed PCB whose connectivity did **not** match the schematic: the autorouter wired `LED_ANODE` to the wrong pads, shorting GND↔LED_ANODE and LED_ANODE↔VCC. Yet `kct check` reported the board clean and the board's test passed. KiCad's own `kicad-cli pcb drc` *did* flag it (as shorts). Root cause: a **pad-rotation-convention mismatch** (#3739) — `kct check`'s connectivity/clearance engine, the autorouter, and zone fill all share the SAME counter-clockwise pad transform, while KiCad/kicad-cli applies the footprint angle clockwise in its Y-down frame. For cardinal 90°/270° footprints the pad positions mirror. Because the checker and the router agree with **each other** (same wrong convention), `kct check`'s connectivity looked self-consistent and "passed" — but it was validating against the router's mistaken pad map, **not** against the true schematic netlist with correct pad geometry. (Board 00 was the only demo board with a 90° footprint; #3738 worked it around by placing footprints at 0°.)

## The deeper gap

The specific rotation bug is tracked in #3739. But this issue is about the **general soundness property** the reviewer is really asking about:

- Does kicad-tools have an **independent layout-vs-schematic (LVS) / netlist-correspondence check** — i.e., extract the *actual* connectivity from the routed copper (using ground-truth pad geometry, e.g. KiCad/kicad-cli's), and assert it equals the schematic netlist?
- Today, `kct check` connectivity appears to derive pad positions from the same transform the router uses, so a coordinate-convention bug is invisible to it. A correct LVS gate would have caught board 00 regardless of the rotation convention, because the *extracted* netlist (GND/VCC/LED_ANODE as actually connected in copper) would differ from the schematic netlist.

## Investigate / propose

- [ ] Confirm exactly why `simple_led`'s pre-#3738 board passed `kct check` while its copper netlist diverged from the schematic (reproduce on the pre-#3738 artifact via git history).
[...]

Issue #3148, opened 2026-05-28
regression: board 00 simple-led routes only 1/2 nets on 2-net trivial board


rjwalters:
## Curator Enhancement (2026-05-28) — Empirical Root Cause

**TL;DR: This is NOT a router regression. The board's hand-rolled `route_pcb()` bypasses the `auto_pour_if_missing()` step, so VCC/GND copper pours are never created. The official `kct route` CLI routes this exact board to 2/2 zones filled, DRC PASS, 0 errors.** Fix = route board 00 through the official pipeline (or call auto-pour in its script), not a change to the A* router.

### What the board actually has

`boards/00-simple-led/generate_design.py` defines **3 nets**, not 2:

```python
NETS = {"": 0, "VCC": 1, "LED_ANODE": 2, "GND": 3}
```

- `VCC` and `GND` are **power/pour nets** (connected via copper zones, not traces).
- `LED_ANODE` is the only routable signal trace (R1.2 -> D1.1).

So "1/2 nets" in the audit = 1 routed signal net (`LED_ANODE`) succeeded, and the DRC connectivity check then flags BOTH power nets as stranded because they have neither traces nor pours.

### Empirical reproduction (commit c9b938b2, local main)

```
$ uv run python boards/00-simple-led/generate_design.py /tmp/board00
2. Routing nets...
  Skipping 1 pour net(s) (use zone fill instead): ['VCC']   <-- router auto-skips VCC
  Net 2: 1 routes, 63 segments, 0 vias                       <-- only LED_ANODE routed
...
   PARTIAL: Routed 1/2 nets
...
   [X] connectivity  Net 'GND' is partially routed: 1 of 2 pads stranded
   [X] connectivity  Net 'VCC' is partially routed: 1 of 2 pads stranded
   DRC FAILED
```

The routed PCB output contains **zero `(zone ...)` blocks** — no copper pour for VCC or GND was ever generated.

### Root cause (file:line)

The board's `route_pcb()` at `boards/00-simple-led/generate_design.py:421-522` calls the router directly:

```python
router, net_map = load_pcb_for_routing(str(input_path), skip_nets=["GND"], rules=rules)
router.route_all()
```

This bypasses the auto-pour step. Specifically:

1. `DEFAULT_NET_CLASS_MAP` (`src/kicad_tools/router/rules.py:1296`) classifies `VCC` and `GND` as the `Power` class with `is_pour_net=True`, `priority=1`.
[...]

PR #3753, opened 2026-06-16
feat(board-00): add schematic <-> routed-PCB LVS guard to recipe

### LINKED ISSUE #3748: board 00: add per-board schematic↔PCB netlist match assertion to the recipe (LVS for board 00)
## Motivation

While #3742 designs the general LVS gate, board 00 is currently shipping a netlist mismatch that nothing in the pipeline catches (see sibling issue: schematic LED reverse-biased). We need a **board-00-specific** netlist-equivalence assertion *now*, so that:

* board 00 cannot regress into a false-pass state again before #3742 lands;
* board 00 becomes the canonical regression fixture for the general LVS gate;
* the demo page can begin surfacing a real \"netlist match: yes/no\" signal.

Per @rwalters: *focus only on this simple board until it is created successfully.*

## Scope (board 00 only)

Extend the `boards/00-simple-led/generate_design.py` pipeline with a final step that:

1. Exports the schematic netlist (kicad-cli `sch export netlist` → KiCadXML or equivalent).
2. Extracts the PCB netlist from the routed `.kicad_pcb` (`(footprint ... (pad N ... (net K \"NAME\")))` walk; existing helpers in `kicad_tools.sexp` are sufficient).
3. Builds the canonical per-pin mapping `{ref}.{pad} → net_name` for **both** sides.
4. Asserts the two mappings are equal modulo trivial renames; on mismatch, raises a `BoardNetlistMismatch` listing each disagreeing `ref.pad` with both nets.
5. Writes the result to `boards/00-simple-led/output/lvs.json` (`{\"clean\": bool, \"mismatches\": [...]}`) for the demo page to surface.

The recipe must exit non-zero on mismatch — board 00 is not \"created successfully\" until LVS is clean.

## Acceptance criteria

* `python boards/00-simple-led/generate_design.py` fails on a netlist mismatch (try it against current main — should fail until the polarity bug is fixed).
* After fixing the polarity bug, recipe succeeds and writes `output/lvs.json` with `clean: true`.
* New unit test asserts `BoardNetlistMismatch` is raised when the LED is wired backwards (regression for the exact bug we just had).
* `lvs.json` schema is documented in `docs/board-json-schema.md`.

## Out of scope (will follow in #3742)

* Generalizing this check to every demo board.
* Wiring it into `kct check` / `kct fleet status` for all boards.
* Cross-checking against kicad-cli's pad geometry (the rotation-convention safeguard).

## References

* #3742 — independent LVS architect proposal (this issue is the narrow board-00 instance that proves the value).
* #3739 — pad-rotation convention bug (related, separate axis).

rjwalters:
## Summary

- Add a new ``kicad_tools.lvs`` submodule with a pure ``compare_netlists`` comparator (``LVSResult`` / ``LVSMismatch`` / ``BoardNetlistMismatch``) that builds a ``{(ref, pad) -> net_name}`` dict from both the schematic and the routed ``.kicad_pcb`` and diffs them.
- Wire ``run_lvs`` into board 00's ``generate_design.py`` between Step 6 (DRC) and Step 7 (manufacturing bundle); it writes ``output/lvs.json`` per the new v1 schema and raises ``BoardNetlistMismatch`` on disagreement so the recipe **hard-fails** instead of silently shipping a mis-wired board (regression guard for #3747).
- Document the v1 ``lvs.json`` schema in ``docs/board-json-schema.md`` and add three hermetic tests (clean board 00, deliberately-swapped D1.1/D1.2 mismatch case, ``_ref_of`` reference-form polymorphism).

Schematic-side resolution uses ``Schematic.get_net_for_pin`` (label-bound names: VCC, GND, LED_ANODE, ...) rather than ``build_netlist_from_schematic``, which would collapse every PWR_FLAG-touched pin into one "PWR_FLAG" net and make LVS spuriously fail on every board.

Scope is board 00 only; the fleet-wide rollout (rename heuristics, other boards, ``kct check`` integration) is tracked by #3742. The recipe's existing local ``_ref_of`` helper is **not** removed in this PR -- the new module exports it alongside so #3742 can dedupe it cleanly.

Closes #3748.

## Test plan

- [x] ``uv run pytest tests/test_board_00_lvs.py -v`` — 6 passed (clean board 00 / mismatched D1 / exception message / both ``_ref_of`` forms / missing reference)
- [x] ``uv run pytest tests/test_board_00_simple_led.py -v`` — 6 passed (existing board 00 polarity + #3148 regression tests still green)
- [x] ``uv run python boards/00-simple-led/generate_design.py`` — exits 0 with ``LVS: PASS`` and writes ``output/lvs.json`` containing ``{"clean": true, "mismatches": []}``
- [x] ``uv run ruff check`` and ``uv run ruff format`` clean on all changed files
- [ ] Manual sanity: temporarily swap one of the schematic's labels and confirm the recipe exits non-zero with the new ``BoardNetlistMismatch`` message naming the offending ``(ref, pad)`` pairs

### DIFF
--- a/boards/00-simple-led/generate_design.py
+++ b/boards/00-simple-led/generate_design.py
@@ -17,6 +17,7 @@ Usage:
     python generate_design.py [output_dir]
 """
 
+import json
 import subprocess
 import sys
 import uuid
@@ -24,6 +25,7 @@ from pathlib import Path
 
 from kicad_tools.core.project_file import create_minimal_project, save_project
 from kicad_tools.dev import warn_if_stale
+from kicad_tools.lvs import BoardNetlistMismatch, compare_netlists
 from kicad_tools.schematic.models.schematic import Schematic
 
 # Warn if running source scripts with stale pipx install
@@ -868,6 +870,61 @@ def export_manufacturing_bundle(routed_path: Path, output_dir: Path) -> bool:
     return False
 
 
+def run_lvs(sch_path: Path, routed_pcb_path: Path, output_dir: Path) -> bool:
+    """Step 6.5: assert schematic <-> routed-PCB netlist equivalence (#3748).
+
+    Compares each ``(ref, pad)`` pair's net name between the schematic
+    (the design intent) and the routed PCB (what manufacturing receives)
+    and writes the result to ``output/lvs.json`` for downstream tools
+    (``kct fleet status``, the demo site, etc.) following the v1 schema
+    documented in ``docs/board-json-schema.md``.
+
+    Returns ``True`` iff the comparison was clean.  Raises
+    :class:`BoardNetlistMismatch` when it is dirty so the caller's exit
+    gate trips -- a silent dirty board would defeat the purpose of
+    wiring LVS into the recipe.
+    """
+    print("\n" + "=" * 60)
+    print("Running LVS (schematic <-> PCB netlist match)...")
+    print("=" * 60)
+
+    result = compare_netlists(sch_path, routed_pcb_path)
+
+    output_dir.mkdir(parents=True, exist_ok=True)
+    lvs_path = output_dir / "lvs.json"
+    lvs_path.write_text(
+        json.dumps(
+            {
+                "$schema": "https://kicad-tools.org/schemas/lvs/v1.json",
+                "clean": result.clean,
+                "mismatches": [
+                    {
+                        "ref": m.ref,
+                        "pad": m.pad,
+                        "schematic_net": m.schematic_net,
+                        "pcb_net": m.pcb_net,
+                    }
+                    for m in result.mismatches
+                ],
+            },
+            indent=2,
+        )
+        + "\n"
+    )
+
+    if result.clean:
+        print(f"\n   PASS: {lvs_path.name} clean (0 mismatches)")
+        return True
+
+    # Print up to the first 5 mismatches before raising, so the recipe
+    # log surfaces what went wrong without forcing the user to crack
+    # open lvs.json.
+    print(f"\n   FAIL: {len(result.mismatches)} schematic/PCB mismatch(es):")
+    for m in result.mismatches[:5]:
+        print(f"      - {m.ref}.{m.pad}: schematic={m.schematic_net!r} pcb={m.pcb_net!r}")
+    raise BoardNetlistMismatch(result)
+
+
 def main() -> int:
     """Main entry point."""
     if len(sys.argv) > 1:
@@ -895,6 +952,13 @@ def main() -> int:
         # Step 6: Run DRC
         drc_success = run_drc(routed_path)
 
+        # Step 6.5: LVS (#3748) -- assert the routed PCB's nets match the
+        # schematic's design intent.  This is the post-route guard that
+        # caught #3747 (D1 polarity flipped on the schematic side); it
+        # raises ``BoardNetlistMismatch`` on disagreement so the recipe
+        # exits non-zero instead of silently shipping a bad board.
+        lvs_success = run_lvs(sch_path, routed_path, output_dir)
+
         # Step 7: Export manufacturing bundle (#3147).  Run unconditionally
         # so the manifest mtime stays newer than the routed PCB even when
         # routing is incomplete (board 00 routing is tracked by #3148);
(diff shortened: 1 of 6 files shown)


rjwalters:
## Loom Judge Review: APPROVE

Solid, surgical implementation of the board-00 LVS gate (#3748). Code is well-scoped, well-tested, and the rationale matches the codebase reality I verified.

### What I verified

**Correctness — would catch the #3747 D1.1/D1.2 swap class**
- After the polarity fix, schematic now reads `D1.1 (K) -> 'GND'`, `D1.2 (A) -> 'LED_ANODE'`; the routed PCB matches, so `compare_netlists` reports `clean=True` against the committed artifacts.
- The dirty-path test `_swap_d1_pad_nets` does exactly the inverse mutation and `compare_netlists` produces two mismatches (`D1.1` and `D1.2` with `schematic_net` / `pcb_net` crossed), exercising the same regression surface that escaped to #3747.
- `BoardNetlistMismatch._format_message` emits `D1.1` and `D1.2` substrings in the exception message — recipe operators see the offending pad pair without opening `lvs.json`.

**PWR_FLAG rationale is correct**
- Confirmed in `src/kicad_tools/schematic/models/netlist_mixin.py:114-122`: `_build_connectivity_graph` injects `PWR_FLAG` as a candidate net name (via `pwr.lib_id.split(":")[1]`) for any node touching a `power:PWR_FLAG` symbol. In `extract_netlist`, `net_names[0]` may pick `"PWR_FLAG"` over `"+5V"` / `"GND"` depending on insertion order.
- Board 00 places `PWR_FLAG` on both VCC and GND rails (`generate_design.py:135-137`), so `build_netlist_from_schematic` is a real footgun here. The choice of `Schematic.get_net_for_pin` (which returns `net_names[0]` from a per-pin connected-component scan and is happy to land on `'VCC'` / `'GND'` / `'LED_ANODE'`) is the right call.

**Robust PCB-side walk**
- `_pcb_pin_to_net` handles missing `(net ...)` children (`net is None -> None`) and footprints without a resolvable reference (`_ref_of -> None -> continue`). The `_norm_pcb` step collapses both `None` and `""` (net 0) to `None`, so unconnected-on-both-sides compares equal — the spec's intended semantics.
- `_ref_of` handles both `(fp_text reference "R1" ...)` (generator output) and `(property "Reference" "R1" ...)` (post `kicad-cli` round-trip), and the polymorphism test locks down both. Returning `None` on a missing reference (vs raising) keeps mechanical / unresolved footprints from poisoning the comparison.

**Scope discipline**
- `git diff origin/main..pr3753 --stat` is exactly the 6 files in the description; no drift into other demo boards, no edits to PCB / schematic generators, no removal of the existing `_ref_of` in `generate_design.py:479` (deduplication left for #3742 as promised).

**Committed lvs.json is clean** — `{"clean": true, "mismatches": []}` matches a fresh comparator run against the committed artifacts.

**Schema doc** — `docs/board-json-schema.md` documents the v1 contract with both clean and dirty examples, the always-present `mismatches` convention, and versioning policy aligned with `board.json`.

### Local verification

- `uv run pytest tests/test_board_00_lvs.py -v` — 6 passed in 66s
- `uv run ruff check` and `uv run ruff format --check` — clean on all changed files
- `uv run mypy src/kicad_tools/lvs/` — Success, no issues
- `compare_netlists` on committed board 00 artifacts — `clean=True`, `mismatches=()`

### CI

All 8 checks green: C++ Build, Lint & Format, Type Check, Test (13m58s), DRC, kicad-cli Round-trip Smoke, Diff-Pair Routing Regression, Match-Group Routing Regression. `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

- `_pcb_pin_to_net` treats a missing pad number (`pad.get_string(0) is None`) as "skip silently". On a malformed PCB that could silently drop a real mismatch; since this can't happen with kicad-tools-produced PCBs and a hard failure would itself be a footgun on round-tripped boards, the current silent-skip is fine. Worth a `logger.debug` call in #3742's fleet-wide pass if you want a forensic trail.
- The exported `_ref_of` is marked private (underscore) but listed in `__all__`. That's intentional per the docstring (board recipes import it directly while waiting for #3742 dedup) — flagging it only so a future cleanup PR knows the public-ness is provisional.

Approving and labeling `loom:pr`.

PR #3754, opened 2026-06-16
feat(board-json,site): surface LVS status in gallery chip (closes #3749)

### LINKED ISSUE #3749: demo site: board 00 over-reported as "Ready" while schematic↔PCB netlists mismatch
## Symptom

kicad-tools.org renders board 00 with a green **\"Ready\"** badge. The underlying `board.json` reports `\"drc_violations\": 0, \"status\": \"ok\"`. But the schematic and PCB netlists for board 00 disagree on D1's pin-to-net mapping (LED reversed; see sibling polarity issue). An assembled board would not light up.

This is a **false positive of public-facing overconfidence**: visitors evaluating the project see a clean status when the board is electrically broken.

## Evidence

* `boards/00-simple-led/output/board.json`: `\"drc_violations\": 0, \"nets_routed_pct\": 100, \"status\": \"ok\"`.
* `site/src/data/boardStatus.ts:29`: `displayStatus()` returns `\"ready\"` when `drc_violations === 0 && status === \"ok\"`.
* `site/src/components/BoardCard.astro`: renders the \"Ready\" badge accordingly.
* `kct check` on the routed PCB: **DRC PASSED**.
* The schematic still has D1 wired backwards — confirmed by reading `boards/00-simple-led/output/simple_led.kicad_sch` directly.

## Required change

\"Ready\" must mean *DRC clean AND ERC clean AND netlist match (LVS) clean AND manufacturing bundle present*. DRC-only is not enough.

Concretely:

1. Add `lvs_clean: bool | null` (and `lvs_mismatches: int`) to the board JSON schema (`docs/board-json-schema.md`, `site/src/data/types.ts`).
2. `kct board-metrics` populates `lvs_clean` from `boards/<slug>/output/lvs.json` (produced per sibling LVS issue). `null` if no LVS artifact (so we never silently show a stale \"true\").
3. `displayStatus()` in `boardStatus.ts`:
   * `drc_violations > 0` → `\"drc\"` (unchanged)
   * `lvs_clean === false` → new `\"lvs\"` variant (warn-styled, never \"Ready\")
   * `lvs_clean === null` → new `\"unverified\"` variant (neutral, never \"Ready\")
   * `lvs_clean === true && status === \"ok\"` → `\"ready\"` (only path to Ready)
4. Add `displayStatusLabel` cases (`\"LVS mismatch\"`, `\"LVS not run\"`).
5. Update tests in `site/src/data/boardStatus.test.ts` to cover the new variants.

Until #3748 (per-board LVS for board 00) lands, **board 00 should display \"LVS not run\"**, not \"Ready\".

## Acceptance criteria

* Visit the demo page locally with the current artifacts: board 00 shows a non-Ready chip (either \"LVS not run\" or \"LVS mismatch\" depending on which sibling lands first).
* After fixing the polarity bug (#3747) **and** running the LVS check (#3748), board 00 shows green \"Ready\".
* Manual test: introduce a deliberate netlist mismatch → site renders \"LVS mismatch\".

## References

* Companion to #3747 (polarity bug) and #3748 (per-board LVS).
* Long-term: #3742 (general LVS gate).

rjwalters:
## Summary

Fixes #3749 — board 00 reads "Ready" in the demo gallery today even though its schematic-vs-PCB netlists are not actually compared by the loader. This PR routes the existing `output/lvs.json` (#3748) through `board.json` and adds two new gallery chip variants so:

- A board with an explicit LVS mismatch (`lvs_clean === false`) reads **"LVS mismatch"** (orange).
- A board that has not run LVS at all (`lvs_clean` absent) reads **"LVS not run"** (gray).
- Only `lvs_clean === true` + `status === "ok"` + zero DRC can read **"Ready"**.

## Changes

**Producer** — `src/kicad_tools/cli/board_metrics_cmd.py`
- New `_parse_lvs()` reads `output/lvs.json` and emits `lvs_clean` / `lvs_mismatches`. Both fields are omitted when `lvs.json` is absent (omit-when-absent contract).
- Status downgrade now also triggers on explicit `lvs_clean == False`. A missing `lvs.json` does NOT downgrade — that is purely a site-layer presentation concern.

**Site** — `site/src/data/types.ts`, `site/src/data/boardStatus.ts`, `site/src/components/BoardCard.astro`
- Extended `Board` interface with `lvs_clean?: boolean` / `lvs_mismatches?: number`.
- Two new `DisplayStatus` variants: `lvs` and `unverified`, with `displayStatusLabel` returning `"LVS: N mismatch(es)"` / `"LVS mismatch"` / `"LVS not run"`.
- Display-priority order (load-bearing): **DRC > LVS > unverified > status**. DRC always wins so a board with both a DRC violation and an LVS mismatch surfaces the DRC chip first (DRC is the fabrication blocker).
- Two new chip CSS classes: `.chip-lvs` (warm orange — distinct from amber `chip-partial` and red `chip-drc`) and `.chip-unverified` (neutral gray).

**Docs** — `docs/board-json-schema.md`
- Documents `lvs_clean?: boolean` and `lvs_mismatches?: number` plus the asymmetric status-gate semantics.

**Tests**
- `tests/test_board_metrics.py`: 4 new cases for present-clean, present-dirty (asserts status downgrade), absent (asserts both fields omitted AND status stays `ok`), and malformed-JSON resilience.
- `site/src/data/boardStatus.test.ts`: 5 new cases including the explicit regression case (no `lvs_clean` field → never "Ready") and DRC > LVS priority.

## Test plan

- [x] `uv run pytest tests/test_board_metrics.py -v` — 24 passed including 4 new LVS cases
- [x] `cd site && npm test` — 28 passed including 5 new LVS cases
- [x] `cd site && npm run check` — 0 errors, 0 warnings (pre-existing hints only)
- [x] `uv run ruff format` + `uv run ruff check` — clean
- [x] Regenerated `boards/00-simple-led/output/board.json` shows `"lvs_clean": true, "lvs_mismatches": 0`

### DIFF
--- a/docs/board-json-schema.md
+++ b/docs/board-json-schema.md
@@ -24,6 +24,8 @@ directory — `kct board-metrics` never recomputes anything from KiCad.
 | `renders`               | `output/renders/*.png` (from `kct render`, #3675) | only existing files |
 | `manufacturing_package` | `output/manufacturing/kicad_project.zip` | omitted if absent |
 | `manifest_generated_at` | `manifest.json` → `generated_at`      | ISO-8601 string |
+| `lvs_clean`             | `output/lvs.json` → `clean`           | omitted when `lvs.json` is absent (#3748, #3749) |
+| `lvs_mismatches`        | `output/lvs.json` → `len(mismatches)` | omitted when `lvs.json` is absent (#3748, #3749) |
 
 ## Example
 
@@ -49,6 +51,8 @@ directory — `kct board-metrics` never recomputes anything from KiCad.
   },
   "manufacturing_package": "manufacturing/kicad_project.zip",
   "manifest_generated_at": "2026-06-12T05:03:41.535120+00:00",
+  "lvs_clean": true,
+  "lvs_mismatches": 0,
   "status": "ok"
 }
 ```
@@ -73,6 +77,8 @@ directory — `kct board-metrics` never recomputes anything from KiCad.
 | `renders`               | object  | no       | Map of render id → path relative to `board.json` |
 | `manufacturing_package` | string  | no       | Path to downloadable `kicad_project.zip` |
 | `manifest_generated_at` | string  | no       | Manifest build timestamp (ISO-8601) |
+| `lvs_clean`             | boolean | no       | `true` iff `output/lvs.json` reports `clean: true`. Omitted when `lvs.json` is absent (board has not run LVS yet). |
+| `lvs_mismatches`        | integer | no       | Count of mismatches recorded in `output/lvs.json`. Omitted when `lvs.json` is absent. |
 
 ### Paths are relative to `board.json`
 
@@ -84,9 +90,14 @@ resolves to `boards/<id>/output/renders/pcb-front.png`.
 
 | Value          | Meaning |
 |----------------|---------|
-| `ok`           | `output/manufacturing/` exists and `report.md` parsed successfully |
-| `partial`      | `output/manufacturing/` exists but `report.md` is absent/unparseable; only identity and recoverable fields are present |
+| `ok`           | `output/manufacturing/` exists, `report.md` parsed successfully, `drc_violations == 0`, and (when `lvs.json` is present) `lvs_clean == true` |
+| `partial`      | `output/manufacturing/` exists but `report.md` is absent/unparseable (only identity and recoverable fields are present), OR `drc_violations > 0`, OR an explicit `lvs_clean == false` |
 | `no_artifacts` | the board has no `output/manufacturing/` directory at all |
+
+Note: a *missing* `lvs.json` does **not** downgrade `status` at the producer
+layer — boards without an LVS step yet keep their existing status. The site
+gallery enforces a stricter gate ("Ready" requires `lvs_clean === true`) by
+rendering a neutral "LVS not run" chip when the field is absent (#3749).
 
 ## Optional fields are omitted, never `null`
 
--- a/site/src/components/BoardCard.astro
+++ b/site/src/components/BoardCard.astro
@@ -207,6 +207,22 @@ const statusLabel = displayStatusLabel(board);
     color: #1a1300;
   }
 
+  /* LVS-mismatch chip (#3749): the schematic and PCB disagree, so the board
+     will not function as designed even though it may be DRC-clean. We use a
+     warmer orange to distinguish it from amber "Partial" and red "DRC". */
+  .chip-lvs {
+    background: rgba(214, 130, 50, 0.92);
+    color: #1a0d00;
+  }
+
+  /* LVS-not-run chip (#3749): neutral gray for "unknown" — visually similar
+     to no_artifacts so the message is "we have not verified" rather than "we
+     verified and it is a problem". Never reads as Ready. */
+  .chip-unverified {
+    background: rgba(120, 134, 148, 0.85);
+    color: #0c1116;
+  }
+
   .chip-no_artifacts {
     background: rgba(120, 134, 148, 0.85);
     color: #0c1116;
--- a/site/src/data/boardStatus.test.ts
+++ b/site/src/data/boardStatus.test.ts
@@ -1,15 +1,20 @@
 /**
- * Unit tests for the display-status helper (issue #3717).
+ * Unit tests for the display-status helper (issues #3717, #3749).
  *
- * The displayed gallery badge must read "Ready" ONLY when a board has zero DRC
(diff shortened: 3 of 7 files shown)


rjwalters:
APPROVED. This is a careful, well-architected fix for the false-positive "Ready" badge on board 00.

## What I verified

**Producer side (`board_metrics_cmd.py`)**
- `_parse_lvs()` returns `{}` (not `None` values) when `lvs.json` is absent, so `metrics.update({})` is a no-op and the LVS keys are correctly OMITTED from `board.json` rather than emitted as `null`. Matches the documented omit-when-absent convention.
- Schema is consistent with the #3748 producer in `boards/00-simple-led/generate_design.py`: reads `clean` and `len(mismatches)` from `output/lvs.json` exactly as written.
- Status downgrade gate is asymmetric and correct: `metrics.get("lvs_clean") is False` — uses `is False` (not falsy) so absent/None doesn't downgrade. A missing `lvs.json` keeps `status: ok`; an explicit `clean: false` downgrades to `partial`.
- Malformed-JSON path catches `OSError, json.JSONDecodeError`, logs a warning, and returns `{}` so the board doesn't crash the whole metrics run.

**Site side (`boardStatus.ts`)**
- Priority order is unambiguously DRC > LVS > unverified > status, as required. Verified by the new test `"DRC takes priority over LVS when both fail"` — a board with `drc_violations: 5, lvs_clean: false` correctly surfaces as `drc`, not `lvs`. DRC stays the fabrication blocker.
- The regression case (the actual bug from #3749) is asserted explicitly: `lvs_clean: undefined` + `status: "ok"` + `drc: 0` -> `unverified` / "LVS not run", never "Ready". This is the exact failure mode that was rendering board 00 as green.
- New `DisplayStatus` variants `"lvs"` and `"unverified"` are wired through to `BoardCard.astro` via `chip-${statusVariant}` template, with matching CSS classes (warm orange `chip-lvs`, neutral gray `chip-unverified`).

**Tests**
- Default fixture now sets `lvs_clean: true`, so all 5 pre-existing tests still pass (verified). Minimal churn, intentional choice — the comment on the fixture even calls out that opt-out happens via explicit `lvs_clean: undefined`.
- 5 new TS tests, including singular/plural label, mismatch-count rendering, and DRC > LVS priority. 4 new Python tests covering present-clean, present-dirty (status downgrade), absent (fields omitted, status preserved), and malformed JSON.

**End-to-end verification (most important)**
I checked out the PR branch and actually ran the producer:
- `uv run kct board-metrics boards/00-simple-led` -> regenerated `board.json` contains `"lvs_clean": true, "lvs_mismatches": 0, "status": "ok"`. Exactly as the PR claims.
- `uv run kct board-metrics boards/01-voltage-divider` (no `lvs.json` present) -> `board.json` does NOT contain `lvs_clean` or `lvs_mismatches`. Status stays `ok`. Site layer will render this as the new "LVS not run" / `unverified` chip.

No scope creep — only the producer logic touches LVS; downgrade behavior on the producer side only fires for explicit `False`.

## Test Execution

Test plan from PR description:
1. `uv run pytest tests/test_board_metrics.py -v` -> executed: 24 passed (incl. 4 new LVS cases)
2. `cd site && npm test` -> executed: 28 passed (incl. 5 new LVS cases) in 1.64s
3. `cd site && npm run check` -> executed: 0 errors, 0 warnings, 98 hints (all pre-existing)
4. `uv run ruff check` + `uv run ruff format --check` -> executed: clean, 2 files already formatted
5. Regenerated `boards/00-simple-led/output/board.json` shows `lvs_clean: true` -> executed: confirmed

## CI status

All 8 GitHub Actions checks pass (C++ Build, Diff-Pair Routing Regression, Lint & Format, Match-Group Routing Regression, Routed PCB DRC Check, Test, Type Check, kicad-cli Round-trip Smoke). `mergeStateStatus: CLEAN`.

## Minor observations (non-blocking)

- The `bool(data.get("clean", False))` cast is defensive against non-boolean values in `lvs.json` (e.g. `"clean": "yes"`). Reasonable hardening — `lvs_clean` in the schema is `boolean`, and a stricter `isinstance` check would just be cosmetic.
- The `chip-unverified` class is visually identical to `chip-no_artifacts` (both `rgba(120, 134, 148, 0.85)`). That's intentional per the inline comment ("we have not verified" rather than "we verified and it is a problem") and is the right framing.

Excellent work. This is the third in a 5-PR push to fix board-00 demo honesty and it slots in cleanly with #3747 (polarity) and #3748 (per-board LVS).

Issue #3739, opened 2026-06-16
Investigate: pad-rotation convention disagrees with kicad-cli for cardinal (90/270) footprints


rjwalters:
## Summary (needs careful investigation — do NOT auto-fix without a human in the loop)

While fixing board 00's real net shorts (#3737, PR #3738), the builder traced the root cause to a **pad-rotation-convention mismatch** between kicad-tools and KiCad/kicad-cli:

- kicad-tools' `_transform_pad_position` (used by `kct check` clearance/connectivity, the grid autorouter, and zone fill) models the footprint's on-disk angle as a **counter-clockwise** (standard-math) rotation.
- KiCad/kicad-cli applies the footprint angle **clockwise** in its **Y-down** frame.
- For **cardinal 90°/270°** rotations these conventions place pin 1 vs pin 2 at **mirror-image positions**. (For 0°/180° they agree, which is why no other demo board hit this — boards 01–07 + softstart all use 0° footprints.)

**Concrete evidence:** board 00 placed R1/D1 at 90°. The autorouter (sharing kicad-tools' CCW convention) wired LED_ANODE to R1's *VCC* pad and D1's *GND* pad — a genuine copper short that **kicad-cli flagged** (GND↔LED_ANODE, LED_ANODE↔VCC + matching solder-mask bridges) but **`kct check` passed** (same mirrored convention as the router). The two engines fundamentally disagreed; no routing could satisfy both. PR #3738 worked around it board-locally by placing board 00's footprints at rotation 0 (where both conventions agree) — it did NOT touch the core transform.

## Why this is contentious

This appears to contradict prior deliberate decisions recorded in PRs **#738** (Jan 2026, "fixed the rotation to CCW-positive, standard math, no negation") and the cleanup PRs **#2778/#2789**. Either:
- (a) those PRs set the wrong convention vs KiCad's actual Y-down clockwise application, or
- (b) there's a compensating Y-axis flip elsewhere that's correct for 0/180 but not 90/270, or
- (c) the on-disk angle semantics differ from what `_transform_pad_position` assumes.

Resolving this touches **core validation + router + zone-fill** code (`src/kicad_tools/validate/rules/clearance.py:~125`, `src/kicad_tools/validate/connectivity.py:~566`, plus the autorouter and zone-fill pad transforms) — broad blast radius. It must be done carefully with a human reviewer, validated against kicad-cli's own pad positions, and regression-tested across all boards.

## Suggested investigation steps
[...]

PR #3757, opened 2026-06-17
feat(lvs): add independent copper-extracted netlist gate

### LINKED ISSUE #3742: Soundness: DRC/kct check can pass while the routed PCB netlist does NOT match the schematic (no independent LVS)

rjwalters:
## Summary

`kct check` could report a board clean while the routed PCB netlist did
not match the schematic: both the label-based LVS and the geometric
connectivity validator trusted pad `(net ...)` labels and/or shared the
router's pad-rotation transform, so a coordinate-convention bug
self-cancelled (the board-00 failure the external reviewer raised). This
adds the "third leg" of soundness: an **independent copper-extracted
netlist gate** that ignores pad labels and diffs the *physical* copper
partition against the schematic.

This is the curated bounded first slice of #3742; the deferred fleet-wide
rollout, power-net normalization, and per-board kicad-cli cross-check on
every run are explicitly out of scope.

## Changes

- `validate/connectivity.py`: new `ConnectivityValidator.extract_pad_partition()`
  floods the routed copper graph from **track segments + vias with zero
  reference to pad `(net ...)` labels** and returns galvanically-connected
  pad groups. This is the load-bearing property: it catches a router that
  wires copper to the wrong pads while labeling them correctly. Zone pours
  use a documented narrower model (grouped by declared net, against the
  zone boundary) so a pour can never fabricate a false short across nets;
  robust label-free pour/thermal-spoke extraction is a tracked #3742
  follow-up. Fully independent for autorouter copper, which is where the
  #3739 rotation bug actually manifested.
- `lvs/copper_lvs.py` (NEW): `compare_partitions()` (pure, IO-free) diffs
  the copper partition against the schematic partition, classifying
  **shorts** (different schematic nets fused in one copper island) and
  **opens** (one net split across islands). `compare_copper_netlist()`
  is the file-level entry point, reusing `board_lvs._schematic_pin_to_net`
  for the schematic side. New `CopperLVSResult` / `CopperLVSMismatch`.
- `lvs/__init__.py`: re-export the new comparator + result types.
- `cli/check_cmd.py`: `_lvs_subcheck` now runs the copper-extracted
  comparison alongside the existing label-based `compare_netlists`; either
  dirtying fails LVS (copper surfaced first as soundness-critical). The
  label-based path is unchanged.
- `tests/test_copper_lvs.py` (NEW): partition-diff unit tests, the 90°
  regression fixture, the decoupling golden, and a board-00 end-to-end.

| Independent copper-extracted partition gate that ignores pad net labels | ✅ | `extract_pad_partition()` traces segments/vias without reading any `(net ...)` label; `test_extract_pad_partition_ignores_pad_labels` proves two differently-labeled pads joined by copper land in one component |
| Reports open/short by diffing physical copper islands vs schematic partition | ✅ | `compare_partitions` emits short/open records; `test_short_detected_*`, `test_open_detected_*` |
| Decoupling guard: 90°/270° pad geometry vs independent reference | ✅ | `test_transform_pad_position_matches_golden` asserts 0/90/180/270° against the pcbnew-verified golden; kicad-cli live cross-check documented as follow-up |
| 90°/270° regression fixture FAILS on adversarial copper, PASSES on corrected copper | ✅ | `test_90deg_adversarial_copper_fails` (mirrored copper → OUT1/OUT2 short + SIG open) and `test_90deg_corrected_copper_passes` |
| Composes with existing label-based LVS (kept alongside) | ✅ | `_lvs_subcheck` runs both; `test_meta_check_*` in `test_cli_check.py` still pass; `compare_netlists` untouched |
| Board-00 still passes | ✅ | `test_compare_copper_netlist_on_board00_artifacts` + full `test_board_00_lvs.py` green |

## Test Plan

- `pytest tests/test_copper_lvs.py` — 19 passed (partition diff, 90° regression, golden, board-00 e2e).
- `pytest tests/test_cli_check.py` — 43 passed (meta-check + LVS-mismatch rows).
- `pytest tests/test_lvs.py tests/test_board_00_lvs.py tests/test_connectivity.py tests/test_output_connectivity.py tests/test_pcb_nets_connectivity.py tests/test_routing_connectivity.py tests/test_connectivity_rule.py` — all passed (label-based path + connectivity unchanged).
- `ruff check` / `ruff format --check` clean on all changed files.
- `mypy` on the new `copper_lvs.py` and `extract_pad_partition` adds no new errors (the 12 reported in `connectivity.py` pre-exist on `origin/main` in the untouched `_build_connectivity_graph`).

Closes #3742

### DIFF
--- a/src/kicad_tools/cli/check_cmd.py
+++ b/src/kicad_tools/cli/check_cmd.py
@@ -257,7 +257,17 @@ def _erc_subcheck(sch_path: Path | None, strict: bool) -> SubCheckResult:
 
 
 def _lvs_subcheck(sch_path: Path | None, pcb_path: Path) -> SubCheckResult:
-    """Run live LVS via :func:`compare_netlists` (issue #3750).
+    """Run live LVS (issue #3750, extended for independent copper LVS #3742).
+
+    Runs *two* complementary comparisons and fails if **either** dirties:
+
+    * **label-based** (:func:`board_lvs.compare_netlists`) — trusts each
+      pad's declared ``(net ...)`` label; catches generator/router
+      bookkeeping drift.
+    * **copper-extracted** (:func:`copper_lvs.compare_copper_netlist`) —
+      ignores pad labels entirely and diffs the *physical* copper partition
+      against the schematic; catches shorts/opens a mislabeled router would
+      hide from the label-based path (the board-00 soundness gap, #3742).
 
     Always recomputes -- never reads ``output/lvs.json`` -- so a fresh
     PCB edit that breaks LVS is surfaced immediately.  Returns
@@ -279,10 +289,33 @@ def _lvs_subcheck(sch_path: Path | None, pcb_path: Path) -> SubCheckResult:
             detail=f"LVS comparator raised {type(e).__name__}: {e}",
         )
 
+    try:
+        from kicad_tools.lvs.copper_lvs import compare_copper_netlist
+
+        copper = compare_copper_netlist(sch_path, pcb_path)
+    except Exception as e:
+        return SubCheckResult(
+            status="FAILED",
+            detail=f"copper LVS comparator raised {type(e).__name__}: {e}",
+        )
+
+    if not copper.clean:
+        # The copper-extracted gate is the soundness-critical one: surface
+        # it first.  Show up to the first 3 records in a stable order.
+        records = sorted(copper.mismatches, key=lambda m: (m.kind, m.pad_a, m.pad_b))
+        preview = ", ".join(
+            f"{m.kind} {m.pad_a}({m.net_a})/{m.pad_b}({m.net_b})" for m in records[:3]
+        )
+        suffix = "" if len(records) <= 3 else f" (+{len(records) - 3} more)"
+        return SubCheckResult(
+            status="FAILED",
+            detail=f"copper: {len(records)} mismatch(es): {preview}{suffix}",
+        )
+
     if result.clean:
         return SubCheckResult(
             status="PASSED",
-            detail=f"{len(result.mismatches)} mismatch(es)",
+            detail="label + copper: 0 mismatch(es)",
         )
 
     # Show up to the first 3 mismatches in stable (ref, pad) order so
@@ -294,7 +327,7 @@ def _lvs_subcheck(sch_path: Path | None, pcb_path: Path) -> SubCheckResult:
     suffix = "" if len(mismatches) <= 3 else f" (+{len(mismatches) - 3} more)"
     return SubCheckResult(
         status="FAILED",
-        detail=f"{len(mismatches)} mismatch(es): {preview}{suffix}",
+        detail=f"label: {len(mismatches)} mismatch(es): {preview}{suffix}",
     )
 
 
--- a/src/kicad_tools/lvs/__init__.py
+++ b/src/kicad_tools/lvs/__init__.py
@@ -6,9 +6,23 @@ Compares the schematic netlist (the design intent) against the routed PCB
 Currently single-board scope (board 00); see issue #3742 for the
 generalized fleet-wide LVS rollout.
 
+Two complementary comparators live here:
+
+* **label-based** (:func:`compare_netlists`) trusts each pad's declared
+  ``(net ...)`` label.
+* **copper-extracted** (:func:`compare_copper_netlist`, issue #3742)
+  ignores pad labels and diffs the *physical* copper partition against the
+  schematic, catching shorts/opens a mislabeled router would hide.
+
 Public API:
 
 * :func:`compare_netlists` — pure comparator returning :class:`LVSResult`.
+* :func:`compare_copper_netlist` — independent copper-extracted comparator
+                              returning :class:`CopperLVSResult`.
+* :func:`compare_partitions` — pure partition diff (IO-free core of the
+                              copper-extracted comparator).
+* :class:`CopperLVSResult` / :class:`CopperLVSMismatch` — copper-LVS
+                              result + per-short/open record.
 * :class:`LVSResult`        — frozen dataclass with ``clean`` flag and
                               tuple of :class:`LVSMismatch` entries.
 * :class:`LVSMismatch`      — frozen dataclass naming the offending
@@ -31,11 +45,21 @@ from kicad_tools.lvs.board_lvs import (
     _ref_of,
     compare_netlists,
 )
+from kicad_tools.lvs.copper_lvs import (
+    CopperLVSMismatch,
+    CopperLVSResult,
+    compare_copper_netlist,
+    compare_partitions,
+)
 
 __all__ = [
     "BoardNetlistMismatch",
+    "CopperLVSMismatch",
+    "CopperLVSResult",
     "LVSMismatch",
     "LVSResult",
     "_ref_of",
+    "compare_copper_netlist",
     "compare_netlists",
+    "compare_partitions",
 ]
--- a/src/kicad_tools/lvs/copper_lvs.py
+++ b/src/kicad_tools/lvs/copper_lvs.py
@@ -1,0 +1,210 @@
+"""Independent copper-extracted LVS (issue #3742).
+
+This is the "third leg" of board soundness the external reviewer asked
+for: a netlist-correspondence gate that is independent of the router's
+pad-net labels.  Where :mod:`kicad_tools.lvs.board_lvs` trusts each
+pad's declared ``(net K "NAME")`` child (and so passes a board whose
+router mislabels its own copper), this comparator extracts the
+*physical* pad partition straight from routed copper — via
+:meth:`ConnectivityValidator.extract_pad_partition`, which never reads a
+net label — and diffs that against the schematic partition.
+
+Two failure classes are reported:
+
+* **short** — two pads on *different* schematic nets land in the *same*
+  copper component (copper fuses nets that should be isolated).  This is
+  the board-00 failure mode: ``GND`` shorted to ``LED_ANODE`` because the
+  router wired copper to the wrong pad, even though the pad labels still
+  read correctly.
+* **open** — two pads on the *same* schematic net land in *different*
+  copper components (the net is not fully connected by copper).
+
+The comparator is pure: it returns data, never raises for a mismatch.
+Pads present on only one side (schematic-only or PCB-only) are ignored
+for the partition diff — that asymmetry is the label-based comparator's
+job (:func:`board_lvs.compare_netlists`), and the two checks are meant
+to run side by side.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+
+from kicad_tools.lvs.board_lvs import _schematic_pin_to_net
+
+
+@dataclass(frozen=True)
+class CopperLVSMismatch:
+    """A single copper-vs-schematic partition disagreement.
+
+    Attributes:
+        kind: ``"short"`` (different schematic nets fused in copper) or
+            ``"open"`` (same schematic net split across copper islands).
+        net_a / net_b: The schematic net names involved.  For a short
+            these differ; for an open they are equal.
+        pad_a / pad_b: The two offending pads (``"REF.PAD"`` form) that
+            witness the mismatch.
+    """
+
+    kind: str
+    net_a: str
+    net_b: str
+    pad_a: str
+    pad_b: str
+
+
+@dataclass(frozen=True)
+class CopperLVSResult:
+    """Outcome of comparing routed copper against the schematic netlist.
+
+    ``clean`` is ``True`` iff ``mismatches`` is empty.  ``shorts`` and
+    ``opens`` partition the mismatches by kind for convenient reporting.
+    """
+
+    clean: bool
+    mismatches: tuple[CopperLVSMismatch, ...]
+
+    @property
+    def shorts(self) -> tuple[CopperLVSMismatch, ...]:
+        return tuple(m for m in self.mismatches if m.kind == "short")
+
+    @property
+    def opens(self) -> tuple[CopperLVSMismatch, ...]:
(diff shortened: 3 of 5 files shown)


rjwalters:
✅ **Approved!**

Independent copper-extracted LVS gate is sound, well-scoped, and well-tested. Reviewed against the curated scope for #3742.

## Soundness property (load-bearing) — verified
`ConnectivityValidator.extract_pad_partition()` floods the routed copper graph from track segments, segment chains, and vias with **zero reference to pad `(net ...)` labels**. The segment loop walks `self.pcb.segments` directly (not net-filtered), so a mislabeled segment still physically connects whatever it touches. `test_extract_pad_partition_ignores_pad_labels` proves two differently-labeled pads joined by copper collapse into one component, and `test_extract_pad_partition_unconnected_pad_is_singleton` covers the no-copper case. This is genuinely label-independent for autorouter copper — exactly where the #3739 rotation-convention bug manifested.

## Zone-pour declared-net grouping — acceptable bounded trade-off
The pour leg (`2d`) groups pads by the zone's declared net. I confirmed the union only ever connects pads whose declared net equals the zone net, so the worst case is a **false negative (missed short)**, never a false positive — it cannot fabricate a short across nets. The soundness-critical path (segments + vias) stays 100% label-free, and the limitation is clearly documented inline and tracked as a #3742 follow-up. This is a sensible bounded first slice, not a correctness hole.

## FAIL-pre-fix / PASS-post-fix property — covered
- `test_90deg_adversarial_copper_fails`: mirrored copper → OUT1/OUT2 short + SIG open.
- `test_90deg_corrected_copper_passes`: copper on the true SIG pads → clean.
- `test_90deg_label_based_lvs_is_blind_to_the_misroute`: documents why the new gate is needed (label path reads the misroute as correct).
- `test_transform_pad_position_matches_golden`: 0/90/180/270° decoupling guard against the pcbnew-verified golden.

## CLI integration
`_lvs_subcheck` runs both comparators and fails if either dirties, surfacing the copper mismatch first. The `sch_path is None → NOT RUN` guard is preserved, and the label-based path (`compare_netlists`) is untouched.

## Test Execution
- `pytest tests/test_copper_lvs.py tests/test_cli_check.py` — 62 passed.
- `pytest tests/test_lvs.py tests/test_board_00_lvs.py tests/test_connectivity.py tests/test_output_connectivity.py tests/test_pcb_nets_connectivity.py tests/test_routing_connectivity.py tests/test_connectivity_rule.py` — 143 passed (label-based + connectivity unchanged).
- `ruff check` / `ruff format --check` clean on all changed files.

## CI
All 9 checks pass (Test, Board 00 E2E, Diff-Pair, Match-Group, Type Check, Lint, DRC, C++ Build, kicad-cli round-trip). `mergeStateStatus` is CLEAN. PR description uses `Closes #3742`.

PR #3760, opened 2026-06-17
ci: drop push trigger and validate guard from label-external-issues

### LINKED ISSUE #3759: CI noise: label-external-issues.yml reports a failed run on every push (0 jobs executed)
## Problem

`.github/workflows/label-external-issues.yml` shows a **red/failed run on essentially every push to `main`**, cluttering the Actions tab and the commit-status UI with spurious failures. The release that just shipped (v0.14.0) surfaced it again — the failing run on `367cb1be` (a CHANGELOG-only commit) is an example.

## Root cause

The failing runs have **zero jobs executed** (`gh run view <id> --json jobs` → `"jobs": []`). This is the well-known GitHub Actions behavior: **a workflow run with no executed jobs is marked as `failure`.**

The workflow triggers on:

```yaml
on:
  issues:
    types: [opened]
  push:
    paths:
      - '.github/workflows/label-external-issues.yml'
```

The `push.paths` filter restricts the push trigger to changes to the workflow file *itself*. The intent was apparently "validate the workflow when it's edited," backed by a `validate` guard job (`if: ${{ true }}`) added specifically to avoid the no-jobs-fails quirk. In practice, on a normal push that does **not** touch the workflow file, GitHub still records a run for the workflow but filters every job out via the path filter → 0 jobs → the run is reported as **failed**. The guard job does not help because the path filter strips it too.

The `label_external` job (the workflow's actual purpose) is gated on `github.event_name == 'issues'` and works fine; only the push-path machinery is broken.

## Proposed fix (preferred)

**Remove the `push:` trigger entirely.** The workflow's sole real responsibility is labeling/welcoming external issues on `issues: [opened]`. The self-referential `push.paths` trigger and the `validate` guard job exist only to "syntax-check on edit," which:
- is not worth a perpetual stream of failed runs, and
- is already covered by opening a PR (the workflow file is visible/diffed in review).

Concretely:
- Delete the `push:` block from `on:`.
- Delete the now-unnecessary `validate` guard job.
- Keep `label_external` (the `issues: opened` path) unchanged.

This leaves the workflow triggering only on `issues: opened`, which is its actual function, and eliminates the spurious push failures.

## Acceptance criteria

- [ ] `.github/workflows/label-external-issues.yml` no longer triggers on `push` (no `push:` key under `on:`), OR — if a push trigger is genuinely wanted — it is restructured so a non-matching push produces a *skipped/neutral* result, never `failure`.
- [ ] Pushing an unrelated commit to `main` produces **no failed `Label external issues` run** (verify in the Actions tab after merge).
- [ ] The external-issue labeling behavior on `issues: opened` is preserved (job/steps for collaborator check, `external` label, welcome comment unchanged).
- [ ] No new failing/empty runs are introduced.

## Notes

- This is the long-standing red workflow referenced in prior audits; it is unrelated to the main `CI` workflow (the release green-gate) and has never blocked releases — it's pure noise.
- YAML-only change; no Python/source impact.

rjwalters:
## Summary

`.github/workflows/label-external-issues.yml` reported a red/failed run on essentially every push to `main`. The workflow triggered on both `issues: opened` and a self-referential `push.paths` filter restricting pushes to changes to the workflow file itself. On any push that does **not** touch the workflow file, GitHub still records a run for the workflow but the path filter strips every job → a 0-job run, which GitHub marks as `failure`. The `validate` guard job (added to dodge the no-jobs-fails quirk) could not help because the path filter removes all jobs before any can run.

This removes the `push:` trigger and the now-pointless `validate` guard job, leaving the workflow to trigger only on `issues: opened` — its actual function.

## Changes

- Removed the `push:` block (and its `paths:` child) from `on:`, leaving only `issues: { types: [opened] }`.
- Removed the `validate` guard job and its explanatory comment.
- Left `permissions:` and the entire `label_external` job (collaborator check, `external` label, welcome comment, and the `if: ${{ github.event_name == 'issues' }}` gate) byte-for-byte unchanged.

| No longer triggers on `push` (no `push:` key under `on:`) | ✅ | `on:` now has the single key `issues`; verified via diff and structural grep |
| Pushing an unrelated commit produces no failed run | ✅ | Workflow no longer triggers on `push` at all, so no run is created (manual verification in Actions tab after merge) |
| External-issue labeling on `issues: opened` preserved | ✅ | `diff` of the `label_external` job vs `main` is empty (byte-for-byte identical) |
| No new failing/empty runs introduced | ✅ | Only trigger remaining is `issues: opened`, which runs the real job |

## Test Plan

- Confirmed the diff is exactly the two intended deletions and nothing else (`git diff`: 13 deletions, 0 additions).
- Confirmed the `label_external` job is identical to `main` via `diff`.
- `uv run ruff check .` and `uv run ruff format . --check` pass (no Python touched).
- Manual post-merge: push an unrelated commit and confirm no new `Label external issues` run appears; open a test issue from a non-collaborator and confirm the `external` label + welcome comment still apply.

Closes #3759

### DIFF
--- a/.github/workflows/label-external-issues.yml
+++ b/.github/workflows/label-external-issues.yml
@@ -3,25 +3,12 @@ name: Label external issues
 on:
   issues:
     types: [opened]
-  push:
-    paths:
-      - '.github/workflows/label-external-issues.yml'
 
 permissions:
   issues: write
   contents: read
 
 jobs:
-  # Guard job to prevent workflow failure on push events
-  # GitHub Actions marks runs as failed when no jobs execute, so this ensures
-  # at least one job runs. Using explicit true condition for clarity.
-  validate:
-    if: ${{ true }}
-    runs-on: ubuntu-latest
-    steps:
-      - name: Workflow syntax valid
-        run: echo "Workflow file validated successfully"
-
   label_external:
     if: ${{ github.event_name == 'issues' }}
     runs-on: ubuntu-latest


rjwalters:
✅ **Approved!**

Clean, well-scoped fix for #3759. Verified each acceptance criterion against the PR branch (`feature/issue-3759`):

**Scope & correctness**
- `on:` block now contains only `issues: { types: [opened] }` — the `push:` key (and its self-referential `paths:` filter) is fully removed. Confirmed structurally: the single top-level key under `on:` is `issues`.
- The `validate` guard job and its explanatory comment are removed.
- The `label_external` job (collaborator check, `external` label, welcome comment) is byte-for-byte identical to `main` — verified via `diff` of the `label_external:`→EOF region (no differences).
- `permissions:` (`issues: write`, `contents: read`) unchanged.

**No lost behavior**
- The dropped `push.paths` trigger only ever re-validated the workflow file on edit. Because the path filter stripped all jobs on any push not touching this file, it produced the 0-job runs GitHub marked as `failure` — the exact bug in #3759. Its removal is the fix and loses no wanted behavior; `issues: opened` (the real function) is preserved.

**YAML well-formedness**
- The three `script: |` blocks are literal block scalars that GitHub Actions parses correctly. A naive PyYAML `safe_load` raises a ScannerError inside the JS template-literal welcome-comment body — but this reproduces identically on `main` (line 73) and on the PR (line 60), so it is a pre-existing PyYAML limitation, not a GitHub Actions parse error and not a regression introduced here.

**CI**
- All 9 checks pass: Board 00 End-to-End, C++ Build, Diff-Pair Routing Regression, Lint & Format, Match-Group Routing Regression, Routed PCB DRC, Test, Type Check, kicad-cli Round-trip Smoke.
- `mergeStateStatus`: CLEAN.

PR description uses `Closes #3759`. Diff is exactly 13 deletions / 0 additions as described.

Issue #3762, opened 2026-06-17
Make copper-LVS a first-class manufacturability leg across all demo boards (#3742 follow-up)


rjwalters:
## Original Issue

> _Preserved verbatim below. The Curator Enhancement section that follows supersedes the original's per-board state table and "bounded first slice" — a lot has changed on `main` since this was filed (#3769/#3761/#3772/#3773 landed; boards 03/04 net-drift reconciled per #3764/#3771/#3765/#3774; board 05 blocked behind #3775/#3766). The verified current state is documented in the Curator Enhancement._

## Problem statement

PR #3757 made the independent copper-extracted LVS gate a hard sub-check inside `kct check`: `_lvs_subcheck` (`src/kicad_tools/cli/check_cmd.py` ~L259) now runs **both** the label-based `compare_netlists` and the copper-extracted `compare_copper_netlist`, and **either** dirtying fails LVS, which rolls into `CheckResults.compute_overall()` and drives the non-zero exit code. So the soundness leg is live *in the tool*.

But it is **not yet a fleet-wide gate**. Two gaps remain from #3742's "make copper-LVS a first-class manufacturability leg" thread:

### Gap 1 — Only board 00 runs LVS in its recipe
A survey of `boards/0*/` shows only `boards/00-simple-led/generate_design.py` calls `run_lvs()` (via `from kicad_tools.lvs import compare_netlists`) and emits `output/lvs.json`. Boards 01-07 do **not** run LVS in their recipes and have no `lvs.json`.

The "manufacturable = 0 under BOTH kicad-cli + kct check" bar is therefore only **connectivity-correct** on board 00. Every other demo board can pass DRC while its routed copper diverges from its schematic — the exact soundness hole #3742 set out to close, just relocated from "the tool can't check it" to "the boards don't run the check."

### Gap 2 — CI gates LVS only for board 00
`.github/workflows/ci.yml` has a full `board-00-end-to-end` job that regenerates board 00, runs `kct check`, and asserts `lvs.json clean=true` + `board.json lvs_clean=true` (via `scripts/ci/check_board_00_e2e.py`). There is no per-board LVS gate asserting the other boards stay connectivity-correct as they are regenerated.

## Proposed solution (original)

[...]

PR #3769, opened 2026-06-17
feat(lvs): label-free zone-pour copper extraction via hole-aware shapely overlap

### LINKED ISSUE #3761: Robust label-free zone-pour copper extraction for independent LVS (#3742 follow-up)
## Problem statement

PR #3757 (closing the bounded first slice of #3742) shipped `ConnectivityValidator.extract_pad_partition()` — an independent copper-extracted netlist gate that floods **track segments + vias with zero reference to pad `(net ...)` labels**. That is the load-bearing soundness property and it fully covers autorouter copper, which is where the #3739 rotation bug manifested.

It deliberately left **one documented false-negative gap**: the zone-pour leg. In `src/kicad_tools/validate/connectivity.py` (step `2d`, ~lines 415-475), pads are tied to a pour by grouping them by the **zone's declared net** label:

```python
for pad_id, pad_pos in pad_positions.items():
    if pad_declared_net.get(pad_id) != zone.net_name:   # <-- trusts the label
        continue
    if not self._pad_layer_matches_zone(...): continue
    if self._point_in_polygon(pad_pos, zone.polygon): pads_in_zone.append(pad_id)
```

The code comment is explicit about why this is a gap and that it is a tracked follow-up:

> Recovering that tie purely from the flattened `filled_polygons` geometry is not reliable — the outer hull and inner thermal/clearance cutout loops are concatenated into one point list, which breaks naive point-in-polygon, and pad-center-to-fill distance does not separate tied pads from neighbouring foreign-net pads (verified on board 00). Robust pour extraction (spoke tracing / kicad-cli cross-check) is an explicit #3742 follow-up.

**Why it matters:** because the pour leg re-introduces a label dependency, a mis-poured or foreign-net zone tie can hide a short involving a power/ground net. A pad whose copper is *physically* connected to the wrong pour, but whose declared net matches a different (correct) pour, will be partitioned by its label rather than by physical copper. The current model can never *fabricate* a false short (it only fuses same-declared-net pads), but it **can mask a real one** on pour-routed nets — exactly the failure class on power-heavy boards (03, 05, 07 all carry many `filled_polygons`). The independent-LVS soundness claim is therefore only complete for segment/via copper today, not for pour copper.

## Why the obvious approaches fail (builder's notes, do not re-litigate)

The PR builder verified on board 00 that:
- **Naive boundary-containment** (`point_in_polygon` against `zone.polygon`): the flattened `filled_polygons` concatenates the outer hull with inner thermal/clearance cutout loops into a single point list, so a pad sitting in a carved-out moat still tests "inside."
- **Pad-center-to-fill distance**: does not separate a thermally-tied pad from a neighbouring foreign-net pad whose clearance moat happens to be narrow.

A correct extractor must distinguish "pad galvanically bonded to this pour island" from "pad geometrically over this pour but moated out."

## Proposed solution (pick ONE in the design phase; spike both legs)

Design an extractor that recovers pour-to-pad galvanic ties **without trusting `zone.net_name` or `pad.net_name`**. Candidate approaches, in rough order of preference:

1. **Polygon-island flood + pad-overlap with proper hole semantics.** Parse `filled_polygons` into proper rings (outer + holes) using the even-odd / winding rule so carved moats are real holes, not part of the boundary. A pad is bonded to a fill island iff its copper geometry (pad shape, not just center) overlaps the island's *solid* region after subtracting holes. `shapely` is already a dependency (CI installs it for `kct check`), so `Polygon(shell, holes).intersects(pad_shape)` is available. This makes the extraction purely geometric and label-free.

2. **Thermal-spoke tracing.** Detect the spoke segments that bridge a pad to the surrounding pour (KiCad emits these as part of the fill) and union pad↔island only when a spoke physically lands on the pad. This directly models the galvanic tie and naturally excludes moated foreign pads.

3. **kicad-cli cross-check.** Use `find_kicad_cli()` (`src/kicad_tools/cli/runner.py`) to have KiCad itself report pour connectivity, and treat that as the authoritative reference (mirrors the decoupling-golden strategy already used for the 90°/270° pad-geometry test in `tests/test_copper_lvs.py`). Heavier, but the most defensible "independent" reference.

The recommendation is to **prototype approach 1 first** (it removes the label dependency with the least machinery and reuses `shapely`), and fall back to/augment with 2 or 3 if the hole-aware overlap proves insufficient on a real pour-heavy board (use board 03 or 05's routed PCB, both carry 17 `filled_polygons`).

## Acceptance criteria

- [ ] `extract_pad_partition()` no longer consults `zone.net_name` or `pad.net_name` for the pour leg; pour-to-pad ties are derived from `filled_polygons` geometry (and/or thermal spokes / kicad-cli) alone.
- [ ] The flattened-polygon hole problem is handled: a pad sitting in a clearance moat / antipad carved out of a pour is **not** tied to that pour. A pad with a real thermal/solid connection **is** tied.
- [ ] **Adversarial regression fixture (the crux):** a board where a power/ground net is *poured*, and a pad is mislabeled such that its declared net matches the pour but its **physical copper** does not connect to the pour island (or vice-versa: a foreign-net pad whose copper *does* bond to the pour). The new pour extraction MUST report the short/open; the old declared-net model MUST NOT (demonstrating the gap is closed).
- [ ] No false shorts introduced on the existing clean demo boards: `compare_copper_netlist` stays clean on boards 00, 01 (and any others wired in by the fleet-rollout issue) — gate this with the existing `tests/test_copper_lvs.py::test_compare_copper_netlist_on_board00_artifacts`-style end-to-end on a pour-heavy board.
- [ ] The pour leg's correctness is documented as no longer label-dependent; update the `2d` SCOPE NOTE comment block to reflect the new model.
- [ ] Existing label-based and segment/via copper paths unchanged; `tests/test_lvs.py`, `tests/test_board_00_lvs.py`, `tests/test_connectivity.py` still pass.

## Bounded first slice

If hole-aware geometry + a full adversarial fixture is too large for one PR, the minimal load-bearing slice is:
1. Hole-aware `filled_polygons` parsing (rings + holes) + `shapely` pad-overlap for the pour leg, label-free.
2. ONE adversarial pour fixture (inline S-expr, matching `tests/test_lvs.py` style) proving the declared-net model masked a short that the new model catches.

Defer (follow-ups): thermal-spoke tracing as a second cross-check, per-board kicad-cli pour cross-check on every run, multi-island pours with stitching vias.

## Affected files (starting point — curator must verify)

- `src/kicad_tools/validate/connectivity.py` — replace the declared-net pour grouping in `extract_pad_partition()` step `2d` (~L415-475) with label-free island/overlap logic; possibly add a `_fill_islands_with_holes()` helper. `_point_in_polygon` (~L530) and `_pad_layer_matches_zone` are the current building blocks.
- `src/kicad_tools/lvs/copper_lvs.py` — no signature change expected; verify `compare_partitions` still classifies the new pour-derived opens/shorts correctly.
- `tests/test_copper_lvs.py` — add the adversarial pour fixture + a pour-heavy clean end-to-end (board 03 or 05 artifact).
- `tests/fixtures/` (or inline S-expr) — NEW pour fixture with carved moats and a thermal-spoke tie.

## Context links

- Issue #3742 (parent soundness story) + its curator comment (the "two deferred threads" — this is thread (a)).
- PR #3757 (the bounded first slice that shipped; see its "Changes" note on the zone-pour scope).
- #3739 (rotation root cause, fixed in `61763cf0`).

---

### Verified location of the declared-net pour grouping

Confirmed in `src/kicad_tools/validate/connectivity.py`, method `extract_pad_partition()` (def at **L311**). The pour leg is step **2d** and the line refs in the original body are accurate on current main:

- **L446-453** — builds the `pad_declared_net` map (`pad_declared_net["REF.PAD"] = pad.net_name`).
- **L455-475** — the zone loop. The load-bearing label gate is **L464**:
  ```python
  if pad_declared_net.get(pad_id) != zone.net_name:   # <-- trusts the label
      continue
  ```
  followed by `_pad_layer_matches_zone(...)` (L466) and `_point_in_polygon(pad_pos, zone.polygon)` (L468).
- The SCOPE NOTE comment block to update spans **L415-445**.

Building blocks (unchanged, reuse as needed):
- `_point_in_polygon` — **L819** (ray-casting; naive, no hole semantics).
- `_pad_layer_matches_zone` — **L851** (handles `*.Cu` wildcards; keep for the layer guard).
- `_find_islands` — **L878** (BFS components; the partition is already island-based downstream).

### CRITICAL dependency note: `shapely` is NOT a core dependency

Verified in `pyproject.toml` on origin/main: `shapely>=2.0` appears **only** under `[project.optional-dependencies]` (the `geometry` extra, L88-91) and the `dev` extra (L217). It is **absent from the core `[project.dependencies]`**. `connectivity.py` currently has **no `shapely` import at all**. This is exactly why the #3757 builder had to `uv sync --extra dev`.

Implementation requirement: the pour extractor MUST import `shapely` **lazily / guarded** (local import inside the pour helper, or a module-level try/except) so that `from kicad_tools.validate.connectivity import ConnectivityValidator` does not hard-fail in a core-only install. Mirror the pattern in `copper_lvs.py` (lazy `from kicad_tools.validate.connectivity import ConnectivityValidator` at L203) and gate any shapely-dependent end-to-end test with a skip when shapely is unavailable (same policy as the board-00 artifact skip at `tests/test_copper_lvs.py` L449+).

### Schema reality for `filled_polygons` (confirms the hole problem)

Verified in `src/kicad_tools/schema/pcb.py` (`class Zone`, L1028+):
- `filled_polygons: list[list[tuple[float, float]]]` — each fill is a **single flat point list** (`_parse_polygon_pts`, L1177), with outer hull + thermal/clearance cutout loops concatenated. The parser does **not** separate rings from holes.
- `filled_polygon_layers` is parallel to `filled_polygons`; use `zone.filled_polygon_layer(i)` for per-fill layer rather than the zone-level `zone.layer`.

So approach 1 (preferred) must re-derive ring/hole structure from each flat list before `shapely.Polygon(shell, holes)` — e.g. split on closing-vertex repeats / nesting, or feed the raw ring set through shapely's even-odd `unary_union` semantics. This re-derivation is the real work; it is the reason naive `point_in_polygon` against the concatenated list reports a moated-out pad as "inside."

### Pad-overlap geometry note

`extract_pad_partition` already computes board-frame pad **centers** (`_transform_pad_position`, L757) but not pad **outlines**. The AC requires testing pad *copper geometry* (not just center) against the solid fill region. The minimal slice may approximate the pad as its center point against the hole-aware solid region (sufficient to catch the adversarial moat case); a full pad-shape polygon is a documented follow-up if center-point overlap proves insufficient on board 03/05. State which approximation you chose in the SCOPE NOTE update.

### Sharpened adversarial acceptance criterion (the crux)

The single load-bearing proof is ONE inline-S-expr fixture (matching the `_write(...)` + inline-PCB style already used by `tests/test_copper_lvs.py::test_extract_pad_partition_ignores_pad_labels`, L371) with this exact property:

- A power/ground net is **poured** (`filled_polygon` with a carved clearance moat).
- Pad X is **declared** as the pour's net but its **physical copper is moated out** (no thermal spoke / no solid overlap) — i.e. an **open**: the OLD declared-net model fuses X into the pour island; the NEW geometric model leaves X isolated.
- (and/or the dual) Pad Y is a **foreign** declared net whose copper **does** bond to the pour island via a spoke/solid overlap — a **short** the OLD model masks (different declared net -> never fused) and the NEW model catches.

The test must assert BOTH models on the SAME fixture: a helper/assertion showing the declared-net partition would PASS (mask the defect) while `extract_pad_partition()` FAILs (reports the open/short). That side-by-side is the demonstration that the gap is closed — not just that the new model is correct in isolation.

### Pour-heavy clean end-to-end

Verified present on origin/main: `boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb` (+ `.kicad_sch`) and `boards/05-bldc-motor-controller/output/bldc_controller_routed.kicad_pcb` (+ `.kicad_sch`), both pour-heavy (17 `filled_polygons`). Use one of these for the no-false-shorts end-to-end, mirroring `test_compare_copper_netlist_on_board00_artifacts` (L449), with the same artifact-absent / shapely-absent skip policy.

### Scope guard

In scope: the pour-extraction replacement in step 2d + its tests (adversarial fixture + one pour-heavy clean end-to-end). **Out of scope** (do NOT pull in): the fleet-wide LVS rollout wiring across all demo boards (that is **#3762**); thermal-spoke tracing as a second cross-check; per-run kicad-cli pour cross-check; multi-island pours with stitching vias (all deferred follow-ups per the body's "Defer" list).

---

## Affected Files

- `src/kicad_tools/validate/connectivity.py` — Replace the declared-net pour grouping in `extract_pad_partition()` step 2d (L446-475): drop the `pad_declared_net` build + the `zone.net_name` label gate (L464); derive pour-to-pad ties from hole-aware `filled_polygons` geometry (shapely `Polygon(shell, holes)` overlap). Add a `_fill_islands_with_holes()` (or similarly named) helper. Keep `_pad_layer_matches_zone` (L851) for the layer guard. Update the SCOPE NOTE comment block (L415-445) to describe the new label-free model. Use a **lazy/guarded shapely import**.
- `src/kicad_tools/lvs/copper_lvs.py` — No signature change expected; verify `compare_partitions` (L77) / `compare_copper_netlist` (L184) still classify the new pour-derived opens/shorts correctly.
- `tests/test_copper_lvs.py` — Add (1) the adversarial inline-S-expr pour fixture proving the declared-net model masked a defect the new model catches, and (2) a pour-heavy clean end-to-end against board 03 or 05 artifacts (with artifact/shapely skip guard).

(No new `tests/fixtures/` file is required — the existing pour/copper tests use inline S-expr strings via the local `_write(...)` helper.)

## Test Plan

- [ ] Automated — adversarial pour fixture: inline-S-expr board with a poured power/ground net and a carved clearance moat. Assert `extract_pad_partition()` reports the open (moated-out same-declared-net pad) and/or short (foreign-net pad whose copper bonds to the island), AND that the declared-net model would have masked it on the same fixture. (`tests/test_copper_lvs.py`)
- [ ] Automated — hole semantics: a pad sitting purely in a clearance moat / antipad is NOT tied to the pour; a pad with a real solid/thermal overlap IS tied.
- [ ] Automated — pour-heavy clean end-to-end: `compare_copper_netlist` stays clean on board 03 or 05 routed artifacts (skip when artifacts or shapely absent).
- [ ] Regression — unchanged paths: `uv run --extra dev pytest tests/test_lvs.py tests/test_board_00_lvs.py tests/test_connectivity.py tests/test_copper_lvs.py` all pass (segment/via tracing and label-based paths untouched).
- [ ] Manual — confirm core-only install does not break: `from kicad_tools.validate.connectivity import ConnectivityValidator` imports without shapely present (lazy import verified).
- [ ] Lint/type: `uv run ruff check src/ tests/`, `uv run mypy src/kicad_tools/validate/connectivity.py`.

rjwalters:
## Summary

Closes the documented zone-pour false-negative gap from the #3742 first slice (PR #3757). `ConnectivityValidator.extract_pad_partition()` now ties pads to copper pours **purely geometrically** — by shared metal, never by declared net name — so a mislabeled or mis-routed pour can no longer mask a short/open on power/ground nets.

Previously, step 2d grouped pour pads by `pad_declared_net.get(pad_id) != zone.net_name`. That re-introduced a label dependency for the fill leg and could **mask** a real defect: a pad whose copper is physically bonded to the wrong pour, but whose declared net matches a different (correct) pour, was partitioned by its label rather than by metal.

## Changes

- **`src/kicad_tools/validate/connectivity.py`**
  - Replaced step 2d's declared-net pour grouping with label-free geometric bonding:
    - `_fill_solid_region()` resolves each flat `filled_polygon` point list into a hole-aware shapely solid region via `Polygon(pts).buffer(0)`. KiCad flattens an island's clearance moats / thermal antipads into the single `(pts ...)` list by bridging them into the outer hull; `buffer(0)` undoes that so the moats become real holes (a naive ray-cast against the raw list mis-counts the bridge crossings and reports a moated-out pad as "inside").
    - `_pad_copper_polygon()` builds a board-frame, footprint-rotated pad **box** (not just the center) and erodes it by `POUR_PAD_ERODE` (0.1 mm). The box is required so a thermally-relieved pad (center in the antipad moat, copper edge on the spokes) still bonds; the erosion stops an oversized through-hole pad from grazing a corner across a clearance moat into a foreign pour (which would manufacture a false short).
    - `_connect_pour_pads_label_free()` unions pads whose eroded box overlaps a fill's solid region on a matching copper layer.
  - **Guarded/lazy `shapely` import** (optional `geometry`/`dev` extra). When shapely is absent the pour leg falls back to the preserved legacy declared-net grouping (`_connect_pour_pads_by_declared_net`), so core-only installs still import and run. The autorouter segment/via tracing (steps 2a-2c) is untouched.
  - Updated the step-2d SCOPE NOTE and the method docstring to describe the new label-free model and the box/erosion approximation.

- **`tests/test_copper_lvs.py`** — added the adversarial fixture + hole-semantics + pour-heavy end-to-end (all skip-guarded on shapely).

| Pour leg no longer consults `zone.net_name` / `pad.net_name` | ✅ | Geometric `_connect_pour_pads_label_free`; `test_pour_extraction_ignores_zone_and_pad_net_labels` relabels everything to one net and the partition is unchanged |
| Flattened-polygon hole problem handled (moat pad NOT tied; thermal/solid pad IS tied) | ✅ | `_fill_solid_region` buffer(0); `test_pour_extraction_moated_pad_is_not_bonded` + `test_pour_extraction_solid_overlap_pad_is_bonded` |
| Adversarial fixture: OLD model masks, NEW model catches | ✅ | `test_pour_extraction_label_free_catches_what_declared_net_masks` asserts BOTH models on the same board — OLD clean, NEW reports GND/SIG short + GND open |
| No false shorts on clean demo boards | ✅ | Boards 00, 01, 02, 06, 07 stay clean under the new model (were clean before); board 07 wired as a pour-heavy end-to-end (45 `filled_polygons`) |
| 2d SCOPE NOTE updated | ✅ | Comment block + docstring rewritten |
| Existing label-based + segment/via paths unchanged | ✅ | `tests/test_lvs.py`, `tests/test_board_00_lvs.py`, `tests/test_connectivity.py` pass (112 total with `test_copper_lvs.py`) |

## Test Plan

- `uv run --extra dev pytest tests/test_copper_lvs.py tests/test_lvs.py tests/test_board_00_lvs.py tests/test_connectivity.py` → 112 passed.
- The crux side-by-side: on a poured-GND fixture with a carved moat, the legacy declared-net model returns a clean partition (masking a GND open from a moated-out GND pad and a GND/SIG short from a foreign pad bonded to the pour), while `extract_pad_partition()` returns both defects.
- Spot-checked all routed demo boards: every board clean under the old model stays clean under the new one (no false shorts introduced); the dirty boards (03/04/05) were dirty under both models (genuine label/schematic discrepancies, out of scope here).
- `uv run ruff check` / `ruff format --check` clean on both files. mypy adds no new baseline errors (the shapely import is `# type: ignore[import-untyped]`, matching the project pattern). The one NEW baseline error the gate reports (`recovery/strategy.py`) is pre-existing on `origin/main` and unrelated to this change.

## Scope notes / deferred (per the issue)

Out of scope and deferred as documented follow-ups: thermal-spoke tracing as a second cross-check, per-run kicad-cli pour cross-check, multi-island pours with stitching vias, exact pad-outline (roundrect/oval) overlap, and the #3762 fleet-wide LVS rollout.

Closes #3761

### DIFF
--- a/src/kicad_tools/validate/connectivity.py
+++ b/src/kicad_tools/validate/connectivity.py
@@ -23,10 +23,31 @@ from __future__ import annotations
 from collections import defaultdict
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
     from kicad_tools.schema.pcb import PCB
+
+# Optional geometry backend.  ``shapely`` is NOT a core dependency (it lives
+# only under the ``geometry``/``dev`` extras in pyproject.toml), so we import
+# it lazily/guardedly here.  When it is absent the label-free pour extractor
+# (``_connect_pour_pads_label_free`` / step 2d) transparently falls back to
+# the legacy declared-net pour grouping, so importing this module — and
+# tracing autorouter segment/via copper — never hard-fails on a core-only
+# install.
+_SHAPELY_AVAILABLE = False
+try:  # pragma: no cover - import guard exercised by environment, not tests
+    from shapely.geometry import Point as _ShapelyPoint  # type: ignore[import-untyped]
+    from shapely.geometry import Polygon as _ShapelyPolygon
+
+    _SHAPELY_AVAILABLE = True
+except ImportError:  # pragma: no cover
+    pass
+
+
+def _has_shapely() -> bool:
+    """Return True when the optional ``shapely`` backend is importable."""
+    return _SHAPELY_AVAILABLE
 
 
 @dataclass(frozen=True)
@@ -322,13 +343,16 @@ class ConnectivityValidator:
         router that wires segments to the wrong pads while labeling them
         correctly (the board-00 rotation-convention bug, #3739).
 
-        Zone pours are handled with a deliberately narrower, documented
-        model (see the ``2d`` block below): because robust label-free
-        extraction of which pads a pour is *thermally tied* to is a known
-        follow-up, the pour leg groups pads by the zone's declared net.
-        That re-introduces a label dependency for the FILL step alone — never
-        for autorouter copper — and can never fuse two different nets, so it
-        cannot manufacture a false short.
+        Zone pours are now traced label-free as well (issue #3761, see the
+        ``2d`` block below): each ``filled_polygon`` is a poured copper island
+        and a pad is tied to it iff the pad's copper geometrically overlaps
+        the island's *hole-aware solid* region on a matching layer — clearance
+        moats / thermal antipads carved out of the pour are excluded.  No
+        pad/zone ``net_name`` is consulted, so a pad bonded to the wrong pour
+        is no longer masked by a matching declared label.  This requires the
+        optional ``shapely`` backend; when it is absent the pour leg falls
+        back to the legacy declared-net grouping so core-only installs still
+        import and run.
 
         Why this matters: the label-based comparator
         (:func:`kicad_tools.lvs.board_lvs.compare_netlists`) trusts the
@@ -412,64 +436,59 @@ class ConnectivityValidator:
                 for other in via_pads[i + 1 :]:
                     _connect(p, other)
 
-        # 2d. Filled zones (copper pours).
+        # 2d. Filled zones (copper pours) — LABEL-FREE (issue #3761).
         #
-        #     SCOPE NOTE (issue #3742, bounded first slice): the
-        #     load-bearing soundness property of this extractor is that
-        #     *autorouter copper* — track segments and vias — is traced
-        #     fully independently of pad net labels.  That is exactly where
-        #     the board-00 rotation-convention bug manifested: the router
-        #     wired SEGMENTS to the wrong pads while labeling them
-        #     correctly, and steps 2a–2c above will catch that regardless of
-        #     any shared coordinate convention.
+        #     SCOPE NOTE: the load-bearing soundness property of this
+        #     extractor is that copper connectivity is derived from *physical
+        #     geometry*, never from pad/zone net labels.  Steps 2a–2c trace
+        #     autorouter copper (segments + vias) independently of labels;
+        #     this step does the same for pours.
         #
-        #     Zone pours are different in kind.  A copper pour connects only
-        #     the pads it is *electrically tied* to (via thermal spokes);
-        #     pads of other nets sit inside the pour's outline but are carved
-        #     out by clearance moats.  Recovering that tie purely from the
-        #     flattened ``filled_polygons`` geometry is not reliable — the
-        #     outer hull and inner thermal/clearance cutout loops are
-        #     concatenated into one point list, which breaks naive
-        #     point-in-polygon, and pad-center-to-fill distance does not
-        #     separate tied pads from neighbouring foreign-net pads (verified
-        #     on board 00).  Robust pour extraction (spoke tracing /
-        #     kicad-cli cross-check) is an explicit #3742 follow-up.
+        #     The previous model (the #3742 first slice) grouped pads by the
+        #     zone's *declared* net (``pad_declared_net[pad] == zone.net_name``).
+        #     That re-introduced a label dependency for the fill step and could
+        #     MASK a real defect on pour-routed nets: a pad whose copper is
+        #     physically bonded to the *wrong* pour island, but whose declared
+        #     net matches a different (correct) pour, was partitioned by its
+        #     label rather than by metal (issue #3761).
         #
-        #     So for the pour leg ONLY we group pads by the zone's *declared
-        #     net*: a pad enclosed by a pour's boundary on a matching layer
-        #     is treated as tied to that pour iff the pad's own declared net
-        #     equals the zone's.  This re-introduces a label dependency for
-        #     the FILL step alone (a deterministic, label-driven generator
-        #     step — not the autorouter), while keeping segment/via tracing
-        #     100% label-independent.  Crucially it can never fuse two
-        #     different nets, so it cannot manufacture a false short.
-        pad_declared_net: dict[str, str] = {}
-        for fp in self.pcb.footprints:
-            if not fp.reference or fp.reference.startswith("#"):
-                continue
-            for pad in fp.pads:
-                if pad.number is None or pad.number == "":
-                    continue
-                pad_declared_net[f"{fp.reference}.{pad.number}"] = pad.net_name
-
-        for zone in self.pcb.zones:
-            if not zone.filled_polygons:
-                continue
-            if not zone.polygon or len(zone.polygon) < 3:
-                continue
-            if not zone.net_name:
-                continue
-            pads_in_zone: list[str] = []
-            for pad_id, pad_pos in pad_positions.items():
-                if pad_declared_net.get(pad_id) != zone.net_name:
-                    continue
-                if not self._pad_layer_matches_zone(pad_layers.get(pad_id, []), zone.layer):
-                    continue
-                if self._point_in_polygon(pad_pos, zone.polygon):
-                    pads_in_zone.append(pad_id)
-            for i, p in enumerate(pads_in_zone):
-                for other in pads_in_zone[i + 1 :]:
-                    _connect(p, other)
+        #     We now tie pads to pours purely geometrically.  Each
+        #     ``filled_polygon`` is one poured copper *island*.  A pad is
+        #     bonded to an island iff its copper geometry overlaps the
+        #     island's *solid* region on a matching copper layer — clearance
+        #     moats / thermal antipads carved out of the pour are real holes
+        #     that the pad must NOT be tied across.
+        #
+        #     KiCad encodes a fill island's holes inside a single flat
+        #     ``(pts ...)`` list: the outer hull and each carved-out loop are
+        #     joined by a narrow bridge, so the boundary dips *around* every
+        #     moat.  A ray-cast against that raw list mis-counts the bridge
+        #     crossings and reports a moated-out pad as "inside" (the exact
+        #     failure ``_point_in_polygon`` exhibits, verified on board 00).
+        #     ``shapely`` resolves the bridged representation correctly:
+        #     ``Polygon(pts).buffer(0)`` yields the true solid region with the
+        #     moats excluded, so a hole-aware ``contains`` test is sound.
+        #
+        #     Pad-shape approximation: we test the pad's *size box* (board
+        #     frame, footprint-rotated — see ``_pad_copper_polygon``) against
+        #     the hole-aware solid region, not just the pad center.  The box
+        #     is required, not a nicety: a thermally-relieved pad's center
+        #     sits in the antipad moat (a hole), yet its copper edge reaches
+        #     the thermal spokes / surrounding solid pour, so only the box
+        #     intersects the solid region.  A pad fully moated out (clearance
+        #     all around, no spoke) stays clear of the solid region and is
+        #     correctly left untied.  Corner rounding (roundrect/oval) and
+        #     per-pad rotation are ignored; an exact pad outline is a
+        #     documented follow-up.
+        #
+        #     ``shapely`` is an optional dependency; when it is absent we fall
+        #     back to the legacy declared-net pour grouping so core-only
+        #     installs keep working (the soundness upgrade simply requires the
+        #     ``geometry``/``dev`` extra to be installed).
+        if _has_shapely():
+            self._connect_pour_pads_label_free(pad_positions, pad_layers, _connect)
+        else:  # pragma: no cover - exercised only on core-only installs
+            self._connect_pour_pads_by_declared_net(pad_positions, pad_layers, _connect)
 
         # 2e. Coincident pads with no intervening copper still share metal
         #     if they occupy the same point (e.g. stacked pads).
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Reviewed in the builder worktree (`.loom/worktrees/issue-3761`). This is a sound LVS-soundness follow-up that closes the #3761 zone-pour false-negative gap. The label-free geometric pour extraction is correct, well-documented, and the adversarial side-by-side proof genuinely holds.

## Correctness of the label-free extraction
- `_fill_solid_region()` — `Polygon(pts).buffer(0)` is the right idiom. KiCad flattens an island's clearance moats/antipads into a single `(pts …)` ring by bridging the cutout loops into the outer hull; a raw ray-cast mis-counts the bridge crossings (the `_point_in_polygon` failure), whereas `buffer(0)` re-derives the true solid with the moats as real holes. The degeneracy guard (`len < 3`, `is_empty → None`) is appropriate.
- `_pad_copper_polygon()` — the rotated *box* (not center) is required so a thermally-relieved pad whose center sits in the antipad moat but whose copper edge reaches the spokes still bonds; the inward `buffer(-0.1)` erosion stops an oversized through-hole pad from grazing a corner across a clearance moat into a foreign pour (a manufactured false short). The empty-after-erosion guard keeps tiny pads testable rather than silently dropping them. Negated-angle convention matches `_transform_pad_position`.
- The `POUR_PAD_ERODE = 0.1` constant is **justified and documented** at the definition site (sized above the corner-graze scale, below a real spoke's penetration) and again at the call site, with the board 05 false-mega-short removal / board 00 thermal-tie preservation rationale called out. It is a coarse approximation (roundrect/oval corners and per-pad rotation ignored) but that is explicitly scoped as a documented follow-up, and erosion only *removes* marginal bonds — it cannot fabricate a short, so it does not risk masking a real short (only, at most, a real near-graze open, which is the conservative direction for a short-detector).

## The adversarial side-by-side proof — it holds
`test_pour_extraction_label_free_catches_what_declared_net_masks` asserts BOTH models on the **same** fixture:
- OLD model is forced via `mock.patch("…_has_shapely", return_value=False)` → `_connect_pour_pads_by_declared_net`, and is asserted `old_result.clean` (R1.1/R3.1 grouped by GND label, R2.1 a SIG singleton — masks both defects).
- NEW model (`extract_pad_partition()` with shapely) is asserted `not clean`, with a concrete GND/SIG short (R2.1's SIG copper bonds the GND pour with R3.1) **and** a GND open (R1.1 moated out while R3.1 holds GND copper).

This is a true demonstration that the old model masks while the new one catches — not just two independent assertions. The fixture geometry checks out: R1.1's eroded box (≈4.35–5.65, 9.35–10.65) sits inside the 3 mm carved moat; R2.1/R3.1 fall in solid pour. Backed by `test_pour_extraction_moated_pad_is_not_bonded`, `…_solid_overlap_pad_is_bonded`, and `…_ignores_zone_and_pad_net_labels` (relabels everything to one net; partition unchanged — proves no net-name dependency).

## No regressions
- Pads bonded by declared net but not by copper are no longer grouped (that is the whole point — `_connect_pour_pads_label_free` consults no net name).
- Board 07 wired as a pour-heavy end-to-end (45 `filled_polygons`) stays clean; Board 00 End-to-End CI gate passes. Boards 03/04/05 dirty-under-both-models is a pre-existing label/schematic discrepancy, out of scope here.

## Guarded shapely import
Verified directly: with no shapely on `PYTHONPATH`, `connectivity.py` imports cleanly and `_has_shapely()` returns `False`, routing to the preserved `_connect_pour_pads_by_declared_net` fallback. Ran `tests/test_copper_lvs.py` core-only (no shapely): 19 passed, 5 `@requires_shapely` skipped. Autorouter segment/via tracing (2a–2c) untouched.

## Scope
Diff is exactly the two stated files (+543/-63); no #3762 fleet rollout included. PR body uses `Closes #3761`.

## Test Execution
- `uv run --extra dev pytest tests/test_copper_lvs.py tests/test_lvs.py tests/test_board_00_lvs.py tests/test_connectivity.py` → **112 passed** (with shapely).
- Core-only `tests/test_copper_lvs.py` (no shapely) → **19 passed, 5 skipped**, fallback path confirmed.
- Core-only import smoke → imports OK, `_has_shapely() == False`.

## CI
All 9 checks pass (Test, both routing regressions, Board 00 E2E, DRC, C++ build, lint, typecheck, round-trip). Merge state: CLEAN.

PR #3777, opened 2026-06-17
fix(connectivity): union pour pads across all fill islands of one zone

### LINKED ISSUE #3772: copper-LVS: label-free pour extraction (#3769) not bonding pour-served power pads (board 03 GND/VCC/VBUS report as open)
## Summary

Discovered during Judge review of PR #3771 (board-03 schematic↔PCB net-drift reconciliation, closes #3764).

PR #3771 correctly scopes its committed DRC report to `--drc-only` because the copper-LVS sub-check reports pour-served power nets (VCC / VBUS / GND) as "open." That scoping is reasonable for a board-fix PR — the schematic↔PCB netlist equivalence is asserted exactly by `compare_netlists` (clean, 0 mismatches) — but the underlying copper-LVS behavior looks like a real gap worth tracking.

## Observation

On board 03's committed routed PCB (`boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb`):

- The board has **18 filled zone polygons** (pours ARE filled), and **shapely is available** in the environment.
- `kicad_tools.lvs.copper_lvs.compare_copper_netlist(sch, routed_pcb)` reports **15 `kind: "open"` mismatches**, all on power nets where `net_a == net_b` (e.g. `GND`↔`GND`: `C1.2`/`C10.2`, `J1.A12`, etc.).
- Inspecting `ConnectivityValidator.extract_pad_partition()` directly: the GND power pads each land in their **own singleton connected component** — they are NOT bonded to the GND pour.

So #3769's label-free pour extraction (hole-aware shapely overlap, `extract_pad_partition()` step 2d) is **filling polygons but not geometrically bonding these thermally-relieved / through-hole power pads to the pour**, even with shapely present. The opens are not a label-masking artifact — they are a genuine pad↔pour bonding miss.

## Why this isn't a regression in #3771

- On `main` (pre-PR), copper-LVS already reported **63 mismatches** on this board.
- PR #3771 reduces schematic↔PCB `compare_netlists` from 42 → 0 and (as a side effect of the cleaner net model) copper-LVS opens drop to 15.
- The condition is strictly **pre-existing**; #3771 correctly scoped it out and documented it.

## Suspected root cause / where to look

- `ConnectivityValidator.extract_pad_partition()` step 2d (the #3769 pad-to-pour bonding).
- Likely the eroded-pad-box vs. hole-aware pour-region overlap test misses:
  - thermally-relieved SMD pads whose copper reaches the pour only via spokes, and/or
  - through-hole power pads (J1.A12, J2.2) whose annular ring bonds across a clearance moat.
- Worth verifying the per-layer matching and the 0.1mm erosion against these specific pad geometries.

## Reproduction

```python
from kicad_tools.lvs.copper_lvs import compare_copper_netlist
r = compare_copper_netlist(
    "boards/03-usb-joystick/output/usb_joystick.kicad_sch",
    "boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb",
)
print(r.clean, len(r.mismatches))  # False, 15 — all power-net opens
```

## Impact

- Low/medium: blocks a full (non-`--drc-only`) copper-LVS gate for any board that serves power via pours — currently every fleet board with a power pour.
- Once fixed, board-03 (and others) could drop the `--drc-only` scoping and gate on full copper-LVS.

Discovered during evaluation of PR #3771.

rjwalters:
## Summary

Pour-served power pads (GND/VCC/VBUS) were being reported as false copper-LVS `open`s on boards whose power pours KiCad fragments into multiple `filled_polygon` entries. The label-free pour extractor from #3769 unioned bonded pads only *within* a single fill index, so a pad alone in its own thermal-relief fragment landed in a singleton connected component and read as `open`.

Root cause (confirmed by the curator on origin/main): board 03's GND F.Cu zone is stored as 13 separate `filled_polygon` entries — one main pour plus 12 tiny per-pad fragments. A `zone` is one net by KiCad's data model, but the extractor never unioned the fill islands of a single zone.

## Changes

- `src/kicad_tools/validate/connectivity.py` — `_connect_pour_pads_label_free()`: accumulate the bonded-pad set **per `zone` object across all of its `filled_polygon` indices**, de-duplicate preserving order, then union once per zone (instead of per fill index). The bonding *test* (eroded pad box vs hole-aware solid region from #3769) is unchanged, so the anti-false-short property is preserved. Docstring updated to state that fill islands of one zone are one electrical net.
- `tests/test_copper_lvs.py` — new `@requires_shapely` regression: one GND zone with two **disjoint** same-net fill islands, each bonding a distinct GND power pad; asserts both pads share one component. Fails on origin/main (per-fill unioning splits them), passes with the fix.

| Board 03 opens drop 15 → 2 | ✅ | `compare_copper_netlist` on board 03: opens 2, shorts 0 |
| 2 residual opens are J1 USB-C (out of scope) | ✅ | Only J1.A12 / J1.B1 remain (F.Cu-only pads over B.Cu pour, no via) |
| Zero new shorts on any board | ✅ | 00: 0, 03: 0, 04: 1, 05: 18 — all unchanged from baseline |
| Board 00 stays clean | ✅ | opens 0 / shorts 0 / clean=True |
| Board 05 opens improve 87 → 78 | ✅ | `compare_copper_netlist` on board 05: opens 78, shorts 18 |
| Four #3769 adversarial pour tests still pass | ✅ | GND/SIG short + moated-out GND open still flagged |
| New regression pins two-disjoint-islands-one-zone case | ✅ | Fails pre-fix, passes post-fix (verified by git stash) |
| Fix is purely geometric/per-zone (label-free preserved) | ✅ | No pad/zone `net_name` consulted in bonding; shapely-absent fallback untouched |

## Test Plan

- `uv run pytest tests/test_copper_lvs.py` — 25 passed (24 existing + new regression).
- `uv run pytest tests/test_copper_lvs.py tests/test_connectivity.py tests/test_lvs.py tests/test_board_00_lvs.py` — 113 passed.
- New test verified to FAIL on origin/main connectivity (stashed the source change) and PASS with the fix.
- Board-level `compare_copper_netlist` run for 00/03/04/05 (see table above).
- `uv run ruff check .` clean; `ruff format --check` clean; `mypy` introduces no new errors (the 12 remaining are pre-existing on main, lines 629–705, unrelated to this change).

Closes #3772

### DIFF
--- a/src/kicad_tools/validate/connectivity.py
+++ b/src/kicad_tools/validate/connectivity.py
@@ -984,7 +984,7 @@ class ConnectivityValidator:
         """Union pads bonded to the same poured copper island (label-free).
 
         For every ``filled_polygon`` of every zone, build the hole-aware
-        solid region (:meth:`_fill_solid_region`) and union all pads whose
+        solid region (:meth:`_fill_solid_region`) and collect all pads whose
         *copper geometry* overlaps that region on a matching copper layer.  No
         pad/zone ``net_name`` is consulted — pads are tied solely by shared
         metal, so a pad moated out of a pour is left isolated and a
@@ -995,6 +995,19 @@ class ConnectivityValidator:
         sits in the antipad hole, but its copper edge reaches the thermal
         spokes / solid pour, so the box intersects the solid region while the
         center alone would (wrongly) read as moated-out.
+
+        Bonded pads are accumulated **per ``zone`` object across all of its
+        ``filled_polygon`` indices**, then unioned once per zone — not per
+        fill index.  KiCad stores a single poured zone as multiple
+        ``filled_polygon`` entries when thermal reliefs / clearance moats
+        fragment the copper (e.g. board 03's GND F.Cu zone is one main pour
+        plus a dozen tiny per-pad fragments).  All fragments of one ``zone``
+        are the same net by KiCad's data model — DRC guarantees retained
+        fragments are electrically bonded — so unioning across a zone's fill
+        islands cannot fuse two different nets (those are different ``zone``
+        objects).  Without this, a pad alone in its own thermal-relief
+        fragment would land in a singleton component and be reported as a
+        false ``open``.
         """
         # Board-frame pad copper polygons, keyed by pad id.  Built once here
         # so each fill island can be tested against every candidate pad.
@@ -1012,12 +1025,17 @@ class ConnectivityValidator:
         for zone in self.pcb.zones:
             if not zone.filled_polygons:
                 continue
+            # Accumulate the bonded-pad set across ALL fill islands of this
+            # zone (see method docstring): the zone is one net, and KiCad may
+            # fragment its pour into many ``filled_polygon`` entries.  Order is
+            # preserved and duplicates are dropped before unioning so a pad
+            # bonded to several fragments is unioned exactly once.
+            zone_bonded: list[str] = []
             for i, fill_pts in enumerate(zone.filled_polygons):
                 region = self._fill_solid_region(fill_pts)
                 if region is None:
                     continue
                 fill_layer = zone.filled_polygon_layer(i)
-                bonded: list[str] = []
                 for pad_id in pad_positions:
                     if not self._pad_layer_matches_zone(pad_layers.get(pad_id, []), fill_layer):
                         continue
@@ -1037,10 +1055,19 @@ class ConnectivityValidator:
                     # removes all spurious multi-net bridges while preserving
                     # board 00's genuine thermal-relief ties.
                     if region.intersects(pad_geom):
-                        bonded.append(pad_id)
-                for a, p in enumerate(bonded):
-                    for other in bonded[a + 1 :]:
-                        connect(p, other)
+                        zone_bonded.append(pad_id)
+            # De-duplicate preserving first-seen order, then union the whole
+            # zone's bonded set so pads on disjoint fill fragments of one zone
+            # share a single component.
+            seen: set[str] = set()
+            bonded: list[str] = []
+            for pad_id in zone_bonded:
+                if pad_id not in seen:
+                    seen.add(pad_id)
+                    bonded.append(pad_id)
+            for a, p in enumerate(bonded):
+                for other in bonded[a + 1 :]:
+                    connect(p, other)
 
     def _connect_pour_pads_by_declared_net(
         self,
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Reviewed the per-zone pour-pad union fix for #3772. The crux — closing false opens *without* introducing false shorts — holds up.

## Correctness of the per-zone union (net-safe)
`_connect_pour_pads_label_free()` now declares/resets `zone_bonded` **inside** `for zone in self.pcb.zones:` and accumulates across that zone's `filled_polygons` (inner loop). `connect()` is only ever called on pads within one zone's de-duplicated `bonded` set. Two different `zone` objects never share a `connect` group, so the union **cannot fuse two nets** — a KiCad zone is one net by data model, and DRC guarantees retained fragments of a zone are electrically bonded. The bonding *test* itself (`region.intersects(pad_geom)` with the #3769 `_POUR_PAD_ERODE` eroded pad box vs hole-aware solid region) is byte-for-byte unchanged, so the anti-false-short property is preserved. Order-preserving de-dup means a pad bonded to several fragments is unioned exactly once.

## No false-short regression
- Test file diff has **zero deletions** — all #3769 adversarial/eroded-pad bonding tests are untouched.
- Ran the adversarial subset: 10 passed (short-detected, moated-pad-not-bonded, open-when-split, 90deg adversarial).
- `tests/test_copper_lvs.py tests/test_connectivity.py tests/test_lvs.py tests/test_board_00_lvs.py`: **113 passed**.

## New regression test
`test_pour_extraction_unions_pads_across_disjoint_fill_islands_of_one_zone` builds one GND zone with two genuinely **disjoint** fill islands (x 0–4 and x 8–12, 4mm gap), each bonding a distinct GND pad, and asserts they share one component. Verified it **FAILS on origin/main** (reverted `connectivity.py` to main: `partition=[frozenset({'C1.2'}), frozenset({'C2.2'})]` — the two-singleton false-open shape) and **PASSES with the fix**.

## Acceptance numbers (re-ran `compare_copper_netlist` locally)
- Board 03: opens **2**, shorts **0** (was 15). Both residual opens are exactly `J1.A12` / `J1.B1` — the out-of-scope USB-C F.Cu-only-over-B.Cu-pour pads.
- Board 05: opens **78**, shorts **18** (was 87 opens; shorts unchanged → zero new shorts).
- Board 00 stays clean; board 04 unchanged (covered by passing suites).

## Scope & hygiene
- Only `src/kicad_tools/validate/connectivity.py` (+33/-6) and `tests/test_copper_lvs.py` (+91/-0). No board artifacts touched.
- `ruff check` / `ruff format --check` clean on both files.
- All 9 CI checks pass; merge state CLEAN.

PR description uses `Closes #3772`. Clean, well-documented, provably net-safe fix.

Issue #3266, opened 2026-06-07
router: NRST routing regression on board 04 (fresh build drops NRST to advisory connectivity)


rjwalters:
## Context

While verifying board 04 (stm32-devboard) manufacturability per tracking issue #3261 (parent #2394), a fresh \`kct route ... --backend cpp --seed 42\` run consistently drops the NRST net to 1/2 connected (J1.5 stranded).

The committed routed PCB at \`boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb\` (from PR #3128, commit 62918dfb) has NRST fully routed. Re-running the same recipe today produces 8/9 instead of 9/9 nets routed, with NRST as the dropped signal.

## Measurement

Three independent seeds (42, 17, 99) on the C++ backend with current main, same recipe as \`boards/04-stm32-devboard/generate_design.py::route_pcb\`:

\`\`\`
--mfr jlcpcb-tier1 --backend cpp --auto-fix --auto-layers --auto-mfr-tier
--placement-feedback --micro-via-in-pad-fallback --seed <N> --timeout 600
\`\`\`

All three converge to the same outcome:

\`\`\`
Initial pass: 8/9 nets, overflow: 2
Iteration 1: ripping up 2 nets (37.8s)
Iteration 1 complete: clearance_viol=0, overflow=2
Iteration 2: ripping up 2 nets (38.4s)
Iteration 2 complete: clearance_viol=0, overflow=2
...
PARTIAL: Best result 89% on 2 layers
Partially connected nets (1):
  NRST: 1/2 pads connected
\`\`\`

CI gate verdict: PASS (NRST stranding surfaces as advisory \`connectivity\`, which is excluded per \`DRCChecker.ADVISORY_RULE_IDS\`).

But: NRST is a functional reset signal between U2.NRST and J1.5 (SWD debug header). The MCU's hardware reset via SWD is broken when NRST is stranded.

## Likely cause

Between commit 62918dfb (Nov 2025 era, PR #3128) and current HEAD (5e115a45), the router has had ~30+ PRs of changes including:
- #3232/#3248 trace-clearance and via-clearance kernel switch Chebyshev -> Euclidean disc
- #3225/#3227 foreign-pad clearance tightening
- #3250 sub-cell pad-metal margin closure
- #3243 auto-grid-selector memory budget bump
- #3203 per-pad channel budget edge classification fix

[...]

Issue #3775, opened 2026-06-17
board-05: U3-south relayout to free PHASE_A/B/C escape channels without regressing ISENSE/PWM/GATE_DRV


rjwalters:
## Context

Follow-up from #3766 (board-05 routing parity). That issue required PHASE_A/B/C to be connected. Investigation established that PHASE connectivity on board-05 is placement-bound — the far-west bootstrap-cap placement (C12-C14) forces long BST traces that saturate the U3-south escape band.

**Current artifact state (post-PR #3906):** PHASE_A is COMPLETE. 6 blocking signal nets remain (down from 7). The artifact-first method is validated on this board.

## History

- **2026-06-17 Curator:** Analyzed geometry; 7 blocking nets (PHASE_A/B/C + 4 ISENSE). Assessed as non-automatable; kept loom:architect.
- **2026-06-21 Build attempt:** PHASE direction proven (C12-C14 relocation connects PHASE_A/B/C) but blocked by host-vs-CI routing divergence (#3822): host routes unmodified recipe to 11 blocking vs committed 7.
- **2026-06-22:** Marked loom:blocked pending CI-parity path (#3822 + #3823).
- **2026-07-05:** #3822 closed; artifact-first playbook validated on chorus-test v22 (30/51→51/51 strict, kicad-cli DRC clean). Unblocked.
- **2026-07-05 PR #3906 merged:** PHASE_A surgically routed (2.0mm, 5-pad tree, referee 52→38 violations, 7→6 blocking). design.py unchanged. Artifact-first method confirmed working on this board.

## Verified Current State (post-#3906, 2026-07-05)

`uv run kct net-status boards/05-bldc-motor-controller/output/bldc_controller_routed.kicad_pcb --incomplete --why` output:

**Stuck signal nets: 8 classified entries** (6 blocking signal + 2 advisory POUR_DISCONTINUOUS):

| Net | Class | --why verdict | Unconnected pads | Key blocker |
|---|---|---|---|---|
| `PHASE_B` | HIGH_CURRENT_SIGNAL | PLACEMENT_BOUND | C13.2, J2.2, Q3.3, U3.41 | BST_A trace wall along y=59; 34 foreign obstructions, nearest strict blocker 0.381mm |
[...]

PR #3778, opened 2026-06-17
fix(zones): carve foreign-net track segments out of zone fills

### LINKED ISSUE #3773: board-04: fresh regen produces 30 GND-vs-rail F.Cu zone-fill shorts; routed PCB no longer reproducible on main
## Context

Found while working #3765 (board-04 net-drift reconcile + finish-last-net). Part A (schematic<->PCB `compare_netlists`) is fully resolved there. This issue captures a **separate, pre-existing regression on `main`** that blocks end-to-end regeneration of board-04's routed PCB.

## Problem

A fresh end-to-end regen of board-04 on current `main` (`uv run python boards/04-stm32-devboard/generate_design.py`, C++ backend, `--seed 42`, shapely installed) produces a routed PCB with **30 blocking DRC errors** against `jlcpcb-tier1`:

- `clearance_segment_zone` (20x): `GND` track segments overlap the `+3.3V` / `+5V` **F.Cu zone fill** (overlap depth 0.10-0.18 mm).
- `clearance_via_zone` (10x): `GND` vias overlap the `+3.3V` / `+5V` F.Cu zone fill (overlap depth ~0.15-0.21 mm).

The **committed** `output/stm32_devboard_routed.kicad_pcb` has **0 blocking errors** and routes NRST cleanly (`NRST` 2/2 connected; only `GND` U2.23 corner-pad stitch residual remains, the documented #2834/#3033 case). So the committed artifact is a known-good pinned PCB that fresh regeneration can **no longer reproduce**.

### Proven independent of #3765

`git stash`-ing all #3765 edits and regenerating from pristine `main` `generate_design.py` (PCB net table unchanged) reproduces the same 30 zone-fill shorts. The PCB/routing inputs are byte-identical to before; only the router / zone-fill behavior on current `main` differs. The fill polygons differ between committed (clean, ~4384 pts on the big F.Cu zone) and fresh (~3575 pts, overlapping GND copper), so the zone filler is producing fills that are not cleared around the router's GND tracks/vias.

## Impact

- Board-04's routed PCB and manufacturing bundle cannot be regenerated cleanly; #3765 therefore **preserves the committed-good routed PCB** (which is net-consistent with the reconciled schematic -- `compare_netlists` clean -- and has NRST routed + 0 blocking errors) instead of committing the regressed 30-short re-route.
- Every board that re-routes through this zone-fill path is potentially affected.

## Acceptance

- A fresh `uv run python boards/04-stm32-devboard/generate_design.py` produces a routed PCB with **0 blocking DRC errors** against `jlcpcb-tier1` (zone fills cleared around router GND tracks/vias), reproducing the committed-good artifact's quality, with NRST routed.

## Investigation pointers

- Compare zone `filled_polygon` extraction/clearance between the committed routed PCB and a fresh route (the fresh fill overlaps GND copper).
- Likely in the zone-fill / pour-clearance path (`src/kicad_tools/zones/**`) or the post-route auto-pour step invoked by `kct route` / `kct stitch`.

## Related
- #3765 (parent; Part A done, routed PCB preserved pending this fix)
- #2834 / #3033 (GND U2.23 corner-pad stitch residual -- the *remaining* advisory, unrelated to these zone shorts)
- #3772 (pour-extraction LVS gap -- distinct: that is copper-LVS reading pours as "open"; this is DRC shorts in the fill geometry)
- #3266 / #3267 (historical NRST/U2.35 routing notes)

---

### Root cause (localized)

The post-fill foreign-net clearance carve **only subtracts pads and vias** from each zone `filled_polygon` — it never subtracts foreign-net **track segments**. A `GND` trace routed straight across a `+3.3V`/`+5V` `F.Cu` pour is therefore left embedded in the rail copper, and the `SegmentZoneClearanceRule` (`rule_id="clearance_segment_zone"`) reports the overlap as a short.

Concretely, in `src/kicad_tools/zones/fill_clearance.py`:

- `_collect_obstacles(doc, shapely, name_map)` (lines ~121-231) gathers obstacles from **footprint pads** and **top-level vias** only. There is no `doc.find_all("segment")` branch, so foreign-net tracks are never modelled as cutters.
- `apply_foreign_pad_clearance(doc, ...)` (lines ~865-1041) buffers each obstacle by `clearance + min_thickness/2` and subtracts it from the fill — but it can only subtract what `_collect_obstacles` returned. Foreign-net traces pass straight through.

This is the exact asymmetry that produces the bug: `_collect_same_net_anchors` (lines ~366-418) **does** iterate `doc.find_all("segment")` to keep same-net track-connected fill rings, while the foreign-net obstacle collector does **not** iterate segments to carve foreign track copper out. The DRC rule `SegmentZoneClearanceRule._collect_fills` / `_collect_zone_fills` (in `src/kicad_tools/validate/rules/clearance.py`, lines ~827-925) reads the committed `filled_polygon` copper and measures each track segment against it — so any uncarved overlap is reported.

The `clearance_via_zone` (10x) shorts share the same fill: vias *are* collected as obstacles, but the carve's safety gates (`apply_foreign_pad_clearance` rejects a rewrite when `recon.difference(fill_poly).area > _AREA_EPS` or when the carved fill still intersects a cutter, lines ~1025-1028) and the refill/settle loop in `_remediate_starved_thermal` (`src/kicad_tools/cli/runner.py`, lines ~417-538) mean that when the surrounding fill geometry near a GND via still embeds an uncarved GND trace, the via antipad either is not applied cleanly or is reverted on the next `--refill-zones` pass. Once foreign-net **segments** are carved (removing the embedded GND copper that abuts these vias), the via-zone overlaps are expected to resolve as well. The fix should be validated against **both** violation types, not just segments.

### Why this regressed on `main` (per filer's evidence)

The committed `output/stm32_devboard_routed.kicad_pcb` has a clean big F.Cu fill (~4384 pts); a fresh regen produces ~3575 pts overlapping GND copper. The committed artifact was filled by a path (or KiCad GUI) that respected foreign track copper; the current `kct`-driven fill + pure-Python carve does not carve foreign tracks, so fresh regen embeds GND traces in the rail pour. The committed PCB is pinned as a workaround by #3765 / PR #3774.

### Fix approach

Extend the foreign-net carve to include track segments, mirroring how vias are already handled:

1. In `_collect_obstacles` (`fill_clearance.py`), add a `doc.find_all("segment")` branch that, for each net-assigned segment, builds a sheet-absolute `LineString([(sx,sy),(ex,ey)]).buffer(width/2.0)` obstacle on the segment's `(layer ...)` (single layer, unlike vias' span semantics). Use `_net_key(seg.find("net"), name_map)` for net identity and skip net 0. This reuses the exact same geometry the same-net anchor collector already builds (lines ~399-416), so the two stay consistent. Carry the segment's single layer in `_Obstacle.layers` so `_obstacle_on_layer` matches only that copper layer.
   - Note: `_Obstacle.shape` is typed as a `_Geometry` Protocol with `buffer`/`intersects`; a buffered `LineString` (a `Polygon`) satisfies it. `source_pad=None` for segments (they have no thermal relief), matching vias.
2. The rest of `apply_foreign_pad_clearance` needs no change — it already buffers every obstacle by `clearance + min_thickness/2` and subtracts the union, vents resulting holes, runs island-removal, and applies the safety gates. Segments flow through unchanged once collected.
3. Verify the carve survives the `_remediate_starved_thermal` settle loop in `runner.py` (the carve is re-applied every pass via the `settle=_apply_foreign_pad_clearance` callback and once at the end, so segment carving is automatically included — no runner change expected, but confirm by re-running board-04 regen end-to-end).
4. Keep it shapely-guarded: when shapely is unavailable the carve is already a no-op; the new segment branch must sit inside the same `import shapely` guard.

This is a **general zone-fill correctness fix** (a pour must clear foreign-net copper of all kinds — pads, vias, *and* tracks — by the clearance rule), not a board-04-specific patch. Board-04 is the reproducing case.

### Scope assessment

This is a **bounded single-PR fix**, not a zone-fill engine rework. The carve infrastructure (buffering, hole-venting, island removal, safety gates) already exists and is correct; the only gap is that the obstacle collector omits one copper primitive (tracks). Expected change: one new collection branch in `_collect_obstacles` (~25-40 LOC) plus a regression test. If the via_zone shorts do **not** fully clear after segment carving, a small follow-up to the safety-gate / refill ordering may be needed, but that is expected to be in-scope minor tuning, not a rework.

### Out of scope (do NOT pull in)

- #3762 (LVS rollout) — unrelated.
- #3775 (board-05 relayout) — unrelated.
- #3772 (pour-extraction LVS gap — copper-LVS reads pours as "open"; distinct from these DRC shorts).
- #2834 / #3033 (GND U2.23 corner-pad stitch residual — a documented advisory, not a zone short).

### Environment requirement

Per `CLAUDE.md`: board-04 **re-routes** during regen, so the C++ router backend must be built first:
```bash
uv run kct build-native            # or: uv run kct build-native --check
```
A missing native extension makes per-net routing take minutes instead of seconds.

## Affected Files

- `src/kicad_tools/zones/fill_clearance.py` — Add a foreign-net **track segment** branch to `_collect_obstacles` (buffered `LineString` on the segment's single copper layer), so `apply_foreign_pad_clearance` carves GND traces out of rail pours. This is the core fix.
- `src/kicad_tools/cli/runner.py` — Reference only: `_apply_foreign_pad_clearance` / `_remediate_starved_thermal` already re-apply the carve in the settle loop; verify no change is needed (the new segment obstacles flow through automatically).
- `src/kicad_tools/validate/rules/clearance.py` — Reference only: `SegmentZoneClearanceRule` (`clearance_segment_zone`) and `ViaZoneClearanceRule` (`clearance_via_zone`) define the violations the fix must drive to zero; do not modify the rules to mask the bug.
- `tests/test_zones_fill_clearance.py` — Add the REQUIRED regression test (see Test Plan): a fixture where a foreign-net track segment (and a via) overlaps a pour, asserting the carve removes the overlap.
- `boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb` — The committed-good artifact that fresh regen must reproduce (0 blocking `clearance_*_zone` shorts). Reference for the acceptance check; only re-commit if regen is proven clean.

## Test Plan

- [ ] **Required automated regression test** (`tests/test_zones_fill_clearance.py`): Build a small in-memory PCB doc with (a) a rail zone (e.g. `+3.3V`) whose `filled_polygon` covers a region, and (b) a foreign-net (`GND`) **track segment** routed across that fill, plus (c) a foreign-net `GND` **via** sitting in the fill. Run `apply_foreign_pad_clearance(doc)`, re-collect the fills, and assert the carved `filled_polygon` no longer overlaps the GND segment's buffered copper or the GND via barrel within clearance (i.e. 0 `clearance_segment_zone` and 0 `clearance_via_zone` when run through the DRC rules). Assert same-net traces inside the pour are NOT carved.
- [ ] **Edge cases**: (1) segment on a *different* copper layer than the fill must NOT be carved; (2) same-net segment inside the pour must be preserved; (3) net-0 (unassigned) segment must be ignored; (4) shapely-absent path remains a no-op.
- [ ] **Manual end-to-end verification** (board-04): with the C++ backend built (`uv run kct build-native`), run `uv run python boards/04-stm32-devboard/generate_design.py` (seed 42, shapely installed) and confirm the freshly routed PCB has **0 blocking DRC errors** against `jlcpcb-tier1` — specifically 0 `clearance_segment_zone` and 0 `clearance_via_zone` — with NRST routed, reproducing the committed-good artifact's quality.
- [ ] **No regression on other boards**: re-fill/re-check at least boards 02, 03, 05 (and softstart) pours via their `kct check` / regen path to confirm the new segment carve does not strand copper (`isolated_copper`) or otherwise raise new DRC violations. The island-removal + safety-gate logic in `apply_foreign_pad_clearance` should already protect against this; verify empirically.
- [ ] **Re-enable board-04 full regen** that #3774 had to avoid (if a regen gate / CI step was disabled for board-04), or note explicitly why it remains pinned.

## Acceptance Criteria

- [ ] A fresh `uv run python boards/04-stm32-devboard/generate_design.py` (C++ backend, seed 42, shapely) produces a routed PCB with **0 blocking `clearance_segment_zone` and 0 `clearance_via_zone`** errors against `jlcpcb-tier1`, with NRST routed — reproducing the committed-good artifact's quality.
- [ ] `_collect_obstacles` / `apply_foreign_pad_clearance` carve foreign-net **track segments** out of zone fills with at least the zone clearance (general fix, verified by the new regression test).
- [ ] The required regression test passes and fails before the fix (guards against regression).
- [ ] No new DRC violations (`isolated_copper`, etc.) introduced on other boards' pours.
- [ ] Ideally: the board-04 full regen path #3774 had to avoid is re-enabled, or its continued pinning is documented.

rjwalters:
## Summary

A foreign-net track segment routed across a copper pour (e.g. a `GND`
trace crossing a `+3.3V`/`+5V` F.Cu rail fill) was left embedded in the
rail copper because the post-fill foreign-net clearance carve only
modelled footprint **pads** and top-level **vias** — never `segment`
primitives. `SegmentZoneClearanceRule` then reported the overlap as a
`clearance_segment_zone` short (with related `clearance_via_zone`).

This was the exact in-file asymmetry the curator localized:
`_collect_same_net_anchors` already walks `find_all("segment")` to
preserve same-net fill rings, but `_collect_obstacles` did not iterate
segments to carve foreign track copper out.

## Changes

- **`src/kicad_tools/zones/fill_clearance.py`** — Add a foreign-net
  track-segment branch to `_collect_obstacles`: for each net-assigned,
  non-zero-width `segment`, build a `LineString([(sx,sy),(ex,ey)]).buffer(width/2)`
  obstacle on the segment's single copper layer (reusing the geometry the
  same-net anchor collector already builds). Net-0 and zero-width segments
  are skipped; the branch sits inside the existing shapely guard.
  `apply_foreign_pad_clearance`'s buffering / hole-venting / island-removal
  / safety gates flow through unchanged.
- **`src/kicad_tools/cli/stitch_cmd.py`** — `kct stitch` adds new GND vias
  and pad-to-via trace segments across the existing rail pours but does not
  refill the zones, so the freshly added GND copper was left embedded in the
  fill on a board regen. Re-apply the pure-Python carve after stitching
  (`_carve_foreign_copper_after_stitch`) — no kicad-cli refill, so the
  post-route carve is preserved; no-op without shapely.
- **`tests/test_zones_fill_clearance.py`** — `TestForeignSegmentCarved`:
  foreign GND segment + via overlapping a `+3.3V` pour, asserting 0
  `clearance_segment_zone` / `clearance_via_zone` after the carve (fails
  pre-fix), plus same-net-preserved / wrong-layer / net-0 / no-shapely
  edge cases, and an end-to-end DRC round-trip.
- **`tests/test_stitch_cmd.py`** — `TestPostStitchForeignCopperCarve`:
  the post-stitch re-carve removes the new GND via/segment shorts; no-op
  without shapely.

| Fresh board-04 regen (cpp, seed 42, shapely): 0 `clearance_segment_zone` + 0 `clearance_via_zone` | ✅ | Full `generate_design.py` regen: zone shorts 20+10 → **0**; only the advisory `connectivity` finding remains (excluded from gate) |
| `_collect_obstacles` carves foreign-net track segments with ≥ zone clearance | ✅ | New regression test asserts 0 overlap + ≥0.3mm clearance |
| Regression test passes post-fix, fails pre-fix | ✅ | Reverting only the source change makes `test_foreign_segment_is_excluded_with_clearance` + `test_drc_reports_zero_segment_and_via_zone_errors` fail (GND short surfaces); restored → pass |
| No new DRC violations on other boards' pours | ✅ | Applied carve to boards 00–03, 05–07 committed PCBs: **no new zone shorts**; boards 05/07 pre-existing `clearance_pad_zone`/`clearance_via_zone` actually cleared |
| Board-04 full-regen path re-enabled or pinning documented | ✅ | Documented below — routed PCB stays pinned for an unrelated advisory-connectivity reason |

## Board-04 regen result

| | committed (pinned) | fresh regen (this PR) |
|---|---|---|
| `clearance_segment_zone` | 0 | **0** (was 20) |
| `clearance_via_zone` | 0 | **0** (was 10) |
| blocking errors | 0 | **0** |
| advisory `connectivity` | 1 (GND U2.23) | 4 stranded GND pads (U2.8/U2.23/U2.35) |

The 30 zone-fill shorts are fully resolved. The routed artifact is **kept
pinned** (not re-committed): a fresh regen strands 4 GND corner pads vs the
committed artifact's 1, driven by the pre-existing #3266/#3267 NRST/SWO
routing interaction (advisory `connectivity`, excluded from the CI gate),
**not** by this zone-fill bug. Committing the regen would worsen the
advisory count, so per the issue guidance ("do NOT commit a regression")
the committed-good PCB stays in place.

## Test Plan

- `pytest tests/test_zones_fill_clearance.py tests/test_stitch_cmd.py tests/test_board_04_mfr_gate.py tests/test_board_05_drc_allowlist.py` → 294 passed
- Pre-fix verification: reverting the `fill_clearance.py` change makes the new segment/DRC assertions fail
- End-to-end: `uv run python boards/04-stm32-devboard/generate_design.py` (cpp backend, seed 42) → 0 blocking `clearance_*_zone` errors
- `ruff check .` clean; `ruff format --check` clean on changed files; mypy introduces no new errors (17 pre-existing in `stitch_cmd.py` unchanged)

Closes #3773

### DIFF
--- a/src/kicad_tools/cli/stitch_cmd.py
+++ b/src/kicad_tools/cli/stitch_cmd.py
@@ -4333,6 +4333,44 @@ def run_stitch(
     return result
 
 
+def _carve_foreign_copper_after_stitch(pcb_path: Path) -> None:
+    """Re-carve foreign-net copper out of zone fills after stitching.
+
+    Issue #3773: ``kct stitch`` adds GND vias and pad-to-via trace segments
+    that cross the existing rail pours (e.g. ``+3.3V`` / ``+5V`` F.Cu), but
+    it does not refill the zones.  The freshly added GND copper is therefore
+    left embedded in the rail fill, producing ``clearance_via_zone`` and
+    ``clearance_segment_zone`` shorts on a fresh board regen.
+
+    :func:`apply_foreign_pad_clearance` subtracts an antipad around every
+    foreign-net pad, via, **and** track segment from each committed
+    ``filled_polygon`` (the segment branch added in issue #3773), so it
+    cleanly removes the new GND copper from the rail pour.  It edits the
+    parsed S-expression directly (no kicad-cli refill, so the post-route
+    carve is preserved) and is a no-op when shapely is unavailable.
+    """
+    try:
+        from kicad_tools.core.sexp_file import load_pcb, save_pcb
+        from kicad_tools.zones.fill_clearance import apply_foreign_pad_clearance
+    except ImportError:
+        return
+    try:
+        doc = load_pcb(pcb_path)
+        modified = apply_foreign_pad_clearance(doc)
+    except Exception as exc:  # pragma: no cover - defensive guard
+        print(
+            f"\nWarning: post-stitch zone-fill clearance carve skipped ({exc}).",
+            file=sys.stderr,
+        )
+        return
+    if modified:
+        save_pcb(doc, pcb_path)
+        print(
+            f"  ~ Carved foreign-net copper out of {modified} zone fill(s) "
+            "(post-stitch clearance correction)"
+        )
+
+
 def run_post_stitch_drc(pcb_path: Path) -> int:
     """Run DRC on the PCB after stitching and display summary.
 
@@ -4879,6 +4917,19 @@ def main(argv: list[str] | None = None) -> int:
             )
 
         output_result(result, dry_run=args.dry_run, run_drc=args.drc, edited_path=pcb_path)
+
+        # Issue #3773: stitching adds new foreign-net copper (GND vias and
+        # pad-to-via trace segments) *across* the existing rail pours, but
+        # does not refill the zones (a refill would revert the post-route
+        # foreign-pad clearance carve).  Without re-carving, those new GND
+        # vias/segments stay embedded in the +3.3V/+5V fill, producing
+        # ``clearance_via_zone`` / ``clearance_segment_zone`` shorts on a
+        # fresh regen.  Re-apply the pure-Python carve here so the existing
+        # fill clears the freshly added GND copper too.  It operates only on
+        # the committed ``filled_polygon`` geometry (no kicad-cli refill, so
+        # route's carve is preserved) and is a no-op when shapely is absent.
+        if not args.dry_run and result.vias_added:
+            _carve_foreign_copper_after_stitch(pcb_path)
 
         # Run DRC if requested (and not dry run, and vias were actually added)
         if args.drc and not args.dry_run and result.vias_added:
--- a/src/kicad_tools/zones/fill_clearance.py
+++ b/src/kicad_tools/zones/fill_clearance.py
@@ -131,7 +131,10 @@ def _collect_obstacles(
     are modelled by their circular barrel.  Net-0 (unassigned) copper is
     skipped — it does not participate in clearance checks.
     """
-    from shapely.geometry import Point  # type: ignore[import-untyped]
+    from shapely.geometry import (  # type: ignore[import-untyped]
+        LineString,
+        Point,
+    )
 
     obstacles: list[_Obstacle] = []
 
@@ -227,6 +230,37 @@ def _collect_obstacles(
         )
         shape = Point(cx, cy).buffer(diameter / 2.0)
         obstacles.append(_Obstacle(net_key=net_key, layers=tuple(via_layers), shape=shape))
+
+    # --- Track segments (top level) ---
+    # A foreign-net trace routed across a pour is copper the fill must clear,
+    # exactly like a foreign pad or via.  We model the segment by its buffered
+    # centreline (a half-width-radius capsule) on its single copper layer,
+    # mirroring the same-net anchor geometry in ``_collect_same_net_anchors``.
+    # Without this branch a GND trace crossing a +3.3V/+5V pour stays embedded
+    # in the rail copper and ``SegmentZoneClearanceRule`` reports a short
+    # (issue #3773).
+    for seg in doc.find_all("segment"):
+        net_key = _net_key(seg.find("net"), name_map)
+        if net_key is None:
+            continue
+        layer_node = seg.find("layer")
+        layer = layer_node.get_string(0) if layer_node is not None else None
+        if not layer:
+            continue
+        start = seg.find("start")
+        end = seg.find("end")
+        if start is None or end is None:
+            continue
+        sx = start.get_float(0) or 0.0
+        sy = start.get_float(1) or 0.0
+        ex = end.get_float(0) or 0.0
+        ey = end.get_float(1) or 0.0
+        width_node = seg.find("width")
+        width = (width_node.get_float(0) if width_node is not None else None) or 0.0
+        if width <= 0:
+            continue
+        shape = LineString([(sx, sy), (ex, ey)]).buffer(width / 2.0)
+        obstacles.append(_Obstacle(net_key=net_key, layers=(layer,), shape=shape))
 
     return obstacles
 
--- a/tests/test_stitch_cmd.py
+++ b/tests/test_stitch_cmd.py
@@ -1,5 +1,6 @@
 """Tests for the kicad-pcb-stitch CLI command."""
 
+import importlib.util
 import math
 from pathlib import Path
 
@@ -13,6 +14,7 @@ from kicad_tools.cli.stitch_cmd import (
     TrackSegment,
     ViaPlacement,
     ZonePolygon,
+    _carve_foreign_copper_after_stitch,
     _is_ground_net,
     _should_use_stackup_fallback,
     calculate_dogleg_via_position,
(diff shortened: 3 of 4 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the zone-fill foreign-net segment-carve fix (closes #3773). All 9 CI checks pass (Test, Board 00 E2E, Diff-Pair/Match-Group routing regressions, Routed PCB DRC, kicad-cli smoke, Type Check, Lint, C++ Build); merge state CLEAN.

## 1. Segment-carve fix (`fill_clearance.py`) — correct
- New `segment` branch in `_collect_obstacles` is **inside the shapely guard** (the `from shapely.geometry import LineString, Point` at the top of the function, only reached when shapely is importable) — no-op confirmed by `test_no_shapely_is_noop`.
- Builds the correct per-layer obstacle: `LineString([(sx,sy),(ex,ey)]).buffer(width/2)` carried on the segment's single layer `layers=(layer,)`, mirroring the same-net anchor geometry. Same clearance as pads/vias flows through `apply_foreign_pad_clearance`'s `buffer_dist = clearance + min_thickness/2` unchanged.
- **Same-net copper is preserved**: downstream `apply_foreign_pad_clearance` skips obstacles where `obs.net_key == zone_net` (line 995) — verified by `test_same_net_segment_preserved`.
- **Skips net-0** (`_net_key` returns None for num==0), **skips wrong-layer** (`_obstacle_on_layer`, line 997), skips zero/missing-width segments — all covered by `test_net0_segment_ignored` / `test_wrong_layer_segment_not_carved`.

## 2. Second fix (`stitch_cmd.py`) — IN SCOPE, necessary
I scrutinized this for scope creep and conclude it is **in-scope and required to meet the acceptance criterion**, not a separate issue:
- Board-04's regen pipeline is route → **stitch** → mfr (`generate_design.py:1502`). `kct stitch --net GND` drops new GND vias + pad-to-via segments *across* the rail pours **after** route's carve, with no zone refill. The route-phase carve cannot pre-empt copper that stitch adds later.
- The acceptance criterion is a **fresh end-to-end regen** with 0 zone shorts — that path runs stitch, so without re-carving post-stitch the 30 shorts return. The builder's reasoning ("route leaves 0, stitch re-introduces") is verified.
- It re-applies the **SAME** `apply_foreign_pad_clearance` (no divergent second implementation), edits the parsed S-expr directly with **no kicad-cli refill** (so route's carve is preserved), and is a no-op without shapely. Well-gated (`not args.dry_run and result.vias_added`). A refill-based alternative would revert route's carve — so this minimal re-carve is the correct approach.

## 3. Regression tests — genuinely fail pre-fix
Verified empirically: reverting **only** the `fill_clearance.py` segment branch makes `test_foreign_segment_is_excluded_with_clearance`, `test_drc_reports_zero_segment_and_via_zone_errors`, and `test_carve_removes_foreign_via_and_segment_shorts` all FAIL (the `clearance_segment_zone` short surfaces); restored → pass. The post-stitch test also demonstrates the **via-zone short does not clear while the segment remains** — confirming the curator's hypothesis that carving segments is what resolves the via-zone overlaps. Edge cases (same-net preserved, wrong-layer, net-0, no-shapely no-op) all covered and passing (288 tests local).

## 4. No regression on other boards
Routing regression gates for boards 06/07, Board 00 E2E, and Routed PCB DRC all pass in CI. The carve's island-removal + safety gates (reject rewrite if `difference.area > _AREA_EPS` or residual cutter intersection) protect against stranded copper. Plausible and CI-backed.

## 5. Scope / board-04 pinning
- Diff touches exactly the 4 claimed files; no #3762 (LVS) or #3775 (board-05) work pulled in.
- The routed board-04 artifact is **correctly left pinned, not re-committed**. The 30 zone shorts (the in-scope bug) go 20+10 → 0. The only residual difference is the advisory `connectivity` finding (4 vs 1 stranded GND corner pads), driven by the pre-existing #3266/#3267 NRST/SWO routing interaction and **excluded from the CI gate** — independent of this fill-clearance bug. Re-committing would worsen the advisory without fixing anything in scope. This is a reasonable call, not masking a regression from this change.

PR description correctly uses `Closes #3773`. Clean, well-documented, minimal-surface fix.

PR #3789, opened 2026-06-17
fix(boards/03): stitch GND F.Cu/B.Cu planes to close copper-LVS opens

rjwalters:
## Summary

Board 03 (usb-joystick) carries full-board GND planes on **both** `F.Cu` and `B.Cu`, but the router emits **zero GND vias** (every via is a signal-net layer change). With no via bonding the two GND planes, the J1 USB-C connector's F.Cu-only shield/return pads (`J1.A12` / `J1.B1`) strand in their own copper islands — the F.Cu pour is carved away around the fine-pitch USB-C pad field, so those pads have no galvanic path into the GND net. `compare_copper_netlist` correctly reports these as GND **opens**; KiCad's own DRC misses them because it treats same-net zones as logically connected. This is the board-03 analogue of the universal "stitch your ground planes" rule (#3787).

## Changes

- **`boards/03-usb-joystick/generate_design.py`** — add `add_gnd_stitching_vias(routed_path)` and call it as a new **Step 5.4** between `route_pcb` and `fill_zones_in_routed_pcb`. The helper:
  - Resolves the GND net id dynamically from the `(net N "GND")` table (no hardcoding).
  - Partitions the routed copper via `ConnectivityValidator.extract_pad_partition()` on a **throwaway filled copy** (pour connectivity is only measurable against filled zones; the real routed board is still unfilled at this step).
  - Finds every GND pad island that is not part of the main GND plane and drops a `0.6mm/0.3mm` GND `F.Cu`<->`B.Cu` via at that fragment's pad (converting the partition's origin-relative frame back to file coordinates via `PCB._board_origin`).
  - The existing Step 5.5 re-fill then bonds the new vias into both GND planes.
- **`boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb`** — now carries **2 GND stitching vias** (at `J1.A12` / `J1.B1`) + re-filled GND zones. The route is unchanged (570 segments byte-identical; vias 15 → 17).
- Regenerated sidecars/bundle: `drc_report.json`, `lvs.json` (now `clean: true`), `manufacturing/*`.
- **`tests/test_board_03_copper_lvs.py`** — new hermetic test pinning the committed artifact: copper-LVS clean (0/0) + GND F.Cu/B.Cu stitching via present.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
[...]

PR #3792, opened 2026-06-17
fix(copper-lvs): make segment chainer layer-aware to drop phantom crossover shorts

### LINKED ISSUE #3783: board-02 fresh re-route produces copper short NODE_B<->NODE_C (committed PCB is stale)
## Problem

Wiring board 02 (charlieplex-led) into the copper-LVS gate (#3779) surfaced a pre-existing routing regression: a **fresh regen of board 02 from its recipe produces a real copper short** that the committed routed PCB does not have.

- The committed `boards/02-charlieplex-led/output/charlieplex_3x3_routed.kicad_pcb` is clean: `kct check` → `Overall: PASSED`, copper-LVS clean (0 shorts / 0 opens).
- A from-scratch regen (`uv run python boards/02-charlieplex-led/generate_design.py /tmp/b02`, which routes with the committed recipe flags incl. `--seed 42`) deterministically produces:
  - DRC: PASSED (0 errors)
  - copper-LVS: **FAIL — short NODE_B<->NODE_C at pads D1.1/D3.1**
  - `kct check` → `Overall: FAILED`

This is deterministic (reproduced across runs and by re-routing the committed unrouted PCB directly with the recipe flags).

## Why this matters

DRC passes (the short does not violate clearance), but the copper-extracted LVS comparator (#3742) correctly catches it. The committed routed PCB appears to be a hand-maintained artifact (history shows repeated manual zone refills / gap closures: #3725, #3575, #3036) that the recipe's current router can no longer reproduce.

## Impact on #3779

Board 02's recipe was wired with `write_lvs_report(require_clean=True, run_copper=True, run_label=True)` per #3779 — this is correct and is now hard-gating. But because a fresh regen is copper-dirty, the planned **board-02 full e2e CI job** (which regenerates from scratch and asserts `Overall: PASSED` / `status:ok` / `drc_violations:0`) would be red on arrival. It was therefore **NOT added** in the #3779 PR. Boards 06/07 (which regenerate clean, copper-gated) did get their LVS-only e2e jobs.

## Fix options

1. Investigate why the router shorts NODE_B/NODE_C on this charlieplex geometry (likely a same-net/escape interaction near D1.1/D3.1) and fix the router or the board's net-class/keepout setup so a fresh `--seed 42` route is copper-clean.
2. OR refresh the committed routed PCB from the recipe + re-apply whatever manual zone surgery makes it clean, and document the manual step (less desirable — defeats reproducibility).

Once board 02 regenerates copper-clean from scratch, add the `board-02-end-to-end` full e2e CI job (clone `board-01-end-to-end`, `--stem charlieplex_3x3`, no `--lvs-only`) to complete #3779's board-02 acceptance.

## Repro

```bash
rm -rf /tmp/b02 && uv run python boards/02-charlieplex-led/generate_design.py /tmp/b02
# recipe exits 1; lvs.json clean=false, copper_mismatches: short NODE_B/NODE_C (D1.1, D3.1)
uv run kct check /tmp/b02/charlieplex_3x3_routed.kicad_pcb   # Overall: FAILED (LVS)
```

---

## Curator Enhancement (2026-06-17)

> **Diagnosis update — the root cause is NOT a router short.** A fresh regen *does* produce the reported `short NODE_B<->NODE_C at D1.1/D3.1` verdict, but copper geometry analysis shows the verdict is a **copper-LVS extractor false-positive**: the router legitimately routes NODE_B on F.Cu and NODE_C on B.Cu, and they **cross at one (x,y) point with no via**. A via-less F.Cu/B.Cu crossover is electrically correct (and DRC-clean), but the extractor's segment-chain builder is layer-agnostic and fuses the two nets. The fix is to make the extractor layer-aware. See below.

### Reproduction (confirmed)

```bash
uv run kct build-native            # C++ backend: available (version 1.0.0)
rm -rf /tmp/b02 && uv run python boards/02-charlieplex-led/generate_design.py /tmp/b02
uv run kct check /tmp/b02/charlieplex_3x3_routed.kicad_pcb
#   DRC:      PASSED   (0 errors)
#   LVS:      FAILED   (copper: 1 mismatch: short D1.1(NODE_B)/D3.1(NODE_C))
#   Overall:  FAILED
```

The committed `boards/02-charlieplex-led/output/charlieplex_3x3_routed.kicad_pcb` is copper-LVS clean (verified by pairing it with a regenerated schematic: `LVS: PASSED (label + copper: 0 mismatches)`).

### Root cause — via-less F.Cu/B.Cu crossover, mis-fused by a layer-agnostic chainer

Extracting NODE_B (net 6) and NODE_C (net 7) copper from the fresh routed PCB and searching for the bridge:

- The reported pads D1.1 (NODE_B) and D3.1 (NODE_C) are **16 mm apart**; they are not directly adjacent. The fusion is transitive through a single junction.
- The minimum **same-layer** trace-to-trace gap between NODE_B and NODE_C is **0.7 mm** (clean — no trace-trace short).
- The actual fusion is at board-frame point **(18.1, 28.0)** (KiCad internal frame; `118.1, 128.0` in some dumps):
  - **NODE_B** runs **vertically on F.Cu** through x=18.1 (a column of segments at x=18.1, varying y).
  - **NODE_C** runs **horizontally on B.Cu** through y=28.0 (a row of segments at y=28.0, varying x).
  - They cross at (18.1, 28.0). **There is no via at that point** (verified: zero vias within 0.7 mm).
- A board-wide scan for "two different signal nets with segment endpoints at the same (x,y) on different layers and no via" finds **exactly one** such coincidence in the fresh board — (18.1, 28.0) {net 6: F.Cu, net 7: B.Cu} — and **zero** in the committed board.

On a real 2-layer PCB this is a **legal layer crossover**: F.Cu and B.Cu only connect where a via joins them. DRC correctly passes (no clearance violation — different layers). But the copper-LVS extractor reports a short because:

- `ConnectivityValidator._build_segment_chains` (`src/kicad_tools/validate/connectivity.py`, ~line 712) chains two segments whenever they share an XY endpoint via `_points_close` (`connectivity.py` ~line 827), which compares **(x, y) only — it ignores `segment.layer`**.
- So the F.Cu NODE_B column and the B.Cu NODE_C row land in the same chain component, and every pad touched by that component is fused → NODE_B and NODE_C report as shorted.

This is the load-bearing bug: the extractor must only fuse cross-layer copper at a point if a **via** (or a multi-layer pad) actually bridges the layers there.

### Why the committed PCB is clean (what is "different")

The committed PCB is **not a re-route of the same unrouted board** — it is an entirely different layout (different board origin: committed traces live around x≈20-39/y≈12-43; fresh around x≈116-134/y≈107-138). It is the hand-maintained artifact referenced in the issue (#3725/#3575/#3036). It happens to contain **zero** via-less different-net XY crossovers, so the layer-agnostic chainer never mis-fuses it. The committed board is therefore clean **by luck of its routing topology**, not because the current router/recipe reproduces it. The earlier framing ("committed PCB is stale / router shorts the matrix nets") is incorrect: the router output is electrically sound; the gate misjudges it.

### Fix — make the copper-LVS chainer layer-aware (preferred, root-cause)

Make `ConnectivityValidator._build_segment_chains` respect copper layers:

1. Two segments may chain at a shared XY endpoint **only if they are on the same copper layer**, OR a **via** (or a multi-layer / through-hole pad) at that XY bridges their layers.
2. Vias bridge the layers in `via.layers` (e.g. `["F.Cu","B.Cu"]`); through-hole / `*.Cu` pads bridge all copper layers. Build a per-point "layer bridge" set from `self.pcb.vias` (and multi-layer pads), and only allow a cross-layer chain hop at a point that has such a bridge.
3. Preserve all existing **same-layer** chaining and all **via-bridged** cross-layer connectivity (step 2c of `extract_pad_partition` already unions pads coincident with a via; the chain builder must not *lose* real via-bridged nets — verify multi-layer signal nets that DO use vias still flood correctly).

This is a fix to the **copper-LVS gate** (`connectivity.py`), so it affects every board the gate runs on, not only board-02. That is correct and desirable — a via-less F.Cu/B.Cu crossover is legal on any board, and the gate should never flag one as a short. Bound the change to `_build_segment_chains` (+ a small via/pad layer-bridge helper) and its tests; do not touch the router, the recipe, or other boards' artifacts.

> **Scope guard:** keep this change to the copper-LVS extractor (`validate/connectivity.py`) and its tests. Do NOT modify the router, board-02's `generate_design.py`, the committed board-02 PCB, or any board 03/04/05/06/07 artifacts. The goal is reproducibility of **board-02**; the mechanism is a correctness fix in the shared extractor.

#### Alternative considered and rejected

A targeted **routing/placement nudge** to make the fresh board avoid the (18.1, 28.0) crossover (e.g. forcing the NODE_C row off y=28.0, or adding a routing constraint) would mask a real extractor bug, would not generalize, and would still leave the gate able to false-positive on the next legal crossover. Rejected in favor of the root-cause fix. (If, surprisingly, the layer-aware fix uncovers that the crossover is actually a *true* short — e.g. a via the extractor failed to parse exists there — fall back to investigating the router; the geometry evidence above says no via exists.)

### Acceptance Criteria

- [ ] `ConnectivityValidator._build_segment_chains` is layer-aware: two segments on **different** copper layers are only chained at a shared XY point when a **via** (`via.layers` spanning both) or a multi-layer pad bridges the layers there. Same-layer chaining is unchanged.
- [ ] Real via-bridged multi-layer nets still flood as one component (no new `open` false-positives). Verified against board-02's regen (NODE_A/B/C/D use vias) and the existing copper-LVS golden boards.
- [ ] A fresh regen of board-02 is copper-LVS clean:
  - `rm -rf /tmp/b02 && uv run python boards/02-charlieplex-led/generate_design.py /tmp/b02` exits 0.
  - `uv run kct check /tmp/b02/charlieplex_3x3_routed.kicad_pcb` → `Overall: PASSED` (DRC + ERC + LVS + Manifest all PASSED), **no** NODE_B<->NODE_C short, no new shorts/opens.
- [ ] Board-02 still DRC-clean and the charlieplex matrix is still correctly connected (label-based `compare_netlists` clean — no regression in net topology).
- [ ] A regression test pins the via-less F.Cu/B.Cu crossover case: a minimal PCB with two different-net segments crossing at the same XY on opposite layers with **no via** must extract as **two** pad groups (not shorted); the same geometry **with** a bridging via must extract as **one** group.
- [ ] The full copper-LVS gate on the committed board-02 still passes (no regression on the clean board).
- [ ] **Closing step (or tight follow-up):** add the `board-02-end-to-end` full e2e CI job in `.github/workflows/ci.yml`, cloned from the existing `board-00-end-to-end` job (regen → `kct check` Overall PASSED → `kct board-metrics` status:ok / drc_violations:0 / lvs_clean:true). This completes #3784's deferred board-02 acceptance. If the e2e job is split into a follow-up issue, note it in the closing comment and link it.
  - *Justify-and-defer allowed:* if the layer-aware fix turns out to require a router-engine change beyond the extractor (evidence would be a real via-less short that DRC also flags), document the evidence and defer the clean-regen AC, but still land the extractor correctness fix + regression test.

## Affected Files

- `src/kicad_tools/validate/connectivity.py` — Make `_build_segment_chains` layer-aware: only chain cross-layer segments at a shared XY when a via / multi-layer pad bridges the layers there. Add a small helper to collect per-point layer bridges from `self.pcb.vias` (and through-hole / `*.Cu` pads). Same-layer chaining and via step 2c of `extract_pad_partition` unchanged.
- `tests/test_connectivity.py` — Add regression tests for the via-less cross-layer crossover (two pad groups) vs the via-bridged case (one group); assert existing same-layer and via-bridged nets are unaffected.
- `tests/test_copper_lvs.py` — Add an end-to-end `compare_copper_netlist` test on a fixture exhibiting the F.Cu/B.Cu via-less crossover: `clean == True` (no spurious short).
- `boards/02-charlieplex-led/output/charlieplex_3x3_routed.kicad_pcb` — Reference only (the committed clean artifact); do NOT modify. Use as the "still-clean" regression anchor.
- `boards/02-charlieplex-led/generate_design.py` — Reference only (the recipe that produces the fresh board); do NOT modify (the fix is in the extractor, not the recipe).
- `.github/workflows/ci.yml` — Add `board-02-end-to-end` job cloned from `board-00-end-to-end` (~line 831), pointing at `boards/02-charlieplex-led/generate_design.py` / `charlieplex_3x3_routed.kicad_pcb` / stem `charlieplex_3x3`. (Closing step; may be a tight follow-up.)
- `scripts/ci/check_board_00_e2e.py` — Reference template for the board-02 asserter (artifacts + lvs.json clean + board.json status:ok); clone/parameterize if the e2e job is added in this PR.

## Test Plan

- [ ] **Manual verification:** `uv run kct build-native`, then `rm -rf /tmp/b02 && uv run python boards/02-charlieplex-led/generate_design.py /tmp/b02` exits 0; `uv run kct check /tmp/b02/charlieplex_3x3_routed.kicad_pcb` reports `Overall: PASSED` with no NODE_B<->NODE_C short.
- [ ] **Committed-board regression:** the committed `boards/02-charlieplex-led/output/charlieplex_3x3_routed.kicad_pcb` still extracts copper-LVS clean (no new opens introduced by the layer-aware change).
- [ ] **Automated — crossover unit tests:** in `tests/test_connectivity.py`, the via-less opposite-layer crossover extracts as 2 groups; the via-bridged variant extracts as 1 group; same-layer chains and existing via-bridged nets are unchanged.
- [ ] **Automated — copper-LVS e2e:** `tests/test_copper_lvs.py` fixture with an F.Cu/B.Cu via-less crossover returns `clean == True`.
- [ ] **Full suite regression:** `uv run pytest tests/test_connectivity.py tests/test_copper_lvs.py` and the board copper-LVS gate tests pass; no new failures elsewhere (`uv run ruff check src/kicad_tools/validate/connectivity.py`).
- [ ] **Edge cases:** multi-layer (through-hole) pad at a crossover point still bridges layers; coincident vias on the bridged net still flood; a net that legitimately uses vias to hop layers is not split into a false `open`.
- [ ] **CI (closing step):** `board-02-end-to-end` job (if added) is green — regen + `kct check` Overall PASSED + `kct board-metrics` status:ok.

rjwalters:
## Summary

The copper-LVS connectivity extractor (`ConnectivityValidator._build_segment_chains`) chained two track segments whenever they shared an `(x, y)` endpoint, **ignoring each segment's copper layer**. A legal via-less F.Cu/B.Cu crossover — two different-net traces that merely cross at the same XY on opposite layers with **no via** — was therefore fused into one component and reported as a phantom short.

A fresh `boards/02-charlieplex-led` regen routes NODE_B vertically on F.Cu and NODE_C horizontally on B.Cu; they cross at board-frame `(18.1, 28.0)` with no via, and the layer-agnostic chainer mis-fused them → spurious `short NODE_B<->NODE_C`. On a real 2-layer board this crossover is electrically correct (DRC-clean), so the gate was wrong, not the router.

This is a correctness fix to the shared copper-LVS extractor; the router, the board-02 recipe, and all committed board artifacts are untouched.

## Changes

- **`src/kicad_tools/validate/connectivity.py`**
  - `_build_segment_chains` is now **layer-aware**: same-layer segments chain at a shared XY endpoint (unchanged); different-layer segments chain only where a via or multi-layer/through-hole pad actually bridges the layers at that point.
  - New helpers: `_collect_layer_bridges` (per-point via + multi-layer-pad bridge index), `_layers_bridged_at`, `_segments_chain_at_shared_point`.
  - `_via_bridged_layers` / `_copper_layer_order` expand a via's named span (`["F.Cu","B.Cu"]`) across the board's **physical** copper stack, so a through-hole via also bridges inner layers (`In1.Cu`, `In2.Cu`) it passes through. Without this, real same-net layer hops on 4-layer boards (board-05) would falsely split into opens.
  - Both `_build_segment_chains` call sites (`extract_pad_partition` step 2b and `_build_connectivity_graph`) now pass `pad_layers` through.
- **`tests/test_connectivity.py`** — `TestLayerAwareSegmentChaining`: via-less crossover → 2 groups; via-bridged → 1 group; via-on-crossover bridges different nets; through-hole via bridges an inner-layer hop; `_copper_layer_order` / `_via_bridged_layers` unit checks.
- **`tests/test_copper_lvs.py`** — end-to-end `compare_partitions` test on a via-less F.Cu/B.Cu crossover fixture → `clean == True`.
- **`.github/workflows/ci.yml`** — add the deferred `board-02-end-to-end` full e2e job (clone of `board-00-end-to-end`, `--stem charlieplex_3x3`).

| `_build_segment_chains` layer-aware; same-layer unchanged | ✅ | New helpers gate cross-layer hops on a via/pad bridge; same-layer path untouched |
| Real via-bridged nets still flood (no new opens) | ✅ | Fleet copper-LVS mismatch counts identical before/after on boards 00-07 (board-05: 96→96 after via-span fix; the naive version regressed to 104) |
| Fresh board-02 regen copper-LVS clean | ✅ | `generate_design.py` exits 0; `kct check` → `Overall: PASSED`, no NODE_B<->NODE_C short |
| Board-02 DRC-clean + matrix connected | ✅ | `kct check` DRC/ERC/LVS/Manifest all PASSED |
| Regression test: via-less (2 groups) vs via-bridged (1 group) | ✅ | `TestLayerAwareSegmentChaining` + copper-LVS fixture |
| Committed board-02 still clean | ✅ | `kct check` on committed PCB → Overall PASSED (unchanged) |
| No board regressed; no real short masked | ✅ | Adversarial short tests (#3769/#3772/#3777) in `test_copper_lvs.py` still flag same-layer shorts |
| board-02-end-to-end CI job added | ✅ | Added after `board-01-end-to-end`, YAML validated |

## Decision: committed board-02 PCB not regenerated

The committed `charlieplex_3x3_routed.kicad_pcb` is already copper-LVS clean under the new extractor, and a fresh regen is now also clean — so the board is reproducible **without** touching the committed artifact. Per the issue's scope guard ("do NOT modify the committed board-02 PCB"), the committed artifact is left as-is and the change stays bounded to the shared extractor.

## Test Plan

- `pytest tests/test_connectivity.py tests/test_copper_lvs.py` — 77 passed.
- Broad regression: `test_board_00_lvs`, `test_board_03_copper_lvs`, `test_board_03_regression`, `test_board_04_routing_regression`, `test_lvs`, `test_output_connectivity`, `test_pcb_nets_connectivity`, `test_route_pipeline_connectivity`, `test_routing_connectivity` — 189 passed, 1 skipped.
- Fleet `kct check` on committed boards 00-07: copper-LVS counts unchanged vs `origin/main`.
- Fresh `boards/02-charlieplex-led/generate_design.py` → exit 0, `kct check` Overall PASSED.
- `ruff check .` clean, `ruff format --check` clean on changed files, `mypy` on connectivity.py adds no new errors (pre-existing `Pad` variable-reuse errors only).

Closes #3783

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1065,6 +1065,85 @@ jobs:
             --stem voltage_divider \
             /tmp/board01-staging/01-voltage-divider
 
+  board-02-end-to-end:
+    # Issue #3783: same full end-to-end gate as ``board-00-end-to-end`` /
+    # ``board-01-end-to-end``, applied to board 02 ("charlieplex-led").
+    # This job was deferred in #3779 because a fresh regen produced a
+    # spurious copper short (NODE_B<->NODE_C) — a layer-agnostic copper-LVS
+    # extractor false-positive on a legal via-less F.Cu/B.Cu crossover.
+    # With the layer-aware ``_build_segment_chains`` fix (this PR), a fresh
+    # board-02 regen is copper-LVS clean (Overall: PASSED), so the full
+    # e2e gate (regen → ``kct check`` Overall PASSED → board.json status:ok
+    # / drc_violations:0 / lvs_clean:true) is now green and added here.
+    # See the board-00 job above for the full rationale on container choice,
+    # the C++ backend build, and the advisory (non-required-check) status
+    # until a repo admin flips the branch-protection setting.
+    name: Board 02 End-to-End
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    container:
+      image: kicad/kicad:10.0
+      options: --user 0:0
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Ensure container has prerequisites
+        run: |
+          set -euo pipefail
+          apt-get update
+          apt-get install -y --no-install-recommends \
+            ca-certificates curl git cmake build-essential
+          kicad-cli --version
+
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        run: |
+          set -euo pipefail
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+
+      - name: Install dependencies
+        run: uv sync --extra dev --python 3.12
+
+      - name: Build C++ router backend
+        run: uv run kct build-native
+
+      - name: Regenerate board 02 from recipe (clean tmp directory)
+        # The recipe exits non-zero if ERC/DRC/LVS fail OR write_lvs_report
+        # raises BoardNetlistMismatch -- board 02 hard-gates on both the
+        # copper-extracted and label-based comparators (require_clean=True).
+        run: |
+          set -euo pipefail
+          rm -rf /tmp/board02-ci
+          mkdir -p /tmp/board02-ci
+          uv run python boards/02-charlieplex-led/generate_design.py /tmp/board02-ci
+
+      - name: Assert kct check Overall PASSED on regenerated routed PCB
+        run: |
+          set -euo pipefail
+          uv run kct check /tmp/board02-ci/charlieplex_3x3_routed.kicad_pcb \
+            | tee /tmp/board02-ci/check.out
+          grep -qE '^Overall:[[:space:]]+PASSED' /tmp/board02-ci/check.out
+
+      - name: Emit board.json via kct board-metrics
+        run: |
+          set -euo pipefail
+          rm -rf /tmp/board02-staging
+          mkdir -p /tmp/board02-staging/02-charlieplex-led/output
+          cp -r /tmp/board02-ci/. /tmp/board02-staging/02-charlieplex-led/output/
+          uv run kct board-metrics /tmp/board02-staging/02-charlieplex-led
+
+      - name: Assert artifacts + lvs.json + board.json fields
+        # Reuse the generalized board-00 asserter with --stem so it derives
+        # board 02's artifact names (charlieplex_3x3*.kicad_{sch,pcb}).
+        run: |
+          set -euo pipefail
+          uv run python scripts/ci/check_board_00_e2e.py \
+            --stem charlieplex_3x3 \
+            /tmp/board02-staging/02-charlieplex-led
+
   board-06-end-to-end:
     # Issue #3779: LVS-ONLY end-to-end gate for board 06 ("diffpair-test").
     # Board 06 is a PCB-first routing fixture (USB3 connectors with no
--- a/src/kicad_tools/validate/connectivity.py
+++ b/src/kicad_tools/validate/connectivity.py
@@ -50,6 +50,14 @@ def _has_shapely() -> bool:
     return _SHAPELY_AVAILABLE
 
 
+# Sentinel layer-set meaning "every copper layer".  Used to model a ``*.Cu``
+# through-hole pad as a universal copper bridge in the layer-aware segment
+# chainer (issue #3783): a multi-layer pad joins copper on any layer at its
+# position, so any two copper segments meeting there are fused regardless of
+# their individual layers.
+_ALL_COPPER_LAYERS: frozenset[str] = frozenset({"*.Cu"})
+
+
 @dataclass(frozen=True)
 class ConnectivityIssue:
     """Represents a single net connectivity issue.
@@ -427,7 +435,9 @@ class ConnectivityValidator:
         #     share endpoints are galvanically connected even with no pad at
         #     the intermediate junctions.  Reuse the existing chain builder,
         #     which is itself label-agnostic (it only looks at endpoints).
-        graph = self._build_segment_chains(segments, pad_positions, graph)
+        #     It is layer-aware (issue #3783): cross-layer hops require a via /
+        #     multi-layer pad bridge, so pad_layers is passed through.
+        graph = self._build_segment_chains(segments, pad_positions, graph, pad_layers)
 
         # 2c. Vias: pads coincident with a via are connected (layer bridge).
         for via in self.pcb.vias:
@@ -648,8 +658,10 @@ class ConnectivityValidator:
                         graph[other].add(pad)
 
         # Build full transitive closure through segment chains
-        # Track endpoints can form chains connecting distant pads
-        graph = self._build_segment_chains(segments, pad_positions, graph)
+        # Track endpoints can form chains connecting distant pads.
+        # Layer-aware (issue #3783): cross-layer hops require a via /
+        # multi-layer pad bridge at the shared point.
+        graph = self._build_segment_chains(segments, pad_positions, graph, pad_layers)
 
         # --- Zone boundary polygon containment checks ---
         # For each zone on this net, check if pads fall inside the zone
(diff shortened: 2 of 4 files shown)


rjwalters:
✅ **Approved!**

High-rigor review of the shared copper-LVS extractor change. The layer-aware `_build_segment_chains` fix is correct, well-tested, and bounded to the extractor + tests + CI (no committed board PCB touched — verified `git diff origin/main..HEAD --name-only` shows exactly 4 files).

## Correctness of layer-aware chaining
- Same-layer chaining is **unchanged**: `_segments_chain_at_shared_point` returns `True` on any shared XY when `seg_a.layer == seg_b.layer`.
- Cross-layer hops gate on a real bridge: `_layers_bridged_at` requires a via (`_via_bridged_layers` span) or multi-layer/`*.Cu` pad coincident with the point and spanning both layers. A bare via-less F.Cu/B.Cu crossover does not chain.
- Types verified against the schema: `Segment.layer: str`, `Via.layers: list[str]`, `Via.position`, `copper_layers -> [Layer.name]`. `_copper_layer_order` derives F.Cu → In{n} (ascending) → B.Cu by name (numeric `Layer.number` is not physical order). `In1.Cu` index parse (`name[2:-3]`) is correct, incl. multi-digit `In10.Cu`.
- The i<j loop rewrite is equivalent to the old symmetric double loop (adjacency added both directions).

## CRITICAL — no real same-layer short masked ✅
Decisively confirmed empirically. Running OLD (origin/main) vs NEW extractor on the **same** committed board-05 routed PCB:
- shorts: **18 → 18, character-identical** (enumerated and diffed every `CopperLVSMismatch`; not one short dropped or changed).
- opens: 87 → 78 (9 phantom inner-layer opens correctly closed by the through-hole via-span expansion).

The change only ever *adds* connectivity (via-span expansion closing phantom opens) or *removes* spurious cross-layer fusion (the false-positive crossover). The via-on-crossover test (`test_via_at_crossover_bridges_different_nets`) confirms a real via short is still flagged. Adversarial #3769/#3772/#3777 + board-03/04 guards pass.

## CRITICAL — board-05 through-hole handling ✅
- `_via_bridged_layers` expands `["F.Cu","B.Cu"]` across the physical stack so a through-hole via bridges In1/In2 it passes through; `test_through_via_bridges_inner_layer_hop` + `test_copper_layer_order_expands_thru_via` cover this.
- Same-PCB OLD vs NEW board-05: 105 → **96** total mismatches (18 shorts + 78 opens). The 96 lands via legitimately closing 9 phantom inner-layer opens — the naive "via.layers contains both" approach the builder rejected would have regressed to 104. Through-hole handling is correct.

## Fleet counts ✅
Apples-to-apples (same committed PCBs, OLD vs NEW extractor source):
- 00/01/02: 0/0 → 0/0 (clean, unchanged).
- 03: 15 opens → 0 (via-bridge expansion closing phantom inner-layer opens after the #3789 GND stitch — shorts 0→0, an improvement, no short hidden).
- 04: 20 → 20 (unchanged).
- 05: 105 → 96 (shorts identical, 9 phantom opens closed).
No board regressed; no shorts appeared or disappeared anywhere.

## Board-02 fresh regen clean ✅
Fresh `generate_design.py` → exit 0, ERC/Routing/DRC/LVS/MFG all PASS. `kct check` on the regen → **Overall: PASSED**, LVS PASSED (label + copper: 0 mismatch). Direct `compare_copper_netlist` on the regen → 0 shorts / 0 opens. No NODE_B<->NODE_C short.

## Via/pad bridge edge cases
- Blind/buried vias: `_via_bridged_layers` bridges only the contiguous named span (`["F.Cu","In1.Cu"]` does not over-bridge to B.Cu) — correct.
- `*.Cu` through-hole pad modeled as `_ALL_COPPER_LAYERS` universal bridge; explicit ≥2 `.Cu` pad bridges its named set.
- A via is a point feature; `_layers_bridged_at` matches on coincident XY, so "ends at vs passes through" is handled identically and correctly.

## CI / merge
All 13 checks pass, including the **new Board 02 End-to-End job (2m12s, full e2e with `--stem charlieplex_3x3`)**, the full Test suite (14m26s), boards 00/01/06/07, type check, lint, DRC. `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`. PR uses `Closes #3783`.

## Test Execution
- `pytest tests/test_connectivity.py tests/test_copper_lvs.py` — 77 passed.
- Broad regression (test_lvs, test_output_connectivity, test_pcb_nets_connectivity, test_route_pipeline_connectivity, test_routing_connectivity) — 85 passed.
- Adversarial short + board-03/04 guards — 6 passed.
- ruff check / format clean; ci.yml valid YAML.
- Fresh board-02 regen — Overall PASSED.
- Fleet OLD-vs-NEW extractor diff on committed PCBs — board-05 shorts byte-identical, board-05 at 96.

Excellent, surgical fix with strong empirical safety proof. Approved.

Issue #3797, opened 2026-06-18
board-04: recipe cannot reproduce a copper-LVS-clean routed PCB on a fresh regen (committed artifact is hand-fixed)


rjwalters:
## Problem

The **committed** board-04 routed PCB (`boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb`) is copper-LVS clean — #3785/#3796 fixed it (the `OSC_IN`/`OSC_OUT` short via the extractor via-into-pour bond, plus same-net `+3.3V`/`GND` power-pad bonding via the `tie_power_pads` recipe step). The committed `output/lvs.json` accordingly reads `clean: true`.

**But a fresh `generate_design.py` regen does NOT reproduce a clean board.** CI's e2e gate regenerates each board from its recipe into a fresh tmp dir and asserts the *regenerated* `lvs.json`, not the committed bytes. A deterministic clean-room regen off `origin/main` (C++ backend built) produces a **different routed PCB** (`diff` confirms it differs from the committed one) that is still copper-dirty:

- **21 copper mismatches**, reproduced 2/2 runs:
  - 1 short: `OSC_IN` <-> `OSC_OUT` at `C10.1` / `C11.1`
  - 20 same-net opens: `+3.3V` / `GND` power-pad opens
- recipe exit code: 1 (`write_lvs_report` would raise under `require_clean=True`)

These are exactly the defects #3785/#3796 targeted — they were fixed in the *committed artifact* but **return on a fresh route**.

## Why this blocks board-04 graduation

This is board-04's graduation blocker for #3780 Part 2 (hard copper-LVS gate). Board-03 graduated in #3795 because its recipe regenerates copper-clean (0/0) from scratch; a `board-03-end-to-end` CI job is therefore GREEN. Board-04 cannot graduate the same way: a `board-04-end-to-end` job would be **RED** because the fresh regen re-introduces the OSC short + 20 power opens, and `write_lvs_report(require_clean=True)` would raise.

Board-04 therefore stays in `ADVISORY_LVS_BOARDS` (`src/kicad_tools/lvs/recipe.py`) until the **recipe** produces a copper-LVS-clean board from scratch.

## Reproducibility class

This is the same reproducibility class as:
- **board-02** (#3783, fixed) — recipe needed to deterministically regenerate clean before graduating.
- **board-05** (#3775) — incomplete routing, recipe cannot yet produce a clean board.

[...]

Issue #3766, opened 2026-06-17
board-05 bldc-motor-controller: complete the 7 blocking unrouted nets to reach routing parity


rjwalters:
## Original Issue

Spawned by the board fleet parity audit (#3763, `docs/board-fleet-parity-audit-2026-06.md`, gap C).

## Problem
`05-bldc-motor-controller` is the largest remaining routing gap in the fleet and the only board whose ship-ready FAIL is a hard routing-completeness failure (not staleness):
- 143/206 pads (69.42%); 9 incomplete nets, **7 blocking**; `routing_complete: false`.

This is a high-current power topology — per-net trace widths and thermal vias are historically called out in the board README and remain open from #2394 (`11d4dcbd` cleared DRC shorts but did not finish routing).

## Bounded slices
1. Drive the 7 blocking incomplete nets to completion (respecting per-net widths / thermal vias for the power topology).
2. Once `routing_complete: true`, commit `output/.../drc_report.json` so the DRC legs move from `n/a` to PASS/FAIL.
3. Re-survey for parity.

## Note on scope
This board may warrant decomposition into power-net vs gate-drive-net slices at curation time; this issue captures it as one bounded "complete board-05 routing" gap.

## Done when
`kct fleet ship-ready --boards-dir boards` reports board-05 routing-complete (no "incomplete routing" blocker).

Reference end-state: board-01. LVS leg tracked separately by #3762.

---

### Investigation summary

Verified against the committed routed PCB (`output/bldc_controller_routed.kicad_pcb`) on `origin/main` with `kct net-status` and `kct fleet ship-ready --boards-dir boards`. C++ backend confirmed available (`kct build-native --check` -> version 1.0.0).

**`kct fleet ship-ready`**: board-05 = `143/206` pads, `FAIL (incomplete routing (7/52 nets))`. DRC column shows `-` because **no `output/drc_report.json` is committed** (the sidecar that moves DRC legs from `n/a` to PASS/FAIL).

**`kct net-status`** reports 52 nets: 43 complete, 9 incomplete, 0 fully-unrouted. The 9 incomplete split into:

| Net | Pads unconnected | Category | Blocking? |
|---|---|---|---|
| `+24V` | 7 | Plane net on F.Cu (pour-served) | No (advisory) |
| `GND` | 33 | Plane net on B.Cu (pour-served) | No (advisory) |
| `PHASE_A` | 4 | Motor phase, **skipped from routing AND no pour zone** | **Yes** |
[...]

PR #3798, opened 2026-06-18
feat(lvs): graduate board-03 to hard copper-LVS gate + add board-03 e2e CI job

rjwalters:
## Summary

Completes the board-03 slice of #3780 Part 2: graduates board 03 (usb-joystick) out of the `ADVISORY_LVS_BOARDS` allowlist and onto a **hard** copper-LVS gate. A fresh clean-room `generate_design.py` regen (C++ backend built) routes board-03 fully and regenerates copper-LVS clean (0 shorts / 0 opens, verified 2/2 runs, recipe exit 0), so "passes CI" now means "copper-LVS clean" for this board too.

Board-04 and board-05 stay advisory. Board-04's graduation is deferred behind a recipe-reproducibility fix (#3797): the **committed** board-04 PCB is hand-fixed (#3785/#3796), but a fresh `generate_design.py` regen re-introduces the `OSC_IN`/`OSC_OUT` short + 20 same-net power opens (21 mismatches), so a `board-04-end-to-end` job would be RED.

## Changes

- **`src/kicad_tools/lvs/recipe.py`** — remove `"03-usb-joystick"` from `ADVISORY_LVS_BOARDS` (04/05 remain); refresh the allowlist comment + module docstring to record board-03's graduation and board-04's still-dirty fresh regen.
- **`boards/03-usb-joystick/generate_design.py`** — flip the `write_lvs_report(...)` call to `require_clean=True` (copper hard gate). `run_copper=True`, `run_label=False` unchanged (USB-C fixture has `schematic_net=None` label noise; label stays advisory). Updated the comment block.
- **`.github/workflows/ci.yml`** — add a `board-03-end-to-end` (LVS-only) job cloned from `board-06-end-to-end`: regenerates board-03 from its recipe, retains the **Build C++ router backend** step (the regen re-routes), emits `board.json` via `kct board-metrics`, and runs `check_board_00_e2e.py --stem usb_joystick --lvs-only`. Uses `--lvs-only` (not the full asserter) because board-03 is `status: partial` (4 advisory `via_in_pad` DRC).
- **`tests/test_lvs_recipe.py`** — update the allowlist membership assertion for board-03's graduation.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| `03-usb-joystick` removed from `ADVISORY_LVS_BOARDS`; 04/05 remain | ✅ | `test_advisory_allowlist_contains_known_dirty_boards` passes |
[...]

PR #3800, opened 2026-06-18
feat(boards): route boards 02/03/04 under deterministic iteration budget

### LINKED ISSUE #3799: router: A* route is non-deterministic despite --seed 42 (blocks board reproducibility / hard-gate graduation)
## Curator Diagnostic (2026-06-18) — VERDICT: bounded-buildable

A full diagnostic pass **reproduced the non-determinism and localized its single root cause**. It is NOT a deep C++ concurrency or float-ordering problem. The fix is bounded and a builder can land it with a regression test. The fix machinery (`--deterministic-budget`, #3538) **already exists in-tree and is already used by board-07** — boards 02/03/04 simply don't opt into it.

### Root cause: the per-net **wall-clock** A* cutoff (`--per-net-timeout`), not the seed

`--seed` only seeds Python's global `random` (`route_cmd.py:7783-7784`), which controls escape-strategy shuffles and the MST trial shuffle. It does **not** control the per-net A* search's **wall-clock deadline**. Each per-net A* search is gated by a `per_net_timeout` wall-clock budget (default **30 s**, `route_cmd.py:6802-6804`) checked inside the C++ A* loop (`cpp_backend.py:1266-1331`, deadline checked every 1024 iterations). On a loaded machine the budget fires mid-search and the net lands less copper; on a fast machine the same search completes. Same seed, different copper → exactly the #3797 symptom (153 vs 145 segments).

### Reproduction (this session)

| Scenario | Board | Flags | Run 1 vs Run 2 (UUID-normalized copper) |
|---|---|---|---|
| Normal, unloaded | 02 | `--per-net-timeout 30 --seed 42` | **IDENTICAL** (397 seg / 24 via) |
| Normal, unloaded | 04 | recipe flags, `--seed 42` | **IDENTICAL** (120 seg / 12 via) |
| **Budget forced to bind** | 04 | `--per-net-timeout 0.05 --seed 42` | **DIVERGENT** (31 vs 29 seg) ← reproduces #3797 |
| **deterministic-budget** | 04 | `--deterministic-budget --max-search-iterations 5000 --seed 42` | **IDENTICAL** (120 seg / 12 via) |

Method: routed each board's unrouted `output/*.kicad_pcb` twice with `PYTHONHASHSEED=42` fixed, stripped `(uuid ...)`/`(tstamp ...)` from `(segment|via|arc)` lines, sorted, and diffed. The C++ backend was built and verified (`uv run kct build-native --check` → `available (version 1.0.0)`). `shapely 2.1.2` present (`--extra dev`).

**Key finding:** the divergence is **wall-clock-load-dependent**, not intrinsic. On an unloaded machine boards 02/04 are already byte-stable; under a binding per-net budget they diverge. Swapping the wall-clock cutoff for the **iteration backstop** (`--deterministic-budget`) makes the abort point machine-independent → byte-stable copper, and routes *more* (the iteration cap lets the search complete deterministically rather than being severed mid-flight).

### Sources RULED OUT (already deterministic)

- **C++ A* tie-breaking**: `AStarNode` already carries a monotonic insertion counter (`search_seq_counter_`, #3144) so heap pop order is deterministic on f-score ties (`pathfinder.cpp:886,1354-1377,1744,1809`).
- **unordered_map/set iteration**: closed-set / g-score containers are keyed by grid coords and never iterated in priority order (`pathfinder.hpp:356` PQ is the only ordering authority).
- **Parallelism**: `parallel.py`'s `ThreadPoolExecutor`+`as_completed` path is **opt-in** and defaults OFF (`region_parallel=False`, `route_cmd.py:3218,4094,6168`); board recipes don't enable it.
- **PYTHONHASHSEED**: pinning it did NOT change the divergent result — hash-seed is not the driver here (though pinning it is cheap insurance and should be in the contract).

## Implementation Guidance (bounded fix)

Two coordinated changes; no router-engine code change required.

1. **Opt boards 02/03/04 into `--deterministic-budget`** (mirror board-07's recipe). Add `"--deterministic-budget"` to each board's `kct route` argv. Keep `--seed 42`. Size the outer `--timeout` generously (it becomes a safety backstop only; the normalizer warns if it would bind — `route_cmd.py:183-198`). Optionally pass an explicit `--max-search-iterations N` per board if the default 12M cap (`DETERMINISTIC_BUDGET_MAX_SEARCH_ITERATIONS`, `route_cmd.py:143`) ever truncates a net that would otherwise route.
2. **Pin `PYTHONHASHSEED=42`** in the CI per-board e2e gates that regenerate (and document it in the determinism contract) so dict/set-iteration entropy can never re-enter, matching board-07's CI convention.
3. **Add a determinism regression test** modeled on `scripts/ci/board06_determinism_smoke.sh` (re-route N times at seed 42 + `--deterministic-budget`, assert identical UUID-stripped content hash) for boards 02 and 04. A focused pytest analogue can route the small board-02 input twice in-process and assert byte-identical normalized copper to keep the gate fast.

This is the prerequisite the cluster needs: with `--deterministic-budget`, "the recipe is the source of truth" holds regardless of machine load, unblocking the #3797 hard-gate graduation and the board-05 work.

## Affected Files

- `boards/02-charlieplex-led/generate_design.py` — add `--deterministic-budget` to the `kct route` argv (route command ~line 575-600); keep `--seed 42`.
- `boards/03-usb-joystick/generate_design.py` — add `--deterministic-budget` to its `kct route` argv; keep `--seed 42`.
- `boards/04-stm32-devboard/generate_design.py` — add `--deterministic-budget` to the `kct route` argv (route command ~line 1215-1244); keep `--seed 42`.
- `src/kicad_tools/cli/route_cmd.py` — reference only; `--deterministic-budget` (`_normalize_deterministic_budget`, lines 146-199), `--per-net-timeout` default (6802-6804), and `--seed` (7783-7784) are the load-bearing knobs. No engine change expected; if the `--seed` help text (6865-6880) is updated to point at `--deterministic-budget` as the reproducibility companion, do it here.
- `scripts/ci/board06_determinism_smoke.sh` — template for a board-02/04 determinism smoke (copy + parameterize per board, or generalize).

## Test Plan

- [ ] Manual verification: route `boards/04-stm32-devboard/output/stm32_devboard.kicad_pcb` twice with the new recipe flags (`--deterministic-budget --seed 42`, `PYTHONHASHSEED=42`), strip `(uuid ...)`/`(tstamp ...)` from `(segment|via|arc)` lines, sort, and `diff` → must be IDENTICAL. Repeat for board-02.
- [ ] Negative control: confirm that WITHOUT `--deterministic-budget`, a tight `--per-net-timeout 0.05` produces DIVERGENT copper on board-04 (proves the test exercises the right failure mode), and WITH `--deterministic-budget --max-search-iterations 5000` it is IDENTICAL. (Both reproduced by the curator this session.)
- [ ] Automated tests: add a determinism regression (new `scripts/ci/board0{2,4}_determinism_smoke.sh` or a pytest analogue under `tests/router/`) asserting same board + `--seed 42` + `--deterministic-budget` + `PYTHONHASHSEED=42` twice → byte-identical UUID-normalized routed copper, for boards 02 and 04. Model on `scripts/ci/board06_determinism_smoke.sh` and the existing `tests/test_route_deterministic_budget.py`.
- [ ] Edge cases: verify the default 12M iteration cap does not lose reach on board-04 (route count >= the seed-42 reference); if it truncates, add an explicit `--max-search-iterations` to that board's recipe.
- [ ] shapely: copper-LVS verification needs `uv sync --extra dev` (shapely); without it ~16 false GND opens appear even on a clean board.
- [ ] C++ backend: `uv run kct build-native --check` must report `available` before running any of the above (fresh worktrees need `uv run kct build-native`).

## Scope guard

- Do NOT fix the boards' routing quality / DRC counts here (#3797 board-04 graduation, #3775/#3766 board-05). This issue is **router determinism only**: make the same seed reproduce the same copper. Whether that copper is *clean* is a separate, downstream concern those issues own.
- No router-engine (C++ or core A*) change is in scope — the engine is already deterministic; the fix is recipe/CI flags + a regression gate.

---

## Problem

The autorouter produces **different copper on repeated runs of the same board with the same `--seed 42`**. Observed on board-04 (#3797 curation, 2026-06-18): three fresh `generate_design.py` regens gave 153 / 145 / 145 segments and 16-short / 1-short / 1-short copper-LVS outcomes — all routing 9/9 nets, but byte-different copper each time.

`--seed 42` is passed and seeds Python's global `random`, yet runs still diverge. Something in the routing pipeline introduces non-determinism not controlled by that seed (candidate sources: dict/set iteration order, parallel/threaded A* in the C++ backend, time- or hash-seeded tie-breaking, floating-point non-associativity in cost accumulation, or `PYTHONHASHSEED` not pinned).

## Why it matters (this is the real blocker)

This is the **root cause** behind the board-reproducibility cluster:
- **#3797** (board-04) — can't graduate to a hard copper-LVS gate because a fresh regen isn't reproducibly clean; the route diverges run-to-run.
- **#3775 / #3766** (board-05) — the relayout/completion work is gated on reproducible routing.
- It also undermines the value of the per-board e2e CI gates (board-00/01/02/03), which regenerate and assert: a non-deterministic router means those gates are only green by luck of the draw, and could flake.

A board's committed routed PCB being clean is much less valuable if `generate_design.py` can't reproduce it. Determinism is the prerequisite for "the recipe is the source of truth."

## Investigate

- [ ] Pin down the non-determinism source: run the same board twice with `PYTHONHASHSEED=42` fixed and the C++ backend, diff the routed copper; bisect across the routing pipeline (global route → detailed A* → rip-up/reroute → zone fill).
- [ ] Check the C++ A* backend (`src/kicad_tools/router/cpp/`) for unseeded tie-breaking, thread-order dependence, or unordered-container iteration in frontier/neighbor expansion.
- [ ] Check Python-side ordering: net ordering, ripup candidate ordering, any `set`/`dict` iteration feeding routing order.
- [ ] Determine whether `--seed` should also pin `PYTHONHASHSEED` (and whether the recipes/CI export it — board-07 already uses `PYTHONHASHSEED=42`, suggesting this is a known lever).

## Acceptance
- [ ] Two fresh routes of the same board with the same seed produce **byte-identical** routed copper (UUID-normalized), across at least boards 02/03/04.
- [ ] A regression test asserts route determinism (same board + seed twice → identical segment/via set).
- [ ] Document the determinism contract (which knobs must be set: `--seed`, `PYTHONHASHSEED`, single-threaded vs parallel A*).

## Notes
- Surfaced during #3797 curation (board-04 recipe reproducibility). The board-02 fix (#3783) and the per-board e2e gates work today, but their reproducibility may be fragile / seed-of-the-day rather than guaranteed.
- Likely a prerequisite for closing #3797 (board-04 graduation) and for the board-05 work (#3775/#3766).
- IMPORTANT for anyone reproducing copper-LVS: the connectivity extractor's pour-membership test needs `shapely` (in the `[dev]`/`[geometry]` extras). Without it, ~16 false GND opens appear even on a clean board. Run `uv sync --extra dev`. CI does this.

rjwalters:
## Summary

The autorouter produced different routed copper on repeated runs of the same board with the same `--seed 42`. Root cause (curator-verified): `--seed` only seeds Python's global `random`; it does NOT control the per-net A* **wall-clock** cutoff (`--per-net-timeout`, default 30 s) checked inside the C++ A* loop. On a loaded machine that budget fires mid-search and the net lands less copper — same seed, different copper (the observed board-04 153/145/145-segment divergence).

The fix already exists in-tree (`--deterministic-budget`, #3538, already used by board-07). This PR opts boards 02/03/04 into it (keeping `--seed 42`), which disables the per-net wall-clock cutoff and pins a fixed node-expansion **iteration** backstop, making the abort point machine-independent. It also pins `PYTHONHASHSEED=42` on each route subprocess (and the board-02 e2e CI job) so dict/set string-iteration entropy cannot re-enter. No router-engine change.

## Changes

- `boards/02-charlieplex-led/generate_design.py` — add `--deterministic-budget`; pin `PYTHONHASHSEED=42` on the route subprocess.
- `boards/03-usb-joystick/generate_design.py` — same; `_sp.run` now passes the pinned env.
- `boards/04-stm32-devboard/generate_design.py` — same.
- `tests/router/test_board03_routing_baseline.py` — kept in lock-step with the board-03 recipe (adds `--deterministic-budget` + pinned `PYTHONHASHSEED`).
- `.github/workflows/ci.yml` — add `PYTHONHASHSEED: "42"` to the board-02 e2e job (board-03/06/07 already had it).
- `scripts/ci/board_route_determinism_smoke.sh` — NEW. Parameterized per-board (02/03/04) routed-copper determinism smoke modeled on `board06_determinism_smoke.sh`.
- `tests/router/test_board_route_determinism.py` — NEW. Routes a board's unrouted PCB twice and asserts byte-identical UUID-normalized copper. Board-02 always-on (PR CI); board-04 behind `KICAD_RUN_SLOW_BOARD04_DETERMINISM=1` (mirrors the board-06 slow gate).

| Boards 02/03/04 route reproducibly (byte-identical UUID-normalized copper twice) | ✅ | board-02 421 lines x2; board-03 585 lines x2; board-04 full-recipe 173 lines x2 — all IDENTICAL via the smoke script + the pytest |
| Routing quality preserved (02/03 copper-LVS clean; 04 no worse) | ✅ | board-02 DRC clean; board-03 recipe exits 0 with `lvs clean: True` and passes its reach floor; board-04 keeps 9/9 nets and **0 blocking DRC errors** against jlcpcb-tier1 (`check_routed_drc.py` OK — same as committed) |
| No net-completion regression | ✅ | board-03 13/13; board-04 9/9; board-02 all signal nets |
| Determinism regression test added + passing | ✅ | `tests/router/test_board_route_determinism.py` — board-02 PASSED, board-04 PASSED with the slow gate enabled |
| Existing e2e gates (00/01/02/03/06/07) still pass | ✅ | recipe flag change only; committed PCBs unchanged so their gate tests still pass; board-02 e2e regenerates fresh under the new deterministic recipe |
| No C++ router engine change; board-05 untouched | ✅ | only recipe/CI/test files changed |

## Test Plan

- C++ backend built and verified: `kct build-native --check` → `available (version 1.0.0)`; `shapely` present (`--extra dev`).
- `scripts/ci/board_route_determinism_smoke.sh 02|03|04 2` → PASS (byte-identical copper) for all three boards.
- `uv run pytest tests/router/test_board_route_determinism.py` → board-02 reproducibility PASS, helper unit test PASS; board-04 PASS under `KICAD_RUN_SLOW_BOARD04_DETERMINISM=1`.
- Full board-03 recipe regen: exit 0, `lvs clean: True`; board-03 slow reach-floor baseline test PASS with the new flags.
- Full board-04 recipe regen x2: byte-identical; `check_routed_drc.py` reports 0 blocking errors (same as committed).
- `ruff check .` clean; `ruff format --check` clean on changed files; `mypy` clean on the new test (board recipe mypy errors are pre-existing, outside edited lines).
- `shellcheck` clean on the new smoke script.

Note on committed PCBs: deliberately NOT regenerated/committed here. The recipe change alone makes regens reproducible; the committed PCBs still pass every gate, and changing board copper/DRC counts is explicitly out of scope (owned by #3797 / #3775 / #3766 per the curator scope guard).

Note on the negative control: WITHOUT `--deterministic-budget`, a tight `--per-net-timeout 0.05` diverges only **under load** (the bug is load-dependent), so it would be a flaky CI assertion. It is documented as the failure mode the positive test guards against, not encoded as a hard assertion.

Closes #3799

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1087,6 +1087,14 @@ jobs:
     defaults:
       run:
         shell: bash
+    env:
+      # Issue #3799: pin PYTHONHASHSEED so the board-02 regen is
+      # reproducible across runner environments (mirrors board-03/06/07).
+      # The recipe ALSO pins it on the route subprocess, but pinning it at
+      # the job level covers the whole regen pipeline (schematic + zone +
+      # LVS) for belt-and-suspenders determinism.
+      PYTHONHASHSEED: "42"
+      PYTHONUNBUFFERED: "1"
     steps:
       - name: Ensure container has prerequisites
         run: |
--- a/boards/02-charlieplex-led/generate_design.py
+++ b/boards/02-charlieplex-led/generate_design.py
@@ -19,6 +19,7 @@ Usage:
 If no output directory is specified, files are written to ./output/
 """
 
+import os
 import subprocess
 import sys
 import uuid
@@ -580,8 +581,17 @@ def route_pcb(input_path: Path, output_path: Path) -> bool:
         "negotiated",
         "--iterations",
         "30",
-        "--per-net-timeout",
-        "30",
+        # Issue #3799: route under an ITERATION budget instead of the
+        # per-net WALL-CLOCK cutoff.  --seed only seeds Python's global
+        # random; it does NOT control the per-net A* deadline checked in
+        # the C++ loop.  On a loaded machine that wall-clock budget fires
+        # mid-search and the net lands less copper -- same seed, different
+        # copper.  --deterministic-budget (#3538) disables the per-net
+        # wall-clock cutoff and pins a fixed node-expansion backstop, so
+        # the seed-42 re-route is byte-identical (UUID-normalized) across
+        # machines.  --timeout 240 below is retained only as a SAFETY
+        # backstop (the normalizer warns if it would bind).
+        "--deterministic-budget",
         "--timeout",
         "240",
         "--seed",
@@ -598,13 +608,22 @@ def route_pcb(input_path: Path, output_path: Path) -> bool:
         "jlcpcb",
     ]
 
+    # Issue #3799: pin PYTHONHASHSEED for the route subprocess so any
+    # string-keyed dict/set iteration in the negotiated router is
+    # reproducible across runner environments (CPython randomizes string
+    # hashing per-process otherwise).  Combined with --seed 42 +
+    # --deterministic-budget this makes the full pipeline deterministic,
+    # not just the A* loop.  Mirrors board-07's convention.
+    env = os.environ.copy()
+    env["PYTHONHASHSEED"] = "42"
+
     print(f"\n1. Input: {input_path}")
     print(f"   Output: {output_path}")
     print(f"   Skipping nets: {skip_nets}")
-    print(f"   Command: {' '.join(cmd)}")
+    print(f"   Command: PYTHONHASHSEED={env['PYTHONHASHSEED']} {' '.join(cmd)}")
     print("\n2. Routing...")
 
-    result = subprocess.run(cmd, capture_output=False, text=True)
+    result = subprocess.run(cmd, capture_output=False, text=True, env=env)
 
     # ``kct route`` returns 0 on full success and a non-zero code on
     # partial / failed routing.  Either way it writes a routed PCB to
--- a/boards/02-charlieplex-led/route_demo.py
+++ b/boards/02-charlieplex-led/route_demo.py
@@ -29,6 +29,7 @@ Example:
 """
 
 import contextlib
+import os
 import subprocess
 import sys
 from pathlib import Path
@@ -194,8 +195,19 @@ def main():
         "negotiated",
         "--iterations",
         "30",
-        "--per-net-timeout",
-        "30",
+        # Issue #3799: route under an ITERATION budget instead of the
+        # per-net WALL-CLOCK cutoff so the seed-42 route is byte-identical
+        # (UUID-normalized) across machines.  --seed only seeds Python's
+        # global random; it does NOT control the per-net A* deadline checked
+        # in the C++ loop.  On a loaded machine that wall-clock budget fires
+        # mid-search and the net lands less copper -- same seed, different
+        # copper.  --deterministic-budget (#3538) disables the per-net
+        # wall-clock cutoff and pins a fixed node-expansion backstop.
+        # --timeout 240 below is retained only as a SAFETY backstop.
+        # This MUST stay in sync with generate_design.py:route_pcb()
+        # (Issue #3207 no-drift guard,
+        # tests/test_board02_route_demo_recipe.py).
+        "--deterministic-budget",
         "--timeout",
         "240",
         "--seed",
@@ -219,7 +231,15 @@ def main():
     # It returns 0 on full success and a non-zero code on partial routing
     # (in which case it still writes the partially-routed PCB).  We capture
     # stdout for net-count parsing by tee-ing through subprocess.PIPE.
-    result = subprocess.run(cmd, capture_output=True, text=True)
+    # Issue #3799: pin PYTHONHASHSEED for the route subprocess so any
+    # string-keyed dict/set iteration in the negotiated router is
+    # reproducible across runner environments (CPython randomizes string
+    # hashing per-process otherwise).  Combined with --seed 42 +
+    # --deterministic-budget this makes the full pipeline deterministic,
+    # mirroring generate_design.py:route_pcb().
+    env = os.environ.copy()
+    env["PYTHONHASHSEED"] = "42"
+    result = subprocess.run(cmd, capture_output=True, text=True, env=env)
     # Echo subprocess output so users can see routing progress, errors, etc.
     if result.stdout:
         print(result.stdout)
--- a/boards/03-usb-joystick/generate_design.py
+++ b/boards/03-usb-joystick/generate_design.py
@@ -23,6 +23,7 @@ Usage:
 If no output directory is specified, files are written to ./output/
 """
 
+import os
 import subprocess
 import sys
 from pathlib import Path
@@ -533,6 +534,17 @@ def route_pcb(input_path: Path, output_path: Path) -> bool:
         "jlcpcb-tier1",
         "--backend",
         "cpp",
+        # Issue #3799: route under an ITERATION budget instead of the
+        # per-net WALL-CLOCK cutoff.  --seed only seeds Python's global
+        # random; it does NOT control the per-net A* deadline checked in
+        # the C++ loop, so on a loaded machine the wall-clock budget fires
+        # mid-search and the net lands less copper -- same seed, different
+        # copper.  --deterministic-budget (#3538) disables the per-net
+        # wall-clock cutoff and pins a fixed node-expansion backstop, so
+        # the seed-42 re-route is byte-identical (UUID-normalized) across
+        # machines.  --timeout 600 below is then a SAFETY backstop only.
+        # KEEP IN SYNC with tests/router/test_board03_routing_baseline.py.
+        "--deterministic-budget",
         "--timeout",
         "600",
         # Issues #3507/#3454: ``--raw`` (skip TraceOptimizer) was
@@ -549,7 +561,21 @@ def route_pcb(input_path: Path, output_path: Path) -> bool:
         # the sidecar-aware ``kct check``).
     ]
     print(f"   $ {' '.join(cmd[1:])}")
-    proc = _sp.run(cmd, capture_output=True, text=True, timeout=1800, check=False)
+    # Issue #3799: pin PYTHONHASHSEED for the route subprocess so any
+    # string-keyed dict/set iteration in the negotiated router is
+    # reproducible across runner environments.  Combined with --seed 42 +
+    # --deterministic-budget this makes the full pipeline deterministic,
+    # not just the A* loop.  Mirrors board-07's convention.
+    _route_env = os.environ.copy()
+    _route_env["PYTHONHASHSEED"] = "42"
+    proc = _sp.run(
+        cmd,
+        capture_output=True,
+        text=True,
+        timeout=1800,
+        check=False,
+        env=_route_env,
+    )
     # Echo the route output so reach parsers (and humans) see the full log.
     print(proc.stdout)
     if proc.returncode in (1, 5):
--- a/boards/04-stm32-devboard/generate_design.py
+++ b/boards/04-stm32-devboard/generate_design.py
@@ -23,6 +23,7 @@ Usage:
 If no output directory is specified, files are written to ./output/
 """
 
+import os
 import subprocess
 import sys
 import uuid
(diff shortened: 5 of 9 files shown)


rjwalters:
Addressed both PR-caused Test-job CI failures.

**Failure 1 — recipe/demo flag-mirror guard** (`test_route_demo_recipe_mirrors_generate_design`):
- `boards/02-charlieplex-led/route_demo.py` now uses `--deterministic-budget` (replacing `--per-net-timeout 30`) and pins `PYTHONHASHSEED=42` on the route subprocess, exactly mirroring `generate_design.py:route_pcb()`.
- `tests/test_board02_route_demo_recipe.py`: removed `--per-net-timeout` from the expected flag pairs, added an assertion that `--deterministic-budget` is present in BOTH recipes, and added a guard that `--per-net-timeout` does not creep back.
- Boards 03/04: board-03's `route_demo.py` delegates to `route_pcb()` directly (cannot drift), board-04 has no `route_demo.py`. No analogous guards needed updating.

**Failure 2 — determinism test CI timeout** (`test_board02_route_is_reproducible`):
- Added `@pytest.mark.timeout(600)` to override CI's suite-wide `--timeout=60` reaper. The two-route board-02 check (~40-60s) now completes under a host-/contention-tolerant budget.
- The determinism regression still runs **unconditionally in the main Test job** (PR CI included) — not deferred to nightly slow-tests. Assertion unchanged: routes twice, asserts byte-identical UUID-normalized copper.

Verified locally:
- `test_route_demo_recipe_mirrors_generate_design` PASSED
- `test_board02_route_is_reproducible` PASSED in 55s under CI-style `--timeout=60` (confirming the marker override)
- `ruff check .` clean, `ruff format --check .` clean

C++ backend confirmed available (v1.0.0). Did not touch the router engine, the e2e gates, or board-05. Ready for re-review.

rjwalters:
✅ **Approved!** (Re-review after Doctor pass — final cycle)

Reviewed the COMPLETE PR (builder commit `fe3fc5c` + Doctor commit `f760ab7d`) together. All 14 CI checks pass and `mergeStateStatus` is CLEAN, including the Test job that previously failed (20312 passed, 87 skipped, 14m05s).

## (a) Determinism achieved — boards 02 / 03 / 04
- **Board 02**: the always-on `tests/router/test_board_route_determinism.py::test_board02_route_is_reproducible` (routes the committed unrouted PCB TWICE, asserts byte-identical UUID-normalized copper) has no `skipif` — it ran in the main Test job as part of the 20312 passed. The e2e job also regenerated fresh under the new recipe.
- **Board 03**: e2e fresh-regen log confirms `[deterministic-budget] Iteration-budgeted routing enabled (#3538): per-net wall-clock cutoff DISABLED, C++ A* iteration backstop pinned to 12,000,000 node expansions. Routed output is reproducible across machines.`
- **Board 04**: `test_board04_route_is_reproducible` (slow gate `KICAD_RUN_SLOW_BOARD04_DETERMINISM=1`, mirrors board-06) opts in with the identical recipe flag set; the smoke script covers it too.

Root cause correctly addressed: `--seed` only seeds Python's global `random`; the divergence came from the per-net WALL-CLOCK `--per-net-timeout` cutoff in the C++ A* loop. Swapping it for `--deterministic-budget` (fixed node-expansion backstop) + pinned `PYTHONHASHSEED=42` makes the abort point machine-independent. No router-engine change.

## (b) Routing quality NOT regressed (the critical gate)
- **Board 02** fresh regen: `DRC PASSED (0 errors)`, `LVS PASSED (label + copper: 0 mismatch)`, `Overall PASSED`, 8/8 nets, `board.json drc_violations=0, lvs_clean=True`.
- **Board 03** fresh regen (HARD copper-LVS gate from #3798): `Routed: 13/13 nets (100%)` and **`copper-LVS PASS: 0 shorts / 0 opens`**. The Board 03 End-to-End (LVS-only) job is green on fresh regeneration — the deterministic budget did not starve nets or produce reproducible-but-worse copper.
- **Board 04**: keeps 9/9 nets + 0 blocking DRC per the recipe/PR evidence.
- All e2e gates green: Board 00/01/02/03/06/07 End-to-End, plus diff-pair (06) and match-group (07) regressions.

## (c) Budget sized to preserve completion
12M-iteration backstop keeps net completion at today's level: board-02 8/8, board-03 13/13 (100%), board-04 9/9. The `--timeout` values (240/600) are retained only as SAFETY backstops; the budget — not wall-clock — bounds the work (normalizer warns if the timeout would bind).

## (d) Doctor's two fixes are correct and not weakening
1. **Recipe/demo/test reconciliation** — `route_demo.py` now mirrors `generate_design.py:route_pcb()` (`--deterministic-budget` + `PYTHONHASHSEED=42`, dropped `--per-net-timeout`). `test_route_demo_recipe_mirrors_generate_design` was tightened, not loosened: it now asserts `--deterministic-budget` is present in BOTH recipes AND adds a creep guard asserting `--per-net-timeout` is absent from both. Sound.
2. **`@pytest.mark.timeout(600)`** — legitimate: the test routes ~40-60s (two routes) and CI's suite-wide `--timeout=60` reaper killed it spuriously. The marker only moves the reaper deadline; the assertion is unchanged (still routes TWICE, asserts byte-identical normalized copper). The determinism regression genuinely still runs unconditionally in the main Test job — not deleted, not deferred to nightly.

## (e) No committed PCB changed; board-05 untouched; no engine change
Diff (9 files) contains zero `.kicad_pcb`, zero `board-05`/`05-bldc`, and zero `src/kicad_tools`/`router/cpp`/`cpp_backend` paths — recipe-flag + CI + test/smoke changes only. The new smoke script is shellcheck-clean per the test plan; the new pytest helper has an order-free normalization unit test.

PR description uses `Closes #3799`. Excellent, well-documented fix with thorough inline rationale and a hard-gated regression test.

PR #3810, opened 2026-06-20
fix(pcb): quote numeric footprint property values on embed

### LINKED ISSUE #3802: create-pcb: unquoted numeric property values make boards unloadable in KiCad
## Problem
`kct create-pcb` emits component property values **unquoted when they are purely numeric**, e.g. a resistor with value `470`:
```
(property "Value" 470
```
KiCad requires a quoted string (`(property "Value" "470")`). With an unquoted value, **KiCad/kicad-cli silently fail to load the entire board** — `kicad-cli pcb drc` prints only `Failed to load board` (exit 3) and `pcbnew.LoadBoard()` returns `None`. The board is unopenable in the GUI too. This is silent and catastrophic for any agentic flow that ends in gerbers.

## Repro
1. Have any component with a unit-less numeric value (e.g. `R 470`).
2. `kct create-pcb design.kicad_sch -o out.kicad_pcb`
3. `kicad-cli pcb drc out.kicad_pcb -o /tmp/d.json --format json` → `Failed to load board`.

## Root cause
Bisected (header loads; first failing footprint = the one whose Value is pure-numeric). The footprint-embed path writes the Value token without quoting. (`pcb/exporter.py` correctly quotes, but the .kicad_mod embed path in the create-pcb/PCBFromSchematic workflow does not.)

## Fix
Always wrap property values (Value, Reference, and any custom props) in double quotes when serializing, regardless of whether the string looks numeric. Add a regression test with a unit-less numeric value (e.g. `470`, `0`, `100`) and assert `kicad-cli pcb drc` loads the board.

## Workaround
`perl -pi -e 's/(\(property "[^"]+") (-?\d[\d.]*)(\s*\n)/$1 "$2"$3/g' board.kicad_pcb`

## Impact
Critical — blocks the entire schematic→PCB→manufacturing path for common designs.

---
_Found during chorus-test-revA agentic-design validation (2026-06-20): generating + placing + routing a real 63-component mixed-signal HAT end-to-end with kicad-tools._

---

### Root cause (confirmed in source)

The defect is in the footprint-embed write path used by `create-pcb` /
`PCBFromSchematic`, specifically `PCB.add_footprint_from_file()` in
`src/kicad_tools/schema/pcb.py`. When it sets the Value/Reference property it calls:

```python
# src/kicad_tools/schema/pcb.py ~3936-3943
for prop in fp_sexp.find_all("property"):
    prop_name = prop.get_string(0)
    if prop_name == "Reference":
        prop.set_value(1, reference)   # <-- creates an UNFLAGGED atom
    elif prop_name == "Value":
        prop.set_value(1, value)       # <-- creates an UNFLAGGED atom
```

`SExp.set_value()` (`src/kicad_tools/sexp/parser.py:791`) builds a fresh atom via
`SExp(value=value)` with the default `_originally_quoted=False`. At serialize time
`_format_atom()` (`parser.py:483`) only quotes a string when
`self._originally_quoted or self._needs_quoting(self.value)` is true. But
`_needs_quoting()` (`parser.py:522`) returns **False for any string that parses as a
float** (lines 664-669):

```python
# Don't quote numeric values (integers and floats including negatives)
try:
    float(s)
    return False
except ValueError:
    pass
```

So `value="470"` serializes as the bare token `470`, which kicad-cli rejects.
Reproduced directly:

```
set_value('470'):  (property "Value" 470)     <-- bug: bare numeric, board unloadable
set_value('470k'): (property "Value" "470k")  <-- fine: non-numeric forces quoting
```

The same hazard applies to the fallback branches that synthesize new
property nodes when Reference/Value are absent (`pcb.py:3956-3980`, via
`SExp.list("property", "Reference", reference)` / `SExp.list("property", "Value", value)`)
and to **any custom property** routed through the same embed path: a unit-less
numeric Value is the common trigger, but a numeric custom-property value would fail
identically.

### Recommended fix

The codebase already has the correct primitive: `SExp.quoted_atom()`
(`parser.py:817`), which sets `_originally_quoted=True` so the atom always
serializes quoted even when its text looks numeric. It was introduced for exactly
this class of bug (pad numbers, `generator_version`, etc.).

Preferred approach — fix at the property-write site in
`add_footprint_from_file()` so Reference/Value (and the synthesized-property
fallback) always emit quoted atoms:

- For the existing-property branch, replace `prop.set_value(1, reference)` /
  `prop.set_value(1, value)` with assignment of a `SExp.quoted_atom(...)` into
  `prop.children[1]` (mirroring `set_value`'s index handling but preserving the
  quoted flag).
- For the synthesized-property fallback, build the value atom with
  `SExp.quoted_atom(...)` rather than passing the bare string to `SExp.list(...)`.

This is the **narrow, low-risk fix** and is sufficient to close the reported
defect. Keep the change scoped to the property name/value atoms; do not change
unrelated numeric tokens (`at`, `size`, layer indices, etc.), which must stay bare.

Optional hardening (consider, but call out the blast radius before doing it):
strengthening the global `_needs_quoting()` heuristic to always quote KiCad
"string-typed" fields is much broader and risks regressing round-trip fidelity of
genuinely-numeric tokens elsewhere; prefer the targeted write-site fix unless a
follow-up explicitly scopes a serializer-wide change.

Note: `add_footprint()` (`pcb.py:4001`) delegates to `add_footprint_from_file()`,
so fixing the latter covers both public entry points and the
`PCBFromSchematic.add_component()` workflow that `create-pcb` uses.

### Acceptance Criteria

- [ ] `PCB.add_footprint_from_file()` (and therefore `add_footprint()` and the
      `create-pcb` / `PCBFromSchematic` path) emits `(property "Value" "470")` —
      quoted — for a unit-less numeric Value such as `470`, `0`, or `100`.
- [ ] Reference designators and any custom property values that happen to be
      numeric are likewise emitted quoted.
- [ ] The synthesized-property fallback branches (when a footprint lacks a
      Reference or Value property) also emit quoted values.
- [ ] A board created via `create-pcb` from a schematic containing a component
      with a pure-numeric value loads successfully under
      `kicad-cli pcb drc` (no "Failed to load board", exit 0).
- [ ] Genuinely-numeric structural tokens (`at`, `size`, net numbers, layer
      indices) remain **unquoted** — the fix must not over-quote.

## Affected Files

- `src/kicad_tools/schema/pcb.py` — `add_footprint_from_file()` (~lines 3936-3980):
  emit Reference/Value (and synthesized-property fallback) as quoted atoms via
  `SExp.quoted_atom()` instead of `set_value`/bare `SExp.list` strings. This is the
  single fix site; `add_footprint()` delegates here.
- `src/kicad_tools/sexp/parser.py` — reference only: `quoted_atom()` (line 817),
  `_format_atom()` (483), `_needs_quoting()` (522). No change expected unless the
  team opts for the broader serializer-level hardening (not recommended for this
  issue).
- `tests/test_kicad_cli_roundtrip.py` — add the regression test (see Test Plan).
  This module already has `_assert_kicad_cli_loads()` and a module-level
  `skipif(find_kicad_cli() is None)` guard, so the new test fits the existing pattern.

## Test Plan

- [ ] Manual verification: Create a schematic with a resistor whose Value is a
      bare number (e.g. `470`), run
      `kct create-pcb design.kicad_sch -o out.kicad_pcb`, confirm the emitted
      file contains `(property "Value" "470")` and that
      `kicad-cli pcb drc out.kicad_pcb -o /tmp/d.json --format json` exits 0
      (no "Failed to load board").
- [ ] Automated unit test: In `tests/test_kicad_cli_roundtrip.py`, add a test
      that calls `add_footprint_from_file()` (or `add_footprint()`) with
      `value="470"`, saves the PCB, and asserts (a) the serialized text contains
      `"470"` quoted and (b) `_assert_kicad_cli_loads()` passes. Parametrize over
      `"470"`, `"0"`, `"100"` and a numeric custom-property value.
- [ ] Edge cases: verify a numeric Reference-like value and a numeric custom
      property are quoted; verify negative/decimal numeric values
      (e.g. `-5`, `3.3`) are quoted; verify structural numeric tokens (`at`,
      `size`) stay unquoted (no over-quoting regression).
- [ ] Unit-level (no kicad-cli) assertion on serialized string so the regression
      is caught even on runners without kicad-cli installed.

rjwalters:
## Summary
`kct create-pcb` / `PCBFromSchematic` embedded footprint Reference/Value
properties via `SExp.set_value()`, which produces an atom with
`_originally_quoted=False`. The serializer's textual heuristic
(`_needs_quoting`) does **not** quote strings that parse as a float, so a
unit-less numeric value like `470` serialized as the bare token
`(property "Value" 470)`. KiCad/kicad-cli reject that with "Failed to load
board" (exit 3), silently breaking the entire schematic -> PCB ->
manufacturing path for common designs (any resistor with a bare numeric value).

## Changes
- `src/kicad_tools/schema/pcb.py` — `add_footprint_from_file()` now routes
  Reference/Value through the existing `SExp.quoted_atom()` helper in all
  three write paths: the existing-`property` branch, the KiCad-7 `fp_text`
  fallback, and the synthesized-`property` fallback (when the footprint lacks a
  Reference/Value node). `add_footprint()` delegates here, so both public
  entry points and the create-pcb workflow are covered.
- `tests/test_pcb.py` — `TestAddFootprintNumericPropertyQuoting`: kicad-cli-free
  unit assertions on the serialized string (parametrized over `470/0/100/-5/3.3`),
  covering both the synthesized and existing-property branches, a numeric
  Reference, and a no-over-quoting check for structural tokens.
- `tests/test_kicad_cli_roundtrip.py` — `TestNumericPropertyValueRoundtrip`:
  loads a board built via `add_footprint_from_file` with a numeric Value/Reference
  through `kicad-cli pcb drc` and asserts it loads (exit 0).

Scope kept narrow per the curated guidance: no change to the global
`_needs_quoting()` heuristic; structural numeric tokens (`at`, `size`, layer
indices, net numbers) remain bare.

| Numeric Value (470/0/100) emits `(property "Value" "470")` quoted | done | `test_synthesized_value_property_is_quoted`, `test_existing_value_property_is_quoted` |
| Numeric Reference / custom values quoted | done | `test_numeric_reference_is_quoted`, `test_numeric_reference_loads_in_kicad_cli` |
| Synthesized-property fallback emits quoted | done | fixture footprint has no Reference/Value -> exercises fallback in `test_synthesized_*` |
| Board with numeric value loads under kicad-cli (exit 0) | done | `TestNumericPropertyValueRoundtrip` (4 passed with kicad-cli present) |
| Structural numerics (at/size) stay unquoted | done | `test_structural_numeric_tokens_stay_unquoted` |

## Test Plan
- `uv run pytest tests/test_pcb.py tests/test_kicad_cli_roundtrip.py` — 209 passed
  (kicad-cli present locally, so the roundtrip load gate runs).
- `uv run ruff check .` — passed; `uv run ruff format . --check` — passed after
  formatting the two touched test files.
- mypy on `src/kicad_tools/schema/pcb.py` reports only pre-existing errors
  (lines 610-1848, 5122); none on the touched lines (~3936-3982).

Closes #3802

### DIFF
--- a/src/kicad_tools/schema/pcb.py
+++ b/src/kicad_tools/schema/pcb.py
@@ -3936,25 +3936,33 @@ class PCB:
         for prop in fp_sexp.find_all("property"):
             prop_name = prop.get_string(0)
             if prop_name == "Reference":
-                prop.set_value(1, reference)
+                # Use quoted_atom so values that look numeric (e.g. a
+                # reference "1") still serialize quoted. A bare numeric
+                # property value makes the board unloadable in kicad-cli.
+                prop.children[1] = SExp.quoted_atom(reference)
                 ref_updated = True
             elif prop_name == "Value":
-                prop.set_value(1, value)
+                # Use quoted_atom so unit-less numeric values (e.g. "470",
+                # "0", "100") serialize as (property "Value" "470") rather
+                # than the bare token (property "Value" 470), which
+                # kicad-cli rejects with "Failed to load board".
+                prop.children[1] = SExp.quoted_atom(value)
                 val_updated = True
 
         # Fall back to KiCad 7 fp_text format
         for fp_text in fp_sexp.find_all("fp_text"):
             text_type = fp_text.get_string(0)
             if text_type == "reference" and not ref_updated:
-                fp_text.set_value(1, reference)
+                fp_text.children[1] = SExp.quoted_atom(reference)
                 ref_updated = True
             elif text_type == "value" and not val_updated:
-                fp_text.set_value(1, value)
+                fp_text.children[1] = SExp.quoted_atom(value)
                 val_updated = True
 
         # If reference/value weren't found, add them as KiCad 8+ properties
         if not ref_updated:
-            ref_prop = SExp.list("property", "Reference", reference)
+            # Quote the value so a numeric reference doesn't serialize bare.
+            ref_prop = SExp.list("property", "Reference", SExp.quoted_atom(reference))
             ref_prop.append(SExp.list("at", 0.0, -1.5))
             ref_prop.append(SExp.list("layer", layer.replace(".Cu", ".SilkS")))
             ref_prop.append(SExp.list("uuid", str(uuid.uuid4())))
@@ -3967,7 +3975,8 @@ class PCB:
             fp_sexp.append(ref_prop)
 
         if not val_updated:
-            val_prop = SExp.list("property", "Value", value)
+            # Quote the value so a unit-less numeric value doesn't serialize bare.
+            val_prop = SExp.list("property", "Value", SExp.quoted_atom(value))
             val_prop.append(SExp.list("at", 0.0, 1.5))
             val_prop.append(SExp.list("layer", layer.replace(".Cu", ".Fab")))
             val_prop.append(SExp.list("uuid", str(uuid.uuid4())))
--- a/tests/test_kicad_cli_roundtrip.py
+++ b/tests/test_kicad_cli_roundtrip.py
@@ -719,3 +719,83 @@ class TestSheetInstancesPageQuoting:
         node = sheet_instances("/00000000-0000-0000-0000-000000000001", "1")
         text = node.to_string()
         assert '(page "1")' in text, f"sheet_instances builder emitted: {text}"
+
+
+# ---------------------------------------------------------------------------
+# Numeric footprint property quoting (issue #3802)
+# ---------------------------------------------------------------------------
+#
+# create-pcb / PCBFromSchematic embed Reference/Value via
+# PCB.add_footprint_from_file(). When a value parses as a float (e.g. a
+# unit-less resistor value "470"), the serializer's textual heuristic used to
+# emit the bare token (property "Value" 470) -- which kicad-cli rejects with
+# "Failed to load board" (exit 3), silently breaking the entire schematic ->
+# PCB -> manufacturing path. The fix forces Reference/Value (and the
+# synthesized-property fallback) through SExp.quoted_atom() so they always
+# serialize quoted. These tests gate the load contract against the real
+# kicad-cli parser; the kicad-cli-free unit assertions live in
+# tests/test_pcb.py:TestAddFootprintNumericPropertyQuoting.
+
+_FIXTURE_FP = (
+    REPO_ROOT / "tests" / "fixtures" / "Test_Library.pretty" / "C_0402_1005Metric.kicad_mod"
+)
+
+
+class TestNumericPropertyValueRoundtrip:
+    """A board with a unit-less numeric footprint Value must load in kicad-cli."""
+
+    @pytest.mark.parametrize("value", ["470", "0", "100"])
+    def test_numeric_value_loads_in_kicad_cli(self, value: str, tmp_path: Path) -> None:
+        """add_footprint_from_file with a numeric Value -> loadable board."""
+        pcb = PCB.create(width=50.0, height=50.0, title="numeric value roundtrip")
+        pcb.add_footprint_from_file(
+            kicad_mod_path=_FIXTURE_FP,
+            reference="R1",
+            x=25.0,
+            y=25.0,
+            value=value,
+        )
+        pcb_path = tmp_path / f"numeric_{value}.kicad_pcb"
+        pcb.save(pcb_path)
+
+        # Regression guard: the value must be quoted before kicad-cli is even
+        # invoked, so the failure is attributed clearly on runners where
+        # kicad-cli's stderr is unhelpful.
+        contents = pcb_path.read_text()
+        assert f'(property "Value" "{value}"' in contents, (
+            f"Regression guard: a numeric Value {value!r} must serialize as "
+            f'(property "Value" "{value}") -- quoted. The bare numeric form is '
+            "rejected by kicad-cli with 'Failed to load board' (exit 3). See "
+            "PCB.add_footprint_from_file in src/kicad_tools/schema/pcb.py "
+            "(issue #3802)."
+        )
+        assert f'(property "Value" {value}' not in contents
+
+        _assert_kicad_cli_loads(
+            pcb_path,
+            producer="PCB.add_footprint_from_file (numeric Value)",
+            tmp_path=tmp_path,
+        )
+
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

The fix is correct, minimal, and well-tested. Root cause and remedy both verified.

**Correctness**
`add_footprint_from_file()` previously embedded Reference/Value via `set_value()`, which produces an atom with `_originally_quoted=False`. The serializer quotes a string only when `self._originally_quoted or self._needs_quoting(self.value)` (`src/kicad_tools/sexp/parser.py:493`), and `_needs_quoting` returns False for numeric-looking strings — so `value="470"` serialized as the bare token `(property "Value" 470)`, which kicad-cli rejects with "Failed to load board". Routing all three write paths (existing-`property`, KiCad-7 `fp_text`, and synthesized-`property` fallback) through the existing `SExp.quoted_atom()` helper sets `_originally_quoted=True` and forces the quote. This matches the established pattern already used for other strict-typed string fields (e.g. `generator_version`, pad numbers). `add_footprint()` delegates to `add_footprint_from_file()`, so the public API and the create-pcb workflow are both covered.

**Scope**
Correctly avoids touching the global `_needs_quoting()` heuristic. `test_structural_numeric_tokens_stay_unquoted` confirms structural numerics (`at`, `size`, `thickness`) stay bare — no over-quoting regression.

**Test coverage**
- `tests/test_pcb.py::TestAddFootprintNumericPropertyQuoting` — kicad-cli-free unit assertions parametrized over `470/0/100/-5/3.3`, covering synthesized + existing-property branches, numeric Reference, and the no-over-quoting guard.
- `tests/test_kicad_cli_roundtrip.py::TestNumericPropertyValueRoundtrip` — load-level gate through the real kicad-cli parser.
- Ran locally in the issue-3802 worktree: **14 passed** (kicad-cli present, so the roundtrip gate executed).

**Quality gates**
- `ruff check` + `ruff format --check` on all three touched files — clean.
- All 14 CI checks pass (Lint & Format, Type Check, Test, kicad-cli Round-trip Smoke, all board E2E + routing regressions).
- Merge state: CLEAN. PR description uses `Closes #3802`.

## Test Scoping
**Strategy**: scoped pytest on the two new test classes (changed files: 1 src + 2 test files)
**Result**: 14 selected, all passed in 7.79s. Full suite confirmed green in CI (Test job pass, 14m8s).

Issue #2706, opened 2026-05-11
drc: false-positive clearance_segment_via at in-pad escape via centers


rjwalters:
## Symptom

After PR #2704 (in-pad via escape for fine-pitch LQFP/QFP, closes #2695) lands, board 04 routes 9/9 with `--manufacturer jlcpcb-tier1`. The routed PCB contains a via at (126.8375, 121.75) and a B.Cu inner-layer escape segment.

However, `kct check --mfr jlcpcb-tier1` reports **2 false-positive `clearance_segment_via` errors** at the via center. The segment electrically IS the via — it's a sub-mm-length stub that emerges from the via on the inner layer to land on the destination pad. The DRC checker models the segment as a separate trace and computes clearance vs. the via as if they were distinct copper features.

This is a DRC-checker modeling gap, not a real clearance violation. The trace endpoint is the via tag, so the geometric distance between them is zero by definition — which the rule reads as "0 < min_clearance → violation."

## Why it matters

- Every fine-pitch in-pad escape on every board going forward will produce these false positives.
- Board 04's 2-layer jlcpcb-tier1 route is structurally clean (9/9 routed), but the DRC report makes it look broken.
- Board 06 (diffpair-test) uses BGA-49 fanout with similar in-pad geometry (Phase 4L); the same false positives apply there.
- The `routed-drc-tolerance.yml` allowlist mechanism papers over the symptom but doesn't fix the underlying model.

## Recommended fix paths

**Option A — segment-via co-location detection** (preferred):
- In `src/kicad_tools/validate/rules/clearance.py` (the `clearance_segment_via` rule), detect when a segment endpoint coincides with a via center (within ~ε).
- When co-located, treat the segment as electrically continuous with the via — skip the clearance check for that segment-via pair.
- The check still fires when the segment is *near but not coincident with* the via (real layout error).

**Option B — escape-marker annotation**:
- When `_try_in_pad_escape` places a via + segment, annotate them both with a custom tag (e.g., a KiCad `(tstamp ...)` matching pair).
- DRC rule reads the annotation and skips co-located pairs.
- Less robust (annotation can be lost in round-trips through KiCad GUI) but cleaner separation.

Option A is recommended — purely geometric, no metadata round-trip risk.

## Reproduction

```
cd boards/04-stm32-devboard
kct route --seed 42 --differential-pairs --auto-layers --manufacturer jlcpcb-tier1 --timeout 240
[...]

PR #3819, opened 2026-06-20
feat(validate): detect net-0 stray-copper bridges between assigned nets

### LINKED ISSUE #3816: drc(D1): internal ClearanceRule skips net-0 copper, missing stray-copper shorts
## Follow-on from #3803 (slice D1)

The native-DRC route gate landed in #3803 (PR #3815). The curator's recommended split included **D1**: the internal `DRCChecker` is structurally blind to a class of shorts that native KiCad catches, so without `kicad-cli` present the internal-only PASS can still ship a shorted board.

## Problem

`ClearanceRule._check_layer` short-circuits any pair where either element is on net 0:

```python
# src/kicad_tools/validate/rules/clearance.py (post-#3815, now lines 643-645)
# Skip net 0 (unconnected) elements
if elem1.net_number == 0 or elem2.net_number == 0:
    continue
```

Router-emitted stray copper / unassigned-net fills therefore never participate in clearance/short checks. Same-layer copper that shorts two different assigned nets *via a net-0 island* (a piece of stray copper that physically bridges net A and net B but carries no net assignment itself) is invisible to the internal engine. KiCad's own DRC catches this through full connectivity-graph analysis; `kct check` does not.

### Why the skip exists (and why we can't just delete it)

The skip is load-bearing. Verified against the routed boards in this repo:

| board | net-0 pads | net-0 segs | net-0 vias |
|-------|-----------:|-----------:|-----------:|
| 01-voltage-divider | 0 | 0 | 0 |
| 02-charlieplex-led | 0 | 0 | 0 |
| 03-usb-joystick | 2 | 0 | 0 |
| 04-stm32-devboard | **31** | 0 | 0 |
| 05-bldc-motor-controller | 4 | 0 | 0 |
| 06-diffpair-test | 0 | 0 | 0 |

Board 04's 31 net-0 pads are the unconnected fine-pitch QFP pins of `U2` (an unused IC): 0.3 mm-wide SMD pads sitting immediately adjacent to *other net-0 pins of the same part*. Boards 03/05 carry analogous unconnected pads (NPTH / unused pins / test points). These are **legitimate unconnected copper**. If the net-0 skip were simply removed (treating net 0 as a foreign net against everything), `ClearanceRule`'s O(n^2) pairwise loop would fire a `clearance_pad_pad` error on every adjacent net-0 QFP pin pair and on every net-0 pad that happens to sit within `min_clearance` of one assigned net. That is exactly the over-firing the issue says to avoid.

### Root cause of the gap

The pairwise loop has no notion of *connectivity*. "net-0 copper shorts two nets" is a connectivity property of a net-0 copper island (does it touch >= 2 distinct assigned nets?), not a property of any single element pair. The current rule cannot express it, so the safest thing it could do was skip net 0 entirely - which masks the genuine short.

## Approach (bounded - bridge detection, not full DRC)

Add a focused **net-0 bridge** check. The principle that avoids false positives:

> A net-0 copper element is only a defect when it is electrically connected (overlapping, or within `< min_clearance` / effectively touching) to copper of **two or more distinct assigned nets** simultaneously. A net-0 element touching zero or one assigned net is legitimate unconnected copper and must NOT fire.

Recommended shape (keep it surgical, reuse existing geometry helpers):

1. Collect net-0 copper elements per layer (segments, pads, vias) using the existing `_collect_elements` machinery (or a net-0 filtered variant). Reuse `_via_spans_layer` for barrel layers exactly as the existing collector does.
2. Build connectivity among net-0 elements: union-find / flood-fill net-0 elements that touch each other (edge-to-edge clearance <= a small touch epsilon, e.g. reuse `DRC_TOLERANCE` / `_COLOCATION_EPSILON_MM` semantics) into **net-0 islands**. This groups the QFP's adjacent net-0 pins into a single island, but an island of mutually-touching net-0 pins that touches no assigned net is harmless.
3. For each net-0 island, find the set of **distinct assigned (nonzero) nets** whose copper the island touches/overlaps (edge-to-edge clearance below a short threshold - use the same `_calculate_clearance` distance math the rule already uses; "touch" means clearance < 0 or effectively zero). If that set has **>= 2 distinct net numbers**, emit ONE short violation per offending island (or per offending net-pair within the island), `severity="error"`, with a distinct `rule_id` such as `clearance_net0_bridge`, naming the two bridged nets and the island location.
4. A net-0 island touching 0 or 1 assigned nets emits nothing.

Notes / guardrails:
- Reuse the existing colocation epsilon (`_COLOCATION_EPSILON_MM`, the router in-pad-escape invariant) so a net-0 segment endpoint that legitimately coincides with a via center is not misread as a bridge - mirror the existing suppression at clearance.py ~696-705.
- Diff-pair set handling is irrelevant here (diff pairs are assigned nets); no interaction.
- Keep the new logic inside `ClearanceRule` (a new private method called from `_check_layer` or `check`) so it ships through the existing `DRCChecker.check_clearances` -> `check_all` path with no checker.py wiring beyond what already exists. The existing strict CI gate counts every blocking `clearance_*` violation, so a `clearance_net0_*` id is picked up with no allowlist surgery (same pattern noted for `clearance_via_zone` / `clearance_pad_zone`).
- Do NOT pull in shapely-zone-fill machinery; this is about discrete net-0 copper elements (segments/pads/vias), parallel to the base `ClearanceRule`, not zone fills. Zone-fill net-0 handling and full geometry-model parity remain the deferred D3 backlog item.

## Acceptance criteria

- [ ] A board with a net-0 copper island (e.g. a stray net-0 segment) physically bridging two distinct assigned nets is flagged as a short by the internal `DRCChecker` with no `kicad-cli` required (surfaces through `kct check` / `check_all`).
- [ ] A net-0 element (pad/segment/via) adjacent to **zero or one** assigned net produces **no** new violation.
- [ ] Adjacent net-0-to-net-0 copper (e.g. board 04 `U2` QFP pins) produces **no** new violation.
- [ ] No new violations on the existing routed boards 01..06 (regression-guard: run `kct check` / `check_all` before and after; counts unchanged).
- [ ] The new violation carries `severity="error"`, a distinct `rule_id` (e.g. `clearance_net0_bridge`), and names both bridged net names in `nets=(...)`.
- [ ] Regression test with a minimal in-memory fixture: (a) positive case - net-0 segment overlapping two assigned-net pads/segments fires exactly one short; (b) negative case - a lone net-0 pad near one assigned net stays silent; (c) negative case - two touching net-0 elements with no assigned-net contact stay silent.

## Test plan

- Unit tests in `tests/` following the existing `MagicMock`-based `CopperElement` fixture pattern (see `tests/test_clearance_net_names.py`) and the in-memory PCB construction used in `tests/test_validate_segment_zone_clearance.py`.
- Positive fixture: two pads on nets 1 and 2, a net-0 segment whose endpoints overlap both -> expect one `clearance_net0_bridge` error naming both nets.
- Negative fixtures: (1) single net-0 pad 0.5 mm from a net-1 pad -> 0 violations; (2) two net-0 pads touching each other but nothing assigned -> 0 violations; (3) replicate the board-04 QFP geometry (a cluster of net-0 pads adjacent to each other and to at most one assigned net) -> 0 violations.
- Board regression: assert `DRCChecker(...).check_clearances()` (or `check_all`) violation count on boards/01..06 is unchanged from baseline.

## Affected Files

- `src/kicad_tools/validate/rules/clearance.py` - add net-0 bridge detection in/around `ClearanceRule` (skip currently at lines 643-645; new private method).
- `tests/test_validate_clearance_net0_bridge.py` (new) - unit + regression coverage.

## References

- Root cause traced in #3803 curation; the net-0 skip at `src/kicad_tools/validate/rules/clearance.py:643-645`.
- Native-DRC route gate: #3803 / PR #3815.
- Prior net-0 / colocation handling precedent: `_COLOCATION_EPSILON_MM` and the segment/via colocation suppression (clearance.py ~696-705, Issue #2706); via-barrel all-layer collection (Issue #3487).
- Deferred follow-on: full geometry-model parity (D3 backlog).

rjwalters:
## Summary

`ClearanceRule`'s O(n^2) pairwise loop skips any pair where either element is on net 0. That skip is load-bearing (it prevents flooding `clearance_pad_pad` false positives on legitimate unconnected copper, e.g. board 04's 31 net-0 QFP pins), but it masked a genuine short class: a piece of stray net-0 copper that physically bridges two distinct assigned nets carries the short while having no net of its own, so no element pair ever fired. KiCad's native DRC catches this via connectivity analysis; the internal engine did not.

This adds a bounded, connectivity-based **net-0 bridge** detector inside `ClearanceRule` — not a reimplementation of full DRC.

## Changes

- `src/kicad_tools/validate/rules/clearance.py`:
  - `_check_net0_bridges()` — runs per copper layer from `check()`, alongside the existing pairwise pass. Groups mutually-touching net-0 copper elements into islands via union-find, then for each island collects the distinct assigned (nonzero) nets it touches. Fires **one** `clearance_net0_bridge` error (severity `error`) only when an island bridges **>= 2 distinct assigned nets**, naming both nets in `nets=(...)`.
  - `_elements_touch()` — reuses the existing `_calculate_clearance` distance math and the `_COLOCATION_EPSILON_MM` in-pad segment/via colocation guard (mirroring the pairwise-loop suppression from Issue #2706) so a net-0 segment endpoint coinciding with a via center is not misread as a bridge.
  - `_create_net0_bridge_violation()` — builds the violation, deterministically naming the two lowest-numbered bridged nets.
- `src/kicad_tools/drc/violation.py`: register `CLEARANCE_NET0_BRIDGE` enum member + explicit `from_string` alias so the new rule_id is not silently miscategorized as the generic `CLEARANCE` by the fuzzy fallback.
- `tests/test_validate_clearance_net0_bridge.py` (new): positive, negative, and board-regression coverage.

The net-0 skip in the pairwise loop is intentionally left in place — `clearance_pad_pad` etc. still skip net 0; the new connectivity check is the only path that examines net-0 copper.

| Net-0 island bridging two assigned nets flagged by internal `DRCChecker` (no kicad-cli) | ✅ | `test_net0_segment_bridging_two_nets_fires_once` via `check_clearances()` |
| Net-0 element adjacent to 0/1 assigned nets produces no new violation | ✅ | `test_lone_net0_pad_near_one_net_silent`, `test_net0_pad_overlapping_one_net_silent` |
| Adjacent net-0-to-net-0 copper (board-04 QFP analogue) produces no new violation | ✅ | `test_two_touching_net0_pads_silent`, `test_qfp_like_net0_cluster_near_one_net_silent` |
| No new violations on routed boards 01..06 (counts unchanged) | ✅ | All six boards report 0 `clearance_net0_bridge` and total clearance count unchanged; `TestNet0BridgeBoardRegression` parametrized over 01..06 |
| Violation has `severity="error"`, distinct `rule_id`, names both nets | ✅ | Asserted in positive test |
| Regression test with in-memory fixtures (positive + two negatives) | ✅ | 12 tests in new file, all passing |

## Test Plan

- `uv run pytest tests/test_validate_clearance_net0_bridge.py` — 12 passed.
- Manual check: all of boards 01..06 routed PCBs report `net0_bridge=0` and unchanged total clearance counts.
- `uv run ruff check` + `ruff format --check` clean on touched files.
- mypy: no new errors from added code (pre-existing shapely-stub / unused-ignore errors in unrelated zone-fill code remain).

Note: pre-existing failures in `tests/test_validate_segment_zone_clearance.py` exist on the base commit (verified via `git stash`) and are unrelated to this change (shapely zone-fill machinery, untouched here).

Closes #3816

### DIFF
--- a/src/kicad_tools/drc/violation.py
+++ b/src/kicad_tools/drc/violation.py
@@ -141,6 +141,13 @@ class ViolationType(Enum):
     # sub-minimum gaps to committed filled_polygon geometry.
     CLEARANCE_SEGMENT_ZONE = "clearance_segment_zone"
     CLEARANCE_VIA_VIA = "clearance_via_via"
+    # Net-0 stray-copper bridge: a piece of unassigned (net 0) copper that
+    # electrically shorts two distinct assigned nets (Issue #3816).  The
+    # base ClearanceRule pairwise loop skips net 0 (legitimate unconnected
+    # copper is common), so this connectivity-based short is reported under
+    # a distinct rule_id and MUST be aliased explicitly below -- the fuzzy
+    # fallback would otherwise miscategorize it as the generic CLEARANCE.
+    CLEARANCE_NET0_BRIDGE = "clearance_net0_bridge"
     # Differential-pair within-pair clearance (Issue #2560, Epic #2556 Phase 1D).
     # Distinct from the generic CLEARANCE family because it validates the
     # *intra-pair* gap (allowed to be tighter than the manufacturer's inter-pair
@@ -282,6 +289,11 @@ class ViolationType(Enum):
             "clearance_segment_segment": cls.CLEARANCE_SEGMENT_SEGMENT,
             "clearance_segment_via": cls.CLEARANCE_SEGMENT_VIA,
             "clearance_via_via": cls.CLEARANCE_VIA_VIA,
+            # Net-0 stray-copper bridge (Issue #3816).  MUST be aliased
+            # explicitly -- the fuzzy "clearance" fallback below would
+            # otherwise drop through to the generic CLEARANCE type and
+            # mask the new short class from downstream exact-type filters.
+            "clearance_net0_bridge": cls.CLEARANCE_NET0_BRIDGE,
             # clearance subtypes using "trace" (synonym for "segment")
             "clearance_pad_trace": cls.CLEARANCE_PAD_SEGMENT,
             "clearance_trace_trace": cls.CLEARANCE_SEGMENT_SEGMENT,
--- a/src/kicad_tools/validate/rules/clearance.py
+++ b/src/kicad_tools/validate/rules/clearance.py
@@ -602,6 +602,16 @@ class ClearanceRule(DRCRule):
             for v in violations:
                 results.add(v)
 
+            # Net-0 bridge detection (Issue #3816).  The pairwise loop in
+            # ``_check_layer`` deliberately skips net 0 to avoid flooding
+            # false positives on legitimate unconnected copper (e.g.
+            # board 04's 31 net-0 QFP pins).  That skip masks a genuine
+            # short class: a piece of stray net-0 copper that physically
+            # bridges two different ASSIGNED nets.  Detect that explicitly
+            # as a connectivity property of net-0 copper islands.
+            for v in self._check_net0_bridges(pcb, layer_name, min_clearance):
+                results.add(v)
+
         # Count rules checked (one per layer)
         results.rules_checked = len(pcb.copper_layers)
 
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the connectivity-based net-0 BRIDGE detector. Verdict: solid, well-scoped, and faithfully guarded against the false-positive class the issue mandates.

## Correctness
- **Union-find** (`_check_net0_bridges`): standard path-halving `_find` + union; net-0 elements grouped into islands by mutual touch, then each island's distinct assigned (nonzero) nets collected. Fires exactly **one** `clearance_net0_bridge` error per island only when `len(touched) >= 2` — verified correct.
- **Colocation guard** (`_elements_touch`): faithfully mirrors the pairwise-loop segment/via in-pad suppression (Issue #2706) using `_COLOCATION_EPSILON_MM` and the same `math.hypot` endpoint-vs-via-center test. A net-0 segment endpoint coinciding with a via center is correctly NOT read as a copper join.
- **Pairwise net-0 skip left untouched** (clearance.py:653-655) — confirmed; the new connectivity check is the only path that examines net-0 copper.
- `_create_net0_bridge_violation` deterministically names the two lowest-numbered bridged nets; `severity="error"`, distinct `rule_id`, both nets in `nets=(...)`.
- `ViolationType.CLEARANCE_NET0_BRIDGE` registered with an **explicit `from_string` alias** — verified `from_string("clearance_net0_bridge")` resolves to its own type, not the generic `CLEARANCE` fuzzy fallback.

## False-positive guard (critical)
- Independently ran all routed boards 01–06: every board reports `net0_bridge=0` **and** total clearance count unchanged (0 on both main and PR source).
- Board 04's 31 net-0 QFP pads do **not** trip the detector.
- `TestNet0BridgeBoardRegression` parametrizes over all 6 boards, **runs (not skipped)**, all pass.

## Tests
- 12/12 pass in `tests/test_validate_clearance_net0_bridge.py`: positive (single-segment bridge + multi-segment island fires once), negative lone pad (near 1 net / overlapping 1 net), negative net-0-only cluster (2 pads + QFP-like 6-pin row).
- 180 passed across related clearance/violation/colocation/filter suites — no regressions.

## Pre-existing failures (independently confirmed NOT introduced by this PR)
- `test_validate_segment_zone_clearance.py` is **byte-identical** between main and this PR (0-line diff) and passes equally (21 passed) on both source versions locally. This PR touches only `clearance.py` + `violation.py` + the new test — no zone-fill/shapely/geometric code.
- mypy's 4 shapely-stub / unused-ignore errors pre-exist on main (verified by checking out main's `clearance.py`).
- The **Match-Group Routing Regression (board 07)** CI failure exists on `main` itself (conclusion: failure) and is out of scope — this PR has no path to board-07 routing.

## CI / lint
- ruff check + format clean on all touched files.
- 14 CI checks pass; only failing check is the pre-existing/out-of-scope board-07 matchgroup regression.

Nice work isolating this as a connectivity property of net-0 islands rather than reimplementing full DRC.

PR #3833, opened 2026-06-21
fix(deps): make shapely a core dependency and fail loud when absent

### LINKED ISSUE #3824: drc: shapely is load-bearing for kct DRC + zone-fill but is only an optional extra (silent shorts + unguarded crash + missing from [all])
## Original Issue

Discovered during a clean-rebuild sweep of all test boards. shapely is declared only in the optional geometry/dev extras (pyproject.toml ~lines 88-90), but the core DRC and zone-fill paths hard-depend on it, with dangerous failure modes when it is absent:

1. SILENT BAD FILLS (HIGH): zone fill skips carving pad-clearance pockets when shapely is missing — the subtraction is a no-op returning 0 (src/kicad_tools/zones/fill_clearance.py:733-737 and :919-923, "A no-op (returns 0) when shapely is unavailable"). Result: pour copper overlaps foreign-net pads. Reproduced on board-00: a shapely-less fill produced real clearance_pad_zone shorts (GND pad overlaps VCC fill on F.Cu, etc.) that a shapely-present fill does not.

2. UNGUARDED CRASH (HIGH): the clearance DRC rule uses bare "from shapely import ..." with no ImportError guard (src/kicad_tools/validate/rules/clearance.py:806-807, 949-951, 1132-1134), so kct check / kct audit on any PCB die with a raw "ModuleNotFoundError: No module named shapely" instead of degrading. Contrast src/kicad_tools/validate/connectivity.py which guards via _has_shapely(). So the one check that would catch the shorts from (1) is also the one that hard-crashes when shapely is missing.

3. PACKAGING GAP (MED): shapely is absent from the [all] extra in pyproject.toml, so pip install kicad-tools[all] yields an install where kct check on a PCB crashes. The failure is a raw ModuleNotFoundError leak rather than a friendly "install kicad-tools[geometry]" message.

Recommended fix: make shapely a core dependency (it is effectively required for correct DRC/fill), OR (b) guard every shapely import to fail-loud with an actionable message AND make zone-fill refuse to claim a clean fill when it could not carve clearances. At minimum add shapely to the [all] extra. Acceptance: a default/[all] install runs kct check on a PCB without crashing; zone fill never silently omits pad clearance (either it carves, or it errors/warns loudly that the fill is not clearance-correct).

---

### Problem Statement

`shapely` is the geometry backend for the project's two correctness-critical copper paths — zone fill clearance carving and the zone-vs-copper clearance DRC rules — yet it is declared *only* under the optional `geometry`/`dev` extras (`pyproject.toml` core `dependencies` at lines 41-47 omit it; `geometry` extra at 88-91; `dev` extra at 217). The project clearly intends DRC and fill to work out of the box (`kct check`, `kct audit`, and the board-build recipes are advertised as primary features), so the current packaging produces three failure modes that range from confusing to dangerous:

1. **Silent bad fills (HIGH, correctness):** `apply_foreign_pad_clearance()` is the function that subtracts foreign-net antipads from each zone `filled_polygon`. When `shapely` is missing it returns `0` and leaves the fill exactly as `kicad-cli` produced it (no clearance carving). The caller cannot distinguish "0 polygons needed carving" from "0 because shapely was absent", so the pipeline reports a clean fill while real `clearance_pad_zone` shorts remain in the copper. This is the worst failure: a fab-ready board with shorts and no warning.
2. **Unguarded crash (HIGH, robustness):** the zone-clearance DRC rules import shapely at the top of their `check()` bodies with bare `import`/`from shapely import ...` and no `ImportError` handling, so `kct check`/`kct audit` raise `ModuleNotFoundError` on any PCB. The check that would *catch* the shorts from (1) is itself the one that hard-crashes when shapely is absent.
3. **Packaging gap (MED):** `shapely` is missing from the `[all]` extra (`pyproject.toml` lines 172-192), so even `pip install kicad-tools[all]` yields an install where `kct check` on a PCB crashes.

### Root Cause Analysis

- Core install (`pip install kicad-tools`) and `[all]` install both lack shapely. Only `[geometry]` and `[dev]` pull it in.
- The codebase has **two inconsistent conventions** for the optional import:
  - **Degrade-silently** (`fill_clearance.py` no-op returns) — wrong for a correctness path because the caller treats "no-op" as "nothing to do".
  - **Crash** (`validate/rules/clearance.py` bare imports) — wrong for a CLI that should degrade gracefully or fail with guidance.
- A **third, correct** convention already exists and should be the template: `validate/connectivity.py` (`_SHAPELY_AVAILABLE` / `_has_shapely()` guard, lines 38-50) and `zones/generator.py:342-359` (catches `ImportError` and emits a friendly `install with: pip install kicad-tools[geometry]` message via `warnings.warn`).

### Curator verification of the enumerations (read against current source on this branch)

The issue's three named DRC crash sites are confirmed, but the enumeration is **incomplete**. A full `git ls-files 'src/**/*.py' | xargs grep -nE 'import shapely|from shapely'` shows the unguarded surface is larger:

- `src/kicad_tools/validate/rules/clearance.py` — bare imports inside DRC code at lines **806-807** (`_repair_fill_polygon`), **849** (`Polygon`), **949-951** (`SegmentZoneClearanceRule.check`), **997** (`nearest_points`), **1132-1134** (`ViaZoneClearanceRule.check`), **1260** and **1285** (`nearest_points`). All seven sites are reachable from `kct check`/`kct audit` and must be covered by the guard, not just the three originally listed.
- `src/kicad_tools/zones/fill_clearance.py` — the two HIGH no-op sites are confirmed: `force_solid_on_isolated_island_pads` at **733-737** and `apply_foreign_pad_clearance` at **919-923**. Additional bare imports (no guard, will crash if reached on a core-only install) at **134, 344-345, 376, 420, 1085-1086**.
- `src/kicad_tools/zones/generator.py` — bare imports at **435, 530, 1209, 1569-1570, 1619, 1681** (some are already wrapped in `try/except ImportError` at 342, 538, 1224, 1571, 1620, 1682 — verify each before changing).
- `src/kicad_tools/pcb/board_geometry.py` — bare imports at **40-42, 191-193, 417, 523**.

Whatever option is chosen, the fix must cover the **complete** set above, not only the three sites named in the original report. The Builder should re-run the `grep` at implementation time to confirm no new sites have landed.

### Recommendation: make `shapely` a CORE dependency

**Recommended: Option A — promote `shapely>=2.0` into `[project] dependencies`.**

Rationale:
- shapely is *effectively required* for correct DRC and fill. The project's headline workflows (`kct check`, `kct audit`, board-build recipes) are not correct without it, so treating it as optional misrepresents what a core install can do.
- shapely 2.x ships manylinux/macOS/Windows wheels (vendored GEOS), so adding it to core does not impose a system-library build burden on typical users — the historical reason to keep geometry optional is largely gone.
- It collapses three failure modes into zero: no silent fills, no crashes, no packaging gap. The two divergent conventions (silent no-op vs crash) stop mattering for the default install.
- It matches the project's stated intent that DRC/fill work out of the box.

**Even with Option A, the silent-no-op guards in `fill_clearance.py` MUST still be hardened** (defense in depth — see "Non-negotiable" below), because a core dep can still be absent in a broken/partial environment, and a correctness path must never *silently* claim success.

**Option B (fallback, if maintainers want to keep shapely optional):** keep shapely under `[geometry]` but (1) add it to `[all]`, (2) replace every bare import in the DRC/fill paths with a fail-loud guard that raises/reports an actionable `install kicad-tools[geometry]` message (model on `zones/generator.py:356`), and (3) make zone fill **refuse to claim a clean fill** when it could not carve clearances. Option B is strictly more work than A and still leaves the default install unable to run a correct DRC.

### Non-negotiable invariant (applies to BOTH options)

The SILENT-BAD-FILL path (failure mode 1) must NEVER silently report a clean fill when it could not carve clearances. Concretely, `apply_foreign_pad_clearance()` and `force_solid_on_isolated_island_pads()` must not return `0` as if there were nothing to do when shapely is the reason they did nothing. Acceptable behaviors when shapely is unavailable:
- raise a clear error that aborts the fill with an actionable message, OR
- emit a loud WARNING/ERROR (surfaced to the CLI and any report/manifest) that the fill is NOT clearance-correct and must not be treated as fab-ready.

A pure return-0 that is indistinguishable from "nothing needed carving" is forbidden.

### Implementation Guidance

If choosing **Option A (recommended):**
1. Add `"shapely>=2.0"` to `[project] dependencies` (`pyproject.toml` lines 41-47). Remove the now-redundant `shapely` line from `[dev]` (line 217) and keep/leave `[geometry]` as an alias for back-compat (optional). It does NOT need to be added to `[all]` separately since core deps are always installed, but adding it there too is harmless and explicit.
2. Harden the two `fill_clearance.py` no-op sites so a missing shapely is loud, not silent (see invariant above). Prefer raising from these functions, or returning a sentinel the caller turns into a surfaced warning — do NOT keep a bare `return 0`.
3. The bare DRC imports in `clearance.py` become safe once shapely is core, but for robustness against partial installs, route them through a shared `_has_shapely()`-style guard or a single helper that raises an actionable error. Reuse/extend the `connectivity.py` `_has_shapely()` pattern (consider lifting it to a small shared module, e.g. `kicad_tools/_shapely.py`, so connectivity, clearance, fill, generator, and board_geometry share one guard + one install message).

If choosing **Option B:**
1. Add `"shapely>=2.0"` to the `[all]` extra (`pyproject.toml` lines 172-192).
2. Introduce one shared guard helper (e.g. `kicad_tools/_shapely.py` with `has_shapely()` and `require_shapely(feature: str)` that raises with `install with: pip install kicad-tools[geometry]`).
3. Replace every bare shapely import in the enumerated DRC/fill/geometry sites with the guard. DRC rules should either skip-with-warning or raise the actionable error rather than `ModuleNotFoundError`.
4. Make `apply_foreign_pad_clearance()` / `force_solid_on_isolated_island_pads()` fail loud (invariant).

Consistency note for either option: align all five modules (`connectivity.py`, `clearance.py`, `fill_clearance.py`, `generator.py`, `board_geometry.py`) on the single guard/helper so the convention is no longer split three ways.

### Affected Files

- `pyproject.toml` - Option A: add `shapely>=2.0` to `[project] dependencies` (lines 41-47); Option B: add to `[all]` extra (lines 172-192). Reconcile the `[dev]`/`[geometry]` entries.
- `src/kicad_tools/zones/fill_clearance.py` - Harden the two no-op guards (`force_solid_on_isolated_island_pads` ~733-737, `apply_foreign_pad_clearance` ~919-923) to fail loud; route the other bare imports (134, 344-345, 376, 420, 1085-1086) through the shared guard.
- `src/kicad_tools/validate/rules/clearance.py` - Replace bare imports at 806-807, 849, 949-951, 997, 1132-1134, 1260, 1285 with the shared guard so DRC degrades/raises with an actionable message instead of `ModuleNotFoundError`.
- `src/kicad_tools/validate/connectivity.py` - Reference template for the guard (`_has_shapely()` lines 38-50); optionally refactor to consume the new shared helper.
- `src/kicad_tools/zones/generator.py` - Reference template for the friendly install message (line 356); align remaining bare imports (435, 530, 1209, 1569-1570, 1619, 1681) on the shared guard.
- `src/kicad_tools/pcb/board_geometry.py` - Bare imports at 40-42, 191-193, 417, 523; align on the shared guard.
- `src/kicad_tools/_shapely.py` (NEW, optional but recommended) - Shared `has_shapely()` / `require_shapely(feature)` helper + single install-message string.

### Acceptance Criteria

- [ ] A **default** install (`pip install kicad-tools`, Option A) — or an `[all]` install (Option B) — can run `kct check` / `kct audit` on a real PCB without raising `ModuleNotFoundError` or any unguarded exception.
- [ ] Zone fill **never silently** omits pad clearance: when shapely is unavailable, the fill path either aborts with a clear error or emits a loud WARNING/ERROR (surfaced to the CLI and any report/manifest) stating the fill is not clearance-correct. A bare `return 0` that masquerades as "nothing to carve" is gone.
- [ ] No bare `import shapely` / `from shapely import ...` remains in the enumerated DRC, fill, generator, or board_geometry code paths reachable from the CLI; each is behind a shared guard with an actionable `install kicad-tools[geometry]` message.
- [ ] (Option A) `shapely>=2.0` is in `[project] dependencies`; (Option B) `shapely>=2.0` is in the `[all]` extra. `[dev]`/`[geometry]` reconciled with no duplicate or contradictory pins.
- [ ] On a shapely-present install, board-00 fill produces zero `clearance_pad_zone` violations (the regression originally observed disappears), confirming the carving path still works.

### Test Plan

- [ ] Manual verification (crash path): in an environment WITHOUT shapely, run `kct check <board>.kicad_pcb` and confirm it does not raise `ModuleNotFoundError`; it either completes or prints an actionable install message.
- [ ] Manual verification (silent-fill path): in an environment WITHOUT shapely, run the zone-fill pipeline and confirm it does NOT report a clean/clearance-correct fill silently — it must warn or error loudly.
- [ ] Manual verification (happy path): in an environment WITH shapely, run a clean-rebuild of board-00 and confirm fill carves foreign-net antipads (zero `clearance_pad_zone` shorts) — i.e., no regression in carving behavior.
- [ ] Automated tests: add a test that monkeypatches/simulates shapely being unimportable (e.g. via `sys.modules`/`importlib` shim or `monkeypatch.setitem(sys.modules, "shapely", None)`) and asserts (a) the DRC clearance rules degrade with the actionable message rather than `ModuleNotFoundError`, and (b) `apply_foreign_pad_clearance` / `force_solid_on_isolated_island_pads` raise or surface a loud warning rather than silently returning 0.
- [ ] Automated tests: add a packaging assertion (parse `pyproject.toml`) verifying shapely is present in the chosen target (core deps for Option A, or `[all]` for Option B).
- [ ] Edge cases: partial/broken shapely install (importable but missing a submodule); non-rectangular board outline path in `generator.py` already has a fallback — verify it still behaves; confirm `nearest_points` sites (clearance.py 997/1260/1285) are also guarded.
- [ ] Regression suite: `uv run pytest` (esp. zones/ and validate/ tests) and `pnpm check:ci` pass.

rjwalters:
## Summary

`shapely` is the geometry backend for the project's two correctness-critical copper paths -- zone-fill clearance carving and the zone-vs-copper clearance DRC rules -- yet it was declared only under the optional `geometry`/`dev` extras. That produced three failure modes (issue #3824):

1. **Silent bad fills (HIGH):** the carve functions returned `0` when shapely was missing, indistinguishable from "nothing to carve", so a board with real `clearance_pad_zone` shorts was reported as a clean fill.
2. **Unguarded crash (HIGH):** the zone-clearance DRC rules imported shapely with bare `from shapely import ...` and died with a raw `ModuleNotFoundError` in `kct check`/`kct audit`.
3. **Packaging gap (MED):** shapely was missing from `[all]`.

This implements the curator's recommended **Option A (shapely as a core dependency)** plus defense-in-depth fail-loud guards.

## Changes

- **Packaging:** promote `shapely>=2.0` into `[project] dependencies`; add it explicitly to the `[all]` extra; drop the now-redundant `[dev]` pin; keep `[geometry]` as a back-compat alias. Regenerated `uv.lock`.
- **New shared guard** `src/kicad_tools/_shapely.py`: `has_shapely()` (cached boolean probe) and `require_shapely(feature)` (raises with a single actionable `install kicad-tools[geometry]` message).
- **Silent-bad-fill path hardened:** `apply_foreign_pad_clearance()` and `force_solid_on_isolated_island_pads()` now `require_shapely()` and raise loud instead of `return 0`.
- **CLI callers no longer swallow the failure:** `runner._apply_foreign_pad_clearance`, the zone-remediation loop, and `stitch_cmd` now propagate `ModuleNotFoundError` loud / warn that the fill is NOT clearance-correct (previously a blanket `except Exception: return` masked it).
- **DRC rules guarded:** `SegmentZoneClearanceRule.check` and `ViaZoneClearanceRule.check` call `require_shapely()` up front, degrading with the actionable message instead of a raw `ModuleNotFoundError`.
- **Convention unified:** `connectivity.py` now consumes the shared `has_shapely()`. `board_geometry.py` already fails loud with a friendly message; `generator.py` already has friendly fallbacks/messages -- left as-is.
- **Tests:** `tests/test_shapely_core_dependency.py` (shapely-absent monkeypatch proving the fill path and both DRC rules fail loud with the actionable hint, plus pyproject packaging assertions); updated `test_zones_fill_clearance.py::test_no_shapely_fails_loud` to assert the new invariant.

| Default/[all] install runs `kct check` without `ModuleNotFoundError` | ✅ | shapely is now a core dep (and in `[all]`); guards raise actionable errors only on broken installs |
| Zone fill never silently omits pad clearance | ✅ | carve functions `require_shapely()` and raise; CLI callers propagate/warn loud instead of swallowing |
| No bare shapely import in reachable DRC/fill paths leaks a raw crash | ✅ | both DRC `check()` bodies and both carve entry points guard before any shapely touch |
| shapely>=2.0 in `[project] dependencies`; `[dev]`/`[geometry]` reconciled | ✅ | `test_shapely_is_core_dependency`, `test_shapely_present_in_all_extra`, `test_shapely_not_duplicated_in_dev_extra` |
| shapely-present fill still carves (no regression) | ✅ | existing `test_zones_fill_clearance.py` carve + zero-violation tests pass |

## Test Plan

- `uv run pytest tests/test_shapely_core_dependency.py` -- 10 passed.
- `uv run pytest -k 'clearance or zone or connectivity or shapely or fill or stitch'` -- 2399 passed, 4 skipped.
- `ruff check` / `ruff format --check` clean on touched files.
- `mypy` clean on new files; remaining errors in `connectivity.py`/`stitch_cmd.py`/`clearance.py` are pre-existing on `main` (unrelated lines).

Closes #3824

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,14 @@ dependencies = [
     "pyyaml>=6.0",
     "numpy>=1.24",
     "pydantic>=2.0",
+    # shapely is load-bearing for the correctness-critical copper paths
+    # (zone-fill clearance carving and the zone-vs-copper clearance DRC
+    # rules), so it is a CORE dependency rather than an optional extra
+    # (issue #3824).  shapely 2.x ships prebuilt manylinux/macOS/Windows
+    # wheels (vendored GEOS), so this imposes no system-library build
+    # burden on typical installs.  The ``geometry`` extra remains as a
+    # back-compat alias.
+    "shapely>=2.0",
 ]
 
 [project.urls]
@@ -85,7 +93,9 @@ kicad-fab-rules = "kicad_tools.cli.fab_rules_cmd:main"
 kicad-project-clean = "kicad_tools.cli.clean_cmd:main"
 
 [project.optional-dependencies]
-# Shapely-based board geometry engine
+# Shapely-based board geometry engine.  shapely is now a CORE dependency
+# (see ``[project] dependencies``); this extra is retained as a
+# back-compat alias so ``pip install kicad-tools[geometry]`` keeps working.
 geometry = [
     "shapely>=2.0",
 ]
@@ -189,6 +199,8 @@ all = [
     "jinja2>=3.1",
     "markdown>=3.5",
     "weasyprint>=60.0",
+    # shapely is a core dependency; listed here too for explicitness.
+    "shapely>=2.0",
 ]
 # Development dependencies
 dev = [
@@ -211,10 +223,8 @@ dev = [
     "jinja2>=3.1",
     "markdown>=3.5",
     "weasyprint>=60.0",
-    # Issue #3413 phase 4: the board 06 recipe's copper-union pour audit
-    # (and its committed-artifact regression test) require shapely.  The
-    # diffpair CI job and the test jobs install via --extra dev.
-    "shapely>=2.0",
+    # shapely is now a CORE dependency (issue #3824) so it no longer needs
+    # an explicit entry here; it is always installed with the package.
 ]
 
 [tool.hatch.build.targets.wheel]
--- a/src/kicad_tools/_shapely.py
+++ b/src/kicad_tools/_shapely.py
@@ -1,0 +1,85 @@
+"""Shared guard for the optional/core ``shapely`` geometry backend.
+
+``shapely`` is a **core** dependency of kicad-tools (declared in
+``[project] dependencies`` in ``pyproject.toml``) because the project's
+correctness-critical copper paths — zone-fill clearance carving and the
+zone-vs-copper clearance DRC rules — cannot produce correct results
+without it.
+
+Historically ``shapely`` lived only under the optional ``geometry`` /
+``dev`` extras, which produced two divergent and both-wrong conventions
+across the codebase:
+
+* **Degrade silently** — return ``0`` / no-op when ``shapely`` is
+  missing.  This is *dangerous* on a correctness path because the caller
+  cannot tell "nothing needed doing" from "we couldn't do anything", so
+  a board with real shorts is reported as a clean fill.
+* **Crash** — bare ``import shapely`` raising ``ModuleNotFoundError`` deep
+  inside a DRC rule, so ``kct check`` / ``kct audit`` die with an opaque
+  traceback instead of an actionable message.
+
+This module is the single, honest convention every geometry-dependent
+module should use:
+
+* :func:`has_shapely` — cheap boolean probe (for fallback paths that have
+  a legitimate pure-Python alternative, e.g. axis-aligned rect inset).
+* :func:`require_shapely` — fail **loud** with an actionable install
+  message when ``shapely`` is genuinely unavailable.  Use this on any
+  path where there is no correct fallback, so a partial/broken
+  environment never silently yields a non-clearance-correct result.
+
+Even though ``shapely`` is now a core dependency, a broken or partial
+install can still leave it unimportable, so these guards remain the
+defense-in-depth layer that keeps a correctness path from ever *silently*
+claiming success.
+"""
+
+from __future__ import annotations
+
+# Friendly, actionable install hint reused everywhere a shapely-dependent
+# feature is requested without shapely available.  shapely is a core
+# dependency, so a missing import almost always means a broken/partial
+# environment; reinstalling the package (or the geometry extra) repairs it.
+SHAPELY_INSTALL_HINT = (
+    "shapely is required for this geometry operation but is not importable. "
+    "shapely is a core dependency of kicad-tools; reinstall it with: "
+    "pip install 'kicad-tools[geometry]' (or 'pip install shapely>=2.0')."
+)
+
+
+_SHAPELY_AVAILABLE: bool | None = None
+
+
+def has_shapely() -> bool:
+    """Return ``True`` when the ``shapely`` backend is importable.
+
+    The result is cached after the first probe.  Use this only for paths
+    that have a *correct* pure-Python fallback; correctness-critical paths
+    must use :func:`require_shapely` instead so a missing backend fails
+    loud rather than silently degrading.
+    """
+    global _SHAPELY_AVAILABLE
+    if _SHAPELY_AVAILABLE is None:
+        try:
+            import shapely  # type: ignore[import-untyped] # noqa: F401
+
+            _SHAPELY_AVAILABLE = True
+        except ImportError:
+            _SHAPELY_AVAILABLE = False
+    return _SHAPELY_AVAILABLE
+
+
+def require_shapely(feature: str) -> None:
+    """Raise :class:`ModuleNotFoundError` with an actionable hint if shapely is absent.
+
+    Call this at the top of any function that cannot produce a correct
+    result without ``shapely``.  ``feature`` names the operation so the
+    error explains *what* could not run, e.g.::
+
+        require_shapely("foreign-net pad clearance carving")
+
+    Raising (rather than a silent ``return 0``) guarantees a correctness
+    path never masquerades a skipped step as success.
+    """
+    if not has_shapely():
+        raise ModuleNotFoundError(f"{feature}: {SHAPELY_INSTALL_HINT}")
--- a/src/kicad_tools/cli/runner.py
+++ b/src/kicad_tools/cli/runner.py
@@ -533,6 +533,11 @@ def _remediate_starved_thermal(
         # refill + carve so the shipped copper reflects both.
         if _run_drc(refill=True) is not None:
             _settle()
+    except ModuleNotFoundError:
+        # shapely unavailable -> the carve/remediation is not clearance-
+        # correct (issue #3824).  Propagate loud rather than shipping a
+        # fill that silently omitted foreign-net clearances.
+        raise
     except Exception:
         # Never let remediation abort a successful fill.
         return
@@ -595,20 +600,30 @@ def _apply_foreign_pad_clearance(pcb_path: Path) -> None:
 
     Loads the filled PCB, applies the pure-Python clearance correction,
     and rewrites the file only when at least one filled_polygon changed.
-    Any failure (e.g. shapely missing) is swallowed — the fill is then
-    left exactly as kicad-cli produced it.
-    """
-    try:
-        from kicad_tools.core.sexp_file import save_pcb
-        from kicad_tools.sexp import parse_file
-        from kicad_tools.zones.fill_clearance import apply_foreign_pad_clearance
 
+    A missing/broken ``shapely`` is propagated **loud** (it is a core
+    dependency — issue #3824): silently leaving the fill as kicad-cli
+    produced it would ship a board with uncarved foreign-net antipads and
+    real ``clearance_pad_zone`` shorts while reporting a clean fill.  Only
+    non-import geometry edge cases are swallowed so a degenerate ring never
+    aborts an otherwise successful fill.
+    """
+    from kicad_tools.core.sexp_file import save_pcb
+    from kicad_tools.sexp import parse_file
+    from kicad_tools.zones.fill_clearance import apply_foreign_pad_clearance
+
+    try:
         doc = parse_file(pcb_path)
         modified = apply_foreign_pad_clearance(doc)
         if modified:
             save_pcb(doc, pcb_path)
+    except ModuleNotFoundError:
+        # shapely unavailable -> the fill is NOT clearance-correct.  Never
+        # swallow this: surface it so the caller cannot mistake an uncarved
+        # fill for a clean one.
+        raise
     except Exception:
-        # Never let the correction abort a successful fill.
+        # Non-import geometry edge case: never let it abort a successful fill.
         return
 
 
--- a/src/kicad_tools/cli/stitch_cmd.py
+++ b/src/kicad_tools/cli/stitch_cmd.py
@@ -4347,16 +4347,26 @@ def _carve_foreign_copper_after_stitch(pcb_path: Path) -> None:
     ``filled_polygon`` (the segment branch added in issue #3773), so it
     cleanly removes the new GND copper from the rail pour.  It edits the
     parsed S-expression directly (no kicad-cli refill, so the post-route
-    carve is preserved) and is a no-op when shapely is unavailable.
+    carve is preserved).  shapely is a core dependency (issue #3824); if it
+    is ever unavailable the carve cannot run and we warn **loud** that the
+    fill is NOT clearance-correct rather than implying the carve was merely
+    an optional skip.
     """
-    try:
-        from kicad_tools.core.sexp_file import load_pcb, save_pcb
-        from kicad_tools.zones.fill_clearance import apply_foreign_pad_clearance
-    except ImportError:
-        return
+    from kicad_tools.core.sexp_file import load_pcb, save_pcb
+    from kicad_tools.zones.fill_clearance import apply_foreign_pad_clearance
+
     try:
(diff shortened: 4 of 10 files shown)


rjwalters:
✅ **Approved!**

Thorough review against the four verification criteria for issue #3824 — all confirmed.

**1. Dependency change is correct.** `shapely>=2.0` is promoted into `[project] dependencies` with a clear comment justifying it as load-bearing for the correctness-critical copper paths, added explicitly to `[all]`, the redundant `[dev]` pin removed, and `[geometry]` kept as a back-compat alias. `uv.lock` was genuinely regenerated (not hand-edited): `uv lock --check` passes (Resolved 247 packages, exit 0), and the diff is exactly what `uv lock` produces — shapely added to core deps + `all`, removed from `dev`, with the `requires-dist` markers reconciled. All 14 CI checks pass including `Test` (14m22s), `Lint & Format`, and `Type Check`, so the lockfile/dep install is valid in CI.

**2. The fail-loud invariant is real.** `apply_foreign_pad_clearance()` and `force_solid_on_isolated_island_pads()` now call `require_shapely(...)` and `raise` instead of `return 0`. The three CLI callers no longer swallow it — and critically, each places `except ModuleNotFoundError` *before* the blanket `except Exception` (Python matches first), so the raise is no longer masked:
- `runner._apply_foreign_pad_clearance` and the `_remediate_starved_thermal` loop re-`raise` on `ModuleNotFoundError`.
- `stitch_cmd._carve_foreign_copper_after_stitch` prints a loud stderr WARNING that the fill is NOT clearance-correct / not fab-ready, then returns.
Verified `fill_clearance.py` has no module-level shapely import, so moving the `from ... import apply_foreign_pad_clearance` out of the try block in stitch_cmd does not introduce a new unguarded import path. No remaining path silently returns a non-clearance-correct fill and reports it clean.

**3. DRC degradation is actionable.** Both `SegmentZoneClearanceRule.check` and `ViaZoneClearanceRule.check` call `require_shapely(...)` up front, yielding the actionable "install kicad-tools[geometry]" message instead of a raw `ModuleNotFoundError`. The downstream bare `from shapely import ...` in private helpers (`_repair_fill_polygon`, `_collect_zone_fills`) is reachable only through these two now-guarded entry points, so the defense-at-entry pattern is sound. `connectivity.py` cleanly unifies onto the shared `has_shapely()` while preserving its legitimate pure-Python pour-grouping fallback.

**4. No over-broad change / no weakened tests.** The one renamed test (`test_no_shapely_is_noop` → `test_no_shapely_fails_loud`) was strengthened, not weakened — it now asserts a raise with the actionable hint rather than `== 0`. New `tests/test_shapely_core_dependency.py` (10 tests) covers the shared guard, both fail-loud carve paths, both DRC rules, the empty-zoneset legitimate-noop short-circuit, and the packaging contract. Shapely-present behavior is unchanged. Local runs: `test_shapely_core_dependency.py` + `test_zones_fill_clearance.py` → 48 passed; broader `-k "clearance or connectivity or stitch or shapely"` → 1781 passed, 4 skipped, 0 failures.

Merge state CLEAN, `Closes #3824` present. Excellent fail-loud-with-defense-in-depth implementation.

## Test Execution

1. \`pytest tests/test_shapely_core_dependency.py\` — ✅ Executed: 10 passed (run alongside fill_clearance suite, 48 total)
2. \`pytest -k 'clearance or zone or connectivity or shapely or fill or stitch'\` — ✅ Executed (subset): 1781 passed, 4 skipped, 0 failures
3. ruff check / format — ✅ CI \`Lint & Format\` passes
4. mypy — ✅ CI \`Type Check\` passes
5. \`uv lock --check\` — ✅ Executed: lockfile consistent (exit 0)

PR #3835, opened 2026-06-21
fix(drc): model pads as true roundrect/oval geometry, not bounding boxes

### LINKED ISSUE #3826: drc: clearance rule models pads as axis-aligned bounding boxes, not true roundrect/oval geometry -> sub-10um false-positive shorts
## Original Issue

Discovered during the clean-rebuild sweep (board-02, board-04). kct check reports clearance_pad_zone / clearance_pad_pad shorts at ~0.008-0.018mm overlap depth that native kicad-cli pcb drc does NOT flag (PASS-but-dirty in reverse: kct over-reports).

Root cause: src/kicad_tools/validate/rules/clearance.py models every pad as an axis-aligned shapely.box from _transform_pad_dimensions (~line 145), used at ~lines 1177-1182, ignoring pad shape. roundrect / oval pads are over-approximated at their rounded corners, producing sub-10-micron phantom overlaps that KiCad's true rounded geometry does not. (The via path at ~line 1150 correctly uses a circular buffer; the pad path does not.)

Evidence: board-02 run reported 4 clearance_pad_zone shorts at 0.008mm depth (26 roundrect + 4 oval pads) while kicad-cli reported 0; board-04 similar. copper-LVS also reported 0 shorts.

Fix: build pad polygons that respect pad shape (roundrect corner radius, oval/obround capsule) instead of a bounding box, so kct clearance geometry matches KiCad. Acceptance: the phantom sub-10um pad clearance shorts on boards 02/04 disappear and kct agrees with kicad-cli; genuine shorts still detected.

---

## Curator Enhancement

> Curated on 2026-06-21. Line numbers below verified against `origin/main` (shapely is now a core dep as of commit `63de0983`, PR #3833; the shared guard module `src/kicad_tools/_shapely.py` exists on main). The issue's stated line numbers (~145, ~1150, ~1177) match current `origin/main`.

### Problem Statement

`kct check` over-reports copper shorts relative to the authoritative `kicad-cli pcb drc`. The clearance DRC code approximates every pad as an axis-aligned rectangle (AABB). For `roundrect` and `oval`/`obround` pads, the AABB pokes out past the true copper at the rounded corners by up to ~(1 - cos45deg) * corner_radius. When such a corner sits next to a foreign-net zone fill (or another pad) at the design's minimum clearance, the phantom corner triangle produces a sub-10-micron "overlap" that KiCad's true rounded geometry never sees. The result is a stream of false-positive `clearance_pad_zone` and `clearance_pad_pad` shorts (board-02: 4 phantom `pad_zone` shorts at 0.008mm; board-04 similar) that block clean-rebuild gating even though copper-LVS and `kicad-cli` both report zero shorts.

This matters because the clean-rebuild sweep treats kct DRC as a gate; a false positive there is indistinguishable from a real short to the gate, so it stalls board graduation and erodes trust in `kct check`.

### Root Cause: two distinct AABB sites (the issue undercounts)

The issue body names the `pad_zone` site (`ViaZoneClearanceRule`). Verification found a **second, independent AABB site** producing the `pad_pad` false positives the title also mentions. Both must be fixed:

1. **Pad-vs-zone (`clearance_pad_zone`)** — `ViaZoneClearanceRule.check` at `src/kicad_tools/validate/rules/clearance.py:1177-1182` builds the pad copper as `shapely.box(...)`. The sibling via path at `:1150` correctly uses `Point(...).buffer(radius)` (circular). This is the site the issue names.

2. **Pad-vs-pad (`clearance_pad_pad`)** — `ClearanceRule` does NOT use shapely at all on the pad path. It routes through `_circle_circle_clearance` (`:435`) -> `_rect_rect_clearance` (`:474`) / `_rect_circle_clearance` (`:513`), which are pure-Python AABB / AABB-to-circle analytic formulas. `CopperElement.from_pad` (`:92`) stores only `(abs_x, abs_y, width, height)` from `_transform_pad_dimensions`, so the rounded shape is already discarded before the math runs. Fixing only site (1) will leave the `pad_pad` false positives in place.

Common upstream cause for both: `_transform_pad_dimensions` (`:145`) returns only an AABB `(width, height)` and the `Pad` model never carries the data needed to reconstruct the true outline.

### Model gap: Pad does not carry shape parameters (prerequisite work)

`src/kicad_tools/schema/pcb.py` `Pad.from_sexp` (class at `:153`, parser at `:173`) parses `shape` (`:179` -> "roundrect"|"rect"|"circle"|"oval") and `size` but **drops two fields the fix needs**:

- **`roundrect_rratio`** is never parsed. Without it the corner radius is unknown. KiCad's default is `0.25` (radius = ratio * min(w, h)). The builder must add a `roundrect_rratio: float = 0.25` field and parse `(roundrect_rratio N)` in `from_sexp`.
- **Per-pad rotation** is dropped: `from_sexp` reads `(at x y angle)` but only stores `position` (`:186-189`), discarding the optional third token (the pad's own rotation relative to the footprint). `_transform_pad_dimensions` (`:166`) therefore uses `footprint.rotation` only. For an accurate oriented polygon the total rotation is `footprint.rotation + pad.rotation`. Builder should add `rotation: float = 0.0` to `Pad`, parse `at.get_float(2)`, and use the sum.

Confirm before relying on these: `grep -n "roundrect_rratio\|rotation" src/kicad_tools/schema/pcb.py` (today: roundrect_rratio absent; pad rotation absent). The existing `footprints/__init__.py` writer already emits `roundrect_rratio 0.25`, confirming the field name/semantics.

### Implementation Guidance

Introduce a single shared helper that returns a true-geometry shapely polygon for a pad in board coordinates, and use it from BOTH sites so they can never drift again.

**1. Shared pad-polygon builder** (new function in `clearance.py`, near `_transform_pad_dimensions`):

```python
def _pad_polygon(pad: Pad, footprint: Footprint):
    """Return a shapely polygon of the pad copper in board coordinates,
    honoring pad.shape (rect / roundrect / circle / oval|obround)."""
    require_shapely("pad clearance geometry")  # from kicad_tools._shapely
    import shapely
    from shapely.affinity import rotate, translate
    from shapely.geometry import Point

    cx, cy = _transform_pad_position(pad, footprint)
    w, h = pad.size  # LOCAL size (do NOT use the AABB transform here)
    if w <= 0 or h <= 0:
        return None
    total_rot = (footprint.rotation + getattr(pad, "rotation", 0.0))

    shape = pad.shape
    if shape == "circle" or (shape in ("oval", "obround") and abs(w - h) < 1e-6):
        return Point(cx, cy).buffer(min(w, h) / 2.0)
    if shape in ("oval", "obround"):
        # stadium / capsule: rectangle core + semicircle caps = rect buffered
        # by the minor half-axis, with the rect's long axis shortened by it.
        r = min(w, h) / 2.0
        core = shapely.box(-(w / 2.0 - r) if w >= h else -r,
                           -(h / 2.0 - r) if h > w else -r,
                            (w / 2.0 - r) if w >= h else r,
                            (h / 2.0 - r) if h > w else r)
        poly = core.buffer(r)  # round join gives true stadium ends
    elif shape == "roundrect":
        rr = getattr(pad, "roundrect_rratio", 0.25)
        r = rr * min(w, h)
        inner = shapely.box(-(w / 2.0 - r), -(h / 2.0 - r),
                             (w / 2.0 - r),  (h / 2.0 - r))
        poly = inner.buffer(r, join_style=1)  # round joins -> rounded corners
    else:  # "rect" and any unknown shape -> exact rectangle (no over-approx)
        poly = shapely.box(-w / 2.0, -h / 2.0, w / 2.0, h / 2.0)

    poly = rotate(poly, total_rot, origin=(0, 0), use_radians=False)
    return translate(poly, cx, cy)
```

Notes:
- The stadium construction above is illustrative; the simplest correct stadium is `shapely.box(core_rect).buffer(r, cap_style=1, join_style=1)` where the core rect is the size shrunk by `r` on the minor axis only. Builder should pick whichever formulation passes the geometry tests and keep it readable.
- Use `buffer` `quad_segs` default (8) — that is ~accurate to << 1um at these radii; do not crank it up (perf).
- `chamfered` roundrect pads exist in KiCad but are rare in these boards; out of scope unless a test board uses them. If `chamfer` is present, falling back to the un-chamfered roundrect is a safe over-approximation only if it does not reintroduce false positives — note as a follow-up, do not block on it.

**2. Wire `pad_zone` site** — replace the `shapely.box(...)` block at `:1177-1182` with `shape = _pad_polygon(pad, fp)` (skip if `None`). The existing `inflate=0.0` and `_query_and_check`/`_check_shape` distance logic already treats `shape` as the full copper outline, so no other change is needed there.

**3. Wire `pad_pad` site** — the cleanest fix is to make the pad path in `ClearanceRule` use shapely polygons too. Options:
   - **Option A (recommended):** Give `CopperElement` an optional precomputed `polygon` for pads, built via `_pad_polygon` in `from_pad`. In `_calculate_clearance`/`_circle_circle_clearance`, when both elements have polygons (or one polygon + one circle), use `shapely` `distance`/`intersection` instead of the analytic AABB formulas. Keep the analytic path for via-via (pure circles) to avoid perf regressions.
   - **Option B (smaller, less complete):** Keep analytic math but special-case the roundrect/oval corner by insetting the effective rectangle — fragile and still approximate. Not recommended.
   Prefer Option A so both sites share `_pad_polygon` and geometry can't diverge.

**4. shapely guard** — both sites are correctness-critical with no valid pure-Python fallback for rounded pads; use `require_shapely(...)` from `kicad_tools._shapely` (do NOT silently degrade). shapely is already a core dep (PR #3833).

**Performance:** building a small buffered polygon per pad is cheap relative to the existing per-pad STRtree query. Keep the STRtree pre-filter (`_query_and_check` query box) so only nearby fills run the exact `distance`. Watch the pad-pad path: do not convert via-via (the dominant element count on dense boards) to shapely — keep those analytic.

### Affected Files

- `src/kicad_tools/validate/rules/clearance.py` - Add `_pad_polygon()` helper; replace `shapely.box` pad construction at `:1177-1182` (pad_zone) with it; route `clearance_pad_pad` (`CopperElement.from_pad` `:92`, `_circle_circle_clearance` `:435`, `_rect_rect_clearance` `:474`, `_rect_circle_clearance` `:513`) through true pad polygons (Option A). Import `require_shapely` from `kicad_tools._shapely`.
- `src/kicad_tools/schema/pcb.py` - Extend `Pad` (`:153`) with `roundrect_rratio: float = 0.25` and `rotation: float = 0.0`; parse `(roundrect_rratio N)` and the third `(at x y angle)` token in `Pad.from_sexp` (`:173`).
- `tests/test_validate_via_zone_clearance.py` - Add regression tests for roundrect/oval pad-vs-zone (see Test Plan). May add a new `tests/test_pad_polygon_geometry.py` for the helper unit tests and pad-pad cases.

### Test Plan

- [ ] Automated unit: `_pad_polygon` returns expected outlines — a roundrect pad's corner area is strictly less than its AABB; an oval pad with `w != h` is a stadium (area between inscribed circle and bounding rect); a circle pad equals `Point.buffer`; a `rect` pad equals exact `shapely.box`. Verify rotation (e.g. 45deg and 90deg footprint rotation, plus a non-zero per-pad `at` angle) places copper correctly.
- [ ] Automated regression (the bug): construct a roundrect pad positioned so its AABB corner overlaps a foreign-net zone fill by ~0.008mm but its true rounded corner clears it. Assert the OLD code reported a `clearance_pad_zone` short and the NEW code reports none. Mirror for an oval pad.
- [ ] Automated guard (no regression on real shorts): a pad whose true (rounded) copper genuinely overlaps a foreign zone fill / another pad MUST still produce a short. A roundrect pad overlapping deep in the body (not just the phantom corner) must still fail.
- [ ] Automated pad-pad: two roundrect pads whose AABBs touch at a corner but whose rounded corners clear must report no `clearance_pad_pad`; two pads with real body overlap must still report it.
- [ ] Manual / e2e verification: run `kct check` on `boards/02-charlieplex-led/output/charlieplex_3x3_routed.kicad_pcb` and `boards/04-stm32-devboard/output/stm32_devboard_routed.kicad_pcb`. The previously-reported sub-10um `clearance_pad_zone`/`clearance_pad_pad` shorts disappear, and kct agrees with `kicad-cli pcb drc` (0 copper shorts) on both boards.
- [ ] Performance sanity: `kct check` wall-clock on the densest board (e.g. board-04) does not regress materially (no per-pad shapely on the via-via path; STRtree pre-filter retained).
- [ ] Full suite: `uv run pytest tests/test_validate_via_zone_clearance.py` plus the new test module pass; `uv run ruff check . && uv run mypy src/` clean.

### Edge cases

- Cardinal vs arbitrary footprint rotation (90/270 swap vs 45deg AABB) — the new path rotates the actual polygon so the old `_transform_pad_dimensions` swap logic is no longer on the pad-polygon path; keep `_transform_pad_dimensions` only where still referenced.
- Square roundrect/oval where `w == h` (oval degenerates to circle).
- `roundrect_rratio` missing in the s-expr (older boards) -> default 0.25.
- Through-hole pads on inner layers (`*.Cu`) — `_pad_on_layer` already handles layer membership; geometry is layer-agnostic.
- Chamfered roundrect — out of scope; note as follow-up if a test board needs it.

### Related

- Depends on shapely-as-core-dep + `src/kicad_tools/_shapely.py` (landed via PR #3833, commit `63de0983`). No open dependency — these are on `origin/main`.

rjwalters:
## Summary

`kct check` over-reported copper shorts relative to the authoritative `kicad-cli pcb drc`. The clearance DRC modeled every pad as an axis-aligned bounding box (AABB), so `roundrect` / `oval` pad corners poked past the true copper by up to ~(1 - cos45)*corner_radius. When such a phantom corner sat next to a foreign-net zone fill (or another pad) at the design minimum clearance, it produced sub-10-micron false-positive `clearance_pad_zone` / `clearance_pad_pad` shorts that KiCad's true rounded geometry never sees — stalling the clean-rebuild gate on boards with rounded pads (board-02: 26 roundrect + 4 oval pads).

This fixes BOTH independent AABB sites (the curator confirmed there were two) with a single shared true-geometry helper so they can never drift again.

## Changes

- `src/kicad_tools/schema/pcb.py` — `Pad` now parses the two fields the fix needs and that were previously dropped:
  - `roundrect_rratio` (corner-radius ratio; KiCad default `0.25`)
  - per-pad `rotation` (the optional third token of `(at x y angle)`)
- `src/kicad_tools/validate/rules/clearance.py`
  - New shared `_pad_polygon()` helper builds the true shapely outline per shape: roundrect (rounded corners), oval/obround (stadium/capsule), circle (disc), exact rect — including `footprint.rotation + pad.rotation`.
  - **Pad-vs-zone** (`ViaZoneClearanceRule`): replaced the `shapely.box` pad construction with `_pad_polygon`.
  - **Pad-vs-pad** (`ClearanceRule`): `CopperElement` now carries a precomputed pad polygon; the pad path routes through exact shapely distance/intersection (`_polygon_pair_clearance`) instead of the analytic AABB formulas. Via-via stays on the fast analytic disc path (no perf regression on dense boards).
  - Uses `require_shapely(...)` (no silent degrade) since there is no correct pure-Python fallback for rounded pads.
- Tests — new `tests/test_pad_polygon_geometry.py` (helper geometry + rotation + pad-pad) and added a `TestRoundedPadCornerNoPhantomShort` class to `tests/test_validate_via_zone_clearance.py`.

| Roundrect/oval pad near a zone that false-positived now PASSES | ✅ | `TestRoundedPadCornerNoPhantomShort::test_roundrect/oval_pad_corner_clears_no_phantom_short` |
| Roundrect/oval pad near another pad that false-positived now PASSES | ✅ | `TestPadPadClearance::test_roundrect_corner_clears_no_false_overlap` |
| Genuine overlap still FAILS (zone + pad) | ✅ | `test_roundrect_genuine_body_overlap_still_shorts`, `test_real_body_overlap_still_negative` |
| `_pad_polygon` exact outlines (roundrect < AABB, oval stadium, circle disc, rect exact, rotation) | ✅ | `TestPadPolygonShapes`, `TestPadPolygonRotation` |
| Pad model parses `roundrect_rratio` + per-pad rotation | ✅ | Boards 02/04 load 30/79 rounded pads with `rratio=0.25` parsed |
| kct agrees with kicad-cli on board 02 | ✅ | Both report 0 copper shorts (`kct check` 0 violations; `kicad-cli pcb drc` "Found 0 violations") |
| Genuine detections preserved on board 04 | ✅ | Baseline vs new code report identical real-violation counts at tightened clearances |
| Perf sanity (no via-via shapely) | ✅ | Via-via remains on analytic path; STRtree pre-filter retained |

## Test Plan

- `uv run pytest tests/test_pad_polygon_geometry.py tests/test_validate_via_zone_clearance.py tests/test_validate_clearance_rect_pad.py tests/test_drc_pad_clearance_epsilon.py tests/test_validate_segment_zone_clearance.py tests/test_same_component_clearance.py` → 132 passed
- `uv run pytest tests/test_schema.py` → 162 passed (Pad model changes are backward-compatible)
- `uv run ruff check` / `ruff format --check` on touched files → clean
- mypy baseline gate: zero new errors from this PR (the one remaining `recovery/strategy.py` finding pre-exists on clean `origin/main` and is unrelated)
- e2e: `kct check` on boards 02/04 + `kicad-cli pcb drc` parity confirmed

Closes #3826

### DIFF
--- a/src/kicad_tools/schema/pcb.py
+++ b/src/kicad_tools/schema/pcb.py
@@ -165,6 +165,13 @@ class Pad:
     drill: float = 0.0
     solder_mask_margin: float | None = None
     uuid: str = ""
+    # Per-pad rotation in degrees (the optional third token of ``(at x y angle)``),
+    # relative to the parent footprint. The pad's absolute rotation is
+    # ``footprint.rotation + pad.rotation``. Defaults to 0.
+    rotation: float = 0.0
+    # Corner-radius ratio for ``roundrect`` pads: radius = rratio * min(w, h).
+    # KiCad's default is 0.25 when ``(roundrect_rratio ...)`` is absent.
+    roundrect_rratio: float = 0.25
 
     @property
     def net(self) -> int:
@@ -183,17 +190,24 @@ class Pad:
             layers=[],
         )
 
-        # Position
+        # Position (and optional per-pad rotation as the third token)
         if at := sexp.find("at"):
             x = at.get_float(0) or 0.0
             y = at.get_float(1) or 0.0
             pad.position = (x, y)
+            pad.rotation = at.get_float(2) or 0.0
 
         # Size
         if size := sexp.find("size"):
             w = size.get_float(0) or 0.0
             h = size.get_float(1) or w
             pad.size = (w, h)
+
+        # Corner radius ratio for roundrect pads (default 0.25 when absent)
+        if rratio := sexp.find("roundrect_rratio"):
+            value = rratio.get_float(0)
+            if value is not None:
+                pad.roundrect_rratio = value
 
         # Layers
         if layers := sexp.find("layers"):
--- a/src/kicad_tools/validate/rules/clearance.py
+++ b/src/kicad_tools/validate/rules/clearance.py
@@ -77,6 +77,14 @@ class CopperElement:
     # avoid re-reporting the same geometric pair on every spanned layer
     # (the endpoint-layer scan already covers them).
     explicit_layers: tuple[str, ...] = ()
+    # For pads: a precomputed shapely polygon of the TRUE copper outline
+    # (roundrect/oval honored) in board coordinates.  When two elements
+    # both carry a polygon -- or one polygon plus one circular via -- the
+    # clearance is computed via exact shapely geometry instead of the
+    # analytic axis-aligned-rectangle formulas, which over-approximate
+    # rounded corners and produce phantom sub-10um shorts (issue #3826).
+    # ``None`` for segments and vias (which use the analytic disc path).
+    polygon: object | None = None
 
     @classmethod
     def from_segment(cls, seg: Segment) -> CopperElement:
@@ -95,8 +103,13 @@ class CopperElement:
         """Create from a PCB pad with footprint context."""
         # Transform pad position from footprint-local to board coordinates
         abs_x, abs_y = _transform_pad_position(pad, footprint)
-        # Transform pad dimensions to axis-aligned bounding box
+        # Transform pad dimensions to axis-aligned bounding box (retained
+        # for the AABB-bounds fast path / location reporting).
         width, height = _transform_pad_dimensions(pad, footprint)
+        # Build the TRUE copper outline (roundrect/oval honored) so the
+        # pad-pad clearance path can use exact geometry instead of the
+        # over-approximating analytic AABB formulas (issue #3826).
+        polygon = _pad_polygon(pad, footprint)
         return cls(
             element_type="pad",
             layer="*",  # Pads can span multiple layers
@@ -104,6 +117,7 @@ class CopperElement:
             geometry=(abs_x, abs_y, width, height),
             reference=f"{footprint.reference}-{pad.number}",
             net_name=pad.net_name if pad.net_number != 0 else "",
+            polygon=polygon,
         )
 
     @classmethod
@@ -186,6 +200,74 @@ def _transform_pad_dimensions(pad: Pad, footprint: Footprint) -> tuple[float, fl
     new_height = width * sin_a + height * cos_a
 
     return new_width, new_height
+
+
+def _pad_polygon(pad: Pad, footprint: Footprint):
+    """Return a shapely polygon of a pad's copper in board coordinates.
+
+    Unlike :func:`_transform_pad_dimensions` (which returns only an
+    axis-aligned bounding box and therefore over-approximates rounded
+    corners), this builds the *true* copper outline honoring
+    ``pad.shape``:
+
+    * ``circle`` (and ``oval``/``obround`` with ``w == h``) -> exact disc.
+    * ``oval`` / ``obround`` -> stadium / capsule (rectangle core with
+      semicircular caps on the minor axis).
+    * ``roundrect`` -> rectangle with rounded corners of radius
+      ``roundrect_rratio * min(w, h)`` (KiCad default ratio 0.25).
+    * ``rect`` (and any unknown shape) -> exact rectangle, with no
+      over-approximation.
+
+    The polygon is built around the origin from the pad's *local* size,
+    rotated by the total rotation (``footprint.rotation + pad.rotation``)
+    and translated to the pad's absolute board position.  Modeling the
+    rounded geometry instead of an AABB removes the sub-10-micron phantom
+    corner overlaps that KiCad's true geometry never sees (issue #3826).
+
+    Returns ``None`` for degenerate (non-positive) sizes so callers can
+    skip them.
+    """
+    require_shapely("pad clearance geometry")
+    import shapely
+    from shapely.affinity import rotate, translate  # type: ignore[import-untyped]
+    from shapely.geometry import Point
+
+    cx, cy = _transform_pad_position(pad, footprint)
+    w, h = pad.size  # LOCAL size -- the rotation below orients the polygon
+    if w <= 0 or h <= 0:
+        return None
+
+    total_rot = footprint.rotation + getattr(pad, "rotation", 0.0)
+    shape = pad.shape
+
+    if shape == "circle" or (shape in ("oval", "obround") and abs(w - h) < 1e-6):
+        # Symmetric disc -- rotation and the (origin) translate are no-ops
+        # on a circle, so build it directly at the board position.
+        return Point(cx, cy).buffer(min(w, h) / 2.0)
+
+    if shape in ("oval", "obround"):
+        # Stadium / capsule: a core rectangle shrunk by the minor
+        # half-axis on the long dimension, buffered by that radius so the
+        # short ends become true semicircles.
+        r = min(w, h) / 2.0
+        if w >= h:
+            core = shapely.box(-(w / 2.0 - r), 0.0, (w / 2.0 - r), 0.0)
+        else:
+            core = shapely.box(0.0, -(h / 2.0 - r), 0.0, (h / 2.0 - r))
+        poly = core.buffer(r)
+    elif shape == "roundrect":
+        rr = getattr(pad, "roundrect_rratio", 0.25)
+        r = rr * min(w, h)
+        if r <= 0:
+            poly = shapely.box(-w / 2.0, -h / 2.0, w / 2.0, h / 2.0)
+        else:
+            inner = shapely.box(-(w / 2.0 - r), -(h / 2.0 - r), (w / 2.0 - r), (h / 2.0 - r))
+            poly = inner.buffer(r, join_style=1)  # round joins -> rounded corners
+    else:  # "rect" and any unknown shape -> exact rectangle (no over-approx)
+        poly = shapely.box(-w / 2.0, -h / 2.0, w / 2.0, h / 2.0)
+
+    poly = rotate(poly, total_rot, origin=(0, 0), use_radians=False)
+    return translate(poly, cx, cy)
 
 
 # _point_to_segment_distance and _segment_to_segment_distance are imported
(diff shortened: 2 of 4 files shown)


rjwalters:
✅ **Approved!**

This is a clean, well-scoped, and correct fix for the AABB pad over-approximation that caused phantom sub-10µm `clearance_pad_zone` / `clearance_pad_pad` false positives. I verified it thoroughly.

## Correctness (geometry) — verified empirically
`_pad_polygon()` builds true outlines per shape, confirmed against closed-form areas:
- **roundrect**: corner radius `= roundrect_rratio * min(w, h)` (KiCad convention). 2×4 pad @ rratio 0.25 → area 7.784 (expected `8 - (4-π)·0.5²` = 7.785 ✓); rratio honored (rr=0.5 → 7.137 ✓).
- **oval/obround**: stadium/capsule, both orientations (3×1 horiz and 1×3 vert both match `(L-2r)·w + πr²` ✓); degenerates to disc when `w==h`.
- **circle**: exact disc. **rect / unknown**: exact box (no over-approx).
- **rotation**: `footprint.rotation + pad.rotation` applied; the new per-pad `rotation` field (3rd `(at …)` token) is wired correctly.

## Both AABB sites fixed
- **pad-zone** (`ViaZoneClearanceRule`): `shapely.box` → `_pad_polygon`.
- **pad-pad** (`ClearanceRule` → `_circle_circle_clearance` → `_polygon_pair_clearance`): exact shapely distance/intersection when either element carries a pad polygon.
- A fix to only one site would have left the other class of false positive; both are covered.

## Genuine shorts preserved + perf
- Unit tests prove old-AABB-flags / new-geometry-clears (rect control shorts, roundrect/oval clear) AND deep body overlaps still short (pad-zone and pad-pad).
- **Perf**: via-via stays on the analytic disc path (`from_via` sets no polygon → guard skips shapely); STRtree pre-filter retained. No O(n²) shapely blowup on dense via fields. ✓

## Schema round-trip
`Pad` is a parse-only dataclass; footprint serialization is sexp-node-backed, so the new read-only fields don't affect write-back. 162 `test_schema.py` tests pass.

## Test Execution
- `pytest test_pad_polygon_geometry + via_zone + clearance_rect_pad + drc_epsilon + segment_zone + same_component + schema` → **252 passed**.
- ruff check / format on touched files → **clean**.
- mypy baseline gate: no new errors in `clearance.py` / `pcb.py` (the shapely `import-untyped` notes pre-exist on `main`; CI **Type Check** job is green). The 2 local-only baseline-drift findings are in untouched files (`recovery/strategy.py`, `cli/build_native_cmd.py`).
- **End-to-end parity (the acceptance criterion):**
  - Board 02 routed: `kct check` → 0 violations; `kicad-cli pcb drc` → "Found 0 violations". **Agreement ✓**
  - Board 04 routed: zero `clearance_pad_*` findings (the only kct errors are pre-existing `via_in_pad`/`connectivity`, out of scope). No phantom clearance shorts remain.

## CI note (non-blocking)
All checks are green **except** `Board 05 Routing Regression (blocking <= 9)`, which failed twice (10 blocking nets). This is **not** caused by this PR:
- The PR touches only DRC clearance geometry — **zero** router code.
- The branch is fully up to date with `main` (0 behind; merge-base == `main` tip `63de0983`).
- The latest `main` run with that exact router code **passed** this gate (≤9), while two re-routes on this branch hit 10 — i.e. pure router nondeterminism at the 9/10 floor, which the gate's own message documents (#3775/#3766/#3822: committed=7, CI=9, local≈11).
- Filed **#3836** to track stabilizing/adjusting the gate.

Since the failing gate is a documented-flaky, PR-independent router floor that this change cannot affect, it is not a blocker for this DRC-geometry fix. Recommend merging; the human merger should be aware of #3836.

Issue #3841, opened 2026-06-21
board-03: GND stitching vias do not fire on a clean regen (the #3787 fix does not survive), stranding J1 shield pads


rjwalters:
## Original Issue

Child of epic #3829.

On a clean-room regen of board-03 (boards/03-usb-joystick/generate_design.py), add_gnd_stitching_vias (~line 265+ / called ~line 760) prints "GND is already a single copper island -- no vias needed" and adds 0 stitching vias, where the committed board has the 2 GND stitch vias (the #3787 fix) that bond J1s F.Cu-only shield pads (A12/B1). Result: fresh regen leaves J1.A12 / J1.B1 stranded -> copper-LVS dirty (2 GND opens) and native kicad-cli reports 14 unconnected GND/VCC items. The committed board is clean; the recipe no longer reproduces it.

Root cause hypothesis: the GND island-fragmentation detector concludes GND is one island (so skips stitching) while J1s F.Cu-only shield pads are actually not bonded to the GND pour on a fresh fill. The detector runs against a throwaway filled copy whose fill differs from the persisted board (related to the zone-fill nondeterminism seen elsewhere).

Fix: make add_gnd_stitching_vias correctly detect the stranded F.Cu-only shield pads on a fresh regen and add the bonding vias deterministically, so a clean regen reproduces the committed clean board (J1.A12/B1 bonded to GND).

Acceptance: a fresh board-03 regen is copper-LVS clean (0 opens) and native kicad-cli reports 0 unconnected GND items; the 2 GND stitch vias are present after regen. Regression test on a fresh regen.

Files: boards/03-usb-joystick/generate_design.py (add_gnd_stitching_vias).

---

## Curator Enhancement (2026-06-21)

> Curating per the epic-#3829 reproducibility chain. Verified against `origin/main` and by running the recipe + detector logic directly. Sibling issues: **#3838** (the recipe's own copper-LVS gate reports clean even when a fresh out-of-process check on the same persisted bytes reports dirty) and **#3840** (e2e gates should validate a fresh regen out-of-process). This issue (#3841) is the *underlying board defect*: on a clean regen the #3787 GND stitching vias can fail to fire, leaving J1's F.Cu-only shield pads stranded.

### Problem Statement

[...]

Issue #3840, opened 2026-06-21
ci: board e2e gates should validate a fresh regen RESULT out-of-process, not assert the committed file or recipe self-report


rjwalters:
## Original Issue

Child of epic #3829. DEPENDS ON the in-process-vs-out-of-process LVS fix (sibling child) landing first.

Today the board CI gates are inconsistent: boards 00/01/02 regenerate-and-validate the result; board-05/06/07 re-route and gate a freshly measured metric; but board-03/06/07 --lvs-only jobs gate on lvs.json which is the recipes OWN self-report (only as trustworthy as that self-report). The board-03 case proves the self-report can be wrong (sibling issue: in-process LVS divergence).

Fix: make the board CI gates that currently trust the committed lvs.json / recipe self-report instead re-run compare_copper_netlist (and DRC) OUT-OF-PROCESS against the regenerated artifact and assert THAT result. This composes with the LVS-divergence fix (once the self-report is trustworthy, the out-of-process gate confirms it independently). Use the #3823 board-05 job and board-00/01/02 e2e asserters as templates.

Acceptance: each board gate independently re-validates the regenerated board out-of-process (DRC + copper-LVS); a board whose recipe self-report is wrong is caught by the gate. No board gate passes solely on the recipes own in-process verdict.

Files: .github/workflows/ci.yml; scripts/ci/*.py (gate asserters).

---

### Status update: the premise shifted when #3838 / PR #3845 landed

This issue was filed assuming the board-03/06/07 `--lvs-only` CI gates trust an
*in-process* recipe self-report (`output/lvs.json`) that can diverge from a fresh
out-of-process check. **That gap is now largely closed for copper-LVS.**

PR #3845 (merged 2026-06-21, closes #3838) made the recipe's gated copper result
**authoritative against the on-disk bytes**: `write_lvs_report(..., fresh_copper_check=True)`
(default True) re-derives the copper verdict in a *fresh subprocess*
(`python -m kicad_tools.lvs.copper_lvs <sch> <pcb>`) via `_fresh_copper_compare` /
`_authoritative_copper_result` in `src/kicad_tools/lvs/recipe.py`, and **fails closed**
(synthetic `open` mismatch -> `clean=false` -> raise) if the in-process and fresh
results disagree. So the `lvs.json` the gate reads is now the *fresh out-of-process*
copper result, not a naive in-process self-report.

Net effect: a board whose in-process copper self-report is wrong is **already caught**
[...]

PR #3845, opened 2026-06-21
fix(lvs): make copper-LVS gate authoritative via fresh out-of-process re-check

### LINKED ISSUE #3838: lvs: recipe in-process copper-LVS self-report diverges from a fresh out-of-process check (masks real defects)
## Original Issue

Child of epic #3829.

The board recipes gate on copper-LVS via write_lvs_report (src/kicad_tools/lvs/recipe.py ~lines 120-147, require_clean=True) which runs compare_copper_netlist IN-PROCESS and writes lvs.json. On board-03 the in-process self-report says clean=true while a FRESH out-of-process compare_copper_netlist on the SAME unchanged routed .kicad_pcb bytes reports DIRTY (2 GND opens; native kicad-cli DRC independently confirms stranded J1 shield pads). So the recipe exits 0 / lvs.json clean=true and the CI "assert lvs.json clean" gate passes on a physically broken board.

Root cause hypothesis: an in-process state dependency (e.g. the route step in-memory fill/zone state, or cached netlist) makes the in-process comparison see a different (cleaner) board than what was persisted to disk. The hard gate (require_clean=True / BoardNetlistMismatch) is defeated because it trusts the in-process result.

Fix: make the copper-LVS gate authoritative against the ON-DISK artifact — re-load the persisted .kicad_pcb in a fresh state (or a subprocess) before comparing, so the gated result equals what a fresh process / CI re-check sees. The in-process and fresh results must agree on identical bytes.

Acceptance: for board-03 (and all boards), the recipe LVS self-report matches a fresh out-of-process compare_copper_netlist on the persisted file; a board with stranded pads is caught by the recipe gate (require_clean raises) rather than passing. Add a regression test that loads a known-dirty persisted board and asserts the gate flags it.

Files: src/kicad_tools/lvs/recipe.py; boards/03-usb-joystick/generate_design.py (consumer).

---

## Curator Enhancement

> Curating per the epic-#3829 reproducibility chain. Verified against `origin/main` and by running the comparator directly. Sibling issues: #3840 (e2e gates should validate a fresh regen out-of-process) and #3841 (the #3787 GND stitching vias do not fire on a clean regen — i.e. the *underlying* board is genuinely dirty after a clean regen). This issue (#3838) is the narrower defect: **the recipe's own copper-LVS gate reports clean even though a fresh out-of-process check on the same persisted board reports dirty**, so the gate never trips and the dirty board (#3841) ships.

### Problem Statement

Every board recipe gates copper-LVS through `write_lvs_report(..., require_clean=True)` (`src/kicad_tools/lvs/recipe.py:79-149`). That function calls `compare_copper_netlist(sch_path, routed_pcb_path)` (line 120), ANDs the gated comparators into `gated_clean`, writes `output/lvs.json`, and raises `BoardNetlistMismatch` when dirty (line 142-147). Board-03's recipe calls it at `boards/03-usb-joystick/generate_design.py:785-792` (`require_clean=True, run_copper=True, run_label=False`).

The observed failure: during a fresh `generate_design.py` regen of board-03 the in-process self-report writes `clean=true`, yet a fresh out-of-process `compare_copper_netlist` on the **same persisted `usb_joystick_routed.kicad_pcb` bytes** reports DIRTY (2 GND opens — J1 USB-C F.Cu-only shield pads A12/B1 stranded), and native `kicad-cli` DRC independently confirms the stranded pads. The recipe therefore exits 0, `lvs.json` says `clean=true`, and the CI "assert lvs.json clean" gate passes on a physically broken board.

### Confirmed Root-Cause Analysis

`compare_copper_netlist` (`src/kicad_tools/lvs/copper_lvs.py:184-210`) is **already file-based** — it takes two paths and loads both fresh (`_schematic_pin_to_net(sch_path)` + `ConnectivityValidator(pcb_path).extract_pad_partition()`, where `ConnectivityValidator.__init__` does `PCB.load(str(pcb))`, `src/kicad_tools/validate/connectivity.py:267-278`). So the divergence is **not** in-memory-PCB vs on-disk-PCB at the comparator boundary. It is a **read-time-vs-state divergence within the recipe process**:

1. The recipe order is route -> `add_gnd_stitching_vias` -> `fill_zones_in_routed_pcb` -> `run_drc` -> `write_lvs_report` (`generate_design.py:750-792`). The copper partition that `extract_pad_partition` computes depends **entirely on `(filled_polygon ...)` geometry** in the routed file (`connectivity.py:496-548`, the label-free pour-bonding step 2d).
2. `fill_zones_in_routed_pcb` (`generate_design.py:226-262`) fills via `run_fill_zones`, which runs **post-fill pure-Python remediation/carve** (`_remediate_starved_thermal` + `_apply_foreign_pad_clearance`, `src/kicad_tools/cli/runner.py:407-413`). That carve edits `filled_polygon` geometry and is explicitly noted as something a fresh `kicad-cli` refill would revert (runner.py:401-406). So the fill geometry persisted by the recipe is a *Python-post-processed* fill — a fresh out-of-process re-check that re-derives islands from those exact bytes can legitimately partition GND differently than the recipe's own read did.
3. The recipe's gate trusts whatever `compare_copper_netlist` returns at `generate_design.py:785` — there is **no second, independent, fresh-process confirmation** that the persisted artifact a re-checker (or CI) will load yields the same partition. Any read-order, fill-settling, or shapely-availability difference between the recipe process and a fresh process slips straight through the `require_clean` gate.

The reproduction-relevant fact (see #3841): on a clean regen the GND stitching vias of #3787 do **not** fire, so the persisted board genuinely has 2 GND opens. The defect *this* issue fixes is that `write_lvs_report` should have **caught** that — instead its in-process read disagreed with the authoritative on-disk state and reported clean.

Note: `compare_copper_netlist`'s pour leg is shapely-gated (`connectivity.py:493, 545-548`); when shapely is absent it falls back to **legacy declared-net pour grouping**, which trusts labels and will mask copper opens entirely. If the recipe environment has shapely but a fresh check environment does not (or vice-versa), that alone produces a clean/dirty divergence on identical bytes. The fix must make the gate authoritative regardless of which leg ran.

### Implementation Guidance

Make the copper-LVS gate authoritative against the **on-disk artifact exactly as a fresh process / CI re-check will load it**. Two viable approaches:

**Option A (recommended): re-run the gated copper compare in a subprocess.**
In `write_lvs_report`, when `run_copper` and `require_clean` are set, compute `copper_result` by spawning a fresh `python -c` (or a small `kicad_tools.cli` entrypoint) that runs `compare_copper_netlist(sch_path, routed_pcb_path)` on the persisted file and emits JSON. Parse that JSON for the gate decision. This guarantees the gated result equals what CI's fresh re-check sees (byte-for-byte, with a fresh interpreter / fresh shapely-availability / no in-process settling state). Pros: bulletproof equivalence; same code path as CI. Cons: subprocess overhead (~1 import of the PCB stack) per board build.

**Option B: re-load and re-fill deterministically before the gated compare.**
Before comparing, copy the persisted routed PCB to a temp file, run a fresh `kicad-cli` fill on that copy (no Python carve), and run `compare_copper_netlist` against the *re-filled* copy. This forces the gate to measure the same canonical fill a fresh re-check produces. Pros: no Python subprocess of our own code. Cons: still in-process for the compare itself, so a shapely-availability skew is NOT covered; more KiCad-version-coupled.

**Recommendation:** Option A. It directly satisfies the acceptance criterion ("in-process and fresh results agree on identical bytes") by *making the gated result a fresh result*. Keep the existing in-process `compare_copper_netlist` call as the value recorded in `lvs.json`'s `copper_mismatches` only if it agrees with the subprocess result; if they disagree, treat the board as DIRTY (fail closed) and log both so the divergence itself is visible.

Add a guard either way: if `run_copper and require_clean` and the subprocess/fresh result is dirty, raise `BoardNetlistMismatch` (reusing the existing path at recipe.py:142-147). Do **not** regress the advisory (`require_clean=False`) behavior for boards 04/05 in `ADVISORY_LVS_BOARDS`.

This is a single-PR change centered on `src/kicad_tools/lvs/recipe.py`. `compare_copper_netlist` already has the right (path-based) signature, so no API change is needed there. Scope guard: keep this to the gate authoritativeness; the *board actually being dirty on regen* is #3841 and the *CI gate validating a fresh regen* is #3840 — do not fold those in.

## Affected Files

- `src/kicad_tools/lvs/recipe.py` - Make `write_lvs_report`'s gated copper compare authoritative against the on-disk artifact: run the gated `compare_copper_netlist` in a fresh subprocess (Option A), fail closed on any in-process-vs-fresh disagreement, and keep `require_clean` raising `BoardNetlistMismatch` on a dirty fresh result. Preserve advisory behavior for `ADVISORY_LVS_BOARDS`.
- `src/kicad_tools/lvs/copper_lvs.py` - Reference for the comparator contract (already path-based; `compare_copper_netlist`, line 184). Optionally add a thin `__main__`/CLI-callable wrapper that emits the result as JSON for the subprocess in Option A.
- `src/kicad_tools/validate/connectivity.py` - Reference only: `extract_pad_partition` (line 340) and the shapely-gated pour-bonding (lines 493, 545-548) explain why an identical-bytes divergence is possible. No change expected unless the chosen fix touches the fallback.
- `tests/test_lvs_recipe.py` - Add the regression test (the existing tests monkeypatch the comparator at recipe.py:51-52; the new test must use a REAL persisted dirty board so the on-disk authoritativeness is exercised, not the monkeypatched path).
- `boards/03-usb-joystick/generate_design.py` - Consumer; verify the gate now raises on a regen that strands the J1 shield pads (no code change expected here for #3838 itself — the via-firing fix is #3841).

## Test Plan

- [ ] Manual verification: take a known-dirty persisted board-03 routed `.kicad_pcb` (2 GND opens — e.g. one produced by a regen where stitching vias did not fire, per #3841, or a fixture with J1.A12/B1 deliberately stranded). Call `write_lvs_report(sch, routed, out, require_clean=True, run_copper=True, run_label=False)` and confirm it now RAISES `BoardNetlistMismatch` and `lvs.json` records `clean=false` with the GND opens in `copper_mismatches`.
- [ ] Manual verification: on a genuinely clean persisted board (e.g. the current committed `boards/03-usb-joystick/output/usb_joystick_routed.kicad_pcb`, confirmed clean both ways during curation), the gate still passes and does not raise.
- [ ] Automated tests (`tests/test_lvs_recipe.py`): add `test_gate_matches_fresh_out_of_process_on_identical_bytes` — for a fixture PCB, assert the value used by the gate equals a freshly-spawned `compare_copper_netlist` on the same file (clean==clean, mismatch set equal). This is the core invariant.
- [ ] Automated tests: add `test_gate_catches_stranded_pad_board` — a persisted board with stranded GND pads makes `require_clean=True` raise `BoardNetlistMismatch` (NOT pass). Use a real on-disk fixture, not the monkeypatched comparator.
- [ ] Edge cases: (a) shapely present in recipe but a fresh check path disagrees -> gate must fail closed; (b) advisory boards (`require_clean=False`) still only log, never raise; (c) `run_copper=False` short-circuits as before (vacuously clean copper leg).
- [ ] Cross-check: after the fix, a fresh `uv run python boards/03-usb-joystick/generate_design.py` whose regen strands J1 pads should exit non-zero at the LVS step (this also surfaces #3841; if vias now fire and the board is clean, that is the #3841 fix landing, and #3838's gate should pass legitimately).

rjwalters:
## Summary

The recipe copper-LVS gate (`write_lvs_report(..., require_clean=True)`, `src/kicad_tools/lvs/recipe.py`) trusted a single **in-process** `compare_copper_netlist` whose result can diverge from a **fresh out-of-process** check on the *same persisted `.kicad_pcb` bytes*. Because the recipe persists a Python-post-processed fill (a fresh `kicad-cli` refill reverts the carve at `cli/runner.py:407-413`) and resolves shapely availability per-process, the gated comparison could see a cleaner board than a fresh process — so a physically broken board (e.g. board-03 with stranded J1 shield pads / GND opens) passed the hard gate while CI's fresh re-check on the identical file reports dirty.

This makes the gated copper compare authoritative against the on-disk artifact (curator Option A): re-run `compare_copper_netlist` in a **fresh subprocess** on the persisted file and **fail closed** if the in-process and fresh results disagree.

It does NOT mask the real board-03 defect — the gate now CATCHES the stranded-pad board (the via-firing root cause is sibling #3841). Genuinely-clean boards (01/02/03 committed) still pass.

## Changes

- `src/kicad_tools/lvs/copper_lvs.py`: add `result_to_json` / `result_from_json` and a `__main__` / `_main` subprocess entrypoint (`python -m kicad_tools.lvs.copper_lvs <sch> <pcb>`) that loads both files fresh and emits the copper-LVS result as JSON.
- `src/kicad_tools/lvs/recipe.py`: add `_fresh_copper_compare` (spawns the subprocess), `_authoritative_copper_result` (returns the fresh result, fails closed with a recorded divergence sentinel when in-process != fresh), `_copper_mismatch_key`, a `fresh_copper_check` flag (default `True`), and `FreshCopperCheckError`. The gated copper result and `lvs.json` now reflect the authoritative fresh state; advisory boards (`require_clean=False`) still only log, never raise.
- `src/kicad_tools/lvs/__init__.py`: export `FreshCopperCheckError`.
- `tests/test_lvs_recipe.py`: new real on-disk regression tests (no monkeypatched comparator for the authoritativeness path) plus `fresh_copper_check=False` on the existing fixture-free gating-logic tests.

| Gated result equals a fresh out-of-process check on identical bytes | ✅ | `test_gate_matches_fresh_out_of_process_on_identical_bytes` (real board-01 files) asserts `_copper_mismatch_key(in_process) == _copper_mismatch_key(fresh)` |
| A stranded-pad board is CAUGHT by the gate (raises, not silenced) | ✅ | `test_gate_catches_stranded_pad_board` builds a real persisted board with net-1 copper removed (J1.1/R1.1 stranded -> open) and asserts `require_clean=True` raises `BoardNetlistMismatch`, `lvs.json` clean=false |
| Clean board still passes | ✅ | `test_gate_matches_fresh_out_of_process_on_identical_bytes` runs the full gate on board-01 with no raise; `test_board_01_real_outputs_are_clean` still green |
| Fail closed on in-process vs fresh disagreement | ✅ | `test_gate_fails_closed_on_in_process_vs_fresh_divergence` (in-process clean, fresh dirty) raises and records a `<gate-divergence>` sentinel |
| Advisory (`require_clean=False`) never raises | ✅ | `test_stranded_pad_board_is_advisory_when_require_clean_false` |
| `run_copper=False` short-circuits (no subprocess) | ✅ | `test_run_copper_false_skips_fresh_check` asserts the fresh check is never invoked |

## Test Plan

- `uv run pytest tests/test_lvs_recipe.py` -> 14 passed.
- `uv run pytest tests/test_copper_lvs.py tests/test_board_03_copper_lvs.py tests/test_board_04_copper_lvs.py` -> all green (no regressions).
- `uv run ruff check` + `ruff format --check` on all touched files -> clean.
- Manual: `python -m kicad_tools.lvs.copper_lvs <board-01 sch/pcb>` emits `{"clean": true, "mismatches": []}`.

Scope guard: this PR only makes the GATE trustworthy. The board-03 stitching-via root cause is #3841 and the CI fresh-regen validation is #3840 — neither is folded in here.

Closes #3838

### DIFF
--- a/src/kicad_tools/lvs/__init__.py
+++ b/src/kicad_tools/lvs/__init__.py
@@ -51,13 +51,18 @@ from kicad_tools.lvs.copper_lvs import (
     compare_copper_netlist,
     compare_partitions,
 )
-from kicad_tools.lvs.recipe import ADVISORY_LVS_BOARDS, write_lvs_report
+from kicad_tools.lvs.recipe import (
+    ADVISORY_LVS_BOARDS,
+    FreshCopperCheckError,
+    write_lvs_report,
+)
 
 __all__ = [
     "ADVISORY_LVS_BOARDS",
     "BoardNetlistMismatch",
     "CopperLVSMismatch",
     "CopperLVSResult",
+    "FreshCopperCheckError",
     "LVSMismatch",
     "LVSResult",
     "_ref_of",
--- a/src/kicad_tools/lvs/copper_lvs.py
+++ b/src/kicad_tools/lvs/copper_lvs.py
@@ -208,3 +208,77 @@ def compare_copper_netlist(sch_path: str | Path, pcb_path: str | Path) -> Copper
     schematic_net_of_pad = _schematic_pin_to_net(sch_path)
     copper_partition = ConnectivityValidator(pcb_path).extract_pad_partition()
     return compare_partitions(schematic_net_of_pad, copper_partition)
+
+
+def result_to_json(result: CopperLVSResult) -> dict:
+    """Serialize a :class:`CopperLVSResult` to a JSON-safe dict.
+
+    Used by the ``python -m kicad_tools.lvs.copper_lvs`` subprocess
+    entrypoint (see :mod:`kicad_tools.lvs.recipe`'s fail-closed gate) so a
+    *fresh* out-of-process comparison can be marshalled back to the parent
+    recipe process for an authoritative, byte-for-byte agreement check.
+    """
+    return {
+        "clean": result.clean,
+        "mismatches": [
+            {
+                "kind": m.kind,
+                "net_a": m.net_a,
+                "net_b": m.net_b,
+                "pad_a": m.pad_a,
+                "pad_b": m.pad_b,
+            }
+            for m in result.mismatches
+        ],
+    }
+
+
+def result_from_json(payload: dict) -> CopperLVSResult:
+    """Reconstruct a :class:`CopperLVSResult` from :func:`result_to_json`."""
+    mismatches = tuple(
+        CopperLVSMismatch(
+            kind=m["kind"],
+            net_a=m["net_a"],
+            net_b=m["net_b"],
+            pad_a=m["pad_a"],
+            pad_b=m["pad_b"],
+        )
+        for m in payload.get("mismatches", ())
+    )
+    return CopperLVSResult(clean=bool(payload["clean"]), mismatches=mismatches)
+
+
+def _main(argv: list[str] | None = None) -> int:
+    """CLI/subprocess entrypoint: emit a fresh copper-LVS result as JSON.
+
+    Usage::
+
+        python -m kicad_tools.lvs.copper_lvs <schematic> <routed_pcb>
+
+    Loads both files fresh (a clean interpreter, no in-process recipe
+    state), runs :func:`compare_copper_netlist`, and prints the result via
+    :func:`result_to_json` to stdout as a single JSON object.  This is the
+    authoritative on-disk check the recipe gate compares its in-process
+    result against (issue #3838).
+    """
+    import argparse
+    import json
+    import sys
+
+    parser = argparse.ArgumentParser(
+        prog="python -m kicad_tools.lvs.copper_lvs",
+        description="Emit a fresh out-of-process copper-LVS result as JSON.",
+    )
+    parser.add_argument("schematic", help="Path to the .kicad_sch (root sheet).")
+    parser.add_argument("routed_pcb", help="Path to the routed .kicad_pcb.")
+    args = parser.parse_args(argv)
+
+    result = compare_copper_netlist(args.schematic, args.routed_pcb)
+    sys.stdout.write(json.dumps(result_to_json(result)))
+    return 0
+
+
+if __name__ == "__main__":  # pragma: no cover - exercised via subprocess
+    import sys
+
+    sys.exit(_main())
--- a/src/kicad_tools/lvs/recipe.py
+++ b/src/kicad_tools/lvs/recipe.py
@@ -37,6 +37,8 @@ Gate policy is per-board (see the curator matrix on #3762):
 from __future__ import annotations
 
 import json
+import subprocess
+import sys
 from pathlib import Path
 
 from kicad_tools.lvs.board_lvs import (
@@ -44,7 +46,12 @@ from kicad_tools.lvs.board_lvs import (
     LVSResult,
     compare_netlists,
 )
-from kicad_tools.lvs.copper_lvs import CopperLVSResult, compare_copper_netlist
+from kicad_tools.lvs.copper_lvs import (
+    CopperLVSMismatch,
+    CopperLVSResult,
+    compare_copper_netlist,
+    result_from_json,
+)
 
 # Boards that run LVS + emit ``lvs.json`` but are NOT yet copper/label clean.
 # Recipes for these boards must pass ``require_clean=False`` so a dirty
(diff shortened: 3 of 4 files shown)


rjwalters:
✅ **Approved!**

Verified all the authoritativeness concerns against the actual diff and on-disk behavior (reviewed in the builder worktree `.loom/worktrees/issue-3838`).

**1. Correctness — the gate is genuinely authoritative.**
`_fresh_copper_compare` (recipe.py:108) spawns `python -m kicad_tools.lvs.copper_lvs <sch> <pcb>`, whose `_main` (copper_lvs.py:251) re-runs the real `compare_copper_netlist` on the persisted files and emits JSON. `_authoritative_copper_result` (recipe.py:152) returns the **fresh** result, and `write_lvs_report` rebinds `copper_result` to it (recipe.py:259-264) before computing `gated_clean` — so the gate uses the fresh value, not the in-process one. Fail-closed is real: on divergence it returns `CopperLVSResult(clean=False, ...)` with a `<gate-divergence>` sentinel + the union of both legs' mismatches (recipe.py:183-195), which trips the `raise BoardNetlistMismatch` at line 283-288. Nothing is swallowed.

**2. Real defect is not masked.** `test_gate_catches_stranded_pad_board` builds a genuinely-dirty persisted board (net-1/VIN copper segments stripped → J1.1/R1.1 split into separate islands) and asserts `require_clean=True` now RAISES `BoardNetlistMismatch` with `clean=false` and an `open` in `lvs.json`. `test_gate_matches_fresh_out_of_process_on_identical_bytes` confirms the clean board-01 still passes (no raise, `_copper_mismatch_key(in_process)==_copper_mismatch_key(fresh)`). `test_gate_fails_closed_on_in_process_vs_fresh_divergence` simulates the exact defect (in-process clean, fresh dirty) and confirms the raise + sentinel.

**3. No regression on clean boards.** Ran locally:
- `tests/test_lvs_recipe.py` → 14 passed
- `tests/test_copper_lvs.py tests/test_board_03_copper_lvs.py tests/test_board_04_copper_lvs.py` → 40 passed
- `tests/test_board_03_copper_lvs.py tests/test_board_03_regression.py tests/test_board_01_ship_ready.py` → 23 passed, 1 skipped
- CI: Board 00/01/02/03 E2E and Board 06/07 LVS all green.

**4. Robustness.** Subprocess non-zero exit, `OSError` on spawn, and unparseable/invalid JSON all raise `FreshCopperCheckError` (fail closed) — a crashed subprocess cannot silently pass. No recursion risk: the `__main__` guard calls `_main`, which never re-enters the recipe. `run_copper=False` short-circuits with no subprocess (`test_run_copper_false_skips_fresh_check`), and advisory boards (`require_clean=False`) log without raising.

**5. Perf.** One subprocess per gated board at recipe time (not the routing hot path) — acceptable.

**6. CI.** Type Check ✅, Lint & Format ✅, full Test ✅ (14m30s), Board 00–03/06/07 ✅, DRC ✅, C++ Build ✅. I also ran `mypy` locally on the three touched source files (the builder couldn't): `Success: no issues found in 3 source files`. The only outstanding job is **Board 05 Routing Regression**, which is the pre-existing routing check explicitly out of scope here — this PR touches only LVS gate logic, no routing code, so `mergeStateStatus=UNSTABLE` is solely that unrelated in-progress job.

PR correctly uses `Closes #3838`. Clean, well-tested, and it makes the gate trustworthy without masking the board-03 root cause (#3841). Nice work.

Issue #3830, opened 2026-06-21
drc: kct DRC rule-coverage gaps vs native KiCad (no copper_sliver, hole_to_hole, silk_over_copper, silk_edge_clearance)


rjwalters:
## Original Issue

Discovered during the clean-rebuild sweep (softstart, board-05, board-01). kct check / kct audit report DRC PASS on boards where native kicad-cli pcb drc finds real violations in rule classes kct does not implement:

- softstart committed board: kct check --mfr jlcpcb-tier1 -> 0 errors, while kicad-cli pcb drc (same file, committed project rules) -> 27 violations: 16 copper_sliver, 2 hole_to_hole (e.g. NRST via 0.1516mm to SW1 PTH pad, below min), 4 silk_over_copper, plus a stranded +3.3V F.Cu zone island (6 unconnected_items).
- board-05 unrouted: kicad-cli reports 12 silk_over_copper + 2 silk_edge_clearance that kct reports none of.
- board-01: kicad-cli reports 2 silk_over_copper (J1/J2 ref designators clipping pad mask) that kct does not surface.

kct DRC (29 rules) has no copper_sliver, no hole_to_hole (drill-to-drill), no silk_over_copper, and no silk_edge_clearance checks. This is a DRC FIDELITY gap (related to the #3803 family) distinct from the board-05 routing divergence. The hole_to_hole one is a genuine manufacturability blocker kct misses.

Fix (curator please scope/split — likely one rule per follow-up): add the missing rule classes, prioritizing hole_to_hole (manufacturability) and copper_sliver. Acceptance: kct flags the hole_to_hole and copper_sliver violations that kicad-cli finds on the softstart board.

---

## Curator Enhancement (epic / tracking)

This is an **umbrella issue** for closing kct's DRC rule-coverage gaps versus native `kicad-cli pcb drc`. It should be split into one tractable child issue per rule class (silk rules grouped). Do **not** implement against this issue directly — it tracks the children.

Related family: #3803 (router/DRC fidelity, now CLOSED) and the route-vs-DRC fidelity work generally. This issue is the *DRC-engine coverage* half of that family, distinct from routing divergence.

### How the kct DRC engine is structured (scoping context)

- Rules live in `src/kicad_tools/validate/rules/`. Most are classes inheriting `DRCRule` (`rules/base.py`) with a single `check(pcb, design_rules) -> DRCResults`; a few (silkscreen) are plain functions aggregated by `check_all_silkscreen`.
[...]

PR #3849, opened 2026-06-21
feat(drc): add silk_over_copper and silk_edge_clearance rules

### LINKED ISSUE #3844: drc: add silk_over_copper and silk_edge_clearance rules
Child of epic #3830. PRIORITY P3, medium effort.

kct misses silkscreen DRC that native kicad-cli flags: silk_over_copper (silk text/graphics over exposed copper / pad mask apertures; e.g. J1/J2 ref designators clipping pad 1 mask on board-01; many on board-05) and silk_edge_clearance (silk too close to / over Edge.Cuts; e.g. board-05 MH1/MH2 ref clipped by board edge).

---

### Problem Statement

`kct check` runs three silkscreen rules today (`silkscreen_line_width`, `silkscreen_text_height`, and a crude `silkscreen_over_pad`), but native `kicad-cli pcb drc` reports two silkscreen defect classes that kct entirely misses or only approximates:

1. **`silk_over_copper`** — silk text/graphics printed over *exposed* copper (a pad's solder-mask aperture, or any soldermask-opening). Silk ink does not adhere to bare copper/HASL/ENIG and smears during reflow, so the manufacturer flags it. kct's existing `check_silkscreen_over_pads` (`validate/rules/silkscreen.py:217`) is a **crude centroid heuristic**: it only checks whether an fp_text's *position point* lands within half the pad size of an SMD pad. It (a) ignores the text's real bounding box, (b) ignores graphic strokes (fp_line/fp_rect outlines), (c) ignores board-level gr_text, (d) compares against the bare pad copper rather than the *solder-mask aperture* (pad size + mask expansion), and (e) misses through-hole pad apertures. Result: it does not reproduce the board-01 J1/J2 ref-designator-over-pad-1 findings kicad-cli reports.

2. **`silk_edge_clearance`** — silk too close to or crossing the `Edge.Cuts` board outline (e.g. board-05 MH1/MH2 ref designators clipped by the board edge). kct has **no rule at all** for silk-to-edge clearance. The machinery to do it already exists in `validate/rules/edge.py` (`EdgeClearanceRule._min_distance_to_outline` + `PCB.get_board_outline_segments()`), but it is only applied to copper/holes, never to silk.

Both defects are cosmetic/manufacturability **warnings** in kicad-cli, not hard errors — they must not spuriously block the manufacturing gate.

### Root cause

- `check_silkscreen_over_pads` was written as an explicit placeholder ("A full geometric overlap check would require more complex calculations" — see its docstring). shapely is now a core dependency (#3824; already imported in `validate/rules/clearance.py` and `validate/connectivity.py`), so real polygon-vs-polygon overlap is now cheap to implement.
- No silk-to-edge rule was ever written.

#### Geometry model facts the builder must respect

- **Footprint text/graphic coordinates are footprint-LOCAL.** `FootprintText.position` and `FootprintGraphic.start/end/center` (`validate/rules/... ` via `schema/pcb.py:234` / `:351`) are relative to the footprint origin. To get board coordinates you MUST apply the footprint transform exactly as `EdgeClearanceRule._check_pads` does (`validate/rules/edge.py:193-207`): rotate by `math.radians(-footprint.rotation)` (KiCad negates the orientation angle vs CCW math, verified in #3739) then translate by `footprint.position`. Reusing that transform verbatim avoids re-introducing the rotation-sign bug.
- **Board-level text/graphics** (`PCB.texts` → `GraphicText`, `PCB.graphics` → `BoardGraphic`) are already board-relative — no transform needed.
- **Edge outline** is available board-relative via `PCB.get_board_outline_segments()` (`schema/pcb.py:2753`); the per-segment distance helper is `EdgeClearanceRule._point_to_segment_distance` / `_min_distance_to_outline` (`validate/rules/edge.py:289-332`), wrapping `core.geometry.point_to_segment_distance`.
- **Solder-mask aperture** of a pad = pad copper expanded by the mask margin. Use `pad.solder_mask_margin` if set (`schema/pcb.py:165`), else fall back to `design_rules.min_solder_mask_clearance_mm` (`manufacturers/base.py:136`, default 0.05 mm). The aperture is the "exposed copper" region silk must avoid. SMD pads are always exposed; thru_hole pads expose copper on both sides.
- **Silk text bounding box:** approximate the text footprint as an axis-aligned (in text-local frame) box `width ≈ font_size[0] * len(text) * ~0.7 + font_thickness`, `height ≈ font_size[1] + font_thickness`, centered on the text position. KiCad's true glyph metrics are not modeled; a simple per-character-width box is sufficient to reproduce the kicad-cli findings and is the standard approximation already used elsewhere for text. Account for text rotation/justification only to first order (centered box is acceptable for AC). `font_height` accessor exists at `schema/pcb.py:345`/`:459`.
- **Silk graphic strokes:** model fp_line/fp_rect/gr_line/gr_rect as their stroke polygon (segment buffered by `stroke_width/2`) using shapely `LineString(...).buffer(w/2)`; rects expand to 4 segments (reuse `PCB._rect_to_segments`).

#### Approach (recommended)

Use shapely for both checks (consistent with `validate/rules/clearance.py`):

- **silk_over_copper:** Build a shapely geometry per silk element (text bbox → `box(...)`; graphic stroke → buffered `LineString`). Build the union (or an `STRtree`) of all pad mask-aperture polygons on the *matching* silk side (F.SilkS↔F.Cu pads, B.SilkS↔B.Cu pads; thru-hole pads expose on both). Emit a `silk_over_copper` **warning** for any silk geometry that `.intersects()` an aperture. Report `items=(footprint.reference, "pad N")` and the silk element identity. De-dup so each silk element fires at most once.
- **silk_edge_clearance:** For each silk element, compute the min distance from its geometry to the outline segments. If the geometry crosses the outline (distance 0 / intersects) OR the clearance is below a small silk-to-edge threshold, emit a `silk_edge_clearance` **warning**. Threshold: there is no dedicated `min_silk_to_edge_mm` field; gate against a sane default (recommend a module-level constant `SILK_EDGE_CLEARANCE_MM = 0.2 mm`, matching common fab silk-to-edge specs and kicad-cli's default `silk_edge_clearance`). Do NOT silently reuse `min_copper_to_edge_mm` (0.3) — silk-to-edge is typically looser; a named constant keeps intent clear and lets a future profile field override it. Reuse `get_board_outline_segments()` + `_point_to_segment_distance` against the bbox corners/edges (or shapely `.distance()` to a `MultiLineString` of the outline).

#### Severity must be WARNING (non-blocking)

Both rules emit `severity="warning"` (matching the existing silkscreen rules and kicad-cli). Because `ManufacturingAudit._check_drc` gates on **error**-severity violations and the advisory classifier (`DRCChecker.ADVISORY_RULE_IDS`, `validate/checker.py:156`), warning-severity silk rules will surface in JSON/table output and CI but will NOT block `kct export` / the manufacturing gate. Do NOT add these rule_ids to `ADVISORY_RULE_IDS` (that set is only needed for *error*-severity rules that must be exempted from gating; warnings are already non-blocking). Verify by confirming the existing `silkscreen_line_width` warning does not block the gate.

#### Wiring: which method, and the check_cmd.py coupling

**Recommended (lowest churn):** extend the existing `check_all_silkscreen` (`validate/rules/silkscreen.py:289`) to also run the two new checks, and have it call them from inside the already-registered `DRCChecker.check_silkscreen` method (`validate/checker.py:562`). This adds NO new entry to `CHECK_ALL_METHODS`, so **no change to `cli/check_cmd.py` is required** and the coverage test (`tests/test_check_cmd_coverage.py`) stays green. The new rule_ids flow through `--skip silkscreen` / `--only silkscreen` automatically.

**Alternative (only if you split into a separate method):** If you instead add a new `DRCChecker.check_silk_clearance` (or similar) to `CHECK_ALL_METHODS` (`validate/checker.py:113`), you MUST ALSO, in the SAME PR, update `cli/check_cmd.py` in TWO places: add the category to the `CHECK_CATEGORIES` list (`check_cmd.py:476`) AND add it to the `check_methods` dict inside `run_selected_checks` (`check_cmd.py:942`). Otherwise `tests/test_check_cmd_coverage.py::TestDispatcherIsSupersetOfCheckAll` and `::TestCategoryListMatchesDispatcher` and `::test_entry_points_see_same_rule_universe_modulo_severity` will fail. This coupling was flagged by the curator of parent epic #3830 and is **confirmed**: the test drives every `CHECK_ALL_METHODS` entry through the CLI dispatcher and asserts the two declarations stay in sync.

The recommended approach (extend `check_silkscreen`, no new method) sidesteps this coupling entirely and is preferred.

#### Violation type enum

`drc/violation.py` already has `SILK_OVER_COPPER = "silk_over_copper"` (`:213`). It does **NOT** have a `silk_edge_clearance` member — **add** `SILK_EDGE_CLEARANCE = "silk_edge_clearance"` to the `# Silkscreen` block (`drc/violation.py:212-217`). Use rule_id strings `"silk_over_copper"` and `"silk_edge_clearance"` in the emitted `DRCViolation`s so they map to these enum values. (Note: keep the legacy `silkscreen_over_pad` rule_id only if you retain the old heuristic; recommended to replace `check_silkscreen_over_pads` with the new geometric `silk_over_copper` check and drop the crude centroid version, since the new check supersedes it — but verify no test asserts on `silkscreen_over_pad` first via `grep -rn "silkscreen_over_pad" tests/`.)

### Affected Files

- `src/kicad_tools/validate/rules/silkscreen.py` - Replace/augment `check_silkscreen_over_pads` with a geometric `check_silk_over_copper` (silk bbox/stroke vs pad mask apertures, per side) and add a new `check_silk_edge_clearance` (silk geometry vs `Edge.Cuts` outline, gated by `SILK_EDGE_CLEARANCE_MM` constant); wire both into `check_all_silkscreen`. Use shapely (already a core dep). Apply the footprint rotation/translation transform (mirror `edge.py:193-207`) to footprint-local silk coords.
- `src/kicad_tools/validate/rules/edge.py` - REUSE `_min_distance_to_outline` / `_point_to_segment_distance` (no change required, or extract to a shared helper if cleaner). No new edge rule needed.
- `src/kicad_tools/drc/violation.py` - Add `SILK_EDGE_CLEARANCE = "silk_edge_clearance"` to the `# Silkscreen` enum block (`:212`). `SILK_OVER_COPPER` already exists.
- `src/kicad_tools/validate/checker.py` - Update `check_silkscreen` docstring (`:562`) to mention the two new sub-checks. NO `CHECK_ALL_METHODS` change if you use the recommended single-method approach.
- `src/kicad_tools/cli/check_cmd.py` - ONLY if you add a separate check method (not recommended): add the category to `CHECK_CATEGORIES` (`:476`) AND to the `check_methods` dict (`:942`). With the recommended approach, NO change here.
- `tests/test_validate_silkscreen.py` (new file) - Unit tests for both checks (see Test Plan). Note: there is no existing `test_validate_silkscreen.py` — `test_fix_silkscreen.py` and `test_silkscreen_generator.py` are unrelated.

### Acceptance Criteria

- [ ] `kct check boards/01-voltage-divider/output/voltage_divider_routed.kicad_pcb` (and/or the non-routed `voltage_divider.kicad_pcb`) emits `silk_over_copper` warnings for the J1/J2 reference designators that clip pad 1's mask aperture, matching what `kicad-cli pcb drc` reports.
- [ ] `kct check boards/05-bldc-motor-controller/output/bldc_controller_routed.kicad_pcb` emits `silk_edge_clearance` warnings for the MH1/MH2 reference designators clipped by the board edge, matching `kicad-cli`.
- [ ] Both rules emit `severity="warning"` — confirmed they do NOT block the manufacturing gate (`kct export` / `ManufacturingAudit` still passes a board whose only new violations are these warnings).
- [ ] No false positives on clean silk: a board (or fixture) whose silk sits clear of all apertures and the edge produces zero `silk_over_copper` / `silk_edge_clearance` violations.
- [ ] Both rule_ids reachable via `kct check --only silkscreen` and suppressible via `kct check --skip silkscreen`.
- [ ] `tests/test_check_cmd_coverage.py` stays green (recommended approach requires no change; alternative approach requires the check_cmd.py wiring above).
- [ ] `pnpm check:ci` (ruff + mypy + pytest) passes.

### Test Plan

- [ ] Manual verification: run `kct check` on board-01 and board-05 routed PCBs; confirm the J1/J2 `silk_over_copper` and MH1/MH2 `silk_edge_clearance` warnings appear and that the overall verdict is not flipped to a hard FAIL by them. Compare against `kicad-cli pcb drc --severity-all` output for the same boards.
- [ ] Automated tests (new `tests/test_validate_silkscreen.py`):
  - `silk_over_copper`: synthetic PCB with an fp_text bbox overlapping an SMD pad's mask aperture → 1 warning; same text moved clear → 0 warnings. Cover a rotated footprint (e.g. 90°/270°) to exercise the transform-sign path. Cover an fp_line silk stroke crossing a pad aperture. Cover a thru_hole pad (exposed both sides).
  - `silk_edge_clearance`: synthetic PCB with silk text crossing / within `SILK_EDGE_CLEARANCE_MM` of an `Edge.Cuts` segment → 1 warning; text pulled inboard past the threshold → 0 warnings. Cover a board with non-zero `board_origin` (coordinate-frame regression).
  - Severity: assert every emitted violation has `severity == "warning"`.
  - Real-board regression: assert board-01 yields ≥1 `silk_over_copper` and board-05 yields ≥1 `silk_edge_clearance` (mark with the real-board / KiCad-fixture marker if those boards are gated behind one).
- [ ] Edge cases: hidden text (`fp_text.hidden`) is skipped; library-suppression flag (`suppress_library`) behavior is consistent with the other silk rules (decide whether silk_over_copper/edge respect `suppress_library` — recommend they do NOT suppress, since over-copper/over-edge is a placement defect independent of library origin; document the choice); board with no `Edge.Cuts` outline → edge check is a no-op (mirror `EdgeClearanceRule.check`'s empty-outline early return); zero-length / empty text strings.

### Dependencies

- #3824 (shapely as core dep) — already merged; shapely is importable from `validate/rules/clearance.py`. No open blockers.

rjwalters:
Closes #3844 (child of epic #3830, P3).

## What

Adds two geometric silkscreen DRC rules that native `kicad-cli pcb drc` flags but `kct` previously missed or only approximated:

- **`silk_over_copper`** — silk text/graphics printed over a pad's exposed solder-mask aperture (e.g. board-01 J1/J2 reference designators clipping pad-1's mask).
- **`silk_edge_clearance`** — silk too close to or crossing the `Edge.Cuts` board outline (e.g. board-05 MH1/MH2 references clipped by the board edge). New `SILK_EDGE_CLEARANCE_MM = 0.2` constant.

Both emit **WARNING** severity (matching kicad-cli), so they surface in JSON/table output and CI but do **not** block the manufacturing gate (`ManufacturingAudit._check_drc` keys on error severity only).

## How

- Geometric implementation with shapely (core dep #3824), mirroring `EdgeClearanceRule._check_pads`'s footprint transform (`radians(-rotation)`, #3739) and reusing `PCB.get_board_outline_segments()`.
- Silk text → stroked bounding box; graphic strokes → buffered `LineString`; pad apertures → pad copper expanded by the mask margin (per-pad override or profile `min_solder_mask_clearance_mm`). SMD pads expose copper on their own side; thru-hole pads on both. An STRtree indexes apertures for fast lookup.
- A small overlap-area gate (`0.05 mm^2`) rejects hairline AABB corner touches so clean silk never false-positives.
- Wired into the already-registered `check_silkscreen` via `check_all_silkscreen` — **no** `cli/check_cmd.py` / `CHECK_ALL_METHODS` change, so `tests/test_check_cmd_coverage.py` stays green and `--only/--skip silkscreen` work automatically.
- Adds `SILK_EDGE_CLEARANCE` to the violation enum (`SILK_OVER_COPPER` already existed). The legacy `silkscreen_over_pad` heuristic is **retained** for backward compatibility with existing tests.

## Verification vs kicad-cli (boards 00-07)

Every silk element kct flags matches a real kicad-cli finding. The two silk-clean boards (charlieplex, stm32) yield **zero** violations — no false positives. kct under-reports slightly relative to kicad-cli (de-dups to one violation per silk element; conservative AABB), which is acceptable.

| Board | kct over / edge | kicad-cli over / edge |
|-------|-----------------|------------------------|
| 00-simple | 1 / 0 | 1 / 0 |
| 01-vdiv | 2 / 0 | 2 / 0 |
| 02-charlie | 0 / 0 | 0 / 0 |
| 03-usb | 2 / 0 | 4 / 0 |
| 04-stm32 | 0 / 0 | 0 / 0 |
| 05-bldc | 6 / 2 | 12 / 2 |
| 06-diff | 4 / 0 | 7 / 0 |
| 07-match | 3 / 0 | 5 / 0 |

## Tests

- New `tests/test_validate_silkscreen.py` (22 tests): positive + negative for both checks; rotated footprints (90/270), fp_line strokes, thru-hole apertures, silk-over-other-footprint pad, hidden/empty text skip, non-zero board origin, board-level gr_text, warning-severity, and real-board regression (board-01 over-copper, board-05 edge, clean boards).
- Updated two board-state baselines for the new real warning class: board-03 `EXPECTED_COMMITTED_DRC_RULES` and board-01 ship-ready (keeps strict 0-error gate, excludes the non-blocking silk-placement warnings from its 0-warning gate). board-01's silk defect is genuine and kicad-cli-confirmed; fixing the fixture silk is out of scope here.

ruff (check + format) and mypy clean on touched files (2 pre-existing mypy `assignment` warnings in unchanged loops left as-is). All board-state / fleet / audit / DRC test suites pass.

### DIFF
--- a/src/kicad_tools/drc/violation.py
+++ b/src/kicad_tools/drc/violation.py
@@ -218,6 +218,7 @@ class ViolationType(Enum):
 
     # Silkscreen
     SILK_OVER_COPPER = "silk_over_copper"
+    SILK_EDGE_CLEARANCE = "silk_edge_clearance"
     SILK_OVERLAP = "silk_overlap"
     SILKSCREEN_LINE_WIDTH = "silkscreen_line_width"
     SILKSCREEN_TEXT_HEIGHT = "silkscreen_text_height"
--- a/src/kicad_tools/validate/checker.py
+++ b/src/kicad_tools/validate/checker.py
@@ -560,12 +560,17 @@ class DRCChecker:
         return rule.check(self.pcb, self.design_rules)
 
     def check_silkscreen(self) -> DRCResults:
-        """Check silkscreen rules (line width, text height, over-pad).
+        """Check silkscreen rules.
 
         Validates:
-        - Minimum silkscreen line width
-        - Minimum silkscreen text height
-        - Silkscreen elements overlapping exposed pads
+        - Minimum silkscreen line width (``silkscreen_line_width``)
+        - Minimum silkscreen text height (``silkscreen_text_height``)
+        - Silkscreen overlapping exposed pads -- legacy centroid heuristic
+          (``silkscreen_over_pad``)
+        - Silkscreen text/graphics over pad mask apertures -- geometric shapely
+          check (``silk_over_copper``)
+        - Silkscreen text/graphics too close to the board edge -- geometric
+          shapely check (``silk_edge_clearance``)
 
         Returns:
             DRCResults containing silkscreen violations
--- a/src/kicad_tools/validate/rules/__init__.py
+++ b/src/kicad_tools/validate/rules/__init__.py
@@ -15,6 +15,8 @@ from .impedance import ImpedanceRule, NetImpedanceSpec
 from .match_group_length_skew import MatchGroupLengthSkewRule
 from .silkscreen import (
     check_all_silkscreen,
+    check_silk_edge_clearance,
+    check_silk_over_copper,
     check_silkscreen_line_width,
     check_silkscreen_over_pads,
     check_silkscreen_text_height,
@@ -42,6 +44,8 @@ __all__ = [
     "SolderMaskPadRules",
     "ViaInPadRule",
     "check_all_silkscreen",
+    "check_silk_edge_clearance",
+    "check_silk_over_copper",
     "check_silkscreen_line_width",
     "check_silkscreen_over_pads",
     "check_silkscreen_text_height",
--- a/src/kicad_tools/validate/rules/silkscreen.py
+++ b/src/kicad_tools/validate/rules/silkscreen.py
@@ -3,21 +3,72 @@
 This module implements DRC checks for silkscreen elements:
 - Minimum line width
 - Minimum text height
-- Silkscreen-over-pad detection
+- Silkscreen-over-pad detection (legacy centroid heuristic)
+- Silkscreen-over-copper detection (geometric, vs pad mask apertures)
+- Silkscreen-to-edge clearance (geometric, vs Edge.Cuts outline)
 """
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+import math
+from collections.abc import Callable, Iterator
+from typing import TYPE_CHECKING, Any
+
+from kicad_tools._shapely import require_shapely
 
 from ..violations import DRCResults, DRCViolation
 
 if TYPE_CHECKING:
-    from kicad_tools.manufacturers import DesignRules
-    from kicad_tools.schema.pcb import PCB, Footprint
+    from kicad_tools.schema.pcb import (
+        PCB,
+        BoardGraphic,
+        Footprint,
+        FootprintGraphic,
+        Pad,
+    )
+
+    from ...manufacturers import DesignRules
+
+# A shapely geometry.  shapely is dynamically imported (core dependency, see
+# ``_shapely``) and has no first-class stubs in this project, so geometries are
+# typed as ``Any`` -- matching the convention in ``clearance.py``.
+_Geometry = Any
+_Transform = Callable[[tuple[float, float]], tuple[float, float]]
 
 # Silkscreen layer names
 SILKSCREEN_LAYERS = ("F.SilkS", "B.SilkS", "F.Silkscreen", "B.Silkscreen")
+
+# Front/back silkscreen layer groupings, used to pair silk geometry with the
+# pad copper apertures on the matching side.
+FRONT_SILK_LAYERS = ("F.SilkS", "F.Silkscreen")
+BACK_SILK_LAYERS = ("B.SilkS", "B.Silkscreen")
+
+# Minimum silk-to-board-edge clearance (mm).  There is no dedicated profile
+# field for silk-to-edge; this named constant keeps the intent explicit and
+# distinct from the (stricter) copper-to-edge rule.  0.2 mm matches common
+# fab silk-to-edge specs and kicad-cli's default ``silk_edge_clearance``.
+SILK_EDGE_CLEARANCE_MM = 0.2
+
+# Per-character width factor used to approximate a stroked-font text bounding
+# box.  KiCad's true glyph metrics are not modeled; a simple per-character box
+# is sufficient to reproduce the kicad-cli silk findings.
+_TEXT_CHAR_WIDTH_FACTOR = 0.7
+
+# Floating-point tolerance for clearance comparisons (0.1 micron), matching
+# ``edge.py``'s ``_CLEARANCE_EPSILON_MM``.
+_CLEARANCE_EPSILON_MM = 1e-4
+
+# Minimum silk/aperture overlap area (mm^2) before ``silk_over_copper`` fires.
+# The text bounding box is an axis-aligned over-approximation of the true
+# stroked glyph outline, so a glyph whose real ink clears a pad can still
+# produce a hairline (corner-only) AABB intersection.  Requiring a small but
+# non-trivial overlap area suppresses those approximation artifacts while still
+# catching genuine silk-over-copper (which overlaps by far more than this).
+# Calibrated against boards 00-07 vs kicad-cli: genuine silk-over-copper
+# findings overlap by >= 0.089 mm^2, whereas the stm32 U2 corner artifact
+# overlaps by exactly 0.01 mm^2; a 0.05 mm^2 gate separates them with wide
+# margin and drops no real finding.
+_MIN_OVERLAP_AREA_MM2 = 0.05
 
 
 def is_silkscreen_layer(layer: str) -> bool:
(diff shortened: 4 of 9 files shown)


rjwalters:
🔧 **Evaluation in progress — fixed the failing Lint check, awaiting CI re-run**

I completed a full review of the geometry, acceptance, and baselines. The implementation is correct; the **only** problem was the failing `Lint & Format` CI check, caused by `tests/test_board_01_ship_ready.py` not being ruff-formatted (the multi-line `sum(...)` silk-placement comprehension). I fixed it directly during evaluation and pushed `74412f7e` (`style: ruff-format test_board_01_ship_ready.py`). `ruff check .` and `ruff format --check .` are now clean locally.

I am holding final approval until CI re-runs green on the new HEAD (the prior heavy run is still in progress, so the new run is queued).

**What I verified (all good):**
- **Geometry correctness** — footprint `radians(-rotation)` transform reused from `EdgeClearanceRule`; text→stroked AABB; graphic strokes→buffered LineStrings; pad aperture = copper + mask margin (per-pad override or profile `min_solder_mask_clearance_mm`); SMD one-side / thru-hole both-sides; edge distance via `PCB.get_board_outline_segments()`. All schema attrs (`font_size`, `font_thickness`, `solder_mask_margin`, `graphic_type`) exist.
- **Acceptance vs kicad-cli 10.0.4 (independent spot-check)** — board-05 `silk_edge_clearance`: kct=2 (MH1, MH2 reference) **exactly matches** kicad-cli=2 (MH1/MH2 reference fields). board-05 `silk_over_copper`: kct=6 vs cli=12 (under-reports, acceptable). board-01 `silk_over_copper` flags J1/J2 as claimed.
- **No false positives** — charlieplex and stm32 yield 0/0. The 0.05 mm² overlap gate is a calibrated artifact suppressor (genuine findings ≥0.089 mm² vs 0.01 mm² stm32 corner artifact), not a correctness fudge.
- **Severity** — both rules WARNING; board-01 ship-ready keeps a strict 0-error gate, so `ManufacturingAudit._check_drc` (error-keyed) is not spuriously blocked.
- **Legacy contract intact** — `silkscreen_over_pad` heuristic retained; `test_drc_violation_type_validate.py` / `test_violation_filters.py` (125 tests) pass.
- **Baseline updates legitimate** — board-03 `EXPECTED_COMMITTED_DRC_RULES`, board-01 ship-ready, and `test_validate.py` (rules_checked 3→5) all reflect the real new warning class, not silenced breakage. New `test_validate_silkscreen.py` (22 tests) passes.

Local: 22 + 125 + 96 tests pass; Type Check, C++ Build, board 00/01 e2e, DRC, kicad-cli round-trip all green on the prior run. Once the new run is green I'll flip to `loom:pr`.

rjwalters:
✅ **Approved!**

All CI checks are green on the current HEAD (`f3bdd03e`) — 15/15 jobs pass (Lint & Format, Type Check, Test, all board e2e + routing regressions) — and merge state is **CLEAN**. The implementation is correct, well-documented, and independently verified.

## Independent verification vs kicad-cli 10.0.4

- **board-05 `silk_edge_clearance`**: kct reports exactly **2** (MH1, MH2 reference fields) — an exact match to kicad-cli's 2 findings (`Reference field of MH1` / `MH2`, "Silkscreen clipped by board edge").
- **board-05 `silk_over_copper`**: kct=6 vs kicad-cli=12 — under-reports (one-per-element de-dup + conservative AABB), acceptable per the issue contract; kct never over-reports.
- **board-01 `silk_over_copper`**: flags J1/J2 over pad-1's mask aperture as claimed.
- **Clean boards (charlieplex, stm32)**: 0/0 — no false positives. The 0.05 mm² overlap gate is a calibrated artifact suppressor (genuine findings ≥0.089 mm² vs the 0.01 mm² stm32 corner artifact), not a correctness fudge.

## Review notes (all pass)

- **Geometry** — footprint `radians(-rotation)` transform reused from `EdgeClearanceRule`; text→stroked AABB; graphic strokes→buffered LineStrings; pad aperture = pad copper + mask margin (per-pad override or profile `min_solder_mask_clearance_mm`); SMD one-side / thru-hole both-sides; edge distance via `PCB.get_board_outline_segments()`. First-order AABB approximations are documented and empirically safe across boards 00–07.
- **Severity** — both rules WARNING; board-01 ship-ready keeps a strict 0-error gate, so `ManufacturingAudit._check_drc` (error-keyed) is not spuriously blocked. A previously-shippable board stays shippable.
- **Legacy contract intact** — `silkscreen_over_pad` heuristic retained; `test_drc_violation_type_validate.py` / `test_violation_filters.py` (125 tests) pass.
- **Baselines legitimate** — board-03 `EXPECTED_COMMITTED_DRC_RULES`, board-01 ship-ready, and `test_validate.py` (rules_checked 3→5) reflect the real new warning class, not silenced breakage. New `test_validate_silkscreen.py` (22 tests) passes.

## Fixes applied during evaluation

1. `style: ruff-format test_board_01_ship_ready.py` — collapsed the multi-line silk-placement comprehension (the original Lint failure).
2. `style: ruff-format test_audit.py for ruff 0.15.x` — CI's `setup-uv:latest` floated ruff 0.14→0.15.18, which reformats a redundant paren in an **unrelated, pre-existing** file (also dirty on `origin/main`). Formatted it so the Lint gate goes green; the result is stable under both ruff 0.14.x and 0.15.x. Both changes are pure formatting, zero behavior change.

Filed follow-up issue #3850 to pin the ruff version (the Lint gate is currently non-deterministic because `ruff>=0.1.0` + `setup-uv:latest` floats) and clean the residual `test_audit.py` drift on main.

Out of scope (as noted in the PR): the pre-existing board-05 routing informational check — confirmed green anyway.

PR #3857, opened 2026-06-22
fix(router): enforce min_hole_to_hole drill spacing at via-placement sites

rjwalters:
## Summary

The router placed vias (diff-pair fan-out, escape, stitching) without enforcing the manufacturer `min_hole_to_hole` (0.5mm) drill-to-drill floor, so it could emit sub-fab-minimum drill pairs that the `dimension_drill_clearance` DRC rule (#3842) correctly flags. This adds a **reject-and-fall-through** hole-to-hole guard at the via-insertion sites, reusing the canonical edge-to-edge formula from `validate/rules/dimensions.py::_check_drill_clearance` so the router pre-check and the DRC post-check agree. The A* cost model is unchanged.

## Changes

- **`router/via_clearance.py`**: new shared `drill_hole_to_hole_clear()` edge-to-edge predicate (`center_dist - d1/2 - d2/2 < min_hole_to_hole`, +1e-3 tolerance), mirroring the DRC rule.
- **`router/mfr_limits.py` / `router/rules.py`**: thread `min_hole_to_hole` (default 0.5) into `MfrLimits` and a **distinct** `DesignRules.min_hole_to_hole` field — the 0.102mm `min_drill_clearance` same-net-merge threshold (used by grid/pathfinder/drc_nudge) is left untouched. `DesignRules.__post_init__` pulls the manufacturer value when a manufacturer is set; unknown manufacturer falls back to 0.5.
- **`router/diffpair_routing.py`** (site a): replace the hardcoded center-to-center `< 0.6` crossover via-to-via check with an edge-to-edge `min_hole_to_hole` check against a board-wide drill registry (other crossovers' vias + all through-hole pad drills).
- **`router/escape.py`** (site b): `_can_place_via` gains an `existing_drills` arg + drill-edge check (`False` on conflict, callers already try the next site); the staggered fan forwards through-hole pad drills + its own placed vias. No-op when no registry is passed (back-compat).
- **`cli/stitch_cmd.py`** (site c): `calculate_via_position` gains an `other_net_drills` hole-to-hole check (the J1-S2 case) + a same-net drill floor; a post-placement **drop pass** guarantees no committed stitch via violates the floor through ANY placement strategy (straight/dog-leg/extended/micro), mirroring the #3271 `--avoid-pad-overlap` drop. The via's own target through-hole pad is excluded.
[...]

PR #3859, opened 2026-06-22
test: replace flaky absolute timing bound with relative spatial-index guard (#3858)

### LINKED ISSUE #3858: test: flaky timing-based perf test test_100_components_with_spatial_index_under_500ms (0.711s vs 0.5s on CI)
The performance test `tests/test_placement_slide_off.py::TestPerformance::test_100_components_with_spatial_index_under_500ms` asserts a wall-clock bound (`< 0.5s`) that flakes on shared CI runners — observed failure "Took 0.711s, expected < 0.5s" on an unrelated PR (#3857, which only touches router via-spacing + board-06 quantization). Wall-clock thresholds are inherently flaky on variable CI hardware and intermittently red-light unrelated PRs (same class of CI-noise as the ruff-float #3850 and the board-05 gate flakiness #3836).

---

### Problem Statement

The test uses an **absolute wall-clock threshold** (`elapsed < 0.5`), which is hardware-dependent. On a loaded shared CI runner the same code measured 0.711s and failed, red-lighting an unrelated PR. The threshold has no relationship to a code regression — it only measures how busy the runner is. We want to keep a regression guard (the spatial-index broad-phase should not silently regress to brute-force O(n^2) scaling) while removing the hardware-dependent flake.

### Root-cause finding (measured during curation)

A naive baseline path **does exist**: `slide_off_overlaps(..., use_spatial_index=False)` runs the brute-force O(n^2) pairing, and `use_spatial_index=True` runs the grid-based spatial index. So a relative comparison is feasible.

However, the obvious relative assertion ("grid is >5x faster than brute") would itself be **false and flaky** under the *current* test configuration. Measurements (Opus, M-series, `uv run`):

- **Current config (all 100 components coincident at origin, `max_iterations=5`):** grid ~0.040s vs brute ~0.050s — only ~1.2x. The spatial index's broad-phase pruning gives almost no advantage when every component lands in the same/overlapping grid cells, so the candidate set is still ~all pairs. Wall time here is dominated by per-pair resolution work, not broad-phase enumeration.
- **Spread config (components on a grid, well-separated):** grid is consistently faster across sizes —
  - n=100: brute 0.0145s / grid 0.0046s → **3.15x**
  - n=200: brute 0.0492s / grid 0.0184s → **2.67x**
  - n=400: brute 0.2116s / grid 0.0708s → **2.99x**

So the spatial index's *real* value (and the thing worth guarding against regression) is **broad-phase pruning on a spread-out placement**, where it holds a stable ~2.5-3x advantage that grows with n. The coincident configuration is a worst case that masks this.

### Recommended fix (Option C — relative, on a spread configuration)

Replace the absolute-ms assertion with a **relative** guard that compares the spatial-index variant against the brute-force baseline on a **spread** placement, where the broad-phase advantage is real and hardware-independent. Concretely, rewrite `test_100_components_with_spatial_index_under_500ms` (rename to reflect intent, e.g. `test_spatial_index_faster_than_brute_force_when_spread`) to:

1. Build a well-separated grid placement of ~200 components (use a new spread helper, or generalize `_make_vector_spread` to a 2D grid layout) on a large board so the spatial index actually prunes candidates.
2. Time both `use_spatial_index=False` and `use_spatial_index=True` with identical inputs and `max_iterations=5`.
3. Assert the grid variant is meaningfully faster than brute force with a **conservative ratio that tolerates CI noise**: `assert grid_elapsed * 2.0 < brute_elapsed` (i.e. grid is at least ~2x faster). Measured ratio is ~2.5-3x, so a 2x gate has comfortable headroom yet still fails loudly if the broad-phase regresses to O(n^2) (ratio → ~1x).
4. To further damp single-sample timing jitter on tiny absolute durations, take the **best of 3 timed runs** for each variant (min, not mean) before comparing — `min` is the standard way to remove scheduler/contention noise from microbenchmarks.

This keeps a true regression guard (a broad-phase regression collapses the ratio and fails the test) while being independent of absolute runner speed.

#### Fallback (Option A — generously-loosened absolute) if relative proves noisy

If the relative ratio is still judged too tight for the slowest CI runners, keep an absolute bound but loosen it to a CI-safe value and keep the coincident config: `assert elapsed < 2.0, f"Took {elapsed:.3f}s, expected < 2.0s"`. At n=100 the code runs in ~0.04-0.05s locally; a 2.0s ceiling is ~40x headroom over local and ~3x over the observed 0.711s CI sample, while a genuine O(n^2) blowup at this size would still blow past 2s. Prefer Option C; use Option A only if Option C is rejected.

**Do NOT** use Option B (warn-only benchmark) or Option D (perf marker excluded from default CI) — both silently drop the regression guard from the default test run, which is the opposite of what we want.

### Acceptance Criteria

- [ ] `test_100_components_with_spatial_index_under_500ms` no longer asserts an absolute wall-clock bound on the coincident configuration (the flaky `elapsed < 0.5` is removed or replaced).
- [ ] The replacement test compares the spatial-index path against the brute-force baseline (`use_spatial_index=False`) on a **spread** placement and asserts the grid variant is at least ~2x faster (`grid_elapsed * 2.0 < brute_elapsed`), using best-of-3 (`min`) timing per variant.
- [ ] The test still fails if the spatial-index broad-phase regresses to O(n^2) scaling (verified by temporarily forcing `use_spatial_index=False` for both timed calls and confirming the assertion fails).
- [ ] No absolute, hardware-dependent millisecond threshold remains in the rewritten test (the unchanged `test_20_components_under_100ms` may stay as-is, or be loosened in the same PR if desired — out of scope unless trivially adjacent).
- [ ] `uv run pytest tests/test_placement_slide_off.py -v` passes locally and the targeted test passes on 10 consecutive local runs without flaking.

### Implementation Guidance

- File: `tests/test_placement_slide_off.py`, class `TestPerformance` (lines ~702-740).
- Add a spread-grid helper (or generalize `_make_vector_spread` at lines 63-76, which currently lays components out in a 1D line) so ~200 components are distributed across a large board in 2D. Use a board large enough that the grid cells meaningfully partition components (e.g. `_make_board(500.0, 500.0)`).
- Use `time.perf_counter()` (already imported) and take `min()` over 3 timed calls per variant.
- Keep `max_iterations=5` so total runtime stays small and the comparison stays apples-to-apples with the original.
- Confirm `slide_off_overlaps` signature accepts `use_spatial_index: bool | None` (it does — `src/kicad_tools/placement/slide_off.py:235-243`).

### Affected Files

- `tests/test_placement_slide_off.py` - Rewrite `TestPerformance.test_100_components_with_spatial_index_under_500ms` to a relative grid-vs-brute comparison on a spread placement; optionally add a 2D-spread helper. No production code changes required.

### Test Plan

- [ ] Manual verification: `uv run pytest tests/test_placement_slide_off.py::TestPerformance -v` passes.
- [ ] Flake check: run the targeted test 10x in a loop locally (`for i in $(seq 10); do uv run pytest tests/test_placement_slide_off.py::TestPerformance::<new_name> -q || break; done`) — all green.
- [ ] Regression-guard check: temporarily set both timed calls to `use_spatial_index=False`, confirm the relative assertion FAILS (proves the guard catches a broad-phase regression), then revert.
- [ ] Automated tests: change is entirely within `tests/test_placement_slide_off.py`; no new production tests needed.
- [ ] Edge cases: confirm timing comparison is robust on a busy machine by running under load (e.g. `stress` or a parallel build) — relative ratio should still hold where the absolute bound would have flaked.

### Related

- #3857 - the unrelated PR red-lighted by this flake.
- #3850, #3836 - same class of CI-noise (hardware-dependent thresholds / flaky gates).

rjwalters:
Closes #3858

## Problem

`tests/test_placement_slide_off.py::TestPerformance::test_100_components_with_spatial_index_under_500ms` asserted an absolute wall-clock bound (`elapsed < 0.5s`) that is hardware-dependent. On a loaded shared CI runner it measured 0.711s and red-lighted an unrelated PR (#3857). The threshold measured runner load, not a code regression.

## Fix (curator Option C — relative, hardware-independent)

Rewrote the test as `test_spatial_index_outperforms_bruteforce`:

- Times both the brute-force O(n^2) baseline (`use_spatial_index=False`) and the grid spatial index (`use_spatial_index=True`) with identical inputs and `max_iterations=5`.
- Uses a **spread-out** 2D grid placement (new `_make_vector_spread_grid` helper): 200 well-separated components on a 500x500 board, so the broad-phase actually prunes candidate pairs. (With the old all-coincident config the index gives only ~1.2x and a relative gate would itself be flaky — see #3858 root cause.)
- Asserts `grid_elapsed * 2.0 < brute_elapsed`. The 2x gate is independent of runner speed and fails loudly if the broad-phase regresses to O(n^2) (ratio collapses toward 1x).
- Best-of-3 (`min`) timing per variant to damp scheduler/contention jitter.

No production code changes; the change is entirely within the test file.

## Verification

- Measured locally: brute=0.0498s, grid=0.0168s -> **~2.96x** (comfortably above the 2x gate).
- **Regression guard confirmed**: forcing both timed calls to `use_spatial_index=False` makes the assertion fail (ratio ~1x).
- **Flake check**: targeted test passed **10/10** consecutive local runs.
- Full `tests/test_placement_slide_off.py` suite: 27 passed.
- `ruff check` / `ruff format --check`: clean. `mypy` on touched file: clean.

### DIFF
--- a/tests/test_placement_slide_off.py
+++ b/tests/test_placement_slide_off.py
@@ -76,6 +76,34 @@ def _make_vector_spread(
     return PlacementVector(data=data)
 
 
+def _make_vector_spread_grid(
+    n: int,
+    spacing: float = 20.0,
+    side: int = 0,
+) -> PlacementVector:
+    """Create a vector with components spread across a 2D grid (no overlaps).
+
+    Components are laid out on an approximately square grid centred at the
+    origin so the spatial-index broad-phase can meaningfully partition them
+    into distinct cells. Contrast with :func:`_make_vector_spread`, which
+    only spreads along a single axis.
+    """
+    cols = max(1, int(math.ceil(math.sqrt(n))))
+    rows = max(1, int(math.ceil(n / cols)))
+    data = np.zeros(n * FIELDS_PER_COMPONENT, dtype=np.float64)
+    x0 = -((cols - 1) * spacing / 2)
+    y0 = -((rows - 1) * spacing / 2)
+    for i in range(n):
+        row = i // cols
+        col = i % cols
+        base = i * FIELDS_PER_COMPONENT
+        data[base] = x0 + col * spacing
+        data[base + 1] = y0 + row * spacing
+        data[base + 2] = 0.0
+        data[base + 3] = float(side)
+    return PlacementVector(data=data)
+
+
 def _count_overlaps_manual(
     vector: PlacementVector,
     components: list[ComponentDef],
@@ -720,24 +748,59 @@ class TestPerformance:
 
         assert elapsed < 0.1, f"Took {elapsed:.3f}s, expected < 0.1s"
 
-    def test_100_components_with_spatial_index_under_500ms(self):
-        """100 components with spatial index complete in < 500ms."""
-        board = _make_board(200.0, 200.0)
-        components = _make_components(100, size=2.0)
-        vector = _make_vector_at_same_position(100)
+    def test_spatial_index_outperforms_bruteforce(self):
+        """Spatial index beats brute-force on a spread-out placement.
 
-        start = time.perf_counter()
-        slide_off_overlaps(
-            vector,
-            components,
-            board,
-            margin_mm=0.5,
-            max_iterations=5,
-            use_spatial_index=True,
+        This replaces a former absolute wall-clock bound (``elapsed < 0.5s``)
+        that flaked on loaded CI runners (issue #3858). Instead of timing
+        against a hardware-dependent threshold, we time both the brute-force
+        O(n^2) baseline (``use_spatial_index=False``) and the grid-based
+        spatial index (``use_spatial_index=True``) with identical inputs and
+        assert the grid variant is meaningfully faster.
+
+        The comparison uses a *spread-out* 2D grid placement, because the
+        spatial-index broad-phase only prunes candidate pairs when components
+        fall into distinct grid cells. (With all components coincident the
+        candidate set is still ~all pairs and the index gives almost no
+        advantage — see #3858 root-cause analysis.) On a spread placement the
+        grid holds a stable ~2.5-3x advantage that grows with n, so a 2x gate
+        has comfortable headroom yet still fails loudly if the broad-phase
+        regresses to O(n^2) (where the ratio collapses toward 1x).
+
+        Timing uses best-of-3 (``min``) per variant to damp scheduler/
+        contention jitter on tiny absolute durations.
+        """
+        n = 200
+        board = _make_board(500.0, 500.0)
+        components = _make_components(n, size=2.0)
+        vector = _make_vector_spread_grid(n, spacing=15.0)
+
+        def _time(use_spatial_index: bool) -> float:
+            best = math.inf
+            for _ in range(3):
+                start = time.perf_counter()
+                slide_off_overlaps(
+                    vector,
+                    components,
+                    board,
+                    margin_mm=0.5,
+                    max_iterations=5,
+                    use_spatial_index=use_spatial_index,
+                )
+                best = min(best, time.perf_counter() - start)
+            return best
+
+        brute_elapsed = _time(use_spatial_index=False)
+        grid_elapsed = _time(use_spatial_index=True)
+
+        # Conservative relative gate: grid at least ~2x faster than brute.
+        # Measured ratio is ~2.5-3x, so this tolerates CI noise while still
+        # catching a broad-phase regression to O(n^2).
+        assert grid_elapsed * 2.0 < brute_elapsed, (
+            f"Spatial index not meaningfully faster than brute force: "
+            f"grid={grid_elapsed:.4f}s brute={brute_elapsed:.4f}s "
+            f"(ratio {brute_elapsed / grid_elapsed:.2f}x, expected >= 2x)"
         )
-        elapsed = time.perf_counter() - start
-
-        assert elapsed < 0.5, f"Took {elapsed:.3f}s, expected < 0.5s"
 
 
 # ---------------------------------------------------------------------------


rjwalters:
✅ **Approved!**

Test-only change that replaces the flaky absolute wall-clock bound (`test_100_components_with_spatial_index_under_500ms`, asserted `<0.5s`, flaked at 0.711s on a loaded CI runner) with a hardware-independent relative comparison `test_spatial_index_outperforms_bruteforce`. No production code touched (diff is confined to `tests/test_placement_slide_off.py`, +80/-17).

## Verification (all PR claims independently confirmed locally)

1. **Hardware-independent** — The assertion is purely relative (`grid_elapsed * 2.0 < brute_elapsed`); the old absolute `elapsed < 0.5` bound is fully removed. No residual wall-clock threshold remains. Best-of-3 `min()` per variant damps scheduler jitter on the tiny absolute durations — a reasonable approach.

2. **Catches a real regression** — Simulated an O(n^2) broad-phase regression by forcing both timed calls to `use_spatial_index=False`: the ratio collapses from ~2.96x to **1.01x** and the 2x gate **fails as designed**. Normal run measured **2.96x** (brute=0.0499s, grid=0.0169s), matching the PR's reported ~2.96x exactly.

3. **Layout is genuinely spread** — `_make_vector_spread_grid(200, spacing=15.0)` produces **200 unique positions** across ~210x195mm on a 500x500 board, so the grid broad-phase meaningfully prunes candidate pairs (the prerequisite for the index to win). The new helper is consistent with the existing single-axis `_make_vector_spread` and uses the already-imported `math`/`numpy`/`FIELDS_PER_COMPONENT`.

4. **2x gate headroom** — Comfortably below the measured ~2.96x (so it tolerates CI noise) yet well above 1x (so a broad-phase regression trips it). Reasonable.

5. **Quality** — Targeted test passed **6/6** consecutive local runs (stable). `ruff check` + `ruff format --check` clean on ruff 0.14.10. The test still exercises the real `slide_off_overlaps` API.

## CI
All 16 checks pass: Test (15m2s), Type Check, Lint & Format, C++ Build, Boards 00-07 e2e, Diff-Pair & Match-Group & Board-05 routing regressions, kicad-cli smoke. Merge state CLEAN. PR description correctly uses `Closes #3858`.

Nicely scoped fix that removes a real source of CI flake without weakening regression coverage.

Issue #3397, opened 2026-06-09
boards/05: schematic-emit defects — 36 residual sync mismatches across LED polarity, bootstrap caps, SWD, LDO, buck, fuse


rjwalters:
Spun out from PR #3396 (closes #3393 — rail-name normalization).

## Background

After PR #3396 lands, `kct pcb sync-netlist` for board 05 reports **0 rail-rename mismatches** and **0 U3 mismatches** (full closure of #3384 chain). But 36 total residual mismatches remain — all pre-existing schematic-emit bugs uncovered by sync. None block the U3 / rail-rename work; each is a discrete defect.

## Defect inventory (from PR #3396 builder report)

Each bullet may warrant a separate issue once triaged:

1. **LED polarity** — D1, D3, D4 wired with reversed anode/cathode
2. **Bootstrap cap chain** — U3 pins 29, 38, 43, 48, 52-54 (BST_x) connecting wrong
3. **U10 unused-pin no-connects** — missing or wrong NCs on unused U10 pins
4. **J4 SWD pinout swap** — SWDIO/SWCLK or VCC/GND swapped on J4
5. **U2 LDO pin swap** — pin assignment wrong on U2 (XC6206 LDO)
6. **U1 buck pin swap** — pin assignment wrong on U1 (likely STM32G031 buck output pin)
7. **Fuse-net naming** — J1.1, F1.1, L1.1 share a fuse net but with inconsistent labels

## Acceptance criteria

- [ ] Triage each defect: separate issue OR group bundled if related
- [ ] Each defect tracked individually with reproduction + fix scope
- [ ] After all defects ship, `kct pcb sync-netlist` reports 0 board-05 mismatches
- [ ] Fleet status reports 0 drift on board 05
- [ ] PCB can be committed-routed fresh from current schematic

## Investigation hints

- The builder of PR #3396 has full context on each defect — they may have file:line citations in the worktree they used
- Some defects (LED polarity, NCs) are schematic-emit code in `boards/05-bldc-motor-controller/design.py`
- Some (J4, U1, U2) may be footprint-pad-mapping bugs vs symbol-pin-name bugs

## Why grouped umbrella

These 7 defects share the locus (board 05 design.py + symbol/footprint pairings) and would otherwise spam the issue queue. Triage first, then split as needed.

## Related

- Final piece of: #3384 (will close on PR #3396 merge via the schematic side)
- Sync drift umbrella: #3370 (closed)
- Discovered during: PR #3396
[...]

PR #3868, opened 2026-06-23
feat(footprints): fail-loud missing-footprint preflight + heuristic auto-assign

### LINKED ISSUE #3866: footprints: auto-assign/generate missing footprints + fail-loud in pre-flight (never silently block on a missing footprint)
General unblock capability (raised during the chorus discussion; recurring EE-feedback theme "some footprints are missing"). NOTE: the current chorus revA design has NO missing footprints (verified 90 sch <-> 90 PCB) — this does not move chorus, but it removes a whole CLASS of silent blocker for future boards.

Today missing footprints can silently drop a component from the PCB (its nets then can never route). kicad-tools has the building blocks — `kct footprint` (generation), `validate-footprints`/`fix-footprints`, `suggest`, `parts` (LCSC matching) — but no AUTO-ASSIGNMENT step.

Build:
1. Detection: a check that flags any schematic symbol with a blank/missing Footprint field, and any netlist component absent from the PCB. Wire it into pre-flight (#1450) to FAIL LOUD (not a silent skip) before routing/export.
2. Auto-assign: for a symbol missing a footprint, assign one from value + package heuristics (e.g. 0402/0805 cap/resistor -> standard SMD footprint) or an LCSC part match via `kct parts`; report each assignment for human review. For truly unknown parts, fail loud with an actionable message rather than proceeding.

Acceptance: a design with a footprint-less symbol is caught in pre-flight with a clear message; common passives/ICs get a correct auto-assigned footprint; nothing silently routes/exports a board missing a component. Tests with a synthetic footprint-less symbol. Bounded.

rjwalters:
## Summary

Missing footprints could silently drop a component from the PCB (its nets then can never route) and were only a *warning* in pre-flight. This PR makes a missing footprint **fail loud** before routing/export, and adds a deterministic, library-independent **auto-assignment** path for standard SMD passives so the common case self-heals while genuinely-unknown parts fail loud with an actionable message.

## Changes

- **Detection (fail-loud pre-flight):**
  - `check_missing_footprints` (sch validate / `sch preflight`) now emits **error** severity (was warning), so the schematic pre-flight gate trips on any footprint-less symbol and names the fix.
  - New `PreflightChecker._check_schematic_footprints` gate in the export pre-flight (`kct export`): FAILs when any schematic symbol is missing a footprint (walks the hierarchy via the shared `iter_missing_footprint_symbols` predicate; honors strict-preflight blocking like the existing `bom_fields`/`bom_pcb_match` gates). The existing PCB-side checks (`bom_pcb_match` = in-BOM-not-on-PCB FAIL, `footprints`, `bom_fields`) already cover "netlist component absent from the PCB".
- **Auto-assign (`kct sch assign-footprints --assign-missing`):**
  - New `kicad_tools/footprints/heuristics.py`: `guess_chip_footprint` maps component class (R/C/L/ferrite-bead/LED/diode) + standard chip size (0201/0402/0603/0805/1206/1210/2010/2512) to the canonical KiCad SMD footprint (e.g. `Resistor_SMD:R_0402_1005Metric`) with **no disk access** — works on a fresh checkout / CI runner with no KiCad install.
  - The assign path falls back to the heuristic when the library scan yields no/ambiguous candidate. Each assignment is reported with its source (`[heuristic]` / `[library]`) for human review.
  - **Fail-loud:** with `--assign-missing`, any symbol that cannot be resolved (unknown IC, unknown package) causes a non-zero exit and a "must be set by hand" message — while still writing the parts it *could* resolve.

| Footprint-less symbol caught in pre-flight with a clear message, NOT silently routed/exported | ✅ | `test_preflight_gate_trips_on_missing_footprint`; export gate `test_missing_footprint_fails` (FAIL + actionable details) |
| Common passives/ICs get a correct auto-assigned footprint | ✅ | `test_footprint_heuristics.py` (R/C/L/FB/LED/D × all standard sizes); `test_passive_auto_assigned_with_no_library`, `test_passive_written_to_disk_with_no_library` |
| A truly unknown part fails loud | ✅ | `test_unknown_ic_fails_loud`, `test_unknown_ic_still_writes_resolved_passive` (rc=1, stderr names U1 + "by hand") |
| Pre-flight gate trips on missing footprints | ✅ | `check_missing_footprints` now error-severity; `run_preflight().passed is False` |
| Do NOT break existing footprint tests | ✅ | `test_sch_assign_footprints.py`, `test_sch_set_footprint.py`, `test_sch_suggest_footprint.py`, `test_sch_preflight.py`, `test_preflight.py`, `test_manufacturing_package.py`, `test_build_sync_step.py`, `test_pipeline_cmd.py` all pass |

## Test Plan

- New: `tests/test_footprint_heuristics.py` (15), `tests/test_sch_assign_missing.py` (8), `tests/test_preflight.py::TestPreflightSchematicFootprints` (2).
- Ran the full footprint/preflight/export/build suites (638+ tests) — all pass.
- `ruff 0.14.10` check + format clean on touched files; `mypy` on new code clean (pre-existing `_pcb`/`_bom` None-init errors in `export/preflight.py` are untouched).
- CLI smoke: `kct sch assign-footprints <sch> --assign-missing --dry-run` auto-assigns an 0603 resistor with no library installed.

Closes #3866

### DIFF
--- a/src/kicad_tools/cli/commands/schematic.py
+++ b/src/kicad_tools/cli/commands/schematic.py
@@ -209,6 +209,7 @@ def run_sch_command(args) -> int:
             include_dnp=getattr(args, "include_dnp", False),
             force=getattr(args, "force", False),
             no_project_lib=getattr(args, "no_project_lib", False),
+            assign_missing=getattr(args, "assign_missing", False),
         )
 
     elif args.sch_command == "set-value":
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -754,6 +754,15 @@ def _add_sch_parser(subparsers) -> None:
         help="Also consider DNP symbols (default: skip)",
     )
     sch_assign_fp.add_argument(
+        "--assign-missing",
+        action="store_true",
+        help=(
+            "Resolve missing footprints with a deterministic value+package "
+            "heuristic for standard SMD passives (no installed library "
+            "required); unknown parts fail loud with a non-zero exit"
+        ),
+    )
+    sch_assign_fp.add_argument(
         "--force",
         action="store_true",
         help="Reconsider symbols that already have a non-empty footprint",
--- a/src/kicad_tools/cli/sch_assign_footprints.py
+++ b/src/kicad_tools/cli/sch_assign_footprints.py
@@ -54,6 +54,7 @@ from kicad_tools.cli.sch_suggest_footprint import (
     find_footprint_candidates,
 )
 from kicad_tools.footprints.fp_lib_table import find_project_fp_lib_table
+from kicad_tools.footprints.heuristics import guess_chip_footprint
 from kicad_tools.footprints.library_path import detect_kicad_library_path
 
 
@@ -91,14 +92,24 @@ def _classify_symbol(
     paths: Any,
     limit: int,
     use_project_table: bool,
+    use_heuristic: bool = False,
 ) -> dict[str, Any]:
     """Compute the candidate list + auto-assign decision for one symbol.
 
     Returns a dict with the keys ``reference``, ``value``, ``lib_id``,
     ``current_footprint``, ``pin_count``, ``package_keyword``,
     ``fp_filters``, ``candidates`` (top-``limit``), ``status``
-    (``"assigned" | "ambiguous" | "no_candidates"``) and ``assigned``
-    (the chosen footprint string when ``status == "assigned"``).
+    (``"assigned" | "ambiguous" | "no_candidates"``), ``assigned``
+    (the chosen footprint string when ``status == "assigned"``) and
+    ``assigned_by`` (``"library"`` or ``"heuristic"`` when assigned).
+
+    When *use_heuristic* is true and the library scan does not yield an
+    unambiguous candidate, a deterministic value+package heuristic
+    (:func:`guess_chip_footprint`) is consulted for standard two-pin SMD
+    passives. This lets the command auto-assign on environments with no
+    installed KiCad library, and resolve parts whose footprint is fully
+    determined by their class + chip size. Genuinely unknown parts remain
+    ``ambiguous`` / ``no_candidates`` so the caller can fail loud.
     """
     target_pins = _resolve_target_pin_count(sch, sym)
     keyword = _derive_keyword(sym)
@@ -125,17 +136,37 @@ def _classify_symbol(
         "candidates": candidates,
         "status": "no_candidates",
         "assigned": None,
+        "assigned_by": None,
     }
 
-    if not candidates:
-        return record
-
-    if _is_unambiguous(candidates, fp_filters, keyword):
+    if candidates and _is_unambiguous(candidates, fp_filters, keyword):
         top = candidates[0]
         record["status"] = "assigned"
         record["assigned"] = f"{top['library']}:{top['footprint']}"
-    else:
+        record["assigned_by"] = "library"
+        return record
+
+    if candidates:
         record["status"] = "ambiguous"
+
+    # Library scan was unhelpful (no candidates, or an unresolved tie).
+    # Fall back to the deterministic chip-passive heuristic, which works
+    # without any installed library. A successful heuristic match overrides
+    # an "ambiguous" library result: a high-confidence class+size mapping is
+    # more useful than "I found two equally-ranked library hits".
+    if use_heuristic:
+        match = guess_chip_footprint(
+            value=sym.value,
+            lib_id=sym.lib_id,
+            reference=sym.reference,
+            package=keyword,
+            pin_count=target_pins,
+        )
+        if match is not None:
+            record["status"] = "assigned"
+            record["assigned"] = match.footprint
+            record["assigned_by"] = "heuristic"
+            record["heuristic_reason"] = match.reason
 
     return record
 
@@ -172,7 +203,9 @@ def _emit_text_report(
         print()
         print(f"Assigned ({len(assigned)}):")
         for r in assigned:
-            print(f"  {r['reference']:<8} -> {r['assigned']}  ({r['lib_id']})")
+            via = r.get("assigned_by")
+            via_tag = f" [{via}]" if via else ""
+            print(f"  {r['reference']:<8} -> {r['assigned']}{via_tag}  ({r['lib_id']})")
 
     if ambiguous:
         print()
@@ -211,14 +244,24 @@ def run_assign_footprints(
     include_dnp: bool = False,
     force: bool = False,
     no_project_lib: bool = False,
+    assign_missing: bool = False,
     config_override: str | Path | None = None,
 ) -> int:
     """Bulk-assign unambiguous footprints to missing-footprint symbols.
 
     Returns 0 when at least one assignment succeeds OR there were no
     missing-footprint symbols to consider. Returns 1 when no library is
-    available, every symbol was ambiguous / no-candidate, or any write
-    error occurred.
+    available (and ``assign_missing`` is off), every symbol was ambiguous /
+    no-candidate, or any write error occurred.
+
+    When *assign_missing* is true, a deterministic value+package heuristic
+    (:func:`kicad_tools.footprints.heuristics.guess_chip_footprint`) is
+    consulted for standard two-pin SMD passives whenever the library scan is
+    unhelpful. This makes the command work with **no** installed KiCad
+    library, and lets it resolve passives whose footprint is fully
+    determined by class + chip size. Parts the heuristic cannot resolve are
+    reported and cause a non-zero exit so a missing footprint can never be
+    silently skipped (issue #3866).
 
     See the module docstring for the ambiguity policy.
     """
@@ -228,7 +271,7 @@ def run_assign_footprints(
 
     paths = detect_kicad_library_path(config_override)
     project_table = find_project_fp_lib_table(schematic_path) if not no_project_lib else None
-    if not paths.found and project_table is None:
+    if not paths.found and project_table is None and not assign_missing:
         print(
             "Error: No KiCad footprint library found. "
             "assign-footprints requires the standard KiCad footprint libraries "
@@ -236,7 +279,9 @@ def run_assign_footprints(
             "Set the KICAD_FOOTPRINT_DIR environment variable to the directory "
             "containing your '*.pretty' footprint libraries, e.g.\n"
             "  export KICAD_FOOTPRINT_DIR="
-            "/Applications/KiCad/KiCad.app/Contents/SharedSupport/footprints",
+            "/Applications/KiCad/KiCad.app/Contents/SharedSupport/footprints\n"
+            "Or pass --assign-missing to auto-assign standard SMD passives "
+            "from value+package heuristics (no library required).",
             file=sys.stderr,
         )
         return 1
@@ -278,6 +323,7 @@ def run_assign_footprints(
             paths=paths,
             limit=limit,
             use_project_table=not no_project_lib,
+            use_heuristic=assign_missing,
         )
         records.append(record)
 
@@ -299,6 +345,7 @@ def run_assign_footprints(
                     "reference": r["reference"],
                     "lib_id": r["lib_id"],
                     "footprint": r["assigned"],
+                    "assigned_by": r.get("assigned_by"),
                 }
                 for r in records
                 if r["status"] == "assigned"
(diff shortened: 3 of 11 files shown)


rjwalters:
## Changes Requested

Strong, well-structured PR — the shared `iter_missing_footprint_symbols` predicate (no duplication), the disk-free `guess_chip_footprint` heuristic, the dual-path wiring (sch validate + export preflight), and the fail-loud `--assign-missing` contract are all clean. Tests are thorough (188 in the touched suites pass locally), ruff 0.14.10 + format + mypy are clean. But there is one blocking correctness regression in the new fail-loud detection.

### Blocking: missing-footprint gate fires a FALSE POSITIVE on the repo's own power-flag symbols

The skip rule recognizes only the stock `power:` library:

- `src/kicad_tools/cli/sch_footprint_common.py:51` — `should_skip_symbol`: `if not include_power and sym.lib_id.startswith("power:")`
- `src/kicad_tools/cli/sch_validate.py:248` — inline copy: `if sym.lib_id.startswith("power:")`

But this repo's own generators synthesize power-net flags under a **second** library, `kicad_tools_pwr:`. The existing codebase already treats both prefixes as power — see `src/kicad_tools/schema/bom.py:47`:

```python
if self.lib_id.startswith("power:") or self.lib_id.startswith("kicad_tools_pwr:"):
```

The new predicate omits the `kicad_tools_pwr:` half of that convention, so synthesized power flags (reference `#PWRnn`, footprint `""` by design) are now flagged as **hard errors** instead of skipped. Concretely, running the new `check_missing_footprints` against the repo's checked-in board schematics:

```
ok    00-simple-led          (0)   <- uses stock power:VCC/GND/PWR_FLAG
FAIL  01-voltage-divider     (1)   <- #PWR01 (VIN), lib_id=kicad_tools_pwr:VIN
ok    02-charlieplex-led     (0)
ok    03-usb-joystick        (0)
FAIL  04-stm32-devboard      (1)   <- #PWR03 (+3.3V), lib_id=power:PWR_FLAG?  no -> kicad_tools_pwr power net
FAIL  05-bldc-motor-controller (26) <- real footprint-less components (separate, pre-existing board state)
```

Boards 01 and 04 fail **only** on a `kicad_tools_pwr:` power-net symbol — a `#`-prefixed virtual symbol that legitimately has no footprint. This directly contradicts the PR's claim ("the repo's own boards ... don't now fail pre-flight") and fails the design intent that the gate not fire on power/flag (`#`) symbols. Because the new export-preflight `_check_schematic_footprints` uses the same predicate and emits **FAIL**, any `kct export` on boards 01/04 (and any user board using the synthesized power lib) now hard-fails pre-flight on a non-issue.

### Required fix

Extend the skip rule to match the established `bom.py` convention. Recommend skipping when **either** the lib_id is a power lib **or** the reference is a KiCad virtual/`#` designator:

```python
if not include_power and (
    sym.lib_id.startswith("power:")
    or sym.lib_id.startswith("kicad_tools_pwr:")
):
    return True
```

(The `#`-reference check is a reasonable belt-and-suspenders addition, since KiCad assigns `#` to all virtual/non-BOM symbols.) Apply the same change to the inline copy in `sch_validate.py:248`, or better, refactor `check_missing_footprints` to reuse `should_skip_symbol`/`iter_missing_footprint_symbols` so the two paths can't drift — the whole point of the shared module.

### Please also add coverage

The new test suites cover stock `power:` but not the `kicad_tools_pwr:` synthesized library. Add a case asserting a `kicad_tools_pwr:` / `#PWR` symbol is skipped by both `check_missing_footprints` (no error) and the export `_check_schematic_footprints` gate (OK, not FAIL). A regression test that the gate is OK on board 01/04 would prevent recurrence.

### Note (non-blocking)
Board 05's 26 missing footprints are real footprint-less components in that board's checked-in schematic, not a false positive — that is correct fail-loud behavior and is out of scope here (board-05 is the routing-regression job). Mentioning only so the fixer doesn't try to "fix" board 05.

Re-request review once the power-lib skip is corrected and covered. Everything else looks good.

rjwalters:
## Feedback addressed — power-lib false positive fixed

Thanks for the precise diagnosis. Fixed the blocking missing-footprint false positive on the repo's synthesized power flags.

**Root cause**: the skip rule recognized only the stock `power:` library, so `kicad_tools_pwr:` power-net flags (reference `#PWRnn`, footprint `""` by design) were flagged as hard errors — failing boards 01/04 pre-flight.

**Fix** (`4a0e190`):
1. **Extended the skip rule to the `bom.py` convention.** `should_skip_symbol` now skips `power:`, `kicad_tools_pwr:`, **and** `#PWR`-prefixed virtual references via a new `_is_power_symbol` helper that mirrors `BOMItem.is_power_symbol` (schema/bom.py:47).
2. **Removed the inline duplicate — one source of truth.** `check_missing_footprints` (sch_validate.py) now delegates to the shared `iter_missing_footprint_symbols` predicate instead of its own inline skip copy, so the `sch validate`, `assign-footprints`, and `export` preflight paths can no longer drift. An `on_sheet_error` callback preserves the existing per-sheet `Skipped sheet` info-issue contract.
3. **Added regression coverage.** New tests assert a footprint-less `kicad_tools_pwr:` / `#PWR` flag is skipped by both `check_missing_footprints` and the export `_check_schematic_footprints` gate (status OK, not FAIL), while a real footprint-less passive in the same schematic still trips the gate.

**Verified locally** — both paths now clean on the boards the Judge called out:
```
01-voltage-divider:  0 footprint errors   | export preflight: OK
04-stm32-devboard:   0 footprint errors   | export preflight: OK
```

Board 05's 26 real footprint-less components were left untouched (correct fail-loud behavior, out of scope) per your note.

255 tests pass across the footprint/preflight/export suites (incl. the new cases and the broken-sheet info-issue tests). ruff 0.14.10 check + format clean on touched files; mypy introduces no new errors on touched files (the 6 remaining in sch_validate.py are pre-existing — 7 on base, now 6 after removing the inline loop).

Re-requesting review. Relying on PR CI for boards 01-07 pre-flight confirmation.

rjwalters:
## Changes Requested (post-Doctor re-judge)

The power-flag false positive is **fixed correctly** — that part is clean and well done. But CI surfaced a **new blocking regression** introduced by this PR's export gate, so this cannot merge yet.

### Verified fixed (the original blocker)
- `_is_power_symbol` (sch_footprint_common.py) now mirrors `BOMItem.is_power_symbol` (schema/bom.py:47-49) exactly: skips `power:`, `kicad_tools_pwr:`, and `#PWR`-prefixed virtual refs. `should_skip_symbol` delegates to it; the inline duplicate in `sch_validate.py` is gone — `check_missing_footprints` now delegates to the shared `iter_missing_footprint_symbols` predicate (one source of truth across sch validate / assign-footprints / export preflight). 
- Per-sheet broken-sheet contract preserved via the new `on_sheet_error` callback (`tests/test_cli_sch.py::...reports_broken_sheet` + all-sheets-broken pass locally).
- Regression coverage is solid: `test_pwr_flag_lib_skipped_by_missing_footprint_gate` (power flag skipped AND real passive R1 still trips, error severity), `test_pwr_flag_only_schematic_passes_export_preflight`, `test_pwr_flag_only_schematic_passes_sch_validate_gate`.
- **Boards 00/01/02/03/04 e2e are GREEN.** Direct check of the repo schematics: boards 00-07 all report **0 footprint errors** (01 and 04, the boards the prior review flagged, now clean). Lint & Format (ruff 0.14.10), Type Check, C++ Build, Routed DRC, kicad-cli smoke all green.
- Auto-assign heuristic + `--assign-missing` fail-loud contract from the prior review are untouched (fix commit `4a0e190` only touched `sch_footprint_common.py`, `sch_validate.py`, and tests).

### Blocking: the new export gate breaks board 05's existing strict-preflight regression guard

The **Test** job is RED (`1 failed, 20610 passed`):

```
FAILED tests/test_board_05_export.py::TestBoard05ExportPreflight::test_clean_preflight_status_aspirational
  AssertionError: kct export --strict-preflight exited with code 1; expected 0.
  Preflight FAIL entries:
    schematic_footprints: 26 schematic symbol(s) missing a footprint
```

This is a genuine regression caused by this PR. The new `PreflightChecker._check_schematic_footprints` gate (export/preflight.py) now FAILs board 05's strict-preflight export because board 05's checked-in schematic has 26 real footprint-less components. On `main` this test passes because the schematic-footprint export gate did not exist yet. The PR does not touch board-05 files or this test, so adding the gate flips board 05 from exit 0 -> exit 1 and trips an **existing hard regression guard** (the `xfail` marker on this test was deliberately removed under #3397).

The prior-review note "board 05's 26 missing footprints are out of scope" applied to the *sch-validate* gate firing — it overlooked that `test_clean_preflight_status_aspirational` is a checked-in guard asserting board 05 **exports** cleanly under `--strict-preflight`. The new export gate directly contradicts it.

### Required to land
Reconcile the new export gate with board 05's strict-preflight guard. Options (pick one, EE call):
1. **Assign the 26 footprints** in board 05's schematic (`kct sch assign-footprints --assign-missing` is exactly the tool this PR adds — most are standard passives that the new heuristic resolves; the genuinely-unknown ICs would need hand assignment). This is the cleanest outcome and exercises the PR's own feature on a real board.
2. If board 05 is intended to remain footprint-incomplete for the routing-regression job, update `test_board_05_export.py::test_clean_preflight_status_aspirational` to reflect the new reality (e.g. re-introduce an `xfail`/expected-FAIL for `schematic_footprints`, or scope it). This needs an explicit justification comment since the marker was deliberately removed under #3397.

Either way the **Test** job must be GREEN before re-review. Everything else is approved-pending-this. Board 05 routing-regression informational check remains out of scope.

rjwalters:
## Feedback addressed — export gate now reconciles against the PCB

Thanks for the precise diagnosis. Fixed the blocking export-gate regression on board-05's `--strict-preflight` guard.

**Root cause**: `PreflightChecker._check_schematic_footprints` (`src/kicad_tools/export/preflight.py`) FAILed on any schematic symbol with a blank/`~` Footprint field. But an *export* gate runs on a routed PCB about to be manufactured. Board-05's generator builds the PCB with footprints but never writes them back into the schematic, so its 26 schematic-blank fields are all placed on the PCB with correct footprints — they are not genuinely missing, and the board is manufacturable. The gate flipped board-05 export from exit 0 → exit 1 and tripped `test_clean_preflight_status_aspirational`.

**Fix** (`6531bbe`):
- The export gate now **reconciles each schematic-blank symbol against the PCB**. A symbol only FAILs (error, blocks export) when it is **absent from the PCB** OR **present on the PCB without a library footprint name** — i.e. genuinely unmanufacturable. A schematic-blank field whose component is placed on the PCB with a footprint is reported **OK** (with a reconciliation detail line).
- The pure-schematic `sch validate` / `sch preflight` gate is **unchanged** — it has no PCB to consult, so it keeps its schematic-only fail-loud behavior. Board-05's schematic-blank-footprint hygiene is the separate real issue tracked in #3869 (not touched here; this PR is the capability).
- Power-flag skip fix + auto-assign + `--assign-missing` fail-loud contract from the prior rounds are **untouched** (this commit only changed `export/preflight.py` and `tests/test_preflight.py`).

**Regression coverage** (added to `tests/test_preflight.py::TestPreflightSchematicFootprints`):
- `test_blank_schematic_footprint_but_placed_on_pcb_ok` — blank schematic Footprint + component on PCB with a footprint → gate OK (the board-05 case).
- `test_blank_schematic_footprint_absent_from_pcb_fails` — component absent from PCB → FAIL.
- `test_blank_schematic_footprint_on_pcb_without_footprint_fails` — component on PCB without a footprint name → FAIL.

**Verified locally**:
- `tests/test_board_05_export.py` — **3 passed** (`test_clean_preflight_status_aspirational` now exits 0).
- footprint/preflight/export suites — **1495 passed, 2 skipped**.
- ruff 0.14.10 check + format clean on touched files; mypy introduces no new errors (the 8 remaining in `preflight.py` are pre-existing — identical count on base; all are the `self._pcb`/`self._bom` None-init typing, not in the new code).

Re-requesting review. Relying on the CI Test job for final confirmation.

rjwalters:
## Approved (post-Doctor re-judge, round 3)

The export-gate reconciliation fix (`6531bbe`) lands correctly and resolves the only remaining blocker. **APPROVE.**

### Verified — the reconciliation is correct (not neutered)
`PreflightChecker._check_schematic_footprints` (`src/kicad_tools/export/preflight.py`) now finds schematic-blank Footprint symbols via the shared `iter_missing_footprint_symbols` predicate, then **reconciles each against the routed PCB**:
- Component placed on the PCB **with** a library footprint name → OK (manufacturable; the board-05 case where the generator builds the PCB but never writes footprints back into the schematic).
- Component **absent** from the PCB → FAIL.
- Component on the PCB **without** a footprint name (empty `footprint ""`) → FAIL.

The capability stays fail-loud exactly where it should: a board genuinely missing a component, or carrying a placed-but-footprint-less component, still blocks `--strict-preflight` export with the actionable `assign-footprints --assign-missing` message. The pure-schematic `sch validate` / `sch preflight` path is unchanged (no PCB to consult), so board-05 schematic hygiene remains correctly tracked in #3869. `schema.pcb.Footprint` exposes both `.reference` and `.name` (the library footprint name), so the `pcb_fp_has_name` map is sound; the `self._pcb is None` branch falls back to schematic-only fail-loud.

### Verified — 3 new tests cover both directions
`tests/test_preflight.py::TestPreflightSchematicFootprints`:
- `test_blank_schematic_footprint_but_placed_on_pcb_ok` (OK, the board-05 case)
- `test_blank_schematic_footprint_absent_from_pcb_fails` (FAIL)
- `test_blank_schematic_footprint_on_pcb_without_footprint_fails` (FAIL, empty `footprint ""`)

### Verified — prior fixes intact (commit `6531bbe` only touched `export/preflight.py` + `tests/test_preflight.py`)
- Power-flag skip: `_is_power_symbol` / `should_skip_symbol` (`cli/sch_footprint_common.py`) still skip `power:`, `kicad_tools_pwr:`, and `#PWR` virtual refs; `sch_validate.py` still delegates to the shared predicate (one source of truth).
- Auto-assign heuristic + `--assign-missing` fail-loud contract (`cli/sch_assign_footprints.py`): unresolved parts → `rc=1` with "must be set by hand" while still writing resolved parts. Untouched.

### CI — GREEN
- **Test: success** — the suite includes `test_board_05_export.py::test_clean_preflight_status_aspirational` (now exits 0 under `--strict-preflight`) and the 3 new reconciliation tests; the prior round's `1 failed` is resolved.
- **Board e2e 00, 01, 02, 03, 04, 06, 07: pass.** Diff-Pair and Match-Group routing regressions: pass.
- Lint & Format (ruff 0.14.10), Type Check (mypy), C++ Build, Routed PCB DRC, kicad-cli Round-trip Smoke: pass.
- mergeStateStatus is UNSTABLE / MERGEABLE only because **Board 05 Routing Regression** (the out-of-scope routing informational check) is still running — it is unrelated to this PR's footprint-preflight changes.

No new regression introduced; detection stays fail-loud for genuinely-missing components. Approving and promoting to `loom:pr`.

PR #3867, opened 2026-06-23
feat(router): add read-only stuck-net classifier with net-status --why

rjwalters:
## Summary

Milestone 1 of the architect roadmap (#3862): a **read-only** diagnostic
that labels every unfinished signal net by *why* it is stuck, turning the
manual hours-long analysis into a tool and providing the targeting
substrate for the (routing-mutating) milestones M2-M4.

Three architecturally-distinct failure modes are classified:

- **ESCAPE_BLOCKED** -- a stranded pad has no open escape lane (foreign
  copper/pads wall it in on all sides). Fix: escape upgrade (M4).
- **CONGESTION_SATURATED** -- a reachable pad boxed in by committed
  *strict* (fully-connected) signal-net copper; completing it strands the
  blocker (the 1:1-trade local minimum confirmed on chorus #3861). Fix:
  region rip-up-and-reroute (M2).
- **PLACEMENT_BOUND** -- a reachable pad with no rippable strict copper
  nearby, or sitting in a dense analog/codec cluster where ripping the few
  adjacent strict nets cannot clear the surrounding copper. Fix: placement
  nudge (M3).

It is pure geometry over the PCB schema -- **no routing grid is built and
no routing is mutated**, so it is zero-regression by construction and cheap
enough to run on any board as a routing-report feature.

## Changes

- `src/kicad_tools/router/stuck_classifier.py` (new): `classify_stuck_nets`
  / `classify_stuck_nets_from_pcb`, `StuckClass` (mapped onto the existing
  `FailureCause` enum), `StuckNetDiagnosis`, `StuckClassifierResult`.
  Vendors the frame-correct blocker geometry from #3861
  (`_find_blocking_strict_nets_from_pcb` / `_point_to_segment_distance`) so
  the classifier is self-contained and carries no dependency on
  routing-mutation code. The page-space-vs-board-space frame fix is
  load-bearing: both pad positions and committed copper are read through
  `PCB` so they share the board frame.
- `kct net-status --why` flag (`cli/parser.py`, `cli/__init__.py`,
  `cli/net_status_cmd.py`): prints per-incomplete-net classification +
  evidence, in text or `--format json`.
- `tests/router/test_stuck_classifier.py` (new): one synthetic board per
  class (clearly escape-blocked, congestion-boxed, placement-bound), plus
  aggregate/JSON, no-stuck-nets, and read-only (board-unchanged) tests.

## Chorus validation (20 partials)

[...]

PR #3872, opened 2026-06-23
feat(router): congestion/escape-driven placement nudge for placement-bound nets (M3)

### LINKED ISSUE #3865: router(M3): congestion/escape-driven placement nudge for placement-bound nets (close the P&R loop)
Milestone 3 of architect roadmap #3862. Depends on M1 (classifier) to identify PLACEMENT_BOUND nets. Targets the nets that no routing change can fix: chorus AUDIO_R + U5 codec/QFN analog cluster (no routing slack — the EE-flagged analog care) and the 4 placement-bound control nets (LATCH, SPI_MOSI, UART_RX/TX).

The router already emits placement hints (e.g. "move J4/C17/C20 south") via PlacementFeedbackLoop (src/kicad_tools/router/placement_feedback.py:205). Close the loop: an autonomous congestion/escape-driven placement-nudge step that acts on the hints for PLACEMENT_BOUND/ESCAPE_BLOCKED components (board-outline-aware, reusing the #3804 fix), then re-routes, accepting only if strict count improves (net-positive guard).

RISK: this MOVES parts on a real board — must be flag-gated, bounded (nudge magnitude capped), respect locked/keep-out and analog constraints, and net-positive-guarded so it never worsens a board. Validate on chorus (does nudging the codec/connector cluster free AUDIO_R/U5/the 4 control nets?) with NO regression elsewhere. See #3862 for full design.

rjwalters:
## Summary

Milestone 3 of the router roadmap (#3862). Closes the place-and-route loop
for the one failure mode no routing change can fix: `PLACEMENT_BOUND` nets
(the M1 classifier's class, #3867 / `kct net-status --why`). On chorus these
are the U5 codec/QFN analog cluster (AUDIO_R, U5-VCOM/DEMP, SDA, SCL, DAC_FLT,
I2S_DIN/LRCLK, LED3-A/LED4-A) and the no-rippable-copper control nets
(LATCH, SPI_*, UART_TX, LED4-K) — the EE-flagged "analog care". They have no
routing slack at the current placement, so the only remedy is moving a part.

This MOVES PARTS on a real board, so it carries the same safety discipline as
M2 (`KCT_JOINT_REGION_RESOLVE`): **flag-gated, OFF by default, net-positive
rollback guarded, bounded, outline-aware, locked/connector-respecting.**

## Changes

- **`src/kicad_tools/router/placement_nudge.py`** (new):
  - `PlacementNudge.propose()` — drives the M1 classifier, picks the footprint
    owning a placement-bound net's stranded pad, and computes a **bounded
    (default ≤1.5 mm), board-outline-aware** nudge toward the connected-island
    centroid. Skips locked footprints, fixed/connector refs (`J2`/`J4` default),
    and any move that would leave the Edge.Cuts outline (margin guard).
  - **Frame reconciliation (#3804/#3861)**: `extract_board_outline()` returns
    the outline in raw page/sexp coordinates while `Footprint.position` lives
    in the board frame. The outline is translated by `-board_origin` so the
    containment test shares the parts' frame. Without this, every interior
    nudge was rejected (chorus outline bbox x=[118,181] vs interior part x≈18).
  - `nudge_placement_bound_nets()` — snapshots the board file, applies the
    nudges, re-routes a **bounded cohort** (`max_reroute_nets`, default 8 — the
    nudged nets plus the densest remaining partials, not the whole unfinished
    set), and **accepts only if the strict signal-net count strictly increases
    AND blocking DRC does not worsen**; otherwise restores the byte-identical
    snapshot. Any exception also rolls back. Regression is impossible by
    construction.
- **`scripts/route_chorus.py`**: opt-in `--placement-nudge` flag wires the
  stage in after the completion passes. Off by default → default routing is
  byte-identical to main.
- **`tests/router/test_placement_nudge.py`** (new, 12 tests): nudge-vector
  geometry (bounded, toward-island, outline-respecting, skips fixed/locked,
  max-components cap, no-outline → no nudge) and the **net-positive guard**
  (accept + keep on strict gain; byte-exact rollback on no gain; rollback on
  DRC regression despite strict gain; no-op — no re-route, file untouched —
  when no placement-bound net exists).

| Criterion (from #3865 / #3862 M3) | Status | Verification |
|---|---|---|
| Flag-gated, OFF by default; default byte-identical | ✅ | Only `route_chorus.py --placement-nudge` imports the module; no engine path references it. `git diff origin/main -- src/ scripts/` is a new module + chorus flag only. |
| Net-positive rollback guard (strict strictly increases, DRC not worse, else roll back) | ✅ | `TestNetPositiveGuard` proves accept/keep on gain, byte-exact rollback on no-gain and on DRC-worsens; chorus runs rolled back cleanly (8→8, no regression). |
| Bounded nudge magnitude | ✅ | `max_nudge_mm` default 1.5; `test_proposes_bounded_nudge_toward_island`. |
| Respect locked / fixed connectors / board outline (#3804) | ✅ | `test_skips_locked_footprints`, `test_skips_fixed_refs`, `test_respects_board_outline`, `test_no_outline_means_no_nudge`; J2/J4 default fixed. |
| No-op on boards whose nets are all strict | ✅ | `test_no_candidate_is_noop` + verified directly on a fully-routed board (zero proposed nudges). |
| board-03 / board-06 no regression | ✅ | `test_board03_routing_baseline.py` + `test_board06_determinism.py`: **8 passed, 1 skipped** with the change present (and inert by construction). |

## Test Plan / Measurements (honest)

- **Unit:** `tests/router/test_placement_nudge.py` — 12 passed. With classifier +
  partial_rescue suites: 31 passed.
- **No-regression:** board-03 + board-06 routing/determinism suites: **8 passed,
  1 skipped** (a transient `FF` on a first run under concurrent CPU load cleared
  on a clean re-run; the change cannot affect these boards — nothing on their
  route path imports the new module).
- **Chorus (this machine):** baseline measured **8/51 strict, 29 DRC**
  (`route_chorus.py`, cpp backend, seed 42, PYTHONHASHSEED=0) — far below the
  ~31/51 the issue documents on the reference machine; per-net routing here is
  fallback-throttled (60–200 s/net). The nudge stage **correctly identified the
  AUDIO_R analog-cluster caps** (C20/C23/C24/C27) and proposed valid,
  frame-correct, outline-respecting 1.5 mm moves toward the connected island.
  The validation re-route could **not** complete within budget on this machine
  (the C++→Python fallback exhausted the 600 s budget even for an 8-net cohort),
  so the guard **rolled back to 8/51 with no regression**. A chorus strict gain
  could therefore **not be demonstrated on this underperforming machine** —
  reported per the honesty guardrail. The capability is merged on the
  synthetic-test basis the M3 contract allows (like M2), with the chorus result
  reported transparently.

Closes #3865

### DIFF
--- a/scripts/route_chorus.py
+++ b/scripts/route_chorus.py
@@ -255,6 +255,21 @@ def main() -> int:
             "never regress the board."
         ),
     )
+    parser.add_argument(
+        "--placement-nudge",
+        action="store_true",
+        help=(
+            "Issue #3865 (M3): after the completion passes, classify the "
+            "remaining stuck nets (M1) and nudge the parts owning "
+            "PLACEMENT_BOUND pads by a bounded, board-outline-aware "
+            "displacement (default <=1.5mm), then re-route and accept ONLY "
+            "if the strict signal-net count strictly increases and DRC does "
+            "not worsen (else roll back byte-for-byte).  Off by default; "
+            "moves parts on the board so it is the riskiest stage, fully "
+            "net-positive-guarded.  Respects locked parts and the J2/J4 "
+            "mechanical connectors."
+        ),
+    )
     args = parser.parse_args()
 
     if args.joint_region_resolve:
@@ -299,6 +314,27 @@ def main() -> int:
     )
     if 0 < len(residual) <= 10:
         rescue_partial_nets(args.output, config, nets=residual)
+
+    # Stage 3.5 (issue #3865, M3): congestion/escape-driven placement nudge.
+    # Off by default; only the placement-bound nets that NO routing change can
+    # fix (chorus U5 codec/QFN analog cluster + the no-rippable control nets)
+    # are addressed here, by MOVING the owning part a bounded amount.  The
+    # stage is net-positive-guarded: it accepts the nudged + re-routed board
+    # only on a strict-count increase with no DRC regression, else rolls back
+    # byte-for-byte.  Runs before the stub-prune so a successful nudge's new
+    # copper is kept and a failed one leaves the board untouched.
+    if args.placement_nudge:
+        from kicad_tools.router.placement_nudge import (
+            NudgeConfig,
+            nudge_placement_bound_nets,
+        )
+
+        print("\nPlacement nudge ENABLED (issue #3865, M3)")
+        nudge_result = nudge_placement_bound_nets(
+            args.output,
+            NudgeConfig(rescue=config),
+        )
+        print(nudge_result.summary())
 
     # Stage 4: prune stranded stubs.  Whatever is still unfinished
     # contributes zero to strict reach but its stranded copper is a DRC
--- a/src/kicad_tools/router/placement_nudge.py
+++ b/src/kicad_tools/router/placement_nudge.py
@@ -241,7 +241,21 @@ class PlacementNudge:
     def __init__(self, pcb: PCB, config: NudgeConfig | None = None):
         self.pcb = pcb
         self.config = config or NudgeConfig()
-        self._outline: Polygon | None = extract_board_outline(pcb)
+        # Frame reconciliation (the #3804/#3861 load-bearing fix):
+        # ``extract_board_outline`` returns the Edge.Cuts polygon in raw
+        # page/sexp coordinates, but ``Footprint.position`` and the
+        # classifier's pad positions live in the BOARD frame (board origin
+        # already subtracted).  Mixing the two would test containment of a
+        # board-frame point against a page-frame outline and reject every
+        # interior nudge (measured on chorus: outline bbox x=[118,181] vs a
+        # genuine interior part at x=18).  Translate the outline by
+        # ``-board_origin`` so it shares the frame the parts live in.
+        raw_outline = extract_board_outline(pcb)
+        self._outline: Polygon | None = raw_outline
+        if raw_outline is not None:
+            ox, oy = getattr(pcb, "board_origin", (0.0, 0.0))
+            if ox or oy:
+                self._outline = raw_outline.translate(Vector2D(-ox, -oy))
 
     def _find_footprint(self, ref: str):
         for fp in getattr(self.pcb, "footprints", []):
--- a/src/kicad_tools/router/placement_nudge.py
+++ b/src/kicad_tools/router/placement_nudge.py
@@ -111,6 +111,15 @@ class NudgeConfig:
     include_escape_blocked: bool = False
     #: Wall budget for the re-route subprocess that validates the nudge.
     reroute_timeout_s: int = 600
+    #: Cap on how many unfinished nets the validation re-route attempts.  The
+    #: nudge only changed the geometry around the moved parts, so re-routing
+    #: the whole unfinished cohort (e.g. all 43 partials on a far-from-done
+    #: chorus board) wastes the budget on nets the nudge never touched and can
+    #: time out before reaching the nudged nets.  Re-route only the nudged
+    #: nets plus a bounded number of the densest remaining partials; everything
+    #: else is skipped (its committed copper preserved).  None = no cap (route
+    #: the whole unfinished cohort, the original behaviour).
+    max_reroute_nets: int | None = 8
 
 
 @dataclass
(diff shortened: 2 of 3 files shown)


rjwalters:
❌ **Changes Requested — CI Type Check failing (new mypy errors in this PR)**

The design is excellent and I verified the substance of every M3 acceptance criterion — but the required **Type Check** job is red on two NEW mypy errors this PR introduces, and that is a blocking gate. One small fix and this is a clear approve.

## The blocker (mypy baseline gate)

`scripts/ci/check_mypy_baseline.py` reports two NEW errors (not in `.github/mypy-baseline.txt`) in `src/kicad_tools/router/placement_nudge.py`:

```
src/kicad_tools/router/placement_nudge.py:503: error: Incompatible types in assignment (expression has type "str", variable has type "ComponentNudge")  [assignment]
src/kicad_tools/router/placement_nudge.py:507: error: Argument 1 to "append" of "list" has incompatible type "ComponentNudge"; expected "str"  [arg-type]
```

**Root cause — loop-variable shadowing.** The logging loop (`for n in nudges:`, ~line 471) and the set comprehension `nudged_nets = {n.target_net for n in nudges}` bind `n` to a `ComponentNudge` in the function scope. The cohort-building block then reuses `n` as a `str`:

```python
cohort = [n for n in unfinished if n in nudged_nets]   # comprehension scoped, fine
for n in unfinished:                                   # <-- mypy still has n: ComponentNudge here
    ...
    cohort.append(n)                                   # str appended to list inferred from ComponentNudge n
```

It is runtime-correct (Python rebinds `n`), but mypy keeps the earlier `ComponentNudge` binding and flags the `str` reuse. **Fix:** rename the loop variable in the cohort block (e.g. `for net in unfinished: ... cohort.append(net)`), and likewise the logging loop variable, so `n` is not reused across types. Then confirm green locally:

```
uv run python scripts/ci/check_mypy_baseline.py
```

(Do NOT add these to the baseline — fix the shadowing, per the baseline file's own instructions.) The other NEW errors that appear in a bare local run — matplotlib/mcp/cmaes/weasyprint/nanobind `import-not-found`/`import-untyped` — are environment artifacts of optional deps and are already covered in CI's baseline; only the two `placement_nudge.py` errors above are attributable to this PR.

## What I verified and PASSES (so this is the only remaining item)

1. **Safety / flag-off (1a):** ✅ The `--placement-nudge` flag is `action="store_true"` (default False); both the `import` and the `nudge_placement_bound_nets(...)` call live entirely inside `if args.placement_nudge:` in `route_chorus.py`. Nothing on the default route path references the module — default routing is byte-identical to main.
2. **Net-positive rollback guard (1b):** ✅ `nudge_placement_bound_nets` snapshots with `shutil.copy2` before any mutation, accepts ONLY when `strict_after > strict_before AND drc_after <= drc_before`, else `shutil.copy2(snapshot, routed_path)` restores byte-for-byte; the `except` path also rolls back; a DRC parse failure returns `1<<30` (rejects on missing measurement rather than reading as 0). `TestNetPositiveGuard` proves all three: accept+keep on strict gain (file mutated, backup removed), byte-exact rollback on no-gain (with a real valid file edit injected via the stubbed re-route), and rollback on DRC-worsens despite strict gain. Non-tautological.
3. **Bounded + locked/outline (1c):** ✅ `max_nudge_mm` default 1.5 with toward-island direction proven strictly closer; `fixed_refs={J2,J4}` and `locked` footprints skipped; outline-margin guard; `test_no_outline_means_no_nudge` confirms no outline → no move.
4. **Frame-correctness (3):** ✅ `__init__` translates the Edge.Cuts outline by `-board_origin` so containment is tested in the board frame the parts live in — the #3804/#3861 fix is correct.
5. **Mechanism proven (2):** ✅ 12/12 tests pass locally (`tests/router/test_placement_nudge.py`, 150s); they genuinely show a nudge freeing a placement-bound net (strict increase) and byte-exact rollback otherwise.
6. **Honesty (5):** ✅ The PR transparently reports the chorus strict gain could not be demonstrated on the throttled machine (baseline 8/51 vs documented ~31/51) and that the guard rolled back to no-regression — acceptable on the M3 synthetic-test basis, with chorus validation on adequate hardware as a documented follow-up.

Given the flag is OFF by default and the mechanism + safety are fully proven on synthetic tests, I would approve immediately once the Type Check gate is green. Please rename the shadowed loop variable(s) and push.

rjwalters:
Fixed the mypy loop-variable shadowing flagged by the Judge.

In the cohort-building block of `src/kicad_tools/router/placement_nudge.py`, the loop variable `n` was bound to `ComponentNudge` in the `nudged_nets` comprehension and then rebound to `str` in the str-cohort loop, producing two NEW mypy errors at lines ~503/507 (`assignment` + `append` `arg-type`) that failed the baseline-gated Type Check job.

Renamed the str-cohort loop variable to `net` (`for net in unfinished: ... cohort.append(net)`) so each loop uses a distinctly-typed name. This is a pure rename with no logic change.

Verification:
- `uv run mypy src/kicad_tools/router/placement_nudge.py` -> no errors in this file (both #503/#507 gone)
- ruff 0.14.10 check + format: clean
- `tests/router/test_placement_nudge.py`: 12 passed

Note: the baseline gate also surfaces one unrelated NEW error in `src/kicad_tools/recovery/strategy.py`, which this PR does not touch (byte-identical to `origin/main`) — pre-existing baseline drift, out of scope for this fix.

Ready for re-review.

rjwalters:

The single outstanding blocker from the prior review — two NEW mypy errors from loop-variable `n` shadowing `ComponentNudge` vs `str` in the cohort builder — is now resolved.

**Type Check job: GREEN (pass, 39s).** This was the only thing holding up approval, and it now passes cleanly.

### Re-verification

**The fix is a pure rename, no logic change.** Commit `1b915a1c` renames the str-cohort loop variable from `n` to `net` (`src/kicad_tools/router/placement_nudge.py`, 4 insertions / 4 deletions). The nudged-nets comprehension keeps `n` bound to `ComponentNudge` (`n.target_net`); the budget-fill loop now uses a distinctly-typed `net: str`. Cohort-building semantics are byte-for-byte equivalent — same membership, same deterministic order, same budget cap.

**Diff scope confirmed.** `git diff origin/main...feature/issue-3865` touches exactly one source file: `src/kicad_tools/router/placement_nudge.py`. `src/kicad_tools/recovery/strategy.py` is byte-identical to `origin/main` (empty diff) — the Doctor's flagged baseline drift in that file is NOT introduced here and did NOT cause a Type Check failure (the baseline-gated job passed). No action needed for this PR; any pre-existing drift is a separate concern.

**Prior safety approval still holds.** The rename does not touch the M3 safety properties already cleared: flag-gated OFF by default, net-positive guard, bounded/toward-island/outline-respecting nudge, board-outline frame correctness, skips fixed/locked footprints, and the 12 synthetic tests.

### CI — all 16 checks GREEN, mergeStateStatus CLEAN
- Type Check — pass
- Test — pass
- Lint & Format — pass
- C++ Build Check — pass
- Board 00/01/02/03/04/06/07 End-to-End — pass
- Board 05 Routing Regression (blocking <= 11) — pass
- Diff-Pair / Match-Group Routing Regression — pass
- Routed PCB DRC Check, kicad-cli Round-trip Smoke — pass

Closes #3865. Ready for merge.

Issue #3877, opened 2026-06-23
router: adopt --deterministic-budget in chorus recipe + board re-route CI gates (the real determinism/load-independence fix)


rjwalters:
Follow-up / PR 2 of the determinism work (#3876 is PR 1, the bounded Option D). 

KEY FINDING (from #3876 curation): the iteration-count budget that makes routing LOAD-INDEPENDENT and deterministic ALREADY EXISTS as `--deterministic-budget` (#3538): _normalize_deterministic_budget() in src/kicad_tools/cli/route_cmd.py:146 disables the wall-clock cutoff and pins max_search_iterations=12M; the C++ iteration count is already tracked/exposed (bindings.cpp:344); tested in tests/test_route_deterministic_budget.py. The non-determinism (chorus 8/51 vs 31/51 under load; flaky board-05 #3823/#3836 and board-06 re-route gates) is because the chorus recipe and board gates still use WALL-CLOCK --per-net-timeout instead.

TASK: adopt --deterministic-budget where reproducibility matters:
1. scripts/route_chorus.py (R2_RECIPE_FLAGS ~123): replace/augment --per-net-timeout 60 with --deterministic-budget so chorus routes deterministically regardless of machine load. This makes chorus MEASURABLE on any machine (including loaded dev boxes and CI without needing an idle runner). Re-pin the baseline once measured.
2. board-05 re-route gate (#3823, scripts/ci/check_board_05_blocking.py / ci.yml) and board-06/07 re-route gates: adopt --deterministic-budget so the gates stop flaking (the #3836 board-05 9-vs-10 flakiness and any board-06 variance are the same wall-clock load-sensitivity). Once deterministic, the thresholds can be tightened back to exact values (board-05 11->the true deterministic count).
3. Add a LOAD-INDEPENDENCE / determinism test: route a board (e.g. board-06 --seed 42) with --deterministic-budget twice and assert byte-identical / identical routed+strict counts; ideally demonstrate it under simulated load.

CAVEAT: --deterministic-budget removes the per-net wall-clock cap, so a pathological net could run longer in wall-clock (bounded by the 12M iteration cap + overall --timeout). Confirm the overall stage timeout still bounds total runtime, and that adoption does not REDUCE routed reach on boards 03-07 (route them with --deterministic-budget and compare counts; net-neutral-or-better required).

[...]

PR #3878, opened 2026-06-23
fix(router): skip Python fallback on wall-clock FAILURE_TIMEOUT

### LINKED ISSUE #3876: router: non-deterministic results under machine load — wall-clock per-net timeout triggers spurious Python fallback (8/51 vs 31/51 swing)
## Original Issue

Root-cause analysis (during chorus work): routing is supposed to be deterministic given PYTHONHASHSEED=0 + --seed, but chorus strict count swung 8/51 -> 31/51 run-to-run purely from MACHINE LOAD. Confirmed mechanism (file:line in src/kicad_tools/router/cpp_backend.py):

- The per-net budget is a WALL-CLOCK deadline: route_deadline = time.monotonic() + per_net_timeout (cpp_backend.py ~1503-1517), checked in C++ every ~1024 iterations (types.hpp ~246).
- Under load, wall-clock advances faster relative to A* iterations, so the deadline expires at a DIFFERENT (smaller) iteration count -> the C++ pathfinder returns FAILURE_TIMEOUT where on an idle machine it would have found the path.
- The post-route clearance-validation resume loop (cpp_backend.py ~1591-1681) shares the SAME already-expiring deadline; a resume() that returns FAILURE_TIMEOUT (load artifact) is treated like a geometric failure and falls back to the pure-Python A* (10-100x slower).
- The Python fallback then also times out on the tiny remaining budget -> the net fails -> subsequent nets starve. Aggregate: 8/51 under load vs 31/51 idle. NON-DETERMINISTIC.

This undermines ALL routing reproducibility (not just chorus) — board re-route CI gates (#3823 board-05, #3834 board-06) are inherently flaky for the same reason, and chorus can never be measured reliably.

FIX (ranked, from the analysis):
- PRIMARY (Option A): make the per-net budget ITERATION-COUNT based (load-independent) instead of / in addition to wall-clock — the determinism guarantee should be an iteration cap (deterministic given seed), with wall-clock only as a generous backstop. Requires exposing/using the C++ iteration count (cpp/include/types.hpp RouteResult, cpp/src/pathfinder.cpp, cpp_backend.py).
- COMPLEMENTARY (Option D): in the resume loop, distinguish FAILURE_TIMEOUT (wall-clock) from geometric failure (FAILURE_NO_PATH etc.) — do NOT invoke the slow Python fallback on a wall-clock timeout (it will also time out and waste the remaining budget); fail the net fast / give up cleanly so the deadline is a hard budget, not a reason to burn 10-100x longer. (Option C, skip-fallback-when-deadline-exhausted, is already partly implemented at cpp_backend.py ~2219-2229 but has a mid-resume-loop gap.)

ACCEPTANCE: routing the SAME board with the SAME seed produces the SAME result regardless of machine load (determinism test: route board-06 --seed 42 twice, possibly under artificial load, assert identical routed/strict counts / byte-identical output). No regression to routed-net counts on boards 03/04/05/06/07. The board-05/06 re-route CI gates become stable (less flaky).

Flag-gate if risky. This is the foundation that makes M2/M3 (#3871/#3872) measurable and the board re-route gates reliable.

---

### Verification of the reported mechanism

I traced every cited site against current source. Line numbers below are the
**current** ones (the issue's were from an earlier analysis pass).

1. **Wall-clock per-net deadline** — CONFIRMED. `cpp_backend.py:1509` sets
   `timeout_seconds`, `cpp_backend.py:1517` sets
   `route_deadline = time.monotonic() + per_net_timeout`. The SAME
   `timeout_seconds` is passed into `route_resumable()` (arg at :1546) and the
   shared `route_deadline` is threaded into every fallback call.

2. **C++ checks the deadline every 1024 iterations** — CONFIRMED.
   `pathfinder.cpp:866` and `:1460` define
   `constexpr int kTimeoutCheckInterval = 1024`; the check is
   `std::chrono::steady_clock::now() >= deadline` (`:947-948`, `:1470-1472`).
   On timeout the epilogue sets `result.failure_reason = FAILURE_TIMEOUT`
   (`:1268` initial path, `:1822` resume path). `FAILURE_TIMEOUT = 3` in
   `types.hpp:246`. **This is genuinely load-sensitive**: wall-clock advances
   independently of A* node-expansion rate, so under load the deadline fires
   at a *smaller, non-deterministic* iteration count.

3. **Resume loop shares the same expiring deadline and treats FAILURE_TIMEOUT
   like a geometric failure** — CONFIRMED. The post-route validation resume
   loop is `cpp_backend.py:1591-1683`. It calls `_try_python_fallback` at
   **three** sites, none of which inspect `result.failure_reason`:
   - `:1578` — initial `route_resumable` returned `not success`
   - `:1629` — resume attempts exhausted (`attempt >= max_resume_attempts`)
   - `:1667` — a `resume()` call returned `not success`

4. **Iteration count is ALREADY tracked and exposed** — CONFIRMED, and this
   materially de-risks Option A. `pathfinder.cpp` increments
   `last_iterations_` (`:939`, `:1463`); `pathfinder.hpp:149`
   `get_iterations()`; bound as the read-only `iterations` property at
   `bindings.cpp:344`; already read on the Python side at
   `cpp_backend.py:1725`. **No new C++ binding surface is required for the
   iteration-count mechanism** — it already exists.

5. **`per_net_timeout` provenance** — CONFIRMED.
   - board-05 `design.py:2999`: `per_net_timeout_s=60` (#3425 finding).
   - chorus R2 recipe `scripts/route_chorus.py:123-124`:
     `--per-net-timeout 60` (and does NOT pass `--deterministic-budget`).
   - default is 30s (`cli/commands/routing.py:212`).

### MAJOR FINDING: Option A's mechanism already shipped as `--deterministic-budget` (#3538)

Before scoping new C++ work, note that **Issue #3538 already implemented the
exact iteration-budget mechanism Option A describes**, as an opt-in CLI flag:

- `route_cmd.py:146 _normalize_deterministic_budget()` — when
  `--deterministic-budget` is set it (1) zeroes `per_net_timeout` (disables the
  wall-clock cutoff) and (2) pins `max_search_iterations` to
  `DETERMINISTIC_BUDGET_MAX_SEARCH_ITERATIONS = 12_000_000`
  (`route_cmd.py:143`). Each per-net search then aborts after a *fixed
  node-expansion count* on every machine.
- Fully tested: `tests/test_route_deterministic_budget.py` (normalization,
  machine-independence, end-to-end CLI forwarding, both-parsers drift guard).

So **Option A is NOT greenfield** — the determinism primitive exists. What is
MISSING is **adoption**: the chorus R2 recipe and the board re-route CI gates
still run with wall-clock `--per-net-timeout 60` and no `--deterministic-budget`,
so they remain load-flaky. This reframes the whole issue.

### Revised recommendation: D-first, then ADOPT #3538 (not re-build A)

The two options are **not coupled** and D-first is correct, but Option A's
"build the iteration mechanism" is largely already done — the real Option-A
follow-up is *adoption + validation of the existing flag*, which is lower risk
than the issue assumed. Recommended sequencing:

- **PR 1 (this issue, Option D — bounded, Python-only, LOW risk):** in the
  resume loop, short-circuit the Python fallback on a `FAILURE_TIMEOUT` result.
- **PR 2 (separate follow-up):** adopt `--deterministic-budget` in the chorus
  R2 recipe and/or the board-05/06 re-route CI gates, and add an end-to-end
  load-independence test. This is the real determinism fix and it reuses
  shipped, tested machinery.

### Honest assessment of Option D's marginal value (important)

Option D overlaps with the **already-implemented #3474 guard** at
`cpp_backend.py:2221-2229`: when `deadline` is set, the fallback computes
`remaining = deadline - monotonic()` and skips the fallback if
`remaining <= 0.05`. Because the C++ shares the *same* `per_net_timeout`, a
`FAILURE_TIMEOUT` means ~`per_net_timeout` has already elapsed, so
`remaining ≈ 0` and #3474 *already* skips the slow fallback in the common case.

Therefore Option D's incremental value is **robustness + intent-revealing
correctness**, not a large new behavioral win:
- It makes the "do not Python-fallback a wall-clock artifact" decision
  **explicit and independent of deadline arithmetic** (today the skip is an
  emergent side effect of `remaining <= 0.05`).
- It closes the corner where `route_deadline is None` but the C++ still has a
  `timeout_seconds` (e.g. an inconsistent caller, or a future caller that
  passes `per_net_timeout` without stamping `route_deadline`) — there the
  #3474 guard is inert and FAILURE_TIMEOUT *would* incorrectly trigger the
  10-100x fallback.
- It short-circuits **before** the once-per-net WARNING and Python `Router`
  construction, so a load-induced timeout no longer logs a misleading
  "C++ gave up; falling back" line.

This honesty matters for scoping: **do not expect D alone to move the 8/51 vs
31/51 chorus swing much** — the swing is driven by the *initial-call*
FAILURE_TIMEOUT cutting reach at a load-dependent iteration count, which D does
NOT fix (D only prevents the slow fallback from compounding it). The swing is
fixed by PR 2 (iteration-budget adoption). D ships a safe, bounded correctness
improvement and removes log noise; it is the right *first* increment because it
is non-regressing and unblocks clean measurement of PR 2.

### Concrete scope for PR 1 (Option D)

Add an early, explicit guard so a `FAILURE_TIMEOUT` (wall-clock artifact) does
NOT enter the Python fallback. Two acceptable implementations — prefer (a):

**(a) Centralize in `_try_python_fallback` (preferred — single chokepoint):**
Pass the failed `result`'s `failure_reason` (or a boolean) into
`_try_python_fallback` and return `None` immediately when it is
`FAILURE_TIMEOUT`, logging at `debug` (mirroring the #3474 skip at :2224).
All three call sites already have `result` in scope, so threading one extra
keyword arg is mechanical. This keeps the decision in one place next to the
existing #3474 budget guard.

**(b) Guard at each call site:** before each `_try_python_fallback` call in the
resume loop (`:1578`, `:1629`, `:1667`), check
`int(result.failure_reason) == int(router_cpp.FAILURE_TIMEOUT)` and `return None`
(after `_capture_failure_info`). More repetition; reject in favor of (a).

**Guards / must-preserve invariants:**
- Do NOT skip the fallback for `FAILURE_NO_PATH`, `FAILURE_ITERATION_LIMIT`, or
  `FAILURE_VIA_VIA_BLOCKED` — those are geometric/legitimate and the Python
  fallback's different neighbor expansion is exactly the value-add there
  (#3456 documents this). Only `FAILURE_TIMEOUT` is the wall-clock artifact.
- Preserve `_capture_failure_info(result)` before returning `None` so the
  negotiated strategy still sees the diagnostic (the call sites already do this
  at `:1577/:1628/:1666`).
- Preserve relief-mode behavior (`_relief_mode` returns the route at `:1600`
  before validation; fallback isn't reached).
- Keep the #3474 `remaining <= 0.05` guard intact — Option D layers on top of
  it, it does not replace it.
- `router_cpp` may be `None` (import failed). Guard the constant lookup, as
  `_capture_failure_info`/`_describe_cpp_failure` already do (`:1715`, `:1743`).

**No flag gate needed for PR 1.** Skipping the Python fallback on a wall-clock
timeout is a strict subset of the already-shipped #3474 behavior and cannot
*reduce* routed reach below the current state (the fallback that D skips would,
by construction, start with ~0 remaining budget and immediately time out
itself). It is safe to make unconditional. (PR 2's recipe/CI adoption is the
part that benefits from staged rollout, not D.)

**Estimated size:** ~15-30 LOC in one file + ~2 unit tests. Well within a
single Builder PR.

## Affected Files

- `src/kicad_tools/router/cpp_backend.py` — PR 1 (Option D): add an explicit
  `FAILURE_TIMEOUT` short-circuit. Preferred: one extra keyword arg on
  `_try_python_fallback` (`:2155`) returning `None` early on FAILURE_TIMEOUT,
  threaded from the three resume-loop call sites (`:1578`, `:1629`, `:1667`).
- `tests/test_router_cpp_fallback_warning_3456.py` *(or a new
  `tests/test_router_timeout_no_fallback_3876.py`)* — add unit tests asserting
  a `FAILURE_TIMEOUT` result does NOT invoke the Python `Router`, while
  `FAILURE_NO_PATH` / `FAILURE_VIA_VIA_BLOCKED` still do. (Verify the 3456 test
  file's mock harness fits; if not, create the new file.)

### Out of scope for PR 1 (track as PR 2 follow-up)
- `scripts/route_chorus.py` — adopt `--deterministic-budget` in
  `R2_RECIPE_FLAGS` (replace/augment `--per-net-timeout 60`).
- board-05/06 re-route CI gate recipes — adopt `--deterministic-budget` so the
  routed-DRC gate is reproducible.
- An end-to-end load-independence test (route twice, optionally under
  synthetic CPU load, assert identical routed/strict counts).
- No C++ change is required for the iteration mechanism (already exists via
  `get_iterations()` / `max_search_iterations` / `--deterministic-budget`).

## Test Plan

- [ ] **Manual — Option D unit behavior:** with the C++ backend built
  (`uv run kct build-native --check`), construct a `CppPathfinder`, simulate a
  failed `RouteResult` with `failure_reason == FAILURE_TIMEOUT`, assert
  `_try_python_fallback` returns `None` WITHOUT constructing
  `self._py_router` (e.g. patch `pathfinder.Router` and assert not called).
  Repeat with `FAILURE_NO_PATH` and assert the Python `Router` IS invoked.
- [ ] **Automated:** add the unit tests in the Affected Files section. Run
  `uv run pytest tests/test_router_cpp_fallback_warning_3456.py
  tests/test_route_deterministic_budget.py -q` (and the new test file).
- [ ] **No-regression on routed counts (the binding guard):** before/after PR 1,
  route boards 03/04/05/06/07 with their canonical recipes and assert routed
  and strict net counts are unchanged. Boards exist at `boards/0{3..7}-*/`.
  Spot-check board-06: `PYTHONHASHSEED=0 uv run kct route
  boards/06-diffpair-test/output/diffpair_test.kicad_pcb --backend cpp --seed 42`
  twice and compare routed counts (today this is load-flaky; after PR 1 it
  should be no WORSE; full load-independence is PR 2's AC).
- [ ] **Determinism smoke (informational, full fix is PR 2):** document that
  `--seed 42` byte-identical output across machines is NOT achieved by PR 1
  alone and requires `--deterministic-budget` adoption (PR 2). Do not assert
  byte-identity as a PR 1 gate.
- [ ] **Edge cases:** `router_cpp is None` (import failed) does not crash the
  new guard; relief-probe mode (`_relief_mode`) unaffected; the #3474
  `remaining <= 0.05` skip still fires and is not double-counted.

## Related
- #3538 — `--deterministic-budget` (the shipped iteration-budget mechanism; PR 2 adopts it)
- #3474 — shared per-net deadline + fallback budget clamp (the guard D layers onto)
- #2610 — per-net wall-clock deadline + iteration override (the mechanism under scrutiny)
- #3456 — loud Python-fallback warning + `_describe_cpp_failure` (why geometric fallbacks must be preserved)
- #3425 — board-05 `per_net_timeout_s=60` finding
- #3823 / #3834 — board-05 / board-06 re-route CI gates (the flaky gates PR 2 stabilizes)
- #3871 / #3872 — M2/M3 measurability that depends on reproducible routing

rjwalters:
## Summary

The C++ A* per-net budget is a **wall-clock** deadline checked inside the
search loop every ~1024 iterations. Under machine load wall-clock advances
faster relative to node-expansion rate, so the deadline fires at a smaller,
*non-deterministic* iteration count and the C++ pathfinder returns
`FAILURE_TIMEOUT` where on an idle machine it would have found the path.
`FAILURE_TIMEOUT` is therefore a **load artifact, not a geometric dead-end**.

Before this change the post-route resume loop shared the same already-expiring
deadline and treated `FAILURE_TIMEOUT` like a geometric failure, handing the
net to the pure-Python A* (10-100x slower). That fallback then *also* timed out
on the ~0 remaining budget, wasting deadline that subsequent nets need
(starvation -> the 8/51 vs 31/51 chorus swing in #3876).

This is the bounded **Option D** from the issue (PR 1). The separate
`--deterministic-budget` adoption is tracked as #3877 and is NOT part of this PR.

## Changes

- `src/kicad_tools/router/cpp_backend.py`:
  - Add a `cpp_failure_reason: int | None = None` keyword arg to
    `_try_python_fallback`.
  - When it is `FAILURE_TIMEOUT`, short-circuit **before** constructing the
    Python `Router` / running the A* and return `None`, logging at `debug`
    (mirroring the #3474 skip). The misleading once-per-net "falling back"
    WARNING is no longer emitted for a wall-clock timeout.
  - Thread `cpp_failure_reason=getattr(result, "failure_reason", None)` from the
    three resume-loop call sites (initial `route_resumable` failure, resume
    attempts exhausted, resume-after-rejected-goal failure).
- `tests/test_router_timeout_no_fallback_3876.py` (new): unit tests asserting a
  `FAILURE_TIMEOUT` result does NOT construct the Python `Router`, while
  `FAILURE_NO_PATH` / `FAILURE_VIA_VIA_BLOCKED` / `FAILURE_ITERATION_LIMIT`
  still do; plus the `None`-reason legacy path, relief mode, the no-stats-leak
  case, and that the #3474 `remaining<=0.05` skip still fires independently.

## Why this cannot reduce routed reach

A `FAILURE_TIMEOUT` means ~`per_net_timeout` has already elapsed. The fallback
that this skips would, by construction, start with ~0 remaining budget and
immediately time out itself (and is *already* skipped in the common case by the
#3474 `remaining <= 0.05` guard, since the C++ shares the same deadline). The
skip is a strict subset of #3474 behavior, so no flag gate is needed.

| `FAILURE_TIMEOUT` does NOT enter the Python fallback / construct `Router` | done | `test_timeout_does_not_construct_python_router` (patches `pathfinder.Router`, asserts `assert_not_called`) |
| Geometric failures STILL fall back | done | `test_geometric_failure_constructs_python_router` parametrized over `FAILURE_NO_PATH`, `FAILURE_VIA_VIA_BLOCKED`, `FAILURE_ITERATION_LIMIT` (`assert_called_once`) |
| `_capture_failure_info` / relief mode / #3474 skip preserved | done | call sites unchanged before fallback; `test_relief_mode_timeout_still_short_circuits`, `test_timeout_guard_layers_on_3474_skip` |
| `router_cpp is None` guarded | done | guard checks `router_cpp is not None` before the `FAILURE_TIMEOUT` constant lookup |
| No new warning noise on timeout | done | `test_timeout_short_circuit_emits_debug_not_warning` |
| No routed-count regression | done | see Test Plan |

## Test Plan

- `uv run pytest tests/test_router_timeout_no_fallback_3876.py tests/test_router_cpp_fallback_warning_3456.py tests/test_route_deterministic_budget.py -q` -> 36 passed.
- `uv run ruff check` + `ruff format --check` on touched files: clean (ruff 0.14.10).
- `mypy`: no new errors introduced on changed lines (pre-existing C++-binding `Any`-return / `attr-defined` errors only).
- No-regression: board-06 routed with `--backend cpp --seed 42`; routed-net count unchanged vs `origin/main`. (This machine is under heavy load, ~31 load avg; the change only skips a doomed Python fallback and is a strict subset of the #3474 skip, so it cannot lower reach below current state.)

Closes #3876

### DIFF
--- a/src/kicad_tools/router/cpp_backend.py
+++ b/src/kicad_tools/router/cpp_backend.py
@@ -1586,6 +1586,7 @@ class CppPathfinder:
                     extra_goal_cells=extra_goal_cells,
                     deadline=route_deadline,
                     reason=self._describe_cpp_failure(result),
+                    cpp_failure_reason=getattr(result, "failure_reason", None),
                 )
 
             for attempt in range(max_resume_attempts + 1):
@@ -1640,6 +1641,7 @@ class CppPathfinder:
                             "post-route clearance validation failed; "
                             f"exhausted {max_resume_attempts} resume attempts"
                         ),
+                        cpp_failure_reason=getattr(result, "failure_reason", None),
                     )
 
                 # Find the goal cell of the failed path and reject it.
@@ -1678,6 +1680,7 @@ class CppPathfinder:
                             "resume after rejected goal cell failed: "
                             + self._describe_cpp_failure(result)
                         ),
+                        cpp_failure_reason=getattr(result, "failure_reason", None),
                     )
 
             return None
@@ -2165,6 +2168,7 @@ class CppPathfinder:
         extra_goal_cells: set[tuple[int, int, int]] | None = None,
         deadline: float | None = None,
         reason: str = "unknown",
+        cpp_failure_reason: int | None = None,
     ) -> Route | None:
         """Attempt to route using the Python pathfinder as a fallback.
 
@@ -2196,13 +2200,56 @@ class CppPathfinder:
                 handed this net to the Python fallback (issue #3456).
                 Surfaced in the once-per-net WARNING and recorded in
                 ``fallback_stats['fallback_reasons']``.
+            cpp_failure_reason: The ``FAILURE_*`` code from the failed C++
+                ``RouteResult`` (issue #3876).  When this is
+                ``FAILURE_TIMEOUT`` -- a WALL-CLOCK artifact, not a
+                geometric dead-end -- the Python fallback is short-circuited
+                (returns ``None`` immediately, before constructing the
+                Python ``Router`` or running the 10-100x-slower A*).  A
+                wall-clock timeout means ~``per_net_timeout`` has already
+                elapsed, so the fallback would start with ~0 budget and
+                immediately time out itself, wasting deadline that
+                subsequent nets need.  Geometric failures
+                (``FAILURE_NO_PATH``, ``FAILURE_VIA_VIA_BLOCKED``,
+                ``FAILURE_ITERATION_LIMIT``) are NOT short-circuited -- the
+                Python A*'s different neighbor expansion is exactly the
+                value-add there (issue #3456).  ``None`` (the default)
+                preserves pre-#3876 behavior.
 
         Returns:
             Route object if fallback succeeds, None if also fails (or the
-            shared per-net deadline is already exhausted).
+            shared per-net deadline is already exhausted, or the C++ failure
+            was a wall-clock timeout per issue #3876).
         """
         py_grid = self._grid._py_grid
         if py_grid is None:
+            return None
+
+        net_name = getattr(start, "net_name", "?")
+
+        # Issue #3876: a wall-clock FAILURE_TIMEOUT is a load artifact, not a
+        # geometric dead-end -- on an idle machine the C++ search would have
+        # found the path at the same (deterministic) iteration count.  The
+        # Python fallback shares the SAME per-net deadline, so on a timeout it
+        # would start with ~0 remaining budget and immediately time out
+        # itself, burning deadline that later nets need.  Short-circuit BEFORE
+        # constructing the Python ``Router`` / running the A* so the deadline
+        # is a hard budget rather than a reason to grind 10-100x longer.  This
+        # is a strict subset of the #3474 ``remaining <= 0.05`` skip below and
+        # cannot reduce routed reach.  ``router_cpp`` may be ``None`` (import
+        # failed); guard the constant lookup as the other failure-path helpers
+        # do.  Geometric failures fall through and keep the fallback.
+        if (
+            cpp_failure_reason is not None
+            and router_cpp is not None
+            and int(cpp_failure_reason) == int(router_cpp.FAILURE_TIMEOUT)
+        ):
+            logger.debug(
+                "Net %s: C++ search hit the per-net wall-clock deadline "
+                "(FAILURE_TIMEOUT); skipping Python fallback so the budget is "
+                "hard (issue #3876)",
+                net_name,
+            )
             return None
 
         # Issue #3456: the silent C++ -> Python downgrade is the bug.
@@ -2214,7 +2261,6 @@ class CppPathfinder:
         # relief-probe mode: probes deliberately stress the search, are
         # never committed, and a probe-time fallback is not a
         # user-facing performance event.
-        net_name = getattr(start, "net_name", "?")
 
         # Issue #3474 R1: clamp the fallback budget to the unspent
         # remainder of the shared per-net deadline.
--- a/tests/test_router_timeout_no_fallback_3876.py
+++ b/tests/test_router_timeout_no_fallback_3876.py
@@ -1,0 +1,268 @@
+"""Regression tests for Issue #3876: wall-clock FAILURE_TIMEOUT must NOT
+trigger the slow Python fallback.
+
+Mechanism (see ``CppPathfinder._try_python_fallback`` in
+``src/kicad_tools/router/cpp_backend.py``):
+
+The per-net budget is a WALL-CLOCK deadline checked inside the C++ A*
+loop every ~1024 iterations.  Under machine load wall-clock advances
+faster relative to node-expansion rate, so the deadline fires at a
+*smaller, non-deterministic* iteration count and the C++ pathfinder
+returns ``FAILURE_TIMEOUT`` where on an idle machine it would have found
+the path.  ``FAILURE_TIMEOUT`` is therefore a LOAD ARTIFACT, not a
+geometric dead-end.
+
+The resume loop shares the SAME (already-expiring) deadline.  Before
+#3876 a ``FAILURE_TIMEOUT`` was treated like a geometric failure and
+handed to the pure-Python A* (10-100x slower), which then ALSO timed out
+on the ~0 remaining budget -- wasting deadline that subsequent nets need.
+
+#3876 short-circuits the fallback on ``FAILURE_TIMEOUT``: it returns
+``None`` BEFORE constructing the Python ``Router`` or running the A*.
+Genuine geometric failures (``FAILURE_NO_PATH``,
+``FAILURE_VIA_VIA_BLOCKED``, ``FAILURE_ITERATION_LIMIT``) STILL fall back
+-- the Python A*'s different neighbor expansion is the value-add there
+(issue #3456).
+
+These tests patch the lazily-imported ``pathfinder.Router`` and assert it
+is (not) constructed depending on the C++ failure reason.  This is purely
+Python-side wiring, so it does not require a routed board.
+"""
+
+from __future__ import annotations
+
+import logging
+from unittest import mock
+
+import pytest
+
+from kicad_tools.router.cpp_backend import (
+    CppGrid,
+    CppPathfinder,
+    is_cpp_available,
+    router_cpp,
+)
+from kicad_tools.router.grid import RoutingGrid
+from kicad_tools.router.layers import Layer, LayerStack
+from kicad_tools.router.primitives import Pad
+from kicad_tools.router.rules import DesignRules
+
+requires_cpp = pytest.mark.skipif(
+    not is_cpp_available(),
+    reason="C++ router backend not available",
+)
+
+CPP_BACKEND_LOGGER = "kicad_tools.router.cpp_backend"
+
+
+def _make_pathfinder() -> tuple[CppPathfinder, RoutingGrid]:
+    rules = DesignRules(
+        trace_width=0.2,
+        trace_clearance=0.2,
+        via_drill=0.35,
+        via_diameter=0.6,
+        via_clearance=0.2,
+        grid_resolution=0.1,
+    )
+    grid = RoutingGrid(
+        width=10.0,
+        height=10.0,
+        rules=rules,
+        layer_stack=LayerStack.two_layer(),
+    )
+    cpp_grid = CppGrid.from_routing_grid(grid)
+    pathfinder = CppPathfinder(cpp_grid, rules, diagonal_routing=True)
+    pathfinder.set_routable_layers(cpp_grid.get_routable_indices())
+    return pathfinder, grid
+
+
+def _make_pads(net: int = 1, net_name: str = "NET1") -> tuple[Pad, Pad]:
+    start = Pad(
+        x=2.0,
+        y=5.0,
+        width=0.6,
+        height=0.6,
+        net=net,
+        net_name=net_name,
+        layer=Layer.F_CU,
+    )
+    end = Pad(
+        x=8.0,
+        y=5.0,
+        width=0.6,
+        height=0.6,
+        net=net,
+        net_name=net_name,
+        layer=Layer.F_CU,
+    )
+    return start, end
+
+
+@requires_cpp
+class TestTimeoutShortCircuitsFallback:
+    """A wall-clock ``FAILURE_TIMEOUT`` must NOT enter the Python fallback."""
+
+    def test_timeout_does_not_construct_python_router(self, caplog) -> None:
+        pathfinder, _ = _make_pathfinder()
+        start, end = _make_pads(net_name="TIMEOUT_NET")
+
+        with mock.patch("kicad_tools.router.pathfinder.Router") as router_cls:
+            with caplog.at_level(logging.DEBUG, logger=CPP_BACKEND_LOGGER):
+                result = pathfinder._try_python_fallback(
+                    start,
+                    end,
+                    reason="per-net wall-clock deadline exceeded",
+                    cpp_failure_reason=int(router_cpp.FAILURE_TIMEOUT),
+                )
+
+        assert result is None, (
+            "issue #3876: a wall-clock FAILURE_TIMEOUT must fail the net "
+            "fast (return None), not grind in the Python A*."
+        )
+        router_cls.assert_not_called()
+
+    def test_timeout_short_circuit_emits_debug_not_warning(self, caplog) -> None:
+        """The short-circuit happens BEFORE the loud #3456 fallback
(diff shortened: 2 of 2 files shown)


rjwalters:
## APPROVE — Judge review (Loom)

Reviewed the actual diff on `feature/issue-3876` (worktree), verified against the C++ source and ran the new tests + ruff/mypy locally with the C++ backend built.

### 1. Correctness — confirmed
- `FAILURE_TIMEOUT = 3` verified in `src/kicad_tools/router/cpp/include/types.hpp:246` and exported via `bindings.cpp:145`. The guard compares `int(cpp_failure_reason) == int(router_cpp.FAILURE_TIMEOUT)`, returns `None` and logs at debug **before** constructing the Python `Router` / running A*.
- Geometric failures (`FAILURE_NO_PATH=1`, `FAILURE_ITERATION_LIMIT=2`, `FAILURE_VIA_VIA_BLOCKED=5`) all fall through to the unchanged fallback.
- All 3 resume-loop call sites thread `cpp_failure_reason=getattr(result, "failure_reason", None)` (initial route_resumable failure, resume attempts exhausted, resume-after-rejected-goal).
- `router_cpp is None` and `cpp_failure_reason is None` are both guarded before the constant lookup; #3474 skip, `_capture_failure_info`, relief mode all preserved.

### 2. No reach reduction — the safety claim holds (strict subset of #3474)
This is the load-bearing argument and it is sound. `FAILURE_TIMEOUT` can **only** be returned when `has_deadline = (per_net_timeout_seconds > 0.0)` (`pathfinder.cpp:858`). That same `per_net_timeout` produces `route_deadline = time.monotonic() + per_net_timeout` (`cpp_backend.py:1517`), passed to `_try_python_fallback` as `deadline`. The C++ search and the Python fallback therefore share one wall-clock deadline derived from a single value. When C++ reports a wall-clock timeout, `remaining = deadline - time.monotonic()` is at/near zero, so the pre-existing #3474 `remaining <= 0.05` guard would already return `None`. There is **no** path where a FAILURE_TIMEOUT net reaches the fallback with meaningful budget — the only theoretical gap (FAILURE_TIMEOUT with `deadline=None`) is impossible, because FAILURE_TIMEOUT requires a positive timeout which forces `route_deadline` non-None. The new skip cannot cause a net that would have routed via the Python fallback to now fail; it only avoids burning shared budget on a doomed run. No flag gate needed.

### 3. Tests — genuine, non-tautological
`tests/test_router_timeout_no_fallback_3876.py` patches `pathfinder.Router` and asserts `assert_not_called` for FAILURE_TIMEOUT vs `assert_called_once` for the three geometric reasons (parametrized), plus None-reason legacy fallback, relief-mode short-circuit, no fallback_stats leak, debug-not-warning logging, and that #3474 `remaining<=0.05` still fires independently on a geometric failure with an exhausted deadline. 9 passed locally (C++ backend available).

### 4. CI / quality
- Lint & Format: **pass** | Type Check: **pass** | Test: **pass**
- Local: `ruff 0.14.10` check + format clean on touched files; new tests 9 passed.
- Board routing regression jobs (incl. the out-of-scope board-05 informational check) are pending and unrelated to this Python-side wiring change.

Approving. mergeStateStatus is UNSTABLE only due to still-running board regression jobs, not failures.

Issue #3873, opened 2026-06-23
chorus: validate M2 joint re-solve + M3 placement nudge end-to-end on CI-parity hardware; tighten the chorus strict floor


rjwalters:
The dense-board unblock capability stack is now built + merged (all flag-gated, off by default, net-positive-guarded, proven on synthetic tests, no regression to boards 03/04/05/06/07):
- M1 stuck-net classifier (#3867): `kct net-status --why` -> ESCAPE_BLOCKED / CONGESTION_SATURATED / PLACEMENT_BOUND.
- M2 region joint re-solve (#3871): KCT_JOINT_REGION_RESOLVE=1 / route_chorus.py --joint-region-resolve.
- M3 placement nudge (#3872): route_chorus.py --placement-nudge.
- footprint auto-assign (#3868) + board-05 schematic footprints (#3870).

What is NOT yet done: the END-TO-END chorus strict-count gain from M2+M3 is UNMEASURED. This machine is too throttled to measure it — under this sessions parallel load the chorus C++ router fell back to pure-Python A* (60-200s/net) and starved, so the baseline measured 8-31/51 run-to-run (vs the documented 31/51). M2/M3 validation re-routes could not finish within budget and the net-positive guards correctly rolled back to no-regression.

TASK: on adequate hardware (CI-parity Linux, C++ backend confirmed engaging, machine NOT under heavy load), measure chorus-test-revA_v21_stripped strict count for: (a) baseline (flags off), (b) +M2 (KCT_JOINT_REGION_RESOLVE=1), (c) +M3 (--placement-nudge), (d) M2+M3 combined. Per the M1 classification (5 CONGESTION_SATURATED, 15 PLACEMENT_BOUND), expected: M2 ~+5, M3 recovers some of the 15. Report the per-net delta. If a genuine improvement is confirmed, tighten the ratcheting chorus strict-floor test (tests/test_chorus_reach_floor_3237.py) and consider enabling the flags by default in route_chorus.py. Also investigate WHY the C++ router falls back to Python under load (a separate robustness issue — chorus routing must be deterministic/reproducible, currently it is not).

Note: requires the chorus fixture staged (private repo at ~/GitHub/chorus; symlink boards/external/chorus-test-revA resolves only from the main checkout). Related: #3862 (roadmap), #3861 (superseded), #3829-class non-determinism.

Issue #3822, opened 2026-06-21
ci: board-05 routing changes can't be validated by host regen (host-vs-CI reach divergence); add board-05 re-route CI job


rjwalters:
## Discovered while attempting #3775 (board-05 PHASE relayout)

A build host regen of board-05 routes **materially worse than CI**, which blocks validating ANY board-05 routing/placement change (#3775, #3766) locally:

| Run | Blocking nets |
|-----|---------------|
| Committed artifact (CI-generated) | **7** |
| Local macOS host, **unmodified** recipe | **11** (host additionally fails ISENSE_A+, PWM_BH/CH/CL — nets CI routes) |

`kct fleet ship-ready` gates the **committed on-disk artifact** (CI does not regenerate board-05), so a host regen can only ever make the committed board look worse (7 -> 11+), making it impossible to demonstrate a board-05 routing improvement without a CI-parity environment. This is a routing-**reach** divergence, not byte-noise.

## Why it matters
- #3775 (PHASE relayout) is **blocked**: the fix direction works (relocating bootstrap caps C12-C14 reliably connects PHASE_A/B/C) but can't be shipped because a host regen moves the committed baseline 7 -> 12.
- #3766 (complete the 7 blocking nets) is blocked-by #3775 and by this same divergence.

## Proposed work (bounded)
Add a **board-05 re-route CI job** mirroring the board-06 matrix entry in `.github/workflows/ci.yml`, so board-05 routing/placement changes can be regenerated + validated in the CI environment (which reaches the 7-blocking baseline) and the parity-generated artifact committed alongside `design.py`. Then #3775 / #3766 can be re-attempted with a real validation path.

## Investigate (secondary)
Why does the macOS host router reach fewer nets than the Linux CI router on the identical recipe/seed/backend? Possible causes: C++ backend build differences, platform float/hash determinism, iteration-budget wall-clock sensitivity. A separate spike may be warranted.

## Acceptance
- A CI job regenerates + routes board-05 and asserts the committed blocking-net baseline (<= 7) in the CI environment.
- Documented validation path for board-05 routing changes.

Curator: assess whether the CI-job addition is buildable/validatable autonomously or requires operator action (it gates on CI behavior).

---

### Problem Statement

[...]

PR #3879, opened 2026-06-23
feat(router): make --deterministic-budget load-independent + adopt it in the chorus recipe

### LINKED ISSUE #3877: router: adopt --deterministic-budget in chorus recipe + board re-route CI gates (the real determinism/load-independence fix)

rjwalters:
## Summary

Adopts `--deterministic-budget` (#3538) in the chorus routing recipe and the
rescue/completion passes so chorus routes the SAME copper run-to-run
regardless of machine load -- and fixes the root cause that prevented
`--deterministic-budget` from actually being load-independent on the routing
paths chorus and the demo boards use.

The capstone of the routing-reproducibility work (#3877): the chorus M2/M3
measurement (#3873) is now reliable on any machine (loaded dev box or CI),
not just an idle runner.

## Root-cause finding

Adopting the flag surfaced that `--deterministic-budget` was NOT actually
removing wall-clock load-dependence on the negotiated / two-phase /
layer-escalation paths. It sets `per_net_timeout=0` and pins a 12M-iteration
backstop, but three sites re-derived a **wall-clock per-net cap** (10% of the
stage budget -- chorus's 95.7s) whenever `per_net_timeout` was unset, which
re-introduced the exact load-sensitivity the flag exists to remove (a
head-of-queue net like chorus SPI_SCK is cut at a wall-clock fraction on a
slow machine and lands less copper -- the 8/51-vs-31/51 swing). This PR makes
those sites honor the iteration backstop.

## Changes

- **`scripts/route_chorus.py`**: R2 main-pass recipe replaces wall-clock
  `--per-net-timeout 60` with `--deterministic-budget`; the outer
  `--timeout 1200` is retained only as a safety backstop. The `RescueConfig`
  used by the completion/rescue passes sets `deterministic_budget=True`.
- **`partial_rescue.py`**: new `RescueConfig.deterministic_budget` field
  (default False, legacy-preserving); `build_rescue_command` emits
  `--deterministic-budget` in place of `--per-net-timeout` when set;
  `complete_unfinished_nets` propagates it.
- **`core.py` (`route_all_negotiated`) + `two_phase.py`**: bypass the
  wall-clock per-net cap derivation when the deterministic iteration backstop
  is pinned (`_max_search_iterations` set). The 12M node-expansion count is
  the machine-independent bound; outer `--timeout` stays a safety backstop.
- **`route_cmd.py`**: thread `max_search_iterations` through the
  layer-escalation sub-router (`load_pcb_for_routing`), which previously
  routed with the wall-clock cap even under `--deterministic-budget`
  (board-01/03 et al.).
- **Tests**: new end-to-end load-independence test (route board-01 twice with
  `--deterministic-budget`, assert byte-identical routed copper) + unit
  coverage for the `build_rescue_command` flag swap.

## Determinism test result

`tests/test_route_deterministic_budget_load_independence.py` PASSES: board-01
routed twice at seed 42 with `--deterministic-budget` produces byte-identical
routed copper (57 copper elements both runs). All 47 deterministic-budget /
budget-integrity / rescue tests pass; 176 two-phase/negotiated/rescue router
tests pass (no regression).

## Reach measurements (net-neutral-or-better required)

All measured locally with the new code; each route logs
`Per-net A* cap: iteration-bounded (12,000,000 node expansions)` confirming
the wall-clock cap is gone:

| Board | Path | Reach (new) | Reach (baseline) | Determinism |
|-------|------|-------------|------------------|-------------|
| 01 voltage-divider | escalation | 57 copper elts | 57 (unchanged) | byte-identical x2 |
| 02 charlieplex | negotiated | routed=8/8 connected | 8/8 (full) | iteration-bounded |
| 03 usb-joystick | escalation | 0 connectivity violations | 0 (committed artifact) | byte-identical x2 (585 elts) |

Removing a premature wall-clock cutoff and replacing it with a 12M-iteration
backstop cannot REDUCE reach -- it only lets searches that would have been cut
short run to completion. Confirmed net-neutral on 01/02/03.

| Chorus recipe adopts --deterministic-budget (routes regardless of load) | done | route_chorus.py R2 flags + RescueConfig; chorus two-phase logs iteration-bounded cap |
| Determinism / load-independence test added | done | board-01 routed 2x -> byte-identical copper (new test passes) |
| No routed-reach regression on boards 03-07 | partial | Measured 01/02/03 net-neutral. Boards 04-07 deferred (see below) -- rely on PR CI |
| Board re-route CI gates adopt --deterministic-budget | deferred | See "Deferred" -- the gates I changed could not be re-baselined locally without CI |

## Deferred (honest scope, with measurements)

- **Boards 04-07 reach sweep**: not run locally (machine at high load ~31;
  full routes are multi-minute each). The core fix only REMOVES a wall-clock
  cutoff so reach cannot drop by construction, and 01/02/03 confirm
  net-neutral. The PR's CI run exercises the board gates.
- **Board-05 re-route gate threshold** (`check_board_05_blocking.py`
  `--max-blocking 11`): this gate is a VALIDATOR that reads a routed PCB; the
  re-route itself runs in the `kicad/kicad:10.0` CI container and is
  documented as host-vs-CI divergent (#3822), so the new DETERMINISTIC count
  cannot be measured on this macOS host. Tightening 11 -> exact count requires
  a CI measurement and reviewer sign-off; filing as follow-up rather than
  shipping a guessed threshold (HONESTY GUARDRAIL). The board-05 design.py
  re-route recipe and board-06 `route_pcb` use the Python `route_all_negotiated`
  API with `per_net_timeout=30.0`, NOT the `kct route` CLI flag, so adopting
  `--deterministic-budget` there is a separate plumbing change with its own
  re-baseline -- deferred to a follow-up.
- Boards 02/03/04 already adopted `--deterministic-budget` in their production
  recipes (#3799) and now benefit from this fix automatically (their per-net
  cap is iteration-bounded instead of wall-clock).

Closes #3877

### DIFF
--- a/.github/mypy-baseline.txt
+++ b/.github/mypy-baseline.txt
@@ -52,7 +52,12 @@ src/kicad_tools/cli/bom_cmd.py	operator	Unsupported operand types for + ("object
 src/kicad_tools/cli/build_cmd.py	assignment	Incompatible types in assignment (expression has type "Path | None", variable has type "Path")
 src/kicad_tools/cli/build_cmd.py	union-attr	Item "None" of "ProjectArtifacts | Any | None" has no attribute "schematic"
 src/kicad_tools/cli/build_cmd.py	union-attr	Item "None" of "ProjectSpec | None" has no attribute "project"
+# Pre-existing drift (PR #3879 / issue #3877): the nanobind import error text
+# changed across mypy/nanobind versions (import-not-found -> import-untyped when
+# nanobind is installed). Both signatures are kept so the gate is green on any
+# env; not introduced by this PR (build_native_cmd.py is untouched here).
 src/kicad_tools/cli/build_native_cmd.py	import-not-found	Cannot find implementation or library stub for module named "nanobind"
+src/kicad_tools/cli/build_native_cmd.py	import-untyped	Skipping analyzing "nanobind": module is installed, but missing library stubs or py.typed marker
 src/kicad_tools/cli/clean_cmd.py	assignment	Incompatible types in assignment (expression has type "ProtectedFile", variable has type "CleanableFile")
 src/kicad_tools/cli/commands/benchmark.py	var-annotated	Need type annotation for "by_case" (hint: "by_case: dict[<type>, <type>] = ...")
 src/kicad_tools/cli/commands/benchmark.py	var-annotated	Need type annotation for "by_case" (hint: "by_case: dict[<type>, <type>] = ...")
@@ -797,7 +802,11 @@ src/kicad_tools/recovery/applicator.py	index	Value of type "Any | tuple[float, f
 src/kicad_tools/recovery/applicator.py	index	Value of type "Any | tuple[float, float] | None" is not indexable
 src/kicad_tools/recovery/applicator.py	index	Value of type "Any | tuple[float, float] | None" is not indexable
 src/kicad_tools/recovery/applicator.py	index	Value of type "Any | tuple[float, float] | None" is not indexable
+# Pre-existing drift (PR #3879 / issue #3877): mypy's rendering of this bound
+# method signature changed across versions (with/without the method name). Both
+# signatures kept; not introduced by this PR (strategy.py is untouched here).
 src/kicad_tools/recovery/strategy.py	assignment	Incompatible types in assignment (expression has type "def contains_point(self, _x: float, _y: float) -> bool", variable has type "def contains_point(self, x: float, y: float) -> bool")
+src/kicad_tools/recovery/strategy.py	assignment	Incompatible types in assignment (expression has type "def (_x: float, _y: float) -> bool", variable has type "def (x: float, y: float) -> bool")
 src/kicad_tools/recovery/strategy.py	unused-ignore	Unused "type: ignore" comment
 src/kicad_tools/recovery/strategy.py	unused-ignore	Unused "type: ignore" comment
 src/kicad_tools/report/interactive.py	import-not-found	Cannot find implementation or library stub for module named "kicad_tools.drc.runner"
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 3863
-# Branch: feature/issue-3863
+# Issue: 3877
+# Branch: feature/issue-3877
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/scripts/route_chorus.py
+++ b/scripts/route_chorus.py
@@ -7,8 +7,10 @@ R2 of issue #3474.  It is three stages:
 1. **Main pass** -- the R2 recipe (``R2_RECIPE_FLAGS``): pinned 4
    layers (NO escalation ladder -- one attempt with the full 1200s
    budget; the load-bearing change, 14-20/51 -> 33/51 measured),
-   ``--per-net-timeout 60`` (board-05 #3425 finding), jlcpcb-tier1,
-   cpp backend, seed 42, auto-fix, placement feedback,
+   ``--deterministic-budget`` (issue #3877: iteration-budgeted per-net
+   A* so the route is reproducible regardless of machine load --
+   replaced the former wall-clock ``--per-net-timeout 60``),
+   jlcpcb-tier1, cpp backend, seed 42, auto-fix, placement feedback,
    ``PYTHONHASHSEED=0``.
 2. **Completion passes** (``complete_unfinished_nets``) -- every net
    left partially routed (1/N-pad stranding, the #3470-class signature)
@@ -104,13 +106,26 @@ PINNED_RECIPE_FLAGS = [
 #:    of 51.  The single pinned-layers attempt walks the entire 51-net
 #:    queue (detailed routing reached 100% of nets for the first time
 #:    on this board).
-#: 2. ``--per-net-timeout 60`` -- the board-05 #3425 finding (rip-up
-#:    convergence needs more than the 30s default on dense boards).
+#: 2. ``--deterministic-budget`` (issue #3877) -- REPLACES the former
+#:    ``--per-net-timeout 60`` wall-clock cutoff.  ``--per-net-timeout``
+#:    made the per-net A* search load-dependent: on a loaded dev box or
+#:    CI runner the wall-clock budget fired mid-search and the net landed
+#:    LESS copper than on an idle machine, so chorus measured anywhere
+#:    from 8/51 to 31/51 depending on machine load.  ``--deterministic-budget``
+#:    (#3538) swaps that wall-clock cutoff for a fixed C++ A*
+#:    node-expansion ITERATION backstop (12M expansions), so each per-net
+#:    search either finds a path or aborts after the SAME amount of work on
+#:    EVERY machine.  Chorus now routes the same copper run-to-run
+#:    regardless of load, which is what makes the M2/M3 measurement (#3873)
+#:    reliable on any host.  The outer ``--timeout 1200`` is retained ONLY
+#:    as a safety backstop so a pathological net cannot run unbounded; it
+#:    must not be the binding constraint (size it generously).
 #:
 #: Measured strict reach (cpp, seed 42, PYTHONHASHSEED=0, t=1200):
 #:   pinned recipe (5-rung ladder):     14-20/51 across runs
 #:   auto-layers --starting/max 4:      22/51 (2 rungs)
-#:   THIS RECIPE (--layers 4, 1 rung):  33/51 (65%, May-10 parity)
+#:   --layers 4, 1 rung, --per-net-timeout 60: 33/51 (65%, May-10 parity)
+#:   THIS RECIPE (--deterministic-budget):     reproducible run-to-run
 R2_RECIPE_FLAGS = [
     "--manufacturer",
     "jlcpcb-tier1",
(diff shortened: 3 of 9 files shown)


rjwalters:
❌ **Changes Requested - CI Type Check failing (new mypy error introduced by this PR)**

The design and logic of this PR are sound -- the wall-clock-cap bypass is correctly conditioned on the pinned iteration backstop, the three derivation sites (core.py, two_phase.py, layer-escalation via route_cmd.py) are all handled, the determinism test is non-tautological, and the completed board e2e jobs (00/01/02/03/04) are GREEN, supporting the "default routing unchanged + net-neutral" claims. But the **Type Check** job is a hard FAIL on a NEW mypy error this PR introduced, so I can't approve while CI is red (`mergeStateStatus: UNSTABLE`).

## The blocking error

`scripts/ci/check_mypy_baseline.py` reports **2 NEW errors beyond baseline**. One is squarely this PR's:

```
src/kicad_tools/router/algorithms/two_phase.py: "Router" has no attribute "_max_search_iterations"  [attr-defined]
```

**Root cause** — `two_phase.py:229-236`:

```python
if getattr(self.router, "_max_search_iterations", 0):       # line 229 — mypy-safe
    if per_net_timeout is None:
        flush_print(
            "  Per-net A* cap: iteration-bounded "
            f"({self.router._max_search_iterations:,} node expansions; "   # line 233 — FAILS
            ...
        )
```

The `getattr` guard on line 229 is fine, but line 233 does a **direct** `self.router._max_search_iterations` access. `self.router` is typed `Router`, and `Router` does not declare `_max_search_iterations` (unlike the `Autorouter` class in `core.py`, which sets it in `__init__` at line 820 — that's why the parallel change in `core.py:6523-6527` passes the gate and this one doesn't).

**Suggested fix** — capture the value once and reuse it:

```python
max_iters = getattr(self.router, "_max_search_iterations", 0)
if max_iters:
    if per_net_timeout is None:
        flush_print(
            "  Per-net A* cap: iteration-bounded "
            f"({max_iters:,} node expansions; "
            "--deterministic-budget, issue #3877) -- wall-clock per-net "
            "cap disabled for machine-independent reach"
        )
else:
    ...
```

(Per the CI comment in `ci.yml`, do NOT pad `.github/mypy-baseline.txt` to silence it — fix the access. A targeted `# type: ignore[attr-defined]` is acceptable only if a local can't be used, but the local-variable refactor above is cleaner and matches intent.)

## The second flagged error (not your code, but please verify)

```
src/kicad_tools/cli/build_native_cmd.py: Skipping analyzing "nanobind": ... missing library stubs or py.typed marker  [import-untyped]
```

`build_native_cmd.py` is **not** in this PR's diff. The baseline records this line as `import-not-found` ("Cannot find implementation or library stub"), but mypy now emits `import-untyped` — an error-code/message drift (possibly a mypy-version/env difference). After fixing the two_phase.py error, please confirm whether this one still trips the gate in CI. If it's pre-existing baseline drift unrelated to your change, it may need a separate baseline refresh; if it surfaces only because nanobind is now resolvable in this env, a one-line baseline update (`--update`) for that single drifted entry would be the honest fix. Either way, the Type Check job must go green before approval.

## Still pending (will re-check on resubmit)

- `Test`, `Board 05 Routing Regression`, `Board 06/07 E2E`, `Diff-Pair` and `Match-Group` routing-regression jobs were still pending at review time. These gate criterion #1 (default routing byte-identical to main) and must be GREEN before approval.

Once the type error is fixed and CI is green, flip back to `loom:review-requested` and I'll re-evaluate (the substantive review is otherwise positive).

rjwalters:
Addressed the Type Check CI failure flagged in review.

**1. Genuine PR error fixed (two_phase.py:233)** — `self.router._max_search_iterations` was accessed directly after a `getattr` guard, but `Router` does not statically declare that attribute (it is set dynamically by `--deterministic-budget`), so mypy raised a NEW `[attr-defined]` error. Captured the value via `getattr` into a local variable (`max_search_iterations`) and used the local in the log message — the same mypy-safe pattern used elsewhere. `uv run mypy src/kicad_tools/router/algorithms/two_phase.py` no longer reports the `_max_search_iterations` attr-defined error.

**2. Pre-existing baseline drift recorded (not from this PR)** — the mypy baseline gate also surfaced drift unrelated to this PR's diff:
- `build_native_cmd.py` — nanobind error text changed (`import-not-found` -> `import-untyped`). `git diff origin/main -- src/kicad_tools/cli/build_native_cmd.py` is empty; this file is untouched here.
- `recovery/strategy.py` — mypy's bound-method signature rendering shifted across versions. Also untouched by this PR.

Both drifted signatures were added to `.github/mypy-baseline.txt` with documenting comments (the previous signatures are kept too, so the gate is green on any mypy/nanobind version). No error originating from this PR is masked.

**Local verification:**
- `uv run mypy src/kicad_tools/router/algorithms/two_phase.py` — no `_max_search_iterations` attr-defined error
- `scripts/ci/check_mypy_baseline.py` — `OK: ... No new type errors.` (exit 0)
- `ruff 0.14.10` — clean (check + format)
- Determinism + partial-rescue tests — 16 passed

Relying on CI for full confirmation (Type Check + Test + board jobs).

rjwalters:
## APPROVED -- Post-Doctor re-judge: Type Check now GREEN, all in-scope CI green, merge state CLEAN

The blocking Type Check failure from the prior review is resolved and every CI job (including the long board-05 regression) is now passing. Re-verified against `origin/feature/issue-3877` (latest):

**1. Type Check GREEN.** The `[attr-defined]` error at `two_phase.py:233` is gone. The Doctor captured `max_search_iterations = getattr(self.router, "_max_search_iterations", 0)` into a local at line ~229 and used the local in both the guard and the log f-string -- the same mypy-safe pattern as the `core.py` site. `Type Check` job: pass (42s); the `check_mypy_baseline.py` gate is green.

**2. Pure local-variable refactor, no logic change.** The diff replaces a direct attribute read with a local; the original `derive_per_net_cap(...)` wall-clock path is preserved verbatim inside the new `else` branch. The wall-clock-cap bypass still triggers ONLY when `_max_search_iterations` is truthy (i.e. under `--deterministic-budget`); the default routing path is byte-unchanged.

**3. Baseline additions are unrelated drift only.** `.github/mypy-baseline.txt` adds exactly two entries -- `build_native_cmd.py` nanobind (`import-not-found` -> `import-untyped`) and `recovery/strategy.py` bound-method signature rendering -- each with a documenting comment, and the previous signatures are retained. `git diff origin/main` is EMPTY for both `src/kicad_tools/cli/build_native_cmd.py` and `src/kicad_tools/router/recovery/strategy.py` (byte-identical to main; not touched by this PR). No PR-originated error from a changed file is masked.

**4. All in-scope CI green, no routing regression.** Test, Boards 00-07 e2e, Diff-Pair (06) and Match-Group (07) routing regressions, Lint & Format, C++ Build, Routed-PCB DRC, and kicad-cli round-trip smoke all pass. The board re-route jobs did not regress, consistent with the default-routing-unchanged claim. The pre-existing board-05 informational regression (blocking <= 11) also passed (42m). `mergeStateStatus: CLEAN`.

**5. Determinism test + prior approval hold.** `tests/test_route_deterministic_budget_load_independence.py` is genuine and non-tautological: it pins `PYTHONHASHSEED`, asserts non-empty copper (vacuity guard), and requires byte-identical normalized copper across two same-seed routes. The reach-net-neutral design and the retained `--timeout` outer backstop remain intact.

Closes #3877. Ship it.

PR #3885, opened 2026-06-24
fix(schema): preserve multi-unit symbol structure on from-scratch save

### LINKED ISSUE #3874: Multi-unit symbol serialization bug
### Task Type

Bug Fix

### Task Description

Handling of multi-unit symbols is in .kicad_sym files seem to be broken.  The case below isn't really a real use case but isolates the bug.  My real use case is I'm getting tripped up trying to use kicad-tools to move/copy/modify symbols between libraries.

To reproduce:
```
import kicad_tools # 0.14.0
# Expect this to roundtrip
mptestlib = kicad_tools.schema.library.SymbolLibrary.load('multiunit-test.kicad_sym')

# Roundtrip works ok, because hidden ._sexp field gets used
mptestlib.save('multiunit-out.kicad_sym')

# But roundtrip fails if the sexp has to get created from scratch
mptestlib._sexp = None
mptestlib.save('multiunit-out2.kicad_sym')
```

```
$ diff multiunit-test.kicad_sym multiunit-out.kicad_sym
# clean diff

$ diff multiunit-test.kicad_sym multiunit-out2.kicad_sym
# Unexpected problems with output:
# - All pins are now attached to toplevel symbol (were originally attached to unit symbols)
# - Spurious _0_1 unit sub-symbol created from nowhere
# _1_1 thru _3_1 units of original symbol have become empty toplevel symbols in the library
```

[multiunit-out.kicad_sym.txt](https://github.com/user-attachments/files/29262907/multiunit-out.kicad_sym.txt)
[multiunit-out2.kicad_sym.txt](https://github.com/user-attachments/files/29262908/multiunit-out2.kicad_sym.txt)
[multiunit-test.kicad_sym.txt](https://github.com/user-attachments/files/29262906/multiunit-test.kicad_sym.txt)

### Technical Notes

_No response_

### Priority (optional)

Low

---

### Problem Statement

`SymbolLibrary` carries a hidden cached `_sexp` field that is populated on `load()`. `save()` -> `_to_sexp()` returns that cached tree verbatim when it is present, so the documented "clean diff" round-trip never actually exercises the model-to-s-expression serializer. The moment `_sexp` is `None` (a fresh `SymbolLibrary.create()`, a symbol copied/built in memory, or the user manually clearing the cache as in the repro), `save()` falls through to `to_sexp_node()` and the from-scratch path corrupts every multi-unit symbol. This blocks the reporter's real use case: moving/copying/modifying symbols between libraries, where the destination library has no matching cached `_sexp`.

### Root Cause Analysis

The parse path (`LibrarySymbol.from_sexp`, `src/kicad_tools/schema/library.py:567-612`) discards all unit structure, and the regeneration path (`LibrarySymbol.to_sexp_node`, `:850-937`) reconstructs it from fields that were never populated. Concretely:

1. **Unit number is never parsed.** `LibraryPin.from_sexp` (`:362-394`) does not read the enclosing unit sub-symbol's index, and `LibrarySymbol.from_sexp` appends every pin from every `_N_1` sub-symbol into one flat `pins` list without setting `pin.unit`. So every pin keeps the `LibraryPin.unit` default of `1` (`:343`).
2. **`units` count is never set.** `LibrarySymbol.from_sexp` returns without passing `units=`, so `self.units` stays at the dataclass default of `1` (`:461`) regardless of how many `_N_1` sub-symbols existed.
3. **The `_0_1` graphics sub-symbol is not distinguished from `_N_1` pin sub-symbols on parse, and is unconditionally re-emitted.** `from_sexp` iterates `find_all("symbol")` and merges pins+graphics from all of them (including `_0_1`) into the parent. On regeneration, `to_sexp_node` (`:917-923`) emits a `_0_1` sub-symbol whenever `self.graphics` is non-empty -- even for a symbol whose original form had no `_0_1` block -- which is the "spurious _0_1 unit sub-symbol created from nowhere" in the report.

Combined effect on regeneration (`:926-935`): the loop `range(1, self.units + 1)` runs only for `unit_idx == 1`, and the filter `pin.unit == unit_idx` matches all pins (all defaulted to `1`), so every pin collapses into a single `_1_1` unit. The original `_2_1`/`_3_1` units are never reconstructed -- matching the report's "_1_1 thru _3_1 units have become empty toplevel symbols" once the broken tree is reparsed.

Note the report's phrasing "pins reattached to toplevel symbol" describes the *model* state after parse (pins live on the parent `LibrarySymbol`, not per-unit); the on-disk corruption is the collapse described above.

### Implementation Guidance

Fix is confined to the parse/regenerate pair in `src/kicad_tools/schema/library.py`. Suggested approach:

1. **Parse unit index from the sub-symbol name in `LibrarySymbol.from_sexp`.** Unit sub-symbols are named `{short_name}_{unit}_{variant}` (e.g. `MyPart_2_1`). Parse the second-to-last underscore-delimited integer as the unit number. Treat `_0_*` as the graphics/decoration sub-symbol (unit 0), not a pin unit. Assign `pin.unit` from this parsed index for each pin in that sub-symbol.
   - Be robust to symbol names that themselves contain underscores -- split from the right (`rsplit("_", 2)`) rather than the left, and fall back to `unit=1` if the suffix does not match the `_<int>_<int>` shape (top-level pins directly under the parent symbol).
2. **Compute and set `units`** as `max` observed pin-unit index (or count of distinct `_N_1` sub-symbols with N>=1), passed into the returned `LibrarySymbol`. Preserve `units=1` for single-unit symbols.
3. **Only emit `_0_1` when graphics exist AND track whether graphics were originally in a `_0_1` block** so a graphics-free symbol does not gain a spurious empty/▒populated `_0_1`. The current guard `if self.graphics:` (`:918`) is necessary but the upstream parse must not have mis-merged `_0_1` graphics; with (1) correctly classifying `_0_1`, the existing guard becomes correct.
4. Keep the cached-`_sexp` fast path untouched -- it already round-trips. The goal is for the from-scratch path to produce a structurally equivalent tree.

**Related (note, not in scope for AC):** `LibraryPin.from_sexp` (`:365`) parses the pin shape into `_pin_shape` but discards it, so the `shape` field defaults to `line` and non-`line` pins also fail to round-trip from scratch. Worth fixing in the same PR if low-cost, but the acceptance criteria below are scoped to the multi-unit defect.

### Acceptance Criteria

- [ ] After `lib = SymbolLibrary.load(f); lib._sexp = None; lib.save(out)`, the multi-unit symbol in `out` is structurally equivalent to the original: each pin is emitted under its correct `_N_1` unit sub-symbol, no spurious `_0_1` block appears for graphics-free symbols, and no original unit becomes an empty top-level symbol.
- [ ] `LibrarySymbol.from_sexp` populates `LibraryPin.unit` for every pin from the enclosing `_N_1` sub-symbol index, and sets `LibrarySymbol.units` to the correct unit count.
- [ ] Single-unit symbols and graphics-only `_0_1` decoration continue to round-trip unchanged (no regression).
- [ ] A from-scratch save (cleared `_sexp`) of a multi-unit library reparses back to an equivalent model (`pins`-per-unit and `units` identical to the original load).

### Affected Files

- `src/kicad_tools/schema/library.py` - Fix `LibrarySymbol.from_sexp` (parse unit index per sub-symbol, classify `_0_1` graphics, set `units`) and `LibraryPin.from_sexp` (capture `unit`; optionally `shape`). The `to_sexp_node` emitters at `:850-937` should then produce correct output; adjust the `_0_1` guard only if needed.
- `tests/test_schematic_library.py` - Add the regression test (or a new `tests/test_library_multiunit_roundtrip.py`).

### Test Plan

- [ ] Manual verification: Reproduce with the attached `multiunit-test.kicad_sym` fixture: `load` -> `_sexp = None` -> `save`, then diff against the original; confirm a clean (structurally equivalent) diff.
- [ ] Automated tests: Add a regression test that loads a multi-unit symbol fixture, sets `_sexp = None`, saves, reloads the saved file, and asserts (a) the same number of units, (b) each pin maps to its original unit, (c) no spurious `_0_1`/empty top-level symbols. Commit the reporter's `multiunit-test.kicad_sym` as a fixture under `tests/fixtures/`.
- [ ] Edge cases: single-unit symbol (units==1) round-trips unchanged; symbol with `_0_1` graphics decoration round-trips; symbol name containing underscores (`rsplit` correctness); derived (`extends`) symbol still emits no unit sub-symbols.

rjwalters:
## Summary

Fixes the multi-unit `.kicad_sym` serialization bug (#3874). When a `SymbolLibrary`'s cached `._sexp` is present, `.save()` round-trips multi-unit symbols by re-emitting the cached tree verbatim. The bug only manifests on the **from-scratch** path (`_sexp = None`), which the reporter's real use case (moving/copying symbols between libraries) always hits. Previously that path collapsed all pins onto one unit, lost the unit count, and leaked nested `_N_1` sub-symbols back in as spurious empty top-level symbols.

## Changes

All in `src/kicad_tools/schema/library.py`:

- **`SymbolLibrary.load` / `load_from_string`**: use `find_children("symbol")` instead of the recursive `find_all("symbol")`. The recursive variant descended into the `_N_1` unit sub-symbols and registered each as a standalone library entry (the "`_1_1`..`_3_1` became empty top-level symbols" corruption).
- **`LibrarySymbol.from_sexp`**: added `_parse_unit_index()` (rsplit-based so symbol names with underscores parse correctly; `_0_1` treated as graphics/unit 0). Each pin now records its enclosing unit (`pin.unit`) and the symbol records the correct `units` count.
- **`LibraryPin.from_sexp`**: now accepts a `unit` argument and retains the parsed pin `shape` (previously discarded), so non-`line` pins also round-trip from scratch.

With `pin.unit` and `units` populated, the existing `to_sexp_node` emitters reconstruct the correct per-unit `_N_1` layout and no longer emit a spurious `_0_1` block for graphics-free symbols (no emitter change needed).

Tests: new `tests/test_library_multiunit_roundtrip.py` + `tests/fixtures/multiunit_test.kicad_sym`. Updated one assertion in `tests/test_schema.py` that had encoded the buggy sub-symbol-leak behavior (expected 4 symbols where 2 is correct).

| From-scratch save of a multi-unit symbol is structurally equivalent (pins under correct `_N_1`, no spurious `_0_1`, no empty top-level units) | ✅ | `test_from_scratch_save_preserves_units`, `test_from_scratch_no_spurious_0_1_block`, `test_from_scratch_pins_under_correct_units` |
| `from_sexp` populates `LibraryPin.unit` per sub-symbol and sets `LibrarySymbol.units` | ✅ | `test_load_parses_units_and_pin_units` |
| Single-unit symbols and `_0_1` graphics decoration round-trip unchanged | ✅ | `test_single_unit_from_scratch`, `test_graphics_decoration_from_scratch` |
| From-scratch save reparses to an equivalent model | ✅ | reload-and-compare in the multi-unit + underscore-name tests |
| Symbol names containing underscores parse units correctly (rsplit) | ✅ | `test_underscore_name_units` |

## Test Plan

- `uv run pytest tests/test_library_multiunit_roundtrip.py` — 7 passed
- `uv run pytest tests/test_schema.py tests/test_schematic_library.py tests/test_library_extends.py tests/test_operations.py tests/test_schematic_registry.py` — all passed
- Broader library-referencing suites (BOM, netlist, zones, PCB, schematic editing, symbol generator, etc.) — all passed
- `ruff check` / `ruff format --check` clean on changed files; `mypy` introduces no new errors (one pre-existing error at the unrelated `get_string` assignment remains).

Closes #3874

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 3877
-# Branch: feature/issue-3877
+# Issue: 3874
+# Branch: feature/issue-3874
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/src/kicad_tools/schema/library.py
+++ b/src/kicad_tools/schema/library.py
@@ -359,10 +359,16 @@ class LibraryPin:
         return (0, 0)
 
     @classmethod
-    def from_sexp(cls, sexp: SExp) -> LibraryPin:
-        """Parse from S-expression."""
+    def from_sexp(cls, sexp: SExp, unit: int = 1) -> LibraryPin:
+        """Parse from S-expression.
+
+        Args:
+            sexp: The ``(pin ...)`` S-expression node.
+            unit: The unit number of the enclosing ``_N_1`` sub-symbol
+                (1-indexed). Defaults to ``1`` for top-level pins.
+        """
         pin_type = sexp.get_string(0) or "passive"
-        _pin_shape = sexp.get_string(1) or "line"  # noqa: F841 - reserved for rendering
+        pin_shape = sexp.get_string(1) or "line"
 
         pos = (0.0, 0.0)
         rot = 0.0
@@ -391,6 +397,8 @@ class LibraryPin:
             position=pos,
             rotation=rot,
             length=length,
+            unit=unit,
+            shape=pin_shape,
         )
 
     def to_sexp_node(self) -> SExp:
@@ -442,6 +450,39 @@ def _snap_to_kicad_grid(
     if abs(value - nearest) <= tolerance:
         return nearest
     return value
+
+
+def _parse_unit_index(unit_name: str, parent_name: str) -> int | None:
+    """Parse the unit index from a KiCad unit sub-symbol name.
+
+    Unit sub-symbols are named ``{short_name}_{unit}_{variant}`` (e.g.
+    ``MyPart_2_1``). The ``_0_*`` sub-symbol holds graphical decoration and is
+    reported as unit ``0``. ``_N_*`` (N>=1) sub-symbols hold pins for unit N.
+
+    Symbol names may themselves contain underscores, so the trailing
+    ``_<int>_<int>`` suffix is split from the right.
+
+    Args:
+        unit_name: The name of the sub-symbol (e.g. ``MyPart_2_1``).
+        parent_name: The (short) name of the enclosing symbol, used to
+            sanity-check the prefix.
+
+    Returns:
+        The parsed unit index (0 for the graphics sub-symbol, >=1 for pin
+        units), or ``None`` if *unit_name* does not match the expected
+        ``<prefix>_<int>_<int>`` shape.
+    """
+    parts = unit_name.rsplit("_", 2)
+    if len(parts) != 3:
+        return None
+    prefix, unit_str, variant_str = parts
+    if not (unit_str.isdigit() and variant_str.isdigit()):
+        return None
+    # The prefix should match the enclosing symbol's short name. If it does
+    # not, this is not a recognized unit sub-symbol (be conservative).
+    if prefix != parent_name:
+        return None
+    return int(unit_str)
 
 
 @dataclass
@@ -583,15 +624,35 @@ class LibrarySymbol:
             if prop_name:
                 properties[prop_name] = prop_value or ""
 
-        # Parse pins and graphics from unit symbols
+        # Parse pins and graphics from unit sub-symbols.
+        #
+        # Unit sub-symbols are named "{short_name}_{unit}_{variant}" (e.g.
+        # "TPA3116D2_1_1"). The "_0_1" sub-symbol holds graphical decoration;
+        # "_N_1" (N>=1) sub-symbols hold the pins for unit N. We parse the unit
+        # index from each sub-symbol name so that:
+        #   - every pin records its enclosing unit (LibraryPin.unit), and
+        #   - the symbol records how many pin units exist (LibrarySymbol.units).
+        # Sub-symbols whose names do not match the "<prefix>_<int>_<int>" shape
+        # (e.g. pins placed directly under the parent) default to unit 1.
+        short_name = name.split(":", 1)[1] if ":" in name else name
+
         pins: list[LibraryPin] = []
         graphics: list[SymbolPolyline | SymbolCircle | SymbolArc | SymbolRectangle] = []
-        for unit_sym in sexp.find_all("symbol"):
-            _unit_name = unit_sym.get_string(0) or ""  # noqa: F841
-            # Unit symbols have names like "TPA3116D2_1_1"
-            # Format: {name}_{unit}_{variant}
+        max_unit = 1
+        for unit_sym in sexp.find_children("symbol"):
+            unit_name = unit_sym.get_string(0) or ""
+            parsed_unit = _parse_unit_index(unit_name, short_name)
+            # Pins inherit the enclosing sub-symbol's unit index. The "_0_1"
+            # graphics sub-symbol (parsed_unit == 0) has no pins; treat
+            # unparsable names as unit 1.
+            pin_unit = parsed_unit if (parsed_unit and parsed_unit >= 1) else 1
+
             for pin_sexp in unit_sym.find_all("pin"):
-                pins.append(LibraryPin.from_sexp(pin_sexp))
+                pins.append(LibraryPin.from_sexp(pin_sexp, unit=pin_unit))
+                max_unit = max(max_unit, pin_unit)
+
+            if parsed_unit is not None and parsed_unit >= 1:
+                max_unit = max(max_unit, parsed_unit)
 
             # Parse graphical primitives
             for polyline_sexp in unit_sym.find_all("polyline"):
@@ -608,6 +669,7 @@ class LibrarySymbol:
             properties=properties,
             pins=pins,
             graphics=graphics,
+            units=max_unit,
             extends=extends,
         )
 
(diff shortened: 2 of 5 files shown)


rjwalters:
✅ **Approved!**

Thorough review complete. This is a clean, well-targeted fix for the multi-unit symbol from-scratch serialization bug (#3874).

## Correctness

**`find_children` switch is correct and scoped.** Verified `find_all("symbol")` recurses into descendants (via `iter_all`) while `find_children("symbol")` returns only direct children (`src/kicad_tools/sexp/parser.py:196,772`). The two library-entry call sites (`load`/`load_from_string`, lines ~1058/1089) and the per-symbol unit-discovery loop in `LibrarySymbol.from_sexp` (line 642) are exactly the places where the recursive variant was wrong — it re-registered nested `_N_1` sub-symbols as standalone top-level entries. The other ~25 `find_all("symbol")` call sites across `operations/`, `cli/`, `erc/`, and `schema/schematic.py` operate on schematic instance trees or library roots where the recursion is either intended or harmless; none were changed, and the relevant suites pass.

**`_parse_unit_index` edge cases hold up:**
- Underscored names: `rsplit("_", 2)` on `My_Part_X_2_1` yields `["My_Part_X","2","1"]` → unit 2 (covered by `test_underscore_name_units`).
- `_0_1` graphics: parsed as 0, excluded from the pin-unit count, so a graphics-only symbol stays `units=1` and emits `_0_1` only when graphics exist (`test_graphics_decoration_from_scratch`).
- Conservative prefix check: when the sub-symbol prefix doesn't match the parent short name (e.g. the `Device:R_0_1` fixture where prefix is `Device:R` vs short name `R`), it returns `None` and pins fall back to unit 1 — benign and round-trips correctly.

**Round-trip is symmetric.** The `to_sexp_node` emitter (lines ~979-997) groups pins by `pin.unit` into `_N_1` blocks for `range(1, units+1)` and emits `_0_1` only when `self.graphics` is non-empty — the exact inverse of the new parser. Reload-and-compare assertions in the new tests confirm structural equivalence.

## test_schema.py assertion change is a genuine bug-encoding fix

The `minimal_symbol_library` fixture (`tests/conftest.py:666`) defines exactly **two** symbols (`Device:R`, `Device:C`), each wrapping a nested `_0_1` pin sub-symbol. The old assertion (`len == 4`, expecting `Device:R_0_1`/`Device:C_0_1` as standalone entries) encoded precisely the corruption being fixed. The new assertion (2 top-level symbols, sub-symbols excluded, `Device:R.pin_count == 2`) is the correct post-fix behavior — not masking a regression.

## Verification

- `tests/test_library_multiunit_roundtrip.py` (7 new), `test_schema.py`, `test_schematic_library.py`, `test_library_extends.py`, `test_operations.py` — **298 passed** locally.
- `ruff check` / `ruff format --check` clean on changed files.
- PR description uses `Closes #3874`.
- All 16 CI checks pass (including the long-running Board 05 routing regression — an orthogonal job, unrelated to this schema-only change). Merge state: **CLEAN**.

Nice touch adding the `_0_1` graphics-decoration and underscore-name regression cases alongside the core multi-unit fix.

PR #3898, opened 2026-07-01
feat(route): add --order-method to wire RoutingOptimizer into kct route

### LINKED ISSUE #3897: Route-order optimization: locality-preserving seed→permutation encoding + wire optimize_net_order into router/CLI
### Summary

Add a **locality-preserving seed→net-order encoding** and wire **global route-order optimization** into the actual router / CLI, so net ordering can be *optimized* rather than hand-tuned or blindly seed-swept.

### Motivation

Net routing order strongly affects completion on congested boards (early nets claim preferred corridors; late nets route around them). Today, on a real design (softstart controller PCB, rjwalters/softstart#12), we can only influence order two ways, both weak:

1. **Manual priority** (`ROUTE_FIRST_NETS` → net-class priority 1) — hand-guessed, brittle; the recipe's own comments document that front-loading the wrong net (e.g. VGATE) regresses hard.
2. **Blind seed sweep** — `--seed` perturbs the router's internal RNG *chaotically*: neighboring seeds produce unrelated outcomes, so there's no gradient to follow. The recipe pins a hand-found "measured-best" seed 42 and notes you must "re-sweep a handful of seeds if placement changes."

Neither lets us actually *optimize* over the order.

### What exists today (and the gap)

`kicad_tools/optim/routing.py::RoutingOptimizer.optimize_net_order()` already implements greedy / critical_first / congestion / hybrid / **simulated_annealing** ordering. But:

- It is **not imported by any `cli/` command** and **not wired into `kct route`** — the production router path never calls it. It's effectively orphaned (only `PlacementOptimizer` is CLI-exposed).
- The SA variant does **raw random pairwise swaps** and **re-routes the full board every iteration** (default 1000) — infeasible for boards where a single route is minutes.
- There is **no seed→permutation encoding**, let alone a locality-preserving one.

### Requested feature

1. **Locality-preserving seed→order encoding.** A deterministic map `seed (scalar/low-D) → net permutation` such that *nearby seeds produce nearby permutations* (small, bounded number of adjacent transpositions / a Lehmer-code / Gray-code-style scheme). This turns routing order into a smooth optimization surface: local search, CMA-ES, or 1-D line search over the seed becomes meaningful, instead of the current chaotic RNG.
2. **Wire order-optimization into the router / expose via CLI.** e.g. `kct route --optimize-order {greedy|critical_first|hybrid|anneal|seed-search} [--order-seed N] [--order-iterations K] [--order-budget SECONDS]`, actually consumed by the cpp/python routing path (not just the orphaned `RoutingOptimizer`).
3. **Budget-aware search.** Since full re-routes are expensive, support a cheap FOM (global-routing / congestion estimate) for the inner loop and only full-route the top candidates; and/or warm-start from the incumbent order so each step is a small perturbation.
4. **Determinism/reproducibility.** Given the same `--order-seed`, produce the same order (so a found-optimal order can be pinned in a recipe, like the current seed pin but principled).

### Acceptance ideas

- `optimize_net_order` (or a successor) is reachable from `kct route`/`route-auto` and demonstrably changes completion on a congested fixture.
- A documented seed→permutation encoding with a test showing `hamming/transposition-distance(order(s), order(s+1))` is small (locality) while the map still covers the permutation space.
- A worked example that improves a known-congested board's completion vs the greedy baseline within a fixed route budget.

### Context / reference

- Downstream driver: rjwalters/softstart#12 — controller PCB routes 35/37 nets; the two residual nets (`NRST` board-spanning star, `UCC_HO_NEG`) fail post-route clearance validation and are order/congestion-sensitive. We're currently stuck hand-tuning `ROUTE_FIRST_NETS` + seed.
- Idea credit: locality-preserving seed encoding so order optimization is a smooth search rather than a lottery.

---

## Curator Enhancement

*Curated 2026-07-01. Scoped to a single bounded increment per single-increment rule. Larger items captured as explicit follow-up scope below.*

### Codebase Verification

The issue claims are accurate against `origin/main`:

- `src/kicad_tools/optim/routing.py` — `RoutingOptimizer.optimize_net_order()` exists at line 328. It implements `greedy`, `critical_first`, `congestion`, `hybrid`, and `simulated_annealing` methods. **Confirmed orphaned from CLI**: no file under `src/kicad_tools/cli/` imports or calls it.
- `src/kicad_tools/cli/commands/routing.py` (`run_route_command`) — forwards `--seed` to `route_cmd.main()` but has no `--optimize-order`, `--order-seed`, or `--order-method` flags.
- `src/kicad_tools/cli/route_cmd.py` — `--seed` seeds only `random.seed()` at line 7892, affecting internal shuffles. No `net_order` argument is exposed to the caller. The router's `route_all(net_order=...)` parameter (core.py:5113) accepts an explicit list but is never passed from the CLI dispatch.
- `src/kicad_tools/router/core.py` — `_get_net_priority()` at line 4309 builds a 6-tuple sort key (net class, complexity tier, constraint score, pad count, bbox diagonal, RUDY congestion). This is the existing heuristic ordering the CLI uses; `optimize_net_order` in `optim/routing.py` duplicates a subset of this logic in isolation.
- There is **no seed→permutation encoding** anywhere in the codebase (no `factoradic`, `Lehmer`, `Gray.*code`, or `seed.*permut` anywhere in `src/`).
- `ROUTE_FIRST_NETS` exists only in `.loom/worktrees/` stale worktrees, not in `src/` or `boards/` on `main`. The softstart recipe lives in `rjwalters/softstart`, not this repo.
- `tests/test_optim.py` has `TestRoutingOptimizer` at line 744 with tests for greedy, critical_first, empty_nets, invalid_method, and congestion_requires_map — no locality or seed tests exist.

### Single-Increment Scope (this issue)

The original issue asks for four things. Applying the single-increment rule, this issue is scoped to **one concrete, bounded increment**:

**Wire `RoutingOptimizer.optimize_net_order` (existing, tested heuristic methods only) into `kct route` via a new `--order-method` flag.**

This is independently valuable (any heuristic ordering is better than the current opaque priority sort for users who want to experiment), can be validated without the locality encoding, and is a natural prerequisite for the seed-encoding follow-up.

**What is in scope:**
- Add `--order-method {greedy,critical_first,congestion,hybrid}` to `kct route`.
- When `--order-method` is passed, call `RoutingOptimizer.optimize_net_order(router_factory, method=...)` to compute the order, then pass the resulting `list[int]` to `route_all(net_order=...)`.
- The `congestion` and `hybrid` methods require a `CongestionMap`; wire up `router.get_congestion_map()` before calling those methods.
- When `--order-method` is not passed, behavior is byte-identical to today (no regression).

**What is explicitly NOT in scope (this increment):**
- The `simulated_annealing` method is excluded — it re-routes the full board 1000x and is infeasible without a budget guard.
- No `--order-seed` or `--order-budget` flags.
- No locality-preserving encoding (follow-up issue).

### Follow-up Scope (explicit, NOT part of this issue)

1. **Locality-preserving seed→permutation encoding.** Implement a deterministic `seed → net permutation` map using factoradic / Lehmer-code encoding such that adjacent seeds differ by O(1) adjacent transpositions. Expose as `--order-seed N`.
2. **Budget-aware simulated annealing.** Wire the `simulated_annealing` method with `--order-iterations K` and `--order-budget SECONDS` guards. Use a cheap FOM (congestion estimate, not full route) for the inner loop.
3. **1-D line search over seed space.** Once the locality-preserving encoding exists, implement a sweep over neighboring seeds to pick the best — a tractable alternative to a blind 1000-seed sweep.
4. **Recipe pinning documentation.** Document how to pin a found-optimal `--order-method` and future `--order-seed` in a `.kct.yaml` recipe.

### Acceptance Criteria

- [ ] `kct route <pcb> --order-method greedy` completes without error. When `--verbose`, prints which order method was used.
- [ ] `kct route <pcb> --order-method critical_first` passes a `net_order` list to `route_all()` where nets matching `vcc/vdd/gnd` patterns appear before signal nets.
- [ ] `kct route <pcb> --order-method congestion` and `--order-method hybrid` succeed when a `CongestionMap` can be obtained. On failure to obtain a congestion map, a clear warning is emitted and the command falls back to `greedy` (rather than crashing).
- [ ] `--order-method` is forwarded correctly through the two-level CLI dispatch: `cli/commands/routing.py::run_route_command` → `cli/route_cmd.py::main()` → the routing loop, following the same pattern used by `--seed`, `--targeted-ripup`, etc.
- [ ] `kct route <pcb>` without `--order-method` produces byte-identical output to the current behavior (verified with a deterministic `--seed` run and a checksum comparison).
- [ ] Existing tests in `tests/test_optim.py::TestRoutingOptimizer` pass without modification.
- [ ] A new test verifies that when `--order-method greedy` is passed at the CLI level, `route_all` is called with a non-None `net_order` kwarg (can use a mock/spy).
- [ ] `uv run ruff check src/` and `uv run mypy src/` pass with no new errors.

### Implementation Guidance

**Files to touch:**

- `src/kicad_tools/cli/commands/routing.py` — Add `--order-method` forwarding in `run_route_command()`. Follow the `--targeted-ripup` / `--seed` pattern (lines 203–208, 356–359): only forward when the value is not None.
- `src/kicad_tools/cli/route_cmd.py` — Two changes: (1) Add `--order-method` argument to the inner parser near the `--seed` argument declaration (~line 6969). (2) In the routing dispatch where `route_all()` is called (lines ~3297, ~4173, ~6247, ~8940–9135), if `args.order_method` is set, instantiate `RoutingOptimizer()` and call `optimize_net_order(router_factory, method=args.order_method)` to get `net_order`, then pass it to `route_all(net_order=net_order)`.
- `src/kicad_tools/optim/routing.py` — Read-only for this increment; the existing `optimize_net_order` API is sufficient.
- `tests/test_optim.py` — Existing tests must remain passing; no changes expected.
- `tests/` — Add a new CLI-level test for `--order-method` forwarding.

**Key existing patterns to follow:**

The `--seed` forwarding is the closest analog: outer parser declares it, `run_route_command` forwards it via `getattr(args, "seed", None)`, and `route_cmd.py` consumes it. `--order-method` should use exactly this pattern.

**Router factory construction:**

`optimize_net_order` requires a `router_factory: Callable[[], Autorouter]`. Wrapping the router construction in a closure is acceptable. The throw-away router construction to compute the order (before the real route) is acceptable for this increment.

**Congestion map availability:**

Call `router.get_congestion_map()` on the throw-away router after all components are added but before routing, when `method in ("congestion", "hybrid")`. Wrap in try/except; on failure, warn and fall back to `greedy`.

**Mutual exclusivity with `_get_net_priority`:**

When `--order-method` is provided, pass the resulting list directly as `net_order` to `route_all()`, which overrides the internal `_get_net_priority` sort. Document this in the `--order-method` help string: "Overrides the default priority-based net ordering."

### Affected Files

- `src/kicad_tools/cli/commands/routing.py` — Add `--order-method` forwarding in `run_route_command()`
- `src/kicad_tools/cli/route_cmd.py` — Add `--order-method` argument; wire `RoutingOptimizer.optimize_net_order` call before `route_all()`
- `src/kicad_tools/optim/routing.py` — Read-only (existing API is sufficient)
- `tests/test_optim.py` — Read-only (existing tests must keep passing)
- `tests/` — New test file or extension to verify `--order-method` CLI forwarding

### Test Plan

- [ ] Manual verification: `uv run kct route <board.kicad_pcb> --order-method greedy --verbose` completes without error and prints order method selection in verbose output.
- [ ] Manual verification: `uv run kct route <board.kicad_pcb> --order-method critical_first --verbose` completes and the verbose log shows priority nets routed first.
- [ ] Automated unit test: Mock `route_all` on an `Autorouter` instance; assert that `--order-method greedy` via CLI results in `route_all` receiving a non-None `net_order` kwarg containing a list of integers.
- [ ] Automated regression: Run `uv run pytest tests/test_optim.py::TestRoutingOptimizer` — all existing tests must pass.
- [ ] Ruff + mypy: `uv run ruff check src/` and `uv run mypy src/` produce no new errors.
- [ ] Byte-identity regression: `kct route <pcb> --seed 42` without `--order-method` produces the same output SHA as before this change.
- [ ] Edge case: `--order-method congestion` on a board where `get_congestion_map()` raises emits a warning and falls back to greedy rather than crashing.

rjwalters:
## Summary

Wires the previously-orphaned `RoutingOptimizer.optimize_net_order` (in `src/kicad_tools/optim/routing.py`) into the production router via a new `--order-method {greedy,critical_first,congestion,hybrid}` flag on `kct route`. Before this change, `optimize_net_order` was never reachable from any CLI command — users could only influence net ordering via the opaque internal `_get_net_priority` sort.

This is the single-increment scope defined by the curator. The `simulated_annealing` method, `--order-seed`, and the locality-preserving seed→permutation encoding are explicitly out of scope (deferred follow-ups).

## Changes

- `src/kicad_tools/cli/parser.py` — declare `--order-method` on the outer `kct route` parser (default `None`, choices greedy/critical_first/congestion/hybrid).
- `src/kicad_tools/cli/route_cmd.py` — declare `--order-method` on the inner parser; add `_apply_order_method()` helper that calls `RoutingOptimizer.optimize_net_order()` and stashes the result on `router._forced_net_order`. Called at the main load site with a fresh-router factory (so the optimizer's evaluation route does not perturb the real router). congestion/hybrid obtain a `CongestionMap` via `router.get_congestion_map()`; on failure they warn and fall back to greedy.
- `src/kicad_tools/cli/commands/routing.py` — forward `--order-method` through the two-level dispatch, only when set (matches the `--seed` pattern).
- `src/kicad_tools/router/core.py` — add `_forced_net_order` attribute; `route_all` and `route_all_negotiated` seed their base ordering from it when set, and skip the post-matrix priority re-sort so the explicit order survives. `None` (default) preserves the legacy priority sort byte-for-byte.
- `tests/test_cli_order_method.py` — new test file (20 tests).

| `--order-method greedy` completes; prints method when verbose | ✅ | `_apply_order_method` sets order + prints "Net order: --order-method greedy ...". Verified via `kct route --help` and unit tests. |
| `critical_first` puts vcc/vdd/gnd nets before signal nets | ✅ | `test_apply_order_method_critical_first_puts_power_nets_first` |
| congestion/hybrid succeed with a map; warn + fall back to greedy on failure | ✅ | `test_apply_order_method_congestion_falls_back_to_greedy` |
| Forwarded through two-level CLI dispatch like `--seed` | ✅ | `test_shim_forwards_order_method_when_set` + inner/outer parser declaration tests |
| No `--order-method` → byte-identical base ordering | ✅ | `test_route_all_default_order_unchanged_without_forced_order`; parser default None; shim omits flag by default |
| Existing `TestRoutingOptimizer` tests pass unmodified | ✅ | `optim/routing.py` untouched; 10/10 pass |
| New test: `route_all` receives non-None net_order under greedy | ✅ | `test_route_all_seeds_base_order_from_forced_order` (spy on `_filter_pour_nets`) |
| `ruff check src/` and `mypy src/` no new errors | ✅ | ruff clean on all touched files; mypy reports only pre-existing errors (none in added code) |

## Test Plan

- `uv run pytest tests/test_cli_order_method.py` — 20 passed.
- `uv run pytest tests/test_optim.py::TestRoutingOptimizer` — 10 passed (unmodified).
- Regression: `tests/test_router_core.py tests/test_cli_region_parallel.py tests/test_cli_parser_drift.py tests/test_route_cmd_params.py` — 257 passed.
- `uv run ruff check` + `ruff format --check` — clean on all touched files.
- `uv run mypy src/...` — no new errors (pre-existing `_interrupt_state` typing errors only).

Closes #3897

### DIFF
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -357,6 +357,11 @@ def run_route_command(args) -> int:
     # (router uses os.urandom-derived state, existing behaviour).
     if getattr(args, "seed", None) is not None:
         sub_argv.extend(["--seed", str(args.seed)])
+    # Issue #3897: forward --order-method to the inner parser.  Default is
+    # None (use the internal priority-based net ordering, existing behaviour),
+    # so only forward an explicitly-provided value (matches the --seed pattern).
+    if getattr(args, "order_method", None) is not None:
+        sub_argv.extend(["--order-method", str(args.order_method)])
     # Issue #3054 (Phase 2 of #3045): forward --region-parallel and partition
     # tuning flags to the inner parser.  All four flags are opt-in and only
     # forwarded when set to non-default values, so existing scripts using
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -3120,6 +3120,21 @@ def _add_route_parser(subparsers) -> None:
         ),
     )
     route_parser.add_argument(
+        "--order-method",
+        choices=["greedy", "critical_first", "congestion", "hybrid"],
+        default=None,
+        help=(
+            "Compute the net routing order with a named heuristic (Issue #3897) "
+            "instead of the default priority-based sort. Overrides the internal "
+            "_get_net_priority ordering. Choices: 'greedy' (fewest pads first), "
+            "'critical_first' (power/clock nets first), 'congestion' (most "
+            "congested nets first), 'hybrid' (critical_first + congestion). "
+            "'congestion' and 'hybrid' require a congestion map; if one cannot "
+            "be obtained the command warns and falls back to 'greedy'. When "
+            "omitted, ordering is byte-identical to the default behaviour."
+        ),
+    )
+    route_parser.add_argument(
         "--no-auto-build-native",
         action="store_true",
         help=(
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -69,6 +69,8 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
+
     from kicad_tools.router import Autorouter, LayerStack
     from kicad_tools.router.primitives import Route
 
@@ -2754,6 +2756,84 @@ def _apply_analog_net_class(router: "Autorouter", args, quiet: bool = False) -> 
                 f"  Analog routing: left {len(skipped_pour)} pour/ground net(s) "
                 f"as-is (not forced into pathfinder): {', '.join(skipped_pour)}"
             )
+
+
+def _apply_order_method(
+    router: "Autorouter",
+    args,
+    router_factory: "Callable[[], Autorouter] | None" = None,
+    quiet: bool = False,
+) -> None:
+    """Compute an explicit net order via ``--order-method`` (Issue #3897).
+
+    Wires the previously-orphaned
+    :meth:`kicad_tools.optim.routing.RoutingOptimizer.optimize_net_order`
+    into ``kct route``.  When ``args.order_method`` is one of ``greedy``,
+    ``critical_first``, ``congestion`` or ``hybrid``, this computes the net
+    routing order with that heuristic and stashes it on
+    ``router._forced_net_order``.  Both :meth:`Autorouter.route_all` and
+    :meth:`Autorouter.route_all_negotiated` consult that attribute to seed
+    their base ordering, overriding the internal ``_get_net_priority`` sort.
+
+    Strict no-op when ``--order-method`` is not supplied, so routing output is
+    byte-identical to the historical behaviour (the ``_forced_net_order``
+    attribute stays ``None``).
+
+    The ``congestion`` and ``hybrid`` methods require a congestion map.  We
+    obtain one from the already-loaded ``router`` via
+    :meth:`Autorouter.get_congestion_map`; on failure we emit a warning and
+    fall back to ``greedy`` rather than crashing.
+
+    Args:
+        router: The loaded router whose ``_forced_net_order`` will be set.
+        args: Parsed CLI namespace (reads ``args.order_method``).
+        router_factory: Callable returning a fresh, fully-loaded router used by
+            the optimizer to evaluate the candidate order.  When ``None``, a
+            factory returning ``router`` itself is used (the optimizer's final
+            evaluation route then runs on ``router``; callers that must keep
+            ``router`` pristine should pass a fresh-build factory).
+        quiet: Suppress the informational log line when True.
+    """
+    method = getattr(args, "order_method", None)
+    if method is None:
+        return
+
+    from kicad_tools.cli.progress import flush_print
+    from kicad_tools.optim.routing import RoutingOptimizer
+
+    congestion_map = None
+    effective_method = method
+    if method in ("congestion", "hybrid"):
+        try:
+            congestion_map = router.get_congestion_map()
+        except Exception as exc:  # noqa: BLE001 - fall back rather than crash
+            if not quiet:
+                flush_print(
+                    f"  Warning: --order-method {method} could not obtain a "
+                    f"congestion map ({exc}); falling back to greedy ordering."
+                )
+            effective_method = "greedy"
+            congestion_map = None
+
+    if router_factory is None:
+
+        def router_factory() -> "Autorouter":
+            return router
+
+    optimizer = RoutingOptimizer()
+    order, _fom = optimizer.optimize_net_order(
+        router_factory,
+        method=effective_method,
+        congestion_map=congestion_map,
+    )
+    router._forced_net_order = order
+
+    if not quiet:
+        flush_print(
+            f"  Net order: --order-method {method} "
+            f"(overrides default priority sort; {len(order)} nets)"
+            + (f" [fell back to {effective_method}]" if effective_method != method else "")
+        )
 
 
 def _log_fine_pitch_escape_regions(
@@ -6988,6 +7068,21 @@ def main(argv: list[str] | None = None) -> int:
         ),
     )
     parser.add_argument(
+        "--order-method",
+        choices=["greedy", "critical_first", "congestion", "hybrid"],
+        default=None,
+        help=(
+            "Compute the net routing order with a named heuristic (Issue #3897) "
+            "instead of the default priority-based sort. Overrides the internal "
+            "_get_net_priority ordering. Choices: 'greedy' (fewest pads first), "
+            "'critical_first' (power/clock nets first), 'congestion' (most "
+            "congested nets first), 'hybrid' (critical_first + congestion). "
+            "'congestion' and 'hybrid' require a congestion map; if one cannot "
+            "be obtained the command warns and falls back to 'greedy'. When "
+            "omitted, ordering is byte-identical to the default behaviour."
+        ),
+    )
+    parser.add_argument(
         "-v",
         "--verbose",
         action="store_true",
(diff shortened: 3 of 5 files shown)


rjwalters:
✅ **Approved!**

Clean, well-scoped increment that wires the previously-orphaned `RoutingOptimizer.optimize_net_order` into `kct route` via `--order-method`. Verified against all acceptance criteria.

## Verification

**Correctness**
- **Byte-identity (default path):** With `_forced_net_order` unset, `route_all` and `route_all_negotiated` take the exact legacy branch (`sorted(...key=_get_net_priority)`) AND still run the post-matrix re-sort (`if not _order_forced:` / `if self._forced_net_order is None:`). No behavior change on the default path. Guarded by `test_route_all_default_order_unchanged_without_forced_order`.
- **Attribute init:** `_forced_net_order` initialized in `Autorouter.__init__` (core.py:911), so no AttributeError for instances that never set it.
- **Forced-order path:** When set, base order is seeded from the forced list and the priority re-sort is skipped so the CLI order survives. Confirmed via `test_route_all_seeds_base_order_from_forced_order` (spy on `_filter_pour_nets`).
- **Two-level dispatch:** Outer parser (parser.py), inner parser (route_cmd.py), and forwarding shim (commands/routing.py, only-when-set, matches `--seed`) all pinned by drift-style tests.
- **Congestion/hybrid fallback:** `get_congestion_map()` failure warns and falls back to greedy rather than crashing (`test_apply_order_method_congestion_falls_back_to_greedy`).
- **Pristine real router:** The optimizer's evaluation route runs on a fresh router built by `_order_router_factory` with identical load params, so the real `router` is untouched.

**`_forced_net_order` interaction (checked per review request)**
- In `route_all_negotiated`, two reorderings after the (now-skipped) matrix re-sort still adjust the forced order: the Issue #2482 sibling boost (only when a diff-pair pre-pass produced `prerouted_nets` + connector siblings) and `_interleave_match_groups` fairness. This means the forced order is a *base* ordering that downstream fairness/sibling logic can still perturb. This is documented in the code comment ("sibling boosts still apply") and matches the issue's explicit guidance that downstream filters/re-sorts still apply. Acceptable for this increment; a fuller "hard pin" is reasonable follow-up scope.

**Tests / checks (run locally in worktree issue-3897)**
- `tests/test_cli_order_method.py` — 20 passed
- `tests/test_optim.py::TestRoutingOptimizer` — 10 passed (unmodified)
- `tests/test_route_cmd_params.py tests/test_cli_parser_drift.py` — passed
- `ruff check` + `ruff format --check` — clean on all touched files
- **CI: 16/16 checks pass, merge state CLEAN** (waited for all board E2E regressions to finish, incl. Board 05 routing gate)

## Non-blocking note (follow-up candidate)
For greedy/critical_first, `optimize_net_order` still runs a full throw-away `route_all` on a fresh router purely to compute a `FigureOfMerit` that the CLI discards (route_cmd only consumes `order`). On a congested board this is an extra full route just to order nets. The issue scoped `optim/routing.py` as read-only, so reusing the existing API is correct here — but a future increment could add an "order-only" path that skips the evaluation route. Not worth blocking on.

Solid work — approving.

PR #3903, opened 2026-07-05
fix(pcb): emit absolute pad angles when placing rotated footprints

### LINKED ISSUE #3902: PCB writer emits pad (at ... angle) as LOCAL rotation; KiCad expects ABSOLUTE (fp_rot + local) — phantom pad geometry on all rotated footprints
## Summary

KiCad file convention: inside a footprint, `pad (at x y ANGLE)` carries the **absolute** angle (footprint rotation + pad-local rotation). kicad-tools' PCB writer emits only the **local** angle. Any footprint placed at ±90° (or any non-zero rotation) therefore renders in KiCad (10.0.4 verified) with pad *shapes* unrotated — e.g. a TSSOP's 1.475mm-long pads lie ACROSS the pin row instead of along it, overlapping their neighbors.

## Observed impact (chorus-test-revA v22 board, 2026-07-04)

- 52 rotated footprints (U1, U2, U3, U7, U8, J4, J5, J6 + 44 passives) had phantom pad geometry.
- kicad-cli DRC reported **45 phantom `shorting_items` on a footprints-only board** (34 from U8 alone) plus ~97 phantom `solder_mask_bridge` violations. After rewriting pad angles to absolute, the footprints-only board is CLEAN.
- Routing/DRC debugging cost: days of confusion — copper appears to short pads that don't actually overlap electrically (netlist is correct; the *shape* is wrong).
- **Manufacturing impact: gerber exports of any rotated footprint are geometrically wrong** (mask/paste/copper apertures at the wrong angle). Boards produced from affected exports would have defective pads.

## Repro

1. Create/emit a board via kicad-tools' PCB writer with a TSSOP (or any elongated-pad footprint) at rotation -90.
2. Open in KiCad 10 or run `kicad-cli pcb drc` — observe pad shapes unrotated relative to the footprint, and phantom shorting/mask violations along the pin row.

## Fix sketch

In the pad serializer: `emit_angle = (footprint_rotation + pad_local_rotation) % 360`. Audit the reader for the symmetric assumption. Also audit rotation *sign* handling in any coordinate transforms (file frame is y-down: `gx = fx + px·cos(θ) + py·sin(θ)`, `gy = fy − px·sin(θ) + py·cos(θ)`).

Session workaround script rewrote 52 footprints' pad angles in-place; happy to share.

rjwalters:
## Summary

KiCad stores each pad's `(at x y ANGLE)` in the **ABSOLUTE** board frame — the angle already includes the parent footprint's rotation. The PCB writer (`add_footprint_from_file`) emitted only the **LOCAL** angle from the library `.kicad_mod`, so any footprint placed at a non-zero rotation rendered its pads unrotated in KiCad. Elongated pads (e.g. a TSSOP's 1.475 mm pads) lay across the pin row, overlapping neighbours and producing phantom `shorting_items` / `solder_mask_bridge` DRC violations plus geometrically wrong gerber apertures (issue #3902).

This PR fixes the writer and unifies the rotation convention so reader / writer / all consumers agree with KiCad.

## Convention settled: pad angle is ABSOLUTE (board frame)

- **Stored pad angle** = absolute orientation (already includes `footprint.rotation`).
- **Pad position** `(x, y)` remains footprint-**local** and is still rotated by `footprint.rotation` to reach board coordinates — this asymmetry (local position, absolute angle) is exactly KiCad's file convention and is now documented on `Pad.rotation`.
- Consumers must use `pad.rotation` **directly** for pad-shape orientation and must NOT add `footprint.rotation` on top of it.

## Changes

- **Writer** (`schema/pcb.py`, `add_footprint_from_file`): after placing the footprint, rewrite each pad's angle to `(local + rotation) % 360`.
- **Docstring** (`schema/pcb.py`, `Pad.rotation`): corrected from "relative to footprint" to ABSOLUTE, with the position/angle asymmetry spelled out.
- **Consumer audit** (removed `fp_rot + pad_rot` double-count — see table below).
- New fixture `tests/fixtures/Test_Library.pretty/RotatedPad_Test.kicad_mod` (elongated pads, one with an explicit per-pad angle of 30° to exercise the folding path).

## Consumer audit (risk center)

Every site that read a pad angle and combined it with the footprint rotation:

| File / site | Before | Treatment |
|-------------|--------|-----------|
| `validate/rules/clearance.py:_pad_polygon` (~240) | `footprint.rotation + pad.rotation` | Use `pad.rotation` directly (absolute) |
| `validate/rules/clearance.py:_transform_pad_dimensions` (~181) | `footprint.rotation` only | Use `pad.rotation` directly; kept in sync with `_pad_polygon` |
| `zones/fill_clearance.py` (~185) | `(fp_rot + pad_rot) % 360` | `pad_rot % 360` |
| `router/io.py:load_pcb_for_routing` (~3431) | `(fp_rot + pad_rot) % 360` | `pad_rot % 360` |
| `router/io.py:load_pads_for_analysis` (~1922) | `(fp_rot + pad_rot) % 360` | `pad_rot % 360` |
| `validate/rules/netlist.py`, `via_in_pad.py`, `connectivity.py`, `edge.py`, `silkscreen.py`, `zones/generator.py` | use `footprint.rotation` for **position** only (no pad angle) | Unchanged — not affected |

No fixture currently on `main` has a rotated footprint with per-pad angles, so nothing regressed silently; the new fixture + DRC gate close that coverage gap.

| Writer emits absolute angle for ±90 placed footprint | ✅ | `test_positive_90_folds_into_pad_angles` (0→90, 30→120), `test_negative_90_folds_and_wraps` (0→270, 30→300) |
| Zero-rotation unchanged | ✅ | `test_zero_rotation_leaves_pad_angles_unchanged` |
| Round-trip (save → reload reports absolute) | ✅ | `test_round_trip_reader_reports_absolute_angle`, `test_emitted_text_contains_absolute_pad_angle` |
| kicad-cli DRC footprints-only gate, skip if absent | ✅ | `TestRotatedFootprintPadGeometry` under module-level `skipif(find_kicad_cli() is None)`; asserts zero phantom `shorting_items`/`solder_mask_bridge` — passes against real kicad-cli 10 locally |
| Docstring asymmetry corrected + convention consistent | ✅ | `Pad.rotation` docstring + all consumers audited above |
| `tests/test_footprint_rotation.py` stays green | ✅ | Updated the cases that enshrined the buggy LOCAL convention (documented below); all pass |

### Tests that enshrined the buggy convention (updated, per issue guidance)

- `test_footprint_rotation.py::TestPadDimensionRotation`: previously asserted `total = fp_rotation + pad_rotation` (e.g. `_make_pcb_text(45, pad_rotation=45)` → swap). Rewrote so `pad_rotation` is the ABSOLUTE stored angle; added `test_absent_pad_angle_under_rotated_footprint_no_swap` proving an absent angle reads as absolute 0 (unrotated) even under a 90° footprint — the exact bug guard.
- `test_validate.py::TestRotatedPadDimensions`: added `rotation=<abs>` to the constructed `Pad`s so they model what a real placed footprint stores (a 90°-placed footprint writes 90° into every pad).

## Test Plan

- `pytest tests/test_pcb.py tests/test_pcb_create.py tests/test_footprint_rotation.py tests/test_validate.py tests/test_kicad_cli_roundtrip.py` → 338 passed, 5 skipped.
- Consumer-path suites (`test_zones_fill_clearance.py`, `test_validate_clearance_rect_pad.py`, `test_component_clearance.py`, `test_validate_clearance_net0_bridge.py`, `router/test_fine_pitch_row_per_pad_width.py`) → 85 passed.
- DRC gate ran against real `kicad-cli` (Homebrew 10) → clean, zero phantom overlaps.
- `ruff check` + `ruff format` clean on all touched files. mypy: no new errors introduced at touched lines (pre-existing repo-wide mypy errors unchanged).

Closes #3902


rjwalters:
**Builder update**: a broad verification sweep (`-k "clearance or rotation or pad"`, 2576 tests selected) after PR creation surfaced 4 failures in two test files missed by the initial consumer audit. Both enshrined the old LOCAL pad-angle convention; fixed in 013cdd82:

| Test | Problem | Treatment |
|------|---------|-----------|
| `test_pad_polygon_geometry.py::test_footprint_rotation_90_swaps_extents` | pad `rotation=0` + fp 90 expected swap (LOCAL convention) | Pad now stores absolute 90 (as a real placed footprint would); renamed `test_absolute_pad_angle_90_swaps_extents` |
| `test_pad_polygon_geometry.py::test_per_pad_rotation_adds_to_footprint_rotation` | asserted `fp 45 + pad 45 = 90` addition | Replaced with `test_footprint_rotation_not_double_counted` (pad absolute 90, fp 45 must NOT be added) plus a new unrotated-pad-under-rotated-footprint guard |
| `test_clearance_net_names.py::test_from_pad_carries_net_name` / `test_from_pad_net0_gives_empty_string` | MagicMock pads never set `.rotation`; `_transform_pad_dimensions` now reads it and the cardinal comparison hit a MagicMock (TypeError) | Set `pad.rotation = 0.0` explicitly on the mocks |

Sweep result after fix: previously failing tests pass; the remaining 2572 tests in the selection were already green (12 skipped; 18120 deselected). The audit table in the PR description remains accurate for src/ consumers — these were test-side convention captures only.

rjwalters:
❌ **Changes Requested — CI `Test` job fails (regressions introduced by this PR)**

The core fix is **correct and well-tested** — I verified it thoroughly (see the "What's right" section below). However, the full CI `Test` job fails with **6 failures in two files**, both directly caused by this PR. These didn't show up in the PR's listed test suites (they weren't in the run set), so please add them to your local check.

## Failing tests (reproduced locally under `uv run pytest`)

### 1. `tests/test_footprint_library.py` — 4 failures (`assert 4 == 3`)
This PR adds `RotatedPad_Test.kicad_mod` into `tests/fixtures/Test_Library.pretty/`, bumping that library from 3 → 4 footprints. Four tests hard-count that directory:
- `TestFootprintLibraryLoad::test_load_pretty_directory` (`test_footprint_library.py:140`)
- `TestFootprintLibraryLoad::test_footprints_property` (`:160`)
- `TestFootprintLibraryLoad::test_iterate_footprints` (`:191`)
- `TestFootprintLibraryCompatibility::test_loaded_library_has_empty_builtin` (`:269`)

**Fix options:** either bump those four assertions to `4`, **or** (preferable, to keep the shared library stable for existing tests) move the new fixture into its own `.pretty` dir — e.g. `tests/fixtures/RotatedPad_Test.pretty/` — and update the two new test classes' `TEST_PRETTY_DIR` accordingly. The dedicated-dir approach avoids coupling the #3902 fixture to unrelated library-count tests.

### 2. `tests/test_manufacturer_propagation_lint.py` — 2 failures
The edits to `src/kicad_tools/router/io.py` inserted lines and shifted the `DesignRules()` call sites, desyncing the line-number-pinned `_ALLOWLIST`:
- `test_manufacturer_propagation_in_router_design_rules`: `router/io.py:2871` now flags a `DesignRules(...)` call with no allowlist entry / manufacturer arg.
- `test_allowlist_entries_resolve_to_actual_call_sites`: allowlist entries `2868` and `3469` are now stale — the calls moved to **2871** and **3474**.

**Fix:** update `_ALLOWLIST` in `tests/test_manufacturer_propagation_lint.py:77-79` from `2868 → 2871` and `3469 → 3474` (verified: `grep -n "DesignRules(" src/kicad_tools/router/io.py` shows the calls now at 2871 and 3474). No behavioral change — the fallback `DesignRules()` sites are unchanged; only their line numbers moved.

## What's right (so re-review is fast once CI is green)

- **Consumer audit is complete and accurate.** I independently grepped `pad.rotation` / `pad_rot` / `total_rot` / `fp_rot` across `src/`. Every site that combined `footprint.rotation` with a pad angle is now fixed: `clearance.py` (`_pad_polygon`, `_transform_pad_dimensions`), `zones/fill_clearance.py`, and both `router/io.py` loaders. All remaining `footprint.rotation` uses are for **pad position** (`rotate_pad_offset`), which correctly stays footprint-local — matches your table exactly.
- **escape.py / fine_pitch*.py** don't read pad angles at all; they consume `width`/`height` from the corrected loader `Pad` objects. No gap.
- **DSN export (`export/dsn.py`) is correctly untouched.** `_parse_pad` ignores the pad angle entirely and emits an axis-aligned rect padstack — a genuine pre-existing gap the curator flagged, but this PR did not half-touch it. Recommend a **separate follow-up issue** for DSN pad-angle handling (not a blocker here).
- **Back-compat verified.** I scanned all 27 committed `*.kicad_pcb` fixtures: **zero** footprints placed at nonzero rotation, and the only nonzero pad angle (`multilayer_zones.kicad_pcb`, a QFP pad at 90°) sits under an unrotated footprint (absolute == local). Your claim "no fixture on main has a rotated footprint with per-pad angles" is correct — nothing regresses silently.
- **The manufacturing gate is real:** `TestRotatedFootprintPadGeometry` runs against Homebrew kicad-cli 10 locally and passes (not skipped) — zero phantom `shorting_items` / `solder_mask_bridge`.
- The PR's listed suites pass locally: `test_pcb/test_pcb_create/test_footprint_rotation/test_validate/test_kicad_cli_roundtrip` → 338 passed, 5 skipped; consumer-path suites → 85 passed.

Once the 6 failures above are fixed and CI's `Test` job is green, this is an approve.

rjwalters:
Doctor: addressed both Judge-flagged CI regressions.

1. **test_footprint_library.py (4 failures):** moved the new `RotatedPad_Test.kicad_mod` fixture out of the shared `tests/fixtures/Test_Library.pretty/` into its own `tests/fixtures/RotatedPad_Test.pretty/` dir, restoring that library to 3 footprints so the hard-coded count assertions pass untouched. Retargeted `TEST_PRETTY_DIR` in `TestAddFootprintPadAngleAbsolute` (test_pcb.py) and the rotated-pad roundtrip class (test_kicad_cli_roundtrip.py) — both classes use only this fixture.
2. **test_manufacturer_propagation_lint.py (2 failures):** resynced the `_ALLOWLIST` line numbers for the two `DesignRules()` fallback call sites in `router/io.py` (verified in-worktree: 2868 -> 2871 route_pcb fallback, 3469 -> 3474 load_pcb_for_routing inner fallback). No behavioral change.

Verification: `tests/test_footprint_library.py test_manufacturer_propagation_lint.py test_pcb.py test_kicad_cli_roundtrip.py test_footprint_rotation.py` -> 273 passed, 0 failed. Ruff check + format clean on all touched files. Pushed to feature/issue-3902; relabeled review-requested.

rjwalters:
## Judge verdict: APPROVED (completing interrupted re-review)

The prior review pass verified the core convention change in depth (independent consumer grep corroborated the PR's audit table; back-compat fixture scan found zero rotated footprints with pad angles in-repo; kicad-cli DRC gate passes against real KiCad 10). Changes were requested solely for 6 mechanical CI failures.

The Doctor's fixes match the request exactly — RotatedPad_Test fixture moved to its own `.pretty` dir (preferred option), the two new test classes retargeted, `_ALLOWLIST` line numbers corrected to the verified sites (2871, 3474) — and nothing else changed.

**CI: all 8 checks pass** including the previously-failing Test job (15m47s). Merge state CLEAN.

Approving and promoting to `loom:pr`.

PR #3905, opened 2026-07-05
feat(net-status): classify pour-carried nets as POUR_DISCONTINUOUS

### LINKED ISSUE #3901: net-status --why never classifies pour-carried nets — stuck pour-net residuals invisible
## Symptom

`kct net-status --why` classifies stuck signal nets (ESCAPE_BLOCKED / CONGESTION_SATURATED / PLACEMENT_BOUND) but **silently skips nets whose connectivity is zone/pour-carried**. On the softstart board, both residual defects (NRST star segment and UCC_HO_NEG) were invisible to `--why` during the 2026-07-03 escape study — the classifier output contained no entry for them, so the investigation had to fall back to manual copper analysis.

## Expected

Pour-carried nets that fail the pour-continuity gate (discontinuous islands) should appear in `--why` output with their own class (e.g. `POUR_DISCONTINUOUS`) and the island inventory, or at minimum be listed as "not classified (pour-carried)" so the operator knows the tool didn't look at them.

## Context

Found during the softstart mfg-readiness push (rjwalters/softstart#12 / #26). The pour-discontinuity failure class is real and recurring (softstart pipeline has a dedicated 16-net power-continuity FAIL-row gate), so the classifier gap is not hypothetical.

rjwalters:
## Summary

`kct net-status --why` never classified pour-carried nets — they were silently dropped or misclassified. This adds a dedicated `POUR_DISCONTINUOUS` class so pour-carried nets with stranded pads surface in `--why` output.

## Root cause

In `classify_stuck_nets_from_pcb`:
- The `net_type != "signal"` gate silently discarded zone-backed **plane/power** nets (e.g. NRST, UCC_HO_NEG) before any analysis — no `--why` entry emitted.
- `analysis.advisory_incomplete_names` (ML-tagged pour-carried **signal** nets) was never consulted, so those nets passed both guards and received an incorrect signal-failure classification.

## Change

- **`stuck_classifier.py`**: add `POUR_DISCONTINUOUS` to `StuckClass` (description + `failure_cause` → `FailureCause.BLOCKED_PATH`). Build `pour_carried_names` = `advisory_incomplete_names` ∪ incomplete plane/power nets. Guard the signal loop against that set (prevents misclassification), then a second collection pass emits POUR_DISCONTINUOUS diagnoses with stranded pads from `NetStatus.unconnected_pads` and the zone layer list. No zone-fill geometry analysis (island counting is a possible follow-on).
- **`net_status_cmd.py`**: render the new class in the `--why` text summary and detail ordering. JSON path unchanged (already generic).
- **`tests/router/test_stuck_classifier.py`**: `TestPourDiscontinuous` (3 tests) + `_pour_discontinuous_board()` fixture; updated `test_counts_and_to_dict` key set.

## Advisory interception

The curated third case — signal-named nets the ML classifier tags `is_pour_net=True` — is handled by unioning `analysis.advisory_incomplete_names` into `pour_carried_names` and adding `if net.net_name in pour_carried_names: continue` immediately after the existing `is_advisory_incomplete` guard in the signal loop. Those nets are then re-collected by the second pass under POUR_DISCONTINUOUS instead of receiving a signal-failure class.

## Tests

- `tests/router/test_stuck_classifier.py`: 9 passed (3 new)
- `tests/test_net_status_cmd.py` + `tests/test_analysis_net_status.py`: 82 passed
- `ruff check` / `ruff format`: clean on touched files
- CLI smoke test confirms `[POUR_DISCONTINUOUS] VCC` renders with stranded pad + layer list, exit code 2

Pre-existing mypy error at `net_status_cmd.py:279` (`output_json`, untouched by this PR) is out of scope.

Closes #3901

### DIFF
--- a/src/kicad_tools/cli/net_status_cmd.py
+++ b/src/kicad_tools/cli/net_status_cmd.py
@@ -304,6 +304,7 @@ def output_why(pcb_path: Path, fmt: str) -> int:
     print(f"  ESCAPE_BLOCKED:       {counts['escape_blocked']}")
     print(f"  CONGESTION_SATURATED: {counts['congestion_saturated']}")
     print(f"  PLACEMENT_BOUND:      {counts['placement_bound']}")
+    print(f"  POUR_DISCONTINUOUS:   {counts['pour_discontinuous']}")
     print()
 
     if not result.diagnoses:
@@ -312,7 +313,12 @@ def output_why(pcb_path: Path, fmt: str) -> int:
         return 0
 
     # Group for readability, escape-blocked first (most upstream).
-    order = {"escape_blocked": 0, "congestion_saturated": 1, "placement_bound": 2}
+    order = {
+        "escape_blocked": 0,
+        "congestion_saturated": 1,
+        "placement_bound": 2,
+        "pour_discontinuous": 3,
+    }
     for diag in sorted(result.diagnoses, key=lambda d: order[d.classification_value]):
         pads = ", ".join(diag.unconnected_pads) or "(none)"
         print(f"[{diag.classification.value.upper()}] {diag.net_name}")
--- a/src/kicad_tools/router/stuck_classifier.py
+++ b/src/kicad_tools/router/stuck_classifier.py
@@ -125,6 +125,7 @@ class StuckClass(Enum):
     ESCAPE_BLOCKED = "escape_blocked"
     CONGESTION_SATURATED = "congestion_saturated"
     PLACEMENT_BOUND = "placement_bound"
+    POUR_DISCONTINUOUS = "pour_discontinuous"
 
     @property
     def description(self) -> str:
@@ -144,6 +145,11 @@ class StuckClass(Enum):
                 "congestion -- no routing order closes it. Fix: placement "
                 "nudge (M3)."
             ),
+            "pour_discontinuous": (
+                "Pour-carried net has stranded pads -- copper zone does not reach "
+                "all pads on this net. Fix: re-pour, add stitching vias, or bridge "
+                "the zone island gap."
+            ),
         }[self.value]
 
     @property
@@ -153,6 +159,7 @@ class StuckClass(Enum):
             "escape_blocked": FailureCause.PIN_ACCESS,
             "congestion_saturated": FailureCause.CONGESTION,
             "placement_bound": FailureCause.ROUTING_ORDER,
+            "pour_discontinuous": FailureCause.BLOCKED_PATH,
         }[self.value]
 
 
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

Solid, well-scoped implementation that matches the curator's acceptance criteria exactly.

## Acceptance criteria — all verified

- **Pour-carried nets surface under POUR_DISCONTINUOUS.** Confirmed on a real board (`boards/05-bldc-motor-controller/output/bldc_controller_routed.kicad_pcb`): `+24V` (F.Cu, 7 stranded pads) and `GND` (B.Cu, 33 stranded pads) now emit `[POUR_DISCONTINUOUS]` entries with the stranded-pad inventory and zone-layer list. Both were previously dropped by the `net_type != "signal"` gate.
- **No pour-carried net in signal classes.** The two collection passes are mutually exclusive by net name: the signal loop `continue`s on `net.net_name in pour_carried_names`, and the second pass requires membership. No double-counting is possible — each net appears at most once across `analysis.nets`.
- **No false positives.** The `status != "incomplete"` guard in the second pass keeps fully-connected pour nets out. Confirmed on `voltage_divider_routed.kicad_pcb`: 0 POUR_DISCONTINUOUS, exit 0.
- **JSON contract.** `summary.counts.pour_discontinuous` present; each pour diagnosis carries `classification: "pour_discontinuous"` and `failure_cause: "blocked_path"`.
- **`counts` covers all classes.** `counts` is built from `{c.value: 0 for c in StuckClass}` (line 218), so the direct `counts['pour_discontinuous']` index in `output_why` can never KeyError even when zero.
- **"no zone definition" fallback** renders sensibly for power-named nets without a zone (used in the synthetic test fixture).
- Existing signal-class behavior unchanged; `TestClassifierAggregate.test_counts_and_to_dict` updated with the new key.

## Tests

- `tests/router/test_stuck_classifier.py` + `tests/test_net_status_cmd.py` + `tests/test_analysis_net_status.py`: **91 passed** locally (worktree, coverage disabled). Includes the 3 new `TestPourDiscontinuous` cases.
- `ruff check` + `ruff format --check` clean on all three touched files.

## CI

All checks green, `mergeStateStatus: CLEAN`. Board 05 Routing Regression passed (38m) — this PR touches only the classifier and CLI, not the router, so no routing impact as expected.

Note (non-blocking, already disclosed in PR body): pre-existing mypy error at `net_status_cmd.py:279` (`output_json`) is untouched by this PR and out of scope. Island counting on zone-fill geometry is correctly deferred as a follow-on.

Issue #3931, opened 2026-07-06
Board 07 ADDR_BUS carries a via-count mismatch (A4/A6 extra via pair) — via-imbalance now visible since #3915


rjwalters:
## Summary

Board 07's `ADDR_BUS` match group carries a **via-count mismatch**: the A4/A6 members route with an extra via pair relative to A0, adding ~1.6 mm of via-transition length that pushes the group past its 0.5 mm length-match tolerance. This surfaces as a single `match_group_length_skew` error.

## Why it is newly visible

Before PR #3915, `DRCChecker.check_match_group_length_skew` was **via-blind** — it did not thread `board_thickness_mm` / `num_copper_layers` into `derive_group_skew_data`, so via-transition length was not counted and board 07's ADDR_BUS reported 0 `match_group_length_skew` violations. PR #3915 threads those stackup sources through the canonical producer path (`MatchGroupTracker.measure_net_from_pcb`), so the pre-existing via imbalance is now correctly measured and flagged.

This is a **real, pre-existing defect that #3915 exposed**, not a regression introduced by #3915.

## Impact on CI (reconciled in #3915)

The `Match-Group Routing Regression (boards/07-matchgroup-test)` job's from-scratch re-route (seed 42) now reports:
- `match_group_length_skew: 1` (was 0 while via-blind)
- total DRC error count `29` (the +1 is this same finding entering the total)

To unblock #3915 without silently widening the gate, that PR:
- adds `boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb` to `MATCHGROUP_VIOLATION_BASELINE = 1` in `scripts/ci/check_matchgroup_coverage.py`, referencing this issue;
- bumps the board 07 entry in `.github/routed-drc-tolerance.yml` `28 -> 29`, referencing this issue.

## Resolution options (to drop the baseline back to 0)

1. **Via-balanced routing** — re-route ADDR_BUS so A4/A6 no longer carry an extra via pair vs A0, bringing the group back within the 0.5 mm length-match tolerance, then re-commit `matchgroup_test_routed.kicad_pcb`; **or**
2. **Artifact refresh** — regenerate the committed routed artifact from a via-balanced route.

[...]

Issue #3925, opened 2026-07-05
Regenerated artifacts are un-reviewable: uuid/format churn, seed-pinned census exemptions, un-gitignored byproducts


rjwalters:
Found during the 2026-07-05 fresh-design sweep: all 8 demo boards were rebuilt from scratch in isolated worktrees (fresh `uv sync` + `kct build-native` + wiped `output/`), cross-gated with both `kct check --mfr jlcpcb` and `kicad-cli pcb drc --refill-zones`. Full findings: `.claude/fresh-sweep-findings.md` on the operator machine.

**Severity: MEDIUM (artifact hygiene / regen workflow)**

Sweep-wide observations blocking any "refresh committed artifacts" PR:

1. **Serialization churn:** rebuilds re-serialize tracked s-expr files with tabs-vs-spaces + fresh uuid4s — board 02's unrouted PCB diff is 955 lines with zero content change; board 04's footprint order differs between identical seeded runs. `git diff` review of a regen is impossible.
2. **Census uuid pinning:** `tests/test_fleet_45_census.py` `DOCUMENTED_OFF_ANGLE` pins a seed-run-specific uuid (board 06: `864fb9ee-...`), so ANY fresh regen fails the census; QUANTIZE_SKIP_UUIDS has the same brittleness. Exemptions need content-addressed keys (net + endpoints), not uuids.
3. **Byproducts neither committed nor gitignored:** `*.kct.json`, routed `.kicad_prl`/`.kicad_dru`/`.kicad_pro`, `drc_report.json`, `erc_report.json`, `lvs.json` (boards 01/02/03/04). renders/ and board.json are correctly ignored — extend the pattern.
4. **Stable-sort + canonical formatting** (or a `kct diff-artifacts` tool) would make regen PRs reviewable; determinism claims should state they are per-router-version.

Sub-items 2 and 3 are small, independently landable fixes.

rjwalters:
## Curator Enhancement

Curating per user request. Filing concrete technical findings for each of the three named sub-problems, a split recommendation, and acceptance criteria/test plan.

---

## Split Recommendation

This issue covers three separable defects of very different cost and risk:

| Sub-problem | Scope | Risk | Effort |
|---|---|---|---|
| (A) UUID/format churn | Generator change (primitives.py, zones/generator.py, io.py serialization) | Medium — must not break board-06 determinism smoke test | 1–2 days |
[...]

PR #3933, opened 2026-07-06
fix(match-group-skew): length-check groups composed only of diff pairs

### LINKED ISSUE #3916: Match groups composed only of diff-pairs are never length-checked (Phase 1B scope skip makes MIPI/HDMI tolerances dead letters)

**Severity: HIGH (checker coverage gap)**

`match_group_skew.py` skips any group with empty single-ended `net_ids` ("no single-ended members in Phase 1B scope"). On board 07 that means MIPI_CSI_LANES (0.05mm tolerance) and HDMI_TMDS_LANES (0.075mm) are **structurally uncheckable** — even fully routed, `rules_checked` stays 0. Measured actuals on the fresh artifact: MIPI group skew 23.19mm, HDMI 8.69mm — hundreds of times over tolerance, zero findings.

Of the 4 bench scenarios on the match-group regression board, only ADDR_BUS is ever gated (and it is via-blind, see companion issue). The bench currently regression-tests almost nothing it was built for.

**Expected:** pair members participate in group skew via their pair-average (or per-leg) lengths; `rules_checked_by_rule.match_group_length_skew` equals the number of declared groups.

**Repro:** fresh build boards/07-matchgroup-test; inspect `rules_checked_by_rule` in kct check JSON.

rjwalters:
## Summary

`derive_group_skew_data` never length-checked a match group whose members
were exclusively differential pairs. After `detect_match_groups` runs
`_extract_pair_ids`, each `_P`/`_N` net is moved out of `net_ids` into
`pair_ids`, so groups like board 07's MIPI_CSI_LANES and HDMI_TMDS_LANES
exit detection with `net_ids=[]`. The `if not grp.net_ids: continue` guard
then silently dropped the whole group and the `match_group_length_skew`
rule never fired on it, regardless of how large the lane-to-lane skew was.

## Changes

- **`src/kicad_tools/validate/match_group_skew.py`** — extend the
  measurement loop to also measure `pair_ids` members. Each diff pair
  contributes the **pair-average** `(L_P + L_N) / 2` as a single
  lane-length entry (not both legs), so group skew is purely between-lane
  and does not double-count within-pair imbalance (already gated by the
  separate `diffpair_length_skew` rule). The early-exit now only fires when
  both `net_ids` and `pair_ids` are empty; the unrouted-member gate applies
  to both legs of a pair; the threshold-map anchor falls back to the P-leg
  of the first pair for pair-only groups. Updated the module docstring.
- **`tests/test_validate_match_group_skew.py`** — new `TestDeriveGroupSkewDataPairOnly`
  class (5 tests) covering pair-only skew, pair-average vs per-leg,
  one-unrouted-leg gating, single-pair below-min, and mixed member groups.
- **`scripts/ci/check_matchgroup_coverage.py`** — raise
  `MATCHGROUP_VIOLATION_BASELINE` for board 07 to **2** to reconcile the
  **reroute** path the Match-Group Routing Regression CI job exercises
  (see verification below), with a comment documenting both gate paths.
- **`tests/test_check_matchgroup_coverage.py`** — update the baseline unit
  test to assert **2** with the corrected composition.

## Baseline verification (both gate paths)

The observed match-group violation count depends on which artifact the
gate runs against:

- **`--skip-route`** (committed-artifact path): **1**. On the committed
  board 07 artifact MIPI/HDMI each still have an unrouted diff-pair leg
  (MIPI_CLK_N, MIPI_DAT0_N, TMDS_D0_*, TMDS_D1_* carry zero geometry) and
  DDR_DATA_BYTE_0 has an unrouted DQ3, so all three are gated out by the
  partial-routing rule — only fully-routed ADDR_BUS fires.
- **full reroute** (the Match-Group Routing Regression CI job, which runs
  `generate_design.py --step route` end-to-end): **2**, verified locally
  with `PYTHONHASHSEED=42 ... --seed 42` (matches CI run 28771297359). The
  reroute lands every MIPI leg, so **MIPI_CSI_LANES** now fires a genuine
  `match_group_length_skew` error on top of ADDR_BUS's via-imbalance.

**Why not 3?** An earlier prediction assumed HDMI_TMDS_LANES would also
fire once routed. It does not: even after the full reroute, HDMI's
`TMDS_D0_N` / `TMDS_D1_N` legs fail to route (the router reports them
"still stranded — NO relief path"), so the unrouted-leg gate correctly
excludes HDMI. The reroute count is therefore **2**, not 3.

A single baseline of **2** satisfies both gate paths: the reroute count of
2 is `<= 2`, and the `--skip-route` count of 1 is also `<= 2`.
Composition of the reroute baseline: 1× ADDR_BUS via-imbalance (#3931) +
1× MIPI_CSI_LANES pair-only skew (#3916). The
`.github/routed-drc-tolerance.yml` allowlist is intentionally NOT touched
(these are `match_group_length_skew` error-severity violations, tracked by
the baseline, not the DRC floor).

| AC1 — coverage for all declared groups | ✅ mechanism landed | Pair-only groups are now measured; the full reroute fires MIPI_CSI_LANES. HDMI stays gated because its TMDS legs do not route even after the reroute. |
| AC2 — MIPI produces findings | ✅ | Full reroute fires MIPI_CSI_LANES; unit test `test_pair_only_group_produces_skew` |
| AC3 — HDMI produces findings | ✅ mechanism landed | Same mechanism; HDMI's TMDS_D0/D1 legs stay unrouted after the reroute so it is correctly gated (not a producer gap) |
| AC4 — DDR/ADDR still checked | ✅ | ADDR_BUS still fires; `test_mixed_single_ended_and_pair_members` guards the mixed path |
| AC5 — CI baseline reconciled | ✅ | Reroute path verified at 2; `MATCHGROUP_VIOLATION_BASELINE` raised to 2 with both-path documentation; `--skip-route` still passes (1 <= 2) |
| AC6 — pair-only group produces skew | ✅ | `test_pair_only_group_produces_skew` |
| AC7 — pair-average semantics | ✅ | `test_pair_average_not_per_leg_semantics` asserts 0.25mm (avg) not 0.30mm (per-leg) |
| AC8 — one unrouted leg skips group | ✅ | `test_pair_with_one_unrouted_leg_omits_group` |

## Test Plan

- `uv run pytest tests/test_validate_match_group_skew.py` — pass (26 prior + 5 new)
- `uv run pytest tests/test_check_matchgroup_coverage.py tests/test_board_07_matchgroup_test.py tests/test_validate_match_group_length_skew.py` — all pass
- Full reroute (`PYTHONHASHSEED=42 uv run python scripts/ci/check_matchgroup_coverage.py boards/07-matchgroup-test --seed 42`) — match-group violations = 2 (== baseline)
- `--skip-route` on committed artifact — match-group violations = 1 (within baseline)
- `ruff check` / `ruff format --check` clean on all changed files

Pre-existing (unrelated, present on `main`): the board 07 committed
artifact's DRC error count drifts above the routed-drc-tolerance allowlist
floor — this artifact-churn drift (#3925) predates and is out of scope for
this change; the reroute path's DRC count (32) is within the allowlist.

Closes #3916

### DIFF
--- a/scripts/ci/check_matchgroup_coverage.py
+++ b/scripts/ci/check_matchgroup_coverage.py
@@ -120,6 +120,23 @@ MATCHGROUP_RULE_IDS: tuple[str, ...] = ("match_group_length_skew",)
 # error.  Dropping this baseline back to strict 0 requires a via-balanced
 # re-route (or an artifact refresh), which is blocked on the artifact-churn
 # problem tracked in #3925.  See #3931 for the resolution plan.
+#
+# Issue #3916 (pair-only groups now length-checked) -- the producer used to
+# skip any match group whose members were exclusively differential pairs
+# (``net_ids=[]`` after ``_extract_pair_ids``), so board 07's MIPI_CSI_LANES
+# and HDMI_TMDS_LANES were never skew-checked.  ``derive_group_skew_data`` now
+# measures diff-pair members via the pair-average ``(L_P + L_N) / 2``
+# contribution, making both groups checkable.  On the COMMITTED board 07
+# artifact this baseline stays at 1: MIPI/HDMI each still have at least one
+# unrouted diff-pair leg (e.g. MIPI_CLK_N, MIPI_DAT0_N, TMDS_D0_*, TMDS_D1_*
+# carry zero geometry) and DDR_DATA_BYTE_0 has an unrouted DQ3, so all three
+# are correctly gated out by the existing partial-routing rule -- only the
+# fully-routed ADDR_BUS group produces a violation.  Once board 07 routes
+# those lanes end-to-end (blocked on the same artifact-churn work as #3931 /
+# #3925), MIPI (measured ~23mm skew vs 0.05mm) and HDMI (~8.7mm vs 0.075mm)
+# will each add a genuine ``match_group_length_skew`` error and this baseline
+# must be raised to 3 with a reroute of the committed artifact.  Composition
+# today: 1 = ADDR_BUS via-imbalance (#3931).
 MATCHGROUP_VIOLATION_BASELINE: dict[str, int] = {
     "boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb": 1,
 }
--- a/scripts/ci/check_matchgroup_coverage.py
+++ b/scripts/ci/check_matchgroup_coverage.py
@@ -126,19 +126,36 @@ MATCHGROUP_RULE_IDS: tuple[str, ...] = ("match_group_length_skew",)
 # (``net_ids=[]`` after ``_extract_pair_ids``), so board 07's MIPI_CSI_LANES
 # and HDMI_TMDS_LANES were never skew-checked.  ``derive_group_skew_data`` now
 # measures diff-pair members via the pair-average ``(L_P + L_N) / 2``
-# contribution, making both groups checkable.  On the COMMITTED board 07
-# artifact this baseline stays at 1: MIPI/HDMI each still have at least one
-# unrouted diff-pair leg (e.g. MIPI_CLK_N, MIPI_DAT0_N, TMDS_D0_*, TMDS_D1_*
-# carry zero geometry) and DDR_DATA_BYTE_0 has an unrouted DQ3, so all three
-# are correctly gated out by the existing partial-routing rule -- only the
-# fully-routed ADDR_BUS group produces a violation.  Once board 07 routes
-# those lanes end-to-end (blocked on the same artifact-churn work as #3931 /
-# #3925), MIPI (measured ~23mm skew vs 0.05mm) and HDMI (~8.7mm vs 0.075mm)
-# will each add a genuine ``match_group_length_skew`` error and this baseline
-# must be raised to 3 with a reroute of the committed artifact.  Composition
-# today: 1 = ADDR_BUS via-imbalance (#3931).
+# contribution, making both groups checkable.  This baseline is now **2**, but
+# the value the gate observes depends on WHICH artifact it runs against:
+#
+#   * ``--skip-route`` (committed-artifact path, used by the diff-driven
+#     routed-pcb-drc-check and by local verification): counts exactly **1**
+#     match-group violation.  On the COMMITTED board 07 artifact MIPI/HDMI
+#     each still have at least one unrouted diff-pair leg (e.g. MIPI_CLK_N,
+#     MIPI_DAT0_N, TMDS_D0_*, TMDS_D1_* carry zero geometry) and
+#     DDR_DATA_BYTE_0 has an unrouted DQ3, so all three are gated out by the
+#     partial-routing rule -- only the fully-routed ADDR_BUS group fires.
+#
+#   * full reroute (the Match-Group Routing Regression CI job, which runs
+#     ``generate_design.py --step route`` end-to-end before the gate):
+#     counts **2**.  The reroute lands every MIPI leg, so MIPI_CSI_LANES now
+#     produces a genuine ``match_group_length_skew`` error on top of
+#     ADDR_BUS's via-imbalance.  HDMI_TMDS_LANES does NOT fire even after the
+#     reroute: its TMDS_D0_N / TMDS_D1_N legs remain unrouted (the router
+#     reports them "still stranded -- NO relief path"), so the unrouted-leg
+#     gate correctly excludes HDMI.  A single baseline of 2 satisfies BOTH
+#     paths (the ``--skip-route`` count of 1 also passes, since 1 <= 2).
+#
+# Composition of the reroute baseline (2):
+#   1x ADDR_BUS via-imbalance (#3931) + 1x MIPI_CSI_LANES pair-only skew
+#   (#3916, the group newly made checkable by this change).
+# (An earlier prediction of 3 assumed HDMI would also fire once routed; the
+# reroute shows HDMI's TMDS legs do not route, so the observed count is 2.
+# Dropping this baseline toward strict 0 requires a via-balanced,
+# fully-routed board 07 -- blocked on the artifact-churn work in #3925.)
 MATCHGROUP_VIOLATION_BASELINE: dict[str, int] = {
-    "boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb": 1,
+    "boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb": 2,
 }
 
 
--- a/src/kicad_tools/validate/match_group_skew.py
+++ b/src/kicad_tools/validate/match_group_skew.py
@@ -51,13 +51,25 @@ imports from router/match_group_length.py only).  Tests live in
 ``tests/test_validate_match_group_skew.py`` and the drift-prevention
 test mirrors PR #2685's pattern.
 
-Group-of-pairs note (Phase 2F reserve):
+Group-of-pairs composition (Issue #3916):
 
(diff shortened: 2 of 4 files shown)


rjwalters:
## Changes Requested — CI gate failing (the fix works, the baseline wasn't reconciled for the reroute path)

**Verdict: request changes.** The core producer change is correct and the test coverage is genuinely strong (AC6/AC7/AC8 are all pinned with good assertions, including the anti-pattern guard that skew must be the pair-average 0.25mm, not the per-leg 0.30mm). But the **Match-Group Routing Regression (boards/07-matchgroup-test)** CI job is failing, and it is failing *because* the fix now works.

### The failure

From the failing job (run 28771297359):

```
[matchgroup-coverage] rules_checked_by_rule: {..., 'match_group_length_skew': 2}
[matchgroup-coverage] match-group error violations: {'match_group_length_skew': 2} (baseline: 1)
##[error]match-group error-severity violations (2) EXCEED the documented baseline (1): match_group_length_skew=2
Gate failed (exit 2).
```

Before this change, pair-only groups were never checked, so a freshly-routed board 07 produced exactly 1 match-group violation (ADDR_BUS). Your fix correctly enables MIPI_CSI_LANES, and on the re-routed artifact MIPI now fires a genuine `match_group_length_skew` error → total **2**, exceeding the baseline of **1**.

### Root cause of the miss

The baseline reconciliation was validated only against the **committed** artifact via `--skip-route` (which stays at 1 because MIPI/HDMI have unrouted legs there). But the **Match-Group Routing Regression** CI job *re-routes* board 07 end-to-end before running the coverage gate — which is exactly the scenario the curator's AC5 flagged: "quantify the new violation count **on the re-routed board 07 artifact** and raise `MATCHGROUP_VIOLATION_BASELINE` accordingly." The PR body even predicts this ("the baseline rises ... with a reroute") but the actual `MATCHGROUP_VIOLATION_BASELINE` value was never updated, so the reroute path trips the gate.

Two additional notes:
- The PR predicted the reroute count would be **3** (ADDR_BUS + MIPI + HDMI). CI observed **2** — only MIPI fired, not HDMI. Please reconcile this discrepancy (HDMI legs may still be unrouted / within tolerance even after the reroute) rather than assuming 3.
- Bumping to the empirically-observed reroute count (2) still passes the `--skip-route` committed-artifact path (`1 <= 2`), so a single baseline value can satisfy both gate paths.

### What to do

1. Run the full **reroute** locally (not `--skip-route`) the way the CI job does, and confirm the count is deterministically 2 (or whatever it stably is): re-generate the routed board 07 artifact, then `python scripts/ci/check_matchgroup_coverage.py boards/07-matchgroup-test`.
2. Raise `MATCHGROUP_VIOLATION_BASELINE["boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb"]` to the verified count with a tracking comment (composition: ADDR_BUS #3931 + the newly-enabled MIPI skew from #3916). The gate explicitly sanctions this ("if this is an accepted, separately-tracked baseline, raise `MATCHGROUP_VIOLATION_BASELINE` ... with a tracking reference").
3. Update the baseline comment: it currently claims the reroute count will be **3**, but CI shows **2** — correct the documented composition to match observed reality.

Once the Match-Group Routing Regression check goes green, I'll re-review. The rest of the diff (measurement loop, pair-average semantics, unrouted-leg gating, threshold-map fallback to the P-leg of the first pair, docstring) all looks correct and reads cleanly.

rjwalters:
Addressed the Match-Group Routing Regression gate feedback.

**Verified the reroute count locally** the same way the CI job does — `PYTHONHASHSEED=42 uv run python scripts/ci/check_matchgroup_coverage.py boards/07-matchgroup-test --seed 42` (full end-to-end reroute, no `--skip-route`). Result is deterministically **2** `match_group_length_skew` violations, matching CI run 28771297359.

**Why 2, not 3 (HDMI does not fire):** even after the full reroute, HDMI_TMDS_LANES' `TMDS_D0_N` / `TMDS_D1_N` legs fail to route — the router reports them "still stranded — NO relief path" — so the unrouted-leg gate correctly excludes HDMI. Only MIPI_CSI_LANES (all legs routed) fires, on top of ADDR_BUS's via-imbalance. This is the fix behaving correctly, not a producer gap, so no code change to `match_group_skew.py` was needed.

**Changes:**
- `scripts/ci/check_matchgroup_coverage.py` — raised `MATCHGROUP_VIOLATION_BASELINE` for board 07 to **2** with a rewritten comment documenting both gate paths (`--skip-route`: 1; reroute: 2) and the corrected composition (ADDR_BUS #3931 + MIPI_CSI_LANES #3916). A single value of 2 satisfies both paths (1 <= 2).
- `tests/test_check_matchgroup_coverage.py` — baseline unit test now asserts 2 with the corrected rationale.
- Corrected the PR body misprediction (reroute count 3 → 2, with the HDMI explanation).

**Local verification:** reroute gate now passes (2 == baseline 2); `--skip-route` committed-artifact path passes (1 <= 2); 127 tests pass across the 4 relevant suites; ruff + format clean. Ready for re-review.

rjwalters:

The prior rejection was purely about the CI gate reconciliation, and that is now correctly resolved.

## Verification

**CI — all green, merge state CLEAN.** The key signal, **Match-Group Routing Regression (boards/07-matchgroup-test)**, now **passes** (13m43s). Every other required check passed as well (Lint, Type Check, Test, all board End-to-End jobs, Routed PCB DRC, Diff-Pair Routing Regression, Board 05 Routing Regression at 33m53s, C++ Build, kicad-cli smoke). No failing or pending checks.

**Baseline bump is justified reconciliation, not violation-masking.** `MATCHGROUP_VIOLATION_BASELINE` for board 07 raised 1 → 2 in `scripts/ci/check_matchgroup_coverage.py`, with a rewritten comment documenting *both* gate paths:
- `--skip-route` (committed artifact): **1** — MIPI/HDMI each still carry an unrouted diff-pair leg and DDR has an unrouted DQ3, so only ADDR_BUS fires.
- full reroute (the CI regression job): **2** — the reroute lands every MIPI leg, so MIPI_CSI_LANES fires a genuine `match_group_length_skew` on top of ADDR_BUS's via-imbalance.

Composition of the reroute baseline is explicit: **1× ADDR_BUS via-imbalance (#3931, pre-existing/separate)** + **1× MIPI_CSI_LANES pair-only skew (#3916, the fix working)**. A single value of 2 satisfies both paths (1 ≤ 2). The earlier "3" misprediction is corrected in the comment and PR body.

**HDMI correctly does not fire.** Even after the full reroute, HDMI's `TMDS_D0_N`/`TMDS_D1_N` legs fail to route ("still stranded — NO relief path"), so the unrouted-leg gate excludes it. This is the producer working as designed, not a coverage gap — confirmed by the Doctor and consistent with the gating logic.

**Core fix unchanged since prior substantive approval.** `src/kicad_tools/validate/match_group_skew.py` is untouched by the Doctor's pass — the only new commit (5aa73a76) touches solely the CI coverage script and its unit test; the producer change lives entirely in the original commit 7baee98a. The pair-average `(L_P + L_N)/2` semantics, both-legs unrouted gating, and P-leg net-class anchor fallback all still read correctly.

**#3925 (board-07 raw DRC drift) is genuinely out of scope.** That is a separate artifact-churn issue about the committed board's DRC error count vs the routed-drc-tolerance floor — unrelated to `match_group_length_skew` error-severity violations, which are tracked by this baseline. The `.github/routed-drc-tolerance.yml` allowlist is correctly left untouched.

The test assertion in `tests/test_check_matchgroup_coverage.py` was updated to expect 2 with the corrected rationale, and the committed board artifact was restored (no artifact appears in the diff). Closes #3916 correctly.

Solid work reconciling the reroute path. Approving.

PR #3930, opened 2026-07-06
fix(stitch): reject connectivity-fallback vias that graze foreign pour

rjwalters:
## Summary

The pour-connectivity fallback (#3633) intentionally drops just-placed sibling stitch geometry from the obstacle pool so a stranded pour pad can reclaim its load-bearing bridge via. That is only safe when the sole blocker was self-inflicted stitch geometry. When the pre-existing poured copper of a **different** net already overlaps the candidate position, dropping the sibling obstacles exposes a via that lands on the foreign fill polygon — once KiCad re-fills zones the boundary resolves against the new via and produces a genuine cross-net short (#3910).

This implements the curator's **recommended Option A** (reject, don't place) as the structural fix, plus **Option C** as an interim guard (build-step warning). Option B (place-then-verify with kicad-cli) was declined per the curator's own assessment ("over-engineered for this failure mode").

## Changes

- **`stitch_cmd.py`**: New helper `_fallback_grazes_foreign_pour()` re-checks the committed fallback via pad circle (radius = `via_size / 2`) against foreign pour fill polygons only, reusing the existing `_check_point_filled_polygon_clearance`. At the fallback site in `run_stitch()`, a graze now degrades the pad to `pads_skipped` (with a descriptive reason) instead of appending to `connectivity_fallback`. The self-inflicted sibling-graze rescue #3633 was designed for is preserved — only foreign **pour** overlap is rejected.
- **`build_cmd.py`**: `_run_step_stitch` now surfaces a non-empty `result.connectivity_fallback` as a visible warning advising a `kicad-cli pcb drc --refill-zones` cross-check.
- **Tests**: helper predicate (graze/clear; straight and dog-leg), forced-fallback rejection into `pads_skipped`, the invariant that no committed fallback via grazes foreign pour, and a build-step warning test.

The pipeline shell path (`pipeline_cmd.py`) inherits the fix automatically: it invokes the same `run_stitch()` via `kct stitch`, so the rejection is entry-point agnostic. No pipeline flag was added (would be scope creep beyond the geometry fix).

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| Fallback via grazing a foreign pour polygon is rejected → pad recorded in `pads_skipped`, not `connectivity_fallback` | ✅ | `test_fallback_on_foreign_pour_lands_in_pads_skipped` |
[...]

Issue #3794, opened 2026-06-18
board-04: power-net (+3.3V/GND) copper-LVS opens — pads not galvanically bonded to the pour


rjwalters:
## Problem

A fresh copper-LVS run (current extractor, post-#3772/#3792) on board-04's committed routed PCB reports **0 shorts / 20 opens**, all of which are **same-net power-pad opens**:

```
open: +3.3V <-> +3.3V (C12.1, U2.24)
open: +3.3V <-> +3.3V (U2.24, U2.36)
open: +3.3V <-> +3.3V (U2.36, U2.48)
open: +3.3V <-> +3.3V (U2.48, U2.9)
open: GND   <-> GND   (C1.2, C10.2)
... (20 total, all +3.3V or GND)
```

These are NOT real electrical defects (the board ships fine at jlcpcb-tier1; `kct check --mfr jlcpcb-tier1` is 0-blocking) and they are NOT shorts — they are *same-net* opens: power pads that the copper extractor does not see galvanically bonded to their pour. The real board-04 defects (OSC_IN<->OSC_OUT short #3785, net drift #3765) are already fixed; this is the last residual keeping board-04's copper-LVS dirty.

## Why it matters

Board-04 is in `ADVISORY_LVS_BOARDS` and shows "LVS mismatch" on the kicad-tools.org gallery chip. Closing these opens is the prerequisite to graduating board-04 from advisory to a hard copper-LVS gate (the sibling of board-03's #3787).

## Facts established

- Board-04 is **2-layer** (F.Cu/B.Cu) — this is NOT a 4-layer internal-plane issue.
- It has F.Cu and B.Cu pour zones. Of 28 vias, **14 are GND, 0 are +3.3V, 0 are +5V**.
- The pattern mirrors board-03 before #3787 (GND F.Cu/B.Cu planes that needed stitching vias to be galvanically one island).

## Investigate / fix (diagnosis is open)

Determine which of these it is, then fix accordingly:
- [ ] **Board fix (most likely, cf #3787)**: the +3.3V (and some GND) pads are served by a pour on one layer but not stitched/bonded to it — add the needed stitching vias / fix the pour so every power pad is galvanically in the pour island. Reuse the `add_gnd_stitching_vias` pattern from board-03's `generate_design.py` (#3789), generalized to +3.3V.
[...]

PR #3946, opened 2026-07-06
fix(lvs): detect via-to-foreign-pour shorts and name nets on negative-clearance overlaps

### LINKED ISSUE #3909: Copper-LVS reports PASS (0 shorts) on boards with real net shorts that kicad-cli detects

**Severity: CRITICAL (correctness — ships electrically wrong boards as clean)**

Two independent reproductions in one sweep:

- **Board 07** (fresh build): pipeline printed `copper-LVS PASS: 0 shorts / 0 opens` and `board.json` says `lvs_clean: true`, while `kicad-cli pcb drc --refill-zones` reports **4 `shorting_items` errors**: 3x "Items shorting two nets (+1V2 and +1V8)" plus a DQ1/DQ6 short.
- **Board 05** (fresh regen via design.py): 2 hard shorts (NRST<->OSC_IN via, PWM_AH<->OSC_OUT via) + one 0.024mm near-short. `kct check` surfaced them only as anonymous negative clearances: `clearance_segment_via -0.188mm < minimum 0.102mm` — **no net names**, no "SHORT" wording; kicad-cli names both nets.

**Repro:** fresh build of boards/07-matchgroup-test (`PYTHONHASHSEED=42 generate_design.py --seed 42`), then compare the recipe's LVS verdict against `kicad-cli pcb drc --refill-zones --format json`.

**Expected:**
1. Copper-LVS detects pour/via-created cross-net connections (it currently misses shorts introduced after/by the stitch+fill stages).
2. `kct check` reports negative clearances as named shorts ("SHORT between X and Y"), severity error, never bundled with ordinary clearance messages.

This is the in-pipeline recurrence of the softstart 2119450 class (defect only caught by kicad-cli cross-check). Until fixed, `lvs_clean: true` cannot be trusted as a shipping gate.

rjwalters:
## Summary

Two correctness gaps let a board ship `lvs_clean: true` while `kicad-cli pcb drc --refill-zones` reported real shorts (issue #3909):

1. **Copper-LVS was blind to via-to-foreign-pour shorts.** `_connect_pour_pads_label_free` bonded each synthetic via node to poured copper by testing the via's bare *centre point* against the fill solid region. KiCad carves an antipad clearance hole centred on every foreign-net via, so the via centre always lands in that hole and the point test never bonded it. A via whose *copper ring* pokes past a marginal / too-small antipad into a foreign net's fill — a physical short — was therefore invisible to the Python-fill-based comparator, even though kicad-cli's native refill flags it as `shorting_items`.

2. **`kct check` reported negative clearances net-anonymously.** A negative clearance means the two copper geometries physically overlap (an electrical short), but `_create_violation` emitted `"Segment to via clearance -0.188mm < minimum 0.102mm"` with the net names only visible under `--verbose`.

## Changes

- **`src/kicad_tools/validate/connectivity.py`** — Represent each synthetic via node as its copper *circle* (radius `size/2`, eroded by `POUR_PAD_ERODE` like a pad box) instead of a bare point when bonding to pours. A via whose ring overlaps a foreign pour is now fused into that island, so `compare_partitions` surfaces the short. Added helpers `_synthetic_via_radii()` and `_via_copper_geom()`. The circle strictly contains the old centre point, so every existing via→pour bond (the `pad → trace → via → opposite-layer pour` path, #3794) still fires.
- **`src/kicad_tools/validate/rules/clearance.py`** — When `actual < 0` between two distinct named nets, promote the message to `SHORT: '<a>' and '<b>' copper overlaps by <d>mm (<t1> to <t2>)` at default verbosity. `rule_id` is left as `clearance_<suffix>` so exact-rule_id filters and the `ViolationType.from_string` alias table stay stable.
- **Tests** — new via-to-foreign-pour crux test (old point model masks / new circle model catches), a net-0-via false-positive guard, and negative/positive/same-net clearance-message tests.

| AC1: via overlapping foreign pour → `clean: False` short naming both nets | ✅ | `test_via_grazing_foreign_pour_is_a_short` (old bare-point model masks, new circle model reports GND/SIG short) |
| AC2: board 07 agrees with kicad-cli | ✅ | kicad-cli `shorting_items = 0` and copper-LVS `shorts = 0` on committed board-07 artifact (post-#3930); new code introduces no false short |
| AC3: negative-clearance message names both nets at default verbosity | ✅ | `test_negative_clearance_message_names_both_nets` → `SHORT: 'GND' and '+3V3' copper overlaps by 0.175mm (segment to via)` |
| AC4: `board.json` `lvs_clean` inherits corrected verdict | ✅ | No change needed — `_parse_lvs` reads `lvs.json` `clean`, now driven by the corrected partition |
| AC5: no false positives on boards 00-04 / 07 | ✅ | Board LVS regression tests pass; kicad-cli vs copper-LVS both 0 shorts on 00/03/07 |

## Test Plan

- `tests/test_copper_lvs.py`, `tests/test_clearance_net_names.py`, `tests/test_connectivity*.py`, `tests/test_board_0{0,3,4}*lvs*.py`, `tests/test_drc_violation_type_validate.py`, clearance/segment-via suites — **291 passed**.
- Ground-truth cross-check with `kicad-cli pcb drc` (10.0.4) on committed artifacts: boards 00 / 03 / 07 report 0 `shorting_items`, matching copper-LVS 0 shorts / `clean=True`.
- `ruff check` + `ruff format` clean on changed files; no new mypy errors (pre-existing shapely-stub / `Pad`-type errors in untouched code only).

Closes #3909

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 3880
-# Branch: feature/issue-3880
+# Issue: 3909
+# Branch: feature/issue-3909
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/src/kicad_tools/validate/connectivity.py
+++ b/src/kicad_tools/validate/connectivity.py
@@ -1198,6 +1198,42 @@ class ConnectivityValidator:
             return None
         return poly
 
+    def _synthetic_via_radii(self, synthetic_nodes: set[str]) -> dict[str, float]:
+        """Map each ``__via{index}`` node to its copper radius (``size / 2``).
+
+        The synthetic via node ids created in :meth:`extract_pad_partition`
+        (step 1b) are ``f"__via{index}"`` in board via order, so we recover the
+        copper radius by re-enumerating ``self.pcb.vias``.  Only nodes present
+        in ``synthetic_nodes`` are returned.
+        """
+        radii: dict[str, float] = {}
+        for via_index, via in enumerate(self.pcb.vias):
+            node_id = f"__via{via_index}"
+            if node_id in synthetic_nodes:
+                radii[node_id] = max(getattr(via, "size", 0.0) or 0.0, 0.0) / 2.0
+        return radii
+
+    def _via_copper_geom(self, pos: tuple[float, float], radius: float) -> Any:
+        """Build a shapely geometry approximating a via's copper (issue #3909).
+
+        Returns the via's copper *circle* (a disk of ``radius`` about ``pos``,
+        eroded inward by :data:`POUR_PAD_ERODE` to match the pad-box treatment)
+        so a via whose copper ring overlaps a foreign pour's solid region bonds
+        into that island and surfaces the short.  Falls back to a bare
+        ``Point`` when the via has no positive size (degenerate) so the legacy
+        centre-in-solid bond (issue #3794) still fires.
+        """
+        if radius <= 0:
+            return _ShapelyPoint(pos)
+        circle = _ShapelyPoint(pos).buffer(radius)
+        if self.POUR_PAD_ERODE > 0:
+            eroded = circle.buffer(-self.POUR_PAD_ERODE)
+            # Erosion can empty a very small via; keep the full circle so the
+            # via is still testable rather than silently dropped.
+            if not eroded.is_empty:
+                return eroded
+        return circle
+
     def _connect_via_in_pad(
         self,
         pad_positions: dict[str, tuple[float, float]],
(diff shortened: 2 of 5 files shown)


rjwalters:
✅ **Approved!**

Reviewed the two-defect fix for #3909. Both fixes are correct, conservative, and well-tested; CI is fully green (16/16 checks pass) and merge state is CLEAN.

## Defect 1 — via-to-foreign-pour short detection (`connectivity.py`)
- **Geometry is right.** `_via_copper_geom` models each synthetic via as a disk of `size/2` eroded inward by `POUR_PAD_ERODE` (0.1mm), matching the pad-box treatment. The eroded disk (radius `r−0.1`) still contains the via centre — and falls back to the full circle when erosion would empty a tiny via — so the circle model **strictly contains** the old bare-point model. Every legacy `pad → trace → via → opposite-layer pour` bond (#3794) still fires; no lost bonds.
- **No false-short risk on DRC-clean boards.** The erosion makes the tested circle *smaller* than the real via copper, so it is conservative: when the antipad hole is wider than the via copper (a clean placement) the eroded ring never reaches the solid region. Guarded by `test_via_on_net0_adjacent_to_pour_is_not_a_false_short`.
- **Enumeration consistency verified.** `_synthetic_via_radii` re-enumerates `self.pcb.vias` with the exact `__via{index}` scheme used at synthetic-node creation (connectivity.py:424), so radii map to the correct nodes.

## Defect 2 — named SHORT message (`clearance.py`)
- Downstream stability preserved: `rule_id` stays `clearance_<suffix>` and the `nets` tuple is unchanged. Only the human-readable `message` is promoted to `SHORT: ...`, and only when `actual < 0` between two **distinct named** nets — exact-rule_id filters and `ViolationType.from_string` aliases stay stable.

## Tests
- The crux test `test_via_grazing_foreign_pour_is_a_short` runs the *same* fixture through both a mocked bare-point via model (masks the short → clean) and the real circle model (catches the GND/SIG short), which is exactly the right way to prove the blind spot is closed.
- Good coverage on the boundaries: net-0 false-positive guard, same-net non-short guard, and positive-clearance (tight-but-legal) guard.

## Scope
Diff is scoped to the two defects (`connectivity.py`, `clearance.py`) plus their tests. The `.loom-managed` marker change is pre-existing Loom worktree noise (already tracked in main with a stale value) — harmless, not a code change.

Nice work — the strict-containment reasoning and the paired old-model/new-model crux test are the right rigor for a shorting-detection change.

PR #3934, opened 2026-07-06
fix(drc): tighten clearance guard band to 0.1um and gate board-06 post-quantize

rjwalters:
## Summary

`kct check` silently passed sub-16 um marginal clearance violations and the board-06 recipe had no post-quantize/post-repair re-validation. Two root causes, fixed here:

1. **Epsilon masking (Bug 2)**: `DRC_TOLERANCE` was 0.005 mm (5 um), a dead band that swallowed any clearance shortfall within 5 um of the floor. The marginal class this issue is about is ~1.6 um wide (0.1000 mm actual vs a 0.1016 mm / 4 mil floor). The constant only exists to suppress IEEE-754 float64 rounding (~1e-9 mm at board coordinates), so it is reduced to **1e-4 mm (0.1 um)** and documented.
2. **Pipeline gap (Bug 1)**: board-06's step 12 (quantize + re-fill) returned with no clearance gate, certifying a -0.337 mm segment-to-via short that `kicad-cli` flagged. `quantize.py`'s `dogleg_points()` docstring already obliges callers that mutate committed copper to re-verify clearances. A **step-12d gate** now does so.

**Bug 3 (JLCPCB YAML floor) is intentionally scoped out.** The curator's premise (YAML says 0.127) refers to the **2-layer 1oz** tier; the **4-layer** tier board 06 uses already declares 0.1016 mm. Per JLCPCB's published capabilities the 2-layer standard minimum is genuinely 5 mil (0.127 mm), so that value is correct, not an error. Changing it would loosen every 2-layer board's clearance floor and risks masking real violations — a repo-wide DRC-verdict change I judged too risky to bundle. A follow-up comment is left on the issue.

## Changes

- `src/kicad_tools/validate/rules/base.py`: `DRC_TOLERANCE` 0.005 -> 1e-4 mm, with a comment explaining the magnitude vs float64 precision at board coordinates.
- `boards/06-diffpair-test/generate_design.py`: add `_clearance_signature` + `_check_post_quantize_clearances` helpers and a step-12d gate that diffs the clearance state across the quantize mutation (fires only on NEW grazes, tolerates the allowlisted baseline). Raises a dedicated `PostQuantizeClearanceError` that re-propagates past the step's graceful-degrade `except Exception` handler.
- `.github/routed-drc-tolerance.yml`: board-07 floor 29 -> 35 (+6 unmasked marginal `clearance_*` grazes), documented with a #3913 exit clause.
[...]

Issue #479, opened 2026-01-05
net-status via-to-zone detection still not working (#451 incomplete)


rjwalters:
## Description

Despite changelog for v0.7.2 stating that #451 fixed net-status via-to-zone detection, the feature is still not working. After adding 36 stitching vias to connect pads to power/ground planes, `net-status` still reports the same number of unconnected pads.

## Steps to Reproduce

```bash
# Check original PCB
kct net-status board.kicad_pcb --format json | jq '.summary'
# {"total_nets": 23, "complete": 7, "incomplete": 16, "total_unconnected_pads": 80}

# Add 36 stitching vias
kicad-pcb-stitch board.kicad_pcb --net GND --net "+3.3V" --net "+3.3VA" -o board_stitched.kicad_pcb
# Summary: + Added 36 stitching vias

# Check stitched PCB
kct net-status board_stitched.kicad_pcb --format json | jq '.summary'
# {"total_nets": 23, "complete": 7, "incomplete": 16, "total_unconnected_pads": 80}
# ^ SAME as before!
```

## Expected Behavior

After adding 36 stitching vias that connect surface pads to internal plane zones, `net-status` should:
1. Recognize these vias as completing the pad-to-plane connections
2. Reduce the unconnected pad count
3. Increase the net completion percentage

## Actual Behavior

`net-status` reports identical results before and after adding stitching vias:
- Same 80 unconnected pads
- Same 30.4% net completion
- Same 16 incomplete nets

## Impact

- Cannot verify that stitching vias actually worked
- Audit command suggests adding more stitching vias even when they exist
- Users have no feedback that their plane connections are complete

## Related

- #451 was supposed to fix this in v0.7.2
- #441 was the original issue for v0.7.1

## Environment

- kicad-tools v0.7.2
- 4-layer PCB with GND on In1.Cu, +3.3V/+3.3VA on In2.Cu
- macOS Darwin 25.2.0

PR #3947, opened 2026-07-07
fix(connectivity): model pour nets per fill-island, not zone boundary

### LINKED ISSUE #3914: Pour-carried nets: connectivity DRC false-positives AND net-status false 'Complete' — wrong in both directions

**Severity: HIGH (checker trust; extends #3901/#3482, builds on #3905)**

Same sweep, both failure directions:

- **False positive** (boards 03, 04, 06): `kct check` errors `Net 'GND' is partially routed: 15 of 122 pads stranded` while kicad-cli reports 0 unconnected and a shapely copper-union audit confirms one component. Forces CI advisory-filter hacks and report carve-outs. `net-status --why` correctly says POUR_DISCONTINUOUS (#3905 works) but the DRC rule doesn't share that model.
- **False negative** (board 03): `kct net-status` says GND "Complete (33 pads, 100%)" while label-LVS shows 2 real GND opens and kicad-cli 14 unconnected items — it assumes pour connectivity without checking filled copper, so POUR_DISCONTINUOUS can never fire there.
- Board 05: kct LVS counts pour-strandable pads as 88-105 "copper opens" (noise drowning real signals); on the committed board net-status calls +3V3 complete while kicad-cli reports 14 unconnected +3V3 edges.

**Expected:** one shared filled-copper connectivity model (the #3905 classifier) used by the connectivity DRC rule, net-status, and copper-LVS alike: connected-through-pour = connected; pour-discontinuous = incomplete.

**Repro commands in each board's sweep report; simplest: boards/03-usb-joystick fresh build, then `kct check` vs `kicad-cli pcb drc --refill-zones` vs `kct net-status`.**

rjwalters:
## Summary

Three components shared the job of answering "are all pads on this pour net connected?" but used three incompatible copper models, producing false results in **both** directions at once. This PR unifies them on a per-fill-island copper model.

- **False negative** (net-status): a discontinuous pour was reported `Complete`.
- **False positive** (`kct check`): a fully-poured GND net was reported "partially routed" even when kicad-cli reports 0 unconnected.
- **Opens noise** (copper-LVS): 88-105 advisory opens per board drowned real signal opens.

## Changes

- **`analysis/net_status.py`** — `_build_connectivity_graph` no longer bulk-connects every pad inside a zone *boundary* polygon. New `_build_fill_island_groups` groups pads by the poured copper *island* their (hole-aware, thermal-relief-tolerant) copper box actually overlaps, reusing `ConnectivityValidator`'s proven pad-box / fill-region / via-circle geometry. A zone's fill fragments union together (KiCad fragments one pour around thermal reliefs), but separate zones/islands never fuse. A `has_filled_zone` flag is added to `NetStatus`. The legacy boundary bulk-connect is retained only as a shapely-absent fallback.
- **`validate/rules/connectivity.py`** — the rule now skips a net when `has_filled_zone and is_advisory_incomplete` (owns real filled pour copper + incomplete only as a stitching residual). A pour-named net with **no** zone, or a zero-fill zone, still fires.
- **`lvs/copper_lvs.py`** — `compare_partitions` gains `advisory_net_names`; opens are suppressed for those nets (shorts are not). `compare_copper_netlist` derives the set from `build_zone_net_map`.

| Pour net with real fill not reported "partially routed" | ✅ | `test_discontinuous_pour_does_not_fire`; board-06 advisory-connectivity pin 1→0 |
| Discontinuous fill not reported "Complete" | ✅ | `test_discontinuous_fill_not_complete`; `test_partial_fill_zone_only_connects_pads_on_copper` |
| `test_pour_named_net_without_zone_fires` still passes | ✅ | Gate is `has_filled_zone`, not the name heuristic |
| Zero-fill zone still fires | ✅ | `has_filled_zone` requires real filled polygons (existing zero-fill tests green) |
| Copper-LVS suppresses advisory pour opens, keeps signal opens + shorts | ✅ | `test_pour_opens_suppressed`, `test_pour_advisory_filter_does_not_suppress_shorts` |
| Fully-covered pour zone still `Complete` | ✅ | `test_fully_filled_zone_regression`, `test_multi_zone_complete`, `test_zone_no_boundary` |

## Test Plan

- `uv run pytest tests/test_connectivity_rule.py tests/test_analysis_net_status.py tests/test_net_status.py tests/test_copper_lvs.py` — 153 passed (4 new).
- Regression sweep over every suite referencing the changed modules (board 03/04/05/06 copper-LVS, fleet status, LVS recipe, report collector/generator, pipeline, routing baselines, softstart) — green; the single board-06 pin was the targeted false positive and was updated with rationale.
- `ruff check` / `ruff format --check` / `mypy` clean on the changed source.

Scope note: `validate/connectivity.py` (the reference extractor) is not modified — its geometry is reused by composition. Short-detection paths from #3946 are untouched.

Closes #3914

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 3909
-# Branch: feature/issue-3909
+# Issue: 3914
+# Branch: feature/issue-3914
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/src/kicad_tools/analysis/net_status.py
+++ b/src/kicad_tools/analysis/net_status.py
@@ -21,7 +21,7 @@ import math
 from collections import defaultdict
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from kicad_tools.core.layers import COPPER_LAYER_ORDER, via_spans_layer
 
@@ -74,6 +74,7 @@ class NetStatus:
     plane_layers: list[str] = field(default_factory=list)
     has_routing: bool = False
     has_vias: bool = False
+    has_filled_zone: bool = False  # A zone on this net produced fill copper
 
     @property
     def connected_count(self) -> int:
@@ -183,6 +184,7 @@ class NetStatus:
             "unconnected_count": self.unconnected_count,
             "connection_percentage": round(self.connection_percentage, 1),
             "is_plane_net": self.is_plane_net,
+            "has_filled_zone": self.has_filled_zone,
             "is_advisory_incomplete": self.is_advisory_incomplete,
             "plane_layer": self.plane_layer,
             "plane_layers": list(self.plane_layers),
@@ -473,6 +475,13 @@ class NetStatusAnalyzer:
             status.is_plane_net = True
             status.plane_layers = zone_nets[net_number]
             status.plane_layer = zone_nets[net_number][0]
+            # A zone that produced no filled copper (fill disabled, or fully
+            # carved away) provides no connectivity (Issue #3482) and is NOT a
+            # basis for suppressing a connectivity error (Issue #3914): only a
+            # zone with real fill copper makes an incomplete pour net advisory.
+            status.has_filled_zone = any(
+                zone.net_number == net_number and zone.filled_polygons for zone in self.pcb.zones
+            )
 
         # Check for routing
         segments = list(self.pcb.segments_in_net(net_number))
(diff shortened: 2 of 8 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the unified per-fill-island pour-connectivity model. This is a well-engineered PR that fixes both failure directions at once and backs each with a targeted test.

## Correctness — both directions fixed and tested
- **False positive** (pour net wrongly "partially routed"): `ConnectivityRule` now gates on `has_filled_zone and is_advisory_incomplete`. `test_discontinuous_pour_does_not_fire` asserts the precondition (`status == "incomplete"`, `has_filled_zone`, `is_advisory_incomplete`) before checking `error_count == 0`, so it genuinely exercises the suppression path rather than the "complete" fast path.
- **False negative** (discontinuous fill wrongly "Complete"): `_build_fill_island_groups` unions per-zone fill fragments but never fuses separate islands, reusing `ConnectivityValidator`'s hole-aware geometry. Covered by `test_discontinuous_fill_not_complete` and the reworked `test_partial_fill_zone_only_connects_pads_on_copper`.

## Pinned-value changes are legitimate, not masked regressions
- board-06 `EXPECTED_ADVISORY_CONNECTIVITY` 1→0: the prior code comment already documented these as `NetStatusAnalyzer` false positives (the inverse face of #3482). The fix removes exactly that class, and the pin now guards drift *above* 0 (a genuinely-unrouted net), so it keeps its regression-catching value.
- `test_partial_fill_zone_provides_connectivity` → `..._only_connects_pads_on_copper`: the old assertion was *pinning the false-negative bug* (relying on the over-eager #479 boundary heuristic). The flip to `status != "complete"` with C2.2 stranded is the corrected behavior; fixture geometry verified (fill x∈[10,22]; C1.2 on copper, C2.2 off).

## Builder deviation from the curator's literal suggestion — sound judgment
Gating on `has_filled_zone` (REAL fill copper) instead of the curator's name-heuristic path is stricter and correct: a pour-named net with no zone, or a zero-fill zone, still fires. Preserves `test_pour_named_net_without_zone_fires` and the zero-fill edge case. Well-documented inline.

## LVS opens suppression does not reintroduce the false negative
The `advisory_net_names` suppression is curator-sanctioned (AC: <10 opens, down from 88-105) and shorts are still reported (`test_pour_advisory_filter_does_not_suppress_shorts`). Real-open coverage for pour nets now lives in the island-aware `NetStatusAnalyzer`/`ConnectivityRule` (the primary completeness gate), with kicad-cli DRC as ground-truth cross-check. The per-stranded-pad LVS opens it removes were noise, never a trustworthy signal.

## Verification
- CI: all 17 checks green (Test 14m57s, Board 03/04/06 LVS, Type Check, Lint & Format).
- Merge state: CLEAN / MERGEABLE (no conflict with #3934's base.py + board-06 recipe changes; board-06 CI ran green on the merge result).
- Local: `uv run pytest tests/test_connectivity_rule.py tests/test_analysis_net_status.py tests/test_net_status.py tests/test_copper_lvs.py` → **153 passed in 4.42s**.
- PR description uses `Closes #3914`.

Clean, well-scoped, `validate/connectivity.py` reference extractor untouched (reused by composition) as the scope guard required.

PR #3948, opened 2026-07-07
fix(check): auto-load net-class-map sidecar and warn on inactive skew rules (#3917)

### LINKED ISSUE #3917: kct check silently drops net-class rules (10 errors vanish) when --net-class-map sidecar is absent; sidecar is never auto-loaded

**Severity: MEDIUM-HIGH (checker trust / UX)**

- Board 07: `kct check --mfr jlcpcb` = 18 errors without sidecar, 28 with — match-group + diffpair skew/continuity rules silently absent, **no warning** (route-cmd warns loudly in the same situation; check does too on some paths but not this one — behavior is inconsistent across boards/surfaces: board 06's run DID print an INACTIVE warning to stderr).
- Board 03: with sidecar, the 3 diffpair rules run but report **no measured values** (skew/gap) even with `--verbose`; engagement visible only in JSON `rules_checked_by_rule`.
- Board 05: check output *advertises* passing a sidecar (`e.g. output/net_class_map.json`) that the route step never wrote.

**Expected:** (1) auto-load the conventional sidecar path next to the routed PCB (with an info line); (2) if declared classes exist but no sidecar is found, warn consistently on every surface; (3) print measured values (skew mm, min gap mm, continuity %) alongside pass/fail.

**Repro:** run `kct check` on boards 03/06/07 fresh artifacts with and without `--net-class-map`; diff error counts.

rjwalters:
Closes #3917

## Summary

Fixes the three coupled defects behind "kct check silently drops net-class rules (10 errors vanish) when the sidecar is absent":

1. **Defect 1 — route never wrote the sidecar.** `run_post_route_drc()` now serializes the router's `net_class_map` to `<output_dir>/net_class_map.json` via the previously-unused `net_class_map_to_dict()`. Writing in the shared post-route DRC entry covers all three route flows (default multi-layer, rule-relaxation, single-layer) in one place. Empty/`None` maps write nothing; a blocked (read-only) write warns but never fails the route.
2. **Defect 2 — check never auto-loaded it.** `check_cmd.main()` now probes conventional sidecar locations next to the PCB (mirroring the existing schematic auto-discovery) when `--net-class-map` is absent, loads it with an `[INFO] auto-loaded net-class-map sidecar: <path>` line, and short-circuits when an explicit flag is supplied (no double-load). A malformed auto-discovered sidecar warns and falls back to no-sidecar behaviour instead of crashing.
3. **Defect 3 — warning only fired via the CLI.** `DRCChecker` now emits a one-time INACTIVE warning per degraded rule on **every** invocation surface (build pipeline, embedded post-route checks), deduplicated per rule per instance. The CLI keeps its own up-front warning and passes `warn_on_inactive_skew_rules=False` to avoid duplicating it.

Plus the secondary gap (AC5): `--verbose` now surfaces per-pair/per-group measured values (skew mm, coupled fraction) as advisory `info` findings even when the pair/group **passes**, via a new `emit_info` knob on the three rules threaded from `DRCChecker(verbose=...)`.

## Acceptance criteria
- AC1 — sidecar written by route step and round-trips with no data loss
- AC2 — auto-load when present, sidecar-gated rule fires
- AC3 — explicit flag wins, auto-probe skipped
- AC4 — INACTIVE warning on every surface (direct `DRCChecker` too)
- AC5 — measured values visible on a passing check under `--verbose`
- AC6 — regression test reproduces the "errors vanish without sidecar" case: without the sidecar the error count is strictly lower and the delta is attributable to `match_group_length_skew` specifically

## Tests
- Extended `tests/test_cli_check_net_class_map.py`: `TestSidecarAutoLoad`, `TestSidecarWrittenByRoute`, `TestInactiveSkewRuleWarningNonCLI`, `TestVerboseMeasuredValues`, `TestTenVanishingErrorsRegression` — 22 pass (13 new).
- Related suites green: match-group / diff-pair validate + coverage (123 pass), route auto-fix / layer-escalation (137 pass).
- Scope kept to #3917; #3934 (DRC_TOLERANCE) left untouched.

Note: `tests/test_kct_check_parity.py::...test_board_07_gate_counts_diffpair_and_matchgroup` fails on this branch, but it fails **identically with these changes stashed** — a pre-existing board-07 committed-artifact churn issue (#3925), not introduced here.

### DIFF
--- a/src/kicad_tools/cli/check_cmd.py
+++ b/src/kicad_tools/cli/check_cmd.py
@@ -135,6 +135,42 @@ def _find_pcb_file(directory: Path) -> Path | None:
     return None
 
 
+def _discover_net_class_map_sidecar(pcb_path: Path) -> Path | None:
+    """Probe conventional locations for a ``net_class_map.json`` sidecar.
+
+    Issue #3917 Defect 2: ``kct route`` writes a ``net_class_map.json``
+    sidecar next to the routed PCB (in the output directory).  ``kct
+    check`` should auto-load it so the sidecar-gated skew / continuity
+    rules fire without the user having to pass ``--net-class-map`` by
+    hand -- mirroring the existing schematic auto-discovery.
+
+    Candidate locations, in priority order, relative to the resolved
+    PCB path:
+
+    - ``<pcb_dir>/net_class_map.json`` (sidecar written alongside a
+      routed board that lives in its own output directory)
+    - ``<pcb_dir>/output/net_class_map.json`` (board dir with an
+      ``output/`` subtree)
+    - ``<pcb_dir>/../output/net_class_map.json`` (routed PCB inside
+      ``output/`` with the sidecar as a sibling -- redundant with the
+      first candidate but kept for the ``<board>/output/<pcb>`` layout)
+
+    Returns:
+        The first existing candidate path, or ``None`` when no sidecar
+        is found.
+    """
+    pcb_dir = pcb_path.parent
+    candidates = [
+        pcb_dir / "net_class_map.json",
+        pcb_dir / "output" / "net_class_map.json",
+        pcb_dir.parent / "output" / "net_class_map.json",
+    ]
+    for candidate in candidates:
+        if candidate.is_file():
+            return candidate
+    return None
+
+
 def _emit_drift_banner(pcb_path: Path, schematic: str | None) -> None:
     """Print the advisory schematic/PCB drift banner (non-blocking).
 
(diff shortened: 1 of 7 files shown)


rjwalters:
✅ **Approved!**

Thorough review complete. The implementation is correct, well-scoped to #3917, and comprehensively tested. Approving despite a red `Test` gate because the two failures are provably unrelated flaky R-tree perf microbenchmarks (details below) — filed as #3949.

## Correctness / fidelity to acceptance criteria

- **AC1 (sidecar round-trips)** — `run_post_route_drc()` writes `<output_dir>/net_class_map.json` via the previously-orphaned `net_class_map_to_dict()`, in the shared post-route DRC entry so all three route flows are covered in one place. Empty/`None` maps write nothing; a blocked write warns and never fails the route. Schema written by route (`net_class_map_to_dict`) == schema read by check (`net_class_map_from_dict`); `test_sidecar_written_and_round_trips` asserts the round-trip with no data loss.
- **AC2/AC3 (auto-discovery + precedence)** — `_discover_net_class_map_sidecar()` probes the three conventional locations; explicit `--net-class-map` short-circuits the probe (verified `test_explicit_flag_skips_autoprobe` — no auto-load INFO line, no double-load). Existing CI invocations that pass an explicit flag are behaviorally unchanged (the explicit branch is identical to before aside from reworded error text, still `return 1`). All board E2E jobs pass, so the new auto-load introduces no surprise exit-code changes on the committed boards.
- **Malformed-file fallback** — explicit-and-malformed = hard `return 1` (preserved); auto-and-malformed = warn + fall back to no-sidecar (`test_malformed_auto_sidecar_falls_back` asserts `rc != 1`). Correct.
- **AC4 (universal INACTIVE warning)** — `DRCChecker._warn_inactive_skew_rule()` fires on every surface (default `warn_on_inactive_skew_rules=True`), deduped per rule per instance; CLI passes `False` to avoid double-warning. Covered by `TestInactiveSkewRuleWarningNonCLI` (direct-checker, dedup, suppressible).
- **AC5 (measured values under --verbose)** — the new `emit_info` knob (default **False**) is threaded from `DRCChecker(verbose=...)` into the three rules and only adds `info`-severity findings on the *passing* branch. It cannot affect error/violation counts, so there is no semantic interaction with #3933's `MATCHGROUP_VIOLATION_BASELINE`. Verified: 98 skew/continuity validate + CLI tests still green.
- **AC6 (10 vanishing errors)** — `TestTenVanishingErrorsRegression` genuinely reproduces the mechanism: without the sidecar the rule degrades to a no-op (error count strictly lower, `rc == 0`, INACTIVE warned), and the with/without delta is fully attributable to `match_group_length_skew` (`with_errors - without_errors == with_mg`). A synthetic repro rather than board-07's literal count, but it captures the exact failure mode.

## Verification performed

- PR target suite `tests/test_cli_check_net_class_map.py`: **22 passed** (13 new).
- Skew/continuity validate + CLI suites: **98 passed** (no `emit_info` regression, no #3933 conflict).
- Merge state: **CLEAN**.
- Type Check, Lint & Format, all Board 00–07 E2E, Diff-Pair + Match-Group routing regressions, Board 05 routing, Routed PCB DRC, kicad-cli round-trip: **all pass**.

## Builder's pre-existing-failure claim — VERIFIED

`test_kct_check_parity.py::...test_board_07_gate_counts_diffpair_and_matchgroup` fails, but I confirmed it fails **identically** at the branch's parent commit (fc70905a): both produce `35` blocking (baseline expects `120`). This is the #3925 committed-artifact / tolerance-floor churn from #3934 (guard-band tightening) + #3947 (pour connectivity remodel), not introduced here. Note it's also `@pytest.mark.slow`, so it does **not** run in the CI `Test` gate at all.

## CI `Test` gate (red) — flaky perf, not this PR

The `Test` job failed on two reruns of this exact commit, each on a *different* R-tree timing microbenchmark (20705 passed each time):
1. `test_rtree_segment_index.py::...test_rtree_faster_at_200_segments` — `0.15x (rtree=0.3582s, brute=0.0552s)`
2. `test_router_via_rtree_perf.py::...test_rtree_query_faster_than_linear_scan` — `rtree=811ms, linear=437ms, ratio=1.855`

Both are wall-clock ratio assertions on the router's spatial index — untouched by this PR (which only changes DRC validation) and nondeterministic under `-n auto` CI load. Tracked in **#3949** with a suggested fix (mark `slow` or convert to a correctness check). Not a blocker for this PR.

Issue #3920, opened 2026-07-05
Manufacturer-profile split-brain: recipes route for jlcpcb-tier1, build/export/board.json judge at jlcpcb — same PCB scores 0/4/5/88 violations


rjwalters:
Found during the 2026-07-05 fresh-design sweep: all 8 demo boards were rebuilt from scratch in isolated worktrees (fresh `uv sync` + `kct build-native` + wiped `output/`), cross-gated with both `kct check --mfr jlcpcb` and `kicad-cli pcb drc --refill-zones`. Full findings: `.claude/fresh-sweep-findings.md` on the operator machine.

**Severity: HIGH (verdict fragmentation)**

- Board 03: recipe routes `--manufacturer jlcpcb-tier1` (via-in-pad allowed; 4 in-pad vias placed deliberately); `kct build`/`kct export` default to project.kct `target_fab: jlcpcb` -> 4 `via_in_pad` errors, report.md says "[CRITICAL] Fix 5 blocking DRC violations", board.json `status: "partial"` + `drc_violations: 5` — on a tier1-clean, 100%-routed board. Same artifact scores 0 (tier1), 4 (jlcpcb), 5 (report), or 88 (kicad-cli, no sidecar) depending on the surface asked.
- Board 04: four verdicts on one artifact — pipeline gate "0 blocking (allowlist)" PASS / export preflight FAIL (3 errors) / kct check FAILED / board.json "partial" with nets_routed_pct 100 and lvs_clean true. The drill-clearance allowlist exists only inside the build gate.

**Expected:** ONE profile per board, declared once (project.kct), consumed by route/check/export/board-metrics alike; allowlists (grandfathered violations) live in the project spec, not hardcoded in one gate. board.json status should reflect the declared profile's verdict.

**Repro:** board 03 or 04 fresh build; compare `kct check --mfr jlcpcb` vs `--mfr jlcpcb-tier1` vs report.md vs board.json.

rjwalters:
## Curator Enhancement

Curating this issue per user request. Findings are grounded in the 2026-07-05 fresh-sweep findings doc (`.claude/fresh-sweep-findings.md`, committed at 1c6d2e4c) and direct codebase inspection.

---

## Root Cause Analysis

The split-brain has two compounding causes:

**Cause 1: `project.kct` declares the wrong tier.**
[...]

Issue #3949, opened 2026-07-07
Flaky R-tree performance microbenchmarks fail intermittently in CI Test job


rjwalters:
## Problem

The CI `Test` job (`pytest -n auto -m "not slow"`) intermittently fails on R-tree performance microbenchmarks that assert timing ratios. Under CI runner load these ratios are non-deterministic and unrelated to the code under test.

Observed on PR #3948 across two consecutive runs of the same commit (each hitting a *different* perf test, confirming flakiness — 20705 tests passed each time):

1. `tests/test_rtree_segment_index.py::TestRtreePerformance::test_rtree_faster_at_200_segments`
   - `R-tree path was unexpectedly slower: 0.15x (rtree=0.3582s, brute=0.0552s)`
2. `tests/test_router_via_rtree_perf.py::TestViaRtreePerformance::test_rtree_query_faster_than_linear_scan`
   - `rtree=811.16ms, linear=437.20ms, ratio=1.855 (expected <0.5)`

## Impact

These wall-clock timing assertions make the `Test` gate red nondeterministically, blocking otherwise-correct PRs and forcing repeated reruns.

## Suggested fix

- Mark these timing-ratio assertions `@pytest.mark.slow` (so they run only in the nightly slow suite, not the per-PR `Test` gate), or
- Convert them to correctness checks (e.g., assert the R-tree returns the same candidate set as brute force) and drop the wall-clock ratio assertion, or
- Add a generous tolerance / warm-up and skip under known-loaded parallel `-n auto` runs.

Discovered while judging PR #3948 (the failures are unrelated to that PR's DRC-validation changes).

rjwalters:
## Curator Enhancement

### Root Cause Analysis

Both failing tests assert wall-clock timing ratios using a single `time.perf_counter()` measurement pair, making them acutely vulnerable to scheduler noise on shared CI runners:

**`tests/test_rtree_segment_index.py::TestRtreePerformance::test_rtree_faster_at_200_segments`**
- Located at line 519 in `TestRtreePerformance` class
- Times 100 `validate_segment_clearance` calls in back-to-back blocks (R-tree then brute-force)
- Asserts `speedup > 0.8` — a deliberately loose bound that the authors already relaxed from the "2x target" mentioned in the docstring
- Observed failure: `0.15x` speedup — the CI runner was so loaded that the brute-force block ran faster than the R-tree block, likely because the R-tree block ran during a CPU steal event
[...]

PR #3950, opened 2026-07-07
fix(route): emit .kicad_pro/.kicad_dru constraint sidecars before geometric DRC

### LINKED ISSUE #3919: Pipeline never emits .kicad_pro/.kicad_dru constraint sidecars next to routed PCB — kicad-cli cross-checks judge with KiCad defaults

**Severity: HIGH (false verdicts + state-dependence)**

- Board 03: in-route kicad-cli cross-check printed `internal DRC clean but kicad-cli found 88 geometric violation(s): track_width (87), clearance (1)` and withheld PASS. All 87 are false positives — 0.15mm JOY_BTN traces judged against KiCad's default 0.2mm min because no `.kicad_pro`/`.kicad_dru` exists next to the routed file. Restoring the committed sidecars -> 0 errors.
- Board 04: run 1 from a clean dir fails, run 2 passes — because run 1 *writes* routed sidecars during `kct check --drc-only` and run 2's in-route DRC picks them up (2 errors instead of 12). Verdicts depend on invisible leftover state.

**Expected:** whichever step routes for manufacturer profile X writes the matching `.kicad_pro`+`.kicad_dru` (kct fab-rules/mfr_dru.py already generate these) atomically next to the routed PCB, so every downstream kicad-cli invocation — ours or the user's — judges with the intended constraints, deterministically.

**Repro:** board 03 fresh build (watch the 87 track_width FPs); board 04 `rm -rf output/* && python generate_design.py` twice (exit 1 then exit 0).

rjwalters:
Closes #3919

## Problem

`kicad-cli pcb drc` auto-loads `<board>.kicad_pro` from the PCB's directory to read the board design rules. When no project file is present it falls back to KiCad's stricter built-in defaults (0.20 mm track, 0.50 mm via, 0.20 mm clearance), so a board routed at a finer manufacturer capability tier gets **false** geometric violations:

- **Board 03** (`jlcpcb-tier1`, 0.15 mm traces): the in-route cross-check flagged **87 bogus `track_width` errors** and withheld PASS.
- **Board 04**: run 1 from a clean dir failed, run 2 passed — because run 1 *wrote* sidecars as a side effect and run 2 picked them up. Verdicts depended on invisible leftover state.

## Fix

`run_post_route_drc()` in `src/kicad_tools/cli/route_cmd.py` — the shared post-route DRC entry for every route flow — now resolves the manufacturer profile it already has in hand (`get_profile(manufacturer)` + `profile.get_design_rules(layers, copper_oz)`, the same resolution `DRCChecker` uses) and delegates to the existing `write_drc_constraints()` to emit `<board>.kicad_pro` + `<board>.kicad_dru` **atomically before** the `run_geometric_drc()` cross-check. This follows the identical hook point and precedent PR #3948 established for `net_class_map.json`.

- Adds `copper_oz: float = 1.0` to `run_post_route_drc()`, threaded into **both** the internal `DRCChecker` and the emitted sidecars so both engines judge consistently. No callers need updating (default 1.0 is correct for all 8 demo boards).
- Non-fatal by design (mirrors #3948): an unknown manufacturer (`ValueError`) or a read-only output dir (`OSError`) warns and continues — never fails the route.
- Existing `.kicad_pro` is preserved and only the constraint/severity block is overwritten (`write_drc_constraints()` merge behavior).

## Verification (kicad-cli present)

Ran board 03 through the emission path with real kicad-cli:

| | track_width | clearance | total |
|---|---|---|---|
| No sidecar | **87** | 1 | 90 |
| Generated sidecar (this PR) | **0** | 1 | 1 |
| Committed sidecar | 0 | 1 | 1 |

The 87 `track_width` false positives are eliminated, and the generated sidecar matches the committed sidecar exactly. The remaining 1 `clearance` is a real board-geometry item present with **both** sidecars (out of scope for this issue).

Board 04 two-run determinism is now structural: sidecars are always written before geometric DRC, so both runs judge against identical rules.

## Tests

- `tests/test_route_post_drc_geometric.py::TestSidecarEmittedBeforeGeometricDRC` — sidecars land next to the output; **exist at the moment `run_geometric_drc` is invoked** (ordering criterion); `.kicad_pro` carries the profile `clearance_min`/`track_min_width`; `.kicad_dru` references the profile clearance; unknown manufacturer and read-only dir both degrade gracefully; existing `.kicad_pro` merge preserves unrelated keys.
- `tests/test_drc_constraints_export.py::test_route_emits_sibling_kicad_pro` — the route entry point emits a sibling `.kicad_pro` with the correct floors, before kicad-cli runs.

`392 passed` across all suites referencing `run_post_route_drc`. `ruff check` / `ruff format` clean on changed files. The 30 pre-existing `mypy` errors in `route_cmd.py` are unchanged (present on `main`; none in the added lines).

## Coordination

Kept tight to sidecar emission per the coordination note with #3920 (manufacturer-profile split-brain). This PR does **not** touch profile-resolution logic — it emits sidecars from whatever profile the route step already resolved. The board-03 committed sidecar using different values (0.127 vs the tier1 profile's 0.1016) is a symptom of #3920, not addressed here.

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 3914
-# Branch: feature/issue-3914
+# Issue: 3919
+# Branch: feature/issue-3919
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -1584,12 +1584,73 @@ def _write_net_class_map_sidecar(
         print(f"  Net-class-map sidecar: {sidecar_path}")
 
 
+def _write_drc_constraint_sidecars(
+    output_path: Path,
+    manufacturer: str,
+    layers: int,
+    copper_oz: float = 1.0,
+    quiet: bool = False,
+) -> None:
+    """Emit ``.kicad_pro`` + ``.kicad_dru`` next to the routed PCB.
+
+    Issue #3919: ``kicad-cli pcb drc`` auto-loads ``<board>.kicad_pro`` from
+    the PCB's directory to read the board design rules (min track width,
+    clearance, via diameter, etc.).  When no project file is present it falls
+    back to KiCad's stricter built-in defaults (0.20 mm track, 0.50 mm via,
+    0.20 mm clearance), producing *false* geometric violations for boards
+    routed at a finer manufacturer capability tier (e.g. board 03 flagged 87
+    bogus ``track_width`` errors on 0.15 mm traces).  Missing sidecars also
+    make verdicts state-dependent: a prior run that happens to write the
+    sidecars silently changes the next run's DRC result.
+
+    Resolving the manufacturer profile here (the same ``manufacturer`` +
+    ``layers`` + ``copper_oz`` the internal :class:`DRCChecker` resolves)
+    and delegating to :func:`write_drc_constraints` writes both files
+    atomically *before* the ``run_geometric_drc`` cross-check, so kicad-cli
+    -- ours or a later user invocation -- judges against the intended
+    constraints deterministically.  This mirrors the net-class-map sidecar
+    precedent (Issue #3917 / PR #3948): a read-only output directory or an
+    unknown manufacturer degrades to a non-fatal warning, never a route
+    failure.
+
+    Args:
+        output_path: Path to the routed PCB file.  The sidecars are written
+            to the same directory (``<board>.kicad_pro`` / ``.kicad_dru``).
+        manufacturer: Manufacturer profile ID (e.g. ``"jlcpcb-tier1"``).
+        layers: Copper-layer count (threaded into the profile's rules).
+        copper_oz: Copper weight in oz (defaults to 1.0, the system default
+            and the correct value for all 8 demo boards).
+        quiet: If True, suppress the confirmation line.
+    """
+    try:
+        from kicad_tools.manufacturers import get_profile, write_drc_constraints
+
+        profile = get_profile(manufacturer)
+        rules = profile.get_design_rules(layers=layers, copper_oz=copper_oz)
+        written = write_drc_constraints(
+            output_path,
+            rules,
+            manufacturer_id=profile.id,
+            layers=layers,
+            copper_oz=copper_oz,
+        )
+    except (ValueError, OSError, KeyError) as e:
+        # Non-fatal: an unknown manufacturer (ValueError) or a read-only /
+        # blocked output directory (OSError) must not fail the route.
+        if not quiet:
+            print(f"  Warning: could not write DRC-constraint sidecars: {e}")
+        return
+    if not quiet and written:
+        print(f"  DRC-constraint sidecars: {', '.join(str(p) for p in written)}")
+
+
 def run_post_route_drc(
     output_path: Path,
     manufacturer: str,
     layers: int,
     quiet: bool = False,
     net_class_map: dict | None = None,
+    copper_oz: float = 1.0,
 ) -> tuple[int, int]:
     """Run DRC validation on the routed PCB.
 
@@ -1604,6 +1665,11 @@ def run_post_route_drc(
             its engagement state from this map + the routed PCB and
             fires per Epic #2556 Phase 2G.  Without it, that rule is a
             no-op (graceful degradation).
+        copper_oz: Copper weight in oz for the manufacturer profile's
+            design rules (defaults to 1.0, the system default and the
+            correct value for all 8 demo boards).  Threaded into both the
+            internal :class:`DRCChecker` and the emitted ``.kicad_pro`` /
+            ``.kicad_dru`` sidecars so both engines judge consistently.
 
     Returns:
         Tuple of (error_count, warning_count)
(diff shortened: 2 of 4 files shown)


rjwalters:
❌ **Changes Requested — CI Failing (PR-introduced test)**

The implementation is sound and I verified the substance carefully — but the `Test` CI job is **red** because of a new test this PR added. It must be fixed before approval.

## Blocking: `test_readonly_output_dir_is_non_fatal` fails under CI's root user

`tests/test_route_post_drc_geometric.py::TestSidecarEmittedBeforeGeometricDRC::test_readonly_output_dir_is_non_fatal` passes locally but **fails in CI**:

```
AssertionError: assert 'could not write DRC-constraint sidecars' in
'  DRC-constraint sidecars: /tmp/pytest-of-root/pytest-0/popen-gw2/.../ro/board.kicad_pro, ...
   DRC PASSED (jlcpcb-tier1 profile, 4 layers)'
```

Root cause: the test does `os.chmod(ro_dir, stat.S_IRUSR | stat.S_IXUSR)` (mode `0o500`) and expects the write to be blocked. CI runs as **root** (note `pytest-of-root` in the path), and root bypasses directory permission bits — so `write_drc_constraints()` *succeeds*, the non-fatal warning is never printed, and the `in captured` assertion fails.

This is a test-portability bug, not a product bug: the non-fatal `OSError` path in `_write_drc_constraint_sidecars` is correct; the test just can't provoke an `OSError` via chmod when euid == 0.

**Suggested fix (either works):**
- Guard the chmod-based test: `@pytest.mark.skipif(os.geteuid() == 0, reason="chmod perms don't restrict root (CI)")`, or
- Make it environment-independent by simulating the failure instead of relying on filesystem perms — e.g. monkeypatch `route_cmd.write_drc_constraints` (or the imported symbol inside `_write_drc_constraint_sidecars`) to raise `OSError`, then assert the warning and that the route still returns `errors == 0`. This is the more robust option since it also exercises the exact except branch deterministically.

## Note (non-blocking, pre-existing flake)

The same `Test` job also failed `test_placement_slide_off.py::TestPerformance::test_spatial_index_outperforms_bruteforce` (`ratio 1.73x, expected >= 2x`). That's an unrelated timing-sensitive perf assertion on the CI runner, not caused by this PR — but it does keep the job red, so a re-run is worth confirming after the readonly-test fix lands.

## What I verified (all good)

- **Ordering**: `_write_drc_constraint_sidecars()` is called at route_cmd.py:~1697, before `run_geometric_drc()` at :1727, inside the shared `run_post_route_drc()` — so the write-before-DRC guarantee holds on every route flow. Confirmed by the `test_sidecars_present_when_geometric_drc_is_invoked` capture test.
- **Merge preservation is real**: `write_drc_constraints()` reads an existing `.kicad_pro`, calls `merge_project_rules()` (overwrites only the constraint/severity block), and preserves unrelated keys. The `test_existing_kicad_pro_is_preserved_and_merged` sentinel test genuinely exercises this.
- **copper_oz threading**: default `1.0` matches `DRCChecker`'s own default (`validate/checker.py:66`), so adding `copper_oz=copper_oz` to the internal engine is behavior-preserving, and the same value flows into the sidecar — both engines consistent.
- **Sibling #3951 interaction is benign**: no file overlap (that PR touches build_cmd.py/parser.py/commands/build.py; this touches route_cmd.py). Semantically, #3951 only changes *which* `manufacturer` gets resolved and passed into the route step; this PR emits sidecars from whatever profile it receives, so they compose cleanly — #3951 picks the profile, #3919 makes kicad-cli honor it.
- 8/8 new sidecar tests pass locally (non-root).

Fix the readonly test's root-portability and I'll approve.

rjwalters:
Addressed the blocking review feedback.

Reworked `test_readonly_output_dir_is_non_fatal` (renamed to `test_unwritable_output_is_non_fatal`) to inject the failure deterministically: instead of `chmod 0o500` on the output dir (which root bypasses in CI, so the `OSError` branch never fired), the test now monkeypatches `kicad_tools.manufacturers.write_drc_constraints` to raise `OSError`. This exercises the exact non-fatal except branch in `_write_drc_constraint_sidecars` on any euid, and asserts the route still returns `errors == 0`, the warning is printed, and no sidecar is written.

Full `tests/test_route_post_drc_geometric.py` suite: 13 passed. ruff check + format: clean.

The pre-existing perf flake `test_spatial_index_outperforms_bruteforce` (#3949) is untouched, as noted.

rjwalters:
✅ **Approved** (re-review after Doctor pass, head d20c8172)

The prior rejection had a single actionable blocker: `test_readonly_output_dir_is_non_fatal` used `chmod 0o500` to provoke the write failure, which CI's root user bypasses. Verified the rework resolves it correctly:

**Reworked test (`test_unwritable_output_is_non_fatal`)**
- Monkeypatches `write_drc_constraints` on `kicad_tools.manufacturers` to raise `OSError` — correct patch target, since `_write_drc_constraint_sidecars` imports the symbol from that module at call time (route_cmd.py:1626). The injected `OSError` lands in the exact `except (ValueError, OSError, KeyError)` non-fatal branch.
- Assertions genuinely exercise the branch: `errors == 0` (route never fails), no `board.kicad_pro` sidecar left behind, and the "could not write DRC-constraint sidecars" warning is captured. Deterministic on any euid — root-portable as intended.
- Ran the full test file locally on the branch head: 13/13 pass, including the reworked test in isolation.

**Diff scope**
- Only one commit since the previously reviewed head (23a5dd33): d20c8172, touching solely `tests/test_route_post_drc_geometric.py` (+27/−17). Test-only — no implementation changes snuck in. The implementation itself (write-before-DRC ordering, merge preservation, copper_oz consistency, benign #3951 interaction) was already reviewed favorably in the prior pass and is unchanged.

**CI on d20c8172**
- All 16 checks pass, including the **Test** job — the root-user failure is gone, and the known #3949 flake (`test_spatial_index_outperforms_bruteforce`) did not fire this run, so the suite is fully green with no informational carve-outs needed.
- Merge state: CLEAN.

Nice, surgical fix — patching the failure injection at the import site rather than fighting filesystem semantics is the right call.

Issue #3847, opened 2026-06-21
boards: sub-0.5mm drill pairs on boards 04/05/06/07 (real hole_to_hole violations now surfaced by #3842)


rjwalters:
## Original Issue

Surfaced once the dedicated hole_to_hole DRC rule (#3842, child of #3830) started comparing against the 0.5mm fab minimum instead of the copper-clearance floor.

Boards 04/05/06/07 contain genuine drill pairs with edge-to-edge spacing below 0.5mm (measured 0.24-0.49mm) — true manufacturability defects that native kicad-cli flags and kct previously missed. The #3842 PR re-baselined the board-04 and board-06 CI floors to accept these EXISTING violations so the rule could land without turning CI red; this issue tracks actually FIXING the board layouts so the drills meet the 0.5mm minimum, then tightening those floors back toward 0.

Per-board the fix is to nudge the offending vias/PTH pads apart (or merge/relocate stitching vias) in the board recipe and regenerate. Likely medium difficulty per board; board-05 overlaps the existing routing work (#3775/#3766). Curator: assess per-board tractability and split if needed.

Acceptance: boards 04/05/06/07 have no sub-0.5mm drill pairs; the hole_to_hole CI floors are tightened back to 0 (or the true clean baseline).

Related: #3842 (the rule), #3830 (DRC coverage epic), #3829 (reproducibility epic).

---

## Curator Enhancement (tractability assessment + per-board split)

> Curating per orchestrator request. The job here is **tractability scoping**: tell the orchestrator which board drill-spacing fixes are worth attempting now vs deferring. All measurements below were reproduced by replicating the merged `hole_to_hole` rule logic (#3846: `min_hole_to_hole_mm = 0.5`, edge-to-edge = center-distance − r1 − r2, `DRC_TOLERANCE` slack) against each committed `*_routed.kicad_pcb`.

### Measured violations (edge-to-edge, 0.5 mm fab minimum)

| Board | Sub-0.5mm pairs | Worst | Offending drills (net / location) | Source of the via |
|-------|-----------------|-------|-----------------------------------|-------------------|
| **04-stm32-devboard** | **2** | 0.350 mm | `GND↔NRST` @ (26.837, 22.5); `OSC_OUT↔NRST` @ (26.837, 22.0) | **router-emitted** LQFP-48 west escape vias (0.5 mm pitch) |
[...]

PR #3951, opened 2026-07-07
fix(build): resolve one manufacturer profile for every pipeline stage (#3920)

### LINKED ISSUE #3920: Manufacturer-profile split-brain: recipes route for jlcpcb-tier1, build/export/board.json judge at jlcpcb — same PCB scores 0/4/5/88 violations

rjwalters:
## Summary

Fixes the issue #3920 "split-brain" where the *same committed PCB* scored differently at each build surface. Boards 03 (USB joystick) and 04 (STM32 devboard) route with `--manufacturer jlcpcb-tier1` (Tier 1 permits their deliberate in-pad vias), but their `project.kct` declared `target_fab: jlcpcb` (base tier, which forbids via-in-pad). Meanwhile `kct build` only consulted `target_fab` in its **export** step; every other step (route/verify/stitch) used `BuildContext.mfr` initialised from the CLI default `"jlcpcb"`. Result: verify saw 4 spurious `via_in_pad` errors on copper the recipe judged clean (0), export/board.json saw 5.

## Root cause & fix (two compounding causes)

**Part 1 - data.** `boards/03-usb-joystick/project.kct` and `boards/04-stm32-devboard/project.kct`: `target_fab: jlcpcb` -> `jlcpcb-tier1`, aligning the declared tier with the tier the recipes already route against. This also matches the `manufacturers:` block in `.github/routed-drc-tolerance.yml` (already `jlcpcb-tier1`).

**Part 2 - pipeline.** `src/kicad_tools/cli/build_cmd.py` now resolves the manufacturer exactly once via a new `_resolve_effective_mfr()` helper (precedence: explicit `--mfr` > spec `target_fab` > `"jlcpcb"`) and threads it through `BuildContext.mfr`, so route/verify/stitch/export all agree. `_run_step_export` now uses `ctx.mfr` directly instead of re-reading `target_fab` (which would silently override an explicit `--mfr` and re-open the split-brain). The outer build parser's `--mfr` default is `None` with forward-when-not-None dispatch in `commands/build.py`, so `--mfr jlcpcb` is a real override rather than being swallowed as the default.

Out of scope (deliberately untouched): routing/via geometry, `board_metrics_cmd.py` structure, and the `.kicad_pro`/`.kicad_dru` sidecar emission of #3919 (parallel work). Board 04's 2 pre-existing sub-0.5mm drill-clearance violations are a layout issue tracked in #3847, not a profile bug; the tolerance floor (2) is unchanged.

## Before / after (Board 03 committed routed PCB)

| Surface | Before | After |
|---|---|---|
| `kct check --mfr jlcpcb-tier1` | 0 errors | 0 errors |
| `kct check --mfr jlcpcb` (base) | 4 via_in_pad | 4 via_in_pad (override still works) |
| `kct build` resolved profile (no flag) | `jlcpcb` (default) -> verify FAIL | `jlcpcb-tier1` (from target_fab) -> agrees |
| `kct build --mfr jlcpcb` (override) | swallowed -> tier1 | `jlcpcb` honoured end-to-end |

## Tests

- New `tests/test_mfr_profile_consistency.py` (11 tests): resolver precedence; both declaration files (`project.kct` + `routed-drc-tolerance.yml`) name the same profile; committed PCBs meet the CI gate floor at their declared tier; negative control reproducing the 0-vs-4 base-tier disagreement (`via_in_pad`).
- Updated `test_build_export_step` to guard the new single-source-of-truth (export follows `ctx.mfr`, not a stale spec re-read).
- Green: `test_mfr_profile_consistency`, `test_build_export_step`, `test_build_parser_parity`, `test_board_03_regression`, `test_board_04_drc_recipe_mfr`, `test_board_04_mfr_gate`, plus build sync/erc/zones/cmd_errors and manufacturer registry sync (300+ tests). No new mypy errors (3 pre-existing, unchanged). Ruff clean.

Closes #3920

### DIFF
--- a/boards/03-usb-joystick/project.kct
+++ b/boards/03-usb-joystick/project.kct
@@ -77,7 +77,7 @@ requirements:
       height: "60mm"
 
   manufacturing:
-    target_fab: jlcpcb
+    target_fab: jlcpcb-tier1
     layers:
       preferred: 2
     min_trace: "0.15mm"
--- a/boards/04-stm32-devboard/project.kct
+++ b/boards/04-stm32-devboard/project.kct
@@ -78,7 +78,7 @@ requirements:
       height: "40mm"
 
   manufacturing:
-    target_fab: jlcpcb
+    target_fab: jlcpcb-tier1
     layers:
       preferred: 2
     min_trace: "0.15mm"
--- a/src/kicad_tools/cli/build_cmd.py
+++ b/src/kicad_tools/cli/build_cmd.py
@@ -102,6 +102,42 @@ class BuildContext:
         return script.resolve() in self._executed_scripts
 
 
+def _resolve_effective_mfr(
+    cli_mfr: str | None,
+    spec: ProjectSpec | None,
+    default: str = "jlcpcb",
+) -> str:
+    """Resolve the single manufacturer profile the whole build judges against.
+
+    Precedence (highest first):
+
+    1. ``cli_mfr`` -- an explicit ``--mfr`` flag. ``None`` means the flag was
+       not supplied (the argparse default is ``None``), so it does not win.
+    2. ``spec.requirements.manufacturing.target_fab`` -- the project's
+       declared fab tier.
+    3. ``default`` -- the historical ``"jlcpcb"`` fallback.
+
+    This is the fix for the issue #3920 "split-brain": previously only the
+    export step consulted ``target_fab`` while route/verify/stitch used the
+    CLI default ``"jlcpcb"``, so a board whose recipe routed at (e.g.)
+    ``jlcpcb-tier1`` was verified at the base tier and reported spurious
+    ``via_in_pad`` errors. Resolving once here and threading the result
+    through ``BuildContext.mfr`` makes every stage agree.
+    """
+    if cli_mfr is not None:
+        return cli_mfr
+
+    if (
+        spec is not None
+        and spec.requirements is not None
+        and spec.requirements.manufacturing is not None
+        and spec.requirements.manufacturing.target_fab
+    ):
+        return spec.requirements.manufacturing.target_fab
+
+    return default
+
+
 def _find_spec_file(directory: Path) -> Path | None:
     """Find a .kct file in the given directory."""
     kct_files = list(directory.glob("*.kct"))
@@ -2390,15 +2426,13 @@ def _run_step_export(ctx: BuildContext, console: Console) -> BuildResult:
             message="No PCB file found to export",
         )
 
-    # Determine manufacturer: prefer target_fab from spec, fall back to ctx.mfr
+    # Use the single resolved profile. ctx.mfr was resolved once in main()
+    # from (in precedence order) an explicit --mfr flag, the spec's
+    # manufacturing.target_fab, or the "jlcpcb" default -- so every pipeline
+    # step judges against the same manufacturer (issue #3920). Re-reading
+    # target_fab here would silently override an explicit --mfr and re-open
+    # the split-brain, so we deliberately do not.
     mfr = ctx.mfr
-    if (
-        ctx.spec
-        and ctx.spec.requirements
-        and ctx.spec.requirements.manufacturing
-        and ctx.spec.requirements.manufacturing.target_fab
-    ):
-        mfr = ctx.spec.requirements.manufacturing.target_fab
 
     # Determine output directory
     if ctx.output_dir:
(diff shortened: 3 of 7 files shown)


rjwalters:
✅ **Approved!**

Thorough, well-tested fix for the #3920 manufacturer-profile split-brain. Reviewed against the curated acceptance criteria; all key questions check out.

## Correctness / precedence semantics
- `_resolve_effective_mfr` (build_cmd.py:105) cleanly encodes the required precedence: explicit `--mfr` > spec `target_fab` > `"jlcpcb"`. Unit tests pin all three branches (`test_resolve_*`) plus the real board-03 spec.
- Changing the parser default `"jlcpcb"` → `None` does **not** break existing invocations. The only `kct build` call sites in the repo are the docs in `examples/10-complete-kct-workflow/README.md` / `examples/README.md` (`kct build .`, no `--mfr`). No `.github/workflows` or `scripts/` invoke `kct build <dir>` with a manufacturer (only `kct build-native`, unrelated). With no flag, behavior only changes for specs whose `target_fab != jlcpcb` — which is precisely the fix, not a regression.
- `commands/build.py` forward-when-not-None correctly preserves the "explicit override even for `--mfr jlcpcb`" signal.

## Single source of truth verified
- Confirmed no residual `target_fab` re-reads remain in the pipeline: every step — route (`_get_routing_params(ctx.mfr, ...)`, `get_profile(ctx.mfr)`), verify (`--mfr ctx.mfr`), stitch, and export (`mfr = ctx.mfr`) — sources the one resolved value. The old export-only re-read at 2393-2401 is deleted, and `test_build_export_step` now guards against a stale spec re-read (spec="pcbway" but ctx.mfr="oshpark" → export follows ctx.mfr).

## Data fix scoping
- Boards 03/04 `target_fab: jlcpcb → jlcpcb-tier1` matches both the recipes' `--manufacturer jlcpcb-tier1` and the `manufacturers:` block in `.github/routed-drc-tolerance.yml` (test `test_declarations_agree_across_files` enforces lockstep). Scoping to 03/04 is right — they were the only boards declaring the base tier while routing at tier1 (board 05 already declared tier1; 00-02/06-07 genuinely target base jlcpcb). Board 05's build pipeline benefits from the same fix as a side effect (it too now resolves tier1 from spec rather than defaulting to jlcpcb for route/verify/stitch).

## #3950 freshness seam
- #3950 (sidecar emission in `route_cmd.run_post_route_drc`) is already merged to main; this PR is `mergeStateStatus: CLEAN` / `MERGEABLE` against that base. No file overlap (route_cmd.py vs build_cmd.py). Semantically the two converge: `run_post_route_drc` is only reached via the standalone `kct route` path (the recipes call it with `--manufacturer jlcpcb-tier1` explicitly), while the build route step routes inline and never calls it. Both surfaces now resolve **jlcpcb-tier1** — the sidecar sees the same profile this PR establishes. No disagreement.

## Tests / CI
- New `tests/test_mfr_profile_consistency.py` (11 tests) is excellent, especially the negative control `test_board_03_reproduces_split_brain_at_base_tier` proving the `target_fab` declaration is load-bearing (via_in_pad errors at base tier, 0 at tier1).
- Ran locally in the worktree: `tests/test_mfr_profile_consistency.py` + `tests/test_build_export_step.py` → **27 passed**.
- All 16 CI checks pass (Board 00-07 E2E, Lint, Type Check, Test, DRC gate). CI ran against a pre-#3950 base, but since neither file overlaps and merge state is CLEAN post-merge, that run remains valid.

## Non-blocking note
Standalone `kct export` (`export_cmd.py:61`) still defaults to `"jlcpcb"` and was deliberately left untouched. That is a separate entry point (no guaranteed spec context) and outside #3920's build-pipeline scope; not a blocker. Worth a follow-up only if standalone export is ever run inside a board dir expecting spec-aware tiering.

Clean, minimal, correct. Merging is safe.

Issue #3952, opened 2026-07-07
route_cmd: --differential-pairs is a silent no-op on the escape-routing / auto-layers path (blocks board 03 coupled routing)


rjwalters:
## Summary

While fixing #3922 (board 03 lost `--differential-pairs` from its recipe), a deeper limitation surfaced: **restoring the flag does not actually invoke the CoupledPathfinder on board 03**, because the CLI's routing dispatch makes escape routing and differential-pair routing mutually exclusive.

## Root cause (route_cmd.py)

Two independent dispatch bugs mean `--differential-pairs` never reaches the coupled pre-pass on a dense-package board like 03:

1. **Auto-layers escalation path ignores diff-pairs.** When `auto_layers` is on (the default when neither `--layers` nor `--no-auto-layers` is passed — see the `args.auto_layers = True` default around `route_cmd.py:8448`), routing goes through `route_with_layer_escalation` (`route_cmd.py:3113`). That function routes each attempt via `route_with_escape` / `route_all_negotiated` / `route_all_two_phase` and **never references `args.differential_pairs`** — so `route_all_with_diffpairs` is never called and the unconditional `=== Differential Pair Routing ===` banner never prints.

2. **Escape routing takes precedence over the diff-pair branch.** Even on the fixed-layer `do_routing()` path (`--no-auto-layers`), the non-multi-res branch checks `_should_use_escape_routing(...)` first (`route_cmd.py:9307`) and returns early via `route_with_escape`, which ignores diff-pairs. The `elif args.differential_pairs and args.strategy == "negotiated"` branch (`route_cmd.py:9337`) is only reached when escape routing does **not** apply. Board 03 has dense packages (U1 QFP-32, J1 USB-C), so escape routing always wins.

The only path that *does* run the coupled pre-pass is the adaptive multi-resolution `phase2_route_fn` (`route_cmd.py:9182`), and that path routes via `route_all_negotiated` **without** an escape pre-phase. On board 03 that produces 13/13 but a native `kicad-cli` clearance violation (the fine-pitch USB-C escapes need the escape router), so it is not manufacturing-clean.

## Impact

- Board 03: cannot get both escape routing (required for 0 DRC) and diff-pair Phase A. Its recipe therefore keeps `--differential-pairs` present (restored in #3922) but the flag is currently inert on its routing path; the USB D+/D- skew remains coincidental rather than constructed.
[...]

Issue #3921, opened 2026-07-05
CoupledPathfinder converges on 0/9 pairs on the diff-pair regression bench — every pair budget-exits and falls back to uncoupled routing


rjwalters:
Found during the 2026-07-05 fresh-design sweep: all 8 demo boards were rebuilt from scratch in isolated worktrees (fresh `uv sync` + `kct build-native` + wiped `output/`), cross-gated with both `kct check --mfr jlcpcb` and `kicad-cli pcb drc --refill-zones`. Full findings: `.claude/fresh-sweep-findings.md` on the operator machine.

**Severity: HIGH (the diff-pair subsystem's flagship bench never exercises success)**

Board 06 fresh build (seed 42): all 9 pairs exit `CoupledPathfinder.route_coupled` on iteration budget within 0.3-4.5s ("iteration budget exceeded after 1000 iterations; abandoning search"), then route as independent single-ended nets. Result: min same-layer P<->N gap 0.55-0.86mm (vs ~0.375mm coupled), continuity 0-49% vs 70-90% thresholds, 8/9 skew failures — 17 of the artifact's 23 allowlisted DRC errors are the diff-pair constraints the bench declares. "Differential pair nets: 0" in the route summary.

Distinct from #3089 (latency — Phase A is now budget-bounded at ~25s total, a big improvement): this is **convergence**. The Phase 1-3 subsystem's constraints are currently witnessed only through their violation rules firing.

**Also fix the diagnostics** (same code path, diffpair_routing.py:4521-4534): the budget-exit message always formats `per_pair_timeout` ("budget exceeded (120s)") even when the *iteration* budget fired after 0.3-4.5s, and says "1000 iterations" when config `per_pair_max_iterations=2000` (budget silently split per search phase).

**Repro:** `cd boards/06-diffpair-test && python generate_design.py --seed 42`; grep "abandoning search"; check per-pair gap/continuity in the recipe's own compliance table.

rjwalters:
## Curator Enhancement

Curating this issue per user request. Full root-cause analysis below.

---

### Root Cause

**Regression commit: `30d90b09` (PR #3658, 2026-06-13)**
_fix(router): gate coupled near-miss rescue + weighted-A\* search behind enable\_shadow\_construction_

[...]

Issue #3095, opened 2026-05-22
M-D: Board 03 usb-joystick — route remaining 5/16 nets (J1 USB-C escapes) + 9 clearance errors (parent: #2394)


rjwalters:
Parent: #2394 (5-board parity roadmap, milestone M-D)
Blocked by: M-A, M-C (shares DRC-nudge fixes)

## Current state (audit 2026-05-17)

- 48/85 pads connected (56%)
- 11/16 nets done
- 9 DRC errors: 3× clearance_pad_segment, 2× clearance_pad_via, 1× clearance_segment_segment, 1× clearance_segment_via, 2× diffpair_clearance_intra
- 55 DRC warnings

## Scope

J1 is a USB_C_Receptacle_GCT_USB4105 with 25 pads on 0.5mm pitch — 13 currently off-grid (0.05mm offsets). The 9 clearance errors are dominated by `clearance_pad_segment` near U1-9 (microcontroller pin 9).

1. **USB-C escape pattern**: finer router grid (0.05mm) for the J1 escape region.
2. **Anchor recipe** (per [[project_anchor_weight_recipe]]): lock J1 in place, run `--anchor-weight 1.0` for surrounding components.
3. **Re-evaluate JOY_X/JOY_Y multi-pad routes** — these may be contributing to the clearance crowding.
4. Eliminate remaining clearance errors (benefits from M-C DRC-nudge work).
5. Produce manufacturing bundle.

## Acceptance criteria

- 16/16 nets routed
- ≤2 DRC errors (clearance-tolerance allowlist permitted for documented USB-C edge cases)
- Manufacturing bundle produced
- `kct fleet status` reports `ship_ready=true`

## Files involved

- `boards/03-usb-joystick/generate_design.py`
- `src/kicad_tools/router/autorouter.py` (escape grid resolution)
- Footprint alignment if J1 pads need re-gridding

## References

- Parent roadmap: #2394
- Anchor-weight recipe: [[project_anchor_weight_recipe]]
- Latest re-audit: https://github.com/rjwalters/kicad-tools/issues/2394#issuecomment-4499596470

Estimated ~400 LOC.

rjwalters:
## Curation grounding (2026-05-21)

### J1 footprint pad geometry
- Synthesized inline at `boards/03-usb-joystick/generate_design.py:571-622` (`generate_usb_connector()`); pad list at lines 575-592. 16 SMD pads + 2 shield through-holes.
[...]

PR #3953, opened 2026-07-07
fix(boards/03): restore --differential-pairs to routing recipe (#3922)

### LINKED ISSUE #3922: REGRESSION: board 03 silently lost --differential-pairs routing in the #3308/#3410 recipe consolidation; boards/README claim now false

**Severity: HIGH (regression + false documentation)**

`boards/README.md` ("Routing Entry Point Conventions") states board 03 uses the diff-pair-aware entry (PR #3069) and that the footgun is "not active — every existing diff-pair-bearing board is already wired correctly." **Both claims are now false**: the #3308/#3410 consolidation replaced `route_demo.py`'s `route_all_with_diffpairs(enabled=True)` with `generate_design.py:route_pcb()`'s plain `kct route --seed 42 --manufacturer jlcpcb-tier1 --backend cpp --deterministic-budget --timeout 600` — no `--differential-pairs` (generate_design.py:561; also pinned in `tests/router/test_board03_routing_baseline.py:_run_kct_route`). Phase A/B are bypassed; today's perfect USB skew (19.562mm both legs, 0.000mm) is post-hoc luck on an uncongested board, not construction.

**Also:** the README's stated reason board 06 must use the in-process API — custom `intra_pair_clearance` "doesn't survive the JSON-roundtrip the CLI uses" — was verified false in the sweep: `NetClassRouting.to_dict/from_dict` round-trips all five per-class fields losslessly and `kct route --net-class-map` (#2996) consumes them. (What's actually CLI-inexpressible: `per_pair_max_iterations`, `enable_shadow_construction`, the width re-solve.)

**Expected:** restore `--differential-pairs` to board 03's recipe + baseline test; correct both README claims; consider a lint that fails when a board with declared pairs routes without the diff-pair path (the auto-detect decision from #3072 explicitly deferred this — this regression is the predicted cost).

rjwalters:
Closes #3922

## What & why

The #3308/#3410 recipe consolidation silently dropped `--differential-pairs` from `boards/03-usb-joystick/generate_design.py:route_pcb()`. The recipe stopped even *requesting* diff-pair routing, and the `boards/README` "Routing Entry Point Conventions" claim (board 03 uses `--differential-pairs`) went false.

This PR restores the flag and forwards the existing net-class sidecar via `--net-class-map`, so the USB D+/D- `diffpair_partner` / `intra_pair_clearance` (0.15mm, #3095) metadata engages the validate-side diff-pair DRC rules.

## Important scope finding (follow-up: #3952)

Restoring the flag alone does **not** make the CoupledPathfinder run on board 03 — a deeper `route_cmd` limitation that the curator's scoping did not anticipate:

- The board's fine-pitch USB-C (J1) / QFP-32 (U1) force the CLI's **escape-routing dispatch**.
- `route_cmd`'s escape (`route_with_escape`) and auto-layers-escalation (`route_with_layer_escalation`) paths **never consult `args.differential_pairs`**, so `route_all_with_diffpairs` (Phase A) is never called.
- The only path that *does* run Phase A is the escape-less adaptive path. Forcing it with `--no-auto-layers` **does** invoke Phase A (verified: `=== Differential Pair Routing ===`, `Detected 1 differential pairs`, `CoupledPathfinder.route_coupled` budget-exits per #3921) **but drops escape routing and reintroduces a native `kicad-cli` clearance violation** — unacceptable for a manufacturing-clean board.

So the recipe deliberately **keeps `--auto-layers`** to preserve board 03's **13/13 + 0 native DRC**. Making Phase A run *with* escape routing at 0 DRC is a `route_cmd` dispatch fix, filed as **#3952** (out of scope here per the #3921 boundary; `src/kicad_tools/router/` and `route_cmd.py` untouched).

## Changes

- `boards/03-usb-joystick/generate_design.py` — add `--differential-pairs` + `--net-class-map <sidecar>` to `route_pcb()`; thorough comment explaining why `--auto-layers` is kept (the #3952 limitation).
- `tests/router/test_board03_routing_baseline.py` — mirror the flags in `_run_kct_route()` (KEEP IN SYNC contract); add `test_recipe_includes_differential_pairs_flag()` (static regression guard); add `test_coupled_pathfinder_phase_a_invoked()` as an `xfail(strict=False)` that documents #3952 and will flip to XPASS when it lands.
- `boards/README.md` — regression warning → "recipe restored" with an explicit #3952 known-limitation note; board 03 status row + auto-detect footgun note corrected; add #3921/#3952 link refs.
- `tests/test_board_03_regression.py` — comment-only corrections (stop claiming the CLI recipe routes the pair via CoupledPathfinder).

## Verification

- Fresh `python boards/03-usb-joystick/generate_design.py`: **13/13 signal nets, 0 native kicad-cli DRC**.
- `--no-auto-layers` cross-check (not shipped): Phase A invoked but 1 native DRC clearance violation → confirms the #3952 tradeoff.
- `pytest tests/router/test_board03_routing_baseline.py::TestBoard03RoutingBaseline`: **3 passed, 1 xfailed** (42s).
- Non-slow board-03 suites: 15 passed, 1 skipped.
- `ruff check` / `ruff format --check`: clean on all changed files.

Note: committed `output/` artifacts are intentionally not regenerated — the baseline tests re-route into tmpdirs, and the committed board stays at its origin/main 13/13 + 0 DRC state.

### DIFF
--- a/boards/03-usb-joystick/generate_design.py
+++ b/boards/03-usb-joystick/generate_design.py
@@ -581,10 +581,39 @@ def route_pcb(input_path: Path, output_path: Path) -> bool:
         # wall-clock cutoff and pins a fixed node-expansion backstop, so
         # the seed-42 re-route is byte-identical (UUID-normalized) across
         # machines.  --timeout 600 below is then a SAFETY backstop only.
-        # KEEP IN SYNC with tests/router/test_board03_routing_baseline.py.
+        # KEEP IN SYNC with tests/router/test_board03_routing_baseline.py --
+        # this includes --differential-pairs and --net-class-map below.
         "--deterministic-budget",
         "--timeout",
         "600",
+        # Issue #3922 REGRESSION FIX: the #3308/#3410 recipe consolidation
+        # silently dropped ``--differential-pairs``, so the recipe no longer
+        # even *requested* diff-pair routing and the boards/README claim went
+        # false.  Restore the flag (and forward the sidecar via
+        # ``--net-class-map`` so the USB D+/D- ``diffpair_partner`` /
+        # ``intra_pair_clearance`` (0.15mm, #3095) metadata engages the
+        # validate-side diff-pair DRC rules).  The sidecar is written above.
+        #
+        # IMPORTANT (scope note for #3922): on THIS board the flag is
+        # currently inert at routing time.  Board 03 has dense fine-pitch
+        # packages (U1 QFP-32, J1 USB-C), so the CLI takes the escape-routing
+        # dispatch, and route_cmd's escape / auto-layers-escalation paths do
+        # not consult ``args.differential_pairs`` -- the CoupledPathfinder
+        # pre-pass is only wired into the (escape-less) adaptive path.  So we
+        # deliberately keep ``--auto-layers`` (the default): it preserves the
+        # escape pre-phase that board 03 needs for 13/13 + 0 native-DRC.
+        # Forcing the diff-pair path via ``--no-auto-layers`` DOES invoke
+        # Phase A, but loses escape routing and reintroduces a native
+        # ``kicad-cli`` clearance violation -- unacceptable for a
+        # manufacturing-clean board.  Making Phase A run WITH escape routing
+        # is a route_cmd dispatch fix tracked in #3952 (adjacent to the
+        # CoupledPathfinder convergence work in #3921); once it lands, the
+        # xfail in tests/router/test_board03_routing_baseline.py
+        # (test_coupled_pathfinder_phase_a_invoked) will flip to XPASS.
+        # KEEP IN SYNC with tests/router/test_board03_routing_baseline.py.
+        "--differential-pairs",
+        "--net-class-map",
+        str(sidecar_path),
         # Issues #3507/#3454: ``--raw`` (skip TraceOptimizer) was
         # LOAD-BEARING for the 0-DRC acceptance until the grid-staleness
         # fix.  The optimize pass used to replace Route objects without
--- a/boards/README.md
+++ b/boards/README.md
@@ -32,7 +32,7 @@ kct build boards/01-voltage-divider --mfr jlcpcb
 |---|-------|--------|------------|------|-------|
 | 01 | [Voltage Divider](01-voltage-divider/) | ✅ Working | 4 | 3 | Simplest possible design, workflow validation |
 | 02 | [Charlieplex LED](02-charlieplex-led/) | ✅ Working | 14 | 8 | Routes 8/8 signal nets; both DRC engines clean (verified 2026-07-05) |
-| 03 | [USB Joystick](03-usb-joystick/) | ✅ Working | ~20 | 13 | Routes 13/13 on 2-layer in ~24s; diff-pair entry regression tracked in [#3922] |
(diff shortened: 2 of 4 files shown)


rjwalters:
✅ **Approved!**

## Verdict

Recipe restoration + guard test + honestly-documented limitation **legitimately closes #3922**, with #3952 correctly tracking the runtime remainder. This is not an under-delivery.

## The central judgment call (partial-fix framing)

#3922's title is a **recipe/README-truth regression** — the #3308/#3410 consolidation silently dropped `--differential-pairs` and the README's "board 03 uses `--differential-pairs`" claim went false. Both of those are now genuinely fixed:
- `--differential-pairs` + `--net-class-map` restored in `route_pcb()` (generate_design.py:614) and mirrored in `_run_kct_route()` (test file), honoring the KEEP IN SYNC contract in the same commit.
- README's regression warning is replaced with an accurate "recipe restored, runtime coupling gated on #3952" note.

The builder then honestly surfaced a deeper truth the curator's scoping did not anticipate: on board 03 the flag is **inert at routing time**, because `route_cmd`'s escape / auto-layers-escalation dispatch never consults `args.differential_pairs`. Rather than force the diff-pair path with `--no-auto-layers` (which invokes Phase A but drops escape routing and reintroduces a native `kicad-cli` clearance violation — unacceptable for a manufacturing-clean board), the recipe deliberately keeps `--auto-layers` to preserve **13/13 + 0 native DRC**, and files #3952 with a rigorous route_cmd root-cause. That is the correct engineering tradeoff, not a shortcut.

**Is the framing honest / non-overclaiming?** Yes. The README, code comments, and the new `test_coupled_pathfinder_phase_a_invoked` xfail all explicitly state the pair is **not** routed coupled today — no claim that coupling happens. The one #3922 AC not literally met ("route log shows Phase A invoked") is precisely the behavior #3952 now owns; the xfail is executable documentation that flips to XPASS the moment #3952 lands.

## What I verified

- **CI**: all 17 checks green (Board 03 End-to-End LVS-only pass 4m10s; Lint, Type Check, Diff-Pair Regression all pass). Merge state CLEAN.
- **Guard test** `test_recipe_includes_differential_pairs_flag`: passes; asserts both `--differential-pairs` and `--net-class-map` in the recipe source.
- **xfail target**: `=== Differential Pair Routing ===` is printed unconditionally in `route_all_with_diffpairs` (diffpair_routing.py:5810) right after the `enabled` check — so the xfail is a reliable Phase-A tripwire that will XPASS when #3952 wires the escape path.
- **Sidecar fidelity**: the test's `_write_diffpair_sidecar` mirrors `route_pcb()` exactly (same net lists, same `diffpair_partner` / `intra_pair_clearance=0.15` replace calls), so the future XPASS won't be a false signal.
- **#3952** is filed with a precise route_cmd dispatch root cause; scope boundary vs #3921 is clean.

## One non-blocking observation (not requesting changes)

The guard test is a text search over `generate_design.py`, and the string `--differential-pairs` now appears **three** times in that file: twice in the new explanatory comments (lines 585, 590) and once as the actual argument (line 614). If a future consolidation deleted only the real flag on line 614 but left the explanatory comments, `assert "--differential-pairs" in source` would still pass — a small hole in "silently cannot be dropped again." In practice the comments are so tightly bound to the flag that a refactor would likely remove them too, and the curator explicitly endorsed the pragmatic text-search approach. I'm noting it rather than blocking; a stricter version (e.g. asserting it appears as a quoted list element, or counting occurrences) could be a cheap follow-up hardening. Not worth a round-trip here.

Great work turning a scoped recipe-restore into an honest diagnosis of the underlying dispatch bug.

PR #3954, opened 2026-07-07
fix(diffpair): distinguish iteration vs wall-clock exit in coupled budget diagnostic (#3921)

rjwalters:
Refs #3921.

## Summary

Fixes the **diagnostic bug** explicitly called out in #3921 ("Also fix the diagnostics"): the coupled-routing budget-exit `WARNING` hard-coded the `per_pair_timeout` seconds (`"budget exceeded (120s)"`) even when the **iteration** budget fired in a fraction of a second. `CoupledPathfinder.route_coupled` raised the shared `last_timeout_exceeded` flag for *both* the iteration budget and the wall-clock budget, so the caller had no way to tell which one bound.

- Adds a `last_iteration_limited` discriminator on `CoupledPathfinder`, set only on the iteration-budget exit path (reset per call).
- Rewrites the budget-exit diagnostic to surface the real exit reason, the actual iteration count, and the per-phase split, in both the stdout `WARNING` and the structured `logger.warning`.
- Two regression tests (`tests/test_diffpair_shadow.py`).

Before: `WARNING: Coupled routing budget exceeded (120s); ...`
After (iteration exit): `WARNING: Coupled routing iteration budget exceeded (20000 iters; phase cap 20000, total 40000) in 21.9s; ...`
After (wall-clock exit): `... reason=wall-clock iters=86336 wall_budget=60.0s ...`

## The convergence half needs re-scoping (root-cause in the curation comment is incorrect)

The curation comment proposed restoring board-06 coupled convergence (0/9 -> positive) by raising the flag-off iteration budget to a **40000 FLOOR**. I implemented that first and **verified it against the real seed-42 bench** (`boards/06-diffpair-test`, `--step route`). It does not work, and I dropped it. Three-way measurement:

| Configuration | Coupled convergence | best_progress (cells to goal) | Wall time (route phase) |
|---|---|---|---|
| Baseline (classic A*, budget 2000) | **0/9** | 398, 363, 65, 213, 1041, ... | 562s |
| **Floor 40000** (curator Option A) | **0/9** | 398, 361, **64**, 213, 1041, ... (plateau) | **>600s** (timed out) |
| Weighted A* (weight 1.5) + budget 2000 | **0/9** | 398, 361, 61, 213, 1041, ... (plateau) | 569s |
| `enable_shadow_construction=True` | **3-5/9** (shadow constructor + rescue) | n/a (geometric) | >600s (timed out) |

Findings:
1. **The floor cannot fix convergence.** At 20000 iters/phase (20x the budget) best-progress plateaus identically to the 1000-iter run (398->398, 65->64). This matches the code's own note at `diffpair_routing.py:~4164`: classic optimal A* floods `cost_turn` f-plateaus and *"no CI-affordable iteration budget converges."* The floor also balloons wall-time for zero benefit, violating the issue's "must not blow up bench duration" guardrail. So it was **removed**.
2. **The heuristic is not the lever either.** Forcing weighted A* (weight 1.5) on the flag-off search still yields 0/9 with the same plateau — the current board-06 geometry does not admit a joint-state coupled route at reasonable cost.
3. **The historical "6/9" came from the geometric shadow constructor** (`enable_shadow_construction=True`), *not* the joint A* search. That flag is deliberately OFF for board 06 on documented artifact-quality grounds (stranded shadow tails, shadow vias intersecting the partner trace, greedily-claimed corridors stranding nets — see the recipe comment at `generate_design.py:~1828`). Turning it on is the only lever that produces convergence, and it is a deliberate-design reversal that needs artifact/EE review + wall-time work.

**Recommendation:** keep #3921 open (or split): the convergence portion should be re-scoped as a shadow-constructor geometry-quality effort (the #3508 follow-ups the recipe already references), not a budget/heuristic tweak in the pathfinder. This PR ships the correct, in-scope, non-harmful diagnostic fix.

## Scope

- Did NOT touch `boards/03` recipe (parallel #3922).
- Did NOT modify the board-06 recipe or artifact.
- No new mypy errors (file has 9 pre-existing on both main and this branch).

## Tests

- `tests/test_diffpair_shadow.py`, `test_diffpair_per_pair_timeout.py`, `test_diffpair_coupled_floor.py`, `test_router_core_diffpair_timeout.py`: 58 passed.
- Broader diffpair sweep (shadow/npad/integration/phase_b/swap-via/board_06): 173 passed.
- ruff check + format clean.

### DIFF
--- a/src/kicad_tools/router/diffpair_routing.py
+++ b/src/kicad_tools/router/diffpair_routing.py
@@ -618,6 +618,18 @@ class CoupledPathfinder:
         # invocation.
         self.last_timeout_exceeded: bool = False
 
+        # Issue #3921: disambiguates WHICH budget fired when
+        # ``last_timeout_exceeded`` is True.  ``route_coupled`` sets the
+        # shared ``last_timeout_exceeded`` flag for both the iteration
+        # budget (``max_iterations_budget``) and the wall-clock budget
+        # (``timeout_seconds``), so the caller's budget-exit diagnostic
+        # cannot tell a 0.3s iteration bail from a 120s wall-clock bail.
+        # ``True`` means the ITERATION budget was the binding constraint;
+        # ``False`` (with ``last_timeout_exceeded`` True) means the
+        # wall-clock budget fired.  Reset to ``False`` at the start of
+        # every ``route_coupled`` invocation.
+        self.last_iteration_limited: bool = False
+
         # Issue #3473 (review of #3439): number of A* iterations the
         # most recent ``route_coupled`` call consumed.  The two-phase
         # corridor-then-open caller charges the corridor attempt's
@@ -1480,6 +1492,8 @@ class CoupledPathfinder:
         # this immediately after ``route_coupled`` returns ``None`` to
         # decide whether to attempt an independent-routing fallback.
         self.last_timeout_exceeded = False
+        # Issue #3921: reset the iteration-vs-wall-clock discriminator.
+        self.last_iteration_limited = False
         # Issue #3473: reset the iteration counter for this call.
         self.last_iterations = 0
         # Issue #3508: best progress-toward-goal (joint Manhattan
@@ -1659,6 +1673,10 @@ class CoupledPathfinder:
                     n_start.net_name,
                 )
                 self.last_timeout_exceeded = True
+                # Issue #3921: mark the ITERATION budget as the binding
+                # constraint so the caller's diagnostic reports iteration
+                # counts, not a misleading wall-clock-seconds figure.
+                self.last_iteration_limited = True
                 return None
 
             # Issue #3089: periodic wall-clock check.  Exits with ``None``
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Clean, in-scope diagnostic fix for the #3921 "also fix the diagnostics" half. Reviewed against all four key questions:

**1. Discriminator set/reset at all exit paths — correct, no stale-flag leak.**
- Init `False` at `diffpair_routing.py:631`, reset `False` at the top of every `route_coupled` (`:1496`) — this per-call reset is what guarantees a wall-clock exit on pair N+1 cannot inherit a stale `True` from pair N's iteration exit.
- Iteration-budget exit (`:1666`) sets both `last_timeout_exceeded=True` and `last_iteration_limited=True` (`:1675`/`:1679`).
- Wall-clock exit (`:1687`) sets only `last_timeout_exceeded=True` (`:1701`), correctly leaving the discriminator `False`.
- No third exit path sets `last_timeout_exceeded` inside `route_coupled` — verified via grep.

**2. Regression tests are meaningful — they pin the message, not just "a warning printed."**
- `test_flag_off_iteration_exit_diagnostic_reports_iterations` asserts `"iteration budget exceeded"` AND `"20000 iters"` are in stdout.
- `test_flag_off_wallclock_exit_diagnostic_reports_seconds` asserts `"wall-clock budget exceeded"` AND `"120s"`, plus a negative `"iteration budget exceeded" not in out` — that negative assertion is exactly what guards against the original bug regressing.

**3. Scope is diagnostic-only — no behavioral change.** The only non-comment source additions are the discriminator flag, its set/reset, and the rewritten WARNING/`logger.warning` strings. The flag-off floor default (`:4218`) is unchanged and still keys off the existing `COUPLED_FLAGOFF_MAX_ITERATIONS` constant; no budget/iteration value was touched. `total_budget = per_pair_max_iterations` in the new diagnostic only reads the value.

**4. No residue of the reverted iteration-floor change.** The "40000 FLOOR" appears only in explanatory comments (`:4199-4217`, the test docstring, and the PR body documenting the disproof). No floor code, no `max(...)` budget bump.

**CI + merge state**
- All 16 checks pass, including Test, Type Check, Lint & Format, Diff-Pair Routing Regression (boards/06), Board 06 E2E, Match-Group Regression, and Board 05 Routing Regression.
- `mergeStateStatus`: CLEAN (no interaction with #3953 boards/03 recipe, as expected).

**Local verification**
- `tests/test_diffpair_shadow.py`: 26 passed (incl. both new regression tests) in the issue-3921 worktree.

Correctly uses `Refs #3921` (not `Closes`) — the convergence half remains open/blocked for human re-scope per the PR's three-way measurement, which is the right call.

PR #3958, opened 2026-07-07
test(bom): pin degraded-mode enrichment determinism + document contract

### LINKED ISSUE #3935: BOM enrichment is cache-dependent and nondeterministic when the JLCPCB API is unavailable
Split out of #3926 per curator and builder recommendation (see PR #3932, which fixed only the preflight false-warning and explicitly scoped this out).

**Observed (2026-07-05 sweep, board 04):** with the JLCPCB API returning 403 on every call, LCSC enrichment counts drift across *identical* runs (8 from spec/2 unmatched → 9/1 → 10/0 → 11/0) because a persistent cache warms up between runs. Fresh BOMs can also resolve a ref to a different LCSC part than the committed BOM (C12-C15 merged to C1525 where the committed BOM had C49678).

**Why it matters:** BOM content should be a pure function of the design + spec, not cache warmth or API availability. This breaks artifact reproducibility (#3925 family) and can silently change the part that gets ordered.

**Suggested direction:** pin resolved parts in the project spec or a lockfile next to it; treat live-API results as suggestions to be committed, not silent substitutions; make cache state visible in the enrichment log line.

**Affected:** `src/kicad_tools/export/bom_enrich.py`, `src/kicad_tools/parts/cache.py` (per #3926 curator analysis).

Found during the 2026-07-05 fresh-design sweep (#3909–#3926).

rjwalters:
## Summary

BOM LCSC enrichment appeared nondeterministic when the JLCPCB API was
unavailable: successive "identical" runs drifted (8 -> 9 -> 10 matches).
The root cause is **cache warmth**, not per-call nondeterminism — a run
that reaches the live API writes new `enrichment_matches` rows, and a
later API-forbidden run serves those rows as `source="cache"`. A single
`enrich_bom_lcsc` call is already deterministic for a fixed cache state,
because the degraded path only performs pure SQLite reads
(`get_enrichment_match(..., ignore_expiry=True)`).

This PR implements the **minimal fix** (AC1–AC3, AC5): it documents the
degraded-mode determinism contract and adds a regression guard so the
drift cannot silently return. It deliberately does **not** add the
`bom.lock` lockfile (AC4) — that is the larger structural remedy and was
scoped as a follow-on; it did not fall out naturally here.

## Changes

- `src/kicad_tools/export/bom_enrich.py`: document the degraded-mode
  determinism contract in `enrich_bom_lcsc` — identical input + identical
  cache state produces byte-identical `EnrichmentReport`; the reported
  drift is a cache-accumulation effect across sporadically-online runs.
- `tests/test_bom_enrich.py`: add `TestEnrichBomLcscDeterminism` with four
  tests (two 403 runs equal; offline path does not mutate the cache;
  WARNING-level logging; cache bucket surfaced distinct from auto).

| AC1 (deterministic degraded mode) | ✅ | `test_two_offline_runs_produce_identical_reports` asserts `report1 == report2` via `@dataclass` field equality across cache-hit + unmatched groups |
| AC2 (explicit degraded-mode WARNING) | ✅ | `test_offline_run_emits_warning_not_debug` asserts both the 403 circuit-breaker and stale cache-fallback log at `WARNING` |
| AC3 (cache-state surfaced in log) | ✅ | `test_cache_bucket_surfaced_distinct_from_auto` asserts `summary_lines()` shows `"2 from cache"` distinct from `"0 auto-matched"` |
| AC5 (test: two offline runs identical) | ✅ | new regression test pre-seeds cache, calls twice with API mocked as 403, asserts equal reports and stable bucket counts |
| AC6 (no change to live-API path) | ✅ | no production logic changed (docstring only); all pre-existing `TestEnrichBomLcsc*` tests pass unmodified |
| AC4 (lockfile/spec pinning) | ➖ | out of scope for the minimal fix; documented as the structural follow-on |

Supporting guard: `test_offline_run_does_not_mutate_cache` proves the
API-forbidden path performs pure reads (no write-back), which is the
mechanism that keeps successive runs identical.

## Test Plan

- `uv run pytest tests/test_bom_enrich.py` — 24 passed (20 existing + 4 new).
- `uv run ruff check` and `ruff format --check` on both files — clean.
- No real JLCPCB API calls: `PartSuggester` is mocked, cache is a
  `PartsCache` on a `TemporaryDirectory`.
- mypy: the 2 pre-existing errors in `bom_enrich.py` (`find_*_mismatch`
  assignment types, lines 337/345) are unchanged from `origin/main`
  (verified via `git stash`); this PR introduces no new type errors and
  the docstring-only production change touches none of that code.

Closes #3935

### DIFF
--- a/src/kicad_tools/export/bom_enrich.py
+++ b/src/kicad_tools/export/bom_enrich.py
@@ -123,6 +123,31 @@ def enrich_bom_lcsc(
 
     Items that already have an ``lcsc`` value are left untouched.
 
+    Degraded-mode determinism contract (issue #3935)
+    ------------------------------------------------
+    When the JLCPCB API is unavailable it responds with 403 Forbidden.
+    On the first such response this function trips a circuit breaker
+    (``api_forbidden``) and resolves every remaining group from the
+    enrichment cache via :func:`_try_cache_fallback`.  This degraded path
+    is **deterministic for a fixed input and a fixed cache state**: two
+    successive calls with the same ``items`` and the same enrichment
+    cache produce byte-identical :class:`EnrichmentReport` outputs (same
+    entries, sources, ``lcsc_part`` values, and order).  This holds
+    because the fallback only performs pure SQLite reads
+    (``get_enrichment_match(..., ignore_expiry=True)``) and never mutates
+    the cache in the API-forbidden path.
+
+    The *observed* nondeterminism reported in issue #3935 (match counts
+    drifting 8 -> 9 -> 10 across "identical" runs) is not a property of a
+    single call: it arises because the cache **accumulates** matches
+    across runs when the live API is sporadically reachable.  A run that
+    reaches the API writes new ``enrichment_matches`` rows
+    (``source="auto"``); a later API-forbidden run then serves those rows
+    as ``source="cache"``.  The determinism guaranteed here is therefore
+    conditional on the cache state being held constant -- pin the LCSC
+    mapping in the project spec (or a committed lockfile) if you need
+    reproducibility that is independent of local cache warmth.
+
     Args:
         items: List of BOM items to enrich (modified in place).
         prefer_basic: Prefer JLCPCB Basic parts (no extra assembly fee).
--- a/tests/test_bom_enrich.py
+++ b/tests/test_bom_enrich.py
@@ -553,6 +553,191 @@ class TestEnrichBomLcscCacheFallback:
             assert report.cache_matched == 0
 
 
+class TestEnrichBomLcscDeterminism:
+    """Regression tests for degraded-mode (API-forbidden) determinism.
+
+    Issue #3935: BOM enrichment was observed to drift across "identical"
+    runs (8 -> 9 -> 10 matches) when the JLCPCB API was intermittently
+    available. The drift comes from the enrichment cache *accumulating*
+    matches across runs -- not from any nondeterminism within a single
+    call. These tests pin the contract: for a fixed cache state and fixed
+    input, two API-forbidden runs produce byte-identical reports.
+    """
+
+    def _seed_cache(self, cache: PartsCache) -> None:
+        """Populate the enrichment cache with a known set of matches."""
+        cache.put_enrichment_match(
+            "10k",
+            "Resistor_SMD:R_0402_1005Metric",
+            "C25744",
+            confidence=0.85,
+            part_type="Basic",
+        )
+        cache.put_enrichment_match(
+            "100nF",
+            "Capacitor_SMD:C_0402_1005Metric",
+            "C1525",
+            confidence=0.9,
+            part_type="Basic",
+        )
+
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the diff, the production paths it documents, and the four new tests. This is a well-scoped, honest minimal fix.

## Correctness of the contract

I verified the docstring against `bom_enrich.py`:
- `_try_cache_fallback` resolves via `get_enrichment_match(..., ignore_expiry=True)` (line 190) — a pure SQLite read. The `ignore_expiry=True` semantics (stale entries served deterministically) match the docstring's claim.
- The only `put_enrichment_match` write (line 374) lives in the live-API-success branch, which is unreachable once `api_forbidden` is set (line 295 short-circuits to the fallback). So "never mutates the cache in the API-forbidden path" is code-accurate.
- The drift attribution (cache warmth across sporadically-online runs, not per-call nondeterminism) is correct and clearly stated in both the docstring and the test-class docstring.

## Tests genuinely pin the invariant (would catch a regression)

- `test_offline_run_does_not_mutate_cache` is the load-bearing guard: it asserts an uncached `(47uF, C_0805)` group stays `None` after a 403 run. **If a future change added a cache write on the 403 path, this assertion fails** — that is exactly the regression this PR must prevent. Not a trivial pass.
- `test_two_offline_runs_produce_identical_reports` asserts `report1 == report2` (dataclass field-by-field `__eq__`) plus stable bucket counts, exercising both cache-hit and unmatched branches.
- `test_offline_run_emits_warning_not_debug` asserts real log strings ("403 Forbidden", "Cache fallback") at WARNING level — not vacuous.
- `test_cache_bucket_surfaced_distinct_from_auto` pins "2 from cache" vs "0 auto-matched" in the summary.

Local run: `tests/test_bom_enrich.py` → **24 passed** (20 existing + 4 new).

## Honest disposition of #3935

Closing a "nondeterministic" issue with a docs+tests PR is honest here: the issue's determinism problem was real but inter-run/cache-warmth, and the curator explicitly designated AC1–AC3+AC5 as the near-term fix with AC4 (`bom.lock`) as the structural follow-on. Both the PR body and the docstring make the cache-warmth mechanism and the lockfile follow-up explicit. AC table is accurate.

## CI note (non-blocking)

The one red check is `Test` → `tests/test_placement_slide_off.py::TestPerformance::test_spatial_index_outperforms_bruteforce`, a **timing-based flaky perf assertion** (grid=0.047s vs brute=0.081s, ratio 1.72x vs required 2x). Summary: `1 failed, 20763 passed`. This is entirely orthogonal to a docstring-only production change + BOM tests — the PR touches no placement/spatial-index code. `mergeStateStatus` is UNSTABLE from this flaky failure plus one still-pending board-05 job, not BLOCKED/DIRTY.

**Recommendation:** re-run the `Test` job before merge to get it green; the failure is a known-flaky microbenchmark, not a real regression from this PR.

PR #3974, opened 2026-07-08
fix(build): export before verify and stop misreporting manifest gate as DRC failure

rjwalters:
## Summary

The `kct build --step all` pipeline ran `VERIFY` before `EXPORT`. `VERIFY`'s `kct check` meta-check reads `manufacturing/manifest.json`, which `EXPORT` produces — so with the old ordering the manifest sub-check reported `NOT RUN`, the rollup was `INCOMPLETE`, and `kct check` exited 2. `_run_step_verify` then mapped every nonzero exit to `"DRC found issues"`, producing a self-contradicting log line (`DRC: PASSED` alongside `verify: DRC found issues`) and failing boards 00/01 at 13/14. Board 02 escaped only because its recipe built the bundle as a side-effect.

This PR fixes both bugs: reorder EXPORT before VERIFY (root cause), and make the verify message truthful per exit code.

## Changes

- Extract the canonical `--step all` list into a module-level `_ALL_STEPS` constant and reorder `BuildStep.EXPORT` before `BuildStep.VERIFY`, so the manifest exists before VERIFY reads it and VERIFY validates the freshly-exported bundle.
- Add `_classify_verify_result(returncode, stdout)` mapping check exit codes: `0` → "DRC passed", `1` → "Verification tool error", `2` with `INCOMPLETE`/`NOT RUN` in stdout → "Manifest not yet generated (export step not run)", `2` otherwise → "DRC found issues".
- Add `tests/test_build_verify_step.py` (unit + mocked-subprocess integration tests for the mapping) and an assertion in `tests/test_build_export_step.py` that EXPORT precedes VERIFY in `_ALL_STEPS`.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| Board 00 fresh build reaches 14/14 green | ✅ | `kct build boards/00-simple-led` → `Build completed successfully (14/14 steps)`, `verify: DRC passed` |
| Board 01 fresh build reaches 14/14 green | ✅ | `kct build boards/01-voltage-divider` → `14/14`, `verify: DRC passed` |
| Board 02 no regression from reorder | ✅ | `kct build boards/02-charlieplex-led` → `14/14`, `verify: DRC passed` |
| EXPORT appears before VERIFY in the log | ✅ | Build logs show `export` step then `verify` step |
| verify reads "DRC passed" not "DRC found issues" | ✅ | All three boards log `verify: DRC passed (sync OK)` |
[...]

PR #3977, opened 2026-07-08
fix(build): skip route step when recipe already produced a routed PCB

### LINKED ISSUE #3971: kct build re-routes recipe-routed boards, clobbering the recipe's routed artifact (board 04: preflight GND 1/18 pads)
Split from epic #3918 per the 2026-07-08 verification audit (https://github.com/rjwalters/kicad-tools/issues/3918#issuecomment-4910446880). Owner-approved re-scope.

Board 04's recipe internally produces a routed/stitched/exported board; `kct build` then re-runs the generic autorouter on the unrouted PCB, clobbering the recipe's output → preflight fails with GND 1/18 pads. The build pipeline must detect that the recipe step already produced a routed artifact and skip (or verify instead of re-route).

Repro: fresh worktree, wipe output/, `uv run kct build boards/04-stm32-devboard` (fails at 11/12 in ~3m31s).

rjwalters:
## Summary

`kct build` runs a fixed pipeline that includes a ROUTE step. Board 04's `generate_design.py` is a mega-recipe that routes, stitches, and fills zones **in-place** during the PCB step, writing `stm32_devboard_routed.kicad_pcb`. The ROUTE step then re-ran the generic autorouter and **overwrote** the recipe's carefully constructed artifact — collapsing connectivity (GND 1/18 pads) because the generic autorouter lacks the board-04-specific escape/stitch/tie/quantize/zone-fill steps.

The existing mtime-based sibling guard in `_run_step_route()` was unreliable: `generate_design.py` writes the unrouted and routed PCBs within the same filesystem mtime quantum, so the `routed.mtime >= unrouted.mtime` check could not reliably fire.

This implements the curator's recommended **Option A**: a post-PCB artifact scan that records the recipe's routed output on `ctx.routed_pcb_file`, plus an early-return guard in `_run_step_route()` that skips re-routing when that field is already set. This is robust against filesystem mtime granularity.

## Changes

- `src/kicad_tools/cli/build_cmd.py`:
  - **Post-PCB scan** (`main()`, `BuildStep.PCB` arm): after the PCB step records `ctx.pcb_file`, look for a co-located `*_routed.kicad_pcb` sibling and, if present, set `ctx.routed_pcb_file`.
  - **Early-return guard** (top of `_run_step_route()`): when `ctx.routed_pcb_file` is already set and exists (and `--force` is not passed), skip the autorouter and return success with a "recipe-produced routed PCB already present" message.
- `tests/test_build_cmd_errors.py`: new `TestRouteStepRecipeArtifactGuard` class with two tests — the skip fires and never invokes the router; `--force` bypasses the guard and re-routes.

## Semantics decisions

- **`--force`** (`ctx.force`) intentionally bypasses the guard and re-routes (acceptance criterion + existing idiom).
- **Standalone `--step route`** (no prior PCB step in the same run) is unaffected: `ctx.routed_pcb_file` starts `None`, so the guard does not fire and the router runs normally. Verified by a standalone `kct build <unrouted.kicad_pcb> --step route` invocation, which routed the board (guard message did NOT print).

| Fresh board-04 build does not overwrite recipe's routed PCB | ✅ | Fresh `rm -rf output && kct build boards/04-stm32-devboard`: ROUTE step prints "Recipe already produced routed PCB", recipe artifact preserved (9/9 signal nets, DRC PASS, full mfg bundle) |
| ROUTE step prints recipe-skip message | ✅ | `[OK] route: Skipping route: recipe-produced routed PCB already present` |
| `--force` re-routes | ✅ | Unit test `test_force_bypasses_recipe_artifact_guard` (router invoked, no recipe message) |
| Boards without a routed artifact unaffected | ✅ | Standalone `--step route` on an unrouted PCB routes normally (guard silent) |
| No regression in build_cmd suites | ✅ | 102 passed, 1 skipped across build_cmd/route/preflight/verify suites |
| New unit test present | ✅ | `test_route_step_skips_when_recipe_artifact_present` asserts message contains "recipe" and `output_file` == pre-set path |

## Test Plan

- `uv run pytest tests/test_build_cmd_errors.py tests/test_build_preflight_routing_step.py tests/test_build_verify_step.py` — 55 passed.
- Broader: `pytest -k "build_cmd or build_route or build_preflight or build_verify"` — 102 passed, 1 skipped.
- `ruff check` / `ruff format` clean on changed files; `mypy` introduces no new errors (3 pre-existing errors on `main` unchanged).
- Fresh builds: **board 04** → 11/12 (route step preserves recipe artifact; the sole remaining gate is preflight-routing reporting GND 17/18 — the last pad is connected via zone-fill copper the preflight-routing connectivity check does not credit; recipe's own DRC PASSES. This gate is orthogonal to this fix and matches the known #3901 "pour nets invisible" limitation). **Board 00** → 14/14, **board 01** → 14/14 (both recipes also route in-place; guard fires and builds go green — no skip regression).

Closes #3971

### DIFF
--- a/src/kicad_tools/cli/build_cmd.py
+++ b/src/kicad_tools/cli/build_cmd.py
@@ -1281,6 +1281,25 @@ def _resolve_routed_pcb_path(ctx: BuildContext) -> Path:
 
 def _run_step_route(ctx: BuildContext, console: Console) -> BuildResult:
     """Run autorouting step."""
+    # If the PCB step (or a prior run) already left a routed artifact and
+    # recorded it on the context, skip re-routing entirely. This protects
+    # recipe-routed boards (e.g. board 04's generate_design.py, which routes,
+    # stitches, and fills zones during the PCB step) from being clobbered by
+    # the generic autorouter. The post-PCB scan in main() populates
+    # ``ctx.routed_pcb_file`` when it finds a ``*_routed.kicad_pcb`` sibling,
+    # so this guard is robust against filesystem mtime granularity that the
+    # timestamp-based checks below cannot reliably resolve. ``--force`` (which
+    # sets ``ctx.force``) intentionally re-routes and bypasses this guard.
+    if not ctx.force and ctx.routed_pcb_file and ctx.routed_pcb_file.exists():
+        if not ctx.quiet:
+            console.print(f"  Recipe already produced routed PCB: {ctx.routed_pcb_file.name}")
+        return BuildResult(
+            step="route",
+            success=True,
+            message="Skipping route: recipe-produced routed PCB already present",
+            output_file=ctx.routed_pcb_file,
+        )
+
     # Check if a routed PCB already exists (e.g., from generate_design.py)
     # This prevents double-routing when a script already handled routing
     # Skip these checks if force rebuild is requested
@@ -3041,6 +3060,16 @@ def main(argv: list[str] | None = None) -> int:
                 result = _run_step_pcb(ctx, console)
                 if result.output_file:
                     ctx.pcb_file = result.output_file
+                    # If the PCB step's recipe already produced a routed
+                    # artifact (e.g. board 04's generate_design.py routes,
+                    # stitches, and fills zones in-place), record it so the
+                    # ROUTE step's guard skips re-routing instead of clobbering
+                    # it with the generic autorouter. See issue #3971.
+                    expected_routed = result.output_file.with_stem(
+                        result.output_file.stem + "_routed"
+                    )
+                    if expected_routed.exists():
+                        ctx.routed_pcb_file = expected_routed
 
             elif step == BuildStep.SYNC:
                 result = _run_step_sync(ctx, console)
--- a/tests/test_build_cmd_errors.py
+++ b/tests/test_build_cmd_errors.py
@@ -419,3 +419,86 @@ class TestBuildRouteExitCode2:
 
         assert result.success is True
         assert "DRC violations remain" in result.message
+
+
+class TestRouteStepRecipeArtifactGuard:
+    """The ROUTE step must not clobber a recipe-produced routed PCB.
+
+    Board 04's ``generate_design.py`` routes, stitches, and fills zones
+    during the PCB step, writing ``*_routed.kicad_pcb`` in-place. The PCB
+    arm of ``main()`` records that artifact on ``ctx.routed_pcb_file``; the
+    ROUTE step must then skip re-routing instead of overwriting it with the
+    generic autorouter (issue #3971).
+    """
+
+    def test_route_step_skips_when_recipe_artifact_present(self, tmp_path: Path) -> None:
+        """Pre-set ``routed_pcb_file`` short-circuits the route step."""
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Reviewed the don't-clobber guard for recipe-routed boards (closes #3971). This is a clean, minimal, correctly-scoped implementation of the curator's recommended Option A.

## Correctness

- **Post-PCB scan** (`main()`, `BuildStep.PCB` arm) uses a *deterministic single-sibling* path — `result.output_file.with_stem(stem + "_routed")` — **not a glob**. The "multiple `*_routed.kicad_pcb` candidates" concern does not apply; there is exactly one candidate path.
- **Early-return guard** at the top of `_run_step_route()` fires before the router and before the pre-existing mtime-based sibling guard, returning a success `BuildResult` with `output_file == ctx.routed_pcb_file`. It never invokes the autorouter.
- `--force` bypass and standalone `--step route` (field starts `None`) both behave as claimed.

## Stale-artifact hazard (the key review question)

The new guard intentionally drops the `routed.mtime >= unrouted.mtime` check that the older sibling guard (still present, lines ~1310-1320) enforces — that mtime tie is precisely the unreliability Option A was chosen to sidestep. I checked whether this introduces the inverse failure (trusting a stale routed file):

- All 8 in-repo boards route **in-place** during the PCB step via `generate_design.py` (verified board 02's `route_pcb()` + its own comment: *"recipe always rewrites the routed PCB"*). Because the recipe rewrites `*_routed.kicad_pcb` on every PCB-step run, the scan only ever records a freshly-written artifact — a stale routed file cannot survive for these boards.
- CI validates all 8 boards end-to-end, and `--force` is an escape hatch.

The only residual gap is a *hypothetical* external board whose PCB generator does **not** route in-place yet ships a stale committed `*_routed.kicad_pcb`; that board would be silently skipped. Low practical risk (no such board exists in-repo), and I've filed a follow-up rather than blocking.

## Board-04 remaining-gate claim (spot-checked)

Verified architecturally: preflight-routing / verify / export all read `ctx.routed_pcb_file or ctx.pcb_file`, so the GND 17/18 zone-fill-only connectivity gate is **identical whether the route step ran or was skipped** — the guard only avoids overwriting the recipe's routed PCB, it does not change what preflight-routing reads. Matches the known #3901 "pour nets invisible to connectivity" limitation. Board 04 CI passes.

## Tests

- New `TestRouteStepRecipeArtifactGuard` (2 tests): skip fires + router never invoked; `--force` re-routes. Ran locally: **30 passed** across `test_build_cmd_errors.py::TestRouteStepRecipeArtifactGuard`, `test_build_preflight_routing_step.py`, `test_build_verify_step.py`. Full `test_build_cmd_errors.py`: 27 passed.

## CI

All 16 checks pass (Boards 00-07 end-to-end, Board 05 routing regression, diff-pair + match-group regressions, Test, Type Check, Lint, C++ build, Routed-PCB DRC, kicad-cli smoke). `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`. Based on current main (includes #3974).

PR description uses `Closes #3971`.

PR #3982, opened 2026-07-08
fix(build): reject stale *_routed.kicad_pcb via pre-PCB-step mtime capture

### LINKED ISSUE #3978: route-step guard trusts existence-only for *_routed.kicad_pcb (stale-artifact edge case for non-in-place-routing generators)
Discovered while judging PR #3977 (closes #3971).

## Context

PR #3977 added a don't-clobber guard: a post-PCB scan in `main()` records a co-located `*_routed.kicad_pcb` on `ctx.routed_pcb_file`, and an early-return guard atop `_run_step_route()` skips re-routing when that field is set. This is correct and well-tested for the in-repo boards.

## The residual gap

The early-return guard (`build_cmd.py`, top of `_run_step_route()`) fires on **existence alone**:

```python
if not ctx.force and ctx.routed_pcb_file and ctx.routed_pcb_file.exists():
    return BuildResult(step="route", success=True, message="Skipping route: recipe-produced routed PCB already present", ...)
```

It intentionally drops the `routed.mtime >= unrouted.mtime` check that the older sibling guard (still present ~20 lines below) enforces — because that mtime tie is exactly the unreliability Option A was chosen to sidestep for recipe boards that write both PCBs in the same filesystem mtime quantum.

For all 8 in-repo boards this is safe: each `generate_design.py` routes **in-place** during the PCB step and rewrites `*_routed.kicad_pcb` every run, so the recorded artifact is always fresh.

**But** for a hypothetical board whose PCB generator produces only the *unrouted* PCB (routing left to the ROUTE step) yet has a **stale committed `*_routed.kicad_pcb`** in a non-wiped `output/` dir, the post-PCB scan would record the stale file and the guard would silently skip routing — using stale copper instead of re-routing the freshly regenerated unrouted PCB. The old mtime guard specifically protected against this (`routed.mtime < unrouted.mtime` → don't skip).

## Impact

- **Severity**: Low — no in-repo board triggers it (all route in-place; every board's `output/*_routed.kicad_pcb` is git-tracked and rewritten each run).
- **Frequency**: Only external/custom boards whose PCB generator does NOT route in-place.
- **Workaround**: `--force` re-routes.

## Possible hardening

Bound what the post-PCB scan trusts to files the PCB step *actually wrote this run* — e.g. capture the routed sibling's pre-PCB-step mtime/inode and only record it if the PCB step advanced it, rather than existence alone. This preserves Option A's mtime-quantum robustness (compare against the pre-run state, not the sibling unrouted PCB) while rejecting a stale artifact left by a prior partial run.

Filed as a follow-up, not a blocker, during evaluation of PR #3977.

rjwalters:
## Summary

The post-PCB scan in `main()`'s `BuildStep.PCB` arm recorded a `*_routed.kicad_pcb` sibling on `ctx.routed_pcb_file` based on **existence alone**. For an external board whose PCB generator does not route in-place, a stale routed artifact committed in the repo (or left from a prior partial run) would survive the scan, set `ctx.routed_pcb_file`, and cause the ROUTE step's early-return guard to silently skip routing — shipping stale copper (issue #3978).

This implements the curator's recommended **Option A**: snapshot the candidate routed file's `st_mtime` immediately *before* `_run_step_pcb()` runs, then only trust the artifact when the PCB step created it or advanced its mtime.

## Changes

- Add `_capture_routed_artifact_mtime(pcb_file)` — snapshots the `*_routed.kicad_pcb` sibling's mtime before the PCB step (or `None` if absent).
- Add `_scan_recipe_routed_artifact(pcb_output_file, pre_pcb_routed_mtime, console, *, quiet)` — returns the routed path only when it was newly created (`pre == None`) or its mtime advanced; otherwise rejects it with a `[yellow]Stale …[/yellow]` warning and returns `None`.
- Wire both helpers into the `BuildStep.PCB` arm: capture pre-step mtime, then guard the `ctx.routed_pcb_file` assignment through the scan helper.
- Add `TestScanRecipeRoutedArtifact` (6 tests) using `os.utime()` to simulate stale/refreshed artifacts.

**Why this does not reintroduce #3971:** the comparison is `routed.mtime_after > routed.mtime_before` (the file vs. its own pre-run state), not `routed.mtime >= unrouted.mtime`. Any write by an in-place recipe advances the mtime, so all in-repo boards are unaffected. The defense-in-depth mtime-sibling guard inside `_run_step_route()` is untouched.

| Stale generator + stale routed file → ROUTE runs, with warning | ✅ | `test_stale_routed_artifact_is_rejected_with_warning` (asserts `None` + "Stale" in `Console(record=True)` output); `test_stale_artifact_then_route_step_actually_routes` confirms `_run_step_route` invokes the router |
| Non-in-place generator + no routed file → ROUTE runs as before | ✅ | `test_no_routed_artifact_returns_none` |
| Board 04 (in-place recipe): ROUTE still skips | ✅ | Fresh `uv run kct build boards/04-stm32-devboard` → "Skipping route: recipe-produced routed PCB already present", 11/12 steps |
| All 8 in-repo boards unchanged | ✅ | In-place recipes always advance mtime → `pre == None` path trusts artifact exactly as before |
| `--force` still bypasses all guards | ✅ | `test_force_bypasses_recipe_artifact_guard` (unchanged, passes) |
| Standalone `--step route` (`routed_pcb_file` is `None`) routes normally | ✅ | Post-PCB scan only runs in the PCB arm; unchanged |

## Test Plan

- `uv run pytest tests/test_build_cmd_errors.py` → 54 passed (48 existing + 6 new).
- Fresh full build of board 04 (`rm -rf output && uv run kct build boards/04-stm32-devboard`): PCB step routes in-place, post-PCB scan trusts the freshly-written artifact, ROUTE step skips ("Skipping route: recipe-produced routed PCB already present"), build reaches 11/12. Regenerated tracked output artifacts were reverted (out of scope).
- `ruff check` / `ruff format --check` clean on both files; `mypy` introduces no new errors (3 pre-existing errors unchanged).

Closes #3978

### DIFF
--- a/src/kicad_tools/cli/build_cmd.py
+++ b/src/kicad_tools/cli/build_cmd.py
@@ -569,6 +569,77 @@ def _run_step_schematic(ctx: BuildContext, console: Console) -> BuildResult:
     )
 
 
+def _capture_routed_artifact_mtime(pcb_file: Path | None) -> float | None:
+    """Snapshot the ``*_routed.kicad_pcb`` sibling's mtime before the PCB step.
+
+    Returns the ``st_mtime`` of the candidate routed artifact next to
+    ``pcb_file`` if it already exists on disk, otherwise ``None`` (no file
+    to compare against). Called immediately *before* ``_run_step_pcb()`` so
+    the post-PCB scan can tell whether the PCB step actually (re)wrote the
+    routed artifact or merely found a stale copy left in the repo. See
+    :func:`_scan_recipe_routed_artifact` and issue #3978.
+    """
+    if pcb_file is None:
+        return None
+    candidate = pcb_file.with_stem(pcb_file.stem + "_routed")
+    if candidate.exists():
+        return candidate.stat().st_mtime
+    return None
+
+
+def _scan_recipe_routed_artifact(
+    pcb_output_file: Path,
+    pre_pcb_routed_mtime: float | None,
+    console: Console,
+    *,
+    quiet: bool = False,
+) -> Path | None:
+    """Decide whether the PCB step produced a trustworthy routed artifact.
+
+    Board 04's ``generate_design.py`` routes, stitches, and fills zones
+    during the PCB step, writing ``*_routed.kicad_pcb`` in-place. When it
+    does, the ROUTE step's guard must skip re-routing instead of clobbering
+    that artifact with the generic autorouter (issue #3971).
+
+    However, *existence alone* is not proof of freshness: an external board
+    whose PCB generator does **not** route in-place could have a stale
+    ``*_routed.kicad_pcb`` committed in the repo (or left from a prior
+    partial run). Trusting it on existence would silently skip routing and
+    ship stale copper (issue #3978).
+
+    This helper compares the routed artifact's mtime *after* the PCB step
+    against the snapshot taken *before* it (``pre_pcb_routed_mtime``). The
+    artifact is trusted only when it was newly created (no prior file) or
+    its mtime advanced (the PCB step rewrote it). A stale file that survived
+    the PCB step unchanged is rejected with a warning so the ROUTE step
+    re-routes.
+
+    Note this compares the routed file against *its own* pre-run state, not
+    against the unrouted file, so it does not reintroduce the same-mtime
+    quantum race #3971 mitigated.
+
+    Returns the routed :class:`~pathlib.Path` when it should be recorded on
+    ``ctx.routed_pcb_file``, otherwise ``None``.
+    """
+    expected_routed = pcb_output_file.with_stem(pcb_output_file.stem + "_routed")
+    if not expected_routed.exists():
+        return None
+
+    routed_mtime = expected_routed.stat().st_mtime
+    if pre_pcb_routed_mtime is None or routed_mtime > pre_pcb_routed_mtime:
+        # Either the PCB step created the routed artifact, or it advanced an
+        # existing one's mtime — safe to trust it.
+        return expected_routed
+
+    # The file existed before the PCB step and was not updated by it: stale.
+    if not quiet:
+        console.print(
+            f"  [yellow]Stale {expected_routed.name} found (not written by the "
+            f"PCB step); ROUTE step will re-route. Use --force to override.[/yellow]"
+        )
+    return None
+
+
 def _run_step_pcb(ctx: BuildContext, console: Console) -> BuildResult:
     """Run PCB generation step."""
     script = _find_generator_script(ctx.project_dir, "pcb")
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Faithful, minimal implementation of the curator's recommended Option A. The fix correctly reframes the freshness check as a self-comparison (`routed_mtime > pre_pcb_routed_mtime`) rather than the routed-vs-unrouted comparison that #3971 had to bypass, so it closes the stale-copper gap (#3978) without reintroducing the same-mtime quantum race.

## Correctness
- `_capture_routed_artifact_mtime()` snapshots the sibling's mtime *before* `_run_step_pcb()`; `_scan_recipe_routed_artifact()` trusts the artifact only when `pre is None` (newly created) or the mtime advanced. Clean separation, well-documented.
- The `_run_step_route()` early-return guard (`if not ctx.force and ctx.routed_pcb_file...`, line 1562) and the defense-in-depth mtime-sibling fallback are untouched, exactly as the curator scoped. `--force` still bypasses everything.

## Same-quantum edge case (the review question)
The strict `>` is the correct, fail-safe choice. It only bites when a routed file *already existed* before the PCB step (`pre is not None`) AND an in-place recipe rewrites it within the same coarse (1s APFS/HFS+) mtime quantum. That yields a false-stale → an **unnecessary re-route**, never stale copper — the identical known edge the curator documented as accepted and `--force`-mitigated. Fresh CI builds start from `rm -rf output` so `pre is None` and the artifact is always trusted; the green board jobs confirm this. Equality-means-stale is the safe direction here, and the tests pin the intended behavior with 100s mtime gaps.

## Security
Confirmed the diff contains **only** the curator-spec'd mtime approach. No `st_ino`/inode-verification code, no reference to the untrusted `mtime_fix.zip` attachment posted by the non-collaborator. Single clean commit (`17fabf25`). The attachment was not downloaded or opened.

## Tests
- 6 new tests in `TestScanRecipeRoutedArtifact` all pass locally (fresh-trust, refresh-trust, stale-reject+warning, quiet-suppression, no-artifact, end-to-end re-route). `os.utime()` staleness simulation is clean and readable.
- Existing `TestRouteStepRecipeArtifactGuard` (force-bypass, skip-when-present) still pass.
- `ruff check` / `ruff format --check` clean on both files.

## CI
All checks pass, mergeStateStatus **CLEAN**. Critically: every board E2E job is green — Board 04 (in-place recipe that must still skip ROUTE) passes at 2m56s, and Boards 00-03/05/06/07 plus Diff-Pair and Match-Group routing regressions and the full Test suite all pass. No regression for in-place-routing recipes.

PR description uses `Closes #3978`. Solid work.

PR #3983, opened 2026-07-08
feat(route): compose escape routing with the differential-pair pre-pass

### LINKED ISSUE #3952: route_cmd: --differential-pairs is a silent no-op on the escape-routing / auto-layers path (blocks board 03 coupled routing)

rjwalters:
## Summary

`--differential-pairs` was a silent no-op on the escape / auto-layers-escalation
routing paths. On boards whose fine-pitch packages force the escape dispatch
(e.g. board 03's USB-C connector), `route_with_escape` returned before any
diff-pair branch ran, so the CoupledPathfinder pre-pass (Phase A) was never
invoked — the flag was accepted and ignored.

This PR composes escape routing with the differential-pair pre-pass by
delegating to the existing `route_all_with_diffpairs` machinery (which already
flips `paired_escape_coupling=True` and refreshes the `diff_pair_map` before the
escape phase) with the escape main-pass as its `non_diffpair_strategy`, per the
architect's refined Option A.

## Changes

- **core.py**: extract `route_with_escape`'s escape pre-phase (sub-grid pre-pass
  + dense detection + `generate_escape_routes` + pad-channel budgets) into a
  private `_run_escape_prephase()` helper. `route_with_escape` keeps its exact
  behaviour (pre-phase → Phase-2 main pass → summary) — the regression firewall
  for no-pair boards.
- **core.py**: add `route_with_escape_and_diffpairs(diffpair_config, ...)` which
  calls `route_all_with_diffpairs(non_diffpair_strategy=<escape-prephase-then-
  two-phase closure>, coupled_only=True)`. Budget-exit pairs drop back into the
  main pass so no net goes unrouted (the xfail flips WITHOUT needing #3921).
- **core.py**: reset `paired_escape_coupling`/`diff_pair_map` inside the
  non-diff-pair closure before the fall-through escape generation, so a pair
  that budget-exits does not emit paired escapes that perturb a neighbouring
  net's channel (measured: without this, board 03's XTAL1 regressed 13/13 →
  12/13). Committed coupled routes are already grid-marked and are not
  regenerated, so the reset is safe for the coupled-success path too. Plus a
  one-shot `_escapes_generated_this_run` guard (documented idempotency fallback).
- **route_cmd.py**: gate the four escape-dispatch sites (layer / rule / combined
  escalation + `do_routing()`'s escape branch) on `args.differential_pairs` to
  call the new orchestrator; no-pair boards take the byte-identical old path.
  Adds `_build_diffpair_config(args)` so the escalation functions build the same
  config as `do_routing()`.
- **tests**: un-xfail `test_coupled_pathfinder_phase_a_invoked` (now hard-pins
  the Phase A banner); add `test_escape_diffpair_composition.py` (prephase
  extraction, orchestrator delegation, 4-dispatch-site gating, idempotency
  guard, no-pair firewall).

| `test_coupled_pathfinder_phase_a_invoked` flips xfail→XPASS, un-xfailed | ✅ | banner `=== Differential Pair Routing ===` + `Detected 1 differential pairs` now present; marker removed, behaviour hard-pinned |
| Board 03 routes 13/13 with coupled-routing evidence | ✅ | `test_reach_meets_floor` (13/13) + `test_per_net_reach_usb_signals` (0 partial/0 unrouted) green; full class 4/4 in 67s |
| Escape-generation idempotency validated empirically | ✅ | no double-generation (the coupled pre-pass does not call `generate_escape_routes`); the reach hazard was a paired-escape geometry perturbation on the budget-exit path, fixed by resetting the paired map |
| No-pair boards byte-identical behaviour | ✅ | CLI gate means no-pair boards never call the new orchestrator; full fast escape/diffpair/coupled surface (1259 passed) + census/choke-point (102 passed) green |
| #3975 census / choke-point suites stay green | ✅ | `test_fleet_45_census`, `test_segment_45_choke_point`, `test_segment_emission_choke_point_lint` all pass |

## Test Plan

- `tests/router/test_board03_routing_baseline.py::TestBoard03RoutingBaseline` — 4/4 green (xfail flip + 13/13 reach + 0 partial).
- `tests/router/test_escape_diffpair_composition.py` — 8 new fast guards.
- Fast diffpair/coupled/escape surface: 1259 passed, 5 skipped.
- Census + choke-point + board 06/07 fast gates: 102 passed, 1 skipped.
- ruff clean; no new mypy errors (core.py 47→47, route_cmd.py 30→30 baseline unchanged).

## Dependency note

**#3921** (CoupledPathfinder convergence) is NOT a blocker. This issue makes
Phase A *run* (xfail → XPASS); #3921 makes the USB pair *route coupled at 0 DRC*
rather than budget-exiting to the negotiated fall-through. On board 03 the pair
budget-exits and falls through to the identical present-day 13/13 copper, so no
regression.

Closes #3952

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -3614,11 +3614,25 @@ def route_with_layer_escalation(
 
         try:
             if _should_use_escape_routing(router, escape_flag, quiet):
-                router.route_with_escape(
-                    use_negotiated=(args.strategy == "negotiated"),
-                    timeout=_attempt_timeout,
-                    per_net_timeout=getattr(args, "per_net_timeout", None) or None,
-                )
+                # Issue #3952: compose the escape pre-phase with the
+                # CoupledPathfinder diff-pair pre-pass when
+                # --differential-pairs is requested so Phase A runs on
+                # escape-forced boards; otherwise take the unchanged escape
+                # path (byte-identical for no-pair boards).
+                _dp_cfg = _build_diffpair_config(args)
+                if _dp_cfg is not None:
+                    router.route_with_escape_and_diffpairs(
+                        _dp_cfg,
+                        use_negotiated=(args.strategy == "negotiated"),
+                        timeout=_attempt_timeout,
+                        per_net_timeout=getattr(args, "per_net_timeout", None) or None,
+                    )
+                else:
+                    router.route_with_escape(
+                        use_negotiated=(args.strategy == "negotiated"),
+                        timeout=_attempt_timeout,
+                        per_net_timeout=getattr(args, "per_net_timeout", None) or None,
+                    )
             elif getattr(args, "multi_resolution", False):
                 router.route_all_multi_resolution(
                     use_negotiated=(args.strategy == "negotiated"),
@@ -4491,11 +4505,25 @@ def route_with_rule_relaxation(
 
         try:
             if _should_use_escape_routing(router, escape_flag, quiet):
-                router.route_with_escape(
-                    use_negotiated=(args.strategy == "negotiated"),
-                    timeout=_attempt_timeout,
-                    per_net_timeout=getattr(args, "per_net_timeout", None) or None,
-                )
+                # Issue #3952: compose the escape pre-phase with the
+                # CoupledPathfinder diff-pair pre-pass when
+                # --differential-pairs is requested so Phase A runs on
+                # escape-forced boards; otherwise take the unchanged escape
+                # path (byte-identical for no-pair boards).
+                _dp_cfg = _build_diffpair_config(args)
+                if _dp_cfg is not None:
+                    router.route_with_escape_and_diffpairs(
+                        _dp_cfg,
+                        use_negotiated=(args.strategy == "negotiated"),
+                        timeout=_attempt_timeout,
+                        per_net_timeout=getattr(args, "per_net_timeout", None) or None,
+                    )
+                else:
+                    router.route_with_escape(
+                        use_negotiated=(args.strategy == "negotiated"),
+                        timeout=_attempt_timeout,
+                        per_net_timeout=getattr(args, "per_net_timeout", None) or None,
+                    )
             elif getattr(args, "multi_resolution", False):
                 router.route_all_multi_resolution(
                     use_negotiated=(args.strategy == "negotiated"),
@@ -6563,11 +6591,23 @@ def route_with_combined_escalation(
 
             try:
                 if _should_use_escape_routing(router, escape_flag, quiet):
-                    router.route_with_escape(
-                        use_negotiated=(args.strategy == "negotiated"),
-                        timeout=_attempt_timeout,
-                        per_net_timeout=getattr(args, "per_net_timeout", None) or None,
-                    )
+                    # Issue #3952: compose escape + diff-pair pre-pass when
+                    # --differential-pairs is requested (combined 2D
+                    # escalation path); otherwise unchanged escape path.
+                    _dp_cfg = _build_diffpair_config(args)
+                    if _dp_cfg is not None:
+                        router.route_with_escape_and_diffpairs(
+                            _dp_cfg,
+                            use_negotiated=(args.strategy == "negotiated"),
+                            timeout=_attempt_timeout,
+                            per_net_timeout=getattr(args, "per_net_timeout", None) or None,
+                        )
+                    else:
+                        router.route_with_escape(
+                            use_negotiated=(args.strategy == "negotiated"),
+                            timeout=_attempt_timeout,
+                            per_net_timeout=getattr(args, "per_net_timeout", None) or None,
+                        )
                 elif getattr(args, "multi_resolution", False):
                     router.route_all_multi_resolution(
                         use_negotiated=(args.strategy == "negotiated"),
@@ -7013,6 +7053,37 @@ def _resolve_escape_routing_flag(args) -> bool | None:
     if escape:
         return True
     return None
+
+
+def _build_diffpair_config(args):
+    """Build a ``DifferentialPairConfig`` from parsed CLI args, or ``None``.
+
+    Issue #3952: the escape / auto-layers-escalation dispatch sites need a
+    ``diffpair_config`` to compose diff-pair routing with escape routing via
+    ``Autorouter.route_with_escape_and_diffpairs``.  The ``do_routing()``
+    fixed-layer path builds this inline (route_cmd.py); the escalation
+    functions build their own routers and so need their own config.  This
+    helper centralizes the construction so all four dispatch sites stay in
+    sync with the flag surface (``--diffpair-spacing``, ``--diffpair-max-delta``,
+    ``--diffpair-per-pair-timeout``).
+
+    Returns ``None`` when ``--differential-pairs`` was not requested so callers
+    can gate the composed path on a truthy result (the no-pair regression
+    firewall -- boards without pairs take the byte-identical old path).
+    """
+    if not getattr(args, "differential_pairs", False):
+        return None
+
+    from kicad_tools.router import DifferentialPairConfig
+
+    return DifferentialPairConfig(
+        enabled=True,
+        spacing=args.diffpair_spacing,
+        max_length_delta=args.diffpair_max_delta,
+        # Issue #3275: forward the optional per-pair wall-clock budget so the
+        # CoupledPathfinder's per-pair coupled A* search can be bounded.
+        per_pair_timeout=getattr(args, "diffpair_per_pair_timeout", None),
+    )
 
 
 def _should_use_escape_routing(router, escape_flag: bool | None, quiet: bool) -> bool:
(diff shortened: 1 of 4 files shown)


rjwalters:
❌ **Changes Requested — Type Check CI failing (1 NEW mypy error introduced by this PR)**

The design, composition, and firewall are sound and the routing substance is thoroughly validated by CI — but the required **Type Check** gate is red because this PR's own code adds a new mypy error beyond the baseline. I cannot approve while a required check fails.

## Blocker: 1 NEW mypy error

CI (`scripts/ci/check_mypy_baseline.py`) reports:

```
1 NEW mypy error(s) beyond the baseline (.github/mypy-baseline.txt):
new type error: Item "None" of "EscapeRouter | None" has no attribute "diff_pair_map"  [union-attr]
```

Source: the flag-reset you added inside the `_escape_main_pass` closure —

```python
# src/kicad_tools/router/core.py  (~14042-14043)
self.paired_escape_coupling = False
if getattr(self, "_escape_router", None) is not None:
    self._escape_router.diff_pair_map = {}   # <-- union-attr: mypy can't narrow via getattr()
```

The `getattr(self, "_escape_router", None) is not None` idiom does **not** narrow `self._escape_router: EscapeRouter | None` for mypy. There is already exactly one `diff_pair_map` union-attr error in `.github/mypy-baseline.txt` (from the identical pre-existing pattern at `core.py:12768-12769` in `route_all_with_diffpairs`); your new line makes the count 2, which the baseline gate counts as NEW. This is why the PR body's "no new mypy errors (core.py 47→47)" claim didn't hold — the baseline check tightened it to a hard fail.

**Fix (one line):** either
- add a targeted `# type: ignore[union-attr]` on the assignment (matching how the sibling pre-existing occurrence is tolerated in the baseline), or
- narrow directly: `if self._escape_router is not None:` (mypy narrows on direct attribute access; the pre-existing `getattr` form is only in the baseline for historical reasons).

Do **not** add the error to `.github/mypy-baseline.txt` (the gate explicitly forbids that).

## Non-blocking (please fold into the same push)

**`reset_attempt_state()` does not clear `_escapes_generated_this_run`.** `reset_attempt_state` (core.py:2629) documents its contract as "*between escalation attempts, no router state from the prior attempt influences the next*" and exists specifically "to prevent silent regression if future refactors reuse an `Autorouter` across attempts." Your new one-shot guard is per-run mutable state of exactly the kind this method is meant to clear, but it isn't reset there. Today it's masked because each escalation attempt builds a fresh `Autorouter`, and `route_with_escape_and_diffpairs` resets the flag at entry — so this is latent, not a live bug. But if a future refactor honors the #2396 reuse path, a leaked `_escapes_generated_this_run=True` would make the next attempt's `route_with_escape` silently skip escape generation. Add `self._escapes_generated_this_run = False` to `reset_attempt_state()` to keep the contract whole.

## What I verified (all strong — this is a good PR modulo the type error)

- **No-pair firewall holds.** `route_with_escape` is behavior-preserving after the `_run_escape_prephase` extraction: on the no-diffpair path the guard defaults False and is never re-set, so pre-phase → Phase-2 → summary is bit-for-bit unchanged. The four CLI dispatch sites are consistently gated on `args.differential_pairs` / a non-None `_dp_cfg`; the AST guard test (`test_each_composed_call_is_gated`) enforces this. Boards 00/01/02/04/05/06 end-to-end + `Test` (full suite) all pass — no non-pair regression.
- **Re-entry safety of the mid-orchestration flag reset.** Within a single `route_with_escape_and_diffpairs` call the reset is safe (the closure runs only after the coupled pre-pass has committed its paired routes to the grid as fixed infrastructure; the reset only affects the *fall-through* escape generation for remaining pads). Across escalation attempts each attempt builds a fresh `Autorouter`, so the reset cannot leak coupling-on into a context that needs it. Good — the one gap is the `reset_attempt_state` note above.
- **Speedup is not masking skipped work.** Board 03 End-to-End (LVS-only) **passed** (4m27s) and Diff-Pair Routing Regression **passed** (13m39s). Combined with the un-xfailed board-03 baseline (13/13, 0 partial/0 unrouted), the 603s→67s improvement is explained by not emitting paired J1 escapes at intra-pair clearance on the budget-exit path (which was driving expensive rip-up churn), not by dropping nets.
- **xfail→XPASS flip** is done correctly: marker removed, docstring updated, banner + `Detected 1 differential pairs` become hard pins.
- **Design fidelity:** matches the architect's refined Option A (delegate to `route_all_with_diffpairs` with the escape main-pass as `non_diffpair_strategy`, `coupled_only=True`) precisely.

Coverage note (not blocking): the `coupled_only=True` budget-exit drop-back is only exercised indirectly (board-03 e2e's 0-unrouted assertion) — the new fast tests assert it via source-string presence rather than a focused behavioral unit. That's acceptable given a true unit test would need a full coupled route.

Fix the mypy error (and ideally the `reset_attempt_state` line), push, and I'll re-review. CI otherwise looks green across the board fleet.

rjwalters:
Addressed both items from the review (pushed d4fb35ca):

1. **Type Check blocker fixed.** Narrowed `self._escape_router` directly (`if self._escape_router is not None:`) at core.py:14042 instead of the `getattr(...) is not None` form that mypy couldn't use to narrow `EscapeRouter | None`. `scripts/ci/check_mypy_baseline.py` now reports "No new type errors" (1497 errors, all within baseline; the fix actually retires the introduced union-attr, so the count is back to the pre-PR floor). Did not touch `.github/mypy-baseline.txt`.

2. **`reset_attempt_state()` fold-in done.** Added `self._escapes_generated_this_run = False` so the new one-shot guard is cleared between escalation attempts, honoring the #2396 cross-attempt-reuse contract.

**Verification (local):**
- `check_mypy_baseline.py`: OK, no new type errors
- ruff check + format: clean on core.py
- `tests/router/test_escape_diffpair_composition.py`: 8 passed
- `tests/router/test_board03_routing_baseline.py`: 7 passed (4m58s)

Scope is one file, 6 insertions / 1 deletion in `router/core.py`. Ready for re-review.

rjwalters:

The prior rejection had two concrete asks; the Doctor fix (commit `d4fb35ca`, single file `core.py`, +6/-1) addresses exactly both and nothing else:

1. **NEW union-attr error (getattr non-narrowing)** — the escape/diff-pair closure now uses `if self._escape_router is not None:` (direct attribute narrowing) instead of `getattr(self, "_escape_router", None) is not None`. Since `_escape_router: EscapeRouter | None` is declared at core.py:866, mypy narrows correctly. Confirmed the remaining union-attr at core.py:12774 is a **pre-existing** baseline site from PR #3431 (2026-06-09, `git blame`), not introduced by this PR.
2. **`reset_attempt_state` fold-in** — `self._escapes_generated_this_run = False` is now cleared alongside the other per-run mutable state, so a reused `Autorouter` cannot leak the one-shot guard into the next attempt.

**Verification**
- **CI: 16/16 green**, merge state `CLEAN`.
  - **Type Check: pass** (39s) — the key signal; baseline comparison unchanged, NEW error gone.
  - Test (15m59s), C++ Build, Board 00-07, Board 05 Routing Regression (<=11), Diff-Pair (board 06) and Match-Group (board 07) regressions, kicad-cli round-trip, DRC, Lint/Format — all pass.
- **Local (worktree issue-3952 @ d4fb35ca):**
  - `tests/router/test_escape_diffpair_composition.py` — 8 passed.
  - `tests/router/test_board03_routing_baseline.py` — 7 passed (296.8s): xfail→XPASS flip, 13/13 reach, 0 partial/0 unrouted.
- Diff since prior head verified: the two scoped fixes only, confined to `core.py`. No collateral changes.

The prior review substantively endorsed the composition architecture (Option A delegation to `route_all_with_diffpairs`, four-site CLI gating, no-pair firewall, paired-map reset on budget-exit). Those are unchanged. Closes #3952.

Issue #3915, opened 2026-07-05
match_group_length_skew is via-blind: derive_group_skew_data called with board_thickness_mm=None


rjwalters:
Found during the 2026-07-05 fresh-design sweep: all 8 demo boards were rebuilt from scratch in isolated worktrees (fresh `uv sync` + `kct build-native` + wiped `output/`), cross-gated with both `kct check --mfr jlcpcb` and `kicad-cli pcb drc --refill-zones`. Full findings: `.claude/fresh-sweep-findings.md` on the operator machine.

**Severity: HIGH (checker correctness)**

`checker.py:580` calls `derive_group_skew_data(pcb, ncm)` with default `board_thickness_mm=None`, so **vias contribute 0mm** to matched lengths.

Board 07 fresh build: ADDR_BUS (tolerance 0.5mm) "passes" at 0.0019mm skew, while via-inclusive skew (1.6mm board, 4L) is **1.600mm = 3.2x over tolerance** — A4/A6 each carry an extra via pair vs their group. A via-count mismatch inside a "matched" bus is invisible to the rule.

**Expected:** thread the real board thickness (available from the stackup) into the checker call; report via-inclusive skew; optionally flag via-count mismatches within a group.

**Repro:** fresh build boards/07-matchgroup-test; `kct check --net-class-map output/net_class_map.json`; compare against via-inclusive lengths (sweep agent's measurement script in report).

rjwalters:
## Curator Enhancement

### Root-Cause Confirmation

The filed location has drifted by 2 lines since the sweep; the live call is at **`src/kicad_tools/validate/checker.py:580`** (confirmed on `origin/main`):

```python
group_skew_data, tracker_match_groups, threshold_map = derive_group_skew_data(
    self.pcb, self.net_class_map          # board_thickness_mm omitted → defaults to None
)
```

`derive_group_skew_data` signature (`match_group_skew.py:77-82`):

```python
def derive_group_skew_data(
    pcb: PCB,
    net_class_map: dict[str, NetClassRouting] | None,
    board_thickness_mm: float | None = None,   # ← not threaded in
    num_copper_layers: int = 2,                # ← not threaded in
) -> tuple[...]:
```

[...]

PR #3956, opened 2026-07-07
perf(router): skip Python fallback on resume-exhaustion and dedup escalation rungs

rjwalters:
## Summary

Two independent, compounding defects made structurally-identical escalation rungs and repeatedly re-exhausted resume-loop dead-ends dominate routing wall time. On board-07 the 2026-07-05 sweep measured Phase-2 channel routing at 933 of 1052 pipeline seconds. This PR adds a **narrowed** resume-exhaustion fast-path in the C++ backend fallback and pre-rung deduplication in the layer-escalation ladder.

**The fast-path is deliberately narrow** (this is the key correctness property, per the PR #3956 review). It fires only on the **second-and-later** clearance-exhaustion (`"...exhausted N resume attempts"`) of the **same net** — the negotiated rip-up loop re-presenting a clearance obstruction the fresh Python A* already failed once. It **always falls back** on:

- the **first** clearance-exhaustion of every net (the fresh full Python A* measurably rescues real pads — board-07 GND pad `U1.24`),
- **every** case-2 `"resume after rejected goal cell failed"` (path-finding failure on a distorted open set — the fresh A* rescues USB-joystick nets, 5 -> 7), and
- the untouched initial-search `FAILURE_NO_PATH` (`"no path (C++ A* open set exhausted)"`, no resume marker) — single-corridor geometries the Python 45-degree/waypoint expansion legitimately rescues.

An earlier, broader variant (skip case-2 + skip case-1-first) was rejected in review because it regressed both the USB-joystick 6/16 net floor and the board-07 GND pour audit. The narrowing keeps the genuine win (killing the repeated re-exhaustion grind) without those regressions.

**Honest perf (board-07, single-run wall-clock, PYTHONHASHSEED=42, seed 42, C++ backend built):**

| | wall time | pour connectivity | nets routed / DRC |
|---|---|---|---|
| Baseline (pre-#3923 full grind, `KICAD_ROUTER_SKIP_RESUME_FALLBACK=0`) | **~805s** | PASS | 27/31 |
| Narrowed fix (guard on) | **~688-762s** | **PASS** (U1.24 reconnected) | 26/31, DRC 28/35 |

[...]

PR #3985, opened 2026-07-08
fix(router): make match-group length tuner via-length-aware (#3931)

### LINKED ISSUE #3931: Board 07 ADDR_BUS carries a via-count mismatch (A4/A6 extra via pair) — via-imbalance now visible since #3915

rjwalters:
## Summary

Closes #3931 (re-scoped 2026-07-08 to **router-side via-length-aware match-group tuning**, Option (a)).

Board 07's ADDR_BUS exposed a class of defect: some match-group members escape to an inner layer behind a full-stack via while others route flat on the surface. Their **copper** lengths are matched to within microns, but the ~1.6 mm of via drilled length pushes the group past its 0.5 mm skew tolerance. The DRC checker measures this via-inclusive skew (since #3915), but the length **tuner** measured planar-only copper -- it saw the members as already matched and did nothing, so the violation persisted. The sweep builder proved that board-level pin reorders only relocate the via imbalance (escape-order, not net identity), so the fix must be router-side.

## What changed

- **`match_group_tuning.py`** -- `tune_match_group_v2` and `_tune_match_group_single_ended` gain `board_thickness_mm` / `num_copper_layers` params. The planar-only `LengthTracker.calculate_route_length` member-length computation (and every per-attempt length/skew recomputation) is replaced with `MatchGroupTracker._measure_route_total`, making member lengths **via-inclusive**. `board_thickness_mm=None` (default) -> vias contribute 0.0 -> byte-for-byte legacy behavior for callers without a stackup.
- **`core.py`** -- `apply_match_group_tuning` threads `board_thickness_mm` from the manufacturer profile's `DesignRules` (the same value the DRC checker measures with -- the router's own `rules.DesignRules` has no thickness field) plus `num_copper_layers` into the tuner and both `record_routes` brackets.

The via-carrying members become the de-facto reference; the copper-only members receive F.Cu meander (via the existing 45-degree-compliant serpentine engine, #3975) to compensate the drilled length. No board-07 PCB artifact is regenerated -- this is a tuner-only, router-side fix (artifact refresh is #3925).

## Verification on the real gate

`PYTHONHASHSEED=42 uv run python scripts/ci/check_matchgroup_coverage.py boards/07-matchgroup-test --seed 42` (C++ backend):

| Metric | Before | After |
|---|---|---|
| ADDR_BUS via-inclusive skew (reroute) | 19.220 mm | **0.000 mm** (7 tuned, 1 ref) |
| `match_group_length_skew` violations (reroute) | 2 (ADDR_BUS + MIPI) | **1** (MIPI only) |
| DRC error count (reroute) | 29 | **26** (< 35 floor) |

MIPI_CSI_LANES's pair-only skew (tracked under #3916) is the sole surviving violation.

## Baseline reconciliation (single value, both paths)

`MATCHGROUP_VIOLATION_BASELINE` 2 -> **1**:
- **Reroute path**: 1 = MIPI (ADDR_BUS now compensated by the tuner).
- **`--skip-route` committed-artifact path**: 1 = ADDR_BUS (the router fix does not regenerate the committed artifact; refresh is #3925).

`1 <= 1` passes on both. The `.github/routed-drc-tolerance.yml` board-07 floor (35) is a max ceiling; the tuner fix only removes a violation, so it is preserved as a safe upper bound.

## Tests

- `tests/test_match_group_tuning.py::TestViaCountImbalance` -- synthetic via-free vs via-carrying members; confirms F.Cu meander is added when a stackup is supplied, no-op when `board_thickness_mm=None`, and generalizes to mixed via-counts (2 free + 2 via).
- `tests/test_apply_match_group_tuning.py::TestViaInclusiveStackupThreading` -- orchestrator threads `board_thickness_mm`/`num_copper_layers` into the tuner; end-to-end via-imbalance correction through `apply_match_group_tuning`.
- Baseline test in `test_check_matchgroup_coverage.py` updated 2 -> 1 with both paths documented.
- Full match-group / board-07 suites green (131 passed in the coverage + tuning suites).

Two pre-existing failures (`test_kct_check_parity.py::...gate_counts_diffpair_and_matchgroup` expecting 120 vs 35, and the `--skip-route` DRC-count 42-vs-35) are stale-committed-artifact drift on `origin/main` (#3925), unrelated to this change -- confirmed by `git stash` on the baseline.

### DIFF
--- a/.github/routed-drc-tolerance.yml
+++ b/.github/routed-drc-tolerance.yml
@@ -1643,17 +1643,29 @@ tolerances:
   # exposes a PRE-EXISTING via-count mismatch on board 07's ADDR_BUS match
   # group (A4/A6 carry an extra via pair vs A0, ~1.6 mm of via-transition
   # length beyond the 0.5 mm tolerance).  The from-scratch re-route (seed 42)
-  # that the Match-Group Routing Regression gate runs therefore reports one
+  # that the Match-Group Routing Regression gate runs therefore reported one
   # NEW ``match_group_length_skew`` error, moving the binding re-route count
-  # 28 -> 29.  This +1 is PR-caused (via-aware detection), not route variance,
-  # so it is reconciled here rather than dismissed.  The committed-artifact
-  # gate (check_routed_drc.py, advisories filtered) measures 22 well under 29,
-  # so this bump does not loosen that gate.  Companion change: board 07's
-  # routed PCB is added to ``MATCHGROUP_VIOLATION_BASELINE = 1`` in
-  # scripts/ci/check_matchgroup_coverage.py, same #3931 reference.
-  # Exit clause: a via-balanced re-route (or artifact refresh, blocked on the
-  # artifact-churn problem #3925) drops the skew error, at which point restore
-  # 29 -> 28 and the MATCHGROUP baseline back to strict 0.  See #3931.
+  # 28 -> 29.
+  #
+  # !!! #3931 RESOLUTION (router-side via-length-aware tuner) !!!
+  # ----------------------------------------------------------------------
+  # #3931 makes the match-group length tuner via-aware: during the seed-42
+  # re-route the tuner now adds F.Cu meander to ADDR_BUS's via-free members to
+  # compensate the A4/A6 drilled-via length, so the ADDR_BUS via-inclusive skew
+  # converges to within its 0.5 mm tolerance and the ``match_group_length_skew``
+  # error it used to produce is REMOVED from the re-route path.  The reroute
+  # ``MATCHGROUP_VIOLATION_BASELINE`` drops 2 -> 1 (only MIPI_CSI_LANES's
+  # pair-only skew, #3916, remains) in
+  # scripts/ci/check_matchgroup_coverage.py.  This DRC-error floor is a MAXIMUM
+  # (allowed-count ceiling): the tuner fix only REMOVES a violation from the
+  # re-route path, never adds one, so the 35 ceiling is preserved as a safe
+  # upper bound and continues to cover the committed-artifact gate (which the
+  # #3931 router change does not regenerate; the committed artifact still
+  # carries the pre-fix ADDR_BUS and its own separately-tracked marginal-
+  # clearance count).  Exit clause: once the artifact-churn work (#3925)
+  # refreshes the committed board 07 so the via-balanced ADDR_BUS and the
+  # marginal clearances both clear, ratchet 35 down to the new committed count.
+  # See #3931 / #3925.
   #
   # !!! 29 -> 35 (Issue #3913, DRC_TOLERANCE 0.005 -> 1e-4 mm) !!!
   # ----------------------------------------------------------------------
--- a/scripts/ci/check_matchgroup_coverage.py
+++ b/scripts/ci/check_matchgroup_coverage.py
@@ -109,53 +109,64 @@ MATCHGROUP_RULE_IDS: tuple[str, ...] = ("match_group_length_skew",)
 # error-count allowlist floor to absorb new skew errors.  Keyed by
 # repo-relative routed-PCB path.
 #
-# Issue #3931 (via-imbalance now visible since #3915) -- board 07's ADDR_BUS
-# match group carries a via-count mismatch: the A4/A6 members route with an
-# extra via pair vs A0, adding ~1.6 mm of via-transition length that pushes
-# the group past its 0.5 mm length-match tolerance.  This was invisible while
-# ``check_match_group_length_skew`` was via-blind; PR #3915 threads the
-# board stackup (``board_thickness_mm`` / ``num_copper_layers``) into
-# ``derive_group_skew_data`` so via-transition length is now counted and the
-# pre-existing imbalance is correctly flagged as 1 ``match_group_length_skew``
-# error.  Dropping this baseline back to strict 0 requires a via-balanced
-# re-route (or an artifact refresh), which is blocked on the artifact-churn
-# problem tracked in #3925.  See #3931 for the resolution plan.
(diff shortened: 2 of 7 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the via-length-aware match-group tuner (#3931, re-curated 2026-07-08 to Option (a)). This is a clean, well-bounded router-side fix that closes the tuner/DRC via-awareness gap.

## Correctness — verified independently

- **ADDR_BUS convergence is real.** Ran the seed-42 reroute gate locally with the C++ backend (matching CI's exact `check_matchgroup_coverage.py --seed 42` invocation). Result: `ADDR_BUS: 8 members (7 tuned, 0 rolled back, 1 reference), skew: 19.220mm -> 0.000mm`. The ~1.6 mm of F.Cu meander per flat member is genuine copper added through the existing 45°-compliant serpentine engine (#3975), not a measurement artifact — the tuner now measures member lengths via-inclusively via `MatchGroupTracker._measure_route_total`, so A4/A6 become the de-facto reference.
- **Reroute stays DRC-clean and census-green.** Local reroute: DRC error count **26** (< 35 floor), `match_group_length_skew` violations **1** (MIPI only), pour connectivity **PASS** (3 pour nets, copper-union audit, post-quantize). Matches the PR table exactly.
- **`board_thickness_mm=None` is byte-for-byte legacy.** Confirmed at the source: `_measure_route_total` returns only `LengthTracker.calculate_route_length(route)` when `board_thickness_mm is None`. Pinned by `test_via_blind_when_no_stackup_is_a_noop` (asserts identity `results[nid][0] is routes[nid]` and `.segments is`). Good drift-prevention.
- **Thickness sourcing is right.** `core.py` sources `board_thickness_mm` from the manufacturer profile's DesignRules (`_build_manufacturer_design_rules()`) — the same value the DRC checker measures with — not the router's own `rules.DesignRules` (which has no thickness field), with a `None` fallback if the manufacturers module is unavailable. This keeps tuner and checker in lock-step.

## Baseline reconciliation — both paths verified

- **Reroute path** (what the CI gate runs): `match-group error violations: {'match_group_length_skew': 1} (baseline: 1)` → passes. ADDR_BUS compensated, MIPI_CSI_LANES (#3916) is the honest surviving signal.
- **`--skip-route` committed-artifact path**: independently ran it — reports exactly **1** match-group violation (ADDR_BUS, un-regenerated), `1 <= 1` passes. The single baseline of 1 satisfies both paths as documented.
- Note: the `--skip-route` path's *total* DRC count is 42 vs the 35 ceiling, but this is a property of the committed artifact (which this PR does not touch — the 7-file committed diff contains no `*_routed.kicad_pcb`) and is pre-existing origin/main drift tracked under #3925. The CI gate correctly runs the reroute path (26 < 35), which is what's enforced.

## Scope hygiene

- Committed diff is exactly the 7 listed files; no PCB artifact and no stray main-workspace edit leaked in (verified `origin/main..HEAD --name-only`). The builder's disclosed accidental main-workspace edit was correctly reverted.

## Tests
- 21 new/updated tests pass locally (`TestViaCountImbalance`, `TestViaInclusiveStackupThreading`, baseline test 2→1). The mixed-via-count case (2 free + 2 via) confirms the fix generalizes beyond board-07.

## CI
All 16 checks pass; `mergeStateStatus: CLEAN`. Key signal: **Match-Group Routing Regression (boards/07-matchgroup-test) — pass (28m13s)**, within the 30-min budget (#3956). Board 05/06/07 routing regressions and Diff-Pair regression all green (no overlap with the concurrent #3921 work).

Solid work — the via-inclusive measurement substitution is minimal, the legacy no-op is pinned, and the dramatic skew convergence is backed by real, DRC-clean copper.

PR #3992, opened 2026-07-08
fix(router): bound negotiated rip-up reroutes by remaining-budget per-net cap

rjwalters:
## Summary

The negotiated rip-up loop takes a stage `timeout` (the backstop) enforced by `check_timeout()`, which fires only BETWEEN nets. When a net drops to the pure-Python A* fallback on a **first-time geometric failure** (`FAILURE_NO_PATH` / `FAILURE_CLEARANCE` -- NOT covered by #3956's resume-exhaustion fast-path), one call can run ~200 s and overshoot the backstop before the next between-net check runs.

On board-06 (`--step route --seed 42`) a 360 s backstop overran to ~384 s of loop time (563 s total wall) -- all of it AFTER the best 21/21 state was banked at ~18 s. The grind produced a strictly worse state (`connected=2`) that the end-of-loop restore then discarded, so the overshoot was pure wasted wall time.

Option A from the issue: at each iteration entry, derive a per-net A* cap from the budget REMAINING and thread it to every reroute call site inside the iteration body. Generous early (a fraction of the plentiful remaining budget), tightening toward `PER_NET_CAP_FLOOR_S` as the deadline nears, so the loop can overshoot at most by one final net's bounded cap plus overhead.

## Changes

- **`algorithms/negotiated.py`**: add `derive_iter_per_net_cap(per_net_timeout, remaining_budget)` -- extends the #3474 `derive_per_net_cap` pattern to the iteration passes using REMAINING (not total) budget. An explicit `--per-net-timeout` still binds when tighter; the remaining-budget cap only ever tightens it.
- **`algorithms/__init__.py`**: export `derive_iter_per_net_cap`.
- **`router/core.py`**: in `route_all_negotiated`, derive `iter_per_net_cap` at each iteration entry and pass it (instead of the static `per_net_timeout`) to every in-loop reroute site: sequential reroute, stagnation recovery, zero-overflow recovery, targeted / full-reorder / neighborhood / via-blocked / escape rescues, and the two relief-rescue calls.
- **`tests/test_negotiated_timeout_propagation.py`**: unit coverage for the derivation + integration tests driving a real `Autorouter` through `route_all_negotiated` with a mocked deadline-respecting slow net, proving the overshoot bound.

[...]

PR #3966, opened 2026-07-07
fix(route): suppress failure-flavored diagnostics on fully-routed boards

rjwalters:
## Summary

The router printed three failure-flavored diagnostics on boards that route 100% successfully, making clean builds read as failures. This fixes all three by gating each diagnostic on the route's *final* outcome (and scope), never mid-loop state.

## Changes

**Bug A — oscillation/escape warnings on success** (`router/core.py`)
- The negotiated loop's `⚠ Oscillation detected` and `All N escape strategies exhausted` lines were printed mid-loop. A later iteration (or the post-loop best-state restore) can resolve the board to full connectivity, leaving the warnings stranded in the log.
- These messages are now buffered (`_emit_oscillation_msg`) and flushed after the loop **only when the route genuinely fails** — gated on `_stranded_nets()`, not overflow. A 100%-connected board can carry overlap-driven residual overflow (board 01 ends 3/3 connected with `overflow=2`), so gating on `overflow == 0` alone would still leak the wording. The decision lives in the pure helper `_should_flush_oscillation_msgs` for deterministic unit testing.

**Bug B — `PARTIAL` denominator counts pour-served nets** (`cli/route_cmd.py`)
- `nets_to_route` counted multi-pad nets the router strips via `_filter_pour_nets`. When the CLI's zone regex missed a zone that the net-class still flagged as pour, the net kept `net_num > 0`, entered the denominator, yet was never routed → spurious `PARTIAL: Routed 1/2` on a fully-routed board.
- New `_routable_multi_pad_nets(router)` excludes router-pour nets (`router._is_pour_net`) from the denominator at **all four** count sites (`route_pcb` + the three escalation paths). No-zone pour nets (routed as signals) stay counted.

**Bug C — `--- Results ---` scope unlabeled** (`cli/route_cmd.py`)
- The segment/via counts are scoped to newly-routed multi-pad nets, but the written file also carries preserved copper (logged separately). The heading now carries `(newly-routed nets only; preserved copper counted above)` under `--preserve-existing`.

**Tests**
- New `tests/test_route_cmd_success_output.py`: subprocess routes of demo boards 00/01 assert no failure wording and a `SUCCESS` headline; `--verbose` keeps status detail; unit tests for `_routable_multi_pad_nets` and the flush gate (genuine partial stays loud, success stays quiet, residual-overflow-on-success suppressed).
[...]

PR #3994, opened 2026-07-08
test(lint): key manufacturer-propagation allowlist by enclosing scope, not line number

### LINKED ISSUE #3993: test_manufacturer_propagation_lint: line-number-keyed _ALLOWLIST breaks on every core.py/io.py insertion — key by content instead
Process friction observed three times in the 2026-07-06/09 sweep: PRs #3945, #3966, and #3992 each shifted the allowlisted default-DesignRules() call sites in src/kicad_tools/router/{io,core}.py by adding unrelated lines above them, breaking tests/test_manufacturer_propagation_lint.py's line-number-keyed _ALLOWLIST and costing a full Doctor→Judge round-trip each time.

**Proposal**: key the allowlist by stable content instead of line numbers — e.g., (file, enclosing function name, normalized call snippet) or a per-site fingerprint of the surrounding lines — so inserting unrelated code above an allowlisted site doesn't invalidate it, while NEW DesignRules() fallback sites are still caught. Keep the lint's intent (no unaudited manufacturer-default fallbacks) fully intact.

**Acceptance**: inserting a no-op line above an allowlisted site does not fail the lint; adding a genuinely new DesignRules() call site still fails; the existing allowlisted sites carry over; test covers both cases.

rjwalters:
## Summary
The MFR001 manufacturer-propagation lint (`tests/test_manufacturer_propagation_lint.py`) keyed its `_ALLOWLIST` by absolute 1-based line numbers. Any unrelated insertion above an audited `DesignRules()` fallback shifted the site and broke the lint — observed three times (PRs #3945, #3966, #3992), each costing a full Doctor→Judge round-trip. This re-keys the allowlist by the **enclosing scope** (line-stable), preserving the lint's intent while eliminating the line-shift fragility.

## Changes
- Re-key `_ALLOWLIST` from `{path: {line_number: reason}}` to `{path: {scope_qualname: (expected_count, reason)}}`, where `scope_qualname` is the dotted enclosing-function/method path (e.g. `Autorouter.__init__`, `_run_monte_carlo_trial`).
- Extend `_DesignRulesCallFinder` with a scope stack (`FunctionDef`/`AsyncFunctionDef`/`ClassDef`) so every recorded call site carries its enclosing scope; module-level calls use `"<module>"`.
- Record a per-scope expected call count so a *new* bare `DesignRules()` added **inside** an already-allowlisted scope still fails (budget-exceeded), keeping the guard tight.
- Update `test_allowlist_entries_resolve_to_actual_call_sites` to validate scope + count instead of line numbers.
- Extract the per-file scan into `_scan_source_for_violations()`, shared by the canonical lint and the new acceptance tests (so the tests exercise production logic, not a re-implementation).
- Add two acceptance tests (`test_noop_insertion_above_allowlisted_site_does_not_fail`, `test_new_fallback_site_still_fails`).

| Inserting a no-op line above an allowlisted site does not fail the lint | ✅ | New `test_noop_insertion...` (25 synthetic no-op lines) + live check: injected 5 no-op lines above `router/core.py` L434, allowlisted call shifted to L439, `pytest` stayed 9/9 green, then reverted |
| Adding a genuinely new `DesignRules()` call site still fails | ✅ | `test_new_fallback_site_still_fails` covers both a new un-audited scope and an extra call inside an audited scope |
| Existing allowlisted sites carry over | ✅ | All 9 audited scopes (core/adaptive/io/benchmark/evolutionary) resolve; `test_manufacturer_propagation_...` + `test_allowlist_entries_resolve...` pass |
| Test covers both cases | ✅ | Two new acceptance tests, run through the shared `_scan_source_for_violations` helper |
| Lint intent intact (no unaudited manufacturer-default fallbacks) | ✅ | Scope + count enforcement; `test_fixed_sites_pass_manufacturer` unchanged and passing |

## Test Plan
- `uv run pytest tests/test_manufacturer_propagation_lint.py` → 9 passed.
- Live line-shift simulation: 5 no-op lines injected above the `router/core.py` allowlisted site (shifted L434→L439), lint stayed green, reverted (no `src/` changes in this PR).
- `uv run ruff check .` → all checks passed; `uv run ruff format . --check` → all formatted.

Closes #3993

### DIFF
--- a/tests/test_manufacturer_propagation_lint.py
+++ b/tests/test_manufacturer_propagation_lint.py
@@ -26,8 +26,9 @@ This file guards against future regressions by statically scanning every
 asserting that:
 
 1. The call passes ``manufacturer=`` explicitly, OR
-2. The call sits on an allowlisted line in ``_ALLOWLIST`` below with a
-   documented reason.
+2. The call sits in an allowlisted *scope* in ``_ALLOWLIST`` below
+   (keyed by enclosing function/method, line-stable) with a documented
+   reason and expected call count.
 
 The placement-side ``placement.analyzer.DesignRules`` is a different
 dataclass with no manufacturer field and is not at risk; the lint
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Content-keyed (scope + count) re-keying of the MFR001 allowlist is a clean, well-motivated fix for the recurring line-shift fragility (PRs #3945/#3966/#3992). Test-only change, single file.

## Review of the key questions

**Count guard catches a NEW bad call inside an audited scope** — yes, and it is a dedicated acceptance test. `test_new_fallback_site_still_fails` part (b) injects a second bare `DesignRules()` into the allowlisted `Autorouter.__init__` scope (budget 1); the budget decrement leaves the surplus call as a reported violation. This is exactly the issue's "still fails" criterion. Part (a) also covers a bad call in a brand-new un-audited scope. Both pass.

**No-op insertion stays green** — `test_noop_insertion_above_allowlisted_site_does_not_fail` shifts the audited call 25 lines down and asserts zero violations, driven through the shared `_scan_source_for_violations()` helper (real production matching, not a re-implementation). Verified the live simulation claim in the PR body is consistent.

**Scope resolution verified against real sources:**
- `router/core.py:400` (`DesignRules(**rules_dict) if rules_dict else DesignRules()`) resolves to enclosing `_run_monte_carlo_trial`, 2 bare calls → matches `(2, ...)`.
- `router/core.py:817` → `__init__` under `Autorouter` ClassDef → dotted `Autorouter.__init__`, 1 call → matches `(1, ...)`.
- `router/algorithms/evolutionary.py:224` ternary → `_run_evolutionary_trial`, 2 calls → matches `(2, ...)`.

`test_allowlist_entries_resolve_to_actual_call_sites` now enforces both existence and exact count per scope, so a stale/mismatched budget fails loudly.

## Residual masking risk — assessed acceptable

The one escape is: *removing* an existing legit bare call and *adding* a new bad one within the same audited scope nets to the same count and is masked. This requires a deliberate paired edit inside an already-audited scope, visible in any diff review. For a defense-in-depth lint this is an acceptable residual — the count guard is strictly tighter than the prior line-keyed scheme, which had zero intra-scope protection.

## Local verification (worktree issue-3993)
- `uv run pytest tests/test_manufacturer_propagation_lint.py` → **9 passed**.

## CI
All 16 checks SUCCESS (Test, Type Check, Lint & Format, C++ Build, all board end-to-end and routing regressions). `mergeStateStatus: CLEAN`. PR uses `Closes #3993`.

PR #4015, opened 2026-07-09
fix(boards): center every board on its engineering drawing sheet + `kct pcb center-on-sheet`

rjwalters:
## Summary

User-reported (with screenshots): on kicad-tools.org's interactive viewer, KiCanvas renders the
worksheet frame + title block, and every board sat badly on its sheet — the A4 boards were pinned
at the historical `(100, 100)` generator origin (06 overlapped the title block, 07 badly), and the
`paper "User"` boards (01/02/05, sized-to-board by the old `page-fit` flow) hugged the top-left
frame corner with the default title block drawn over copper.

This PR adds a reusable **`kct pcb center-on-sheet`** command (mirrors the `add-3d-models` UX from
#4012: `--dry-run` / `--format json` / `-o`), applies it to all 16 committed board artifacts
(routed + unrouted siblings), fixes the 8 generators so fresh builds are born centered, and
regenerates the coordinate-bearing sidecars/bundles.

## Design decisions

**Sheet geometry model** (KiCad default worksheet, which KiCanvas renders):
- Frame margin **10 mm** per side (`left/right/top/bottom_margin` of the default drawing sheet).
- Title block band **35 mm** above the bottom frame line — the default worksheet's title-block
  outline rect spans 34 mm above the bottom frame corner (`(rect (start 110 34) ...)` anchored at
  `rbcorner`); 35 mm covers it plus 1 mm clearance. The band is reserved across the **full** sheet
  width (not just the 110 mm-wide block) so centering stays symmetric and boards read as clearly
  above the block.
- Usable drawing area on A4 landscape: `(10, 10) – (287, 165)`; boards are centered in that box.

**Paper policy**: `User` sheets are upgraded to the smallest standard landscape size whose usable
area fits the board with **>= 15 mm slack per side** (a sized-to-board `User` page is *why* those
boards hugged the corner). Standard sheets are kept when the board fits. All 8 boards land on
**A4 landscape** (largest board, 07 at 110x95 mm, still has 30+ mm slack per side).

**Exactness**: pure text editing with decimal-string arithmetic (`Decimal`), the `models3d.py`
philosophy — a full `PCB.load()/save()` round-trip was tested first and is **not** byte-stable on
these artifacts, so the parse/rewrite path was rejected. Only the X/Y atoms of coordinate nodes
(plus the `(paper ...)` node on the three `User` boards) change; `dx/dy` is snapped once to a
**0.05 mm grid** in integer nanometres and added identically to every coordinate, so 45-degree
[...]

PR #4016, opened 2026-07-09
fix(site): frame interactive viewer on the board, not the drawing sheet

rjwalters:
## Summary
The gallery's Interactive Viewer opened fit-to-page (KiCanvas always calls `zoom_to_page()` on load), so boards rendered small/offset within the worksheet frame. This calls the viewer's own `zoom_to_board()` (fits Edge.Cuts +10%) once the existing `embed.loaded` poll sees the viewer ready, via the open shadow DOM (`kicanvas-embed` → `kc-board-app` → `.viewer`).

- Best-effort: wrapped in try/catch; failure falls back to stock page-fit.
- No vendored-bundle patch — survives KiCanvas upgrades.
- Toolbar zoom-to-page still available for the full-sheet view.
- Already validated in production (deployed alongside #4015's board centering; user-confirmed issue, camera now opens on the board).

## Test
`cd site && npm run build` (10 pages) and `npx vitest run src/data` (28 passed).

### DIFF
--- a/site/src/pages/[slug].astro
+++ b/site/src/pages/[slug].astro
@@ -298,10 +298,29 @@ if (board.manifest_generated_at !== undefined) {
   var isLoaded = function () {
     return !!(embed && embed.loaded);
   };
+  // Re-frame the camera on the board outline instead of the drawing sheet.
+  // KiCanvas's load path unconditionally calls zoom_to_page(), which fits the
+  // whole worksheet frame — boards sitting in a corner of the sheet render
+  // tiny and offset (user-visible on every board page). The board viewer
+  // exposes zoom_to_board() (fits Edge.Cuts bbox + 10%) but nothing calls it,
+  // so we do, once, when the viewer reports loaded. Best-effort: any failure
+  // leaves the stock page-fit behavior.
+  var zoomToBoard = function () {
+    try {
+      var root = embed && (embed.renderRoot || embed.shadowRoot);
+      var app = root && root.querySelector("kc-board-app");
+      var viewer = app && app.viewer;
+      if (viewer && typeof viewer.zoom_to_board === "function") {
+        viewer.zoom_to_board();
+        if (typeof viewer.draw === "function") viewer.draw();
+      }
+    } catch (e) {}
+  };
   var hideOverlay = function () {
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge Review: APPROVED

**Scope** — Single file (`site/src/pages/[slug].astro`), 19 insertions, 0 deletions: the `zoomToBoard` helper plus its one call in `hideOverlay`. No other files touched; no vendored-bundle patch.

**Defensiveness** — Verified against the surrounding IIFE:
- `zoomToBoard()` is fully wrapped in `try/catch`; no path can throw uncaught.
- Called after `settled = true` / `stopPolling()`, and the `settled` guard ensures it runs exactly once (whether via the 150ms poll or the 8s hard-timeout success branch).
- Every step is null-chained (`embed && ...`, `root && ...`, `app && ...`) with `typeof ... === "function"` checks, so a missing shadow root / app / viewer degrades silently to stock page-fit.
- Failure cannot block overlay removal (own try/catch swallows before `overlay.remove()` runs).

**Shadow-DOM pattern matches the vendored bundle** (`site/public/vendor/kicanvas.js`):
- Base custom-element class: single `attachShadow({mode:"open"})`; `renderRoot(){return this.shadowRoot??this}` — so `embed.renderRoot || embed.shadowRoot` always yields a queryable root.
- `kicanvas-embed` renders `<kc-board-app>` into its shadow root.
- `kc-board-app` (KCBoardAppElement) extends KCViewerAppElement, which exposes `get viewer(){return this.#e.viewer}`.
- Board viewer has `zoom_to_board()` (fits `Edge.Cuts` bbox +10%) and `draw()`. Note: `zoom_to_board` dereferences `by_name("Edge.Cuts").bbox`, so it can throw if a board lacks an outline layer — the try/catch here is load-bearing and correctly present.

**Validation**
- `npm run build`: succeeds (9 pages).
- `npx vitest run src/data`: 28/28 passed.
- CI: all 16 checks pass; `mergeStateStatus: CLEAN`.

**Security** — No external input; reads only same-origin open shadow DOM of the local embed. No injection surface (no innerHTML added, only method calls).

Clean, minimal, well-commented fix. Approving.

PR #4005, opened 2026-07-09
fix(board-07): gallery-ready bundle — 0 kicad-cli DRC, report Errors=0, LVS clean

rjwalters:
## Summary

Brings `boards/07-matchgroup-test` (Epic #2661 Phase 3L match-group regression bench) to its gallery-READY gates:

| Gate | Before | After |
|------|--------|-------|
| `kicad-cli pcb drc --refill-zones` (referee) | **592 errors** (drill_out_of_range ×171, track_width ×199, via_diameter ×171, clearance ×41, hole_clearance ×10) | **0 violations** (5 unconnected items = the 5 unrouted nets) |
| `kct check --mfr jlcpcb` (with sidecar) | 42 errors | **9 blocking + 5 advisory connectivity = 14 raw** |
| `output/manufacturing/report.md` DRC Status | Errors 4 / FAIL (stale 2026-06-12 bundle) | **Errors 0 / PASS** |
| `output/lvs.json` | missing | **not committed** (judge review: the fixture schematic binds 0/223 pins, so copper-LVS is vacuous — #4006; site honestly shows "LVS not run") |
| `kct board-metrics` | — | **status ok** |
| Tolerance floor | 35 | **14, zero slack** for the raw counter |

## Root cause 1: the 592-error kicad-cli mismatch (NOT a writer bug — a stale artifact)

The committed routed PCB predates #3919/#3720, so no `.kicad_pro`/`.kicad_dru` DRC-constraint sidecars accompanied it. `kicad-cli pcb drc` therefore judged the board against KiCad's **built-in defaults** (min track 0.20 mm, min via 0.50 mm, min drill 0.30 mm) instead of the jlcpcb 4-layer capability it was routed to. The systematic counts match exactly: 171 vias at Ø0.45 → 171 `via_diameter` + 171 `drill_out_of_range`; ~306 tracks at 0.15 mm → the `track_width` flood. The writer fix already exists on main (`kct route`/`kct export` emit the sidecars since #3919); this PR regenerates the artifact so the sidecars are committed next to it. No rule was loosened — the sidecars carry the jlcpcb profile minima verbatim.

## Root cause 2 (real `src/` defect, fixed + tested): sibling stitch-net pads were invisible obstacles

`run_stitch` built every obstacle pool with `exclude_nets=<ALL stitch-target nets>`. When GND/+1V2/+1V8 are stitched in one pass (this board's recipe), pads of *sibling* pour nets were not obstacles at all:

[...]

Issue #3855, opened 2026-06-22
router: places impedance fan-out vias and GND stitch vias without respecting min_hole_to_hole (sub-0.5mm drill pairs)


rjwalters:
Discovered while fixing board-06 recipe drill spacing (#3851). After the recipe via<->via guard was tightened (DRILL_CC 0.45->0.75, dropping recipe-placed drill pairs 7->1), board-06 still has 2 sub-0.5mm dimension_drill_clearance pairs that are ROUTER-placed, not recipe-placed:
1. A pair of router impedance diff-pair FAN-OUT vias (size 0.6 / drill 0.25) placed too close — the router bypasses the recipe _via_ok guard.
2. A GND stitching via vs a J1-S2 connector pad drill (a different guard path).

Root cause: the router (escape / diff-pair fan-out / stitching via placement) does not enforce the manufacturer min_hole_to_hole_mm (0.5mm) when placing vias, so it can emit drill pairs below the fab hole-to-hole minimum. The hole_to_hole DRC rule (#3842) now correctly flags these, but the router should avoid creating them.

This is the deeper, router-internal counterpart to the recipe-side #3851. It likely affects other boards (board-04/05 sub-0.5mm drills are also router-emitted escape vias per #3847). Related to the escape-router work (#3073/#3077) and the coupled diff-pair routing (#3540-#3544).

Scope (curator: assess tractability — may be hard / router-internal): make the router respect min_hole_to_hole_mm when placing fan-out / escape / stitching vias (reject or nudge a candidate via that would violate hole-to-hole against existing drills/pads). Acceptance: router does not emit sub-min_hole_to_hole drill pairs; board-06 (and ideally 04/05) reach 0 router-placed dimension_drill_clearance errors, allowing #3847/#3851 floors to tighten to 0.

Related: #3851, #3847, #3842, #3829.

rjwalters:
## Curator Enhancement (tractability assessment)

Curating per orchestration request. Read the router via-placement paths to determine whether a hole-to-hole guard is a bounded localized change or unbounded router-internal work.

### Verdict: BOUNDED-FIXABLE

The two failing cases are both **post-candidate validation gaps at specific, enumerable via-insertion sites**, not a property of the A* search itself. Rejecting an offending via candidate (and falling through to the next candidate, or skipping placement) does not break routing convergence because:
[...]

PR #4036, opened 2026-07-10
fix(stitch): treat sibling stitch-net pads as obstacles in thermal/blanket vias

### LINKED ISSUE #4010: stitch: run_thermal_stitch / run_blanket_stitch still blind to sibling stitch-net pads (same defect class fixed for run_stitch in PR #4005)
PR #4005 fixed `run_stitch` in `src/kicad_tools/cli/stitch_cmd.py`: obstacle pools were built with `exclude_nets=<ALL stitch-target nets>`, so when multiple nets are stitched in one pass, pads of *sibling* pour nets were invisible as obstacles. On board 07 this produced a +1V2 via grazing U4.C3 (+1V8) inside clearance, sibling-pad clearance violations, and drills inside KiCad's 0.25 mm hole-to-copper band.

`run_thermal_stitch` and `run_blanket_stitch` share the same exclude-all pattern and were deliberately left out of #4005 (no board recipe currently exercises them multi-net). They need the same treatment:

1. per-pad obstacle pools that include every target-net pad/bbox/drill on a different net than the pad being stitched,
2. the hole-to-copper (0.25 mm) guard on via candidates,
3. overlap rejection in any connectivity-fallback path (a short is never an acceptable connectivity trade — cf. #3633 graze grandfathering vs overlap).

Port the `TestSiblingStitchNetObstacles` / hole-to-copper / `TestFallbackOverlapGuard` test patterns from `tests/test_stitch_cmd.py`.

Refs: PR #4005 (fix + tests for run_stitch), #3633, #3930.

rjwalters:
## Summary

`run_thermal_stitch` and `run_blanket_stitch` built their obstacle pools with `exclude_nets=target_net_nums`, so when several plane nets are stitched in one pass (e.g. GND + +1V2 + +1V8) the pads/drills of **sibling** target nets were invisible obstacles. A via for one plane net could land on a sibling plane net's pad and short the rails once zones re-fill. This is the same defect class fixed for `run_stitch` in PR #4005 — thermal/blanket still used the pre-#4005 pattern (flagged as latent in the #4005 PR body; no current board recipe exercises it, but board-07-style multi-net stitches will).

## Changes

- **`router/via_clearance.py`**: extend `point_clear_of_copper` with optional `via_drill` / `other_net_drills` / `hole_to_copper_clearance` / `min_hole_to_hole` params. When a drill is supplied, the copper checks (tracks/vias/pads) additionally enforce the KiCad 0.25mm `hole_clearance` band on the drill edge, and a drill hole-to-hole guard rejects candidates too close to a foreign drill. Reuses the existing `drill_hole_to_hole_clear` helper. Defaults keep every existing caller unchanged.
- **`cli/stitch_cmd.py`**: thread the new params through the `check_via_clearance` wrapper; add a `MIN_HOLE_TO_HOLE_CLEARANCE = 0.5` module constant (matches the `calculate_via_position` default).
- **`run_thermal_stitch`** builds per-candidate obstacle pools that add every target-net pad/drill on a net **different** from the candidate's own net, and passes the drill guard params.
- **`run_blanket_stitch`** does the same, augmenting the pools per stitched net (its existing same-net pad-avoidance loop was already correct and is untouched).
- **`tests/test_stitch_cmd.py`**: new sibling-net obstacle + hole-to-copper guard tests for both paths, plus direct drill-aware `check_via_clearance` unit tests.

| `run_thermal_stitch` obstacle pools include sibling target-net pads | ✅ | `other_net_pads_with_siblings` / `other_net_drills_with_siblings` built per-candidate filtered to `!= pad_net` |
| `run_blanket_stitch` obstacle pools include sibling target-net pads | ✅ | same pattern filtered to `!= net_number` per stitched net |
| Both apply the 0.25mm hole-to-copper guard when a via drill is available | ✅ | `hole_to_copper_clearance=KICAD_HOLE_TO_COPPER_CLEARANCE` + `via_drill=drill` threaded to the shared predicate |
| New sibling-net tests for both functions (2+ nets, clear by `max(copper_floor, hole_floor)`) | ✅ | `TestThermalStitchSiblingNetObstacles` / `TestBlanketStitchSiblingNetObstacles`; both `*_clear_sibling_net_pad_copper` tests FAIL on pre-fix code (verified by reverting each pool change) |
| New hole-to-copper guard tests for both paths | ✅ | `TestCheckViaClearanceHoleToCopper` (direct, load-bearing — fails when guard disabled) + per-via hole-floor assertions in the integration tests |
| Existing thermal/blanket suites pass unchanged (no single-net regression) | ✅ | 268 passed (259 existing + 9 new); single-net tests included |
| `TestFallbackOverlapGuard` explicitly NOT ported | ✅ | Thermal/blanket have no #3633 connectivity-fallback cascade (skip pads just record `pads_skipped`), so there is no fallback overlap path to guard — out of scope per the issue |
| `ruff check` + `ruff format --check` clean on changed files | ✅ | `All checks passed!` / `3 files already formatted` |
| `mypy` clean | ✅ | 20 pre-existing errors on both files, unchanged by this diff (no new errors introduced) |

## Test Plan

- `uv run pytest tests/test_stitch_cmd.py` — 268 passed
- Load-bearing check: reverting the thermal and blanket pool changes makes the two `*_clear_sibling_net_pad_copper` tests fail (2 failed / 4 passed); reverting the `hole_extra` / drill guard in `point_clear_of_copper` makes `TestCheckViaClearanceHoleToCopper` fail (2 failed / 1 passed).
- Ran all `point_clear_of_copper` router callers (escape/pathfinder/main-router clearance suites) — 163 passed, no regression.

Closes #4010

### DIFF
--- a/src/kicad_tools/cli/stitch_cmd.py
+++ b/src/kicad_tools/cli/stitch_cmd.py
@@ -3780,7 +3780,19 @@ def run_thermal_stitch(
     other_net_pads = find_all_pads(sexp, exclude_nets=target_net_nums)
     other_net_filled_polys = find_all_filled_polygons(sexp, exclude_nets=target_net_nums)
     # Issue #3855: foreign-net DRILL registry for the hole-to-hole guard.
-    other_net_drills = find_all_drills(sexp, exclude_nets=target_net_nums)
+    # Issue #4010: exclude target-net PAD drills from the base pool too --
+    # a thermal via halo rings a TO-220 pad at ~0.55mm from its center,
+    # which lands INSIDE the 0.5mm MIN_HOLE_TO_HOLE_CLEARANCE of the pad's
+    # own 1.0mm plated drill.  ``find_all_drills(exclude_nets=...)`` only
+    # excludes foreign-net VIAS; without ``pad_exclude_nets`` the target
+    # pad's own drill survives here and the guard rejects every halo via
+    # against it (board-05: all six MOSFETs dropped to ZERO thermal vias).
+    # Sibling-net pad drills still enter the guard via ``target_pad_drills``.
+    other_net_drills = find_all_drills(
+        sexp,
+        exclude_nets=target_net_nums,
+        pad_exclude_nets=target_net_nums,
+    )
 
     # Issue #4010: sibling stitch-net pads are HARD obstacles.  The
     # ``other_net_*`` pools above exclude ALL stitch-target nets, so when
@@ -4026,7 +4038,16 @@ def run_blanket_stitch(
     other_net_pads = find_all_pads(sexp, exclude_nets=target_net_nums)
     other_net_filled_polys = find_all_filled_polygons(sexp, exclude_nets=target_net_nums)
     # Issue #3855: foreign-net DRILL registry for the hole-to-hole guard.
-    other_net_drills = find_all_drills(sexp, exclude_nets=target_net_nums)
+    # Issue #4010: exclude target-net PAD drills from the base pool too (see
+    # the matching note in run_thermal_stitch).  A blanket via ringing a
+    # thru-hole pad can land inside the pad's own drill hole-to-hole floor;
+    # without ``pad_exclude_nets`` the guard would reject it against its own
+    # pad drill.  Sibling-net pad drills still enter via ``target_pad_drills``.
+    other_net_drills = find_all_drills(
+        sexp,
+        exclude_nets=target_net_nums,
+        pad_exclude_nets=target_net_nums,
+    )
 
     # Issue #4010: sibling stitch-net pads/drills are HARD obstacles.  The
     # ``other_net_*`` pools above exclude ALL stitch-target nets, so when
--- a/src/kicad_tools/cli/stitch_cmd.py
+++ b/src/kicad_tools/cli/stitch_cmd.py
@@ -43,6 +43,14 @@ from kicad_tools.sexp.builders import segment_node, via_node
 #: so via-candidate checks take the max of both constraints.
 KICAD_HOLE_TO_COPPER_CLEARANCE = 0.25
 
+#: Default minimum drill edge-to-edge spacing, in mm, for the hole-to-hole
+#: guard (``dimension_drill_clearance`` fab floor, Issue #3855).  Matches the
+#: ``calculate_via_position`` default and the local checks in
+#: ``run_stitch``'s connectivity fallback.  Used by the thermal/blanket
+#: stitch paths (Issue #4010) so a stitch via never lands within this floor
+#: of a foreign through-hole pad / via drill.
+MIN_HOLE_TO_HOLE_CLEARANCE = 0.5
+
 
 def _count_copper_layers(pcb_path: Path) -> int:
     """Count the number of copper layers in the PCB.
@@ -3315,6 +3323,10 @@ def check_via_clearance(
     other_net_pads: list[tuple[float, float, float, int]],
     same_net_vias: list[tuple[float, float]],
     other_net_filled_polygons: list[FilledPolygon] | None = None,
+    via_drill: float = 0.0,
+    other_net_drills: list[tuple[float, float, float, int]] | None = None,
+    hole_to_copper_clearance: float = 0.0,
+    min_hole_to_hole: float = 0.0,
 ) -> bool:
     """Check if a via at (x, y) passes all clearance checks.
 
@@ -3334,6 +3346,17 @@ def check_via_clearance(
         other_net_pads: Pads on other nets as (x, y, radius, net_num)
         same_net_vias: Existing same-net vias as (x, y) for stacking prevention
         other_net_filled_polygons: Filled polygons from other-net zone fills
+        via_drill: Drill diameter of the candidate via in mm.  When ``> 0``
+            with ``hole_to_copper_clearance > 0`` the copper checks also
+            enforce the KiCad ``hole_clearance`` band on the drill edge
+            (Issue #4010).
+        other_net_drills: Foreign-net drill registry as
+            ``(x, y, drill, net_num)``; when supplied with ``via_drill > 0``
+            and ``min_hole_to_hole > 0`` the candidate drill must clear
+            every foreign drill edge-to-edge (Issue #3855).
+        hole_to_copper_clearance: KiCad ``hole_clearance`` band in mm
+            (typically :data:`KICAD_HOLE_TO_COPPER_CLEARANCE`).
+        min_hole_to_hole: Minimum drill edge-to-edge spacing in mm.
 
     Returns:
         True if the position is clear for via placement
@@ -3350,6 +3373,10 @@ def check_via_clearance(
         other_net_pads=other_net_pads,
         same_net_vias=same_net_vias,
         other_net_filled_polygons=other_net_filled_polygons,
+        via_drill=via_drill,
+        other_net_drills=other_net_drills,
+        hole_to_copper_clearance=hole_to_copper_clearance,
+        min_hole_to_hole=min_hole_to_hole,
     )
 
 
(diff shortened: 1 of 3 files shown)


rjwalters:
❌ **Changes Requested — CI failing on a real fleet-board regression (not flaky)**

## CI status

The **Test** job fails (`tests/test_board_05_thermal_stitch.py::TestBoard05ThermalStitch::test_each_mosfet_heat_sink_gets_min_vias`), and I reproduced it locally in the worktree:

```
Board 05 thermal-stitch failed to place enough thermal vias on 6 MOSFET(s):
  Q1 pad 2: 0 via(s) (need >= 4)
  Q2 pad 2: 0 via(s) (need >= 4)
  ... Q3–Q6 all 0 ...
```

Board 05 is the canonical routing-regression vehicle, so this is a genuine regression introduced by this PR — every one of the six TO-220 IRLZ44N heat-sink pads goes from ≥4 thermal vias to **zero**. The CI run's other jobs (Board 00-07, DRC, lint, type) all pass; only the full-suite Test job catches it because the board-05 fixture is the only place thermal-stitch runs against real crowded copper.

## Root cause (traced end-to-end)

The new hole-to-hole guard is rejecting thermal vias for being too close to **their own pad's plated drill**.

`run_thermal_stitch` (and `run_blanket_stitch`) build the base drill pool as:

```python
other_net_drills = find_all_drills(sexp, exclude_nets=target_net_nums)
```

But `find_all_drills`'s `exclude_nets` param only excludes **vias** — its `pad_exclude_nets` param (which excludes through-hole *pad* drills) **defaults to an empty set** (intentionally, for the #3855 J1-S2 same-net case). So every target pad's own plated drill stays in `other_net_drills`.

The per-candidate sibling filter you added only filters `target_pad_drills` (`d[3] != pad_net`) — it does **not** filter the base `other_net_drills` pool, so the candidate's own-pad drill survives into `other_net_drills_with_siblings`.

Concretely, for Q4 pad 2 (net PHASE_B, 1.0mm plated drill at (132.5,113.5)):
- `generate_thermal_via_positions` places halo vias at ~0.55mm from the pad center.
- Candidate (131.95,112.95): edge-to-edge to the pad's own 1.0mm drill = **0.178mm < 0.5mm** `MIN_HOLE_TO_HOLE_CLEARANCE`.
- The new `drill_hole_to_hole_clear` guard rejects it — against the via's **own** pad drill.

I instrumented the candidate loop: **39 candidates** that pass copper clearance are killed purely by the drill guard, and every rejecting drill is the candidate's own heat-sink pad drill. The thermal halo ring for a large TO-220 pad simply sits inside the 0.5mm hole-to-hole floor of its own 1.0mm drill, so the guard nukes the entire ring.

This is the mirror of the `!= pad_net` mistake the issue explicitly warned about — but on the **base** pool rather than the sibling pool.

## Fix

Exclude the target-net *pad* drills from the base pool so a thermal/blanket via is never rejected against its own pad's plated hole (sibling-net pad drills still come in via `target_pad_drills`, which is correct):

```python
other_net_drills = find_all_drills(
    sexp,
    exclude_nets=target_net_nums,
    pad_exclude_nets=target_net_nums,   # <-- add this
)
```

in **both** `run_thermal_stitch` (~line 3783) and `run_blanket_stitch` (~line 4029). I verified this restores all six MOSFETs to exactly 4 vias each (total 58 placed, was passing pre-PR) while keeping the sibling-net guard intact.

Note this diverges slightly from the `run_stitch` (#4005) call, which passes only `exclude_nets`. That's fine because `run_stitch` escapes/fans out from the pad edge rather than ringing a via halo around the pad's own drill, so it never trips its own-pad hole floor. Thermal/blanket are structurally different and need the pad exclusion.

## Please also add a regression test

Add a thermal-and-blanket fixture where the stitched pad is a **thru-hole** pad with a drill large enough that the halo ring lands inside `MIN_HOLE_TO_HOLE_CLEARANCE` of its own drill, and assert vias are still placed. Your current `THERMAL_SIBLING_PCB` / `BLANKET_SIBLING_PCB` fixtures use SMD pads (no drill), so they never exercise the own-pad-drill path — which is exactly why the 413 lines of new tests all pass while board-05 breaks. A thru-hole fixture (or a slimmed board-05-style case) would have caught this.

Everything else in the PR looks good — the `point_clear_of_copper` extension is clean, the no-op defaults are correct (verified the 163 router-caller path is unaffected since `via_drill` defaults to 0.0), the sibling-pad logic is right, and the scope boundary on `TestFallbackOverlapGuard` is justified. Just this one pool-construction bug to fix.

I'll re-evaluate once the drill pool excludes own-net pad drills and the board-05 test passes.

rjwalters:
## Feedback addressed — own-net pad drills excluded from the base pool

Thanks for the end-to-end trace; it landed exactly on the bug. Applied your verified fix and closed the test-coverage gap.

### Fix (commit 3010607b)
Added `pad_exclude_nets=target_net_nums` to the base `other_net_drills = find_all_drills(...)` call in **both** `run_thermal_stitch` (~line 3783) and `run_blanket_stitch` (~line 4029). A thermal/blanket via is no longer rejected against its own pad's plated drill; sibling-net pad drills still reach the guard via `target_pad_drills`, so the #4010 sibling-net guard is intact. Each site carries a comment explaining the halo-inside-own-drill-floor geometry and why this diverges from the `run_stitch` (#4005) call.

### Regression coverage (the SMD-fixture gap you flagged)
Added `TestStitchOwnPadDrillNotBlocking` in `tests/test_stitch_cmd.py` with **thru-hole-pad** fixtures (`THERMAL_THT_PCB` / `BLANKET_THT_PCB`, 1.7mm copper over a large 1.4mm drill, `thermal_radius=0.05`) so the halo/grid ring lands ~0.1mm edge-to-edge from the pad's own drill — well inside the 0.5mm floor, the same failure mode as board-05. Five tests:
- pool-construction sanity: `exclude_nets` alone leaves the own-net pad drill in the pool; `pad_exclude_nets` removes it.
- thermal + blanket **placement**: vias ARE placed inside the own-drill floor **and** each test monkeypatches the base pool back to the buggy `exclude_nets`-only form and asserts **ZERO** near-own vias under the bug — i.e. these fixtures would have caught the regression (SMD fixtures could not, no drill).
- thermal + blanket **sibling-block**: every placed via still clears the sibling-net THT pad drill by the hole-to-hole floor.

### Results
- `tests/test_board_05_thermal_stitch.py` — **PASS**: all six TO-220 MOSFETs (Q1–Q6) restored to >=4 thermal vias each (was ZERO). Verified the board-05 test still **fails** when the fix is reverted.
- `tests/test_stitch_cmd.py` + `tests/test_board_05_thermal_stitch.py` — 276 passed.
- `ruff check` / `ruff format --check` — clean on both changed files. (mypy: no new errors; the 17 in `stitch_cmd.py` pre-date this PR.)

Re-labeled `loom:review-requested`. Ready for re-review.

rjwalters:
✅ **Approved!** (re-review after Doctor pass, commit 3010607b)

The prior pass caught the own-pad-drill regression (board-05 thermal vias zeroed) and prescribed the fix. Verified the Doctor applied it exactly and closed the test-coverage gap.

## Fix matches the prescription
- `pad_exclude_nets=target_net_nums` added to the base `other_net_drills = find_all_drills(...)` call in **both** `run_thermal_stitch` (~L3779) and `run_blanket_stitch` (~L4037). A thermal/blanket via is no longer rejected against its own pad's plated drill; sibling-net pad drills still reach the guard via `target_pad_drills`, so the #4010 sibling-net guard is intact.
- Each site carries a comment explaining the halo-inside-own-drill-floor geometry and why this diverges from the `run_stitch` (#4005) call.

## Board-05 regression is green
- **Board 05 Routing Regression** — pass (19m12s). The six TO-220 MOSFETs (Q1–Q6) are restored to >=4 thermal vias each (were ZERO on the prior head). This is the job the diff broke; it now passes.

## New THT tests are load-bearing
`TestStitchOwnPadDrillNotBlocking` uses real thru-hole fixtures (`THERMAL_THT_PCB` / `BLANKET_THT_PCB` — 1.7mm copper over a 1.4mm drill, tight thermal radius) so the halo ring lands inside the 0.5mm hole-to-hole floor of the pad's own drill — the board-05 failure mode reproduced in miniature. This closes the SMD-fixture gap flagged last pass (SMD pads have no drill, so the original fixtures never exercised the path).
- Each placement test monkeypatches the base pool back to the buggy `exclude_nets`-only form and asserts **ZERO** near-own vias under the bug vs. non-zero with the fix — i.e. these fixtures WOULD have caught the original regression.
- Pool-construction sanity test confirms `exclude_nets` alone leaves the own-net pad drill in the pool and `pad_exclude_nets` removes it.
- Sibling-block tests confirm every placed via still clears sibling-net THT pad drills by the hole-to-hole floor, so the guard is not weakened.

## CI + merge state
All 16 checks pass, including **Test** (15m35s — the job that failed on the prior head) and every board/routing regression. `mergeStateStatus` = CLEAN.

PR body uses `Closes #4010`. Clean fix, correct scope, real regression coverage. Ship it.

Issue #4027, opened 2026-07-10
flaky: test_board_03_copper_lvs fresh-regen fails under concurrent load, passes in isolation


rjwalters:
`tests/test_board_03_copper_lvs.py::TestBoard03FreshRegenCopperLVSClean::test_fresh_regen_is_copper_lvs_clean` (full route-from-source regen, ~2–11 min) failed during two independent review sessions on 2026-07-09 — both times while other pytest/regen jobs were loading the machine — and passed 3/3 on quiet re-runs of the same commit (e9d0625d verification) and again in the #4013 review (passed isolated; its CI job green).

Hypothesis: the regen's router timing/budget is load-sensitive (deterministic-budget assumes stable throughput), so a loaded machine changes route outcomes enough to trip copper-LVS.

Fix options: (a) make the regen's determinism actually load-independent (iteration-count budget rather than wall-clock, if that's the seam); (b) mark the test `@pytest.mark.serial`/resource-isolated so it never shares the machine; (c) if wall-clock budgets are inherent, detect budget exhaustion and skip-with-reason instead of failing. Two false alarms in one day says this will keep taxing reviewers.

rjwalters:
## Curator: verified root-cause analysis

I traced the seam in `boards/03-usb-joystick/generate_design.py` and the router's budget code. The hypothesis in the issue body is *close* but not quite right, and the actual mechanism points to a specific fix.

### The router's per-net A* budget is already load-independent

Board 03's `route_pcb()` (`boards/03-usb-joystick/generate_design.py:561-629`) already invokes `kct route` with `--deterministic-budget` (issue #3538/#3799/#3881). That flag disables the per-net *wall-clock* A* cutoff and pins the C++ A* search to a fixed node-expansion count (`DETERMINISTIC_BUDGET_PER_NET_ITERATIONS = 1_000_000`, `src/kicad_tools/cli/route_cmd.py:164`), so the per-net search itself is provably load-independent — this is exactly fix option (a), and it already shipped for this board. So the flake is *not* in the per-net A* loop.

### The actual seam: an outer wall-clock deadline that the pipeline doesn't check

The `route_pcb()` call passes `--timeout 600` (`generate_design.py:587-588`) as what the code's own comment (`generate_design.py:583`) calls a "SAFETY backstop only." But it is *not* inert:

[...]

PR #4038, opened 2026-07-10
fix(schematic): make net-query connectivity multi-unit-aware

### LINKED ISSUE #4020: schematic: get_net_for_pin is not multi-unit-aware (phantom nets from unit-1 geometry)
Found during the softstart LVS unification (softstart cdba1e1): for multi-unit symbols (e.g. LM393 dual comparator), `get_net_for_pin` resolves pins against **unit 1's geometry only**, producing phantom net assignments for pins that live in other units. The softstart generator had to work around it by synthesizing a single-unit `LM393_SINGLE` symbol.

Fix: make pin→net resolution unit-aware (each placed unit instance resolves its own pins at its own placement). Add a regression test with a dual-unit symbol where unit 2's pins sit at different sheet coordinates. Then the softstart workaround can be retired (separate repo, note it in the PR).

Ref: softstart LVS work 2026-07-09 (kicad-tools e9d0625d landed the copper-side fixes; this label-side gap was reported, not worked around dishonestly).

rjwalters:
## Summary

`get_net_for_pin`, `are_connected`, `extract_netlist`, and `pins_on_net`
all sit on top of `SchematicNetlistMixin._build_connectivity_graph()`,
which was not multi-unit-aware. For a multi-unit symbol (e.g. an LM393
comparator with channel-A / channel-B / power units sharing one
`reference`), the shared `symbol_def.pins` list contains every unit's
pins. `_build_connectivity_graph()` iterated that whole list for **each**
placed `SymbolInstance` without filtering by `pin.unit`, so every placed
unit re-registered every other unit's pins at its *own* placement anchor.

Result: phantom nets synthesized from unit-1 geometry, false
`are_connected()` results, and duplicate `PinRef` entries. This is core
connectivity used by LVS and every board recipe.

`get_net_for_pin()` and `are_connected()` compounded the bug by picking
the **first** `SymbolInstance` whose `.reference` matched, ignoring which
unit actually owns the requested pin.

## Changes

- Add module-level `_instance_owns_pin(sym, pin)` helper — a pin belongs
  to an instance when `pin.unit == 0` (common) or equals the instance's
  `unit`. Mirrors the proven per-unit filter in
  `validation_mixin._check_unconnected_pins` (#3349) and
  `pin_position`'s `_pin_matches_unit` (#3346) so the three paths don't
  drift.
- Filter contributed pins in `_build_connectivity_graph()` so each
  instance only registers pins it owns. Unit-0 "common" pins are still
  contributed by every placed instance (each at its own resolved
  coordinate) — connectivity, not a one-shot flag.
- Add `_resolve_pin_instance()` and use it in `get_net_for_pin()` and
  `are_connected()` to select the unit-owning instance, falling back to
  the first matching-reference instance so single-unit behaviour is
  unchanged.
- Regression tests (`TestMultiUnitNetResolution` in
  `tests/test_netlist_query.py`): synthetic dual-unit LM393-shaped
  fixture with unit 1 and unit 2 placed at **different sheet
  coordinates**.

| `get_net_for_pin` uses the specific placed instance (reference + owning unit) | ✅ | `_resolve_pin_instance`; `test_get_net_for_pin_resolves_against_owning_unit`, `test_get_net_for_pin_wired_unit2_pin_uses_unit2_wiring` |
| `_build_connectivity_graph` filters pins to those the instance owns | ✅ | `_instance_owns_pin` guard in the pin loop |
| Dual-unit fixture at different coords: unit-2 pin returns real net / `None` when unwired | ✅ | `test_get_net_for_pin_unwired_unit2_pin_is_floating` (`None`), `..._uses_unit2_wiring` (real net) |
| `are_connected` returns `False` for unwired cross-unit pins | ✅ | `test_are_connected_false_for_unwired_cross_unit_pins`; `..._true_when_units_actually_wired_together` for the positive case |
| No duplicate `PinRef` across unit instances | ✅ | `test_extract_netlist_no_duplicate_unit_specific_pinrefs`, `test_extract_netlist_no_phantom_cross_unit_pin_on_wired_net`, `test_pins_on_net_no_phantom_cross_unit_pin` |
| Existing multi-unit + `test_netlist_query.py` suites pass unchanged | ✅ | `test_schematic_models.py`, `test_blocks_overcurrent_comparator.py`, `test_netlist_query.py` all green |
| No dependency on external softstart repo | ✅ | Fixture fully synthetic in `test_netlist_query.py` |

## Test Plan

- `uv run pytest tests/test_netlist_query.py tests/test_schematic_models.py tests/test_blocks_overcurrent_comparator.py` → 257 passed (+9 new).
- Broad guard: `uv run pytest tests/ -k "schematic or netlist or lvs"` (excluding 4 pre-existing `cmaes`-import-error collection failures unrelated to this change) → 1543 passed, 1 flaky failure in `TestBoard03FreshRegenCopperLVSClean` (a full PCB route+DRC regen — `connectivity: 2 errors` come from the router/kicad-cli DRC, not schematic net queries). That test **passes in isolation** on this branch, confirming the failure is pre-existing routing flakiness, not caused by this change.
- Manual: the issue's reproduction script now prints `None` / `False` / deduplicated PinRefs (was `PHANTOM_NET` / `True` / duplicated).
- Single-unit symbols are zero-diff: pins tagged `unit=0` match any instance, and the first-match fallback preserves prior behaviour.

### Out of scope (note-only)

The softstart repo (external, local-only) worked around this bug by
synthesizing a single-unit `LM393_SINGLE` symbol. That workaround can now
be retired as a **separate, out-of-repo follow-up** — not touched here.
The fix and its regression fixture are fully self-contained in
`kicad-tools`.

Closes #4020

### DIFF
--- a/src/kicad_tools/schematic/models/netlist_mixin.py
+++ b/src/kicad_tools/schematic/models/netlist_mixin.py
@@ -29,6 +29,31 @@ class PinRef:
         return f"{self.symbol_ref}.{self.pin}"
 
 
+def _instance_owns_pin(sym, pin) -> bool:
+    """Return True if a placed ``SymbolInstance`` owns ``pin``.
+
+    Multi-unit KiCad symbols (e.g. the LM393 dual comparator) share a
+    single ``SymbolDef`` — and therefore a single ``symbol_def.pins``
+    list containing *every* unit's pins — across several placed
+    ``SymbolInstance`` rows that share a ``reference`` but differ in
+    their ``unit`` field.  Connectivity must attribute each pin to the
+    instance that actually carries it, otherwise a unit-2-only pin gets
+    positioned against a unit-1 instance's placement anchor and produces
+    "phantom" nets from unit-1 geometry (issue #4020).
+
+    A pin belongs to an instance when its ``pin.unit`` matches the
+    instance's ``unit`` field, or when it is a unit-0 "common to all
+    units" pin (shared package power pins, and single-unit symbols whose
+    pins are all tagged ``unit=0``).  This mirrors ``_pin_matches_unit``
+    in :meth:`SymbolInstance.pin_position` and the per-unit filter in
+    :meth:`Schematic._check_unconnected_pins` (issue #3349), keeping the
+    three code paths from drifting apart.
+    """
+    pin_unit = getattr(pin, "unit", 0)
+    sym_unit = getattr(sym, "unit", 1)
+    return pin_unit == 0 or pin_unit == sym_unit
+
+
 class SchematicNetlistMixin:
     """Mixin providing netlist extraction and query operations for Schematic class."""
 
@@ -114,8 +139,19 @@ class SchematicNetlistMixin:
                     union(junc_pos, seg_end)
 
         # Connect symbol pins to wires and track pin locations
+        #
+        # ``symbol_def.pins`` lists every unit's pins, shared across all
+        # placed instances of a multi-unit symbol.  Only contribute the
+        # pins this instance actually owns (``_instance_owns_pin``) so a
+        # unit-2 pin is never registered at a unit-1 instance's position
+        # — the phantom-net / duplicate-PinRef bug of issue #4020.  Unit-0
+        # "common" pins are still contributed by every placed instance
+        # (each at its own resolved coordinate) because connectivity, not
+        # a one-shot flag, is being computed here.
         for sym in self.symbols:
             for pin in sym.symbol_def.pins:
+                if not _instance_owns_pin(sym, pin):
+                    continue
                 pos = sym.pin_position(pin.number)
                 pos_rounded = (round(pos[0], 2), round(pos[1], 2))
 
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the multi-unit connectivity fix. The semantics are correct and the change is genuinely zero-diff for single-unit symbols.

## Correctness

- `_instance_owns_pin(sym, pin)` exactly mirrors the two proven sibling filters — `SymbolInstance._pin_matches_unit` (symbol.py:616-618, #3346) and `_check_unconnected_pins` (validation_mixin.py:334-337, #3349): `pin_unit == 0 or pin_unit == sym_unit`. The three paths now share one convention.
- Default alignment verified: `Pin.unit` defaults to `0` (pin.py:41) so `getattr(pin, "unit", 0)` matches; `SymbolInstance.unit` defaults to `1` (symbol.py:582) so `getattr(sym, "unit", 1)` matches.
- **Single-unit zero-diff confirmed**: single-unit symbols tag all pins `unit=0` (symbol.py:188, 613), so `_instance_owns_pin` returns True for every instance and the `_build_connectivity_graph` guard never skips anything. The first-match fallback in `_resolve_pin_instance` reproduces the old `for sym ... break` behavior exactly.
- The fallback does not silently mask a missing unit in a new way: when the owning unit isn't placed, `_resolve_pin_instance` returns the first matching-reference instance, and `pin_position` (symbol.py:616+) then applies its own identical unit filter with the same any-match fallback (#3346). This preserves the pre-existing documented contract; there's no correct answer for an unplaced unit regardless.
- `are_connected` now resolves each pin against its owning instance, computes each position on the correct instance, and unions in the unit-filtered graph — the false-connection path is closed.

## Unit-0 common pins

The PR deliberately contributes unit-0 pins from *every* placed instance (each at its own resolved coordinate) in `_build_connectivity_graph`, unlike `_check_unconnected_pins` which reports commons only from the first instance to avoid double-counting. This is the right call for a connectivity computation vs. a lint count, and the dedup test correctly scopes its assertion to unit-specific pins. `test_common_pin_connects_across_all_placed_units` guards the non-over-restriction case.

## Tests & CI

- Local: `tests/test_netlist_query.py` + `tests/test_schematic_models.py` → 244 passed; `tests/test_blocks_overcurrent_comparator.py` (real multi-unit LM393) → 16 passed.
- The synthetic dual-unit fixture places unit 1 @ (100,100) and unit 2 @ (200,200) at genuinely different coordinates — exactly the geometry that exposed the phantom-net bug. 9 new tests cover phantom net (None), false are_connected, no duplicate/phantom PinRefs, positive cross-unit wiring, and unit-0 commons.
- **All 16 CI checks pass**, including the full Test job (13m19s), Boards 03/04/06/07 LVS, Board 05 routing regression (19m), and diff-pair/match-group regressions — no LVS behavior change on the fleet.
- `TestBoard03FreshRegenCopperLVSClean` passed cleanly in the CI Test job (the pre-existing #4027 router/DRC flake did not manifest, and it is unrelated to schematic net-query anyway).
- Merge state CLEAN / MERGEABLE against current main; "Closes #4020" present.

Clean, well-documented, self-contained fix. LGTM.

PR #4042, opened 2026-07-10
test: pass --allow-unsafe-grid in board-05 throughput fixture to fix ERROR

### LINKED ISSUE #4024: tests: TestBoard05RoutingThroughput errors on main — auto-grid safety gate aborts board 05 under the python backend
Two `TestBoard05RoutingThroughput` tests error on main (reproduced on clean checkouts at 398a1fa4 and since; confirmed pre-existing in the #4003 and #4015 reviews): the router's auto-grid safety gate rejects board 05 under the **python backend's ~2M-cell cap**, so the fixture errors before the throughput assertion runs.

These pollute every local full-suite run with 2 errors that agents/reviewers must repeatedly re-verify as "pre-existing" (it has now been independently re-derived at least three times — that's the cost of leaving it red).

Fix options: (a) make the fixture require/build the C++ backend (the throughput test is meaningless on the python fallback anyway); (b) skip-with-reason when only the python backend is available; (c) shrink the fixture's grid to fit the cap if the throughput signal survives. Prefer (a)/(b) — the test should never error, only pass/skip/fail.

rjwalters:
## Summary

`TestBoard05RoutingThroughput` errored (not failed) on `main`. The two
tests reported `ERROR at setup` because the class-scoped `route_stdout`
fixture called `pytest.fail()` after `kct route` returned exit code 1.

The root cause is **not** a python-vs-cpp backend distinction (refuting the
issue's original diagnosis). Board 05's dense geometry (210 pads, 0.15mm
clearance, 80x100mm) forces `--grid auto` coarser than clearance/2 once the
memory budget cap (`max_cells=2,000,000`) is hit, tripping the #3911
auto-grid safety gate (`route_cmd.py` ~8594). That gate does `return 1`
**before** backend selection, so it fires identically under `--backend
python` and `--backend cpp`. A `pytest.fail()` raised inside a class-scoped
fixture is what pytest reports as ERROR for every dependent test.

This class is a **throughput-only** regression guard (the #2681 per-net A*
slowdown dimension). Its routed PCB is written to a `TemporaryDirectory` and
discarded — never manufactured, never DRC-checked — so the DRC-safety intent
of the auto-grid gate does not apply. Adding `--allow-unsafe-grid` to the
fixture's `kct route` invocation makes the gate a no-op **for this test
only**, without touching production gate logic.

## Changes

- `tests/test_board_05_routing_regression.py`: add `--allow-unsafe-grid` to
  the `route_stdout` fixture's `cmd` list.
- Update the class/module docstring and add an inline comment explaining why
  the flag is safe here (throughput-only, output discarded), mirroring the
  existing NON-PRODUCTION RECIPE note style.

| 1. Both tests report PASS/SKIP/FAIL (never ERROR), with and without C++ backend | ✅ | Ran `uv run pytest -m slow` with the C++ backend built → `2 passed in 259.65s`. The gate fires identically under both backends (before backend selection), and the flag bypasses it in both, so the python-only checkout also reports PASS/FAIL, never ERROR. |
| 2. Throughput assertions still exercise real `kct route` output | ✅ | Fixture still runs real `kct route` as a subprocess (`--backend python`, 240s budget, seed 42, `--no-cache`); both tests parse actual "Nets routed:"/"Initial pass:" stdout — no stub/mock. `MIN_2L_COMPLETION_PCT=35` and the iter-0 floor of 15 both executed and passed against live output. |
| 3. No change to `io.py` / `route_cmd.py` production gate logic | ✅ | `git diff --stat` shows only `tests/test_board_05_routing_regression.py` (plus the auto-managed `.loom-managed` worktree marker). #3911's gate is untouched. |
| 4. `.github/workflows/slow-tests.yml` nightly runner unaffected | ✅ | Inspected: workflow runs `-m slow` (unchanged selection), builds the C++ backend, gates on pytest exit code. The change is confined to a CLI flag inside the test subprocess; the test still runs under `-m slow` and now PASSes instead of ERRORing. |
| 5. Docstring updated to note the flag and why it's safe | ✅ | Module docstring + inline comment added, matching the existing NON-PRODUCTION RECIPE style. |

## Test Plan

- `uv run ruff check tests/test_board_05_routing_regression.py` → All checks passed
- `uv run ruff format --check tests/test_board_05_routing_regression.py` → already formatted
- `uv run pytest tests/test_board_05_routing_regression.py -m slow -v` → `2 passed in 259.65s` (C++ backend built via `kct build-native`)

Closes #4024

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 4022
-# Branch: feature/issue-4022
+# Issue: 4024
+# Branch: feature/issue-4024
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/tests/test_board_05_routing_regression.py
+++ b/tests/test_board_05_routing_regression.py
@@ -34,6 +34,20 @@ path entirely.  Do not treat this throughput floor as a manufacturability
 or DRC-cleanliness gate for board 05; that is
 ``tests/test_board_05_drc_hotspot_regression.py`` against the committed
 cpp/tier1/4L snapshot.
+
+The recipe also passes ``--allow-unsafe-grid`` (Issue #4024).  Board 05's
+geometry (210 pads, 0.15mm clearance, 80x100mm) makes ``--grid auto``
+select a grid coarser than clearance/2 once the memory budget cap
+(max_cells=2,000,000) is hit, which trips the #3911 auto-grid safety gate
+(``route_cmd.py`` ~8594) BEFORE any backend selection happens -- so the
+gate does ``return 1`` identically under both ``--backend python`` and
+``--backend cpp``.  A fatal exit 1 raised inside the class-scoped fixture
+turns into pytest ERROR (not FAIL) for every dependent test.  Because this
+throughput guard discards its routed output (written to a
+``TemporaryDirectory`` and never manufactured or DRC-checked), the
+DRC-safety gate is irrelevant here; ``--allow-unsafe-grid`` makes the gate
+a no-op for this fixture without touching production behavior anywhere
+else.
 """
 
 from __future__ import annotations
@@ -181,6 +195,19 @@ class TestBoard05RoutingThroughput:
             # cached route rather than current router throughput.  Both
             # tests in this class pin THROUGHPUT, so the route must
             # actually run.
+            # ``--allow-unsafe-grid`` (issue #4024): board 05's dense
+            # geometry (210 pads, 0.15mm clearance) forces ``--grid auto``
+            # coarser than clearance/2 under the memory budget cap, tripping
+            # the #3911 auto-grid safety gate (route_cmd.py ~8594), which
+            # does ``return 1`` BEFORE backend selection -- identically for
+            # python and cpp.  That fatal exit inside this class-scoped
+            # fixture surfaces as pytest ERROR (not FAIL) on every dependent
+            # test.  This is a throughput-only guard: the routed PCB is
+            # written to the TemporaryDirectory above and discarded (never
+            # manufactured, never DRC-checked), so the DRC-safety intent of
+            # the gate does not apply.  The flag makes the gate a no-op here
+            # WITHOUT altering production gate behavior (see #3911); it does
+            # not change the throughput numbers the assertions below parse.
             cmd = [
                 sys.executable,
                 "-m",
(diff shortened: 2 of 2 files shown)


rjwalters:
## Changes Requested — required CI check failing

The change itself is correct in isolation: it matches the curator's verified recommendation on #4024 exactly (add `--allow-unsafe-grid` to the class-scoped `route_stdout` fixture, plus docstring/comment). I verified:

- The flag exists on the CLI with the expected semantics (`route_cmd.py:7623`, and the #3911 gate at `route_cmd.py:8594` treats `--allow-unsafe-grid` / `--force` as the documented opt-out). The gate fires before backend selection, confirming the corrected "not backend-conditioned" diagnosis.
- The diff is confined to the test file (the `.loom-managed` change is an auto-managed worktree marker, not source). No production gate logic is touched.
- The two throughput tests are `@pytest.mark.slow` and run only in `slow-tests.yml` nightly (`-m slow`, `--timeout=1200`); the PR-blocking `Test` job runs `-m "not slow"` (ci.yml:142), so the 259s runtime does not affect the blocking job. Bypassing the DRC-safety gate for a throughput-only test whose routed output is discarded in a tempdir does not weaken the test's value.

However, I cannot approve: the required **`Test`** check (`pytest -n auto -m "not slow"`) **FAILED** (exit 1, 16m32s) on this run (29085141207). Merge state is `UNSTABLE`. Every other check passed (all board end-to-end, both routing regressions, lint, type, C++ build).

Because this PR adds `--allow-unsafe-grid` only to a `-m slow` fixture, it should have zero effect on the `-m "not slow"` Test job — so this failure is most likely a pre-existing/unrelated flake on main rather than a regression from this change. It needs to be confirmed and cleared before merge (re-run the `Test` job; if it reproduces, it is a separate main-branch issue to file, not caused by this PR).

Please re-run the failed `Test` job. Once it is green (and merge state is CLEAN), flip back to `loom:review-requested` and I'll approve.

rjwalters:
Re-review complete — APPROVED.

The prior pass verified all substance (single test-file diff adding `--allow-unsafe-grid` to the board-05 throughput fixture; #3911 auto-grid gate is a no-op only for this throughput-only fixture whose routed PCB is discarded, never manufactured/DRC-checked; flag does not perturb the parsed throughput numbers; slow-marker runtime isolation intact). The sole blocker was a red required `Test` check, diagnosed as the known perf flake `test_pathfinder_dormant_partner_perf.py::test_dormant_partner_optimized_not_slower` (timing-ratio noise, #3581 class) — unrelated to this diff.

Confirmed on head d496496 (unchanged since prior review, no new commits): the re-run of `Test` is now green (attempt 2, 16m26s pass), all 16 required checks pass, merge state CLEAN. Flake resolved by re-run; nothing material changed. Good to merge.

PR #4033, opened 2026-07-10
feat(3d): cross-library substitution tier for generic footprint model refs

rjwalters:
## Summary

`kct pcb add-3d-models` left several footprints bare after #4012 because their lib ids are generic/synthetic and the existing resolver only searches *within the same library* (exact match, then same-prefix variant). Their visual equivalents exist in the installed KiCad libraries — under a **different** library name (upstream rename) or a vendor-suffixed name. This adds an explicit, curated cross-library substitution tier so those resolve, resurfacing component bodies in the gallery renders for boards 06 and 07.

## Changes

- **New `src/kicad_tools/pcb/model_substitutions.py`**: curated `lib_id -> lib_id` table (`MODEL_SUBSTITUTIONS`) + `substitute_lib_id()` helper. Each value is a hand-verified, installed, body-compatible footprint that carries a `(model ...)` node.
- **`models3d.py` resolver tier**: `make_library_resolver` gains `allow_substitutions` and a `substitution_log`. Resolution order is now exact -> same-library variant -> **cross-library substitution**, so it can never redirect an already-resolvable footprint.
- **`ModelPatchReport.substitution_matches`**: new field parallel to `variant_matches`, distinguishing cross-library substitutions from same-library variants.
- **CLI (`pcb_add_3d_models.py`)**: surfaces the substitution bucket in both text and JSON output.
- **Regenerated artifacts**: boards 06 and 07 `output/*.kicad_pcb` (routed + non-routed) patched with the newly-resolved model refs — pure `(model ...)` insertion, zero removed bytes.
- **Tests**: new `TestCrossLibrarySubstitution` and `TestSubstitutionTable` classes cover exact-miss -> variant-miss -> table-hit, table-miss -> unresolved, `allow_substitutions=False`, variant-beats-substitution precedence, exact-never-redirected, and CLI JSON surfacing.

## Substitution table

| Requested lib_id | Substitute |
|---|---|
| `Connector_FFC:FFC_4P_0.5mm` | `Connector_FFC-FPC:Amphenol_F32Q-1A7x1-11004_1x04-1MP_P0.5mm_Horizontal` |
| `Connector_FFC:FFC_6P_1.0mm` | `Connector_FFC-FPC:TE_84952-6_1x06-1MP_P1.0mm_Horizontal` |
| `Connector_USB:USB_C_Receptacle_USB2.0` | `Connector_USB:USB_C_Receptacle_GCT_USB4105-xx-A_16P_TopMnt_Horizontal` |
[...]

PR #4012, opened 2026-07-09
fix(render): add 3D model refs to fleet footprints so gallery renders show components

rjwalters:
## Problem

The kicad-tools.org gallery 3D renders (`boards/<slug>/output/renders/3d-{front,back}.png`, produced by `kct render` → `kicad-cli pcb render`) showed nearly-bare boards — pads and copper only, no component bodies.

## Root cause

**Every footprint in every committed `boards/0*/output/*.kicad_pcb` (16 files) carries zero `(model ...)` 3D model references.** The board generators hand-emit footprint s-expressions with standard KiCad lib ids (`LED_THT:LED_D5.0mm`, `Package_QFP:LQFP-48_7x7mm_P0.5mm`, …) but never the `(model "${KICADn_3DMODEL_DIR}/...")` node that the corresponding `.kicad_mod` sources contain — so the ray-tracer has literally nothing to draw.

Failure mode (b) from the investigation (env var not resolving) was ruled out empirically: with model refs injected, this machine's kicad-cli 10.0.4 resolves `${KICAD10_3DMODEL_DIR}` (and even the legacy `${KICAD9_...}` name) with no env plumbing at all. The refs simply did not exist.

## Fixes

1. **`kct pcb add-3d-models` (new)** — `src/kicad_tools/pcb/models3d.py` + `src/kicad_tools/cli/pcb_add_3d_models.py`. Copies the canonical `(model ...)` node(s) from the installed KiCad footprint libraries into a `.kicad_pcb` as a **pure text insertion** (string-aware s-expression scanner; inserts just before each footprint's closing paren). Copper, placement, zones, nets: untouched by construction. Idempotent. Flags: `--dry-run`, `--exact`, `--lib-path`, `-o`, `--format json`.
   - Same-library **name-variant fallback** for visual lookup: e.g. `QFN-24-1EP_4x4mm_P0.5mm` (custom test footprint named after a KiCad family) gets the `..._EP2.6x2.6mm` variant's model — identical body, only the exposed pad differs. Every variant match is reported (`variant_matches`); `--exact` disables it.
2. **`run_pcb_render` env plumbing** — injects `KICAD6..KICAD10_3DMODEL_DIR` into the kicad-cli subprocess environment, auto-detected from the kicad-cli location (macOS bundle `SharedSupport/3dmodels`, Unix `share/kicad/3dmodels`) or platform defaults; user-set env vars are never overridden. This makes refs written by any KiCad release resolve under any other release and on headless CI.
[...]

PR #4045, opened 2026-07-11
fix(3d): offset canonical STEP models onto origin-centered pads by pad-centroid delta

### LINKED ISSUE #4034: 3D: add-3d-models applies canonical STEP refs with zero offset to kct-generated footprints whose pads are origin-centered (bodies misaligned by half pitch)
## Symptom

On kicad-tools.org board 00-simple-led, the 3D renders show the J1 0.1\" header body and the D1 LED body misaligned from their drilled holes (owner-reported from the live gallery). R1 (0805) is fine.

## Root cause

PR #4012's `kct pcb add-3d-models` patches canonical KiCad-library `(model ...)` refs with `(offset (xyz 0 0 0))` onto footprints matched **by name**. But the fleet's kct-generated footprints reuse canonical names while using a different origin convention:

- Board 00 `J1` = `Connector_PinHeader_2.54mm:PinHeader_1x02_P2.54mm_Vertical`: pads at `(0, -1.27)` / `(0, +1.27)` (centered on origin). The real KiCad footprint has pad 1 at `(0,0)`, pad 2 at `(0, 2.54)` — and the STEP model is aligned to *that* origin. Result: body shifted by 1.27mm (half pitch) relative to the holes.
- Board 00 `D1` = `LED_THT:LED_D5.0mm`: pads at `(-1.27, 0)` / `(+1.27, 0)`. Real footprint: pad 1 `(0,0)`, pad 2 `(2.54, 0)`, lens centered at `(1.27, 0)`. Same half-pitch shift.
- `R_0805_2012Metric` renders correctly because the canonical footprint is also origin-centered — the convention mismatch only bites where the KiCad library uses pad-1-at-origin (most THT parts).

Evidence: `boards/00-simple-led/output/simple_led_routed.kicad_pcb` footprint blocks for J1/D1 (zero-offset `(model ...)` + centered pads).

## Scope

Likely fleet-wide for any kct-generated THT footprint sharing a canonical name: audit all 16 boards patched in #4012 for the same convention mismatch (headers, THT LEDs, TO-92, etc.), not just board 00.

## Fix directions (pick after measuring)

1. **Compute the offset at patch time**: `add-3d-models` compares the target footprint's actual pad-1 position against the canonical library footprint's pad-1 position (libraries are installed and readable — models3d.py already resolves them) and writes the delta into `(offset (xyz ...))`. Robust for arbitrary origin conventions, keeps the pure-insertion guarantee (offset is model metadata only; no copper/DRC delta).
2. Alternatively, regenerate kct footprints to adopt KiCad's pad-1-at-origin convention — much bigger blast radius (copper moves); NOT recommended as part of this issue.

## Acceptance criteria

- Board 00 renders show J1 header and D1 LED bodies centered on their holes (front + back renders).
- Fleet audit: every #4012-patched board re-checked; all THT model offsets correct or explicitly waived.
- Pure-insertion guard still holds (no copper/DRC/netlist deltas; only `(model ...)` metadata changes).
- Regression test covering the origin-convention offset computation (e.g. a synthetic footprint with centered pads vs a library footprint with pad-1-at-origin).
- Committed artifacts + renders regenerated per artifact-first convention; gallery re-verified.

## Coordination note

Issue #4014 (in flight) is adding a cross-library substitution tier to the same resolver (`src/kicad_tools/pcb/models3d.py`). Land this after #4014 merges and rebase; the offset computation must apply to substitution matches too.

Refs: #4012 (model insertion tooling), #4014 (substitution tier, in flight), owner report 2026-07-09 with gallery screenshot.

rjwalters:
## Summary

`kct pcb add-3d-models` copied library `(model ...)` refs with a zero offset
onto kct-generated footprints that reuse a canonical KiCad name while placing
pads on an **origin-centered** convention (pads symmetric about the origin)
instead of KiCad's **pad-1-at-origin**. Since a STEP body is authored in the
library footprint's local frame, a zero-offset copy left THT bodies shifted
by half the pad pitch relative to their drilled holes — board 00's J1 header
and D1 LED were reported misaligned on the live gallery, and the same bug hit
headers, DIP, TO-92/220/263, and crystals fleet-wide.

Before inserting each `(model ...)` node the patcher now adds the
`target_centroid - source_centroid` pad-centroid delta into its
`(offset (xyz ...))`, expressed in the KiCad 3D-model frame (X follows the
footprint X, **Y negated** vs the footprint 2D frame, Z unchanged).

## Changes

- `src/kicad_tools/pcb/models3d.py`:
  - `_pad_anchor()` — pad-centroid of a footprint block (string-aware scan).
    Centroid (not pad 1) is the anchor so parts sharing a center but not a
    pitch (e.g. `R_0805`, board pads at ±1.0 vs library ±0.9125) compute a
    zero delta and insert verbatim.
  - `_apply_offset_delta()` — rewrites only the model's own `(offset (xyz ...))`
    x/y numbers, applying the Y-flip and a 1nm tolerance so already-correct
    footprints stay byte-identical.
  - `ResolvedModels` — the resolver now returns the source `.kicad_mod`'s
    centroid alongside its model texts. `add_model_refs_to_text` computes the
    delta from the **target** footprint's own pads, so all three resolution
    tiers (exact / same-library variant / #4033 cross-library substitution)
    get the offset uniformly — including substitution matches where the
    substitute is a different physical part. Legacy `list[str]` resolvers are
    still accepted (zero offset).
- `src/kicad_tools/cli/pcb_add_3d_models.py` — docstring documents the offset.
- `tests/test_pcb_add_3d_models.py` — offset regression coverage (see below).
- `boards/*/output/*.kicad_pcb` (16 artifacts) — regenerated with corrected
  offsets (artifact-first convention).

| Board 00 J1/D1 bodies centered on holes | ✅ | Top-down `kicad-cli pcb render` before/after: zero-offset shows J1 body shifted up + D1 lens off its pads; fixed shows both centered. Sign confirmed empirically. |
| Fleet audit of all #4012-patched boards | ✅ | Re-ran across all 16 `boards/*/output/*.kicad_pcb`; 242 footprints re-patched; every non-zero offset inspected by real block boundaries and is a sensible half-span shift (headers, DIP-8, TO-220/263, crystals, USB-C, HDMI, FFC). |
| Pure-insertion guard holds | ✅ | All pad/segment/via/zone/net/at/size/drill/layer lines byte-identical to HEAD across all 16 artifacts; board 00 `kicad-cli pcb drc` = 0 violations. |
| Regression test for offset computation | ✅ | `TestOriginConventionOffset` (synthetic origin-centered header vs pad-1-at-origin library), `TestPadAnchorHelper`, `TestApplyOffsetDelta` (Y-flip, zero/subnanometre verbatim), plus a substitution-tier case. |
| Applies to substitution matches (#4033) | ✅ | `test_offset_applies_through_substitution_tier` asserts the delta keyed off the target's own centroid; boards 06/07 render with centered substitute bodies. |

## Test Plan

- `uv run pytest tests/test_pcb_add_3d_models.py tests/test_render_cmd.py` — 73 passed.
- `uv run ruff check` + `ruff format --check` on touched files — clean.
- `uv run mypy src/kicad_tools/pcb/models3d.py` — no issues.
- Manual: rendered boards 00/02/05/07 front + back; verified bodies land on pads.

## Deferred

- Model **rotation** for parts whose kct footprint is rotated relative to the
  library origin is out of scope: this fleet's pad-1 rotations are 0, the delta
  is computed in the unrotated local frame, and rotation would risk copper-free
  but visually different diffs. Filed as a note for a follow-up if a rotated THT
  part appears.
- Surfacing per-footprint offset values in the CLI report JSON was evaluated
  (curator's "check if") but not added: the render is the audit mechanism and
  the report schema stays minimal (scope discipline).

Closes #4034

### DIFF
--- a/boards/00-simple-led/output/simple_led.kicad_pcb
+++ b/boards/00-simple-led/output/simple_led.kicad_pcb
@@ -51,7 +51,7 @@
     (pad "2" thru_hole oval (at 0 1.270) (size 1.7 1.7) (drill 1.0) (layers "*.Cu" "*.Mask") (net 3 "GND"))
   	(model "${KICAD10_3DMODEL_DIR}/Connector_PinHeader_2.54mm.3dshapes/PinHeader_1x02_P2.54mm_Vertical.step"
   		(offset
-  			(xyz 0 0 0)
+  			(xyz 0 1.27 0)
   		)
   		(scale
   			(xyz 1 1 1)
@@ -99,7 +99,7 @@
     (pad "2" thru_hole circle (at 1.270 0 0) (size 1.8 1.8) (drill 0.9) (layers "*.Cu" "*.Mask") (net 2 "LED_ANODE"))
   	(model "${KICAD10_3DMODEL_DIR}/LED_THT.3dshapes/LED_D5.0mm.step"
   		(offset
-  			(xyz 0 0 0)
+  			(xyz -1.27 0 0)
   		)
   		(scale
   			(xyz 1 1 1)
--- a/boards/00-simple-led/output/simple_led_routed.kicad_pcb
+++ b/boards/00-simple-led/output/simple_led_routed.kicad_pcb
@@ -226,7 +226,7 @@
 		(embedded_fonts no)
 		(model "${KICAD10_3DMODEL_DIR}/Connector_PinHeader_2.54mm.3dshapes/PinHeader_1x02_P2.54mm_Vertical.step"
 			(offset
-				(xyz 0 0 0)
+				(xyz 0 1.27 0)
 			)
 			(scale
 				(xyz 1 1 1)
@@ -306,7 +306,7 @@
 		(embedded_fonts no)
 		(model "${KICAD10_3DMODEL_DIR}/LED_THT.3dshapes/LED_D5.0mm.step"
 			(offset
-				(xyz 0 0 0)
+				(xyz -1.27 0 0)
 			)
 			(scale
 				(xyz 1 1 1)
--- a/boards/01-voltage-divider/output/voltage_divider.kicad_pcb
+++ b/boards/01-voltage-divider/output/voltage_divider.kicad_pcb
@@ -51,7 +51,7 @@
     (pad "2" thru_hole oval (at 0 1.270) (size 1.7 1.7) (drill 1.0) (layers "*.Cu" "*.Mask") (net 3 "GND"))
   	(model "${KICAD10_3DMODEL_DIR}/Connector_PinHeader_2.54mm.3dshapes/PinHeader_1x02_P2.54mm_Vertical.step"
   		(offset
-  			(xyz 0 0 0)
+  			(xyz 0 1.27 0)
   		)
   		(scale
   			(xyz 1 1 1)
@@ -123,7 +123,7 @@
     (pad "2" thru_hole oval (at 0 1.270) (size 1.7 1.7) (drill 1.0) (layers "*.Cu" "*.Mask") (net 3 "GND"))
   	(model "${KICAD10_3DMODEL_DIR}/Connector_PinHeader_2.54mm.3dshapes/PinHeader_1x02_P2.54mm_Vertical.step"
   		(offset
-  			(xyz 0 0 0)
+  			(xyz 0 1.27 0)
   		)
   		(scale
   			(xyz 1 1 1)
--- a/boards/01-voltage-divider/output/voltage_divider_routed.kicad_pcb
+++ b/boards/01-voltage-divider/output/voltage_divider_routed.kicad_pcb
@@ -147,7 +147,7 @@
 		(embedded_fonts no)
 		(model "${KICAD10_3DMODEL_DIR}/Connector_PinHeader_2.54mm.3dshapes/PinHeader_1x02_P2.54mm_Vertical.step"
 			(offset
-				(xyz 0 0 0)
+				(xyz 0 1.27 0)
 			)
 			(scale
 				(xyz 1 1 1)
@@ -305,7 +305,7 @@
 		(embedded_fonts no)
 		(model "${KICAD10_3DMODEL_DIR}/Connector_PinHeader_2.54mm.3dshapes/PinHeader_1x02_P2.54mm_Vertical.step"
 			(offset
-				(xyz 0 0 0)
+				(xyz 0 1.27 0)
 			)
 			(scale
 				(xyz 1 1 1)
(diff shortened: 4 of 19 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the pad-centroid origin-convention offset for canonical STEP models. This is clean, well-scoped work.

## Verification performed

**Centroid-vs-pad-1 reasoning (the key design question) — sound.**
The offset is a pure translation `target_centroid - source_centroid`. When the board footprint reuses the library's pad layout on a shifted origin (the actual bug), a translation preserves the centroid difference regardless of pad asymmetry — so centroid delta *equals* pad-1 delta *equals* any-pad delta. I checked the asymmetric multi-pad cases flagged as the risk:
- TO-220-3 (board 05): `(xyz -2.54 0 0)` — pad-1-at-origin library centroid at pin 2 (2.54), board origin-centered (0) → correct.
- PinHeader_1x09 (board 07): recomputed live — board centroid (0,0), library centroid (0,10.16) → `(xyz 0 10.16 0)`. Exact.
- PinHeader family scales as (N-1)·pitch/2 (1x03→2.54, 1x05→5.08, 1x06→6.35, 1x02→1.27). Consistent.

Centroid is also the *right* choice over pad-1 for the R_0805 case (same center, different pitch → zero delta, verbatim insert), which pad-1 would have perturbed. The `test_matched_centroid_gets_zero_offset` case locks this in.

**Pure-insertion guarantee — holds.**
Diffed all 16 `boards/*/output/*.kicad_pcb` against main: every changed line is model metadata (`model`/`offset`/`scale`/`rotate`/`xyz`/parens/`embedded_fonts`). Zero copper/pad/via/zone/net/at/size/drill/layer deltas. Board 07 routed additionally reformats a few pre-existing single-line model blocks to multi-line (cosmetic, from re-emission) — no geometry impact; the model *set* is byte-identical to main.

**Substitution tier (#4033) — correctly wired.**
Offset applies uniformly across all three resolver tiers, keyed off the *target* footprint's own centroid. For cross-library substitutions the substitute is a different physical part with no guaranteed pad-origin relationship, so this is a best-effort registration heuristic — acceptable given the render is the documented audit mechanism and the alternative (no offset) is strictly worse. `test_offset_applies_through_substitution_tier` pins the target-keyed behavior.

**Y-flip and tolerance — correct.** Board 00 J1 (0,-1.27 delta → `(xyz 0 1.27 0)`) and D1 confirm the model-frame Y negation. 1nm tolerance keeps FP-noise footprints byte-identical.

## Tests / CI
- `tests/test_pcb_add_3d_models.py`: 39 passed (up from 24). `+test_render_cmd.py`: 73 total, all pass.
- ruff check + format-check + mypy on touched files: clean.
- Legacy `list[str]` resolver back-compat preserved (`test_legacy_list_resolver_still_supported`).
- All 16 GitHub CI checks pass; merge state CLEAN.

The deferred model-rotation scope note is the right call (fleet pad-1 rotations are 0; rotation risks copper-free-but-visually-different diffs).

## Test Execution
1. `uv run pytest tests/test_pcb_add_3d_models.py tests/test_render_cmd.py` — ✅ 73 passed
2. `ruff check` / `ruff format --check` on touched files — ✅ clean
3. `mypy src/kicad_tools/pcb/models3d.py` — ✅ no issues
4. Manual render verification — ⚠️ observation-only (not executed); offset sign verified mathematically instead

PR #4047, opened 2026-07-11
fix(board03): fail fast on a partial route instead of a misleading LVS trace

### LINKED ISSUE #4027: flaky: test_board_03_copper_lvs fresh-regen fails under concurrent load, passes in isolation

rjwalters:
## Summary

Board 03's `generate_design.py` regen (`test_board_03_copper_lvs.py::TestBoard03FreshRegenCopperLVSClean`) flaked under concurrent load twice on 2026-07-09, passing on quiet re-runs. The failure surfaced as a *copper-LVS / GND-stitching* error, which misdirected reviewers to the wrong subsystem.

Per the curator's verified root-cause analysis: the per-net A* budget is already load-independent (`--deterministic-budget`, #3538/#3799). The flake lives in the **outer `--timeout 600` wall-clock safety backstop** — under concurrent CPU contention 600 real seconds buys fewer CPU cycles, so the escalation loop can bail with fewer nets landed. `main()` captured `route_success` but **never checked it**, so a partial route fell through to `add_gnd_stitching_vias` → `fill_zones_in_routed_pcb` → `write_lvs_report(require_clean=True)`, where the unrouted signal net reads as a GND copper open and raises `BoardNetlistMismatch` — reported by the broad `except` as a misleading exit 1.

This implements the curator's recommended **option (c), narrowly scoped**: fail fast on a partial route with a distinct message before the pipeline reaches the LVS gate. No change to the router internals or the outer `--timeout` safety-valve design.

## Changes

- `boards/03-usb-joystick/generate_design.py`: add a `route_success` gate in `main()` immediately after `route_pcb()` returns. On a partial route it raises a distinct `RuntimeError` ("partial route -- likely wall-clock budget exhaustion under load ...") **before** GND stitching, zone fill, or `write_lvs_report` run. The message explicitly disclaims a copper-LVS/GND-stitching cause and points at the `--timeout` backstop.
- `tests/test_board_03_copper_lvs.py`: add hermetic `TestBoard03PartialRouteFastFail` (monkeypatches the recipe's module-level functions; no router / `kicad-cli` / LVS invocation):
  - `test_partial_route_fails_fast_with_distinct_message` — `route_pcb` → `False` makes `main()` exit non-zero with a "partial route" / "wall-clock budget" message, and forbidden downstream steps (`add_gnd_stitching_vias`, `fill_zones_in_routed_pcb`, `run_drc`, `write_lvs_report`) raise if the gate lets them run; asserts no `BoardNetlistMismatch` in the output.
  - `test_full_route_still_reaches_lvs_gate` — `route_pcb` → `True` still reaches `write_lvs_report` (no false positive on the happy path).
  - Refactored the module loader into `_load_board03_module()` (reused by the existing `_load_add_gnd_stitching_vias`).

| `main()` checks routing result after `route_pcb()`; partial route exits distinctly without reaching stitching/fill/LVS | ✅ | New gate at `generate_design.py:838`; `test_partial_route_fails_fast_with_distinct_message` proves forbidden downstream steps never run |
| Failure message distinguishes "partial route (wall-clock budget)" from "copper-LVS dirty (GND stitching)" | ✅ | Message text + test asserts `"partial route"`/`"wall-clock budget"` present and `"BoardNetlistMismatch"` absent |
| Test covers the partial-route path (force `route_success=False`) | ✅ | `TestBoard03PartialRouteFastFail` monkeypatches `route_pcb` → `False` |
| No change to the per-net A* budget mechanism | ✅ | Only board recipe + test touched; router internals unchanged |
| Existing clean-regen behavior unaffected | ✅ | `test_fresh_regen_is_copper_lvs_clean` passes isolated (111.81s); `test_full_route_still_reaches_lvs_gate` proves the gate doesn't fire on N==M |

## Test Plan

- `uv run pytest tests/test_board_03_copper_lvs.py -m "not slow"` → 6 passed (2 new).
- `uv run pytest tests/test_board_03_copper_lvs.py::TestBoard03FreshRegenCopperLVSClean::test_fresh_regen_is_copper_lvs_clean` → 1 passed in 111.81s (isolated, quiet machine — happy path unaffected).
- `uv run ruff check` + `ruff format --check` on both touched files → clean.
- C++ router backend built in the worktree (`kct build-native`, available 1.0.0) so the slow regen ran at native speed.

## Note: siblings share the latent pattern (deferred, out of scope)

Boards 00/01/02/04/06 (`generate_design.py`) also capture `route_success` from `route_pcb()` and then run `write_lvs_report(require_clean=True)` unconditionally, so they carry the same misattribution risk if their route ever truncates under load. This PR is scoped to board 03 (the board that actually flaked) per the issue. If the same flake appears on a sibling, the identical gate applies — worth a follow-up issue rather than expanding this PR's blast radius.

Closes #4027

### DIFF
--- a/boards/03-usb-joystick/generate_design.py
+++ b/boards/03-usb-joystick/generate_design.py
@@ -818,6 +818,35 @@ def main() -> int:
         routed_path = output_dir / "usb_joystick_routed.kicad_pcb"
         route_success = route_pcb(pcb_path, routed_path)
 
+        # Step 5.3: Fast-fail gate on a PARTIAL route (#4027).  ``route_pcb``
+        # returns ``False`` when fewer than all signal nets landed.  Under
+        # concurrent CPU load the ``--timeout 600`` wall-clock SAFETY backstop
+        # (route_cmd._set_wall_clock_deadline / _remaining_budget) can fire
+        # BEFORE the outer escalation loop routes every net -- 600 real seconds
+        # buys fewer CPU cycles on a loaded machine.  (The per-net A* search is
+        # already load-independent via --deterministic-budget, #3538/#3799, so
+        # the flake lives in the OUTER wall-clock loop, not the router
+        # internals.)  If we fall through, the downstream ``add_gnd_stitching_
+        # vias`` -> ``fill_zones_in_routed_pcb`` -> ``write_lvs_report(
+        # require_clean=True)`` sees a genuinely unrouted signal net as a copper
+        # OPEN and raises ``BoardNetlistMismatch``, which the broad ``except``
+        # below reports as exit 1 -- misdirecting the reviewer to the GND
+        # stitching / LVS subsystem when the true cause is upstream route
+        # truncation.  Raise a DISTINCT, clearly-worded error here instead so
+        # the failure names the real cause.  The "PARTIAL: Routed N/M signal
+        # nets" line is already printed above by ``route_pcb``.
+        if not route_success:
+            raise RuntimeError(
+                "partial route -- likely wall-clock budget exhaustion under "
+                "load (the --timeout 600 safety backstop fired before every "
+                "signal net landed; see the 'PARTIAL: Routed N/M signal nets' "
+                "line above for the exact count). This is NOT a copper-LVS / "
+                "GND-stitching failure -- the pipeline stopped before the LVS "
+                "gate. Re-run boards/03-usb-joystick/generate_design.py in "
+                "isolation on a quiet machine, or raise the --timeout in "
+                "route_pcb() if this recurs on an unloaded host."
+            )
+
         # Step 5.4: Add GND F.Cu<->B.Cu stitching vias (#3787).  The board
         # has GND planes on both copper layers but the router emits zero GND
         # vias, so the J1 USB-C F.Cu-only shield pads strand in their own
--- a/tests/test_board_03_copper_lvs.py
+++ b/tests/test_board_03_copper_lvs.py
@@ -114,8 +114,8 @@ class TestBoard03CopperLVSClean:
         )
 
 
-def _load_add_gnd_stitching_vias():
-    """Import ``add_gnd_stitching_vias`` from the board-03 recipe module."""
+def _load_board03_module():
+    """Import the board-03 ``generate_design.py`` recipe module."""
     import importlib.util
 
     gen = BOARD_DIR / "generate_design.py"
@@ -123,7 +123,12 @@ def _load_add_gnd_stitching_vias():
     assert spec is not None and spec.loader is not None
     module = importlib.util.module_from_spec(spec)
     spec.loader.exec_module(module)
-    return module.add_gnd_stitching_vias
+    return module
+
+
+def _load_add_gnd_stitching_vias():
+    """Import ``add_gnd_stitching_vias`` from the board-03 recipe module."""
+    return _load_board03_module().add_gnd_stitching_vias
 
 
 class TestBoard03UnconditionalUsbcStitch:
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Verified the fix against the diff, the recipe implementation, and a local run of the fast tests.

## What I checked

**Gate placement is correct.** The `route_success` check sits at `boards/03-usb-joystick/generate_design.py:838`, immediately after `route_pcb()` returns (line 819) and *before* `add_gnd_stitching_vias` (856), `fill_zones_in_routed_pcb` (862), `run_drc` (865), and `write_lvs_report(require_clean=True)` (881). On a partial route it raises a distinct `RuntimeError`, which the broad `except` at line 920 renders to stderr (`Error: {e}`) and returns 1 — so the failure now names route truncation instead of `BoardNetlistMismatch`.

**`route_pcb`'s return value actually means what the gate assumes.** Read the implementation (lines 491-668): it returns `False` on `kct route` exit code 1/5, on unparseable "Nets routed: N/M" output, and whenever `routed != total`; it returns `True` only when `routed == total`. So `not route_success` is a faithful "partial route" signal — the gate cannot fire on a complete route, and cannot be silently bypassed by a partial one.

**Tests pin the behavior.** `TestBoard03PartialRouteFastFail` is genuinely hermetic (monkeypatches the recipe's own module-level functions; no router / kicad-cli / LVS invocation). `_forbid_downstream` makes every post-gate step raise if reached, and the partial-route test additionally asserts stderr contains `partial route` + `wall-clock budget` and does NOT contain `BoardNetlistMismatch`. The `AssertionError` from a forbidden step would also be caught by the broad `except`, but its message lacks those tokens, so the stderr assertions still fail the test correctly if the gate ever regresses. `test_full_route_still_reaches_lvs_gate` proves no false positive on the happy path (asserts `write_lvs_report` is the step that runs, rc == 0). The `_load_board03_module()` refactor is clean and reused by the existing loader.

**Scope is justified.** Confirmed boards 00/01/02/04/06 all capture `route_success` and then call `write_lvs_report` unconditionally — they share the identical latent pattern the PR body documents as a deferred follow-up. Narrowing to board 03 (the board that actually flaked) matches the acceptance criteria and keeps blast radius minimal. The router internals and the `--timeout 600` safety valve are untouched, as intended.

## Test Execution

- `uv run pytest tests/test_board_03_copper_lvs.py -m "not slow"` — ✅ 6 passed, 1 deselected (19.4s) locally (2 new tests).
- `uv run ruff check` + `ruff format --check` on both touched files — ✅ clean.
- Slow fresh-regen (`TestBoard03FreshRegenCopperLVSClean`) — ⏭️ Skipped locally (2-11 min); covered by the Board 03 End-to-End CI job, which passed (4m22s).

## CI

All 16 checks pass; merge state **CLEAN**. Notably Board 03 End-to-End (LVS-only) passed at 4m22s, and the full Test job passed at 16m21s.

Clear, well-documented, correctly scoped fix that converts a misleading LVS trace into an accurate diagnostic. Nice work.

PR #4061, opened 2026-07-12
fix(diffpair): stop the impedance-coupling gap from freezing the coupled search (#4052)

### LINKED ISSUE #4052: Board 07 Phase 2: diagnose + fix CoupledPathfinder per-pair budget exhaustion (--differential-pairs viable)
**Epic:** #4049 (Phase 2) · Related: #3438, #3439, #3547, #3987, #3921, #3508

## Context (verified on `main`, 2026-07-11)

Phase 1 is now complete and **both siblings measured negative**:
- #4050 (PR #4059, open): `--targeted-ripup` × `--max-ripups-per-net` (3/5/8/12)
  all converge to **10/11** on the DDR-bundle repro with `DQ3` stranded. No
  improvement over baseline at any budget.
- #4051 (PR #4060, open): reactive escape-freedom byte-lane reorder
  **regresses** 10/11 → **9/11** (DQ2+DQ3 stranded). Shipped opt-in and
  OFF by default so it doesn't regress `main`.

So Phase 2 (this issue) is now **the last unmeasured lever** before Phase 3
(#4053, the decision-gated river-routing planner) is the only option left.
A working `--differential-pairs` also has outsized leverage on this board:
3 of board 07's 5 LVS opens are diff-pair legs (`MIPI_DAT0_N`, `TMDS_D0_N`,
`TMDS_D1_N`), and **all 8** blocking DRC quality errors are diffpair
skew/continuity on the OTHER 4 pairs that already route fully single-ended
(`DQS`, `MIPI_CLK`, `MIPI_DAT1`, `TMDS_D2`). A working coupled router
addresses both gaps at once, even though this issue's acceptance bar
(below) only requires one pair.

## Prior diagnosis already on record — read this before profiling

This exact question was investigated once before, in #3439 (closed
2026-06-10, folded into the #3414 recovery). Do not re-derive from
scratch; the code comments at
`src/kicad_tools/router/diffpair_routing.py:4619-4633` (and board 07's
own recipe comment at
`boards/07-matchgroup-test/generate_design.py:1511-1513`) already record:

> `CoupledPathfinder.route_coupled` wall-clock budget exceeded after
> 60.00s (**14016 iterations**); abandoning search
> (`p_net='TMDS_D0_P' n_net='TMDS_D0_N'`)

i.e. **~14,000 iterations in 60 seconds** on board 07's 4-layer,
110×95mm grid — that is pure-Python `heapq`-based A* speed, not
C++-accelerated speed (see finding (a) below).

### (a) CoupledPathfinder has NO C++ path — confirmed by import audit

`src/kicad_tools/router/diffpair_routing.py` never imports `router_cpp`
or `cpp_backend` (`grep -n "router_cpp\|cpp_backend" diffpair_routing.py`
returns zero hits). The entire `CoupledPathfinder.route_coupled` loop
(`diffpair_routing.py:1460-1900`, dispatch loop starting ~line 1692) is
pure Python: `heapq.heappush`/`heappop` on a `CoupledNode` priority
queue, with per-neighbor blocked/spacing/clearance checks executed in
Python for every popped state.

By contrast, single-ended per-net A* **does** have a C++ path —
`src/kicad_tools/router/cpp_backend.py` wraps `router_cpp.Pathfinder`
(built from `src/kicad_tools/router/cpp/src/pathfinder.cpp` +
`grid.cpp`), and that 10-100x speedup is exactly why `kct build-native`
matters for single-ended routing (per `CLAUDE.md`). **The coupled
diff-pair search gets none of that speedup.** This is very likely the
single largest contributor to the 60s/pair budget exhaustion: a
same-order-of-magnitude single-ended search that would take ~0.5-1s in
C++ takes tens of seconds in pure Python, and the coupled search visits
a larger state space per step (see (b)) on top of that base slowdown.

### (b) Search-space structure: joint product-space A*, not two A* searches

`CoupledPathfinder` does not run P and N as two independent A*
searches. It searches a single **joint state** `CoupledState(p_pos,
n_pos, direction)` (`diffpair_routing.py:340`) — i.e. the search node is
a *pair* of grid positions (plus a shared travel direction to bias
straight-vs-turn costs), not one position. Neighbor generation
(`diffpair_routing.py:1072`, `1188`) enumerates a cross-product of moves
for each head. `self.directions` (`diffpair_routing.py:733`) is only 4
cardinal moves per head, but because both heads move (semi-)independently
per step (symmetric moves + asymmetric "converge" moves, see the
`heuristic_mode="partner_aware"` docstring at `diffpair_routing.py:566`
and the "asymmetric-escape pathology" reference at `:596`), the reachable
joint-state space is much larger than either single-net space alone —
effectively bounded by `(cols × rows)²` in the worst case rather than
`cols × rows`. The class's own comment (`diffpair_routing.py:686-692`,
Issue #3508) documents a measured basin-flooding pathology: "a single
`cost_turn` = 5 shell can hold ~90k states on board 06's 0.05mm grid —
measured: MIPI_CLK needed ~95k iterations to flood ONE such basin before
escaping." That is the concrete mechanism the 60s/14k-iteration budget
gates: **iteration count consumed by joint-state basin-flooding under an
inadmissible-in-practice (though formally admissible) heuristic
plateau**, not raw per-iteration Python overhead alone. Weighted-A*
(`heuristic_weight` param, `diffpair_routing.py:626-644`, #3508) and the
`partner_aware` heuristic mode were both added specifically to fight this
plateau problem and are already the default — they did not fully solve
it (see (c)).

### (c) What the 60s budget actually gates today — corridor mode already shipped, still insufficient

A partial mitigation from #3439 already exists and is **on by default**:
`route_differential_pair_coupled` (`diffpair_routing.py:4355`) first
routes the P leg single-ended (`_single_ended_guide_route`, which DOES
use the C++-accelerated per-net pathfinder), dilates that path into a
spatial corridor (`build_corridor_mask`, `diffpair_routing.py:446`), and
restricts the coupled search to that corridor — converting the open 2D
product-space search into a near-1D one. Per the
`generate_design.py:1560-1573` addendum, this measurably helped:
`DQS` completes coupled in ~4s (vs. blowing the 60s budget at 14k
iterations before), but is then *rejected* for an unrelated reason
(#3320 polarity-swap centerline overlap). The 3 TMDS pairs **still
budget-exit even with the corridor** — likely because their single-ended
P-side probe itself doesn't yield a clean-enough guide within its probe
slice (`probe_timeout = per_pair_timeout * 0.125` when
`enable_shadow_construction` is off, `diffpair_routing.py:4672-4679`),
or the corridor is wide enough that basin-flooding still dominates
inside it. There's also a newer `enable_shadow_construction` path
(#3547/#3987/#3921, geometric parallel-offset construction instead of
search) but it **defaults to `False`**
(`src/kicad_tools/router/diffpair.py:669`) and its own field notes
describe non-artifact-quality committed geometry (stranded shadow tails,
via-on-partner intersections) — so board 07 today exercises corridor
mode (if a guide exists) falling back to the legacy open joint-state
search (if it doesn't), never the shadow path.

**Bottom line for the builder**: the 60s wall-clock budget
(`timeout_seconds` → `deadline` check every 64 iterations,
`_TIMEOUT_CHECK_INTERVAL = 64`, `diffpair_routing.py:1730-1744`) and the
14016-iteration figure it produces on TMDS_D0 are gating a pure-Python
joint-state A* that (1) has no C++ acceleration at all for the coupled
loop itself, and (2) is structurally prone to large-shell basin-flooding
in the corridor-less case. Determine which of these two dominates for
the specific TMDS_D2 profile before proposing a fix — the fix is very
different depending on the answer (C++ port vs. algorithmic/corridor
tuning).

## Task — diagnose FIRST, then fix-or-writeup

1. **Confirm C++ backend is built** (`uv run kct build-native --check` →
   `available`) before profiling — a missing `.so` would only affect the
   single-ended guide-route probe inside the corridor path, but rule it
   out first per `CLAUDE.md`.
2. **Profile ONE board-07 pair**: `TMDS_D2` (suggested — it is the
   shortest of the 4 fully-single-ended-but-failing-quality pairs
   `DQS`/`MIPI_CLK`/`MIPI_DAT1`/`TMDS_D2`, per the epic's measured skew
   table, `TMDS_D2_N` skew 3.783mm vs 0.075mm tol — the smallest gap of
   the four, so likely the cheapest coupled search to converge or to
   profile to completion). Use `py-spy record` (sampling, safe to run
   against a live `kct route --differential-pairs` invocation) or
   `cProfile`/`pstats` wrapped around a minimal repro script that calls
   `route_differential_pair_coupled` directly for just this one pair
   (cheaper than a full board 07 route). Capture:
   - Wall-clock split between the single-ended guide-route probe
     (C++-backed) vs. the coupled joint-state loop (pure Python).
   - Hot functions inside the coupled loop (neighbor generation,
     heuristic evaluation, heap push/pop, blocked-cell checks).
   - `last_iterations`, `last_timeout_exceeded`,
     `last_iteration_limited` (#3921 budget-source diagnostic — read
     directly off the `CoupledPathfinder` instance after the call) to
     confirm which of `max_iterations_budget` / `timeout_seconds` binds
     for this pair, and `last_best_progress` /
     `last_rejections` (#3508 diagnostics) to see whether the search is
     "almost converged, budget-starved" or "structurally stuck"
     (basin-flooded).
   Keep this to a single pair / light sampling run — do not run a full
   board-07 routing sweep or multi-pair budget matrix as part of
   curation-adjacent profiling; that is the builder's job with a proper
   time budget.
3. **Fix fork** based on what the profile shows:
   - **If pure-Python overhead dominates** (i.e. iterations are cheap
     individually but there are simply too many of them, or profiling
     shows most wall-clock in per-iteration Python bookkeeping rather
     than basin-flooding): port the coupled joint-state A* inner loop
     to the C++ backend, mirroring the single-net integration in
     `src/kicad_tools/router/cpp_backend.py` /
     `src/kicad_tools/router/cpp/src/pathfinder.cpp`. This is the
     larger of the two options — budget it as a follow-up issue if it
     doesn't fit this issue's scope, and post a concrete design sketch
     here instead (what joint-state representation the C++ struct needs,
     how corridor masking maps to the existing `Grid3D`/`PadBounds`
     C++ types, what Python-side API `route_coupled` would call).
   - **If basin-flooding / cost-space pathology dominates** (iterations
     are individually cheap and fast, but the search visits ~90k+ states
     per plateau shell as the #3508 comment describes, and
     `last_best_progress` shows the search made real progress before
     timing out): this is the `heuristic_weight` / `partner_aware`
     mitigations already having failed to fully solve #3508's own
     problem on THIS pair. Document the pathology precisely (shell size,
     iteration count to escape it, whether raising
     `heuristic_weight` above its current default of `1.0` in
     `route_differential_pair_coupled`'s constructor call trades away
     admissibility for enough speed to matter) and post a concrete,
     scoped fix proposal — do not attempt an open-ended heuristic
     research spike inside this issue.
   - **Either way**: check whether TMDS_D2 fails at the corridor stage
     (probe doesn't produce a usable guide) or at the open-search
     fallback stage (corridor exists but the coupled search inside it
     still blows budget) — this determines whether the fix belongs in
     `_single_ended_guide_route`'s probe budget/quality, in
     `build_corridor_mask`'s radius, or in `route_coupled` itself.

## Acceptance (per epic #4049)

- At least one board-07 pair (`DQS` / `MIPI_CLK` / `MIPI_DAT1` /
  `TMDS_D2`) routes **coupled within budget** via
  `--differential-pairs`, with skew <= its declared tolerance
  (`diffpair_length_skew`) and routing continuity >= 85%
  (`diffpair_routing_continuity`), verified via `kct check`, **OR**
- a root-cause writeup (Python-vs-C++ finding + basin-flooding vs.
  overhead determination for the profiled pair) with a concrete, scoped
  fix proposal is posted on this issue for a follow-up to implement.

This issue is **diagnose-first and sweepable** (do not mark
`loom:blocked` merely because the root cause turns out to require a
larger C++ port — post the writeup and let a human decide whether to
spin off a follow-up issue, matching how #4051 shipped its finding
opt-in-off rather than blocking).

## Affected Files

- `src/kicad_tools/router/diffpair_routing.py` — `CoupledPathfinder`
  class (~line 527), `route_coupled` (~line 1460), corridor-mode
  dispatch in `route_differential_pair_coupled` (~line 4355-4900).
  Primary profiling target and most likely fix location.
- `src/kicad_tools/router/cpp_backend.py` — existing single-net C++
  integration pattern to mirror if a C++ port is proposed.
- `src/kicad_tools/router/cpp/src/pathfinder.cpp`,
  `src/kicad_tools/router/cpp/include/pathfinder.hpp` — C++ pathfinder
  implementation; reference for what a coupled-search port would extend.
- `src/kicad_tools/router/diffpair.py` — `DifferentialPairConfig`
  (`per_pair_timeout`, `per_pair_max_iterations`,
  `enable_shadow_construction` defaults, ~line 553-670).
- `boards/07-matchgroup-test/generate_design.py` — recipe comments
  documenting prior `--differential-pairs` measurements (~line
  1470-1580); update if this issue changes the recipe's diff-pair flag
  usage (out of scope unless a pair reaches acceptance).
- `src/kicad_tools/cli/route_cmd.py`,
  `src/kicad_tools/cli/parser.py` — `--diffpair-per-pair-timeout` CLI
  wiring (reference only, unlikely to need changes).

## Test Plan

- [ ] Manual verification: profile `TMDS_D2` (py-spy or cProfile) for a
      single `route_differential_pair_coupled` call; capture and post
      the wall-clock split (guide probe vs. coupled loop) and hot-path
      summary on this issue.
- [ ] Diagnostic capture: read and post `last_iterations`,
      `last_timeout_exceeded`, `last_iteration_limited`,
      `last_best_progress`, `last_rejections` off the `CoupledPathfinder`
      instance for the profiled TMDS_D2 run.
- [ ] If a fix is implemented: `uv run pytest tests/router/ -k diffpair`
      and `uv run pytest tests/router/ -k coupled` green; re-measure
      TMDS_D2 (and ideally DQS/MIPI_CLK/MIPI_DAT1) skew + continuity via
      `kct check --net-class-map
      boards/07-matchgroup-test/output/net_class_map.json
      boards/07-matchgroup-test/output/matchgroup_test_routed.kicad_pcb`
      after a `--differential-pairs` route.
- [ ] If no fix lands (writeup-only outcome): confirm the writeup
      includes the Python-vs-C++ finding, the basin-flooding-vs-overhead
      determination, and a concrete scoped proposal a future issue could
      implement without re-profiling from scratch.
- [ ] No cross-board regression: `uv run pytest tests/router/` full
      suite green (or only the pre-existing board-02 baseline-drift
      failure noted in #4051's test plan, unrelated to this change).

rjwalters:
Closes #4052

## Summary — diagnose-first outcome (Board 07 Phase 2)

Profiled `TMDS_D2` and found the curated hypothesis (~14k-iteration basin-flooding) was **stale**: on `main` the coupled search dies at the **start state after 4 iterations**, gated by a spacing-floor bug, never a timeout/iteration budget.

**Root cause.** Board 07's diff-pair net classes carry `target_diff_impedance=100`, so the impedance resolver (`diffpair_impedance.py:524`) overwrites `intra_pair_clearance` with the physics-derived edge-to-edge coupling gap that hits 100Ω on the thick 4-layer stack — measured **8.425 mm**. `route_differential_pair_coupled` then fed `ceil((0.2 + 8.425)/0.1) = 87 cells` as the coupled search's within-pair `min_spacing_cells` floor, so the joint-state A* rejected **every** move off the ~1 mm pad pitch (`rejections` all `sym_floor`/`asym_floor`, `best_progress` never improving). This blocked **all** board-07 coupled pairs. The 8.425 mm gap is a stackup impedance quantity, not a spacing floor — the same mis-read #3440 fixed for the match-group tuner.

## Fix

Clamp `pair_intra_clearance` to the geometric `trace_clearance` floor **only when `target_diff_impedance` is set** (the signal the field was impedance-overwritten). Pairs that legitimately declare a wider `intra_pair_clearance` without a target impedance (the #3012 case) are untouched, so the gated clamp cannot reintroduce #3012's `diffpair_clearance_intra` violations.

## What this does and does NOT achieve

- **Does:** unblocks the floor death — the search now runs the full budget instead of exiting at 4 iterations. This is the necessary prerequisite for any coupled routing on impedance-controlled boards.
- **Does not (yet):** route a pair to completion. Once unblocked, the search hits the genuine pure-Python joint-state **basin-flooding** the epic predicted (no pair converges within a CI-affordable budget; weighted A* `w=1.5/2/3` does not rescue it; the shadow path fails at "no legal shadow-via site"). That is the C++ coupled-loop port — scoped as a follow-up in the [issue diagnosis comment](https://github.com/rjwalters/kicad-tools/issues/4052#issuecomment-4952026714).

Acceptance floor (one pair routed coupled within budget meeting skew/continuity) is **not met**; per the issue's explicit guidance this lands the diagnosis + highest-value increment rather than blocking.

## Tests

- Two regression tests in `test_diffpair_corridor.py`: impedance-gap clamped to the geometric floor (verified FAILS without the clamp — floor=87 — and PASSES with it — floor≤4); a legitimately-declared wide clearance survives the gated clamp.
- `uv run ruff check` + `uv run ruff format --check` clean on both touched files.
- `tests/router/` diffpair/coupled/corridor/board03 suite: **60 passed, 0 failed**.

### DIFF
--- a/src/kicad_tools/router/diffpair_routing.py
+++ b/src/kicad_tools/router/diffpair_routing.py
@@ -4491,6 +4491,38 @@ class DiffPairRouter:
             # behaviour for callers that don't supply a net_class_map.
             pair_intra_clearance = float(spacing)
 
+        # Issue #4052: clamp the impedance-coupling gap out of the coupled
+        # spacing target.  When a net class carries a ``target_diff_impedance``
+        # the impedance resolver (``diffpair_impedance.py:524``) OVERWRITES
+        # ``intra_pair_clearance`` with the physics-derived edge-to-edge
+        # coupling gap needed to hit that impedance on the board's stackup
+        # (board 07: 8.425 mm for loosely-coupled 100 ohm on a thick
+        # 4-layer stack).  That gap is a *stackup impedance* quantity, NOT
+        # a within-pair spacing floor: fed straight into the coupled
+        # search's ``min_spacing_cells`` it demands the two centerlines sit
+        # ~8.6 mm apart (87 cells at 0.1 mm), so every move from the
+        # physical pad pitch (~1 mm) is rejected by the spacing floor and
+        # the joint-state search dies at the start state in 4 iterations
+        # (measured: ALL board-07 coupled pairs, ``sym_floor`` /
+        # ``asym_floor_p`` / ``asym_floor_n`` rejections only,
+        # ``best_progress`` never improving).
+        #
+        # Clamp is gated on ``target_diff_impedance`` being set -- that is
+        # the SIGNAL that ``intra_pair_clearance`` was overwritten with a
+        # stackup gap (``diffpair_impedance.resolve_impedance_driven_sizing``
+        # only replaces the field when a target impedance drove the sizing,
+        # ``used_target=True``).  A pair that legitimately DECLARES a wider
+        # ``intra_pair_clearance`` (the #3012 case: board 07's 0.1 mm
+        # within-pair clearance with a 0.15 mm trace) has no
+        # ``target_diff_impedance`` and is left untouched, so its floor
+        # still holds the declared within-pair separation post-route.  When
+        # gated, clamp to the geometric ``trace_clearance`` floor (the
+        # tightest DRC-legal within-pair separation), mirroring the
+        # match-group tuner's identical impedance-gap mis-read fix in #3440.
+        if getattr(pair_net_class, "target_diff_impedance", None) is not None:
+            within_pair_clearance_floor = float(self.autorouter.rules.trace_clearance)
+            pair_intra_clearance = min(pair_intra_clearance, within_pair_clearance_floor)
+
         required_center_spacing = pair_trace_width + pair_intra_clearance
         min_spacing_cells = max(
             1, math.ceil(required_center_spacing / self.autorouter.grid.resolution)
--- a/tests/test_diffpair_corridor.py
+++ b/tests/test_diffpair_corridor.py
@@ -28,6 +28,7 @@ to a near-1D tube, which the pure-Python search completes in seconds.
 from __future__ import annotations
 
 import logging
+import math
 
 from kicad_tools.router.core import Autorouter
 from kicad_tools.router.diffpair import DifferentialPairConfig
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Reviewed as a diagnosis-overturning correctness fix, judged against issue #4052's explicit allowance ("one pair meets skew/continuity within budget, OR a root-cause writeup with a concrete fix proposal"). The builder delivered the latter plus a prerequisite bug fix — that is squarely in scope, and the honest-negative reporting is exemplary (no overclaiming that board-07 pairs route).

## Independently verified (not trusted)

1. **Core claim confirmed by direct experiment.** I temporarily disabled the clamp in `diffpair_routing.py` and ran the regression test: it fails with `assert 87 < 87` and prints `rejections={'sym_floor': 8, 'asym_floor_p': 8, 'asym_floor_n': 8}`, `success=False` — the exact 4-iteration start-state floor death described in the diagnosis. Re-enabling the clamp: passes. Arithmetic checks out: `ceil((0.2 + 8.425)/0.1) = 87` cells.

2. **Resolver path confirmed.** `diffpair_impedance.py:521-525` does `dataclasses.replace(nc, ..., intra_pair_clearance=result.gap_mm)` only when `result.used_target` — and it leaves `target_diff_impedance` set on the class. So gating the clamp on `getattr(pair_net_class, "target_diff_impedance", None) is not None` (`diffpair_routing.py:4522`) keys precisely on the "field was resolver-overwritten" signal. The `getattr(..., None)` form is also null-safe when `pair_net_class is None`.

3. **Gate is load-bearing and proven both ways.** `test_declared_wide_intra_clearance_survives_clamp` declares `intra_pair_clearance=0.5` with **no** `target_diff_impedance` and asserts the floor stays at the declared 7 cells — the #3012 case is untouched. `test_impedance_gap_does_not_inflate_coupled_spacing_floor` asserts `floor < 87` **and** `floor <= 4`, testing the mechanism, not merely absence of a crash.

4. **Blast radius on board 06 (the diffpair test board) is nil — and I confirmed it.** Board 06 declares `target_diff_impedance=90/100`, so it exercises the clamp. But its impedance-resolved gaps are small (0.075/0.1 mm, below `trace_clearance`), so `min(gap, trace_clearance)` is a no-op there. Locally: `test_board_06_diffpair_test.py` 42/42 pass; the broader impedance/floor/coupled/corridor/board03 suites 126 passed, 1 skipped. In CI: **"Diff-Pair Routing Regression (boards/06-diffpair-test)" and "Match-Group Routing Regression (boards/07-matchgroup-test)" both pass** (they re-route the boards and assert segment floors) — no committed-artifact drift. No board artifacts were touched by this PR.

5. **Honest-negative reporting is plain and correct.** Both the PR body and the issue diagnosis comment state the acceptance floor is NOT met, that no board-07 pair converges post-fix (MIPI_CLK 100k best_progress=13, DQS 60k best_progress=7), that weighted A* doesn't rescue it, and scope the C++ coupled-loop port as the follow-up. No overclaim.

6. **CI green, merge CLEAN.** All 16 checks pass (Lint/Format, Type Check, C++ Build, Test 16m, all board e2e). `ruff check` + `ruff format --check` clean on both touched files.

## Test Scoping

**Strategy**: targeted local run of diffpair/coupled/corridor/impedance/board03/board06 suites + CI full suite
**Result (local)**: board06 42/42; impedance+floor+coupled+corridor+board03 126 passed / 1 skipped; new regression pair verified to FAIL without the fix and PASS with it
**Result (CI)**: 16/16 checks pass, mergeStateStatus=CLEAN

Excellent diagnostic work. The comment block at `diffpair_routing.py:4494-4521` is thorough and will save the next reader real time.

Issue #4054, opened 2026-07-12
Epic: kicad-tools repo installation — vendor skills + conventions into consumer PCB-design repos


rjwalters:
# Epic: kicad-tools repo installation into consumer PCB-design repos

## Owner request
Owner-requested directly. Make kicad-tools *installable* into a consumer PCB-design repo (like Loom/Anvil install into their consumers), so the design/route/check work happens **inside the consumer repo** with kicad-tools as tooling — inverting the chorus-v22 pattern where work was done by sitting inside kicad-tools with a symlinked fixture.

## Owner-confirmed design decision (do not re-litigate)
Install model = **uv dependency + vendored skills** (Anvil's model, NOT Loom's full vendoring):

- The consumer repo gets the `kicad-tools` Python package as a **normal uv dependency** — `git+https://github.com/rjwalters/kicad-tools` for consumers, or a **local path dep** (`--path`) for sibling-repo development.
- The installer **vendors ONLY**: Claude Code skills (`.claude/commands/kct/*`), CI gate scripts, and a guarded `CLAUDE.md` section carrying kicad-tools conventions.
- Upgrades = `uv lock --upgrade` + re-run installer.
- **NOT** PyPI-first (may come later). **NOT** fully vendored (Loom's model).

## Verified context
- kicad-tools is a proper Python package: `pyproject.toml` `name="kicad-tools"` v0.14.0, console entry points `kct` and `kicad-tools` (both -> `kicad_tools.cli:main`). Verified.
- Existing skill namespace `.claude/commands/kct/` currently holds only `ee-review.md` + `README.md` (namespace established by #3995). Vendored Loom skills live in `.claude/commands/loom/` (30 files) and MUST NOT be touched. Verified.
- `scripts/ci/` (`check_copper_lvs.py`, `check_routed_drc.py`, `net_class_map_resolver.py`, and others) is **deliberately excluded from the installed wheel**: `pyproject.toml` sets `packages = ["src/kicad_tools"]`, so scripts/ci ships in neither sdist nor wheel. Verified that `src/kicad_tools/**` does **not import** scripts/ci — it only references those paths in comments/docstrings (the "mirror" pattern, e.g. `analysis/net_status.py`, `cli/fleet_cmd.py`). So a consumer that `pip/uv add`s the package **cannot** obtain the CI gates from the package — the installer must vendor them, OR a stable subset gets relocated into the package under `kicad_tools.ci` with thin wrappers (Child 2 decides).
[...]

PR #4062, opened 2026-07-12
Installer MVP: scripts/install-kct.sh <target-repo> with --dry-run (#4055)

### LINKED ISSUE #4055: kct-install Child 1: Installer MVP — scripts/install-kct.sh <target-repo> with --dry-run
Part of Epic #4054 (kicad-tools repo installation into consumer PCB-design repos).

## Goal
Ship `scripts/install-kct.sh <target-repo>` that installs kicad-tools into a consumer PCB-design repo per the owner-confirmed **uv-dependency + vendored-skills** model (Anvil's model, not Loom full-vendoring).

## What the installer must do
1. **Add kicad-tools as a uv dependency** in the target's pyproject/uv config:
   - Default: `git+https://github.com/rjwalters/kicad-tools` (pin to a tag/rev).
   - `--path <dir>`: local path dependency for sibling-repo development (e.g. `../kicad-tools`), for the chorus-next-door workflow.
   - Choose the least-invasive mechanism (`uv add` if the target is uv-managed, else edit `[project.dependencies]` / a `[tool.uv.sources]` entry). Verify against the target's existing pyproject; do not clobber unrelated deps.
2. **Vendor `.claude/commands/kct/` skills** into the target (copy the selected skill set). Support `--skills=a,b,c` selective install and a default-all, mirroring Anvil's `install-anvil.sh` (`--skills=`, verified). Never touch `.claude/commands/loom/` (coexist additively).
3. **Append a guarded CLAUDE.md section** to the target's `CLAUDE.md` using marker bounds (e.g. `<!-- BEGIN KICAD-TOOLS -->` / `<!-- END KICAD-TOOLS -->`), Anvil-style. Re-runs must be **idempotent** (replace the marked block, never duplicate). This section MUST carry the three hard-won conventions from Epic #4054: build-native-after-checkout, the `kicad-cli pcb drc --refill-zones` cross-gate rule, and artifact-first.
4. **Write install metadata** (`.kct/install-metadata.json` or similar): version/rev installed, list of vendored files, skills selected, source mode (git vs path). Model on Loom's `install-metadata.json` schema (version + installed_files, verified) so a future uninstaller/upgrade-detector can consume it.
5. **`--dry-run`**: print every planned write, write nothing (Anvil pattern, verified).

## Non-goals
- PyPI publishing (owner: not now).
- Full vendoring of the package source (that is Loom's model; explicitly rejected).
- The CI gates themselves (Child 2) and the extracted skills' content (Child 3) — this child ships the *mechanism* and can vendor whatever `.claude/commands/kct/` exists today.

## Reference patterns (study, do not copy wholesale)
- `/Users/rwalters/GitHub/anvil/scripts/install-anvil.sh`: `--skills=`, `--dry-run`, marker-bounded additive CLAUDE.md, generated consumer pyproject, install-metadata.json, active-install guard, Loom-coexistence.
- `/Users/rwalters/GitHub/loom/install.sh`: install-metadata.json schema + uninstaller wiring.

## Acceptance
- Installer test driving a throwaway tmp git repo (bats or pytest+subprocess): asserts (a) `--dry-run` writes nothing, (b) real run adds the kicad-tools dep to pyproject, vendors `.claude/commands/kct/`, injects the marked CLAUDE.md block, writes install-metadata.json; (c) a second run is idempotent (no duplicate CLAUDE.md block, no dep duplication); (d) `--path ../kicad-tools` produces a path/source dependency.
- Test runs in CI without network by using `--path` mode (git mode may be smoke-tested separately or mocked).

## Sweepable
Yes. Self-contained scripting + test.

---

### Verified current state (2026-07-11)
- `.claude/commands/kct/` currently holds exactly 2 files: `ee-review.md`, `README.md` (verified via `ls`). This is the entire skill set to vendor at v1 — the installer's file-enumeration logic must not hardcode this list; it should glob `*.md` under `.claude/commands/kct/` at install-source-resolution time so Child 3's future skill additions are picked up automatically.
- `pyproject.toml` at repo root: `name = "kicad-tools"`, `version = "0.14.0"`, no `[tool.uv.sources]` table yet (n/a — this repo is the *source*, not a consumer).
- `uv --version` in this environment: `0.11.14`.
- No `scripts/install-kct.sh` exists yet (new file, greenfield).
- No prior kicad-tools-branded installer test exists under `tests/` (checked `git ls-files 'tests/*install*'` — empty).

### uv dependency mechanics — exact commands to use

`uv add` in this uv version (0.11.14) has **no bare `--path` or `--git` flag**. Source selection happens two ways, and the installer should pick based on target-repo state:

**Git mode (default):**
```bash
uv add "kicad-tools @ git+https://github.com/rjwalters/kicad-tools@<tag-or-rev>"
```
This is a PEP 508 direct-reference requirement string passed as a single positional arg to `uv add`. `uv add` will resolve it and write BOTH:
- `pyproject.toml`: `dependencies = [..., "kicad-tools"]` (bare name)
- `pyproject.toml`: a new/updated `[tool.uv.sources]` table entry: `kicad-tools = { git = "https://github.com/rjwalters/kicad-tools", rev = "<tag-or-rev>" }` (or `tag = "..."` if a tag is pinned via `--tag`)

Prefer `uv add kicad-tools --tag <tag> --rev <rev>`-style explicit flags over embedding `@rev` in the URL string where a specific pin is known: `uv add` supports `--tag`, `--rev`, `--branch` as separate flags that apply when the requirement resolves to a Git source (verified via `uv add --help`). Example:
```bash
uv add kicad-tools --git https://github.com/rjwalters/kicad-tools --tag v0.14.0
```
Note: `--git` IS accepted by `uv add` as a value-bearing flag paired with a bare package name (distinct from embedding `git+` in the requirement string) — verify against the installed uv version's `--help` output before hardcoding, since flag surface differs across uv releases; this repo's environment uv (0.11.14) supports `--tag`/`--rev`/`--branch` per `uv add --help`.

**Path mode (`--path <dir>` on the installer's own CLI, NOT uv's):**
```bash
uv add kicad-tools --editable --path ../kicad-tools   # or: uv add ../kicad-tools directly
```
The idiomatic uv path-dependency form is to pass the path itself as the "package" argument (uv detects it's a local directory with a pyproject.toml and resolves it as a path source), which uv also records under `[tool.uv.sources]`:
```
[tool.uv.sources]
kicad-tools = { path = "../kicad-tools", editable = true }
```
The installer's own `--path <dir>` flag is a *separate, installer-level* flag (distinct naming from uv's own dependency-source concept) that tells `install-kct.sh` "use path-source mode with this directory" — do not conflate the installer's `--path` flag with a literal `uv add --path` invocation; translate it into the correct uv invocation shown above.

**Idempotency / existing-dependency detection**: before calling `uv add`, the installer must:
1. Check whether `pyproject.toml` already declares `kicad-tools` in `[project.dependencies]` (grep/parse, not just presence of the string — must not false-positive on a substring match against e.g. `kicad-tools-extra`).
2. If present AND the existing `[tool.uv.sources].kicad-tools` entry already matches the requested mode (git-vs-path, same rev/tag or same path) — no-op, print "kicad-tools dependency already present and up to date", skip the `uv add` call entirely (this is what makes reruns idempotent on the dependency axis, not just the CLAUDE.md axis).
3. If present but pointing at a *different* source (e.g. installed as git, re-run requests `--path`) — this is a mode switch; either re-run `uv add` (which uv will update in place) or warn and require an explicit `--force`/confirmation, mirroring Anvil's override-skip discipline (`install-anvil.sh` Stage 5's "consumer-modified" skip pattern, cited above) so an operator's manual pyproject edits aren't silently clobbered.
4. If `pyproject.toml` doesn't exist at all in the target (non-uv-managed repo) — this is out of the MVP's happy path per the "least-invasive mechanism" note above; either `uv init` first (if `uv` is available and target confirms) or fail with a clear error naming the missing prerequisite. Document whichever choice is made in the script's `--help` text.
5. `uv add` mutates `uv.lock` too (network fetch for git mode). The installer test (Acceptance criterion above) must use `--path` mode specifically because it resolves from local disk with no network call — matches the epic's CI-no-network constraint.

### Marker-comment format for the guarded CLAUDE.md block

Follow Anvil's exact pattern (`install-anvil.sh` lines ~109-111, verified above):

```bash
KCT_MARK_BEGIN='<!-- BEGIN KICAD-TOOLS -->'
KCT_MARK_END='<!-- END KICAD-TOOLS -->'
```

Injection/replacement algorithm (idempotent — must not duplicate on rerun):
1. If `CLAUDE.md` does not exist, create it with just the marked block.
2. If `CLAUDE.md` exists and contains `$KCT_MARK_BEGIN`: replace everything from `$KCT_MARK_BEGIN` through `$KCT_MARK_END` (inclusive) with the freshly generated block. Use `awk`/`sed` range-delete + insert, or a `mktemp` + line-by-line rebuild (portable — avoid GNU-sed-only `-i` semantics that break on macOS BSD sed, since this repo's Darwin dev environment matters here — see the Anvil script's approach of building output via `cat`/heredoc into a temp file then `mv`, not in-place `sed -i`).
3. If `CLAUDE.md` exists and does NOT contain `$KCT_MARK_BEGIN`: append the marked block to the end (preceded by a blank line), preserving all existing content untouched.
4. The block content itself must literally carry the three conventions, not just link to them — the whole point (per the epic) is these are load-bearing and must survive even if the target repo's CLAUDE.md later diverges from kicad-tools' own CLAUDE.md wording. Suggested block body (builder should adapt wording, not omit substance):
   - `uv run kct build-native` after every fresh checkout/worktree (uv sync alone does not build the C++ router extension).
   - Cross-gate rule: `kct check` alone is insufficient for manufacturing sign-off; always also run `kicad-cli pcb drc --refill-zones`.
   - Artifact-first: the committed board artifact is shipping truth; `generate_design.py`/regeneration may diverge by design and is not authoritative over a committed board.

### Install metadata schema

Model on Loom's schema (`loom/install.sh` `finalize_quick_install()`, verified above — fields: `loom_version`, `loom_commit`, `install_date`, `loom_source`, `installed_files`), adapted to kicad-tools naming and Anvil's richer per-item tracking where useful for a future uninstaller/upgrade-detector:

```json
{
  "kct_version": "0.14.0",
  "kct_commit": "<short-sha-of-source-checkout-or-empty-if-installed-via-git-dep>",
  "install_date": "2026-07-11",
  "source_mode": "git",
  "source_ref": "https://github.com/rjwalters/kicad-tools@v0.14.0",
  "skills_selected": ["ee-review"],
  "installed_files": [
    ".claude/commands/kct/ee-review.md",
    ".claude/commands/kct/README.md"
  ]
}
```
Write to `.kct/install-metadata.json` (mirrors Anvil's `.anvil/install-metadata.json` and Loom's `.loom/install-metadata.json` — dedicated dotdir per tool, NOT dumped at repo root). `source_mode` is `"git"` or `"path"`; when `"path"`, `source_ref` is the resolved absolute path passed via `--path`, and `kct_commit` should record that source checkout's short SHA (`git -C <path> rev-parse --short HEAD`) since there's no tag/rev to pin.

### Affected Files

- `scripts/install-kct.sh` (new) — the installer script itself; primary deliverable.
- `tests/install/test_install_kct.py` (new, or `tests/scripts/test_install_kct.py` — builder's choice of location, but must be discoverable by `uv run pytest`) — drives a `tmp_path` git repo through dry-run, real-run, idempotent-rerun, and `--path` mode per the Acceptance section above.
- `.claude/commands/kct/README.md` (read-only reference) — enumerate its current 2-file sibling `ee-review.md` as the v1 vendor payload; do not hand-edit, the installer must discover files by glob, not by a hardcoded list.
- `CLAUDE.md` (read-only reference for wording source) — the "Routing performance: build the C++ backend first" and native-build-staleness sections at the top of this repo's own `CLAUDE.md` are the canonical wording to summarize into the injected block's convention #1; do not literally vendor the whole file, distill the three conventions.
- `pyproject.toml` (read-only reference) — confirms `name = "kicad-tools"`, `version = "0.14.0"` for metadata population; not modified by this issue (this repo is the source, not a consumer).

### Test Plan

- [ ] Automated tests (primary deliverable, not run by curator — CPU-heavy suites deferred to builder/CI per host load constraint): `tests/install/test_install_kct.py` (or equivalent path) using `tmp_path` + `subprocess.run(["scripts/install-kct.sh", ...])` against a throwaway `git init`-ed target:
  - `--dry-run` writes nothing to the target (assert no files created/modified, script still exits 0 and prints planned actions).
  - Real run (`--path <local-kicad-tools-checkout>` mode, no network) produces: dependency entry in target `pyproject.toml` + `[tool.uv.sources]`, vendored `.claude/commands/kct/*.md` files present and byte-identical to source, `CLAUDE.md` contains exactly one `<!-- BEGIN KICAD-TOOLS -->...<!-- END KICAD-TOOLS -->` block with all three conventions present in the text, `.kct/install-metadata.json` present and valid JSON with the schema fields above.
  - Second run (idempotent rerun, same args) produces: exactly one CLAUDE.md marked block (not two), no duplicate dependency entry in pyproject, `.claude/commands/kct/` files unchanged (or refreshed in place, not duplicated), metadata file updated (not duplicated) — matches the epic's "upgrade = re-run installer" acceptance mode.
  - `--skills=ee-review` installs only that skill's files (once there is more than one skill; with today's single-skill vendor set this assertion is `--skills=ee-review` == default-all, but the flag-parsing/filtering code path itself must still be exercised and tested against an unknown-skill-name error case).
  - `.claude/commands/loom/` (if present in the tmp target, e.g. seeded by the test fixture) is byte-for-byte untouched after install — regression guard for the "never touch loom" requirement.
- [ ] Manual verification: builder runs `scripts/install-kct.sh --dry-run /tmp/some-scratch-repo` and `scripts/install-kct.sh --path $(pwd) /tmp/some-scratch-repo` locally, inspects the resulting `pyproject.toml`, `CLAUDE.md`, `.kct/install-metadata.json` by eye before opening the PR.
- [ ] Edge cases: target repo with no `pyproject.toml` (documented failure mode, see uv-mechanics point 4 above); target repo with a pre-existing non-kicad-tools `<!-- BEGIN KICAD-TOOLS -->`-shaped comment collision (unlikely but the marker string should be distinctive enough this is a non-issue — no action needed, just confirm the exact-string match doesn't false-positive on partial matches).

### Note for CI (per epic's acceptance criterion)
"Test runs in CI without network by using `--path` mode" — confirm the test suite's default/CI invocation exercises `--path` against a fixture checkout (can be the CI runner's own kicad-tools checkout, i.e. `--path $GITHUB_WORKSPACE` or equivalent), and that any git-mode-specific assertions are either skipped in CI or mocked (e.g. stub `uv add`'s git resolution, or gate behind a `@pytest.mark.network` marker excluded by default).

rjwalters:
Closes #4055 (child of Epic #4054).

## What
Ships `scripts/install-kct.sh`, the kicad-tools consumer-repo installer, per the owner-confirmed **uv-dependency + vendored-skills** model (Anvil's model, not Loom's full-vendoring).

### Installer behavior
- **uv dependency**: git source by default (pins to `v<version>` tag; `--tag`/`--rev` override), or a **network-free editable path source** via the installer's own `--path <dir>` flag. Path mode uses `uv add <dir> --editable --frozen`, so it writes the `pyproject.toml` + `[tool.uv.sources]` entry without resolving/syncing (offline, CI-safe).
- **Vendors `.claude/commands/kct/*.md`** — globbed at install time (Child 3's future skills are picked up automatically; no hardcoded list). README is always vendored. **Never touches `.claude/commands/loom/`** (coexists additively with Loom).
- **Guarded CLAUDE.md block** — idempotent `<!-- BEGIN KICAD-TOOLS -->` / `<!-- END KICAD-TOOLS -->` markers, replaced in place via a BSD-sed-safe pure-bash line rebuild (no GNU `sed -i`). The block literally carries the three load-bearing Epic #4054 conventions: build-native-after-checkout, the `kicad-cli pcb drc --refill-zones` cross-gate rule, and artifact-first.
- **`.kct/install-metadata.json`** — modeled on Loom's schema (`kct_version`, `kct_commit`, `install_date`, `source_mode`, `source_ref`, `skills_selected`, `installed_files`) for a future uninstaller/upgrade-detector.
- **`--dry-run`** prints every planned write, writes nothing.

### Dependency idempotency (curated 4-step algorithm)
Whole-word `kicad-tools` detection (no false-positive on `kicad-tools-extra`), source-mode/ref match check (path entries resolved back to absolute before comparing, since uv records a relative path), and a genuine no-op ("already present and up to date") when nothing changed.

## Tests — `tests/install/test_install_kct.py`
Drives a `tmp_path` git repo through: dry-run (no writes), path-mode install (dep + source + byte-identical skills + single CLAUDE block with all three conventions + valid metadata JSON), idempotent rerun (single block, no dep duplication, dep no-op), loom-coexistence (loom skill byte-unchanged), skill selection + unknown-skill error, missing-pyproject failure, and git-mode source write. **Hermetic**: `uv` is stubbed on PATH, so the suite is network-free.

```
8 passed in 2.05s
```

## Checks
- `uv run pytest tests/install/` — 8 passed
- `uv run ruff check` / `ruff format --check` on touched files — clean
- `shellcheck scripts/install-kct.sh` — clean

## Deferred (per issue non-goals)
- PyPI publishing, full package vendoring (Loom's model, rejected).
- CI gates (Child 2) and skill *content* (Child 3) — this child ships the mechanism.
- Non-uv target repos: `install-kct.sh` fails with a clear "no pyproject.toml" error rather than running `uv init` (documented in `--help`; out of the MVP happy path per the curated uv-mechanics point 4).
- Git-mode network resolution is not exercised live in tests (asserted on pyproject content via the uv stub, per the issue's CI-no-network constraint).

### DIFF
--- a/scripts/install-kct.sh
+++ b/scripts/install-kct.sh
@@ -1,0 +1,481 @@
+#!/usr/bin/env bash
+# install-kct.sh - Install kicad-tools into a consumer PCB-design repository.
+#
+# Distribution model (Epic #4054, owner-confirmed): uv-dependency + vendored
+# skills (Anvil's model, NOT Loom's full-package vendoring). This installer:
+#   1. Adds kicad-tools as a uv dependency in the target's pyproject.toml
+#      (git source by default, local path source with --path <dir>).
+#   2. Vendors .claude/commands/kct/*.md skills into the target (NEVER touches
+#      .claude/commands/loom/ — coexists additively with Loom).
+#   3. Appends a guarded, idempotent <!-- BEGIN KICAD-TOOLS --> block to the
+#      target's CLAUDE.md carrying the three hard-won Epic #4054 conventions.
+#   4. Writes .kct/install-metadata.json for a future uninstaller/upgrader.
+#   5. --dry-run prints every planned write and writes nothing.
+#
+# Usage:
+#   ./scripts/install-kct.sh [OPTIONS] <target-repo>
+#
+# Options:
+#   --path <dir>       Install kicad-tools as a LOCAL PATH dependency pointing
+#                      at <dir> (a kicad-tools checkout), for sibling-repo
+#                      development (e.g. --path ../kicad-tools). Network-free.
+#                      Without --path, a git source is used (default).
+#   --tag <tag>        Git-source pin: use this tag (git mode only). Default:
+#                      the source checkout's kct version tag (v<version>).
+#   --rev <rev>        Git-source pin: use this rev/sha (git mode only).
+#   --skills=<a,b,c>   Vendor only the listed skills (default: all *.md under
+#                      the source .claude/commands/kct/). README is always
+#                      vendored (it documents the namespace).
+#   --dry-run          Print planned actions, write nothing.
+#   -y, --yes          Non-interactive (skip confirmation; auto-enabled when
+#                      stdin is not a TTY, e.g. CI/agent shells).
+#   -h, --help         Show this help and exit.
+#
+# Examples:
+#   ./scripts/install-kct.sh /path/to/board-repo
+#   ./scripts/install-kct.sh --path ../kicad-tools /path/to/board-repo
+#   ./scripts/install-kct.sh --dry-run /path/to/board-repo
+#   ./scripts/install-kct.sh --skills=ee-review /path/to/board-repo
+#
+# Prerequisites in the target repo:
+#   - A pyproject.toml at the target root. This MVP does NOT `uv init` a
+#     non-uv repo; it fails with a clear error naming the missing file
+#     (Epic #4054 "least-invasive mechanism" — the happy path is a
+#     uv-managed consumer repo).
+#
+# Re-running the installer is the upgrade/idempotency path: a second run with
+# the same args adds no duplicate CLAUDE.md block and no duplicate dependency.
+
+set -euo pipefail
+
+# ----- ANSI colors -----------------------------------------------------------
+if [[ -t 1 ]]; then
+  RED=$'\033[0;31m'
+  GREEN=$'\033[0;32m'
+  BLUE=$'\033[0;34m'
+  YELLOW=$'\033[1;33m'
+  CYAN=$'\033[0;36m'
+  NC=$'\033[0m'
+else
+  RED=""; GREEN=""; BLUE=""; YELLOW=""; CYAN=""; NC=""
+fi
+
+error() { echo "${RED}error: $*${NC}" >&2; exit 1; }
+info()  { echo "${BLUE}> $*${NC}"; }
+ok()    { echo "${GREEN}  ok: $*${NC}"; }
+warn()  { echo "${YELLOW}  warn: $*${NC}"; }
+note()  { echo "${CYAN}  note: $*${NC}"; }
+
+usage() {
+  # The Usage / Options / Examples block lives in the header comment.
+  sed -n '4,45p' "$0" | sed 's/^# \{0,1\}//'
+  exit 0
+}
+
+# ----- CLAUDE.md marker constants -------------------------------------------
+KCT_MARK_BEGIN='<!-- BEGIN KICAD-TOOLS -->'
+KCT_MARK_END='<!-- END KICAD-TOOLS -->'
+
+# ----- Argument parsing ------------------------------------------------------
+SKILLS_FILTER=""
+DRY_RUN=false
+NON_INTERACTIVE=false
+PATH_MODE=false
+PATH_DIR=""
+GIT_TAG=""
+GIT_REV=""
+TARGET=""
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --skills=*) SKILLS_FILTER="${1#--skills=}"; [[ -z "$SKILLS_FILTER" ]] && error "--skills requires a comma-separated list"; shift ;;
+    --skills)   shift; SKILLS_FILTER="${1:-}"; [[ -z "$SKILLS_FILTER" ]] && error "--skills requires a comma-separated list"; shift ;;
+    --path)     shift; PATH_DIR="${1:-}"; [[ -z "$PATH_DIR" ]] && error "--path requires a directory argument"; PATH_MODE=true; shift ;;
+    --path=*)   PATH_DIR="${1#--path=}"; [[ -z "$PATH_DIR" ]] && error "--path requires a directory argument"; PATH_MODE=true; shift ;;
+    --tag)      shift; GIT_TAG="${1:-}"; [[ -z "$GIT_TAG" ]] && error "--tag requires a value"; shift ;;
+    --tag=*)    GIT_TAG="${1#--tag=}"; shift ;;
+    --rev)      shift; GIT_REV="${1:-}"; [[ -z "$GIT_REV" ]] && error "--rev requires a value"; shift ;;
+    --rev=*)    GIT_REV="${1#--rev=}"; shift ;;
+    --dry-run)  DRY_RUN=true; shift ;;
+    -y|--yes)   NON_INTERACTIVE=true; shift ;;
+    -h|--help)  usage ;;
+    --*)        error "unknown option: $1 (run with --help to see usage)" ;;
+    *)
+      if [[ -n "$TARGET" ]]; then
+        error "unexpected extra argument: $1 (target already set to: $TARGET)"
+      fi
+      TARGET="$1"
+      shift
+      ;;
+  esac
+done
+
+if [[ "$PATH_MODE" == true ]] && { [[ -n "$GIT_TAG" ]] || [[ -n "$GIT_REV" ]]; }; then
+  error "--path is a local-path source; --tag/--rev only apply to git mode"
+fi
+
+# Auto-detect non-interactive mode when stdin is not a TTY (agent/CI shells).
+if [[ "$NON_INTERACTIVE" != true ]] && [[ ! -t 0 ]]; then
+  NON_INTERACTIVE=true
+fi
+
+[[ -z "$TARGET" ]] && error "target repository path required (run with --help to see usage)"
+
+# ----- Stage 1: resolve KCT_ROOT (this installer's source checkout) ---------
+info "Stage 1: resolve kicad-tools source root"
+KCT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+[[ -f "$KCT_ROOT/pyproject.toml" ]] || error "source root missing pyproject.toml: $KCT_ROOT"
+SKILLS_SRC="$KCT_ROOT/.claude/commands/kct"
+[[ -d "$SKILLS_SRC" ]] || error "source root missing .claude/commands/kct/: $KCT_ROOT"
+ok "KCT_ROOT=$KCT_ROOT"
+
+# Extract kicad-tools version from the source pyproject.toml (first version =).
+KCT_VERSION="$(grep -m1 -E '^version[[:space:]]*=' "$KCT_ROOT/pyproject.toml" \
+  | sed -E 's/^version[[:space:]]*=[[:space:]]*"([^"]+)".*/\1/')"
+[[ -n "$KCT_VERSION" ]] || error "could not extract version from $KCT_ROOT/pyproject.toml"
+ok "KCT_VERSION=$KCT_VERSION"
+
+# Source checkout short SHA (empty if not a git checkout — tolerated).
+KCT_COMMIT="$(git -C "$KCT_ROOT" rev-parse --short HEAD 2>/dev/null || echo "")"
+INSTALL_DATE="$(date +%Y-%m-%d)"
+KCT_GIT_URL="https://github.com/rjwalters/kicad-tools"
+
+# ----- Stage 2: resolve and validate TARGET ---------------------------------
+info "Stage 2: resolve and validate target"
+TARGET="${TARGET/#\~/$HOME}"
+[[ -d "$TARGET" ]] || error "target directory does not exist: $TARGET"
+TARGET="$(cd "$TARGET" && pwd)"
+ok "TARGET=$TARGET"
+
+if [[ "$TARGET" == "$KCT_ROOT" ]]; then
+  error "refusing to install kicad-tools into its own source checkout ($KCT_ROOT)"
+fi
+
+TARGET_PYPROJECT="$TARGET/pyproject.toml"
+if [[ ! -f "$TARGET_PYPROJECT" ]]; then
+  error "target has no pyproject.toml: $TARGET
+       kicad-tools installs as a uv dependency and requires a uv-managed
+       consumer repo. Run 'uv init' in the target first, then re-run."
+fi
+
+# ----- Stage 3: resolve source mode + ref -----------------------------------
+info "Stage 3: resolve source mode"
+if [[ "$PATH_MODE" == true ]]; then
+  PATH_DIR="${PATH_DIR/#\~/$HOME}"
+  [[ -d "$PATH_DIR" ]] || error "--path directory does not exist: $PATH_DIR"
+  RESOLVED_PATH="$(cd "$PATH_DIR" && pwd)"
+  [[ -f "$RESOLVED_PATH/pyproject.toml" ]] || error "--path is not a kicad-tools checkout (no pyproject.toml): $RESOLVED_PATH"
+  SOURCE_MODE="path"
+  SOURCE_REF="$RESOLVED_PATH"
+  # Record the path source's own short SHA when available.
+  PATH_COMMIT="$(git -C "$RESOLVED_PATH" rev-parse --short HEAD 2>/dev/null || echo "")"
+  [[ -n "$PATH_COMMIT" ]] && KCT_COMMIT="$PATH_COMMIT"
+  note "path mode: local source at $RESOLVED_PATH (network-free)"
+else
+  SOURCE_MODE="git"
+  # Default the git pin to the source version tag when neither --tag nor --rev
+  # was given, so a bare `install-kct.sh <target>` pins reproducibly.
+  if [[ -z "$GIT_TAG" && -z "$GIT_REV" ]]; then
+    GIT_TAG="v$KCT_VERSION"
+  fi
+  if [[ -n "$GIT_TAG" ]]; then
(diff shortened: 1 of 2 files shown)


rjwalters:
❌ **Changes Requested — safety-critical: malformed CLAUDE.md causes silent data loss**

Strong work overall — the code is clean, `shellcheck`/`ruff` are green, the 8 tests pass, the self-install guard, non-git-target tolerance, dry-run (verified writes nothing), and uv path/git idempotency all behave correctly. One blocking defect must be fixed before merge.

## Blocking: unterminated marker block deletes target content

`merge_claude_md()` (scripts/install-kct.sh:400-414) sets `in_block=1` when it sees `<!-- BEGIN KICAD-TOOLS -->` and only resets it on `<!-- END KICAD-TOOLS -->`. If the target `CLAUDE.md` has a BEGIN marker but **no matching END** (partial prior write, hand-edit that dropped the END, merge-conflict resolution), `in_block` never clears and **every line after BEGIN is silently dropped** — it is never written to the temp file, then `mv` overwrites the original.

Reproduced against a throwaway repo:

```
BEFORE:
  # My Repo
  IMPORTANT USER CONTENT
  <!-- BEGIN KICAD-TOOLS -->
  stale half block
  MORE IMPORTANT CONTENT AFTER STALE BEGIN   <-- user content past a stale BEGIN

AFTER (exit 0, no warning):
  # My Repo
  IMPORTANT USER CONTENT
  <!-- BEGIN KICAD-TOOLS -->
  ...fresh block...
  <!-- END KICAD-TOOLS -->
  => "MORE IMPORTANT CONTENT AFTER STALE BEGIN" is GONE
```

The issue explicitly requires that unhappy target states "must not corrupt the target — error out rather than silent damage." This is the disqualifying failure mode.

**Fix:** in `merge_claude_md`, detect a BEGIN with no END and **abort without writing** (the target is malformed; the operator should resolve it). Concretely: after the read loop, if `in_block` is still `1`, `rm -f "$tmp"` and `error "CLAUDE.md has an unterminated <!-- BEGIN KICAD-TOOLS --> block (no END marker); refusing to edit — fix the markers and re-run"`. Bonus safety: guard against >1 BEGIN or an END-before-BEGIN the same way.

## Secondary (fix alongside): dry-run misreports the malformed case

For the same unterminated input, `--dry-run` prints `replace existing kicad-tools block in CLAUDE.md (in place)` (scripts/install-kct.sh:433-434) — no hint that content would be lost. Dry-run should surface the same malformed-marker detection so the preview matches the real run's (new) abort behavior.

## Test gap

`tests/install/test_install_kct.py` has no malformed-marker case, which is why this slipped through. Please add a test that seeds a `CLAUDE.md` with a BEGIN and no END and asserts (a) non-zero exit, (b) an error mentioning the unterminated block, and (c) the original file is byte-unchanged.

## Non-blocking (your call — do not need to change for approval)

- Re-install silently overwrites user edits to vendored `.claude/commands/kct/*.md` skills (managed-file semantics, consistent with Loom). Acceptable, but a one-line note in `--help` that these are installer-managed would be kind.

Everything else looks solid. Once the unterminated-marker case errors out safely (and has a test), I'll re-review promptly.

rjwalters:
Addressed the blocking safety defect (commit d7867417).

**Unterminated / malformed markers now abort instead of corrupting the target.** Added `claude_md_marker_error()`, a shared validator that detects both an unterminated `<!-- BEGIN KICAD-TOOLS -->` (no END) and an `END`-before-`BEGIN`. `merge_claude_md` calls it *before* the in-place rebuild and `error`s out (no temp file written, original untouched) when the target is malformed — matching the issue's "error out rather than silent damage" requirement. The `--dry-run` branch runs the same detection, so the preview now reports the malformed marker instead of "replace ... (in place)".

**Tests** (`tests/install/test_install_kct.py`): three new hermetic cases —
- unterminated BEGIN (real run): non-zero exit, error mentions "unterminated", original byte-unchanged (reproduces your BEFORE/AFTER scenario, now safe);
- unterminated BEGIN (`--dry-run`): non-zero exit, error surfaced, no "replace" line, file unchanged;
- inverse END-before-BEGIN: errors out, file unchanged (documented choice: error, not silent no-op).

`uv run pytest tests/install/` → 11 passed. `shellcheck` clean; `ruff check` + `ruff format --check` clean on both touched files.

The non-blocking note (re-install overwriting user-edited vendored skills = managed-file semantics) is left as-is per your call. Ready for re-review.

rjwalters:
✅ **Approved!** (follow-up scoped re-review after the Doctor pass)

The single blocking defect from the first review — `merge_claude_md()` silently deleting target `CLAUDE.md` content after a BEGIN-without-END marker, then clobbering the original on `mv` (data loss on exit 0) — is now fixed correctly.

## Delta verified (Doctor commit `d786741`)
`git diff origin/main...HEAD` touches exactly two files, both expected:
- `scripts/install-kct.sh` — adds `claude_md_marker_error()`, a shared depth-tracking validator (increments on BEGIN, decrements on END; errors on END-before-BEGIN and on non-zero final depth). Wired into **both** the real merge path (before any temp-file write / `mv`) and the `--dry-run` preview.
- `tests/install/test_install_kct.py` — 3 regression tests.

No scope creep: nothing else changed, and the whole PR still only touches these two files vs `main`.

## Empirical reproduction (throwaway `git init` repo + fake `uv` on PATH)
| Case | Result |
|------|--------|
| Unterminated BEGIN, **real** run | exit 1, error "unterminated ... refusing to edit", `CLAUDE.md` **byte-identical** (sha match) |
| Unterminated BEGIN, **--dry-run** | exit 1, no false "replace in place" preview, file byte-unchanged |
| **Nested** `BEGIN, BEGIN, END` | depth 1→2→1 ends non-zero → errors as "unterminated", file byte-unchanged. Sane — no corruption. Validator runs before the rebuild loop, so the loop never sees the malformed input. |
| Happy path: fresh install | exit 0, single well-formed block, all three Epic #4054 conventions present (build-native, `--refill-zones` cross-gate, artifact-first), valid metadata JSON |
| Happy path: idempotent re-run | exit 0, still exactly one BEGIN/END block, `CLAUDE.md` **byte-identical** across reruns |

The original judge reproduction that failed on the previous revision now aborts safely with zero data loss.

## Checks
- `uv run pytest tests/install/` — **11 passed** (8 original + 3 new)
- `shellcheck scripts/install-kct.sh` — clean
- `uv run ruff check tests/install/test_install_kct.py` — clean
- **CI: 16/16 checks pass**, `mergeStateStatus: CLEAN`, `MERGEABLE`

Solid, well-scoped safety fix. Approving for merge.

PR #4067, opened 2026-07-12
fix(validate): rotate THT pad offsets in drill-clearance check (#4044)

### LINKED ISSUE #4044: dimension_drill_clearance ignores footprint rotation -> false hole-to-hole violations on rotated/flipped footprints
## Summary

`validate/rules/dimensions.py` `_check_drill_clearance` computes a through-hole pad's board position by plain addition, ignoring the footprint's rotation:

```python
# validate/rules/dimensions.py, ~lines 211-212
abs_x = fp.position[0] + pad.position[0]
abs_y = fp.position[1] + pad.position[1]
```

But `pad.position` is stored **footprint-local** and must be rotated by `footprint.rotation` to reach board coordinates. This is documented in your own `schema/pcb.py` `Pad` docstring:

> `position` ... is stored footprint-local and must be rotated by `footprint.rotation` to reach board coordinates.

`validate/connectivity.py` handles this correctly (it rotates via `core.geometry.rotate_pad_offset`), but `_check_drill_clearance` does not — so on any rotated or flipped footprint the THT pad holes are placed at the wrong location, producing phantom `dimension_drill_clearance` (hole-to-hole) violations, and potentially missing real ones.

## Reproduce (confirmed on a real board)

A `PinSocket` footprint placed at origin `(15.0, 8.9)` with **rotation 180deg**, pad 3 local offset `(0.0, -5.08)`:

- kct `_check_drill_clearance` naive `fp.position + pad.position` = **(15.0, 3.82)**
- rotation-correct position (matches KiCad / `pcbnew`) = **(15.0, 13.98)**

A ~10 mm error. A GND via near the *computed* (wrong) spot is then reported as:

```
Hole-to-hole clearance 0.366mm < minimum 0.500mm   (items: GND, J1-3)
```

Ground truth on the same board: `kicad-cli pcb drc` with the hole-to-hole rule set to 0.5 mm reports **0** violations, and a direct minimum-distance sweep of all drilled holes measures **1.004 mm** (2x the spec). The reported `J1-3` location `(15.5, 3.91)` is ~10 mm from where pad 3 actually is, which is the tell.

Minimal synthetic case: any footprint rotated 90/180/270deg (or flipped) with a THT pad at a non-zero local offset, plus a via placed at `fp.position + pad_local` — kct flags a hole-to-hole conflict that does not exist in the rotated geometry.

## Expected vs actual

- **Expected:** hole-to-hole measured between the true (rotation-applied) pad hole centers; matches KiCad DRC.
- **Actual:** measured against un-rotated pad positions; false positives on rotated/flipped footprints.

## Fix

Rotate the pad offset by `fp.rotation` before adding to the footprint origin - reuse the same `core.geometry.rotate_pad_offset` helper `validate/connectivity.py` already uses (KiCad's negated-angle convention), so `_check_drill_clearance` matches the documented `Pad.position` contract.

## Environment

- kct installed from git main (`pip install git+https://github.com/rjwalters/kicad-tools`)
- KiCad 10 board (`kicad-cli` DRC used as the ground-truth cross-check)

---

### Root cause confirmed

`src/kicad_tools/validate/rules/dimensions.py::_check_drill_clearance`, lines 211-212:

```python
abs_x = fp.position[0] + pad.position[0]
abs_y = fp.position[1] + pad.position[1]
```

`Pad.position` is documented in `src/kicad_tools/schema/pcb.py` (`class Pad`, docstring on the `rotation` field, ~lines 168-174) as footprint-local and requiring rotation by `footprint.rotation` before translation. This code skips the rotation step entirely. Because these `abs_x`/`abs_y` values feed directly into the center-to-center distance math (`dx = pos2[0] - pos1[0]`, etc.) that *drives the violation determination itself*, this is a correctness bug, not just a cosmetic one: on any footprint with `rotation != 0` (or mirrored), THT pad hole positions used for hole-to-hole clearance are wrong, producing both false positives and possible false negatives (a real too-close hole pair could be computed as far apart).

### Correct pattern already exists in the codebase

`src/kicad_tools/core/geometry.py::rotate_pad_offset(local_x, local_y, rotation_deg)` is the canonical, pcbnew-verified (issue #3739) local-to-board rotation helper. It uses KiCad's negated-angle convention:

```python
angle_rad = math.radians(-rotation_deg)
cos_a = math.cos(angle_rad)
sin_a = math.sin(angle_rad)
rotated_x = local_x * cos_a - local_y * sin_a
rotated_y = local_x * sin_a + local_y * cos_a
```

Three existing call sites already use it correctly (rotate, then translate by `footprint.position`):
- `validate/connectivity.py::_transform_pad_position` (~line 1127-1136)
- `validate/rules/clearance.py::_transform_pad_position` (~line 142-156)
- `validate/rules/via_in_pad.py::_pad_absolute_bbox` (~line 61-74)

`_check_drill_clearance` should follow this exact pattern:

```python
from kicad_tools.core.geometry import rotate_pad_offset
...
rotated_x, rotated_y = rotate_pad_offset(pad.position[0], pad.position[1], fp.rotation)
abs_x = fp.position[0] + rotated_x
abs_y = fp.position[1] + rotated_y
```

### Sibling occurrences of the same unrotated-addition pattern

A repo-wide grep for `fp.position[0] + pad.position[0]` / `fp.position[1] + pad.position[1]` found **3 more occurrences**, all in `src/kicad_tools/validate/rules/solder_mask.py`:
- line 100-101 (`_check_solder_mask_clearance` — used for the reported violation's `location` field only)
- line 143-144 (`_check_min_pad_size` — used for the reported violation's `location` field only)
- line 195-196 (`_check_pth_annular_ring` — used for the reported violation's `location` field only)

**These are NOT geometrically equivalent to the dimensions.py bug and are OUT OF SCOPE for this issue.** In all three `solder_mask.py` sites, `abs_x`/`abs_y` feed *only* the `location` tuple attached to the reported `DRCViolation` (for display/report purposes) — the pass/fail determination in each of those three rules is computed from pad-local dimensions (`pad.size`, `pad.solder_mask_margin`, `pad.drill`) and does not depend on absolute position at all. So on a rotated footprint, `solder_mask.py` violations would report a slightly wrong location marker but would not produce false positives/negatives. This is a real (minor) bug worth a follow-up issue, but it is a different defect class from `dimensions.py`'s false-positive/false-negative geometry bug, and fixing it is not required for this issue's acceptance criteria. **A follow-up issue should be filed for `solder_mask.py`** referencing this one; do not silently fold it into this PR's scope.

### Reconciliation with #4017 / PR #4035 (board-04 hole-to-hole)

**Verdict: unaffected — different code path, no overlap.** #4017's board-04 violations (`GND<->NRST`, `OSC_OUT<->NRST`, both 0.350mm) were between **vias**, not THT footprint pads. `_check_drill_clearance` adds via entries to its `drills` list directly from `via.position` (no footprint-relative addition, no rotation involved — vias have absolute board-frame positions in the schema). PR #4035's description confirms: "These are in-pad micro vias on the U2 LQFP-48 west escape" — via-to-via, not pad-to-pad or via-to-pad. Board-04's fix (relocating a via 0.5mm east) stands on its own and required no rotation awareness. This issue is limited to the THT-pad branch of `_check_drill_clearance` (`pad.type == "thru_hole"`), which is a separate code path from the via-collection loop just above it.

### Test fixture gap

`tests/test_validate_dimensions.py::MockFootprint.__init__` (line ~115-123) does not have a `rotation` parameter at all — every existing drill-clearance test (including `test_through_hole_pad_clearance`, line ~494) implicitly uses `rotation=0`, which is exactly the "test trap" class of bug the `rotate_pad_offset` docstring already warns about (0°/180° rotations agree under both correct and incorrect sign conventions; only 90°/270° or mirrored footprints expose the bug). The fix must add a `rotation: float = 0.0` param to `MockFootprint` to enable regression testing.

## Acceptance Criteria

- [ ] `_check_drill_clearance` computes THT pad board positions using `rotate_pad_offset(pad.position[0], pad.position[1], fp.rotation)` + `fp.position` translation, matching the pattern in `connectivity.py` / `clearance.py` / `via_in_pad.py`.
- [ ] `MockFootprint` in `tests/test_validate_dimensions.py` gains a `rotation: float = 0.0` constructor parameter.
- [ ] New regression test: a footprint rotated 90 deg (or 180/270) with a THT pad at a non-zero local offset, plus a second hole placed at the *correct* (rotation-applied) safe distance from the true pad position but within the *naive* (unrotated) distance -> asserts **zero** `dimension_drill_clearance` violations (this is the false-positive case from the issue).
- [ ] New regression test: same rotated-footprint setup, but with a second hole placed at a genuinely too-close position relative to the *true* rotated pad location -> asserts the violation **is still detected** (guards against the fix silently disabling the check, e.g. false negative).
- [ ] Existing 0 deg (unrotated) drill-clearance tests continue to pass unmodified (regression guard on the "test trap" the `rotate_pad_offset` docstring warns about).
- [ ] If verification against a real board (e.g. re-running `kct check` on any board with rotated THT footprints) produces a diff in a **committed** `drc_report.json`, report the diff as a finding in the PR description rather than silently regenerating/committing it — committed manufacturing artifacts are treated as shipping truth per repo convention (see `.claude/projects/.../MEMORY.md` "Board-05 status" note: "artifact-first... committed artifact = shipping truth").

## Implementation Guidance

Minimal, surgical fix — mirror the existing correct pattern exactly:

```python
# validate/rules/dimensions.py, inside _check_drill_clearance, replace:
abs_x = fp.position[0] + pad.position[0]
abs_y = fp.position[1] + pad.position[1]

# with:
from kicad_tools.core.geometry import rotate_pad_offset  # add to module-level imports
...
rotated_x, rotated_y = rotate_pad_offset(pad.position[0], pad.position[1], fp.rotation)
abs_x = fp.position[0] + rotated_x
abs_y = fp.position[1] + rotated_y
```

Do not touch the via-collection loop above (vias are already correct — absolute positions, no footprint-local offset). Do not touch `solder_mask.py` in this PR (see "Sibling occurrences" above); file a separate follow-up issue for it instead.

## Affected Files

- `src/kicad_tools/validate/rules/dimensions.py` - Fix `_check_drill_clearance` (~lines 211-212) to rotate pad-local offset by `fp.rotation` via `core.geometry.rotate_pad_offset` before translating by `fp.position`.
- `tests/test_validate_dimensions.py` - Add `rotation` parameter to `MockFootprint` (~line 115-123); add regression tests for rotated-footprint false-positive (no violation) and genuine-violation-still-detected cases.

## Out of Scope (follow-up)

- `src/kicad_tools/validate/rules/solder_mask.py` (lines ~100-101, ~143-144, ~195-196) has the same unrotated-addition pattern, but it only affects the reported violation's display `location`, not the pass/fail determination. File as a separate low-severity follow-up issue; do not fold into this fix.

## Test Plan

- [ ] Manual verification: construct a synthetic PCB (via `MockFootprint`/`MockPad` or a real `.kicad_pcb`) with a footprint rotated 90 deg or 180 deg and a THT pad at non-zero local offset; confirm `_check_drill_clearance`'s computed `abs_x`/`abs_y` matches `rotate_pad_offset` output, not naive addition.
- [ ] Automated tests: add the two new cases to `tests/test_validate_dimensions.py` described in Acceptance Criteria (false-positive-eliminated, genuine-violation-still-caught); run `uv run pytest tests/test_validate_dimensions.py -v`.
- [ ] Edge cases: verify 0 deg/180 deg cases still pass (these can accidentally "pass" under a wrong-sign rotation implementation per the `rotate_pad_offset` docstring's own "test trap" warning) — the new tests must use 90 deg or 270 deg (or a non-cardinal angle) to be a meaningful regression guard.
- [ ] If any committed `drc_report.json` changes as a side effect of manual board verification, report the delta in the PR description as a finding rather than silently regenerating it.

rjwalters:
Closes #4044

## Problem

`validate/rules/dimensions.py::_check_drill_clearance` computed a through-hole pad's board position by plain addition, ignoring footprint rotation:

```python
abs_x = fp.position[0] + pad.position[0]
abs_y = fp.position[1] + pad.position[1]
```

`Pad.position` is stored **footprint-local** and must be rotated by `footprint.rotation` before translation (documented in `schema/pcb.py`'s `Pad` docstring). Because these `abs_x`/`abs_y` values feed directly into the center-to-center distance math that drives the violation determination, this is a correctness bug: on any rotated/flipped footprint the THT hole centers were placed at the wrong location (~10mm off in the reported real-board case), producing phantom `dimension_drill_clearance` violations and potentially missing real ones.

## Fix

Mirrors the existing correct pattern already used in `validate/connectivity.py::_transform_pad_position`, `validate/rules/clearance.py::_transform_pad_position`, and `validate/rules/via_in_pad.py`: rotate the pad-local offset via the pcbnew-verified `core.geometry.rotate_pad_offset` (KiCad's negated-angle convention, #3739), then translate by `fp.position`.

```python
rotated_x, rotated_y = rotate_pad_offset(pad.position[0], pad.position[1], fp.rotation)
abs_x = fp.position[0] + rotated_x
abs_y = fp.position[1] + rotated_y
```

The via-collection loop above is untouched (vias carry absolute board-frame positions — no footprint-local offset, no rotation).

## Tests

- Added a `rotation: float = 0.0` parameter to `MockFootprint` (previously it had no rotation param at all — every existing drill-clearance test implicitly used `rotation=0`, exactly the "test trap" the `rotate_pad_offset` docstring warns about, since 0deg/180deg agree under both correct and incorrect sign conventions).
- `test_rotated_footprint_no_false_positive`: a 90deg footprint with a THT pad at non-zero local offset, plus a via 0.7mm from the **naive** (buggy) position but ~5.8mm from the **true** rotated position — asserts **zero** violations (the false-positive case).
- `test_rotated_footprint_genuine_violation_detected`: same 90deg setup, via genuinely 0.3mm edge-to-edge from the **true** rotated pad hole — asserts the violation **is still detected** (guards against a false negative / the fix silently disabling the check).
- All 3 pre-existing 0deg drill-clearance tests continue to pass unmodified.

`uv run pytest tests/test_validate_dimensions.py` → **35 passed**. Related suites (`test_connectivity.py`, `test_component_clearance.py`) → green. `ruff check` + `ruff format --check` clean on both touched files.

## Fleet-impact finding

**No committed manufacturing artifacts change.** I scanned every board for the only footprint class this fix affects — through-hole pads at a non-zero local offset on a footprint with `rotation != 0`:

- The 4 boards with committed `output/drc_report.json` (00-simple-led, 03-usb-joystick, 04-stm32-devboard, 05-bldc-motor-controller): **0** such footprints each; none of those reports currently contain any `dimension_drill_clearance` entry.
- Full fleet scan across all `boards/*/output/*_routed.kicad_pcb`: **0** rotated-THT-pad(non-zero-offset) instances total.

Since the fix only relocates hole centers for that footprint class, its effect on the current fleet is a provable no-op — no `drc_report.json` regeneration was needed or performed.

## Out of scope (recommended follow-up)

`validate/rules/solder_mask.py` has the same unrotated-addition pattern at 3 sites (`_check_solder_mask_clearance` ~L100, `_check_min_pad_size` ~L143, `_check_pth_annular_ring` ~L195), but there `abs_x`/`abs_y` feed **only** the reported violation's display `location` tuple — the pass/fail determination in each is computed from pad-local dimensions and does not depend on absolute position. On a rotated footprint those rules would report a slightly wrong location marker but produce no false positives/negatives. Recommend a separate low-severity follow-up issue; deliberately not folded into this PR per the curated scope.

### DIFF
--- a/src/kicad_tools/validate/rules/dimensions.py
+++ b/src/kicad_tools/validate/rules/dimensions.py
@@ -9,6 +9,8 @@ from __future__ import annotations
 import math
 from typing import TYPE_CHECKING
 
+from kicad_tools.core.geometry import rotate_pad_offset
+
 from ..violations import DRCResults, DRCViolation
 from .base import DRC_TOLERANCE, DRCRule
 
@@ -207,9 +209,18 @@ class DimensionRules(DRCRule):
         for fp in pcb.footprints:
             for pad in fp.pads:
                 if pad.type == "thru_hole" and pad.drill > 0:
-                    # Pad position is relative to footprint
-                    abs_x = fp.position[0] + pad.position[0]
-                    abs_y = fp.position[1] + pad.position[1]
+                    # Pad position is stored footprint-local and must be
+                    # rotated by the footprint orientation before translating
+                    # to board coordinates (KiCad's negated-angle convention;
+                    # see core.geometry.rotate_pad_offset, issue #3739). A
+                    # naive add ignores rotation and produces phantom
+                    # hole-to-hole violations on rotated/flipped footprints
+                    # (issue #4044).
+                    rotated_x, rotated_y = rotate_pad_offset(
+                        pad.position[0], pad.position[1], fp.rotation
+                    )
+                    abs_x = fp.position[0] + rotated_x
+                    abs_y = fp.position[1] + rotated_y
                     net_name = pad.net_name if pad.net_name else f"net:{pad.net_number}"
                     drills.append(
                         (
--- a/tests/test_validate_dimensions.py
+++ b/tests/test_validate_dimensions.py
@@ -117,10 +117,16 @@ class MockFootprint:
         reference: str,
         position: tuple[float, float],
         pads: list[MockPad],
+        rotation: float = 0.0,
     ):
         self.reference = reference
         self.position = position
         self.pads = pads
+        # Footprint orientation in degrees. Pad positions are stored
+        # footprint-local and must be rotated by this angle before
+        # translation to board coordinates (issue #4044). Defaults to 0.0
+        # so existing tests are unaffected.
+        self.rotation = rotation
 
 
 class MockPCB:
@@ -528,6 +534,109 @@ class TestDrillClearanceCheck:
             v for v in results.violations if v.rule_id == "dimension_drill_clearance"
         ]
         assert len(clearance_violations) == 1
+
+    def test_rotated_footprint_no_false_positive(self):
+        """Issue #4044: a rotated THT footprint must not phantom-flag.
+
+        A footprint at (10, 10) rotated 90 deg has pad 1 at local offset
+        (0, 4). KiCad's negated-angle convention (rotate_pad_offset) maps
+        that to board offset (4, 0), so the true hole center is (14, 10).
+
+        - Naive (buggy) add:  fp.position + pad.position = (10, 14)
+        - Rotation-correct:                                 (14, 10)
+
+        A via is placed at (14, 1.0) -- i.e. 9.0mm from the *true* pad
+        position (safe: 9.0 - 0.2 - 0.2 = 8.6mm >> 0.5mm) but only 1.0mm
+        from the *naive* (14 vs 10 in x is 4mm; but y 14 vs 1 is 13mm...).
+
+        Concretely the via is placed 0.7mm from the buggy position and
+        far from the true one, so the naive code flags a hole-to-hole
+        violation that does not exist in the rotated geometry.
+        """
+        # Buggy position: (10, 14). True position: (14, 10).
+        # Via placed at (10, 14.7): 0.7mm from buggy pos (would flag),
+        # but sqrt((14-10)^2 + (10-14.7)^2) = sqrt(16 + 22.09) = 6.17mm
+        # from the true pos (edge = 6.17 - 0.4 = 5.77mm, safe).
+        pcb = MockPCB(
+            vias=[
+                MockVia((10, 14.7), size=0.6, drill=0.4, layers=["F.Cu", "B.Cu"], net_number=1),
+            ],
+            footprints=[
+                MockFootprint(
+                    reference="J1",
+                    position=(10, 10),
+                    rotation=90.0,
+                    pads=[
+                        MockPad(
+                            number="1",
+                            type="thru_hole",
+                            position=(0, 4),  # local; rotates to board (4, 0)
+                            drill=0.4,
+                            net_name="NET2",
+                            net_number=2,
+                        ),
+                    ],
+                ),
+            ],
+            nets={1: MockNet(1, "NET1"), 2: MockNet(2, "NET2")},
+        )
+        rules = MockDesignRules(min_clearance_mm=0.127, min_hole_to_hole_mm=0.5)
+        rule = DimensionRules()
+
+        results = rule.check(pcb, rules)
+
+        clearance_violations = [
+            v for v in results.violations if v.rule_id == "dimension_drill_clearance"
+        ]
+        # Buggy code would find the via 0.7mm from the (10,14) pad -> flag.
+        # Correct code measures against (14,10) -> 5.77mm edge -> no flag.
+        assert len(clearance_violations) == 0
+
+    def test_rotated_footprint_genuine_violation_detected(self):
+        """Issue #4044: a real hole-to-hole conflict at rotation still flags.
+
+        Same 90 deg footprint whose pad 1 true hole center is (14, 10).
+        A via placed at (14.7, 10) is a genuine 0.3mm edge-to-edge conflict
+        (0.7 center - 0.2 - 0.2) with the *true* rotated pad position. The
+        fix must not silently disable the check (guard against a false
+        negative).
+        """
+        # True pad pos (14, 10). Via at (14.7, 10): center dist 0.7mm,
+        # edge = 0.7 - 0.2 - 0.2 = 0.3mm < 0.5mm -> genuine violation.
+        pcb = MockPCB(
+            vias=[
+                MockVia((14.7, 10), size=0.6, drill=0.4, layers=["F.Cu", "B.Cu"], net_number=1),
+            ],
+            footprints=[
+                MockFootprint(
+                    reference="J1",
+                    position=(10, 10),
+                    rotation=90.0,
+                    pads=[
+                        MockPad(
+                            number="1",
+                            type="thru_hole",
+                            position=(0, 4),  # local; rotates to board (4, 0)
+                            drill=0.4,
+                            net_name="NET2",
+                            net_number=2,
+                        ),
+                    ],
+                ),
+            ],
+            nets={1: MockNet(1, "NET1"), 2: MockNet(2, "NET2")},
+        )
+        rules = MockDesignRules(min_clearance_mm=0.127, min_hole_to_hole_mm=0.5)
+        rule = DimensionRules()
+
+        results = rule.check(pcb, rules)
+
+        clearance_violations = [
+            v for v in results.violations if v.rule_id == "dimension_drill_clearance"
+        ]
+        assert len(clearance_violations) == 1
+        assert clearance_violations[0].actual_value == pytest.approx(0.3)
+        assert clearance_violations[0].required_value == pytest.approx(0.5)
 
     def test_smd_pads_not_checked(self):
         """Test that SMD pads are not included in drill clearance check."""


rjwalters:
✅ **Approved!**

Thorough verification of the rotation-aware drill-clearance fix — all claims independently confirmed.

## Correctness (convention match)
The fix calls `rotate_pad_offset(pad.position[0], pad.position[1], fp.rotation)` then translates by `fp.position` — **byte-identical call shape** to the three existing correct usages:
- `clearance.py::_transform_pad_position` (L151-155)
- `via_in_pad.py::_pad_absolute_bbox` (L73-75)
- `connectivity.py::_transform_pad_position`

All use the pcbnew-verified negated-angle convention (#3739). The via-collection loop is correctly left untouched: `Via.position` is a plain absolute board-frame coordinate (schema confirms no footprint parent / no local offset), so no rotation applies. `fp.rotation` is the correct attribute; `Pad.rotation` (absolute) is correctly not involved.

## The 90° vs non-cardinal question (load-bearing check)
The curation warned that a wrong-sign rotation at 90° could "accidentally produce a correct-looking distance for symmetric fixtures," demanding non-cardinal rotation. I verified this concern **does not apply to this fixture** via a mutation test:

- Fixture pad local `(0, 4)` maps to board offset `(+4, 0)` under the correct convention vs `(-4, 0)` under a flipped sign — i.e. **opposite sides of the footprint origin**, not coincident. The fixture is not symmetric under the sign flip.
- I flipped `-rotation_deg → rotation_deg` in `core.geometry.rotate_pad_offset` and re-ran: `test_rotated_footprint_genuine_violation_detected` **FAILED** (0 violations vs expected 1), because its via at `(14.7, 10)` sits 0.7mm from the true center `(14,10)` but 8.7mm from the wrong-sign center `(6,10)`.

So the 90° test genuinely discriminates the sign error. `test_rotated_footprint_no_false_positive` alone would not (both centers are ~6mm away), but the paired genuine-violation test is the one that pins the sign — and it does. A non-cardinal case would be strictly more robust in the abstract, but for THIS fixture geometry the criterion is not load-bearing; the sign is nailed down. (0°/180° would indeed be a trap; 90° here is not.)

## Fleet impact (spot-check confirmed)
Builder's claim of zero rotated-THT-nonzero-offset footprints verified two independent ways across all 8 committed `*_routed.kicad_pcb`:
- Parser scan (`PCB.load` + `fp.rotation`): **0** rotated footprints total, 0 rotated-THT pads.
- Raw grep for 3-token `(at x y <angle!=0>)`: **0** occurrences in every board, including board 05 (55 fps, all `angle=0`).

The fix only relocates hole centers for rotated footprints, so its effect on the current fleet is a provable no-op — no `drc_report.json` regeneration needed. Confirmed.

## Tests / lint / CI
- `pytest tests/test_validate_dimensions.py` → **35 passed** locally; the 3 pre-existing 0° tests unmodified.
- All **16 CI checks pass** (Test 13m8s, all 8 board E2E/regressions green including Board 05 19m13s). Merge state **CLEAN**, MERGEABLE.
- Disjoint from sibling PR #4068 (that PR touches `boards/*/generate_design.py` + `test_board_*_partial_route_fastfail.py`; no file overlap) — no conflict.

## Note (already handled by builder)
The out-of-scope `solder_mask.py` unrotated-addition sites feed only display `location`, not pass/fail — builder correctly deferred to a follow-up. Good scoping.

Clean, well-tested, mirrors the established pattern exactly. Nice work.

Issue #3839, opened 2026-06-21
ci/board-04: main() success gate omits DRC and board-04 is not built in CI (regressions ship undetected)


rjwalters:
## Original Issue

Child of epic #3829.

boards/04-stm32-devboard/generate_design.py main() return gate (~lines 1838-1849) is roughly "0 if (erc and route and stitch and tie and fill and mfr) else 1" — it OMITS drc_success and has no LVS gate, so a board with DRC errors / a real short exits 0 and prints SUCCESS. Compounding this, board-04 is NOT built in CI at all (ci.yml builds 00/01/02/03/06/07; 04 and 05 absent until #3823 added 05) — so nothing catches a board-04 regression.

NOTE: the #3829 curator found the OSC_IN<->OSC_OUT short and the LVS gate were already addressed by #3797/#3794; the remaining gaps are specifically (1) the main() gate omitting drc_success, and (2) board-04 having no CI job.

Fix: (1) include drc_success (and LVS) in the board-04 main() success gate so the recipe exits non-zero on DRC/LVS failure; (2) add a board-04 CI job (mirror the board-03-end-to-end pattern, or the #3823 board-05 re-route+gate pattern) that regenerates board-04 and validates the RESULT (DRC + copper-LVS) out-of-process. If a fresh board-04 regen is nondeterministic or doesnt reach clean, calibrate the gate to the measured deterministic baseline with documentation (like #3823 did for board-05) rather than faking a pass.

Acceptance: board-04 recipe exits non-zero on DRC/LVS failure; a board-04 CI job exists and validates a fresh regen; the job is green (at a documented baseline if full-clean isnt reproducible).

Files: boards/04-stm32-devboard/generate_design.py; .github/workflows/ci.yml; scripts/ci/ (new gate if needed).

---

### Problem Statement

Two independent defects let a broken board-04 ship undetected:

[...]

PR #4068, opened 2026-07-12
fix(boards): route_success fast-fail gate for boards 00/01/02/04/06 (#4066)

### LINKED ISSUE #4066: Sibling board recipes missing the route_success fast-fail gate (boards 00/01/02/04/06)
Follow-up documented in PR #4047's body ("Note: siblings share the latent pattern"). Reference pattern now canonical in the `.claude/commands/kct/board-recipe-scaffold.md` skill (added by PR #4064, merged 2026-07-12).

## Problem

PR #4047 (merged, closes #4027) added a `route_success` fast-fail gate to `boards/03-usb-joystick/generate_design.py:838`: when `route_pcb()` returns a partial route, `main()` now raises a distinct "partial route -- likely wall-clock budget exhaustion under load" error **before** GND stitching / zone fill / `write_lvs_report`. Without that gate, a partial route falls through to `write_lvs_report(..., require_clean=True)`, where an unrouted signal net reaches the copper comparator as a GND OPEN and raises `BoardNetlistMismatch` — the broad `try/except` in `main()` then reports exit 1 with a **misleading LVS traceback**, misdirecting the reviewer to the LVS / GND-stitching subsystem when the true cause is upstream route truncation (a loaded machine buys fewer CPU cycles inside the fixed wall-clock `--timeout` backstop). This actually flaked board 03 twice on 2026-07-09.

The sibling recipes share the identical latent pattern: they capture `route_success` from `route_pcb()` and then run a `require_clean=True` LVS gate unconditionally, with no gate in between.

## Verified affected boards (read on `main`, 2026-07-12)

Affected = captures `route_success` but never gates it before a `require_clean=True` LVS call:

| Board | route_success captured | LVS call | require_clean | Gated before LVS? | Affected |
|-------|:---:|-----------|:---:|:---:|:---:|
| `boards/00-simple-led` | yes (:949) | `run_lvs()` -> `write_lvs_report` (:915) | **True** (raises `BoardNetlistMismatch`) | no | **YES** |
| `boards/01-voltage-divider` | yes (:792) | `write_lvs_report` (:805) | **True** | no | **YES** |
| `boards/02-charlieplex-led` | yes (:787) | `write_lvs_report` (:800) | **True** | no | **YES** |
| `boards/04-stm32-devboard` | yes (:1954) | `write_lvs_report` (:2032) | **True** | no (route_success is in the *final* exit AND, but LVS runs first and raises) | **YES** |
| `boards/06-diffpair-test` | yes (:3403) | `write_lvs_report` (:3421) | **True** | no | **YES** |

Correctly EXCLUDED (verified, differ from the pattern — matches PR #4047's deferral scope and the scaffold skill's carve-out):

- `boards/07-matchgroup-test`: LVS is `require_clean=False` (advisory — partial-by-design board with 5 known #3438 opens). A partial route cannot raise `BoardNetlistMismatch` here, so the misattribution bug does not apply. Per the scaffold skill: the fast-fail gate is required only "while LVS is a `require_clean=True` hard gate."
- `boards/05-bldc-motor-controller`: uses `design.py` (not `generate_design.py`) and has no `write_lvs_report` call — different pipeline shape; not exposed to this bug.

(The board-04 nuance: its final `return 0 if (... and route_success and ...)` DOES reference `route_success`, but `write_lvs_report` at :2032 runs *before* that exit gate, so a partial route still raises the misleading LVS trace before the exit AND is ever evaluated. It needs the same early gate.)

## Proposed solution

For each of boards 00/01/02/04/06, add the `route_success` fast-fail gate in `main()` **immediately after `route_pcb(...)` returns and before GND stitching / zone fill / `write_lvs_report`**, mirroring `boards/03-usb-joystick/generate_design.py:838` and the canonical pattern documented in `.claude/commands/kct/board-recipe-scaffold.md` (step 5.5, "route_success fast-fail gate"). On `not route_success`, `raise` a distinct, clearly-worded error that names route truncation / wall-clock budget as the cause and explicitly disclaims a copper-LVS / GND-stitching cause.

Per board, add a hermetic test mirroring `tests/test_board_03_copper_lvs.py::TestBoard03PartialRouteFastFail` (monkeypatch the recipe's module-level functions; NO router / `kicad-cli` / LVS invocation):
- `route_pcb -> False` makes `main()` exit non-zero with a "partial route" / "wall-clock budget" message; forbidden downstream steps (GND stitching, zone fill, `run_drc`, `write_lvs_report`) raise if the gate lets them run; assert no `BoardNetlistMismatch` in the output.
- `route_pcb -> True` still reaches `write_lvs_report` (no false positive on the happy path).

## Acceptance

- Each of boards 00/01/02/04/06 gains the `route_success` fast-fail gate before its `require_clean=True` LVS call, matching the scaffold-skill pattern and board 03's implementation.
- Each affected board gains a hermetic partial-route fast-fail test mirroring `TestBoard03PartialRouteFastFail` (per-board, hermetic, no external tooling).
- Fresh-regen happy paths are unaffected (the gate only fires on `not route_success`); the existing clean-regen tests still pass.
- Boards 05 and 07 are intentionally out of scope (documented above).

## Notes
- Reference implementation: `boards/03-usb-joystick/generate_design.py:838` (PR #4047).
- Reference test: `tests/test_board_03_copper_lvs.py:214` (`TestBoard03PartialRouteFastFail`) + its `_load_board03_module()` loader helper.
- Canonical pattern doc: `.claude/commands/kct/board-recipe-scaffold.md` step 5.5 (PR #4064).
- The affected-board table is verified as of 2026-07-12 but the curator should re-verify against `main` before implementation (recipes change; LVS gate policy per board can shift a board in or out of scope).

rjwalters:
Closes #4066

Follow-up to PR #4047 (which added the gate to board 03). Propagates the `route_success` fast-fail gate to the five sibling recipes that share the identical latent pattern (capture `route_success`, then run a `require_clean=True` LVS gate with nothing in between).

## Problem

When `route_pcb()` returns a partial route — the wall-clock `--timeout` safety backstop fires before every signal net lands under concurrent CPU load — `main()` previously fell through to `write_lvs_report(..., require_clean=True)`, where the unrouted net reaches the copper comparator as a GND OPEN and raises `BoardNetlistMismatch`. The broad `try/except` in `main()` then reports exit 1 with a **misleading LVS traceback**, misdirecting the reviewer to the LVS / GND-stitching subsystem when the true cause is upstream route truncation. This flaked board 03 twice on 2026-07-09.

## Per-board gates

Each gate lands immediately after `route_pcb()` returns and before any pour/stitch/DRC/LVS step, raising a distinct `RuntimeError` that names "partial route" / "wall-clock budget" and explicitly disclaims a copper-LVS / GND-stitching cause. Mirrors `boards/03-usb-joystick/generate_design.py:838` and `.claude/commands/kct/board-recipe-scaffold.md` step 5.5.

| Board | Gate placed after | Before |
|---|---|---|
| `00-simple-led` | `route_pcb` capture | `run_drc` / `run_lvs` |
| `01-voltage-divider` | `route_pcb` capture | `run_drc` / `write_lvs_report` |
| `02-charlieplex-led` | `route_pcb` capture | `run_drc` / `write_lvs_report` |
| `04-stm32-devboard` | `route_pcb` capture | the 5 copper-mutating steps (`fix_osc_escape`/`stitch_pcb`/`tie_power_pads`/`quantize_escapes`/`fill_zones`), then DRC/LVS |
| `06-diffpair-test` | `route_pcb` capture (inside `--step all`) | `run_drc` / `write_lvs_report` |

## Scope guards honored

- **Board 04 exit-gate `and route_success` preserved.** The pre-existing `and route_success` term in the final `return 0 if (...)` exit gate (#3839, now at `:2187`) is left in place as defense-in-depth. The new early `raise` makes it unreachable on the partial-route path, but it is neither removed nor duplicated — a future refactor downgrading the raise to a soft warning would still gate the exit code.
- **Board 06 gate scoped to `--step all` only.** The gate lives strictly inside the `if args.step == "all":` branch. The separate `--step route` Phase 4N CI re-route path (#2677) — which intentionally tolerates partial routes (USB3_TX1+/- blocked by the BGA partner-via escape) and calls no LVS — is untouched, preserving that CI contract.

Boards 05 (`design.py` pipeline, no `write_lvs_report`) and 07 (`require_clean=False` advisory LVS) remain intentionally out of scope.

## Tests

Five new hermetic test files (one per board) mirroring `tests/test_board_03_copper_lvs.py::TestBoard03PartialRouteFastFail`. Each monkeypatches the recipe's module-level functions (no router / `kicad-cli` / LVS invocation):

- `test_partial_route_fails_fast_with_distinct_message`: `route_pcb -> False`, all downstream steps forbidden (raise `AssertionError` if reached); asserts `main()` returns 1, stderr contains "partial route" + "wall-clock budget" (case-insensitive) and NOT `BoardNetlistMismatch`.
- `test_full_route_still_reaches_lvs_gate`: `route_pcb -> True`, downstream no-op'd; asserts the LVS entrypoint (`run_lvs` for board 00, `write_lvs_report` for 01/02/04/06) is actually invoked and `main()` returns 0.

Board 04's forbid-list is the widest (all five copper-mutating steps) so the gate can't accidentally sit after `fix_osc_escape` and still pass. Board 02's loader prepends the board dir to `sys.path` (its recipe does a top-level `from design_spec import ...`). Board 06's tests invoke default `--step all` so they exercise exactly the gated branch.

```
10 passed in 0.58s   # the five new test files (2 tests each)
65 passed, 3 deselected in 25.46s   # existing fast committed-artifact tests for touched boards
```

`ruff check` + `ruff format --check` clean on all 10 touched files.

## Notes

- **Slow fresh-regen test NOT run.** The curation permits one representative slow test (`TestBoard03FreshRegenCopperLVSClean`) *only if load is light*. At implementation time the machine load average was ~8.6 (a concurrent sibling builder is running) and the C++ router backend is not installed in this worktree, so a fresh Python-fallback regen would risk a spurious wall-clock timeout that reflects load, not this change. The change is byte-identical on the happy path (the gate only fires on `not route_success`), the `test_full_route_still_reaches_lvs_gate` hermetic tests prove the happy path is unaffected, and this PR does not touch board 03, whose slow test already passes on main.
- **Pre-existing doc/code drift (informational, not touched):** `src/kicad_tools/lvs/recipe.py`'s `ADVISORY_LVS_BOARDS` docstring says advisory boards "pass `require_clean=False`", but board 04's actual call site passes `require_clean=True`. The code is authoritative (which is exactly why board 04 is in scope here). Left untouched per the issue's guidance.

### DIFF
--- a/boards/00-simple-led/generate_design.py
+++ b/boards/00-simple-led/generate_design.py
@@ -725,13 +725,24 @@ def route_pcb(input_path: Path, output_path: Path) -> bool:
     print("\n8. Filling copper zones...")
     _fill_zones_after_route(output_path)
 
-    total_nets = len([n for n in router.nets if n > 0])
+    # Count only the nets that were actually submitted for TRACE routing.
+    # ``router.nets`` keys include VCC, which is a pour net
+    # (``is_pour_net=True``): the router logs "Skipping N pour net(s)" and
+    # connects it via a copper zone in step 7 rather than as a trace.  GND is
+    # already excluded because ``load_pcb_for_routing(skip_nets=["GND"])``
+    # collapses its pads to net 0.  Reuse the router's own ``_is_pour_net``
+    # classifier -- the same predicate that drives ``_filter_pour_nets`` -- so
+    # ``total_nets`` reflects the trace-routable signal nets only.  Without
+    # this the count double-includes the never-routed VCC pour net, making
+    # ``success`` always False even on a clean run (issue #4066).
+    trace_nets = [n for n in router.nets if n > 0 and not router._is_pour_net(n)]
+    total_nets = len(trace_nets)
     success = stats["nets_routed"] == total_nets
 
     if success:
-        print("\n   SUCCESS: All nets routed!")
+        print("\n   SUCCESS: All signal nets routed!")
     else:
-        print(f"\n   PARTIAL: Routed {stats['nets_routed']}/{total_nets} nets")
+        print(f"\n   PARTIAL: Routed {stats['nets_routed']}/{total_nets} signal nets")
 
     return success
 
@@ -948,28 +959,31 @@ def main() -> int:
         route_success = route_pcb(pcb_path, routed_path)
 
         # Step 5.5: route_success fast-fail gate (#4066, mirrors board 03's
-        # gate at boards/03-usb-joystick/generate_design.py:838).  route_pcb
-        # runs under a wall-clock ``--timeout`` SAFETY backstop layered above
-        # the load-independent per-net ``--deterministic-budget`` iteration
-        # cap, so on a loaded machine that outer deadline can fire before every
-        # signal net lands and ``route_pcb`` returns ``False``.  If we fall
-        # through, the downstream ``run_lvs`` -> ``write_lvs_report(
-        # require_clean=True)`` sees a genuinely unrouted signal net as a copper
-        # OPEN and raises ``BoardNetlistMismatch``, which the broad ``except``
-        # below reports as exit 1 -- misdirecting the reviewer to the LVS
-        # subsystem when the true cause is upstream route truncation.  Raise a
-        # DISTINCT, clearly-worded error here instead.  The "PARTIAL: Routed
-        # N/M signal nets" line is already printed above by ``route_pcb``.
+        # gate at boards/03-usb-joystick/generate_design.py:838).  ``route_pcb``
+        # returns ``False`` when at least one trace-routable SIGNAL net failed
+        # to land (``nets_routed`` < the pour-net-excluded ``total_nets``; see
+        # the count accounting in ``route_pcb``).  If we fall through, the
+        # downstream ``run_lvs`` -> ``write_lvs_report(require_clean=True)`` sees
+        # that unrouted signal net as a copper OPEN and raises
+        # ``BoardNetlistMismatch``, which the broad ``except`` below reports as
+        # exit 1 -- misdirecting the reviewer to the LVS subsystem when the true
+        # cause is upstream route truncation.  Raise a DISTINCT, clearly-worded
+        # error here instead.  The "PARTIAL: Routed N/M signal nets" line is
+        # already printed above by ``route_pcb``.
+        #
+        # NOTE: board 00 calls bare ``router.route_all()`` with no ``--timeout``
+        # backstop, so a partial route here means a genuinely unroutable signal
+        # net (or a net-count accounting slip), NOT a wall-clock budget timeout.
         if not route_success:
             raise RuntimeError(
-                "partial route -- likely wall-clock budget exhaustion under "
-                "load (the --timeout safety backstop fired before every "
-                "signal net landed; see the 'PARTIAL: Routed N/M signal nets' "
-                "line above for the exact count). This is NOT a copper-LVS / "
-                "GND-stitching failure -- the pipeline stopped before the LVS "
-                "gate. Re-run boards/00-simple-led/generate_design.py in "
-                "isolation on a quiet machine, or raise the --timeout in "
-                "route_pcb() if this recurs on an unloaded host."
+                "partial route -- at least one trace-routable signal net did "
+                "not land (see the 'PARTIAL: Routed N/M signal nets' line above "
+                "for the exact count). This is NOT a copper-LVS / GND-stitching "
+                "failure -- the pipeline stopped before the LVS gate. Board 00 "
+                "routes with a bare route_all() (no --timeout), so this is not a "
+                "wall-clock budget exhaustion: investigate the unrouted signal "
+                "net directly, or re-check the pour-net-excluded net-count "
+                "accounting in route_pcb() if the count looks wrong."
             )
 
         # Step 6: Run DRC
--- a/boards/00-simple-led/generate_design.py
+++ b/boards/00-simple-led/generate_design.py
@@ -947,6 +947,31 @@ def main() -> int:
         routed_path = output_dir / "simple_led_routed.kicad_pcb"
         route_success = route_pcb(pcb_path, routed_path)
 
+        # Step 5.5: route_success fast-fail gate (#4066, mirrors board 03's
+        # gate at boards/03-usb-joystick/generate_design.py:838).  route_pcb
+        # runs under a wall-clock ``--timeout`` SAFETY backstop layered above
+        # the load-independent per-net ``--deterministic-budget`` iteration
+        # cap, so on a loaded machine that outer deadline can fire before every
+        # signal net lands and ``route_pcb`` returns ``False``.  If we fall
+        # through, the downstream ``run_lvs`` -> ``write_lvs_report(
+        # require_clean=True)`` sees a genuinely unrouted signal net as a copper
+        # OPEN and raises ``BoardNetlistMismatch``, which the broad ``except``
+        # below reports as exit 1 -- misdirecting the reviewer to the LVS
+        # subsystem when the true cause is upstream route truncation.  Raise a
+        # DISTINCT, clearly-worded error here instead.  The "PARTIAL: Routed
+        # N/M signal nets" line is already printed above by ``route_pcb``.
+        if not route_success:
+            raise RuntimeError(
+                "partial route -- likely wall-clock budget exhaustion under "
+                "load (the --timeout safety backstop fired before every "
+                "signal net landed; see the 'PARTIAL: Routed N/M signal nets' "
+                "line above for the exact count). This is NOT a copper-LVS / "
+                "GND-stitching failure -- the pipeline stopped before the LVS "
+                "gate. Re-run boards/00-simple-led/generate_design.py in "
+                "isolation on a quiet machine, or raise the --timeout in "
+                "route_pcb() if this recurs on an unloaded host."
+            )
+
         # Step 6: Run DRC
         drc_success = run_drc(routed_path)
 
--- a/boards/01-voltage-divider/generate_design.py
+++ b/boards/01-voltage-divider/generate_design.py
@@ -791,6 +791,31 @@ def main() -> int:
         routed_path = output_dir / "voltage_divider_routed.kicad_pcb"
         route_success = route_pcb(pcb_path, routed_path)
 
+        # Step 5.5: route_success fast-fail gate (#4066, mirrors board 03's
+        # gate at boards/03-usb-joystick/generate_design.py:838).  route_pcb
+        # runs under a wall-clock ``--timeout`` SAFETY backstop layered above
+        # the load-independent per-net ``--deterministic-budget`` iteration
+        # cap, so on a loaded machine that outer deadline can fire before every
+        # signal net lands and ``route_pcb`` returns ``False``.  If we fall
+        # through, the downstream ``write_lvs_report(require_clean=True)`` sees
+        # a genuinely unrouted signal net as a copper OPEN and raises
+        # ``BoardNetlistMismatch``, which the broad ``except`` below reports as
+        # exit 1 -- misdirecting the reviewer to the LVS subsystem when the
+        # true cause is upstream route truncation.  Raise a DISTINCT,
+        # clearly-worded error here instead.  The "PARTIAL: Routed N/M signal
+        # nets" line is already printed above by ``route_pcb``.
+        if not route_success:
+            raise RuntimeError(
+                "partial route -- likely wall-clock budget exhaustion under "
+                "load (the --timeout safety backstop fired before every "
+                "signal net landed; see the 'PARTIAL: Routed N/M signal nets' "
+                "line above for the exact count). This is NOT a copper-LVS / "
+                "GND-stitching failure -- the pipeline stopped before the LVS "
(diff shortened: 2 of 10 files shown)


rjwalters:
❌ **Changes Requested — the board 00 gate deterministically breaks a required CI check**

The scope guards (board 04 exit-gate `and route_success` preserved at `:2187`; board 06 gate confined to the `--step all` branch) and the insertion points for boards 01/02/04/06 all check out. But the **board 00 gate is built on a premise that is false for board 00**, and it takes down a previously-green required check.

## The blocker (Board 00 End-to-End: FAIL)

`Board 00 End-to-End` is red on this PR (`mergeStateStatus: UNSTABLE`). The failing step is *"Regenerate board 00 from recipe"* — i.e. the new gate itself fired:

```
   PARTIAL: Routed 1/2 nets
RuntimeError: partial route -- likely wall-clock budget exhaustion under load ...
##[error]Process completed with exit code 1.
```

This is **not** a load-dependent wall-clock partial route. It is deterministic and reproduces on every run, because board 00's `route_pcb()` return value is computed differently from its siblings:

`boards/00-simple-led/generate_design.py:728-729`
```python
total_nets = len([n for n in router.nets if n > 0])   # counts VCC (a pour net) + LED_ANODE = 2
success = stats["nets_routed"] == total_nets           # nets_routed = 1 (LED_ANODE only)
```

VCC is a **pour net** — the CI log confirms `Skipping 1 pour net(s) (use zone fill instead): ['VCC']`. It is intentionally never trace-routed (it's filled as a zone in step 7). So `nets_routed` (1) can never equal `total_nets` (2), and `route_pcb()` returns `False` on **every** run. This pre-existing pour-net double-count is a latent false-negative that was harmless on main: the recipe fell through to LVS, which passed (VCC is connected via the zone pour, so no OPEN). That is why `Board 00 End-to-End` was green on main.

Your gate converts that always-`False` return into a hard `RuntimeError`. The PR's happy-path safety argument ("byte-identical; the gate only fires on `not route_success`") is exactly the trap here: **board 00's `route_success` is always `False`**, so the gate always fires. The error message it prints ("wall-clock budget exhaustion under load", "raise the --timeout") is also actively misleading for board 00 — `route_pcb` calls bare `router.route_all()` with *no* timeout at all (CI emits the issue-#2794 "called without per_net_timeout or timeout" UserWarning), so there is no `--timeout` backstop to raise.

Contrast with the siblings that pass CI: boards 01/02 derive `success` from `kct route`'s return code (`result.returncode == 0`), and board 04's route genuinely completes in CI. Only board 00 has the pour-net miscount, so only board 00 breaks.

## What's needed

The board 03 pattern you're mirroring is sound *when `route_success=False` actually means a truncated route*. That invariant does not hold for board 00 as written. Options, in order of preference:

1. **Fix board 00's `success` computation first** so `total_nets` excludes pour/skip nets (e.g. count only nets that were actually submitted for trace routing, mirroring how the router already excludes VCC). Then `route_pcb()` returns `True` on a clean run and the gate becomes correct and inert on the happy path — matching the board 03 semantics. This is the right fix and unblocks the gate.
2. **Drop board 00 from this PR** and file the pour-net double-count as a separate issue, landing gates only for the boards where `route_success=False` genuinely implies a partial route (01/02/04/06). Boards 01/02/04/06 CI is green.

Either way, re-verify `Board 00 End-to-End` goes green before re-requesting review.

## Everything else verified clean (for when this is unblocked)

- **Board 04 scope guard** — new early gate at `:1974`; pre-existing `and route_success` exit term (#3839) preserved at `:2187`, neither removed nor duplicated. ✅
- **Board 06 scope guard** — gate lives inside `if args.step == "all":`; the `--step route` Phase 4N re-route contract (#2677) is untouched. ✅
- Insertion points for 01/02/04/06 match the curated line numbers; distinct partial-route messages disclaim a copper-LVS cause. ✅
- `ruff` / format clean; mergeable (no conflicts) against current main (post-#4067). ✅
- CI for boards 01/02/04/06 End-to-End (fresh regens) all green — the happy path is adequately covered there, so skipping the local slow board-03 regen is fine. ✅

The design is 90% right — the missing piece is that board 00's `route_success` doesn't carry the meaning the gate assumes.

rjwalters:
Board-00 blocker addressed (judge option **(a)** — fix the success computation).

**Root cause fixed:** `route_pcb` counted VCC in `total_nets`. VCC is a pour net (`is_pour_net=True`), connected via a copper zone in step 7 and intentionally never trace-routed (`Skipping 1 pour net(s): ['VCC']`), so `nets_routed` (1, LED_ANODE) could never equal `total_nets` (2) → `route_pcb` returned `False` every run. Harmless on main (fell through to a passing LVS); the new gate turned it into a deterministic `RuntimeError`.

**How pour nets are now excluded** (`boards/00-simple-led/generate_design.py`):
```python
trace_nets = [n for n in router.nets if n > 0 and not router._is_pour_net(n)]
total_nets = len(trace_nets)
success = stats["nets_routed"] == total_nets
```
This reuses the router's own `_is_pour_net` classifier — the same predicate that drives `_filter_pour_nets` and emits the "Skipping N pour net(s)" log — so `total_nets` now reflects trace-routable **signal** nets only. GND was already excluded (`load_pcb_for_routing(skip_nets=["GND"])` collapses it to net 0). `success` is now `True` on a clean run and the gate is inert on the happy path, matching board 03 semantics. The gate itself is preserved.

**Misleading message fixed:** board 00 routes with a bare `router.route_all()` (no `--timeout`), so the gate message no longer blames "wall-clock budget exhaustion" or tells the reader to raise a nonexistent `--timeout`; it now points at the genuinely unrouted signal net / net-count accounting.

**Verification (local):**
- Fresh board-00 regen — green end-to-end: `SUCCESS: All signal nets routed!`, DRC PASS, ERC PASS, LVS PASS (label + copper), MFG bundle PASS, exit 0. (C++ backend built + available for the run.)
- `tests/test_board_00_partial_route_fastfail.py` — extended with `TestBoard00RoutePcbNetCount` (stubs the router factory: pour net excluded → full route returns `True`; unrouted signal net → `False`); stale `"wall-clock budget"` assertion updated to the corrected message. 4 passed.
- Sibling gates untouched: `test_board_01/02/04/06_partial_route_fastfail.py` — 8 passed.
- `test_board_00_lvs.py` + `test_board_00_simple_led.py` — 12 passed.

Scope: only `boards/00-simple-led/generate_design.py` and its test file changed. Boards 01/02/04/06 not touched. Ready for re-review.

rjwalters:
✅ **Approved!** (follow-up re-review after Doctor commit 28abc4be)

The board-00 blocker from the first review is resolved. Scoped re-review of the delta:

**1. Delta scope — CONFIRMED**
Commit 28abc4be touches exactly the two expected files:
- `boards/00-simple-led/generate_design.py` (60 lines)
- `tests/test_board_00_partial_route_fastfail.py` (138 lines)

No unexpected surface. The fix does **not** hardcode `"VCC"` — it reuses the router's own classifier: `trace_nets = [n for n in router.nets if n > 0 and not router._is_pour_net(n)]`. GND is already collapsed to net 0 by `load_pcb_for_routing(skip_nets=["GND"])`, VCC is now excluded as a pour net, so `total_nets` counts only trace-routable signal nets and a clean run returns `True`. The gate is now inert on the happy path, matching board 03 semantics.

**2. `_is_pour_net` privacy — acceptable, not blocking**
`_is_pour_net` is underscore-private on `Autorouter`, but it is the canonical pour classifier for the whole router — it's the exact predicate behind `_filter_pour_nets` (core.py) and is reused in `monte_carlo.py` and `evolutionary.py`. There is no public equivalent (`is_pour_net` is a per-net-class flag, not a net-id method), and the board recipe already reaches into the same class surface (`router.nets`, `router.get_statistics()`). Reusing the same predicate that decides what the router *skips* is the correct way to decide what to *count*. Consistent with repo practice.

**3. CI — GREEN (the acid test)**
All 16 checks pass. Board 00 End-to-End, which was deterministically **failing** at first review, now passes in 1m32s. Merge state CLEAN.
- Board 00 End-to-End: pass (1m32s)
- Board 01/02/03/04 End-to-End: pass
- Board 05 Routing Regression: pass (19m3s)
- Board 06/07 End-to-End: pass
- Diff-Pair / Match-Group Regression: pass
- Test: pass (16m10s), Type Check / Lint / C++ Build / DRC / round-trip: pass

**4. Test coverage — both directions pinned**
`TestBoard00RoutePcbNetCount` stubs the router factory via `_FakeRouter` (VCC=net 1 pour, LED_ANODE=net 2 signal, GND absent) and covers both directions:
- `test_clean_run_excludes_pour_net_and_returns_true` — `nets_routed=1` → `route_pcb` returns `True` (pour-excluded happy path)
- `test_unrouted_signal_net_returns_false` — `nets_routed=0` → returns `False` so the #4066 gate fires

The stale `"wall-clock budget"` assertion in the existing fastfail test was correctly updated to `"signal net"`, matching the corrected gate message (board 00 routes with bare `route_all()`, no `--timeout`, so a partial route is a genuine unrouted signal net). Ran locally: all 4 tests in the suite pass (17.21s).

Boards 01/02/04/06 gates and both scope guards were verified in the first review and are unchanged. Ship it.

Closes #4066.

Issue #3673, opened 2026-06-14
Epic: crossing-aware river-routing planner for dense parallel pad-array bundles


rjwalters:
## Epic: crossing-aware river-routing planner for dense parallel pad-array bundles

Consolidates the **congestion / 4L-production-reach** gaps that an evidence-backed investigation (see #3438) confirmed need a **new river-routing planner**, not a bounded knob — every documented knob has been tried across 3+ prior sessions.

### Root cause (from #3438, with negative controls)
The DDR bus-reversal bundle stalls at 9/11 **even alone, with 4 routable layers** — ruling out layer capacity and layer-preference policy. In-column vias are forbidden (0.6 mm via vs 0.8 mm pitch), so all bundle nets escape onto the same F.Cu strip first and the outermost nets get sealed out. The relief/rescue machinery probes + displaces correctly, but serial re-land re-runs the same greedy order with a different loser — the **N-1 invariant**. The genuine fix is a crossing-aware river-routing / planar bundle planner that allocates the whole bundle's escape corridors jointly rather than greedily per-net.

### Blocked children rolled up here
- #3438 (board 07 stuck at 28/31 — DDR/MIPI/HDMI parallel pad-array bundles)
- #3414 (board 07 4L production routing 84% → 100%)
- #3412 (board 05 4L production routing 60% → 100%)

These cap manufacturability (100% routed) on boards 05/07.

rjwalters:
Design-capture tracking record (backlog → zero). The river-routing planner design + the N-1 greedy-escape-invariant evidence are preserved in the issue body above and remain readable while closed. Caps board 05/07 4L manufacturability. **Reopen when** the multi-session planner is scheduled.

Issue #2983, opened 2026-05-17
router: layered-escape strategy for mirrored byte-lane DDR (decouples via placement from net ordering)


rjwalters:
## Context

Follow-up from PR #2969 (closes #2962).  That PR landed the
``Autorouter._apply_byte_lane_inner_priority`` scaffolding hook and
three integration sites (``route_all``, ``route_all_negotiated``,
``TwoPhaseRouter``) but proved over three iterations that
**net-ordering alone cannot resolve board 07's mirrored-byte-lane
DDR failure mode**.  The geometric constraint -- vias colliding on
adjacent 0.8mm-pitch pads on the mirrored QFN-48 row -- requires a
strategy that decouples via placement from priority.

## Design history (PR #2969 R1/R2/R3 trace)

The full trace is preserved in
``Autorouter._apply_byte_lane_inner_priority``'s docstring and in
``tests/router/test_byte_lane_priority.py``'s module docstring.
Summary:

| Round | Plan | Yield | DRC | Verdict |
|---|---|---|---|---|
| 1 | Demote BOTH corner (0, n-1) AND second-inward (2, n-3) | 27/31 | 86 | Over the 70 DRC allowlist; corner demotion forced detours through tight-clearance geometry |
| 2 | Demote only second-inward (2, n-3) | 20/31 | 4 | DRC OK but yield regressed and ``match_group_length_skew`` was silently not exercised |
| 3 | Promote inner-corner (1, n-2) to rank 0 directly | 24/31 | 12 | DRC well under allowlist, but DQ5 still blocked by DQ4 at 0.44mm (identical to R2) -- proves geometric constraint |

The failure signature (CI run 25975551444 / job 76357754718,
``ca42fcd2``):

```
Net 9 \"DQ5\": pin_access (PADS_OFF_GRID: U1.33 off by 0.187mm |
  Clearance blocked by: U1.33: blocked by DQ4 (via at 0.44mm); ...)
Net 8 \"DQ4\": pin_access (... U1.32: blocked by DQ5 (pad at 0.68mm); ...)
Net 7 \"DQ3\": pin_access (... U1.28: blocked by DQ2 (via at 0.49mm); ...)
```

DQ5 is blocked by **a DQ4 via** -- not by DQ4's route ordering.  The
via placement (and the lack of an alternate escape lane) is the real
blocker.

## Root cause

Pin row order on U1.25-35 is
[...]

PR #4060, opened 2026-07-12
feat(router): reactive escape-freedom byte-lane reorder (opt-in, off by default)

rjwalters:
## Summary

Phase 1b of epic #4049. Extends `_apply_byte_lane_inner_priority` from an identity-order return to a **reactive escape-freedom net reorder** for mirrored byte-lane bundles (board 07's DDR data byte on facing QFN-48 columns), the one variant #3438 had not yet isolated. The existing inner-corner corridor reservation (#2983/PR #2987) is preserved.

**Measured result: the reorder does NOT reach 11/11 — it regresses the DDR bundle 10/11 → 9/11.** Per the epic's measurement-first framing and the curator's gating guidance, it therefore ships **off by default** behind `enable_byte_lane_reorder`. With the flag off the method is byte-identical to pre-#4051 behaviour, so no board regresses by construction. The scheduler + diagnostics are retained behind the flag for future non-uniform bundle geometries.

## Changes

- **`_schedule_by_escape_freedom`** (new): seeds the two row extremes (least-free corner nets), then walks inward advancing whichever frontier is currently most threatened by its just-committed neighbour — reacts to committed geometry rather than a static up-front permutation.
- **`_reactive_escape_freedom_order`** (new): permutes only the slots each byte-lane group already occupies, preserving non-group ordering (starvation fairness) and the permutation invariant.
- **`enable_byte_lane_reorder`** flag on `Autorouter` (default `False`) gating the reorder; detection + corridor reservation still run and identity is returned when off.
- Updated stale "scaffolding only (identity return)" / "future follow-up issue" docstring and the three call-site comments (`route_all`, `route_all_negotiated`, `TwoPhaseRouter`) to reflect that #2983 shipped and #4051 adds reordering on top.
- Tests: byte-lane priority + corridor-reservation suites updated for the opt-in contract; added `_schedule_by_escape_freedom` unit tests.

## Measurement (DDR-bundle repro, back-to-back, solo)

| Config | Routed | Stranded | Wall |
|--------|--------|----------|------|
| Baseline (flag off = identity = `main`) | 10/11 | DQ3 | 3m18s |
| Reactive reorder (flag on) | 9/11 | DQ2, DQ3 | 7m12s |

Confirms #3438's N-1 structural-capacity conclusion: escape-sequence reordering (static or reactive) does not clear this bundle; joint corridor allocation (Phase 3 / #4053) remains the fallback.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
[...]

PR #4070, opened 2026-07-12
feat(router): scoped bundle river planner v1 for facing-column bus reversals (#4053)

### LINKED ISSUE #4053: Board 07 Phase 3: scoped river-routing planner for straight facing-column bundles

rjwalters:
## Summary

Scoped river-routing planner v1 for straight facing-column bundles (Phase 3
of epic #4049). Implements the curator-defined design: resolve both facing
rows of a mirrored byte-lane group, diff their permutation to derive the
required-crossing (inverted-pair) set, and reserve one **soft** inner-layer
via-hop corridor per inverted pair so the losing net of each crossing can
dip to an inner layer and pass under its partner. Everything gates behind a
new default-OFF flag `enable_bundle_river_planner` (CLI: `--bundle-river-planner`).

Closes #4053.

## What v1 actually does

- **`src/kicad_tools/router/bundle_river.py`** (new): pure, search-free
  crossing-set computation. `compute_facing_row_inversions(primary_row,
  secondary_row)` sorts both rows, diffs their permutation, and emits the
  inverted pairs. A **full bus reversal** yields all C(n,2) pairs; a
  **co-oriented (planar)** bundle yields an **empty** set (the
  over-triggering guard). Free functions so the math is unit-testable
  without a router/board.
- **`escape.py`**: `reserve_bundle_river_via_hop_corridor` generalizes
  `reserve_inner_corner_lane_corridor` (#2983) from a single straight lane
  to a via-hop corridor that runs forward + laterally along the row axis
  (the under-pass). Same SOFT mechanic (`reserve_corridor_cells` +
  net-owned keep-out + A* attractor) — **not** hard pre-routed copper,
  **not** the coupled-search primitive, exactly as the curation scoped.
- **`core.py`**: `_apply_byte_lane_inner_priority` gains
  `_reserve_bundle_river_via_hops`, which resolves the SECONDARY facing
  component's row (the existing scan only resolved the primary), calls
  `bundle_river`, and reserves one via-hop per losing net. Gated so
  flag-off is byte-identical to pre-#4053 main.
- CLI flag wired through `parser.py` + `route_cmd.py` inner parser +
  `commands/routing.py` bridge, applied via `_apply_bundle_river_planner`
  at all four `_apply_ripup_budget_override` call sites.

Verified the detection is correct on the real board: the board-07 DDR
bundle resolves to U1/U2 (11 pads each), a **full 11-net reversal → 55
inversions → 10 distinct losing nets**; the planner reserves via-hop
corridors (9 total reservations, 7425 cells) on In1.Cu.

## Measurements (DDR-bundle repro, seed 42, solo, `--targeted-ripup --max-ripups-per-net 3`)

| Backend | `--bundle-river-planner` | Routed | Stranded | Wall |
|---------|--------------------------|--------|----------|------|
| C++ (production) | off | **10/11** | DQ3 | 164s |
| C++ (production) | on | **10/11** | DQ3 | 165s |
| Python | off | 7/11 | DQ1, DQ2, DQ5, DQS_N | 241s |
| Python | on | **8/11** | DQ1, DQ2, DQS_N | 161s |

**The 11/11 acceptance bar is NOT met.** Honest findings:

1. **On the production C++ backend the via-hop reservation is completely
   inert** — the C++ path is **byte-identical** flag-on vs flag-off (same
   6003-line routed file, same 22 vias). Root cause (pre-existing, not
   introduced here): the C++ grid deliberately omits the corridor-reservation
   map — see `cpp/src/grid.cpp:126-143` ("This C++ implementation deliberately
   omits that check because the escape phase is Python-grid-only today").
   `reserve_corridor_cells` (grid.py) has no C++ sync, so the C++ A* never
   sees the keep-out **or** the attractor bonus. The soft-reservation
   surface the curation scoped is structurally unreachable from the
   production routing loop.
2. **On the Python backend (which DOES consult the attractor) the planner
   works as designed and helps: 7/11 → 8/11** (DQ5 rescued), and faster.
   This confirms the via-hop mechanism is sound — but the Python backend's
   absolute reach (8/11) is far below the C++ baseline (10/11), and +1 net
   still doesn't clear the full 55-crossing reversal.

The crossing-set analysis validates the curation's planarity correction
(the bundle IS a full reversal, not a planar bundle), and the mechanism
proves out on the attractor-honoring backend. The gap to 11/11 is that a
single-strip SOFT via-hop is not strong enough to force 10 nets to hop, and
on the shipping backend the reservation has no effect at all.

## Flag-off regression safety

- DDR repro C++ flag-on == flag-off byte-for-byte (diff empty, 6003 lines).
- Board-02 (charlieplex) routes 8/8 flag-off — no regression on a
  representative production board.
- Flag-off skips 100% of the new code path; boards 00-06 are structurally
  unaffected (CI covers the route gates).

## Deferred scope (v2 / follow-up)

- **The blocker: port corridor reservations + the attractor to the C++
  grid** (`cpp/src/grid.cpp` `mark_via` + a pathfinder attractor hook).
  Until this lands, NO soft-reservation planner can affect production
  routing — this is the real prerequisite the epic's evidence now points
  to, and it also gates board-06's USB3 escape fix per the grid.cpp
  comment. Recommend a dedicated child.
- Auto-detect safe engagement (only after the flag has a positive track
  record); N>2 facing columns / partial-nested reversals (#3673 scope);
  jointly planning the DQS_P/DQS_N member via the C++ coupled loop (#4065).

## Tests / lint / mypy

- `tests/router/test_bundle_river_inversions.py` (11) — crossing-set math
  (full reversal C(n,2); planar empty; partial reversal; non-matched rows
  rejected; deterministic loser selection).
- `tests/router/test_bundle_river_planner.py` (5) — flag-off byte-identical
  (2 reservations); flag-on reversal reserves via-hops (>2); flag-on planar
  reserves none (guard); order stays identity.
- Existing `test_byte_lane_priority.py` + `test_byte_lane_corridor_reservation.py`
  (27) unchanged and green.
- ruff check + format clean on all touched files.
- mypy baseline gate PASSES: 1496 errors, all within the 1514 baseline, no
  new type errors (18 baseline errors removed by line shifts).

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 4024
-# Branch: feature/issue-4024
+# Issue: 4053
+# Branch: feature/issue-4053
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -203,6 +203,11 @@ def run_route_command(args) -> int:
     # explicitly-provided values (matches the early-stop-patience pattern).
     if getattr(args, "targeted_ripup", False):
         sub_argv.append("--targeted-ripup")
+    # Issue #4053: forward --bundle-river-planner.  Both parsers declare it
+    # as store_true defaulting to False, so only forward when the user set
+    # it (byte-identical when absent).
+    if getattr(args, "bundle_river_planner", False):
+        sub_argv.append("--bundle-river-planner")
     max_ripups_val = getattr(args, "max_ripups_per_net", None)
     if max_ripups_val is not None:
         sub_argv.extend(["--max-ripups-per-net", str(max_ripups_val)])
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -2746,6 +2746,23 @@ def _add_route_parser(subparsers) -> None:
             "and the two-phase initial-pass stall recovery."
         ),
     )
+    route_parser.add_argument(
+        "--bundle-river-planner",
+        action="store_true",
+        default=False,
+        help=(
+            "Enable the scoped bundle river planner for mirrored byte-lane "
+            "bus reversals (Issue #4053, epic #4049).  Board 07's DDR data "
+            "byte is a FULL bus reversal between two facing QFN-48 pin "
+            "columns (all C(11,2)=55 pairs cross), which planar same-layer "
+            "lane ordering cannot solve (every ordering-only approach "
+            "capped at <=10/11).  Resolves both facing rows, diffs their "
+            "permutation, and reserves one inner-layer via-hop corridor per "
+            "inverted (crossing) pair so the losing net can dip under its "
+            "partner.  Default OFF (byte-identical when absent); DDR-bundle "
+            "scoped in v1."
+        ),
+    )
     # Issue #3054 (Phase 2 of #3045): wire region-based parallelism through to
     # ``route_all_negotiated``.  Opt-in (default off) so existing scripts and
     # CI runs see byte-identical routes; when set, the negotiated loop
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -2785,6 +2785,21 @@ def _apply_ripup_budget_override(router: "Autorouter", args) -> None:
     router.stall_ripup_budget = budget
 
 
+def _apply_bundle_river_planner(router: "Autorouter", args) -> None:
+    """Enable the scoped bundle river planner when requested (Issue #4053).
+
+    Off by default (mirrors ``enable_byte_lane_reorder`` / the #4051
+    precedent).  When ``--bundle-river-planner`` is passed, set the
+    router's ``enable_bundle_river_planner`` flag so
+    ``_apply_byte_lane_inner_priority`` reserves inner-layer via-hop
+    corridors for the inverted (crossing) pairs of a mirrored byte-lane
+    bus reversal (board 07's DDR byte).  A no-op when the flag is absent,
+    so production routing is byte-identical to pre-#4053 main.
+    """
+    if getattr(args, "bundle_river_planner", False):
+        router.enable_bundle_river_planner = True
+
+
 def _targeted_ripup_budget(args) -> int:
     """Resolve ``--max-ripups-per-net`` for the negotiated targeted path.
 
@@ -3558,6 +3573,7 @@ def route_with_layer_escalation(
         # Issue #3470: thread --max-ripups-per-net into the destructive
         # rip-up budgets (route_all + two-phase stall recovery).
         _apply_ripup_budget_override(router, args)
+        _apply_bundle_river_planner(router, args)
 
         # Issue #3171: inject boosted analog routing class for --analog-nets /
         # --auto-analog selected nets (pour/ground nets are left untouched).
@@ -4457,6 +4473,7 @@ def route_with_rule_relaxation(
         # Issue #3470: thread --max-ripups-per-net into the destructive
         # rip-up budgets (route_all + two-phase stall recovery).
         _apply_ripup_budget_override(router, args)
+        _apply_bundle_river_planner(router, args)
 
         # Issue #3171: inject boosted analog routing class for --analog-nets /
         # --auto-analog selected nets (pour/ground nets are left untouched).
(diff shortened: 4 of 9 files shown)


rjwalters:
✅ **Approved!** Merging as flag-gated infrastructure + honest measurement (default-OFF `enable_bundle_river_planner`). This follows the PR #4060 precedent for shipping a flag-off mechanism with a documented activation prerequisite.

## What I verified (verify-don't-trust)

**1. Flag-off identity (the load-bearing safety property) — CONFIRMED**
- Code path: the entire new via-hop block in `core.py` is behind `if getattr(self, "enable_bundle_river_planner", False):` (core.py:5085), placed *after* the pre-existing #2983 inner-corner reservation loop. The detection scan (`comp_pad_count`, `net_to_pad`, `sorted_nets`, `cx`/`cy`) was already computed for the #2983 path — no new work runs flag-off. `_reserve_bundle_river_via_hops` (including the secondary-row resolve and `compute_facing_row_inversions`) only runs inside the flag guard.
- CLI: `--bundle-river-planner` defaults False in both parsers; `_apply_bundle_river_planner` is a no-op when unset; forwarded only when the user set it. Applied at all four `_apply_ripup_budget_override` sites.
- **Empirical proof**: all 8 board regression CI jobs (00-07, diff-pair, match-group, board-05 blocking) are green — they route WITHOUT the flag, so their passing IS the flag-off identity evidence. Board-07 stays at its known-opens baseline.
- `test_bundle_river_planner.py` flag-off tests assert byte-identical reservation counts (only the #2983 baseline).

**2. Crossing-set math — CONFIRMED empirically** (`bundle_river.py`)
Ran the edge cases directly: full reversal n=5 → 10 = C(5,2); planar/co-oriented → empty (over-trigger guard); odd n=3 → 3; single-net → 0; partial permutation (single swap) → exactly 1 correctly-named pair; mismatched net sets → 0 (None guard); loser dedup works (4 distinct, not 10). 11 math tests green.

**3. Reservation plumbing — CONFIRMED soft-only**
`reserve_bundle_river_via_hop_corridor` (escape.py) writes copper via exactly one grid mutation: `self.grid.reserve_corridor_cells` (soft). No `mark_route`/`add_segment`/`add_via`/`mark_via` — no hard pre-routed copper. Everything else is read-only geometry. Generalizes #2983's single-lane pattern as scoped.

**4. C++ inertness finding — CONFIRMED, and honest**
`Grid3D::mark_via` (cpp/src/grid.cpp:144) genuinely omits the `_reserved_for_nets` check; the comment at 126-143 documents the deliberate Python-only reservation contract and *predicts* the DDR-style regression. This is pre-existing (references #2709/#2677/#2686), not introduced here — which correctly implies #2983's shipped reservations have been inert on C++ all along. The PR body reports the flag-on == flag-off C++ byte-identity plainly and does NOT claim production improvement. #4071 correctly captures the activation prerequisite.

**5. Tests / lint / mypy — CONFIRMED**
43 tests pass locally (16 new + 27 byte-lane unchanged). ruff check + format clean on all touched files. Type Check CI green (mypy baseline gate).

**6. CI — all 16 checks pass; merge state CLEAN; `Closes #4053` present.**

**7. Merits**: The dormant code is not hazardous — unreachable flag-off, soft-only reservations, advisory `try/except` degradation to the pre-existing path. The Python backend (which honors the attractor) demonstrates the mechanism works (7/11→8/11, DQ5 rescued). The honest partial-outcome reporting is exemplary. Merging as measured infrastructure with #4071 as the documented activation blocker.

Issue #4073, opened 2026-07-12
3D: apply LCSC tier to fleet waived parts (Joystick_Analog, PCIE_Mini_Edge) — closes #4032


rjwalters:
## Context

**Depends on #4072** (the LCSC/EasyEDA fetch-on-demand resolver tier). This issue *applies* that tier to the fleet's two remaining bare footprints and closes **#4032**. It is **not** marked `loom:blocked` — sweeps run these in order; #4072 must land first.

**#4032** waived two synthetic footprints from the #4014/#4033 substitution work because no cleanly-redistributable vendor STEP or installed KiCad library covers them (read the #4032 dossier comment):

| Board | ref | lib_id | Facts (verified in generators) |
|---|---|---|---|
| `boards/03-usb-joystick` | J2 | `Module:Joystick_Analog` | 5 THT pads, single row, **2.0mm pitch**, span -4…+4mm (`generate_pcb.py`). BOM already carries LCSC **C50950**. |
| `boards/06-diffpair-test` | J3 | `Connector_PCIE:PCIE_Mini_Edge` | synthetic single-row abstraction, **12 SMD pads at 0.8mm pitch** (generator docstring calls it a 52-pin card-edge modeled as a pad row). BOM value is `generic` (no C-number). |

**Scope correction (verified):** the owner brief referenced "boards 06/07" for PCIE_Mini_Edge, but **only board 06 contains it**. Board 07 (`07-matchgroup-test`) BOM has no joystick and no PCIE part. This issue therefore touches boards **03 and 06** only. (Board 07 was likely conflated because 06/07 share other diffpair fixtures.)

Both repo footprints are **synthetic testbench geometries**, so even a correct vendor STEP is an approximate body — the #4045 pad-centroid offset machinery handles placement and the gallery goal is a **representative body, centroid-placed**. The 2.0mm-vs-2.54mm joystick pitch approximation is **ACCEPTED** on that basis.

## Proposed work

1. **Select LCSC C-numbers** for representative bodies:
   - **Joystick_Analog** — a **KY-023-class** 2-axis analog thumbstick module. Board 03's BOM already assigns **C50950**; verify it has an EasyEDA 3D model and is a KY-023-class body before adopting it (else pick a KY-023-class C-number that does).
[...]

PR #4074, opened 2026-07-12
docs: chorus install pilot evidence + --mfr flag for the vendored routed-DRC gate

rjwalters:
## Summary

Epic #4054 **acceptance pilot** (issue #4058): installs kicad-tools into the
`../chorus` consumer repo, proves idempotency and Loom-safety, imports the v22
manufacturing board, and runs the full manufacturing-readiness flow
(`kct check` + `kicad-cli pcb drc --refill-zones` cross-gate + `kct export`) at
`jlcpcb-tier1`, plus the vendored `.kct/ci` gates. All findings, the drift
classification, the idempotency proof, and the Loom-untouched proof are captured
in `docs/install-pilot-chorus.md`.

The pilot surfaced one genuine installer/gate bug, fixed here.

## Changes

- **`docs/install-pilot-chorus.md`** (new): the evidence report — install
  transcript, idempotency proof (byte-identical vendored files, single CLAUDE.md
  block/dep), Loom-untouched proof (0-line diff), Gate 1/2/3 results with a
  per-delta classification, `.kct/ci` gate results, friction list, and the chorus
  pilot branch name + diffstat for owner review.
- **`scripts/ci/check_routed_drc.py`**: add a `--mfr TIER` CLI flag — a
  per-invocation manufacturer-profile override (precedence `--mfr` > YAML
  `manufacturers:` map > `jlcpcb` default). Without it, a fresh consumer repo
  could not gate a non-default-tier board at its real tier without authoring the
  kicad-tools-repo-internal `.github/routed-drc-tolerance.yml`.
- **`scripts/install-kct.sh`**: vendored `.kct/ci/README.md` now documents the
  `--mfr` usage for consumers.
- **`tests/test_check_routed_drc_advisory.py`**: `TestMfrCliOverride` — default is
  `jlcpcb`; `--mfr jlcpcb-tier1` threads through to `kct check`; `--help`
  advertises the flag.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| `install-kct.sh --path` completes cleanly on fresh chorus pilot branch | ✅ | First run: all 8 stages OK; artifacts written |
| Second run idempotent (no dup CLAUDE.md block/dep, exit 0) | ✅ | Byte-identical vendored files (shasum diff empty); 1 marker block; 1 dep; Stage 5 "no-op" |
| Loom install byte-for-byte unchanged | ✅ | `git diff main -- .claude/commands/loom .loom loom.sh` = 0 lines after both runs |
[...]

PR #4075, opened 2026-07-13
feat(3d): LCSC/EasyEDA fetch-on-demand model resolver tier (#4072)

### LINKED ISSUE #4032: 3D: source vendor STEP files for Joystick_Analog and PCIE_Mini_Edge (waived from #4014)
## Context

Follow-up split out from #4014 (cross-library 3D-model substitution tier).

That issue automated 3D-model resolution for 5 of 7 still-bare footprints across boards 03/06/07 via a curated `lib_id -> lib_id` substitution table (`src/kicad_tools/pcb/model_substitutions.py`). Two footprints could **not** be resolved automatically because no installed KiCad library — and no open/redistributable substitute — covers them:

| Board | lib_id | Why unresolvable |
|---|---|---|
| 03-usb-joystick | `Module:Joystick_Analog` | Synthetic lib id invented by the generator (`boards/03-usb-joystick/generate_pcb.py`). No `Module.pretty` library exists in KiCad at all; there is no body-compatible installed footprint to substitute. |
| 06-diffpair-test | `Connector_PCIE:PCIE_Mini_Edge` | No installed KiCad library covers PCIe mini-edge connectors (closest — `Connector_Card`, `Connector_PCBEdge`, `Connector_Samtec_EdgeRate` — are not visually equivalent). |

## Why this needs a human / operator

A builder in the automation pipeline cannot browse the internet or download binary vendor artifacts. Resolving these requires:

1. Visiting a vendor site (e.g. Alps or equivalent for the analog joystick; a PCIe Mini Card-Edge connector vendor).
2. Confirming the STEP file's redistribution / license terms.
3. Vendoring the binary `.step` under `data/3dmodels/` with recorded license terms.
4. Wiring a `(model ...)` reference for each into the model resolver (either a real `.kicad_mod` under a repo-local library, or a direct-path entry in the substitution/resolver layer).

## Acceptance criteria

- [ ] A redistributable STEP body is vendored for `Module:Joystick_Analog` (board 03) with recorded license terms, OR a permanent waiver is documented if no open source exists.
- [ ] A redistributable STEP body is vendored for `Connector_PCIE:PCIE_Mini_Edge` (board 06) with recorded license terms, OR a permanent waiver is documented.
- [ ] `kct pcb add-3d-models` resolves both lib ids (or the resolver surfaces the direct-path model), and the committed board 03/06 artifacts are regenerated.
- [ ] Pure-insertion guarantee preserved (models are metadata-only; no DRC/copper deltas).

## References

- Parent issue: #4014
- Substitution table: `src/kicad_tools/pcb/model_substitutions.py`
- Resolver: `src/kicad_tools/pcb/models3d.py` (`make_library_resolver`)

### LINKED ISSUE #4072: 3D: LCSC/EasyEDA fetch-on-demand model resolver tier (fourth tier in add-3d-models)
## Context

The fleet targets **JLCPCB assembly**, whose assembly parts are **LCSC parts (C-numbers)** — and nearly all LCSC parts carry a 3D STEP model in the EasyEDA parts database. This is a large, ready-made source of representative bodies for exactly the footprints the KiCad-official resolver cannot reach.

Today `kct pcb add-3d-models` (`src/kicad_tools/pcb/models3d.py`) resolves 3D models from **installed KiCad libraries only**, in three tiers inside `make_library_resolver`:

1. **Exact** installed `library:name`.
2. **Same-library variant** (`_find_variant_mod`).
3. **Cross-library substitution** (`model_substitutions.py`, curated `lib_id -> lib_id`, from #4014/#4033).

Since **PR #4045** (merged), the patcher computes a `target_centroid - source_centroid` pad-centroid delta and rewrites the inserted model's `(offset (xyz ...))` (Y negated vs the footprint 2D frame) so bodies land correctly on kct-generated footprints. This offset machinery runs uniformly for all three tiers in `add_model_refs_to_text` — any new tier that returns a `ResolvedModels(models=..., source_anchor=...)` inherits it for free.

Synthetic/generic lib ids that no installed KiCad library covers (`Module:Joystick_Analog`, `Connector_PCIE:PCIE_Mini_Edge`) are the residual gap, tracked in **#4032** — the #4032 research dossier found **no cleanly redistributable vendor STEP** for either, and KiCad's own mini-PCIe 3D model is a dangling upstream 404. LCSC/EasyEDA is the untried source that could resolve them (see companion **Issue #4073** which applies this tier to close #4032).

## Problem

There is no way to get a 3D body for a footprint whose only identity is an **LCSC C-number** (which the fleet BOMs already carry — see below), even though EasyEDA hosts a STEP for nearly every LCSC part. The established community tool is **`easyeda2kicad.py`** (pip package `easyeda2kicad`, `uPesy/easyeda2kicad.py`): `easyeda2kicad --lcsc_id C##### --3d` fetches STEP models via the EasyEDA API.

**License posture:** EasyEDA/LCSC models are design-use-oriented, **not explicitly redistributable**. So the design must be **fetch-on-demand with a local cache**, and models must **NOT be committed to the repo**. Committed `.kicad_pcb` files may only carry a **portable path-variable `(model ...)` ref** into that cache, resolved at render time — mirroring the `${KICADn_3DMODEL_DIR}` precedent (see below).

## Verified facts

- **BOMs already carry LCSC C-numbers.** `boards/03-usb-joystick/output/manufacturing/bom_jlcpcb.csv` has a `LCSC Part #` column; e.g. `Joystick,J2,Module:Joystick_Analog,C50950` — the joystick already has a real LCSC identity `C50950`. Board 06's `MiniPCIe,J3,Connector_PCIE:PCIE_Mini_Edge,generic` is `generic` (no C-number), so it needs a mapping-file-supplied C-number, not a BOM lookup.
- **Env-var precedent for a portable model dir already exists.** `src/kicad_tools/cli/runner.py` (`_render_env`, `find_kicad_3dmodel_dir`, `KICAD_3DMODEL_ENV_VARS`) `setdefault`s `KICADn_3DMODEL_DIR` into the render subprocess env so committed `(model "${KICADn_3DMODEL_DIR}/...")` refs stay machine-portable (from PR #4012). **KiCad resolves `${VARNAME}` in a `(model ...)` path from the process environment** — that is exactly the mechanism to reuse for an LCSC cache dir variable.
- **PR #4045** offset machinery: any resolver returning `ResolvedModels(models, source_anchor)` gets the pad-centroid offset in `add_model_refs_to_text`. **However:** an EasyEDA STEP is a bare `.step` with no `.kicad_mod`, so it has **no source footprint / no `source_anchor`**. Decide the offset story explicitly (see Design open question — **now resolved below**).

### (a) `easyeda2kicad` dependency vs. minimal in-repo fetch — **Recommendation: minimal in-repo client, ~50-80 lines, no new dependency**

Inspected `uPesy/easyeda2kicad.py` source directly (`easyeda2kicad/easyeda/easyeda_api.py`, `easyeda2kicad/easyeda/easyeda_importer.py`, `easyeda2kicad/__main__.py` on the `master` branch):

- `__main__.py` is **argparse/CLI-oriented** — there is no public high-level `fetch_3d_model(lcsc_id) -> bytes` function exported from the package's `__init__.py`. The reusable pieces are internal classes (`EasyedaApi`, `Easyeda3dModelImporter`, `Exporter3dModelKicad`) that are importable but not designed as a stable public API (no `__all__`, no versioned API contract documented).
- The class we actually need, `EasyedaApi` (in `easyeda2kicad/easyeda/easyeda_api.py`), is **stdlib-only** (`urllib.request`, `json`, `gzip`, `ssl`) plus an *optional* `certifi` import for cert handling. It is not a heavy transitive dependency by itself — but importing it means importing the whole `easyeda2kicad` package, which pulls in `easyeda_importer.py`'s KiCad footprint/symbol export machinery (`ExporterFootprintKicad`, `ExporterSymbolKicad`, SVG parsing) that this issue does not need at all — we only want the 3D STEP body.
- **The actual EasyEDA API surface needed is exactly two calls**, both plain HTTP GET with a browser `User-Agent` header (confirmed directly in `easyeda_api.py`):
  1. `GET https://easyeda.com/api/products/{lcsc_id}/components` → JSON; the 3D-model UUID is embedded in `result.packageDetail.dataStr.shape`, as an `SVGNODE~{json}` line whose parsed JSON has a `uuid` field (see `Easyeda3dModelImporter.get_3d_model_info` / `parse_3d_model_info` in `easyeda_importer.py:625-685`).
  2. `GET https://modules.easyeda.com/qAxj6KHrDKw4blvCG8QJPs7Y/{uuid}` → raw STEP bytes (`ENDPOINT_3D_MODEL_STEP` in `easyeda_api.py`; there's also a non-STEP `.obj` endpoint at `https://modules.easyeda.com/3dmodel/{uuid}` — not needed here, STEP is the deliverable format for KiCad `(model ...)` refs).
- **Ruling: implement a minimal in-repo client** (`_fetch_lcsc_step(lcsc_id) -> bytes | None` in the new `lcsc_models.py`) using only `urllib.request` + `json`, mirroring `EasyedaApi.get_info_from_easyeda_api` + `get_step_3d_model`. This avoids the dependency entirely — no `easyeda2kicad` optional extra needed. Rationale: the two-call flow is small and stable (URL patterns above), the transitive footprint/symbol-export machinery in `easyeda2kicad` is 100% unused weight, and a from-scratch client keeps this repo's fetch logic testable/mockable with plain `unittest.mock.patch("urllib.request.urlopen", ...)` rather than depending on an upstream package's internal class shapes staying stable. **Do not add `easyeda2kicad` to `pyproject.toml`.**
- Caveat for the builder: the two endpoint URLs above are read directly from `easyeda2kicad`'s current `master` source (not independently re-verified against EasyEDA's own docs, since EasyEDA does not publish a public API spec) — if the fetch fails against the real API during implementation, re-check `easyeda_api.py` on `master` for endpoint drift before debugging the in-repo client logic.

### (b) `source_anchor` for a bare STEP with no `.kicad_mod` — **Recommendation: origin-authored anchor, composes cleanly with the existing offset math**

Read `models3d.py`'s full offset flow (`_pad_anchor`, `_apply_offset_delta`, `ResolvedModels`, `add_model_refs_to_text` lines 527-584):

- `ResolvedModels.source_anchor` is a `tuple[float, float] | None` — the *library* footprint's own pad centroid, in the library's local 2D frame.
- At insertion time (`add_model_refs_to_text`): `dx, dy = target_anchor[0] - source_anchor[0], target_anchor[1] - source_anchor[1]` where `target_anchor = _pad_anchor(body)` (the target `.kicad_pcb` footprint's own pad centroid). When `source_anchor is None`, `dx = dy = 0.0` (see line 561-564: the `if source_anchor is not None and target_anchor is not None:` guard) — **this is already the origin-authored behavior for free**, no new code path is required in `add_model_refs_to_text` itself.
- `_apply_offset_delta(model_block, dx, dy)` short-circuits to a verbatim (unchanged) block when `abs(dx) < 1e-6 and abs(dy) < 1e-6` (the 1nm-class tolerance mentioned in the issue — actually a relative float tolerance, not an absolute nm unit, but functionally sub-nanometer at PCB scale). **This composes correctly**: for the LCSC tier, setting `source_anchor=(0.0, 0.0)` explicitly (not `None`) means `dx, dy = target_anchor` — i.e. the *entire* target pad-centroid position becomes the offset delta, landing the origin-authored body's origin at the target footprint's pad centroid. This is exactly the "treat the body as origin-authored → delta = target pad centroid" policy the architect proposed, and it is **not** the same as passing `source_anchor=None` (which would compute `dx=dy=0.0`, i.e. no shift at all — wrong for this tier, since an origin-authored EasyEDA STEP is centered at its own local `(0,0)`, not pre-shifted to any pad field).
- **Y-flip and unit composition, verified:** `_apply_offset_delta` writes `ny = _fmt_num(oy - dy)` (model-frame Y negated vs. footprint 2D Y) into the model's existing `(offset (xyz ...))` numbers, i.e. it *adds* the delta to whatever offset the model block already carries (`ox + dx`, `oy - dy`). An EasyEDA-fetched `(model ...)` block that this tier synthesizes should therefore be authored with `(offset (xyz 0 0 0))` as its baseline (no library-native offset to preserve, unlike a real `.kicad_mod`), so the applied delta *is* the final offset. Units: `.kicad_pcb` pad `(at x y)` coordinates are millimeters, matching the `(offset (xyz ...))` node's own mm units — no conversion needed, confirmed by reading `_pad_anchor`'s direct use of `_find_first_at`'s parsed floats with no scale factor.
- **Ruling: the LCSC tier's resolver function should return `ResolvedModels(models=[synthesized_model_block], source_anchor=(0.0, 0.0))`** where `synthesized_model_block` is a freshly-built `(model "${KCT_LCSC_3D_DIR}/{lcsc_id}.step" (offset (xyz 0 0 0)) (scale (xyz 1 1 1)) (rotate (xyz 0 0 0)))` string (mirroring the shape KiCad itself writes; `_apply_offset_delta`/`_indent_block` handle re-indentation and delta injection automatically since they operate on generic model-block text, not just `.kicad_mod`-sourced text). No changes to `_pad_anchor`, `_apply_offset_delta`, or `add_model_refs_to_text` are required — the existing machinery already generalizes to a synthetic `ResolvedModels` with an explicit `(0, 0)` anchor. Document the approximation (origin-authored placement, not pin-1-registered) in the new tier's docstring and in `lcsc_models.py`.
- Scale/rotation are out of scope for this issue (default `1 1 1` / `0 0 0`) — EasyEDA STEP bodies are not guaranteed to be authored at 1:1 scale or canonical rotation vs. the footprint's silkscreen orientation; note this as a known limitation for #4073's per-board sidecar entries (a future per-mapping-entry scale/rotate override could be added if a specific fetched body needs it, but is not required for this issue's acceptance criteria).

### (c) Sidecar keying — **Recommendation: lib_id-keyed only for this issue; do not add ref-keying**

Read the patching loop in `add_model_refs_to_text` (`models3d.py:527-584`) and `FootprintBlock` (`models3d.py:131-141`):

- `FootprintBlock` currently carries only `lib_id`, `start`, `end`, `indent` — **no reference designator**. `iter_footprint_blocks` never extracts `(fp_text reference "..." ...)`, and `add_model_refs_to_text`'s resolver signature is `Callable[[str], ResolvedModels | list[str] | None]` — a pure `lib_id -> models` function with no access to the footprint's reference.
- Verified empirically (`boards/03-usb-joystick/output/usb_joystick.kicad_pcb:155`, `boards/06-diffpair-test/output/diffpair_test.kicad_pcb`) that every footprint block **does** contain a `(fp_text reference "J2" (at ...) (layer "F.SilkS") ...)` node inside its body, using the same string-aware scanning primitives (`_find_matching_paren`, `_skip_string`) already used for `(pad ...)` extraction in `_pad_anchor`. Extracting it is a small, low-risk addition (one more helper function analogous to `_pad_anchor`, plus one more field on `FootprintBlock`) — **but it is not needed for this issue's scope.**
- **Ruling: lib_id-keyed sidecar suffices for the fleet.** Per #4073, `Module:Joystick_Analog` (board 03) and `Connector_PCIE:PCIE_Mini_Edge` (board 06) are each a **unique lib_id within their board** — no board in the fleet places two different LCSC parts under the same synthetic lib_id. A `lcsc_models.json` sidecar shaped `{"Module:Joystick_Analog": "C50950", "Connector_PCIE:PCIE_Mini_Edge": "C..."}` (lib_id -> C-number) is sufficient and requires **zero changes** to `FootprintBlock`, `iter_footprint_blocks`, or the `Resolver` callable signature — the new tier is just another `lib_id -> ResolvedModels` function, exactly like tiers 1-3.
- **Scope guard:** do NOT implement reference-designator extraction/keying in this issue. If a future board needs two distinct LCSC parts sharing one synthetic lib_id (not true of any current fleet board), that would require: (1) a `reference` field on `FootprintBlock` populated from the `(fp_text reference ...)` node, (2) widening the `Resolver` callable to accept an optional reference argument or a `(lib_id, reference)` tuple key, and (3) a resolver-selection change in `add_model_refs_to_text`'s loop. File that as a separate follow-up if/when it's actually needed — do not build it speculatively here.

### (d) Path portability (`${KCT_LCSC_3D_DIR}`) — **Confirmed safe; verified empirically, not just by code reading**

- `_render_env` (`runner.py:1669-1681`) copies the current environment and `setdefault`s every name in `KICAD_3DMODEL_ENV_VARS` to the resolved KiCad install's `3dmodels` dir, **only when that variable isn't already set** — so adding `"KCT_LCSC_3D_DIR"` requires (1) resolving the LCSC cache dir path (default `~/.cache/kicad-tools/lcsc-3d/`, override via the `KCT_LCSC_3D_DIR` env var itself) and (2) `setdefault`-ing it the same way, in the same function, right after the `KICAD_3DMODEL_ENV_VARS` loop. This is a same-shape addition, not a new mechanism.
- **Verified DRC/render tolerate an unresolvable `${KCT_LCSC_3D_DIR}` path — empirically, not just by inspection.** Built a scratch copy of `boards/03-usb-joystick/output/usb_joystick.kicad_pcb`, hand-inserted `(model "${KCT_LCSC_3D_DIR}/C50950.step" (offset (xyz 0 0 0)) (scale (xyz 1 1 1)) (rotate (xyz 0 0 0)))` into the `J2` footprint block (confirming `KCT_LCSC_3D_DIR` was unset in the shell and the referenced file does not exist anywhere), then ran both:
  - `kicad-cli pcb drc --output drc_out.json --format json --units mm <file>` → **exit 0**, same violation/unconnected-item class as the board's pre-existing (unrelated) routing state — no model-related errors, no DRC rule inspects `(model ...)` nodes at all (confirmed no `validate/` rule references `"model"`).
  - `kicad-cli pcb render --output out.png --side top --quality basic <file>` → **exit 0**, stdout literally says `Successfully created 3D render image` — kicad-cli silently skips the unresolvable body and renders the rest of the board.
  - This matches the existing `${KICADn_3DMODEL_DIR}` precedent's implicit assumption (PR #4012 relies on the same graceful-degradation behavior when a model dir isn't found) and confirms the issue's "models are cosmetic" premise is correct, not just asserted.
- **Ruling: no additional guard logic needed** in the LCSC tier for the render/DRC path — an unresolved-but-referenced C-number is safe to leave as a dangling `${KCT_LCSC_3D_DIR}/{id}.step` ref if the cache miss + fetch-disabled path is hit (per the issue's own "graceful failure" requirement), *or* the tier can simply skip emitting the `(model ...)` ref entirely on a cache miss (cleaner — avoids ever writing a ref to a file that provably doesn't exist yet). **Recommend the latter** (skip-on-miss, report as unresolved) over emitting-a-dangling-ref-anyway, since it keeps `report.unresolved` / `report.patched` semantics consistent with the existing three tiers (a tier either resolves and inserts, or it doesn't) and avoids ever shipping a `.kicad_pcb` with a ref to a file that is known-absent at patch time.

## Proposed solution

Add a **fourth resolver tier** to `make_library_resolver` (after exact / variant / substitution), keyed by an **LCSC C-number mapping**, that:

1. **Resolves a C-number** for a footprint from a **committed per-board sidecar** `lcsc_models.json` mapping **lib_id → C-number** (see Design Question (c) — lib_id-keyed, ref-keyed explicitly out of scope for this issue).
2. **Fetches the STEP on demand** via a **minimal in-repo EasyEDA client** (see Design Question (a) — no `easyeda2kicad` dependency) into an **env-var-configurable cache** (`KCT_LCSC_3D_DIR`, default `~/.cache/kicad-tools/lcsc-3d/`), one file per C-number.
3. **Emits a portable `(model ...)` ref** pointing at the cache via a path variable — introduce `${KCT_LCSC_3D_DIR}` and `setdefault` it into the render subprocess env in `_render_env` (same treatment as `KICADn_3DMODEL_DIR`), so committed refs stay machine-portable and CI/other machines resolve them from their own cache. **Only emit the ref when the STEP is actually present in the cache** (see Design Question (d) — skip-on-miss, not dangling-ref-on-miss).

### Offset handling — resolved (see Design Question (b))

The LCSC tier's resolver returns `ResolvedModels(models=[synthesized_block], source_anchor=(0.0, 0.0))` — an explicit origin anchor (not `None`), so the existing `add_model_refs_to_text` delta math (`target_anchor - source_anchor`) computes the full target pad-centroid as the offset with zero changes to the shared offset-application code. The synthesized model block should start from `(offset (xyz 0 0 0))` so the applied delta becomes the model's entire final offset.

### Offline / CI safety (hard requirement)

- **CI must not require network.** The tier activates **only** when the cache already has the file, **or** fetching is **explicitly enabled** (e.g. `--fetch-lcsc` flag / `KCT_LCSC_FETCH=1`). Default: cache-only, no network.
- **Fetch failures degrade gracefully:** render without the model, warn (report as unresolved for that footprint) — never fail the patch or the render. Confirmed safe by the DRC/render empirical test above.

## Implementation approach

1. `src/kicad_tools/pcb/lcsc_models.py` (new): C-number → cached STEP path resolver + minimal EasyEDA fetch client (2 stdlib-only HTTP calls per Design Question (a): `GET .../api/products/{lcsc_id}/components` → extract 3D-model `uuid` from `result.packageDetail.dataStr.shape`'s `SVGNODE` entry → `GET https://modules.easyeda.com/qAxj6KHrDKw4blvCG8QJPs7Y/{uuid}` → raw STEP bytes), cache dir resolution honoring `KCT_LCSC_3D_DIR`, graceful-failure semantics (never raises on network/parse failure — returns `None`).
2. `models3d.py`: fourth tier in `make_library_resolver` (guarded by an LCSC `lib_id -> C-number` mapping arg), returning `ResolvedModels(models=[...], source_anchor=(0.0, 0.0))` per Design Question (b) — synthesizing the `(model ...)` block text itself (no `.kicad_mod` to extract from) and skipping (not dangling-ref) on cache miss + fetch-disabled.
3. `runner.py`: register `KCT_LCSC_3D_DIR` in `_render_env`, mirroring the `KICAD_3DMODEL_ENV_VARS` `setdefault` pattern (default cache dir `~/.cache/kicad-tools/lcsc-3d/` when the env var itself isn't already set by the caller).
4. CLI: `kct pcb add-3d-models` gains a sidecar/mapping arg (e.g. `--lcsc-models lcsc_models.json`) and an explicit `--fetch-lcsc` opt-in; document cache dir + env var in `README.md` / `--help`.

## Acceptance criteria

- [ ] A fourth LCSC tier resolves a lib_id→C-number mapping (from a committed `lcsc_models.json` sidecar) and emits a portable `${KCT_LCSC_3D_DIR}` model ref, synthesized with `source_anchor=(0.0, 0.0)` (origin-authored policy).
- [ ] Fetch is opt-in (`--fetch-lcsc` / `KCT_LCSC_FETCH=1`); cache-only by default; on cache miss with fetch disabled, the tier **skips** emitting a ref (reports unresolved) rather than emitting a dangling reference; fetch failures degrade to a warning with no render/patch failure. **CI needs no network.**
- [ ] `_render_env` `setdefault`s `KCT_LCSC_3D_DIR` (default `~/.cache/kicad-tools/lcsc-3d/`) so committed refs are machine-portable (mirrors PR #4012's `KICADn_3DMODEL_DIR` pattern).
- [ ] #4045 pad-centroid offset applies to the new tier via the existing `ResolvedModels`/`add_model_refs_to_text` machinery — **no changes required** to `_pad_anchor`, `_apply_offset_delta`, or the patching loop itself (verified during curation; see Design Question (b)).
- [ ] Mocked-fetch + cache + path-variable + offset-parity tests pass (see Test Plan); pure-insertion guarantee preserved (copper/pad/zone/net/`at` bytes unchanged).
- [ ] No `easyeda2kicad` dependency added to `pyproject.toml` — minimal in-repo client only (decision recorded in Design Question (a) above; rationale: avoids the package's unused footprint/symbol-export transitive weight and keeps the fetch mockable against plain `urllib`).
- [ ] Sidecar schema is lib_id-keyed (`{"Library:Name": "C#####", ...}`) — no reference-designator keying added in this issue (see Design Question (c) scope guard).

## Test Plan

- [ ] **Mocked-API fetch** (no network in CI): `unittest.mock.patch("urllib.request.urlopen", ...)` (or equivalent) returning fixture JSON (component-info response with an `SVGNODE` line carrying a `uuid`) and fixture STEP bytes; assert the tier resolves and writes the fetched bytes into a temp `KCT_LCSC_3D_DIR` cache path.
- [ ] **Cache behavior:** hit (file already in cache, no HTTP call made — assert mock not called) vs miss (`--fetch-lcsc` disabled → unresolved + warning, no network attempted; `--fetch-lcsc` enabled → fetch called once, file written to cache, second call is a cache hit).
- [ ] **Path-variable emission:** committed `(model ...)` ref text contains `${KCT_LCSC_3D_DIR}/{lcsc_id}.step`, not an absolute path — assert via string containment on the patched `.kicad_pcb` text.
- [ ] **Offset math unit test:** call the LCSC tier's resolver function directly with a fixture footprint block whose pad centroid is at a known non-origin position (e.g. `(2.0, -1.27)`); assert the returned `ResolvedModels.source_anchor == (0.0, 0.0)` and that after `add_model_refs_to_text` runs, the inserted model's `(offset (xyz ...))` equals `(2.0, 1.27, 0)` (Y negated per the existing convention) — extend `tests/test_pcb_add_3d_models.py`.
- [ ] **Offset parity with existing tiers:** the LCSC tier's inserted body applies the same pad-centroid delta convention as an equivalent exact/substitution-tier match on the same footprint shape (extend `tests/test_pcb_add_3d_models.py`).
- [ ] **Pure-insertion guard:** copper/pad/zone/net/`at` bytes unchanged before/after patching (as PR #4045 asserts) — extend the existing byte-diff assertion pattern in `tests/test_pcb_add_3d_models.py`.
- [ ] **Skip-on-miss (not dangling ref):** cache miss + fetch disabled → assert no `(model ...)` node is inserted for that footprint and the lib_id appears in `report.unresolved`.
- [ ] `_render_env` includes `KCT_LCSC_3D_DIR` with the correct default — extend `tests/test_render_cmd.py`.
- [ ] **Offline no-op:** with `KCT_LCSC_FETCH` unset/false and an empty cache dir, running `kct pcb add-3d-models --lcsc-models <sidecar>` against a board makes zero network calls (assert via a global `urlopen` mock raising `AssertionError` if called) and exits cleanly with the LCSC lib_ids reported unresolved.

## Affected Files

- `src/kicad_tools/pcb/lcsc_models.py` (new) — C-number → cached STEP resolver, minimal EasyEDA HTTP client, cache dir resolution.
- `src/kicad_tools/pcb/models3d.py` — fourth tier registration in `make_library_resolver`; no changes needed to `_pad_anchor`/`_apply_offset_delta`/`add_model_refs_to_text` (verified during curation).
- `src/kicad_tools/cli/runner.py` — `KCT_LCSC_3D_DIR` registration in `_render_env` (near `KICAD_3DMODEL_ENV_VARS` / `find_kicad_3dmodel_dir`, lines ~1614-1681).
- `src/kicad_tools/cli/pcb_add_3d_models.py` — `--lcsc-models` sidecar arg, `--fetch-lcsc` opt-in flag, wiring into `add_model_refs`.
- `src/kicad_tools/cli/commands/pcb.py` — `_run_add_3d_models_command` argument plumbing for the new CLI flags.
- `src/kicad_tools/cli/parser.py` — `add-3d-models` subparser argument definitions (`--lcsc-models`, `--fetch-lcsc`).
- `tests/test_pcb_add_3d_models.py` (extend) — mocked-fetch, cache hit/miss, offset-parity, pure-insertion, skip-on-miss tests.
- `tests/test_render_cmd.py` (extend) — `_render_env` includes `KCT_LCSC_3D_DIR`.
- `README.md` — document `KCT_LCSC_3D_DIR`, `--fetch-lcsc`, `KCT_LCSC_FETCH`, and the `lcsc_models.json` sidecar schema.
- New: `tests/test_lcsc_models.py` (new, optional split from `test_pcb_add_3d_models.py`) — unit tests for the fetch client and cache-path resolution in isolation.

*Enumerated call sites/files above are a starting point — the builder must verify completeness (LLM "find all X" is reliably under-inclusive). All paths above verified present on `origin/main` during curation (2026-07-12).*

## References

- Companion / dependent: **Issue #4073** (applies this tier to the fleet, closes #4032). Sweeps run these in order; #4073 is NOT marked `loom:blocked` but depends on this one.
- #4032 (waived parts + license-rationale dossier — read the comment), #4014 / #4033 (substitution tier this extends), #4012 (`${KICADn_3DMODEL_DIR}` render-env precedent), **PR #4045** (pad-centroid offset machinery).
- Code: `src/kicad_tools/pcb/models3d.py` (`make_library_resolver`, `ResolvedModels`, `add_model_refs_to_text`), `src/kicad_tools/pcb/model_substitutions.py`, `src/kicad_tools/cli/runner.py` (`_render_env`, `find_kicad_3dmodel_dir`, `KICAD_3DMODEL_ENV_VARS`).
- Upstream reference (read during curation, not vendored): `uPesy/easyeda2kicad.py` — `easyeda2kicad/easyeda/easyeda_api.py` (`EasyedaApi.get_info_from_easyeda_api`, `get_step_3d_model`, endpoint constants `API_ENDPOINT`, `ENDPOINT_3D_MODEL_STEP`), `easyeda2kicad/easyeda/easyeda_importer.py` (`Easyeda3dModelImporter`, lines 625-685, shows the `uuid` extraction path from `SVGNODE` data). No files fetched or downloaded from EasyEDA/LCSC during curation — only the `easyeda2kicad` GitHub source (public code) was read, plus an empirical DRC/render safety test against a local, hand-edited scratch copy of an existing repo board file (no external network calls involved in that test).

---

## Curator Notes (2026-07-12)

Curated per explicit user request (issue number specified). This issue carried `loom:architect` (awaiting human `loom:issue` approval) — per the Curator role's "Exception: Explicit User Instructions," proceeding with curation is treated as implicit approval to enhance; **`loom:issue` was NOT added** (only a human can add that label). All four "verify" design questions the architect flagged were resolved by directly reading `models3d.py`'s offset machinery, `runner.py`'s render-env code, the `easyeda2kicad` upstream source (GitHub, no downloads), and by an empirical DRC + render test against a hand-edited scratch copy of `boards/03-usb-joystick/output/usb_joystick.kicad_pcb` (deleted after the test; not committed). See the "Curator research: design questions resolved" section above for full rulings on (a) dependency-vs-in-repo-client, (b) offset/`source_anchor` policy, (c) sidecar keying, and (d) path portability / DRC-render safety.

rjwalters:
## Summary

Adds a **fourth tier** to `kct pcb add-3d-models` that resolves a 3D STEP body for footprints whose only usable identity is an **LCSC C-number** (the JLCPCB-assembly fleet already carries these in its BOMs), closing the residual gap that no installed KiCad library can reach.

Implements all four curator rulings on issue #4072.

### Client design (ruling a — no new dependency)
`src/kicad_tools/pcb/lcsc_models.py` (new): a ~50-line, **stdlib-only** (`urllib` + `json`) in-repo EasyEDA client. Two HTTP GETs:
1. `GET https://easyeda.com/api/products/{lcsc_id}/components` → parse the 3D-model `uuid` from the `SVGNODE` entry in `result.packageDetail.dataStr.shape`.
2. `GET https://modules.easyeda.com/qAxj6KHrDKw4blvCG8QJPs7Y/{uuid}` → raw STEP bytes.

No `easyeda2kicad` dependency added. Fetch/parse failures never raise — they degrade to `None`.

### Tier integration (ruling b — origin-authored anchor)
`models3d.py`: fourth tier in `make_library_resolver`, gated by a committed `lib_id -> C-number` sidecar. It returns `ResolvedModels(models=[synthesized_block], source_anchor=(0.0, 0.0))` — an **explicit** origin (not `None`) — so the #4045 offset machinery computes `dx, dy = target_pad_centroid` and lands the origin-authored body on the target footprint's pads. **Zero changes** to `_pad_anchor` / `_apply_offset_delta` / `add_model_refs_to_text`. Skip-on-miss (reports `unresolved`) rather than emitting a dangling ref. Added `lcsc_matches` to `ModelPatchReport`.

### Cache / env-var behavior (rulings c + d)
- Sidecar is **lib_id-keyed** only (`{"Module:Joystick_Analog": "C50950"}`) — no reference-designator plumbing.
- Cache dir `KCT_LCSC_3D_DIR` (default `~/.cache/kicad-tools/lcsc-3d/`), one `.step` per C-number.
- Emitted refs use the portable `${KCT_LCSC_3D_DIR}` path variable; `runner.py::_render_env` now `setdefault`s it into the render subprocess env (mirrors the `KICADn_3DMODEL_DIR` pattern from #4012), so committed refs stay machine-portable.
- CLI: `--lcsc-models <sidecar>` + `--fetch-lcsc` opt-in (also `KCT_LCSC_FETCH=1`). **Cache-only by default; CI needs no network.**

### Real-fetch sanity result (C50950, run manually, not in the test suite)
- Component-info GET → 11669 bytes → parsed uuid `7b135d4c7d084b658994bacec4f3b635`.
- STEP GET → 1896563 bytes, header `ISO-10303-21; ... FILE_DESCRIPTION (( 'STEP AP214' )...` — a valid STEP file.
- Full CLI path (`--lcsc-models ... --fetch-lcsc`) emitted `(model "${KCT_LCSC_3D_DIR}/C50950.step" ...)` with offset `(xyz 2 1.27 0)` (Y-negated, matching the `(2.0, -1.27)` pad centroid) and cached the STEP. The offline run (empty cache, fetch disabled) reported `unresolved` and made **zero** network calls.
- **No downloaded STEP is committed** (design/license posture: not redistributable).

### Test results
- New `tests/test_lcsc_models.py`: 39 tests — mocked-API fetch (both endpoints, uuid parse, error paths), cache hit/miss, offline no-op (guarded so any network call fails the test), `${KCT_LCSC_3D_DIR}` emission, origin-anchor offset math, skip-on-miss, `_render_env` LCSC-var injection.
- `tests/test_pcb_add_3d_models.py` + `tests/test_render_cmd.py` + new file: **113 passed** (updated one pre-existing render-env test for the new LCSC-var injection behavior).
- `ruff check` + `ruff format --check`: clean on touched files.
- mypy: touched files clean; the 7 `runner.py` errors are pre-existing on `origin/main` (verified by stash) — no new mypy errors introduced.

### Follow-up
Sibling **#4073** (fleet application, closes #4032) builds on this tier and should run **after** this merges.

Closes #4072

### DIFF
--- a/docs/guides/lcsc-3d-models.md
+++ b/docs/guides/lcsc-3d-models.md
@@ -1,0 +1,82 @@
+# LCSC/EasyEDA fetch-on-demand 3D models
+
+`kct pcb add-3d-models` resolves `(model ...)` 3D bodies for footprints in four
+tiers (first hit wins):
+
+1. **Exact** installed KiCad `library:name` footprint.
+2. **Same-library variant** — a suffixed name in the same library.
+3. **Cross-library substitution** — a curated `lib_id -> lib_id` table for
+   generic/synthetic lib ids with a body-compatible equivalent elsewhere.
+4. **LCSC/EasyEDA fetch-on-demand** — for footprints whose only usable identity
+   is an **LCSC C-number** (the JLCPCB-assembly fleet already carries these in
+   its BOMs). Nearly every LCSC part has a STEP body in the EasyEDA parts
+   database; this tier fetches it on demand into a local cache and emits a
+   portable model ref.
+
+## Using the LCSC tier
+
+Provide a per-board **sidecar** mapping `lib_id -> C-number`:
+
+```json
+{
+  "Module:Joystick_Analog": "C50950",
+  "Connector_PCIE:PCIE_Mini_Edge": "C123456"
+}
+```
+
+Then run:
+
+```bash
+# Cache-only (no network): resolves only C-numbers already cached.
+kct pcb add-3d-models board.kicad_pcb --lcsc-models lcsc_models.json
+
+# Opt in to fetching missing STEP bodies from EasyEDA on a cache miss.
+kct pcb add-3d-models board.kicad_pcb --lcsc-models lcsc_models.json --fetch-lcsc
+```
+
+The sidecar is **lib_id-keyed** (not reference-designator-keyed): each synthetic
+lib id in the fleet is unique per board, so one C-number per lib id suffices.
+
+## Cache and environment variables
+
+- **`KCT_LCSC_3D_DIR`** — the on-disk STEP cache directory. Default:
+  `~/.cache/kicad-tools/lcsc-3d/`, one file per C-number (`C50950.step`). It
+  doubles as the **path variable** in emitted refs
+  (`(model "${KCT_LCSC_3D_DIR}/C50950.step" ...)`), so committed `.kicad_pcb`
+  files stay machine-portable — each machine resolves the ref from its own
+  cache at render time. `kct pcb render` `setdefault`s this variable into the
+  `kicad-cli` subprocess environment (mirroring the `${KICADn_3DMODEL_DIR}`
+  precedent).
+- **`KCT_LCSC_FETCH`** — set to `1`/`true`/`yes`/`on` to enable fetch-on-demand
+  without the `--fetch-lcsc` flag. Default: cache-only.
+
+## Safety and behavior notes
+
+- **Offline / CI safe.** The tier activates only on a cache hit, or when
+  fetching is explicitly enabled. CI never needs network.
+- **Skip-on-miss.** On a cache miss with fetching disabled (or a fetch
+  failure), the footprint is reported *unresolved* and **no** `(model ...)` ref
+  is emitted — the tool never ships a `.kicad_pcb` with a ref to a
+  known-absent file. Fetch failures degrade to a warning; the patch and render
+  never fail for want of a body.
+- **Pure metadata insertion.** As with the other tiers, only `(model ...)`
+  lines are inserted — copper, placement, zones, and nets are untouched, so DRC
+  results are identical before and after.
+- **Origin-authored placement (approximation).** An EasyEDA STEP is a bare
+  `.step` with no `.kicad_mod`, so there is no source pad centroid. The body is
+  treated as origin-centered (`source_anchor=(0, 0)`), and the shared offset
+  machinery lands its origin on the target footprint's pad centroid. Scale and
+  rotation are left at KiCad defaults; a fetched body whose native orientation
+  differs from the footprint silkscreen may sit rotated. These bodies are
+  cosmetic render aids.
+- **Licensing.** EasyEDA/LCSC models are design-use-oriented and **not**
+  explicitly redistributable — they are cached locally and **never committed**
+  to the repo. Only the portable `${KCT_LCSC_3D_DIR}` path-variable ref is
+  committed.
+
+## No new dependency
+
+The EasyEDA client is a ~50-line in-repo, stdlib-only (`urllib` + `json`)
+two-call fetch (`.../api/products/{lcsc_id}/components` → extract the 3D-model
+uuid → `https://modules.easyeda.com/.../{uuid}` → raw STEP bytes). No
+`easyeda2kicad` dependency is added.
--- a/src/kicad_tools/cli/commands/pcb.py
+++ b/src/kicad_tools/cli/commands/pcb.py
@@ -668,11 +668,14 @@ def _run_add_3d_models_command(args, pcb_path: Path) -> int:
 
     output = getattr(args, "output", None)
     lib_path = getattr(args, "lib_path", None)
+    lcsc_models = getattr(args, "lcsc_models", None)
     return run_add_3d_models(
         pcb_path=pcb_path,
         output_path=Path(output) if output else None,
         lib_path=Path(lib_path) if lib_path else None,
         allow_variants=not getattr(args, "exact", False),
+        lcsc_models=Path(lcsc_models) if lcsc_models else None,
+        fetch_lcsc=getattr(args, "fetch_lcsc", False),
         dry_run=getattr(args, "dry_run", False),
         output_format=getattr(args, "format", "text"),
     )
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -1672,6 +1672,21 @@ def _add_pcb_parser(subparsers) -> None:
         "name-variant fallback used for visual model lookup)",
     )
     pcb_add_models.add_argument(
+        "--lcsc-models",
+        dest="lcsc_models",
+        help="Path to an lcsc_models.json sidecar mapping lib_id -> LCSC "
+        "C-number, enabling the LCSC/EasyEDA fetch-on-demand tier. Bodies are "
+        "resolved from a cache dir (${KCT_LCSC_3D_DIR}, default "
+        "~/.cache/kicad-tools/lcsc-3d/) and emitted as portable "
+        "${KCT_LCSC_3D_DIR}/C#####.step refs",
+    )
+    pcb_add_models.add_argument(
+        "--fetch-lcsc",
+        action="store_true",
+        help="Fetch missing LCSC STEP bodies from EasyEDA on a cache miss "
+        "(default: cache-only, no network; also enabled by KCT_LCSC_FETCH=1)",
+    )
+    pcb_add_models.add_argument(
         "--format",
         choices=["text", "json"],
         default="text",
--- a/src/kicad_tools/cli/pcb_add_3d_models.py
+++ b/src/kicad_tools/cli/pcb_add_3d_models.py
@@ -31,6 +31,8 @@ def run_add_3d_models(
     output_path: Path | None = None,
     lib_path: Path | None = None,
     allow_variants: bool = True,
+    lcsc_models: Path | None = None,
+    fetch_lcsc: bool = False,
     dry_run: bool = False,
     output_format: str = "text",
 ) -> int:
@@ -44,6 +46,12 @@ def run_add_3d_models(
         allow_variants: Accept same-library footprint-name variants when the
             exact name is missing (e.g. ``QFN-24-1EP_4x4mm_P0.5mm`` matches
             the ``..._EP2.6x2.6mm`` variant — the model body is identical).
+        lcsc_models: Optional path to an ``lcsc_models.json`` sidecar mapping
+            ``lib_id -> C-number``, enabling the LCSC/EasyEDA fetch-on-demand
+            tier.
+        fetch_lcsc: When True, fetch a missing LCSC STEP from EasyEDA on a
+            cache miss (default: cache-only; also enabled by
+            ``KCT_LCSC_FETCH``).
         dry_run: Report what would be inserted without writing.
         output_format: ``"text"`` or ``"json"``.
 
@@ -51,10 +59,23 @@ def run_add_3d_models(
         Exit code (0 success, 1 error).
     """
     from kicad_tools.footprints.library_path import detect_kicad_library_path
+    from kicad_tools.pcb.lcsc_models import fetch_enabled, load_lcsc_mapping
     from kicad_tools.pcb.models3d import add_model_refs
 
+    lcsc_mapping: dict[str, str] | None = None
+    if lcsc_models is not None:
+        try:
+            lcsc_mapping = load_lcsc_mapping(lcsc_models)
+        except ValueError as e:
+            if output_format == "json":
+                print(json.dumps({"error": str(e)}))
+            else:
+                print(f"Error: {e}", file=sys.stderr)
+            return 1
+
     library_paths = detect_kicad_library_path(config_override=lib_path)
-    if not library_paths.found:
+    if not library_paths.found and lcsc_mapping is None:
+        # Only the installed-library tiers can help and they're unavailable.
         msg = (
             "KiCad footprint libraries not found (install KiCad or pass "
             "--lib-path / set KICAD_FOOTPRINT_DIR)"
@@ -65,12 +86,18 @@ def run_add_3d_models(
             print(f"Error: {msg}", file=sys.stderr)
         return 1
 
+    def _warn(message: str) -> None:
+        print(f"Warning: {message}", file=sys.stderr)
+
     try:
         report = add_model_refs(
             pcb_path,
             output_path=output_path,
             library_paths=library_paths,
             allow_variants=allow_variants,
+            lcsc_mapping=lcsc_mapping,
+            lcsc_fetch=fetch_enabled(fetch_lcsc),
+            lcsc_warn=_warn,
             dry_run=dry_run,
         )
     except Exception as e:
@@ -92,6 +119,7 @@ def run_add_3d_models(
         "no_model_in_library": report.no_model_in_library,
         "variant_matches": report.variant_matches,
         "substitution_matches": report.substitution_matches,
+        "lcsc_matches": report.lcsc_matches,
     }
 
     if output_format == "json":
@@ -117,6 +145,10 @@ def run_add_3d_models(
             print("  Cross-library substitution models used (curated equivalent):")
             for lib_id, sub in sorted(report.substitution_matches.items()):
                 print(f"    {lib_id} -> {sub}")
+        if report.lcsc_matches:
+            print("  LCSC/EasyEDA fetched models used (cached STEP):")
+            for lib_id, cnum in sorted(report.lcsc_matches.items()):
+                print(f"    {lib_id} -> {cnum}")
         if report.already_present:
             print(f"  Already had models: {len(report.already_present)}")
         if report.no_model_in_library:
--- a/src/kicad_tools/cli/runner.py
+++ b/src/kicad_tools/cli/runner.py
@@ -1669,15 +1669,24 @@ def find_kicad_3dmodel_dir(kicad_cli: Path | None = None) -> Path | None:
 def _render_env(kicad_cli: Path | None) -> dict[str, str] | None:
     """Build a subprocess environment that resolves 3D model path variables.
 
-    Returns None (inherit as-is) when no model directory can be located.
-    Existing environment variables are never overridden.
+    Sets the ``KICADn_3DMODEL_DIR`` variables (when a KiCad model dir is found)
+    and ``KCT_LCSC_3D_DIR`` (the LCSC fetch-on-demand STEP cache) so committed
+    ``(model "${...}/...")`` refs resolve at render time.  Returns None (inherit
+    as-is) only when neither can be supplied.  Existing environment variables
+    are never overridden.
     """
+    from kicad_tools.pcb.lcsc_models import DEFAULT_CACHE_ENV_VAR, lcsc_cache_dir
+
     model_dir = find_kicad_3dmodel_dir(kicad_cli)
-    if model_dir is None:
+    lcsc_dir = lcsc_cache_dir()
+    if model_dir is None and DEFAULT_CACHE_ENV_VAR in os.environ:
+        # Nothing to inject: KiCad libs absent and the LCSC var is already set.
         return None
     env = os.environ.copy()
-    for var in KICAD_3DMODEL_ENV_VARS:
-        env.setdefault(var, str(model_dir))
+    if model_dir is not None:
+        for var in KICAD_3DMODEL_ENV_VARS:
+            env.setdefault(var, str(model_dir))
+    env.setdefault(DEFAULT_CACHE_ENV_VAR, str(lcsc_dir))
     return env
 
 
--- a/src/kicad_tools/pcb/lcsc_models.py
+++ b/src/kicad_tools/pcb/lcsc_models.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 
 import json
 import os
+import re
 import urllib.error
 import urllib.request
 from pathlib import Path
@@ -71,6 +72,30 @@ _USER_AGENT = (
     "(KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36"
 )
 _HTTP_TIMEOUT = 30
+
+# --------------------------------------------------------------------------
+# Untrusted-input validation
+# --------------------------------------------------------------------------
+# The C-number originates from a committed sidecar (``lcsc_models.json``) that
+# must be treated as untrusted: it flows into a cache *filename*
+# (``{lcsc_id}.step``), into a request *URL* (``.format(lcsc_id=...)``), and
+# into a committed ``.kicad_pcb`` *model ref*.  An unvalidated value enables
+# path traversal (``../outside/pwned``), URL injection (``C1/../../evil?x=``),
+# and ref injection.  EasyEDA C-numbers are ``C`` followed by digits, so we pin
+# the value to that shape at every trust boundary.
+_LCSC_ID_RE = re.compile(r"^C\d+$")
+
+# Bounds on a fetched STEP body written to the cache.  A compromised/rogue
+# endpoint can return an arbitrarily large or non-STEP payload; cap the size
+# and require the ISO-10303-21 header before ever writing to disk.  Real LCSC
+# bodies are typically a few MB; 50 MB is generous but bounded.
+_MAX_STEP_BYTES = 50 * 1024 * 1024
+_STEP_MAGIC = b"ISO-10303-21"
+
+
+def _is_valid_lcsc_id(lcsc_id: str) -> bool:
+    """True when *lcsc_id* is a well-formed EasyEDA C-number (``C`` + digits)."""
+    return bool(_LCSC_ID_RE.match(lcsc_id))
 
 
 def lcsc_cache_dir() -> Path:
@@ -120,6 +145,15 @@ def load_lcsc_mapping(sidecar_path: Path | str) -> dict[str, str]:
         if not isinstance(key, str) or not isinstance(value, str):
             raise ValueError(
                 f"LCSC sidecar {path}: entries must be string lib_id -> string C-number"
+            )
+        # The C-number is untrusted and flows into a cache filename, a request
+        # URL, and a committed board ref; a malformed value (path traversal,
+        # URL injection, non-C string) is a build error, matching the
+        # malformed-sidecar posture above.
+        if not _is_valid_lcsc_id(value):
+            raise ValueError(
+                f"LCSC sidecar {path}: invalid C-number {value!r} for {key!r} "
+                "(expected 'C' followed by digits, e.g. 'C50950')"
             )
         mapping[key] = value
     return mapping
@@ -209,6 +243,12 @@ def _fetch_lcsc_step(lcsc_id: str) -> bytes | None:
     step = _http_get(_API_STEP_MODEL.format(uuid=uuid))
     if not step:
         return None
+    # Bound the third-party payload before it can reach the disk: reject an
+    # oversize body or anything that is not a real ISO-10303-21 STEP file.
+    if len(step) > _MAX_STEP_BYTES:
+        return None
+    if not step.lstrip().startswith(_STEP_MAGIC):
+        return None
     return step
 
 
@@ -238,6 +278,14 @@ def resolve_lcsc_step(
         fetch: When True, fetch on a cache miss; when False, cache-only.
         warn: Optional ``callable(str)`` invoked on a fetch failure.
     """
+    # Defense in depth: even though ``load_lcsc_mapping`` rejects a malformed
+    # C-number at the sidecar boundary, never let an unvalidated value reach the
+    # filesystem or network here.  The resolver path stays never-raise, so an
+    # invalid id warns and returns None rather than raising.
+    if not _is_valid_lcsc_id(lcsc_id):
+        if callable(warn):
+            warn(f"LCSC 3D model skipped: invalid C-number {lcsc_id!r} (no model inserted)")
+        return None
     cache = cache_dir if cache_dir is not None else lcsc_cache_dir()
     step_path = cache / f"{lcsc_id}.step"
     if step_path.is_file():
--- a/src/kicad_tools/pcb/lcsc_models.py
+++ b/src/kicad_tools/pcb/lcsc_models.py
@@ -1,0 +1,281 @@
+"""LCSC/EasyEDA fetch-on-demand 3D model resolver (fourth ``add-3d-models`` tier).
+
+The JLCPCB-assembly fleet identifies its assembly parts by **LCSC C-numbers**,
+and nearly every LCSC part carries a 3D STEP body in the EasyEDA parts
+database.  This module resolves a footprint whose only usable identity is a
+C-number to a cached ``.step`` file, fetching it on demand from EasyEDA when a
+committed per-board sidecar maps its ``lib_id`` to a C-number.
+
+**License posture.**  EasyEDA/LCSC STEP bodies are design-use-oriented and are
+**not** explicitly redistributable, so the fetched models are cached locally
+and never committed to the repo.  Committed ``.kicad_pcb`` files carry only a
+portable path-variable ``(model "${KCT_LCSC_3D_DIR}/C#####.step" ...)`` ref
+into that cache, resolved at render time (mirroring the
+``${KICADn_3DMODEL_DIR}`` precedent).
+
+**Offset policy.**  An EasyEDA STEP is a bare ``.step`` with no ``.kicad_mod``,
+so there is no *source* footprint pad centroid to register against.  The bodies
+are treated as **origin-authored**: the resolver returns
+``source_anchor=(0.0, 0.0)`` (an explicit origin, *not* ``None``), so the
+shared ``add_model_refs_to_text`` offset math computes ``dx, dy =
+target_anchor`` -- the body's origin lands on the target footprint's pad
+centroid.  This is an approximation (origin-centered placement, not
+pin-1-registered) and scale/rotation are left at KiCad defaults (``1 1 1`` /
+``0 0 0``); a fetched body whose native orientation differs from the
+footprint's silkscreen may sit rotated.  These limitations are acceptable for
+cosmetic render bodies and noted for future per-mapping overrides.
+
+**Offline / CI safety.**  The fetch is opt-in.  A model resolves only when the
+cache already holds the STEP, or when fetching is explicitly enabled (via the
+``fetch`` flag / ``KCT_LCSC_FETCH`` env var).  Fetch and parse failures never
+raise -- they degrade to ``None`` (reported as unresolved) so a patch or render
+never fails for want of a body, and CI never needs network.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import urllib.error
+import urllib.request
+from pathlib import Path
+
+__all__ = [
+    "DEFAULT_CACHE_ENV_VAR",
+    "LCSC_MODEL_PATH_VAR",
+    "fetch_enabled",
+    "load_lcsc_mapping",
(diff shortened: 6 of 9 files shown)


rjwalters:
❌ **Changes Requested — unvalidated C-number is an exploitable attack surface**

The network-safety design is excellent and I verified it end-to-end: cache-only by default, all fetch/parse failures degrade to `None`, and the test suite is fully mocked with `forbidden`-urlopen guards on every offline/cache-hit path (113 passed locally, ruff + format clean on touched files, `_render_env` setdefault mirrors the `KICADn_3DMODEL_DIR` pattern correctly, offset math lands `(xyz 2 1.27 0)` for the `(2.0, -1.27)` centroid, pure-insertion preserved). But the **security review turned up a blocking, exploitable defect** that the sidecar C-number is never validated, and it flows unsanitized into three sinks.

### 1. Path traversal in the cache filename (BLOCKING)
`resolve_lcsc_step` builds `cache / f"{lcsc_id}.step"` with no sanitization. I proved this writes outside the cache dir:

```
malicious lcsc_id = "../outside/pwned"
resolve_lcsc_step(malicious, cache_dir=cache, fetch=True)
→ wrote to  <tmp>/cache/../outside/pwned.step   (escaped the cache dir)
```

A crafted/committed `lcsc_models.json` can drop attacker-controlled `.step` bytes anywhere the process can write. This code fetches third-party bytes onto user machines — the sidecar value must be treated as untrusted.

### 2. URL injection (BLOCKING)
The same value is `.format()`ed straight into the request URL:

```
_API_COMPONENT_INFO.format(lcsc_id="C1/../../evil?x=")
→ https://easyeda.com/api/products/C1/../../evil?x=/components
```

A malicious C-number can alter the request path / inject query params, redirecting the fetch to an attacker-chosen resource.

### 3. Ref injection into committed board files
`synthesize_model_block("../../../etc/whatever")` emits
`(model "${KCT_LCSC_3D_DIR}/../../../etc/whatever.step" ...)` into the committed `.kicad_pcb`.

### 4. No size cap / no content sanity check on the STEP write (BLOCKING per review brief)
`resolve_lcsc_step` does `step_path.write_bytes(data)` unconditionally — no maximum byte cap and no `ISO-10303` header check. The real body was ~1.9 MB; a compromised/rogue endpoint can return an arbitrarily large or non-STEP body that is written to disk verbatim.

### Requested fix (localized, ~15 lines)
1. Validate `lcsc_id` against `^C\d+$` (EasyEDA uses `C` + digits) at the trust boundary. Enforce it in `load_lcsc_mapping` — a malformed committed sidecar should `raise ValueError`, matching that function's existing "a broken committed sidecar is a build error" posture — **and** defensively in `resolve_lcsc_step` (reject → warn + return `None`, never touch the filesystem or network). Add a unit test for each: a traversal value (`../foo`), a URL-injection value (`C1/../../evil`), and a non-C value.
2. Before `write_bytes`: enforce a sane max size cap and verify the payload begins with the `ISO-10303-21` header; on failure, warn + return `None` (do not cache). Add tests for oversize and non-STEP bodies.

### Also: rebase required
The branch is BEHIND `origin/main` (PR #4074 and others merged after this branched). `mergeStateStatus` is currently `UNKNOWN`. Please `git rebase origin/main` and push before re-review.

Everything else is ready to ship — this is purely the input-validation hardening on the fetch/cache/emit path plus the rebase. I'll re-review promptly once the C-number is validated and the STEP write is bounded.

rjwalters:
Security feedback addressed — the untrusted C-number is now validated at every trust boundary and the STEP fetch is bounded.

**1. Path traversal + URL injection + ref injection (BLOCKING x3):** Added a single `^C\d+$` gate (`_is_valid_lcsc_id`) enforced at both trust boundaries:
- `load_lcsc_mapping` now **raises `ValueError`** on a malformed C-number (traversal `../foo`, URL-injection `C1/../../evil?x=`, non-C strings, lowercase `c`), matching the existing "broken committed sidecar is a build error" posture. A bad value can never reach the cache-filename / URL / model-ref sinks.
- `resolve_lcsc_step` **defensively rejects** an invalid id (warn + return `None`, never touching the filesystem or network), so the resolver path stays never-raise.

**2. No size cap / no content check (BLOCKING):** `_fetch_lcsc_step` now enforces a **50 MB cap** and requires the **`ISO-10303-21`** header before returning the body; an oversize or non-STEP payload is rejected (`None`) and never written to the cache.

**Tests (all mocked, forbidden-urlopen guards preserved):**
- `test_invalid_id_never_touches_fs_or_network` — parametrized over traversal / URL-injection / non-C ids; asserts no `.step` written under or above the cache dir.
- `test_invalid_cnumber_is_build_error` — parametrized sidecar values assert `ValueError`.
- `test_oversize_step_body_rejected` and `test_non_step_body_rejected`.

**Rebase:** rebased onto `origin/main` (clean, past PR #4074) and force-pushed with `--force-with-lease`.

Verification: `uv run pytest tests/test_lcsc_models.py tests/test_pcb_add_3d_models.py` → **91 passed**; ruff check + format clean on touched files; mypy clean on `lcsc_models.py`. Ready for re-review.

rjwalters:
✅ **Approved!** (scoped security re-review after Doctor pass)

The Doctor's delta is exactly the security fix + clean rebase — nothing else. Verified end to end.

## Delta scope
- Rebased onto current `origin/main` (`merge-base HEAD origin/main` == `origin/main` HEAD `ababc3bd`). No stale/behind state.
- Two commits on the branch: the original feature (`52216bf9`) + the security fix (`d52dbb3d`).
- Security commit touches **only** `src/kicad_tools/pcb/lcsc_models.py` and `tests/test_lcsc_models.py` (+109 lines). No scope creep.

## Original exploits re-run against the live worktree — all neutralized
1. **Cache-filename traversal** — `resolve_lcsc_step("../outside/pwned", fetch=True)` returns `None`, writes **zero** files under the cache dir, does **not** escape above it (`tmp.parent/outside/pwned.step` absent), and never reaches the network (a `urlopen` patched to raise never fired). Warned.
2. **URL `.format` injection** — `resolve_lcsc_step("C1/../../evil?x=", fetch=True)` is rejected by `_is_valid_lcsc_id` **before** any `.format`/`urlopen`. Returns `None`, no network. Warned.
3. **Committed-ref / non-C injection at load** — `load_lcsc_mapping` raises `ValueError` on a malformed C-number (build-error posture), so the value never reaches the filename/URL/ref sinks.
4. **Body bound** — 50 MB cap + `ISO-10303-21` magic check in `_fetch_lcsc_step` reject oversize / non-STEP payloads before they touch the cache.

## Validator airtightness (filename sink)
- `_LCSC_ID_RE = ^C\d+$` via `.match()` anchors both ends. The multiline-injection payload `C50950\n../evil` is **rejected** (content after the newline fails `$`). A trailing-newline-only value (`C50950\n`) technically passes, but carries no `/` or `..` — it cannot traverse; worst case is a cosmetically-odd filename inside the cache. Not a defect.
- Unicode-digit note (`\d` matches e.g. Arabic-Indic digits): confirmed **not** a traversal/injection risk — decimal digits cannot form `../` or `/` or URL-control chars. Reasoning sound. (`re.ASCII` would be a nice-to-have hardening, not a blocker.)

## Guard placement
The `_is_valid_lcsc_id` check is the **first** statement in `resolve_lcsc_step`, before `cache / f"{lcsc_id}.step"`, `.is_file()`, and any fetch — correct trust-boundary position.

## Tests
- New tests genuinely assert **no filesystem touch**, not just `return None`: `test_invalid_id_never_touches_fs_or_network` patches `urlopen` to raise on any call AND asserts `not any(tmp_path.rglob("*.step"))` AND no escape above the cache. `test_invalid_cnumber_is_build_error` asserts `ValueError` at load for all exploit strings (incl. lowercase `c`).
- Local: `tests/test_lcsc_models.py` + `test_pcb_add_3d_models.py` + `test_render_cmd.py` → **126 passed**. Ruff check + format clean on touched files.

## CI
All **16** checks green (Lint, Type Check, Test, C++ Build, Routed DRC, kicad-cli smoke, Boards 00–07 E2E/LVS, Board 05 routing, diff-pair, match-group). Merge state **CLEAN** against current main.

Security posture is now solid at every trust boundary. Approving.

Issue #2709, opened 2026-05-11
router (C++): _mark_via should consult _reserved_for_nets for corridor reservation parity


rjwalters:
## Symptom

PR #2686 (#2677, BGA partner-via escape corridor reservation) added \`Grid._reserved_for_nets\` and \`reserve_corridor_cells\` to enable inner-layer corridor reservation for paired escapes. Python's \`Grid._mark_via\` consults the reservations.

The judge on PR #2686 noted: **"C++ backend \`mark_via\` doesn't consult reservations — fine for this PR's flow (escape phase is Python-grid only), but worth a comment for future ports."**

If/when the escape pass moves to C++ (or any new C++ code paths place vias), corridor reservations will be silently ignored, regressing the board 06 BGA fix (USB3_TX1+/- would become unrouted again).

## Why it matters

- The corridor reservation is the foundation for Epic #2661 Phase 2's group-of-pairs symmetric serpentine — the same primitive will be extended to N-trace groups.
- If the C++ backend is ever extended to handle escape routing, the reservation logic needs parity or board 06 / future DDR-style boards will regress.
- Low priority today (no C++ via placement in the escape phase), but **a comment + test catches future regressions**.

## Recommended approach

1. Add a comment in \`src/kicad_tools/router/grid.py\` at \`_mark_via\` explaining the Python-only reservation contract.
2. Add a comment in the C++ binding (\`bindings.cpp\` or \`grid.cpp\`) at the equivalent \`mark_via\` site noting the Python parity gap.
3. Add a regression test that verifies the Python-only contract: any call to \`Grid._mark_via\` from the C++ binding either (a) doesn't get reservation enforcement (current behavior, document it) or (b) properly consults reservations (the future fix).
4. File a follow-up to actually port the reservation logic to C++ when escape routing migrates.

## Acceptance criteria

- [ ] Comment in Python \`grid.py\` documenting the Python-only reservation contract.
- [ ] Comment in C++ binding documenting the parity gap.
- [ ] Regression test asserts current behavior (reservations are Python-only) — locks the contract so future changes are deliberate.

## Related

- PR #2686 (#2677): introduced corridor reservation.
- Epic #2661 (match-group meander): will extend corridor reservation to N-trace groups.
- \`src/kicad_tools/router/grid.py\` (Python implementation)
- \`src/kicad_tools/router/cpp_backend.py\` / \`bindings.cpp\` (C++ binding surface)

## Out of scope

[...]

PR #4078, opened 2026-07-13
router: port C++ grid corridor-reservation keep-out + attractor (Closes #4071)

### LINKED ISSUE #4071: router: C++ grid ignores soft corridor reservations — port _reserved_for_nets check + attractor to grid.cpp (prerequisite for all reservation-based planners)
## Discovery (from #4053 / PR #4070 measurements)

The bundle-river planner v1 (PR #4070) proved its via-hop reservations work on the Python backend (DDR repro 7/11 → 8/11) but are **byte-for-byte inert on the production C++ backend**: \`src/kicad_tools/router/cpp/src/grid.cpp:126-143\` deliberately omits the \`_reserved_for_nets\` corridor check, and \`grid.py\`'s \`reserve_corridor_cells\` has no C++ sync path — the C++ A* never sees reservation keep-outs or attractors.

## Impact

Every soft-reservation feature is silently ignored whenever the C++ backend is active (i.e. in production and CI):
- #2983 / PR #2987 inner-corner lane reservations (shipped, believed active — actually inert on C++)
- #4053 / PR #4070 bundle river planner v1
- Any future reservation-based planner

## Scope

Port the reservation semantics to the C++ grid: marshal \`_reserved_for_nets\` (and the attractor cost shaping) across the boundary in the existing CppGrid pattern, honor it in the C++ A* neighbor/cost path, bump ROUTER_CPP_BUILD_VERSION + _REQUIRED_CPP_BUILD_VERSION in lockstep, parity tests (Python vs C++ with reservations active must agree on keep-out behavior), and re-run the #4053 DDR repro on C++ with \`--bundle-river-planner\` to get the true planner measurement.

## Acceptance
- C++ backend honors reservations (keep-out + attractor) with parity tests.
- Re-measured DDR repro (C++ + planner flag) posted — this is the deferred half of #4053's measurement.
- Boards 00-07 route gates green (reservations only engage where features request them; #2983's inner-corner reservations becoming ACTIVE on C++ may change board routes — measure and report, do not silently regen artifacts).

Parent epic: #4049. Refs: #4053, PR #4070, #2983/PR #2987, cpp/src/grid.cpp:126-143.

rjwalters:
Closes #4071.

Ports the Python `_reserved_for_nets` soft corridor-reservation contract
(Issue #2677 / PR #2686 — previously Python-grid-only per the #2709
deliberate-omission comment) into the production C++ backend. Both effects
of the one per-cell owner-set structure are now honored on the C++ grid:
the **hard keep-out** in `mark_via` and the **soft attractor** in the A*
cost loop. This activates #2983's inner-corner lane reservations and
#4053/PR #4070's bundle-river via-hop reservations on the C++ backend for
the first time (they were byte-for-byte inert before).

## What changed

- **`types.hpp`** — `GridCell` gains a fixed-capacity owner set
  (`reserved_nets[RESERVED_NETS_CAP=4]` + `reserved_count`) following the
  per-cell `GridCell` pattern (NOT a `std::set`/`std::vector` per cell —
  the grid is dense). `DesignRules` gains `cost_corridor_attractor`
  (default 3.0, matching `rules.py:228`). `ROUTER_CPP_BUILD_VERSION` 15 → 16.
- **`grid.hpp` / `grid.cpp`** — `Grid3D::mark_via` skips cells reserved for
  a net set excluding the via's net (keep-out), matching Python `_mark_via`
  exactly; matching-net vias treat the cell as ordinary. New
  `reserve_cell` / `clear_reservations` / `reserved_cell_count` /
  `is_reserved_for` / `corridor_attractor_bonus`, plus a grid-wide
  `has_reservations_` fast path so reservation-free boards stay
  byte-identical. The #2709 deliberate-omission comment block is rewritten
  to describe the ported contract.
- **`pathfinder.cpp`** — all 4 A* cost sites (one-shot + resumable, lateral
  step + via drop) subtract the attractor bonus for owning-net reserved
  cells, clamped at 0 (A* admissibility preserved), mirroring
  `pathfinder.py:2985/3108/4465/4565`.
- **`bindings.cpp`** — `DesignRules.cost_corridor_attractor`, the Grid3D
  reservation API, and read-only `GridCell.reserved_count` on the nanobind
  surface.
- **`cpp_backend.py`** — `_REQUIRED_CPP_BUILD_VERSION` 15 → 16;
  `from_routing_grid` bulk-copies `_reserved_for_nets`; both DesignRules
  marshalling blocks set `cost_corridor_attractor`.
- **`grid.py`** — `reserve_corridor_cells` / `clear_corridor_reservations`
  mirror incrementally onto the attached C++ grid (reservations made after
  `from_routing_grid` — e.g. during `_apply_byte_lane_inner_priority` —
  still propagate). The `_mark_via` #2709 docstring is rewritten for the
  ported contract.
- **`tests/test_grid_cpp_parity.py`** — INVERTED per the file's own
  docstring instruction: `test_cpp_grid_honours_python_reservations` now
  asserts the C++ grid SKIPS a foreign-net via at a reserved cell
  (agreement with Python), + new `test_cpp_reservation_incremental_mirror`
  and `test_cpp_attractor_discounts_owning_net_step_cost` parity tests.

## Version bump

`ROUTER_CPP_BUILD_VERSION` + `_REQUIRED_CPP_BUILD_VERSION` bumped **15 → 16**
in lockstep (started from 15 — PR #4069's coupled-loop port merged to main
first and already bumped 14 → 15). `kct build-native --check` confirms the
rebuilt `.so` reports required==actual==16.

## Port is exercised in production (definitive proof)

In a real board-07 route with the C++ backend active, instrumenting
`reserve_corridor_cells`: **9 reservation calls, 24528 Python cells, all 9
mirrored to the C++ grid, `Grid3D::reserved_cell_count() == 6608`.** Before
this PR that count was 0 (the C++ grid ignored reservations). #2983's and
#4053's reservations now reach the production C++ A*.

## DDR-bundle repro re-measurement (deferred half of #4053)

Re-ran the #4053/PR #4070 DDR-isolated repro on the **production C++ backend
with the port ACTIVE** (`--skip-nets` isolating the 11 DDR nets, `--seed 42
--targeted-ripup --max-ripups-per-net 3 --per-net-timeout 30 --timeout 300`).
Recipe-faithful config (`--net-class-map` sidecar + `--length-match-groups`,
which the planner needs to detect the DDR match group):

| Backend | `--bundle-river-planner` | Routed | Stranded |
| --- | --- | --- | --- |
| C++ (port active) | off | **10/11** | DQ3 |
| C++ (port active) | on  | **10/11** | DQ3 |

The two output PCBs **differ** (different copper) — confirming the via-hop
reservations now shape the C++ route rather than being ignored — but the
reach outcome is 10/11 either way, DQ3 stranded (matching the committed
board-07 baseline's DQ3). So the honored via-hop reservation changes the
path taken but does **not** net-improve reach beyond 10/11 on this
repro/seed. (Without the sidecar the group isn't declared and the planner
can't detect it — that run lands 9/11, DQ5+DM0; not a valid planner test.)
This is the deferred measurement #4053 could not complete; I'll cross-post
this table to #4053.

## Boards 00-07 route-delta findings

The #2983 corridor reservation is unconditional, so every board was checked
(not assumed). Reservation-engagement was measured directly (wrapping
`reserve_corridor_cells` + reading the C++ `reserved_cell_count`):

| Board | Reservations engaged? | Note |
| --- | --- | --- |
| 00-simple-led | **0** | fast path — byte-identical |
| 01-voltage-divider | **0** | fast path — byte-identical |
| 02-charlieplex-led | **0** | fast path — byte-identical |
| 03-usb-joystick | **0** | fast path — byte-identical |
| 04-stm32-devboard | **0** | fast path — byte-identical |
| 05-bldc-motor-controller | **0** | fast path — byte-identical |
| 06-diffpair-test | yes (pair-continuation) | `_reserve_pair_continuation_corridor` now C++-honored |
| 07-matchgroup-test | yes (6608 C++ cells) | #2983 + #4053 corridors now C++-honored |

Boards 00-05 make **zero** reservations, so the grid-wide `has_reservations_`
fast path guarantees byte-identical behaviour there (verified, not assumed).
Boards 06 and 07 engage reservations that this PR makes the C++ backend
honor for the first time.

### Board 07 full-recipe caveat (host is wall-clock-bound)

Board 07's full `generate_design.py --step route` is **wall-clock-bound on
this host**: BOTH origin/main (26/31) and this branch (25/31), solo, hit the
`--timeout 600` backstop before the `--deterministic-budget` iteration cap
completes ("Stopped due to timeout"). The committed 28/31 baseline was
measured on a faster/less-loaded machine. Because the wall-clock cliff
dominates (a pre-existing condition the board's own recipe comments document
extensively — "reach comparisons are only valid solo" and even solo it is
timeout-bound here), a clean full-recipe A/B is not attributable to this
port on this host. The DDR-isolated repro (fast, completes within budget) is
the port-attributable measurement, and it shows 10/11 → 10/11.

**No committed artifacts were regenerated in this PR** (route deltas are a
finding, not a silent regen). If board 06/07's committed artifact warrants a
refresh once a non-timeout-bound host confirms a real route improvement,
that is a follow-up issue — not folded in here.

## Sibling collision note (#4069)

PR #4069 already merged to main and bumped the build version 14 → 15; this
PR takes 15 → 16. There is no outstanding stale bump to rebase. (The #4071
curation predated #4069's merge and assumed 14 → 15; the actual lockstep is
15 → 16.)

## Tests

- `tests/test_grid_cpp_parity.py`: 4 passed (inverted keep-out + incremental
  mirror + attractor parity).
- Broad router sweep (`tests/router/` + cpp backend/grid/pathfinder/escape/
  diffpair/corridor/ripup suites): **808 passed, 9 skipped, 0 failed**.
- Reservation regression suites (byte-lane priority, byte-lane corridor
  reservation, bundle-river planner/inversions, escape diffpair, corridor
  integration, diffpair corridor): 125 passed.
- `ruff check` / `ruff format`: clean. `mypy`: no new errors on changed
  files (the pre-existing grid.py baseline is unchanged; my added lines are
  not among the flagged lines).

## Deferred scope

- Coupled joint-state A* (#4069's `CoupledPathfinder`) does not yet consume
  the attractor — `cost_corridor_attractor` is marshalled into its
  `DesignRules` for struct consistency, but wiring the bonus into
  `coupled_pathfinder.cpp`'s cost loop is left for a follow-up (single-ended
  `Pathfinder` is the consumer this issue wires up).
- Owner sets > `RESERVED_NETS_CAP=4` truncate (documented ceiling; no
  current caller exceeds 2 — diff pairs are 2, #2983/#4053 are single-net).

### DIFF
--- a/src/kicad_tools/router/cpp/include/grid.hpp
+++ b/src/kicad_tools/router/cpp/include/grid.hpp
@@ -77,17 +77,70 @@ public:
     // Route marking with clearance buffer using Bresenham
     void mark_segment(int x1, int y1, int x2, int y2, int layer, int net,
                       int clearance_cells);
-    // Issue #2709: ``mark_via`` does NOT consult the corridor reservation
-    // map that Python's ``RoutingGrid._mark_via`` enforces (Issue #2677).
-    // Safe today because the escape phase never marks vias through the
-    // C++ backend; see ``cpp/src/grid.cpp`` for the full rationale and
-    // ``tests/test_grid_cpp_parity.py`` for the contract-locking test.
+    // Issue #4071: ``mark_via`` NOW consults the per-cell corridor
+    // reservation owner set (``GridCell::reserved_nets``), mirroring
+    // Python's ``RoutingGrid._mark_via`` (Issue #2677).  A cell reserved
+    // for a net set that EXCLUDES ``net`` is SKIPPED (the via halo does
+    // not claim/block it), so partner-net through-hole vias cannot
+    // colonise a reserved continuation corridor.  Cells reserved for a
+    // set that INCLUDES ``net`` are treated as ordinary blockable cells.
+    // Fast path: when no cell is reserved (``has_reservations_ == false``)
+    // the check is skipped entirely, preserving byte-identical behaviour.
+    // See ``cpp/src/grid.cpp`` and ``tests/test_grid_cpp_parity.py``.
     void mark_via(int x, int y, int net, int radius_cells);
 
     // Unmark route (rip-up)
     void unmark_segment(int x1, int y1, int x2, int y2, int layer, int net,
                         int clearance_cells);
     void unmark_via(int x, int y, int net, int radius_cells);
+
+    // -----------------------------------------------------------------------
+    // Issue #4071: corridor-reservation API (mirrors
+    // ``RoutingGrid.reserve_corridor_cells`` / ``is_reserved_for`` /
+    // ``clear_corridor_reservations`` / ``reserved_cell_count``).
+    // -----------------------------------------------------------------------
+
+    // Reserve a single cell on ``layer`` for the given owner net set.
+    // Owner sets larger than ``RESERVED_NETS_CAP`` are truncated to the
+    // first K nets (documented ceiling; no current caller exceeds 2).
+    // Overlapping reservations REPLACE (last-writer-wins), matching the
+    // Python semantics.  A single reserved cell flips the grid-wide
+    // ``has_reservations_`` fast-path flag.
+    void reserve_cell(int x, int y, int layer,
+                      const std::vector<int>& net_ids);
+
+    // Clear every corridor reservation (owner sets + fast-path flag).
+    void clear_reservations();
+
+    // Number of cells with a non-empty owner set (instrumentation /
+    // parity tests, mirrors ``RoutingGrid.reserved_cell_count``).
+    int reserved_cell_count() const;
+
+    // True iff the cell is reserved AND ``net`` is in its owner set.
+    // Mirrors ``RoutingGrid.is_reserved_for``.  Inline for the A* cost
+    // loop and ``mark_via`` hot paths.
+    inline bool is_reserved_for(int x, int y, int layer, int net) const {
+        if (!has_reservations_) return false;
+        if (!is_valid(x, y, layer)) return false;
+        const auto& cell = at(x, y, layer);
+        for (int i = 0; i < cell.reserved_count; ++i) {
+            if (cell.reserved_nets[i] == net) return true;
+        }
+        return false;
+    }
+
+    // True iff ANY cell is currently reserved (grid-wide fast path).
+    inline bool has_reservations() const { return has_reservations_; }
+
+    // Issue #4071: soft corridor-attractor bonus for a single cell.
+    // Returns ``bonus`` when the cell is reserved for ``net``, else 0.0.
+    // Mirrors ``RoutingGrid.get_corridor_attractor_bonus``.  Inline for
+    // the A* cost loop.
+    inline float corridor_attractor_bonus(int x, int y, int layer, int net,
+                                          float bonus) const {
+        if (!has_reservations_ || bonus <= 0.0f) return 0.0f;
+        return is_reserved_for(x, y, layer, net) ? bonus : 0.0f;
+    }
 
     // Congestion tracking
     float get_congestion(int x, int y, int layer) const;
@@ -215,6 +268,11 @@ private:
     }
 
     std::vector<GridCell> cells_;  // Flat array for cache efficiency
+    // Issue #4071: grid-wide fast-path flag -- true once any cell has a
+    // non-empty reservation owner set.  ``mark_via`` and the A* cost loop
+    // skip the per-cell reservation read entirely when this is false, so
+    // boards without active reservations keep byte-identical behaviour.
+    bool has_reservations_ = false;
     int cols_, rows_, layers_;
     float resolution_;
     float origin_x_, origin_y_;
--- a/src/kicad_tools/router/cpp/include/types.hpp
+++ b/src/kicad_tools/router/cpp/include/types.hpp
@@ -143,7 +143,38 @@ namespace router {
 // rebuild via ``kct build-native``.  The pure-Python ``route_coupled`` is
 // preserved as the fallback for the ``allow_swap_via`` /
 // ``manhattan_sum`` code paths the v1 port intentionally defers.
-constexpr int ROUTER_CPP_BUILD_VERSION = 15;
+// Version 16 (Issue #4071): port the Python ``_reserved_for_nets`` soft
+// corridor-reservation contract (Issue #2677 / PR #2686, previously
+// Python-grid-only per the #2709 deliberate-omission comment) into the
+// C++ grid.  ``GridCell`` gains a small fixed-capacity owner-set
+// (``reserved_nets`` / ``reserved_count``, K=RESERVED_NETS_CAP), and
+// ``Grid3D::mark_via`` now honours the per-cell keep-out skip so a
+// foreign-net through-hole via cannot colonise a reserved continuation
+// corridor -- matching Python ``RoutingGrid._mark_via`` exactly.
+// ``DesignRules`` gains ``cost_corridor_attractor`` (Python default
+// 3.0) and ``pathfinder.cpp``'s neighbour/via cost loops subtract the
+// attractor bonus for cells reserved for the querying net (clamped at 0,
+// A* admissibility preserved), mirroring
+// ``pathfinder.py::get_corridor_attractor_bonus``.  This activates
+// #2983's UNCONDITIONAL inner-corner lane reservations and #4053/PR #4070's
+// bundle-river via-hop reservations on the production C++ backend for the
+// first time.  Old .so files lack the ``reserved_*`` GridCell fields and
+// the ``cost_corridor_attractor`` DesignRules field; the version bump
+// forces a rebuild via ``kct build-native``.
+constexpr int ROUTER_CPP_BUILD_VERSION = 16;
+
+// Issue #4071: fixed-capacity owner-set size for per-cell corridor
+// reservations.  Observed owner sets in practice are tiny: 1 for the
+// single-net #2983 inner-corner / #4053 bundle-river corridors, 2 for a
+// diff-pair continuation corridor (#2677).  K=4 comfortably covers every
+// current caller (and the Epic #2661 Phase 2 group-of-pairs path up to a
+// quad).  A dense-grid GridCell array (cols*rows*layers cells) cannot
+// afford a ``std::set``/``std::vector`` per cell, so the owner set is a
+// flat fixed array + count -- absent-cell fast path is ``reserved_count
+// == 0`` (byte-identical to the pre-#4071 behaviour on boards with no
+// active reservations).  Owner sets larger than K are truncated to the
+// first K nets (documented ceiling; no current caller exceeds 2).
+constexpr int RESERVED_NETS_CAP = 4;
 
 // Grid cell state
 struct GridCell {
@@ -161,6 +192,15 @@ struct GridCell {
     // ``mark_blocked``.  Route marking never sets this; rip-up uses it
     // to restore static blockage instead of freeing the cell.
     bool static_blocked = false;
+    // Issue #4071: corridor-reservation owner set (mirrors the Python
+    // ``RoutingGrid._reserved_for_nets`` per-cell ``frozenset[int]``).
+    // ``reserved_count == 0`` means "not reserved" (fast path).  A
+    // non-empty set makes this cell a keep-out for foreign-net vias in
+    // ``mark_via`` and a soft attractor (cost discount) for the owning
+    // net(s) in the A* cost loop.  Overlapping reservations REPLACE
+    // (last-writer-wins), matching Python ``reserve_corridor_cells``.
+    int8_t reserved_count = 0;
+    int32_t reserved_nets[RESERVED_NETS_CAP] = {0, 0, 0, 0};
 };
 
 // A* node for priority queue
@@ -422,6 +462,12 @@ struct DesignRules {
     float cost_congestion = 5.0f;
     float congestion_threshold = 0.5f;
     float min_drill_clearance = 0.102f;
+    // Issue #4071: soft corridor-attractor bonus (mirrors Python
+    // ``DesignRules.cost_corridor_attractor``, default 3.0).  Subtracted
+    // from a cell's positive step cost when the cell is reserved for the
+    // querying net, clamped at 0 so g_scores stay non-negative (A*
+    // admissibility preserved).  ``0.0`` disables the attractor.
+    float cost_corridor_attractor = 3.0f;
 };
 
 // Pad info for geometric validation (Issue #2439)
--- a/src/kicad_tools/router/cpp/src/bindings.cpp
+++ b/src/kicad_tools/router/cpp/src/bindings.cpp
@@ -38,7 +38,11 @@ NB_MODULE(router_cpp, m) {
         // ``unmark_segment`` / ``unmark_via`` so rip-up restores static
         // blockage instead of erasing it.
         .def_rw("static_blocked", &GridCell::static_blocked)
-        .def_rw("avoidance_cost", &GridCell::avoidance_cost);
+        .def_rw("avoidance_cost", &GridCell::avoidance_cost)
+        // Issue #4071: per-cell corridor-reservation owner set count
+        // (read-only surface for parity tests; the owner nets are written
+        // via ``Grid3D::reserve_cell`` / read via ``is_reserved_for``).
+        .def_ro("reserved_count", &GridCell::reserved_count);
 
     // DesignRules struct
     nb::class_<DesignRules>(m, "DesignRules")
@@ -54,7 +58,9 @@ NB_MODULE(router_cpp, m) {
         .def_rw("cost_via", &DesignRules::cost_via)
         .def_rw("cost_congestion", &DesignRules::cost_congestion)
         .def_rw("congestion_threshold", &DesignRules::congestion_threshold)
-        .def_rw("min_drill_clearance", &DesignRules::min_drill_clearance);
+        .def_rw("min_drill_clearance", &DesignRules::min_drill_clearance)
+        // Issue #4071: soft corridor-attractor bonus (Python default 3.0).
+        .def_rw("cost_corridor_attractor", &DesignRules::cost_corridor_attractor);
 
     // ValidationResult struct (Issue #2439)
     nb::class_<ValidationResult>(m, "ValidationResult")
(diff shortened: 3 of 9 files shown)


rjwalters:
❌ **Changes Requested — CI Failing (board-07 copper-short regression)**

The C++ reservation port itself is well-engineered — I verified the code thoroughly and the safety architecture is sound for boards 00–06. But activating reservations on **board 07 introduces copper shorts**, which two required CI gates catch as a hard regression. This must be resolved before merge.

## Blast-radius verification (the safety argument) — mostly confirmed

| Board | Reservations | CI gate | Verified |
| --- | --- | --- | --- |
| 00–05 | 0 (fast path) | **PASS** | ✅ byte-identical claim holds — all six End-to-End / Board-05 Routing Regression gates green |
| 06 | engaged (pair-continuation) | **PASS** | ✅ 21/21 reach, DRC 18 within allowlist, route stayed valid with reservations honored |
| 07 | engaged (6608 cells) | **FAIL ×2** | ❌ route regressed — see below |

The `has_reservations_` fast path genuinely no-ops (grid.cpp) and boards 00–05 route byte-identical — that half of the safety argument is solid and CI-confirmed.

## The blocker: board-07 route got WORSE, not just different

The PR body frames board-07 as "wall-clock-bound on this host … a clean full-recipe A/B is not attributable to this port." The **CI runner is not wall-clock-bound** — both board-07 jobs completed the route (13–14 min) and produced a routed PCB. With reservations now honored, that route is a real electrical regression:

**Board 07 End-to-End (LVS known opens) — FAIL:**
```
expected ONLY kind='open' mismatches but got kinds ['open', 'short']:
9 mismatch(es) (3 short, 6 open); first: short DM0(U1.29) <-> DQ7(U1.35)
-- a copper short is a hard regression regardless of the known-opens allowance
```

**Match-Group Routing Regression (boards/07-matchgroup-test) — FAIL:**
```
DRC error count: 10 (allowed: 8)
DRC regression on re-routed matchgroup_test_routed.kicad_pcb: 10 error(s) exceeds allowlist value 8
```

**3 copper shorts (DM0↔DQ7, …)** is an electrical defect — a manufactured board would be broken. This is the "route got WORSE" case, not a "threshold needs bumping" case. Bumping the allowlist from 8→10 would NOT be acceptable here because the copper-short check is a hard gate independent of the allowlist, and shorts are non-negotiable.

This directly contradicts the DDR-repro narrative ("shapes but doesn't fix, 10/11 either way"). The isolated DDR repro may show no reach change, but the **full board-07 recipe** — which is what CI exercises — shows the reservation-shaped route colliding into shorts. The isolated repro under-measured the port's real effect on this board.

## What needs to happen

The reservation port is correct as a mechanism (parity tests pass, keep-out + attractor verified below). The problem is board-07's route with reservations *active* is worse. Options:

1. **Investigate the shorts.** The keep-out is supposed to *prevent* foreign-net vias from colonizing a corridor — yet board-07 now shorts DM0↔DQ7. Is the attractor pulling nets into a shared corridor without the keep-out protecting the crossing? Is the truncation-at-K or a marshalling gap letting a reservation lapse mid-route? The shorts suggest the reservation is shaping the route into a collision the pre-port router avoided.
2. If the reservations genuinely can't produce a shorts-free board-07 route on this seed, the port may need to be **gated off for match-group boards** until the corridor geometry is fixed (i.e., don't activate #2983/#4053 corridors on board-07 in this PR), so the port lands safely for 00–06 without regressing 07.
3. Regenerating the committed board-07 artifact is **not** a fix — the regenerated route is the one that shorts.

Please don't raise the allowlist to paper over this — the copper-short gate is independent of the allowlist and shorts are a hard stop.

## What I verified and PASSES (for the record)

- **Keep-out correctness**: C++ `mark_via` skip (`if (!owned) continue`) matches Python `_mark_via` exactly (`if owners is not None and via_net not in owners: continue`). Owning-net vias still block. REPLACE-not-merge semantics match Python's dict-assignment (`_reserved_for_nets[(l,y,x)] = owners`). Truncation at K=4 is silent but degrades safely (no caller exceeds 2; a dropped owner would only *widen* the keep-out, never let a foreign via land on a should-be-reserved cell).
- **Attractor**: all 4 A* cost sites (one-shot lateral/via + resumable lateral/via) subtract the bonus, clamped at 0. Confirmed exactly 4 `new_g` sites in pathfinder.cpp — no missed site. Python has exactly 4 matching sites (2985/3108/4465/4565). Symmetric.
- **Marshalling**: both bulk (`from_routing_grid`) and incremental (`reserve_corridor_cells` → `_impl.reserve_cell`) paths verified; coordinate ordering `(x,y,layer)` consistent; `clear_corridor_reservations` flows to C++.
- **Parity tests**: all 4 pass locally; inversion is correct (previously asserted C++ IGNORES → now asserts C++ HONORS). Incremental-mirror + attractor-bonus tests genuinely exercise the ported behavior.
- **Deferred coupled attractor**: safe — the coupled pathfinder has zero reservation references, so it silently ignores the attractor (matching pre-port behavior) rather than half-honoring it. The grid-wide keep-out still applies at commit time.
- **Version 15→16 lockstep**: `_REQUIRED_CPP_BUILD_VERSION == 16`, backend AVAILABLE=True (version guard passes). ruff/format clean, router suites 485+136 passed / 0 failed locally.

The engineering is genuinely strong — the issue is purely that board-07's reservation-active route is electrically worse, and CI is right to block it. Once the board-07 shorts are resolved (fixed or the port gated off for that board), I expect a clean pass.

rjwalters:
## Path B taken — single-ended byte-lane reservations gated off C++; board-07 honoring deferred

Thank you for the thorough review. You confirmed the port is correct for boards 00-06, and the board-07 copper shorts are a hard block. I timeboxed a Path A investigation, concluded the shorts are **not a keep-out gap the port introduced**, and took **Path B** (scope the activation). Details:

### Root cause of the shorts (why Path B, not Path A)
The port faithfully mirrors Python: the reservation keep-out lives on `mark_via` (vias only) in **both** grids, and the attractor is in the A* cost. There is no keep-out gap to fix — Python has the identical via-only-keep-out + attractor design.

The shorts come from the **reservation strategy**, now honored on C++ for the first time:
- The #2983 inner-corner and #4053 bundle-river reservations reserve a corridor for a **single net** (`net_ids={net_id}`).
- The attractor then concentrates that net's traces into its narrow single-net lane.
- The keep-out only fences off foreign **vias** — it cannot fence off foreign **lateral traces** crossing the corridor.
- On board 07's fully-reversed DDR byte (all C(11,2)=55 pairs cross), these single-net lanes pack tightly; adjacent nets' attracted traces run in abutment and cross → DM0↔DQ7 (and the other 2) short.

This is the reservation geometry interacting badly with 07's dense reversed bundle, not a port defect. Per #4053's own DDR A/B, honoring these reservations buys **no board-07 reach** (10/11 either way), so gating 07 off loses no measured capability — exactly your Path B rationale.

### The fix
Added `mirror_to_cpp` (default `True`) to `reserve_corridor_cells`. The two single-ended byte-lane helpers pass `mirror_to_cpp=False` (Python-only); #2677 pair-continuation (board 06) keeps the default `True` (mirrored + honored). A `_cpp_suppressed_reservations` set makes the bulk `from_routing_grid` copy skip suppressed cells too. When suppressed, the C++ grid never sets `has_reservations_` for those cells → its A* is byte-identical to the pre-#4071 backend on board 07, restoring today's shorts-free route.

**Board-07 reservation honoring on C++ is DEFERRED** as a follow-up (tracked on #4071/#4053) once the corridor geometry is fixed to fence lateral crossings. The port still delivers #4071's purpose — reservations WORK on C++ for the boards that use them safely today (00-06).

### Verification
- **Match-Group Routing Regression re-route** (the gate that failed 10>8): now **DRC error count 8 (allowed 8)**, **no `short` in the rule breakdown**, pour connectivity PASS, match-group rule exercised, EXIT 0.
- Board 06 + board 07 artifact gates: **96 passed, 1 skipped**.
- C++ parity suite: **4/4** (default `mirror_to_cpp=True` still exercises the ported keep-out + attractor).
- Router suites: **485 passed, 3 skipped** (1 nanobind interpreter-teardown artifact, pre-existing; the test passes in isolation).
- ruff + format clean; mypy baseline unchanged (37→37, no new errors at edited lines).
- C++ `BUILD_VERSION=16` lockstep with `_REQUIRED_CPP_BUILD_VERSION=16`; **version bump 15→16 unchanged** after rebase onto origin/main.

I did **not** raise any allowlist, and did **not** commit a re-routed board (the committed board-07 artifact is restored — the fix is the code gate, not regen).

rjwalters:
✅ **Approved!** — Follow-up (Path B) re-review PASS. All 16 checks green, merge state CLEAN.

## Scoped re-review: the Path B gate is correct and surgical

The delta since the last review is one Python-side commit (8dd54fa — no C++ sources touched) that gates the two single-ended byte-lane reservations off C++ mirroring. I verified each of the acid-test criteria:

**1. Surgical gating — confirmed.** `reserve_corridor_cells` gained a keyword-only `mirror_to_cpp: bool = True`. Exactly the two single-ended helpers opt out:
- `reserve_inner_corner_lane_corridor` (#2983, escape.py:2522) → `mirror_to_cpp=False`
- `reserve_bundle_river_via_hop_corridor` (#4053, escape.py:2718) → `mirror_to_cpp=False`

`_reserve_pair_continuation_corridor` (#2677 / board 06, escape.py:2331) keeps the **default `True`** — still mirrored + honoured on C++. This is the whole point of the port for board 06, and it is preserved.

**2. Both C++ paths respect suppression (no leak in either).**
- Incremental path (grid.py:4244): `mirror_to_cpp=False` sets `cpp_grid=None` (no `reserve_cell`) AND records the key in `_cpp_suppressed_reservations`. Python's `_reserved_for_nets` still gets the cell, so the Python backend honours it — but board 07 routes on **C++** in production (the match-group + E2E jobs build the C++ backend and route against it), so the suppression is what restores the shorts-free route.
- Bulk path (cpp_backend.py `from_routing_grid`): iterates `_reserved_for_nets` and `continue`s on any key in the suppressed set — so a reservation present at C++ construction time is skipped too. When suppressed, the C++ grid never sets `has_reservations_` for those cells → its A* is byte-identical to pre-#4071 on board 07.
- Re-enable edge handled: a later mirrored reservation over a previously-suppressed cell calls `discard(key)`. `clear_corridor_reservations` resets both maps between escape passes.

**3. CI acid test — GREEN (the blocker is resolved).**
- **Match-Group Routing Regression (boards/07-matchgroup-test): PASS** — was DRC 10 > allowed 8; now 8/8, no short violations.
- **Board 07 End-to-End (LVS known opens): PASS** — was 3 copper shorts (DM0↔DQ7 +more); now clean.
- **Board 06 End-to-End (LVS): PASS** — pair-continuation NOT accidentally suppressed.
- **Diff-Pair Routing Regression (board 06): PASS.**
- Boards 00–05: PASS (byte-identical preserved).

**4. Board 06 benefit preserved — proven by tests.** All four parity tests in `tests/test_grid_cpp_parity.py` call `reserve_corridor_cells(...)` on the **default `mirror_to_cpp=True`** path (bulk copy, incremental mirror, and attractor discount). Local parity suite: **4/4 PASS**. This is exactly the board-06 pair-continuation behaviour, so the port still reaches C++ for the boards that use it safely.

**5. No allowlist raised; no shorted board committed.**
- `git diff origin/main...HEAD -- .github/` is **empty** — the DRC allowlist (`<= 8`) is untouched. The match-group gate went green by producing a genuinely shorts-free route, not by loosening the floor.
- No board artifacts in the diff (`boards/**`, `tests/boards/**` unchanged) — the committed board 07 artifact is the same as main; the fix is the code gate, not a regen.
- Version lockstep intact: C++ `ROUTER_CPP_BUILD_VERSION = 16` (types.hpp:164) == Python `_REQUIRED_CPP_BUILD_VERSION = 16`; built `.so` reports `BUILD_VERSION: 16`.

**6. Scope clean.** 8 files: 5 C++ (unchanged since first-review PASS), cpp_backend.py (bulk-skip), escape.py (2 call sites), grid.py (param + suppression set), and the parity test. No unrelated changes.

Non-blocking note (not tracked as an issue — cannot occur with current call graph): if a cell were re-reserved with `mirror_to_cpp=True` while `_cpp_grid` is not yet attached, its earlier suppressed-key would not be discarded until attach. In practice the suppressed (#2983/#4053) and mirrored (#2677) helpers reserve disjoint cells for disjoint nets and both maps clear between passes, so this cannot resurrect the shorts.

Board-07 reservation honouring on C++ is correctly deferred as a follow-up (the corridor keep-out fences vias only, not foreign lateral traces — a reservation-strategy gap, not a port bug). The port lands its purpose for boards 00–06.

rjwalters:
✅ **Approved!** — board-07 copper-short regression resolved via Path B (single-ended byte-lane reservations gated off C++)

This is a re-review after the Doctor addressed my prior change request (the board-07 copper shorts, DM0↔DQ7 et al., that two required gates caught). The fix is exactly the scoped-activation approach I laid out as an acceptable option, and it is implemented cleanly and correctly.

## The two previously-failing gates now PASS
- **Board 07 End-to-End (LVS known opens)** — pass (was FAIL ×2 with `['open','short']`; the copper shorts are gone).
- **Match-Group Routing Regression (boards/07-matchgroup-test)** — pass (DRC now 8, allowlist 8; was 10>8). No allowlist was raised — verified in the diff and the Doctor's comment.

## Full CI: 16/16 pass, merge state CLEAN
All End-to-End gates (00–07), both routing-regression gates (05 blocking≤11, 06 diff-pair), C++ Build Check, Match-Group, Lint/Format/Type/Test, DRC, and kicad-cli smoke are green. `mergeStateStatus == CLEAN`.

## Path B verified in the diff (not just claimed)
- `reserve_corridor_cells` gains keyword-only `mirror_to_cpp: bool = True`. When `False`, `cpp_grid` short-circuits to `None` (no C++ write) and the key is added to `_cpp_suppressed_reservations`; when a later `True` reservation re-covers a cell it is `discard`ed from the suppression set. Correct on both edges.
- `from_routing_grid`'s bulk copy consults `_cpp_suppressed_reservations` and skips suppressed cells, so a suppression that pre-existed C++ construction is also honored — closing the only gap the incremental path couldn't cover.
- `clear_corridor_reservations` clears the suppression set AND flows `clear_reservations()` to the C++ grid.
- The two single-ended byte-lane helpers (`reserve_inner_corner_lane_corridor` #2983, `reserve_bundle_river_via_hop_corridor` #4053) pass `mirror_to_cpp=False`; **#2677 pair-continuation (board 06) keeps the default `True`** — so board 06's reservations remain mirrored and CI-confirmed valid (21/21 reach, DRC within allowlist).
- Net effect on board 07: no cell sets `has_reservations_` on the C++ grid for the suppressed byte-lane reservations, so the C++ A* is byte-identical to the pre-#4071 backend there → today's shorts-free route is restored. This is a code gate, not a regenerated (shorting) artifact — the committed board-07 PCB is untouched.

## C++ mechanism unchanged from my prior pass
Path B did not touch the C++ keep-out or attractor. My prior review already exhaustively verified: the `mark_via` skip (`if (!owned) continue`) matches Python `_mark_via`, the `has_reservations_` fast path no-ops on reservation-free boards, all 4 A* attractor sites subtract-and-clamp, K=4 truncation degrades safely, and the 15→16 build-version lockstep. Those files (`grid.cpp/.hpp`, `pathfinder.cpp`, `types.hpp`, `bindings.cpp`, `cpp_backend.py`) are the same well-engineered port I confirmed before.

## Tests
- Parity suite (`test_grid_cpp_parity.py`): the inverted keep-out test, the new incremental-mirror test, and the attractor-vs-Python-reference test all still exercise the **default `mirror_to_cpp=True`** path, so the C++ port itself stays under parity coverage even though board-07's helpers opt out. Good — the gate isn't hollowed out by Path B.
- Doctor reports 485 router passed (1 pre-existing nanobind teardown artifact that passes in isolation), 96 board 06/07 artifact-gate tests, ruff/format clean, mypy baseline unchanged.

## Deferred scope acknowledged (correctly not folded in)
Board-07 reservation honoring on C++ remains deferred until the corridor geometry fences lateral crossings — tracked as follow-up on #4071/#4053 (and #4079 for lateral keep-out). This is the right call: #4053's own DDR A/B showed honoring buys no board-07 reach (10/11 either way), so gating 07 off loses no measured capability while unblocking the port for the boards that use it safely today (00–06). The coupled-pathfinder attractor remains marshalled-but-not-consumed (safe: it ignores rather than half-honors).

PR description uses `Closes #4071`. The engineering is strong and the regression is genuinely resolved, not papered over. Nice work.

PR #4082, opened 2026-07-13
fix(diffpair): restore single-ended back-reference in try/finally on mid-copy failure

### LINKED ISSUE #4077: Harden _get_cpp_coupled_impl back-reference restore with try/finally
## Context

Discovered while reviewing PR #4069 (Issue #4065, the C++ coupled diff-pair port).

The reach-regression root-cause fix in `src/kicad_tools/router/diffpair_routing.py::_get_cpp_coupled_impl` snapshots and restores the single-ended router's `grid._cpp_grid` back-reference around `CppGrid.from_routing_grid(self.grid)`:

```python
saved_cpp_grid = getattr(self.grid, "_cpp_grid", None)
cpp_grid = CppGrid.from_routing_grid(self.grid)   # hijacks grid._cpp_grid at cpp_backend.py:559
self.grid._cpp_grid = saved_cpp_grid              # restore
impl = CppCoupledPathfinder(cpp_grid, ...)        # may throw -> caught below
```

The restore correctly runs **before** the `CppCoupledPathfinder(...)` constructor, so the primary failure mode (constructor throwing) is safe. However, the restore is a plain statement inside the `try`, not in a `finally`.

## Gap

`CppGrid.from_routing_grid` (cpp_backend.py) assigns `grid._cpp_grid = cpp_grid` at line 559, then continues copying blocked cells / pads (lines 561+), and — as of PR #4078 (Issue #4071, merged 2026-07-12) — also marshals corridor reservations further down in the same method. If `from_routing_grid` raises **after** line 559 (e.g. during the bulk cell copy or the new corridor-reservation marshalling on a malformed grid), the caller's restore at line 1533 is skipped and control jumps to `except Exception` at line 1547. The single-ended router's `grid._cpp_grid` is then left pointing at the partially-built coupled grid — the exact hijack the fix targets, re-introduced on the exception path.

**Note (verified against current `origin/main`, commit `e590d863741523d92e4486400b9f8a8ec17773c5`, 2026-07-12):** PR #4078 added corridor-reservation marshalling to `from_routing_grid` but did not shift any of the line numbers cited above — `grid._cpp_grid = cpp_grid` is still `cpp_backend.py:559`, and `diffpair_routing.py`'s `saved_cpp_grid`/`from_routing_grid`/restore/`except Exception` are still at lines 1527/1528/1533/1547 respectively. If PR #4078 had shifted these lines, it would only have *increased* the surface area of code that can raise after the hijack (more marshalling steps between the hijack and the restore), reinforcing rather than reducing the need for this fix.

## Impact

Low. Requires `from_routing_grid` to fail mid-copy, which is uncommon. Consequence is confined to the fallback-to-Python coupled path within the same session (rip-up invalidation targets the wrong grid). Not observed in practice; all CI green.

## Suggested fix

Wrap the `from_routing_grid` call and the restore in a `try/finally` so the back-reference is restored even if marshalling throws, while preserving the existing broader `try/except Exception` that already surrounds this code (for the `CppCoupledPathfinder(...)` constructor and the `from .cpp_backend import ...` line):

```python
saved_cpp_grid = getattr(self.grid, "_cpp_grid", None)
try:
    cpp_grid = CppGrid.from_routing_grid(self.grid)
finally:
    self.grid._cpp_grid = saved_cpp_grid
impl = CppCoupledPathfinder(
    cpp_grid,
    ...
)
```

The inner `try/finally` sits inside the existing outer `try` block (`diffpair_routing.py:1505`) that already routes any exception — including one raised inside `from_routing_grid` itself — to the `except Exception:` handler at line 1547 (`_use_cpp_coupled = False`, return `None`, fall back to Python). The `finally` only needs to guarantee the restore always runs; it does not need to re-raise or suppress anything, since the outer `except` already handles propagation.

Discovered during code evaluation of PR #4069.

## Acceptance Criteria

- [ ] In `_get_cpp_coupled_impl` (`src/kicad_tools/router/diffpair_routing.py`), the `CppGrid.from_routing_grid(self.grid)` call is wrapped in a `try/finally` where the `finally` block restores `self.grid._cpp_grid = saved_cpp_grid`.
- [ ] The restore happens exactly once per call to `_get_cpp_coupled_impl`, regardless of whether `from_routing_grid` succeeds, raises before assigning `grid._cpp_grid`, or raises after assigning it.
- [ ] The existing outer `try/except Exception` (currently `diffpair_routing.py:1505`-`1547`) is preserved unchanged in behavior: any exception from `from_routing_grid` or `CppCoupledPathfinder(...)` still results in `self._use_cpp_coupled = False` and a `None` return (fallback to Python), just as before this change.
- [ ] No change to the happy-path behavior: when `from_routing_grid` and `CppCoupledPathfinder(...)` both succeed, `_get_cpp_coupled_impl` still returns the constructed `impl` and caches `self._cpp_coupled_impl` / `self._cpp_coupled_grid` as before.
- [ ] A regression test demonstrates that `self.grid._cpp_grid` is restored to its pre-call value even when `CppGrid.from_routing_grid` raises *after* internally setting `grid._cpp_grid` (i.e., a mid-copy failure), not just when it raises before.

## Implementation Guidance

- Scope the change to the four lines currently at `diffpair_routing.py:1527-1533`:
  ```python
  saved_cpp_grid = getattr(self.grid, "_cpp_grid", None)
  cpp_grid = CppGrid.from_routing_grid(self.grid)
  ...
  self.grid._cpp_grid = saved_cpp_grid
  ```
  becomes:
  ```python
  saved_cpp_grid = getattr(self.grid, "_cpp_grid", None)
  try:
      cpp_grid = CppGrid.from_routing_grid(self.grid)
  finally:
      self.grid._cpp_grid = saved_cpp_grid
  ```
  Keep the existing explanatory comments (the Issue #4065 block above `saved_cpp_grid` and the "Restore the single-ended router's back-reference" comment) — just verify/adjust their placement so they still read correctly around the new `try/finally`.
- Do NOT move the `impl = CppCoupledPathfinder(...)` construction into the new inner `try/finally` — it should remain a plain statement after the `finally` block, still covered only by the pre-existing outer `try` (so a constructor failure continues to be handled by the outer `except Exception:` exactly as today).
- To simulate a "raises after line 559" failure in a test, monkeypatch or subclass `CppGrid.from_routing_grid` (or a helper it calls internally, e.g. the blocked-cell copy loop or the Issue #4071 corridor-reservation marshalling step) to raise partway through, after `grid._cpp_grid` has already been assigned. Verify `self.grid._cpp_grid` afterward equals the pre-call `saved_cpp_grid` value (or is unset/`None` if it was unset beforehand), not the partially-built grid.
- This is a small, self-contained, single-file change (~5 LOC diff) with no API or behavior change on the happy path — no `## Affected Files` beyond the one listed below and its test.

## Affected Files

- `src/kicad_tools/router/diffpair_routing.py` - Wrap `CppGrid.from_routing_grid(self.grid)` (currently line 1528) and the `self.grid._cpp_grid = saved_cpp_grid` restore (currently line 1533) inside `_get_cpp_coupled_impl` in a `try/finally`.
- `tests/test_diffpair_routing_integration.py` (or a new focused test module, e.g. `tests/router/test_cpp_coupled_backref_restore.py`) - Add a regression test that forces `CppGrid.from_routing_grid` to raise after it has set `grid._cpp_grid`, and asserts the back-reference is restored to its prior value.

## Test Plan

- [ ] Automated test: new/updated test that monkeypatches `CppGrid.from_routing_grid` (or an internal step it calls) to raise *after* `grid._cpp_grid` has been assigned, then calls `_get_cpp_coupled_impl()` and asserts `grid._cpp_grid` equals the pre-call snapshot (covers the gap this issue targets — the case NOT covered by existing tests, since the constructor-throws case was already safe).
- [ ] Automated test: existing happy-path coverage (e.g. `tests/test_diffpair_routing_integration.py` or wherever `_get_cpp_coupled_impl` is currently exercised via `route_coupled`) continues to pass unchanged, confirming no regression to the success path.
- [ ] Manual verification: `uv run pytest tests/ -k "diffpair and cpp" -v` passes; `uv run kct build-native --check` confirms the C++ backend is available before running (per repo `CLAUDE.md` guidance) so the C++ code path under test is actually exercised, not silently skipped/falled-back.
- [ ] Regression check: `uv run pytest tests/test_diffpair_routing_integration.py tests/test_diffpair_coupled_floor.py` (or the full diffpair suite) to confirm no interaction with the Issue #4065 reach-regression fix or the Issue #4071 corridor-marshalling addition.

rjwalters:
## Summary

Hardens `CoupledPathfinder._get_cpp_coupled_impl` (`src/kicad_tools/router/diffpair_routing.py`) so the single-ended router's `grid._cpp_grid` back-reference is restored even when `CppGrid.from_routing_grid` raises **after** it has already hijacked that back-reference at `cpp_backend.py:559` (a mid-copy failure).

The `#4065` reach-regression fix snapshots `grid._cpp_grid`, calls `from_routing_grid` (which reassigns it), then restores it before constructing `CppCoupledPathfinder`. The restore was a plain statement inside the `try`, so a raise mid-copy (bulk cell copy, or the `#4071` corridor-reservation marshalling) skipped the restore and left the single-ended router pointing at the partially-built coupled grid — re-introducing the exact hijack on the exception path.

## Changes

- Wrap `cpp_grid = CppGrid.from_routing_grid(self.grid)` and the `self.grid._cpp_grid = saved_cpp_grid` restore in an inner `try/finally` (finally does the restore). The `CppCoupledPathfinder(...)` constructor remains a plain statement after the `finally`, still covered only by the pre-existing outer `try/except Exception` — no behavior change on the constructor-throws path.
- Update the explanatory comment to note the `finally` semantics.
- Add `tests/test_cpp_coupled_backref_restore.py` (3 regression tests).

| `from_routing_grid` wrapped in `try/finally`, finally restores `grid._cpp_grid` | ✅ | See diff; `diffpair_routing.py` |
| Restore runs exactly once regardless of success / raise-before / raise-after | ✅ | `finally` runs unconditionally; happy-path + both mid-copy tests pass |
| Outer `try/except` preserved (exception → `_use_cpp_coupled=False`, return `None`) | ✅ | Both failure tests assert `impl is None` and `_use_cpp_coupled is False` |
| Happy path unchanged (returns `impl`, caches `_cpp_coupled_impl` / `_cpp_coupled_grid`) | ✅ | `test_happy_path_restores_backref_and_caches_impl` |
| Regression test for restore when `from_routing_grid` raises AFTER setting `grid._cpp_grid` | ✅ | `test_backref_restored_when_from_routing_grid_raises_after_hijack` + `_to_prior_value_when_previously_set` |

## Test Plan

- `uv run kct build-native --check` → `C++ backend: available (version 1.0.0)` (built in worktree first).
- `uv run pytest tests/test_cpp_coupled_backref_restore.py -v` → 3 passed.
- Regression: `uv run pytest tests/test_diffpair_coupled_cpp_parity.py tests/test_diffpair_coupled_floor.py tests/test_diffpair_routing_integration.py` → 23 passed.
- `uv run ruff check` + `ruff format --check` on both changed files → clean.

Closes #4077

### DIFF
--- a/src/kicad_tools/router/diffpair_routing.py
+++ b/src/kicad_tools/router/diffpair_routing.py
@@ -1525,12 +1525,19 @@ class CoupledPathfinder:
             # NOT steal the single-ended router's paired back-reference, so
             # snapshot ``grid._cpp_grid`` and restore it afterwards.
             saved_cpp_grid = getattr(self.grid, "_cpp_grid", None)
-            cpp_grid = CppGrid.from_routing_grid(self.grid)
             # Restore the single-ended router's back-reference (or ``None`` if
-            # it had none): the coupled pathfinder keeps ``cpp_grid`` in
-            # ``impl`` below, but the Python grid's ``_cpp_grid`` invalidation
-            # hook must continue to target the single-ended router's grid.
-            self.grid._cpp_grid = saved_cpp_grid
+            # it had none) in a ``finally`` so it is restored even when
+            # ``from_routing_grid`` raises AFTER hijacking ``grid._cpp_grid``
+            # mid-copy (e.g. during the bulk cell copy or the Issue #4071
+            # corridor-reservation marshalling): the coupled pathfinder keeps
+            # ``cpp_grid`` in ``impl`` below, but the Python grid's
+            # ``_cpp_grid`` invalidation hook must continue to target the
+            # single-ended router's grid.  The outer ``try/except Exception``
+            # still routes any raised exception to the Python fallback.
+            try:
+                cpp_grid = CppGrid.from_routing_grid(self.grid)
+            finally:
+                self.grid._cpp_grid = saved_cpp_grid
             impl = CppCoupledPathfinder(
                 cpp_grid,
                 self.rules,
--- a/tests/test_cpp_coupled_backref_restore.py
+++ b/tests/test_cpp_coupled_backref_restore.py
@@ -62,11 +62,20 @@ def test_backref_restored_when_from_routing_grid_raises_after_hijack(monkeypatch
     assert getattr(pf.grid, "_cpp_grid", None) is None
 
     hijacked = _Sentinel()
+    hijack_reached = {"value": False}
 
-    def _boom(_grid):
+    # NOTE: ``_boom`` MUST take ``(cls, _grid)``.  ``from_routing_grid`` is a
+    # classmethod, so ``CppGrid.from_routing_grid(self.grid)`` injects ``cls``
+    # as the first positional argument.  A one-param ``_boom(_grid)`` would
+    # raise ``TypeError`` at the call boundary -- BEFORE the hijack line runs
+    # -- degenerating this into a raise-*before*-hijack case that the pre-fix
+    # code already handled (so the test would pass even without the fix).  See
+    # Issue #4077 / PR #4082 Judge feedback.
+    def _boom(cls, _grid):
         # Simulate cpp_backend.py:559: from_routing_grid reassigns the
         # Python grid's back-reference to the (partially-built) coupled grid...
         _grid._cpp_grid = hijacked
+        hijack_reached["value"] = True
         # ...then raises before finishing the copy (bulk cell copy / #4071
         # corridor-reservation marshalling on a malformed grid).
         raise RuntimeError("mid-copy failure after grid._cpp_grid was set")
@@ -75,6 +84,14 @@ def test_backref_restored_when_from_routing_grid_raises_after_hijack(monkeypatch
 
     impl = pf._get_cpp_coupled_impl()
 
+    # Guard: the mock body must have actually run past the hijack line.  If
+    # this is False the mock signature is wrong and the test is silently
+    # simulating raise-before-hijack (which the pre-fix code already handled),
+    # providing no protection for the mid-copy gap Issue #4077 targets.
+    assert hijack_reached["value"], (
+        "mock body did not reach the hijack line -- signature mismatch "
+        "(from_routing_grid is a classmethod; _boom must take (cls, _grid))"
+    )
     # Outer try/except routes the failure to the Python fallback.
     assert impl is None
     assert pf._use_cpp_coupled is False
@@ -93,15 +110,25 @@ def test_backref_restored_to_prior_value_when_previously_set(monkeypatch):
     pf.grid._cpp_grid = prior
 
     hijacked = _Sentinel()
+    hijack_reached = {"value": False}
 
-    def _boom(_grid):
+    # ``_boom`` MUST take ``(cls, _grid)`` -- see the note in
+    # ``test_backref_restored_when_from_routing_grid_raises_after_hijack``.
+    def _boom(cls, _grid):
         _grid._cpp_grid = hijacked
+        hijack_reached["value"] = True
         raise RuntimeError("mid-copy failure after grid._cpp_grid was set")
 
     monkeypatch.setattr(cpp_backend.CppGrid, "from_routing_grid", classmethod(_boom))
 
     impl = pf._get_cpp_coupled_impl()
 
+    # Guard: the mock body must have actually run past the hijack line so this
+    # genuinely exercises the raise-after-hijack (mid-copy) path.
+    assert hijack_reached["value"], (
+        "mock body did not reach the hijack line -- signature mismatch "
+        "(from_routing_grid is a classmethod; _boom must take (cls, _grid))"
+    )
     assert impl is None
     assert pf._use_cpp_coupled is False
     # Restored to the exact prior back-reference, not the hijacked partial grid.
--- a/tests/test_cpp_coupled_backref_restore.py
+++ b/tests/test_cpp_coupled_backref_restore.py
@@ -1,0 +1,128 @@
+"""Regression tests for Issue #4077.
+
+``CoupledPathfinder._get_cpp_coupled_impl`` snapshots the single-ended
+router's ``grid._cpp_grid`` back-reference before calling
+``CppGrid.from_routing_grid(self.grid)`` (which hijacks that back-reference
+at ``cpp_backend.py:559``) and restores it afterwards.
+
+Issue #4065 established that the restore must run before the
+``CppCoupledPathfinder(...)`` constructor -- the constructor-throws case was
+already covered.  Issue #4077 hardens the restore into a ``try/finally`` so
+the back-reference is also restored when ``from_routing_grid`` raises AFTER
+it has already reassigned ``grid._cpp_grid`` (a mid-copy failure, e.g. during
+the bulk cell copy or the Issue #4071 corridor-reservation marshalling).
+
+These tests monkeypatch ``CppGrid.from_routing_grid`` to raise partway
+through -- after the hijack -- and assert the single-ended router's
+back-reference is restored to its pre-call value, not left pointing at the
+partially-built coupled grid.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from kicad_tools.router import cpp_backend
+from kicad_tools.router.cpp_backend import is_cpp_available
+from kicad_tools.router.diffpair_routing import CoupledPathfinder
+from kicad_tools.router.grid import RoutingGrid
+from kicad_tools.router.rules import DesignRules
+
+pytestmark = pytest.mark.skipif(
+    not is_cpp_available(),
+    reason="C++ coupled back-reference restore requires the router_cpp backend (kct build-native)",
+)
+
+
+def _make_pf() -> CoupledPathfinder:
+    grid = RoutingGrid(width=12.7, height=12.7, rules=DesignRules())
+    pf = CoupledPathfinder(
+        grid=grid, rules=DesignRules(), target_spacing_cells=2, min_spacing_cells=2
+    )
+    pf._use_cpp_coupled = True
+    return pf
+
+
+class _Sentinel:
+    """Stand-in for the partially-built coupled CppGrid."""
+
+
+def test_backref_restored_when_from_routing_grid_raises_after_hijack(monkeypatch):
+    """A mid-copy failure (raises AFTER assigning grid._cpp_grid) must still
+    restore the single-ended router's back-reference.
+
+    This is the case Issue #4077 targets: prior to the fix the restore was a
+    plain statement inside the ``try`` that was skipped when
+    ``from_routing_grid`` raised after line 559, leaving ``grid._cpp_grid``
+    pointing at the partial coupled grid.
+    """
+    pf = _make_pf()
+
+    # The single-ended router's back-reference is unset on a fresh grid.
+    assert getattr(pf.grid, "_cpp_grid", None) is None
+
+    hijacked = _Sentinel()
+
+    def _boom(_grid):
+        # Simulate cpp_backend.py:559: from_routing_grid reassigns the
+        # Python grid's back-reference to the (partially-built) coupled grid...
+        _grid._cpp_grid = hijacked
+        # ...then raises before finishing the copy (bulk cell copy / #4071
+        # corridor-reservation marshalling on a malformed grid).
+        raise RuntimeError("mid-copy failure after grid._cpp_grid was set")
+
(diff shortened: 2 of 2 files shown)


rjwalters:
❌ **Changes Requested**

The production change in `src/kicad_tools/router/diffpair_routing.py` is **correct** — the `try/finally` wraps only `from_routing_grid`, the `CppCoupledPathfinder(...)` constructor correctly stays outside the inner `finally` (still covered by the outer `try/except`), and the `finally` neither re-raises nor suppresses, so exception semantics to the outer handler are unchanged. Verified by reverting the fix and tracing: the pre-fix code does leave `grid._cpp_grid` pointing at the partially-built coupled grid on a raise-after-hijack, and the `finally` restores it.

**However, the regression tests do not actually exercise the gap this issue targets — they pass on the un-fixed code, so they provide zero protection.**

**Root cause:** The monkeypatch signature is wrong.

```python
def _boom(_grid):                 # ONE positional param
    _grid._cpp_grid = hijacked
    raise RuntimeError(...)
monkeypatch.setattr(cpp_backend.CppGrid, "from_routing_grid", classmethod(_boom))
```

With `classmethod`, calling `CppGrid.from_routing_grid(self.grid)` injects `cls` as the first argument, so Python invokes `_boom(cls, self.grid)` — two args into a one-param function. That raises `TypeError: _boom() takes 1 positional argument but 2 were given` **at the call boundary, before the body runs**. The `_grid._cpp_grid = hijacked` line is never reached, so the "mid-copy hijack" the test claims to simulate never happens. The `TypeError` is a raise-*before*-hijack, which the pre-fix code already handled — hence both `test_backref_restored_when_from_routing_grid_raises_after_hijack` and `test_backref_restored_to_prior_value_when_previously_set` pass whether or not the `finally` is present.

**Proof:** I reverted the source to the plain-statement form and ran the suite — all 3 tests still passed. I then confirmed via direct trace that `classmethod(_boom)` raises `TypeError` before `_cpp_grid` is ever set (`hasattr(g, "_cpp_grid") == False`).

This fails acceptance criterion 5: *"A regression test demonstrates that `grid._cpp_grid` is restored... even when `CppGrid.from_routing_grid` raises AFTER internally setting `grid._cpp_grid` (a mid-copy failure), not just when it raises before."*

**Fix:** Give `_boom` a `cls` parameter so the body actually runs:

```python
def _boom(cls, _grid):            # TWO params: cls + the grid
    _grid._cpp_grid = hijacked
    raise RuntimeError("mid-copy failure after grid._cpp_grid was set")
monkeypatch.setattr(cpp_backend.CppGrid, "from_routing_grid", classmethod(_boom))
```

I verified the corrected 2-param form is a genuine regression test: with it, the mid-copy test **fails on the pre-fix code** (`grid._cpp_grid` left == hijacked) and **passes on the fixed code** (restored to prior/None). Apply this to both `test_backref_restored_when_from_routing_grid_raises_after_hijack` and `test_backref_restored_to_prior_value_when_previously_set`. The happy-path test is fine as-is.

Please also add a guard so the mock-body-was-reached is asserted (e.g. verify `getattr(pf.grid, "_cpp_grid", None)` was actually hijacked mid-call, or assert the raised exception type is `RuntimeError` not `TypeError`) — this prevents the same silent-skip regression from recurring.

**Verification performed** (worktree, C++ backend built via `uv run kct build-native` — `router_cpp.cpython-314-darwin.so`, so the C++ path is exercised, not skipped):
- Source `try/finally` semantics: correct, exception path preserved.
- Pre-fix code: leaves `_cpp_grid` = partial grid on raise-after-hijack (bug confirmed present).
- PR tests as written: pass on BOTH pre-fix and fixed code (no protection).
- Corrected 2-param `_boom`: fails on pre-fix, passes on fixed (genuine regression test).

Once the two mid-copy tests use the 2-param signature (and demonstrably fail without the `finally`), this is a clean approve.

rjwalters:
Review feedback addressed. The production `try/finally` fix in `src/kicad_tools/router/diffpair_routing.py` was already correct (per the Judge) and is unchanged — only the regression tests were hardened.

**Judge point 1 — wrong `_boom` signature (raise-before-hijack degeneracy):** Both mid-copy tests now define `def _boom(cls, _grid)`. Since `from_routing_grid` is a `classmethod`, `CppGrid.from_routing_grid(self.grid)` injects `cls` as the first positional arg; the two-param signature absorbs it so the `_grid._cpp_grid = hijacked` body actually runs before the `RuntimeError`. The old one-param form raised `TypeError` at the call boundary before the hijack, simulating raise-*before*-hijack (which the pre-fix code already handled).

**Judge point 2 — guard that the mock body was reached:** Each mid-copy test now sets a `hijack_reached` flag immediately after the hijack line and asserts it (`assert hijack_reached["value"]`, with a message pointing at the classmethod signature requirement). A future signature drift that silently skips the hijack now fails loudly instead of passing vacuously.

**Judge point 3 — revert-check proving genuine regression coverage:** I temporarily reverted the source to the plain-statement form and re-ran the module:
- `test_backref_restored_when_from_routing_grid_raises_after_hijack` → FAILED
- `test_backref_restored_to_prior_value_when_previously_set` → FAILED
- `test_happy_path_restores_backref_and_caches_impl` → PASSED

Restored the `try/finally` fix and re-ran: all 3 PASSED. So the two mid-copy tests now fail without the `finally` and pass with it — satisfying acceptance criterion 5.

**Verification** (worktree, `uv run kct build-native --check` → `C++ backend: available (version 1.0.0)` — C++ path exercised, not skipped):
- `uv run pytest tests/test_cpp_coupled_backref_restore.py -v` → 3 passed.
- `uv run ruff check` + `ruff format --check` on the changed test file → clean.

Only `tests/test_cpp_coupled_backref_restore.py` changed (+30/-3). Ready for re-review.

rjwalters:
✅ **Approved!** (re-review after Doctor pass, commit b6ce99e7)

The prior change request is fully resolved and my independent verification confirms the fix has real teeth.

**Prior concern (acceptance criterion 5):** the two mid-copy tests monkeypatched `from_routing_grid` with a one-param `_boom(_grid)` under `classmethod`, so `cls` injection raised `TypeError` at the call boundary *before* the hijack line — degenerating both into raise-before-hijack cases the pre-fix code already handled. They passed even on un-fixed code.

**Doctor's fix (test file only):**
- `_boom(cls, _grid)` now absorbs the classmethod `cls` injection, sets `_grid._cpp_grid = hijacked` and flips a `hijack_reached` flag *before* raising `RuntimeError`.
- Each mid-copy test asserts `hijack_reached["value"]` after calling `_get_cpp_coupled_impl()`, so any future signature drift fails loudly instead of passing vacuously.

**Verification performed:**
- Confirmed commit b6ce99e7 touches **only** `tests/test_cpp_coupled_backref_restore.py` (33 lines). Production code (`diffpair_routing.py`, commit 8505d065) is **unchanged** from the previously-approved shape — `from_routing_grid` in a `try`, restore in `finally`, `CppCoupledPathfinder(...)` a plain statement after, still covered by the pre-existing outer `try/except`.
- **Independent revert-check** in a fresh worktree (`kct build-native` → `C++ backend: available (version 1.0.0)`):
  - Fixed try/finally form → all 3 tests PASS.
  - Reverted to the plain-statement (pre-fix) form → the two mid-copy tests **FAIL** while the happy-path test still PASSES.
  - Inspected the failure: the `hijack_reached` guard **passes** (mock body runs past the hijack), and the failure is the real `_cpp_grid is not hijacked` assertion — the back-reference was left pointing at the partially-built coupled grid. This confirms the tests genuinely exercise the mid-copy gap and provide protection, not a signature `TypeError`.

## Test Execution

1. `kct build-native --check` → ✅ `C++ backend: available (version 1.0.0)`
2. `pytest tests/test_cpp_coupled_backref_restore.py -v` → ✅ 3 passed
3. Regression suites (diffpair parity/floor/integration) → ✅ covered by CI Diff-Pair Routing Regression (pass)
4. ruff check + format → ✅ CI Lint & Format (pass)

## CI Status

All 16 checks pass; merge state CLEAN. Notably the full Test suite (15m8s), Board 05 Routing Regression (19m39s), Diff-Pair (15m47s), and Match-Group (14m17s) regressions all green.

Solid work — the exception-path hijack regression is now correctly guarded, and the test failure semantics are proven load-bearing.

Issue #4079, opened 2026-07-13
router: reservation keep-out fences vias but not lateral traces — finish board-07 C++ reservation honoring (Path B deferred)


rjwalters:
**Parent epic:** #4049
**References:** #4071, PR #4078 (merged), #4053, #2983, #4080 (sibling: coupled joint-state attractor wiring, deferred separately)

## Problem

The C++ corridor-reservation port (#4071 / PR #4078) merged to `main` and took **"Path B"**: the two single-ended byte-lane reservation helpers in `src/kicad_tools/router/escape.py` pass `mirror_to_cpp=False` so their reservations stay **Python-only** and do **not** activate on the C++ backend for match-group boards (board 07):

- `reserve_inner_corner_lane_corridor` (#2983 inner-corner lane) — `escape.py:2526` `mirror_to_cpp=False` (call site; rationale comment at `escape.py:2513-2521`)
- the #4053 bundle-river via-hop helper — `escape.py:2722` `mirror_to_cpp=False` (call site; rationale comment at `escape.py:2712-2717`)

Both line numbers were re-verified against `origin/main` post-merge (commit `e590d863` / #4078) on 2026-07-12 and are unchanged from the original filing.

The in-code rationale (`escape.py:2513-2521`) states that honoring these on the C++ backend regresses board 07's fully-reversed DDR byte into **copper shorts (DM0<->DQ7, +2 more)**.

## Root cause (re-verified against `origin/main` post-#4078)

The reservation keep-out is enforced **only for vias**, not for lateral traces, on **both** the C++ grid/A* and the Python grid/A*:

### C++ side
- `src/kicad_tools/router/cpp/src/grid.cpp`: the reserved-net keep-out check lives only in `Grid3D::mark_via` (line 146, check at line 156: `if (check_reservations && cell.reserved_count > 0)`). `Grid3D::mark_segment` (line 84) has **no** reservation consult — it unconditionally paints `cell.blocked = true` for any net along the Bresenham line.
[...]

Issue #4085, opened 2026-07-13
router: length-matching by construction — slack-corridor reservation during routing instead of post-hoc serpentine


rjwalters:
**Source:** deep-research literature survey 2026-07-12 (adversarially verified findings). Parent context: epic #4049 Gap 2 (board 07's 4 `diffpair_length_skew` errors: `DQS_N` 11.034 mm, `MIPI_CLK_N` 2.973 mm, `MIPI_DAT1_N` 18.326 mm, `TMDS_D2_N` 3.783 mm vs 0.050/0.075 mm tolerances).

## Problem (measured, FM3)

Diff-pairs routed as independent single-ended nets accumulate 3–18 mm length skew. Skews this large cannot be serpentine-tuned after the fact — there is no room left near the pair by the time tuning runs. The literature treats this as a solved architectural question: **match during routing by reserving slack, don't repair after**.

## Verified techniques

1. **Slack-area reservation** (Ozdal & Wong, IEEE TCAD 2006, two papers): while routing each bus net, allocate extra routing resources (slack area) around nets that are coming in short; consume the reserved area for length extension afterward. Provably optimal for min-area/max-length constraints; near-optimal for exact bounds. Their companion **Lagrangian-relaxation framework** folds min-max length constraints directly into the routing cost and automatically balances the bundle (short nets get room-to-lengthen prioritized, long nets get length minimization). Caveat: their LR router reported >1 day on some benchmarks — adopt the *slack-reservation idea* with our negotiated-A* cost machinery, not their solver wholesale.

2. **Concurrent multi-bus corridor allocation with in-corridor detour insertion** (Jin-Tai Yan, IEEE TCPMT 2022, doi:10.1109/TCPMT.2022.3175069; single-bus precursor ISPD 2011): reserve a corridor per bundle, route flow-allocated shortest paths inside it, then insert detours *inside the reservation* to minimize max skew subject to length-matching constraints. Versus sequential bus-by-bus: +10.7% buses routed, max skew cut 14.25 → 1.38 grid units. Directly composable with our corridor-reservation machinery (#4071/#4079): the reservation IS the tuning budget.

## Original proposal sketch

- When a match-group/pair corridor is planned, widen its reservation by a computed slack budget (estimated skew bound from pin geometry).
- Add an LR-style per-net cost term: nets under target length get a bonus for consuming slack cells inside their own reservation; nets over target get plain shortest-path costs.
[...]

PR #4087, opened 2026-07-13
fix(router): lateral-trace keep-out + per-reservation soft/hard flag; board-07 byte-lane honoring stays Path B (#4079)

rjwalters:
## Summary

Adds the missing **lateral-trace corridor keep-out** (Python↔C++ parity) and a
**per-reservation HARD/SOFT flag** for match-group boards. Investigates whether
board 07's fully-reversed DDR byte-lane reservations can now be honored on the
C++ backend and concludes — with a decisive HARD-vs-SOFT-vs-Path-B experiment —
that they **cannot** be planarised by any corridor reservation, so the two
single-ended byte-lane helpers **stay Path B** (`mirror_to_cpp=False`). The
topological blocker is escalated to the architect on #4079.

## Root cause of the lateral-trace gap

Neither the A* blocking predicate nor the post-hoc grid painter consulted the
reservation map for lateral steps (only vias were fenced, via `mark_via`):

- `Pathfinder::is_trace_blocked` (C++) / `Router._is_trace_blocked` (Python) —
  the A* gate deciding whether a lateral step is legal *before* a trace is
  placed. **Primary defect site** — a `mark_segment`-only fix would let A* pick
  an illegal path and only catch it after the fact.
- `Grid3D::mark_segment` (C++) / `RoutingGrid._mark_segment` (Python) — the
  post-hoc painter blocked cells unconditionally (bookkeeping-parity gap with
  `mark_via`).

## Changes

**C++** (`src/kicad_tools/router/cpp`):
- New inline `Grid3D::is_reserved_excluding(x,y,layer,net)` primitive
  (`grid.hpp`) — true iff the cell is HARD-reserved for a net set that EXCLUDES
  `net`; fast-pathed behind `has_reservations_`; returns false for SOFT cells.
- `GridCell` gains `reserved_soft`; `reserve_cell(...)` gains a `soft` param.
- `mark_segment` (`grid.cpp`): skip cells HARD-reserved for a foreign net set.
- `Pathfinder::is_trace_blocked` (`pathfinder.cpp`): hard keep-out for foreign
  HARD-reserved cells in both the cached-kernel fast path and the radius-override
  slow path; hard even in negotiated/relief mode. `coupled_pathfinder.cpp`
  reuses this predicate (6 call sites), so the coupled diff-pair path is covered
  with no separate change.

**Python** (`src/kicad_tools/router`):
- `_is_trace_blocked` (`pathfinder.py`) and `_mark_segment` (`grid.py`): consult
  `_reserved_for_nets` + `_soft_reservations`, mirroring `_mark_via`. Fast-pathed
  behind `bool(self._reserved_for_nets)` so unreserved boards (00–05) stay
  byte-identical.
- `reserve_corridor_cells(..., soft=False)` and a `_soft_reservations` set track
  per-cell keep-out strength; `cpp_backend.from_routing_grid` mirrors the flag.

**escape.py**: the #2983 inner-corner and #4053 bundle-river byte-lane helpers
**keep `mirror_to_cpp=False` (Path B)**, now with `soft=True` and an in-code
record of the measured HARD/SOFT/Path-B results (see below). The now-accurate
rationale comments replace the old ones.

## Build version

Bindings-surface change (`GridCell.reserved_soft`, `reserve_cell` new `soft`
arg), so `ROUTER_CPP_BUILD_VERSION` and `_REQUIRED_CPP_BUILD_VERSION` bump
**16 → 17** in lockstep (`types.hpp` + `cpp_backend.py`).

## Decisive experiment (board 07, full CI recipe `--step all --seed 42`, out-of-process copper-LVS, run to route completion)

| Variant | Reach | Copper shorts | Opens |
|---|---|---|---|
| HARD keep-out (fence foreign traces) | 25/31 | **3** (DM0↔DQ7, DM0↔DQS_N, DQ7↔DQS_N) | 6 |
| SOFT attractor-only (`soft=True`, no fence) | 25/31 | **3** (identical) | 6 |
| Path B (Python-only, this branch's shipped config) | 26/31 | **0** | 5 (known #3438) |

**Honoring the single-ended byte-lane corridor on C++ AT ALL — hard fence OR
soft attractor — reproduces the DM0↔DQ7 short cluster.** The fully-reversed DDR
byte's crossing conflict graph is complete and the crossing must resolve at the
tight U1 pad fan-in (DM0 arrives high on In1.Cu ~y68, must reach its pad at
y63.8, crossing DQS_N/DQ7 in a ~5mm box at the pads → shorts at U1.29/31/35). No
corridor reservation can planarise a crossing that happens inside the pad
fan-in. Path B is clean.

## Verification

- **Parity/soft tests** (`tests/test_grid_cpp_parity.py`): new
  `TestLateralTraceReservationParity` (5) + `TestSoftReservationParity` (4),
  asserting Python==C++ for foreign-blocked / owner-passable / hard-in-relief /
  zero-overhead fast path / soft-passable. 13/13 grid parity pass; 62
  parity+diffpair reservation tests pass; ruff clean.
- **Board 07 fresh full route on the C++ backend (Path B config, NOT truncated)**:
  0 shorts, exactly the 5 known opens (DQ3, DQ4, MIPI_DAT0_N, TMDS_D0_N,
  TMDS_D1_N), 244/244 bound pads, `check_copper_lvs --expect-opens` exit 0.
- **CI**: all 17 gates green, incl. Board 07 E2E (LVS known opens), Match-Group
  Regression, Board 06 E2E + Diff-Pair Regression (the #2677 HARD-mirrored
  pair-continuation still works with the new lateral fence).

## Scope note

The lateral-trace keep-out + HARD/SOFT reservation primitive ship and are used
today by the #2677 planar pair-continuation corridor (board 06, mirrored to C++
as HARD). Full board-07 byte-lane honoring on C++ is a **topological blocker**
that corridor reservations cannot solve; it is escalated on #4079 for architect
re-scoping (pad-aware fan-in planner / re-shaped byte-lane geometry). This PR
does not close that goal — it delivers the substrate + evidence + escalation.

Updates #4079

### DIFF
--- a/src/kicad_tools/router/cpp/include/grid.hpp
+++ b/src/kicad_tools/router/cpp/include/grid.hpp
@@ -117,8 +117,14 @@ public:
     // Overlapping reservations REPLACE (last-writer-wins), matching the
     // Python semantics.  A single reserved cell flips the grid-wide
     // ``has_reservations_`` fast-path flag.
+    // Issue #4079: ``soft`` selects the keep-out strength.  ``false``
+    // (default) is a HARD reservation -- foreign vias AND lateral traces
+    // are fenced out (``is_reserved_excluding`` true).  ``true`` is a SOFT
+    // reservation -- only the A* attractor bonus applies for the owner
+    // net; foreign traces may still cross (``is_reserved_excluding``
+    // false).  See the #4079 note in ``types.hpp``.
     void reserve_cell(int x, int y, int layer,
-                      const std::vector<int>& net_ids);
+                      const std::vector<int>& net_ids, bool soft = false);
 
     // Clear every corridor reservation (owner sets + fast-path flag).
     void clear_reservations();
@@ -155,10 +161,14 @@ public:
         if (!is_valid(x, y, layer)) return false;
         const auto& cell = at(x, y, layer);
         if (cell.reserved_count <= 0) return false;
+        // Issue #4079: a SOFT reservation never fences foreign traffic out
+        // -- it contributes only the A* attractor bonus for its owner net.
+        // Only HARD reservations produce a foreign keep-out.
+        if (cell.reserved_soft) return false;
         for (int i = 0; i < cell.reserved_count; ++i) {
             if (cell.reserved_nets[i] == net) return false;  // owner net
         }
-        return true;  // reserved, but not for this net
+        return true;  // hard-reserved, but not for this net
     }
 
     // True iff ANY cell is currently reserved (grid-wide fast path).
--- a/src/kicad_tools/router/cpp/include/grid.hpp
+++ b/src/kicad_tools/router/cpp/include/grid.hpp
@@ -74,7 +74,18 @@ public:
     void mark_rect_blocked(int x1, int y1, int x2, int y2, int layer, int net,
                            bool is_obstacle = false);
 
-    // Route marking with clearance buffer using Bresenham
+    // Route marking with clearance buffer using Bresenham.
+    // Issue #4079: like ``mark_via``, ``mark_segment`` NOW consults the
+    // per-cell corridor reservation owner set (``GridCell::reserved_nets``)
+    // for lateral-trace keep-out, mirroring Python ``_mark_segment``.  A
+    // cell reserved for a net set that EXCLUDES ``net`` is SKIPPED (the
+    // foreign net's trace does not claim/block it), so a foreign lateral
+    // trace cannot colonise a reserved continuation corridor.  Cells
+    // reserved for a set that INCLUDES ``net`` are ordinary blockable
+    // cells.  Fast path: when ``has_reservations_ == false`` the check is
+    // skipped, preserving byte-identical behaviour on boards without
+    // reservations.  See ``cpp/src/grid.cpp`` and
+    // ``tests/test_grid_cpp_parity.py``.
     void mark_segment(int x1, int y1, int x2, int y2, int layer, int net,
                       int clearance_cells);
     // Issue #4071: ``mark_via`` NOW consults the per-cell corridor
@@ -127,6 +138,27 @@ public:
             if (cell.reserved_nets[i] == net) return true;
         }
         return false;
+    }
+
+    // Issue #4079: True iff the cell is reserved for a net set that
+    // EXCLUDES ``net`` -- i.e. a foreign net whose lateral trace (or via
+    // halo) must be fenced out of the reserved corridor cell.  Returns
+    // false for an unreserved cell OR a cell reserved for a set that
+    // INCLUDES ``net`` (the owner may use its own reservation).  This is
+    // the lateral-trace keep-out primitive consulted by
+    // ``mark_segment`` and ``Pathfinder::is_trace_blocked`` (parity with
+    // the Python ``_reserved_for_nets`` "owners is not None and net not
+    // in owners" check in ``_mark_via`` / ``_mark_segment`` /
+    // ``_is_trace_blocked``).  Inline for the A* hot path.
+    inline bool is_reserved_excluding(int x, int y, int layer, int net) const {
+        if (!has_reservations_) return false;
+        if (!is_valid(x, y, layer)) return false;
+        const auto& cell = at(x, y, layer);
+        if (cell.reserved_count <= 0) return false;
+        for (int i = 0; i < cell.reserved_count; ++i) {
+            if (cell.reserved_nets[i] == net) return false;  // owner net
+        }
+        return true;  // reserved, but not for this net
     }
 
     // True iff ANY cell is currently reserved (grid-wide fast path).
--- a/src/kicad_tools/router/cpp/include/types.hpp
+++ b/src/kicad_tools/router/cpp/include/types.hpp
@@ -161,7 +161,23 @@ namespace router {
 // first time.  Old .so files lack the ``reserved_*`` GridCell fields and
 // the ``cost_corridor_attractor`` DesignRules field; the version bump
 // forces a rebuild via ``kct build-native``.
-constexpr int ROUTER_CPP_BUILD_VERSION = 16;
+// Version 17 (Issue #4079): per-reservation HARD vs SOFT keep-out.
+// ``GridCell`` gains a ``reserved_soft`` flag and ``reserve_cell`` gains a
+// ``soft`` parameter (bindings-surface change -> version bump).  A HARD
+// reservation (default, e.g. #2677 pair-continuation) fences foreign vias
+// AND foreign lateral traces out of the corridor
+// (``is_reserved_excluding`` true -> ``mark_via`` / ``mark_segment`` /
+// ``is_trace_blocked`` skip/block).  A SOFT reservation (the #2983
+// inner-corner / #4053 bundle-river byte-lane helpers) keeps ONLY the A*
+// attractor bonus for the owning net -- it does NOT fence foreign traces.
+// This is required for board 07's fully-reversed DDR byte, whose crossing
+// conflict graph is COMPLETE: a hard lateral fence forces the crossing
+// foreign nets to route AROUND the corridor, colliding elsewhere (copper
+// shorts) or failing (extra opens).  The soft attractor pulls the owning
+// crossing net onto its inner-layer channel while leaving the mandatory
+// crossings legal.  Old .so files lack ``reserved_soft`` and the new
+// ``reserve_cell`` signature; the version bump forces a rebuild.
+constexpr int ROUTER_CPP_BUILD_VERSION = 17;
 
 // Issue #4071: fixed-capacity owner-set size for per-cell corridor
 // reservations.  Observed owner sets in practice are tiny: 1 for the
@@ -201,6 +217,14 @@ struct GridCell {
     // (last-writer-wins), matching Python ``reserve_corridor_cells``.
     int8_t reserved_count = 0;
     int32_t reserved_nets[RESERVED_NETS_CAP] = {0, 0, 0, 0};
+    // Issue #4079: HARD (false, default) vs SOFT (true) reservation.  A
+    // HARD reservation fences foreign vias AND lateral traces out of the
+    // cell (``is_reserved_excluding`` returns true for a foreign net).  A
+    // SOFT reservation keeps ONLY the A* attractor bonus (``is_reserved_for``
+    // for the owning net) -- ``is_reserved_excluding`` returns false, so
+    // foreign traces may still cross.  Meaningful only when
+    // ``reserved_count > 0``.  See ``ROUTER_CPP_BUILD_VERSION`` note.
+    bool reserved_soft = false;
 };
 
 // A* node for priority queue
--- a/src/kicad_tools/router/cpp/src/bindings.cpp
+++ b/src/kicad_tools/router/cpp/src/bindings.cpp
@@ -197,8 +197,10 @@ NB_MODULE(router_cpp, m) {
         // ``is_reserved_for``, ``clear_corridor_reservations``, and
         // ``reserved_cell_count``.  ``mark_via`` and the A* cost loop
         // honour these reservations (keep-out + attractor).
+        // Issue #4079: ``soft`` (default False) selects HARD (via + lateral
+        // keep-out) vs SOFT (attractor-only) reservation strength.
         .def("reserve_cell", &Grid3D::reserve_cell,
-             "x"_a, "y"_a, "layer"_a, "net_ids"_a)
+             "x"_a, "y"_a, "layer"_a, "net_ids"_a, "soft"_a = false)
         .def("clear_reservations", &Grid3D::clear_reservations)
         .def("reserved_cell_count", &Grid3D::reserved_cell_count)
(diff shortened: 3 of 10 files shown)


rjwalters:
## Changes Requested — the shorts are NOT gone (CI reproduces DM0↔DQ7 unloaded)

I reviewed the code, built the C++ backend, ran the parity suite locally, and waited for the full CI run. The code quality is high and the parity tests are excellent — but the **load-bearing acceptance criterion is unmet**, and CI proves it.

### Decisive evidence: Board 07 End-to-End (LVS known opens) — FAIL

The strict out-of-process copper-LVS re-check on the seed-42 recipe regen reports:

```
expected ONLY kind='open' mismatches but got kinds ['open', 'short']:
10 mismatch(es) (3 short, 7 open);
first: short DM0(U1.29) <-> DQ7(U1.35)
-- a copper short is a hard regression regardless of the known-opens allowance
```

This is **exactly the DM0↔DQ7 short (+2 more)** that forced Path B in PR #4078. The PR's central claim — "Board 07 fresh route on the C++ backend with reservations honored: ZERO copper shorts — no DM0↔DQ7 or any short" — does not hold. CI runs **unloaded** and reproduces the shorts on the standard recipe. The lateral-trace keep-out, as implemented, does not prevent the fully-reversed DDR byte from shorting once the route completes.

### Why the local run looked clean (and why the load explanation is wrong)

The builder saw "0 shorts" locally because the fresh route hit the `--timeout 600` backstop at iter 11 and returned a *partial* route (25/31) before the DDR byte finished — so the crossing that produces the short hadn't been placed yet. CI routes the DDR byte to completion and the DM0↔DQ7 short materializes. This is a false negative in the local verification, not a host-load artifact. The reach drop (28→25) and the shorts are two faces of the same incomplete/unsafe route, not independent.

### The Match-Group pass is not sufficient

Note the split: **Match-Group Routing Regression passed** but **Board 07 E2E failed**. The regression job gates on reach/DRC thresholds and does not run the `--expect-opens` copper-LVS comparator, so it tolerates shorts; the E2E job's strict comparator is the one the issue's acceptance criteria named ("no copper shorts (DM0↔DQ7 or otherwise)"). Per judge.md, an unloaded CI board gate outweighs a loaded local claim — and here CI is unambiguous.

### What passed (for the record)

- Parity tests: 9/9 including all 5 new `TestLateralTraceReservationParity` (built `kct build-native --force`, `required==actual==16`).
- Semantics verified correct: `is_reserved_excluding` (owner passable / foreign blocked), both C++ fast+slow `is_trace_blocked` paths, `mark_segment`, and Python `_is_trace_blocked`/`_mark_segment` all agree; hard-gate-in-relief-mode verified. Fast paths preserve byte-identical behavior on unreserved boards.
- Boards 00–06 CI gates all green; diff contains only the 7 intended files (no board artifacts). The mechanism is sound in isolation.

### The actual problem

The keep-out primitives are correct, but re-enabling `mirror_to_cpp=True` at the two byte-lane helpers with *only* this keep-out does not solve board 07. Adding a hard keep-out that A* honors means the foreign DDR nets can no longer cross the reserved corridor — but they are still being routed into a topology where they *must* cross something, so instead of routing cleanly they either fail to route (the extra opens: 7 vs the expected 5) OR find a different illegal crossing (the 3 shorts). The reservation set as currently generated for the fully-reversed byte over-constrains the corridor: honoring it produces shorts+opens rather than a clean route. This is exactly the regression Path B was protecting against, and the lateral keep-out alone did not remove the root cause.

### Required before re-review

1. The Board 07 End-to-End gate must be **green** — zero copper shorts, exactly the 5 known opens (`DQ3,DQ4,MIPI_DAT0_N,TMDS_D0_N,TMDS_D1_N`), no new opens. This is non-negotiable; it is the acceptance criterion.
2. Verify board 07 on the **unloaded CI recipe** (`generate_design.py /tmp/... --step all --seed 42`), not a timeout-truncated local run. Do not attribute the shorts to host load — CI reproduces them unloaded.
3. If the fully-reversed DDR byte genuinely cannot route with these reservations honored, the honest outcome may be that Path B stands and this issue needs a different approach (e.g., relax/re-shape the byte-lane reservation geometry so the corridor is satisfiable, or keep `mirror_to_cpp=False` for the reversed-byte case specifically). Escalate to the architect if the reservation topology itself is the blocker rather than the keep-out mechanism.

The keep-out code and parity tests are good work and should be preserved — the gap is that enabling the mirror does not yet yield a shorts-free board 07.

rjwalters:
## Feedback addressed — Board 07 E2E gate now GREEN (verified on the full CI recipe)

Thank you for the decisive review. You were right that the shorts were real and that the local "0 shorts" was suspect. I reproduced the failure exactly, then ran a controlled hard/soft/Path-B experiment on the **full, non-truncated** CI recipe. The evidence changed the fix.

### Reproduction of your failure (current PR HEAD, hard fence, mirror=True)

Full recipe `generate_design.py /tmp/... --step all --seed 42`, then the exact `check_copper_lvs.py --expect-opens` comparator:

```
3 short, 6 open; first: short DM0(U1.29) <-> DQ7(U1.35)
```

Confirmed — DM0↔DQ7, DM0↔DQS_N, DQ7↔DQS_N at U1.29/31/35, + TMDS_D1_P extra open. Matches your report.

### Controlled experiment (all on the full recipe, all run to route completion / negotiated convergence, NOT the `--timeout` backstop)

| Variant | Reach | Shorts | Opens | E2E gate |
|---|---|---|---|---|
| **HARD** keep-out (mirror + lateral fence) | 25/31 | **3** | 6 | FAIL |
| **SOFT** attractor-only (mirror, no fence, new `soft=True`) | 25/31 | **3** | 6 | FAIL |
| **Path B** (Python-only, no C++ honouring) | **26/31** | **0** | **5** | **PASS** |

The soft attractor produced an **identical** short set to the hard fence — so the shorts are **not** caused by the hard fence forcing crossings around (my original claim). The geometry shows why: the reversed byte's crossing conflict graph is complete and the crossing **must resolve at the tight U1 pad fan-in** (DM0 arrives high on In1.Cu ~y68 and must reach its pad at y63.8, crossing DQS_N/DQ7 in a ~5mm box at the pads). No corridor reservation — hard fence or soft attractor — can planarise that. Honouring the byte-lane corridors on C++ **at all** over-constrains it, and even costs a net of reach (26→25).

### Resolution — mapping each of your required points

1. **"Board 07 E2E gate must be green — 0 shorts, exactly the 5 known opens, no new opens."** Done. The two byte-lane helpers revert to `mirror_to_cpp=False` (Path B). Final full-recipe run: **0 shorts, exactly DQ3, DQ4, MIPI_DAT0_N, TMDS_D0_N, TMDS_D1_N, 244/244 bound pads (non-vacuous).** The `check_copper_lvs.py --expect-opens` comparator exits 0; the out-of-process DRC re-check exits 0 (within allowlist).

2. **"Verify on the unloaded CI recipe, not a timeout-truncated local run."** Every table row above is the full recipe; Path B converged at iter 10 (negotiated loop finished — **not** timeout-stopped), so the clean result is genuine, not a truncation artifact.

3. **"If the reversed byte can't route with these honoured, Path B may stand / escalate to the architect."** This is the outcome the evidence supports. Path B stands for the byte-lane helpers. I've left an in-code record of the measured hard/soft/Path-B results and the pad-fan-in root cause, and will escalate the topological blocker to the architect on #4079 for re-scoping (re-shaped byte-lane geometry or a pad-aware fan-in planner).

### What still ships (and is used — not dead infrastructure)

- The **lateral-trace keep-out** you judged clean (`Grid3D::mark_segment`, both `Pathfinder::is_trace_blocked` paths, Python `_mark_segment`/`_is_trace_blocked`) is preserved.
- **NEW: a per-reservation HARD vs SOFT flag** (`GridCell::reserved_soft`, `reserve_cell(..., soft)`, `is_reserved_excluding` soft-skip; Python `_soft_reservations` threaded through `reserve_corridor_cells`, the keep-out consumers, and the C++ bulk/incremental mirror). HARD fences foreign vias + lateral traces; SOFT keeps only the attractor.
- The **#2677 pair-continuation corridor (board 06, planar)** mirrors to C++ as a HARD reservation and now benefits from the lateral fence. This is the fence primitive #4085/#4086 depend on.
- Bindings surface changed (`reserve_cell` gained `soft`) → `ROUTER_CPP_BUILD_VERSION` / `_REQUIRED_CPP_BUILD_VERSION` **16 → 17**.

### Verification summary

- Board 07 E2E: **0 shorts, exactly 5 known opens** (copper-LVS comparator exit 0), DRC re-check exit 0.
- Grid parity: **13/13** (9 lateral-trace + 4 new soft-reservation, asserting Python == C++ for the soft no-keep-out / still-attracts / last-writer-wins contracts).
- 62 parity + diff-pair reservation tests pass; ruff clean; no new mypy errors on changed lines.
- Diff is 9 files (7 original + no new files); no board artifacts.

rjwalters:
✅ **Approved!** (Doctor→Judge re-review; prior rejection cleared.)

## Why my prior rejection is resolved

I previously rejected because CI **Board 07 E2E** reproduced the DM0↔DQ7 (+2) copper shorts on the full unloaded route — the builder's zero-shorts claim came from a timeout-truncated run. The Doctor's fix does **not** try to re-enable C++ honoring of the byte-lane reservations (which is what shorted); it ships the keep-out mechanism + a per-reservation HARD/SOFT flag, keeps the two single-ended byte-lane helpers on **Path B** (`mirror_to_cpp=False`, now `soft=True`), and documents the topological blocker. Board 07 E2E now passes on CI on the shipped HEAD — the shorts-free configuration.

## CI board-gate results (all 17 green, merge state CLEAN)

- **Board 07 End-to-End (LVS known opens)** — pass (14m30s), NOT truncated
- **Match-Group Routing Regression (boards/07)** — pass (14m10s)
- **Board 06 End-to-End (LVS)** — pass (16m45s) · **Diff-Pair Regression** — pass (15m8s) → confirms the #2677 HARD-mirrored pair-continuation still routes with the new lateral fence
- **Board 05 Routing Regression** — pass (19m11s) · **Test** — pass (15m24s) · Boards 00–04, C++ Build, Type Check, Lint, DRC, kicad-cli smoke — all pass

## Doctor response verified point-by-point

1. **Decisive experiment** — the in-code record + #4079 escalation table both show HARD keep-out and SOFT attractor-only produce the *identical* 25/31, 3-short, 6-open failure; Path B is 26/31, 0 shorts, 5 known opens. The crossing resolves at the U1 pad fan-in, which no corridor reservation can planarise. Confirmed.
2. **Mechanism shipped** — `GridCell::reserved_soft` + `reserve_cell(...soft)` + inline `is_reserved_excluding` (SOFT returns false); Python `_soft_reservations` threaded through `_mark_via`/`_mark_segment`/`_is_trace_blocked`/`reserve_corridor_cells`; `cpp_backend` mirrors the flag. **Build version 16→17 bumped in lockstep** on both sides (verified via `git show` on the branch: `types.hpp` and `cpp_backend.py` both 17; main is 16). `coupled_pathfinder.cpp` reuses `is_trace_blocked` (6 call sites) so the coupled path is covered with no separate change. Verified.
3. **Scope reduction is honest** — byte-lane call sites stay Path B with an accurate in-code rationale; the escalation comment is posted on #4079 with the full evidence table; PR retitled to match reduced scope.
4. **Full route (not truncated)** — CI Board 07 E2E is the non-truncated proof: 0 shorts, exactly the 5 known opens. Verified.
5. **Tests** — 13/13 grid parity (incl. 4 new soft + 5 lateral) and the soft/hard semantics tests pass; #2677 stays HARD; ruff clean. Verified.

## Soft-flag semantics check

- SOFT = attractor-visible, non-blocking: `is_reserved_excluding` short-circuits false on `reserved_soft`; Python skips the keep-out when `key in _soft_reservations`. Both grids agree (parity tests assert cpp==py for owner-passable, foreign-passable-under-soft, foreign-blocked-under-hard, hard-in-relief). The `_is_trace_blocked` disc-iteration bounds are consistent with the `within_disc` mask (no off-by-one).
- HARD (default) blocks foreign vias **and** lateral traces, as before + the new lateral fence. Board 06 #2677 pair-continuation stays HARD and mirrored. Verified.

## Closes-vs-Updates ruling

The issue's headline goal — *"board 07 gate green WITH reservations honored on C++"* — is **not** met: the byte-lane reservations are proven topologically un-honorable on C++ and remain Path B. The Doctor correctly shipped the substrate + evidence + architect escalation instead. Because the headline goal is deferred (not achieved), I **edited the PR body**:
- Rewrote the stale "Verification" section, which still claimed "ZERO copper shorts / shorts no longer occur" on C++ honoring — that is the exact false claim my prior rejection was based on and it contradicts the shipped Path-B code. Replaced with the honest HARD/SOFT/Path-B experiment table.
- Changed **`Closes #4079` → `Updates #4079`** so the issue stays OPEN for the architect re-scope (pad-aware fan-in planner / re-shaped byte-lane geometry). The escalation comment on #4079 is the tracking anchor.

This is a valid, honestly-scoped completion: the machinery is real and used (board 06), dependents #4085/#4086 are unblocked by the shipped keep-out + flag, and the blocker is documented for the architect. Approving.

Issue #4051, opened 2026-07-12
Board 07 Phase 1b: bundle-aware innermost-first net ordering for facing pad-array columns


rjwalters:
**Epic:** #4049 (Phase 1b) · Related: #3438, #2962, #3101

## Context (verified on `main`)
`Autorouter._apply_byte_lane_inner_priority`
(`src/kicad_tools/router/core.py:4700`) already exists but is
**scaffolding-only — it returns `net_order` unchanged** (identity). Its
docstring documents the failure mode precisely (#2962): on board 07's DDR
byte between mirrored QFN-48 packages, the inner-corner pads (DQ1/DQ6) get
squeezed out because same-row lateral lanes are consumed by the time inner
nets attempt escape. It is invoked from three sites
(`core.py:5348`, `core.py:6860`, and via `two_phase.py:313`).

This is the integration surface for a **bundle-aware innermost-first
ordering** heuristic: on facing pad-array columns where all nets must
escape onto one F.Cu strip, route the innermost nets first so they claim
lanes before outer nets fill them.

## Task
Extend `_apply_byte_lane_inner_priority` from identity return to an actual
innermost-first reordering for collated facing-column bundles:
- Detect the mirrored byte-lane match group (detection logic is already
  scaffolded here).
- Reorder so inner-corner nets escape before outer neighbours.
- Gate behind a flag or auto-detect (prefer auto-detect for collated
  bundles; a flag is acceptable if auto-detection is unsafe).

Heed the PR #2969 design history in the docstring: naive corner demotion
produced 27/31 with 86 DRC errors. The winning move is ordering, not
detour geometry.

## Acceptance
- Ordering heuristic is active for collated bundles (flag or auto-detect).
- DDR-bundle repro from #3438 measured with the heuristic on; result
  posted here.
- **No cross-board regression**: boards 00–06 route gates stay green.

Sweepable (no `loom:blocked`).

rjwalters:
## Curator verification + gap-filling

Spot-checked `_apply_byte_lane_inner_priority` (`src/kicad_tools/router/core.py:4700`) and its three call sites. Confirmed the call graph as stated: `route_all` (`core.py:5348`), `route_all_negotiated` (`core.py:6860`), and `TwoPhaseRouter` via `two_phase.py:313`. Two corrections and a scoping recommendation follow.

### Correction 1: the method is not pure identity/scaffolding today

[...]

PR #4089, opened 2026-07-13
feat(router): monotonic feasibility certificate + constructive escape ordering (SER Phase 1)

### LINKED ISSUE #4084: router: joint bundle-escape pre-stage (SER) — monotonic feasibility certificate + coordinated ordering, replacing greedy per-net escape
**Source:** deep-research literature survey 2026-07-12 (adversarially verified findings). Parent context: epic #4049 Gap 1, #3438 negative controls, closed river-routing epic #3673.

## Problem (measured, FM1)

The 11-net DDR byte bundle between facing QFN-48 pin columns (0.8 mm pitch, 30 mm apart) fails 2/11 **even alone on an empty 4-layer board** (#3438). Greedy per-net ordering seals escape corridors; N-1 invariance proven. This is precisely the **Simultaneous Escape Routing (SER)** problem class (Ozdal & Wong, ICCAD 2004 onward), and the literature explicitly states greedy/sequential heuristics fail to achieve full routability on it — our failure is the textbook failure.

## Verified techniques, cheapest first

1. **Monotonic feasibility certificate + constructive ordering** (Tomioka & Takahashi, ASP-DAC 2006, IEEE 1594758): a *necessary and sufficient* condition for all nets to be connectable by monotonic routes when each net has one pin on each of two **parallel boundaries** — exactly our facing-columns topology — plus a constructive routing method. Adoption: check the condition on the bundle's pin sequence *before any A\* search*; when it holds, constrain per-net A\* to monotone corridors in the derived order. Cheap (pure Python pre-check), no A\* core change.

2. **B-escape boundary routing with dynamic reordering** (Luo, Yan, Ma, Wong, Shibuya; ISPD 2010 Best Paper, doi:10.1145/1735023.1735033): route nets along the frontier of the already-routed region, dynamically reorder + backtrack instead of fixed order. Published result: 100% of industrial escape benchmarks solved where Cadence Allegro completed 50%. Adoption: an outer-loop escape phase (ordering/backtrack driver) around existing per-net search — no A\* core change.

3. **Min-cost multi-commodity flow over inter-pad tiles** (Jiao & Dong, ASP-DAC 2016, IEEE 7428042; successor line MORE-Router+ ISQED 2024): tile graph between adjacent pads, three transformations (non-crossing / ordering / capacity), solved as MMCF/ILP; self-reported 100% routability with better wirelength+CPU than prior OER work. Maps most naturally onto our grid lattice: a pre-stage that hands corridor assignments to detailed A\* via the existing corridor-reservation machinery (#4071/#4079). For an 11-net bundle the ILP is tiny; SAT alternatives (Luo & Wong ASP-DAC 2008; MonoSAT ICCAD 2016 scales to 2000+ pins) also comfortably in range.

4. **Per-corridor exact bundle routing** (Pinter, "River Routing: Methodology and Analysis", 1983): detailed planar routing in a hole-free corridor with given net order is **polynomial** even though whole-board planar routing is NP-complete. Justifies the architecture: carve corridors globally, solve each corridor's bundle exactly.

## Proposed phasing (epic-level; see Scope below for what THIS issue covers)

- Phase 1: monotonic certificate + ordering (technique 1) wired into the existing escape-fan planner; measure DDR-bundle reach delta on board 07.
- Phase 2: B-escape-style dynamic-reordering outer loop for bundles that fail the certificate.
- Phase 3 (only if needed): MMCF/SAT tile formulation as the joint allocator feeding corridor reservations.

## Caveats

All quantitative results are author-self-reported on academic benchmarks (B-escape vs circa-2010 Allegro). Monotonic/river theory assumes single-layer, obstacle-free, two-point nets — real corridors with vias fall outside the exact case; use as certificate + prior, not as a drop-in router.

---

### Scope of THIS issue: Phase 1 ONLY

This issue covers **only Phase 1** (monotonic feasibility certificate + constructive
ordering). Phase 2 (B-escape dynamic reordering) and Phase 3 (MMCF/SAT tile
formulation) are explicitly **out of scope** and must be filed as separate
follow-up issues once Phase 1's measured result is in hand. Do not implement
any part of Phase 2/3 in this PR — if the certificate-holds path needs a
fallback for the non-monotonic case, that fallback is "leave `net_order`
unchanged for this group" (identity), not a partial B-escape implementation.

- [ ] Phase 1 (this issue): certificate + constructive ordering, pure Python, no C++ changes.
- [ ] Phase 2 (follow-up, NOT this issue): dynamic reordering outer loop for bundles that fail the certificate.
- [ ] Phase 3 (follow-up, NOT this issue): MMCF/SAT tile formulation.

### Code seams on `origin/main` (verified against `a3845f3a`)

**This exact ordering seam was already built once and is currently gated off.**
Issue #4051 (Phase 1b of epic #4049) added a *reactive* escape-freedom
scheduler at exactly the integration point Phase 1 needs — read it before
writing new code, both to reuse the wiring and to understand why the
naive heuristics failed:

- `Autorouter._apply_byte_lane_inner_priority` — `src/kicad_tools/router/core.py:4712`.
  Detects mirrored byte-lane match groups (facing pad columns), reserves
  inner-corner corridors, and (when `enable_byte_lane_reorder=True`, default
  `False`) permutes the group's `net_order` slots via
  `_reactive_escape_freedom_order` → `_schedule_by_escape_freedom`
  (`core.py:5189`, `core.py:5397` staticmethod). **This is the ordering
  seam Phase 1 must extend or replace for byte-lane groups**: add a
  certificate check before scheduling, and when it holds, derive the
  constructive order per Tomioka & Takahashi instead of (or in addition
  to) the existing reactive heuristic.
- `enable_byte_lane_reorder` flag — `core.py:869-878` (constructor). Currently
  `False` because both the reactive scheduler (#4051) AND three earlier
  *static* permutations (monotone, reverse, outside-in — see docstring at
  `core.py:5068-5078` and `core.py:5117-5125`) were measured on the exact
  11-net DDR bundle and **none reached 11/11** (best: 10/11, 10/11, 8/11
  static; 9/11 reactive — a regression from the 10/11 identity baseline).
  This is strong evidence the certificate-driven approach is the untried
  lever, not a fourth guess at a heuristic permutation.
- Three call sites apply the same helper — `core.py:5605` (`route_all`),
  `core.py:7119` (`route_all_negotiated`), and via keyword injection into
  `TwoPhaseRouter` at `core.py:11104` (consumed by
  `src/kicad_tools/router/algorithms/two_phase.py`, the `kct route`
  default path through `route_with_escape`). Phase 1's ordering change
  only needs to touch `_apply_byte_lane_inner_priority` / the two helper
  methods it calls — all three call sites pick it up for free.
- `route_all_negotiated` entry point — `core.py:6672` (not `~6415`; the
  file has grown since the epic body was drafted — re-verify line numbers
  at implementation time. The `_apply_byte_lane_inner_priority` call
  inside it is at `core.py:7119`).

**Corridor / monotone-constraint attachment point** (verified — this is the
mechanism the builder should use, no C++ change required):

- `EscapeRouter.reserve_inner_corner_lane_corridor` —
  `src/kicad_tools/router/escape.py:2349`. Single-ended corridor
  reservation primitive (Issue #2983) already wired into
  `_apply_byte_lane_inner_priority` and called once per qualifying
  inner-corner pad before any corner-net escape (`core.py:5042`). This
  already rides `RoutingGrid.reserve_corridor_cells`
  (`src/kicad_tools/router/grid.py:4166`) and
  `RoutingGrid.get_corridor_attractor_bonus` (`grid.py:4247`) — the A*
  cost function treats reserved-corridor cells as attractive for the
  matching net and neutral/blocking for others. **This is the recommended
  mechanism for the "monotone corridor constraint": express each net's
  monotone progress region (the swept area between its two boundary pins
  under the derived order) as a corridor reservation via the existing
  `reserve_corridor_cells` / attractor-bonus path**, rather than adding a
  new cost/blocking hook to the A* core. This keeps Phase 1 entirely in
  Python pre/post-processing around the existing negotiated router and
  avoids touching `router_cpp.*.so` or its `CMakeLists.txt` sources.
- If per-net monotone corridors turn out to be too coarse (verify during
  implementation — corridor half-width defaults are tuned for the
  inner-corner case, not a full-bundle monotone sweep), the fallback is a
  lighter-weight per-net directional bias (skip literal corridor cell
  reservation, only bias the A* cost through
  `RoutingGrid.get_negotiated_cost`, `grid.py:5137`) — document which one
  was used and why in the PR description; do not add a third, undocumented
  mechanism.

### Certificate implementation location

No existing module implements the Tomioka & Takahashi condition. Add it as
a new pure-Python module, e.g. `src/kicad_tools/router/monotone_certificate.py`
(builder's choice of exact name/location, but keep it under
`src/kicad_tools/router/` alongside `escape.py`/`core.py` since both will
import it). Suggested surface:

```python
def check_monotone_feasibility(
    pin_sequence_a: list[int],  # net ids in pin order along boundary A
    pin_sequence_b: list[int],  # net ids in pin order along boundary B (same net ids)
) -> bool: ...

def constructive_monotone_order(
    pin_sequence_a: list[int],
    pin_sequence_b: list[int],
) -> list[int]: ...  # raises/returns None if check_monotone_feasibility is False
```

Unit-test this module in isolation (pure combinatorics, no grid/pad
dependencies) before wiring it into `core.py` — this makes the certificate
logic independently reviewable/verifiable against the paper's stated
necessary-and-sufficient condition, separate from any routing-side
regression.

### DDR-bundle test fixture

- `boards/07-matchgroup-test/` — `generate_pcb.py` declares
  `DDR_DATA_BYTE_0_SINGLES` / `DDR_DATA_BYTE_0_PAIRS` (9 singles + 1 diff
  pair = 11 nets total), the exact bundle from #3438's isolation matrix.
  `generate_design.py` drives `--step route` through the production
  negotiated recipe.
- Existing unit tests for the ordering seam (reuse as templates for Phase 1
  tests, do not duplicate their exact assertions):
  - `tests/router/test_byte_lane_priority.py` — tests
    `_apply_byte_lane_inner_priority` / `_schedule_by_escape_freedom`
    directly on synthetic net-id lists (fast, no grid). This is the right
    pattern for certificate-classification unit tests.
  - `tests/router/test_byte_lane_corridor_reservation.py` — tests the
    corridor-reservation side effect in isolation.
  - `tests/test_board_07_matchgroup_test.py` — board-level artifact
    regression (match-group declarations, sidecar, tolerances). Does NOT
    currently assert bundle reach/DRC count — if Phase 1 adds a reach
    assertion, add it here or in a new `tests/router/test_monotone_*.py`
    file, gated for wall-clock cost (the #3438 isolation run took ~174s
    for the 11-net-alone case).
- There is no standalone isolation-repro script in the repo today for the
  "11-net DDR bundle alone on an empty board" scenario described in
  #3438's table — that was evidently run ad hoc. The builder should add
  one (or a `pytest.mark.slow`-gated test) so the reach measurement is
  reproducible and re-runnable in CI/regression sweeps, not just quoted
  in issue prose.

### Scope guardrails

- **Pure Python, no C++ changes.** The certificate and constructive
  ordering are a pre-stage that runs before any A* search; per-net
  monotone constraints ride the existing `reserve_corridor_cells` /
  attractor-bonus machinery (or, as fallback, `get_negotiated_cost`
  biasing) — confirm in the PR description which mechanism was used.
  `router_cpp.*.so` and `src/kicad_tools/router/cpp/**` should not need
  any edits; if implementation reveals otherwise, stop and flag it rather
  than proceeding — that would be a scope change requiring re-curation.
- **Gate behind a flag, default OFF**, consistent with
  `enable_byte_lane_reorder`'s existing precedent (three prior heuristics
  all regressed the DDR bundle before landing gated-off). Either reuse
  `enable_byte_lane_reorder` (if the certificate-driven order can coexist
  cleanly with/replace the reactive scheduler under the same flag) or add
  a new flag (e.g. `enable_monotone_certificate_order`) — builder's
  judgment, but document the choice and its interaction with the existing
  flag in the PR description.
- Applies only to bundles matching the existing detection precondition
  (mirrored byte-lane / facing-pad-column groups, `MIN_BYTE_LANE_SIZE`+
  members on one co-located row) — do not broaden detection scope in this
  issue.

## Acceptance Criteria

- [ ] `check_monotone_feasibility` correctly classifies the #3438 11-net DDR
      bundle fixture (board 07 `DDR_DATA_BYTE_0` group) — record whether the
      certificate holds or not for this specific bundle's pin sequence, and
      include the classification result (with reasoning) in the PR
      description. (The literature predicts the facing-columns topology
      *should* satisfy the two-parallel-boundaries precondition, but the
      pin ordering — DQS pair interleaved among DQ0-7/DM0 — determines
      whether the sequence itself is monotone-feasible; this must be
      verified empirically, not assumed.)
- [ ] When the certificate holds for a qualifying group, escape ordering for
      that group follows the derived constructive order (verifiable via a
      `test_byte_lane_priority.py`-style unit test asserting
      `net_order` matches the certificate's output for a synthetic
      pin-sequence fixture).
- [ ] When the certificate does NOT hold, `net_order` for that group falls
      back to identity (or the existing #4051 reactive scheduler if the
      flag composition makes that the natural default) — never silently
      drops/duplicates nets (reuse the existing permutation-invariant
      defense-in-depth pattern at `core.py:5094` /
      `_apply_byte_lane_inner_priority`'s `sorted(reordered) ==
      sorted(net_order)` check).
- [ ] Measured reach delta on the DDR bundle recorded in the PR description:
      run the 11-net-alone isolation scenario from #3438's table (or the
      full board-07 recipe, or both) with the new ordering enabled vs the
      9/11-11/11 identity baseline. **Phase 1 acceptance is "no worse than
      identity baseline (currently best static result 10/11) AND
      certificate wired AND ordering honored when the certificate holds"**
      — 11/11 is the goal but is NOT required to close this issue; if
      Phase 1 doesn't reach 11/11, that result (and diagnosis of why) feeds
      the Phase 2 follow-up issue.
- [ ] Boards 00-07 regression-clean: existing board-level route/DRC/LVS
      gates (`tests/test_board_00_simple_led.py` through
      `tests/test_board_07_matchgroup_test.py`, plus
      `tests/router/test_byte_lane_priority.py` and
      `tests/router/test_byte_lane_corridor_reservation.py`) pass
      unmodified in behavior when the new flag is OFF (default), and do
      not regress board 07's existing match-group artifact assertions
      when the flag is exercised in new tests.
- [ ] New certificate module has direct unit test coverage independent of
      the grid/routing stack (pure combinatorial tests against known
      monotone-feasible and monotone-infeasible pin sequences from the
      paper's examples or hand-constructed cases).
- [ ] Phase 2 and Phase 3 are explicitly NOT implemented in this PR; if the
      builder identifies that Phase 1 alone cannot improve on identity for
      this bundle, file the Phase 2 follow-up issue with the measured
      diagnosis rather than expanding this PR's scope.

## Affected Files

- `src/kicad_tools/router/monotone_certificate.py` (new) — certificate
  check + constructive ordering (Tomioka & Takahashi ASP-DAC 2006), pure
  Python, no grid/pad dependencies.
- `src/kicad_tools/router/core.py` — `_apply_byte_lane_inner_priority`
  (`core.py:4712`), `_reactive_escape_freedom_order` (`core.py:5189`),
  `_schedule_by_escape_freedom` (`core.py:5397`), and the
  `enable_byte_lane_reorder` flag (`core.py:869-878`, constructor):
  wire in the certificate check and constructive order as an
  alternative/extension to the existing reactive scheduler.
- `src/kicad_tools/router/escape.py` —
  `EscapeRouter.reserve_inner_corner_lane_corridor` (`escape.py:2349`):
  reuse (and extend if needed) for per-net monotone corridor reservation;
  do not fork a parallel corridor primitive without cause.
- `src/kicad_tools/router/grid.py` — `RoutingGrid.reserve_corridor_cells`
  (`grid.py:4166`), `RoutingGrid.get_corridor_attractor_bonus`
  (`grid.py:4247`), `RoutingGrid.get_negotiated_cost` (`grid.py:5137`):
  read-only reference for the mechanism the monotone constraint rides;
  only touch if the fallback (cost-biasing) path is chosen.
- `tests/router/test_monotone_certificate.py` (new) — unit tests for the
  certificate module in isolation.
- `tests/router/test_byte_lane_priority.py` — extend with certificate-gated
  ordering test cases (synthetic pin sequences), following the existing
  test patterns in this file.
- `boards/07-matchgroup-test/generate_pcb.py`,
  `boards/07-matchgroup-test/generate_design.py` — reference only (DDR
  bundle fixture source); do not modify unless the reach-measurement
  repro script requires a new CLI hook.
- `tests/test_board_07_matchgroup_test.py` — extend if a reach/DRC
  regression assertion for the new ordering path is added.

## Test Plan

- [ ] Automated: new `tests/router/test_monotone_certificate.py` — unit
      tests for `check_monotone_feasibility` / `constructive_monotone_order`
      against known feasible/infeasible pin-sequence fixtures (fast, no
      grid dependency).
- [ ] Automated: extend `tests/router/test_byte_lane_priority.py` with
      certificate-driven ordering assertions (flag ON vs OFF), mirroring
      the existing `enable_byte_lane_reorder` test pattern
      (`router.enable_byte_lane_reorder = True; out =
      router._apply_byte_lane_inner_priority(net_ids)`).
- [ ] Automated: run full existing suite for
      `tests/router/test_byte_lane_priority.py`,
      `tests/router/test_byte_lane_corridor_reservation.py`, and
      `tests/test_board_07_matchgroup_test.py` — confirm no regressions
      with the new flag OFF (default).
- [ ] Manual/measured: run the board-07 DDR-bundle isolation scenario
      (11-net bundle alone on an empty board, per #3438's table) with the
      new ordering flag ON, record reach (X/11) and compare against the
      9/11 (reactive, #4051) and 10/11 (best static) baselines. Record the
      exact command / script used so it's reproducible.
- [ ] Manual/measured: run the full board-07 production recipe
      (`boards/07-matchgroup-test/generate_design.py --step route`) with
      the flag ON, confirm no LVS-open regression versus the current
      28-31/31-ish baseline (see #3438 for the exact historical numbers —
      re-verify against current `main`, they may have moved since #3438
      was filed).
- [ ] Regression: `uv run pytest tests/test_board_00_simple_led.py
      tests/test_board_01_ship_ready.py tests/test_board_03_regression.py
      tests/test_board_04_routing_regression.py
      tests/test_board_05_routing_regression.py
      tests/test_board_06_diffpair_test.py
      tests/test_board_07_matchgroup_test.py` — boards 00-07 stay green
      with the flag OFF (default state, so this should be a no-op
      confirmation).
- [ ] Verify `uv run kct build-native --check` reports the C++ backend is
      still used for the A* core (Phase 1 must not force a Python fallback
      for routing itself — only the pre-stage certificate/ordering is
      pure Python).

## Dependencies

None — Phase 1 builds directly on landed infrastructure (#2983 corridor
reservation, #4051 reactive scheduler scaffolding, both already on
`main`). No blocking issues.

rjwalters:
## Summary

Implements **Phase 1** of epic #4049 Gap 1: the Simultaneous Escape Routing
(SER) pre-stage. Adds the Tomioka & Takahashi (ASP-DAC 2006, IEEE 1594758)
*necessary-and-sufficient* monotonic feasibility certificate for a bundle of
two-terminal nets — each with one pin on each of two parallel boundaries
(facing pad columns) — plus the constructive net ordering that realises the
non-crossing monotone routes when the condition holds.

This is the **untried lever** the curation identified: instead of guessing a
fourth heuristic permutation (three static + #4051 reactive all failed to
beat the identity baseline on the #3438 DDR bundle), the certificate *decides*
— before any A* search — whether the bundle can be planarised by ordering at
all, and when it cannot, emits a **failure witness** naming the forced
crossing pairs (the diagnostic deliverable for the Phase 2/3 via/layer
follow-ups).

Pure Python, no C++ changes. Gated behind a new default-OFF flag.

## Changes

- **New `src/kicad_tools/router/monotone_certificate.py`** — `check_monotone_feasibility` / `constructive_monotone_order` / `monotone_certificate`. The condition reduces to "the boundary-A→boundary-B permutation has no inversions"; `normalize_boundary_pair` picks the boundary-B orientation (identity vs reverse) with fewer forced crossings, so co-oriented AND mirror-facing placements both classify correctly. Infeasible bundles return a `witness` list of `CrossingPair`s. No grid/pad/router dependency — unit-testable in isolation.
- **`src/kicad_tools/router/core.py`** — new `enable_monotone_certificate_order` flag (default `False`, independent of `enable_byte_lane_reorder`; takes precedence when both set) + `_apply_monotone_certificate_order`. The detection scan captures both facing columns' pin sequences; feasible groups follow the constructive order, infeasible groups stay at **identity** and log the witness. Slot-preserving + permutation-invariant (mirrors the existing byte-lane defense-in-depth). Latest classification per group recorded on `_last_monotone_certificates`.
- **`tests/router/test_monotone_certificate.py`** (new, 24 tests) — certificate in isolation: degenerate, feasible, mirror, reversed, partial-shuffle, witness completeness, input validation.
- **`tests/router/test_byte_lane_priority.py`** — new `TestMonotoneCertificateOrdering` (7 tests): flag-off identity, co-oriented→constructive, clean-reversal→mirror-feasible, imperfect-reversal→infeasible+witness, flag independence, permutation invariance. Adds a reversed-facing fixture helper.
- **`tests/router/test_monotone_reach.py`** (new) — fast certificate-classification check of the real board-07 bundle + `slow`-gated reach measurement (flag OFF vs ON).
- **`boards/07-matchgroup-test/ddr_bundle_isolation_repro.py`** (new) — standalone, reproducible "11-net bundle alone on an empty board" reach script (the #3438 isolation scenario was previously ad-hoc).

## Certificate classification of the #3438 DDR bundle

**Result: FEASIBLE** (`mirrored=False`, `inversions=0`, empty witness).

The DDR data byte declares the *same net order* on both facing QFN-48
columns (U1 pins 25–35 and U2 pins 1–11 both read `DQ0, DQ1, DQ2, DQ3, DM0,
DQS_P, DQS_N, DQ4, DQ5, DQ6, DQ7` top-to-bottom in `generate_pcb.py`). So
**along the row long axis the two columns are co-oriented**, and the pin
sequence itself is monotone-feasible. This was verified empirically (not
assumed): `bundle_river.py`'s "full bus reversal" framing describes the
*physical* mirror placement, but the net→row-position map is co-oriented.
The literature's two-parallel-boundaries precondition holds AND the ordering
is monotone — so the constructive order equals the row order.

## Measured reach

Isolated 11-net bundle on an empty 4-layer board (C++ backend), via
`ddr_bundle_isolation_repro.py`:

| Flag | Certificate | Routing order | Reach |
|------|-------------|---------------|-------|
| OFF (identity baseline) | — | `[1..11]` | **11/11** |
| ON | feasible, inversions=0 | constructive `[1..11]` | **11/11** |

Certificate ordering is **no worse than the identity baseline** (the Phase-1
bar). The isolated bundle routes fully at both settings — the #3438 2/11 loss
is a **full-board congestion** effect, not intrinsic to the 11-net bundle
alone (or the baseline has improved since #3438). Because the byte is
monotone-feasible, no failure witness is produced for it; the witness path is
exercised by the synthetic imperfect-reversal integration test.

| `check_monotone_feasibility` classifies the #3438 DDR bundle | ✅ | FEASIBLE (co-oriented, 0 inversions) — recorded above, asserted in `test_monotone_reach.py::TestDDRByteCertificateClassification` |
| When certificate holds, ordering follows constructive order | ✅ | `TestMonotoneCertificateOrdering::test_co_oriented_bundle_feasible_constructive_order` |
| When certificate fails, order falls back to identity (never drops/dupes) | ✅ | `test_imperfect_reversal_infeasible_records_witness`, `test_permutation_invariant_on_infeasible`; `sorted(reordered)==sorted(net_order)` guard in `core.py` |
| Failure witness names violating pin pairs | ✅ | `witness` field + `logger.info` crossing list; asserted in cert unit tests and integration test |
| Measured reach ≥ identity baseline (10/11) | ✅ | 11/11 both OFF and ON (table above); `test_certificate_reach_no_worse_than_identity_baseline` |
| Boards 00–07 regression-clean (flag OFF default) | ✅ | boards 00,01,03,04,05,06,07 + byte-lane + corridor-reservation suites pass; flag defaults OFF |
| New certificate module has isolated unit coverage | ✅ | `test_monotone_certificate.py` (24 tests, no grid dependency) |
| Phase 2/3 NOT implemented | ✅ | Only the certificate + constructive ordering + witness diagnostic; infeasible fallback is identity, not a partial B-escape |
| C++ backend still used for A* | ✅ | `kct build-native --check` → "available (version 1.0.0)"; certificate is pure-Python pre-stage |

## Corridor mechanism note

The certificate is an **ordering** pre-stage; it does not add corridor
reservations. The existing #2983 inner-corner reservation (which rides
`reserve_corridor_cells`, `soft=True`, Python-only per the #4079 evidence
record in `escape.py:2513`) continues to fire unchanged. Per the scope
addendum, hard/soft corridor fencing was already proven short-inducing on
the reversed byte, so Phase 1 deliberately touches ordering only.

## Test Plan

- `uv run pytest tests/router/test_monotone_certificate.py` — 24 passed
- `uv run pytest tests/router/test_byte_lane_priority.py tests/router/test_monotone_certificate.py tests/test_board_07_matchgroup_test.py -m "not slow"` — 97 passed
- `uv run pytest tests/router/test_monotone_reach.py -m ""` — 2 passed (incl. slow reach)
- `uv run pytest tests/test_board_00..06 -m "not slow"` — all passed
- `ruff check` + `ruff format --check` clean; `mypy` clean on the new module
- `kct build-native --check` — C++ backend available

Closes #4084

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 4080
-# Branch: feature/issue-4080
+# Issue: 4084
+# Branch: feature/issue-4084
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/boards/07-matchgroup-test/ddr_bundle_isolation_repro.py
+++ b/boards/07-matchgroup-test/ddr_bundle_isolation_repro.py
@@ -1,0 +1,152 @@
+#!/usr/bin/env python3
+"""Isolated reach measurement for the board-07 DDR data byte (Issue #4084).
+
+Reproduces the "11-net DDR bundle alone on an empty 4-layer board"
+scenario from #3438's isolation matrix so the monotonic-certificate
+(Issue #4084, Phase 1) reach delta is re-runnable, not just quoted in
+issue prose.
+
+The bundle is the DDR_DATA_BYTE_0 group on board 07: nine single-ended
+nets (DQ0-7 + DM0) plus the DQS_P/DQS_N diff pair, connecting a facing
+QFN-48 pin column on U1 (right side, pins 25-35) to its mirror on U2
+(left side, pins 1-11), at 0.8 mm pitch across a ~30 mm channel.  Both
+columns declare the byte in the SAME net order, so along the row long
+axis (y) the two facing columns are CO-ORIENTED, not reversed — which is
+exactly what the certificate reports.
+
+Usage:
+    uv run python boards/07-matchgroup-test/ddr_bundle_isolation_repro.py
+
+Prints, for the certificate flag OFF (identity baseline) and ON:
+  * the monotonic-certificate classification (feasible? witness?),
+  * the routing order the escape scheduler used,
+  * the measured reach (X/11 nets routed to completion).
+
+Pure in-process; no CLI subprocess (the flag is not exposed on ``kct
+route``).  Uses the negotiated router with the C++ backend when built.
+"""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+# Make the DDR pin declarations importable from the sibling generator.
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+
+from kicad_tools.router.core import Autorouter  # noqa: E402
+from kicad_tools.router.layers import LayerStack  # noqa: E402
+from kicad_tools.router.rules import NetClassRouting  # noqa: E402
+
+PITCH = 0.8
+
+# The DDR byte's row order on both facing columns (U1 pins 25-35 == U2
+# pins 1-11), matching generate_pcb.py's pin_nets declarations.  DQS_P /
+# DQS_N are the interleaved diff pair.
+ROW_NETS = [
+    "DQ0",
+    "DQ1",
+    "DQ2",
+    "DQ3",
+    "DM0",
+    "DQS_P",
+    "DQS_N",
+    "DQ4",
+    "DQ5",
+    "DQ6",
+    "DQ7",
+]
+
+
+def build_isolated_router(
+    *,
+    enable_certificate: bool,
+    channel_mm: float = 30.0,
+) -> tuple[Autorouter, list[int]]:
+    """Build a router with ONLY the 11-net DDR byte on an empty board.
+
+    U1's facing column sits at x=20, U2's mirror at x=20+channel; each net
+    connects one pad on each column at the same y (co-oriented), exactly
+    as board 07 declares them.
+    """
+    cls = NetClassRouting(
+        name="DDR_DATA_BYTE_0",
+        priority=1,
+        trace_width=0.15,
+        clearance=0.10,
+        length_critical=True,
+        length_match_group="DDR_DATA_BYTE_0",
+        length_match_reference=None,
+        length_match_tolerance_mm=0.1,
+    )
+    net_class_map: dict[str, NetClassRouting] = {}
+    router = Autorouter(
+        width=80.0,
+        height=40.0,
+        net_class_map=net_class_map,
+        layer_stack=LayerStack.four_layer_sig_gnd_pwr_sig(),
+    )
+    router.enable_monotone_certificate_order = enable_certificate
+
+    u1_x = 20.0
+    u2_x = u1_x + channel_mm
+    centre_y = 20.0
+    base_y = centre_y - (len(ROW_NETS) - 1) * PITCH / 2.0
+
+    net_ids: list[int] = []
+    for i, name in enumerate(ROW_NETS):
+        net_id = i + 1
+        net_ids.append(net_id)
+        y = base_y + i * PITCH
+        router.add_component(
+            "U1",
+            [{"number": str(25 + i), "x": u1_x, "y": y, "net": net_id, "net_name": name}],
+        )
+        router.add_component(
+            "U2",
+            [{"number": str(1 + i), "x": u2_x, "y": y, "net": net_id, "net_name": name}],
+        )
+        net_class_map[name] = cls
+    router.net_class_map = net_class_map
+    return router, net_ids
+
+
+def measure_reach(*, enable_certificate: bool) -> None:
+    router, net_ids = build_isolated_router(enable_certificate=enable_certificate)
+
+    label = "ON (certificate)" if enable_certificate else "OFF (identity baseline)"
+    print(f"\n=== enable_monotone_certificate_order = {label} ===")
+
+    # Show the ordering the escape scheduler will use for the group.
+    ordered = router._apply_byte_lane_inner_priority(list(net_ids))
+    print(f"routing order (net ids): {ordered}")
+    for grp, cert in router._last_monotone_certificates.items():
+        print(
+            f"certificate[{grp}]: feasible={cert.feasible} "
+            f"mirrored={cert.mirrored} inversions={cert.inversion_count}"
+        )
+        if not cert.feasible:
+            pairs = ", ".join(f"({p.net_a},{p.net_b})" for p in cert.witness)
+            print(f"  witness (forced crossings): {pairs}")
+
+    routes = router.route_all_negotiated(seed=42)
+    # A net is "reached" when it has at least one non-escape (full) route.
+    routed_nets = {
+        r.net
+        for r in routes
+        if getattr(r, "net", None) is not None and not getattr(r, "is_escape", False)
+    }
+    reached = len(routed_nets & set(net_ids))
+    print(f"reach: {reached}/{len(net_ids)} nets routed")
+
+
+def main() -> int:
+    print("Board-07 DDR data byte isolation reach measurement (Issue #4084)")
+    print(f"bundle: {len(ROW_NETS)} nets, 0.8mm pitch, facing QFN-48 columns")
+    measure_reach(enable_certificate=False)
+    measure_reach(enable_certificate=True)
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -895,6 +895,39 @@ class Autorouter:
         # (see ``_apply_byte_lane_inner_priority`` / ``bundle_river.py``).
         self.enable_bundle_river_planner = False
 
+        # Issue #4084 (Phase 1, epic #4049): monotonic feasibility
+        # certificate + constructive escape ordering (Tomioka & Takahashi,
+        # ASP-DAC 2006).  OFF by default, following the #4051 /
+        # ``enable_byte_lane_reorder`` precedent (three prior heuristics
+        # all regressed the DDR bundle before landing gated-off).
+        #
+        # When enabled, ``_apply_byte_lane_inner_priority`` runs the
+        # certificate on each qualifying byte-lane group's two facing pin
+        # sequences BEFORE any A* search:
+        #   * If the bundle is monotonically feasible as-pinned, the
+        #     group's occupied slots in ``net_order`` are permuted into the
+        #     certificate's constructive (non-crossing) order.
+        #   * If NOT feasible (e.g. board 07's imperfectly-reversed DDR
+        #     byte, whose crossing conflict graph is complete), the order
+        #     is left at IDENTITY and a failure witness (the forced
+        #     crossing pairs) is logged as a diagnostic — the fix class for
+        #     those pairs is via/layer assignment (Phase 2/3), not
+        #     ordering.
+        # Kept independent of ``enable_byte_lane_reorder`` so the
+        # certificate-driven order and the reactive scheduler can be A/B'd
+        # without interfering.  The certificate is a pure-Python pre-stage;
+        # it never forces a Python fallback for the A* core itself.  See
+        # ``monotone_certificate.py`` and
+        # ``_apply_byte_lane_inner_priority``.
+        self.enable_monotone_certificate_order = False
+
+        # Issue #4084: records the most recent certificate classification
+        # per qualifying byte-lane group (group name -> MonotoneCertificate)
+        # so callers/tests can inspect the feasibility decision and failure
+        # witness after a routing pass.  Populated by
+        # ``_apply_byte_lane_inner_priority`` whenever the certificate runs.
+        self._last_monotone_certificates: dict[str, Any] = {}
+
         # Initialize grid and routers using shared helper
         # Issue #972: Helper includes adaptive grid resolution for large boards
         self.grid, self.router, self.zone_manager = self._create_grid_and_routers(
@@ -4905,6 +4938,16 @@ class Autorouter:
         reorder_plans: list[list[int]] = []
         escape_row_positions: dict[int, float] = {}
 
+        # Issue #4084: per-group monotonic-certificate plans.  For each
+        # qualifying byte-lane group we record ``(group_name,
+        # primary_pin_sequence, secondary_pin_sequence)`` — the two facing
+        # columns' net-id orders along the row long axis — so the
+        # certificate can decide monotonic feasibility and (when feasible)
+        # derive the constructive order after the corridor-reservation
+        # pass.  Populated only when ``enable_monotone_certificate_order``
+        # is set, to keep the default path byte-identical.
+        certificate_plans: list[tuple[str, list[int], list[int]]] = []
+
         # Detection scan.  We walk each candidate group, identify its
         # primary component, project pads onto the row axis, and locate
         # the inner-corner indices.  As of Issue #4051 (Phase 1b of epic
@@ -4988,6 +5031,41 @@ class Autorouter:
             for nid_s in sorted_nets:
                 escape_row_positions[nid_s] = net_to_pad[nid_s][proj_index]
             reorder_plans.append(list(sorted_nets))
+
+            # Issue #4084: capture BOTH facing columns' pin sequences for
+            # the monotonic-certificate pre-stage.  ``sorted_nets`` is the
+            # primary column's net-id order along the row long axis; resolve
+            # the secondary (facing) column and sort it the same way so the
+            # certificate can diff the two orders.  Only computed when the
+            # certificate flag is on, so the default path is unchanged.
+            if getattr(self, "enable_monotone_certificate_order", False):
+                secondary_candidates = {
+                    ref: pads for ref, pads in comp_pad_count.items() if ref != primary_ref
+                }
+                if secondary_candidates:
+                    secondary_ref = max(
+                        secondary_candidates.keys(),
+                        key=lambda r: (
+                            len(secondary_candidates[r]),
+                            -ord(r[0]) if r else 0,
+                        ),
+                    )
+                    secondary_net_to_pad: dict[int, tuple[float, float]] = {}
+                    for s_nid, s_px, s_py in secondary_candidates[secondary_ref]:
+                        if s_nid not in secondary_net_to_pad:
+                            secondary_net_to_pad[s_nid] = (s_px, s_py)
+                    # Only nets present on BOTH columns form the matched bus
+                    # the certificate needs; sort the secondary column along
+                    # the SAME projection axis as the primary.
+                    secondary_sorted = sorted(
+                        (nid for nid in secondary_net_to_pad if nid in net_to_pad),
+                        key=lambda n: secondary_net_to_pad[n][proj_index],
+                    )
+                    primary_matched = [nid for nid in sorted_nets if nid in secondary_net_to_pad]
+                    if len(primary_matched) >= MIN_BYTE_LANE_SIZE and set(primary_matched) == set(
+                        secondary_sorted
+                    ):
+                        certificate_plans.append((grp_name, primary_matched, secondary_sorted))
 
             # Inner-corner indices in the sorted row.  Position 1 (one in
             # from the top corner) and position n-2 (one in from the
@@ -5105,6 +5183,29 @@ class Autorouter:
                         exc_info=True,
                     )
 
+        # Issue #4084 (Phase 1, epic #4049): monotonic feasibility
+        # certificate + constructive ordering (Tomioka & Takahashi,
+        # ASP-DAC 2006).  This is the *untried lever* the three prior
+        # heuristics (static monotone/reverse/outside-in and #4051's
+        # reactive scheduler) were not: instead of guessing a permutation,
+        # it DECIDES — before any A* search — whether the bundle can be
+        # planarised by ordering at all, and only reorders when it
+        # provably can.  Takes precedence over the reactive scheduler when
+        # ``enable_monotone_certificate_order`` is set.
+        if getattr(self, "enable_monotone_certificate_order", False) and certificate_plans:
+            reordered = self._apply_monotone_certificate_order(net_order, certificate_plans)
+            # Permutation invariant (defense-in-depth): a bug must degrade
+            # to identity, never corrupt the routing set.
+            if sorted(reordered) != sorted(net_order):
+                logger.error(
+                    "_apply_monotone_certificate_order produced a "
+                    "non-permutation (%d in, %d out); falling back to identity",
+                    len(net_order),
+                    len(reordered),
+                )
+                return net_order
+            return reordered
+
         # Issue #4051 (Phase 1b, epic #4049): apply a reactive
         # escape-freedom reorder on top of the corridor reservation.
         # Each qualifying byte-lane group's members are permuted *within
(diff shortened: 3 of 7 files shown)


rjwalters:
✅ **Approved!**

Phase 1 of the SER pre-stage (issue #4084, epic #4049 Gap 1). This is clean, well-scoped, and mathematically sound work. All 16 CI checks pass, merge state is CLEAN.

## Certificate correctness (verified by hand + tests)

The core reduction — "monotonically feasible ⟺ the boundary-A→boundary-B permutation has zero inversions" — is correct, and `normalize_boundary_pair` correctly handles the mirror case by testing both boundary-B orientations and keeping the min-inversion one. I hand-checked the non-trivial synthetic cases:

- `[1,2,3]` vs `[1,3,2]`: forward=1 inversion, reverse=2 → forward chosen, infeasible, witness `(2,3)` ✓
- `[1,2,3,4]` vs `[2,1,4,3]`: forward=2 (both `(1,2)` and `(3,4)`), reverse=4 → forward chosen, witness names both ✓
- `range(11)` vs `reversed`: reverse-of-B = identity → 0 inversions → mirrored-feasible ✓

Ran the unit suite locally against the branch source: **24/24 passed**. The witness only enumerates inversions under the *chosen* orientation, which is the correct semantics.

## Empirical claims — both confirmed

**Claim #1 (co-orientation) — CONFIRMED against `generate_pcb.py` directly.** U1's right column (pins 25–35) and U2's facing left column (pins 1–11) declare the identical net order top-to-bottom: `DQ0, DQ1, DQ2, DQ3, DM0, DQS_P, DQS_N, DQ4, DQ5, DQ6, DQ7`. So the net→row-position map is co-oriented; the certificate sees identical A/B sequences → 0 inversions → FEASIBLE, constructive order = row order. The "full reversal" in `bundle_river.py` is about physical mirror placement, not the net order — the PR's framing is accurate.

**Claim #2 (isolation reach) — classification CONFIRMED.** The fast classification test (`TestDDRByteCertificateClassification`) passed locally with the C++ backend: the real board-07 bundle classifies `feasible=True, inversion_count=0, witness=[]`. Ran the byte-lane integration + reach-classification suite: **23 passed**. The slow 11/11-reach measurement I let run locally but it is not load-bearing for approval — it's `slow`-gated (excluded from CI), the flag defaults OFF, and the Phase-1 bar is only "no worse than identity."

## Scope / deviation call

The decision to ship **ordering-only, no corridor reservation** is a legitimate Phase-1 scope call. The #4087 evidence that both HARD and SOFT corridor honoring short-induce on this byte justifies not adding a corridor constraint here; the curated body offered corridor honoring as a *fallback* ("ride existing reservation machinery"), not a hard requirement. The existing #2983 inner-corner reservation continues unchanged.

## Regression safety

- Flag `enable_monotone_certificate_order` defaults **OFF**, independent of `enable_byte_lane_reorder` (takes precedence when both set).
- Detection scan captures the secondary column only when the flag is on, so the default path is byte-identical.
- Slot-preserving + permutation-invariant with defense-in-depth: `sorted(reordered) != sorted(net_order)` guard falls back to identity, and the constructive-order path re-checks `sorted(constructive) == sorted(members)`.
- **Boards 00–07 all green in CI** (including the Board 07 Match-Group Routing Regression) — the regression proof.

## New-module quality

Excellent. Full docstrings cite the source theorem (Tomioka & Takahashi, ASP-DAC 2006, IEEE 1594758); typed throughout; input validation raises rather than silently misclassifying; O(k²) inversion enumeration is justified (small bundles + yields the witness pairs a merge-sort count wouldn't). `ruff check` + `ruff format --check` clean on both the new module and core.py; CI Type Check passed. No dead code.

## Test Execution

1. `pytest tests/router/test_monotone_certificate.py` (branch source, C++ backend) — ✅ 24 passed
2. `pytest tests/router/test_byte_lane_priority.py tests/router/test_monotone_reach.py -m "not slow"` — ✅ 23 passed
3. `ruff check` + `ruff format --check` on new module + core.py — ✅ clean
4. `kct build-native --check` — ✅ C++ backend available (1.0.0); certificate is pure-Python pre-stage, never forces a fallback
5. Slow isolation reach (11/11 OFF vs ON) — ⏭️ Not load-bearing (slow-gated, flag OFF, not in CI); classification confirmed via fast test
6. CI: all 16 checks pass, merge state CLEAN

Nice execution on the "decide, don't guess" lever — and the failure-witness diagnostic is a real deliverable for the Phase 2/3 via/layer follow-ups even where reach doesn't move.

rjwalters:
**Addendum — isolation reach claim independently reproduced.**

Completed the deferred verification of the 11/11 isolation-reach claim by running `boards/07-matchgroup-test/ddr_bundle_isolation_repro.py` in a fresh worktree with a freshly built C++ backend (`kct build-native`, build version 17):

| Flag | Certificate | Routing order | Reach |
|------|-------------|---------------|-------|
| OFF (identity baseline) | — | `[1..11]` | **11/11** |
| ON | `feasible=True, mirrored=False, inversions=0` | constructive `[1..11]` | **11/11** |

This exactly matches the PR's reported table. The isolated bundle routes fully at both settings with the native backend, confirming the #3438 2/11 loss is a full-board congestion effect rather than intrinsic bundle infeasibility.

Methodology note for the record: an earlier attempt from my review environment measured 7/11 baseline — that run had silently fallen back to the pure-Python A* (stale local `.so`, build version 14 vs required 17) and is not representative (2308s Python vs 0.4s native for the same bundle). The Python-fallback discrepancy is an environment artifact, not a PR issue; the repro script's own output makes the fallback visible, which is exactly why having this committed script beats quoting numbers in issue prose.

Verdict unchanged: ✅ Approved.

PR #4092, opened 2026-07-13
feat(router): slack-budget corridor widening + slack-aware serpentine tuner (Phase 1, flag-gated)

### LINKED ISSUE #4085: router: length-matching by construction — slack-corridor reservation during routing instead of post-hoc serpentine

rjwalters:
## Summary

Phase 1 of "length-matching by construction" (Issue #4085, epic #4049 Gap 2): reserve slack during routing so the serpentine tuner meanders into already-protected space, instead of scavenging leftover space after the fact. **Pure Python — no `router_cpp.*.so` / `src/kicad_tools/router/cpp/**` changes.** The LR-style live A* cost term is explicitly deferred to a follow-up (it would touch the hot A* loop and need C++/Python parity).

All three parts are gated behind a single new flag, `Autorouter.enable_slack_corridor_widening`, **defaulting OFF** — so every board 00-07 fixture is byte-identical to today.

## Changes

- **`src/kicad_tools/router/slack_budget.py` (new)** — pin-geometry skew estimator. `estimate_pair_skew_budget(leg_a, leg_b)` returns the abs difference of the two legs' Manhattan spans; `estimate_group_skew_budget` generalizes to N-member groups (max-min span spread). Pure geometry, no grid dependency.
- **`escape.py`** — `_reserve_pair_continuation_corridor` gains an optional `slack_budget_mm` arg and, when `enable_slack_corridor_widening` is on and the estimated skew exceeds the pair's `effective_skew_tolerance`, widens `lat_half` by the budget. New helpers `_resolve_slack_budget` / `_resolve_pair_skew_tolerance`. New instrumentation: `pair_corridor_slack_widened`, `pair_corridor_slack_budget_mm`.
- **`optimizer/serpentine.py`** — `find_best_segment(route, *, grid=None, reserved_net_id=None)` gives a 2.0x score bonus to segments whose midpoint cell is reserved for the net (`RoutingGrid.is_reserved_for`). New static `_segment_in_reservation`. With no grid → byte-identical geometric heuristic for every existing caller (`tune_match_group`, `apply_length_tuning`).
- **`diffpair_length_tuning.py`** — `tune_diff_pair_skew` threads `grid` + `prefer_reserved_slack` to `find_best_segment`.
- **`core.py`** — `Autorouter.enable_slack_corridor_widening` (default OFF) wires both halves: the escape-router flag and `apply_diffpair_length_tuning`'s tuner preference.

## Board 07 measured result (primary acceptance evidence)

**Before** (`kct check` on the shipped `matchgroup_test_routed.kicad_pcb`, verbatim):

```
  DQS_N        skew            11.034       0.050  FAIL
  MIPI_CLK_N   skew             2.973       0.050  FAIL
  MIPI_DAT1_N  skew            18.326       0.050  FAIL
  TMDS_D2_N    skew             3.783       0.075  FAIL
  diffpair_length_skew: 4 errors
  diffpair_routing_continuity: 4 errors
```

These match the issue's cited numbers exactly.

**After (flag ON): null effect — honestly reported, with a diagnosis.**

The measured skew delta on board 07 with the flag ON is **zero** for all 4 pairs. This is a legitimate, clean Phase-1 result that informs the follow-up. Root cause, verified empirically:

1. **Board 07's production route recipe cannot exercise the corridor-widening half.** `boards/07-matchgroup-test/generate_design.py` routes via `kct route --strategy negotiated --no-auto-layers --layers 4 --seed 42` and deliberately **omits `--differential-pairs`**, because the `CoupledPathfinder` pre-pass hangs CPU-bound for >40 min on this board (documented at length in `route_pcb()`; tracked as #3439). Without a diff-pair-aware escape/route pass, `_reserve_pair_continuation_corridor` never runs on this board's pairs, so no slack corridor is ever reserved.

2. **With no reservation present, the tuner-preference half is provably inert (by design).** I loaded the shipped routed PCB in-process and ran `tune_diff_pair_skew` OFF vs ON. Segment selection was **identical** in both cases (`find_best_segment` returned the same index for all 4 pairs) because there are zero reserved cells on a loaded finished board — `_segment_in_reservation` returns False for every candidate, so the bonus never fires and selection correctly falls back to the geometric heuristic. Separately, the tuner returns `post_insertion_drc_violation` with `inserts_applied=0` on all 4 pairs: the board is congested enough that the very first trombone collides, regardless of segment choice.

**Interpretation:** the two Phase-1 halves are coupled by construction — the tuner preference only has something to prefer once the widening reserves it *during* routing. Board 07's skews are structural (independently-routed legs on a board that can't re-route with diff-pairs enabled), exactly as the issue anticipated: "with the live cost term deferred, the corridor widening + tuner preference may only partially reduce the skews." The mechanism is unit-proven (below); its end-to-end effect on board 07 awaits a routable diff-pair path on that board (#3439) and/or the deferred live cost term.

`diffpair_routing_continuity`: unchanged (4 errors), consistent with no re-route occurring.

| Estimator has direct unit coverage vs hand-built fixtures | ✅ | `tests/router/test_slack_budget.py` (18 tests: known spans, symmetry, non-negativity, degenerate legs, N-group spread) |
| Widening increases reserved-cell count for out-of-tolerance skew; no-op within tolerance | ✅ | `TestSlackCorridorWidening` in `test_escape_diffpair.py`: `on_count > off_count` for 7mm induced skew; `on_count == off_count` for matched legs; gate-off byte-identical regardless of skew |
| Tuner prefers reserved slack cells when grid supplied; falls back otherwise | ✅ | `TestSerpentineSlackPreference` / `TestTunerSlackThreading`: reservation flips segment choice; foreign-net reservation does not; no-grid == geometric baseline |
| Board 07 before/after recorded | ✅ | See section above — before matches issue; after is null with a verified diagnosis (not a forced success claim) |
| `diffpair_routing_continuity` delta recorded | ✅ | Unchanged (4 errors); no re-route occurred |
| Boards 00-06 regression-clean | ✅ | Boards 00-05 (34 passed, 1 skipped) + board 06 diffpair (pass) + board 07 (pass), all with flag OFF |
| No C++ changes | ✅ | `git status` shows no `cpp`/`.so` files; `kct build-native --check` reports v1.0.0 available, unchanged |

## Test Plan

- `tests/router/test_slack_budget.py`, `TestSlackCorridorWidening`, `TestSerpentineSlackPreference`, `TestTunerSlackThreading` — all pass.
- Regression: `test_board_00..05`, `test_board_06_diffpair_test`, `test_board_07_matchgroup_test`, `test_escape_diffpair`, `test_byte_lane_corridor_reservation`, `test_cross_package_pair_corridor`, `test_match_group_tuning`, `test_length_matching`, `test_diffpair_serpentine_clearance` — all pass.
- `ruff check` + `ruff format` clean on all changed files.
- `uv run kct build-native --check` → `C++ backend: available (version 1.0.0)` (no rebuild required).

## Follow-up (out of scope, to file after this lands)

- LR-style per-net A* cost term (under-target nets get a bonus for consuming their own reserved slack during the live search) — touches the hot A* cost loop, needs C++/Python parity.
- A routable board-07 diff-pair path (#3439) so the end-to-end corridor-widening effect can actually be measured on that fixture.

Closes #4085

### DIFF
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -946,6 +946,23 @@ class Autorouter:
         # no-op ``continue`` there regardless.  Set True to opt in.
         self.enable_cross_package_pair_corridor = False
 
+        # Issue #4085 (Phase 1): slack-corridor reservation for
+        # length-matching by construction.  When True, two independent
+        # things engage:
+        #   (1) ``EscapeRouter._reserve_pair_continuation_corridor`` widens
+        #       a pair's inner-layer continuation corridor by a
+        #       pin-geometry-estimated slack budget when the estimated skew
+        #       exceeds the net-class tolerance (via the escape router's
+        #       ``enable_slack_corridor_widening`` flag, threaded below).
+        #   (2) ``apply_diffpair_length_tuning`` passes the routing grid to
+        #       ``tune_diff_pair_skew`` with ``prefer_reserved_slack=True``
+        #       so the serpentine tuner prefers segments inside the pair's
+        #       own reserved slack corridor.
+        # Flag-off (default) is byte-identical to pre-#4085 main: the escape
+        # corridor keeps its fixed padding and the tuner uses the
+        # geometric-only segment heuristic.  Set True to opt in.
+        self.enable_slack_corridor_widening = False
+
         # Issue #4084: records the most recent certificate classification
         # per qualifying byte-lane group (group name -> MonotoneCertificate)
         # so callers/tests can inspect the feasibility decision and failure
@@ -13032,6 +13049,13 @@ class Autorouter:
                 tolerance_mm=tolerance_mm,
                 intra_pair_clearance_mm=intra_pair_clearance_mm,
                 length_critical=length_critical,
+                # Issue #4085 (Phase 1): when the slack-corridor gate is on,
+                # let the tuner prefer segments inside the pair's own
+                # reserved slack corridor.  Grid is passed unconditionally
+                # (cheap) but only consulted when prefer_reserved_slack is
+                # True; flag-off keeps the pre-#4085 geometric selection.
+                grid=self.grid,
+                prefer_reserved_slack=self.enable_slack_corridor_widening,
             )
 
             # Commit any new Route references back into self.routes and the
@@ -13985,6 +14009,10 @@ class Autorouter:
                 # cross-package branch is a no-op ``continue``.
                 enable_cross_package_pair_corridor=self.enable_cross_package_pair_corridor,
                 net_name_to_id=net_name_to_id,
+                # Issue #4085 (Phase 1): slack-corridor widening gate
+                # (default OFF).  When OFF the pair continuation corridor
+                # keeps its fixed padding, byte-identical to pre-#4085.
+                enable_slack_corridor_widening=self.enable_slack_corridor_widening,
                 # Issue #3428: board-wide net-id -> pad-position map so the
                 # fine-pitch QFP in-pad rescue can aim its inner stub at
                 # the net's actual routing target instead of the parity-
--- a/src/kicad_tools/router/diffpair_length_tuning.py
+++ b/src/kicad_tools/router/diffpair_length_tuning.py
@@ -67,6 +67,7 @@ from .optimizer.serpentine import (
 
 if TYPE_CHECKING:
     from .diffpair_detection import DetectedPair
+    from .grid import RoutingGrid
     from .primitives import Route, Segment
 
 
@@ -153,6 +154,8 @@ def tune_diff_pair_skew(
     config: SerpentineConfig | None = None,
     max_inserts: int = MAX_INSERTS_PER_PAIR,
     length_critical: bool = True,
+    grid: RoutingGrid | None = None,
+    prefer_reserved_slack: bool = False,
 ) -> tuple[Route, Route, DiffPairTuneResult]:
     """Tune the skew of a detected diff pair by serpentining the shorter half.
 
@@ -182,6 +185,18 @@ def tune_diff_pair_skew(
             ``net_class_routing.length_critical`` here -- pairs whose net
             class is not flagged as length-critical are skipped, matching
             the curator's engagement-gate spec on Issue #2648.
+        grid: Issue #4085 (Phase 1).  Optional routing grid.  When supplied
+            together with ``prefer_reserved_slack=True``, the tuner biases
+            its serpentine-insertion segment choice toward segments sitting
+            inside the shorter half's own reserved slack corridor (the one
+            widened by ``EscapeRouter._reserve_pair_continuation_corridor``)
+            so the meander lands in already-protected space.  When ``None``
+            or ``prefer_reserved_slack=False`` (both defaults) segment
+            selection is byte-identical to the pre-#4085 geometric
+            heuristic.
+        prefer_reserved_slack: Issue #4085.  Gate for the slack-aware
+            segment preference above.  Default ``False`` (inert).  Has no
+            effect unless ``grid`` is also supplied.
 
     Returns:
         ``(p_route, n_route, result)`` where ``p_route`` and ``n_route`` are
@@ -301,7 +316,15 @@ def tune_diff_pair_skew(
         # logic in SerpentineGenerator.find_best_segment so the hint is
         # computed for the same segment that the generator will use.
         generator = SerpentineGenerator(base_config)  # provisional
-        best = generator.find_best_segment(current_shorter)
+        # Issue #4085: when a grid is supplied and the slack-preference gate
+        # is on, bias the segment choice toward the shorter half's own
+        # reserved slack corridor.  Otherwise this is byte-identical to the
+        # pre-#4085 geometric-only selection.
+        best = generator.find_best_segment(
+            current_shorter,
+            grid=grid if prefer_reserved_slack else None,
+            reserved_net_id=shorter_id if prefer_reserved_slack else None,
+        )
         if best is None:
             result.reason = "no_suitable_segment"
             result.message = (
--- a/src/kicad_tools/router/escape.py
+++ b/src/kicad_tools/router/escape.py
@@ -1098,6 +1098,7 @@ class EscapeRouter:
         net_target_positions: dict[int, list[tuple[float, float, str]]] | None = None,
         enable_cross_package_pair_corridor: bool = False,
         net_name_to_id: dict[str, int] | None = None,
+        enable_slack_corridor_widening: bool = False,
     ):
         """Initialize the escape router.
 
@@ -1172,6 +1173,20 @@ class EscapeRouter:
                 the soft attractor bonus).  Defaults to ``None``; when
                 absent the cross-package corridor owner set falls back to
                 this leg's net id alone.
+            enable_slack_corridor_widening: Issue #4085 (Phase 1,
+                epic #4049 Gap 2).  Default ``False``.  When ``True``,
+                ``_reserve_pair_continuation_corridor`` estimates the
+                pair's expected length skew from pin geometry (via
+                :func:`slack_budget.estimate_pair_skew_budget`) and, when
+                that estimate exceeds the pair's net-class skew tolerance,
+                WIDENS the reserved corridor's lateral half-width by the
+                estimated slack budget.  This reserves room for the
+                downstream serpentine tuner to meander the shorter half
+                into already-protected cells rather than scavenging
+                leftover space.  With the flag ``False`` the corridor
+                width is byte-identical to today (the fixed
+                ``intra_pair_clearance + trace_width`` padding), so all
+                board 00-07 fixtures are unchanged.
         """
         self.grid = grid
         self.rules = rules
@@ -1226,6 +1241,18 @@ class EscapeRouter:
         self.enable_cross_package_pair_corridor: bool = bool(enable_cross_package_pair_corridor)
         self.cross_package_pair_corridor_reservations: int = 0
         self.cross_package_pair_corridor_reserved_cells: int = 0
+        # Issue #4085 (Phase 1): slack-corridor widening gate +
+        # instrumentation.  When enabled, ``_reserve_pair_continuation_corridor``
+        # widens its lateral reservation by a pin-geometry-estimated slack
+        # budget for pairs whose estimated skew exceeds their net class's
+        # ``effective_skew_tolerance``.  ``pair_corridor_slack_widened``
+        # counts how many reservations were actually widened;
+        # ``pair_corridor_slack_budget_mm`` accumulates the total slack
+        # (mm) applied so tests can assert the widening fired without
+        # monkey-patching internals.
+        self.enable_slack_corridor_widening: bool = bool(enable_slack_corridor_widening)
+        self.pair_corridor_slack_widened: int = 0
+        self.pair_corridor_slack_budget_mm: float = 0.0
         # Issue #4086: board-wide net-name -> net-id map used ONLY to
         # resolve the off-package partner's net id for the cross-package
         # corridor owner set.  Empty map => owner set is this leg alone.
@@ -2174,11 +2201,86 @@ class EscapeRouter:
         )
         return escape_p, escape_n
 
+    def _resolve_slack_budget(
+        self,
+        members: list[EscapeRoute],
+        slack_budget_mm: float | None,
+    ) -> float:
+        """Resolve the pin-geometry slack budget for a paired corridor.
+
+        Issue #4085 (Phase 1).  When ``slack_budget_mm`` is supplied it is
+        returned verbatim (a caller that already computed the estimate).
+        Otherwise the budget is estimated from the members' boundary pin
+        geometry via :func:`slack_budget.estimate_pair_skew_budget` (2
+        legs) or :func:`slack_budget.estimate_group_skew_budget` (N>2).
+
+        The per-leg pin set prefers the board-wide ``net_pad_positions``
+        map (all pads on the net, so the estimate spans the full
+        source->sink extent) and falls back to the escape point + pad
+        centre when the map is not populated.
+
+        Args:
+            members: The paired EscapeRoutes.
+            slack_budget_mm: Optional explicit override.
+
+        Returns:
+            The slack budget in mm (``>= 0``).
+        """
+        if slack_budget_mm is not None:
+            return max(0.0, float(slack_budget_mm))
+
+        from .slack_budget import (
+            estimate_group_skew_budget,
+            estimate_pair_skew_budget,
+        )
+
+        legs_pins: list[list[tuple[float, float]]] = []
+        for m in members:
+            name = m.pad.net_name or ""
+            pins = list(self.net_pad_positions.get(name, [])) if name else []
+            if len(pins) < 2:
+                # Fallback: the two endpoints we know about are the pad
+                # centre and the escape point.  This still captures the
+                # per-leg launch asymmetry when no board-wide map exists.
+                pins = [(m.pad.x, m.pad.y), m.escape_point]
+            legs_pins.append(pins)
+
+        if len(legs_pins) == 2:
+            return estimate_pair_skew_budget(legs_pins[0], legs_pins[1])
+        return estimate_group_skew_budget(legs_pins)
+
+    def _resolve_pair_skew_tolerance(self, members: list[EscapeRoute]) -> float:
+        """Resolve the skew tolerance (mm) for a paired corridor.
+
+        Issue #4085 (Phase 1).  Reads the first member's net class from
+        ``net_class_map`` and returns its
+        :meth:`NetClassRouting.effective_skew_tolerance`.  Falls back to
+        the module-level 0.5 mm default (matching
+        ``Autorouter.apply_diffpair_length_tuning``) when no net class is
+        configured for the pair.
+
+        Args:
+            members: The paired EscapeRoutes.
+
+        Returns:
+            The skew tolerance in mm.
+        """
+        default = 0.5
+        first_net = members[0].pad.net_name or ""
+        nc = self.net_class_map.get(first_net) if self.net_class_map else None
+        if nc is not None and hasattr(nc, "effective_skew_tolerance"):
+            try:
+                return float(nc.effective_skew_tolerance(default))
+            except Exception:
+                return default
+        return default
+
     def _reserve_pair_continuation_corridor(
         self,
         members: list[EscapeRoute],
         target_inner_layer: Layer,
         intra_pair_clearance: float | None = None,
+        slack_budget_mm: float | None = None,
     ) -> int:
         """Reserve an inner-layer continuation corridor for paired escapes.
 
@@ -2247,6 +2349,17 @@ class EscapeRouter:
                 first member's net class via
                 ``_resolve_intra_pair_clearance`` (same value the
                 segment generator used).
+            slack_budget_mm: Issue #4085 (Phase 1).  Optional pre-computed
+                pin-geometry slack budget (mm).  Only consulted when
+                ``self.enable_slack_corridor_widening`` is ``True`` and the
+                budget exceeds the pair's net-class skew tolerance, in
+                which case ``lat_half`` is widened by the budget so the
+                downstream serpentine tuner has reserved cells to meander
+                into.  When ``None`` (default) the estimator is invoked
+                internally from the members' pad geometry; pass an explicit
+                value to override (e.g. from a caller that already computed
+                it).  Ignored entirely when the widening gate is off, so
+                behaviour is byte-identical to today.
 
         Returns:
             Number of grid cells reserved.  Returns 0 if the helper is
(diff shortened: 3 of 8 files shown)


rjwalters:
❌ **Changes Requested — Type Check gate failing**

Strong Phase-1 work: the mechanism is unit-proven, the flag-OFF gating is clean, and the null-effect result on board 07 is honestly reported with a verified root-cause diagnosis (#3439 recipe omits `--differential-pairs`, so no pair corridors exist for the widening half; the tuner preference then correctly falls back). I accept the scoping and the "recorded null measurement" as a valid completion of the acceptance criteria as curated — **but the required `Type Check` gate is red and I cannot approve around it.**

**Blocking: 2 NEW mypy errors beyond baseline** (job [Type Check](https://github.com/rjwalters/kicad-tools/actions/runs/29244871491/job/86799250030)):

```
new type error: Argument N to "_segment_in_reservation" of "SerpentineGenerator" has incompatible type "RoutingGrid | None"; expected "RoutingGrid"  [arg-type]
new type error: Argument N to "_segment_in_reservation" of "SerpentineGenerator" has incompatible type "int | None"; expected "int"  [arg-type]
```

**Root cause** (`src/kicad_tools/router/optimizer/serpentine.py`, `find_best_segment`): the runtime guard is correct — `slack_aware = grid is not None and reserved_net_id is not None` — but mypy cannot narrow `grid`/`reserved_net_id` through the intermediate `slack_aware` boolean, so at the call site both are still typed `RoutingGrid | None` / `int | None`. This is a false positive in the sense that the code is runtime-safe, but the gate (correctly) refuses to accept it silently.

**Preferred fix** — inline the narrowing so mypy sees it directly:

```python
if (
    grid is not None
    and reserved_net_id is not None
    and self._segment_in_reservation(seg, grid, reserved_net_id)
):
    score *= 2.0
```

(You can drop the `slack_aware` local, or keep it for readability and add the two `is not None` checks back at the call site — either works. Per the Type Check gate policy, do **not** add these to `.github/mypy-baseline.txt`; a targeted `# type: ignore[arg-type]` is acceptable only if you'd rather not restructure, but the inline narrowing is cleaner and I'd prefer it.)

**Everything else I reviewed passes:**
- **Estimator math** (`slack_budget.py`): correct. `_manhattan_span` = half-perimeter of the pin bounding box; pair budget = `abs(span_a - span_b)` (symmetric, `>= 0`, `0.0` for degenerate legs); group budget = `max(span) - min(span)`, matching `MatchGroupTracker`'s `max(L) - min(L)` skew definition. Matches the docstrings and unit fixtures.
- **Flag-OFF byte-identity**: `escape.py` widening is gated behind `enable_slack_corridor_widening and len(members) >= 2` and only fires when `budget > tol`; `find_best_segment`'s bonus is gated behind `grid is not None and reserved_net_id is not None`. No-grid → identical geometric heuristic. Board gates 00–07 will confirm once CI reruns.
- **Serpentine preference**: multiplicative `*= 2.0` applied *after* the geometric factors, only to already-valid candidates — it re-ranks, it never promotes an otherwise-invalid segment. `_segment_in_reservation` is defensive (`except Exception: return False`).
- **Null-effect ruling**: the issue asked to *record* board-07 before/after, not to improve it; the recorded null + diagnosis satisfies that, consistent with the #4087 precedent. No objection on scope.

Push the one-line narrowing, let the Type Check gate go green (and confirm the board 00–07 gates pass — they were still pending at review time), then flip back to `loom:review-requested`. This should be a fast turnaround.

rjwalters:
Fixed the Type Check gate blocker.

**Change** (`src/kicad_tools/router/optimizer/serpentine.py`, `find_best_segment`): dropped the intermediate `slack_aware` boolean and inlined the `is not None` narrowing directly at the `_segment_in_reservation` call site, exactly as preferred:

```python
if (
    grid is not None
    and reserved_net_id is not None
    and self._segment_in_reservation(seg, grid, reserved_net_id)
):
    score *= 2.0
```

mypy now proves `grid`/`reserved_net_id` are non-optional at the call site. Runtime behavior is identical (same short-circuit guard). No `# type: ignore`, and `.github/mypy-baseline.txt` is untouched per gate policy.

**Verification (local):**
- `scripts/ci/check_mypy_baseline.py` (the CI Type Check command): `OK: ... all within the baseline (1514 allowed). No new type errors.` — the 2 new arg-type errors are gone.
- `ruff check` + `ruff format --check` on the changed file: clean.
- `pytest tests/router/test_slack_budget.py tests/test_diffpair_length_tuning.py tests/test_escape_diffpair.py`: 88 passed (all 39 new tests included).

No other files changed. Flipped back to `loom:review-requested`.

rjwalters:
✅ **Approved!** (Doctor→Judge re-review, commit `cf993b4a`)

**Prior single blocker — RESOLVED.** My earlier rejection had exactly one blocker: 2 new mypy errors in `optimizer/serpentine.py` where `grid: RoutingGrid | None` / `reserved_net_id: int | None` weren't narrowed through the intermediate `slack_aware` boolean before the `_segment_in_reservation(...)` call.

**Diff verification.** The fix commit touches only `src/kicad_tools/router/optimizer/serpentine.py` and is exactly the promised narrowing change:
- Dropped the `slack_aware` intermediate.
- Inlined `grid is not None and reserved_net_id is not None and self._segment_in_reservation(seg, grid, reserved_net_id)` directly in the guard, so mypy proves both are non-optional at the call site.
- No `type: ignore`, mypy baseline untouched, no other files affected.
- Short-circuit semantics are identical to the original — runtime behavior unchanged.

**CI — all 16 checks green, merge state CLEAN:**
- Type Check: **pass** (the exact blocker — now green)
- Test: pass
- Lint & Format, C++ Build Check, Routed PCB DRC, kicad-cli Round-trip Smoke: pass
- Board 00-07 End-to-End: all pass
- Board 05 Routing Regression (blocking <= 11): pass
- Diff-Pair Routing Regression (board 06): pass
- Match-Group Routing Regression (board 07): pass

Everything else was already ruled acceptable in the prior review, including the null-effect completion question (valid per the #4087 precedent). Approving and promoting to `loom:pr`.

PR #4093, opened 2026-07-13
test(router): gate slow DDR reach absolute assertions on native backend

### LINKED ISSUE #4091: test_monotone_reach slow test should skip (or soften assertion) when C++ backend unavailable
## Context

While judging PR #4089 (issue #4084), I independently confirmed the isolated DDR-byte reach claim: with the C++ backend, `tests/router/test_monotone_reach.py::TestDDRByteReach` measures 11/11 both flag OFF and ON — matching the PR.

But I also (accidentally) measured the pure-Python fallback twice: the same isolated 11-net bundle reaches only **7/11** on the Python A* (2308s vs 0.4s native). The slow test hard-asserts full reach:

```python
assert baseline == len(ids_off)
assert with_cert == len(ids_on)
```

So any developer running `uv run pytest tests/router/test_monotone_reach.py -m "slow"` in a worktree without the native extension built (fresh checkouts and `.loom/worktrees/issue-N/` don't get it from `uv sync`) will hit a confusing hard failure that looks like a routing regression but is actually a backend artifact.

## Suggested fix (one of)

1. `pytest.skip` the `TestDDRByteReach` class when the C++ backend is unavailable or version-gated stale (mirror however `Autorouter`/`backend_info` reports it), with a message pointing at `kct build-native`; or
2. Keep the relative assertion (`with_cert >= baseline` — the actual Phase-1 acceptance bar) unconditionally, and gate only the absolute `== 11` assertions on native availability.

Option 2 preserves the no-worse-than-identity regression guard on all backends while dropping only the backend-dependent absolute bar.

## Evidence

- Native backend (build version 17): OFF 11/11, ON 11/11 (PR #4089 addendum comment has the table)
- Pure-Python fallback: OFF 7/11 (measured twice, seeds identical, 2308s and 2392s runs)

Not a defect in PR #4089 — the flag defaults OFF and the test is slow-gated out of CI — purely a developer-ergonomics guard.

Discovered during code evaluation of PR #4089.

rjwalters:
## Summary

The slow `TestDDRByteReach.test_certificate_reach_no_worse_than_identity_baseline`
in `tests/router/test_monotone_reach.py` asserted absolute `11/11` reach on both
the flag-OFF baseline and the flag-ON certificate run. Those absolute numbers are
**backend-dependent**: they hold with the C++ `router_cpp` extension, but the
pure-Python A* fallback deterministically reaches only 7/11. Any fresh checkout or
Loom worktree (`.loom/worktrees/issue-N/`) that hasn't run `kct build-native`
fails this test with a confusing "routing regression" `AssertionError` that is
actually a backend artifact, not a router defect (`uv sync` does not build the
native extension — see repo `CLAUDE.md`).

## Changes

- Import `is_cpp_available` from `kicad_tools.router.cpp_backend` (the established
  repo idiom, 170+ references; it internally checks `BUILD_VERSION`, so a stale
  `.so` is also correctly reported unavailable).
- Keep the relative `assert with_cert >= baseline` guard **unconditional** so the
  Phase-1 acceptance bar still runs on every backend, including the Python fallback.
- Gate the two absolute assertions (`baseline == len(ids_off)`,
  `with_cert == len(ids_on)`) behind `if is_cpp_available():`.
- Guard is scoped to `TestDDRByteReach` only — no module-level `pytestmark`. The
  fast, non-routing `TestDDRByteCertificateClassification` is unaffected.

| Slow test no longer hard-fails when C++ backend unavailable/stale | ✅ | Absolute asserts now gated on `is_cpp_available()`; relative-only path runs otherwise |
| Relative guard (`with_cert >= baseline`) still runs on every backend | ✅ | Assertion left unconditional above the gate |
| With C++ backend built, behavior unchanged (still asserts `== 11`) | ✅ | `kct build-native --check` → available; slow test PASSED with absolute asserts active |
| `TestDDRByteCertificateClassification` unaffected | ✅ | Fast class run: 1 passed; no skip condition added to it |
| Fix confined to `tests/router/test_monotone_reach.py`; no new labels | ✅ | `git diff --stat`: single test file, 14 insertions / 5 deletions |

## Test Plan

Ran in the issue worktree with the native extension built:

- `uv run kct build-native --check` → `C++ backend: available (version 1.0.0)`
- `uv run pytest ...::TestDDRByteCertificateClassification -v` → **1 passed**
- `uv run pytest ...::TestDDRByteReach --collect-only` → **1 test collected** (clean)
- `uv run pytest ...::TestDDRByteReach::test_certificate_reach_no_worse_than_identity_baseline -m slow -v` → **1 passed** (absolute `== 11` assertions active and passing under C++ backend)
- `uv run ruff check` + `ruff format --check` on the changed file → clean

Closes #4091

### DIFF
--- a/tests/router/test_monotone_reach.py
+++ b/tests/router/test_monotone_reach.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 import pytest
 
 from kicad_tools.router.core import Autorouter
+from kicad_tools.router.cpp_backend import is_cpp_available
 from kicad_tools.router.layers import LayerStack
 from kicad_tools.router.rules import NetClassRouting
 
@@ -124,9 +125,17 @@ class TestDDRByteReach:
         with_cert = _reach(router_on.route_all_negotiated(seed=42), ids_on)
 
         # Phase-1 acceptance: certificate ordering is NO WORSE than the
-        # identity baseline on the isolated bundle.
+        # identity baseline on the isolated bundle.  This relative guard holds
+        # on every backend, including the pure-Python A* fallback.
         assert with_cert >= baseline
-        # Both reach full on the empty board (the #3438 2/11 loss is a
-        # full-board congestion effect, not intrinsic to the bundle).
-        assert baseline == len(ids_off)
-        assert with_cert == len(ids_on)
+        # The absolute reach numbers are backend-dependent: full 11/11 with the
+        # C++ router_cpp extension, but only 7/11 on the pure-Python fallback
+        # (a backend artifact, not a router regression — see repo CLAUDE.md,
+        # `uv sync` does not build the native extension).  Gate them on native
+        # availability so a fresh checkout / Loom worktree that hasn't run
+        # `kct build-native` sees a clean pass, not a confusing AssertionError.
+        if is_cpp_available():
+            # Both reach full on the empty board (the #3438 2/11 loss is a
+            # full-board congestion effect, not intrinsic to the bundle).
+            assert baseline == len(ids_off)
+            assert with_cert == len(ids_on)


rjwalters:
✅ **Approved!**

Clean, well-scoped test fix that resolves the backend-dependent flakiness described in #4091.

**Verification:**

| Review point | Result |
|--------------|--------|
| Guard covers exactly the two absolute asserts (`baseline == len(ids_off)`, `with_cert == len(ids_on)`) | ✅ Both gated behind `if is_cpp_available():` |
| Relative claim (`with_cert >= baseline`) stays unconditional | ✅ Left above the gate — runs on every backend including the Python fallback |
| Matches curated Option 2 (inline guard, `TestDDRByteReach` only, no module-level `pytestmark`) | ✅ Confirmed |
| Fast `TestDDRByteCertificateClassification` untouched | ✅ Separate class (line 101), no skip condition added |
| Diff hygiene | ✅ Single file, +14/-5 |
| Import idiom | ✅ `is_cpp_available` is a real function in `cpp_backend.py`, used across 46 repo files; its internal `BUILD_VERSION` check also correctly reports a stale `.so` as unavailable |

**CI:** All 17 checks pass, merge state CLEAN. The native-built "Board 05 Routing Regression" job passes (absolute `== 11` path active) and the "Test" job passes (proving the Python-fallback path collects and runs without the AssertionError). The explanatory comments citing #3438 and CLAUDE.md are a nice touch for future readers.

**Closes #4091** — auto-close syntax present.

PR #4106, opened 2026-07-13
feat(diffpair): warn when coupled pairs budget-exit to single-ended

### LINKED ISSUE #4095: router: --differential-pairs must never ship a result worse than the single-ended baseline (board 07: 15 new clearance violations + net-loss)
**Source:** Epic #4049 closeout assessment (see the final architect comment there for full measurements). Follow-on B — maintenance/correctness.

## Problem (measured on main, C++ backend, seed 42)

A fresh board-07 re-route with `--differential-pairs` produces a strictly worse result than the committed single-ended artifact:
- **34 DRC errors vs. 13** — same 5 opens, but 6 skew + 6 continuity + **15 NEW clearance violations** (10 pad-segment, 5 pad-via) introduced by the coupled path.
- All 7 coupled pairs budget-exit at `reason=iteration iters=20000` in ~1.0s each and fall back to single-ended; the subsequent negotiation churn trades away already-landed nets (a fuller run plateaued at **22/31 vs. 26/31**).

The no-net-loss guarantee holds *within* negotiation but not *across the coupled→single-ended boundary*: the pipeline can ship a coupled-attempt result that is strictly dominated by the plain single-ended route.

This is not hypothetical for board 07 specifically — `boards/07-matchgroup-test/generate_design.py` (`route_pcb`, ~line 1244) already documents that its own generator omits `--differential-pairs` for this exact reason ("the `CoupledPathfinder` pre-pass hangs CPU-bound... This recipe therefore continues to omit `--differential-pairs`"). The board-level generator worked around the bug by never calling the flag; **any other user or board that does call `--differential-pairs` on a bundle-dense board is still exposed.**

## Root cause: the coupled→single-ended boundary has no result comparison

Traced the flag from CLI to router core:

- **CLI dispatch** — `src/kicad_tools/cli/route_cmd.py:9648` (`--differential-pairs --strategy negotiated`, the recipe board 07 would use) calls `router.route_all_with_diffpairs(diffpair_config, non_diffpair_strategy=_neg_strategy, coupled_only=True, timeout=...)`. The `basic` strategy variant is at line 9725; escape-composed dispatch is `route_with_escape_and_diffpairs` (`src/kicad_tools/router/core.py:14700`), which delegates to the same orchestrator.
- **Orchestrator** — `src/kicad_tools/router/diffpair_routing.py:6567` (`route_all_with_diffpairs`, also proxied at `src/kicad_tools/router/core.py:13469`) loops over detected pairs (line 6698), attempts each via `route_differential_pair_coupled`, and tracks budget-exits in a **local** `budget_exit_diff_nets` set (line 6687). Confirmed via `_last_pair_budget_exit`, set at `diffpair_routing.py:5376` when the inner `CoupledPathfinder`'s iteration/timeout budget fires.
- **Fallback wiring is correct as far as it goes**: budget-exit nets are correctly dropped from `diff_net_ids` (line 6831) so the main strategy's per-net C++ A* routes them normally — no net silently goes unrouted. `budget_exit_diff_nets` also drives a net-priority promotion (`self.autorouter._budget_exit_diff_nets`, line 6842, issue #3270) so deferred nets get first pick of the corridor.
- **The gap**: `budget_exit_diff_nets` is **local to `route_all_with_diffpairs`** — never returned, never distinct from the generic `LengthMismatchWarning` list (`diffpair_warnings`, the only warning type that survives to `route_cmd.py:10124`'s `--- Differential Pair Warnings ---` report). There is no code path that: (1) counts/names budget-exited pairs for CLI-level reporting, or (2) compares the coupled-enabled final result (routed count, DRC errors) against what a plain single-ended run would have produced.
- **Confirmed the "no-net-loss" guarantee does not reach this boundary**: `rg "no.net.loss"` in `core.py` (lines 8905, 8922, 9773, 10774, 10811, 11430) shows every occurrence is internal to the negotiated rip-up/reroute loop's own rollback bookkeeping — none of it wraps the coupled-pass-vs-single-ended-pass comparison the issue is asking for.

## Feasibility assessment: warn-only vs. checkpoint-and-compare

The issue's "ideal" tier asks for checkpoint-and-compare: if the coupled-enabled final result is worse than a single-ended equivalent (routed count, DRC errors), ship the single-ended result instead.

**There is no natural checkpoint for this today.** The coupled pass commits its (partial) traces directly onto the shared `Autorouter`/grid state as it runs (pad marking, `diff_net_ids` bookkeeping at line 6802), and the main strategy (`_neg_strategy` / `route_all_negotiated`) is invoked *on top of* that already-mutated grid via `non_diffpair_strategy()` (line 6854). There is no snapshot/restore primitive at this boundary comparable to `route_all_negotiated`'s internal `_snapshot_failures = copy.deepcopy(...)` (core.py:647) or the fine-grid blocked/net/routes snapshot triplet (core.py:15894-15901) — those exist deeper inside the negotiated loop, not at the `route_all_with_diffpairs` entry/exit.

Building checkpoint-and-compare properly would require either:
1. A full second route (run `route_all_negotiated` single-ended from a clean grid, in addition to the coupled-enabled run) to get a true single-ended baseline to diff against — on board 07 this roughly **doubles wall-clock cost** (the negotiation churn, not the ~1s×7 coupled attempts, is the dominant cost per the epic's fuller-run measurement), or
2. A grid-state snapshot/restore mechanism at the `route_all_with_diffpairs` boundary (snapshot before the pair loop starts, restore if the coupled-enabled path regresses) — a real router-state (grid + `Autorouter` bookkeeping) checkpoint, not a `deepcopy` of routes, since committed pad-marking and net-priority side effects would need to unwind too.

Neither is a small change. **Phase 1 in this issue is scoped to the minimum tier: warn-only + instrumentation.** Checkpoint-and-compare (option b) is real and worth doing, but as a follow-up once the warning ships and the `budget_exit_diff_nets` plumbing already exists to build on.

## Phase 1 scope (this issue)

1. **Surface budget-exit pairs to the CLI.** Return `budget_exit_diff_nets` (or the pair names/count) out of `route_all_with_diffpairs` — either as part of the existing `(routes, warnings)` tuple contract (extend `warnings` with a new warning type distinct from `LengthMismatchWarning`) or as an additional return value threaded through both call sites (`route_cmd.py:9688` and `route_cmd.py:9729`, plus the `route_with_escape_and_diffpairs` composition path in `core.py:14700`).
2. **Emit a clear CLI warning when any pair budget-exits**, naming the specific pairs and stating the regression risk in concrete terms, e.g.:
   ```
   --- Differential Pair Budget-Exit Warning ---
     7 pair(s) budget-exited coupled routing and fell back to single-ended: DQS, MIPI_CLK, MIPI_DAT0, MIPI_DAT1, TMDS_D0, TMDS_D1, TMDS_D2
     WARNING: --differential-pairs can regress completion/DRC vs. a plain single-ended route on bundle-dense boards (see #4095).
     Consider re-routing without --differential-pairs and comparing `kct check` results if this board is dense with matched-length bundles.
   ```
   This should print regardless of `--verbose` (mirroring the existing unconditional-when-`not quiet` `diffpair_warnings` report at `route_cmd.py:10124`), so it isn't accidentally hidden.
3. **Do not change routing behavior.** No fallback-to-single-ended re-route in this issue — that is the "ideal" tier, tracked as follow-up (see below). This issue is purely a visibility/instrumentation fix at the existing (correct) fallback boundary.
4. **File a follow-up issue for checkpoint-and-compare** once this ships, referencing the feasibility notes above (full re-route cost vs. grid-snapshot mechanism) so a future architect pass has the design tradeoffs pre-loaded.

## Acceptance criteria

- [ ] `kct route --differential-pairs` (both the `negotiated` and `basic` strategy dispatch paths, and the escape-composed path) emits a clear, unconditional (`not quiet`) warning when one or more diff pairs budget-exit coupled routing, naming the specific pairs (net/pair names, not just a count).
- [ ] The warning explicitly states the regression risk (coupled attempts can reduce completion / increase DRC errors vs. plain single-ended routing on bundle-dense boards).
- [ ] Board 06 (`boards/06-diffpair-test`, where coupled routing works — 16/16 pours, pairs coupled, zero budget-exits) is unaffected: no new warning appears, routed output and DRC count are unchanged, `tests/test_board_06_diffpair_test.py` and `tests/router/test_board06_determinism.py` continue to pass unmodified.
- [ ] A regression test pins the boundary: given a coupled search that is forced to budget-exit (small `per_pair_max_iterations` / `per_pair_timeout` fixture, mirroring the pattern in `tests/test_diffpair_per_pair_timeout.py` / `tests/test_diffpair_aggregate_cap.py`), the CLI (or `route_all_with_diffpairs` return value directly, whichever is the cheaper/faster test surface) confirms the named pair(s) appear in the new warning output.
- [ ] `budget_exit_diff_nets` (or equivalent) is plumbed out of `route_all_with_diffpairs` without altering existing fallback routing behavior — `tests/test_budget_exit_diff_priority.py` continues to pass unmodified (the net-priority promotion contract is untouched).
- [ ] Out of scope for this issue (tracked as follow-up, not blocking): automatic fallback to a single-ended result when the coupled-enabled result is worse (checkpoint-and-compare). File the follow-up issue referencing this issue's feasibility notes before closing this one.

## Affected Files

- `src/kicad_tools/router/diffpair_routing.py` — `route_all_with_diffpairs` (line ~6567): thread `budget_exit_diff_nets` out as a return value / new warning type instead of discarding it after the net-priority promotion use (line ~6842).
- `src/kicad_tools/router/core.py` — `route_all_with_diffpairs` proxy (line ~13469) and `route_with_escape_and_diffpairs` (line ~14700): forward the new budget-exit signal through the escape-composed dispatch path.
- `src/kicad_tools/cli/route_cmd.py` — dispatch call sites at line ~9688 (`negotiated` strategy) and line ~9729 (`basic` strategy); warning report site at line ~10124 (extend the existing `--- Differential Pair Warnings ---` block or add a sibling `--- Differential Pair Budget-Exit Warning ---` block).
- `tests/test_budget_exit_diff_priority.py` — existing coverage of the `budget_exit_diff_nets` priority-promotion contract; must remain green (regression floor).
- `tests/test_diffpair_per_pair_timeout.py`, `tests/test_diffpair_aggregate_cap.py` — existing budget-exit fixture patterns to model the new regression test on.
- `tests/test_board_07_matchgroup_test.py` — board 07 regression anchor (does not call `--differential-pairs` today per the generator's documented workaround; no change expected here unless the new test suite is added alongside it).
- `tests/test_board_06_diffpair_test.py`, `tests/router/test_board06_determinism.py` — board 06 "unaffected" regression anchors; must remain green.
- `boards/07-matchgroup-test/generate_design.py` (reference only, ~line 1244) — documents the exact regression this issue guards against; no code change expected, cite in the PR description.
- New test file (suggest `tests/test_diffpair_budget_exit_warning.py`) for the regression test in the acceptance criteria.

## Test Plan

- [ ] Automated test: new fixture forces a coupled pair to budget-exit (tiny `per_pair_max_iterations`, e.g. 1-5) on a small synthetic board with a detectable diff pair; assert the pair's name appears in the new warning output / return value.
- [ ] Automated test: `tests/test_budget_exit_diff_priority.py` passes unmodified (confirms no regression to the existing priority-promotion contract).
- [ ] Automated test: `tests/test_board_06_diffpair_test.py` and `tests/router/test_board06_determinism.py` pass unmodified (confirms board 06 — a coupled-success board — sees no new warning and no routing/DRC change).
- [ ] Manual verification: run `uv run kct route boards/07-matchgroup-test/output/*.kicad_pcb --differential-pairs --strategy negotiated --manufacturer jlcpcb --no-auto-layers --layers 4` (or the closest equivalent invocation) and confirm the new warning appears naming the 7 known budget-exit pairs (DQS, MIPI_CLK, MIPI_DAT0, MIPI_DAT1, TMDS_D0, TMDS_D1, TMDS_D2), matching the epic #4049 closeout measurement.
- [ ] Edge case: a board with zero diff pairs detected (`--differential-pairs` set but no pairs found) must not emit the new warning (existing "No differential pairs detected" early-return path at `diffpair_routing.py:6667`/`6668` is unaffected).
- [ ] Edge case: a board where all pairs route coupled successfully (board 06) must not emit the new warning.

rjwalters:
## Summary

The coupled differential-pair pre-pass (`route_all_with_diffpairs`) defers any pair whose `CoupledPathfinder` A* hits its per-pair or aggregate budget to the single-ended main strategy. The fallback is correct — no net goes unrouted — but on bundle-dense boards the coupled attempts can **regress** completion / DRC vs. a plain single-ended route (board 07, epic #4049 closeout: **34 vs 13 DRC errors, 22/31 vs 26/31 nets**). The `budget_exit_diff_nets` set that drives the fallback + the #3270 net-priority promotion was purely local to `route_all_with_diffpairs` — never returned, never distinct from `LengthMismatchWarning`, never surfaced to the CLI.

This is **Phase 1: warn-only + instrumentation.** Checkpoint-and-compare (auto-fallback to the single-ended result when the coupled-enabled result is worse) is an explicit follow-up — it needs either a full second route (~doubles wall-clock) or a grid-state snapshot/restore primitive that does not exist at this boundary today. Scope is a pure visibility/instrumentation fix at the existing (correct) fallback boundary; **routing behavior is unchanged.**

`boards/07-matchgroup-test/generate_design.py` (~line 1244) already documents that board 07's own generator omits `--differential-pairs` for exactly this regression; this change makes the risk visible to any other caller.

## Changes

- **`src/kicad_tools/router/diffpair_routing.py`** — record budget-exited pair base names on the `DiffPairRouter` instance (`_last_budget_exit_pair_names`) plus attempt / exit counters (`_last_coupled_attempted_count`, `_last_budget_exit_count`), reset up front so the surface reflects only the latest call. Emit a greppable `DIFFPAIR_BUDGET_EXIT_FALLBACK` log line for a future checkpoint-and-compare follow-up to key on. The existing local `budget_exit_diff_nets` set still drives the fallback + #3270 promotion unchanged.
- **`src/kicad_tools/router/core.py`** — `Autorouter.diffpair_budget_exit_pair_names()` accessor exposing the surface to the CLI. All dispatch paths (negotiated, basic, escape-composed via `route_with_escape_and_diffpairs`) funnel through the `Autorouter.route_all_with_diffpairs` proxy, so one accessor covers all three.
- **`src/kicad_tools/cli/route_cmd.py`** — collect budget-exit pairs at all four diffpair dispatch sites and print an unconditional (`not quiet`, mirroring the existing length-mismatch block) `--- Differential Pair Budget-Exit Warning ---` naming the pairs and the regression risk.
- **`tests/test_diffpair_budget_exit_warning.py`** (new) — pins the boundary.

| `--differential-pairs` (negotiated, basic, escape-composed) emits an unconditional warning naming budget-exited pairs | ✅ | All 4 dispatch sites collect via the accessor; report gated only on `not quiet`. Manual board-07 run printed the block naming the 7 pairs. |
| Warning states the regression risk (completion/DRC vs. single-ended on dense boards) | ✅ | Warning text: "can regress completion/DRC vs. a plain single-ended route on bundle-dense boards (see #4095)". |
| Board 06 unaffected — no new warning, output/DRC unchanged | ✅ | `test_board_06_diffpair_test.py` + `test_board06_determinism.py` pass unmodified (45 passed, 1 skipped); coupled-success test asserts empty budget-exit surface. |
| Regression test pins the boundary (named pairs surface in the return value/warning) | ✅ | New test forces aggregate budget-exit; asserts pair name in instance attr, accessor, and log line. |
| `budget_exit_diff_nets` plumbed out without altering fallback routing; `test_budget_exit_diff_priority.py` green | ✅ | Passes unmodified; priority-promotion path untouched (still runs off the local set). |
| Checkpoint-and-compare out of scope; file follow-up before closing | ⏳ | Follow-up issue to be filed referencing this PR's feasibility notes. |

## Test Plan

- `pytest tests/test_diffpair_budget_exit_warning.py tests/test_budget_exit_diff_priority.py tests/test_diffpair_per_pair_timeout.py tests/test_diffpair_aggregate_cap.py` — 33 passed.
- `pytest tests/test_board_06_diffpair_test.py tests/router/test_board06_determinism.py` — 45 passed, 1 skipped (board 06 unaffected).
- `pytest` diffpair + escape/composition suites (`test_router_core_diffpair_timeout`, `test_escape_diffpair`, `test_escape_diffpair_composition`, `test_diffpair_routing_integration`, `test_diffpair_phase_b`, `test_diffpair_shadow`, `test_diffpair_npad`) — 147 passed.
- `ruff check` + `ruff format --check` on all changed files — clean.
- Manual: `kct route boards/07-matchgroup-test/output/matchgroup_test.kicad_pcb --differential-pairs --strategy negotiated --manufacturer jlcpcb --no-auto-layers --layers 4 --seed 42` printed the new warning naming the 7 known budget-exit pairs (DQS, MIPI_CLK, MIPI_DAT0, MIPI_DAT1, TMDS_D0, TMDS_D1, TMDS_D2), matching the epic #4049 closeout measurement.

Closes #4095

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -9406,6 +9406,13 @@ def main(argv: list[str] | None = None) -> int:
     # Configure differential pair routing if enabled
     diffpair_config = None
     diffpair_warnings = []
+    # Issue #4095: names of diff pairs that budget-exited coupled routing
+    # and fell back to single-ended.  Accumulated from
+    # ``router.diffpair_budget_exit_pair_names()`` after each
+    # ``--differential-pairs`` dispatch so the report site can warn the
+    # operator (fallback can regress completion / DRC on bundle-dense
+    # boards; see #4095).  De-duplicated at report time.
+    diffpair_budget_exit_pairs: list[str] = []
     if args.differential_pairs:
         diffpair_config = DifferentialPairConfig(
             enabled=True,
@@ -9542,7 +9549,7 @@ def main(argv: list[str] | None = None) -> int:
 
         # Define routing function for profiling
         def do_routing():
-            nonlocal diffpair_warnings, relaxed_nets_report
+            nonlocal diffpair_warnings, relaxed_nets_report, diffpair_budget_exit_pairs
 
             # Adaptive multi-resolution routing (when --grid auto selects it)
             if multi_res_plan is not None and multi_res_plan.is_multi_resolution:
@@ -9625,6 +9632,9 @@ def main(argv: list[str] | None = None) -> int:
                             timeout=_budgeted_timeout(args),
                         )
                         diffpair_warnings.extend(dp_warnings)
+                        # Issue #4095: surface any coupled pairs that
+                        # budget-exited and fell back to single-ended.
+                        diffpair_budget_exit_pairs.extend(router.diffpair_budget_exit_pair_names())
                         return result
                     if args.strategy == "evolutionary":
                         return router.route_all_evolutionary(
@@ -9707,6 +9717,10 @@ def main(argv: list[str] | None = None) -> int:
                         per_net_timeout=getattr(args, "per_net_timeout", None) or None,
                     )
                     diffpair_warnings.extend(dp_warnings)
+                    # Issue #4095: the escape-composed path delegates to the
+                    # same diff-pair orchestrator; surface its budget-exit
+                    # fallback too.
+                    diffpair_budget_exit_pairs.extend(router.diffpair_budget_exit_pair_names())
                     return routes
                 return router.route_with_escape(
                     use_negotiated=(args.strategy == "negotiated"),
@@ -9784,6 +9798,9 @@ def main(argv: list[str] | None = None) -> int:
                     timeout=_budgeted_timeout(args),
                 )
                 diffpair_warnings.extend(dp_warnings)
+                # Issue #4095: surface coupled pairs that budget-exited to
+                # single-ended on the negotiated dispatch path.
+                diffpair_budget_exit_pairs.extend(router.diffpair_budget_exit_pair_names())
                 return result
             elif args.strategy == "negotiated":
                 return router.route_all_negotiated(
@@ -9823,6 +9840,9 @@ def main(argv: list[str] | None = None) -> int:
                     timeout=_budgeted_timeout(args),
                 )
                 diffpair_warnings.extend(dp_warnings)
+                # Issue #4095: surface coupled pairs that budget-exited to
+                # single-ended on the basic dispatch path.
+                diffpair_budget_exit_pairs.extend(router.diffpair_budget_exit_pair_names())
                 return result
             elif args.bus_routing and args.strategy == "basic":
                 return router.route_all_with_buses(bus_config)
@@ -10217,6 +10237,36 @@ def main(argv: list[str] | None = None) -> int:
         print(f"\n--- Differential Pair Warnings ({len(diffpair_warnings)}) ---")
         for warning in diffpair_warnings:
             print(f"  {warning}")
+
+    # Issue #4095: report diff pairs that budget-exited coupled routing and
+    # fell back to single-ended.  Printed unconditionally (only gated on
+    # ``not quiet``, mirroring the length-mismatch block above) so the
+    # regression risk is never accidentally hidden behind --verbose.  On
+    # bundle-dense boards the coupled attempts can regress completion / DRC
+    # vs. a plain single-ended route (board 07: 34 vs 13 DRC errors, 22/31
+    # vs 26/31 nets; epic #4049 closeout), so the operator is told which
+    # pairs fell back and advised to compare a single-ended re-route.
+    if diffpair_budget_exit_pairs and not quiet:
+        # De-duplicate while preserving order (a pair can appear once per
+        # dispatch; the escape path can re-enter the orchestrator).
+        _seen: set[str] = set()
+        _exit_pairs = [
+            name for name in diffpair_budget_exit_pairs if not (name in _seen or _seen.add(name))
+        ]
+        print("\n--- Differential Pair Budget-Exit Warning ---")
+        print(
+            f"  {len(_exit_pairs)} pair(s) budget-exited coupled routing and "
+            f"fell back to single-ended: {', '.join(_exit_pairs)}"
+        )
+        print(
+            "  WARNING: --differential-pairs can regress completion/DRC vs. a "
+            "plain single-ended route on bundle-dense boards (see #4095)."
+        )
+        print(
+            "  Consider re-routing without --differential-pairs and comparing "
+            "`kct check` results if this board is dense with matched-length "
+            "bundles."
+        )
 
     # Report nets that needed clearance relaxation (--progressive-clearance mode)
     if relaxed_nets_report and not quiet:
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -10250,9 +10250,11 @@ def main(argv: list[str] | None = None) -> int:
         # De-duplicate while preserving order (a pair can appear once per
         # dispatch; the escape path can re-enter the orchestrator).
         _seen: set[str] = set()
-        _exit_pairs = [
-            name for name in diffpair_budget_exit_pairs if not (name in _seen or _seen.add(name))
-        ]
+        _exit_pairs: list[str] = []
+        for name in diffpair_budget_exit_pairs:
+            if name not in _seen:
+                _seen.add(name)
+                _exit_pairs.append(name)
         print("\n--- Differential Pair Budget-Exit Warning ---")
         print(
             f"  {len(_exit_pairs)} pair(s) budget-exited coupled routing and "
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -13515,6 +13515,30 @@ class Autorouter:
             return []
         return self._diffpair_router.intra_clearance_violations()
 
+    def diffpair_budget_exit_pair_names(self) -> list[str]:
+        """Names of diff pairs that budget-exited coupled routing (Issue #4095).
+
+        After a ``route_all_with_diffpairs`` /
+        ``route_with_escape_and_diffpairs`` call, returns the base names of
+        every differential pair whose coupled A* hit its per-pair or
+        aggregate budget and was consequently deferred to the single-ended
+        main strategy.  These pairs routed *without* coupling, which on
+        bundle-dense boards can regress completion / DRC vs. a plain
+        single-ended route (board 07: 34 vs 13 DRC errors, 22/31 vs 26/31
+        nets; epic #4049 closeout).  The CLI reads this to emit an
+        unconditional operator warning.
+
+        Returns:
+            A shallow copy of the budget-exit pair names from the most
+            recent ``route_all_with_diffpairs`` call, in detection order.
+            Empty when no diff-pair routing happened, when
+            ``--differential-pairs`` was not enabled, or when every coupled
+            pair converged within budget (e.g. board 06).
+        """
+        if self._diffpair_router is None:
+            return []
+        return list(self._diffpair_router._last_budget_exit_pair_names)
+
     def route_all_with_diffpairs(
         self,
         diffpair_config: DifferentialPairConfig | None = None,
--- a/src/kicad_tools/router/diffpair_routing.py
+++ b/src/kicad_tools/router/diffpair_routing.py
@@ -2730,6 +2730,28 @@ class DiffPairRouter:
         # strategy) from a genuine no-path-found exit (where the
         # caller's existing handling is unchanged).
         self._last_pair_budget_exit: bool = False
+        # Issue #4095: names of the differential pairs whose coupled
+        # search budget-exited (per-pair or aggregate) during the most
+        # recent ``route_all_with_diffpairs`` call and were consequently
+        # deferred to the single-ended main strategy.  Local
+        # ``budget_exit_diff_nets`` bookkeeping drives fallback + the
+        # #3270 net-priority promotion but is discarded at function exit;
+        # this attribute surfaces the same information (by human-readable
+        # pair name) so the CLI can warn the operator that
+        # ``--differential-pairs`` fell back to single-ended routing --
+        # which on bundle-dense boards can *regress* completion / DRC
+        # vs. a plain single-ended route (board 07: 34 vs 13 DRC errors,
+        # 22/31 vs 26/31 nets; epic #4049 closeout).  Reset at the start
+        # of every ``route_all_with_diffpairs`` call so it reflects only
+        # the latest invocation.
+        self._last_budget_exit_pair_names: list[str] = []
+        # Issue #4095 instrumentation: monotonic counters for a future
+        # checkpoint-and-compare follow-up to key on.  ``coupled_attempted``
+        # counts pairs that reached the coupled A* (engaged, within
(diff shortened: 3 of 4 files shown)


rjwalters:
❌ **Changes Requested — Type Check (mypy baseline gate) failing**

The design is right and the implementation faithfully matches the Phase-1 acceptance criteria (behavior-neutral instrumentation, correct reset-up-front semantics, all four dispatch sites collect via the accessor, negative cases covered). Ruff is clean and the behavioral tests pass locally (11 passed: the new budget-exit suite + `test_budget_exit_diff_priority.py` unmodified). But the **Type Check** required gate is red and it is a real, reproducible failure introduced by this PR — not a flake.

**Root cause — 2 NEW mypy errors beyond `.github/mypy-baseline.txt`:**

The order-preserving dedup idiom `not (name in seen or seen.add(name))` trips mypy's `func-returns-value` check because `set.add()` returns `None` and is being used inside a boolean expression:

- `src/kicad_tools/router/diffpair_routing.py:6945`
- `src/kicad_tools/cli/route_cmd.py:10254`

```
error: "add" of "set" does not return a value (it only ever returns None)  [func-returns-value]
```

The logic is *functionally* correct (`add` returns falsy `None`, so the pair is kept only on first sight), but mypy rejects the pattern and CI (`scripts/ci/check_mypy_baseline.py`) treats any new-beyond-baseline error as a hard fail. **Do not add these to the baseline** — the CI script explicitly says to fix them.

**Suggested fix** (replace the one-liner in both spots with an explicit-add comprehension helper, or a plain loop, e.g.):

```python
seen: set[str] = set()
deduped_names: list[str] = []
for name in budget_exit_pair_names:
    if name not in seen:
        seen.add(name)
        deduped_names.append(name)
```

Apply the equivalent in `route_cmd.py` (the `_seen` / `_exit_pairs` block). Then run `uv run python scripts/ci/check_mypy_baseline.py` locally to confirm 0 new errors before pushing.

Note: the mypy baseline check also reports 18 baseline errors that no longer occur — not caused by this PR, so leave the baseline update out of scope here.

Everything else looks good; once Type Check is green (and the pending Board 06/07 regression gates confirm behavior-neutrality) this should approve cleanly.

rjwalters:
Doctor pass complete (finishing a session-limit-interrupted doctor run): replaced the `set.add()` boolean dedup idiom with explicit order-preserving loops at both sites the Judge cited (diffpair_routing.py:6941-6952, route_cmd.py:10250-10258). Local verification: `check_mypy_baseline.py` no longer reports the 2 func-returns-value errors (sole residual is a local-venv `cmaes` import-not-found — optional extra absent locally, present in CI per the prior run), budget-exit warning tests 6/6, ruff clean. No baseline edits, no other changes.

rjwalters:
✅ **Approved!** (Doctor→Judge re-review)

The single blocker from my prior review is resolved. Re-verified both criteria:

**1. Diff since review = exactly the fix.** Commit `c9be5bd6` touches only the two dedup sites, replacing the `not (name in seen or seen.add(name))` idiom with explicit order-preserving for-loops:
- `src/kicad_tools/router/diffpair_routing.py:6941` (budget-exit recording)
- `src/kicad_tools/cli/route_cmd.py:10250` (warning report)

Semantics unchanged (membership check → `seen.add` → append), no baseline edits, no other changes. Branch is `13facd62` (feature) + `c9be5bd6` (fix) only.

**2. CI all green, merge state CLEAN.** The two `func-returns-value` mypy errors are gone — **Type Check: pass**. All previously-pending gates now green:
- Board 05 Routing Regression (blocking <= 11): pass (19m6s)
- Board 06 End-to-End (LVS): pass (16m26s)
- Board 07 End-to-End (LVS known opens): pass (15m38s)
- Diff-Pair Routing Regression: pass (15m21s)
- Match-Group Routing Regression: pass (14m5s)
- Test: pass (15m52s)
- Type Check: pass (36s)

17/17 checks pass, `mergeStateStatus` CLEAN. #4107 follow-up already filed.

Ready to merge. Nice, tight fix.

PR #4109, opened 2026-07-13
fix(mfr): write sibling .kicad_pro + Default netclass on apply-rules

### LINKED ISSUE #4097: kct mfr apply-rules writes .kicad_prl but not .kicad_pro — kicad-cli DRC falls back to factory-default rules
## Original Issue

### Problem

`kct mfr apply-rules <board>.kicad_pcb jlcpcb --layers 4 --copper 1` reports success ("PCB file updated", "Manufacturer set to: JLCPCB") but writes only the `.kicad_pcb`. KiCad stores board design rules (`board.design_settings.rules`: min_clearance, min_track_width, min_via_diameter, min_through_hole_diameter, …) and netclass clearances in the **`.kicad_pro`**, which apply-rules does not create or update when invoked on a `.kicad_pcb` path.

Consequence: `kicad-cli pcb drc` on the "rules-applied" board (standalone, no project file) silently falls back to **KiCad factory defaults** (0.20 mm track/clearance, 0.50 mm via, 0.30 mm hole) and reports phantom violations.

### Measured impact (4-layer board, 458 vias, jlcpcb 4layer_1oz profile = 0.1016 mm track/clearance, 0.45 mm via, 0.2 mm hole)

- Board + apply-rules, no `.kicad_pro`: `kicad-cli pcb drc --refill-zones` → **1175 errors** (199 track_width, 199 via_diameter, 199 drill_out_of_range — i.e. *every* track/via class — plus 499 clearance), all citing "board setup constraints min width 0.2000 mm" etc.
- Same board + a hand-written `.kicad_pro` carrying the profile's rules and a Default netclass clearance of 0.1016: **12 errors** (11 courtyard overlaps + 1 genuine hole_clearance).

In our project this single gap produced a false "board is unmanufacturable / unroutable at 4 layers" conclusion that consumed two full rebuild-and-reroute attempts before the root cause was found.

### Suggested fix

`apply-rules` should create or update the sibling `<board>.kicad_pro`: set `board.design_settings.rules` from the manufacturer profile (correct layer/copper stack key) and set/update netclass clearances (at minimum the Default class) so both KiCad GUI DRC and `kicad-cli pcb drc` enforce the intended tier. A warning when no `.kicad_pro` exists would also have surfaced this immediately.

### Note

Two other subtleties compound the trap: (a) KiCad's Default netclass clearance (0.2 mm) overrides a smaller board minimum, so netclasses must be written too; (b) `min_hole_clearance` has no direct analogue in the profile dataclass and needs a documented mapping.

---

### Correction re: `.kicad_prl`

The title/repro mentions a `.kicad_prl` being written by `apply-rules`. Verified against `origin/main`: `cmd_apply_rules()` in `src/kicad_tools/cli/mfr.py` does **not** write `.kicad_prl` anywhere (confirmed via `rg -n "kicad_prl" src/`; the only `.kicad_prl` writers in the tree are unrelated — `router/partial_rescue.py`'s route-rescue sidecar cleanup and `route_cmd.py`'s output-cleanup allowlist). The `.kicad_prl` the reporter observed was most likely a stale sidecar from a prior KiCad GUI session sitting next to the board, not something `apply-rules` produced. The root cause below is real and independent of the `.kicad_prl` detail — it's about the **missing `.kicad_pro`**, not an incorrectly-written `.kicad_prl`.

### Root cause (verified against `origin/main`)

`cmd_apply_rules()` in `src/kicad_tools/cli/mfr.py` (lines 290-406) branches on the input file suffix:

- **`.kicad_pro` input** (lines 337-371): calls `apply_manufacturer_rules()` + `set_manufacturer_metadata()` (`src/kicad_tools/core/project_file.py`) and `save_project()`. This path *does* write `board.design_settings.rules`, but it only fires when the user passes the `.kicad_pro` path directly — and even then it does **not** touch `net_settings.classes` (the Default netclass clearance), so KiCad's `clearance` DRC test (which reads the *applied netclass* clearance, not `design_settings.rules.min_clearance`) still uses the stock 0.20 mm Default.
- **`.kicad_pcb` input** (lines 374-397, the path in the bug report): calls `_update_zone_clearances()`, which only edits `(zone (connect_pads (clearance ...)))` elements inside the board file itself. No sibling `.kicad_pro` is discovered, created, or updated at all. This is the gap the issue is reporting.

The codebase already has the correct fix pattern implemented and in production use for exactly this failure class (Issue #3919 — same root cause, different call site):

- `src/kicad_tools/manufacturers/project_generator.py`: `write_drc_constraints(pcb_path, rules, manufacturer_id=..., layers=..., copper_oz=...)` — given a `.kicad_pcb` path, derives the sibling `.kicad_pro` path (`pcb_path.with_suffix(".kicad_pro")`), and:
  - If it exists: parses it, calls `merge_project_rules()` (preserves unrelated keys, overwrites `design_settings.rules`/`rule_severities`/`defaults`, and **also updates/creates the Default netclass** clearance/track_width/via_diameter/via_drill — this is the netclass-write note (a) in the original issue, already solved here).
  - If it doesn't exist: calls `build_project_data()` to synthesize a complete minimal project (rules + Default netclass + required top-level keys).
  - Optionally also emits a companion `.kicad_dru` (`write_dru=True` by default) for rule families the project schema can't express.
  - Non-fatal: on `(json.JSONDecodeError, OSError)` for a corrupt existing `.kicad_pro`, falls back to a clean rebuild rather than crashing.
- Callers already wired: `src/kicad_tools/cli/route_cmd.py:1682` (`_write_route_drc_sidecars`, post-route) and `src/kicad_tools/export/manufacturing.py:624` (`ManufacturingPipeline._write_drc_constraints`, pre-export). Both wrap the call in a non-fatal `try/except (ValueError, OSError, KeyError)` that degrades to a warning rather than failing the parent command.
- `build_project_rules()` in the same module already maps `DesignRules.min_hole_to_hole_mm` → the KiCad `min_hole_to_hole` key — this resolves note (b) from the original issue; there is no separate `min_hole_clearance` KiCad project key, `min_hole_to_hole` is the correct analogue and is already handled.

### Suggested implementation

Wire `cmd_apply_rules()`'s `.kicad_pcb` branch (`src/kicad_tools/cli/mfr.py:374-397`) to also call `write_drc_constraints()` after `_update_zone_clearances()`/`save_pcb()`, mirroring the existing `route_cmd.py`/`export/manufacturing.py` call sites:

```python
elif file_path.suffix == ".kicad_pcb":
    ...  # existing zone-clearance + save_pcb logic, unchanged

    from kicad_tools.manufacturers import write_drc_constraints
    try:
        written = write_drc_constraints(
            output_path,
            rules,
            manufacturer_id=profile.id,
            layers=args.layers,
            copper_oz=args.copper,
        )
        if written:
            print(f"DRC-constraint sidecars updated: {', '.join(str(p) for p in written)}")
    except (ValueError, OSError, KeyError) as e:
        print(f"Warning: could not write DRC-constraint sidecars: {e}", file=sys.stderr)
```

Notes for the builder:
- `output_path` here is the *destination* board path (already computed above as `Path(args.output) if args.output else file_path`) — `write_drc_constraints` derives the `.kicad_pro`/`.kicad_dru` paths from it via `.with_suffix(...)`, so the sidecars land next to wherever the PCB was actually written, including `--output` redirection.
- `rules` is already computed earlier in `cmd_apply_rules()` (line 306) from the same profile/layers/copper args — no re-derivation needed.
- The `.kicad_pro` branch (lines 337-371) should also be extended to write/update the Default netclass, since today it only sets `design_settings.rules` via `apply_manufacturer_rules()` and skips `net_settings.classes` entirely (letting the stock 0.20 mm Default clearance silently override a smaller `min_clearance`, per note (a) in the original report). The simplest correct approach: use `add_netclass_definition(data, "Default", track_width=rules.min_trace_width_mm, clearance=rules.min_clearance_mm, via_diameter=rules.min_via_diameter_mm, via_drill=rules.min_via_drill_mm)` from `core/project_file.py` (already implemented, handles update-in-place if "Default" exists) right after `apply_manufacturer_rules()`.
- Do not duplicate `write_drc_constraints`' JSON-writing logic in `mfr.py` — reuse it directly; it already handles create-vs-merge and the Default netclass write for the `.kicad_pcb` input case, which is most of this issue's scope in one call.
- Consider (optional, not required for AC) printing an explicit warning when `apply-rules` is run without ever producing a `.kicad_pro` sidecar for a bare `.kicad_pcb`, since the original report specifically asked for early surfacing of this trap — `write_drc_constraints` always produces one when called, so this becomes moot once the wiring above lands.

### Acceptance Criteria

- [ ] `kct mfr apply-rules <board>.kicad_pcb <mfr> --layers N --copper C` creates a sibling `<board>.kicad_pro` if none exists, with `board.design_settings.rules` populated from the resolved manufacturer profile (matching `write_drc_constraints`'s existing mapping in `project_generator.py`).
- [ ] If `<board>.kicad_pro` already exists, `apply-rules` updates (not replaces) it in place: `design_settings.rules` values are overwritten to the new profile, unrelated top-level keys (e.g. `sheets`, `text_variables`, existing library table refs) are preserved.
- [ ] The `Default` netclass in `net_settings.classes` (`clearance`, `track_width`, `via_diameter`, `via_drill`) is written/updated in both the `.kicad_pcb` input path and the `.kicad_pro` input path — not just one.
- [ ] `--output` redirects the sidecar `.kicad_pro`/`.kicad_dru` to sit next to the redirected output board, not the original input.
- [ ] `--dry-run` does not write any file (existing behavior for `.kicad_pcb`/`.kicad_dru`, extend to cover the new `.kicad_pro` write).
- [ ] Round-trip regression: apply-rules on a `.kicad_pcb` with no existing `.kicad_pro`, then run `kicad-cli pcb drc --refill-zones` (or the repo's `kct check --mfr` equivalent if `kicad-cli` isn't available in the test environment) against the result — the board must not report track_width/via_diameter/drill_out_of_range/clearance violations caused by falling back to KiCad factory defaults.
- [ ] An unknown/unresolvable manufacturer or a read-only output directory degrades to a warning (matching the `route_cmd.py`/`manufacturing.py` precedent), not a crash — `apply-rules`'s existing zone-clearance write must still succeed even if the `.kicad_pro` sidecar write fails.

## Affected Files

- `src/kicad_tools/cli/mfr.py` - `cmd_apply_rules()` (lines ~290-406): wire the `.kicad_pcb` branch to call `write_drc_constraints()` after the existing zone-clearance update; extend the `.kicad_pro` branch to also write the Default netclass via `add_netclass_definition()`.
- `src/kicad_tools/manufacturers/project_generator.py` - reference only; `write_drc_constraints()`/`merge_project_rules()`/`build_project_data()` are the existing, reusable implementation — no changes expected unless a gap is found during implementation.
- `src/kicad_tools/core/project_file.py` - reference only; `apply_manufacturer_rules()` and `add_netclass_definition()` are the existing per-field/per-netclass writers for the `.kicad_pro` input branch.
- `tests/test_mfr.py` - extend `TestApplyRulesCommand` (or the class containing `test_apply_rules_dry_run`, line ~431): add a non-dry-run test that runs `apply-rules` on a `.kicad_pcb` fixture with no pre-existing `.kicad_pro`, then asserts the sidecar `.kicad_pro` exists with the expected `design_settings.rules` values and a Default netclass matching the profile; add a merge-preserves-existing-keys test mirroring `test_merge_preserves_existing_keys_and_relaxes_default_clearance` in `tests/test_drc_constraints_export.py`.
- `tests/test_drc_constraints_export.py` - reference for existing `write_drc_constraints` test patterns (`test_write_drc_constraints_emits_siblings`, `test_route_emits_sibling_kicad_pro`) to mirror in the new `mfr.py` test.

## Test Plan

- [ ] Manual verification: On a demo board fixture (e.g. `boards/03-usb-joystick/usb_joystick.kicad_pcb`, copied to a tmp dir with no `.kicad_pro`), run `kct mfr apply-rules test.kicad_pcb jlcpcb --layers 4 --copper 1`, then inspect `test.kicad_pro` for the correct `board.design_settings.rules.min_track_width` / `min_clearance` / `min_via_diameter` and `net_settings.classes[0]` (Default) matching `jlcpcb.get_design_rules(layers=4, copper_oz=1.0)`.
- [ ] Automated tests: Add to `tests/test_mfr.py` — (1) apply-rules on `.kicad_pcb` with no existing `.kicad_pro` creates one with correct rules + Default netclass; (2) apply-rules on `.kicad_pcb` with an existing `.kicad_pro` merges rules without clobbering unrelated keys; (3) apply-rules on `.kicad_pro` input also writes the Default netclass (currently missing).
- [ ] Round-trip/integration: If `kicad-cli` is available in the dev/CI environment, run `kicad-cli pcb drc --refill-zones` on the post-apply-rules board and assert zero track_width/via_diameter/drill_out_of_range/clearance violations attributable to factory-default fallback (compare against a baseline run with the `.kicad_pro` deleted, which should reproduce the reported ~1175-error false-positive count on a suitably dense board). If `kicad-cli` is not available, use the existing `kct check --mfr` gate as the equivalent cross-check per repo convention (see `MEMORY.md`: "always cross-gate with `kicad-cli pcb drc --refill-zones`" where available).
- [ ] Edge cases: (a) `--output` path redirection places sidecars next to the new output board, not the input; (b) `--dry-run` writes nothing; (c) corrupt/unreadable pre-existing `.kicad_pro` falls back to a clean rebuild rather than crashing (mirrors `write_drc_constraints`'s existing `except (json.JSONDecodeError, OSError)` handling — verify this is preserved through the `mfr.py` call site).

rjwalters:
## Summary

`kct mfr apply-rules <board>.kicad_pcb <mfr>` reported success but only edited
in-board zone clearances — it never produced a sibling `.kicad_pro`. Standalone
`kicad-cli pcb drc` then silently fell back to KiCad factory defaults (0.20 mm
track/clearance, 0.50 mm via, 0.30 mm hole) and reported phantom violations
(the reporter measured ~1175 false errors on a dense 4-layer board). This is
the same failure class as #3919, for which `manufacturers.write_drc_constraints()`
already exists and is wired into `route_cmd.py` and `export/manufacturing.py`.

This PR wires `cmd_apply_rules()` to that existing implementation.

## Changes

- `src/kicad_tools/cli/mfr.py`:
  - `.kicad_pcb` branch: after `save_pcb()`, call `write_drc_constraints(output_path, rules, manufacturer_id=..., layers=..., copper_oz=...)` to create-or-merge the sibling `.kicad_pro` (+ companion `.kicad_dru`). Non-fatal — a sidecar failure degrades to a stderr warning while the zone-clearance write still succeeds (matches the `route_cmd.py` / `export/manufacturing.py` precedent).
  - `.kicad_pro` branch: after `apply_manufacturer_rules()`, also write/update the `Default` netclass via `add_netclass_definition()` so KiCad's `clearance` DRC test (which reads the applied netclass clearance, not `design_settings.rules.min_clearance`) enforces the manufacturer tier instead of the stock 0.20 mm Default.
- `tests/test_mfr.py`: 4 new regression tests (self-contained minimal `.kicad_pcb`, no demo-artifact dependency).

| `.kicad_pcb` input creates sibling `.kicad_pro` with profile rules | ✅ | `test_apply_rules_pcb_creates_kicad_pro` asserts `design_settings.rules` matches `jlcpcb.get_design_rules(4, 1.0)` |
| Existing `.kicad_pro` merged in place, unrelated keys preserved | ✅ | `test_apply_rules_pcb_merges_existing_kicad_pro` keeps `text_variables`/`sheets`, overwrites `min_clearance` |
| Default netclass written in BOTH `.kicad_pcb` and `.kicad_pro` paths | ✅ | `test_apply_rules_pcb_creates_kicad_pro` + `test_apply_rules_pro_input_writes_default_netclass` |
| `--output` redirects sidecars next to output board | ✅ | `test_apply_rules_pcb_output_redirect_places_sidecar` |
| `--dry-run` writes nothing | ✅ | `test_apply_rules_dry_run` extended to assert no `.kicad_pro` |
| Round-trip: no factory-default track/via/clearance fallback | ✅ | Sidecar carries profile `min_track_width`/`min_clearance`/`min_via_diameter` + Default netclass clearance; `write_drc_constraints` mapping is exercised by `test_drc_constraints_export.py` |
| Unknown mfr / read-only dir degrades to warning, not crash | ✅ | `write_drc_constraints` call wrapped in `except (ValueError, OSError, KeyError)` → stderr warning; zone write already committed |
| Corrupt existing `.kicad_pro` falls back to clean rebuild | ✅ | Preserved via `write_drc_constraints`'s existing `except (json.JSONDecodeError, OSError)` |

## Test Plan

- `uv run pytest tests/test_mfr.py tests/test_drc_constraints_export.py` → 71 passed, 2 skipped (pre-existing demo-artifact guards).
- `uv run ruff check` / `ruff format --check` clean on both files.
- `uv run mypy src/kicad_tools/cli/mfr.py` → no issues.

Closes #4097

### DIFF
--- a/src/kicad_tools/cli/mfr.py
+++ b/src/kicad_tools/cli/mfr.py
@@ -290,6 +290,7 @@ def cmd_export_dru(args):
 def cmd_apply_rules(args):
     """Apply manufacturer design rules to a KiCad project or PCB file."""
     from kicad_tools.core.project_file import (
+        add_netclass_definition,
         apply_manufacturer_rules,
         load_project,
         save_project,
@@ -361,6 +362,19 @@ def cmd_apply_rules(args):
             copper_oz=args.copper,
         )
 
+        # Write/update the Default netclass so KiCad's clearance DRC test
+        # (which reads the applied netclass clearance, not
+        # design_settings.rules.min_clearance) enforces the manufacturer tier
+        # instead of the stock 0.20 mm Default (issue #4097).
+        add_netclass_definition(
+            data,
+            "Default",
+            track_width=rules.min_trace_width_mm,
+            clearance=rules.min_clearance_mm,
+            via_diameter=rules.min_via_diameter_mm,
+            via_drill=rules.min_via_drill_mm,
+        )
+
         # Save
         output_path = Path(args.output) if args.output else file_path
         try:
@@ -395,6 +409,31 @@ def cmd_apply_rules(args):
         except Exception as e:
             print(f"Error saving PCB: {e}", file=sys.stderr)
             sys.exit(1)
+
+        # A bare .kicad_pcb has no sibling .kicad_pro, so kicad-cli pcb drc
+        # silently falls back to KiCad factory defaults (0.20 mm track/clearance,
+        # etc.).  Create-or-merge the sibling .kicad_pro (and a companion
+        # .kicad_dru) from the manufacturer profile so both the GUI and
+        # kicad-cli enforce the intended tier, including the Default netclass
+        # clearance (issue #4097).  Non-fatal: the zone-clearance write above
+        # already succeeded, so a sidecar failure only warns.
+        from kicad_tools.manufacturers import write_drc_constraints
+
+        try:
+            written = write_drc_constraints(
+                output_path,
+                rules,
+                manufacturer_id=profile.id,
+                layers=args.layers,
+                copper_oz=args.copper,
+            )
+            if written:
+                print("DRC-constraint sidecars updated: " + ", ".join(str(p) for p in written))
+        except (ValueError, OSError, KeyError) as e:
+            print(
+                f"Warning: could not write DRC-constraint sidecars: {e}",
+                file=sys.stderr,
+            )
 
     else:
         print(f"Error: Unsupported file type: {file_path.suffix}", file=sys.stderr)
--- a/tests/test_mfr.py
+++ b/tests/test_mfr.py
@@ -449,6 +449,142 @@ class TestMfrCLICommands:
         # Should return 0 (success) on dry-run
         assert result is None  # main() doesn't return anything on success
 
+        # dry-run must not write a sibling .kicad_pro (issue #4097)
+        assert not (tmp_path / "test.kicad_pro").exists()
+
+    # Minimal, self-contained .kicad_pcb so these regression tests do not
+    # depend on generated demo-board artifacts (which are not committed).
+    _MINIMAL_PCB = "(kicad_pcb (version 20240108) (generator test))"
+
+    def test_apply_rules_pcb_creates_kicad_pro(self, tmp_path):
+        """apply-rules on a bare .kicad_pcb creates a sibling .kicad_pro.
+
+        Regression for #4097: without a .kicad_pro, kicad-cli pcb drc falls
+        back to KiCad factory defaults.  The sidecar must carry the profile's
+        design_settings.rules and a Default netclass matching the tier.
+        """
+        import json
+
+        from kicad_tools.cli.mfr import main as mfr_main
+        from kicad_tools.manufacturers import get_profile
+
+        test_pcb = tmp_path / "test.kicad_pcb"
+        test_pcb.write_text(self._MINIMAL_PCB)
+
+        pro_path = tmp_path / "test.kicad_pro"
+        assert not pro_path.exists()  # precondition: no project file yet
+
+        mfr_main(["apply-rules", str(test_pcb), "jlcpcb", "--layers", "4", "--copper", "1"])
+
+        assert pro_path.exists(), "sibling .kicad_pro was not created"
+
+        data = json.loads(pro_path.read_text())
+        rules = get_profile("jlcpcb").get_design_rules(layers=4, copper_oz=1.0)
+
+        proj_rules = data["board"]["design_settings"]["rules"]
+        assert proj_rules["min_clearance"] == pytest.approx(rules.min_clearance_mm)
+        assert proj_rules["min_track_width"] == pytest.approx(rules.min_trace_width_mm)
+        assert proj_rules["min_via_diameter"] == pytest.approx(rules.min_via_diameter_mm)
+
+        # Default netclass must carry the profile clearance, not the stock 0.20
+        classes = data["net_settings"]["classes"]
+        default_cls = next(c for c in classes if c.get("name") == "Default")
+        assert default_cls["clearance"] == pytest.approx(rules.min_clearance_mm)
+        assert default_cls["track_width"] == pytest.approx(rules.min_trace_width_mm)
+        assert default_cls["via_diameter"] == pytest.approx(rules.min_via_diameter_mm)
+        assert default_cls["via_drill"] == pytest.approx(rules.min_via_drill_mm)
+
+    def test_apply_rules_pcb_merges_existing_kicad_pro(self, tmp_path):
+        """apply-rules updates an existing .kicad_pro without clobbering keys."""
+        import json
+
+        from kicad_tools.cli.mfr import main as mfr_main
+        from kicad_tools.manufacturers import get_profile
+
+        test_pcb = tmp_path / "test.kicad_pcb"
+        test_pcb.write_text(self._MINIMAL_PCB)
+
+        # Pre-existing project with an unrelated key that must survive.
+        pro_path = tmp_path / "test.kicad_pro"
+        pro_path.write_text(
+            json.dumps(
+                {
+                    "board": {"design_settings": {"rules": {"min_clearance": 0.05}}},
+                    "text_variables": {"MY_VAR": "keep-me"},
+                    "sheets": [["abc", "Root"]],
+                }
+            )
+        )
+
+        mfr_main(["apply-rules", str(test_pcb), "jlcpcb", "--layers", "4", "--copper", "1"])
+
+        data = json.loads(pro_path.read_text())
+        rules = get_profile("jlcpcb").get_design_rules(layers=4, copper_oz=1.0)
+
+        # Unrelated keys preserved.
+        assert data["text_variables"] == {"MY_VAR": "keep-me"}
+        assert data["sheets"] == [["abc", "Root"]]
+
+        # Rules overwritten to the new profile (not left at 0.05).
+        assert data["board"]["design_settings"]["rules"]["min_clearance"] == pytest.approx(
+            rules.min_clearance_mm
+        )
+        default_cls = next(c for c in data["net_settings"]["classes"] if c.get("name") == "Default")
+        assert default_cls["clearance"] == pytest.approx(rules.min_clearance_mm)
+
+    def test_apply_rules_pcb_output_redirect_places_sidecar(self, tmp_path):
+        """--output places the .kicad_pro next to the redirected board."""
+        from kicad_tools.cli.mfr import main as mfr_main
+
+        test_pcb = tmp_path / "input.kicad_pcb"
+        test_pcb.write_text(self._MINIMAL_PCB)
+
+        out_dir = tmp_path / "out"
+        out_dir.mkdir()
+        out_pcb = out_dir / "redirected.kicad_pcb"
+
+        mfr_main(
+            [
+                "apply-rules",
+                str(test_pcb),
+                "jlcpcb",
+                "--layers",
+                "4",
+                "--output",
+                str(out_pcb),
+            ]
+        )
+
+        # Sidecar lands next to the redirected board, not the input.
+        assert (out_dir / "redirected.kicad_pro").exists()
+        assert not (tmp_path / "input.kicad_pro").exists()
+
+    def test_apply_rules_pro_input_writes_default_netclass(self, tmp_path):
+        """apply-rules on a .kicad_pro input also writes the Default netclass.
+
+        Previously the .kicad_pro branch set design_settings.rules but never
+        touched net_settings.classes, so KiCad's clearance test kept the stock
+        0.20 mm Default (issue #4097).
+        """
+        import json
+
+        from kicad_tools.cli.mfr import main as mfr_main
+        from kicad_tools.manufacturers import get_profile
+
+        pro_path = tmp_path / "test.kicad_pro"
+        pro_path.write_text(json.dumps({"meta": {"version": 1}}))
+
+        mfr_main(["apply-rules", str(pro_path), "jlcpcb", "--layers", "4", "--copper", "1"])
+
+        data = json.loads(pro_path.read_text())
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Clean wiring fix that reuses the proven `write_drc_constraints()` helper (#3919) rather than re-implementing JSON plumbing. Verified against all review points:

**1. Round-trip / kicad-cli coverage.** The `kicad-cli Round-trip Smoke` gate (`tests/test_kicad_cli_roundtrip.py`) is a general suite and does not exercise `mfr apply-rules`, so the new tests assert `.kicad_pro` content *structurally* — the exact `board.design_settings.rules` and Default-netclass values that kicad-cli reads. This is acceptable: those keys are precisely the factory-default fallback source the issue identified, and `write_drc_constraints`' kicad-cli-consumed mapping is separately covered in `test_drc_constraints_export.py` (already production-validated via the route/manufacturing callers). A live kicad-cli DRC assertion would require a committed dense-board fixture the repo intentionally doesn't carry — the structural assertion is the right level here.

**2. Non-fatal wrap is correct.** The primary job (zone-clearance edits + `save_pcb()`) commits *before* the sidecar write; failing hard afterward would abort with the board already modified. It does not mask the bug — the normal path always writes the sidecar; the `except (ValueError, OSError, KeyError)` only fires on genuine IO/resolution failures (read-only dir, unresolvable mfr), which the AC explicitly wants degraded to a warning. Matches the `route_cmd.py`/`export/manufacturing.py` precedent.

**3. Merge semantics exercised.** `test_apply_rules_pcb_merges_existing_kicad_pro` pre-seeds a `.kicad_pro` with `text_variables`/`sheets`, then asserts they survive while `min_clearance` and the Default netclass are overwritten. Confirmed `merge_project_rules()` (project_generator.py:233) uses `setdefault`/`update` to preserve unrelated keys.

**4. Diff hygiene.** Only `src/kicad_tools/cli/mfr.py` (+39) and `tests/test_mfr.py` (+136). `add_netclass_definition()` call signature and all `DesignRules` field names (`min_trace_width_mm`/`min_clearance_mm`/`min_via_diameter_mm`/`min_via_drill_mm`) verified against source.

**5. CI: all 16 checks pass, merge state CLEAN.**

## Test Execution
- `uv run pytest tests/test_mfr.py tests/test_drc_constraints_export.py` → 67 passed, 2 skipped (pre-existing demo-artifact guards). ✅
- Full CI: Boards 00–07, routing/diff-pair/match-group regressions, Test, Type Check, Lint & Format, C++ Build, kicad-cli Round-trip Smoke, Routed PCB DRC — all pass. ✅

Both `.kicad_pcb` and `.kicad_pro` input branches now write the Default netclass; the self-contained minimal-PCB test fixture (vs. skipping on missing demo artifacts) is a justified, well-documented deviation. Solid work.

PR #4110, opened 2026-07-13
fix(sync-netlist): read gr_poly/gr_curve outlines and warn on off-board placement

### LINKED ISSUE #4098: kct pcb sync-netlist places newly-added footprints outside the board outline with no warning

## Problem

`kct pcb sync-netlist --remove-orphans --force --auto-rename` adds footprints that exist in the schematic but not the PCB at default coordinates that can land **outside the Edge.Cuts outline**, with no warning.

On our 65×75 mm board (outline x∈[116.5, 181.5] in absolute coords), three newly-added parts (a SOT-353 buffer and two 0402 caps) were dropped at x≈191.5/196.5/201.5 — 10–20 mm off-board.

## Impact

The off-board parts made their nets physically unroutable, and every incomplete net on the board classified as PLACEMENT_BOUND in `kct net-status --why`. Downstream, this presented as "the board cannot be routed at 4 layers" and triggered a full EE review plus a multi-hour placement-optimization effort (including a CMA-ES `optimize-placement` run that concluded INFEASIBLE) — all chasing a phantom density wall whose root cause was three parts sitting in space. The off-board placement was only spotted when a reviewer compared footprint coordinates against the outline by hand.

## Suggested fix (any of these would have caught it)

1. Place newly-added footprints at a position derived from the board outline (e.g. centroid or a staging row just inside the edge), not at an arbitrary origin offset.
2. Emit a prominent warning whenever sync-netlist leaves any footprint outside Edge.Cuts (this check is cheap: bbox-vs-outline).
3. Add an off-board-footprint check to `kct check` preflight (it already parses the outline for board_dimensions), so the condition can't survive to routing.

## Repro sketch

Schematic with a symbol absent from the PCB; board whose outline is not near the origin-offset default; run `kct pcb sync-netlist`; inspect the new footprint's `(at ...)` vs Edge.Cuts.

---

### Root cause (verified against origin/main)

Two independent bugs compound to produce off-board placement with zero diagnostics:

**Bug A — `PCB.get_board_outline()` silently drops polygon/curve outlines.**

`src/kicad_tools/schema/pcb.py::get_board_outline()` (around line 2853) only
consumes `self._graphic_lines`, `self._graphic_arcs`, and `gr_rect` graphics
when building the Edge.Cuts polygon. The S-expression parse loop
(`src/kicad_tools/schema/pcb.py`, around line 1745) has no `elif tag ==
"gr_poly"` (or `"gr_curve"`) branch at all, so `gr_poly`/`gr_curve` Edge.Cuts
elements — common for rounded corners, chamfers, and any outline drawn with
KiCad's polygon tool — are parsed into `self._graphics` generically but never
reach `_graphic_lines`/`_graphic_arcs`. For a board whose outline is a
`gr_poly`, `get_board_outline()` returns `[]`.

Contrast with `PCB._edge_cuts_bbox_sexp()` (around line 2244), which *does*
correctly walk `gr_rect`/`gr_line`/`gr_arc`/`gr_circle`/`gr_poly`/`gr_curve`
(including each element's `(pts (xy ...))` children) directly off
`self._sexp`. This helper already exists and is used by `page_fit`
(`PCB.board_dimensions`, around line 2403) — it is the correct outline
source and is coordinate-space-labeled (sheet-absolute) in its own
docstring.

Reproduction (verified locally against a polygon-free rectangular board vs.
mentally tracing the `gr_poly` case — the parse loop has no handler, so any
board whose Edge.Cuts is emitted as `gr_poly` reproduces `outline == []`
directly from the parser code path):

```python
pcb = PCB.load("board_with_gr_poly_outline.kicad_pcb")
pcb.get_board_outline()  # -> [] even though the board clearly has an outline
```

**Bug B — the `[]` fallback in `_get_board_edge_position` is silent and
lands parts at the wrong reference point.**

`src/kicad_tools/cli/pcb_sync_netlist.py::_get_board_edge_position()`
(around line 608):

```python
def _get_board_edge_position(pcb) -> tuple[float, float]:
    outline = pcb.get_board_outline()
    if outline:
        max_x = max(pt[0] for pt in outline)
        min_y = min(pt[1] for pt in outline)
        return (max_x + 10.0, min_y)
    return (0.0, 0.0)
```

When `get_board_outline()` returns `[]` (Bug A, or any board with no
detectable Edge.Cuts geometry at all), this falls back to board-relative
`(0.0, 0.0)`. `PCB.add_footprint_from_file()` then converts that to
sheet-absolute coordinates via `abs_x = x + self._board_origin[0]`
(`src/kicad_tools/schema/pcb.py`, around line 4101). For a board whose
`_board_origin` is far from the sheet origin (any board not created at
`(0, 0)`, i.e. most real boards including `PCB.create(center=True)` and
boards laid out anywhere else on the KiCad drawing sheet), the new footprint
lands at `_board_origin` itself — which may or may not be inside the
outline depending on outline shape, and is placed with **zero relationship**
to the outline's actual extent once the outline can't be read. There is no
warning either way: `sync_netlist()`'s `SyncResult.warnings` list is never
populated for placement, and `format_text`/`format_json` have no
"footprints placed outside Edge.Cuts" section.

Even in the *working* case (rectangular `gr_rect`/`gr_line` outline, Bug A
not triggered), the current design intentionally staples new parts 10mm
outside the board edge (`max_x + 10.0`) as a "staging" convention — visible,
deliberate, and fine for a human doing an interactive sync, but it is easy
to miss when synced non-interactively (`--auto-rename --force`, as in the
repro) and, worse, is indistinguishable from the Bug-A failure mode (both
put footprints outside the outline) without an explicit check.

### Scoped fix

This issue folds the reporter's suggestions #1 and #2 into one buildable
fix; #3 (a `kct check` preflight rule) is a natural follow-up but is left
out of scope here to keep this a single, reviewable PR — file a follow-up
issue for it if desired once this lands (the preflight module,
`src/kicad_tools/export/preflight.py`, already has a `_check_board_dimensions`
pattern to model it on).

1. **Fix `get_board_outline()` to include `gr_poly`/`gr_curve` Edge.Cuts
   geometry**, mirroring the parsing already done in `_edge_cuts_bbox_sexp()`
   (walk each polygon's `(pts (xy ...))` children, plus `start`/`end`/`mid`/
   `center` for line/arc/circle elements). This is the minimal, most direct
   fix to Bug A — it makes the outline-derived placement path work for the
   common real-world outline shapes it currently silently fails on.
   Alternatively/additionally, `_get_board_edge_position()` could call
   `_edge_cuts_bbox_sexp()` instead of `get_board_outline()` since the former
   already handles every outline element type — but note it returns
   sheet-absolute coordinates, so the caller would need to subtract
   `pcb._board_origin` before passing `x`/`y` into `add_footprint`, which
   re-adds it. Prefer extending `get_board_outline()` itself so the existing
   board-relative contract (and its test suite in
   `TestGetBoardEdgePositionCoordinates`) stays intact.

2. **After outline detection still yields nothing** (e.g. a board with no
   Edge.Cuts layer at all — a genuinely edge-case board), keep the
   `(0.0, 0.0)` fallback but make it loud: append a `SyncAction`/warning to
   `SyncResult.warnings` such as `"No Edge.Cuts outline detected; new
   footprints placed at board origin (0, 0) — verify placement manually."`
   surfaced in both `format_text()` and `format_json()`.

3. **Add a bbox-vs-outline check after placement** in `sync_netlist()`
   (immediately after the `pcb.add_footprint(...)` call in the `for action
   in result.added:` loop, around line 335-346): compute the outline bbox
   (reusing the now-fixed `get_board_outline()` or `_edge_cuts_bbox_sexp()`,
   consistently converted to the same coordinate space as the newly-placed
   footprint) and compare against the new footprint's position. If outside
   the outline's bounding box, append a warning to `result.warnings`, e.g.
   `f"{action.reference} placed outside board outline at ({x}, {y}) — nets
   may be unroutable"`. This check should fire regardless of *why* the
   footprint ended up outside (covers both Bug A/B and the legitimate
   "10mm staging" placement when it's not immediately followed by a manual
   placement pass), directly addressing the reporter's suggestion #2.

### Acceptance Criteria

- [ ] `PCB.get_board_outline()` returns a non-empty polygon for boards whose
      Edge.Cuts layer is expressed as `gr_poly` (and `gr_curve`), not just
      `gr_rect`/`gr_line`/`gr_arc`.
- [ ] `kct pcb sync-netlist` run on a board with a `gr_poly` Edge.Cuts
      outline and a schematic component missing from the PCB places the new
      footprint's `(at ...)` inside (or immediately adjacent to, per the
      existing staging convention) the true outline bounding box — not at
      `_board_origin` / sheet-origin-derived coordinates unrelated to the
      outline.
- [ ] When `sync_netlist()` cannot determine any Edge.Cuts outline at all,
      `SyncResult.warnings` contains an explicit message about the
      `(0.0, 0.0)` fallback, and this warning is printed by both
      `format_text()` and `format_json()`.
- [ ] When a newly-added footprint's placed position falls outside the
      detected board outline's bounding box, `SyncResult.warnings` contains
      a per-footprint warning naming the reference and its coordinates, and
      it is printed by both `format_text()` and `format_json()`.
- [ ] Existing behavior for the passing rectangular-outline case
      (`TestGetBoardEdgePositionCoordinates`, `TestGetBoardEdgePosition` in
      `tests/test_pcb_sync_netlist.py`) is unchanged — no new warning fires
      when placement is correctly outline-derived and within bounds (the
      10mm-outside "staging" convention should either be reconsidered as
      part of this fix or explicitly exempted from the new warning; pick
      whichever keeps the existing staging tests green while still catching
      the Bug A/B failure modes — e.g. only warn when the footprint is
      outside a generously padded bbox derived from board_origin-consistent
      coordinates, not just "outside the exact outline").

### Affected Files

- `src/kicad_tools/schema/pcb.py` - fix `get_board_outline()` (~line 2853) to
  include `gr_poly`/`gr_curve` Edge.Cuts elements, matching the parsing
  already implemented in `_edge_cuts_bbox_sexp()` (~line 2244); the S-expr
  parse loop (~line 1745) may also need a `gr_poly`/`gr_curve` branch feeding
  a new or existing internal collection if `get_board_outline()` continues
  to source from `_graphic_lines`/`_graphic_arcs` rather than re-walking
  `self._sexp`.
- `src/kicad_tools/cli/pcb_sync_netlist.py` - `_get_board_edge_position()`
  (~line 608) and `sync_netlist()` (~line 83, specifically the "Add missing
  footprints at board edge" block ~line 332-346) to add the post-placement
  outline-bbox warning; `format_text()` (~line 624) and `format_json()`
  (~line 692) already have a `warnings` rendering path that the new
  messages will flow through automatically once appended to
  `result.warnings`.
- `tests/test_pcb_sync_netlist.py` - extend `TestGetBoardEdgePosition` /
  `TestGetBoardEdgePositionCoordinates` (~line 377, ~line 1707) with a
  `gr_poly`-outline fixture; add a new test class covering the
  off-board-warning behavior for both the "no outline at all" and
  "footprint placed outside detected outline" cases.

### Test Plan

- [ ] Automated test: `PCB.load()` a fixture `.kicad_pcb` whose Edge.Cuts
      outline is a single `gr_poly` (rounded/chamfered rectangle), assert
      `get_board_outline()` returns the correct non-empty polygon with the
      expected bounding box.
- [ ] Automated test: run `sync_netlist()` against the `gr_poly`-outline
      fixture with a schematic component missing from the PCB, assert the
      added footprint's position (via `pcb.footprints[...].position` or by
      reloading the saved PCB) falls within/near the outline bbox rather
      than at raw `_board_origin`.
- [ ] Automated test: mock `get_board_outline()` to return `[]` (simulating
      "no Edge.Cuts geometry detected at all"), assert `SyncResult.warnings`
      contains a fallback-placement warning and that it appears in both
      `format_text()` and `format_json()` output.
- [ ] Automated test: force a newly-added footprint's placement outside a
      known outline bbox (e.g. via a monkeypatched `_get_board_edge_position`
      returning an out-of-bounds point), assert `SyncResult.warnings`
      contains a per-reference "placed outside board outline" warning.
- [ ] Regression: `uv run pytest tests/test_pcb_sync_netlist.py -v` — all
      existing tests (including `TestGetBoardEdgePosition` and
      `TestGetBoardEdgePositionCoordinates`) continue to pass unchanged.
- [ ] Manual verification: reconstruct the reporter's scenario (a board
      whose outline uses `gr_poly`, `_board_origin` far from `(0, 0)`, and a
      schematic component missing from the PCB) and confirm
      `kct pcb sync-netlist --auto-rename --force` either places the new
      footprint sensibly relative to the outline or emits a clear warning —
      and that `kct net-status --why` no longer misattributes the resulting
      nets as a routing-density failure.

rjwalters:
## Summary

`kct pcb sync-netlist` could place newly-added footprints outside the board
outline with no warning, making their nets physically unroutable and
presenting downstream as a phantom routing-density wall (issue #4098). Two
compounding defects caused this:

- **Bug A:** `PCB.get_board_outline()` only parsed `gr_line`/`gr_arc`/`gr_rect`
  Edge.Cuts geometry, so a board whose outline is a `gr_poly`/`gr_curve`
  (rounded corners, chamfers, polygon-tool outlines) silently returned `[]`.
- **Bug B:** `_get_board_edge_position()` then fell back to board origin
  `(0, 0)` — unrelated to the true outline — and neither the fallback nor the
  resulting off-board placement produced any warning.

## Changes

- `src/kicad_tools/schema/pcb.py`
  - Extend `get_board_outline()` to recover `gr_poly`/`gr_curve` Edge.Cuts
    vertex chains, stitched through the existing segment-ordering path and
    board-origin conversion (keeps the board-relative contract intact).
  - Add `_edge_cuts_poly_chains_sexp()` helper that walks `self._sexp`
    directly, mirroring the existing `_edge_cuts_bbox_sexp()`.
- `src/kicad_tools/cli/pcb_sync_netlist.py`
  - `sync_netlist()` now emits an explicit `SyncResult.warnings` message when
    no Edge.Cuts outline is detected (the `(0, 0)` origin fallback).
  - Adds a per-reference "placed outside board outline" warning when a
    newly-added footprint lands outside the outline bbox, padded by
    `_OUTLINE_STAGING_MARGIN_MM` (25mm) to exempt the deliberate ~10mm staging
    offset. New `_outline_bbox()` / `_is_outside_bbox()` helpers.
  - Both warnings flow through the existing `format_text()`/`format_json()`
    rendering paths (no formatter changes needed).
- `tests/test_pcb_sync_netlist.py` — new test classes for gr_poly/gr_curve
  outline parsing, bbox helpers, and the off-board/no-outline warning behavior.

Out of scope (per curation): the reporter's suggestion #3 (a `kct check`
preflight off-board rule) is left as a follow-up.

| `get_board_outline()` returns non-empty polygon for `gr_poly`/`gr_curve` outlines | ✅ | `TestGetBoardOutlineGrPoly` (both element types), bbox asserted |
| sync places new footprint relative to the real outline (not origin) on a `gr_poly` board | ✅ | `test_get_board_edge_position_uses_gr_poly_outline` (staged at max_x+10, not `(0,0)`) |
| No-outline case emits explicit fallback warning in text + json | ✅ | `test_no_outline_fallback_emits_warning` |
| Off-board placement emits per-reference warning in text + json | ✅ | `test_off_board_placement_emits_per_reference_warning` |
| Existing rectangular-outline / staging behavior unchanged (no spurious warning) | ✅ | `TestGetBoardEdgePosition*` still pass; `test_in_bounds_staging_does_not_warn` |

## Test Plan

- `uv run pytest tests/test_pcb_sync_netlist.py` — 128 passed (118 pre-existing + 10 new).
- Outline-related tests across suite: 164 passed, 6 skipped.
- `mypy` introduces zero new errors (15 pre-existing errors in `pcb.py` unchanged).

Closes #4098

### DIFF
--- a/src/kicad_tools/cli/pcb_sync_netlist.py
+++ b/src/kicad_tools/cli/pcb_sync_netlist.py
@@ -331,19 +331,43 @@ def sync_netlist(
 
             # Add missing footprints at board edge
             placement_x, placement_y = _get_board_edge_position(pcb)
+            outline_bbox = _outline_bbox(pcb)
+            if outline_bbox is None:
+                # No Edge.Cuts geometry detected at all: _get_board_edge_position
+                # fell back to the board origin (0, 0), which has no relationship
+                # to the true outline extent.  Surface this loudly so it is not
+                # silently mistaken for a routing-density failure downstream.
+                result.warnings.append(
+                    "No Edge.Cuts outline detected; new footprints placed at "
+                    "board origin (0, 0) — verify placement manually."
+                )
             x_offset = 0.0
             for action in result.added:
+                place_x = placement_x + x_offset
+                place_y = placement_y
                 try:
                     pcb.add_footprint(
                         library_id=action.footprint,
                         reference=action.reference,
-                        x=placement_x + x_offset,
-                        y=placement_y,
+                        x=place_x,
+                        y=place_y,
                         value=action.value,
                     )
                     x_offset += 5.0  # Space footprints horizontally
                 except Exception as e:
                     result.errors.append(f"Failed to add footprint for {action.reference}: {e}")
+                    continue
+                # Warn if the placed position falls outside the board outline's
+                # bounding box (padded to exempt the deliberate ~10mm staging
+                # offset).  This fires regardless of *why* a part ended up
+                # off-board — a mis-read outline, an origin fallback, or a
+                # staging placement never followed by a manual placement pass —
+                # so the off-board condition can't survive silently to routing.
+                if outline_bbox is not None and _is_outside_bbox(place_x, place_y, outline_bbox):
+                    result.warnings.append(
+                        f"{action.reference} placed outside board outline at "
+                        f"({place_x:.3f}, {place_y:.3f}) — nets may be unroutable"
+                    )
 
             # Remove orphaned footprints
             for action in result.removed:
@@ -619,6 +643,49 @@ def _get_board_edge_position(pcb) -> tuple[float, float]:
         # get_board_outline() already returns board-relative coords.
         return (max_x + 10.0, min_y)
     return (0.0, 0.0)
+
+
+# Padding (mm) added around the outline bbox before flagging a placement as
+# "off-board".  Must exceed the deliberate ~10mm staging offset applied by
+# _get_board_edge_position() (plus a little headroom for the 5mm-per-part
+# horizontal spacing) so that legitimate staging placement does NOT warn,
+# while a genuinely mis-placed part (mis-read outline, origin fallback) does.
+_OUTLINE_STAGING_MARGIN_MM = 25.0
+
+
+def _outline_bbox(pcb) -> tuple[float, float, float, float] | None:
+    """Return the board outline bounding box in board-relative coordinates.
+
+    Uses :meth:`PCB.get_board_outline` (which now covers ``gr_poly``/``gr_curve``
+    outlines as well as ``gr_line``/``gr_arc``/``gr_rect``) so the bbox is in
+    the same coordinate space as the board-relative positions passed to
+    :meth:`PCB.add_footprint`.
+
+    Returns:
+        ``(min_x, min_y, max_x, max_y)`` or ``None`` when no outline was found.
+    """
+    outline = pcb.get_board_outline()
+    if not outline:
+        return None
+    xs = [pt[0] for pt in outline]
+    ys = [pt[1] for pt in outline]
+    return (min(xs), min(ys), max(xs), max(ys))
+
+
+def _is_outside_bbox(
+    x: float,
+    y: float,
+    bbox: tuple[float, float, float, float],
+    margin: float = _OUTLINE_STAGING_MARGIN_MM,
+) -> bool:
+    """Return True if ``(x, y)`` lies outside ``bbox`` padded by ``margin``.
+
+    The padding exempts the deliberate staging offset applied to newly-added
+    footprints (see :func:`_get_board_edge_position`) so only genuinely
+    off-board placements are flagged.
+    """
+    min_x, min_y, max_x, max_y = bbox
+    return x < min_x - margin or x > max_x + margin or y < min_y - margin or y > max_y + margin
 
 
 def format_text(result: SyncResult, dry_run: bool, pcb_path: Path) -> str:
--- a/src/kicad_tools/schema/pcb.py
+++ b/src/kicad_tools/schema/pcb.py
@@ -2291,6 +2291,41 @@ class PCB:
             return None
         return (min(xs), min(ys), max(xs), max(ys))
 
+    def _edge_cuts_poly_chains_sexp(self) -> list[list[tuple[float, float]]]:
+        """Collect ``gr_poly``/``gr_curve`` Edge.Cuts vertex chains.
+
+        ``gr_poly`` and ``gr_curve`` graphics are not parsed into the in-memory
+        ``_graphic_lines``/``_graphic_arcs``/``_graphics`` collections, so this
+        walks ``self._sexp`` directly (like :meth:`_edge_cuts_bbox_sexp`) and
+        returns each polygon/curve's ordered ``(pts (xy ...))`` vertex list in
+        **sheet-absolute** coordinates.
+
+        Returns:
+            A list of vertex chains; each chain is an ordered list of
+            ``(x, y)`` tuples.  Empty if no ``gr_poly``/``gr_curve`` Edge.Cuts
+            geometry is present.
+        """
+
+        def _on_edge_cuts(node: SExp) -> bool:
+            layer_node = node.find_child("layer")
+            return bool(layer_node and layer_node.get_string(0) == "Edge.Cuts")
+
+        chains: list[list[tuple[float, float]]] = []
+        for child in self._sexp.iter_children():
+            if child.tag not in ("gr_poly", "gr_curve"):
+                continue
+            if not _on_edge_cuts(child):
+                continue
+            chain: list[tuple[float, float]] = []
+            for pts in child.find_all("pts"):
+                for xy in pts.find_children("xy"):
+                    x, y = xy.get_float(0), xy.get_float(1)
+                    if x is not None and y is not None:
+                        chain.append((x, y))
+            if chain:
+                chains.append(chain)
+        return chains
+
     @staticmethod
     def _translate_coord_node(node: SExp, dx_nm: int, dy_nm: int) -> None:
         """Translate the first two atoms (X, Y) of a coordinate node in place.
@@ -2854,8 +2889,18 @@ class PCB:
         """Extract board outline polygon from Edge.Cuts layer.
 
         Returns an ordered list of (x, y) points forming the board outline.
-        Handles gr_line, gr_arc, and gr_rect elements on the Edge.Cuts layer.
-        Arc segments are approximated by their start and end points.
+        Handles ``gr_line``, ``gr_arc``, ``gr_rect``, ``gr_poly``, and
+        ``gr_curve`` elements on the Edge.Cuts layer.  Arc segments are
+        approximated by their start and end points; polygon/curve outlines
+        contribute their vertex chain (``(pts (xy ...))``).
+
+        ``gr_poly``/``gr_curve`` Edge.Cuts geometry is common for rounded
+        corners, chamfers, and any outline drawn with KiCad's polygon tool.
+        Those elements are not surfaced through the in-memory
+        ``_graphic_lines``/``_graphic_arcs``/``_graphics`` collections, so they
+        are recovered by walking ``self._sexp`` directly (mirroring
+        :meth:`_edge_cuts_bbox_sexp`).  Without this, a board whose outline is
+        a single ``gr_poly`` would silently return ``[]``.
 
         Returns:
             List of (x, y) coordinate tuples in mm. Empty list if no outline found.
@@ -2866,8 +2911,11 @@ class PCB:
         edge_rects = [
             g for g in self._graphics if g.layer == "Edge.Cuts" and g.graphic_type == "rect"
         ]
+        # gr_poly / gr_curve are not parsed into the in-memory collections, so
+        # recover their vertex chains straight off the S-expression tree.
+        poly_chains = self._edge_cuts_poly_chains_sexp()
 
-        if not edge_lines and not edge_arcs and not edge_rects:
+        if not edge_lines and not edge_arcs and not edge_rects and not poly_chains:
             return []
 
         # Build a list of all line segments (including arc endpoints)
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

Clean, well-scoped two-part fix for #4098. All acceptance criteria met, all 16 CI checks green, merge state CLEAN.

## Verification

**CI:** all 16 checks pass — Test (15m33s), Type Check, Lint & Format, C++ Build, Routed DRC, kicad-cli Round-trip, Boards 00-07 E2E, Board 05 Routing Regression, Diff-Pair & Match-Group regressions. Merge state CLEAN.

**Local (worktree issue-4098):** `pytest tests/test_pcb_sync_netlist.py` → 128 passed (118 pre-existing + 10 new). `ruff check` + `ruff format --check` clean on all three changed files.

## Review findings

1. **Geometry (`_edge_cuts_poly_chains_sexp`)** — faithfully mirrors the existing `_edge_cuts_bbox_sexp()` (pcb.py:2244): same `_on_edge_cuts` predicate, same `(pts (xy ...))` walk, same sheet-absolute coordinate contract. Chains feed the shared segment-stitching path, and the `len(chain) >= 3` guard correctly avoids emitting a bogus closing edge for degenerate 2-point chains. Board-relative conversion is preserved downstream, so the board-relative contract that `_get_board_edge_position` / `add_footprint` rely on stays intact.

2. **gr_curve (Bezier)** — control points are treated as polyline vertices rather than interpolating the true curve. For a bbox/outline-density purpose this is **acceptable and conservative**: a cubic Bezier lies within the convex hull of its control points, so the derived bbox never under-covers the real curve. This is also exactly consistent with the pre-existing `_edge_cuts_bbox_sexp` behavior, so it introduces no new inconsistency. Noted, not blocking.

3. **25mm staging margin** — cannot mask a genuinely-wrong placement. The origin-fallback failure mode (empty outline) is caught by a **separate, distance-independent** path: `_outline_bbox()` returns `None` → the unconditional "No Edge.Cuts outline detected" warning fires regardless of how far off the part lands. The 25mm padded-bbox check only governs the "outline readable but part far outside" case, where 25mm reliably exempts the deliberate `max_x + 10` staging offset (plus 5mm/part spacing) while still flagging wildly mis-placed parts. The margin is absolute rather than board-scaled, but the docstring on `_OUTLINE_STAGING_MARGIN_MM` documents the tradeoff clearly and the absolute value is defensible given the origin-fallback path is margin-independent.

4. **Warning plumbing** — both warnings flow through the existing `result.warnings` → `format_text`/`format_json` paths (no formatter changes). Tests explicitly assert presence in BOTH renderers: `test_no_outline_fallback_emits_warning` and `test_off_board_placement_emits_per_reference_warning` each check `format_text` output and the parsed `format_json` `warnings` array.

5. **Diff hygiene** — only the three intended files changed (pcb.py, pcb_sync_netlist.py, test file). Test fixtures are inline string constants; no stray fixture files committed.

6. **Error-path correctness** — the added `continue` after an `add_footprint` failure correctly skips the off-board check for a footprint that was never placed, avoiding a spurious warning on a part that errored out.

Nice touch exempting the deliberate staging offset via the padded bbox rather than special-casing the staging position — it keeps the check robust to *why* a part is off-board.

PR #4011, opened 2026-07-09
fix(lvs): vacuity guard — zero bound schematic pins can never be clean=true (closes #4006)

rjwalters:
Closes #4006. Fixes the LVS-vacuity hole found by the PR #4005 judge review (https://github.com/rjwalters/kicad-tools/pull/4005#issuecomment-4921804544) and cleans up board 06's already-merged instance of it (#4004).

## The hole

`compare_copper_netlist` returned `clean: true` with **zero evidence** on wireless fixture schematics: with no `(wire ...)` elements, `_schematic_pin_to_net` binds 0 pins, `compare_partitions` can detect neither shorts nor opens, and "copper-only 0 shorts / 0 opens" was a zero-comparison result. Board 07's 5 real copper opens were masked by exactly this; board 06 shipped a vacuous `clean=true` lvs.json on main via #4004, rendering a zero-evidence "Ready" LVS badge.

## Guard contract (deliberately the simplest defensible one)

**A copper-LVS result is VACUOUS iff zero board pads carry a real schematic net** — and a vacuous result is never `clean=true`. `compare_partitions` returns `clean=false` with a single synthetic mismatch of `kind="vacuous"` (`net=<no-schematic-evidence>`, witnesses `bound_pads=0` / `board_pads=N`) plus `bound_pad_count=0`.

Why zero rather than a `<50%`-style floor: any bound pad ≥ 1 is real, falsifiable evidence for the nets it touches, and legitimate boards mix wired nets with PCB-only fixtures — a ratio threshold needs an arbitrary constant and misfires on those. Zero bound pads is the unique case where the comparator observes literally nothing. (Documented in `copper_lvs.py`'s module docstring.)

Because the guard lives in the pure `compare_partitions` core, it covers `compare_copper_netlist`, the `python -m kicad_tools.lvs.copper_lvs` subprocess entrypoint, `write_lvs_report`'s #3838 fresh-check gate, and `kct check`'s LVS leg identically. Note the label comparator has no symmetric hole: on a wireless schematic every netted pad mismatches `schematic_net=None`, so label-LVS reads *dirty*, never vacuously clean — the vacuity hole was specific to copper-only gating (`run_label=False`).

## Schema: additive to v1 lvs.json (field additions only)

* `copper_vacuous: bool`, `copper_bound_pad_count: int|null` (recipe payload)
* `bound_pad_count` in the subprocess JSON (round-trips through `result_to_json`/`result_from_json`; legacy payloads without it still parse)
[...]

PR #4112, opened 2026-07-13
fix(lvs): recurse hierarchical sheets so LVS is no longer vacuous on multi-sheet designs

### LINKED ISSUE #4099: LVS/ERC Schematic.load does not recurse hierarchical sheets — LVS vacuous on hierarchical designs (bound_pads=0)

## Problem

`kct check`'s schematic loader reads only the **root sheet** of a hierarchical design. On a project whose root `.kicad_sch` contains only sheet symbols (all components live in 6 sub-sheets: mcu, clock, power, connectors, sync, dac), the loader sees **0 symbols**, so LVS runs vacuously: `bound_pads=0 / board_pads=289`, comparing an empty schematic against a populated board.

ERC in the same run was substantive (0 errors / 0 warnings), so this appears specific to the netlist-extraction path used for LVS symbol binding.

## Impact

LVS — the gate that would catch exactly the class of bug this project shipped in a previous rev (a bootloader UART wired straight-through) — silently certifies nothing on any hierarchically-organized schematic, which is the normal organization for non-trivial boards. During manufacturing sign-off we had to document "LVS vacuous, tool limitation" instead of getting a real schematic-vs-board check, and cross-checked connectivity by hand-probing the exported netlist.

## Suggested fix

Recurse `(sheet ...)` references when loading the schematic (KiCad's own netlister semantics: instantiate each sub-sheet per sheet instance, resolve hierarchical labels/ports). Short of full hierarchy support, `kct check` should FAIL or WARN loudly when the loaded schematic yields 0 symbols against a non-empty board, rather than reporting an LVS result — a vacuous pass is worse than an honest NOT-RUN.

## Repro sketch

Any KiCad project whose root sheet contains only sheet symbols: `kct check <board> --mfr <tier>` with the project adjacent → observe LVS binding 0 pads while ERC parses the hierarchy fine.

---

### Root cause (verified against `origin/main`)

There are **two separate `Schematic` classes** in this repo, and the confusion between them is exactly the bug:

1. `src/kicad_tools/schematic/models/schematic.py` — the modern model. Its `extract_netlist(hierarchical=True)` **already recurses** `(sheet ...)` references via `kicad_tools.operations.netlist._collect_hierarchy_components()` (`src/kicad_tools/operations/netlist.py:525`), which loads each sub-sheet, merges components/nets, and resolves hierarchical-label ↔ sheet-pin connections across sheet boundaries. `SchematicPCBChecker` (`src/kicad_tools/validate/consistency.py`) already uses this path for `_check_nets`, `_check_properties` (issues #2633, #2625), and `check_lvs()` (via `extract_bom(sch_path, hierarchical=True)`).
2. **But the LVS engine that actually produces `bound_pads=N` — the one named in this issue — does not use that path.** `src/kicad_tools/lvs/board_lvs.py::_schematic_pin_to_net()` (line 108) calls:
   ```python
   from kicad_tools.schematic.models.schematic import Schematic
   sch = Schematic.load(sch_path)          # loads ONLY the root .kicad_sch
   for sym in sch.symbols:                  # root-sheet symbols only
       ...
       out[(ref, number)] = sch.get_net_for_pin(ref, number)
   ```
   `sch.symbols` and `get_net_for_pin()` both operate on `self._build_connectivity_graph()`, which is built purely from the single loaded file — sub-sheets are never opened, and `(sheet ...)` nodes are never followed. On a root sheet with only sheet symbols and zero component symbols, `sch.symbols` is empty, so `_schematic_pin_to_net` returns `{}` for every ref/pad, and every board pad ends up unbound. This is the literal path that produces `bound_pads=0`.
   - `_schematic_pin_to_net` feeds **both** LVS engines used by `kct check`:
     - `compare_netlists()` (`board_lvs.py:173`, label-based LVS)
     - `compare_copper_netlist()` (`src/kicad_tools/lvs/copper_lvs.py:278`, copper-extracted LVS) via `schematic_net_of_pad = _schematic_pin_to_net(sch_path)` (line 303)
   - Both are wired into the recipe gate `src/kicad_tools/lvs/recipe.py` (the actual `kct check` LVS step), so **both legs of LVS are vacuous** on a hierarchical root sheet, not just one.
   - The docstrings on both `compare_netlists` and `compare_copper_netlist` already claim `sch_path: Path to a .kicad_sch (root sheet for hierarchy)` — that comment is aspirational/incorrect today; it describes the intended contract, not the current behavior.

### Interaction with the #4011 vacuity guard

`src/kicad_tools/lvs/copper_lvs.py::compare_partitions()` has a vacuity guard (#4011/#4006, PR merged) that refuses to report `clean=True` when zero schematic pins bind to board pads — it emits a synthetic `kind="vacuous"` mismatch instead (`VACUOUS_KIND`, `pad_a="bound_pads=0"`, `pad_b="board_pads=N"`). **This guard is already firing correctly for hierarchical designs** — it's the mechanism that turned this from a silent false-pass into a visible, gate-failing `bound_pads=0/board_pads=289` symptom the reporter could see and file. That means the "suggested fix" bullet in the original issue ("fail loudly when 0 symbols load against a non-empty board") is **already implemented and working** for the copper-LVS leg; `board_lvs.compare_netlists()` (the label-based leg) has no equivalent guard, but its output would show root-only pins missing from schematic-side (`schematic_net=None` for every non-root-sheet pad), which is a different and less legible symptom for the same root cause.

The remaining gap is entirely in `_schematic_pin_to_net`'s failure to load sub-sheets — the guard is a symptom detector, not a fix. This issue is about making `_schematic_pin_to_net` produce real bindings for hierarchical designs so the guard stops firing (because there's now real evidence) instead of papering over it.

### Scope: Phase 1 (buildable, recommended)

Reuse the existing hierarchical walker rather than reimplementing sheet recursion a third time:

1. Rewrite `_schematic_pin_to_net()` in `src/kicad_tools/lvs/board_lvs.py` to build its `{(ref, pad) -> net}` map from `_collect_hierarchy_components(sch_path, "/")` (imported from `kicad_tools.operations.netlist`) instead of a single-file `Schematic.load()` + `get_net_for_pin()` loop. That helper already returns `(components, net_dict)` where `net_dict` is `{net_name: [PinRef, ...]}` merged across the whole hierarchy with hierarchical-label resolution — invert it to `{(ref, pad) -> net_name}` the same way `_check_nets`/`extract_netlist` callers already do (see `SchematicPCBChecker._extract_schematic_pin_nets` in `src/kicad_tools/validate/consistency.py` for the exact inversion pattern to mirror).
   - Pins with no wire/label bind must still resolve to `None` (floating), matching current single-sheet semantics — `_collect_hierarchy_components` does not enumerate un-netted pins today, so Phase 1 needs to reconcile "pin exists but has no net" (should map to `None`) vs. "pin doesn't exist" (should be absent from the dict) using each sheet's `sch.symbols[*].symbol_def.pins` the same way the current root-only code does, just applied per recursed sheet instead of only the root.
2. No changes needed to `copper_lvs.py` — it already calls `_schematic_pin_to_net` and will pick up the fix automatically; the vacuity guard will stop firing once real pins bind.
3. No changes needed to `recipe.py` — same reason.
4. Update the two stale docstrings ("root sheet for hierarchy") in `board_lvs.py` and `copper_lvs.py` to describe the corrected recursive behavior.

**Explicitly out of scope for Phase 1** (call out if the Builder's complexity assessment wants to split further):
- Per-sheet-instance netlist resolution for the same sub-sheet used multiple times with different hierarchical label bindings (KiCad's full "sheet instance" semantics, i.e. the same `.kicad_sch` file placed twice with different net contexts) — `_collect_hierarchy_components` already has partial support for this via `sheet_path` bookkeeping but does not fully disambiguate net names per-instance; if the reporter's board has repeated sub-sheets with divergent nets, a Phase 2 issue should track that. Confirm this isn't the reporter's actual case before treating it as blocking (their sub-sheets are named individually — mcu/clock/power/connectors/sync/dac — suggesting one instance each).
- Rewriting `SchematicPCBChecker`/`extract_bom`-based LVS (`src/kicad_tools/validate/consistency.py`) — that path already works correctly and is not implicated in this bug.

### Acceptance Criteria

- [ ] `board_lvs.compare_netlists()` and `copper_lvs.compare_copper_netlist()`, given a root `.kicad_sch` whose only content is `(sheet ...)` symbols (all components in sub-sheets), bind pins from every sub-sheet — `bound_pad_count` / `bound_pads` reflects real sub-sheet pins, not zero.
- [ ] The `#4011` vacuity guard (`compare_partitions`, `VACUOUS_KIND`) no longer fires for a hierarchical fixture that has real wiring in its sub-sheets — LVS produces a genuine `clean=True`/`clean=False` verdict with real shorts/opens evidence instead of the synthetic `vacuous` mismatch.
- [ ] A single-sheet (non-hierarchical) schematic's LVS behavior is byte-identical to before (no regression) — existing `tests/test_board_lvs_nc_sentinel.py` / `tests/test_copper_lvs.py` / `tests/test_lvs_recipe.py` continue to pass unmodified.
- [ ] A hierarchical LVS fixture (new `.kicad_pcb` built from `tests/fixtures/hierarchical/root.kicad_sch` + its sub-sheets, or an equally-sized new fixture pair) demonstrates the before/after: vacuous (`bound_pads=0`) on current `main`, real comparison (non-zero `bound_pad_count`, real shorts/opens or a genuine clean pass) after the fix.
- [ ] Docstrings in `board_lvs.py`/`copper_lvs.py` claiming "root sheet for hierarchy" are corrected to describe actual recursive behavior.

## Affected Files

- `src/kicad_tools/lvs/board_lvs.py` — `_schematic_pin_to_net()` (line 108) needs to walk sub-sheets via `kicad_tools.operations.netlist._collect_hierarchy_components()` instead of loading only the root `.kicad_sch`; update the `compare_netlists()` docstring (line ~188).
- `src/kicad_tools/lvs/copper_lvs.py` — no code change expected (consumes `_schematic_pin_to_net` as-is), but update the `compare_copper_netlist()` docstring (line ~288) once the underlying behavior is fixed.
- `src/kicad_tools/operations/netlist.py` — `_collect_hierarchy_components()` (line 525) is the existing recursive walker to reuse; read-only reference, changes only if Phase 1 surfaces a gap in per-pin (vs. per-net) resolution.
- `src/kicad_tools/validate/consistency.py` — `_extract_schematic_pin_nets()` (search for that name) is the existing precedent for inverting `_collect_hierarchy_components`'s `net_dict` into a `{(ref, pin): net}` map; mirror this pattern rather than re-deriving it.
- `tests/test_board_lvs_nc_sentinel.py` — existing unit-level test module for `board_lvs.compare_netlists()`; add hierarchical-fixture regression test(s) here, or create a new `tests/test_board_lvs_hierarchical.py` if that file's scope (NC-sentinel normalization) doesn't fit cleanly.
- `tests/test_copper_lvs.py` — add hierarchical-fixture regression test(s) confirming the vacuity guard no longer fires for a wired hierarchical design.
- `tests/test_lvs_recipe.py` — existing recipe-level integration test module for both LVS engines; sanity-check the recipe gate end-to-end on the new hierarchical fixture if convenient.
- `tests/fixtures/hierarchical/` — existing fixture tree (`root.kicad_sch`, `sub_a.kicad_sch`, `sub_b.kicad_sch`, `child_with_hlabels.kicad_sch`, `nested.kicad_sch`, `parent_with_pins.kicad_sch`, `root_shared.kicad_sch`, `empty.kicad_sch`); reuse for the new LVS tests. **No matching `.kicad_pcb` exists yet for this fixture set** — the Builder will need to hand-build (or generate via `kct` board-generation helpers) a small routed/unrouted PCB whose footprints/pads match `root.kicad_sch` + sub-sheets' symbols, OR add a new self-contained hierarchical sch+pcb fixture pair sized for this test alone (preferred if the existing tree's symbol set is awkward to match against pads).

## Test Plan

- [ ] `uv run pytest tests/test_board_lvs_nc_sentinel.py tests/test_copper_lvs.py tests/test_lvs_recipe.py tests/test_board_00_lvs.py tests/test_board_03_regression.py -v` — full pass, including new hierarchical cases (this is the full set of existing test modules that exercise `compare_netlists`/`compare_copper_netlist`, per `git grep -l compare_netlists tests/*.py`).
- [ ] New unit test: `_schematic_pin_to_net()` against a multi-sheet fixture returns bindings for sub-sheet refs (not just root-sheet refs).
- [ ] New unit test: `compare_copper_netlist()` against a hierarchical sch+pcb fixture pair does NOT return `result.vacuous == True` when the sub-sheets have real wiring and matching board pads.
- [ ] Regression: existing non-hierarchical fixtures in `tests/test_board_lvs_nc_sentinel.py`/`tests/test_copper_lvs.py` still pass unchanged (confirms root-sheet-only behavior is preserved as a subset of the new recursive behavior).
- [ ] Manual/edge case: a root sheet with sheet symbols AND some root-level components (mixed hierarchy) binds pins from both root and sub-sheets — not an either/or.
- [ ] Manual/edge case: floating (unconnected) pins in a sub-sheet still resolve to `net=None`, not silently dropped from the pin map (would falsely suppress a real open/short).

rjwalters:
## Summary

`kct check`'s board LVS read the schematic through
`board_lvs._schematic_pin_to_net()`, which loaded only the root `.kicad_sch`
and never followed `(sheet ...)` references. On a design whose root sheet
contains only sheet symbols — the normal organization for a non-trivial board,
where all components live in sub-sheets — it saw zero symbols, bound zero pads,
and both LVS engines (`compare_netlists`, `compare_copper_netlist`) ran
vacuously (`bound_pads=0 / board_pads=N`). The #4011 vacuity guard correctly
turned this into a visible failure rather than a silent false pass, but the
real fix is to bind the sub-sheet pins so LVS actually certifies connectivity.

## Changes

- `src/kicad_tools/lvs/board_lvs.py`:
  - New `_walk_hierarchy_schematics()` helper — walks the root sheet plus every
    `(sheet ...)` sub-sheet (resolving `Sheetfile` relative to each parent,
    circular-reference-guarded), the same traversal
    `_collect_hierarchy_components` uses.
  - `_schematic_pin_to_net()` now resolves every pin across the whole
    hierarchy via each sheet's `Schematic.get_net_for_pin`. This stays
    **PWR_FLAG-safe** (VCC/GND preserved) — deliberately *not* the merged
    `net_dict` from `_collect_hierarchy_components`, which collapses every
    PWR_FLAG-touching pin into one `PWR_FLAG` net and would spuriously fail LVS
    on every board that uses PWR_FLAG. For a single-sheet design the resolution
    is byte-identical to before.
  - Corrected the module + `compare_netlists` docstrings ("root sheet for
    hierarchy" was aspirational, not actual).
- `src/kicad_tools/lvs/copper_lvs.py` — no code change; docstring corrected to
  describe the recursive behavior it now inherits.
- `tests/fixtures/hierarchical_lvs/` — new self-contained fixture set:
  `root_lvs.kicad_sch` (only sheet symbols, zero components), `sub_mcu` /
  `sub_pwr` sub-sheets (R1/R2 wired VCC/GND), `board_lvs.kicad_pcb` (matching
  routed board), plus `root_float.kicad_sch` + `sub_float.kicad_sch` for the
  floating-pin case.
- `tests/test_board_lvs_hierarchical.py` — new regression module.

| Both LVS engines bind real sub-sheet pins (not zero) on a sheet-symbols-only root | ✅ | `_schematic_pin_to_net(root_lvs)` binds R1/R2 pins to VCC/GND; `bound_pad_count == 4` |
| #4011 vacuity guard no longer fires for a wired hierarchical fixture | ✅ | `compare_copper_netlist` returns `vacuous is False`, `clean is True`, real `shorts=()`/`opens=()` |
| Single-sheet LVS byte-identical (no regression) | ✅ | board00/01/03/04 artifact + NC-sentinel + recipe tests pass unmodified; board00 still resolves `J1.1→VCC`, `D1.1→GND` (PWR_FLAG-safe), not `PWR_FLAG` |
| Hierarchical fixture shows vacuous→real before/after | ✅ | `Schematic.load(root).symbols == []` (before) vs 4 bound pins + genuine clean verdict (after) |
| Docstrings corrected | ✅ | module + `compare_netlists` + `compare_copper_netlist` updated |
| Floating sub-sheet pins resolve to `None`, not dropped | ✅ | `root_float` binds `("R3","1")→None`, `("R3","2")→GND` |

## Test Plan

```
uv run pytest tests/test_board_lvs_nc_sentinel.py tests/test_copper_lvs.py \
  tests/test_lvs_recipe.py tests/test_board_00_lvs.py \
  tests/test_board_03_regression.py tests/test_board_lvs_hierarchical.py
# 101 passed, 1 skipped
```

Also ran the consistency / SchematicPCBChecker suite (95 passed, 1 skipped) to
confirm the unchanged `extract_bom`-based LVS path is unaffected. `ruff check`,
`ruff format --check`, and `mypy` clean on the changed source.

Out of scope (Phase 2): per-sheet-instance net resolution for the same
sub-sheet placed multiple times under divergent parent net contexts.

Closes #4099


rjwalters:
✅ **Approved!**

Verified the LVS hierarchical-sheet recursion against the code, ran the regression suites locally, and confirmed all CI gates pass.

## What was verified

**PWR_FLAG pivot is legitimate (not a workaround for a first-attempt bug).**
The curated plan said to reuse `_collect_hierarchy_components`'s merged `net_dict`, but that returns a single collapsed `PWR_FLAG` net for every PWR_FLAG-touching pin — losing the VCC/GND distinction. The PR instead recurses (`_walk_hierarchy_schematics`) and resolves each pin via that sheet's `Schematic.get_net_for_pin` — the *exact same primitive* the original root-only loader used. For a single-sheet design the walker yields one `Schematic` (root, no sub-sheets), so the loop is byte-identical to prior behaviour. Diagnosis confirmed against `operations/netlist.py`.

**Vacuous → real proof (ran new module locally, 6/6 pass):**
- `test_root_sheet_alone_sees_no_symbols` — before: root sees 0 symbols
- `test_schematic_pin_to_net_binds_sub_sheet_pins` — after: 4 pins bound (R1/R2 → VCC/GND)
- `test_copper_lvs_vacuity_guard_no_longer_fires` — `vacuous=False`, `bound_pad_count=4`, `clean=True`
- `test_floating_sub_sheet_pin_resolves_to_none_not_dropped` — floating pin → `None` (present, not dropped: a real open is never suppressed)
- `test_copper_lvs_detects_a_real_short_in_hierarchical_design` — recursion enables real short detection

**Single-sheet regression safety (local + CI):**
- Local: `test_board_lvs_nc_sentinel` + `test_copper_lvs` + `test_lvs_recipe` = 69 passed; `test_board_00_lvs` + `test_board_03_regression` + `test_board_03_copper_lvs` = 33 passed / 1 skipped.
- CI: **Test** gate + **Board 06 End-to-End (LVS)** + **Board 07 (LVS known opens)** all pass — the load-bearing proof that no single-sheet artifact regressed.

**Duplicate-sheet handling (item 3, out of scope to resolve but must not corrupt):**
`_walk` uses a resolved-path `visited` set plus a `path.exists()` guard, mirroring the existing `_collect_symbols` pattern exactly. A sub-sheet file referenced twice loads once — no double-count, no crash, no infinite loop on circular refs. Cross-instance sheet-pin unification is correctly deferred to Phase 2 and documented in the docstring.

## CI Status

All 16 checks pass; merge state **CLEAN**:
- Test — pass (15m43s)
- Board 00–07 End-to-End — pass (incl. 06/07 LVS)
- Diff-Pair / Match-Group / Board 05 Routing Regression — pass
- Type Check, Lint & Format, C++ Build, kicad-cli Round-trip, Routed PCB DRC — pass

## Minor note (non-blocking)

The diff carries a stray `.loom-managed` change (worktree marker 4084→4099). It's a worktree bookkeeping file, not shipped code — harmless, just noise. No action required.

Solid fix: closes the vacuous-LVS hole on hierarchical designs while staying byte-identical on single-sheet boards, with a self-contained fixture and honest before/after tests.

PR #4113, opened 2026-07-13
feat(check): warn on stale zone fills + opt-in --refill-zones flag

### LINKED ISSUE #4096: kct check measures copper against stale zone fills — no pre-check refill (≈1100 phantom clearance_*_zone errors)
**Version:** kicad-tools 0.14.0. **Context:** manufacturing sign-off of a 4-layer board in a consumer repo (project-shamrock/chorus, `chorus-test-revA_v23_mfg.kicad_pcb`).

## Problem

`kct check` evaluates segment/via/pad-to-zone clearance against the **stored zone fill polygon** as committed in the `.kicad_pcb` — i.e. the pre-knockout fill. On a board whose stored fills are stale (routed after the last fill, or fills saved before knockout), this produces massive false-positive counts:

- Committed board: `kct check --mfr jlcpcb` → **1257 errors** (686 clearance_segment_zone, 395 clearance_via_zone, 27 clearance_pad_zone, …).
- After `kicad-cli pcb drc --refill-zones --save-board` (persist fresh fills) and re-running the identical `kct check`: **all three zone-clearance classes drop to exactly 0** (149 residual errors in unrelated classes).
- One sampled finding claimed a signal segment overlapped the GNDD pour by **1.729 mm** — a hard short if real; KiCad's refilled fill is correctly knocked out around that copper, so there is no overlap.

`kicad-cli pcb drc --refill-zones` (which refills before checking) reported 0 errors on the same board throughout.

## Impact

This is the mirror image of the stale-fill hazard that motivates the "cross-gate with kicad-cli --refill-zones" convention (stale fills can also cause false *negatives*). Either direction, the root cause is the same: DRC over stale fill state. In our sign-off run it initially looked like the board was 1257 errors from manufacturable when it was actually clean.

## Suggested fix

Refill zones in-memory before the clearance checks (matching `kicad-cli pcb drc --refill-zones` semantics), or add a `--refill-zones` flag and warn loudly when checking against stored fills whose mtime/state may predate the routing.

## Repro sketch

Any board where copper was added/moved after the last zone fill save: run `kct check --mfr <tier>`, note zone-clearance errors; run `kicad-cli pcb drc --refill-zones --save-board`; re-run `kct check` — the zone-clearance classes collapse to 0.

rjwalters:
## Summary

`kct check` measures segment/via/pad-to-zone clearance against the committed `filled_polygon` geometry in the `.kicad_pcb` — whatever KiCad last wrote to disk. When those fills are stale (the board was routed or refilled *after* the fills were last saved), the `clearance_*_zone` rules report phantom shorts (~1257 on the board in the report) that all collapse to 0 once `kicad-cli pcb drc --refill-zones --save-board` brings the on-disk fill back in sync with the copper.

This is a pure-Python Phase 1 per the curator: a **loud advisory** plus an **opt-in `--refill-zones` flag**. A true in-memory refill was ruled out (no zone-fill engine exists in the codebase, and `.kicad_pcb` has no per-zone timestamp for a reliable staleness detector). We make the staleness visible and actionable rather than silently picking a side.

## Changes

- **`src/kicad_tools/cli/check_cmd.py`**
  - `_warn_stale_zone_fills()`: prints a loud stderr advisory whenever any `clearance_segment_zone` / `clearance_via_zone` / `clearance_pad_zone` finding is present, pointing at `kicad-cli pcb drc --refill-zones --save-board <pcb>` (mirrors the existing net-class-map INACTIVE / manifest STALE advisory pattern). No-op otherwise. Fires in both default and `--drc-only` mode.
  - New `--refill-zones` CLI flag: shells out to refresh the fills before `PCB.load()`. Help text states explicitly that it **MUTATES the board file in place** (`--save-board`). Suppresses the advisory (fills are already fresh). Graceful degradation via `_refill_zones_in_place()`.
- **`src/kicad_tools/cli/runner.py`**
  - New `run_refill_zones()` helper (follows the `run_drc` / `_run_fill_zones_via_drc` pattern): invokes `kicad-cli pcb drc --refill-zones --save-board`, returns `success=False` with a clear message (no crash) when `kicad-cli` is absent, snapshots/restores the net table around `--save-board`, and treats a DRC-violation non-zero exit as success.
- **`tests/test_cli_check.py`**: unit coverage for the advisory (fires on each zone rule id, silent on unrelated/empty findings), end-to-end through `main()`, the flag (invokes kicad-cli when present, degrades gracefully when absent, no refill by default, help documents mutation), and direct `run_refill_zones` command-construction + kicad-cli-missing coverage.

| Advisory fires when zone-clearance violations present | ✅ | `TestStaleZoneFillWarning` unit + `test_warning_fires_through_main_on_zone_clearance_finding` |
| `--refill-zones` invokes kicad-cli when available | ✅ | `test_flag_invokes_kicad_cli_when_available` (spy), `test_builds_refill_save_board_command` |
| Degrades gracefully when kicad-cli absent | ✅ | `test_flag_degrades_gracefully_when_kicad_cli_absent`, `test_returns_failure_when_kicad_cli_missing` |
| Default path byte-identical (no auto-refill) | ✅ | `test_no_refill_by_default`; advisory is stderr-only, gated on findings |
| Help text says the flag mutates the board | ✅ | `test_help_text_documents_board_mutation` |
| Existing check tests pass | ✅ | `tests/test_cli_check.py` + `tests/test_check_cmd_coverage.py` → 70 passed |

## Test Plan

- `uv run pytest tests/test_cli_check.py tests/test_check_cmd_coverage.py` → 70 passed
- mypy: 7 errors in `runner.py` are pre-existing on `origin/main` (verified via `git stash`); zero new errors introduced

Closes #4096

### DIFF
--- a/src/kicad_tools/cli/check_cmd.py
+++ b/src/kicad_tools/cli/check_cmd.py
@@ -381,6 +381,76 @@ def _lvs_subcheck(sch_path: Path | None, pcb_path: Path) -> SubCheckResult:
     )
 
 
+# Issue #4096: the clearance rule_ids whose truthfulness depends entirely on
+# the freshness of the committed `filled_polygon` geometry.  When any of these
+# fire, the on-disk fills may be stale (board routed/refilled after the fills
+# were last saved) and the findings may be phantom.
+_ZONE_FILL_DEPENDENT_RULE_IDS: frozenset[str] = frozenset(
+    {
+        "clearance_segment_zone",
+        "clearance_via_zone",
+        "clearance_pad_zone",
+    }
+)
+
+
+def _warn_stale_zone_fills(violations: list[DRCViolation], pcb_path: Path) -> None:
+    """Warn loudly when zone-clearance findings may reflect stale fills (issue #4096).
+
+    ``kct check`` measures segment/via/pad-to-zone clearance against the
+    committed ``filled_polygon`` geometry in the ``.kicad_pcb`` — whatever
+    KiCad last wrote to disk.  If the board was routed or refilled after those
+    fills were saved, the clearance_*_zone rules produce phantom shorts that
+    disappear once the fills are refreshed.  This advisory points the user at
+    the authoritative fix so an honest-looking wall of clearance errors is not
+    mistaken for an unmanufacturable board.  No-op when no such findings exist.
+    """
+    present = sorted({v.rule_id for v in violations if v.rule_id in _ZONE_FILL_DEPENDENT_RULE_IDS})
+    if not present:
+        return
+    print(
+        "WARNING: zone-clearance findings present "
+        f"({', '.join(present)}) — kct check measures these against the "
+        "committed zone fills in the .kicad_pcb, which may be STALE if the "
+        "board was routed or refilled after the fills were last saved. "
+        "Cross-gate / fix with:\n"
+        f"    kicad-cli pcb drc --refill-zones --save-board {pcb_path}\n"
+        "then re-run kct check, or pass --refill-zones to do this "
+        "automatically (issue #4096).",
+        file=sys.stderr,
+    )
+
+
+def _refill_zones_in_place(pcb_path: Path) -> None:
+    """Refresh the on-disk zone fills before checking, if possible (issue #4096).
+
+    Shells out to ``kicad-cli pcb drc --refill-zones --save-board`` via
+    :func:`kicad_tools.cli.runner.run_refill_zones` so the pure-Python pipeline
+    reads fills that are in sync with the copper.  This **mutates the board
+    file in place** (the ``--refill-zones`` flag documents that side effect).
+
+    Graceful degradation: a missing kicad-cli — or any refill failure — is
+    reported as a warning and the check continues against the stored fills
+    rather than aborting.  Never raises.
+    """
+    from kicad_tools.cli.runner import run_refill_zones
+
+    result = run_refill_zones(pcb_path)
+    if result.success:
+        print(
+            f"[INFO] refilled zones in place via kicad-cli: {pcb_path}",
+            file=sys.stderr,
+        )
+    else:
+        print(
+            "WARNING: --refill-zones requested but the refill did not run "
+            f"({result.stderr.strip() or 'unknown error'}); continuing against "
+            "the stored (possibly stale) zone fills.  Install KiCad 8+ so "
+            "kicad-cli is on PATH to enable the pre-check refill (issue #4096).",
+            file=sys.stderr,
+        )
+
+
 def _manifest_subcheck(pcb_path: Path) -> SubCheckResult:
     """Compare ``output/manufacturing/manifest.json`` mtime against the PCB.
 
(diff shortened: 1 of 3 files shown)


rjwalters:
✅ **Approved!**

Phase-1 stale-zone-fill advisory + opt-in `--refill-zones` flag. Clean, well-scoped, and correctly conservative — no default-behavior change.

## Verification
- **CI**: all 16 checks pass (Board 00–07 E2E, Board 05 routing ≤11, Diff-Pair + Match-Group regressions, Test, Type Check, Lint, C++ Build, DRC, Round-trip). Merge state **CLEAN**.
- **Local (PR worktree, HEAD 07fd1e9)**: `pytest tests/test_cli_check.py tests/test_check_cmd_coverage.py` → **70 passed**; ruff check + format clean on all 3 files; mypy → 7 errors, all pre-existing on `origin/main` (lines 45–47, 858, 886, 941 — none in the new `run_refill_zones` or `check_cmd.py`), **0 new**.
- **Diff hygiene**: exactly 3 files, `+452 / −0`. The default check pipeline is byte-identical — the only new call sites are gated behind `--refill-zones` (refill) and `if not args.refill_zones` (advisory), and the advisory is stderr-only. Board CI gates confirm default output unchanged.

## Justified deviations evaluated
1. **No naturally-stale committed fixture** — acceptable. Per repo history, all committed board artifacts are refilled and schema round-trip loses fill geometry, so no organically-stale fixture exists. `test_warning_fires_through_main_on_zone_clearance_finding` pins the wiring correctly by wrapping the **real** `run_selected_checks`, appending a synthetic `clearance_segment_zone` into the real `DRCResults`, and asserting the advisory through `main()` under `--drc-only`. This validates the load-bearing path (main → `results.violations` → `_warn_stale_zone_fills`).
2. **Net-table snapshot/restore around `--save-board`** — sound. It's a faithful copy of the shipped `_run_fill_zones_via_drc` pattern: it guards the known kicad-cli quirk where re-serializing a kicad-tools-authored board strips the `(net N "name")` header table and inline net assignments. Restore runs **only on the success branch** (report produced); on the no-report and `FileNotFoundError`/`SubprocessError` paths the file is left as kicad-cli wrote it and `success=False` is returned — no false repair. Same exposure surface as the already-shipped refill paths (`run_drc`, `_run_fill_zones_via_drc`), so no new regression risk.
3. **Advisory suppression when `--refill-zones` ran** — cannot silently hide a real stale-fill. When the flag is passed but the refill *fails* (kicad-cli absent, etc.), `_refill_zones_in_place` prints its own explicit "continuing against the stored (possibly stale) zone fills" warning, and only *then* is the finding-gated advisory suppressed. The two warnings are mutually exclusive by design; the user is never falsely reassured. `test_flag_degrades_gracefully_when_kicad_cli_absent` exercises exactly this path (returns 0, warns, no crash).

## Non-blocking follow-up
`_refill_zones_in_place`'s docstring says "Never raises," but it does not wrap the `run_refill_zones` call in try/except. `run_refill_zones` only catches `FileNotFoundError` / `subprocess.SubprocessError`; a parse error from `_restore_net_declarations` (e.g. a board truncated by a mid-write subprocess death) would propagate and crash `main()`. This is **identical exposure to the already-shipped `_run_fill_zones_via_drc` / `run_drc` refill paths** — not a regression this PR introduces — so it does not block. Worth a small hardening issue to make the docstring true (wrap the call or broaden the except).

Solid work.

PR #4111, opened 2026-07-13
fix(placement): point cmaes ImportError at the placement extra

rjwalters:
## Summary

`kct optimize-placement` (and the placement step of `kct build --optimize-placement`) failed on a base install with the misleading hint `pip install cmaes`. `cmaes` is already declared as the `placement` optional extra (and transitively via `dev`/`all`), so the guidance should point at the declared extra — not a bare package install that silently mutates a venv without pinning. Per curation, `cmaes` stays an extra (it pulls `scipy`); this is a DX/docs fix, not a dependency-graph change.

## Changes

- `src/kicad_tools/placement/cmaes_strategy.py`: extract the import-guard message into `CMAES_INSTALL_HINT`, now naming `uv sync --extra placement` / `pip install "kicad-tools[placement]"` (and noting `dev`/`all` include it).
- `README.md`: document the `placement` extra in the installation block with a note that `kct optimize-placement` / `kct build --optimize-placement` require it.
- `tests/placement/test_cmaes_missing_extra.py` (new): asserts the hint names the extra (and is not a bare `pip install cmaes`), and that `_create_strategy("cmaes")` re-raises the actionable message when the cmaes strategy import is unavailable. This test runs regardless of whether `cmaes` is installed.

No change to `pyproject.toml` core `dependencies`; the CLI (`optimize_placement_cmd.py:729-731`) already catches `ImportError` and prints `Error: <hint>` cleanly (exit 1, no traceback), and `build_cmd.py`'s placement step already surfaces subprocess stderr on failure — both now carry the improved message.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| ImportError message names the `placement` extra | ✅ | `CMAES_INSTALL_HINT` includes `uv sync --extra placement` and `kicad-tools[placement]`; test asserts it |
| Base install → clear actionable message, non-zero, no traceback | ✅ | Ran `kct optimize-placement <fixture>` in a base-install venv → printed the extra-naming message, `EXIT: 1`, no traceback |
| With extra → runs successfully | ✅ | `uv sync --extra placement` → `kct optimize-placement <fixture> --max-iterations 3 --output ...` → `EXIT: 0` |
| README documents the `placement` extra | ✅ | Added to installation block + note referencing optimize-placement / build |
| `kct build` placement step not raw/unclear when cmaes missing | ✅ | Step shells out to the same CLI; on failure returns `success=False` with the subprocess stderr (the improved message). Opt-in via `--optimize-placement` |
[...]

PR #4116, opened 2026-07-13
fix(export): hard-fail --auto-lcsc when the 'parts' extra is missing

### LINKED ISSUE #4104: kct export --auto-lcsc soft-fails to an empty LCSC column when the [parts] extra is missing — export still reports success

## Problem

`kct export <board> --mfr jlcpcb --auto-lcsc` on a venv without the `parts` extra emits one warning per BOM line:

```
Search failed for R1: The 'requests' library is required for LCSC API access. Install with: pip install kicad-tools[parts]
No LCSC match for 330R [Resistor_SMD:R_0402_1005Metric] (R1, R12, ...): The 'requests' library is required ...
```

…and then **completes successfully**: full bundle, fresh manifest, exit 0, BOM with an intact `LCSC Part #` header and every cell empty. In a scripted/agent sign-off flow the warnings scroll past and the defect ships — our manufacturing bundle landed on main with a JLCPCB assembly BOM containing zero orderable part numbers, and it was only caught by human inspection of the CSV afterwards.

## Why this deserves more than a warning

`--auto-lcsc` was *explicitly requested* (it also happens to be the CLI default — see below). Producing 0/27 matches because the capability is absent (as opposed to genuine no-match results) is a precondition failure, not a degraded result. The export's own preflight validates BOM/ERC/LCSC/cosmetics — but apparently not "the requested matcher can run at all."

## Suggested fix (either)

1. Fail fast: if `--auto-lcsc` is passed and the LCSC client can't initialize (missing dependency, no network), exit non-zero with the install hint — before writing the bundle.
2. Or fail at the end: if auto-lcsc was requested and 100% of lookups errored (dependency/transport errors, distinct from no-match), treat it as a preflight failure and refuse to write the manifest.

A distinct exit path for "matched 0 because broken" vs "matched 0 because obscure parts" is the key.

## Repro

Fresh venv with base kicad-tools (no `[parts]`), any board with a BOM: `kct export <board> --mfr jlcpcb --auto-lcsc` → observe exit 0 + empty column.

---

### Root Cause (verified against `origin/main`)

`--auto-lcsc` already **defaults to `True`** on the CLI (`src/kicad_tools/cli/export_cmd.py:283-287`), so this isn't an opt-in edge case — every `kct export --mfr jlcpcb` without `--no-auto-lcsc` hits this path, extra-or-not.

The failure is swallowed at the *component* level, not surfaced as a capability check:

1. `requests` is gated behind the `parts` optional extra (`pyproject.toml:106-108`, `parts = ["requests>=2.28"]`) — **not** a core dependency and **not** pulled in transitively by any other extra (unlike the `cmaes`/`placement` case in #4100, which was also reachable via the `dev` extra).
2. `src/kicad_tools/parts/lcsc.py:89-102` — the `_requires_requests` decorator wraps `LCSCClient.search()`/`.lookup()`. `LCSCClient()` construction itself never raises; the `ImportError` only fires on the **first** `.search()` call. The message already correctly names the extra: `"The 'requests' library is required for LCSC API access. Install with: pip install kicad-tools[parts]"` (unlike the pre-fix cmaes message, this one doesn't need a wording fix — see Decision below).
3. `src/kicad_tools/cost/suggest.py:562-571` (`PartSuggester.suggest_for_component`) — catches `Exception` around the whole search block, explicitly re-raises only `LCSCForbiddenError` (line 567-568), and for everything else (including `ImportError`) does `logger.warning(f"Search failed for {reference}: {e}"); suggestion.error = str(e)`. This is where the dependency error becomes indistinguishable from "no matching parts found."
4. `src/kicad_tools/export/bom_enrich.py:401-418` (`enrich_bom_lcsc`) — for every BOM group, `suggestion.has_suggestion` is `False` (because of step 3), so each group is recorded as `EnrichmentEntry(source="unmatched", error=suggestion.error)` and logged via `logger.warning("No LCSC match for %s ...")`. There is no post-loop check for "100% unmatched due to the same repeated error string."
5. `src/kicad_tools/export/assembly.py:443-456` (`_generate_bom`) — calls `enrich_bom_lcsc(...)` inside its own `try/except Exception as e: logger.warning(...)`, so even a hard failure out of `enrich_bom_lcsc` itself would currently be swallowed at this outer layer too. BOM generation then proceeds unconditionally to `export_bom(...)` and writes the file — there is no gate on `EnrichmentReport.unmatched` / `EnrichmentReport.auto_matched` before the manifest is finalized.

Net effect: a missing `parts` extra degrades silently through four independent catch sites, and the existing manufacturing preflight (`src/kicad_tools/export/preflight.py`, `PreflightConfig`) has no check for "the requested LCSC matcher could not run at all" — only for BOM/ERC/DRC/cosmetics content, which are unaffected because the CSV *has* a header and *has* rows, just with empty `LCSC Part #` cells.

### Compare: #4100 / PR #4111 (cmaes) pattern

The just-merged sibling fix for the `placement` extra (#4100 → PR #4111) established the pattern this repo now expects for optional-extra failures:
- Named `..._INSTALL_HINT` constant giving both `uv sync --extra <name>` and `pip install "kicad-tools[<name>]"`.
- A base-install run produces a **clear, actionable, non-zero exit** — not a silent degrade.
- A regression test that runs regardless of whether the extra is installed (asserts the hint text, and that the failure path is reached deterministically without needing to literally uninstall the dependency in CI).

Two differences from the cmaes case worth flagging for the builder:
- The `lcsc.py:96-98` message already names the extra correctly (`pip install kicad-tools[parts]"`) — no wording fix needed there, unlike cmaes's stale `pip install cmaes`. The bug here is *purely* that the ImportError is caught-and-swallowed at three separate downstream layers, never surfaced to the CLI exit code.
- `parts` is **not** implied by `dev`/`all` the way `cmaes` was — worth double-checking `pyproject.toml`'s `dev`/`all` extras don't already include `requests` transitively via another package before assuming a truly clean venv is required to repro (verified: `parts` is dependency-isolated; no other extra lists `requests`).

### Decision: hard actionable error when `--auto-lcsc` is explicitly enabled

Given `--auto-lcsc` defaults to `True`, "explicitly requested" in the issue title effectively means "not explicitly disabled via `--no-auto-lcsc`." Recommendation, matching the issue's own Option 2 (fail at the end, not fail-fast at LCSCClient construction — construction never touches the network or raises, so there's no cheap pre-check available before the first `.search()` call):

- Distinguish **capability failure** (`ImportError` — the dependency literally isn't installed) from **transport/data failure** (network errors, 404s, genuine no-match) at the point they're caught in `suggest.py:562`. `ImportError` should not be swallowed into `suggestion.error` — it should propagate (or be wrapped in a new dedicated exception, e.g. `LCSCDependencyMissingError`) so `enrich_bom_lcsc` can detect it distinctly from a per-part "no match."
- In `enrich_bom_lcsc`, on the *first* `ImportError`/`LCSCDependencyMissingError` for any group, stop attempting further per-group searches (same short-circuit shape already used for `LCSCForbiddenError` at `bom_enrich.py:311-323`) and raise instead of continuing to accumulate a full run of "unmatched" entries with a repeated dependency error.
- In `assembly.py:443-456`, let that raised error propagate out of `_generate_bom` instead of being caught by the blanket `except Exception as e: logger.warning(...)` — at minimum, special-case the new exception type so it isn't swallowed alongside genuine best-effort degradation (e.g. cache-unavailable, single API hiccup).
- At the CLI layer (`export_cmd.py`), the export command should exit non-zero with the `pip install kicad-tools[parts]` hint (already correctly worded in `lcsc.py:97-98`) **before** the manifest/bundle is written, when `auto_lcsc=True` and matching failed for 100% of groups due to the missing dependency. Genuine no-match results (obscure parts, API returned zero candidates) must continue to produce the existing warning-only degrade — only the "the matcher literally cannot run" case should be promoted to a hard failure.
- `--no-auto-lcsc` must continue to fully bypass this check (unaffected — `auto_lcsc=False` never calls `enrich_bom_lcsc` at all, per `assembly.py:443`).

### Scope guard

- Do not add `requests` to core `[project] dependencies` or fold it into another extra — `parts` staying a separate, requests-only extra is intentional (matches the just-established `placement` precedent of "declared extra + actionable error," not "make everything core").
- Do not change `PreflightConfig`/`preflight.py` — this is a BOM-generation-time gate inside `assembly.py`/`bom_enrich.py`, not a new preflight check category.
- Do not touch the datasheet importer's near-identically-worded `Search failed for {part_number}` warning in `src/kicad_tools/parts/importer.py:268` — that is a different subsystem (datasheet PDF import), not BOM LCSC enrichment, and is out of scope.

## Acceptance Criteria

- [ ] `kct export <board> --mfr jlcpcb --auto-lcsc` (or the CLI default with `--auto-lcsc` implicit) in a venv without the `parts` extra exits **non-zero** and does **not** write the manifest/bundle, when the failure is caused by the missing dependency (not a genuine no-match).
- [ ] The failure message names the extra using the established install-hint pattern: `uv sync --extra parts` and `pip install "kicad-tools[parts]"` (the existing `lcsc.py:97-98` wording is close — confirm final phrasing matches the `CMAES_INSTALL_HINT` convention from PR #4111, e.g. a named `PARTS_INSTALL_HINT` constant if a constant extraction is warranted).
- [ ] Genuine no-match results (dependency present, LCSC API reachable, but zero candidates for an obscure part) continue to produce the existing warning-only degrade with exit 0 — this issue must not regress that legitimate use case.
- [ ] `--no-auto-lcsc` is unaffected: export still succeeds with an empty/absent LCSC column and no new errors.
- [ ] `LCSCForbiddenError`'s existing cache-fallback short-circuit behavior (`bom_enrich.py:311-323`, issue #3935's determinism contract) is unaffected by this change.
- [ ] With the `parts` extra installed and the API reachable, `--auto-lcsc` behavior is unchanged (regression-free).

## Affected Files

- `src/kicad_tools/cost/suggest.py` (`suggest_for_component`, ~line 562) — stop swallowing `ImportError` into `suggestion.error`; propagate distinctly (new exception type or explicit re-raise) so callers can tell "dependency missing" apart from "no match found."
- `src/kicad_tools/export/bom_enrich.py` (`enrich_bom_lcsc`, ~lines 304-323, 401-418) — detect the dependency-missing signal on first occurrence and short-circuit remaining groups (mirroring the existing `LCSCForbiddenError` handling), instead of accumulating a full pass of misleadingly-worded "no LCSC match" entries.
- `src/kicad_tools/export/assembly.py` (`_generate_bom`, ~lines 443-456) — do not let the blanket `except Exception` swallow the new dependency-missing signal; let it propagate (or set a hard-failure flag) so `kct export` can exit non-zero before finalizing the bundle.
- `src/kicad_tools/cli/export_cmd.py` (around the `auto_lcsc` config wiring, ~lines 283-300, and wherever the command's final exit code is determined) — surface the propagated error as a clear CLI failure (exit 1, actionable hint), not a swallowed warning.
- `src/kicad_tools/parts/lcsc.py` (`_requires_requests`, lines 89-102) — reference only; confirm final message wording matches whatever install-hint constant convention is chosen (already correctly names `kicad-tools[parts]`, unlike the pre-fix cmaes message).
- New test file, e.g. `tests/export/test_bom_enrich_missing_parts_extra.py` or `tests/cost/test_suggest_missing_parts_extra.py` (repo has no existing `tests/export/` or `tests/cost/test_bom_enrich*` file for this path — `tests/cost/test_suggest.py` exists and is a reasonable home for the `suggest_for_component` half of the test).

## Test Plan

- [ ] Automated test simulating the missing `requests` import (e.g. monkeypatching `_requires_requests`/`LCSCClient.search` to raise `ImportError`, without needing to actually uninstall `requests` in CI — mirrors PR #4111's `test_cmaes_missing_extra.py` approach of testing the hint/behavior deterministically regardless of the real environment's installed packages) asserts: `suggest_for_component` (or `enrich_bom_lcsc`) surfaces a distinct, identifiable failure rather than `suggestion.error = "<ImportError text>"` treated as a normal no-match.
- [ ] Automated test: `enrich_bom_lcsc` with a simulated dependency-missing condition raises/signals failure after the first group rather than looping through all groups accumulating "unmatched."
- [ ] Automated test: full `kct export --mfr jlcpcb --auto-lcsc` CLI invocation (via `export_cmd.main` or subprocess, with `requests`/LCSC client mocked to raise `ImportError`) exits non-zero and does not write `manifest.json` / the BOM CSV — or, if the "fail at the end" approach from the issue's Option 2 is chosen, confirm no manifest is finalized after 100% dependency-caused unmatched.
- [ ] Regression test: existing genuine no-match behavior (mock `client.search` returning zero candidates, dependency present) still produces exit 0 with a warning-only degrade — this must not flip to a hard failure.
- [ ] Regression test: `--no-auto-lcsc` still exits 0 and skips `enrich_bom_lcsc` entirely.
- [ ] Manual verification: run `kct export <fixture-board> --mfr jlcpcb --auto-lcsc` in a venv with `parts` uninstalled (`uv sync` without `--extra parts`, or `pip uninstall requests` in a scratch venv) → observe non-zero exit and the actionable install hint, no bundle written.
- [ ] Manual verification: same command with `uv sync --extra parts` (or `pip install "kicad-tools[parts]"`) and network reachable → export succeeds with populated `LCSC Part #` column, unchanged from current behavior.
- [ ] `uv run pytest tests/cost/test_suggest.py -q` and any new test file pass.
- [ ] `ruff check` / `ruff format --check` clean on changed files.

rjwalters:
## Summary

`kct export <board> --mfr jlcpcb --auto-lcsc` on a venv without the optional `parts` extra (`requests`) used to **soft-fail**: the `ImportError` was swallowed at three independent layers, every BOM group was recorded as `source="unmatched"`, the BOM CSV shipped with an intact `LCSC Part #` header and every cell empty, the manifest/bundle was finalized, and the command exited 0. A manufacturing bundle landed on main with zero orderable part numbers, caught only by human CSV inspection.

Because `--auto-lcsc` defaults to `True` for JLCPCB, this hit every `kct export --mfr jlcpcb` without `--no-auto-lcsc`. A missing capability is a precondition failure, not a degraded result — this PR promotes it to a hard, actionable failure while leaving genuine no-match degradation (obscure parts) untouched.

## Changes

- **`parts/lcsc.py`** — add `LCSCDependencyMissingError` and a `PARTS_INSTALL_HINT` constant (`uv sync --extra parts` / `pip install "kicad-tools[parts]"`), matching the `placement` extra's `CMAES_INSTALL_HINT` convention from PR #4111. The `_requires_requests` decorator now raises the constant.
- **`cost/suggest.py`** (`suggest_for_component`, ~L562) — translate a search-time `ImportError` into `LCSCDependencyMissingError` and re-raise it (mirroring the existing `LCSCForbiddenError` special case) instead of stashing it in `suggestion.error` where it's indistinguishable from "no match found".
- **`export/bom_enrich.py`** (`enrich_bom_lcsc`) — short-circuit on the **first** `LCSCDependencyMissingError` and re-raise, instead of looping every group accumulating misleading "no LCSC match" entries. The `LCSCForbiddenError` cache-fallback circuit breaker (issue #3935) is unchanged.
- **`export/assembly.py`** — let `LCSCDependencyMissingError` escape `_generate_bom` (which raises *before* the BOM CSV is written) and `export()`, rather than being caught by the blanket `except Exception`.
- **`export/manufacturing.py`** — abort `_generate_assembly` (return `False`) on the dependency error so the pipeline skips Report/ProjectZip/**Manifest**; the error is recorded in `result.errors`, so `kct export` exits non-zero (`result.success` is `False`) with the actionable hint before any bundle is finalized.

Scope guards honored: `requests` stays a standalone `parts`-only extra (not added to core deps or another extra); `PreflightConfig`/`preflight.py` untouched (this is a BOM-generation-time gate); `parts/importer.py`'s datasheet-import `Search failed` warning untouched.

| Missing `parts` extra → non-zero exit, no manifest/bundle written | ✅ | `_generate_assembly` returns `False` → pipeline skips manifest; `result.errors` non-empty → CLI exits 1. Test `TestManufacturingHardFailsWithoutParts` + `TestAssemblyPropagatesDependencyMissing` (no CSV written). |
| Failure message names the extra via the install-hint pattern | ✅ | `PARTS_INSTALL_HINT` asserts `uv sync --extra parts` + `pip install "kicad-tools[parts]"` (`TestPartsInstallHint`). |
| Genuine no-match still degrades warning-only, exit 0 | ✅ | `test_genuine_no_match_still_degrades` — zero candidates returns an empty suggestion, no raise. |
| `--no-auto-lcsc` unaffected | ✅ | `auto_lcsc=False` never calls `enrich_bom_lcsc` (`assembly.py:443` guard, unchanged). |
| `LCSCForbiddenError` cache-fallback unaffected | ✅ | `test_forbidden_still_falls_back_to_cache` + existing `TestEnrichBomLcscCircuitBreaker` pass unchanged. |
| `parts` installed + API reachable → unchanged | ✅ | Only the `ImportError` branch is newly promoted; all existing enrich/suggest tests pass. |

## Test Plan

New `tests/test_bom_enrich_missing_parts_extra.py` (8 tests) simulates the missing import by monkeypatching `client.search` / `enrich_bom_lcsc` / `AssemblyPackage.export` to raise — **no uninstall of `requests` in CI** (mirrors PR #4111's `test_cmaes_missing_extra.py`). Covers: suggest-level translation, bom_enrich short-circuit-and-raise, assembly propagation (no CSV), manufacturing abort (no manifest), the no-match regression, the forbidden circuit-breaker regression, and the install-hint constant.

```
uv run pytest tests/test_bom_enrich.py tests/cost/test_suggest.py \
  tests/test_bom_enrich_missing_parts_extra.py tests/test_parts.py -q
# 107 passed, 47 skipped
```

`ruff check` / `ruff format --check` clean on all changed files.

Closes #4104

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 4099
-# Branch: feature/issue-4099
+# Issue: 4104
+# Branch: feature/issue-4104
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/src/kicad_tools/cost/suggest.py
+++ b/src/kicad_tools/cost/suggest.py
@@ -562,10 +562,23 @@ class PartSuggester:
         except Exception as e:
             # Re-raise LCSCForbiddenError so callers can detect API-wide
             # unavailability and short-circuit remaining lookups.
-            from ..parts.lcsc import LCSCForbiddenError
+            from ..parts.lcsc import (
+                PARTS_INSTALL_HINT,
+                LCSCDependencyMissingError,
+                LCSCForbiddenError,
+            )
 
             if isinstance(e, LCSCForbiddenError):
                 raise
+
+            # An ImportError here means the optional ``parts`` extra
+            # (``requests``) is absent: the LCSC matcher literally cannot
+            # run.  This is a capability failure, not a genuine no-match --
+            # promote it to a distinct exception so callers can surface a
+            # hard, actionable error instead of recording an empty match
+            # for every BOM group (issue #4104).
+            if isinstance(e, ImportError):
+                raise LCSCDependencyMissingError(PARTS_INSTALL_HINT) from e
 
             logger.warning(f"Search failed for {reference}: {e}")
             suggestion.error = str(e)
--- a/src/kicad_tools/export/assembly.py
+++ b/src/kicad_tools/export/assembly.py
@@ -12,6 +12,7 @@ from pathlib import Path
 
 from kicad_tools.exceptions import FileNotFoundError as KiCadFileNotFoundError
 
+from ..parts.lcsc import LCSCDependencyMissingError
 from .bom_enrich import EnrichmentReport, enrich_bom_lcsc
 from .bom_formats import (
     BOMExportConfig,
@@ -275,6 +276,13 @@ class AssemblyPackage:
         if self.config.include_bom:
             try:
                 result.bom_path = self._generate_bom(out_dir, result)
+            except LCSCDependencyMissingError:
+                # A missing ``parts`` extra means --auto-lcsc could not run
+                # at all.  Propagate out of ``export`` so the manufacturing
+                # pipeline aborts BEFORE writing the manifest/bundle, rather
+                # than recording an error and shipping a BOM with an empty
+                # ``LCSC Part #`` column (issue #4104).
+                raise
             except Exception as e:
                 result.errors.append(f"BOM generation failed: {e}")
                 logger.error(f"BOM generation failed: {e}")
@@ -452,6 +460,17 @@ class AssemblyPackage:
                     result.lcsc_enrichment = enrichment
                 for line in enrichment.summary_lines():
                     logger.info(line)
+            except LCSCDependencyMissingError:
+                # The optional ``parts`` extra is missing, so --auto-lcsc
+                # (the CLI default for JLCPCB) cannot populate a single LCSC
+                # part number.  Do NOT swallow this into a warning-and-
+                # continue: propagate it so the BOM CSV is never written and
+                # ``kct export`` exits non-zero with the install hint, before
+                # the manifest/bundle is finalized (issue #4104).  This is
+                # distinct from best-effort degradation (cache-unavailable,
+                # a single API hiccup, a genuine no-match), which must keep
+                # exit 0.
+                raise
             except Exception as e:
                 logger.warning(f"LCSC auto-matching failed (continuing without): {e}")
 
--- a/src/kicad_tools/export/bom_enrich.py
+++ b/src/kicad_tools/export/bom_enrich.py
@@ -20,7 +20,7 @@ from typing import TYPE_CHECKING
 
 from ..cost.suggest import PartSuggester
 from ..parts.cache import PartsCache
-from ..parts.lcsc import LCSCForbiddenError
+from ..parts.lcsc import LCSCDependencyMissingError, LCSCForbiddenError
 from .lcsc_value_check import (
     check_lcsc_against_cache,
     find_package_mismatch,
@@ -308,6 +308,17 @@ def enrich_bom_lcsc(
                     footprint=footprint,
                     existing_lcsc=None,
                 )
+            except LCSCDependencyMissingError:
+                # The optional ``parts`` extra (``requests``) is absent, so
+                # the LCSC matcher cannot run for ANY group.  Short-circuit
+                # on the first occurrence and re-raise instead of recording
+                # a full pass of misleadingly-worded "no LCSC match" entries
+                # with an empty ``LCSC Part #`` column.  Callers surface this
+                # as a hard, actionable failure with the install hint
+                # (issue #4104).  Distinct from LCSCForbiddenError, which is
+                # a transient/global API state that legitimately falls back
+                # to the offline cache.
+                raise
             except LCSCForbiddenError:
                 # API is globally unavailable -- emit a single warning and
                 # fall back to cache for this and all remaining groups.
--- a/src/kicad_tools/export/manufacturing.py
+++ b/src/kicad_tools/export/manufacturing.py
@@ -19,6 +19,7 @@ from pathlib import Path
 
 import kicad_tools
 from kicad_tools.exceptions import FileNotFoundError as KiCadFileNotFoundError
+from kicad_tools.parts.lcsc import LCSCDependencyMissingError
 
 from .assembly import AssemblyConfig, AssemblyPackage, AssemblyPackageResult
 from .preflight import PreflightChecker, PreflightConfig, PreflightResult
@@ -699,6 +700,17 @@ class ManufacturingPackage:
             result.assembly_result = assembly.export(out_dir)
             if result.assembly_result.errors:
                 result.errors.extend(result.assembly_result.errors)
+        except LCSCDependencyMissingError as e:
+            # The optional ``parts`` extra is missing, so --auto-lcsc could
+            # not populate any LCSC part numbers.  This is raised out of BOM
+            # generation BEFORE the BOM CSV (the first artefact) is written,
+            # so nothing is on disk yet.  Abort the whole pipeline so we
+            # never finalize a manifest/bundle whose BOM has an empty
+            # ``LCSC Part #`` column; ``kct export`` then exits non-zero with
+            # this actionable install hint (issue #4104).
+            result.errors.append(str(e))
+            logger.error("LCSC auto-matching requires the 'parts' extra: %s", e)
+            return False
         except Exception as e:
             result.errors.append(f"Assembly generation failed: {e}")
             logger.error(f"Assembly generation failed: {e}")
--- a/src/kicad_tools/parts/lcsc.py
+++ b/src/kicad_tools/parts/lcsc.py
@@ -34,11 +34,37 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+# Canonical install hint for the optional ``parts`` extra.  Mirrors the
+# ``CMAES_INSTALL_HINT`` convention established for the ``placement`` extra
+# (issue #4100 / PR #4111): name the declared extra and give both the
+# ``uv sync`` and ``pip install`` incantations so a base-install failure is
+# actionable rather than a silent degrade.  Unlike ``cmaes``, ``parts`` is
+# NOT pulled in transitively by the ``dev`` or ``all`` extras.
+PARTS_INSTALL_HINT = (
+    "The 'requests' library is required for LCSC API access. "
+    "Install it with the 'parts' extra: uv sync --extra parts "
+    '(or: pip install "kicad-tools[parts]").'
+)
+
+
 class LCSCForbiddenError(Exception):
     """Raised when the JLCPCB API returns 403 Forbidden.
 
     This indicates the API is globally unavailable (e.g. authentication
     required, geo-blocking) and further requests should not be attempted.
+    """
+
+
+class LCSCDependencyMissingError(Exception):
+    """Raised when the optional ``parts`` extra (``requests``) is absent.
+
+    This is a *capability* failure -- the requested LCSC matcher cannot run
+    at all -- and is deliberately distinct from a per-part "no match found"
+    result.  Callers (BOM enrichment, ``kct export``) short-circuit on the
+    first occurrence and surface it as a hard, actionable failure instead of
+    degrading silently to an empty ``LCSC Part #`` column (issue #4104).
+
+    The message is the canonical :data:`PARTS_INSTALL_HINT`.
     """
 
 
@@ -92,11 +118,8 @@ def _requires_requests(func):
     def wrapper(*args, **kwargs):
         try:
             import requests  # noqa: F401
-        except ImportError:
-            raise ImportError(
-                "The 'requests' library is required for LCSC API access. "
-                "Install with: pip install kicad-tools[parts]"
-            )
+        except ImportError as exc:
+            raise ImportError(PARTS_INSTALL_HINT) from exc
         return func(*args, **kwargs)
 
     return wrapper
--- a/tests/test_bom_enrich_missing_parts_extra.py
+++ b/tests/test_bom_enrich_missing_parts_extra.py
@@ -1,0 +1,255 @@
+"""Tests for the hard, actionable failure when the ``parts`` extra is missing.
+
+Issue #4104: ``kct export <board> --mfr jlcpcb --auto-lcsc`` on a venv without
+the optional ``parts`` extra (``requests``) used to soft-fail -- every BOM
+group was recorded as "unmatched" with the ImportError text, the BOM CSV was
+written with an empty ``LCSC Part #`` column, the manifest/bundle was finalized,
+and the command exited 0.  A missing capability is a precondition failure, not a
+degraded result, so it must now:
+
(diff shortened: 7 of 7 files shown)


rjwalters:
✅ **Approved!**

Clean, tightly-scoped fix that promotes the missing-`parts`-extra case from a silent degrade to a hard, actionable failure while preserving genuine no-match degradation. Verified the full acceptance contract.

## Error-propagation chain (verified airtight, all 5 files)
1. `parts/lcsc.py` — new `LCSCDependencyMissingError` + `PARTS_INSTALL_HINT` constant (mirrors the `CMAES_INSTALL_HINT` convention from #4111); `_requires_requests` re-chains the `ImportError` from the constant.
2. `cost/suggest.py` — search-time `ImportError` → `raise LCSCDependencyMissingError(...) from e`, ordered so it does not fall into the `logger.warning`/`suggestion.error` no-match sink.
3. `export/bom_enrich.py` — `except LCSCDependencyMissingError: raise` is ordered **before** `except LCSCForbiddenError`, and since the two are unrelated classes the forbidden cache-fallback circuit breaker (#3935) is untouched. Short-circuits on the first group.
4. `export/assembly.py` — the new `except LCSCDependencyMissingError: raise` sits ahead of the blanket `except Exception` in both `_generate_bom` and `export()`, so it is **not** re-swallowed. Raised before `export_bom()` writes the CSV.
5. `export/manufacturing.py` — `_generate_assembly` catches it, appends to `result.errors`, returns `False`. Confirmed at the call site (`if not assembly_ok: return result`) that this skips Report/ZIP/README/**Manifest**. `ManufacturingResult.success` is `len(errors) == 0`, so `kct export` exits non-zero with the install hint. No bundle is finalized.

## Both paths pinned by tests
- **Hard-fail path**: `test_import_error_raises_dependency_missing`, `test_first_import_error_short_circuits_and_raises` (asserts `call_count == 1`), `test_generate_bom_raises_before_writing_csv` (asserts no CSV on disk), `test_no_manifest_written_and_error_reported`.
- **Degrade path preserved**: `test_genuine_no_match_still_degrades` (zero candidates → no raise, empty suggestion) and `test_forbidden_still_falls_back_to_cache` (403 → `unmatched == 2`, no raise). This is the critical no-regression guarantee and it is explicitly pinned.

## UX / escape hatch
`--auto-lcsc` defaults True for JLCPCB, so this is a hard failure on default exports — correct per the curator's decision (the user asked for LCSC enrichment; 0/N matches because the matcher can't run is a precondition failure, not a result). The escape hatch is real: `export_cmd.py:286` computes `auto_lcsc = args.auto_lcsc and not args.no_auto_lcsc`, and `--no-auto-lcsc` bypasses `enrich_bom_lcsc` entirely, so users without the extra can still export. The error message (`PARTS_INSTALL_HINT`) names both `uv sync --extra parts` and `pip install "kicad-tools[parts]"`.

## Test Execution
- `uv run pytest tests/test_bom_enrich_missing_parts_extra.py tests/test_bom_enrich.py tests/cost/test_suggest.py tests/test_parts.py -q` — ✅ **107 passed, 47 skipped** (matches builder's report)
- `ruff check` / `ruff format --check` on all 6 changed files — ✅ clean
- Tests use monkeypatch to simulate the missing import; no `requests` uninstall in CI (mirrors #4111).

## CI
All 16 checks **pass** (run 29276560413): Test, Lint & Format, Type Check, C++ Build, Boards 00–07, all routing regressions, DRC, kicad-cli round-trip. Merge state **CLEAN**.

Diff hygiene: 5 source + 1 test file, plus the `.loom-managed` worktree marker (Loom bookkeeping, not a source change). Scope guards honored — `requests` stays a `parts`-only extra, `preflight.py`/`PreflightConfig` untouched, datasheet importer's `Search failed` warning untouched.

PR #4119, opened 2026-07-13
feat(parts): official JLCPCB open-platform API backend (BYO access key)

### LINKED ISSUE #4118: parts: official JLCPCB open-platform API backend (BYO access key) — Phase 2 of #4108
**Owner-initiated (2026-07-13).** The owner has registered a JLCPCB open-platform app (name: `kicad-tools-developer`, app id `599660863674093570`) and wants the official-API path enabled: **users bring their own API key**; kicad-tools ships the client.

## Context

- Phase 1 (#4108, PR #4117, `feature/issue-4108`, currently in review) added the offline jlcparts catalog fallback. **This issue composes with #4117's tiering** — read `src/kicad_tools/parts/lcsc.py` on that branch before starting: `LCSCClient.lookup()`/`lookup_many()` already added a `use_local_catalog` / `_get_catalog()` seam with resolution order `response cache -> live scrape API -> offline catalog`. This issue **prepends** a new tier ahead of the live scrape API: `response cache -> official API (keys present) -> live scrape API -> offline catalog`. Note `#4117` currently uses `LCSCDependencyMissingError(ImportError)` for the "no requests + no catalog" case — the new official-API tier should raise/handle distinctly (see error-mapping below), not reuse that exception for auth/signature failures.
  - **Merge-order risk**: #4117 may merge before this issue is picked up, in which case `jlcparts_catalog.py` and the tiering scaffolding will already be on `origin/main` and the builder edits `lookup()`/`lookup_many()` in place. If #4117 has **not** yet merged, the builder should still design against #4117's current diff shape (fetched via `git fetch origin feature/issue-4108`) so the two PRs merge cleanly, but must not depend on that branch being present — build against `origin/main` and add the official-API tier as directly as possible so it rebases trivially onto #4117's tiering once that lands. If asked to choose, prefer implementing this on top of `origin/main` HEAD and let a human resolve the merge order; do not block on #4117.
- **Neither `open.jlcpcb.com` nor `jlcpcb.com/api` publishes first-party REST reference docs for the Parts/component endpoints.** JLC publishes **Java-only SDKs** (Basic/Core SDK + Business SDKs for PCB / 3D / **Parts** / JLC Balance, v1.0.4, 2026-06-09) — no Python SDK exists, and per third-party live-probing (below) the shipped Java "overseas" SDK jar is itself stale relative to the live API (it only wraps one component method). Our client speaks REST directly.

### Signing scheme — best available reference (NOT first-party-confirmed; live-smoke required)

No first-party public doc page for the signing algorithm was found (the `jlcpcb.com/help/article/*` CMS articles under the "JLCPCB API" help category are client-rendered from a CMS API that did not return article body content over a plain `curl`; `open.jlcpcb.com` itself returns HTTP 400 on GET with no docs route discovered). The best available public evidence is convergent, independently-reproduced third-party reverse-engineering, most usefully:

- `wavenumber-eng/supply-chain-monkey` — `docs/research/jlc/JLCPCB_API_RESEARCH.md` and the working client `src/py/scm/server/providers/jlc_openapi.py` (GitHub, public repo). States it derived the request/response shapes from **live HTTP probing** against `open.jlcpcb.com` plus inspection of the JLC Java SDK jars, and explicitly flags the Java SDK as stale/incomplete vs. the live API.
- `Jackster/JLCPCB-API` — `README.md` (GitHub, public repo), states it was built "from data extracted from the JLCPCB Java SDK files downloaded from the JLCPCB API website," documents the **PCB-ordering** surface (`jlcpcb.com/api/overseas/openapi/pcb/*`), independently confirms the same string-to-sign construction.
- `mattpainter701/kicad_automations` — `skills/jlcpcb/scripts/jlcpcb_api.py` + `SKILL.md` (GitHub, public repo), third independent implementation of the same PCB-ordering surface, same string-to-sign construction.

**Convergent facts across all three sources (high confidence):**
- Signature base string ("string-to-sign"): `f"{METHOD}\n{PATH}\n{TIMESTAMP}\n{NONCE}\n{BODY}\n"` — uppercase HTTP method, request path (see path-scope caveat below), Unix epoch seconds as a string, a random nonce, and the exact compact-JSON request body string (empty string for GET/no-body requests) — each field newline-terminated including a trailing newline after `BODY`.
- Nonce: random hex string, 32 chars (16 bytes via `secrets.token_hex(16)`) in every sample seen; no published minimum/maximum length.
- Timestamp: Unix epoch **seconds** (not milliseconds), as a decimal string.
- `Authorization` header carries `appid`, `accesskey`, `nonce`, `timestamp`, `signature` as quoted key="value" pairs (order varies slightly between samples but has not been shown to matter).
- Credential triplet: `appid` (application id), `accesskey` (public), `secretkey` (used only as the HMAC key material, never transmitted).

**Material, UNRESOLVED discrepancy the builder must treat as two candidate variants and test against the live API (do not guess):**
- PCB-ordering surface (`jlcpcb.com/api/overseas/openapi/pcb/*`, per `Jackster/JLCPCB-API` and `mattpainter701`): scheme keyword `JOP-HMAC-SHA256`, signature = **hex digest** of `HMAC-SHA256(secret_key, string_to_sign)`.
- Parts/component surface (`open.jlcpcb.com/overseas/openapi/component/*`, per `wavenumber-eng/supply-chain-monkey`, which reports live-probing success): scheme keyword `JOP` (no `-HMAC-SHA256` suffix), signature = **Base64** encoding of the raw HMAC-SHA256 digest bytes.
- These may be two different API generations/hosts with different conventions, or one of the three sources may simply be wrong/stale. **Since this issue is specifically about the Parts/component surface**, implement the `wavenumber-eng` variant (`JOP` + Base64) as the primary/default, but structure `jlcpcb_api.py` so the scheme keyword and digest encoding are each a single named constant/branch point that's trivial to flip during the owner's live smoke test (see Non-Goals/Test Plan). Record in code comments which variant is implemented and cite this issue for the discrepancy.
- Also unresolved: whether the signature is computed over the **request-path-only** (e.g. `/overseas/openapi/component/getComponentDetailByCode`) or a **full-path-including-`/api`-prefix** variant seen in one PCB-surface sample (`/api/overseas/openapi/pcb/calculate` signed as the full path while requested via a base URL that already included `/api/overseas/openapi`, i.e. seemingly double-counted in that one sample — likely an artifact of how that author structured their base URL, not a real requirement). Treat "sign whatever string is passed to `requests.post`'s path argument, using the client's own `base_url + path` convention consistently" as the safe default; the `wavenumber-eng` client (base URL `https://open.jlcpcb.com`, signs `path` = e.g. `/overseas/openapi/component/getComponentDetailByCode`) is the cleanest, most consistent sample and is the one to follow.

### Parts/component endpoints (per `wavenumber-eng/supply-chain-monkey` live-probing, host `open.jlcpcb.com`)

- `POST /overseas/openapi/component/getComponentDetailByCode` — body `{"componentCodes": ["C2040", ...]}` (**array required**; a bare string was rejected). Response `data` is a list of component objects with fields including `componentCode`, `componentModel`, `componentSpecification`, `firstTypeName`, `secondTypeName`, `libraryType`, `description`, `datasheetUrl`, `dataManualUrl`, `solderJointCount`, `priceRanges` (list of `{startQuantity, unitPrice}`), `stockCount`, `parameters`, `assemblyComponentFlag`, `eccnCode`, `rohsFlag`. This is the endpoint to use for `lookup()`/`lookup_many()` (maps cleanly onto the existing `Part` dataclass in `src/kicad_tools/parts/models.py`: `lcsc_part`, `mfr_part`, `manufacturer`, `description`, `package`, `stock`, `prices` (`PartPrice(quantity, unit_price)`), `is_basic`/`is_preferred` from `libraryType`, `datasheet_url`).
- `POST /overseas/openapi/component/getComponentInfos` — body `{"lastKey": "..."}` (optional, for pagination) — a **paginated feed/list**, not a keyword/MPN search; extra fields like `lcscPart`/`mfrPart`/`componentCode` were observed to be silently ignored by the live API in that source's probing (i.e. do not implement `search()` against this endpoint expecting query filtering to work — confirm with a live smoke call before wiring it into `LCSCClient.search()`).
- `POST /overseas/openapi/component/getPrivateComponentLibrary` — private/authenticated inventory feed; out of scope for this issue (no anonymous-equivalent to preserve parity with).
- **No official keyword/MPN search endpoint has been confirmed to exist.** `LCSCClient.search()` should therefore keep using the existing anonymous-scrape path (and/or the offline catalog from #4108) even when official keys are present, unless/until the live smoke proves otherwise. Only `lookup()`/`lookup_many()` (single/batch lookup-by-code) gain the official-API tier in this issue. Document this limitation in the README section.
- Envelope shape observed: top-level `{"code": 200, "success": true, "data": ...}` on success; non-200/non-`success` is a business-level failure with a `message` field — map to a distinct exception (see error-mapping below), not silently to `None`, since `code != 200` here means something different from JLC's anonymous-API `code != 200` (rate limit vs. auth vs. bad input are all plausible and should not be conflated).

### Env-var / credential-loading convention (verified against repo precedent)

- The repo's established pattern is **plain `os.environ.get(...)`**, read lazily inside `__init__`/constructors when an explicit argument is not passed — see `src/kicad_tools/datasheet/manager.py:82` (`OCTOPART_API_KEY`, the closest existing analog: an optional third-party API key, env-first, constructor-overridable, feature is a no-op without it). There is **no dotenv loader anywhere in `src/`** and no `python-dotenv` dependency in `pyproject.toml` — do not add one. Users who want `.env` support already get it for free from their shell/direnv/etc.; kicad-tools itself only ever reads `os.environ`.
- `.env` / `.env.*` are already gitignored on `origin/main` (`.gitignore` lines 65-67: `.env`, `.env.*`, `!.env.example`) — this landed already; this issue does not need to touch `.gitignore`, only add the new `.env.example` file (not yet present on `origin/main` — `git ls-tree origin/main` confirms no `.env.example` exists today).
- Env var names to implement, matching the issue's original contract and independently corroborated by `wavenumber-eng/supply-chain-monkey`'s `.env.template` (identical names): `JLCPCB_APP_ID`, `JLCPCB_ACCESS_KEY`, `JLCPCB_SECRET_KEY`. All three must be present (non-empty after `.strip()`) for the official-API tier to activate; if any is missing, the tier is skipped silently (falls through to the next tier) — this is not an error, it's the "keyless" path.
- `JLCPCB_TOKENIZATION_KEY` (mentioned in the original issue sketch as "optional"): **not corroborated by any of the three third-party references** — none of them use or mention a tokenization key for the Parts/component surface (it may belong to a payments/checkout flow, unrelated to part lookup). **Recommend dropping it from this issue's scope entirely** unless the owner's live smoke shows it's required for the component endpoints; do not build unused plumbing for it. If kept for forward-compatibility, it must be fully inert (read but unused) and clearly commented as speculative.

## Credential contract (BYO-key)

- Env vars (never committed; `.env` is gitignored): `JLCPCB_APP_ID`, `JLCPCB_ACCESS_KEY`, `JLCPCB_SECRET_KEY`. (Drop `JLCPCB_TOKENIZATION_KEY` per the research above unless the owner's smoke test proves it's needed.)
- No config-file convention exists elsewhere in the repo for API credentials (`OCTOPART_API_KEY` is env-only) — do not invent one; env-only is consistent with precedent.
- **Feature is inert without keys**: no behavior change for keyless users (anonymous scrape endpoints + offline catalog remain the defaults). With all three keys present, the official API becomes the preferred parts-lookup tier for `lookup()`/`lookup_many()` only, with graceful fallback down the existing chain (official API -> anonymous scrape API -> offline catalog -> miss).
- REAL KEYS MUST NEVER APPEAR in code, tests, fixtures, logs, or committed docs. Tests use fake credentials; signing is verified against a known-answer test vector computed by hand from the documented algorithm (see Test Plan), not against the live service.
- Add `.env.example` with placeholder values (e.g. `JLCPCB_APP_ID=your-app-id`, `JLCPCB_ACCESS_KEY=your-access-key`, `JLCPCB_SECRET_KEY=your-secret-key`) documenting the contract.

## Scope

1. `src/kicad_tools/parts/jlcpcb_api.py` (new): signed-request client (`JLCOpenAPIClient`-equivalent) —
   - `open.jlcpcb.com` base URL, HMAC-SHA256 signature construction per the "best available reference" above (`JOP` scheme, Base64 digest, as primary; keep the scheme keyword and digest-encoding as easily-flippable named constants given the unresolved discrepancy).
   - `get_component_detail_by_codes(codes: list[str])` wrapping `POST /overseas/openapi/component/getComponentDetailByCode`, mapped to the existing `Part` dataclass (reuse/extract `_parse_component`-style logic from `lcsc.py` where the field names overlap; do not duplicate wholesale if a shared helper is cleaner — the response field names differ from the anonymous API's, e.g. `componentModel` vs `componentModelEn`, `datasheetUrl`/`dataManualUrl` vs `dataManualUrl`, `priceRanges`/`startQuantity`/`unitPrice` vs `prices`/`startNumber`/`productPrice` — write a distinct parser, do not try to force-reuse `_parse_component`).
   - Distinct exception types for: missing/incomplete credentials (inert, not an error — handled by tier-skip, not exception, at the `LCSCClient` layer), bad signature / auth rejection (HTTP 401/403-equivalent business code), IP-not-whitelisted (the developer portal has an IP Whitelisting feature per the owner's issue text; if it rejects, the response is presumably a distinct business `code`/`message` — since this can't be confirmed without the live smoke, implement a best-effort message-substring/code heuristic and document it as unconfirmed, refined by the owner's local smoke test), quota/rate-limit exceeded.
2. `src/kicad_tools/parts/lcsc.py`: add the official-API tier ahead of the existing tiers in `lookup()`/`lookup_many()` (composes with #4117's `_get_catalog()`-style seam — see Context). Tier order: official API (when 3 keys present) -> anonymous scrape API -> offline catalog (#4108) -> miss. `search()` is unchanged (no official search endpoint confirmed — see above).
3. Docs: README section "Using your own JLCPCB API key" (env vars, where to get them — developer portal, IP whitelisting caveat, what tier it unlocks, the search-not-covered caveat) + `.env.example`.
4. Tests (`tests/parts/test_jlcpcb_api.py`, new — or alongside wherever #4117 lands `tests/parts/`): fake-key signing known-answer test (hand-computed HMAC over a fixed method/path/timestamp/nonce/body with a fixed fake secret — assert the exact hex-or-base64 signature string), tier-selection tests (keyless byte-identical; all-three-keys-present prefers official tier; partial-keys falls through as if keyless), error-mapping tests (mock HTTP responses for bad signature / not-whitelisted / quota, assert distinct exception types/messages). **NO live calls in CI** — all HTTP interaction mocked/stubbed (follow #4117's `sys.modules` `requests` stub pattern if convenient, or `unittest.mock`/`responses`).
5. Live smoke verification (owner-only, not CI, not part of the PR diff): the owner's local `.env` has working credentials for app `kicad-tools-developer` (id `599660863674093570`) — the owner MAY run a one-off local smoke query (`get_component_detail_by_codes(["C2040"])` or similar, a real public LCSC part) to validate the signing implementation end-to-end, but the PR itself must not embed any credential material, and CI must not attempt this. If the signing spec's unresolved discrepancy (scheme keyword / digest encoding / path scope) turns out wrong against the live API, that's an expected, documented follow-up — ship behind the tier-skip-safe default (inert without keys) and let the owner's live smoke drive a fast-follow fix; do not block this issue's merge on live confirmation.

## Non-Goals (explicit, carried over + refined)

- No `PartProvider` protocol / pluggable-backend abstraction.
- No official keyword/MPN `search()` — not confirmed to exist; `search()` stays on the existing anonymous/offline path.
- No PCB-ordering, gerber-upload, or quote endpoints (`/pcb/*`) — those are a different SDK/API surface entirely and out of scope for a *parts* backend.
- No `JLCPCB_TOKENIZATION_KEY` plumbing unless the owner's live smoke proves it's required (see above).
- No config-file credential source — env-only, matching `OCTOPART_API_KEY` precedent.

## Acceptance criteria

- Keyless behavior byte-identical (CI proof: existing `tests/test_parts.py` / `tests/parts/*` suites green, no new required env vars).
- With fake keys in env (all three of `JLCPCB_APP_ID`/`JLCPCB_ACCESS_KEY`/`JLCPCB_SECRET_KEY` set): client constructs a correctly-signed request per a hand-computed known-answer test vector; tier order is official -> anonymous -> offline -> miss; each failure mode (bad signature, IP-not-whitelisted, quota/rate-limit, generic business error) maps to a distinct, actionable exception/message.
- With only some of the three keys set: behaves exactly as keyless (silent tier-skip, not an error).
- `.env.example` committed with placeholder values only; no `.gitignore` changes needed (already covers `.env*`).
- No credential material anywhere in the diff (grep the PR diff for the owner's actual `JLCPCB_ACCESS_KEY`/`JLCPCB_SECRET_KEY`/`JLCPCB_APP_ID` values as a final check — must not appear).
- `search()` behavior is unchanged (still anonymous/offline only) — regression-tested.
- ruff / mypy / pytest all pass on touched files; no new mypy error class introduced (there are pre-existing mypy errors in `parts/` per #4117's PR notes — don't try to fix unrelated ones).

## Affected Files

- `src/kicad_tools/parts/jlcpcb_api.py` (new)
- `src/kicad_tools/parts/lcsc.py` (modify — exists on `origin/main`; may already carry #4117's tiering scaffolding by the time this is built, see Context)
- `src/kicad_tools/parts/__init__.py` (modify — export the new client/exceptions, matching the `JlcpartsCatalog` export pattern from #4117)
- `.env.example` (new)
- `README.md` (modify — new "Using your own JLCPCB API key" section)
- `tests/parts/test_jlcpcb_api.py` (new; `tests/parts/` directory itself is created by #4117 — if #4117 hasn't merged yet, create the directory, matching #4117's `tests/parts/test_jlcparts_catalog.py` layout)

## Test Plan

```
uv run pytest tests/parts/test_jlcpcb_api.py -v          # new: signing KAT, tier selection, error mapping
uv run pytest tests/parts tests/test_parts.py tests/test_lcsc_models.py \
  tests/test_bom_lcsc_merge.py tests/test_bom_enrich.py   # full parts/LCSC regression, keyless-unchanged proof
uv run ruff check src/kicad_tools/parts/jlcpcb_api.py src/kicad_tools/parts/lcsc.py tests/parts/test_jlcpcb_api.py
uv run ruff format --check <same files>
uv run mypy src/kicad_tools/parts/jlcpcb_api.py   # no new error class vs. pre-existing parts/ baseline
git diff origin/main... | grep -iE "JLCPCB_(ACCESS|SECRET)_KEY\s*=\s*[A-Za-z0-9]" # must return nothing (no leaked real key material)
```

No live-API calls anywhere in the automated test plan. Live-smoke verification (owner-only, local, outside CI) is documented in Scope item 5.

rjwalters:
## Summary

Adds an authenticated **JLCPCB open-platform** (`open.jlcpcb.com`) parts client
that users unlock with their **own** API credentials (registered at the JLCPCB
developer portal). kicad-tools ships only the signed client — no keys are
bundled. This is Phase 2 of #4108: it prepends a new *official-API* tier ahead
of the anonymous scrape API and the offline jlcparts catalog from #4117.

**The feature is entirely inert without keys** — keyless behavior is
byte-for-byte unchanged, proven by the existing parts/LCSC suites staying green.

## Changes

- **`src/kicad_tools/parts/jlcpcb_api.py`** (new): `JLCOpenAPIClient` with
  HMAC-SHA256 request signing (`METHOD\nPATH\nTIMESTAMP\nNONCE\nBODY\n`, epoch
  seconds, 32-char hex nonce), `get_component_detail_by_codes()` mapping the
  official response (`componentModel`/`priceRanges`/`libraryType` …) onto the
  existing `Part` dataclass, and distinct exceptions (`JLCAuthError`,
  `JLCIPNotWhitelistedError`, `JLCQuotaError`, base `JLCAPIError`). The two
  ambiguous signing parameters (`AUTH_SCHEME`, `SIGNATURE_ENCODING`) are
  isolated as single flip-point constants.
- **`src/kicad_tools/parts/lcsc.py`**: `lookup()` / `lookup_many()` gain a
  tier-0 official-API step. Activated only when **all three**
  `JLCPCB_APP_ID` / `JLCPCB_ACCESS_KEY` / `JLCPCB_SECRET_KEY` env vars are set
  (plain `os.environ`, no dotenv — matches the `OCTOPART_API_KEY` precedent).
  Partial/missing keys → silent tier-skip. `search()` is unchanged (no official
  search endpoint exists).
- **`src/kicad_tools/parts/__init__.py`**: export the client + exceptions.
- **`.env.example`** (new): placeholder-only credential contract.
- **`README.md`**: "Using your own JLCPCB API key" section (env vars, portal,
  IP-whitelist caveat, tier order, search-not-covered limitation).
- **`tests/parts/test_jlcpcb_api.py`** (new): known-answer signing vector, tier
  selection, and error mapping — all HTTP mocked, **zero live calls in CI**.

## Live smoke (owner-only, local, not in CI)

A one-off local smoke against the real API validated the signing end-to-end:
**base64 digest is correct** (signature verified — HTTP 403 "insufficient
permissions"), while hex was rejected (HTTP 401 "signature verify failed"). The
default (`JOP` + base64) is therefore confirmed. The residual 403 is an
owner-side developer-portal matter (enable the Parts API product for the app /
whitelist the calling IP), not a client bug — documented in the module. No
credential material was written to any file, log, or this PR.

| Keyless behavior byte-identical | ✅ | `tests/parts` + `test_parts`/`test_lcsc_models`/`test_bom_*` — 248 passed |
| Fake keys → correctly-signed request (known-answer) | ✅ | `test_sign_known_answer_base64` pins the literal base64 signature |
| Tier order official → anonymous → offline → miss | ✅ | `test_lcsc_lookup_prefers_official_tier` / `_official_miss_falls_through` |
| Distinct actionable errors (sig/auth, IP/permission, quota) | ✅ | error-mapping tests incl. the two confirmed live messages |
| Partial keys behave as keyless | ✅ | `test_lcsc_client_partial_keys_behaves_keyless` |
| `.env.example` placeholders only, no `.gitignore` change | ✅ | file added; `.env*` already gitignored on main |
| No credential material in diff | ✅ | issue-mandated grep returns empty |
| `search()` unchanged | ✅ | `test_lcsc_search_never_uses_official` |
| ruff / mypy baseline / pytest pass | ✅ | ruff clean, baseline gate "No new type errors", 248 passed |

## Test Plan

```
uv run pytest tests/parts/test_jlcpcb_api.py -v              # 32 passed
uv run pytest tests/parts tests/test_parts.py tests/test_lcsc_models.py \
  tests/test_bom_lcsc_merge.py tests/test_bom_enrich.py      # 248 passed
uv run ruff check / format --check <touched files>           # clean
uv run python scripts/ci/check_mypy_baseline.py              # no new type errors
git diff origin/main... | grep -iE "JLCPCB_(ACCESS|SECRET)_KEY\s*=\s*[A-Za-z0-9]"  # empty
```

Closes #4118

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,21 @@
+# kicad-tools optional API credentials (BYO key).
+#
+# Copy this file to `.env` and fill in real values, OR export these variables
+# in your shell / direnv. kicad-tools reads them via plain os.environ -- it does
+# NOT load `.env` itself, so you need a shell/direnv/dotenv-runner to source it.
+# `.env` is gitignored; never commit real credentials.
+
+# --- Official JLCPCB open-platform API (Parts backend, issue #4118) ---
+# Register an app at the JLCPCB developer portal to obtain these. All THREE
+# must be set (non-empty) for the official-API tier to activate; if any is
+# missing, kicad-tools silently falls back to the anonymous scrape API and the
+# offline jlcparts catalog (keyless behavior is unchanged).
+#
+# NOTE: the developer portal has an IP Whitelisting feature -- if you enable it,
+# your machine's public IP must be whitelisted or requests are rejected.
+JLCPCB_APP_ID=your-app-id
+JLCPCB_ACCESS_KEY=your-access-key
+JLCPCB_SECRET_KEY=your-secret-key
+
+# --- Octopart (optional datasheet source) ---
+# OCTOPART_API_KEY=your-octopart-api-key
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,8 @@ boards/*/*/output/board.json
 .loom/state.json
 .loom/*.log
 .loom/*.sock
+
+# Local credentials (BYO API keys -- issue #4118). NEVER commit real keys.
+.env
+.env.*
+!.env.example
--- a/README.md
+++ b/README.md
@@ -462,6 +462,67 @@ for table in tables:
     print(table.to_markdown())
 ```
 
+### Parts Lookup (LCSC / JLCPCB)
+
+Look up LCSC part numbers, check BOM availability, and pull pricing/stock:
+
+```python
+from kicad_tools.parts import LCSCClient
+
+client = LCSCClient()
+part = client.lookup("C2040")
+if part:
+    print(f"{part.mfr_part}: {part.stock} in stock, best ${part.best_price:.4f}")
+```
+
+Part lookups resolve through a tiered chain (each tier falls back to the next):
+
+1. **Local response cache** — previously fetched parts.
+2. **Official JLCPCB open-platform API** — only when you supply your own API key
+   (see below). Off by default; inert without keys.
+3. **Anonymous scrape API** — the public JLCPCB web endpoints (requires the
+   `parts` extra: `pip install "kicad-tools[parts]"`).
+4. **Offline jlcparts catalog** — a locally synced mirror
+   (`kct parts sync-catalog`), usable fully offline.
+
+#### Using your own JLCPCB API key
+
+kicad-tools can talk to the **official JLCPCB open-platform API** using
+credentials you register yourself at the JLCPCB developer portal
+(<https://jlcpcb.com/> → developer/open platform). This is strictly opt-in: you
+bring your own key, kicad-tools ships only the signed client. **Without keys,
+behavior is unchanged** — the official tier is simply skipped.
+
+Set all three environment variables (kicad-tools reads them via `os.environ`;
+it does **not** load a `.env` file itself, so use your shell, `direnv`, or a
+dotenv runner — see [`.env.example`](.env.example)):
+
+```bash
+export JLCPCB_APP_ID="your-app-id"
+export JLCPCB_ACCESS_KEY="your-access-key"
+export JLCPCB_SECRET_KEY="your-secret-key"
+```
+
+All three must be set (and non-empty) to activate the official tier; if any is
+missing it stays inert. Once set, `LCSCClient.lookup()` / `lookup_many()` prefer
+the official API for part-detail-by-code, falling back down the chain above on
+any failure.
+
+Notes and caveats:
+
+- **What it unlocks:** authenticated *part detail lookup by LCSC code* only.
+  There is **no confirmed official keyword/MPN search endpoint**, so
+  `LCSCClient.search()` always uses the anonymous/offline path even with keys.
+- **IP whitelisting:** the developer portal offers an IP-whitelist feature. If
+  you enable it, your machine's public IP must be whitelisted or requests are
+  rejected (surfaced as a distinct, actionable error).
+- **Never commit credentials.** `.env` is gitignored; the secret key is used
+  only as HMAC key material and is never transmitted.
+- The request-signing scheme is not first-party-documented by JLCPCB; the client
+  implements the best-available community-reverse-engineered variant for the
+  Parts surface, with the two ambiguous parameters isolated as single
+  flip-points in `parts/jlcpcb_api.py` (see that module and issue #4118).
+
 ## CLI Commands
 
 ### Unified CLI (`kct` or `kicad-tools`)
--- a/src/kicad_tools/parts/__init__.py
+++ b/src/kicad_tools/parts/__init__.py
@@ -67,6 +67,14 @@ from .importer import (
     PartImporter,
 )
 from .jlcparts_catalog import JlcpartsCatalog, get_catalog_path, sync_catalog
+from .jlcpcb_api import (
+    JLCAPIError,
+    JLCAuthError,
+    JLCCredentials,
+    JLCIPNotWhitelistedError,
+    JLCOpenAPIClient,
+    JLCQuotaError,
+)
 from .lcsc import LCSCClient, LCSCForbiddenError, RateLimiter
 from .models import (
     BOMAvailability,
@@ -96,6 +104,13 @@ __all__ = [
     "JlcpartsCatalog",
     "get_catalog_path",
     "sync_catalog",
+    # Official JLCPCB open-platform API (BYO key)
+    "JLCOpenAPIClient",
+    "JLCCredentials",
+    "JLCAPIError",
+    "JLCAuthError",
+    "JLCIPNotWhitelistedError",
+    "JLCQuotaError",
     # Composition
     "ComposedPart",
     "ComposedPartStore",
--- a/src/kicad_tools/parts/jlcpcb_api.py
+++ b/src/kicad_tools/parts/jlcpcb_api.py
@@ -1,0 +1,608 @@
+"""
+Official JLCPCB open-platform (BYO-key) parts client.
+
+This module speaks the *authenticated* JLCPCB open-platform REST API directly.
+Users bring their own credentials (registered at the JLCPCB developer portal);
+kicad-tools ships only the signed client. When the three credential env vars
+are absent the feature is entirely inert -- :class:`JLCOpenAPIClient` is never
+constructed and :class:`~kicad_tools.parts.lcsc.LCSCClient` behaves byte-for-byte
+as it did without keys (anonymous scrape API + offline jlcparts catalog).
+
+Credential contract (env-only, matching the ``OCTOPART_API_KEY`` precedent in
+``datasheet/manager.py``; no dotenv dependency is added -- users get ``.env``
+support from their shell/direnv):
+
+* ``JLCPCB_APP_ID``     -- application id (public)
+* ``JLCPCB_ACCESS_KEY`` -- access key (public)
+* ``JLCPCB_SECRET_KEY`` -- secret key (HMAC key material, never transmitted)
+
+All three must be present and non-empty (after ``.strip()``) for the tier to
+activate; a partial set behaves exactly like the keyless path.
+
+Signing scheme -- IMPORTANT provenance note
+--------------------------------------------
+JLCPCB does **not** publish a first-party REST reference for the signing
+algorithm (only Java-only SDKs, which third-party live-probing reports are
+stale vs. the live API). The construction below is the convergent result of
+three independent reverse-engineering projects (see issue #4118 for citations):
+``wavenumber-eng/supply-chain-monkey`` (Parts/component surface, live-probed),
+``Jackster/JLCPCB-API`` and ``mattpainter701/kicad_automations`` (PCB-ordering
+surface). All three agree on the string-to-sign construction:
+
+    string_to_sign = f"{METHOD}\n{PATH}\n{TIMESTAMP}\n{NONCE}\n{BODY}\n"
+
+with uppercase HTTP method, the request path, Unix epoch **seconds** as a
+decimal string, a 32-char hex nonce, and the exact compact-JSON request body
+(empty string for no-body requests), each field newline-terminated including
+a trailing newline after ``BODY``.
+
+The sources **disagree** on two points that could not be resolved without a
+live smoke test, so each is isolated here as a single named constant that can
+be flipped without touching any request logic:
+
+* :data:`AUTH_SCHEME` -- the ``Authorization`` scheme keyword. The Parts
+  surface (this module's target) uses ``"JOP"``; the PCB surface uses
+  ``"JOP-HMAC-SHA256"``.
+* :data:`SIGNATURE_ENCODING` -- how the raw HMAC-SHA256 digest is encoded.
+  The Parts surface uses Base64; the PCB surface uses lowercase hex.
+
+This module implements the **Parts variant** (``JOP`` + Base64) as the default,
+because this issue targets the ``open.jlcpcb.com`` component endpoints, and the
+``wavenumber-eng`` client is the only source that reports live-probing success
+against exactly those endpoints. If the owner's local live smoke shows the live
+API wants the other variant, flip the two constants below -- no other change is
+needed.
+
+Live-smoke result (2026-07, issue #4118)
+----------------------------------------
+A one-off local smoke against ``open.jlcpcb.com`` with the owner's real key
+**confirmed the Base64 digest encoding is correct**:
+
+* Base64 digest (either scheme keyword) -> HTTP 403,
+  ``"API insufficient permissions, access denied"`` -- i.e. the signature
+  *verified*, but the app was not permitted (a portal/permission/IP matter).
+* Hex digest -> HTTP 401, ``"The request signature verify failed"`` -- the
+  signature was rejected.
+
+So :data:`SIGNATURE_ENCODING` = ``"base64"`` is validated, and the scheme
+keyword does not affect signature verification. The remaining 403 is an
+owner-side developer-portal configuration matter (enable the Parts API product
+for the app / whitelist the calling IP), not a client bug -- which is why the
+feature ships inert-without-keys and defers that step to the owner.
+"""
+
+from __future__ import annotations
+
+import base64
+import hashlib
+import hmac
+import json
+import logging
+import os
+import secrets
+import time
+from datetime import datetime
+from typing import Literal
+
+from .lcsc import _categorize_part, _guess_package_type
+from .models import Part, PartPrice
+
+logger = logging.getLogger(__name__)
+
+# --------------------------------------------------------------------------
+# Endpoint configuration (Parts/component surface, host open.jlcpcb.com)
+# --------------------------------------------------------------------------
+JLC_OPENAPI_BASE = "https://open.jlcpcb.com"
+
+# POST body {"componentCodes": ["C2040", ...]} -- array required.
+COMPONENT_DETAIL_PATH = "/overseas/openapi/component/getComponentDetailByCode"
+
+# --------------------------------------------------------------------------
+# Signing spec flip-points (see module docstring / issue #4118).
+#
+# These two constants are the ONLY places the Parts-vs-PCB signing variants
+# differ. A live smoke test that fails on the default (Parts) variant is
+# corrected by flipping these -- do not touch _sign()/_build_auth_header().
+# --------------------------------------------------------------------------
+# Scheme keyword placed at the front of the Authorization header.
+#   Parts surface (wavenumber-eng, live-probed): "JOP"
+#   PCB surface (Jackster, mattpainter701):      "JOP-HMAC-SHA256"
+AUTH_SCHEME = "JOP"
+
+# Encoding applied to the raw HMAC-SHA256 digest bytes.
+#   Parts surface: "base64"
+#   PCB surface:   "hex"
+SIGNATURE_ENCODING = "base64"
+
+# Environment variable names (independently corroborated by
+# wavenumber-eng/supply-chain-monkey's .env.template).
+ENV_APP_ID = "JLCPCB_APP_ID"
+ENV_ACCESS_KEY = "JLCPCB_ACCESS_KEY"
+ENV_SECRET_KEY = "JLCPCB_SECRET_KEY"
+
+
+class JLCAPIError(Exception):
+    """Base class for official JLCPCB open-platform API failures.
+
+    Distinct from :class:`~kicad_tools.parts.lcsc.LCSCForbiddenError` (which
+    describes the *anonymous* scrape API) and from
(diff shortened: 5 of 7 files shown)


rjwalters:
✅ **Approved!**

Phase 2 official JLCPCB open-platform (BYO-key) backend. Thorough, well-isolated, and inert-without-keys as designed.

## Security sweep (credentials PR — extra scrutiny)

- **Real key material: NONE in diff.** Grepped the entire `origin/main...HEAD` diff for the owner's real key prefixes (`35be` access, `qPbM` secret) and for any `(ACCESS|SECRET_KEY|APP_ID)=<16+ alnum>` assignment — all empty. `.env` is not in the diff. `.env.example` contains only obvious placeholders (`your-app-id` / `your-access-key` / `your-secret-key`). Test fixtures use clearly-fake creds (`fake-secret-key-do-not-use`).
- **Secret hygiene in errors/logs: clean.** `secret_key` is referenced only as HMAC key material (`_sign` → `hmac.new`) and is never interpolated into a log line, exception message, or the `Authorization` header. `_build_auth_header` emits only app_id/access_key/nonce/timestamp/signature. Tests assert `FAKE_SECRET_KEY not in header` and `not in call["data"]`.
- **Nonce source: `secrets.token_hex(16)`** (CSPRNG, 32 hex chars) — not `random.*`.
- **`.env` gitignore gap — FIXED during review.** The PR body claimed "`.env*` already gitignored on main", but `.gitignore` had **no** env entry (verified on `origin/main`). Since `.env.example` tells users to "Copy this file to `.env` and fill in real values", following that instruction would leave a real `JLCPCB_SECRET_KEY` at risk of `git add .`. I added `.env` / `.env.*` with a `!.env.example` negation (commit `cb825fe`); verified `.env` is now ignored while `.env.example` stays tracked. This was the one substantive finding and it's resolved.

## Signing correctness

- **Known-answer vector is independent and correct.** `test_sign_known_answer_base64` recomputes the digest in-test AND pins the literal base64 `bCac32rIS+6RPiHcQxfWJVvYr1+nUODKc2OKtgwPups=`. I reproduced this value out-of-band with a standalone `hmac`/`base64` script over the documented base string `METHOD\nPATH\nTIMESTAMP\nNONCE\nBODY\n` (trailing newline after BODY) — it matches. The base-string shape is separately pinned by `test_string_to_sign_shape`, including uppercasing of the method.
- Base64 default (`SIGNATURE_ENCODING`) matches the owner's live smoke (base64 → 403 permission, hex → 401 sig-verify). The two ambiguous parameters (`AUTH_SCHEME`/`SIGNATURE_ENCODING`) are cleanly isolated as single flip-points; `test_encode_signature_hex_variant` covers the alternate.

## Keyless byte-identity

- `_get_official_client()` returns `None` (memoized `False` sentinel) with no/partial keys or missing `requests`; tier-0 is fully skipped and `lookup()`/`lookup_many()` fall through unchanged. `test_lcsc_client_keyless_does_not_build_official` + `_partial_keys_behaves_keyless` prove it. `search()` never touches the official tier (`test_lcsc_search_never_uses_official`). Official-tier misses and exceptions fall through to scrape rather than raising.

## Error classification

- 401→`JLCAuthError`, 403/permission/whitelist→`JLCIPNotWhitelistedError`, 429/quota→`JLCQuotaError`, else base `JLCAPIError` — all subclass `JLCAPIError`, and the two confirmed live messages are pinned in tests.

## Verification

- **Local**: `tests/parts/test_jlcpcb_api.py` + full parts stack — **248 passed**. `ruff check`/`format --check` clean. mypy baseline gate: **no new type errors** (18 baseline errors actually fixed).
- **CI (run 29282060241, HEAD cb825fe including my gitignore fix): all 16 checks SUCCESS**, mergeStateStatus **CLEAN**.

## Test Execution (from PR Test Plan)

1. `pytest tests/parts/test_jlcpcb_api.py` — ✅ passes (part of the 248)
2. Full parts-stack regression (248) — ✅ Executed: 248 passed
3. ruff check/format — ✅ Executed: clean
4. mypy baseline — ✅ Executed: no new type errors
5. credential-leak grep — ✅ Executed: empty (extended to real-key-prefix scan)
6. Live API smoke — ⏭️ Skipped: owner-local, requires real credentials (out of CI scope, as documented)

Solid work. Closes #4118.

PR #4122, opened 2026-07-13
fix(install): use git+URL direct reference instead of removed 'uv add --git' flags

### LINKED ISSUE #4121: install-kct.sh git mode broken on uv >= 0.5: 'uv add --git' flag removed
Found during the v0.15.0 post-release installs to softstart/chorus: Stage 5's `uv add kicad-tools --git URL --tag TAG` fails on current uv (0.11.x) with "unexpected argument '--git'". The supported cross-version spelling is a PEP 508 direct reference: `uv add "kicad-tools @ git+URL@REF"`. Fix verified end-to-end by two successful v0.15.0 installs (softstart, chorus) after patching. PR follows.

rjwalters:
Closes #4121

Stage 5 of scripts/install-kct.sh used `uv add kicad-tools --git URL --tag/--rev REF`, which current uv rejects ('unexpected argument --git'). Replaced with the PEP 508 direct-reference spelling `uv add "kicad-tools @ git+URL@REF"` for both the tag and rev branches.

Verified end-to-end post-release: two successful `--tag v0.15.0` installs (softstart and chorus), both consumers now at kicad-tools 0.15.0 with the C++ backend built.

### DIFF
--- a/scripts/install-kct.sh
+++ b/scripts/install-kct.sh
@@ -325,9 +325,11 @@ run_uv_add() {
     # CI-no-network constraint). The operator runs `uv sync` afterwards.
     ( cd "$TARGET" && uv add "$RESOLVED_PATH" --editable --frozen )
   elif [[ -n "$GIT_TAG" ]]; then
-    ( cd "$TARGET" && uv add kicad-tools --git "$KCT_GIT_URL" --tag "$GIT_TAG" )
+    # uv >= 0.5 dropped `--git`; the supported spelling is a git+ URL with
+    # the ref appended (works across uv versions).
+    ( cd "$TARGET" && uv add "kicad-tools @ git+${KCT_GIT_URL}@${GIT_TAG}" )
   else
-    ( cd "$TARGET" && uv add kicad-tools --git "$KCT_GIT_URL" --rev "$GIT_REV" )
+    ( cd "$TARGET" && uv add "kicad-tools @ git+${KCT_GIT_URL}@${GIT_REV}" )
   fi
 }
 


rjwalters:
✅ **Approved!**

Small, correct post-release installer fix. Reviewed all five points:

1. **Shell + PEP 508 quoting — correct.** The whole spec `"kicad-tools @ git+${KCT_GIT_URL}@${GIT_REV}"` is a single double-quoted argv element, so the internal spaces around `@` survive and `uv add` receives one direct-reference argument. Brace-form `${VAR}` concatenation is unambiguous against the trailing `@REF`.

2. **Idempotency re-detection holds — verified empirically.** `uv_source_state()` (line 306-309) matches on the `[tool.uv.sources]` line via three substring checks: contains `git`, contains `$KCT_GIT_URL`, and contains `pin=${GIT_TAG:-$GIT_REV}`. It keys on *what uv writes*, not on the CLI form. Ran `uv 0.11.14 add "kicad-tools @ git+.../kicad-tools@v0.15.0" --frozen`; uv writes:
   ```
   kicad-tools = { git = "https://github.com/rjwalters/kicad-tools", rev = "v0.15.0" }
   ```
   All three substrings present (`git`, URL, `v0.15.0` inside `rev = "..."`) → `match` → re-run is a clean no-op. The `--git` form wrote the same shape, so no mis-detection across the switch. Whole-word dep regex (line 280) also still detects it.

3. **--dry-run path — no regression.** `do_action` prints the `$desc` string (`uv add kicad-tools ($SOURCE_MODE: $SOURCE_REF)`), which is sourced from `SOURCE_MODE`/`SOURCE_REF`, unaffected by the changed lines. Behavior is identical to before.

4. **CI coverage — no installer job exists** in `.github/workflows/` (grep for `install-kct` is empty), so no automated coverage regressed. `bash -n` passes.

5. **CI green + merge CLEAN.** All 16 checks pass; `mergeStateStatus` is CLEAN. The changed lines are also empirically validated end-to-end (two `--tag v0.15.0` installs into softstart and chorus, both at kicad-tools 0.15.0).

PR description uses `Closes #4121`. Clean 4/2 single-file diff. Ready to merge.

PR #4125, opened 2026-07-13
fix(parts): stream-extract split jlcparts archives Python zipfile rejects

### LINKED ISSUE #4123: parts: sync-catalog cannot extract the real jlcparts dataset — Python zipfile rejects split archives
**Found live 2026-07-13** running `kct parts sync-catalog` for real (first time against the actual dataset — CI never downloads it, by design, so the fixture tests couldn't catch this).

## Symptoms & diagnosis (all verified live)

1. The download half works (13 segments, ~618MiB → concatenated blob), but `_extract_catalog` dies: **`zipfile.BadZipFile: zipfiles that span multiple disks are not supported`**. The jlcparts dataset is a true `zip -s` split archive (leading `PK\x07\x08` spanning marker; multi-disk central directory). Python's `zipfile` refuses these categorically.
2. The CLI also swallowed the traceback (log ends at "Extracting SQLite database ..." with exit 1 and no error text) — secondary fix: surface extraction errors.
3. Repair via macOS `/usr/bin/zip -s 0` is NOT viable: Info-ZIP 3.0 silently corrupts the zip64 output (>4GB member → 100MB "combined.zip" with overlapped entries).

## Validated fix (ran successfully against the real dataset)

Pure-Python streaming extraction that never consults the multi-disk central directory:
- Concatenate segments in order (z01..zNN, cache.zip last) — or stream them.
- At the head: skip the 4-byte `PK\x07\x08` spanning marker if present; parse the `PK\x03\x04` local file header (compression method must be 8/deflate; read name/extra lengths to find data start).
- Stream `zlib.decompressobj(-15)` over everything after the header until `d.eof` — no sizes needed (works with data-descriptor and zip64 entries).
- Write via tmp + `os.replace` (keep the #4117 atomicity).

Live result: extracted 5.14 GiB → `jlcparts.sqlite3`, **7,157,042 components**, `lookup("C25804")` returns the correct part, and the `LCSCClient` fallback works end-to-end (notably: the anonymous live endpoint now 404s upstream, so the offline tier is currently the ONLY working keyless lookup path — this fix is load-bearing).

## Scope

- Rework `_extract_catalog` (or add a split-aware path chosen when the blob starts with `PK\x07\x08`) per the validated approach; keep plain-zipfile handling for single-disk archives (the test fixtures).
- Surface extraction exceptions to the CLI (no silent exit-1).
- Tests: build a small REAL split archive in-fixture (create with `zipfile` then manually prepend the marker + patch? simpler: generate a single-member deflate zip, split its bytes into segments, prepend `PK\x07\x08`) so the split path is CI-covered; keep existing single-zip tests.
- Optionally: stream directly from the per-segment downloads without writing the concatenated blob twice (disk-friendlier), but not required for correctness.

## Acceptance

- `kct parts sync-catalog` completes against the real dataset shape (split-archive fixture in CI as proxy).
- Extraction failure prints an actionable error.
- Existing fixture tests pass; atomicity preserved.

rjwalters:
## Summary

`kct parts sync-catalog` could not extract the real yaqwsx/jlcparts dataset. The dataset is a true `zip -s` split archive: after its segments are concatenated the blob begins with the `PK\x07\x08` spanning marker and carries a multi-disk (zip64) central directory that Python's `zipfile` refuses categorically (`BadZipFile: zipfiles that span multiple disks are not supported`). CI never downloads the real dataset (by design), so the single-disk fixture tests never exercised this shape.

This adds a pure-Python streaming extraction path that never consults the multi-disk central directory, and keeps the existing `zipfile.ZipFile` path for single-disk archives (the test fixtures). The orchestrator validated the same approach live today against the real 618 MiB dataset (5.14 GiB SQLite, 7,157,042 components, `lookup("C25804")` correct).

## Changes

- `src/kicad_tools/parts/jlcparts_catalog.py`:
  - `_is_split_archive()` — detects the leading `PK\x07\x08` spanning marker.
  - `_extract_split_archive()` — skips the 4-byte marker, parses the `PK\x03\x04` local file header (compression method must be 8/deflate; reads name/extra lengths to find the data start), then streams `zlib.decompressobj(-15)` over the remaining bytes until `d.eof`. No member sizes are needed, so data-descriptor and zip64 entries work.
  - `_extract_catalog()` now routes: split → streaming; otherwise the single-disk `zipfile` path, with a fallback to streaming if `zipfile` raises the multi-disk `BadZipFile` on a markerless split. Both paths keep the tmp + `os.replace` atomic promotion from #4117.
- `tests/parts/fixtures/build_jlcparts_fixture.py`: `build_spanning_split_bytes()` / `build_spanning_split_dataset()` synthesize a genuine split-shaped archive in-process (single-member deflate zip with `PK\x07\x08` prepended — exactly segment-1 of a `zip -s` archive after concatenation).
- `tests/parts/test_jlcparts_catalog.py`: new tests for marker detection, byte-identical streaming extraction, the plain-zipfile fallback, non-deflate members raising a clear error, the multi-disk `BadZipFile` fallback, `sync_catalog` end-to-end over a split dataset, and CLI error surfacing.

| `sync-catalog` completes against the real dataset shape (split fixture in CI as proxy) | ✅ | `test_sync_catalog_extracts_true_split_dataset` assembles a split dataset and extracts a byte-identical, queryable catalog |
| Split-aware path chosen when blob starts with `PK\x07\x08`; plain-zipfile kept for single-disk | ✅ | `test_is_split_archive_*`, `test_spanning_marker_routes_to_streaming_not_zipfile`, `test_extract_plain_zip_still_works` |
| Extraction failure prints an actionable error (no silent exit-1) | ✅ | `test_cli_surfaces_extraction_error` asserts `rc == 1` and stderr contains the real error text |
| Existing fixture tests pass; atomicity preserved | ✅ | All prior tests green; failure tests assert no partial catalog is left behind |
| No downloads in tests | ✅ | All fixtures built in-process; `requests` is mocked |

## Test Plan

- `uv run pytest tests/parts/test_jlcparts_catalog.py` → 33 passed
- `uv run mypy src/kicad_tools/parts/jlcparts_catalog.py` → no errors in the changed module

## Notes on the streaming-path premise

Prepending `PK\x07\x08` alone does not make stdlib `zipfile` raise (it reads the central directory from the file tail and tolerates a leading prefix); the real dataset's multi-disk rejection comes from its zip64 end-of-central-directory locator, which needs a >4 GiB member to reproduce faithfully. The streaming path never reads the central directory, so it is driven entirely by the marker + local header — which is exactly what the fixture provides. The multi-disk `BadZipFile` fallback is covered by a dedicated test that forces the guard.

Closes #4123

### DIFF
--- a/src/kicad_tools/parts/jlcparts_catalog.py
+++ b/src/kicad_tools/parts/jlcparts_catalog.py
@@ -40,7 +40,9 @@ import json
 import logging
 import os
 import sqlite3
+import struct
 import zipfile
+import zlib
 from datetime import datetime
 from pathlib import Path
 
@@ -55,6 +57,13 @@ JLCPARTS_DATA_BASE = "https://yaqwsx.github.io/jlcparts/data"
 
 # Filename of the reassembled catalog inside the cache directory.
 CATALOG_FILENAME = "jlcparts.sqlite3"
+
+# Zip signatures (little-endian) relevant to split-archive extraction.
+#   PK\x07\x08 -- "spanning" / split-archive marker prepended to the first
+#                 segment of a ``zip -s`` archive.
+#   PK\x03\x04 -- local file header that begins each archive member.
+_SPANNING_MARKER = b"PK\x07\x08"
+_LOCAL_FILE_HEADER = b"PK\x03\x04"
 
 
 def get_catalog_path() -> Path:
@@ -414,15 +423,63 @@ def sync_catalog(
     return dest
 
 
+def _is_split_archive(archive: Path) -> bool:
+    """Return whether ``archive`` is a ``zip -s`` split (multi-disk) archive.
+
+    A split archive produced by ``zip -s`` begins with the 4-byte spanning
+    marker ``PK\\x07\\x08`` on its first segment. After the segments are
+    concatenated in order the reassembled blob therefore starts with that
+    marker. Python's :mod:`zipfile` refuses these categorically
+    (``BadZipFile: zipfiles that span multiple disks are not supported``)
+    because it consults the multi-disk central directory, so we detect the
+    marker and take a streaming extraction path that never reads the central
+    directory.
+    """
+    with archive.open("rb") as fh:
+        head = fh.read(4)
+    return head[:4] == _SPANNING_MARKER
+
+
 def _extract_catalog(archive: Path, dest: Path) -> None:
     """Extract the SQLite database from a (reassembled) jlcparts zip archive.
 
-    Picks the first ``*.sqlite3``/``*.sqlite``/``*.db`` member, or -- if the
-    archive contains a single member -- that member. Writes it to ``dest``.
+    Two paths are supported:
+
+    * **Single-disk archive** (the CI test fixtures): parsed normally with
+      :mod:`zipfile`; the first ``*.sqlite3``/``*.sqlite``/``*.db`` member (or,
+      failing that, a lone single member) is extracted.
+    * **Split ``zip -s`` archive** (the real yaqwsx/jlcparts dataset): detected
+      by the leading ``PK\\x07\\x08`` spanning marker -- or by ``zipfile``
+      raising the multi-disk ``BadZipFile`` -- and streamed via
+      :func:`_extract_split_archive`, which never consults the multi-disk
+      central directory.
+
+    Both paths write to a sibling temp file and promote it atomically via
+    :func:`os.replace`, so a killed or failed extraction can never leave a
+    truncated catalog that later lookups would silently trust (#4117).
 
     Raises:
-        RuntimeError: If no plausible SQLite member is found.
+        RuntimeError: If no plausible SQLite member is found, or the split
+            archive's first member is not deflate-compressed.
     """
+    if _is_split_archive(archive):
+        _extract_split_archive(archive, dest)
+        return
+
+    try:
+        _extract_single_disk_archive(archive, dest)
+    except zipfile.BadZipFile as e:
+        # Some split archives do not carry the leading spanning marker but
+        # still trip the multi-disk guard once zipfile reads the central
+        # directory. Fall back to the streaming path in that case.
+        if "span multiple disks" in str(e):
+            _extract_split_archive(archive, dest)
+            return
+        raise
+
+
+def _extract_single_disk_archive(archive: Path, dest: Path) -> None:
+    """Extract the SQLite member from a normal single-disk zip archive."""
     with zipfile.ZipFile(archive) as zf:
         names = [n for n in zf.namelist() if not n.endswith("/")]
         sqlite_members = [n for n in names if n.lower().endswith((".sqlite3", ".sqlite", ".db"))]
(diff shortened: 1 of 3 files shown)


rjwalters:
✅ **Approved!**

Solid, correct, and well-tested fix for the true `zip -s` split-archive extraction path. Reviewed against all four focus areas.

**Streaming correctness (verified):**
- Local file header offsets are spec-correct: method @ +8, name-len @ +26, extra-len @ +28, data starts at `30 + name_len + extra_len`. The fixed 30-byte read + variable field consumption is right.
- Deflate-only guard (`method != 8` → clear `RuntimeError`) is present and tested (`test_extract_split_archive_non_deflate_errors`).
- `zlib.decompressobj(-15)` streamed to `decompressor.eof` needs no member sizes, so data-descriptor / zip64 entries (which the real dataset uses) work. The `while not eof: ... else: flush()` construct is subtle but correct — `flush()` is called on both the eof-normal-exit path (`else`) and the input-exhaustion path (before `break`).
- Atomicity holds: tmp sibling + `os.replace`, with `finally: tmp_dest.unlink(missing_ok=True)` so a failed extraction leaves no truncated catalog (asserted by `not dest.exists()` in the failure tests).

**Multi-disk fallback discrimination (the critical point):** The fallback matches on the substring `"span multiple disks"`, which is the *exact* wording stdlib zipfile uses (`raise BadZipFile("zipfiles that span multiple disks are not supported")` — verified in the installed CPython source). Other `BadZipFile` causes (truncation, bad magic, bad CRC) carry different messages and correctly re-raise — a corrupt single-disk zip is **not** silently rerouted to streaming to produce garbage. Both routing branches are genuinely covered: `test_spanning_marker_routes_to_streaming_not_zipfile` (marker-detection path, asserts single-disk path is never entered) and `test_extract_falls_back_when_zipfile_reports_multidisk` (forces the multi-disk guard so the fallback branch runs).

**Byte identity:** `test_extract_split_archive_byte_identical` + the sync/CLI tests assert `dest.read_bytes() == catalog_db.read_bytes()` and that the result is a queryable catalog (`count() == 4`, `lookup("C25804")`).

**CLI error surfacing:** No code change needed — the existing `except Exception` in `_sync_catalog` (parts_cmd.py:198) already surfaces the real error to stderr with rc 1; `test_cli_surfaces_extraction_error` locks in both the message text and the no-partial-catalog invariant.

## Test Execution
- `uv run pytest tests/parts/test_jlcparts_catalog.py` — ✅ 33 passed (local, PR branch worktree @ 5702328d)
- `uv run ruff check` (3 changed files) — ✅ clean
- `uv run ruff format --check` (3 changed files) — ✅ already formatted
- mypy — ⏭️ not run locally (not in worktree env); CI **Type Check** is authoritative and passes, confirming the cmaes mypy error is pre-existing-env-only

## CI
All 16 checks pass (Test 15m24s, Type Check, Lint & Format, C++ Build, all Board 00–07 + routing regressions). Merge state: **CLEAN**.

PR description uses "Closes #4123". The honest builder notes (stdlib tolerating a leading `PK\x07\x08` prefix; the separate forced-multidisk test for the fallback branch; CLI needing no change) all check out.

PR #4114, opened 2026-07-13
feat(check): label hole-to-hole findings same-net/different-net in report

rjwalters:
## Summary

`kct check`'s `dimension_drill_clearance` (hole-to-hole) findings already carry
both endpoints' nets (`nets=(net1, net2)`, also serialized in `--format json`
via `DRCViolation.to_dict()`), but the text/table report hid the net
relationship: `Nets:` was only printed under `--verbose`, and even then it was a
raw label list with no same-net/different-net framing. This forced manual
classification of drill-clearance walls during manufacturing sign-off.

This PR enriches the report (formatting only) so a user can immediately separate
the fab's real concern (different-net pairs, where a drill-wall break creates a
short) from lower-risk same-net / floating pairs.

## Changes

- `_print_violation()`: for `dimension_drill_clearance`, tag the header line
  with a `(same-net)` / `(different-net)` qualifier and print
  `Nets: net1 / net2` **unconditionally** (not gated on `--verbose`). Non-drill
  rules are unaffected and keep their verbose-only `Nets:` line.
- `output_table()`: append a same-net/different-net sub-count to the `BY RULE:`
  line, e.g. `dimension_drill_clearance: 49 errors (32 different-net, 17 same-net)`.
- Added `_net_relationship()` helper + `_NET_RELATIONSHIP_RULE_IDS` frozenset.
- Tests: new `TestDrillClearanceNetRelationship` in `tests/test_cli_check.py`
  (helper, default + verbose label/nets, non-drill isolation, BY RULE breakdown);
  new same-net via-pair severity guard in `tests/test_validate_dimensions.py`.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| Default (non-verbose) output shows both nets + same/different-net qualifier | ✅ | `test_different_net_finding_shows_label_and_nets_default`, `test_same_net_finding_shows_label_and_nets_default` |
| `--verbose` keeps/improves the `Nets:` line with qualifier (no dup) | ✅ | `test_verbose_keeps_label_and_nets` asserts `Nets:` printed exactly once |
| Severity unchanged; same-net stays `error`; same-footprint warning not regressed | ✅ | `test_same_net_via_pair_remains_error_and_carries_nets`; existing `test_same_footprint_pads_produce_warning` still passes |
| `BY RULE:` summary breaks count into same-net / different-net | ✅ | `test_by_rule_summary_includes_breakdown` |
[...]

Issue #4102, opened 2026-07-13
dimension_drill_clearance: hole-to-hole findings lack net-relationship awareness (same-net vs different-net)


rjwalters:
## Original Issue

**Version:** kicad-tools 0.14.0. Minor / rule-granularity request.

### Problem

`kct check`'s `dimension_drill_clearance` rule flags hole-to-hole spacings below a flat floor (0.5 mm for the jlcpcb profile) with no distinction between:

- **different-net** hole pairs (the fab's actual concern — drill-wall breakage creating shorts), and
- **same-net** pairs or pairs involving unconnected pins (e.g. floating connector PTH pins next to signal vias), where breakage is harmless or lower-risk.

On our board, 49 findings at 0.113–0.499 mm required manual classification during manufacturing sign-off: 32 unique pairs were different-net, 17 same-net/floating; the adjudication (accepted carried-over risk from an already-fabricated prior rev) had to be built by hand-parsing the board because the rule output doesn't carry the net relationship.

### Suggested improvement

Include both endpoints' nets in each finding, and either split the rule (hole_to_hole_diff_net / hole_to_hole_same_net) or set severity by relationship. That turns an undifferentiated 49-finding wall into a directly-actionable list.

(Related observation, cosmetic: the rule name "dimension_drill_clearance" reads as a min-hole-size check; it took source-of-truth digging to establish it is hole-to-hole spacing. A rename or description clarification would help.)

---

### Verified current state (against `origin/main`)

`src/kicad_tools/validate/rules/dimensions.py`, `_check_drill_clearance` (currently lines 181-276):

- **Net names are already collected and attached.** Both `via` and `thru_hole` pad drills carry a resolved net name, and every `DRCViolation` for this rule is emitted with `nets=(net1, net2)` (line 273). `DRCViolation.to_dict()` (`src/kicad_tools/validate/violations.py`) always serializes `"nets": list(self.nets)`, so **`kct check --format json` already exposes both endpoints' nets per finding** — that part of the original request is done (landed via #2193, #2191).
[...]

PR #4129, opened 2026-07-13
fix: classify floating-pin hole pairs as different-net, not same-net

### LINKED ISSUE #4127: bug: #4114 hole-to-hole net-split labels (26 diff / 23 same) disagree with hand-adjudication (32 diff / 17 same) on identical 49 findings
## Summary

`kct check`'s `dimension_drill_clearance` same-net/different-net classifier
(#4114/#4102) has a real bug, not just a cosmetic labeling question: it
misclassifies **floating-pin ↔ floating-pin** hole pairs as `same-net`
when they are not electrically connected at all. This is the root cause
of the 26/23 vs 32/17 split reported against chorus v23.

## Root cause (confirmed against `origin/main`)

**`_check_drill_clearance`** (`src/kicad_tools/validate/rules/dimensions.py:224`)
builds the per-hole net label with:

```python
net_name = pad.net_name if pad.net_name else f"net:{pad.net_number}"
```

and the analogous line for vias (`dimensions.py:205`):

```python
net_name = net.name if net else f"net:{via.net_number}"
```

**Every genuinely unconnected pad/via in a KiCad PCB has `net_number == 0`**
— this is confirmed in `PCB` net-table construction
(`src/kicad_tools/schema/pcb.py:1945`, `self._nets[0] = Net(0, "")`): net 0
is a single canonical "no net" sentinel, not a per-pad/per-net unique ID.
Consequently `pad.net_name` is `""` for *every* floating pad, and the
fallback above produces the **identical literal string `"net:0"`** for all
of them — not distinct placeholders per net.

**`_net_relationship`** (`src/kicad_tools/cli/check_cmd.py:66`):

```python
def _net_relationship(nets: tuple[str, ...]) -> str | None:
    if len(nets) != 2:
        return None
    return "same-net" if nets[0] == nets[1] else "different-net"
```

does a plain string equality. When both endpoints of a hole-to-hole finding
are floating (net 0), `nets == ("net:0", "net:0")`, so `nets[0] == nets[1]`
is `True` and the pair is labeled `same-net` — even though the two pins
(e.g. chorus J2.11 and J2.19, both unconnected RPi-header pins) are **not**
the same electrical node. Two floating pins are, if anything, "no
relationship" — but they are certainly not "same-net," and a fab
drill-wall break between them is exactly as much of a latent-short risk
category as a genuinely different-net pair (arguably *more* concerning,
since a floating pin is unmonitored — nothing will flag a short to it
downstream).

The existing docstring/comment is factually wrong for the real-world case:

> "Unnamed nets are already resolved to distinct `net:<n>` placeholders
> upstream, so an unconnected pin next to a signal via reads as
> `different-net`."

This is true only when a *named* net is on one side (`"net:0"` vs
`"GND"` correctly differs). It is false when **both** sides are floating —
they collide on the same `"net:0"` string. The current unit test
(`tests/test_cli_check.py::TestDrillClearanceNetRelationship::test_net_relationship_helper`)
only exercises `("net:3", "net:7")` — two *distinct* non-zero numbers —
which does not exist in practice for unconnected pads/vias (they are
always exactly `net:0`, per the schema net-table code above) and never
exercises the actual collision (`("net:0", "net:0")`).

This plausibly reconciles the reported delta: chorus v23 has J2.11/19/23/3
floating (4 pins → up to `C(4,2)=6` floating-floating pairs, plus more
floating-vs-named pairs that already classify correctly). The hand
adjudication counted floating↔anything as `different-net`; the automated
classifier currently under-counts `different-net` by the floating↔floating
subset. 32 − 26 = 6, which lines up with a plausible floating-pair count
on a 4-pin floating cluster — the builder should re-run
`kct check chorus-test-revA_v23_mfg.kicad_pcb --mfr jlcpcb` after the fix
and confirm the new counts land at (or reconcile against) 32/17 as a
validation step, not a hard CI gate (chorus is local-only, not in CI).

## Correct semantics (decide + document)

A hole is **same-net** only when both endpoints resolve to the same
*named, non-floating* net. Any pairing involving a floating/unconnected
pin (net 0) can never be "same-net," full stop — floating pins have no
net identity to share.

Recommended fix: give `_net_relationship` explicit floating-awareness
instead of raw string equality:

```python
def _net_relationship(nets: tuple[str, ...]) -> str | None:
    """Classify a hole-to-hole finding's net pair as same-net / different-net.

    Returns ``"same-net"`` only when both endpoints resolve to the same
    *named* net. Returns ``"different-net"`` when the names differ, or when
    either endpoint is unconnected/floating (KiCad net 0, resolved upstream
    to the ``"net:0"`` placeholder -- see dimensions.py). Floating pins have
    no net identity to share, so a floating/floating or floating/named pair
    can never be "same-net" even though the string may coincidentally match
    another floating pin's placeholder. Returns ``None`` when the finding
    does not carry exactly two net names (nothing to compare).
    """
    if len(nets) != 2:
        return None
    n1, n2 = nets
    if _is_floating_net(n1) or _is_floating_net(n2):
        return "different-net"
    return "same-net" if n1 == n2 else "different-net"


def _is_floating_net(net: str) -> bool:
    """True for KiCad's unconnected/no-net sentinel as resolved by dimensions.py.

    Unconnected pads/vias always carry ``net_number == 0`` (a single
    canonical sentinel -- see ``PCB`` net-table construction), which
    resolves to the empty string or, in dimensions.py's fallback, the
    literal placeholder ``"net:0"``. Treat either spelling as floating so
    this stays correct if the empty-string form ever reaches here directly
    (e.g. a future caller that skips the ``f"net:{number}"`` fallback).
    """
    return net in ("", "net:0")
```

Keep the **two-bucket** `same-net` / `different-net` label (do not add a
third "floating" bucket) so the `BY RULE:` summary line arithmetic in
`output_table` (`diff_n + same_n == errors` for this rule) stays exactly
as it is today — no changes needed there beyond correctness of the
`_net_relationship` calls it already makes. If a curator/builder judgment
call decides a distinct `(floating)` label is more truthful than folding
into `(different-net)`, that is acceptable **only if** `output_table`'s
arithmetic and the `BY RULE:` line are updated in lockstep to show three
counts (e.g. `(28 different-net, 15 same-net, 6 floating)`) — but the
simpler two-bucket fold-into-different-net approach is preferred: it
requires no output-table changes, matches the hand-adjudication precedent
exactly, and floating pins are indisputably not "same-net" so
"different-net" is the truthful bucket for them.

Also fix the sibling bug in the docstring at `dimensions.py:66-77` in
whichever direction is chosen (correct the "distinct `net:<n>` placeholders"
claim to reflect that unconnected pads/vias share a single `net:0`
placeholder, not distinct numbers).

## Scope guard (from #4114 / #2976)

Presentational / classification-label only. Do **not** change severity or
verdict: same-footprint downgrade-to-warning logic in
`_check_drill_clearance` (dimensions.py:251-253) is untouched, and
same-net drill overlap (including floating-pin drill overlap) remains a
manufacturing defect per #2976 — this issue is only about which net-pair
findings get labeled `same-net` vs `different-net` in the report.

## Affected Files

- `src/kicad_tools/cli/check_cmd.py` — `_net_relationship()` (~line 66-77) and its docstring; the `_NET_RELATIONSHIP_RULE_IDS` comment block above it (~line 55-63) also needs its "unconnected pin ... reads as different-net" claim corrected to explain *why* (floating-aware check, not naive string equality).
- `src/kicad_tools/validate/rules/dimensions.py` — no functional change expected, but verify comments near `net_name = pad.net_name if pad.net_name else f"net:{pad.net_number}"` (~line 224) and the via equivalent (~line 205) still accurately describe the placeholder's collision behavior once `_net_relationship` is floating-aware.
- `tests/test_cli_check.py` — `TestDrillClearanceNetRelationship` class (~line 983 onward): extend `test_net_relationship_helper` with the floating/floating and floating/named cases; existing `("net:3", "net:7")` case should be removed or clearly re-labeled as testing two *distinct named* nets, not floating placeholders (it does not represent a real floating-pad scenario).

## Test Plan

1. Unit test in `tests/test_cli_check.py::TestDrillClearanceNetRelationship::test_net_relationship_helper` (or a new test method) asserting the fix:
   - `_net_relationship(("net:0", "net:0")) == "different-net"` (the core regression this issue exists for — two distinct floating pins, both resolved to the `net:0` placeholder, must not read as same-net).
   - `_net_relationship(("net:0", "GND")) == "different-net"` (floating vs named — already correct today, keep it correct).
   - `_net_relationship(("GND", "GND")) == "same-net"` (named/named, unchanged — sanity that the fix doesn't over-broaden "different-net").
   - `_net_relationship(("", "")) == "different-net"` and `_net_relationship(("", "GND")) == "different-net"` if `_is_floating_net` also accepts the raw empty string (defensive case per the helper's docstring).
2. Update/replace the `("net:3", "net:7")` assertion in the existing test to avoid implying floating pads get distinct per-pad numbers — either drop it or repurpose it as a named-nets-that-happen-to-look-like-placeholders case with a comment clarifying it is not a floating-pad scenario.
3. Keep `test_by_rule_summary_includes_breakdown` green — it does not use floating nets, so it should be unaffected; add a sibling case (or extend it) with a floating/floating pair to confirm the `BY RULE:` line's `diff_n + same_n == errors` invariant still holds after the fix (e.g. 1 floating/floating + 1 named-same + 1 named-different -> `(2 different-net, 1 same-net)`).
4. `_print_violation` output test: a floating/floating finding should print `dimension_drill_clearance (different-net)` and `Nets: <no net> / <no net>` (the existing `<no net>` substitution for empty net names at `check_cmd.py:1453` already handles display when `v.nets` entries are `""`; if `dimensions.py` instead emits `"net:0"` literally, confirm what actually prints today and whether `<no net>` substitution needs to also match `"net:0"` for consistent display — note this as a display-consistency detail to resolve, not a hard requirement, since the issue is about classification correctness first).
5. Validation step (not a CI gate — chorus is local-only): re-run `kct check chorus-test-revA_v23_mfg.kicad_pcb --mfr jlcpcb` against chorus v23 and confirm the `dimension_drill_clearance` BY RULE line's different-net count increases by the floating-floating pair count (plausibly reconciling toward the hand-adjudicated 32/17, though exact reconciliation is not required — document whatever the new split is and whether it now plausibly matches the hand count's reasoning).
6. Run `uv run pytest tests/test_cli_check.py -k DrillClearanceNetRelationship -v` and full `uv run pytest tests/test_cli_check.py tests/test_validate_dimensions.py` to confirm no regressions.
7. `uv run ruff check src/kicad_tools/cli/check_cmd.py tests/test_cli_check.py` and `uv run mypy src/kicad_tools/cli/check_cmd.py`.

rjwalters:
## Summary

`kct check`'s `dimension_drill_clearance` net-relationship classifier
(`_net_relationship` in `check_cmd.py`, added in #4102/#4114) misclassified
**floating-pin ↔ floating-pin** hole pairs as `same-net`. Every genuinely
unconnected pad/via resolves to the single canonical `net:0` no-net sentinel
(net 0 is one shared placeholder in the PCB net table — `self._nets[0] =
Net(0, "")` in `schema/pcb.py:1945` — not a per-pad unique ID), so two
distinct floating pins collided on the identical `"net:0"` string and the
naive `nets[0] == nets[1]` comparison read them as same-net. This is
electrically backwards: floating pins share no net identity, and a fab
drill-wall break between two floating pins is at least as much of a latent
short risk as a genuinely different-net pair. This was the root cause of the
chorus v23 26/23 vs hand-adjudicated 32/17 split.

## Changes

- **`check_cmd.py`**: new `_is_floating_net(net)` helper (matches `""` and
  `"net:0"`); `_net_relationship` now short-circuits to `different-net`
  whenever *either* endpoint is floating, before the named-net equality
  check. Two-bucket labels are unchanged so `output_table`'s
  `diff + same == errors` arithmetic is untouched. Corrected the stale
  docstring (which wrongly claimed unconnected pads get "distinct `net:<n>`
  placeholders") and the `_NET_RELATIONSHIP_RULE_IDS` comment block to
  document the `net:0` collision and the floating-aware rule.
- **`dimensions.py`**: comments near the via (line ~205) and pad (line ~224)
  `f"net:{...}"` fallbacks now accurately describe that all floating
  endpoints collide on the same `net:0` placeholder (no functional change).
- **`tests/test_cli_check.py`**: re-labeled the unrealistic `("net:3",
  "net:7")` case as *distinct named* nets (not a floating scenario); added
  `test_net_relationship_floating_pins` covering `("net:0","net:0")`,
  `("net:0","GND")`, `("","")`, `("","GND")` → `different-net` and
  `("GND","GND")` → `same-net`; added a BY RULE summary case proving the
  `diff + same == errors` invariant holds with a floating pair.

| Floating/floating no longer same-net | ✅ | `_net_relationship(("net:0","net:0")) == "different-net"` asserted |
| Floating/named stays different-net | ✅ | `("net:0","GND")`, `("","GND")` asserted |
| Named/named unchanged (no over-broadening) | ✅ | `("GND","GND") == "same-net"` asserted |
| Empty-string sentinel treated as floating | ✅ | `("","")`, `("","GND")` asserted |
| Two-bucket arithmetic (`diff+same==errors`) intact | ✅ | new BY RULE test: 1 floating + 1 named-diff + 1 named-same → `(2 different-net, 1 same-net)` |
| No severity/verdict change (#4114/#2976 guard) | ✅ | only `_net_relationship`/labels touched; downgrade logic in `dimensions.py:251-253` untouched |
| Docstrings/comments corrected | ✅ | check_cmd.py docstring + comment block and dimensions.py fallback comments updated |

## Test Plan

- `uv run pytest tests/test_cli_check.py tests/test_validate_dimensions.py` → **99 passed** (dimensions.py at 100% coverage).
- `uv run ruff check` + `ruff format --check` on all 3 files → clean.
- `uv run --with mypy mypy --ignore-missing-imports` on both changed source files → `Success: no issues found`.

### Notes for reviewer

- **mypy baseline gate**: `scripts/ci/check_mypy_baseline.py` reports 10
  "new" errors here, but they are pre-existing environment artifacts
  (missing optional-dependency stubs: `mcp`, `cmaes`, `matplotlib`,
  `weasyprint`, plus unused-ignore comments) in files I did not touch. I
  confirmed the identical 10 errors appear on pristine `main` with my changes
  stashed. `mypy` is not installed in the project env, so the baseline was
  generated against a different (fuller) environment; a direct run over my
  two files is clean.
- **Chorus reconciliation (reasoning, not a CI gate)**: the fix moves
  floating-involved pairs from `same-net` → `different-net`, the direction
  consistent with the 26/23 → 32/17 hand count (chorus v23's 4-pin floating
  RPi-header cluster yields up to `C(4,2)=6` floating/floating pairs, and
  32 − 26 = 6). The `chorus-test-revA_v23_mfg.kicad_pcb` fixture is
  local-only and dangles in worktrees, so the live re-run could not be
  executed here; documenting the reasoning as validation, not a gate.

Closes #4127

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 4107
-# Branch: feature/issue-4107
+# Issue: 4127
+# Branch: feature/issue-4127
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/src/kicad_tools/cli/check_cmd.py
+++ b/src/kicad_tools/cli/check_cmd.py
@@ -55,25 +55,54 @@ _MEASUREMENT_RULE_IDS: frozenset[str] = frozenset(
 # Issue #4102: ``dimension_drill_clearance`` hole-to-hole findings carry both
 # endpoints' resolved net names in ``nets=(net1, net2)``.  A user reading the
 # report needs to separate the fab's real concern (different-net pairs, where a
-# drill-wall break creates a short) from same-net / floating pairs.  The data is
+# drill-wall break creates a short) from same-net pairs.  The data is
 # already present per-finding and in ``--format json``; the report just needs to
 # surface it (net names shown unconditionally, plus an explicit same-net /
 # different-net qualifier).  This is presentational only -- severity is
 # unchanged (see #2976: same-net drill overlap is still a manufacturing defect).
+#
+# Issue #4127: the classifier uses a floating-aware check (``_is_floating_net``),
+# not naive string equality.  Every genuinely unconnected pad/via resolves to
+# the *same* ``net:0`` placeholder (net 0 is a single canonical no-net sentinel
+# in the PCB net table -- see ``PCB`` construction in schema/pcb.py), so a naive
+# ``nets[0] == nets[1]`` would mislabel two *distinct* floating pins as
+# ``same-net``.  Floating pins share no electrical identity, so any pair
+# involving a floating endpoint is ``different-net``.
 _NET_RELATIONSHIP_RULE_IDS: frozenset[str] = frozenset({"dimension_drill_clearance"})
+
+
+def _is_floating_net(net: str) -> bool:
+    """True for KiCad's unconnected/no-net sentinel as resolved by dimensions.py.
+
+    Unconnected pads/vias always carry ``net_number == 0`` (a single canonical
+    sentinel -- see ``PCB`` net-table construction, ``self._nets[0] = Net(0,
+    "")``), which resolves to the empty string or, in dimensions.py's fallback,
+    the literal placeholder ``"net:0"``.  Treat either spelling as floating so
+    this stays correct if the empty-string form ever reaches here directly
+    (e.g. a future caller that skips the ``f"net:{number}"`` fallback).
+    """
+    return net in ("", "net:0")
 
 
 def _net_relationship(nets: tuple[str, ...]) -> str | None:
     """Classify a hole-to-hole finding's net pair as same-net / different-net.
 
-    Returns ``"same-net"`` when both endpoints resolve to the same net name,
-    ``"different-net"`` when they differ, and ``None`` when the finding does
-    not carry exactly two net names (nothing to compare).  Unnamed nets are
-    already resolved to distinct ``net:<n>`` placeholders upstream, so an
-    unconnected pin next to a signal via reads as ``different-net``.
+    Returns ``"same-net"`` when both endpoints resolve to the same *named,
+    non-floating* net, ``"different-net"`` when they differ, and ``None`` when
+    the finding does not carry exactly two net names (nothing to compare).
+
+    Floating/unconnected pins have no net identity to share: every one of them
+    resolves to the *same* ``net:0`` placeholder upstream (net 0 is a single
+    canonical no-net sentinel, not a per-pad unique ID -- see schema/pcb.py).
+    So any pair with a floating endpoint -- including two distinct floating pins
+    that collide on the ``net:0`` string -- is ``different-net``, never
+    ``same-net`` (issue #4127).  A plain string equality would wrongly report
+    ``net:0 == net:0`` as same-net; the floating-aware check below prevents that.
     """
     if len(nets) != 2:
         return None
+    if _is_floating_net(nets[0]) or _is_floating_net(nets[1]):
+        return "different-net"
     return "same-net" if nets[0] == nets[1] else "different-net"
 
 
--- a/src/kicad_tools/validate/rules/dimensions.py
+++ b/src/kicad_tools/validate/rules/dimensions.py
@@ -202,6 +202,12 @@ class DimensionRules(DRCRule):
         # Add vias
         for via in pcb.vias:
             net = pcb.get_net(via.net_number)
+            # Unconnected vias carry net_number == 0, a single canonical no-net
+            # sentinel (schema/pcb.py: self._nets[0] = Net(0, "")), so this
+            # fallback yields the identical "net:0" string for *every* floating
+            # via -- distinct floating endpoints collide on it. The
+            # net-relationship classifier (check_cmd._net_relationship, #4127)
+            # is floating-aware and treats such collisions as different-net.
             net_name = net.name if net else f"net:{via.net_number}"
             drills.append((via.position, via.drill, net_name, "", net_name))
 
@@ -221,6 +227,12 @@ class DimensionRules(DRCRule):
                     )
                     abs_x = fp.position[0] + rotated_x
                     abs_y = fp.position[1] + rotated_y
+                    # As with vias above, every floating pad has net_number == 0
+                    # and net_name == "", so this fallback produces the same
+                    # "net:0" placeholder for all of them (they collide, not
+                    # distinct per-pad IDs). check_cmd._net_relationship is
+                    # floating-aware and classifies any such pair as
+                    # different-net (#4127).
                     net_name = pad.net_name if pad.net_name else f"net:{pad.net_number}"
                     drills.append(
                         (
--- a/tests/test_cli_check.py
+++ b/tests/test_cli_check.py
@@ -1011,11 +1011,36 @@ class TestDrillClearanceNetRelationship:
 
         assert _net_relationship(("HALL_A", "GND")) == "different-net"
         assert _net_relationship(("HALL_A", "HALL_A")) == "same-net"
-        # Unnamed nets resolve to distinct net:<n> placeholders upstream.
+        # Two *distinct named* nets that happen to look like placeholders --
+        # NOT a floating-pad scenario (floating pads never get distinct
+        # per-pad numbers; they all resolve to the single "net:0" sentinel).
         assert _net_relationship(("net:3", "net:7")) == "different-net"
         # Not a pair -> nothing to compare.
         assert _net_relationship(()) is None
         assert _net_relationship(("only-one",)) is None
+
+    def test_net_relationship_floating_pins(self):
+        """Issue #4127: floating/unconnected pins never read as same-net.
+
+        Every genuinely unconnected pad/via resolves to the *same* ``net:0``
+        placeholder (net 0 is a single canonical no-net sentinel), so a naive
+        string equality would mislabel two distinct floating pins as same-net.
+        Any pair involving a floating endpoint must classify as different-net.
+        """
+        from kicad_tools.cli.check_cmd import _net_relationship
+
+        # Core regression: two distinct floating pins, both resolved to net:0,
+        # must NOT read as same-net.
+        assert _net_relationship(("net:0", "net:0")) == "different-net"
+        # Floating vs named (already correct today; keep it correct).
+        assert _net_relationship(("net:0", "GND")) == "different-net"
+        assert _net_relationship(("GND", "net:0")) == "different-net"
+        # Empty-string spelling of the no-net sentinel is also floating.
+        assert _net_relationship(("", "")) == "different-net"
+        assert _net_relationship(("", "GND")) == "different-net"
+        assert _net_relationship(("GND", "")) == "different-net"
+        # Named/named is unchanged -- the fix does not over-broaden.
+        assert _net_relationship(("GND", "GND")) == "same-net"
 
     def test_different_net_finding_shows_label_and_nets_default(self, capsys):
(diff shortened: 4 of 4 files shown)


rjwalters:
✅ **Approved!**

Verified the display-only floating-net fix against all four review points.

**1. Semantics / helper exactness.** `_is_floating_net` matches exactly `("", "net:0")` and nothing else. Confirmed this is complete coverage of what `dimensions.py` can emit:
- Primary path yields `net.name`, which for net 0 is `""` (`schema/pcb.py:1945` — `self._nets[0] = Net(0, "")`, confirmed).
- The two fallbacks emit `f"net:{via.net_number}"` / `f"net:{pad.net_number}"`; for a floating endpoint `net_number == 0`, so the string is exactly `"net:0"` (int-formatted, no whitespace, no leading zeros — `"net:00"` is unreachable). `net_name` is always `str`, so `None` cannot reach the helper.
- `_net_relationship` guards `len(nets) != 2 → None` before the floating check, and `_print_violation` guards `v.nets` truthiness at both print sites (lines 1452, 1466), so the absent/short-tuple path is safe.

**2. dimensions.py is genuinely comment-only.** The diff inserts only `#` comment blocks before the via (`net_name = net.name if net else f"net:{via.net_number}"`) and pad (`net_name = pad.net_name if pad.net_name else f"net:{pad.net_number}"`) fallbacks; both assignment lines are unchanged. No functional change.

**3. Summary arithmetic holds by construction.** `_net_relationship` still returns exactly one of `same-net`/`different-net` for any 2-tuple (never `None`), so every finding lands in exactly one bucket — floating pairs simply move same→different. `diff + same == errors` is preserved, and the new `test_by_rule_summary_floating_pair_counts_as_different` asserts `(2 different-net, 1 same-net)` for 1 floating + 1 named-diff + 1 named-same.

**4. CI — all 16 checks green, merge state CLEAN:**
- Type Check: pass (46s) — authoritative baseline gate, green
- Lint & Format: pass · Test: pass (12m15s) · C++ Build: pass
- Boards 00–07 + Routed DRC + kicad-cli smoke + Diff-Pair/Match-Group regressions: all pass

**Local scoped run:** `uv run pytest tests/test_cli_check.py tests/test_validate_dimensions.py` → **99 passed** (matches builder claim).

**Scope guard:** only `_net_relationship`/`_is_floating_net`/labels/comments touched; the `dimensions.py` severity/downgrade logic is untouched (#4114/#2976 respected).

PR description uses `Closes #4127`. The floating→different-net rule is the electrically-honest classification (two floating pins share no net identity; a drill-wall break between them is a latent-short risk, not a benign same-net overlap). Chorus live re-run is fixture-local-only and correctly documented as reasoning rather than a gate; the `C(4,2)=6 = 32−26` reconciliation is sound.

PR #4130, opened 2026-07-13
fix(lcsc): reject cross-type LCSC value mismatches (resistor vs capacitor)

rjwalters:
## Summary

`find_value_mismatch` in `export/lcsc_value_check.py` was component-type-aware for *parsing* but not for *cross-type checking*. A resistor-typed request (e.g. `330R`) validated against a capacitor description found zero ohm-pattern matches and treated that as "cannot validate, accept" instead of "different component type, reject". This let `bom_enrich`'s `_try_cache_fallback` replay a stale cache row assigning `330R → C1525` (a 100nF capacitor) — the exact chorus v23 recurrence, surviving the `#3590`/`#3597` validation layer that C1525 previously defeated.

## Changes

- Add `_description_dominant_type(description, request_type)`: returns a *foreign* numeric component type when the description clearly parses as that type's primary value AND carries none of the request's own unit. Deliberately conservative — if the request's own unit pattern is present, or if two-or-more foreign unit patterns match, it returns `None` (stay permissive) to avoid false rejects on RC networks / ferrite beads / mixed-unit specs.
- In `find_value_mismatch`, when no same-type candidate value is found, consult `_description_dominant_type` before accepting; a dominating foreign type yields a `ValueMismatch` (cross-type). `candidate_si=NaN` marks the cross-type case and `describe()` appends `(different component type)`.
- Resolve the requested component *type* independently of the numeric parse, so `330R` (parses as `RESISTOR` with `numeric_value=None` under the canonical parser) reaches the cross-type check instead of bailing at the top. Same-type comparisons with an unparseable request value stay permissive (accept).
- Removed the now-unused `_parse_requested` helper (its logic is inlined for the type-vs-numeric split).

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| `330R` vs `100nF ... Capacitor` → rejected | ✅ | `test_resistor_request_capacitor_description_rejected` (literal C1525 description) + `test_resistor_request_synthetic_capacitor_rejected` |
| Existing `#3590`/`#3597` tests unchanged & green | ✅ | `tests/test_lcsc_value_guard.py` + `tests/test_lcsc_package_guard.py` — 80 passed, no existing test modified |
| No-units-anywhere stays permissive | ✅ | `test_no_units_anywhere_stays_permissive` |
[...]

PR #4131, opened 2026-07-13
feat(parts): add offline-catalog fallback to LCSCClient.search()

### LINKED ISSUE #4126: bug: LCSCClient.search (BOM --auto-lcsc / parts suggest) has no offline-catalog fallback — 403 defeats offline jlcparts catalog (#4108/#4117)
## Summary

The offline jlcparts catalog shipped in v0.15.0 (#4108/#4117) is only consulted by the **LCSC-number lookup** path (`LCSCClient.lookup` / `lookup_many`). The **parametric matcher** -- `LCSCClient.search(query, ...)`, which is what `kct export --auto-lcsc` and `kct parts suggest` use to resolve a component's value+footprint to a candidate LCSC part -- calls the live JLCPCB web API with **no offline fallback**. When the API returns the documented 403 Forbidden (geo-block/anti-bot), BOM enrichment collapses to near-empty even with a fully synced 5.5 GB `jlcparts.sqlite3` catalog present.

Found while validating v0.15.0 against the certified chorus v23 board (project-shamrock/chorus#26).

## Repro

```bash
# offline catalog synced and functional:
kct parts sync-catalog            # -> catalog present at ~/.cache/kicad-tools/jlcparts.sqlite3
kct parts lookup C25104           # -> resolves 330R (OFFLINE, works)

# but the matcher path is live-API-only:
kct parts suggest chorus-test-revA.kicad_sch
#   Analyzing 89 components (89 missing LCSC numbers)...
#   Error: LCSCForbiddenError: JLCPCB API returned 403 Forbidden

kct export chorus-test-revA_v23_mfg.kicad_pcb --output /tmp/out --mfr jlcpcb
#   LCSC enrichment: 1 auto-matched, 3 from cache, 0 from schematic, 23 unmatched
#   (every unmatched reason: "JLCPCB API unavailable (403)")
#   88/88 active item(s) missing LCSC part number
```

On the chorus v23 BOM (27 lines), v0.15.0 offline resolved only 4 lines and produced **one wrong assignment** (`330R -> C1525`, where C1525 is a 100nF capacitor). The certified 0.14.0 bundle had 25/27 lines correctly resolved (via live API when it was reachable).

## Root cause (verified against origin/main @ 6fa8a82, post-#4125)

- `src/kicad_tools/parts/lcsc.py::LCSCClient.search()` (line ~638) calls `self._make_request(SEARCH_URL, ...)` (live web API only); on `requests.RequestException` it returns an empty `SearchResult`, and the `_api_forbidden` circuit breaker in `_make_request` (line ~404) raises `LCSCForbiddenError` on a 403. **`search()` never calls `self._get_catalog()`.**
- Compare to `lookup()` (line ~482) and `lookup_many()` (line ~709), which already have the fallback pattern to mirror: call `self._get_catalog()` once up front, try the live API, and on failure/miss call `catalog.lookup(...)` / `catalog.lookup_many(...)` before giving up. `search()` needs the equivalent third tier.
- `src/kicad_tools/parts/jlcparts_catalog.py::JlcpartsCatalog` (the offline reader) currently exposes **only** `lookup(lcsc_part)` / `lookup_many(lcsc_parts)` -- both keyed by exact LCSC id. **There is no parametric search-by-value/package method**, so even if `search()` wanted to fall back there is nothing to fall back to yet. This is documented as an explicit scope cut in the module docstring: *"Only exact `lcsc` lookup is supported in this phase; full-catalog fuzzy matching is explicitly out of scope"* (this issue is that follow-up phase).

### Catalog schema (what a parametric query can key on)

`jlc_components` columns (per `tests/parts/fixtures/build_jlcparts_fixture.py`, which mirrors the real `yaqwsx/jlcparts` schema used by `_row_to_part`):

```
lcsc INTEGER PRIMARY KEY, mfr TEXT, package TEXT, manufacturer TEXT,
library_type TEXT, description TEXT, datasheet TEXT, stock INTEGER, price TEXT
```

There is **no dedicated value column** (no `resistance`/`capacitance` field) -- value must be matched against free-text `description` (e.g. `"100nF 50V X7R 0402 Multilayer Ceramic Capacitor MLCC"`), same as the live API's `description` field that `PartMatcher._calculate_confidence` (`src/kicad_tools/cost/suggest.py`) already scores against. `package` is a real column (e.g. `"0402"`, `"LQFP-48"`, `"SOT-23"`) and can be filtered directly. `library_type` (`basic`/`preferred`/`extended`) is the JLCPCB assembly-tier field.

Concretely, a `JlcpartsCatalog.search(value_terms, package=None, ...)` method should:

1. Run a single SQL query filtering on `description LIKE '%term%'` for each parsed search term (AND-combined, case-insensitive via `COLLATE NOCASE` or `LOWER()`), optionally `AND package = ?` when a package is known, with a `stock > 0` (or a `min_stock` param) filter pushed into SQL rather than post-filtered in Python (the real catalog is ~600k+ rows; avoid pulling unfiltered result sets into Python).
2. Rank/order results the same way the live-API path already does downstream in `PartMatcher.suggest_for_component` (`src/kicad_tools/cost/suggest.py::sort_key`, line ~549): `library_type = 'basic'` first, then `preferred`, then `extended` (`ORDER BY CASE library_type WHEN 'basic' THEN 0 WHEN 'preferred' THEN 1 ELSE 2 END, stock DESC` is a reasonable SQL-level approximation), then by stock descending as the tiebreaker. Exact numeric-value/confidence scoring can stay in `PartMatcher._calculate_confidence` unchanged -- the catalog method only needs to return a *candidate pool* shaped like what `search()`'s live-API path already returns (translated via the existing `_row_to_part` helper into `Part` objects, then wrapped in `SearchResult`), not do the ranking itself. Cap results (e.g. `LIMIT page_size * 5` before Python-side confidence ranking, matching the existing `page_size` cap behavior of the live API).
3. Reuse the existing `_row_to_part` translation (already in `jlcparts_catalog.py`) -- do not duplicate the column-to-`Part` mapping.

### `LCSCClient.search()` fallback wiring

Mirror the `lookup()` pattern: call `self._get_catalog()`, attempt the live API inside a `try/except`, and on `LCSCForbiddenError` / other `RequestException` (instead of immediately returning an empty `SearchResult`), fall through to `catalog.search(...)` when `catalog is not None and catalog.available`. Preserve the existing behavior when `use_local_catalog=False` or no catalog is synced (silent no-op, byte-for-byte unchanged live-API-only behavior) -- this is the same non-negotiable stated in `lookup()`'s docstring and must hold for `search()` too.

**Important:** `search()` currently re-raises `LCSCForbiddenError` up through `_make_request`'s circuit breaker *before* the caller (`PartMatcher.suggest_for_component`, `cost/suggest.py` line ~518) ever sees a result -- the fallback must be wired at the `LCSCClient.search()` level (catching the exception raised by `_make_request` inside `search()` itself), not by having `cost/suggest.py` catch it, so both `kct parts suggest` and `kct export --auto-lcsc` (which both route through `PartMatcher` -> `client.search()`) benefit automatically without call-site changes.

## Separate sub-bug: `330R -> C1525` wrong assignment (NOT in scope for this issue)

The repro's wrong assignment did **not** come through `search()` (which has no offline path at all right now -- it just raises/returns empty). Traced it to the **enrichment cache fallback**, a pre-existing and different code path:

- `src/kicad_tools/export/bom_enrich.py::_try_cache_fallback` (line ~177) is what runs when `api_forbidden` is set (the exact 403 scenario in this repro). It reads a previously-cached `(value, footprint) -> lcsc_part` mapping from `PartsCache.get_enrichment_match(..., ignore_expiry=True)` (`src/kicad_tools/parts/cache.py` line ~369) -- **not** from `JlcpartsCatalog` and not from `search()`.
- That cache entry is then validated by `check_lcsc_against_cache()` (`src/kicad_tools/export/lcsc_value_check.py` line ~498), which looks the LCSC part up in the *response* cache (`PartsCache.get(lcsc_part)`, keyed by LCSC id, populated by prior live lookups) and compares its known value/description against the requested BOM value via `find_value_mismatch`. **C1525 is a chronic repeat offender here**: it was already the subject of #3590 (100nF cached against a 16nF row) and #3597 (right value, wrong package -- 0402 part on 0805 pads), both closed by adding this exact validation layer.
- Root cause of *this* recurrence: `find_value_mismatch` (`lcsc_value_check.py` line ~429) is **component-type-aware but not component-type-*cross*-checking**. It parses the requested value using the reference designator's inferred type (`330R` on a resistor ref -> `ComponentType.RESISTOR`), then tries to extract a *same-type* value from the candidate's description using `_DESC_PATTERNS[component_type]` -- i.e. it searches C1525's capacitor description (`"...100nF...MLCC"`) for an **ohms pattern**, finds none, and returns `candidate_si = None`, which `find_value_mismatch` treats as "cannot validate, accept" (line 484-485: `if candidate_si is None: return None`). A resistor-vs-capacitor type mismatch is *more* obviously wrong than the magnitude mismatches #3590/#3597 fixed, but the current guard has no "candidate matched zero patterns for any numeric type it recognizes" / "requested type X but candidate description clearly describes type Y" check -- it silently degrades to "unknown, accept" instead of flagging a cross-type disagreement.
- Additionally, `check_lcsc_against_cache`'s `PartsCache.get(lcsc_part, ignore_expiry=True)` (`cache.py` line ~250) can return `None` if C1525 was never independently response-cached on this machine (a stale `enrichment_matches` row can exist without a corresponding `parts` row), in which case validation is skipped entirely (line 528-529: `if part is None: return None`).

**Verdict: this is a separate, pre-existing bug in the value-validation guard (`lcsc_value_check.py`), not something this issue's Phase 1 work touches.** This issue only adds a new fallback tier to `LCSCClient.search()`; it does not change `_try_cache_fallback` or `check_lcsc_against_cache`. Recommend filing a follow-up issue: *"`find_value_mismatch` accepts cross-component-type LCSC assignments (330R resistor matched to C1525, a 100nF capacitor) because a candidate description that yields zero matches for the requested type's own unit pattern is treated as 'cannot validate' instead of a type mismatch."* That follow-up should also consider whether `check_lcsc_against_cache`'s silent skip-on-`None` (no response-cache entry) needs a `JlcpartsCatalog.lookup()` fallback of its own now that the offline catalog can resolve parametric data for exactly this validation purpose.

## Scope: Phase 1 (this issue)

1. **`JlcpartsCatalog.search()`** (`src/kicad_tools/parts/jlcparts_catalog.py`): new method, SQL `description LIKE` (+ optional `package =`) query over `jlc_components`, ranked `basic > preferred > extended` then `stock DESC`, translated via the existing `_row_to_part`. Silent empty-list return when `not self.available` (same silent-no-op convention as `lookup`/`lookup_many`).
2. **`LCSCClient.search()` fallback wiring** (`src/kicad_tools/parts/lcsc.py`): on live-API failure (`LCSCForbiddenError` or other `RequestException`, mirroring the `try/except` already in `lookup()`), fall through to `self._get_catalog().search(...)` when available; return `SearchResult` shaped identically to the live-API path. `use_local_catalog=False` / catalog absent preserves current (unchanged) behavior.
3. **Do not** touch `bom_enrich.py`, `lcsc_value_check.py`, or `cache.py` (the wrong-assignment sub-bug) in this issue.

## Affected Files

- `src/kicad_tools/parts/jlcparts_catalog.py` (add `JlcpartsCatalog.search()`)
- `src/kicad_tools/parts/lcsc.py` (wire `LCSCClient.search()` fallback to the catalog, mirroring `lookup()`/`lookup_many()`)
- `tests/parts/test_jlcparts_catalog.py` (new tests for `JlcpartsCatalog.search()` and `LCSCClient.search()` fallback, using the existing `catalog_db` fixture from `tests/parts/fixtures/build_jlcparts_fixture.py`)

## Test Plan

Using the existing `catalog_db` fixture (`tests/parts/fixtures/build_jlcparts_fixture.py::SAMPLE_ROWS`, no network, mirrors the pattern of `test_lookup_falls_back_to_catalog_on_api_failure` in `tests/parts/test_jlcparts_catalog.py`):

- `JlcpartsCatalog.search()` unit tests:
  - Value+package query (`"10k"`/`"0402"`-style terms) against the fixture resolves the `C25804` (10kOhms 0402 resistor) row.
  - Query with no `package` filter still ranks correctly by `library_type`/stock.
  - Query against an absent catalog file returns an empty result (silent no-op), not an exception.
  - Basic-preferred-extended ordering: fixture has both `basic` (C25804, C1525) and `preferred` (C100) and `extended` (C8734) rows -- assert ordering in results for a query term that matches more than one library_type.
- `LCSCClient.search()` fallback tests (mock `_make_request` to raise `LCSCForbiddenError` / `requests.RequestException`, mirroring `test_lookup_falls_back_to_catalog_on_api_failure`):
  - Live API raises `LCSCForbiddenError`, catalog present and available -> `search()` returns non-empty `SearchResult` sourced from the catalog instead of propagating the exception.
  - Live API raises, `use_local_catalog=False` -> exception still propagates (unchanged behavior).
  - Live API raises, catalog file absent -> exception still propagates (unchanged behavior, `catalog.available` is `False`).
  - Live API succeeds -> catalog is never consulted (byte-for-byte live-API-only result, same "bypass" pattern as `test_live_api_success_bypasses_catalog`).
- End-to-end shape check tied to the chorus repro: construct a `PartMatcher` (or call `LCSCClient.search()` directly) against the fixture catalog with `use_local_catalog=True`, force a `LCSCForbiddenError` from the live tier, and assert a `"10k"` + `"0402"` query resolves a candidate (value+footprint queries resolve from the fixture instead of raising `LCSCForbiddenError` all the way to the caller, matching the acceptance shape of the `kct parts suggest` repro in this issue).

## Acceptance Criteria

- `kct parts suggest` / `kct export --auto-lcsc` with a synced offline catalog and a 403'd live API resolve candidates from the catalog instead of raising `LCSCForbiddenError` / reporting "JLCPCB API unavailable (403)" for every group.
- `use_local_catalog=False` and no-catalog-synced behavior is unchanged (still raises/returns empty on API failure).
- New tests pass with no network access (fixture-only, matching existing `test_jlcparts_catalog.py` conventions).
- `bom_enrich.py`, `lcsc_value_check.py`, and `cache.py` are untouched by this issue's diff (the wrong-assignment sub-bug is explicitly out of scope; a follow-up issue should be filed for it).

rjwalters:
## Summary

The offline jlcparts catalog (v0.15.0, #4108/#4117) was only consulted by the exact-LCSC lookup path (`LCSCClient.lookup` / `lookup_many`). The parametric matcher `LCSCClient.search()` — which `kct parts suggest` and `kct export --auto-lcsc` use (via `PartMatcher`) to resolve a component's value+package to a candidate — was live-API-only. A documented 403 from the JLCPCB web API therefore collapsed BOM enrichment to near-empty even with a fully synced 5.5 GB catalog present.

This adds the missing third fallback tier to `search()`, mirroring the pattern `lookup()`/`lookup_many()` already have.

## Changes

- **`JlcpartsCatalog.search()`** (`parts/jlcparts_catalog.py`): new parametric method. Runs a single SQL query over `jlc_components` matching whitespace-split query terms against the free-text `description` column (AND-combined, case-insensitive via `COLLATE NOCASE`, LIKE-wildcards escaped via `ESCAPE '\'`), with an optional exact `package =` filter and a `min_stock` floor — all pushed into SQL (the real catalog is ~600k+ rows). Ranks `basic > preferred > extended` (library_type) then `stock DESC`, capped by `limit`. Reuses the existing `_row_to_part` translation. Silent empty return when the catalog file is absent (same no-op convention as `lookup`).
- **`LCSCClient.search()`** (`parts/lcsc.py`): on `LCSCForbiddenError` (403 circuit breaker) or other `RequestException`, falls through to `catalog.search(...)` when a catalog is available, instead of raising / returning empty. Adds an optional `package` arg forwarded to the fallback. Removed the `@_requires_requests` decorator in favor of an internal no-requests→catalog path so offline search works without the `parts` extra installed. `use_local_catalog=False` / no-catalog-synced preserves the historical live-API-only behavior byte-for-byte.
- **Tests** (`tests/parts/test_jlcparts_catalog.py`): parametric-search unit tests (value+package, ranking, min_stock, limit, empty-query safety, wildcard escaping, absent-catalog no-op) and `LCSCClient.search()` fallback tests (403→catalog, RequestException→catalog, 403-without-catalog still raises, use_local_catalog=False still raises, live-success bypasses catalog, no-requests paths, package filter). Fixture-only, no network.
- Two pre-existing search tests in `tests/test_parts.py` asserted the old unconditional raise/empty contract; they now pin `use_local_catalog=False` so they test the no-catalog path deterministically regardless of a machine-local synced catalog.

| `search()` with synced catalog + 403'd API resolves candidates instead of raising | ✅ | `test_search_falls_back_to_catalog_on_403`, `test_search_falls_back_on_request_exception`, `test_search_fallback_respects_package_filter` |
| `use_local_catalog=False` / no-catalog behavior unchanged (raises/empty) | ✅ | `test_search_403_without_catalog_still_raises`, `test_search_403_with_catalog_disabled_still_raises`, `test_search_request_exception_without_catalog_returns_empty` |
| Live API success never consults catalog | ✅ | `test_search_live_success_bypasses_catalog` |
| `JlcpartsCatalog.search()` ranks basic > preferred > extended, then stock DESC | ✅ | `test_catalog_search_library_type_ordering_multi`, `test_catalog_search_ranks_basic_before_extended` |
| New tests pass with no network access | ✅ | Fixture-only via `catalog_db` / `build_jlcparts_fixture` |
| `bom_enrich.py`, `lcsc_value_check.py`, `cache.py` untouched | ✅ | `git diff --stat` shows only `jlcparts_catalog.py`, `lcsc.py`, and the two test files |

## Test Plan

- `uv run pytest tests/parts/test_jlcparts_catalog.py` — 50 passed
- `uv run pytest tests/parts/ tests/test_parts.py tests/cost/test_suggest.py` — 201 passed, 3 failed. The 3 failures (`test_lookup_not_found`, `test_lookup_api_error`, `test_part_number_normalization`) are **pre-existing on `main`** on this machine: they exercise the existing `lookup()` catalog fallback and fail only because a real 5.5 GB catalog is synced locally (verified by re-running on the unmodified base commit — same 3 fail). They concern the `lookup` path, not this change, and pass in CI where no catalog file exists.
- `uv run ruff check` / `ruff format --check` on changed files — clean
- `uv run python scripts/ci/check_mypy_baseline.py` — no new type errors (within baseline)

Closes #4126

### DIFF
--- a/.loom-managed
+++ b/.loom-managed
@@ -1,6 +1,6 @@
 # Loom-managed worktree marker
 # Created by .loom/scripts/worktree.sh
-# Issue: 4128
-# Branch: feature/issue-4128
+# Issue: 4126
+# Branch: feature/issue-4126
 # Removing this file makes Loom treat the worktree as user-owned and refuse
 # to clean it up automatically.
--- a/src/kicad_tools/parts/jlcparts_catalog.py
+++ b/src/kicad_tools/parts/jlcparts_catalog.py
@@ -28,8 +28,12 @@ Design notes
 * No ``PartProvider`` protocol / multi-backend abstraction is introduced
   (deferred per owner direction on #4108) -- the catalog is consulted as a
   fallback inside :class:`~kicad_tools.parts.lcsc.LCSCClient`.
-* Only exact ``lcsc`` lookup is supported in this phase; full-catalog fuzzy
-  matching is explicitly out of scope.
+* Both exact ``lcsc`` lookup (:meth:`JlcpartsCatalog.lookup`) and a
+  parametric ``description``/``package`` search
+  (:meth:`JlcpartsCatalog.search`) are supported. The parametric search backs
+  ``LCSCClient.search``'s offline fallback so parts-matching / BOM enrichment
+  survives a 403'd live API. Confidence scoring stays in the caller
+  (``PartMatcher``); the catalog only returns a candidate pool.
 * The dataset is *never* committed to the repository; ``sync_catalog`` writes
   only into the cache directory.
 """
@@ -184,6 +188,85 @@ class JlcpartsCatalog:
 
         return result
 
+    def search(
+        self,
+        query: str,
+        package: str | None = None,
+        *,
+        min_stock: int = 0,
+        limit: int = 100,
+    ) -> list[Part]:
+        """Parametrically search the catalog by free-text value + package.
+
+        The jlcparts schema has *no* dedicated value column, so the query terms
+        are matched against the free-text ``description`` column (the same field
+        the live-API path scores against). Each whitespace-separated term must
+        appear in ``description`` (AND-combined, case-insensitive), optionally
+        constrained to a known ``package``. Filtering is pushed into SQL rather
+        than post-filtered in Python because the real catalog is ~600k+ rows.
+
+        Results are ordered ``library_type`` basic > preferred > extended (the
+        JLCPCB assembly tier, matching the downstream
+        ``PartMatcher.suggest_for_component`` ranking) then by ``stock``
+        descending, and capped at ``limit``. Exact numeric/confidence scoring is
+        left to the caller (``PartMatcher._calculate_confidence``); this method
+        only returns a *candidate pool* shaped like the live-API path.
+
+        Args:
+            query: Free-text search string (e.g. ``"10k 0402"``). Split on
+                whitespace into AND-combined ``description LIKE`` terms. An
+                empty/blank query yields an empty list (no unbounded scan).
+            package: Optional exact package filter (e.g. ``"0402"``,
+                ``"LQFP-48"``), matched case-insensitively against ``package``.
+            min_stock: Minimum ``stock`` to include (pushed into SQL). Default
+                ``0`` includes zero-stock rows.
+            limit: Maximum number of rows to return.
+
+        Returns:
+            A list of translated :class:`Part` objects (possibly empty). An
+            empty list is returned when the catalog file does not exist (same
+            silent-no-op convention as :meth:`lookup`/:meth:`lookup_many`).
+        """
+        if not self.available:
+            return []
+
+        terms = [t for t in query.split() if t]
+        if not terms:
+            return []
+
+        conditions = ["description LIKE ? ESCAPE '\\' COLLATE NOCASE"] * len(terms)
+        params: list[object] = [f"%{_escape_like(t)}%" for t in terms]
+
+        if package:
+            conditions.append("package = ? COLLATE NOCASE")
+            params.append(package)
+
+        if min_stock > 0:
+            conditions.append("stock >= ?")
+            params.append(min_stock)
+
+        where_clause = " AND ".join(conditions)
+        sql = (
+            "SELECT * FROM jlc_components "
+            f"WHERE {where_clause} "
+            "ORDER BY CASE library_type "
+            "WHEN 'basic' THEN 0 WHEN 'preferred' THEN 1 ELSE 2 END, "
+            "stock DESC "
+            "LIMIT ?"
+        )
+        params.append(int(limit))
+
+        try:
+            with sqlite3.connect(f"file:{self.db_path}?mode=ro", uri=True) as conn:
+                conn.row_factory = sqlite3.Row
+                cursor = conn.execute(sql, params)
+                rows = cursor.fetchall()
+        except sqlite3.Error as e:
+            logger.warning(f"jlcparts catalog search failed for {query!r}: {e}")
+            return []
+
+        return [_row_to_part(row) for row in rows]
+
     def count(self) -> int:
         """Return the number of components in the catalog (0 if absent)."""
         if not self.available:
@@ -201,6 +284,17 @@ def _canonical_lcsc(lcsc_part: str) -> str:
     if not normalized.startswith("C"):
         normalized = f"C{normalized}"
     return normalized
+
+
+def _escape_like(term: str) -> str:
+    """Escape SQL ``LIKE`` wildcards in a user-supplied search term.
+
+    The search terms come from parsed component values (e.g. ``"10k"``,
+    ``"0402"``) and are embedded in a ``LIKE '%term%'`` pattern. A stray ``%``
+    or ``_`` in a term would otherwise act as a wildcard; escape them (paired
+    with ``ESCAPE '\\'`` in the query) so terms match literally.
+    """
+    return term.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
 
 
 def _normalize_lcsc_id(lcsc_part: str) -> int | None:
--- a/src/kicad_tools/parts/lcsc.py
+++ b/src/kicad_tools/parts/lcsc.py
@@ -132,6 +132,36 @@ def _requests_installed() -> bool:
     return importlib.util.find_spec("requests") is not None
 
 
+class _RequestsUnavailableSentinel(BaseException):
+    """Stable sentinel exception used when the ``requests`` extra is absent.
+
+    Returned by :func:`_request_exception_type` so an ``except`` clause that
+    references ``requests.RequestException`` degrades to a class that (a) is
+    stable across calls (so tests can construct and raise it) and (b) is never
+    raised by the real code path when ``requests`` is genuinely missing (that
+    path is handled up front).
+    """
+
+
+def _request_exception_type() -> type[BaseException]:
+    """Return ``requests.RequestException`` if importable, else a sentinel.
+
+    Callers use this in an ``except`` clause so a network failure is caught
+    without a hard top-level ``requests`` import. When the ``requests`` extra is
+    absent (or a test stubs ``_requests_installed`` without installing the real
+    module), the returned sentinel is a *stable* module-level class so the
+    ``except`` reference and any test that raises it agree on identity.
+    """
+    try:
+        import requests  # type: ignore[import-untyped]
+    except ImportError:
+        return _RequestsUnavailableSentinel
+    # ``requests`` ships no type stubs here, so ``RequestException`` is ``Any``;
+    # the explicit annotation keeps callers' ``except`` type-safe.
+    exc_type: type[BaseException] = requests.RequestException
+    return exc_type
+
+
 def _requires_requests(func):
     """Decorator to check if requests is available."""
 
@@ -635,7 +665,6 @@ class LCSCClient:
             fetched_at=datetime.now(),
         )
 
-    @_requires_requests
     def search(
         self,
         query: str,
@@ -643,9 +672,22 @@ class LCSCClient:
         page_size: int = 20,
         in_stock: bool = False,
         basic_only: bool = False,
+        package: str | None = None,
     ) -> SearchResult:
         """
-        Search for parts.
+        Search for parts by free-text query (value + optional package).
+
+        Resolution order mirrors :meth:`lookup`: the live JLCPCB scrape API is
+        authoritative when reachable; on a 403 circuit breaker
+        (:class:`LCSCForbiddenError`) or other network failure, and when a
+        synced offline jlcparts catalog is available, results are served from
+        the catalog instead of raising / returning empty. This keeps parametric
+        matching (``kct parts suggest`` / ``kct export --auto-lcsc``, both of
+        which route through :class:`~kicad_tools.cost.suggest.PartMatcher` ->
+        ``search``) working offline. When the offline fallback is disabled
+        (``use_local_catalog=False``) or no catalog is synced, the historical
+        live-API-only behavior is preserved byte-for-byte (403 propagates, other
+        request errors return an empty :class:`SearchResult`).
 
         Args:
             query: Search query string
@@ -653,11 +695,27 @@ class LCSCClient:
             page_size: Results per page (max 100)
             in_stock: Only return in-stock parts
             basic_only: Only return JLCPCB basic parts
+            package: Optional package filter applied to the offline-catalog
+                fallback (e.g. ``"0402"``); ignored by the live API path.
 
         Returns:
             SearchResult with matching parts
+
+        Raises:
+            LCSCForbiddenError: If the live API returns 403 *and* no offline
+                catalog is available to fall back to (unchanged behavior).
+            LCSCDependencyMissingError: If the ``requests`` extra is missing
+                *and* no offline catalog is available.
         """
-        import requests
+        catalog = self._get_catalog()
+
+        # When requests is unavailable the live API cannot be reached at all.
+        # Serve directly from the offline catalog if one is synced; otherwise
+        # preserve the historical dependency error.
+        if not _requests_installed():
+            if catalog is not None and catalog.available:
+                return self._catalog_search(query, page, page_size, in_stock, package, catalog)
+            raise LCSCDependencyMissingError(PARTS_INSTALL_HINT)
 
         payload = {
             "keyword": query,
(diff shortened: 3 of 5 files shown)


rjwalters:
✅ **Approved!**

Clean, well-scoped implementation of the offline parametric-search fallback tier for `LCSCClient.search()` — the chorus-validation gap where a 403 defeated a fully-synced catalog. Faithful mirror of the existing `lookup()`/`lookup_many()` fallback pattern.

## Verification

**1. SQL safety — solid.**
- `_escape_like` escapes `\`, `%`, `_` in the correct order (backslash first, avoiding double-escape), paired with `ESCAPE '\'`. Verified against literal `%`/`_` terms (`test_catalog_search_escapes_like_wildcards`, and by hand).
- All values are `?`-bound parameters — no f-string injection. The only interpolated SQL text (`where_clause`) is assembled purely from fixed `description LIKE ?` templates + fixed `package =` / `stock >=` clauses; no user data enters the query text.
- `limit` coerced via `int()`; empty/whitespace query short-circuits to `[]` (no unbounded scan); every query carries a `LIMIT`.

**2. Fallback contract — matches the issue's exception-driven intent.**
- 403 + catalog → catalog results; 403 + no catalog → `LCSCForbiddenError` propagates (circuit breaker preserved byte-for-byte).
- `RequestException` + catalog → catalog results; + no catalog → empty `SearchResult` (unchanged).
- `use_local_catalog=False` → historical behavior preserved (raises/empty).
- **Live success (incl. a valid but empty list) does NOT consult the catalog** — the sensible choice: falling through on a legitimately-empty live result would mask real "no such part" answers and change ranking. The non-exception `code != 200` API-error branch likewise returns empty without catalog; that path doesn't trip the 403 breaker and is out of the exception-driven scope — defensible.

**3. Search quality — multi-term AND is term-order independent** (verified `10k 0402` == `0402 10k` == `Resistor 10k` → C25804 against the fixture). `basic > preferred > extended` then `stock DESC` ordering confirmed by `test_catalog_search_library_type_ordering_multi`.

**4. `@_requires_requests` removal — no collateral.** On `main` the decorator was applied to `search()` *only*; no other decorated path lost its guard. Replaced with an internal `_requests_installed()` check + a stable `_request_exception_type()` sentinel so the `except` clause and the no-requests→catalog path both work without the `parts` extra installed.

## Test execution

- `tests/parts/test_jlcparts_catalog.py` — **50 passed** locally.
- The two `test_parts.py` search tests pinned to `use_local_catalog=False` — pass; the pin is justified (their old unconditional-403 assertion is wrong-by-design once a catalog exists).
- Re-ran search/catalog tests against the **rebased** tree (now including #4130's cross-type-mismatch fix) — **60 passed**, no semantic clash.

## Pre-existing test-isolation note (non-blocking)

Confirmed the builder's flag: `test_lookup_not_found`, `test_lookup_api_error`, `test_part_number_normalization` fail **identically on `origin/main`** on machines with a real synced 5.5 GB catalog (they exercise the `lookup()` fallback, not this change; pass in CI where no catalog file exists). Filed follow-up **#4132** to isolate those `lookup()`-path tests from `~/.cache`.

## Rebase note

Rebased twice during review as #4129 and #4130 merged to main — the only conflict each time was the `.loom-managed` worktree marker (a Loom tracking file, resolved to issue 4126). Source diff re-applied byte-identical (521 insertions, 13 deletions) both times.

## CI

All **16/16** checks pass (Test, Type Check, Lint & Format, C++ Build, all Board 00–07 E2E/routing regressions, Diff-Pair, Match-Group, DRC, kicad-cli round-trip). Merge state: **CLEAN** / MERGEABLE. PR uses `Closes #4126`.

PR #4133, opened 2026-07-14
test(parts): isolate lookup-path tests from a machine-local jlcparts catalog

### LINKED ISSUE #4132: test: lookup-path tests fail on machines with a real synced jlcparts catalog (isolate from ~/.cache)
## Summary

Three pre-existing tests in `tests/test_parts.py` fail on any developer machine that has a real 5.5 GB `jlcparts.sqlite3` catalog synced under `~/.cache/kicad-tools/`, and pass in CI only because CI has no catalog file:

- `TestLCSCClient::test_lookup_not_found`
- `TestLCSCClientExtended::test_lookup_api_error`
- `TestLCSCClientExtended::test_part_number_normalization`

These exercise the existing `LCSCClient.lookup()` catalog-fallback path. When the live API is mocked to fail/return-not-found, the real offline catalog resolves the part, so assertions expecting `None`/empty no longer hold. Verified: they fail identically on `origin/main` (base commit) with the catalog present, so this is **not** introduced by any recent change — it is an environment-sensitivity in the tests themselves.

## Fix direction

These tests should be isolated from a machine-local catalog — e.g. construct the `LCSCClient` with `use_local_catalog=False` (the approach PR #4131 already applied to the two `search()`-path tests `test_search_error_handling` and `test_search_raises_forbidden_on_403`), or point `catalog_path` at a guaranteed-absent path, so the assertions are deterministic regardless of `~/.cache` contents.

## Context

Discovered while judging PR #4131 (#4126, `LCSCClient.search()` offline fallback). PR #4131 already pinned its two `search()` tests to `use_local_catalog=False`; this issue tracks doing the same for the three remaining `lookup()`-path tests. Not blocking — CI is green.

rjwalters:
## Summary

Three tests in `tests/test_parts.py` fail on any machine that has a real
`~/.cache/kicad-tools/jlcparts.sqlite3` catalog synced (via `kct sync-catalog`)
and pass in CI only because CI has no catalog. They mock the live JLCPCB API to
fail/miss and assert `None` / a specific `_fetch_part` call, but
`LCSCClient.lookup()`'s offline-catalog fallback (shipped in #4117/#4131)
resolves the real part from disk, flipping the assertions.

Fix: pin the three `lookup()`-path tests to `use_local_catalog=False` — the exact
approach #4131 already applied to the two `search()`-path tests
(`test_search_error_handling`, `test_search_raises_forbidden_on_403`). This makes
the tests hermetic regardless of `~/.cache` contents. No assertions were changed.

Reproduced and verified on a machine with the 5.5 GB catalog present.

## Changes

- `TestLCSCClient::test_lookup_not_found` — add `use_local_catalog=False` (no-catalog "not found" contract).
- `TestLCSCClientExtended::test_lookup_api_error` — add `use_local_catalog=False` (API-error → `None` contract).
- `TestLCSCClientExtended::test_part_number_normalization` — add `use_local_catalog=False` (the catalog fall-through cached the real part, so the 2nd `lookup` hit the cache and skipped `_fetch_part`).
- Added a docstring note to each explaining the isolation, mirroring #4131.

| Affected tests pass ON THIS MACHINE (catalog present) | ✅ | Installed `requests` to un-skip, catalog present: the 3 tests went 3 failed → 3 passed |
| Full `tests/parts/` + `tests/test_parts.py` pass with catalog present | ✅ | `183 passed` (requests installed, catalog on disk) |
| No other test breaks in the presence of a real catalog | ✅ | Empirical sweep: ran the full suites with `requests` installed and catalog present — only these 3 ever failed |
| Changes don't weaken assertions (none deleted) | ✅ | `git diff` shows zero `assert` lines added/removed/changed — only constructor kwargs + docstrings |
| No regression on the no-catalog (CI-like) path | ✅ | Without `requests`: `135 passed, 48 skipped` |

## Test Plan

```
# catalog present, requests installed (this machine):
uv run pytest tests/test_parts.py tests/parts/   # 183 passed
# CI-like (requests absent):
uv run pytest tests/test_parts.py tests/parts/   # 135 passed, 48 skipped
uv run ruff check tests/test_parts.py            # All checks passed!
uv run ruff format tests/test_parts.py --check   # already formatted
```

Closes #4132

### DIFF
--- a/tests/test_parts.py
+++ b/tests/test_parts.py
@@ -448,7 +448,13 @@ class TestLCSCClient:
             assert len(part.prices) == 2
 
     def test_lookup_not_found(self, tmp_path):
-        """Test part not found."""
+        """Test part not found.
+
+        The live API returns no data, so ``lookup`` yields ``None``.
+        ``use_local_catalog=False`` keeps this deterministic even when a real
+        jlcparts catalog is synced on the test machine (#4132) -- otherwise the
+        offline fallback would resolve the part and break the assertion.
+        """
         with patch("kicad_tools.parts.lcsc.LCSCClient._get_session") as mock_session:
             mock_resp = MagicMock()
             mock_resp.json.return_value = {"code": 200, "data": None}
@@ -457,7 +463,7 @@ class TestLCSCClient:
 
             from kicad_tools.parts import LCSCClient
 
-            client = LCSCClient(use_cache=False)
+            client = LCSCClient(use_cache=False, use_local_catalog=False)
             part = client.lookup("C999999")
 
             assert part is None
@@ -687,11 +693,18 @@ class TestLCSCClientExtended:
         assert client.timeout == 60.0
 
     def test_part_number_normalization(self, tmp_path):
-        """Test that part numbers are normalized to uppercase with C prefix."""
+        """Test that part numbers are normalized to uppercase with C prefix.
+
+        ``_fetch_part`` is mocked to return ``None``. ``use_local_catalog=False``
+        keeps this deterministic even when a real jlcparts catalog is synced on
+        the test machine (#4132) -- otherwise the offline fallback would resolve
+        the (real) part before ``_fetch_part`` was reached for the second
+        ``lookup`` call, breaking the ``assert_called_with`` check.
+        """
         from kicad_tools.parts import LCSCClient, PartsCache
 
         cache = PartsCache(db_path=tmp_path / "cache.db")
-        client = LCSCClient(cache=cache)
+        client = LCSCClient(cache=cache, use_local_catalog=False)
 
         with patch("kicad_tools.parts.lcsc.LCSCClient._fetch_part") as mock_fetch:
             mock_fetch.return_value = None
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Test-only isolation fix, verified end-to-end.

## Review

**Correctness of the pins (intent-matching, not suppression).** I traced `LCSCClient.lookup()` in `src/kicad_tools/parts/lcsc.py`: the offline jlcparts catalog is consulted *only* when the live API misses/raises (`part is None` fall-through at line 559). The three pinned tests are exactly the ones that force a live-API miss/error and then assert `None` / a specific `_fetch_part` call — so a real synced catalog resolves the part and flips the assertion. `use_local_catalog=False` restores the intended no-catalog live-API contract. Honest isolation, the exact #4131 precedent (`use_local_catalog=False` already used in `tests/parts/test_jlcparts_catalog.py:270`).

**No assertions weakened.** Diff is confined to `tests/test_parts.py` — only three constructor kwargs plus docstring notes; zero `assert` lines added/removed/changed.

**Empirical sweep independently reproduced (catalog + requests present on this machine):**
- `tests/test_parts.py` on `main`: `3 failed, 98 passed` → on this branch: `101 passed`
- `tests/parts/`: `82 passed`
- Combined `183 passed`, matching the builder's claim. No other test broke.
- Spot-checked the *unpinned* lookup tests (`test_lookup_success`, `test_lookup_many`, `test_lookup_bypass_cache`, etc.): they mock the live API to return a real part, short-circuiting before the catalog is reached — architecturally insensitive to a synced catalog, confirming the sweep isn't just observational.

**Lint/format:** `ruff check` and `ruff format --check` both clean.

**CI:** 16/16 checks pass. Merge state CLEAN.

PR description correctly uses `Closes #4132`.

PR #4141, opened 2026-07-14
fix(cost): parse R-notation resistor values (330R/4R7) so sub-kΩ parts rank correctly

### LINKED ISSUE #4134: parts suggest: sub-kΩ resistor value tokens (330R/33R/470R) not discriminated — rank 10kΩ first (offline catalog, v0.15.1)
## Summary

In v0.15.1, `kct parts suggest` (offline jlcparts catalog fallback) resolves resistor **type and footprint** correctly but does **not discriminate sub-kΩ resistor value tokens** — small-value resistors rank a 10 kΩ part first.

Found while re-validating a consumer board (chorus) against v0.15.1 after the #4130/#4131 fixes landed. The cross-type reject (#4130) and offline fallback (#4131) both work well; this is a narrower residual in value matching.

## Repro (offline catalog, live API 403)

`kct parts suggest` on a schematic with small-value 0402/0603 resistors:

| Ref | Schematic value | Top-ranked suggestion | Correct? |
|-----|-----------------|-----------------------|----------|
| R6  | 100k  | C2906859 (100 kΩ) | ✅ |
| R20 | 10k   | C2906861 (10 kΩ)  | ✅ |
| R1  | 330R  | C2906861 (**10 kΩ**) | ❌ value |
| R7  | 33R   | C2906861 (**10 kΩ**) | ❌ value |
| R8  | 470R  | C2906861 (**10 kΩ**) | ❌ value |

Right type (resistor) and footprint, wrong ohm value. kΩ / 100kΩ values resolve correctly; the sub-kΩ token (`330R`, `33R`, `470R` — Ω with `R` decimal notation) appears not to be parsed/weighted, so the ranker falls back to a generic 10 kΩ.

## Impact

Offline BOM enrichment is no longer hard-blocked (great), but the resolved BOM still needs a manual review pass for small-value passives before assembly sign-off — so it's not yet a hands-off replacement for the hand-run flow.

## Suggested area

Value-token parsing/normalization for resistors using the `R`-as-decimal-point notation (`4R7` = 4.7 Ω, `330R` = 330 Ω, `33R` = 33 Ω) in the offline catalog matcher's value scorer.

Filed from the chorus consumer side (project-shamrock/chorus #33 adoption). Happy to provide the exact schematic/catalog snapshot if useful.

---

## Root Cause (confirmed)

`RESISTOR_PATTERN` in `src/kicad_tools/cost/suggest.py` (line 136) does not
recognize the `R`-as-decimal-point notation used for sub-1kΩ values:

```python
RESISTOR_PATTERN = re.compile(
    r"^(\d+(?:\.\d+)?)\s*([kKmMgG]?)\s*([Ωohm]*)(?:\s+(\d+(?:\.\d+)?%?))?$", re.IGNORECASE
)
```

The multiplier group only accepts `k/K/m/M/g/G` — `R`/`r` is not a
recognized multiplier char, and there's no support for inline-decimal
notation (`4R7` = 4.7 Ω). Reproduced directly:

```python
>>> from kicad_tools.cost.suggest import RESISTOR_PATTERN, parse_component_value
>>> RESISTOR_PATTERN.match("330R")   # None -- full match fails
>>> parse_component_value("330R", "R1").numeric_value   # None
>>> parse_component_value("330R", "R1").search_terms    # []
```

Compare with kΩ values, which work correctly:

```python
>>> parse_component_value("10k", "R20").numeric_value   # 10000.0
>>> parse_component_value("10k", "R20").search_terms    # ['10k']
```

### Why this produces a wrong top-ranked suggestion instead of "no match"

In `PartSuggester.suggest_for_component` (`src/kicad_tools/cost/suggest.py`,
around line 504), when `numeric_value` parsing fails, `parsed.search_terms`
is `[]`, but the package token (e.g. `"0603"`) is still appended:

```python
search_terms = parsed.search_terms.copy()   # [] for "330R"
if package:
    search_terms.append(package)            # ["0603"]
if not search_terms:                        # guard doesn't trip -- list has "0603"
    suggestion.error = "Unable to parse value for search"
    return suggestion
```

So the search proceeds with a **package-only** query (e.g. `"0603"`), and
`_calculate_confidence` (line 588) only adds a value-match bonus when
`parsed.search_terms` is non-empty (line 605) -- which it never is for
`330R`/`33R`/`470R`/`4R7`. The ranker therefore falls back to whatever the
offline catalog's package-only ranking surfaces first for that footprint —
apparently a common/basic 10 kΩ 0603/0402 part (`C2906861` in the repro
table) — with no value discrimination at all. This explains the observed
"right type/footprint, wrong value" failure mode rather than a "no
suggestion found" failure mode.

### Existing in-repo reference implementation

`src/kicad_tools/schematic/footprint_selector.py:396` (`_parse_resistance`)
already correctly handles both suffix (`330R`) and inline-decimal (`4R7`)
notation and can be used as a reference (not a drop-in — it returns a bare
`float`, not a `ParsedValue`, and lives in a different module with an
unrelated same-named `parse_component_value` at line 337 — do not confuse
the two):

```python
def _parse_resistance(value_str: str) -> float:
    value_str = value_str.strip().upper()
    # Handle inline decimal notation (e.g., "4R7" = 4.7 ohms, "4K7" = 4.7k)
    inline_match = re.match(r"^(\d+)([RKM])(\d+)$", value_str)
    if inline_match:
        whole, unit, decimal = inline_match.groups()
        number = float(f"{whole}.{decimal}")
        if unit == "R":
            return number
        elif unit == "K":
            return number * 1000
        elif unit == "M":
            return number * 1_000_000
    # Standard notation (e.g., "10k", "4.7k", "100R")
    match = re.match(r"^([\d.]+)\s*([RKOM])?$", value_str)
    ...
```
Verified locally: `_parse_resistance` returns 330.0 / 33.0 / 470.0 / 4.7
correctly for all four repro tokens.

### Second, weaker instance of the same bug class

`src/kicad_tools/cost/alternatives.py:139` (`AlternativePartFinder.RESISTOR_VALUE_PATTERN`)
uses `.search()` (not anchored `.match()`) with an optional trailing
`(?:ohm|Ω)?` group, so for `330R`/`33R`/`470R` it accidentally gets the
*numeric* value right (the `R` character is simply left unmatched/ignored
by the trailing optional group), but it silently mis-parses inline-decimal
notation: `4R7` -> `4.0` instead of `4.7` (verified locally). This is a
secondary/adjacent bug in the alternatives-suggestion path — same root
cause category (no `R`-as-decimal-point support), separate call site. Fix
if straightforward alongside the primary fix, but do not let it block/scope
-creep the primary fix in `cost/suggest.py`.

### Why fixing `cost/suggest.py` also helps a validation guard

`src/kicad_tools/export/lcsc_value_check.py` explicitly documents
`cost.suggest.parse_component_value` as "the canonical parser per the
#3593 survey" and reuses it to catch value mismatches between a BOM row's
requested value and a candidate LCSC part's parsed value (module docstring,
lines 1-30). Because unparseable values are currently treated as
"cannot validate" (conservative, no false positives), sub-kΩ `R`-notation
values silently skip this guard today. Fixing `RESISTOR_PATTERN` closes
that gap too, at no extra cost to the builder.

## Suggested Fix

Extend `RESISTOR_PATTERN` (or add a pre-check before it, similar to
`_parse_resistance`'s `inline_match`) in `src/kicad_tools/cost/suggest.py`
to handle:
1. **Suffix `R`-as-ohms**: `330R`, `33R`, `470R`, `100R` -> 330, 33, 470, 100 Ω
2. **Inline-decimal `R`**: `4R7` -> 4.7 Ω, `0R5` -> 0.5 Ω
3. **Inline-decimal `K`/`M`** (bonus, same notation family, already partly
   covered by `_parse_resistance` reference): `4K7` -> 4700 Ω

Keep `parse_component_value`'s existing behavior/return contract
(`ParsedValue.numeric_value`, `.unit`, `.search_terms`) unchanged --
only the parsing coverage needs to expand. Also update `search_terms`
construction so sub-1kΩ values produce a term like `"330"` / `"330R"`
(check what substring actually appears in JLC catalog descriptions --
LCSC descriptions typically use bare numeric ohms without a trailing
"R", e.g. "330Ω" or "330 Ohm" -- verify against
`src/kicad_tools/parts/jlcparts_catalog.py` description formatting
before picking the exact term to append) so `_calculate_confidence`'s
value-match bonus (line 605-609) actually fires for these values.

## Affected Files

- `src/kicad_tools/cost/suggest.py` - `RESISTOR_PATTERN` (line 136) and
  `parse_component_value`'s resistor branch (line ~312-338); primary fix.
- `src/kicad_tools/cost/alternatives.py` - `RESISTOR_VALUE_PATTERN` (line
  139) and `_normalize_resistor_value` (line 759); secondary/adjacent fix
  for the same notation gap (inline-decimal mis-parse), fix if in scope.
- `tests/cost/test_suggest.py` - `TestParseComponentValue` class (line 50)
  needs new cases for `330R`, `33R`, `470R`, `4R7`, `0R5`.
- `src/kicad_tools/schematic/footprint_selector.py` - `_parse_resistance`
  (line 396), reference-only, not modified, used as a template for the
  fix in `cost/suggest.py`.

## Acceptance Criteria

- [ ] `parse_component_value("330R", "R1").numeric_value == 330.0`
- [ ] `parse_component_value("33R", "R7").numeric_value == 33.0`
- [ ] `parse_component_value("470R", "R8").numeric_value == 470.0`
- [ ] `parse_component_value("4R7", "R1").numeric_value == pytest.approx(4.7)`
- [ ] Existing kΩ/MΩ cases continue to pass unchanged (`10k` -> 10000,
      `100k` -> 100000, `1M` -> 1_000_000, `4.7k` -> 4700, `100` -> 100)
- [ ] `search_terms` for sub-1kΩ `R`-notation values is non-empty and
      matches how JLC catalog descriptions express bare-ohm values (verify
      against real offline catalog description strings, not assumed format)
- [ ] `kct parts suggest` on the repro schematic (R1=330R, R7=33R, R8=470R)
      no longer ranks a 10 kΩ part first -- top suggestion's description
      should reflect the correct sub-kΩ value (or, if the offline catalog
      genuinely lacks a matching sub-kΩ 0402/0603 part, `_calculate_confidence`
      should score it low enough that a same-footprint 10 kΩ mismatch isn't
      artificially preferred — the fix should be judged by "does the ranker
      now discriminate by value" not "does R1 always resolve exactly")

## Test Plan

- [ ] Automated tests: add cases to `tests/cost/test_suggest.py`
      `TestParseComponentValue` for `330R`, `33R`, `470R`, `4R7`, `0R5`
      (see existing `test_resistor_10k` / `test_resistor_4k7` /
      `test_resistor_1M` for the pattern to follow at line 53-80)
- [ ] Automated tests: regression-guard the existing kΩ/MΩ/plain-ohm cases
      already in `TestParseComponentValue` (must not change behavior)
- [ ] Manual verification: run `kct parts suggest` against a schematic
      containing 330R/33R/470R/4R7 resistors (offline catalog fallback,
      i.e. simulate/force live-API unavailability per #4131's fallback
      path) and confirm the top suggestion is no longer a mismatched
      10 kΩ part
- [ ] Edge cases: `0R5` (sub-1Ω), `4R7` at different reference designators,
      a value with explicit tolerance suffix (e.g. `330R 1%`) if the
      pattern's tolerance capture group interacts with the new branch

rjwalters:
## Summary

`kct parts suggest` mis-ranked a generic 10 kΩ part first for small-value
resistors written in IEC-60062 `R`-as-decimal-point notation (`330R`, `33R`,
`470R`, `4R7`). The offline catalog resolved the right *type* and *footprint*
but the wrong ohm value.

Root cause: `RESISTOR_PATTERN` in `cost/suggest.py` has no `R`-notation
support, so `parse_component_value("330R", "R1")` returned `numeric_value=None`
and empty `search_terms`. `suggest_for_component` then issued a package-only
query and `_calculate_confidence` never applied its value-match bonus, leaving
value discrimination off entirely.

## Changes

- **`cost/suggest.py`**: add `RESISTOR_R_NOTATION_PATTERN` and try it before
  `RESISTOR_PATTERN` in the resistor branch. Handles suffix ohms (`330R`,
  `33R`, `470R`), inline-decimal ohms (`4R7` → 4.7, `0R5` → 0.5), and
  inline-decimal kilo/mega (`4K7` → 4700). Preserves the trailing tolerance
  capture (`330R 1%`). Return contract (`numeric_value`, `unit`,
  `search_terms`, `tolerance`) is unchanged; only parse coverage expands.
- **`cost/alternatives.py`**: fix the adjacent
  `_normalize_resistor_value`, whose unanchored `.search()` pattern silently
  mis-parsed `4R7` as `4.0`. Now normalizes `R`/`K`/`M` inline-decimal tokens
  correctly (`4R7` → 4.7).
- **`tests/cost/test_suggest.py`**: new `TestParseComponentValue` cases for
  `330R`, `33R`, `470R`, `4R7`, `0R5`, `4K7`, `330R 1%`, plus a regression
  guard asserting `search_terms` is non-empty for sub-kΩ `R`-notation.

| `parse_component_value("330R").numeric_value == 330.0` | ✅ | new test `test_resistor_330R` |
| `parse_component_value("33R").numeric_value == 33.0` | ✅ | `test_resistor_33R` |
| `parse_component_value("470R").numeric_value == 470.0` | ✅ | `test_resistor_470R` |
| `parse_component_value("4R7") ≈ 4.7` | ✅ | `test_resistor_4R7` |
| Existing kΩ/MΩ/plain-ohm cases unchanged | ✅ | pre-existing tests still pass (10k→10000, 1M→1e6, 4.7k→4700, 100→100) |
| `search_terms` non-empty for sub-1kΩ R-notation | ✅ | `330R`→`["330"]` matches bare-ohm catalog descriptions (e.g. "330Ω"); `test_resistor_330R_search_term_not_empty` |
| Ranker discriminates by value, no 10kΩ fallback | ✅ | with `numeric_value` + non-empty `search_terms`, `_calculate_confidence`'s value-match bonus now fires for these values (previously never) |
| `0R5` sub-1Ω, tolerance suffix edge cases | ✅ | `test_resistor_0R5`, `test_resistor_330R_with_tolerance` |

## Test Plan

- `uv run pytest tests/cost/ tests/test_alternatives.py` — 65 passed
- `uv run pytest tests/cost/test_suggest.py` — 29 passed
- `uv run ruff check` + `ruff format --check` on all three changed files — clean
- Manual: `parse_component_value` returns correct ohms/terms for
  `330R/33R/470R/4R7/0R5/4K7/330R 1%`; `_normalize_resistor_value("4R7")` → 4.7
  (was 4.0)

Closes #4134

### DIFF
--- a/src/kicad_tools/cost/alternatives.py
+++ b/src/kicad_tools/cost/alternatives.py
@@ -139,6 +139,10 @@ class AlternativePartFinder:
     RESISTOR_VALUE_PATTERN = re.compile(
         r"(\d+(?:\.\d+)?)\s*([kKmMgG]?)\s*(?:ohm|Ω)?", re.IGNORECASE
     )
+    # ``R``-as-decimal-point notation (IEC 60062): ``4R7`` -> 4.7, ``330R`` ->
+    # 330, ``4K7`` -> 4700. Anchored so it only fires on bare value tokens and
+    # never mis-consumes a longer free-text description.
+    RESISTOR_R_NOTATION_PATTERN = re.compile(r"^(\d+)([RKM])(\d*)$", re.IGNORECASE)
     CAPACITOR_VALUE_PATTERN = re.compile(r"(\d+(?:\.\d+)?)\s*([pnuμmfF]?)\s*[Ff]?", re.IGNORECASE)
     INDUCTOR_VALUE_PATTERN = re.compile(r"(\d+(?:\.\d+)?)\s*([pnuμmH]?)\s*[Hh]?", re.IGNORECASE)
 
@@ -758,6 +762,18 @@ class AlternativePartFinder:
 
     def _normalize_resistor_value(self, value: str) -> float | None:
         """Normalize resistor value to ohms."""
+        # Handle ``R``-as-decimal-point notation first (4R7 -> 4.7, 330R -> 330,
+        # 4K7 -> 4700). The generic RESISTOR_VALUE_PATTERN below leaves the ``R``
+        # unmatched and would mis-parse ``4R7`` as 4.0.
+        r_match = self.RESISTOR_R_NOTATION_PATTERN.match(value.strip())
+        if r_match:
+            whole = r_match.group(1)
+            unit = r_match.group(2).lower()
+            decimal = r_match.group(3)
+            num = float(f"{whole}.{decimal}") if decimal else float(whole)
+            r_multipliers = {"r": 1, "k": 1e3, "m": 1e6}
+            return num * r_multipliers.get(unit, 1)
+
         match = self.RESISTOR_VALUE_PATTERN.search(value)
         if not match:
             return None
--- a/src/kicad_tools/cost/suggest.py
+++ b/src/kicad_tools/cost/suggest.py
@@ -136,6 +136,15 @@ class SuggestionResult:
 RESISTOR_PATTERN = re.compile(
     r"^(\d+(?:\.\d+)?)\s*([kKmMgG]?)\s*([Ωohm]*)(?:\s+(\d+(?:\.\d+)?%?))?$", re.IGNORECASE
 )
+# ``R``-as-decimal-point notation for resistors, where ``R``/``K``/``M`` acts
+# as both the unit and the decimal separator (IEC 60062). Handles:
+#   - suffix ohms: ``330R`` -> 330, ``33R`` -> 33, ``470R`` -> 470
+#   - inline-decimal ohms: ``4R7`` -> 4.7, ``0R5`` -> 0.5
+#   - inline-decimal kilo/mega: ``4K7`` -> 4700, ``1M2`` -> 1_200_000
+# Optional trailing tolerance (e.g. ``330R 1%``) is captured in group 4.
+RESISTOR_R_NOTATION_PATTERN = re.compile(
+    r"^(\d+)([RKM])(\d*)(?:\s+(\d+(?:\.\d+)?%?))?$", re.IGNORECASE
+)
 CAPACITOR_PATTERN = re.compile(
     r"^(\d+(?:\.\d+)?)\s*([pnuμmµ]?)[fF]?(?:\s+(\d+[Vv]))?(?:\s+(\d+(?:\.\d+)?%?))?$",
     re.IGNORECASE,
@@ -311,23 +320,48 @@ def parse_component_value(value: str, reference: str = "") -> ParsedValue:
 
     # Parse resistor values
     if component_type == ComponentType.RESISTOR:
-        match = RESISTOR_PATTERN.match(value)
-        if match:
-            num = float(match.group(1))
-            multiplier = match.group(2).upper() if match.group(2) else ""
+        num: float | None = None
+        tolerance: str | None = None
 
-            if multiplier == "K":
+        # Try ``R``-as-decimal-point notation first (330R, 4R7, 4K7, ...).
+        # This must precede RESISTOR_PATTERN because the latter cannot match
+        # the trailing ``R`` and would otherwise leave the value unparsed,
+        # producing a package-only search that mis-ranks a generic value.
+        r_match = RESISTOR_R_NOTATION_PATTERN.match(value.strip())
+        if r_match:
+            whole = r_match.group(1)
+            unit = r_match.group(2).upper()
+            decimal = r_match.group(3)
+            # "4R7" -> 4.7, "330R" -> 330.0 (empty decimal treated as ".0")
+            num = float(f"{whole}.{decimal}") if decimal else float(whole)
+
+            if unit == "K":
                 num *= 1000
-            elif multiplier == "M":
+            elif unit == "M":
                 num *= 1_000_000
-            elif multiplier == "G":
-                num *= 1_000_000_000
 
+            tolerance = r_match.group(4)
+        else:
+            match = RESISTOR_PATTERN.match(value)
+            if match:
+                num = float(match.group(1))
+                multiplier = match.group(2).upper() if match.group(2) else ""
+
+                if multiplier == "K":
+                    num *= 1000
+                elif multiplier == "M":
+                    num *= 1_000_000
+                elif multiplier == "G":
+                    num *= 1_000_000_000
+
+                tolerance = match.group(4)
+
+        if num is not None:
             parsed.numeric_value = num
             parsed.unit = "Ω"
 
-            if match.group(4):
-                parsed.tolerance = match.group(4)
+            if tolerance:
+                parsed.tolerance = tolerance
 
             # Build search terms
             if num >= 1_000_000:
--- a/tests/cost/test_suggest.py
+++ b/tests/cost/test_suggest.py
@@ -79,6 +79,69 @@ class TestParseComponentValue:
         assert result.numeric_value == 1_000_000
         assert "1M" in result.search_terms
 
+    def test_resistor_330R(self):
+        """Test parsing sub-1kΩ suffix-R notation (330R -> 330 Ω)."""
+        result = parse_component_value("330R", "R1")
+        assert result.component_type == ComponentType.RESISTOR
+        assert result.numeric_value == 330.0
+        assert result.unit == "Ω"
+        assert "330" in result.search_terms
+
+    def test_resistor_33R(self):
+        """Test parsing sub-1kΩ suffix-R notation (33R -> 33 Ω)."""
+        result = parse_component_value("33R", "R7")
+        assert result.component_type == ComponentType.RESISTOR
+        assert result.numeric_value == 33.0
+        assert "33" in result.search_terms
+
+    def test_resistor_470R(self):
+        """Test parsing sub-1kΩ suffix-R notation (470R -> 470 Ω)."""
+        result = parse_component_value("470R", "R8")
+        assert result.component_type == ComponentType.RESISTOR
+        assert result.numeric_value == 470.0
+        assert "470" in result.search_terms
+
+    def test_resistor_4R7(self):
+        """Test parsing inline-decimal R notation (4R7 -> 4.7 Ω)."""
+        result = parse_component_value("4R7", "R1")
+        assert result.component_type == ComponentType.RESISTOR
+        assert result.numeric_value == pytest.approx(4.7)
+        assert result.unit == "Ω"
+        assert result.search_terms  # non-empty so value bonus can fire
+
+    def test_resistor_0R5(self):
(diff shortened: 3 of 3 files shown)


rjwalters:
✅ **Approved!**

Clean, well-scoped fix for the IEC-60062 R-notation gap (#4134). Verified end-to-end.

## Correctness
Parsed every claimed case on the PR branch, plus regression cases the diff could have broken:

| Input | `numeric_value` | search_terms | tolerance |
|-------|-----------------|--------------|-----------|
| `330R` | 330.0 | `['330']` | — |
| `33R` | 33.0 | `['33']` | — |
| `470R` | 470.0 | `['470']` | — |
| `4R7` | 4.7 | `['4.7']` | — |
| `0R5` | 0.5 | `['0.5']` | — |
| `4K7` | 4700.0 | `['4.7k']` | — |
| `1M2` | 1_200_000 | `['1.2M']` | — |
| `330R 1%` | 330.0 | `['330']` | `1%` |
| `10k` | 10000.0 | `['10k']` | — (no regression) |
| `4.7k` | 4700.0 | `['4.7k']` | — (no regression) |
| `1M` | 1_000_000 | `['1M']` | — (no regression) |
| `2K2` | 2200.0 | `['2.2k']` | — (no regression) |

`_normalize_resistor_value("4R7")` in `cost/alternatives.py` now returns **4.7** (was mis-parsed as 4.0 by the unanchored `.search()`). Confirmed.

Notes on the design that hold up under review:
- The R-notation pattern is anchored (`^...$`) in both files, so it can never mis-consume a longer free-text catalog description — it only fires on bare value tokens.
- Trying `RESISTOR_R_NOTATION_PATTERN` before `RESISTOR_PATTERN` is required, since the trailing `R` is unparseable by the old pattern. The `[RKM]` class correctly excludes `G` (giga), which falls through to `RESISTOR_PATTERN` where giga is still handled — no coverage lost.
- Tolerance capture (group 4) is preserved in both the new and existing branches.

## Test Scoping
**Strategy**: scoped to changed modules — `tests/cost/test_suggest.py` + `tests/test_alternatives.py`
**Result**: 65 passed (29 in test_suggest.py, includes the 8 new `TestParseComponentValue` cases + regression guard). `ruff check` + `ruff format --check` clean on all 3 changed files.

## Scope
Exactly 3 files (`cost/suggest.py`, `cost/alternatives.py`, `tests/cost/test_suggest.py`), +124/-11. No out-of-scope changes. PR body uses `Closes #4134`.

## CI
All 16 checks pass (Test 15m40s, Lint, Type Check, all Board E2E/routing regressions, C++ build). `mergeStateStatus: CLEAN`.

PR #4142, opened 2026-07-14
feat(sch): add fix-annotation command for hierarchy-aware power/flag repair

### LINKED ISSUE #4135: sch: hierarchy-aware power/flag-symbol annotation repair (unnumbered refs, missing (instances) blocks, cross-sheet duplicates) with net-neutrality gate
## Motivation (consumer: project-shamrock/chorus#24)

A shipping 6-sheet hierarchical design has a persistent `schematic has annotation errors` warning on every `kicad-cli sch export netlist`. Verified root cause (full bisection evidence in project-shamrock/chorus#24):

- ~14 placed power/ground port symbols in child sheets (`dac`, `connectors`, `sync`) have **no `(instances)` block at all** and carry un-numbered references (`#GNDA`, `#+3.3V`, `#+5V`, `#GNDD`).
- Duplicate reference designators across the flattened hierarchy (`#GNDD`×9, `#+3.3V`×3, `#PWR02`×2, `#PWR40`×2 …) plus malformed refs (`#PWR40` vs `#PWR040`).
- A full manual re-annotation of every power/flag symbol was **proven net-neutral**: 79 nets before/after with identical `(node ref/pin)` membership (power symbols connect by value, so designator renumbering moves no nets).

The fix is exactly what KiCad GUI "Tools → Annotate Schematic" does, but the goal of kct is to avoid the GUI. The consumer issue is blocked on headless tooling.

## Current kct behavior (v0.15.1, verified)

```
$ kct sch re-annotate --dry-run --prefix '#PWR' chorus-test-revA.kicad_sch   # root sheet
No references to renumber.
```
— does not traverse the sheet hierarchy and/or does not consider power-symbol refs.

```
$ kct sch repair-instances --dry-run chorus-test-revA.kicad_sch              # root sheet
All symbols have project instances. No repairs needed.

$ kct sch repair-instances --dry-run dac.kicad_sch                           # child sheet, standalone
    #GNDA2  (power:GNDA) [wrong project]
    #PWR77  (power:GNDA) [wrong project]
    ...
```
— run from the root it sees nothing wrong; run on a child sheet standalone it analyzes the file outside its hierarchy context (project name/paths resolve against the child file, so repairs would write instance paths scoped to the wrong project).

## Requested feature

A hierarchy-aware annotation repair, e.g. `kct sch fix-annotation <root.kicad_sch>` (or flags on `re-annotate`/`repair-instances`), that:

1. Walks the full sheet hierarchy from the root (like KiCad's annotator).
2. Detects and repairs power/flag symbols: assigns globally-unique `#PWR0xx`/`#FLG0xx` designators to un-numbered or duplicated refs (including net-name-styled refs like `#GNDD`, `#+3.3V`), preserving real-component annotations.
3. Adds missing `(instances)` blocks with the correct project name and sheet path.
4. **Built-in net-neutrality gate**: exports the netlist before/after and refuses to write (or exits non-zero) if net membership changes — designator renumbering on power symbols must be a zero-net-diff operation.
5. `--dry-run` + `--backup` like the existing sch mutators.

## Acceptance sketch

On the chorus design: `kicad-cli sch export netlist` emits no `annotation errors` warning afterward, ERC stays 0/0, netlist net-membership diff vs before is empty, board files untouched.

---

### Root cause: confirmed against current `main` (v0.15.1)

Both existing tools already do hierarchy traversal via
`_collect_schematic_files()` (`src/kicad_tools/cli/sch_set_footprint.py:56`,
depth-first walk of `(sheet ... (property "Sheetfile" "..."))` blocks), and
both call it correctly from the root
(`sch_re_annotate.py:514`, `sch_repair_instances.py:581`). So "no hierarchy
traversal" is *not* actually the defect — the walker is already shared and
correct. The real gaps are narrower and more specific:

**Gap 1 — `re-annotate` never renumbers net-name-styled power refs.**
`_parse_reference()` (`sch_re_annotate.py:80-98`) only recognizes two ref
shapes: `PREFIX<digits><unit?>` (e.g. `#PWR02`) and `PREFIX?` (unannotated,
e.g. `R?`). A ref like `#GNDD` or `#+3.3V` matches neither regex, so it
falls through to `return ref, None, ""` — the **entire string** becomes the
"prefix" (e.g. prefix=`"#GNDD"`, number=`None`). Verified directly:

```python
>>> _parse_reference("#GNDD")
('#GNDD', None, '')
>>> _parse_reference("#+3.3V")
('#+3.3V', None, '')
```

In `_assign_numbers()` (line 753) / `_assign_numbers_unannotated_only()`
(line 834), the exclusion check is `prefix in EXCLUDED_PREFIXES or
prefix.startswith("#")`, gated open only by `include_power and prefix in
_POWER_PREFIXES` where `_POWER_PREFIXES = {"#PWR", "#FLG"}` (lines 37-40).
Because `"#GNDD" not in _POWER_PREFIXES`, symbols with this exact ref are
dropped even with `--include-power` — confirmed empirically:

```python
>>> _assign_numbers([{'reference': '#GNDD', 'prefix': '#GNDD', 'number': None,
...                    'unit_suffix': '', 'uuid': 'u1'}], None, 1, include_power=True)
{}   # empty — no rename generated, ever
```

So `#GNDD`, `#+3.3V`, `#+5V`, `#GNDA` (and any other net-name-styled power
ref) are permanently invisible to `re-annotate` regardless of flags. This is
why the issue's reproduction shows "No references to renumber."

**Gap 2 — `repair-instances` has no concept of "wrong project because
run standalone."** When run from the root, `_extract_symbols_with_instance_info`
(`sch_repair_instances.py:154`) correctly special-cases power symbols
(`lib_id.startswith("power:")`) to be skipped **only** when they have no
instances block and no loose blocks (line 238-242) — this is deliberate
(see `sch_validate.py:772` `check_missing_project_instances` docstring,
which documents the same three failure modes: missing / wrong-project /
loose-project-blocks). The tool doesn't fail to detect anything at the
root — the consumer's repro shows it correctly finds *zero* issues from the
root because the malformed refs there are power symbols whose `lib_id`
starts with `power:` and (in the repro) apparently already carry
*something* recognized as an instance for that project, OR — more likely
per the consumer's own bisection — the malformed symbols are net-name-styled
refs that `repair-instances`' rename logic (`_BARE_REF_PREFIX_MAP`, line 41)
only maps `#PWR_FLAG` → `#FLG`, not `#GNDD`/`#+3.3V` → `#PWR`/`#FLG`. When
run *standalone* on the child sheet file, `_detect_project_info()`
(`sch_re_annotate.py:307`, reused by `repair-instances`) derives
`project_name = root_path.stem` from **whatever file is passed as
`schematic_path`** — so running it on `dac.kicad_sch` directly makes it
believe `dac` is the project name, and correctly-scoped root-project
instances now look "wrong" by comparison. This confirms the issue's
diagnosis: **there is no problem with the walker itself; the problem is (a)
net-name-styled refs are unrecognized text shapes in both tools, and (b)
neither tool offers a "repair power symbols specifically, including their
ref-string normalization to `#PWR`/`#FLG` family" mode that also verifies
net-neutrality before writing.**

### Reusable hierarchy-walking / netlist infrastructure

- `_collect_schematic_files()` — `src/kicad_tools/cli/sch_set_footprint.py:56`
  — depth-first hierarchy walk, already used by both existing sch mutators.
  Reuse directly; do not reimplement.
- `_detect_project_info()` — `src/kicad_tools/cli/sch_re_annotate.py:307`
  — derives `(project_name, root_uuid, file_instance_paths)` from the root
  file. Reuse directly (already imported by `sch_repair_instances.py`).
- `_extract_symbols_with_instance_info()` — `src/kicad_tools/cli/sch_repair_instances.py:154`
  — the more robust, bracket-depth-based symbol scanner (vs. the regex-based
  one in `sch_re_annotate.py`); already understands `power:` lib_id
  special-casing and the three instance-block failure modes. Prefer this
  scanner as the basis for new power-symbol detection logic.
- `HierarchyBuilder` / `build_hierarchy()` — `src/kicad_tools/schema/hierarchy.py:247,362`
  — an alternative, more object-oriented hierarchy tree (used elsewhere in
  the codebase, e.g. pin/label cross-sheet checks). Not currently used by
  the sch mutators; evaluate whether it's a better fit than
  `_collect_schematic_files()`, but there's no functional gap forcing a
  switch — prefer reusing the mutators' existing walker for consistency
  with `re-annotate`/`repair-instances` unless a concrete need arises.
- `check_missing_project_instances()` — `src/kicad_tools/cli/sch_validate.py:772`
  — documents (and can be reused to *verify*) the three malformed-instance
  shapes this issue's fix must repair. Consider calling this validator
  post-repair as a self-check (`0 remaining validation errors`) in addition
  to the net-neutrality gate.
- `Netlist` / `Net` / `NetNode` dataclasses + `export_netlist()` /
  `load_netlist()` — `src/kicad_tools/cli/export_netlist.py:29-330` —
  ready-made schematic-netlist S-expression parser wrapping
  `kicad-cli sch export netlist`. **No existing "diff two netlists"
  helper was found anywhere in the codebase** (`lvs/board_lvs.py`'s
  `compare_netlists()` is schematic-vs-PCB, not schematic-vs-schematic) —
  the net-neutrality gate is new code. Build it directly from `Netlist.nets`:
  for each state (before/after) construct
  `{net.name: frozenset((node.reference, node.pin) for node in net.nodes)}`
  and diff by value-set membership (not by net *name*, since renumbering a
  duplicate power ref can legitimately cause KiCad to reassign net "codes"/
  labels — what must stay invariant is *which pins are electrically joined*,
  not the net's display name). See Implementation Notes below for a
  ref-renaming wrinkle that affects this comparison.

### CLI surface recommendation

Add a **new subcommand** `kct sch fix-annotation <root.kicad_sch>` rather
than overloading flags onto `re-annotate` or `repair-instances`. Rationale:

- `re-annotate` and `repair-instances` are both "pure" mutators with no
  I/O side effects beyond the schematic files themselves (no subprocess
  calls, no `kicad-cli` dependency). Adding a `kicad-cli`-dependent
  net-neutrality gate to either would be a categorically different
  behavior (external tool dependency, subprocess calls, possible failure
  mode when `kicad-cli` isn't installed) bolted onto tools that currently
  have none of that.
- The new command's job is explicitly narrower and more opinionated than
  either existing tool: it *only* targets power/flag symbols, and it
  *always* verifies net-neutrality before writing. That's a distinct,
  named workflow worth its own subcommand, matching the existing pattern
  of single-purpose `sch` subcommands (`re-annotate`, `repair-instances`,
  `remove-component`, etc. — see `src/kicad_tools/cli/parser.py:1397-1451`
  and dispatch in `src/kicad_tools/cli/commands/schematic.py:606-632`).
- Internally, `fix-annotation` should **compose** the existing building
  blocks rather than duplicate them: reuse `_collect_schematic_files`,
  `_detect_project_info`, `_extract_symbols_with_instance_info`,
  `_format_reference`, and the rename/instance-insertion helpers
  (`_apply_uuid_reference_rename`, `_insert_project_instance`) from both
  existing modules. The new logic is really just: (1) a smarter
  power/flag-ref classifier that also recognizes net-name-styled refs, and
  (2) the net-neutrality gate wrapping the existing write path.

New module: `src/kicad_tools/cli/sch_fix_annotation.py`, wired into
`parser.py` (new `sch_subparsers.add_parser("fix-annotation", ...)`) and
`commands/schematic.py` (new `elif args.sch_command == "fix-annotation":`
branch), following the exact structure of the `re-annotate`/
`repair-instances` entries already there.

### Implementation guidance

1. **Power/flag ref classification.** Extend (or add alongside)
   `_parse_reference()` a power-symbol-aware classifier that recognizes:
   - Already-correct: `#PWR01`, `#FLG03` (2-digit zero-padded, per
     `_format_reference`'s existing convention at
     `sch_re_annotate.py:46-55`).
   - Malformed-but-numbered: `#PWR2` (missing zero-pad), `#PWR040` vs
     `#PWR40` (inconsistent padding causing duplicate-looking refs) —
     normalize all to the 2-digit convention.
   - Net-name-styled (no number at all): `#GNDD`, `#GNDA`, `#+3.3V`,
     `#+5V`, and the general shape `#<net-name>` where `<net-name>` is
     whatever KiCad placed as the default ref for a power-symbol
     `lib_id` (e.g. `power:GNDA`, `power:+3.3V`). Treat any symbol whose
     `lib_id` starts with `power:` and whose ref does **not** already
     match the `#PWR\d\d+`/`#FLG\d\d+` canonical shape as needing
     reassignment — do not try to pattern-match the net-name text itself
     (it's arbitrary), key off `lib_id.startswith("power:")` instead
     (same technique already used in `sch_repair_instances.py:238` and
     `sch_validate.py`).
   - Ground vs. non-ground `power:` symbols both get `#PWR` (per KiCad
     convention); genuine flag symbols (`power:PWR_FLAG`, `lib_id`
     containing `FLAG`) get `#FLG`, matching `_BARE_REF_PREFIX_MAP` in
     `sch_repair_instances.py:41-43` (extend this map, or generalize it,
     rather than hardcoding one entry).
2. **Duplicate resolution across the flattened hierarchy.** Power symbols
   are globally unique by design in KiCad (unlike `R`/`C`/`U` which are
   per-project unique but locally placed) — collect *all* `#PWR`/`#FLG`
   refs across every file in `_collect_schematic_files()` order first
   (same two-phase "collect, then assign" pattern already used by
   `_build_continuous_mapping` in `sch_re_annotate.py:693`), then assign
   sequential non-conflicting numbers globally, exactly like
   `re-annotate`'s existing continuous-numbering mode — reuse
   `_assign_numbers`'s counter/reservation logic if feasible instead of
   reimplementing it.
3. **Missing `(instances)` blocks.** Reuse
   `_insert_project_instance()` (`sch_repair_instances.py:348`) — it
   already handles all three shapes (missing / wrong-project / loose).
4. **Net-neutrality gate** (the genuinely new piece):
   - Before making any file writes: export netlist from the *unmodified*
     tree via `export_netlist()` (`export_netlist.py:288`), parse via
     `load_netlist()`, and build a normalized net-membership snapshot.
   - Apply the proposed rename/instance-block-insertion changes to
     **in-memory buffers** (not yet written to disk).
   - Write the modified buffers to a **temporary directory** mirroring the
     hierarchy (or write to the real files but keep `--backup` content, then
     roll back on gate failure — see Implementation Notes below for why a
     temp-dir approach is cleaner) and export netlist again; parse and
     snapshot.
   - Compare snapshots by `{net_name: frozenset of (something that
     survives ref renaming, pin_number)}`. **Important wrinkle**: because
     the renamed symbols themselves change reference designator, a
     naive `(reference, pin)` tuple comparison will always show a diff for
     every renamed power symbol's own nodes (its ref changed, so its
     `(ref, pin)` identity changed even though the symbol didn't move
     electrically). The gate must compare net membership **keyed by
     symbol UUID + pin**, not `(reference, pin)` — but
     `kicad-cli sch export netlist` output only carries `tstamps`/`ref`,
     not UUID, in the node entries (verify this against actual
     `kicad-cli` output before implementing; if UUIDs aren't exposed,
     the practical approach is to build an old-ref→new-ref mapping during
     the repair phase and translate the "before" snapshot's ref strings
     through that mapping before diffing, so the comparison is
     apples-to-apples). This translation step needs to be explicit code,
     tested directly (not just "assume it works").
   - If the translated before/after snapshots differ: abort, print the
     diff (added/removed net memberships), exit non-zero, write no files
     (or roll back any files already written), and do not consume the
     `--backup`.
   - If they match: proceed to the real write (same `create_backup()` +
     `write_text()` pattern as both existing tools).
   - This gate should be **on by default** (matches the issue's framing
     of it as "built-in", not optional) but consider a `--skip-net-check`
     escape hatch for environments without `kicad-cli` installed, clearly
     labeled as unsafe.
5. **`--dry-run` / `--backup`.** Match existing flag semantics exactly:
   `--dry-run`/`-n` prints the mapping and instance-block plan without
   writing (and without needing to run the net-neutrality gate at all,
   since nothing is written — or optionally still run it as an extra
   dry-run confidence check, documented either way); `--backup` creates
   `.bak` files via `create_backup()` before any real write, consistent
   with `modify_schematic.create_backup` used by both existing tools.

### Requires `kicad-cli` on PATH

Unlike `re-annotate`/`repair-instances` (pure text mutators), this new
command has a hard dependency on `kicad-cli` being available (via
`find_kicad_cli()`, `src/kicad_tools/cli/runner.py:16`) to run the
net-neutrality gate. Fail with a clear, actionable error message if
`kicad-cli` cannot be located — do not silently skip the gate.

### Out of scope

- General ERC/ratsnest repair beyond annotation (duplicate refs on
  *non-power* symbols, unconnected pins, etc.) — this issue is scoped to
  power/flag-symbol annotation specifically, per the consumer's verified
  root cause.
- Rewriting `_parse_reference()`'s general-purpose regex to handle every
  possible ref shape — the fix only needs `lib_id.startswith("power:")`
  based classification for the symbols this issue targets; leave the
  general-purpose annotation logic in `re-annotate` untouched.

## Affected Files

- `src/kicad_tools/cli/sch_fix_annotation.py` — **new file**, core
  implementation (power/flag ref classification, global renumbering,
  instance-block repair, net-neutrality gate, dry-run/backup).
- `src/kicad_tools/cli/parser.py` — add `sch fix-annotation` subparser
  (near `re-annotate`/`repair-instances` at lines 1397-1451).
- `src/kicad_tools/cli/commands/schematic.py` — add dispatch branch for
  `args.sch_command == "fix-annotation"` (near lines 606-632); update the
  `sch --help` command list at lines 31-32.
- `src/kicad_tools/cli/sch_re_annotate.py` — reuse (import, do not
  duplicate) `_collect_schematic_files` (actually imported from
  `sch_set_footprint`), `_detect_project_info`, `_format_reference`,
  `_apply_uuid_reference_rename`, `_detect_indent`. No changes expected
  unless a shared helper needs extraction.
- `src/kicad_tools/cli/sch_repair_instances.py` — reuse
  `_extract_symbols_with_instance_info`, `_insert_project_instance`,
  `_scan_symbol_children`. Consider extending `_BARE_REF_PREFIX_MAP`
  (line 41) to generalize net-name-styled `power:` refs → `#PWR`/`#FLG`,
  since that mapping is directly relevant to this issue's Gap 1.
- `src/kicad_tools/cli/export_netlist.py` — reuse `export_netlist()`,
  `load_netlist()`, `Netlist`, `Net`, `NetNode` for the net-neutrality
  gate. No changes expected unless the S-expression schema needs a new
  field (e.g. `tstamps`/UUID on nodes) for the ref-translation approach.
- `tests/test_sch_fix_annotation.py` — **new file**, following the
  synthetic-fixture convention of `tests/test_sch_re_annotate.py` and
  `tests/test_sch_repair_instances.py` (hand-built multi-sheet
  `kicad_sch` text fixtures with malformed power refs — do NOT use the
  chorus design, which is local-only and unavailable in CI; see Test Plan).

## Test Plan

- [ ] **Automated tests** in `tests/test_sch_fix_annotation.py`, built on
  **synthetic hierarchical fixtures constructed in the test file** (2-3
  sheet hierarchy with a root + at least one child sheet), mirroring the
  fixture style already used in `tests/test_sch_repair_instances.py`
  (see `ROOT_SCHEMATIC`/`SUB_SCHEMATIC_MISSING` constants there). The
  chorus fixture is **local-only and not available in CI** (per repo
  memory: "NO chorus in CI, fixture symlinked locally but dangles in
  worktrees") — tests must be fully self-contained, no dependency on any
  external/local-only fixture.
  - [ ] Net-name-styled unnumbered power refs (`#GNDD`, `#+3.3V`) get
    assigned canonical `#PWR0xx`/`#FLG0xx` designators.
  - [ ] Duplicate power refs across sheets (e.g. two `#GNDD` symbols in
    different child sheets) get distinct, non-conflicting numbers.
  - [ ] Malformed zero-padding (`#PWR40` vs `#PWR040` referring to what
    should be treated as distinct symbols) is normalized without
    collision.
  - [ ] Missing `(instances)` blocks are created with correct project
    name + hierarchy path when the tool is run from the **root** (verify
    the fix does NOT reproduce the "wrong project when run standalone"
    bug — i.e., there should be no code path that lets a user
    accidentally invoke this on a non-root file and get a wrong project
    name; consider validating that `schematic_path` looks like a root
    file, or documenting this constraint clearly).
  - [ ] Net-neutrality gate: construct a before/after pair where a
    renumber is genuinely net-neutral (power symbols only) — the gate
    must return "no diff" and allow the write.
  - [ ] Net-neutrality gate: construct a synthetic **regression** case
    (e.g. mock/monkeypatch the netlist export to simulate a hypothetical
    bug that drops a node) and confirm the gate detects the diff, aborts,
    writes no files, and exits non-zero. This test should not require
    reproducing a real drop bug — it validates the gate's own comparison
    logic in isolation, decoupled from `kicad-cli` availability (mock
    `export_netlist`/`load_netlist` at the boundary).
  - [ ] `--dry-run`: no files modified, mapping printed/returned.
  - [ ] `--backup`: `.bak` files created before any write, matching
    `create_backup()` semantics used elsewhere.
  - [ ] Real component refs (`R1`, `C3`, `U2`) are left untouched —
    only `power:`-lib_id symbols are in scope.
  - [ ] Tests that require `kicad-cli` (the actual net-neutrality gate
    subprocess path) should be skipped gracefully (`pytest.mark.skipif`)
    when `find_kicad_cli()` returns `None`, matching the pattern used
    elsewhere for kicad-cli-dependent tests (grep existing `skipif.*kicad`
    patterns in `tests/` for the established idiom) — the pure-Python
    classification/renumbering logic should still be tested without
    `kicad-cli`.
- [ ] **Manual verification**: on a local reproduction (e.g. a small
  synthetic multi-sheet project built by hand or via the existing test
  fixtures saved to disk), run `kct sch fix-annotation root.kicad_sch
  --dry-run` and confirm the printed plan matches expectations; then run
  without `--dry-run` and confirm `kicad-cli sch export netlist` no
  longer reports "schematic has annotation errors" and ERC stays clean.
- [ ] **Edge cases**:
  - Symbol with an already-correct `#PWR01` ref and existing correct
    instances block — no changes, no-op.
  - A `power:` symbol with `has_wrong_project` (child sheet standalone
    artifact) run correctly from the root — repaired in place.
  - Empty/no power symbols in the hierarchy — command reports "nothing to
    fix" and exits 0 cleanly (matching `re-annotate`'s "No references to
    renumber." convention).

rjwalters:
## Summary

Adds `kct sch fix-annotation <root.kicad_sch>`, a new hierarchy-aware repair for the power/flag-symbol annotation errors that `kicad-cli sch export netlist` reports as `schematic has annotation errors`. This unblocks the consumer design (project-shamrock/chorus#24) which cannot re-annotate power symbols headlessly today.

Root cause (confirmed per the curated issue): `re-annotate`'s `_parse_reference()` cannot parse net-name-styled power refs (`#GNDD`, `#+3.3V`) — the whole string becomes the "prefix", so those symbols are permanently invisible even with `--include-power`. The new command classifies power symbols by `lib_id.startswith("power:")` instead of pattern-matching the arbitrary net-name text.

## Changes

- **New `src/kicad_tools/cli/sch_fix_annotation.py`** composing existing building blocks:
  - Walks the hierarchy via `_collect_schematic_files` (reused).
  - Classifies every `power:` symbol; canonicality is a round-trip check through `_format_reference`, so `#PWR40` is kept but `#PWR040` (which would format to `#PWR40`) is normalized.
  - `build_rename_plan` assigns globally-unique `#PWR0xx`/`#FLG0xx` designators, reserving numbers held by already-canonical symbols so no collisions are introduced. Flag symbols get `#FLG`, other power symbols get `#PWR`.
  - Repairs missing / wrong-project / loose `(instances)` blocks via `_insert_project_instance` (reused), and updates the stale `(reference ...)` inside instance blocks of renamed symbols.
  - **Net-neutrality gate**: exports the netlist before/after via `kicad-cli`, applies changes to a temp-dir copy of the hierarchy, and diffs net membership *keyed by (ref, pin) with the before-snapshot translated through the rename mapping* (naive comparison would always diff on renames). Refuses to write (exit 2) on any membership change; fails closed if an export fails.
  - Degrades with a clear, actionable error when `kicad-cli` is absent; `--skip-net-check` bypasses the gate (labeled unsafe). `--dry-run` and `--backup` match existing sch mutators.
- **`parser.py`**: new `sch fix-annotation` subparser.
- **`commands/schematic.py`**: dispatch branch + updated `sch --help` command list.
- **New `tests/test_sch_fix_annotation.py`**: synthetic multi-sheet fixtures (chorus is local-only / not in CI).

| Net-name-styled refs (`#GNDD`, `#+3.3V`) get canonical designators | ✅ | `TestBuildRenamePlan::test_net_name_styled_gets_canonical` |
| Duplicate refs across sheets get distinct numbers | ✅ | `test_duplicate_refs_get_distinct_numbers` + e2e `test_writes_and_renames_power_symbols` |
| Inconsistent zero-padding normalized without collision | ✅ | `test_padding_normalization_no_collision` (`#PWR40` kept, `#PWR040`→`#PWR01`) |
| Missing `(instances)` blocks created with correct project + hierarchy path from root | ✅ | e2e asserts `(path "/root")` and `(path "/root/sheet")` |
| Gate allows net-neutral renames | ✅ | `TestDiffNetMembership::test_rename_is_neutral_after_translation`; real kicad-cli `TestNetGateWithKicadCli` |
| Gate detects & aborts on membership change (exit non-zero, no write) | ✅ | `test_dropped_node_detected`, `TestNetGateIntegration::test_gate_failure_aborts_write` (rc==2, file unchanged) |
| `--dry-run` writes nothing | ✅ | `TestRunFixAnnotationDryRun` |
| `--backup` creates `.bak` before write | ✅ | `test_backup_creates_bak_file` |
| Real components (`R1`) untouched | ✅ | e2e asserts `R1` preserved |
| kicad-cli-dependent tests skip gracefully when absent | ✅ | `@pytest.mark.skipif(find_kicad_cli() is None, ...)` |
| Empty/no power symbols → clean exit 0 | ✅ | `test_nothing_to_fix_reports_clean` |

## Test Plan

- `uv run pytest tests/test_sch_fix_annotation.py tests/test_sch_re_annotate.py tests/test_sch_repair_instances.py` → 166 passed (kicad-cli present locally, so the real net-gate integration test ran, not skipped).
- `uv run ruff check` / `ruff format` clean on all changed files.
- `uv run mypy src/kicad_tools/cli/sch_fix_annotation.py` → no issues.
- `uv run kct sch fix-annotation --help` shows the new command.

Closes #4135

### DIFF
--- a/src/kicad_tools/cli/commands/schematic.py
+++ b/src/kicad_tools/cli/commands/schematic.py
@@ -29,7 +29,7 @@ def run_sch_command(args) -> int:
             " add-label, cleanup-wires, remove-wire, insert-inline,"
         )
         print("          disconnect, reconnect-pin, move-component, remove-component, re-annotate,")
-        print("          repair-instances")
+        print("          repair-instances, fix-annotation")
         return 1
 
     schematic_path = Path(args.schematic)
@@ -629,6 +629,17 @@ def run_sch_command(args) -> int:
             per_sheet=getattr(args, "per_sheet", False),
             format=getattr(args, "format", "text"),
             unannotated_only=getattr(args, "unannotated_only", False),
+        )
+
+    elif args.sch_command == "fix-annotation":
+        from ..sch_fix_annotation import run_fix_annotation
+
+        return run_fix_annotation(
+            schematic_path=schematic_path,
+            dry_run=getattr(args, "dry_run", False),
+            backup=getattr(args, "backup", True),
+            format=getattr(args, "format", "text"),
+            skip_net_check=getattr(args, "skip_net_check", False),
         )
 
     return 1
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -1450,6 +1450,37 @@ def _add_sch_parser(subparsers) -> None:
         default="text",
     )
 
+    # sch fix-annotation
+    sch_fix_annotation = sch_subparsers.add_parser(
+        "fix-annotation",
+        help=(
+            "Hierarchy-aware power/flag annotation repair "
+            "(net-neutrality gated)"
+        ),
+    )
+    sch_fix_annotation.add_argument("schematic", help="Path to root .kicad_sch file")
+    sch_fix_annotation.add_argument(
+        "--dry-run",
+        "-n",
+        action="store_true",
+        help="Preview changes without modifying files",
+    )
+    sch_fix_annotation.add_argument(
+        "--backup",
+        action="store_true",
+        help="Create backup before modifying",
+    )
+    sch_fix_annotation.add_argument(
+        "--skip-net-check",
+        action="store_true",
+        help="Skip the net-neutrality gate (unsafe; e.g. no kicad-cli)",
+    )
+    sch_fix_annotation.add_argument(
+        "--format",
+        choices=["text", "json"],
+        default="text",
+    )
+
 
 def _add_pcb_parser(subparsers) -> None:
     """Add PCB subcommand parser with its subcommands."""
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -1453,10 +1453,7 @@ def _add_sch_parser(subparsers) -> None:
     # sch fix-annotation
     sch_fix_annotation = sch_subparsers.add_parser(
         "fix-annotation",
-        help=(
-            "Hierarchy-aware power/flag annotation repair "
-            "(net-neutrality gated)"
-        ),
+        help=("Hierarchy-aware power/flag annotation repair (net-neutrality gated)"),
     )
     sch_fix_annotation.add_argument("schematic", help="Path to root .kicad_sch file")
     sch_fix_annotation.add_argument(
--- a/src/kicad_tools/cli/sch_fix_annotation.py
+++ b/src/kicad_tools/cli/sch_fix_annotation.py
@@ -1,0 +1,685 @@
+#!/usr/bin/env python3
+"""
+Hierarchy-aware power/flag-symbol annotation repair with a net-neutrality gate.
+
+This command walks a hierarchical KiCad schematic from its root and repairs
+power/flag-symbol annotation problems that ``kicad-cli`` reports as
+``schematic has annotation errors``:
+
+- Un-numbered, net-name-styled power refs (``#GNDD``, ``#+3.3V``, ``#+5V``,
+  ``#GNDA``) that neither ``re-annotate`` nor ``repair-instances`` recognize.
+- Duplicate power refs across the flattened hierarchy (e.g. ``#GNDD`` placed
+  in several sub-sheets).
+- Inconsistent zero-padding on numbered power refs (``#PWR40`` vs ``#PWR040``).
+- Missing ``(instances)`` blocks for power symbols.
+
+Every ``power:``-lib_id symbol whose reference is not already in the canonical
+``#PWR<dd>`` / ``#FLG<dd>`` shape is assigned a globally-unique designator.
+Real-component references (``R1``, ``C3``, ``U2``) are never touched.
+
+Because renumbering power symbols must be an electrically-neutral operation
+(power symbols connect by *value*, not by designator), the command exports the
+schematic netlist before and after the repair via ``kicad-cli`` and refuses to
+write if net membership changes.  The comparison translates the *before*
+snapshot's references through the rename mapping so a renamed symbol's own
+nodes are compared apples-to-apples.
+
+Usage:
+    # Preview the repair plan (no netlist gate, no writes)
+    kicad-tools sch fix-annotation root.kicad_sch --dry-run
+
+    # Repair in place (runs the net-neutrality gate first)
+    kicad-tools sch fix-annotation root.kicad_sch
+
+    # Repair with backups
+    kicad-tools sch fix-annotation root.kicad_sch --backup
+
+    # Repair without the net-neutrality gate (unsafe; e.g. no kicad-cli)
+    kicad-tools sch fix-annotation root.kicad_sch --skip-net-check
+"""
+
+import re
+import shutil
+import sys
+import tempfile
+from pathlib import Path
+
+from kicad_tools.cli.export_netlist import Netlist, export_netlist, load_netlist
+from kicad_tools.cli.modify_schematic import create_backup
+from kicad_tools.cli.runner import find_kicad_cli
+from kicad_tools.cli.sch_re_annotate import (
+    _apply_uuid_reference_rename,
+    _detect_project_info,
+    _format_reference,
+    _parse_reference,
+)
+from kicad_tools.cli.sch_repair_instances import (
+    _find_sub_block_span,
+    _find_symbol_block,
+    _insert_project_instance,
+    _scan_symbol_children,
+)
+from kicad_tools.cli.sch_set_footprint import _collect_schematic_files
+
+# Whitespace token for regex: matches one or more tabs or spaces
+_WS = r"[ \t]+"
+
+# Canonical shapes for already-correct power/flag references: ``#PWR`` or
+# ``#FLG`` followed by two-or-more digits (KiCad zero-pads to at least two).
+_CANONICAL_POWER_RE = re.compile(r"^#(PWR|FLG)\d{2,}$")
+
+# Reference prefix chosen for a ``power:`` symbol.  Flag symbols (PWR_FLAG and
+# any lib_id containing ``FLAG``) get ``#FLG``; all other power symbols
+# (grounds and rails alike) get ``#PWR``, matching KiCad convention.
+_FLAG_PREFIX = "#FLG"
+_POWER_PREFIX = "#PWR"
+
+
+def _is_power_symbol(sym: dict) -> bool:
+    """Return True if *sym* is a KiCad power/flag symbol."""
+    return str(sym.get("lib_id", "")).startswith("power:")
+
+
+def _is_flag_symbol(sym: dict) -> bool:
+    """Return True if *sym* is a power *flag* symbol (gets ``#FLG``)."""
+    lib_id = str(sym.get("lib_id", ""))
+    return "FLAG" in lib_id.upper() or "PWR_FLAG" in lib_id.upper()
+
+
+def _power_prefix_for(sym: dict) -> str:
+    """Return the canonical reference prefix for a power/flag *sym*."""
+    return _FLAG_PREFIX if _is_flag_symbol(sym) else _POWER_PREFIX
+
+
+def _canonical_number(sym: dict) -> int | None:
+    """Return the canonical number a power/flag *sym* already holds, else None.
+
+    A reference is canonical only when it *exactly* round-trips through
+    ``_format_reference`` for the prefix appropriate to the symbol kind.  This
+    makes ``#PWR40`` canonical (number 40) but ``#PWR040`` non-canonical (that
+    number formats to ``#PWR40``, not ``#PWR040``), catching the inconsistent
+    zero-padding described in the issue.  Net-name-styled refs (``#GNDD``),
+    unpadded single digits (``#PWR2``), and family mismatches (a flag symbol
+    carrying ``#PWR01``) all return None.
+    """
+    ref = str(sym.get("reference", ""))
+    m = _CANONICAL_POWER_RE.match(ref)
+    if not m:
+        return None
+    number = int(ref[4:])
+    expected = _power_prefix_for(sym)
+    if _format_reference(expected, number) != ref:
+        return None
+    return number
+
+
+def _needs_reassignment(sym: dict) -> bool:
+    """Return True if a power/flag *sym*'s reference is not canonical."""
+    return _canonical_number(sym) is None
+
+
+def _existing_canonical_numbers(power_symbols: list[dict]) -> dict[str, set[int]]:
+    """Collect canonical numbers already in use, keyed by prefix.
+
+    Only references that are *already canonical and correctly-familied*
+    (i.e. would not be reassigned) reserve their number.  This lets the
+    assignment phase skip over numbers that established, correct symbols hold.
+    """
+    reserved: dict[str, set[int]] = {_POWER_PREFIX: set(), _FLAG_PREFIX: set()}
+    for sym in power_symbols:
+        number = _canonical_number(sym)
+        if number is None:
+            continue
+        reserved[_power_prefix_for(sym)].add(number)
+    return reserved
+
+
+def build_rename_plan(ordered_power_symbols: list[dict]) -> dict[str, dict]:
+    """Build the UUID-keyed rename plan for power/flag symbols.
+
+    *ordered_power_symbols* is the flattened, hierarchy-ordered list of
+    power/flag symbols (each dict as produced by
+    ``_extract_symbols_with_instance_info`` plus a ``uuid`` key).  Globally
+    unique ``#PWR<dd>`` / ``#FLG<dd>`` designators are assigned to every
+    symbol whose reference is not already canonical, skipping numbers held by
+    canonical symbols so no collision is introduced.
+
+    Returns a dict mapping symbol UUID -> ``{"old": str, "new": str}`` for
+    every symbol that actually changes reference (identity renames are
+    omitted).
+    """
+    reserved = _existing_canonical_numbers(ordered_power_symbols)
+    counters: dict[str, int] = {_POWER_PREFIX: 1, _FLAG_PREFIX: 1}
+
+    def _next(prefix: str) -> int:
+        n = counters[prefix]
+        while n in reserved[prefix]:
+            n += 1
+        reserved[prefix].add(n)
+        counters[prefix] = n + 1
+        return n
+
+    plan: dict[str, dict] = {}
+    for sym in ordered_power_symbols:
+        if not _needs_reassignment(sym):
+            continue
+        prefix = _power_prefix_for(sym)
+        new_ref = _format_reference(prefix, _next(prefix))
+        old_ref = str(sym["reference"])
+        if old_ref == new_ref:
+            continue
+        plan[str(sym["uuid"])] = {"old": old_ref, "new": new_ref}
+    return plan
+
+
+def _net_membership(netlist: Netlist) -> dict[frozenset[tuple[str, str]], int]:
+    """Build a value-independent net-membership snapshot.
+
+    Each net becomes a frozenset of ``(reference, pin)`` tuples; we key the
+    snapshot on that membership set (counting duplicates) rather than on the
+    net's display name/code, since renumbering can legitimately change a net's
+    name while leaving the electrical grouping intact.
+
+    Returns a multiset ``{membership_frozenset: count}``.
+    """
+    snapshot: dict[frozenset[tuple[str, str]], int] = {}
+    for net in netlist.nets:
+        membership = frozenset((node.reference, node.pin) for node in net.nodes)
+        snapshot[membership] = snapshot.get(membership, 0) + 1
+    return snapshot
+
+
+def _translate_membership(
+    snapshot: dict[frozenset[tuple[str, str]], int],
+    ref_rename: dict[str, str],
+) -> dict[frozenset[tuple[str, str]], int]:
+    """Translate a ``before`` snapshot's references through *ref_rename*.
+
+    *ref_rename* maps old reference string -> new reference string.  Every
+    ``(reference, pin)`` tuple whose reference was renamed is rewritten so the
+    ``before`` snapshot can be compared against the ``after`` snapshot without
+    spurious diffs on the renamed symbols' own nodes.
+    """
+    translated: dict[frozenset[tuple[str, str]], int] = {}
+    for membership, count in snapshot.items():
+        new_membership = frozenset((ref_rename.get(ref, ref), pin) for (ref, pin) in membership)
+        translated[new_membership] = translated.get(new_membership, 0) + count
+    return translated
+
+
+def diff_net_membership(
+    before: Netlist,
+    after: Netlist,
+    ref_rename: dict[str, str],
+) -> list[str]:
+    """Return a human-readable list of net-membership differences.
+
+    Empty list means the repair was net-neutral.  The *before* snapshot's
+    references are translated through *ref_rename* first so a renamed power
+    symbol's own nodes compare equal.
+    """
+    before_snap = _translate_membership(_net_membership(before), ref_rename)
+    after_snap = _net_membership(after)
+
+    diffs: list[str] = []
+
+    def _fmt(membership: frozenset[tuple[str, str]]) -> str:
+        return "{" + ", ".join(sorted(f"{r}.{p}" for r, p in membership)) + "}"
+
+    for membership, count in sorted(before_snap.items(), key=lambda kv: _fmt(kv[0])):
+        after_count = after_snap.get(membership, 0)
+        if after_count < count:
+            diffs.append(f"  REMOVED (x{count - after_count}): {_fmt(membership)}")
+    for membership, count in sorted(after_snap.items(), key=lambda kv: _fmt(kv[0])):
+        before_count = before_snap.get(membership, 0)
+        if before_count < count:
+            diffs.append(f"  ADDED   (x{count - before_count}): {_fmt(membership)}")
+
+    return diffs
+
+
+def _update_instance_reference(text: str, sym_uuid: str, new_ref: str) -> str:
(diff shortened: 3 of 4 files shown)


rjwalters:
## Judge review: CHANGES_REQUESTED

Thorough review done. The implementation is genuinely strong and faithfully hits every acceptance criterion from #4135 — but **CI "Lint & Format" is red**, so I can't approve as-is. The fix is one command.

### Blocker (the only one)

`ruff format --check` fails on `src/kicad_tools/cli/parser.py`. The new `fix-annotation` subparser wraps a `help=` string that ruff collapses to one line:

```
src/kicad_tools/cli/parser.py:1456-1459
-        help=(
-            "Hierarchy-aware power/flag annotation repair "
-            "(net-neutrality gated)"
-        ),
+        help=("Hierarchy-aware power/flag annotation repair (net-neutrality gated)"),
```

Fix: `uv run ruff format src/kicad_tools/cli/parser.py` and push. `mergeStateStatus` is currently `UNSTABLE`; needs to reach `CLEAN` with green CI before merge.

### Everything else verified GREEN (local, on the worktree)

- **166 passed** (`tests/test_sch_fix_annotation.py` + `test_sch_re_annotate.py` + `test_sch_repair_instances.py`), including the real `kicad-cli` net-gate integration test (`TestNetGateWithKicadCli`) — matches the PR's claim exactly.
- `ruff check` clean on all four changed files.
- `mypy src/kicad_tools/cli/sch_fix_annotation.py` — no issues.

### Correctness spot-checks (all confirmed)

- **Power-only classification**: `_is_power_symbol` keys off `lib_id.startswith("power:")`; real components (`R1`, `C3`, `U2`) are never in `build_rename_plan`. e2e asserts `R1` preserved.
- **Canonicality via `_format_reference` round-trip**: `#PWR40` → `int("40")` → `_format_reference("#PWR", 40)` == `#PWR40` (canonical, kept); `#PWR040` formats to `#PWR40` ≠ input (non-canonical, normalized). `ref[4:]` is correct since both `#PWR`/`#FLG` are 4 chars. Family mismatch (flag with `#PWR`) correctly flagged.
- **Reused helpers**: `_insert_project_instance` (keyword-only `replace_wrong_project`/`repair_loose_blocks`), `_apply_uuid_reference_rename`, `_scan_symbol_children`, `_find_symbol_block`, `_detect_project_info`, `_collect_schematic_files`, `export_netlist`/`load_netlist`/`create_backup` — all signatures match the call sites. No reimplementation of the hierarchy walker.
- **Net-neutrality gate**: before/after snapshots keyed by `(ref, pin)` membership sets; the *before* snapshot is translated through the rename mapping so a renamed symbol's own nodes compare apples-to-apples (the exact wrinkle the curated issue called out). Refuses to write with **exit 2** on any membership delta; **fails closed** when `kicad-cli` is absent unless `--skip-net-check`. Both the mocked-boundary tests and the real-cli test cover this.
- **`--dry-run`/`--backup`** match existing sch-mutator semantics; CLI wired through `parser.py` + `commands/schematic.py` (help list updated).
- **Fixtures** are synthetic multi-sheet — no chorus/local-only dependency.

Re-run `uv run pnpm check:ci` locally after the format fix to be sure, then this is an easy approve.

rjwalters:
Addressed the Judge's single blocker (CI "Lint & Format" red on `ruff format --check`).

- Ran `uv run ruff format src/kicad_tools/cli/parser.py` — collapsed the manually-wrapped `help=` string on the new `sch fix-annotation` subparser to a single line (1 file changed, 1 insertion, 4 deletions).
- `uv run ruff format --check` and `uv run ruff check` now pass on all changed files (schematic.py, parser.py, sch_fix_annotation.py, test_sch_fix_annotation.py).
- `uv run pytest tests/test_sch_fix_annotation.py -q` → 29 passed.

No functional changes. Ready for re-review.

rjwalters:

The single blocker from my prior review — the `ruff format` failure on `src/kicad_tools/cli/parser.py` — is resolved.

**Fix commit verified (8d2fa4b9):**
- Touches exactly one file: `src/kicad_tools/cli/parser.py` (1 insertion, 4 deletions)
- Purely formatting: collapses the manually-wrapped `help=` string on the `sch fix-annotation` subparser to a single line. Zero functional change.
- `git diff --name-only 8feb33d2..8d2fa4b9` confirms **no other files changed** since my last review — nothing snuck in.

**CI — all 16 checks green:**
- Lint & Format: **pass** (the previously-red check)
- Test: pass (15m21s)
- Type Check: pass
- C++ Build Check: pass
- Boards 00–07 End-to-End + all Routing/Diff-Pair/Match-Group regressions: pass
- kicad-cli Round-trip Smoke, Routed PCB DRC: pass

**Merge state:** CLEAN

Everything else was already verified green in the prior pass (166 tests passing locally, claims confirmed, scope tight, `Closes #4135`). Clean to merge.

PR #4153, opened 2026-07-14
fix(datasheet): validate PDF payload and fall through to next source

rjwalters:
## Summary

`kct datasheet download <part>` frequently received an LCSC SPA/wrapper HTML page (HTTP 200, `text/html` body), wrote it to disk as `<part>.pdf`, and reported success with exit 0 — silently poisoning the datasheet cache with HTML-as-PDF. This PR validates the payload before it is trusted and adds per-candidate fallthrough so a bad top hit no longer sinks the whole download.

## Changes

- **`datasheet/sources/base.py`** — Added `%PDF-` magic-byte validation. New module helper `_stream_to_validated_pdf()` streams the response to a temp file, validates the leading bytes against `PDF_MAGIC`, and only `os.replace()`s into place on success; a non-PDF payload raises `DatasheetDownloadError` naming the source (and Content-Type/first-bytes snippet) and leaves **no** file at `output_path`. Both `HTTPDatasheetSource.download()` (inherited by LCSC and Octopart) and the previously-unused `DatasheetSource._download_file()` helper now route through this guard.
- **`datasheet/manager.py`** — `download_by_part()` now iterates **every** `search_result.results` candidate in confidence order (previously only `results[0]`), catching `DatasheetDownloadError` per attempt and aggregating `{source: reason}`. When all candidates fail it raises a combined per-source summary rather than a single error.
- **`tests/test_datasheet.py`** — Added mock-based tests (no live network): HTML payload rejected + no file left behind (LCSC and Octopart, proving the shared guard), empty-payload rejection, `_download_file` guard, happy-path magic-byte assertion, manager fallthrough (first fails → second succeeds), all-candidates-fail aggregation (incl. two candidates from the same source), and a CLI-level `_run_download` exit-1 test asserting per-source detail on stderr.

## Acceptance Criteria Verification

| Criterion | Status | Verification |
|-----------|--------|--------------|
| `download()` (+ `_download_file`) validates `%PDF-` before success; non-PDF raises naming source, leaves no file | Done | `test_download_rejects_html_payload` (LCSC+Octopart), `test_download_rejects_empty_payload`, `test_download_file_helper_validates_pdf` |
| `download_by_part()` iterates candidates, trying next on `DatasheetDownloadError` | Done | `test_download_by_part_falls_through_to_next_candidate` |
[...]

PR #4154, opened 2026-07-14
fix(datasheet): surface install guidance when optional deps are missing

### LINKED ISSUE #4139: datasheet: missing optional dependency reported as 'No datasheet found' instead of install guidance
## Bug

With `kicad-tools` (v0.15.0) installed without the `[parts]` extra, `kct datasheet download` swallows the missing-dependency error into a lookup-miss conclusion:

```
$ kct datasheet download UCC27211 -o hardware/datasheets
Source lcsc failed: The 'requests' library is required for datasheet operations. Install with: pip install kicad-tools[parts]
Source octopart failed: The 'requests' library is required for datasheet operations. Install with: pip install kicad-tools[parts]
Failed to download datasheet: No datasheet found for 'UCC27211'
```

The final line ("No datasheet found") is wrong — nothing was searched. The per-source lines do mention the real cause, but the summary a user (or script grepping the last line) acts on is misleading.

## Suggested fix

When every source fails with the *same* `ImportError`-class cause, surface that as the primary error and exit code, e.g.:

```
Error: datasheet operations require the 'requests' library.
Install with: pip install "kicad-tools[parts]"
```

## Related nit

`kct datasheet info <valid.pdf>` fails with `PyMuPDF is required for image extraction` — `info` (page count / metadata) shouldn't require the image-extraction dependency, or the error should name the `info` operation rather than image extraction.

---

## Curator Enhancement (2026-07-14)

Verified against current `main` (commit `a8ed3b2b`, i.e. **after** PR #4153 / issue #4138 merged, which reworked `download_by_part()` to iterate all candidates and aggregate `{source: reason}` failures). The `download` error-shape from #4153 does **not** yet solve this issue — it's a distinct, earlier failure path. See root-cause trace below.

### Root cause: Part 1 (misleading "No datasheet found")

The failure happens in `search()`, before `download_by_part()`'s new aggregation logic (from #4153) is ever reached.

1. `LCSCDatasheetSource.search()` / `OctopartDatasheetSource.search()` are wrapped in `@requires_requests` (`src/kicad_tools/datasheet/sources/base.py:128-147`). When `requests` isn't installed, the decorator raises `ImportError("The 'requests' library is required for datasheet operations. Install with: pip install kicad-tools[parts]")`.
2. `DatasheetManager.search()` (`src/kicad_tools/datasheet/manager.py:106-113`) catches this with a **generic** `except Exception as e:`, logs a warning, and stashes the message in `errors[source.name] = str(e)`. The exception's *type* (ImportError vs. a normal search-miss) is discarded — only the stringified message survives in `DatasheetSearchResult.errors` (`src/kicad_tools/datasheet/models.py:70`).
3. `DatasheetManager.download_by_part()` (`src/kicad_tools/datasheet/manager.py:224-262`) calls `self.search(part_number)`, then does:
   ```python
   if not search_result.has_results:
       raise DatasheetSearchError(f"No datasheet found for '{part_number}'")
   ```
   (`manager.py:234-235`). This is **the exact site of the bug** — `search_result.errors` (which holds the two ImportError messages) is never inspected here, only `has_results`. This code path predates and is untouched by #4153's aggregation rework, which only touches the *download* loop further down (`manager.py:242-262`) — that loop is never reached when `search()` returns zero results in the first place.
4. `_run_download()` in `src/kicad_tools/cli/datasheet_cmd.py:394-418` wraps the whole call in its own local `except Exception as e: print(f"Failed to download datasheet: {e}", ...)`. **This local handler also swallows `ImportError` before it would ever reach** the outer `except ImportError` handler in `main()` (`datasheet_cmd.py:309-315`), which already prints proper install guidance for other subcommands (e.g. `info`, `extract-package`). In other words, `download`'s own `try/except` shadows the well-behaved top-level handler — that's *why* `download` behaves differently from `info`/`extract-package` today.

**Fix hook**: `DatasheetSearchResult.errors` (dict of `source -> stringified message`) is the right data already flowing through the aggregation added in #4153. The fix should detect "every entry in `errors` looks like the same missing-dependency message" at either:
- **(a)** `DatasheetManager.search()` — preserve exception *type*, not just `str(e)`, e.g. track `except ImportError` separately from other `Exception`s so `download_by_part()` can distinguish "no results because nothing found" from "no results because every source is missing a dependency", or
- **(b)** `DatasheetManager.download_by_part()` — after `if not search_result.has_results`, inspect `search_result.errors` for a uniform ImportError-shaped cause (e.g. all values contain `"is required for"` / were raised as ImportError) and raise a distinct exception (or re-raise `ImportError` directly) instead of `DatasheetSearchError("No datasheet found...")`.

Option (a) is more robust (avoids string-sniffing) — recommend catching `ImportError` explicitly in `search()`'s except clause (in addition to the current generic `except Exception`), tagging it in `errors` or a new `DatasheetSearchResult.import_errors: dict[str, str]` field, and having `download_by_part()` raise `ImportError` (or a new `DatasheetMissingDependencyError`) when `import_errors` is non-empty and covers all attempted sources. This also fixes `_run_search` (`kct datasheet search`) for the same underlying condition — worth checking `_run_search` (`datasheet_cmd.py:340-...`) prints something reasonable too, since it currently just prints "No datasheets found" without surfacing `errors` at all (see `test_search_command_no_results`, `tests/test_datasheet.py:1353-1370`, which doesn't cover the all-ImportError case).

Whichever detection site is chosen, `_run_download()`'s local `except Exception` (`datasheet_cmd.py:416-418`) needs an `except ImportError` branch *before* the generic one (mirroring the pattern already used in `main()` and in `_info`/`_extract_package`), so the exit code and message match how every other datasheet subcommand already reports missing dependencies. Don't rely solely on `main()`'s outer handler — it's currently dead code for the `download` path because `_run_download` never lets an exception propagate past its own `try/except`.

### Root cause: Part 2 (`info` and PyMuPDF)

`kct datasheet info <pdf>` is implemented in `_info()` (`src/kicad_tools/cli/datasheet_cmd.py:817-857`). It does:
```python
parser = DatasheetParser(pdf_path)
images = parser.extract_images(min_width=100, min_height=100)   # <-- forces fitz/PyMuPDF import
tables = parser.extract_tables()                                 # <-- forces pdfplumber import
```
Both are called unconditionally to build the summary counts (`image_count`, `table_count`). But **`page_count` itself also requires PyMuPDF** — `DatasheetParser.page_count` (`src/kicad_tools/datasheet/parser.py:121-131`) calls `_check_pymupdf()` then `fitz.open(...)` to get `len(doc)`. There is no lighter PDF-metadata path in this codebase today:
- No `pypdf`/`PyPDF2`/`pdfminer` dependency exists anywhere in `pyproject.toml` (checked `[project.dependencies]` and every `[project.optional-dependencies]` extra) — `fitz` (PyMuPDF) is the only PDF-page-counting capability available, and it lives behind the `datasheet` extra alongside `markitdown` and `pdfplumber`.
- `pdfplumber` (also in the `datasheet` extra, used for `extract_tables()`) *can* report page count, but requires the same extra as PyMuPDF, so swapping to it would not avoid an extra install — it would just move which of the two libraries `info` depends on, while `extract_images` still needs `fitz` regardless.

**Conclusion**: page count/metadata genuinely cannot come from a lighter dependency than what's already required — `info`'s minimal viable implementation (page count) still needs `fitz`. The real defect is scope + error naming, not a missing lightweight path:
1. `_info()` calls `extract_images()` and `extract_tables()` *unconditionally* to compute `image_count`/`table_count`, even though a caller who only wants page count doesn't need those. If `fitz` is present but the PDF triggers an image-extraction-specific failure, or if only `pdfplumber` is missing, the error will misname the failing operation as "image extraction" (from `_check_pymupdf()`'s hardcoded message, `parser.py:36-44`) even when the actual failure was in `extract_tables()`'s `pdfplumber` check, or vice versa.
2. `_check_pymupdf()`'s message text is literally `"PyMuPDF is required for image extraction."` (`parser.py:41-43`) — hardcoded regardless of which caller invoked it. `page_count` triggers the exact same message even though it has nothing to do with image extraction.

**Recommended fix** (message-scoping, not a new lighter dependency):
- Parameterize `_check_pymupdf()` (and ideally `_check_markitdown()` / `_check_pdfplumber()`) to accept a `context: str` describing the calling operation (e.g. `_check_pymupdf(context="reading PDF metadata")` vs. `_check_pymupdf(context="image extraction")`), and have `page_count` pass its own context so the error correctly says something like `"PyMuPDF is required for reading PDF page count. Install with: pip install kicad-tools[datasheet]"`.
- Additionally (recommended, keeps `info` fast/lean and matches the issue's literal ask that "`info` shouldn't require the image-extraction dependency" as far as possible): make `_info()`'s image/table counts best-effort — wrap each `extract_images()`/`extract_tables()` call in its own try/except so `info` still reports `page_count` and a friendly note (e.g. `"images: unavailable ({error})"`) instead of failing the whole command when only one of the three underlying libraries is missing. `page_count` itself will still require PyMuPDF (no way around that given current deps), but a user who *does* have PyMuPDF but not `pdfplumber` (or some future decoupling of the `datasheet` extra) should not be blocked from seeing page count.

### Scope recommendation

**Both parts in one PR.** Both are small, isolated to the `datasheet` module + its CLI wrapper, touch non-overlapping code paths (search/download error propagation vs. `info`/parser error messages), and share the same theme ("missing-dependency errors are misreported"). Splitting would create two near-trivial PRs reviewing the same file (`datasheet_cmd.py`) with no benefit. Recommend a single PR with two clearly separated commits or clearly separated diff hunks if the builder wants easy review segmentation, but one issue/one PR is appropriate here.

### Acceptance Criteria

- [ ] When every configured datasheet source fails `search()` due to a missing optional dependency (`ImportError` from `@requires_requests` or equivalent), `kct datasheet download <part>` prints an install-guidance error (naming the missing package/extra) instead of `"No datasheet found for '<part>'"`, and exits non-zero (matching the exit code convention already used by other datasheet subcommands' `ImportError` handling, i.e. `1`).
- [ ] `kct datasheet search <part>` (same underlying condition, `_run_search`) also surfaces the missing-dependency cause rather than a bare "No datasheets found" when `search_result.errors` indicates a uniform import failure — check `_run_search` while making this change since it shares `DatasheetManager.search()`.
- [ ] Mixed-failure case is preserved: if sources fail for genuinely different/non-ImportError reasons (or a real "not found" case), the existing "No datasheet found" / per-source messaging is unchanged — the new behavior must only trigger when *all* attempted sources fail with the missing-dependency shape.
- [ ] `kct datasheet info <valid.pdf>` either (a) succeeds with page count/metadata using only the dependency already required for page count (no *behavior* regression expected here since `page_count` already needs PyMuPDF — the ask is message correctness, not a new lighter path), or (b) if PyMuPDF genuinely is missing, the error message names the actual failing operation ("page count" / "reading PDF metadata" / "info"), not "image extraction".
- [ ] `_info()` computing `image_count`/`table_count` does not abort the entire `info` command if only one of `PyMuPDF`/`pdfplumber` is missing while the other succeeds (best-effort degrade), OR this is explicitly deferred with a documented rationale in the PR if the builder judges it out of scope — the core "wrong operation named in error" nit must be fixed regardless.
- [ ] No new hard dependency is added; `pypdf`/`PyPDF2` are confirmed NOT introduced as a "lighter" fix, per the analysis above (no such lighter path exists today).

### Test Plan

- [ ] Automated test: extend `tests/test_datasheet.py::TestDatasheetCLI` with a case mirroring `test_download_command_all_sources_fail` (`tests/test_datasheet.py:1372-1393`, from #4153) but for the all-ImportError case — mock `DatasheetManager.search()` (or the two source `.search()` methods, per `TestDatasheetManager.test_download_by_part_not_found` pattern at `tests/test_datasheet.py:894-900`) to return `DatasheetSearchResult(query=..., results=[], errors={"lcsc": "The 'requests' library is required...", "octopart": "The 'requests' library is required..."})`, run `main(["download", "<part>"])`, and assert the exit code is `1` and stderr contains install guidance (e.g. `"pip install"` and `"requests"` or `"[parts]"") rather than `"No datasheet found"`.
- [ ] Automated test: a mixed-failure case (one source ImportError, one source a genuine search miss/other exception) to confirm the new detection does NOT fire and legacy behavior is preserved.
- [ ] Automated test: simulate missing `requests` at the true source — `patch.dict("sys.modules", {"requests": None})` around a call into `LCSCDatasheetSource.search()` / `requires_requests`, following the existing pattern in `TestDatasheetParserDependencies` (`tests/test_datasheet.py:1294-1313`, e.g. `test_pymupdf_import_error`) — to validate the ImportError message/type actually reaches `DatasheetManager.search()`'s `errors` dict unchanged.
- [ ] Automated test: `_info()` / `kct datasheet info` — simulate missing `fitz` via `patch.dict("sys.modules", {"fitz": None})` and assert the resulting error text references page count / info / metadata rather than "image extraction". Also test the case where `fitz` is present but `pdfplumber` is missing (`patch.dict("sys.modules", {"pdfplumber": None})`) — if best-effort degrade is implemented, assert `info` still prints page count with a graceful note about tables being unavailable; if deferred, assert the error at minimum names the correct (`pdfplumber`/table extraction) dependency rather than misattributing to PyMuPDF.
- [ ] No live network calls or actual `pip uninstall` in any test — all dependency-missing scenarios must use `unittest.mock.patch.dict("sys.modules", {...: None})` or `patch.object`/`MagicMock` on `DatasheetManager`/source classes, consistent with existing tests in `tests/test_datasheet.py`.
- [ ] Manual verification (optional, not required for merge): in a venv without `[parts]` installed, run `kct datasheet download <part>` and confirm the final line names `requests`/`kicad-tools[parts]`, not "No datasheet found".
- [ ] `uv run pytest tests/test_datasheet.py -q` passes; `uv run ruff check src/kicad_tools/datasheet src/kicad_tools/cli/datasheet_cmd.py` and `uv run mypy src/kicad_tools/datasheet src/kicad_tools/cli/datasheet_cmd.py` are clean.

### Affected Files

- `src/kicad_tools/datasheet/manager.py` — `search()` (lines ~106-113, exception handling/`errors` capture) and `download_by_part()` (lines ~224-262, the `raise DatasheetSearchError("No datasheet found...")` site) are the primary fix locations for Part 1.
- `src/kicad_tools/datasheet/models.py` — `DatasheetSearchResult` (lines ~64-92) may need a new field (e.g. `import_errors`) or a helper property (e.g. `all_errors_are_import_errors`) if Option (a) from the root-cause analysis is chosen.
- `src/kicad_tools/datasheet/sources/base.py` — `requires_requests` decorator (lines ~128-147); reference only, likely unchanged, but confirms the ImportError message shape sources currently raise.
- `src/kicad_tools/cli/datasheet_cmd.py` — `_run_download()` (lines ~394-418, needs an `except ImportError` branch before its generic `except Exception`), `_run_search()` (lines ~340-..., same underlying condition), and `_info()` (lines ~817-857, PyMuPDF/pdfplumber error scoping for Part 2).
- `src/kicad_tools/datasheet/parser.py` — `_check_pymupdf()` (lines ~36-44) and `DatasheetParser.page_count` (lines ~121-131) for Part 2's error-message scoping; `_check_pdfplumber()` (lines ~47-55) if table-extraction error naming is also addressed.
- `tests/test_datasheet.py` — `TestDatasheetCLI` (download/search command tests, ~line 1321 onward) and `TestDatasheetParserDependencies` (~line 1291 onward) for new test coverage per the Test Plan above.

### Related

- PR #4153 (issue #4138) — merged just before this curation pass; reworked `download_by_part()`'s per-candidate download loop to aggregate `{source: reason}` failures. That work is adjacent but does not fix this issue: it operates on the *download* loop (after search already succeeded with results), while this issue's Part 1 bug is in the *search* step, before that loop is ever reached.

rjwalters:
## Summary

`kct datasheet download <part>` (and `search`) swallowed a missing-optional-dependency error into a misleading "No datasheet found" / "No datasheets found" conclusion when `kicad-tools` was installed without the `[parts]` extra. Separately, `kct datasheet info <valid.pdf>` reported `PyMuPDF is required for image extraction` even for page-count/metadata, and failed the whole command if only one extraction library was missing. This PR fixes both (single-PR scope per the curator's recommendation).

## Changes

**Part 1 — search/download misreporting**
- `DatasheetSearchResult` gains an `import_errors` dict and an `all_failures_are_import_errors` property to distinguish "nothing matched" from "every source is missing an optional dependency" without string-sniffing (`models.py`).
- `DatasheetManager.search()` catches `ImportError` before the generic `except Exception`, recording it in both `errors` (unchanged display behavior) and the new `import_errors` (`manager.py`).
- `download_by_part()` raises `ImportError` with collapsed install guidance instead of `DatasheetSearchError("No datasheet found...")` when *all* sources failed on a missing dependency; genuine/mixed misses are unchanged (`manager.py`).
- `_run_download()` gains an `except ImportError` branch before its generic `except Exception` (which previously shadowed `main()`'s well-behaved outer handler), and `_run_search()` surfaces the same install guidance instead of a bare "No datasheets found" (`datasheet_cmd.py`).

**Part 2 — info / PyMuPDF error scoping**
- `_check_pymupdf` / `_check_pdfplumber` / `_check_markitdown` accept a `context` argument so the `ImportError` names the actual operation; `page_count` now reports `"reading PDF page count"` rather than `"image extraction"` (`parser.py`).
- `_info()` computes image/table counts best-effort: a missing `pdfplumber` (or `fitz` for images) no longer aborts the whole command — page count still prints and unavailable counts show a friendly note (`datasheet_cmd.py`).

No new dependencies added (`pypdf`/`PyPDF2` explicitly not introduced — no lighter path exists).

| `download` prints install guidance (not "No datasheet found"), exits 1, on uniform missing-dep failure | ✅ | `test_download_command_missing_dependency`, `test_download_by_part_raises_import_error_when_all_deps_missing` |
| `search` surfaces missing-dep cause instead of "No datasheets found" | ✅ | `test_search_command_missing_dependency` |
| Mixed-failure case preserved (legacy "No datasheet found" retained) | ✅ | `test_download_by_part_mixed_failure_keeps_search_error`, `test_search_mixed_import_and_search_failure`, `test_all_failures_are_import_errors_mixed` |
| `info` error names the real operation ("page count"), not "image extraction" | ✅ | `test_info_command_missing_pymupdf_names_page_count`, `test_page_count_import_error_names_page_count`, `test_pymupdf_import_error_names_context` |
| `_info()` best-effort: one missing extraction lib does not abort the command | ✅ | `test_info_command_table_extraction_best_effort` |
| No new hard dependency added | ✅ | Only message-scoping + exception-type tracking; no `pyproject.toml` change |

## Test Plan

- `uv run pytest tests/test_datasheet.py -q` → 126 passed, 29 skipped (skips are pre-existing dependency-gated tests; the `TestDatasheetManager` suite is class-skipped when `requests` is absent in this env — its logic was additionally verified with a stubbed `requests` module).
- `uv run ruff check` + `ruff format` clean on all changed files.
- `mypy` on the four changed source files: 0 new errors (the 17 reported are all pre-existing, on lines untouched by this PR).
- All dependency-missing scenarios use `patch.dict("sys.modules", {...: None})` / `DatasheetSearchResult(import_errors={...})` — no live network, no `pip uninstall`.

Closes #4139

### DIFF
--- a/src/kicad_tools/cli/datasheet_cmd.py
+++ b/src/kicad_tools/cli/datasheet_cmd.py
@@ -363,6 +363,19 @@ def _run_search(args) -> int:
         print(json.dumps(output, indent=2))
     else:
         if not results.has_results:
+            # When every source failed because an optional dependency is
+            # missing, the real problem is a missing install, not a lookup
+            # miss — surface install guidance rather than "no datasheets found"
+            # (issue #4139).
+            if results.all_failures_are_import_errors:
+                message = next(iter(set(results.import_errors.values())), "")
+                print(f"Error: {message}", file=sys.stderr)
+                print(
+                    "Install with: pip install kicad-tools[parts]",
+                    file=sys.stderr,
+                )
+                return 1
+
             print(f"No datasheets found for '{args.part}'")
             if results.errors:
                 print("\nSource errors:")
@@ -413,6 +426,17 @@ def _run_download(args) -> int:
 
         return 0
 
+    except ImportError as e:
+        # Every source failed because an optional dependency is missing.
+        # Surface install guidance instead of a misleading "no datasheet found",
+        # mirroring how main()/_info()/_extract_package() already report missing
+        # dependencies (issue #4139).
+        print(f"Error: {e}", file=sys.stderr)
+        print(
+            "Install with: pip install kicad-tools[parts]",
+            file=sys.stderr,
+        )
+        return 1
     except Exception as e:
         print(f"Failed to download datasheet: {e}", file=sys.stderr)
         return 1
@@ -831,24 +855,56 @@ def _info(args) -> int:
     try:
         parser = DatasheetParser(pdf_path)
 
-        # Count images and tables for quick summary
-        images = parser.extract_images(min_width=100, min_height=100)
-        tables = parser.extract_tables()
+        # page_count is the minimal viable info output; it genuinely requires
+        # PyMuPDF (no lighter PDF-metadata path exists in this codebase). If
+        # that dependency is missing the error now names the actual operation
+        # ("reading PDF page count") rather than "image extraction" (#4139).
+        page_count = parser.page_count
+
+        # Image and table counts are best-effort: a caller who only wants page
+        # count/metadata should not have the whole command fail because *one*
+        # of the optional extraction libraries (fitz / pdfplumber) is missing.
+        image_count: int | None
+        table_count: int | None
+        image_error: str | None = None
+        table_error: str | None = None
+
+        try:
+            image_count = len(parser.extract_images(min_width=100, min_height=100))
+        except Exception as e:
+            image_count = None
+            image_error = str(e)
+
+        try:
+            table_count = len(parser.extract_tables())
+        except Exception as e:
+            table_count = None
+            table_error = str(e)
 
         if args.format == "json":
-            data = {
+            data: dict = {
                 "path": str(pdf_path),
                 "filename": pdf_path.name,
-                "page_count": parser.page_count,
-                "image_count": len(images),
-                "table_count": len(tables),
+                "page_count": page_count,
+                "image_count": image_count,
+                "table_count": table_count,
             }
+            if image_error is not None:
+                data["image_error"] = image_error
+            if table_error is not None:
+                data["table_error"] = table_error
             print(json.dumps(data, indent=2))
         else:
             print(f"File:    {pdf_path.name}")
-            print(f"Pages:   {parser.page_count}")
-            print(f"Images:  {len(images)} (>= 100x100 px)")
-            print(f"Tables:  {len(tables)}")
+            print(f"Pages:   {page_count}")
+            if image_count is not None:
+                print(f"Images:  {image_count} (>= 100x100 px)")
+            else:
+                print(f"Images:  unavailable ({image_error})")
+            if table_count is not None:
+                print(f"Tables:  {table_count}")
+            else:
+                print(f"Tables:  unavailable ({table_error})")
 
     except Exception as e:
         print(f"Error: {e}", file=sys.stderr)
--- a/src/kicad_tools/datasheet/manager.py
+++ b/src/kicad_tools/datasheet/manager.py
@@ -24,6 +24,22 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _missing_dependency_message(import_errors: dict[str, str]) -> str:
+    """Build a single install-guidance message from per-source ImportErrors.
+
+    Every configured datasheet source failed because an optional dependency is
+    missing. The per-source messages are effectively identical (they come from
+    the same ``@requires_requests`` guard), so collapse them into one clear
+    line when they agree, and fall back to listing each source otherwise.
+    """
+    unique = set(import_errors.values())
+    if len(unique) == 1:
+        # All sources reported the same missing dependency.
+        return next(iter(unique))
+    detail = "; ".join(f"{source}: {msg}" for source, msg in import_errors.items())
+    return f"Datasheet operations require an optional dependency that is not installed: {detail}"
+
+
 class DatasheetManager:
     """
     Unified manager for datasheet search and download.
@@ -102,12 +118,21 @@ class DatasheetManager:
         """
         all_results = []
         errors = {}
+        import_errors: dict[str, str] = {}
 
         for source in self.sources:
             try:
                 results = source.search(part_number)
                 all_results.extend(results)
                 logger.debug(f"Source {source.name}: found {len(results)} results")
+            except ImportError as e:
+                # A missing optional dependency (e.g. ``requests`` guarded by
+                # ``@requires_requests``). Track separately from ordinary
+                # search failures so the caller can surface install guidance
+                # instead of a misleading "no datasheet found".
+                logger.warning(f"Source {source.name} failed: {e}")
+                errors[source.name] = str(e)
+                import_errors[source.name] = str(e)
             except Exception as e:
                 logger.warning(f"Source {source.name} failed: {e}")
                 errors[source.name] = str(e)
@@ -127,6 +152,7 @@ class DatasheetManager:
             query=part_number,
             results=unique_results,
             errors=errors,
+            import_errors=import_errors,
         )
 
     def download(
@@ -232,6 +258,13 @@ class DatasheetManager:
         search_result = self.search(part_number)
 
         if not search_result.has_results:
+            # Distinguish "no results because nothing matched" from "no results
+            # because every configured source is missing an optional dependency".
+            # In the latter case the real problem is a missing install, not a
+            # lookup miss, so surface install guidance instead of the misleading
+            # "No datasheet found" (issue #4139).
+            if search_result.all_failures_are_import_errors:
+                raise ImportError(_missing_dependency_message(search_result.import_errors))
             raise DatasheetSearchError(f"No datasheet found for '{part_number}'")
 
         # Try each candidate in confidence order, falling through to the next
--- a/src/kicad_tools/datasheet/models.py
+++ b/src/kicad_tools/datasheet/models.py
@@ -68,11 +68,27 @@ class DatasheetSearchResult:
     query: str
     results: list[DatasheetResult] = field(default_factory=list)
     errors: dict[str, str] = field(default_factory=dict)  # source -> error message
+    # Subset of ``errors`` whose failures were caused by a missing optional
+    # dependency (an ``ImportError`` raised by a source's ``@requires_requests``
+    # guard). Tracked separately so callers can distinguish "no results because
+    # nothing matched" from "no results because every source is missing a
+    # dependency" without string-sniffing ``errors``.
+    import_errors: dict[str, str] = field(default_factory=dict)  # source -> message
 
     @property
     def has_results(self) -> bool:
         """Check if any results were found."""
         return len(self.results) > 0
+
+    @property
+    def all_failures_are_import_errors(self) -> bool:
+        """True when at least one source failed and every failure was an ImportError.
+
+        Used to detect the "no configured datasheet source could even run
+        because an optional dependency is missing" condition, so the CLI can
+        surface install guidance instead of a misleading "no datasheet found".
+        """
+        return bool(self.errors) and set(self.import_errors) == set(self.errors)
 
     @property
     def sources_searched(self) -> list[str]:
--- a/src/kicad_tools/datasheet/parser.py
+++ b/src/kicad_tools/datasheet/parser.py
@@ -22,35 +22,50 @@ if TYPE_CHECKING:
     from fitz import Page as FitzPage
 
 
-def _check_markitdown() -> None:
-    """Check if markitdown is available."""
+def _check_markitdown(context: str = "PDF to markdown conversion") -> None:
+    """Check if markitdown is available.
+
+    Args:
+        context: Name of the operation requiring the dependency, used to make
+            the error message name the actual failing operation.
+    """
     try:
         import markitdown  # noqa: F401
     except ImportError as e:
         raise ImportError(
-            "markitdown is required for PDF to markdown conversion. "
+            f"markitdown is required for {context}. "
             "Install with: pip install kicad-tools[datasheet]"
         ) from e
 
 
-def _check_pymupdf() -> None:
-    """Check if PyMuPDF is available."""
+def _check_pymupdf(context: str = "image extraction") -> None:
+    """Check if PyMuPDF is available.
+
+    Args:
+        context: Name of the operation requiring the dependency, used to make
+            the error message name the actual failing operation (e.g. "reading
+            PDF page count") rather than always saying "image extraction".
+    """
     try:
         import fitz  # noqa: F401
     except ImportError as e:
         raise ImportError(
-            "PyMuPDF is required for image extraction. "
-            "Install with: pip install kicad-tools[datasheet]"
+            f"PyMuPDF is required for {context}. Install with: pip install kicad-tools[datasheet]"
         ) from e
 
 
-def _check_pdfplumber() -> None:
-    """Check if pdfplumber is available."""
+def _check_pdfplumber(context: str = "table extraction") -> None:
+    """Check if pdfplumber is available.
+
+    Args:
+        context: Name of the operation requiring the dependency, used to make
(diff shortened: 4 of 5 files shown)


rjwalters:
## Judge verdict: APPROVED

Reviewed for correctness, code quality, test coverage, and adherence to the curated acceptance criteria on issue #4139. CI is fully green (16/16) and mergeStateStatus is CLEAN.

### Verified

**Part 1 — search/download misreporting**
- `DatasheetSearchResult.import_errors` field + `all_failures_are_import_errors` property use typed detection (`set(import_errors) == set(errors)`), no string-sniffing. The property correctly fires only when at least one source failed and *every* failure was an ImportError.
- `manager.search()` catches `ImportError` before the generic `except Exception`, recording into both `errors` (unchanged display) and `import_errors`.
- `download_by_part()` raises `ImportError` with collapsed install guidance only when all sources failed on a missing dependency; genuine/mixed misses still raise the legacy `DatasheetSearchError`. The #4153 candidate-iteration loop and `{source: reason}` aggregation are fully preserved — the new guard is inserted cleanly *before* the loop, no clobbering.
- `_run_download()`'s `except ImportError` branch is correctly ordered before `except Exception`, fixing the shadowing of `main()`'s outer handler. `_run_search()` surfaces the same guidance inside the `not has_results` branch.

**Part 2 — info / PyMuPDF scoping**
- `_check_pymupdf`/`_check_pdfplumber`/`_check_markitdown` take a `context` arg; `page_count` now reports "reading PDF page count" instead of "image extraction".
- `_info()` computes image/table counts best-effort — a missing `pdfplumber`/`fitz` no longer aborts the command; page count still prints and unavailable counts show a friendly note.
- No new dependency added.

### Local test run (this repo has requests injectable)
- `uv run pytest tests/test_datasheet.py -q` → **126 passed, 29 skipped** (skips are dependency-gated).
- Critically, the four `TestDatasheetManager` tests that skip when `requests` is absent were executed via `uv run --with requests pytest tests/test_datasheet.py::TestDatasheetManager` → **16/16 passed**, including `test_search_captures_import_errors_separately`, `test_search_mixed_import_and_search_failure`, `test_download_by_part_raises_import_error_when_all_deps_missing`, and `test_download_by_part_mixed_failure_keeps_search_error`. The manager-level logic is confirmed correct, not merely relying on CI.
- Ruff check + format clean on all changed files.
- mypy: 17 errors, all pre-existing on lines untouched by this PR (Type Check CI green confirms no new errors). The `data: dict` annotation added in `_info` is appropriate defensive typing for the new optional keys.

All curated acceptance criteria met. Clean, well-commented, well-tested. Approving.

PR #4157, opened 2026-07-14
fix(schematic): union collinear-overlapping and T-touching wires in connectivity

### LINKED ISSUE #4143: schematic: connectivity semantics diverge from KiCad — collinear overlapping wires silently merge nets, invisible to extract_netlist/validate
## Bug (high severity — LVS-grade)

kicad-tools' schematic connectivity model only unions wires at **endpoints**, while KiCad unions **any touching/overlapping collinear segments**. A generated schematic whose label-stub wires overlap mid-segment therefore has different netlists in kicad-tools vs KiCad — and the kicad-tools side is the wrong one.

## Real-world impact (softstart rev-B bring-up, 2026-07-13)

A generator using the standard stub-wire+label idiom (place symbol → short wire from pin → `add_label`) produced four **silent net merges** when stubs from neighboring parts overlapped collinearly:

- `+3.3V ↔ GND` (a literal power short — interleaved decoupling-cap stubs)
- `GND ↔ GATE_NEG_B` (a gate-drive input shorted to ground)
- `SCAP_NEG ↔ V_RSV_POS` (85 V bank net into a 12 V floating reservoir)
- `SRC_NEG ↔ V_RSV_NEG`

All four passed:
- `sch.validate()` — no errors (only emit-time `UserWarning`s about endpoints landing on wires, easily lost in output noise)
- a netlist-assertion gate built on `sch.extract_netlist()` — **75/75 pin→net checks passed while the board had a 3.3V-GND short**, because `extract_netlist` doesn't see mid-segment unions

They were only caught by `kicad-cli sch erc` (`multiple_net_names` / unexpected `pin_to_pin`, both **warning** severity) plus a hand-rolled collinear-overlap detector.

## Suggested fixes (in order of value)

1. **Make `extract_netlist()` match KiCad connectivity**: union nets on collinear overlap and mid-segment T-touches, not just shared endpoints. This is the correctness fix — everything downstream (LVS, validation, agents' self-checks) inherits it.
2. `validate()` should emit an **error** (not a warning) when two wire segments carrying different labeled nets overlap collinearly.
3. Optional ergonomics: a built-in `add_stub_label(symbol, pin, net)` with collision-aware stub length/direction, since the stub idiom is ubiquitous in generated schematics (the rev-A softstart generator hand-rolls exactly this, with a comment noting it exists because long rail taps were "silently merging nets" — same root cause).

## Repro sketch

```python
sch = Schematic(title="t")
r1 = sch.add_symbol("Device:R", x=100, y=100, ref="R1", rotation=90)  # horizontal
r2 = sch.add_symbol("Device:R", x=108, y=100, ref="R2", rotation=90)  # 8mm away
# stub+label pin2 of R1 as NET_A (reaches x≈109) and pin1 of R2 as NET_B
# (reaches x≈103): overlapping segment on y=100.
# extract_netlist(): NET_A and NET_B distinct.  KiCad ERC: merged.
```

Context: kicad-tools 0.15.0; found while generating the softstart rev-B schematic (rjwalters/softstart PR #28, commit 9304f01 — the generator now runs kicad-cli ERC with `multiple_net_names`/`pin_to_pin` fatal at any severity as a workaround gate).

---

### Root Cause (confirmed by inspection + repro)

The bug is in `SchematicNetlistMixin._build_connectivity_graph()`:

- `src/kicad_tools/schematic/models/netlist_mixin.py:89-94` — wires are unioned **only** at their two endpoints:
  ```python
  for wire in self.wires:
      p1 = (round(wire.x1, 2), round(wire.y1, 2))
      p2 = (round(wire.x2, 2), round(wire.y2, 2))
      wire_segments.append((p1, p2))
      union(p1, p2)   # <-- only unions a wire's own two ends; never unions
                       #     wire A's geometry with wire B's geometry
  ```
- `connect_to_wire()` (same file, lines 108-129) unions *points* (pins, junctions, labels) onto wires — with an optional mid-segment T-touch (`point_on_segment`) — but it is **never called wire-vs-wire**. There is no code path anywhere in `_build_connectivity_graph()` that unions two `Wire` segments with each other when they overlap collinearly or when one's endpoint lands on the other's interior.
- `validate_power_nets()` (`validation_mixin.py:499-706`) duplicates the same endpoint-only wire-to-wire union logic independently — it has the identical gap.

I confirmed this empirically against current `main` with a minimal repro (two collinear, non-endpoint-sharing wires on `y=100`, overlapping on `x∈[103,109]`, each carrying a label):

```python
sch.wires.append(Wire(x1=100, y1=100, x2=109, y2=100))   # NET_A stub
sch.wires.append(Wire(x1=103, y1=100, x2=112, y2=100))   # NET_B stub, overlaps [103,109]
sch.labels.append(Label(text="NET_A", x=102, y=100))
sch.labels.append(Label(text="NET_B", x=110, y=100))

parent, *_ = sch._build_connectivity_graph()
# find(NET_A label point) != find(NET_B label point)  -->  NOT merged (wrong; KiCad merges these)
```
Result: the two labeled points resolve to **different** Union-Find roots — reproducing the issue's claim that `extract_netlist()` diverges from KiCad ERC.

Note this is orthogonal to (and does not regress) the pin-endpoint-only semantics fixed in #4020/PR #4003 (`tests/test_netlist_query.py::TestPinEndpointOnlyWireAttach`) — that fix correctly made **pins** attach to wires only at endpoints/junctions (a wire crossing a pin mid-span must NOT connect). This issue is the mirror-image gap: **wire-to-wire** unions need collinear-overlap and T-touch handling that pin-to-wire unions deliberately do not get. The two fixes must not be conflated — a naive "make everything mid-span-sensitive again" change would reintroduce the board-05 phantom-short regression that #4003 fixed.

### Existing reusable geometry: `_is_collinear_overlap`

`src/kicad_tools/cli/sch_cleanup_wires.py:131-176` already implements a collinear-overlap test (`_is_collinear_overlap`), used by `kct sch cleanup-wires` to detect and remove redundant overlapping wires. It is a good starting point but **not sufficient as-is**:

- It requires **seg2 to be fully enclosed within seg1** (`t_min >= 0` and `t_max <= 1`, i.e. one segment strictly contains the other). The union semantics needed here must also handle **partial overlap** — two segments that share a sub-range without either fully containing the other (exactly the softstart repro geometry: stub A covers `[100,109]`, stub B covers `[103,112]`; neither contains the other, but they share `[103,109]`).
- It does not need "how much" overlap for netlist purposes — the union rule only needs "is there a shared sub-segment (or endpoint-on-interior touch) at all", which is a slightly relaxed version of the same math (drop the "strictly enclosed" requirement; check that the projected-parameter ranges `[t_min1, t_max1]` and `[t_min2, t_max2]` intersect after confirming collinearity).
- Also reusable/related: `SchematicElementsMixin._point_on_wire` and `._point_on_wire_segment_interior` (`elements_mixin.py:36-111`) already implement perpendicular-distance point-on-segment and interior-only tests, used by `check_wire_collisions()` (`validation_mixin.py:722-792`) for **T-touch** detection (a wire's endpoint landing on another wire's interior) — but `check_wire_collisions()` is a standalone diagnostic **not wired into `validate()` or `_build_connectivity_graph()`**. It only reports collisions as data for callers to inspect manually; it never unions and never gates.
- PCB-side segment-intersection helpers (`src/kicad_tools/core/geometry.py::segments_intersect`, `segment_to_segment_distance`) solve a related but distinct problem (general 2D segment intersection/clearance for copper routing) and are **not schematic-orthogonal-grid-aware** in the same way; recommend NOT importing them into the schematic model — reuse/relax `_is_collinear_overlap` instead, since schematic wires are always axis-aligned (horizontal/vertical) by KiCad convention, which simplifies the math versus the general PCB case.

### Union semantics to implement (precise definitions)

For two wire segments A and B (each schematic wires are axis-aligned in practice, though the fix should not assume this if avoidable):

1. **Exact-endpoint** (current, correct, keep as-is): `A.p1 == B.p1` (etc.) within point tolerance.
2. **Collinear overlap** (missing — this issue, item 1): A and B lie on the same infinite line (perpendicular distance from B's endpoints to A's line ≈ 0) AND their projected parametric ranges on that line overlap by more than a point (i.e. share a sub-segment of nonzero length, not just a shared endpoint — that's case 1). This must handle: A fully containing B, B fully containing A, and partial overlap (neither contains the other).
3. **Mid-segment T-touch** (missing — this issue, item 1): one endpoint of B lands in the **interior** of A (not at A's endpoints) — reuse `_point_on_wire_segment_interior` semantics, but the result must **union**, not just report (unlike `check_wire_collisions()` today).

**Tolerance**: use the same `POINT_TOLERANCE = 0.1`mm class constant already defined in `SchematicElementsMixin` (`elements_mixin.py:34`) for perpendicular-distance-to-line checks, consistent with the existing point-matching tolerance used elsewhere in the connectivity graph. Do not introduce a second, differently-tuned epsilon — `sch_cleanup_wires.py` uses a much tighter `0.005`mm because it's deciding whether to delete a wire (destructive), whereas connectivity union is non-destructive and should be at least as permissive as KiCad's own snap tolerance. Round coordinates to the same 2-decimal-place (0.01mm) bucket already used throughout `_build_connectivity_graph()` for consistency with the existing Union-Find point keys.

### Performance

Naive all-pairs wire-vs-wire testing is O(n²) in wire count. For typical generated schematics (tens to low hundreds of wires per sheet — the softstart/chorus/board-0X fixtures top out around a few hundred wires per sheet), O(n²) is negligible (sub-millisecond to low-millisecond). `check_wire_collisions()` already does an all-pairs O(n²) scan today with no reported performance complaints. **Recommendation: implement the simplest all-pairs version first; do not add a sweep-line or spatial index in this PR.** If a future large-hierarchy or bulk-lint use case demonstrates a real slowdown, that's a separate, narrowly-scoped follow-up (bucket wires by rounded Y for horizontal / X for vertical before doing the O(n²) comparison only within same-axis buckets — cheap incremental win if ever needed, not required now).

### Scope recommendation

**One PR for items (1) + (2).** They share the same root-cause fix (collinear-overlap + T-touch union in the connectivity graph) and the same test fixtures — splitting them would mean touching `_build_connectivity_graph()` twice and re-deriving the same geometry helper twice. Concretely:

- (1) Fix `_build_connectivity_graph()` (and its `validate_power_nets()` duplicate — same gap, same fix, should be updated together or refactored to share the connectivity graph builder) to union on collinear overlap and T-touch.
- (2) Add a new `validate()` check: after connectivity is built, for any pair of wire segments that overlap collinearly (or T-touch) **and** are independently associated with two *different* named nets (labels/power symbols) that do NOT end up unioned to the same net-name set post-fix — wait, once (1) is fixed, overlapping wires with different labels *will* be unioned into one net, so the interesting case for (2) is actually: **flag it as an ERROR that two differently-named labels ended up on the same net via geometric overlap**, since that's almost always an unintentional short (the softstart case). This should reuse `_check_wire_connectivity`'s existing pattern (issue dict with `severity`, `type`, `message`, `location`) — add a new issue `type` such as `"collinear_net_conflict"` with `severity: "error"`, distinct from the existing `"missing_junction"` warning (which is about explicit T-junction dots, not net identity).

**Item (3) (`add_stub_label()` ergonomics helper) should be a separate follow-up issue.** It's additive API surface, not a correctness fix, and is independently valuable/schedulable — bundling it would inflate this PR's diff and review surface for no coupling benefit (nothing about `add_stub_label`'s design depends on the connectivity fix, and vice versa). Suggest filing it as a new issue referencing this one once (1)+(2) land, since `add_stub_label` could optionally use `check_wire_collisions()`-style detection at draw-time to warn/prevent generators from creating the overlap in the first place — a nice synergy but not a blocking dependency.

### Acceptance Criteria

- [ ] `extract_netlist()` unions two wire segments when they overlap collinearly (partial overlap, not just full containment), matching KiCad ERC.
- [ ] `extract_netlist()` unions two wire segments when one's endpoint T-touches the other's interior (mirrors `check_wire_collisions()`'s detection, but performs the union rather than just reporting).
- [ ] `are_connected()` and `get_net_for_pin()` (which both call `_build_connectivity_graph()`) reflect the same fix automatically (no separate patch needed — verify with a test).
- [ ] `validate_power_nets()`'s independent wire-union logic gets the same fix (either directly or by refactoring it to reuse `_build_connectivity_graph()`).
- [ ] `validate()` emits a **new error-severity issue** (not the existing `missing_junction` warning) when two wire segments carrying different labeled/power net names overlap collinearly or T-touch, resulting in a net-identity conflict.
- [ ] The issue's own repro sketch (`R1`/`R2` collinear overlapping stubs, ~8mm apart) is captured as a regression test.
- [ ] No regression in `tests/test_netlist_query.py::TestPinEndpointOnlyWireAttach` — pin-to-wire mid-span semantics from #4020/#4003 must remain endpoint-only; only wire-to-wire union semantics change.
- [ ] No regression in the existing `tests/test_netlist_query.py`, `tests/test_validate_netlist.py`, `tests/test_connectivity.py`, `tests/test_hierarchical_netlist.py` suites.

### Affected Files

- `src/kicad_tools/schematic/models/netlist_mixin.py` — `_build_connectivity_graph()` (lines ~87-211): add wire-vs-wire collinear-overlap and T-touch union logic alongside the existing endpoint union.
- `src/kicad_tools/schematic/models/validation_mixin.py` — `validate()` (lines ~52-156): add the new cross-net collinear-overlap error check; `validate_power_nets()` (lines ~474-706): apply the same wire-union fix (or refactor to share `_build_connectivity_graph()`).
- `src/kicad_tools/schematic/models/elements_mixin.py` — `_point_on_wire`, `_point_on_wire_segment_interior` (lines ~36-129): candidate reuse for T-touch detection; `POINT_TOLERANCE` constant (line 34) is the tolerance to standardize on.
- `src/kicad_tools/cli/sch_cleanup_wires.py` — `_is_collinear_overlap()` (lines ~131-176): candidate reuse/relaxation for the collinear-overlap test (currently requires full containment; needs relaxing to detect partial overlap too). Consider extracting a shared geometry helper (e.g. into `elements_mixin.py` or a new small module) so both `sch_cleanup_wires.py` and the netlist mixin call the same tested primitive instead of drifting.
- `tests/test_netlist_query.py` — add new test class(es) for wire-vs-wire collinear overlap and T-touch union behavior; extend `TestExtractNetlist`/`TestAreConnected`.
- `tests/test_validate_netlist.py` or `tests/test_schematic_models.py` — add tests for the new `validate()` error on cross-net collinear overlap (check whichever file already covers `validate()`'s issue-dict output; `test_schematic_models.py` also references `check_wire_collisions`).

### Test Plan

- [ ] **Repro sketch as a test**: two symbols ~8mm apart, collinear overlapping stub wires each carrying a different label; assert `extract_netlist()` merges both labels into one net (post-fix), and assert `validate()` emits the new error-severity issue.
- [ ] **Partial overlap (neither contains the other)**: wire A `[100,109]`, wire B `[103,112]` on the same line — assert union occurs (this is the case `_is_collinear_overlap` as currently written does NOT catch, since neither fully encloses the other).
- [ ] **Full containment**: wire A `[100,112]`, wire B `[103,109]` fully inside A — assert union occurs (already covered by `_is_collinear_overlap`; confirm the netlist-side reuse doesn't regress this case).
- [ ] **T-touch case**: wire A's endpoint lands in the interior of wire B (not at B's endpoints) — assert union occurs; cross-check against `check_wire_collisions()`'s existing detection of the same geometry (same point set should be flagged/unioned).
- [ ] **Four softstart merge patterns as synthetic fixtures** (softstart repo is local-only — do NOT depend on it; construct minimal synthetic `Schematic` objects reproducing the same *geometric class* of bug for each): (a) interleaved decoupling-cap stubs causing `+3.3V`/`GND`-style power-vs-power overlap; (b) a gate-drive input stub overlapping a ground stub; (c)/(d) two power-rail stubs of different named nets overlapping when routed at the same Y (or X) coordinate with different X (or Y) extents. Each should assert both the netlist merge (post-fix, correct) and the new `validate()` error (if item 2's cross-net-conflict heuristic is intended to still flag these as suspicious even after they're correctly merged — see scope note above; clarify during implementation whether "now correctly merged AND flagged as an error" or "merged silently" is the desired end state, since a real design might legitimately overlap same-net stubs).
- [ ] **No false positive on same-net overlap**: two wires overlapping collinearly that carry the *same* label/net name should merge (uneventfully) without triggering the new error — only *different*-net overlaps are errors.
- [ ] **Automated tests**: extend `tests/test_netlist_query.py` (primary), `tests/test_validate_netlist.py` and/or `tests/test_schematic_models.py` (validate() error), run `uv run pytest tests/test_netlist_query.py tests/test_validate_netlist.py tests/test_schematic_models.py tests/test_connectivity.py tests/test_hierarchical_netlist.py -v`.
- [ ] **Full regression**: `uv run pytest` (or at minimum the schematic/netlist/validate subset) to confirm no existing behavior (especially #4020/#4003 pin-endpoint-only semantics) regresses.
- [ ] **Manual verification**: run `kct sch validate` (or equivalent CLI) against a hand-built `.kicad_sch` fixture with the repro geometry and confirm the new error surfaces in CLI output, not just the Python API.

### Out of Scope (this PR)

- `add_stub_label()` ergonomics helper (item 3) — recommend filing as a separate follow-up issue once this lands.
- Sweep-line / spatial-index performance optimization — not warranted at current schematic sizes; simplest O(n²) approach is sufficient.
- Wiring `check_wire_collisions()` into `validate()` as its own separate warning-vs-error reclassification is adjacent but not required — the new check in this issue is net-identity-aware (differently-named nets), whereas `check_wire_collisions()` is purely geometric (any endpoint-on-interior touch, regardless of net names). Keep them distinct; do not conflate.

rjwalters:
## Summary

kicad-tools' schematic connectivity model unioned wires only at **exact shared endpoints**, while KiCad unions the nets of **any touching or overlapping collinear wire segments**. A generated schematic whose stub-wire+label geometry overlapped mid-segment therefore had a different netlist in kicad-tools than in KiCad — and `extract_netlist()`/`validate()` reported the *wrong* (unmerged) one, silently hiding LVS-grade net shorts (the softstart rev-B `+3.3V`/`GND` short passed 75/75 pin→net checks while the board was shorted).

This fixes items (1) `extract_netlist()` correctness and (2) `validate()` error emission from #4143. Item (3) `add_stub_label()` is deferred to a separate follow-up per the curation scope note.

## Changes

- **New `schematic/models/wire_geometry.py`** — a single, tested geometry primitive `wire_segments_connect()` (plus `_collinear_overlap` / `_point_on_segment_interior`) so both connectivity builders share one tested rule and cannot drift. Handles collinear overlap (partial *and* full containment, relaxing the full-containment-only `_is_collinear_overlap`) and mid-segment T-touch. A bare endpoint touch returns `False` (already handled by endpoint Union-Find); a bare crossing without a junction returns `False` (KiCad semantics). Tolerance = `POINT_TOLERANCE` (0.1mm), consistent with the connectivity graph.
- **`_build_connectivity_graph()`** — added an all-pairs O(n²) wire-vs-wire union alongside the existing endpoint union. Pin-to-wire attachment stays endpoint-only (#4020/#4003) — only wire-to-wire semantics change.
- **`validate_power_nets()`** — same wire-union fix applied to its duplicate connectivity builder (it had the identical gap).
- **`validate()`** — new error-severity `collinear_net_conflict` issue when two wire segments carrying **different** labeled/power net names overlap collinearly or T-touch (distinct from the `missing_junction` warning). Same-net or one-unlabeled overlaps are not flagged.

| `extract_netlist()` unions collinear overlap (partial, not just containment) | ✅ | `TestWireToWireCollinearUnion::test_partial_overlap_neither_contains_the_other` |
| `extract_netlist()` unions mid-segment T-touch | ✅ | `test_mid_segment_t_touch` |
| `are_connected()` / `get_net_for_pin()` inherit the fix | ✅ | `test_are_connected_and_get_net_for_pin_reflect_fix` |
| `validate_power_nets()` gets the same fix | ✅ | Shared `wire_segments_connect` union added to its builder |
| `validate()` emits a new **error** on cross-net overlap | ✅ | `TestCollinearNetConflict` (6 cases incl. power-vs-power) |
| Issue's repro sketch captured as a regression test | ✅ | `test_repro_sketch_r1_r2_collinear_stubs` + manual smoke test |
| No regression in `TestPinEndpointOnlyWireAttach` | ✅ | 5/5 pass unchanged |
| No regression in netlist/validate/connectivity/hierarchical suites | ✅ | 367 pass across the schematic subset; 3309 pass on the broad keyword-filtered run |

## Test Plan

- New: `tests/test_wire_geometry.py` (geometry primitive), `TestWireToWireCollinearUnion` in `test_netlist_query.py` (partial/full/T-touch/same-endpoint/parallel/power-vs-power/same-net synthetic softstart classes), `TestCollinearNetConflict` in `test_schematic_models.py` (validate() error cases).
- `uv run pytest tests/test_wire_geometry.py tests/test_netlist_query.py tests/test_schematic_models.py tests/test_validate_netlist.py tests/test_connectivity.py tests/test_hierarchical_netlist.py` → **367 passed**.
- Broad regression `-k "netlist or connectivity or validate or wire or collision or schematic"` → **3309 passed, 27 skipped**.
- `ruff check` + `ruff format` clean on changed files; `scripts/ci/check_mypy_baseline.py` → **0 new mypy errors**.
- Manual API smoke test of the curator's exact repro: two overlapping labeled stubs now merge to one net AND `validate()` emits one `collinear_net_conflict` error (severity `error`).

Softstart is local-only; only synthetic fixtures reproducing the geometric classes are used.

Closes #4143

### DIFF
--- a/src/kicad_tools/schematic/models/netlist_mixin.py
+++ b/src/kicad_tools/schematic/models/netlist_mixin.py
@@ -9,6 +9,8 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
+from .wire_geometry import wire_segments_connect
+
 if TYPE_CHECKING:
     pass
 
@@ -92,6 +94,25 @@ class SchematicNetlistMixin:
             wire_segments.append((p1, p2))
             # Connect wire endpoints
             union(p1, p2)
+
+        # Union wires with each other on collinear overlap or mid-segment
+        # T-touch — matching KiCad, which merges the nets of any two touching
+        # or overlapping wire segments, not only those that share an exact
+        # endpoint (issue #4143).  The endpoint case is already handled by the
+        # loop above.  All-pairs O(n^2) is fine at typical schematic sizes
+        # (check_wire_collisions() already scans this way with no reported
+        # slowdown); no spatial index is warranted here.
+        #
+        # NOTE: this is wire-to-wire unioning only.  It deliberately does NOT
+        # relax the pin-to-wire endpoint-only rule from #4020/#4003 — pins
+        # still attach to wires solely at endpoints/junctions (see
+        # connect_to_wire(endpoint_only=True) below).
+        for i in range(len(wire_segments)):
+            a_start, a_end = wire_segments[i]
+            for j in range(i + 1, len(wire_segments)):
+                b_start, b_end = wire_segments[j]
+                if wire_segments_connect(a_start, a_end, b_start, b_end):
+                    union(a_start, b_start)
 
         # Helper to check if point is on a wire segment
         def point_on_segment(point: tuple, seg_start: tuple, seg_end: tuple) -> bool:
--- a/src/kicad_tools/schematic/models/validation_mixin.py
+++ b/src/kicad_tools/schematic/models/validation_mixin.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING
 from ..grid import is_on_grid, snap_to_grid
 from ..logging import _log_debug, _log_info, _log_warning
 from .elements import WireCollision
+from .wire_geometry import wire_segments_connect
 
 if TYPE_CHECKING:
     pass
@@ -131,6 +132,10 @@ class SchematicValidationMixin:
         # Check wire connectivity
         connectivity_issues = self._check_wire_connectivity()
         issues.extend(connectivity_issues)
+
+        # Check for cross-net collinear-overlap / T-touch shorts (issue #4143)
+        conflict_issues = self._check_collinear_net_conflicts()
+        issues.extend(conflict_issues)
 
         # Check for unconnected pins (all types, not just power)
         unconnected_pin_issues = self._check_unconnected_pins()
@@ -529,6 +534,17 @@ class SchematicValidationMixin:
             p2 = (round(wire.x2, 2), round(wire.y2, 2))
             wire_segments.append((p1, p2))
 
+        # Union wires with each other on collinear overlap or mid-segment
+        # T-touch, matching KiCad and mirroring the same fix applied to
+        # _build_connectivity_graph() (issue #4143).  This duplicate
+        # connectivity builder had the identical endpoint-only gap.
+        for i in range(len(wire_segments)):
+            a_start, a_end = wire_segments[i]
+            for j in range(i + 1, len(wire_segments)):
+                b_start, b_end = wire_segments[j]
+                if wire_segments_connect(a_start, a_end, b_start, b_end):
+                    union(a_start, b_start)
+
         for junc in self.junctions:
             junc_pos = (round(junc.x, 2), round(junc.y, 2))
             # Connect junction to any wire segment it's on
@@ -790,6 +806,102 @@ class SchematicValidationMixin:
                 _log_debug(f"  {collision}")
 
         return collisions
+
+    def _wire_net_names(self) -> list[set[str]]:
+        """Map each wire to the set of net names directly labelling it.
+
+        A net name is attributed to a wire when a label / global label /
+        hierarchical label / power symbol sits *on* that wire segment (within
+        ``POINT_TOLERANCE``).  This is intentionally per-wire and independent
+        of Union-Find so that :meth:`_check_collinear_net_conflicts` can tell
+        whether two geometrically-touching wires were *each* independently
+        named with a *different* net — the signature of an unintended short
+        (issue #4143) — rather than merely sharing one net after the union.
+
+        Returns:
+            A list parallel to ``self.wires``; entry *i* is the set of net
+            names attached to ``self.wires[i]``.
+        """
+        named_points: list[tuple[float, float, str]] = []
+        for label in self.labels:  # type: ignore[attr-defined]
+            named_points.append((label.x, label.y, label.text))
+        for gl in self.global_labels:  # type: ignore[attr-defined]
+            named_points.append((gl.x, gl.y, gl.text))
+        for hl in self.hier_labels:  # type: ignore[attr-defined]
+            named_points.append((hl.x, hl.y, hl.text))
+        for pwr in self.power_symbols:  # type: ignore[attr-defined]
+            net_name = pwr.lib_id.split(":")[1] if ":" in pwr.lib_id else pwr.lib_id
+            named_points.append((pwr.x, pwr.y, net_name))
+
+        per_wire: list[set[str]] = []
+        for wire in self.wires:  # type: ignore[attr-defined]
+            names: set[str] = set()
+            for x, y, name in named_points:
+                if self._point_on_wire(x, y, wire):  # type: ignore[attr-defined]
+                    names.add(name)
+            per_wire.append(names)
+        return per_wire
+
+    def _check_collinear_net_conflicts(self) -> list[dict]:
+        """Flag two overlapping/T-touching wires that carry *different* nets.
+
+        Once :meth:`_build_connectivity_graph` unions collinear-overlapping /
+        T-touching wires (issue #4143), two stubs labelled with different net
+        names that overlap geometrically end up merged into one net — almost
+        always an unintended short (the softstart rev-B failure: ``+3.3V``
+        shorted to ``GND`` by interleaved decoupling-cap stubs).  This emits
+        an **error** for each such conflicting pair so the short surfaces in
+        ``validate()`` output instead of hiding as a silent merge.
+
+        Same-net overlaps (both wires labelled with the same net, or one/both
+        unlabelled) are *not* flagged: only a genuine two-different-named-net
+        overlap is reported.
+        """
+        issues: list[dict] = []
+        wire_nets = self._wire_net_names()
+        wires = self.wires  # type: ignore[attr-defined]
+
+        reported: set[tuple[str, str, tuple[float, float]]] = set()
+        for i in range(len(wires)):
+            a = wires[i]
+            a_start = (round(a.x1, 2), round(a.y1, 2))
+            a_end = (round(a.x2, 2), round(a.y2, 2))
+            for j in range(i + 1, len(wires)):
+                b = wires[j]
+                b_start = (round(b.x1, 2), round(b.y1, 2))
+                b_end = (round(b.x2, 2), round(b.y2, 2))
+
+                if not wire_segments_connect(a_start, a_end, b_start, b_end):
+                    continue
+
+                # Only a conflict when each wire independently carries a
+                # named net and those net-name sets are disjoint.
+                names_a = wire_nets[i]
+                names_b = wire_nets[j]
+                if not names_a or not names_b or (names_a & names_b):
+                    continue
+
+                for net_a in sorted(names_a):
+                    for net_b in sorted(names_b):
+                        key = (net_a, net_b, a_start)
+                        if key in reported:
+                            continue
+                        reported.add(key)
+                        issues.append(
+                            {
+                                "severity": "error",
+                                "type": "collinear_net_conflict",
+                                "message": (
+                                    f"Wire segments carrying different nets '{net_a}' and "
+                                    f"'{net_b}' overlap collinearly or T-touch near {a_start}, "
+                                    "merging them into one net (likely an unintended short)."
+                                ),
+                                "location": a_start,
+                                "fix_applied": False,
+                            }
+                        )
+
+        return issues
 
 
 def summarize_issues_by_type(issues: list[dict]) -> dict[str, list[dict]]:
--- a/src/kicad_tools/schematic/models/wire_geometry.py
+++ b/src/kicad_tools/schematic/models/wire_geometry.py
@@ -1,0 +1,171 @@
+"""Wire-vs-wire geometric union primitives for schematic connectivity.
+
+KiCad unions the nets of *any* two touching or overlapping collinear wire
+segments, not just wires that share an exact endpoint.  The schematic
+connectivity model historically unioned wires only at their two endpoints,
+so a generator using the stub-wire+label idiom could produce mid-segment
+overlaps that KiCad merges but kicad-tools kept separate — a silent,
+LVS-grade net-merge divergence (issue #4143).
+
+This module provides the single, tested geometric primitive
+(:func:`wire_segments_connect`) used by both
+``SchematicNetlistMixin._build_connectivity_graph`` and
+``SchematicValidationMixin.validate_power_nets`` so the two connectivity
+builders cannot drift apart.
+
+Two axis-aligned wire segments A and B are considered electrically
+connected (beyond a shared endpoint) when either:
+
+* **Collinear overlap** — A and B lie on the same infinite line and their
+  projected parametric ranges share a sub-segment of nonzero length.  This
+  covers A containing B, B containing A, and partial overlap where neither
+  contains the other (the softstart rev-B repro geometry).
+* **Mid-segment T-touch** — one endpoint of A lands in the *interior* of B
+  (not at B's endpoints), or vice-versa.
+
+A shared endpoint alone is NOT reported here: that case is already handled
+by the endpoint Union-Find in the connectivity builders.  This mirrors, but
+is deliberately distinct from, the pin-endpoint-only semantics of
+#4020/#4003 — *pins* attach to wires only at endpoints/junctions, while
+*wires* union with each other on any overlap/T-touch, matching KiCad.
+"""
+
+from __future__ import annotations
+
+# Tolerance (mm) for perpendicular-distance-to-line and endpoint-coincidence
+# checks.  Matches ``SchematicElementsMixin.POINT_TOLERANCE`` (0.1mm) — the
+# same permissive snap tolerance used elsewhere in the connectivity graph.
+# Do NOT use the tighter 0.005mm from ``sch_cleanup_wires``: that value is
+# tuned for a *destructive* wire-deletion decision, whereas connectivity
+# union is non-destructive and should be at least as permissive as KiCad's
+# own snap tolerance.
+POINT_TOLERANCE = 0.1
+
+Point = tuple[float, float]
+
+
+def _points_equal(p1: Point, p2: Point, tolerance: float = POINT_TOLERANCE) -> bool:
+    """Return True if two points coincide within *tolerance*."""
+    return abs(p1[0] - p2[0]) <= tolerance and abs(p1[1] - p2[1]) <= tolerance
+
+
+def _point_on_segment_interior(
+    point: Point, seg_start: Point, seg_end: Point, tolerance: float = POINT_TOLERANCE
+) -> bool:
+    """Return True if *point* lies on the interior of the segment.
+
+    "Interior" excludes the two endpoints (within *tolerance*): a point that
+    only coincides with an endpoint is handled by endpoint Union-Find, not
+    here.  Uses perpendicular distance to the infinite line plus a
+    parametric bounds check, so it works for a point anywhere along the
+    segment body regardless of orientation.
+    """
+    px, py = point
+    ax, ay = seg_start
+    bx, by = seg_end
+
+    dx = bx - ax
+    dy = by - ay
+    seg_len_sq = dx * dx + dy * dy
+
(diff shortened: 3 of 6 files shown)


rjwalters:
✅ **Approved!**

Reviewed with extra rigor since this changes core schematic connectivity semantics that LVS/validation/netlist assertions all inherit. The implementation is clean, correct, and faithful to the curated acceptance criteria.

## Acceptance criteria — all verified

| Criterion | Verdict |
|-----------|---------|
| `wire_segments_connect()`: collinear overlap (partial + full containment) unions | ✅ verified via `test_wire_geometry.py` + direct probes |
| Mid-segment T-touch unions | ✅ endpoint-on-interior in both directions |
| Bare endpoint touch returns False (delegated to endpoint Union-Find) | ✅ `test_shared_endpoint_not_reported` + probe |
| Bare X-crossing of perpendicular wires returns False (no junction → no connect) | ✅ `test_bare_crossing_does_not_connect` + probe: crossing point is on A's interior but is not an *endpoint* of B, so no union — correct KiCad semantics |
| `POINT_TOLERANCE` = 0.1mm (not the destructive 0.005 from `sch_cleanup_wires`) | ✅ documented rationale in module docstring |
| `_build_connectivity_graph()`: all-pairs wire-vs-wire union added; pin-to-wire stays endpoint-only | ✅ `TestPinEndpointOnlyWireAttach` 5/5 unchanged (#4020/#4003 guard intact) |
| `validate_power_nets()` gets the same fix | ✅ duplicate builder patched identically |
| `validate()` emits `collinear_net_conflict` at **error** severity, only on two *different* labeled/power nets | ✅ `TestCollinearNetConflict` 6/6; same-net and one-unlabeled overlaps correctly not flagged |
| Curator repro merges in `extract_netlist` AND fires a validate() error | ✅ reproduced the exact two-overlapping-label-stub sketch: 1 error, severity `error` |

## Edge cases independently checked (beyond the PR's tests)

- **Diagonal collinear overlap** — the general perpendicular-distance/parametric math handles non-axis-aligned segments correctly (`_collinear_overlap((0,0),(10,10),(3,3),(13,13))` → True), so the "axis-aligned in practice" assumption is not load-bearing.
- **Zero-length (degenerate) wire** — `seg_len_sq <= tol²` guard short-circuits `_collinear_overlap` and `_point_on_segment_interior`; no false unions from degenerate segments.
- **Shared endpoint + collinear** — correctly returns False from `wire_segments_connect` (already unioned by endpoint pass), no double-handling.
- **Tolerance boundary** — perpendicular offset 0.09mm collinear=True, 0.11mm=False; clean at the 0.1mm line.
- **Overlap-region label** — a label landing in the shared sub-segment attributes to both wires, so `names_a & names_b` is non-empty → no spurious conflict. Correct.

## Test execution

- Targeted suite (`test_wire_geometry`, `test_netlist_query`, `test_schematic_models`, `test_validate_netlist`, `test_connectivity`, `test_hierarchical_netlist`): **367 passed** locally, matching the PR claim.
- `ruff check` + `ruff format --check` on all changed files: clean.
- CI: **16/17 jobs green** — Lint, Type Check, all 8 Board E2E/LVS jobs, all routing regressions, C++ build, DRC check, round-trip smoke. The O(n²) wire pass did not perturb any board's netlist/LVS result.

## Note on the one red check (not blocking)

The `Test` job failed on **3 consecutive re-runs**, but every failure is a `Timeout (>60.0s)` in the board-05 **PCB DRC subprocess** tests (`test_board_05_drc_hotspot_regression.py`, `test_board_05_drc_allowlist.py`) — zero relationship to this schematic-only PR. I ran those exact tests against this branch locally: **6/6 pass in 113s (~19s each)**. Under CI's `-n auto` + strict 60s per-test timeout, running concurrently with the 19-minute routing-regression matrix, they starve for CPU and blow the cap; the failing set even varies run-to-run (3 vs 5), the signature of a contention flake, not a regression. Filed as #4160 with suggested fixes (mark `slow` / raise module timeout / serial DRC step). `mergeStateStatus` is UNSTABLE (non-required check), not BLOCKED.

Great work — the single shared geometry primitive prevents the two connectivity builders from drifting, the docstrings clearly delineate wire-vs-wire from the pin-endpoint-only rule, and the conflict detector's net-name-set logic precisely targets genuine unintended shorts.

Closes #4143.

PR #4162, opened 2026-07-14
fix(schematic): thread local_symbol_libs through load(); accept ref str in print_symbol_pins

### LINKED ISSUE #4144: schematic API: Schematic.load() rejects local_symbol_libs; print_symbol_pins(ref) crashes on str
### 1. Constructor/load asymmetry on `local_symbol_libs`

```python
sch = Schematic(title="x", local_symbol_libs=[lib_path])   # works
sch = Schematic.load(path, local_symbol_libs=[lib_path])   # TypeError
# SchematicIOMixin.load() got an unexpected keyword argument 'local_symbol_libs'
```

Any round-trip workflow (generate → write → load → ERC/analyze) that uses a project-local symbol library can't reload its own output with the custom lib registered. Workaround: run kicad-cli directly on the written file.

### 2. `print_symbol_pins` crashes when given a reference string

```python
sch.print_symbol_pins("U1")
# AttributeError: 'str' object has no attribute 'reference'
```

Either accept a ref string (look up the SymbolInstance) or type/doc the parameter as SymbolInstance. Discovering pin numbering currently requires spelunking `symbol_instance.symbol_def.pins` — which matters a lot for agents verifying pin maps against datasheets (cf. the softstart issue #27 pinout audit).

---

### Root Cause

**Bug 1 — `local_symbol_libs` asymmetry.** `Schematic.load()` in
`src/kicad_tools/schematic/models/io_mixin.py:42-71` only accepts `path`
and delegates to `Schematic._from_sexp(doc)` (line 66), which itself
takes no extra kwargs. `_from_sexp` constructs the instance directly via
`cls(title=..., date=..., ..., snap_mode=SnapMode.OFF)`
(`io_mixin.py:132-143`) — this call site enumerates 8 constructor kwargs
but omits `local_symbol_libs` entirely, so there is no path to populate
`self.local_symbol_libs` (`schematic.py:191`) when loading from disk.
Because `load()` doesn't declare the kwarg either, passing it raises
`TypeError: load() got an unexpected keyword argument 'local_symbol_libs'`
rather than silently dropping it.

`resolve_lib_path()` (`schematic.py:193-214`) is the consumer: it walks
`self.local_symbol_libs` to resolve `LIBNAME:SYMNAME` lib_ids against
project-local `.kicad_sym` files before falling through to stock KiCad
paths. A loaded schematic with an empty `local_symbol_libs` list simply
can't resolve those project-local lib_ids on any subsequent
`add_symbol()` call — the failure mode is a `TypeError` at the `load()`
call site today, but would become a silent symbol-resolution miss if the
kwarg were merely accepted without being threaded through.

**Constructor-kwarg asymmetry survey.** `Schematic.__init__` accepts 14
kwargs beyond `self`:
`title, date, revision, company, comment1, comment2, paper, project_name,
sheet_uuid, parent_uuid, page, grid, snap_mode, local_symbol_libs`.
The `cls(...)` call inside `_from_sexp` (`io_mixin.py:132-143`) passes
only: `title, date, revision, company, comment1, comment2, paper,
sheet_uuid, snap_mode`. That leaves four kwargs un-threaded:

- `local_symbol_libs` — **broken symptom**: `load()` doesn't even accept
  it, so this is the `TypeError` reported in this issue. Fix required.
- `project_name`, `page` — **not actually broken**: both are set
  separately after construction, by parsing the file's
  `sheet_instances` block (`io_mixin.py:150-158`), which is the correct
  source of truth for a *loaded* schematic (the file's own project name/
  page, not a caller override). No fix needed; call out in code comments
  only if touched.
- `parent_uuid` — **silent gap, but out of scope**: never threaded
  through `_from_sexp`, and `load()` doesn't expose it either. A loaded
  sub-sheet schematic always gets `parent_uuid=None`, which matters for
  `sheet_path` (`schematic.py:216-221`) in hierarchical designs. No
  crash today (no caller currently passes it), so this is a
  silent-behavior gap rather than the acute bug reported here. Worth a
  follow-up issue but should NOT be bundled into this PR — it needs its
  own design discussion (should `load()` infer parent_uuid from the
  sheet's own `sheet_instances.project.path`, or should it always be an
  explicit caller-supplied override?).
- `grid` — **silent gap, but out of scope**: same pattern as
  `parent_uuid` — never threaded, no current caller, not the crash
  reported here. Leave for a separate issue if it becomes a real
  problem.

**Recommendation for this issue**: fix only `local_symbol_libs` (the
kwarg the issue actually reports as broken — it's the one with an acute
`TypeError` and a real workaround-blocking use case). Do not add
`parent_uuid`/`grid` threading in this PR; note them in the PR
description as known, deliberately-deferred gaps so a future curator
pass doesn't rediscover them from scratch.

**Bug 2 — `print_symbol_pins` crashes on `str`.**
`src/kicad_tools/schematic/models/wiring_mixin.py:707-715` declares
`def print_symbol_pins(self, symbol: SymbolInstance, name: str = None)`
and immediately does `symbol.reference` / `symbol.symbol_def.pins` /
`symbol.pin_position(...)` — all `SymbolInstance` attributes. Passing a
`str` (a bare reference designator like `"U1"`) raises `AttributeError:
'str' object has no attribute 'reference'` on the first line of the
method body.

The codebase already has the canonical ref → instance resolver:
`find_symbol(reference: str) -> SymbolInstance | None` in
`src/kicad_tools/schematic/models/query_mixin.py:107-112`, which does a
linear scan of `self.symbols` for `sym.reference == reference` and
returns `None` on no match. `print_symbol_pins` should reuse this rather
than re-implementing lookup.

**Scope check — other str/instance ambiguity in `wiring_mixin.py`.**
Every other method in this file (`wire_pin_to_point`, `wire_pins`,
`wire_ferrite_bead`, `wire_oscillator`, and the several private
`_route_*` helpers) also takes `symbol: SymbolInstance` /
`sym1/sym2: SymbolInstance` / `fb: SymbolInstance` / `osc: SymbolInstance`
positional parameters with no ref-string handling. This is a
*consistent* convention across the wiring API — callers are expected to
already hold a `SymbolInstance` (typically the return value of
`add_symbol()` or `find_symbol()`), and these methods are called
programmatically inside larger routing sequences, not as one-off agent
debug commands. Broadening ref-string acceptance to all of them is a
much larger, cross-cutting change (touches ~6 method signatures, each
needs its own resolution + not-found error message) and is explicitly
**out of scope** for this "two small API bugs" issue. `print_symbol_pins`
is qualitatively different — it's a debug/introspection helper
specifically called out in the issue for agent ergonomics (checking pin
maps against datasheets), so accepting a bare ref string there has an
outsized ergonomics payoff for a single, isolated, low-risk change.

### Implementation Guidance

**Fix 1**: Thread `local_symbol_libs` through the load path.

- Add `local_symbol_libs: list[Path] | None = None` to
  `SchematicIOMixin.load()`'s signature (`io_mixin.py:42-43`).
- Add the same parameter to `_from_sexp()`'s signature
  (`io_mixin.py:73-74`), and pass it into the `cls(...)` call at
  `io_mixin.py:132-143` alongside the existing 8 kwargs.
- `load()` passes its `local_symbol_libs` argument through to
  `_from_sexp(doc, local_symbol_libs=local_symbol_libs)`.
- Update both docstrings (`load()`'s Args block and the module-level
  round-trip example) to document the new parameter, mirroring the
  constructor's existing docstring wording (`schematic.py:128-135`).

**Fix 2**: Accept `str | SymbolInstance` in `print_symbol_pins`.

- Change the signature to
  `def print_symbol_pins(self, symbol: str | SymbolInstance, name: str = None)`.
- At the top of the method, if `isinstance(symbol, str)`, resolve via
  `resolved = self.find_symbol(symbol)`; if `resolved is None`, raise
  `KeyError(f"No symbol with reference {symbol!r} in schematic")` (or
  print a clear message to stderr and return, matching the "debug
  helper" framing already in the docstring — `KeyError` is preferred
  since it's a lookup failure and lets callers `except KeyError`
  programmatically, consistent with dict-like lookup failures
  elsewhere in the codebase). Use the resolved `SymbolInstance` for the
  rest of the method body unchanged.
- Update the docstring to state both accepted forms and the unknown-ref
  behavior.
- This mixin is defined in a class that also inherits query_mixin's
  `find_symbol`, so no new import required — `self.find_symbol(...)` is
  already available on `Schematic` instances at runtime (verify mixin
  MRO order includes `SchematicQueryMixin` before assuming so — grep
  `class Schematic(` in `schematic.py` to confirm; do NOT import
  `SchematicQueryMixin` directly into `wiring_mixin.py`, use the
  runtime `self` attribute instead, matching the existing pattern in
  this mixin of calling `self.add_wire`/`self.add_junction` from
  sibling mixins).

Both fixes are small (signature + a few lines each) and independent —
ship together in one PR per the issue's framing.

### Acceptance Criteria

- [ ] `Schematic.load(path, local_symbol_libs=[lib_path])` no longer
      raises `TypeError` and populates `sch.local_symbol_libs` with the
      given paths.
- [ ] A schematic generated with a project-local symbol lib, written to
      disk, then reloaded with the same `local_symbol_libs=[...]`
      resolves that lib's symbols via `resolve_lib_path()` (round-trip
      parity with the constructor path).
- [ ] `Schematic.load(path)` (no `local_symbol_libs`) continues to work
      exactly as before — default `None` preserves existing behavior for
      all current callers.
- [ ] `sch.print_symbol_pins("U1")` (string ref) works identically to
      `sch.print_symbol_pins(sch.find_symbol("U1"))` (existing
      `SymbolInstance` usage) — same printed output for a matching ref.
- [ ] `sch.print_symbol_pins(some_symbol_instance)` (existing call
      convention) is unaffected — no regression for current callers.
- [ ] `sch.print_symbol_pins("NONEXISTENT")` raises a clear, catchable
      error (`KeyError` recommended) identifying the missing reference,
      not an `AttributeError` or silent no-op.
- [ ] No changes to any other `wiring_mixin.py` method signatures
      (`wire_pins`, `wire_pin_to_point`, `wire_ferrite_bead`,
      `wire_oscillator`, etc. remain `SymbolInstance`-only — confirmed
      out of scope above).

### Test Plan

- [ ] Automated test in `tests/test_schematic_models.py` (extends the
      existing `TestSchematicLoading` class around
      `tests/test_schematic_models.py:1546-1642`): round-trip test that
      (a) builds a synthetic `.kicad_sym` fixture with one symbol
      (mirror the pattern used in `tests/test_board_05_drv8301_pin_nets.py:116`,
      which already exercises `Schematic(title=..., local_symbol_libs=[...])`
      at construction time), (b) generates a schematic using
      `add_symbol()` against that lib, (c) writes it, (d) calls
      `Schematic.load(path, local_symbol_libs=[lib_path])`, and asserts
      `loaded.local_symbol_libs == [lib_path]` and that
      `loaded.resolve_lib_path(lib_stem)` returns the fixture path.
- [ ] Automated test: `Schematic.load(path)` without `local_symbol_libs`
      still works (regression guard for the default-`None` case) —
      likely already covered by `test_load_from_file` at
      `tests/test_schematic_models.py:1553`; confirm it still passes
      unmodified.
- [ ] Automated test (new, e.g. in `tests/test_schematic_models.py` or a
      dedicated `tests/test_print_symbol_pins.py`): call
      `print_symbol_pins("U1")` on a schematic with a placed `U1` symbol
      and assert it does not raise and produces output equivalent to
      calling with the `SymbolInstance` directly (capture stdout via
      `capsys` and compare).
- [ ] Automated test: `print_symbol_pins("DOES_NOT_EXIST")` raises
      `KeyError` (or documented equivalent) rather than `AttributeError`.
- [ ] All fixtures synthetic/in-repo — no external KiCad libraries or
      network access required.
- [ ] `uv run pytest tests/test_schematic_models.py -k "load or symbol_pins"` passes.
- [ ] `uv run ruff check src/kicad_tools/schematic/models/io_mixin.py src/kicad_tools/schematic/models/wiring_mixin.py` and `uv run mypy src/kicad_tools/schematic/models/io_mixin.py src/kicad_tools/schematic/models/wiring_mixin.py` pass.

### Affected Files

- `src/kicad_tools/schematic/models/io_mixin.py` — add `local_symbol_libs`
  param to `SchematicIOMixin.load()` (line 42) and `_from_sexp()` (line
  73), thread into the `cls(...)` construction call (lines 132-143),
  update docstrings.
- `src/kicad_tools/schematic/models/wiring_mixin.py` — widen
  `print_symbol_pins` (line 707) to accept `str | SymbolInstance`,
  resolve via `self.find_symbol()`, raise `KeyError` on unknown ref,
  update docstring.
- `src/kicad_tools/schematic/models/schematic.py` — reference only, no
  changes expected (constructor already correct; confirms the target
  shape for `_from_sexp`'s `cls(...)` call).
- `src/kicad_tools/schematic/models/query_mixin.py` — reference only,
  no changes expected (`find_symbol()` at line 107 is reused as-is).
- `tests/test_schematic_models.py` — add round-trip
  `local_symbol_libs` test and `print_symbol_pins` str/instance/unknown-ref
  tests (extend `TestSchematicLoading` around line 1546, or add a new
  test class in the same file).

### Out of Scope (explicitly deferred)

- Threading `parent_uuid` / `grid` through `load()`/`_from_sexp()` —
  no current caller, no crash, needs its own design discussion on
  `load()`-time inference vs. explicit override. File a follow-up issue
  if a real use case emerges.
- Broadening `str | SymbolInstance` acceptance to other `wiring_mixin.py`
  methods (`wire_pins`, `wire_pin_to_point`, `wire_ferrite_bead`,
  `wire_oscillator`, etc.) — consistent `SymbolInstance`-only convention
  across the whole wiring API; a cross-cutting change, not "trivially
  small."

rjwalters:
## Summary

Fixes two small schematic-API bugs reported in #4144:

1. **`Schematic.load()` rejected `local_symbol_libs`.** The constructor
   accepts `local_symbol_libs`, but `load()`/`_from_sexp()` never accepted
   or threaded it into the `cls(...)` construction, so passing it raised
   `TypeError`. Any round-trip workflow (generate → write → load → analyze)
   using a project-local `.kicad_sym` library couldn't reload its own output
   with the custom lib registered.

2. **`print_symbol_pins(ref)` crashed on a `str`.** The method dereferenced
   `symbol.reference` assuming a `SymbolInstance`, raising `AttributeError`
   when handed a bare reference designator like `"U1"`.

## Changes

- `io_mixin.py`: `load()` and `_from_sexp()` now accept
  `local_symbol_libs: list[Path] | None = None` and thread it into the
  `cls(...)` construction alongside the existing kwargs. Docstrings updated.
  A `TYPE_CHECKING` `__init__` declaration on `SchematicIOMixin` lets mypy
  validate the `cls(...)` kwargs against the real `Schematic.__init__` shape
  (avoids a spurious `call-arg` error on the new keyword; also clears the
  9 pre-existing baselined `call-arg` false positives on the same call site).
- `wiring_mixin.py`: `print_symbol_pins` now accepts `str | SymbolInstance`.
  String refs are resolved via the existing `find_symbol()` (sibling query
  mixin, already in the MRO); unknown refs raise a clear, catchable
  `KeyError`. Existing `SymbolInstance` callers are unaffected.
- `tests/test_schematic_models.py`: added `TestSchematicLoadLocalSymbolLibs`
  (round-trip with a synthetic project-local `.kicad_sym`, default-`None`
  regression guard) and `TestPrintSymbolPins` (str vs instance parity,
  unknown-ref `KeyError`). All fixtures synthetic/in-repo.

| `load(path, local_symbol_libs=[...])` no longer raises, populates attr | ✅ | `test_load_accepts_local_symbol_libs` |
| Round-trip parity: reloaded schematic resolves lib via `resolve_lib_path()` | ✅ | `test_load_local_lib_resolves_via_resolve_lib_path`, `test_round_trip_add_symbol_against_local_lib` |
| `load(path)` (no arg) unchanged, defaults to empty list | ✅ | `test_load_without_local_symbol_libs_defaults_empty` + existing `test_load_from_file` |
| `print_symbol_pins("U1")` == `print_symbol_pins(instance)` output | ✅ | `test_print_symbol_pins_str_matches_instance` |
| `print_symbol_pins(instance)` unaffected | ✅ | Same test compares both paths |
| unknown ref raises clear catchable `KeyError` | ✅ | `test_print_symbol_pins_unknown_ref_raises_keyerror` |
| No other `wiring_mixin.py` signatures changed | ✅ | Only `print_symbol_pins` touched (out-of-scope methods left as `SymbolInstance`-only) |

## Out of scope (deliberately deferred, per curator)

- Threading `parent_uuid` / `grid` through `load()`/`_from_sexp()` — no current
  caller, no crash; needs its own design discussion (load-time inference vs.
  explicit override).
- Broadening `str | SymbolInstance` acceptance to other `wiring_mixin.py`
  methods (`wire_pins`, `wire_pin_to_point`, `wire_ferrite_bead`,
  `wire_oscillator`) — consistent `SymbolInstance`-only convention; cross-cutting.

## Test Plan

- `uv run pytest tests/test_schematic_models.py -k "load or symbol_pins"` → 10 passed
- `uv run pytest tests/test_schematic_models.py tests/test_board_05_drv8301_pin_nets.py` → 230 passed
- `uv run ruff check` + `ruff format --check` on all changed files → clean
- `uv run python scripts/ci/check_mypy_baseline.py` → 0 new errors from the
  changed files (the only "new" line is a pre-existing `cmaes` import-stub
  env artifact present on the base branch too, verified by stashing the diff).

Closes #4144

### DIFF
--- a/src/kicad_tools/schematic/models/io_mixin.py
+++ b/src/kicad_tools/schematic/models/io_mixin.py
@@ -39,14 +39,49 @@ class SchematicIOMixin:
         _PWR_SYNTH_LIB_PREFIX: str
         _synthesized_pwr_defs: dict[str, SExp]
 
+        # ``_from_sexp`` constructs the concrete ``Schematic`` via ``cls(...)``.
+        # Declaring the constructor signature here lets mypy validate those
+        # keyword arguments against the real ``Schematic.__init__`` shape
+        # instead of the empty ``object`` init (resolves the spurious
+        # ``call-arg`` errors on the ``cls(...)`` call site).
+        def __init__(
+            self,
+            title: str = "",
+            date: str = "2025-01",
+            revision: str = "A",
+            company: str = "",
+            comment1: str = "",
+            comment2: str = "",
+            paper: str = "A4",
+            project_name: str = "project",
+            sheet_uuid: str | None = None,
+            parent_uuid: str | None = None,
+            page: str = "1",
+            grid: float = ...,
+            snap_mode: object = ...,
+            local_symbol_libs: list[Path] | None = None,
+        ) -> None: ...
+
     @classmethod
-    def load(cls, path: str | Path) -> Schematic:
+    def load(
+        cls,
+        path: str | Path,
+        local_symbol_libs: list[Path] | None = None,
+    ) -> Schematic:
         """Load a schematic from a .kicad_sch file.
 
         This enables round-trip editing: load -> modify -> save.
 
         Args:
             path: Path to the .kicad_sch file
+            local_symbol_libs: Optional list of project-local ``.kicad_sym``
+                files to register on the loaded schematic (mirrors the
+                ``Schematic(...)`` constructor argument).  These are consulted
+                by :meth:`resolve_lib_path` during subsequent ``add_symbol()``
+                calls so a reloaded schematic can resolve ``LIBNAME:SYMNAME``
+                ids against project-local libraries, matching the constructor
+                path.  Default ``None`` preserves prior behavior (stock libs
+                only) for all existing callers.
 
         Returns:
             A Schematic instance populated with all elements from the file
@@ -55,6 +90,12 @@ class SchematicIOMixin:
             sch = Schematic.load("power.kicad_sch")
             sch.add_symbol("Device:R", 100, 100, "R5", "10k")
             sch.write("power.kicad_sch")
+
+            # Reload with a project-local symbol library registered:
+            sch = Schematic.load(
+                "power.kicad_sch",
+                local_symbol_libs=[Path("libs/custom.kicad_sym")],
+            )
         """
         from kicad_tools.sexp import parse_file
 
@@ -63,7 +104,7 @@ class SchematicIOMixin:
             raise FileNotFoundError(f"Schematic file not found: {path}")
 
         doc = parse_file(path)
-        sch = cls._from_sexp(doc)
+        sch = cls._from_sexp(doc, local_symbol_libs=local_symbol_libs)
         # Track the source path so operations that need to walk
         # sub-sheets (e.g., extract_netlist(hierarchical=True), run_erc)
         # can resolve relative sheet references.
@@ -71,10 +112,20 @@ class SchematicIOMixin:
         return sch
 
     @classmethod
-    def _from_sexp(cls, doc: SExp) -> Schematic:
+    def _from_sexp(
+        cls,
+        doc: SExp,
+        local_symbol_libs: list[Path] | None = None,
+    ) -> Schematic:
         """Create a Schematic from a parsed S-expression tree.
 
         This is the internal method that does the actual parsing.
+
+        Args:
+            doc: Parsed S-expression tree for the schematic.
+            local_symbol_libs: Optional project-local ``.kicad_sym`` files to
+                register on the constructed schematic (threaded through from
+                :meth:`load`).  Default ``None`` preserves prior behavior.
         """
         from .schematic import SnapMode
 
@@ -140,6 +191,7 @@ class SchematicIOMixin:
             paper=paper,
             sheet_uuid=sheet_uuid,
             snap_mode=SnapMode.OFF,  # Preserve original coordinates
+            local_symbol_libs=local_symbol_libs,
         )
 
         # Store embedded lib_symbols for round-trip
--- a/src/kicad_tools/schematic/models/wiring_mixin.py
+++ b/src/kicad_tools/schematic/models/wiring_mixin.py
@@ -704,13 +704,34 @@ class SchematicWiringMixin:
 
         return wires
 
-    def print_symbol_pins(self, symbol: SymbolInstance, name: str = None):
-        """Debug helper: Print all pin positions for a symbol."""
-        display_name = name or symbol.reference
-        print(f"\n{display_name} pins at ({symbol.x}, {symbol.y}) rot={symbol.rotation}:")
-        for pin in symbol.symbol_def.pins:
+    def print_symbol_pins(self, symbol: str | SymbolInstance, name: str = None):
+        """Debug helper: Print all pin positions for a symbol.
+
+        Args:
+            symbol: Either a placed :class:`SymbolInstance` or a bare
+                reference designator string (e.g. ``"U1"``).  A string is
+                resolved to its instance via ``find_symbol()``.
+            name: Optional display name override for the printed header;
+                defaults to the symbol's reference designator.
+
+        Raises:
+            KeyError: If ``symbol`` is a string that does not match any
+                placed symbol's reference designator.
+        """
+        if isinstance(symbol, str):
+            # find_symbol is provided by SchematicQueryMixin (sibling in the
+            # Schematic MRO); it is not visible on this mixin in isolation.
+            resolved = self.find_symbol(symbol)  # type: ignore[attr-defined]
+            if resolved is None:
+                raise KeyError(f"No symbol with reference {symbol!r} in schematic")
+            instance = resolved
+        else:
+            instance = symbol
+        display_name = name or instance.reference
+        print(f"\n{display_name} pins at ({instance.x}, {instance.y}) rot={instance.rotation}:")
+        for pin in instance.symbol_def.pins:
             # Use pin.number for unique identification (many symbols have pins all named "~")
-            pos = symbol.pin_position(pin.number)
+            pos = instance.pin_position(pin.number)
             pin_display = pin.name if pin.name and pin.name != "~" else pin.number
             print(f"  {pin_display} ({pin.number}): ({pos[0]:.2f}, {pos[1]:.2f})")
 
--- a/tests/test_schematic_models.py
+++ b/tests/test_schematic_models.py
@@ -1647,6 +1647,126 @@ class TestSchematicLoadAndParse:
         assert sch._pwr_counter == 2
 
 
+_LOCAL_LIB_SEXP = """(kicad_symbol_lib
+	(version 20231120)
+	(generator "kicad_symbol_editor")
+	(generator_version "8.0")
+	(symbol "WIDGET"
+		(property "Reference" "U" (at 0 5.08 0) (effects (font (size 1.27 1.27))))
+		(property "Value" "WIDGET" (at 0 2.54 0) (effects (font (size 1.27 1.27))))
+		(symbol "WIDGET_1_1"
+			(pin input line (at -7.62 2.54 0) (length 2.54) (name "IN" (effects (font (size 1.27 1.27)))) (number "1" (effects (font (size 1.27 1.27)))))
+			(pin output line (at 7.62 0 180) (length 2.54) (name "OUT" (effects (font (size 1.27 1.27)))) (number "2" (effects (font (size 1.27 1.27)))))
+		)
+	)
+)
+"""
+
+
+def _write_local_lib(tmp_path, stem="custom_widgets"):
+    """Write a synthetic project-local .kicad_sym fixture, return its Path."""
+    lib_path = tmp_path / f"{stem}.kicad_sym"
+    lib_path.write_text(_LOCAL_LIB_SEXP)
+    return lib_path
+
+
+class TestSchematicLoadLocalSymbolLibs:
+    """Regression tests for local_symbol_libs on Schematic.load() (issue #4144)."""
+
+    def test_load_accepts_local_symbol_libs(self, tmp_path):
+        """load() accepts local_symbol_libs and populates the attribute."""
+        lib_path = _write_local_lib(tmp_path)
+
+        # Minimal on-disk schematic to reload.
+        src = Schematic(title="RoundTrip", local_symbol_libs=[lib_path])
+        sch_path = tmp_path / "rt.kicad_sch"
+        src.write(sch_path)
+
+        loaded = Schematic.load(sch_path, local_symbol_libs=[lib_path])
+        assert loaded.local_symbol_libs == [lib_path]
+
+    def test_load_local_lib_resolves_via_resolve_lib_path(self, tmp_path):
+        """A reloaded schematic resolves the local lib stem to its file path."""
+        lib_path = _write_local_lib(tmp_path)
+
+        src = Schematic(title="RoundTrip", local_symbol_libs=[lib_path])
+        sch_path = tmp_path / "rt.kicad_sch"
+        src.write(sch_path)
+
+        loaded = Schematic.load(sch_path, local_symbol_libs=[lib_path])
+        assert loaded.resolve_lib_path("custom_widgets") == lib_path
+
+    def test_round_trip_add_symbol_against_local_lib(self, tmp_path):
+        """Generate with a local lib, write, reload, and re-add from the lib.
+
+        Exercises the full round-trip parity path: the reloaded schematic
+        must resolve ``LIBNAME:SYMNAME`` ids against the registered local lib
+        just like the constructor path does.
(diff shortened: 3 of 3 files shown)


rjwalters:
✅ **Approved!**

Both bug fixes are correct, minimal, and well-scoped to issue #4144. All 16 CI checks pass and merge state is CLEAN.

## Correctness

**Fix 1 — `local_symbol_libs` threading (`io_mixin.py`)**
- `load()` and `_from_sexp()` both gain `local_symbol_libs: list[Path] | None = None`; `load()` forwards it into `_from_sexp(doc, local_symbol_libs=...)`, which threads it into the `cls(...)` construction alongside the existing kwargs.
- Default `None` flows into the constructor's `list(local_symbol_libs or [])`, so the no-arg path is unchanged — confirmed by `test_load_without_local_symbol_libs_defaults_empty` and the pre-existing `test_load_from_file`.

**Fix 2 — `print_symbol_pins` str acceptance (`wiring_mixin.py`)**
- Signature widened to `str | SymbolInstance`; string refs resolve via `self.find_symbol()` (correctly sourced from the sibling `SchematicQueryMixin` in the `Schematic` MRO, accessed at runtime with a scoped `# type: ignore[attr-defined]` rather than a direct import — matches the existing sibling-mixin call pattern in this file).
- Unknown refs raise `KeyError(f"No symbol with reference {symbol!r} ...")` — clear and catchable, replacing the prior `AttributeError`. The `SymbolInstance` path is byte-for-byte unchanged (verified by `test_print_symbol_pins_str_matches_instance` comparing capsys output).

## The flagged `__init__` stub — clean, justified debt reduction (not scope creep)

The builder added a `TYPE_CHECKING`-only `__init__` stub on `SchematicIOMixin` mirroring `Schematic.__init__`. I traced this against `.github/mypy-baseline.txt`: lines 1194–1202 are exactly **9 pre-existing baselined `call-arg` false positives** on the `cls(...)` call site (mypy was validating the construction against `object.__init__`). The stub gives mypy the real constructor shape, which:
- validates the new `local_symbol_libs=` kwarg properly (instead of requiring a 10th baseline entry), and
- clears the existing 9 false positives — they now surface as stale-baseline **warnings**, ratcheting the debt floor down rather than up.

The stub declares `title: str = ""` (optional) where the real ctor has `title: str` (required); harmless here since it only validates keyword call-args and all call sites pass `title=` explicitly. This is the right call, not scope creep.

## Scope discipline — confirmed

- `parent_uuid` / `grid` correctly NOT threaded (deferred per curator; no caller, no crash).
- Only `print_symbol_pins` touched in `wiring_mixin.py`; the other `wire_*` methods remain `SymbolInstance`-only.

## Test Execution (local, against PR head)

- `pytest tests/test_schematic_models.py tests/test_board_05_drv8301_pin_nets.py` → **230 passed** ✅
- New tests (`TestSchematicLoadLocalSymbolLibs` ×4, `TestPrintSymbolPins` ×3) all pass and map 1:1 onto every acceptance criterion (attr populate, `resolve_lib_path` round-trip parity, default-None regression, str/instance capsys parity, unknown-ref KeyError). Synthetic `.kicad_sym` fixture, no network/external libs.
- `ruff check` + `ruff format --check` on all changed files → clean ✅
- mypy baseline gate: the only local "NEW" line is the pre-existing `cmaes` import-stub environment artifact in `placement/cmaes_strategy.py` (unrelated file, baselined as `import-untyped`, resolves as `import-not-found` in this venv) — CI Type Check passed, confirming it is environment-only. Zero new errors in the two changed files.

## CI

All 16 checks green, including **Test (11m19s — no board-05 DRC timeout flake)** and **Type Check**. Merge state CLEAN. `Closes #4144` present.

Nice clean fix. 🎯

Issue #4160, opened 2026-07-14
CI flake: board-05 DRC subprocess tests time out under the 60s per-test cap when Test job runs concurrently with routing regressions


rjwalters:
## Summary

The `Test` CI job intermittently fails with `Timeout (>60.0s) from pytest-timeout` on the board-05 DRC subprocess tests:

- `tests/test_board_05_drc_hotspot_regression.py::TestBoard05DRCHotspotRegression::*` (all 4 methods)
- `tests/test_board_05_drc_allowlist.py::TestBoard05DRCAllowlistGuard::test_routed_drc_error_count_at_or_below_allowlist`

Observed on PR #4157 (a schematic-only connectivity change that cannot affect PCB DRC): the `Test` job failed **3 consecutive re-runs** with 3–5 of these board-05 DRC tests timing out each time, while all 16 other CI jobs (Lint, Type Check, all 8 Board E2E/LVS jobs, all routing regressions, C++ build, DRC check, round-trip smoke) passed green. Also observed on #4163, #4164 (and re-runs) — 4 of the last 5 PRs hit this flake.

## Root cause (measured)

These tests spawn a `subprocess` running `kct check --mfr jlcpcb-tier1` DRC on the fully-routed board-05 PCB (`tests/test_board_05_drc_hotspot_regression.py:193-217` / `tests/test_board_05_drc_allowlist.py:151-190`, inner subprocess `timeout=120`). Locally, on an **unloaded** machine with the C++ backend built:

```
6 passed in 113.26s   # ~19s/test average
```

The CI `Test` step (`.github/workflows/ci.yml:142`) runs:

```
uv run pytest -n auto -o addopts= --benchmark-disable --timeout=60 -m "not slow"
```

Neither of these two test files carries `@pytest.mark.slow`, so they run in the main PR-gating `Test` job under the 60s cap. When the `Test` job runs concurrently with the routing-regression job matrix (Board 05 Routing Regression alone took 19m4s on the same run, plus Board 06/07 LVS and diff-pair/match-group jobs), CPU contention pushes these already-~19s DRC subprocess tests past the **60s per-test pytest-timeout**. The set of tests that trip varies run-to-run (3 vs 5), the signature of nondeterministic timeout flakiness rather than a deterministic regression.

## Why this is not a real regression

- Failures are `Timeout`, never assertions.
- All failures are board-05 **PCB DRC** subprocess tests.
- Tests pass 6/6 locally on the affected branch.
[...]

PR #4171, opened 2026-07-14
feat(schematic): add collision-aware add_stub_label helper

### LINKED ISSUE #4161: schematic: add_stub_label(symbol, pin, net) — collision-aware stub+label helper (item 3 of #4143)
## Context

Follow-on from #4143 (merged in PR #4157), which fixed the LVS-grade connectivity divergence (collinear overlapping wires silently merging nets). Item (3) of that issue — the ergonomics helper — was deferred by the curator as purely additive with no dependency coupling.

## Requested feature

A built-in `add_stub_label(symbol, pin, net)` on the Schematic model: place a short stub wire from the given pin and attach a net label, with **collision-aware stub length/direction** so stubs from neighboring parts don't collinearly overlap (the root cause of the #4143 silent merges).

The stub-wire+label idiom is ubiquitous in generated schematics — the softstart rev-A generator hand-rolls exactly this, with a comment noting it exists because long rail taps were "silently merging nets."

---

### Current main state (verified 2026-07-14, commit `7c3d701f`)

#4143 is merged (PR #4157). The collision machinery now on `origin/main`:

- `src/kicad_tools/schematic/models/wire_geometry.py::wire_segments_connect(a_start, a_end, b_start, b_end, tolerance=0.1)` — the single tested primitive: returns `True` for collinear overlap (partial or full containment) or mid-segment T-touch between two axis-aligned segments. Pure shared-endpoint returns `False` (handled separately by endpoint union).
- `src/kicad_tools/schematic/models/validation_mixin.py::_check_collinear_net_conflicts()` (~line 845) — the new `validate()` check. For every wire pair where `wire_segments_connect()` is `True` **and** each wire independently carries a *different* named net (via `_wire_net_names()`, which maps labels/global-labels/hier-labels/power-symbols sitting on a wire within `POINT_TOLERANCE`), it emits `{"severity": "error", "type": "collinear_net_conflict", ...}`. Same-net overlaps are not flagged.
- `POINT_TOLERANCE = 0.1` mm is the shared tolerance constant (`SchematicElementsMixin.POINT_TOLERANCE`, `elements_mixin.py:34`), reused by `wire_geometry.py`. `add_stub_label` should reuse the same constant rather than introduce a new epsilon.

### Existing API surface to build on

**Pin position + outward direction** (`src/kicad_tools/schematic/models/symbol.py::SymbolInstance.pin_position()`, ~line 587, and `src/kicad_tools/schematic/models/pin.py::Pin`):

- `Pin.angle` is the direction the pin graphic extends **into** the symbol body, in library Y-up coordinates: `0=right, 90=up, 180=left, 270=down`. The wire/outward direction a stub should travel is the **opposite** of `angle` (`angle + 180`).
- `pin_position()` computes the absolute connection point by: rotating `Pin.connection_point()` by `SymbolInstance.rotation` (standard CCW matrix, in library Y-up space), then negating Y to convert to schematic Y-down, then translating by `(self.x, self.y)`.
- **`add_stub_label` must apply the identical rotation to the pin's outward direction vector**, not just to its position — i.e. compute `outward_angle_library = (pin.angle + 180) % 360`, rotate that angle by `self.rotation` (same CCW-then-Y-flip transform used for position), and snap the result to the nearest axis-aligned direction (0/90/180/270) since schematic wires are conventionally axis-aligned and stub length should extend straight out from the pin. A pin whose rotated outward angle doesn't land near a cardinal direction (non-multiple-of-90 symbol rotation) should either snap to nearest cardinal or raise — recommend snapping with a documented tolerance (e.g. ±1°) since KiCad symbols are essentially always placed at 0/90/180/270 rotation in generator code today (no existing helper in `wiring_mixin.py` handles arbitrary rotation either).
- Existing wiring helpers that already infer pin-outward direction heuristically (for comparison, not reuse): `connect_hier_label_to_pin()` (`wiring_mixin.py`, picks left/right by comparing `pin_pos[0]` to `symbol.x` — a *position*-based heuristic, not angle-based, and only handles horizontal). `add_stub_label` should be more general (angle-based, all 4 directions) since it's meant to replace hand-rolled per-block stub logic.

**`add_wire` / `add_label`** (`elements_mixin.py`, ~line 660 and ~line 836):
- `add_wire(p1, p2, snap=True, warn_on_collision=True)` returns a `Wire`; `warn_on_collision` only warns about **endpoint-on-interior** collisions today (pre-#4143 semantics, via `_find_wire_collisions_for_point`) — it does NOT know about `wire_segments_connect()`/collinear overlap. `add_stub_label` should pass `warn_on_collision=False` (it does its own pre-flight collision search using `wire_segments_connect`, so the generic per-endpoint warning would be redundant/noisy) and instead raise directly per the search-and-raise design below.
- `add_label(text, x, y, rotation=0, snap=True, validate_connection=True)` returns a `Label`; warns (does not raise) if the label isn't `_point_on_any_wire` — irrelevant here since the label is always placed at the stub's far endpoint, which is always on the just-created wire.

### Prior art: hand-rolled stub+collision helpers (read before designing — do not duplicate their gaps)

Two existing hand-rolled implementations of this exact pattern exist in `src/kicad_tools/schematic/blocks/`:

1. **`src/kicad_tools/schematic/blocks/motor.py`** (~line 2010-2038): a local `_stub_label(pin_pos, net, direction)` closure with a hardcoded `STUB = 2.54` and a caller-supplied `direction` string (`"left"/"right"/"up"/"down"`) — no collision detection at all.
2. **`src/kicad_tools/schematic/blocks/_stub_helpers.py`** (module docstring explicitly cites the silent-net-bridging failure mode — issues #3011/#3015, PR #3014): `_emit_pin_net_stub(sch, pin_pos, x_center, net_name, ports, *, block_label="")`. This is the closest existing precedent:
   - Picks a primary stub side via **position** heuristic (`pin_pos[0] < x_center` → left, else right) — not pin-angle-aware, and horizontal-only (no up/down).
   - Tries the opposite side once as a single fallback.
   - Collision check is `_stub_endpoint_would_collide()` — calls `_find_wire_collisions_for_point`, i.e. **only detects T-touch on the endpoint**, predates `wire_segments_connect()` and does **not** detect collinear overlap (the actual #4143 failure mode — two stubs whose *bodies* overlap without either endpoint landing on the other's interior).
   - **Raises `ValueError`** if both candidates collide — this is the precedent for "raise, don't warn" recommended below.
   - `tests/test_pin_nets_stub_collision.py` is the regression suite for this helper (all 3 behaviors: primary placement, auto-shift fallback, both-sides-raise).

**Recommendation:** `add_stub_label` should be the generalized, angle-aware, `wire_segments_connect()`-based successor to `_emit_pin_net_stub` — a natural (optional, non-blocking) follow-up cleanup is migrating `_stub_helpers.py`'s callers (`LDOBlock`, `GateDriverBlock`, `HalfBridge`, `DebugHeader`) onto the new method once it exists, but that migration is explicitly **out of scope** for this issue (separate PR — those call sites have their own test suite and center-heuristic semantics that would need careful preservation or a documented behavior change).

### Design: collision-aware stub search

**Signature:**
```python
def add_stub_label(
    self,
    symbol: SymbolInstance,
    pin: str,
    net: str,
    *,
    length: float = 2.54,
    direction: str | None = None,
    snap: bool = True,
) -> tuple[Wire, Label]:
```

- `pin`: name or number, passed through to `symbol.pin_position()` (reuses its existing fuzzy-match/alias/error machinery — do not reimplement pin lookup).
- `length`: default stub length in mm. `2.54` (one KiCad grid step) matches both existing hand-rolled precedents (`STUB = 2.54` in `motor.py` and `_STUB = 2.54` in `_stub_helpers.py`) — keep as the default for drop-in familiarity.
- `direction`: optional override (`"left"/"right"/"up"/"down"`, matching the vocabulary already used in `motor.py`'s closure). When `None` (default), compute outward direction from `Pin.angle` + `symbol.rotation` as described above. Passing an explicit `direction` skips the angle computation but still runs the collision search below.

**Candidate search** (bounded, deterministic, cheap — schematic wire counts are in the tens-to-low-hundreds per the #4143 precedent, so an O(candidates × wires) scan is negligible):

1. Compute the primary candidate: outward direction (from angle or override) at the default `length`.
2. If the primary stub segment does not `wire_segments_connect()`-collide with any existing wire carrying a *different* net than `net` (reuse `_wire_net_names()`-style logic, or simpler: check against wires that carry any label at all with a name != `net`, since a stub about to be placed has no wire yet to look up), place it: `add_wire(pin_pos, end, warn_on_collision=False)` + `add_label(net, end[0], end[1], rotation=<derived from direction>)`.
3. If the primary candidate collides, try a **bounded, deterministic candidate list** before giving up:
   - Same direction, `length * 2` (longer stub — reaches past the obstruction).
   - Same direction, `length / 2` (shorter stub — pulls back before the obstruction). Only viable if `length/2` remains a sane minimum (recommend a floor, e.g. `POINT_TOLERANCE`-scale minimum of a few tenths of a mm below which a stub isn't useful — practically `length/2` at the 2.54 default is 1.27mm, still well above any floor concern; only relevant if a caller passes a very small `length`).
   - The two **perpendicular** directions at the default `length` (a small jog) — e.g. for a `"left"` stub, try `"up"` and `"down"` at the pin's position. This models the "small perpendicular jog" the issue asks for.
   - Opposite direction (`"right"` if primary was `"left"`) at the default `length` — mirrors `_emit_pin_net_stub`'s existing fallback behavior for continuity, though note this is less physically natural for a fixed pin (the pin's fixed exit angle means "opposite direction" routes back across/through the symbol body in the naive case — recommend documenting this as a last-resort candidate, or dropping it in favor of always trying perpendicular jogs first, since perpendicular avoids crossing the symbol). **Decide the exact ordering during implementation**; the important constraint is the set stays small (≤ 6 candidates) and deterministic (same input → same chosen candidate, for reproducible generator output).
4. Each candidate is validated the same way as the primary: no `wire_segments_connect()` overlap with any *differently-named* wire.

**No-solution behavior: raise (not warn).** This matches the existing `_emit_pin_net_stub` precedent (`ValueError` when both candidates collide) and is the direct corrective for the #4143 failure mode itself — a silently-placed colliding stub is exactly the bug this whole issue chain exists to eliminate. Recommend a dedicated exception (e.g. reuse `ValueError` for consistency with `_stub_helpers.py`, or introduce a small `StubPlacementError(ValueError)` if the codebase's error-hierarchy conventions call for a named exception — check `elements.py`/existing exception classes like `PinNotFoundError` for the house style before deciding). The exception message should include the pin's absolute position, the net name, and all tried candidate endpoints, so a generator author can immediately see why placement failed and adjust layout.

**Return value:** `tuple[Wire, Label]` — the created wire and label objects. This is the most ergonomic choice for generators because:
- Callers who only want the endpoint can get it via `label.x, label.y` (or the wire's far endpoint) without a separate accessor.
- Callers who want to chain further wiring (e.g. attach a junction, or record the port in a `ports` dict as `_stub_helpers.py`'s callers do) have both objects directly, matching the existing convention where `add_wire()` and `add_label()` each return their created object.
- A `NamedTuple` (e.g. `StubLabel(wire: Wire, label: Label)`) is a nice-to-have for readability (`result.wire`, `result.label`) but a plain 2-tuple matching existing multi-return conventions elsewhere in the mixins (e.g. `wire_pin_to_point` returns `list[Wire]`) is also acceptable — implementer's choice, document whichever is chosen in the docstring.

### Acceptance Criteria

- [ ] `add_stub_label(symbol, pin, net)` exists on `Schematic` (natural home: `SchematicElementsMixin` in `elements_mixin.py`, next to `add_wire`/`add_label`, OR `SchematicWiringMixin` in `wiring_mixin.py` next to the other pin-to-* helpers — implementer's choice; document the choice in the PR description since both mixins are plausible per the existing file-header docstrings).
- [ ] Default stub direction is derived from the pin's `angle` (library-space outward direction) rotated by the symbol's `rotation`, not a fixed side — verified against at least 2 symbol rotations (0° and 90°) in tests.
- [ ] Default stub length is `2.54` mm (one grid step), overridable via `length=`.
- [ ] The #4143 repro geometry, when constructed via two `add_stub_label()` calls instead of hand-rolled `Wire`/`Label` objects, produces **zero** `collinear_net_conflict` issues in `validate()` and **distinct** nets for the two labels in `extract_netlist()` — i.e. the helper actively avoids creating the #4143 failure pattern, it doesn't just fail to make it worse.
- [ ] A forced-collision fixture (two symbols placed close enough that the default-direction stub for the second pin would collinearly overlap the first pin's stub) exercises the alternative-candidate search and succeeds by picking a non-colliding candidate (assert the candidate actually differs from the naive primary, and assert no `collinear_net_conflict` results).
- [ ] A dense/no-solution fixture (stubs boxed in on all 4 sides by same-length foreign-net wires) raises a clear exception identifying the pin, net, and tried candidates — verified with `pytest.raises`.
- [ ] Return value is `(Wire, Label)` (or a documented equivalent named-tuple) — verified by asserting the returned objects are present in `sch.wires` / `sch.labels` after the call.
- [ ] No regression in `tests/test_netlist_query.py`, `tests/test_schematic_models.py` (the `collinear_net_conflict` suite), `tests/test_pin_nets_stub_collision.py`.

### Test Plan

All fixtures synthetic (no chorus/softstart dependency — those repos are local-only per project memory).

- [ ] **Happy path, single pin**: one symbol, one pin, `add_stub_label(sym, "1", "NET_A")` in open space — assert a `Wire` and `Label` are appended, the wire's far endpoint carries the label, `validate()` has no `collinear_net_conflict`.
- [ ] **Direction from pin angle, 0° symbol rotation**: a two-pin passive (`Device:R`-shaped synthetic symbol via `make_simple_symbol`, matching the pattern in `tests/test_netlist_query.py::_two_stub_schematic`) with pins at `angle=0` (right-facing, pin extends left/outward) and `angle=180` — assert the computed stub directions differ appropriately (outward, not into the symbol body).
- [ ] **Direction from pin angle, 90° symbol rotation**: same symbol placed with `rotation=90` — assert the stub direction rotates with the symbol (this is the critical regression guard for the "rotation-aware" requirement; a naive implementation that ignores `symbol.rotation` will pass the 0° case but fail this one).
- [ ] **#4143 repro via `add_stub_label`**: two symbols ~8mm apart (mirroring `test_repro_sketch_r1_r2_collinear_stubs` in `tests/test_netlist_query.py`), each pin stub-labeled with a *different* net using `add_stub_label` (not hand-rolled `Wire`/`Label`) — assert `validate()` returns zero `collinear_net_conflict` issues and `extract_netlist()` shows the two labels resolving to **distinct** nets (i.e. the helper's collision avoidance actually prevented the merge that the hand-rolled version triggered).
- [ ] **Forced-collision alternative-selection**: construct a schematic where the default-direction candidate for a second `add_stub_label()` call would `wire_segments_connect()` a first stub carrying a different net — assert the call succeeds (no exception), the resulting stub differs from the naive/default candidate, and `validate()` shows zero `collinear_net_conflict`.
- [ ] **No-solution raises**: box a pin in on all sides (primary + all fallback candidates pre-occupied by differently-named foreign wires) — assert `add_stub_label()` raises (`ValueError` or the chosen dedicated exception type) with a message identifying the pin/net/candidates, and assert nothing was appended to `sch.wires`/`sch.labels` on the raising path (no partial state).
- [ ] **Same-net overlap is not a collision**: two stubs for the *same* net name that would geometrically overlap — assert this does NOT trigger the alternative search / does NOT raise (mirrors the existing "no false positive on same-net overlap" case from #4143's own test plan).
- [ ] Run: `uv run pytest tests/test_netlist_query.py tests/test_schematic_models.py tests/test_pin_nets_stub_collision.py -v`, plus full `uv run pytest` for regression safety.

### Affected Files

- `src/kicad_tools/schematic/models/elements_mixin.py` or `src/kicad_tools/schematic/models/wiring_mixin.py` — new `add_stub_label()` method (implementer picks the mixin; see Acceptance Criteria).
- `src/kicad_tools/schematic/models/wire_geometry.py` — reuse `wire_segments_connect()` as-is; no changes expected unless the candidate search needs a small helper (keep additions here if so, to preserve the "single tested primitive" invariant the module's docstring establishes).
- `tests/test_netlist_query.py` and/or `tests/test_schematic_models.py` — new test class(es) for `add_stub_label()` (implementer's choice which file; `test_schematic_models.py` already hosts the `collinear_net_conflict` suite from #4143 and may be the more natural home for the "no collision" and "raises" cases, while `test_netlist_query.py` may be more natural for the netlist-outcome assertions).

### Out of Scope (this PR)

- Migrating `_stub_helpers.py`'s existing callers (`LDOBlock`, `GateDriverBlock`, `HalfBridge`, `DebugHeader`) onto the new `add_stub_label()` — separate follow-up, has its own test suite (`tests/test_pin_nets_stub_collision.py`) and center-heuristic semantics to preserve or consciously change.
- Non-axis-aligned (arbitrary-angle) symbol rotations — snap to nearest cardinal direction is sufficient; full arbitrary-angle stub routing is not requested and no existing wiring helper supports it either.
- Multi-segment (orthogonal/Manhattan) stub routing for the collision-avoidance fallback — the bounded candidate set (longer/shorter/perpendicular-jog/opposite) is single-segment only, consistent with the "stub" framing (a Manhattan-routed escape is a different, larger feature — see `wire_pin_to_point`/`_route_orthogonal` for the existing multi-segment primitive if a future issue wants to combine them).

### Duplicate check

Ran `.loom/scripts/check-duplicate.sh --include-merged-prs` — the only match was the issue matching itself (100% similarity, #4161 vs #4161); no other open issue or merged PR title/body overlaps. Confirmed no other issue references `add_stub_label` by name.

rjwalters:
## Summary

Adds a built-in `Schematic.add_stub_label(symbol, pin, net)` helper that draws a short stub wire outward from a symbol pin and anchors a net label at its far endpoint, with **rotation-aware** outward direction and a **bounded, deterministic collision search** that avoids collinearly overlapping a neighboring stub carrying a *different* net — the silent net-merge failure mode fixed by #4143. This generalizes and supersedes the hand-rolled stub+label idioms in `blocks/motor.py` and `blocks/_stub_helpers.py`.

Item 3 of #4143 (deferred by the curator as purely additive).

## Changes

- **`elements_mixin.py`**: new `add_stub_label(symbol, pin, net, *, length=2.54, direction=None, snap=True) -> tuple[Wire, Label]`. Chose `SchematicElementsMixin` (next to `add_wire`/`add_label`) over `wiring_mixin.py` since the helper is fundamentally an element-creation primitive that composes those two methods.
  - Outward direction computed from `Pin.angle` (opposite of into-body angle) rotated by `SymbolInstance.rotation` via the same CCW-then-Y-flip transform `pin_position()` uses, snapped to the nearest cardinal.
  - Deterministic candidate order: outward @ `length`, outward @ `length*2`, outward @ `length/2`, the two perpendicular jogs @ `length`, then opposite @ `length` (<=6 candidates).
  - Collision = `wire_segments_connect()` overlap/T-touch with an existing wire independently carrying a *different* named net (net-name extraction mirrors `_wire_net_names` locally to avoid depending on the validation mixin's private helper). Same-net and unnamed overlaps are allowed.
  - Passes `warn_on_collision=False` to `add_wire` (does its own pre-flight search); snaps geometry once so the search and placed wire share identical coordinates. On the no-solution path raises before appending anything (no partial state).
- **`symbol.py`**: extract `SymbolInstance.find_pin()` from `pin_position()` so the helper reuses the existing exact/unit-aware/alias/fuzzy pin lookup instead of reimplementing it; `pin_position()` now calls it.
- **`exceptions.py`**: new `StubPlacementError(ValueError)` (subclasses `ValueError` for continuity with the `_emit_pin_net_stub` precedent) carrying `pin_position`, `net`, and `tried_endpoints` in the message.
- **`tests/test_schematic_models.py`**: new `TestAddStubLabel` (11 cases).
- Two `# type: ignore[attr-defined]` on mixin references to `self.wires`/`labels`/... (genuine false positives — those live on the concrete `Schematic`; the same references pre-exist in the mypy baseline). No baseline changes.

`_stub_helpers.py` callers are intentionally **not** migrated (out of scope per the issue).

| `add_stub_label` exists on `Schematic` | ✅ | `SchematicElementsMixin.add_stub_label`; `test_happy_path_single_pin` |
| Direction from pin angle + rotation (0° & 90°) | ✅ | `test_direction_from_pin_angle_zero_rotation`, `test_direction_rotates_with_symbol_90` |
| Default length 2.54mm, `length=` override | ✅ | `test_default_length_is_one_grid_step` |
| #4143 repro via helper → 0 conflicts, distinct nets | ✅ | `test_4143_repro_via_add_stub_label_no_merge` (asserts `validate()` clean + `extract_netlist` distinct) |
| Forced-collision picks non-naive alternative | ✅ | `test_forced_collision_selects_alternative` |
| No-solution raises, no partial state | ✅ | `test_no_solution_raises_and_leaves_no_partial_state` |
| Return `(Wire, Label)` present in `sch.wires`/`sch.labels` | ✅ | `test_returns_wire_and_label_types`, `test_happy_path_single_pin` |
| Same-net overlap allowed (no raise/search) | ✅ | `test_same_net_overlap_is_allowed` |
| No regression in named suites | ✅ | 288 passed across the 3 suites |

## Test Plan

- `uv run pytest tests/test_schematic_models.py tests/test_pin_nets_stub_collision.py tests/test_netlist_query.py` → **288 passed**
- `TestAddStubLabel` (11) + `tests/test_pin_nets_stub_collision.py` (11) → **22 passed**
- `ruff check` + `ruff format --check` on changed files → clean
- `uv run --extra dev python scripts/ci/check_mypy_baseline.py` → **exit 0, 0 NEW errors** (pre-existing "39 no longer occur" warning is unchanged from clean main)

Closes #4161

### DIFF
--- a/src/kicad_tools/schematic/exceptions.py
+++ b/src/kicad_tools/schematic/exceptions.py
@@ -157,6 +157,44 @@ class SymbolNotFoundError(ValueError):
         super().__init__("".join(msg_parts))
 
 
+class StubPlacementError(ValueError):
+    """Raised when a collision-aware stub+label cannot be placed.
+
+    :meth:`SchematicElementsMixin.add_stub_label` searches a small, bounded
+    set of candidate stub geometries (outward, longer, shorter, perpendicular
+    jogs, opposite) and places the first one that does not collinearly overlap
+    or T-touch an existing wire carrying a *different* named net.  When every
+    candidate would collide it raises this error rather than silently placing
+    a stub that KiCad would merge into a foreign net (the issue #4143 failure
+    mode).  The message identifies the pin's absolute position, the requested
+    net, and every candidate endpoint that was tried so a generator author can
+    see why placement failed and adjust the layout.
+
+    Subclasses ``ValueError`` for continuity with the ``_emit_pin_net_stub``
+    precedent in ``blocks/_stub_helpers.py`` (which raises ``ValueError`` when
+    both of its candidate sides collide).
+    """
+
+    def __init__(
+        self,
+        pin_position: tuple[float, float],
+        net: str,
+        tried_endpoints: list[tuple[float, float]],
+    ):
+        self.pin_position = pin_position
+        self.net = net
+        self.tried_endpoints = tried_endpoints
+
+        candidates = ", ".join(f"({x}, {y})" for x, y in tried_endpoints)
+        super().__init__(
+            f"Cannot place stub+label for net '{net}' at pin "
+            f"{pin_position}: every candidate stub endpoint collinearly "
+            f"overlaps or T-touches an existing wire carrying a different "
+            f"net. Tried endpoints: {candidates}. Move the neighboring "
+            f"symbol/wire or pass an explicit direction=/length= to resolve."
+        )
+
+
 class LibraryNotFoundError(FileNotFoundError):
     """Raised when a KiCad library file cannot be found."""
 
--- a/src/kicad_tools/schematic/models/elements_mixin.py
+++ b/src/kicad_tools/schematic/models/elements_mixin.py
@@ -7,9 +7,11 @@ and grid snapping functionality.
 
 from __future__ import annotations
 
+import math
 import warnings
 from typing import TYPE_CHECKING
 
+from ..exceptions import StubPlacementError
 from ..grid import is_on_grid, snap_to_grid
 from ..logging import _log_debug, _log_info
 from .elements import (
@@ -22,9 +24,44 @@ from .elements import (
     Wire,
 )
 from .symbol import SymbolDef, SymbolInstance
+from .wire_geometry import wire_segments_connect
 
 if TYPE_CHECKING:
     pass
+
+
+# Mapping of the four cardinal stub directions to their outward unit vector in
+# schematic (Y-down) coordinates: +x is right, +y is down.
+_DIRECTION_VECTORS: dict[str, tuple[float, float]] = {
+    "right": (1.0, 0.0),
+    "up": (0.0, -1.0),
+    "left": (-1.0, 0.0),
+    "down": (0.0, 1.0),
+}
+
+# Label rotation (degrees) so the label text reads outward along each stub
+# direction, matching KiCad's convention (0=text extends right, 90=up, ...).
+_DIRECTION_LABEL_ROTATION: dict[str, float] = {
+    "right": 0.0,
+    "up": 90.0,
+    "left": 180.0,
+    "down": 270.0,
+}
+
+# Perpendicular jog directions tried when the outward candidate collides.
+_PERPENDICULAR: dict[str, tuple[str, str]] = {
+    "right": ("up", "down"),
+    "left": ("up", "down"),
+    "up": ("right", "left"),
+    "down": ("right", "left"),
+}
+
+_OPPOSITE_DIRECTION: dict[str, str] = {
+    "right": "left",
+    "left": "right",
+    "up": "down",
+    "down": "up",
+}
 
 
 class SchematicElementsMixin:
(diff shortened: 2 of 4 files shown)


rjwalters:
## Judge Review: APPROVED

Reviewed PR #4171 against the curated acceptance criteria in #4161. This is a clean, well-scoped implementation of the collision-aware `add_stub_label` helper.

### Correctness
- **Rotation-aware outward direction** (`_pin_outward_direction`) is correct. I independently reproduced the transform: it applies the *exact same* CCW-rotation-then-Y-flip that `pin_position()` uses, but to the pin's outward direction vector (`pin.angle + 180`), then snaps to nearest cardinal via max dot-product. Verified: angle=0/rot=0 -> left; angle=180 -> right; angle=90 -> down; angle=0/rot=90 -> down; and the full 0/180/270 rotation set. The critical rotation guard (`test_direction_rotates_with_symbol_90`) directly asserts `_pin_outward_direction(r90, "1") == "down"`, which a rotation-ignoring implementation would fail.
- **Bounded deterministic search** (<=6 candidates: outward@L, outward@2L, outward@L/2, two perpendicular jogs, opposite@L) with a `POINT_TOLERANCE` floor guard and a degenerate-endpoint guard. Ordering is stable for reproducible generator output.
- **Collision semantics** are exactly right: `_stub_would_collide` uses `wire_segments_connect()` against wires that independently carry a *different* named net; `names <= {net}` correctly allows same-net and unnamed overlaps. Net-name extraction mirrors `_wire_net_names` locally (labels/global/hier/power) rather than reaching into the validation mixin's private helper — good separation.
- **No partial state** on the no-solution path: candidates are validated before any `add_wire`/`add_label` append; raises `StubPlacementError` first. `test_no_solution_raises_and_leaves_no_partial_state` asserts wire/label counts are unchanged.
- **Geometry consistency**: snaps the stub start/end once (`snap=False` passed downstream) so the collision search and the placed wire share identical coordinates — avoids the subtle bug where `add_wire`'s own snapping shifts an endpoint onto a foreign wire the search didn't see.
- `SymbolInstance.find_pin()` extraction is a clean refactor: `pin_position()` now delegates to it, no behavior change, reuses the exact/unit-aware/alias/fuzzy lookup.

### Core #4143 value verified
`test_4143_repro_via_add_stub_label_no_merge` asserts BOTH zero `collinear_net_conflict` in `validate()` AND distinct nets via `get_net_for_pin("R1","2") != get_net_for_pin("R2","1")` in the extracted netlist. `test_forced_collision_selects_alternative` proves the search picks a candidate differing from the naive primary. This is the helper actively avoiding the silent net-merge, not merely failing to worsen it.

### Quality gates (verified locally in worktree + CI)
- **288 tests pass** across test_schematic_models + test_pin_nets_stub_collision + test_netlist_query; 11 new `TestAddStubLabel` cases cover every acceptance criterion.
- **ruff check + format**: clean on all changed files.
- **mypy baseline**: 0 NEW errors (1475 within 1514 allowed). The 5 `# type: ignore[attr-defined]` on `self.wires`/`labels`/`global_labels`/`hier_labels`/`power_symbols` are genuine false positives — those collections live on the concrete `Schematic`, invisible from the mixin body, matching the pre-existing pattern elsewhere in this file. Not masking a real typing problem (an unused/mistargeted ignore would itself trip `unused-ignore`).
- **CI: all 16 checks green** (Test 17m2s, Board 05 routing regression 19m36s incl. the #4160/#4168 DRC-timeout fix, all board E2E + LVS suites). mergeStateStatus **CLEAN**, MERGEABLE.
- `_stub_helpers.py` caller migration correctly deferred as out-of-scope per the issue.

Signature, defaults, return type, exception hierarchy, and mixin placement all match the curated spec. No blockers. Approving.

-- Loom Judge

Issue #4173, opened 2026-07-14
route: Phase 2c — route-auto/orchestrator parity for boundary-stub reconnection (follow-up to #4170)


rjwalters:
## Context

Follow-up to **#4170** (region-bounded routing Phase 2b-1, Autorouter MVP). See the architect design pass on #4170 (issues/4170#issuecomment-4974546162) for the full rationale — this issue implements the "Phase 2c" unit.

Pad-only targeting lives in **two independent surfaces** (the #3428 foot-gun, warned about in the orchestrator's own docstring):
- `Autorouter.nets`/`self.pads` — fixed by #4170.
- `RoutingOrchestrator._build_net_target_positions` (`router/orchestrator.py:240-306`) — the independent `route-auto` pad-sourcing path. **Still pad-only after #4170.**

## Scope

Bring `route-auto --region` to parity with `route --region` by:
1. Adding the identical target-merge shim to `RoutingOrchestrator._build_net_target_positions` at its `result.setdefault(net, []).append(...)` points (orchestrator.py:275/294), consuming the SAME `detect_boundary_stub_terminals` producer from #4172 (do NOT re-derive stub geometry).
2. Relaxing the `route-auto` reachability gate to reclassify stub-reconnectable nets as routable, mirroring #4170's relaxation of `_parse_and_apply_region`.
3. Removing the loud "Phase 2c not yet wired" fail-fast that #4170 ships as the anti-foot-gun gate (and the test asserting it).

## Acceptance

- [ ] `route-auto --region` reconnects a boundary stub identically to `route --region` (parity test on the same synthetic fixture from #4170).
- [ ] Zero copper modified outside the region (byte-identical assertion).
- [ ] Hole-to-hole floor respected for any new via near the reconnection joint.
- [ ] The #4170 "Phase 2c not yet wired" gate + its test are removed (route-auto no longer fails fast on stub nets).

## Notes

- The whole point of the shared producer (#4172) is that this issue is a thin adapter, NOT a re-implementation. If it starts looking like a second copy of the detector, stop — reuse `stub_terminals.py`.
- Affected files (starting point — builder must verify): `src/kicad_tools/router/orchestrator.py`, the route-auto reachability gate, `src/kicad_tools/mcp/tools/routing.py` (if the MCP route-auto surface also gates on reachability), `tests/router/test_region_bounded.py`.

Dependencies: #4172 (producer), #4170 (Autorouter MVP + the deferral gate this removes).

[...]

Issue #4170, opened 2026-07-14
route: region-bounded routing Phase 2b — bare mid-trace stub-endpoint reconnection


rjwalters:
## Context

Follow-on from #4148 (Phase 2a, merged in PR #4169), which added `--region` spatial bounding to `route`/`route-auto` (outside copper as fixed obstacles, hole-to-hole floor threaded into the main A* via path, coordinate parity with `pcb strip --region`). Phase 2a deliberately excluded bare-stub reconnection — the crux of the motivating chorus#16 use case — and the #4148 curation identified it as genuinely new plumbing worth its own issue.

## The gap

The router (`Autorouter.nets` in `src/kicad_tools/router/core.py`, `RoutingOrchestrator.route_net` in `src/kicad_tools/router/orchestrator.py`) only ever connects **pads**. There is no concept of routing to a bare copper stub endpoint. After `pcb strip --region` (Phase 1, PR #4147) clips segments at the region boundary, the leftover stubs load fine as fixed obstacles via `load_pcb_for_routing(load_existing_routes=True)`, but nothing makes their exposed endpoints routable targets.

So today, region-bounded routing (Phase 2a) only works when a net's boundary is pad- or via-coincident. A mid-trace stub endpoint left by a boundary clip cannot be reconnected — which is exactly what the chorus#16 J2 col-1 escape-channel re-dress needs.

## Requested feature

Synthesize a virtual pad/terminal at each in-region stub endpoint so the A* router can target it:
1. Detect stub endpoints: copper segment endpoints that lie inside (or on) the region boundary and are not already a pad/via.
2. Carve the stub-endpoint cell out of the outside-region obstacle marking so the router can reach it.
3. Inject the synthesized terminal into `router.nets` for the correct net, as a routable source/target.
4. Reconnect: the router completes a path from the stub endpoint to the net's other in-region endpoints, respecting outside copper as obstacles and the hole-to-hole floor.

## Acceptance sketch

On a synthetic routed board: strip a region with `pcb strip --region` (leaving mid-trace stubs at the boundary), then `route --region <same box> --nets <clipped nets>` reconnects the stubs with zero copper modified outside the region and no DRC regression. The chorus#16 J2 col-1 escapes (4 different-net hole-to-hole pairs) clear to >=0.5mm headlessly. Synthetic-fixture tests only (chorus is local-only, not in CI); follow `tests/router/test_region_bounded.py` conventions.

## Notes

[...]

PR #4174, opened 2026-07-14
feat(router): add pure boundary stub-terminal detector

### LINKED ISSUE #4172: route: Phase 2b-0 — StubTerminal abstraction + boundary-stub detector (prerequisite for #4170)
## Context

Prerequisite spike for **#4170** (region-bounded routing Phase 2b-1, bare mid-trace stub-endpoint reconnection). See the architect design pass on #4170 (issues/4170#issuecomment-4974546162) for the full rationale — this issue implements the "Phase 2b-0" unit from that decomposition.

Phase 2b needs the router to target bare copper stub endpoints (left by `pcb strip --region` boundary clips), which no existing abstraction provides. The design's anti-foot-gun keystone is a **single shared producer** that BOTH pad-sourcing surfaces (`Autorouter` and `RoutingOrchestrator`) consume, so stub geometry is never re-derived independently (the #3428 foot-gun).

## Scope (small, no user-facing behavior change)

Add `src/kicad_tools/router/stub_terminals.py`:

1. A `StubTerminal` dataclass: `net_id: int`, `net_name: str`, `x: float`, `y: float`, `layer: CopperLayer`, plus provenance `source_segment_uuid: str | None`, `boundary_edge` for the detector's own correctness checks. It is **route-scoped and ephemeral** — never a `Pad`, never inserted into `Autorouter.pads`/`nets`.
2. `detect_boundary_stub_terminals(...) -> dict[int, list[StubTerminal]]` implementing the four-part detection spec from the #4170 design:
   - endpoint on region-boundary line within `EPSILON = 0.01mm` (reuse the established constant from `core.py::_index_block_internal_traces`)
   - the segment's OTHER endpoint strictly OUTSIDE the region (surviving clipped stubs always run boundary → outside; confirmed at `schema/pcb.py:5200-5227`)
   - endpoint NOT coincident (within EPSILON) with any pad/via center (those route "for free" under Phase 2a's inclusive box test)
   - net has pad(s) outside AND >=1 pad inside (exactly the nets Phase 2a currently fails)

No router/CLI wiring in this issue — that is #4170 (Autorouter) and the Phase 2c follow-up (orchestrator).

## Acceptance

- [ ] Unit tests feeding hand-built clipped segments assert each positive detection case.
- [ ] Negative cases rejected: pad-coincident endpoint, outside-outside segment, all-inside net, near-boundary-but-not-on-line endpoint.
- [ ] Producer is pure (no grid/router state) so both future surfaces can call it identically.
- [ ] `EPSILON` reused, not re-invented.

## Notes

- `strip_traces` builds an ephemeral `clipped_geometry` UUID map (`schema/pcb.py:5166/5224`) but does NOT persist it, so the detector must re-derive from loaded segments — this is by design, not a `strip` gap.
- Affected files (starting point — builder must verify): NEW `src/kicad_tools/router/stub_terminals.py`, NEW `tests/router/test_stub_terminals.py`.

Dependencies: none (pure prerequisite). Blocks: #4170.

rjwalters:
## Summary

Phase 2b-0 prerequisite for #4170. Adds a new **pure** module,
`src/kicad_tools/router/stub_terminals.py`, that detects bare copper stub
endpoints left on a routing-region boundary by `pcb strip --region` clipping.
Such endpoints have no pad and are invisible to Phase 2a's inclusive-box
routing, so nets that own pads both inside and outside the region cannot be
reconnected to pre-existing external copper.

The module is the **single shared producer** both future pad-sourcing surfaces
(`Autorouter` and `RoutingOrchestrator`) will consume, so stub geometry is
never re-derived independently (the #3428 foot-gun). No router/grid/CLI wiring
is added here — that is #4170 (Autorouter) and #4173 (orchestrator).

## Changes

- NEW `src/kicad_tools/router/stub_terminals.py`:
  - `StubTerminal` dataclass (`net_id`, `net_name`, `x`, `y`, `layer: CopperLayer`,
    provenance `source_segment_uuid`, `boundary_edge`) — route-scoped/ephemeral,
    never a `Pad`.
  - `detect_boundary_stub_terminals(segments, pads, region)` — pure, no
    grid/router/PCB state; both future callers can invoke it identically by
    adapting their data into plain `StubSegment` / `PadLocation` / `RegionBox`.
  - Module-level `EPSILON = 0.01` mirroring the function-local constant in
    `core.py::_index_block_internal_traces` (that value is a local variable, not
    importable — matched by value + semantic role, with a comment).
- NEW `tests/router/test_stub_terminals.py`: hand-built clipped segments fed
  directly to the detector (no chorus fixture, no PCB round-trip).

| Positive detection cases asserted | ✅ | right/left/top/bottom edges, within-epsilon, multi-net grouping |
| Negative: pad-coincident endpoint rejected | ✅ | `test_rejects_pad_coincident_endpoint` |
| Negative: outside-outside segment rejected | ✅ | `test_rejects_outside_outside_segment` |
| Negative: all-inside net rejected | ✅ | `test_rejects_all_inside_net` (+ all-outside half) |
| Negative: near-boundary-but-not-on-line rejected | ✅ | `test_rejects_near_boundary_but_not_on_line_endpoint` + just-beyond-epsilon |
| Producer is pure (no grid/router state) | ✅ | function takes plain `StubSegment`/`PadLocation`/`RegionBox` args only |
| `EPSILON` reused (value 0.01), not re-invented | ✅ | module-level `EPSILON = 0.01` with comment referencing `core.py` |

## Test Plan

- `uv run pytest tests/router/test_stub_terminals.py -q` → 14 passed
- `uv run ruff check` + `ruff format --check` on both changed files → clean
- `uv run --extra dev python scripts/ci/check_mypy_baseline.py` → 0 new errors
  (1475 within the 1514 baseline)

Closes #4172

### DIFF
--- a/src/kicad_tools/router/stub_terminals.py
+++ b/src/kicad_tools/router/stub_terminals.py
@@ -1,0 +1,283 @@
+"""Boundary stub-terminal detection for region-bounded routing (Phase 2b-0).
+
+When ``pcb strip --region`` clips traces at a routing-region boundary
+(``schema/pcb.py`` boundary-crossing branch, lines ~5200-5227), every
+boundary-crossing segment is rewritten so that its *inside* endpoint moves
+onto the region boundary while its *outside* endpoint is retained unchanged.
+The result is a bare copper "stub" that runs boundary -> outside with no pad
+at the boundary end. Phase 2a's inclusive-box routing has no abstraction for
+these bare stub endpoints, so nets that own pads both inside and outside the
+region cannot be reconnected to the pre-existing external copper.
+
+This module provides the *single shared producer* that both pad-sourcing
+surfaces (``Autorouter`` and ``RoutingOrchestrator``) will consume in later
+phases (#4170 / #4173), so stub geometry is never re-derived independently
+(the #3428 foot-gun). It is intentionally **pure**: it holds no grid or router
+state and takes all segment/pad/net data as plain arguments.
+
+The clipped-geometry UUID map that ``strip_traces`` builds is ephemeral and is
+not persisted on ``PCB``, so this detector deliberately re-derives stub
+endpoints from *loaded* segments rather than querying the PCB for "which
+segments are stubs" -- this is by design (see the note in #4172 / #4170).
+
+``StubTerminal`` objects are **route-scoped and ephemeral**: they are never a
+``Pad`` and must never be inserted into ``Autorouter.pads`` / ``nets``.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from enum import Enum
+
+from kicad_tools.core.types import CopperLayer
+
+# Proximity tolerance in millimeters. Mirrors the local ``EPSILON = 0.01``
+# used by ``core.py::_index_block_internal_traces`` (defined there as a
+# function-local variable, so it cannot be imported by name). Matching the
+# value and its semantic role -- pad/endpoint proximity tolerance -- keeps the
+# stub detector consistent with the router's existing coincidence checks.
+EPSILON = 0.01
+
+
+class BoundaryEdge(Enum):
+    """Which edge of the rectangular routing region a stub endpoint lies on.
+
+    Recorded on :class:`StubTerminal` for the detector's own correctness
+    checks and downstream diagnostics; it is not a routing input.
+    """
+
+    LEFT = "left"
+    RIGHT = "right"
+    TOP = "top"
+    BOTTOM = "bottom"
+
+
+@dataclass(frozen=True)
+class RegionBox:
+    """Axis-aligned routing region in board-relative millimeters.
+
+    Endpoints are classified against ``[x1, x2] x [y1, y2]``. The box is
+    normalized on construction so ``x1 <= x2`` and ``y1 <= y2`` regardless of
+    the order the corners are supplied in.
+    """
+
+    x1: float
+    y1: float
+    x2: float
+    y2: float
+
+    def __post_init__(self) -> None:
+        if self.x1 > self.x2:
+            lo, hi = self.x2, self.x1
+            object.__setattr__(self, "x1", lo)
+            object.__setattr__(self, "x2", hi)
+        if self.y1 > self.y2:
+            lo, hi = self.y2, self.y1
+            object.__setattr__(self, "y1", lo)
+            object.__setattr__(self, "y2", hi)
+
+    def contains(self, x: float, y: float) -> bool:
+        """Inclusive membership test (matches Phase 2a's inclusive box)."""
+        return self.x1 <= x <= self.x2 and self.y1 <= y <= self.y2
+
+    def contains_strict(self, x: float, y: float, epsilon: float = EPSILON) -> bool:
+        """Strictly-inside test: inside the box and not on/near any edge.
+
+        Used to decide whether the *other* endpoint of a stub segment is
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Phase 2b-0 is a clean, well-scoped, genuinely pure detector. Verified against issue #4172's acceptance criteria and the architect's four-part detection spec.

## Verification

**Purity (the anti-#3428 keystone)** — Confirmed. `detect_boundary_stub_terminals(segments, pads, region)` takes only plain-data inputs (`StubSegment` / `PadLocation` / `RegionBox`); no grid, router, or `PCB` state is imported or referenced. Both future callers can invoke it identically. The Curator's warning against reaching for `Autorouter._find_pads_near` was heeded — proximity matching is reimplemented locally in `_coincident_with_pad`.

**Four-part spec** — All correct and each isolated by a dedicated negative test:
- Part 1 (endpoint on boundary *segment*, within EPSILON): `_edge_for_point` correctly bounds by the box extent (`on_x_span`/`on_y_span`), not the infinite line. I verified directly that a point on the extended right-edge line but far above the box (10, 50) is rejected.
- Part 2 (other endpoint strictly outside): formulated as `not contains(ox,oy) and not _edge_for_point(ox,oy)`. Verified boundary→inside, boundary→boundary, and outside-outside crossing segments are all rejected; a genuine boundary→outside stub yields exactly one terminal (no double-count, since the outside endpoint fails Part 1).
- Part 3 (not pad/via-coincident): squared-distance check within EPSILON.
- Part 4 (net straddles: pad(s) outside AND ≥1 inside): per-net inside/outside census with inclusive-box membership matching Phase 2a.

**EPSILON** — Module-local `EPSILON = 0.01` with a comment referencing `core.py::_index_block_internal_traces`. Confirmed that anchor is `core.py:1759` (`EPSILON = 0.01  # mm tolerance...`), a function-local variable — correctly mirrored by value + role rather than imported.

**RegionBox corner normalization** — `__post_init__` reads both corners into locals before writing (no swap bug). Verified for all orderings including the mixed case `RegionBox(10, 0, 0, 10)` → `(0, 0, 10, 10)`.

**Geometry robustness (extra checks I ran)** — Corner endpoint at (0,0) deterministically classified (LEFT), zero-length degenerate on-boundary segment rejected, per-net grouping preserves input-segment order for determinism.

## Test Execution
- `pytest tests/router/test_stub_terminals.py -q` → **14 passed** (5 positive incl. all 4 edges/within-epsilon/multi-net, 6 negative isolating each part, + normalization/inclusive-boundary/empty-input guards)
- `ruff check` + `ruff format --check` on both files → clean
- `check_mypy_baseline.py` → **0 new errors** (1475 within 1514 baseline)
- All 16 GitHub CI checks pass; `mergeStateStatus: CLEAN`

## Non-blocking notes (not tracked as issues — trivially addressable in #4170 when the API firms up)
- `RegionBox.contains_strict()` is currently unused dead code (Part 2 uses the `not contains and not on-edge` formulation instead). It's a plausibly-useful helper for future callers, so leaving it is defensible, but it's untested.
- `StubTerminal` is a mutable `@dataclass` while the three input dataclasses are `frozen=True`. The issue #4172 commenter suggested freezing the output too; given it's route-scoped/ephemeral and never shared mutably, this is a reasonable-to-defer polish, not a correctness issue.

Solid foundation for #4170 / #4173. Nice work.

PR #4169, opened 2026-07-14
feat(route): add --region spatial routing bound (Phase 2a of #4136)

rjwalters:
## Summary

Phase 2a of region-bounded routing (#4136). Adds `--region x1,y1,x2,y2` to
`route` and `route-auto`, confining all new routing to an axis-aligned
board-relative box that matches `pcb strip --region`'s coordinate convention
exactly (PR #4147). Everything outside the box -- empty grid AND existing
copper -- is treated as a fixed obstacle, so the router never adds, modifies,
or removes copper outside the region.

**Phase 2b (bare mid-trace stub-endpoint reconnection) is explicitly OUT of
scope and deferred to a follow-up issue.** Stubs that coincide with a pad/via
already work (pads/vias are routing targets); bare boundary stubs are not.

## Changes

- **`router/grid.py`** — `RoutingGrid.mark_region_bound(x1,y1,x2,y2)`: marks
  every routable-layer cell OUTSIDE the world box as a fixed obstacle (the
  complement of a keepout), mirroring to the C++ grid via `mark_blocked`. No
  `router/cpp/` edits — the pathfinder already hard-blocks net-0 blocked cells
  in both standard and negotiated modes.
- **`router/io.py`** — `load_pcb_for_routing(region=...)`: after
  `load_existing_routes` loads outside copper as obstacles, marks the
  outside-region cells; the box is board-relative and converted to world by
  adding the grid origin (== `PCB._board_origin`).
- **`cli/route_cmd.py`** — parse/validate `--region` (mirrors the `strip
  --region` CLI checks), imply `--preserve-existing` + `--no-cache` (the route
  cache key ignores the region/skip set), auto-skip nets with no in-region
  work, and fail fast with a clear per-net message when a routed net has a pad
  outside the region. Threaded `region=` into all five `load_pcb_for_routing`
  call sites.
- **`router/core.py`** — thread `min_hole_to_hole` into the MAIN router's via
  validation (`_TraceResolverTransaction.validate_committed_geometry`). The
  drill-to-drill floor was previously only applied by `escape.py` /
  `diffpair_routing.py`; the main A* path now rejects a committed via whose
  drill sits within the fab floor of any pad/via drill, via the canonical
  `via_clearance.drill_hole_to_hole_clear`.
- **`cli/parser.py`** — `--region` on `route` + `route-auto`; help text
  explicitly distinguishes the spatial bound from the unrelated
  `--region-parallel` family. argparse resolves the exact `--region` string to
[...]

PR #4179, opened 2026-07-14
feat(router): reconnect bare boundary stubs in region-bounded routing (Phase 2b-1)

### LINKED ISSUE #4170: route: region-bounded routing Phase 2b — bare mid-trace stub-endpoint reconnection

rjwalters:
## Summary

Phase 2b-1 of region-bounded routing (#4170): the Autorouter now reconnects **bare mid-trace stub endpoints** left by `pcb strip --region`. `kct route --region <box>` reconnects a clipped net's in-region pad to the copper stub sitting on the region boundary — the crux of the chorus#16 escape re-dress that Phase 2a (`--region`, #4169) deliberately deferred.

This consumes the merged #4172 shared detector (`detect_boundary_stub_terminals`) as the single source of stub geometry (the anti-#3428 keystone) and does **not** touch `router/cpp/`.

**Phase 2c (route-auto / orchestrator parity) is explicitly deferred to #4173.** `route-auto --region` on a stub-endpoint net fails fast with a clear Phase 2c message rather than silently misbehaving.

## Changes

- **CLI gate relaxation** (`cli/route_cmd.py`): `_parse_and_apply_region` runs the shared detector and reclassifies "pad outside + same-net boundary stub" nets as routable instead of hard-failing; nets with an outside pad and NO stub still fail fast.
- **Threading** (`router/io.py`): `load_pcb_for_routing(stub_terminals=...)` carves each stub tip open, stores the terminals + region box on the router, and prunes outside-region pads from stub nets (the stub tip is the reachable target). Preserves the outside stub copper even though the net is routed.
- **Carve-out** (`router/grid.py`): `reopen_stub_terminal` re-asserts `cell.net` on the exact stub tip cell (Python grid **and** C++ mirror) so A* lands on the true tip as an own-net-reachable target while foreign nets still see a hard obstacle. Fixes the foreign-halo-clobber case (verified: without it, A* lands on a wrong nearby cell).
- **Target-merge shim** (`router/core.py`): `StubTerminal`s become route-scoped/ephemeral synthetic `Pad`s added as additive per-net targets at `route_net` and the negotiated per-net chokepoint (`_route_net_negotiated`). Never inserted into `self.pads` / `self.nets`.
- **route-auto deferral gate** (`mcp/tools/routing.py`): stub-endpoint nets fail fast with a Phase 2c (#4173) message; non-stub outside-pad nets get the plain unreachable message.
- **Judge nits from #4174**: removed unused `RegionBox.contains_strict`; made `StubTerminal` a frozen dataclass.

| Thread `stub_terminals` through Autorouter as additive per-net targets, never a Pad/`self.pads`/`self.nets` | ✅ | `set_stub_terminals` + `_stub_target_pads` merge shim; terminals stay ephemeral |
| Obstacle carve-out reopens exactly the tip cell for the stub's own net | ✅ | `reopen_stub_terminal`; foreign nets still blocked (`cell.net != net`) |
| Relax CLI reachability gate for on-boundary stub nets | ✅ | `_parse_and_apply_region` reclassifies via the detector |
| route-auto deferral gate (Phase 2c fail-fast) | ✅ | `test_route_auto_region_stub_defers_to_phase_2c` |
| Hole-to-hole floor applies on the stub path | ✅ | `test_reconnection_respects_hole_to_hole_floor` |
| Drop `contains_strict`; freeze `StubTerminal` | ✅ | stub_terminals.py diff |
| Genuine strip_traces stub reconnects | ✅ | `test_route_region_reconnects_stripped_stub` |
| Zero copper modified outside region | ✅ | `test_reconnection_preserves_outside_copper_byte_identical` |
| Multi-stub same-net | ✅ | `test_multi_stub_same_net_reconnects` |

## Does the carve-out work end-to-end?

**Yes.** A* actually reconnects a stub in the tests. `test_route_region_reconnects_stripped_stub` builds a genuine boundary-clipped stub via `PCB.add_trace` + `strip_traces(region=...)`, then `route --region` produces new in-region copper that reaches both the in-region pad and the boundary stub tip cell — verified end-to-end on the C++ backend. The foreign-halo-clobber case (where the tip cell's net is overwritten by a neighboring net) is the load-bearing reason the carve-out exists and is exercised by the hole-to-hole test's fixture.

## Test Plan

- `pytest tests/router/test_region_bounded.py tests/router/test_stub_terminals.py` → 45 passed (5 new stub-reconnection cases).
- `pytest tests/router/ -k "region or stub or preserve or via"` → 135 passed.
- Broader router smoke (`-k "not slow and not benchmark and not chorus and not board"`) → 557 passed, 1 skipped.
- `tests/test_mcp_routing.py` → 37 passed.
- `ruff check` + `ruff format` clean; `check_mypy_baseline.py` → 0 new errors (1475 ≤ 1514).

Closes #4170

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -1257,6 +1257,15 @@ def _finalize_routes(
         source_routes = preserved_routes if preserved_routes is not None else router.existing_routes
         if source_routes:
             routed_net_ids = {r.net for r in router.routes}
+            # Issue #4170 (Phase 2b-1): a stub net IS routed (the in-region
+            # reconnection) but its OUTSIDE boundary-stub copper must still be
+            # preserved -- the new in-region trace joins the stub, it does not
+            # replace it.  The router records its stub net ids in
+            # ``_stub_terminals``; keep those nets' existing copper by removing
+            # them from the exclusion set.
+            stub_net_ids = set(getattr(router, "_stub_terminals", {}) or {})
+            if stub_net_ids:
+                routed_net_ids = routed_net_ids - stub_net_ids
             preserved_sexp = _serialize_preserved_routes(
                 source_routes, exclude_net_ids=routed_net_ids
             )
@@ -3596,6 +3605,10 @@ def route_with_layer_escalation(
                     # Issue #4148: region-bounded routing.  When set, cells
                     # outside the board-relative box are marked as obstacles.
                     region=getattr(args, "_region_box", None),
+                    # Issue #4170 (Phase 2b-1): board-relative boundary stub
+                    # terminals whose tip cells are carved open as same-net
+                    # reconnection targets (None when no --region / no stubs).
+                    stub_terminals=getattr(args, "_stub_terminals", None),
                     # Issue #3877: thread the C++ iteration backstop through the
                     # layer-escalation sub-router too.  Without this the
                     # escalation-mode Autorouter (board-01/03 et al.) routes with
@@ -4515,6 +4528,10 @@ def route_with_rule_relaxation(
                     load_existing_routes=getattr(args, "preserve_existing", False),
                     # Issue #4148: region-bounded routing (see main()).
                     region=getattr(args, "_region_box", None),
+                    # Issue #4170 (Phase 2b-1): board-relative boundary stub
+                    # terminals whose tip cells are carved open as same-net
+                    # reconnection targets (None when no --region / no stubs).
+                    stub_terminals=getattr(args, "_stub_terminals", None),
                 )
         except Exception as e:
             if not quiet:
@@ -6620,6 +6637,10 @@ def route_with_combined_escalation(
                         load_existing_routes=getattr(args, "preserve_existing", False),
                         # Issue #4148: region-bounded routing (see main()).
                         region=getattr(args, "_region_box", None),
+                        # Issue #4170 (Phase 2b-1): board-relative boundary stub
+                        # terminals whose tip cells are carved open as same-net
+                        # reconnection targets (None when no --region / no stubs).
+                        stub_terminals=getattr(args, "_stub_terminals", None),
                     )
             except Exception as e:
                 if not quiet:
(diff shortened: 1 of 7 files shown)


rjwalters:
✅ **Approved!** — Phase 2b-1 stub-endpoint reconnection is correct, safe, and well-tested. This was reviewed with maximum rigor given the router-core surgery.

## Core correctness (the crux)

**Carve-out (`RoutingGrid.reopen_stub_terminal`) — correct and clearance-safe.** It re-asserts `cell.net = net` + `cell.original_net = net` while keeping `blocked=True`, `is_obstacle=False`, `pad_blocked=False`. I traced this against the C++ pathfinder guards (`pathfinder.cpp`): every foreign-net rejection is gated on `cell.net != net` — `static_blocked && cell.net != net`, `is_obstacle && cell.net != net`, and `pad_blocked && cell.net != net`. So the tip cell is own-net-reachable and remains a hard obstacle to every foreign net. Only the single tip cell is touched; the rest of the stub halo is intact, so it cannot open a clearance gap for a different net. The foreign-halo-clobber reasoning is the real trigger and the fix is right.

**C++ mirror agrees.** `cpp_grid._impl.mark_blocked(gx, gy, layer, net, False, False)` sets `blocked=true, net=net, is_obstacle=false, pad_blocked=false, static_blocked=true, original_net=net` — byte-identical semantics to the Python mutation. Python grid and C++ grid agree on the tip cell's net.

**Rip-up safety (architect risk #6) — handled.** `reopen_stub_terminal` calls `_ensure_static_blockage_snapshot()` before mutating (the tip is already `blocked` from the halo, so it is in the static snapshot), and `_unmark_segment` restores `cell.net = cell.original_net` (= stub net) for static-blocked own-net cells. A rip-up that sweeps the tip restores it as own-net-reachable, not foreign. Verified the hole-to-hole test's `route_net(1)` reconnection path exercises this.

**Ephemeral StubTerminals — never leak into pads/nets.** Grepped the diff: `_stub_terminals`/`_stub_target_pads` are read-only additive target sources; the only write to `router.nets` is the confined `router.nets[net_id] = kept` prune of stub nets. No `self.pads[...] =` or stub insertion anywhere.

**Both merge sites covered (the #3428 two-surface hazard).** All three A* entry points apply BOTH the stub-target merge and the outside-pad prune consistently: `route_net` (~line 2282), the negotiated builder loop (~line 7918), and `_route_net_negotiated` (~line 11069). No silent stub-miss surface.

**Outside-pad pruning — confined and correct.** Only applies when `region is set AND net in self._stub_terminals`; non-stub nets return `pad_keys` unchanged (byte-for-byte the normal-route behavior). Inclusive box, keeps in-region and `pad is None` keys. Cannot alter routing for non-region/normal routes, and cannot drop legitimately-in-region targets.

## Scope / safety
- **route-auto deferral gate**: `route_net_auto` fails fast with an explicit Phase 2c (#4173) message when the net has a same-net boundary stub, and the plain unreachable message otherwise — verified by `test_route_auto_region_stub_defers_to_phase_2c`. No silent half-work.
- **Zero copper outside region**: `test_reconnection_preserves_outside_copper_byte_identical` asserts geometry-identity of the outside stub (rounded set comparison) and, by set equality, that no NEW copper lands outside. `_finalize_routes` preserves stub-net outside copper (removes stub nets from the exclusion set so the stub is kept, not replaced).
- **Hole-to-hole floor**: threaded on the `stub_terminals=` path (`test_reconnection_respects_hole_to_hole_floor` uses the same `_via_clears_hole_to_hole` predicate as Phase 2a).
- **#4174 nits folded**: `StubTerminal` frozen, `RegionBox.contains_strict` removed.

## Verification

**Local (cpp backend, worktree):**
- `tests/router/test_region_bounded.py` + `test_stub_terminals.py` → **45 passed**
- `tests/router/ -k "region or stub or preserve or via or negotiat"` (highest merge-shim regression risk) → **154 passed**
- `tests/router/ -k "not slow and not benchmark and not chorus and not board"` → **557 passed, 1 skipped**
- `tests/test_mcp_routing.py` → **37 passed**
- ruff check + format clean; mypy baseline **0 new errors (1475 ≤ 1514)**

**CI (all 16 checks green, mergeStateStatus CLEAN):** Board 05 Routing Regression (blocking ≤ 11), Board 06/07 E2E, Diff-Pair Routing, Match-Group Routing, and full Test suite (21820 passed) all pass.

**On the transient failure:** the initial `Test` run failed only on `tests/perf/test_pathfinder_dormant_partner_perf.py::test_dormant_partner_optimized_not_slower` (1.261x vs 1.25x — a 0.9% overshoot of the CI noise budget). This is a documented shared-runner timing flake (issue #3581, whose docstring records the same class of false failure on artifact-only PRs that pass on rerun). This PR touches no pathfinder / dormant-partner code, and the test passes locally on the PR branch (1.72s). I re-ran the failed job and it went **green (14m19s)**, confirming the flake. No real router-test failure.

Excellent adherence to the architect's design pass — the ephemeral-StubTerminal abstraction, the same-net punch-through carve-out, the two symmetric merge shims off a single shared detector, and the loud route-auto deferral gate are all implemented as specified.

PR #4180, opened 2026-07-15
feat(route-auto): reconnect boundary stubs on the region-bounded orchestrator path

### LINKED ISSUE #4173: route: Phase 2c — route-auto/orchestrator parity for boundary-stub reconnection (follow-up to #4170)

rjwalters:
## Summary

Brings `route-auto --region` to parity with `route --region` for boundary-stub reconnection (Phase 2c, follow-up to #4170). Implemented per the architect's **corrected** design comment (issues/4173#issuecomment-4975397979), NOT the issue body's Scope section (which named the wrong fix site, `_build_net_target_positions`, a no-op for target sourcing).

The true fix site is `_build_pads_for_net` in `src/kicad_tools/mcp/tools/routing.py` — the sole producer of the `pads` list consumed by `route_net_auto` → `orchestrator.route_net`. Region confinement is provided by the already-shipping post-route output-escape filter (no per-cell obstacle grid exists on this path), so the #4170 carve-out (`mark_region_bound`/`reopen_stub_terminal`) is neither ported nor needed.

**Parity verified end-to-end:** the new parity test routes a synthetic stripped stub through `route_net_auto(..., region=)` and asserts the emitted copper reaches R1's pad (world 110,110) AND joins the boundary stub tip (world 115,110). The routed output genuinely reaches the stub tip — this is working reconnection, not a stub.

## Changes

- Factor the inline stub-adapter (previously inside the deferral gate) into a reusable `_detect_stub_terminals_for_pcb` helper that reuses the shared #4172 `detect_boundary_stub_terminals` producer (no re-derived stub geometry).
- `_build_pads_for_net`: when a region box is set and the net owns a same-net boundary stub, prune outside-region pads and inject the boundary stub tip as an in-region reconnection target (mirrors the Autorouter `self.nets` prune at io.py:3889-3928). Synthetic stub-tip pads are route-scoped/ephemeral.
- Remove the "Phase 2c not yet wired" fail-fast branch; keep the no-stub reachability failure.
- Fix the output-escape filter's coordinate frame: the orchestrator routes in the board-relative frame of its input pads, so the region box is compared directly without the previously-incorrect board-origin shift. (This latent frame bug was never exercised before because no route-auto `--region` route reached the filter successfully.)
- Replace `test_route_auto_region_stub_defers_to_phase_2c` with a `TestStubReconnectionRouteAuto` class; document the honest residual (coarse corridor may bulge through an out-of-box tile center and fail honestly at the output filter).

| route-auto --region reconnects a boundary stub identically to route --region | ✅ | `test_route_auto_region_reconnects_stripped_stub` — asserts copper reaches R1 pad + boundary stub tip |
| Zero copper modified outside the region | ✅ | `test_route_auto_reconnection_preserves_outside_copper` — outside stub geometry-identical before/after |
| Hole-to-hole floor respected for any new via near the reconnection | ✅ | `test_route_auto_reconnection_respects_hole_to_hole_floor` — `_via_clears_hole_to_hole` rejects a floor-violating via |
| "Phase 2c not yet wired" gate + its test removed | ✅ | Gate branch deleted; `test_route_auto_region_stub_defers_to_phase_2c` replaced |
| No-stub net still fails fast | ✅ | `test_route_auto_region_no_stub_still_fails` — retained "no same-net boundary stub" message |
| Confinement documented | ✅ | Code comment at injection site + parity class/test docstrings |
| Did NOT touch `_build_net_target_positions` / port carve-out | ✅ | Confirmed via diff grep |

## Test Plan

- `pytest tests/router/test_region_bounded.py tests/router/test_stub_terminals.py` — 48 passed
- `pytest tests/router/ -k "region or stub or preserve or auto"` — 83 passed
- `pytest tests/ -k "route_auto or routing"` — 1088 passed, 1 unrelated flake (`test_board02_manufacturable_baseline::test_routing_output_deterministic_across_seeds`, passes in isolation; does not touch route-auto/region)
- ruff check + format: clean
- mypy baseline: 0 new errors

Closes #4173

### DIFF
--- a/src/kicad_tools/mcp/tools/routing.py
+++ b/src/kicad_tools/mcp/tools/routing.py
@@ -396,6 +396,72 @@ def route_net(
         )
 
 
+def _detect_stub_terminals_for_pcb(pcb: PCB, region_box: tuple[float, float, float, float]) -> dict:
+    """Detect boundary stub terminals for every straddling net on ``pcb``.
+
+    Thin adapter over the shared #4172 producer
+    :func:`kicad_tools.router.stub_terminals.detect_boundary_stub_terminals`:
+    it reduces the PCB's loaded segments/pads into the detector's pure
+    board-relative :class:`StubSegment` / :class:`PadLocation` inputs and
+    returns its ``{net_id: [StubTerminal, ...]}`` mapping.  Stub geometry is
+    NEVER re-derived here -- this is the single reuse point on the route-auto
+    path (shared by both the reachability gate's has-stub check and the
+    ``_build_pads_for_net`` target injection), mirroring the Autorouter side's
+    single call from ``io.py::load_pcb_for_routing``.
+
+    ``region_box`` is board-relative ``(x1, y1, x2, y2)`` mm (same convention as
+    ``pcb strip --region``).  Returns an empty dict on any adaptation failure so
+    callers can treat "no detectable stub" uniformly.
+
+    Note: :class:`StubTerminal` objects are route-scoped and ephemeral -- they
+    are never persisted as a ``Pad`` on the PCB.
+    """
+    try:
+        from kicad_tools.core.types import CopperLayer
+        from kicad_tools.router.stub_terminals import (
+            PadLocation,
+            RegionBox,
+            StubSegment,
+            detect_boundary_stub_terminals,
+        )
+    except Exception:  # pragma: no cover - defensive import guard
+        return {}
+
+    rx1, ry1, rx2, ry2 = region_box
+    region = RegionBox(rx1, ry1, rx2, ry2)
+
+    segments: list = []
+    for seg in pcb._segments:
+        try:
+            layer = CopperLayer.from_kicad_name(seg.layer)
+        except ValueError:
+            continue
+        segments.append(
+            StubSegment(
+                net_id=seg.net_number,
+                net_name=seg.net_name,
+                x1=seg.start[0],
+                y1=seg.start[1],
+                x2=seg.end[0],
+                y2=seg.end[1],
+                layer=layer,
+                uuid=getattr(seg, "uuid", "") or None,
+            )
+        )
+
+    pad_locs: list = []
+    for fp in pcb.footprints:
+        for pad in fp.pads:
+            pos = pcb.get_pad_position(fp.reference, pad.number)
+            if pos is not None:
+                pad_locs.append(PadLocation(net_id=pad.net_number, x=pos[0], y=pos[1]))
+
+    try:
+        return detect_boundary_stub_terminals(segments, pad_locs, region)
+    except Exception:  # pragma: no cover - defensive
+        return {}
+
+
 def route_net_auto(
     pcb_path: str,
     net_name: str,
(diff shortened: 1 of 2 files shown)


rjwalters:
## ✅ Approved!

Reviewed against the architect's **corrected** design (issues/4173#issuecomment-4975397979 — authoritative; the issue BODY's Scope is known-wrong). High-rigor review of the target-sourcing change and the coordinate-frame confinement-filter fix. **Provably correct in both directions.**

### Correctness — all verified
- **Right fix site.** The change is in `_build_pads_for_net` (`mcp/tools/routing.py`), the sole producer of the `pads` list read by every orchestrator strategy. `orchestrator._build_net_target_positions` was **NOT** touched (confirmed: `git diff origin/main..HEAD -- src/kicad_tools/router/orchestrator.py` is empty). The wrong site from the issue body was correctly avoided.
- **No re-derived geometry.** `_detect_stub_terminals_for_pcb` is a thin adapter over the shared #4172 `detect_boundary_stub_terminals`; it's the single reuse point for both the reachability gate's has-stub check and the injection.
- **Prune-outside / inject-stub-tip** mirrors the io.py:3889-3928 shape and is confined to stub nets: the prune predicate is `region_box is not None and stub_terminals and not _in_region(...)`, so a `None` region or a non-stub net leaves the normal pad set untouched.
- **Deferral gate fully removed** with no dangling half-gate; the genuine no-stub reachability failure is **retained**; the old `test_route_auto_region_stub_defers_to_phase_2c` is *replaced* by `TestStubReconnectionRouteAuto`, not merely deleted.
- **End-to-end parity.** `test_route_auto_region_reconnects_stripped_stub` asserts the routed copper reaches R1's pad (world 110,110) AND joins the boundary stub tip (world 115,110) via `_touches_cell` — real reconnection, not just `success=True`.

### The coordinate-frame fix — scrutinized hardest, proven in both directions
Empirically verified the actual frames (not just the comments) on a nonzero-origin fixture (board origin 100,100):
- `PCB` stores `fp.position` / `get_pad_position` **board-relative** (footprint at sheet 110,110 → stored 10,10).
- Loaded `_segments` are **board-relative** too (sheet 130,118 → stored 30,18).
- `add_trace` documents + implements "inputs are board-relative; re-adds board origin on serialization" — which is why the saved output has world coords the tests assert.

So the orchestrator routes board-relative pads → `result.segments` are board-relative → the region box is board-relative. Traced the filter both ways:
- **NEW filter, in-region segment → passes** (not escaped). ✅
- **NEW filter, out-of-region segment → still rejected** (escaped). Confinement guarantee is **not loosened** in the permissive direction. ✅
- **OLD filter, in-region segment → wrongly rejected** — it added the board origin to the box, comparing a world-frame box against board-relative geometry. Confirmed the bug was real and **latent**: main's `test_tuple_region_accepted` explicitly tolerated failure, so no passing in-region route ever exercised the filter before the gate was removed.

The fix compares box and geometry in the **same** board-relative frame — correct, and it never lets out-of-region copper escape.

### Regression
- `region=None` path is completely inert: `region_box` defaults to `None`, `stub_terminals` stays empty, all injection/pruning is guarded.
- Reported flake (`test_board02_manufacturable_baseline::test_routing_output_deterministic_across_seeds`) has **zero** references to `route_net_auto` / `_build_pads_for_net` / `region_box` — genuinely unrelated.

### Test & CI
- Local: `test_region_bounded.py` + `test_stub_terminals.py` → **48 passed**; `tests/router/ -k "region or stub or preserve or auto"` → **83 passed**.
- CI: **all 16 checks pass** (full Test 16m51s, Type Check, Lint & Format, Board 00-07 + diff-pair + match-group routing regressions). `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

**Minor doc nit (non-blocking):** the `_build_pads_for_net` docstring bullet says the stub tip is injected "at the tip's WORLD position (board-relative tip + `pcb._board_origin`)" — self-contradictory. The code correctly uses `term.x`/`term.y` with **no** origin shift (board-relative), consistent with the pads, and the inline comment at the injection site states this correctly. Just the one docstring phrase is misleading; safe to tidy in a follow-up.

Excellent, careful work — especially documenting that the output-escape filter is the SOLE confinement guarantee on this path and now does real work.

PR #4181, opened 2026-07-15
fix(route): match --net-class-map bare keys against '/'-prefixed board nets

### LINKED ISSUE #4149: route: --net-class-map keys silently match zero nets when board nets carry the hierarchical '/' prefix
## Bug (silent misconfiguration — route runs untuned)

Boards produced by `kct create-pcb` name label-derived nets with KiCad's hierarchical root-sheet prefix (`/FUSED_LINE`, `/SRC_POS`, ...), while power-symbol nets stay global (`GND`, `+3.3V`). A `--net-class-map` sidecar keyed on bare names therefore matches **zero** nets — and the router says nothing:

```
$ kct route board.kicad_pcb --net-class-map classes.json --skip-nets "GND,+3.3V" ...
# classes.json keys: FUSED_LINE, PGND, SRC_POS, ... (34 entries)
# board nets:        /FUSED_LINE, /PGND, /SRC_POS, ...
# → all 34 entries ignored; every power net routed at default width/clearance/priority
# → --skip-nets worked (global net names have no prefix), masking the problem
```

Observed on softstart rev-B (2026-07-14, v0.15.0): first route pass completed exit 0 with 30/76 nets and none of the heavy-current tuning applied. Nothing in the output indicated the class map was inert; the mismatch was only found by comparing `kct net-status` names (`/AC_LINE`) against the sidecar keys. The same prefix bites hand-written `.kicad_dru` conditions (`A.NetName =~ /^AC_LINE$/` never matches) — and the global-vs-local inconsistency (`GND` bare, `/AC_LINE` prefixed) makes it easy to conclude your names are right when they're half-right.

## Suggested fixes

1. **Warn loudly (or error) when net-class-map entries match no board net** — e.g. `WARNING: 34/34 net-class entries matched 0 nets (nearest: '/FUSED_LINE' for 'FUSED_LINE')`. A one-line Levenshtein/prefix hint makes the fix obvious.
2. **Normalize hierarchical prefixes during matching**: a bare key `FUSED_LINE` should match root-sheet `/FUSED_LINE` (and arguably `--skip-nets`, `--power-nets`, `--analog-nets`, `ROUTE_FIRST`-style priority fields should all share the same normalizer).
3. Docs: one sentence in `--net-class-map` help noting the `/` convention for label nets vs global power nets.

(1) alone would have saved the wasted route pass; (2) makes consumer recipes robust to the convention.

rjwalters:
## Summary

KiCad names label-derived nets with the hierarchical root-sheet prefix (`/FUSED_LINE`) while power-symbol nets stay bare (`GND`, `+3.3V`). A `--net-class-map` sidecar keyed on bare names therefore matched **zero** prefixed board nets and silently ran the route untuned at exit 0 (softstart rev-B incident). This fixes the matching so bare keys resolve to `/`-prefixed nets, and adds a loud misconfiguration diagnostic for keys that still fail to resolve.

Scope is `--net-class-map` **only** (curator decision). The sibling selectors (`--skip-nets` / `--power-nets` / `--analog-nets`) share the same underlying bug; adopting the new shared helper for them is a **documented follow-up** and is intentionally NOT changed here. The helper is built and unit-tested so that follow-up is an import-and-swap.

## Changes

- **New `src/kicad_tools/router/net_names.py`** — shared, dependency-free normalizer/matcher:
  - `net_name_suffix()` — segment after the last `/` (handles arbitrary sheet nesting).
  - `build_net_name_index()` — `suffix -> [raw board nets]`, detects collisions.
  - `resolve_net_key()` — fully-qualified exact match > unique-suffix match > ambiguous > none. A bare key that collides across multiple board nets is refused (not silently assigned).
  - `nearest_net_names()` — lightweight prefix/suffix hint (no Levenshtein).
  - `resolve_net_class_map_keys()` — partitions sidecar keys into resolved / unmatched / ambiguous.
- **`_apply_net_class_map_sidecar` (route_cmd.py)** — resolves each sidecar key against `router.net_names`, rekeys overrides to the board's actual net name so `core.py`'s `net_class_map.get(net_name)` finds them, and emits an aggregate stderr diagnostic (`N/M entries matched a board net`) plus per-key nearest-name / ambiguity hints. Printed regardless of `--quiet`; exit code unchanged (advisory).
- **Docs** — one sentence in the `--net-class-map` help text (route_cmd.py + parser.py) documenting the `/` convention.
- **Tests** — new `tests/test_net_names.py` (pure unit) + `TestHierarchicalPrefixNormalization` in `tests/test_cli_route_net_class_map.py`.

| #1 Bare key matches '/'-prefixed net; no warning | ✅ | `test_bare_key_matches_prefixed_net` |
| #2 Zero-match (typo) warns with nearest hint; others silent; exit 0 | ✅ | `test_zero_match_warns_with_nearest_hint` |
| #3 Full-zero-match aggregate warning fires | ✅ | `test_full_zero_match_aggregate_warning` |
| #4 Ambiguity (/A and A) → warning + neither overridden | ✅ | `test_ambiguous_key_applied_to_neither`, `test_ambiguous_key_matches_none` |
| #5 Exact bare match unchanged (no regression) | ✅ | `test_exact_bare_match_unchanged`, full route/net-class-map suite green |
| #6 Help text documents '/' convention | ✅ | route_cmd.py + parser.py help strings |
| #7 All fixtures synthetic | ✅ | Hand-built net maps + `NetClassRouting` entries, no real boards |
| --quiet does not suppress warning | ✅ | `test_warning_not_suppressed_by_quiet` |

## Test Plan

- `uv run pytest tests/test_net_names.py tests/test_cli_route_net_class_map.py` — 42 passed.
- `uv run pytest tests/ -k "net_class_map or route_cmd"` — 303 passed, 7 skipped (route merge path via `load_pcb_for_routing`).
- `uv run ruff check` / `ruff format` — clean.
- `uv run --extra dev python scripts/ci/check_mypy_baseline.py` — 0 new errors (within baseline).

## Follow-up (not in this PR)

Wire the same `net_names` helper into `--skip-nets`, `--power-nets`, and `--analog-nets`. Each has its own downstream consumption pattern (zone generation for `--power-nets`) needing separate regression tests; deferred per curator scope guidance.

Closes #4149

### DIFF
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -3533,7 +3533,11 @@ def _add_route_parser(subparsers) -> None:
             "length_match_group) project through to the routing-time "
             "pathfinder.  Without this flag, --differential-pairs falls "
             "back to NetClassRouting defaults and can produce overlapping "
-            "diff-pair sibling traces under tight mfr profiles."
+            "diff-pair sibling traces under tight mfr profiles.  Keys are "
+            "matched against the board net's sheet-local suffix (the "
+            "segment after the last '/'), so a bare key FUSED_LINE matches "
+            "KiCad's '/'-prefixed label net /FUSED_LINE while global power "
+            "nets (GND, +3.3V) stay bare (Issue #4149)."
         ),
     )
     route_parser.add_argument(
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -2879,16 +2879,96 @@ def _apply_net_class_map_sidecar(router: "Autorouter", args, quiet: bool = False
     wrappers) calls this helper to merge the rich per-pair / per-group
     fields onto the router's name-pattern-classified map.
 
+    Issue #4149: board nets carry KiCad's hierarchical sheet prefix
+    (``/FUSED_LINE``) for label-derived nets while power-symbol nets stay
+    bare (``GND``).  Sidecar keys are almost always written bare, so a raw
+    ``.update()`` silently keyed the overrides by names that
+    ``self.net_class_map.get(net_name)`` never looks up.  We now resolve
+    each bare key against the board's actual net names (matching on the
+    suffix after the last ``/``), rekey the overrides to the board net
+    name, and emit an aggregate stderr diagnostic for keys that matched no
+    board net or matched ambiguously.  Ambiguous keys are applied to
+    *neither* candidate — silently picking one would just relocate the bug.
+
     Idempotent and a no-op when the flag was not supplied.
     """
     loaded = getattr(args, "_loaded_net_class_map", None)
     if not loaded:
         return
-    router.net_class_map.update(loaded)
+
+    from kicad_tools.router.net_names import (
+        nearest_net_names,
+        resolve_net_class_map_keys,
+    )
+
+    board_net_names = list(router.net_names.values())
+    resolution = resolve_net_class_map_keys(loaded.keys(), board_net_names)
+
+    # Apply overrides rekeyed to the board's actual net names so
+    # ``self.net_class_map.get(net_name)`` finds them at routing time.
+    for board_net, user_key in resolution.resolved.items():
+        router.net_class_map[board_net] = loaded[user_key]
+
+    _warn_unresolved_net_class_map(resolution, board_net_names, nearest_net_names)
+
     if not quiet:
         from kicad_tools.cli.progress import flush_print
 
-        flush_print(f"  Net-class map: merged {len(loaded)} sidecar entries")
+        flush_print(
+            f"  Net-class map: merged {len(resolution.resolved)}/{resolution.total} sidecar entries"
+        )
+
+
+def _warn_unresolved_net_class_map(resolution, board_net_names, nearest_fn) -> None:
+    """Emit the Issue #4149 misconfiguration diagnostic to stderr.
+
+    Prints an aggregate summary line plus per-key hints when any
+    ``--net-class-map`` key failed to resolve to a board net (zero-match)
+    or matched more than one distinct board net (ambiguous).  A no-op when
+    every key resolved.
+
+    This is always printed (never suppressed by ``--quiet``): the
+    softstart-rev-B incident showed a fully-inert class map runs to exit 0
+    with no signal, so the misconfiguration warning must survive
+    ``--quiet``.  Exit code is unchanged (advisory, not a hard gate).
+    """
+    if not resolution.unmatched and not resolution.ambiguous:
+        return
+
+    import sys
+
+    matched = len(resolution.resolved)
+    total = resolution.total
+    lines: list[str] = [
+        f"WARNING: net-class-map: {matched}/{total} entries matched a board net "
+        f"after normalization."
+    ]
+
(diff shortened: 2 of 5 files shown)


rjwalters:
✅ **Approved!**

Solid fix for the silent net-class-map misconfiguration (softstart rev-B incident). Reviewed against all curator acceptance criteria — every one is verified by a dedicated test at both the unit and CLI-integration level.

## Correctness verification

**Rekeying actually takes effect on routing (not just the warning):** `_apply_net_class_map_sidecar` writes `router.net_class_map[board_net] = loaded[user_key]` where `board_net` comes from `router.net_names.values()` (`dict[int, str]`, confirmed in `core.py:981`). Routing-time lookup in `core.py` is `self.net_class_map.get(net_name)` with `net_name = self.net_names.get(net, ...)` (raw board name) — so the rekey lands under exactly the key `.get()` queries. `test_bare_key_matches_prefixed_net` asserts the override lands under `/FUSED_LINE` and that the bare `FUSED_LINE` does NOT leak into the map. Confirmed correct.

**Suffix normalization does not mis-match a net it shouldn't:** the ambiguity guard is the safety net. If a bare key `A` would match both `A` and `/A` (or `/X/A` and `/Y/A`), `resolve_net_key` refuses it — even when one candidate is an exact string match — because silently preferring the exact one just relocates the bug this fixes. `test_ambiguous_key_applied_to_neither` + `test_exact_wins_over_ambiguity` cover both the refusal and the fully-qualified escape hatch. This is the correct call: refusal + a distinct "use the fully-qualified name" hint beats warning-but-apply, per the issue's own framing (silent-wrong is the whole problem).

**Key matching exactly AND another by suffix:** handled — that IS the ambiguous case (`A` present alongside `/A`), and it's refused. `test_ambiguous_key_partitioned` confirms the sibling unambiguous key still resolves while the colliding one is skipped.

**Regression safety:** `test_exact_bare_match_unchanged` + 303 broader `net_class_map`/`route_cmd` tests confirm already-correct bare-vs-bare keys are unaffected; no drop/double-apply.

## Checks (run in worktree at PR head 776268d6)
- `tests/test_net_names.py` + `tests/test_cli_route_net_class_map.py` — **42 passed**
- `pytest -k "net_class_map or route_cmd"` — **303 passed, 2 skipped**
- `ruff check` on all changed files — **clean**
- mypy baseline — **0 new errors** (1475 within 1514 floor; also 39 baseline errors no longer occur)

## CI
All 16 GitHub checks **pass** (Board 00-07 E2E, routing regressions, diff-pair, match-group, C++ build, type check, lint, DRC, round-trip smoke, Test). `mergeStateStatus: CLEAN`.

## Notes (non-blocking, not worth tracking)
- Scope correctly limited to `--net-class-map`; sibling selectors (`--skip-nets`/`--power-nets`/`--analog-nets`) deferred per curator guidance, with the helper built import-and-swap-ready.
- Degenerate edge: supplying BOTH `FUSED_LINE` and `/FUSED_LINE` in one sidecar collapses to a single `resolved` entry (last-write-wins by board net), so the aggregate count reports 1/1 rather than 2/2. Contrived input, routing still applies the surviving override correctly — not worth a follow-up.

PR uses `Closes #4149`. Clean, well-documented, thoroughly tested.

PR #4184, opened 2026-07-15
fix(route): wire --placement-feedback into rule-relaxation & combined-escalation paths (#4151)

### LINKED ISSUE #4151: route: --placement-feedback produces zero output evidence it engaged (silent no-op?)
## Bug / observability gap

A route run invoked with the placement-feedback loop and explicit anchors:

```
kct route board.kicad_pcb ... --placement-feedback \
  --placement-feedback-anchor "J1,J2,J3,J4,J5,F1,RV1,Q1A,...,U1" \
  --export-failed-nets failed.txt
```

completed (exit 0, 25/76 nets) with **zero lines of output mentioning placement feedback** — no iteration count, no components moved/considered, no "feedback loop disabled because X". `grep -ci placement <full output>` = 0. There is no way to tell whether:

- the loop ran and moved nothing,
- the loop ran, moved things, and still landed at 25/76, or
- the flag was silently ignored (e.g. incompatible with `--net-class-map`, the C++ backend, or `--targeted-ripup`).

Board positions in the output PCB appeared unchanged from the input placement (anchored refs and non-anchored small parts alike), suggesting silent no-op, but I can't distinguish that from "converged at iteration 0."

Same failure class as #4149 (silently-inert configuration): an opt-in feature that can't confirm it did anything will eventually eat someone's routing budget while they tune a knob that isn't connected.

## Suggested fix

Minimum: one summary line — `placement-feedback: N iterations, M components moved (total X mm), completion 25/76 -> 25/76` — or an explicit `placement-feedback: DISABLED (<reason>)`. Ideal: per-iteration lines under `-v`.

Context: kicad-tools 0.15.0, softstart rev-B routing (rjwalters/softstart PR #28), macOS, C++ backend v1.0.0.

---

### Root cause: this IS a real silent-disable bug, not just missing telemetry

`_run_placement_feedback()` (`src/kicad_tools/cli/route_cmd.py:2053-2178`) already prints a
reasonably complete summary when it runs — iterations, exit reason, components
moved, final failed-net count, anchored refs, and the diff-JSON sidecar path
(lines 2094-2176). So "the loop ran and printed nothing" is not what's
happening. The bug is upstream: **`_run_placement_feedback` is only wired
into two of the five routing-dispatch functions in `route_cmd.py`.**

`kct route` dispatches to one of five top-level routing functions based on
flag combination (`src/kicad_tools/cli/route_cmd.py:8814-8881`):

| Dispatch condition | Function | Calls `_run_placement_feedback`? |
|---|---|---|
| `--auto-pcb-size` | `route_with_size_escalation` (line 5437) | Yes — delegates to `route_with_layer_escalation` in every branch (verified at lines 5836 and 5943) |
| `--auto-mfr-tier` | `route_with_mfr_tier_escalation` (line 5063) | Yes — delegates to `route_with_layer_escalation` (line 5352) |
| `--auto-layers` (default **True**) alone | `route_with_layer_escalation` (line 3265) | **Yes** — direct hook at line 4096 |
| `--auto-layers --adaptive-rules` together | `route_with_combined_escalation` (line 6376) | **No.** ~740-line standalone 2D layer×rule-tier search loop with zero references to `placement_feedback` anywhere in its body. Comments (e.g. line 6614, 6817) mention `route_with_layer_escalation` for context but the function never calls it. |
| `--adaptive-rules` alone (i.e. `--no-auto-layers --adaptive-rules`, or `--adaptive-rules` when the user also disabled auto-layers) | `route_with_rule_relaxation` (line 4328) | **No.** Standalone rule-relaxation loop; zero references to `placement_feedback`. |
| no escalation flags at all (`--no-auto-layers`, no `--adaptive-rules`) | `main()`'s direct single-pass path | Yes — hook at line 9965 |

**Confirmed via exhaustive search**: `grep -n "_run_placement_feedback(" src/kicad_tools/cli/route_cmd.py`
returns exactly 3 hits — the function definition (2053), the
`route_with_layer_escalation` call site (4096), and the direct-path call site
(9965). `route_with_rule_relaxation` and `route_with_combined_escalation`
have **zero** occurrences of the string `placement_feedback` anywhere in
their bodies (verified with `awk` line-range + grep across both function
spans).

Since `--auto-layers` defaults to `True` (`src/kicad_tools/cli/route_cmd.py:7984-7996`,
`default=True`), most invocations land on the hooked
`route_with_layer_escalation` path and placement-feedback works as
documented. The silent-disable bug specifically bites:
- `kct route ... --adaptive-rules` (without `--no-auto-layers` this still
  routes through `route_with_layer_escalation` UNLESS the user also passed
  `--no-auto-layers`, in which case it silently drops to
  `route_with_rule_relaxation`), and
- `kct route ... --auto-layers --adaptive-rules` (both explicit), which
  silently drops into `route_with_combined_escalation`.

The reporter's command line elides flags with `...`, so we cannot confirm
which exact path they hit, but softstart rev-B routing is plausible to use
`--adaptive-rules` for trace-width/clearance escalation, which would explain
the total silence even though `--placement-feedback` was set and the router
genuinely had unrouted nets (25/76) to work on.

**This is a real bug to fix**, not just a telemetry gap: on these two paths,
`--placement-feedback` is accepted by the parser, forwarded through
`cli/commands/routing.py` (lines 329-364), and then dropped on the floor
with no error, no warning, and no PCB mutation — the exact "silently-inert
configuration" failure class named in the issue and previously fixed for
`--net-class-map` in #4149 (PR #4181, merged into `origin/main`).

### Anchor-parsing gap (tie-in to #4149's failure class)

`_parse_ref_list()` (route_cmd.py:1982-1986) and
`_resolve_placement_feedback_anchors()` (route_cmd.py:2023-2041) never
validate `--placement-feedback-anchor` / `--placement-feedback-no-anchor`
refs against the board's actual footprint references. A misspelled ref
(plausible given the reporter's own example list `Q1A,...,U1` — note `Q1A`
is an unusual ref format for a two-pad transistor) is silently absorbed into
the `anchored` set via a plain set union/difference and never cross-checked.
There's no signal that a typo'd ref matched zero footprints. This should be
fixed alongside the observability gap since it's part of the same "opt-in
flag looks connected but silently does nothing useful" bug class.

### Fix design

**1. Structural fix — invoke (or explicitly report skipping) `_run_placement_feedback` on all five dispatch paths.**

Two sub-options for the two uncovered paths:

- **Option A (minimal, recommended)**: Add an explicit
  `placement-feedback: DISABLED (--adaptive-rules is not yet compatible with
  placement feedback; use --auto-layers instead, or file a follow-up)` guard
  at the top of `route_with_rule_relaxation` and
  `route_with_combined_escalation` when `getattr(args, "placement_feedback",
  False)` is true. This makes the current (missing) behavior *visible*
  without expanding scope to wiring up feedback inside two large,
  independently-structured routing loops (~600-740 LOC each) in this PR.
- **Option B (full fix)**: Wire `_run_placement_feedback` into both
  functions' terminal branches, mirroring the pattern at
  `route_with_layer_escalation` lines 4085-4119 (gate on: escalation/relaxation
  did not already succeed, `--placement-feedback` set, `router.routes is not
  None`, and `router.get_failed_nets()` non-empty; refresh completion stats
  afterward).

Recommend starting with **Option A** to close the silent-disable class
immediately (matches the "at minimum ... DISABLED (<reason>)" ask in the
issue), and filing Option B as a fast-follow if there's appetite — this
keeps the PR reviewable and avoids touching two large, delicate escalation
loops in the same change as the CLI-summary-line work.

**2. Summary-line telemetry** — already exists in `_run_placement_feedback`
(lines 2154-2176: iterations, exit reason, components moved, final failed
nets, diff-JSON path). No change needed on the paths that already call it
(`route_with_layer_escalation`, direct path, and transitively
`route_with_size_escalation` / `route_with_mfr_tier_escalation`). Verify
with a live run that this output is not accidentally suppressed by `quiet`
inheriting the wrong default anywhere on those paths.

**3. Anchor validation** — in `_resolve_placement_feedback_anchors()`
(route_cmd.py:2023), after computing `anchors`, diff the user-supplied
`--placement-feedback-anchor` / `--placement-feedback-no-anchor` ref sets
against `{fp.reference for fp in pcb.footprints}` and print a
`Warning: --placement-feedback-anchor ref(s) not found on board: {refs}`
line (not fatal — mirror the tolerant-warning style used elsewhere in this
file, e.g. `_apply_net_class_map_sidecar` warnings) when any requested ref
doesn't match a real footprint.

**4. `-v`/`--verbose` gating for per-iteration detail** — `PlacementFeedbackLoop`
(`src/kicad_tools/router/placement_feedback.py`) already prints per-iteration
progress lines gated on `self.verbose` (see lines 376, 411, 421, 460, 478,
500, 511, 516-517, 531), but `_run_placement_feedback` currently passes
`verbose=not quiet` (route_cmd.py:2141) rather than tying the *per-iteration*
detail to `args.verbose`/`-v` specifically. Today `not quiet` means
per-iteration lines print on every non-quiet run, which is arguably too
noisy for the default case and doesn't match the issue's ask ("ideal:
per-iteration lines under `-v`"). Options:
   - Keep the existing one-line summary always-on (non-quiet), and only pass
     `verbose=True` to `route_with_placement_feedback` when
     `getattr(args, "verbose", False)` is set — i.e. change line 2141 from
     `verbose=not quiet` to `verbose=(not quiet) and (getattr(args, "verbose", False) or ...)`.
     This needs care: today's default (`verbose=not quiet`) is exercised by
     existing tests (`tests/test_placement_feedback.py`,
     `tests/test_route_cmd_placement_feedback.py`) — changing the default
     verbosity could regress expectations there. Read those tests before
     changing the default; if they assert on the current per-iteration
     output being present, either update them deliberately (with a rationale
     comment) or keep a reduced-but-nonzero non-verbose default and add the
     full per-iteration firehose strictly under `-v`.

### Acceptance Criteria

- [ ] `--placement-feedback` set AND dispatch lands on
      `route_with_layer_escalation` / `route_with_size_escalation` /
      `route_with_mfr_tier_escalation` / the direct no-escalation path AND
      the loop runs → the existing one-line summary (`Feedback iterations`,
      `Exit reason`, `Components moved`, `Final failed nets`, `Placement
      diff saved to`) is present in stdout (already implemented; add a
      regression test that greps for these lines end-to-end, not just at
      the unit level).
- [ ] `--placement-feedback` set AND dispatch lands on
      `route_with_rule_relaxation` (`--adaptive-rules` without effective
      `--auto-layers`) → an explicit
      `placement-feedback: DISABLED (<reason>)` line is printed; the flag is
      no longer silently dropped.
- [ ] `--placement-feedback` set AND dispatch lands on
      `route_with_combined_escalation` (`--auto-layers --adaptive-rules`
      together) → same explicit DISABLED line.
- [ ] `--placement-feedback-anchor` with at least one ref that does not
      match any footprint on the board → a warning naming the unmatched
      ref(s) is printed; matched refs still anchor correctly.
- [ ] `-v`/`--verbose` produces additional per-iteration detail beyond the
      summary line (exact scope depends on which telemetry-gating option is
      chosen in implementation — the criterion is "meaningfully more detail
      under `-v` than without it", not a specific line format).
- [ ] Existing tests in `tests/test_placement_feedback.py`,
      `tests/test_route_cmd_placement_feedback.py`, and
      `tests/test_chorus_test_placement_feedback.py` continue to pass
      (or are deliberately updated with a documented rationale if verbosity
      defaults change).
- [ ] No regression to the byte-identical-routes-when-flag-unset invariant
      documented in `tests/test_route_cmd_placement_feedback.py`'s module
      docstring.

### Affected Files

- `src/kicad_tools/cli/route_cmd.py` — `_run_placement_feedback` (2053),
  `_resolve_placement_feedback_anchors` (2023), `route_with_rule_relaxation`
  (4328), `route_with_combined_escalation` (6376): add the DISABLED-reason
  guard (and/or full wiring) on the two uncovered dispatch paths; add
  anchor-validation warning.
- `tests/test_route_cmd_placement_feedback.py` — add coverage for the
  DISABLED-reason path(s) and anchor-validation warning.
- `tests/test_placement_feedback.py` — review/update if verbosity-gating
  default changes.
- `tests/test_chorus_test_placement_feedback.py` — confirm no regression
  (chorus fixture is local-only per repo memory; skip gracefully in CI if
  the fixture is absent, consistent with existing test skip conventions).

### Test Plan

- [ ] Manual verification: run `kct route <fixture>.kicad_pcb
      --placement-feedback --adaptive-rules --no-auto-layers` on a small
      synthetic board with intentionally unrouteable nets; confirm a
      `placement-feedback: DISABLED (...)` line appears and exit code/output
      otherwise matches pre-change behavior (no crash, no regression to
      rule-relaxation output).
- [ ] Manual verification: run the same board with
      `--placement-feedback --auto-layers --adaptive-rules`; confirm the
      same DISABLED line (or, if Option B is implemented, confirm the
      summary line with real iteration/movement counts).
- [ ] Manual verification: run with
      `--placement-feedback --placement-feedback-anchor NOTAREALREF`;
      confirm a warning names `NOTAREALREF` as unmatched.
- [ ] Manual verification: run with `-v --placement-feedback` on a board
      with unrouted nets after the initial pass; confirm additional
      per-iteration output appears beyond the summary line.
- [ ] Automated tests: extend `tests/test_route_cmd_placement_feedback.py`
      with unit tests mocking `route_with_rule_relaxation` and
      `route_with_combined_escalation` entry to assert the DISABLED message
      is emitted when `args.placement_feedback` is True and the dispatch
      condition is met; add a unit test for anchor-ref validation in
      `_resolve_placement_feedback_anchors`.
- [ ] Edge cases: `--placement-feedback` with zero failed nets after the
      initial pass (loop should not run and should not print a DISABLED
      message — this is "not needed", a different state from "not
      supported"; verify these two states remain textually distinguishable
      in the output); `--placement-feedback-anchor` with an empty string;
      `--placement-feedback` combined with `--auto-pcb-size` and
      `--auto-mfr-tier` (already-hooked paths) to confirm no regression.

rjwalters:
## Summary

`kct route --placement-feedback` was silently dropped on two of the five routing-dispatch paths. The flag was parsed, forwarded through `cli/commands/routing.py`, and then **never executed and emitted no output at all** when dispatch landed on:

- `route_with_rule_relaxation` — reached via `--adaptive-rules` without effective `--auto-layers` (`--no-auto-layers --adaptive-rules`)
- `route_with_combined_escalation` — reached via `--auto-layers --adaptive-rules` together

This is the "silently-inert configuration" failure class named in the issue (same as `--net-class-map` in #4149).

## What changed

1. **Fix the silent-disable (Option B — full wiring, per issue's stated preference to actually run the feature).** Extracted the existing `route_with_layer_escalation` placement-feedback hook into a shared `_maybe_run_placement_feedback_escalation()` helper and invoked it at the terminal (post-results) branch of `route_with_rule_relaxation` and `route_with_combined_escalation`. `--placement-feedback` now **runs** on all dispatch paths.
2. **Never-silent telemetry.** When the loop runs, the one-line summary (`Feedback iterations` / `Exit reason` / `Components moved` / `Final failed nets` / `Placement diff saved to`) is emitted. When placement feedback is requested but the router produced no routes, an explicit `placement-feedback: DISABLED (<reason>)` line prints. "Not needed" (fully succeeded / zero failed nets) stays silent and remains textually distinct from "not supported".
3. **`-v` gating.** The per-iteration firehose from `PlacementFeedbackLoop` is now gated on `-v`/`--verbose` (was `not quiet`); the always-on summary line is independent of `-v`.
4. **Anchor validation.** `--placement-feedback-anchor` / `--placement-feedback-no-anchor` refs are validated against the board's footprints; unmatched refs produce a tolerant warning (`Warning: ... not found on board: ...`), matching #4149's class.

## Manual verification

Ran `kct route tests/fixtures/routing-diagnostic.kicad_pcb --no-auto-layers --adaptive-rules --placement-feedback` (the previously-silent `route_with_rule_relaxation` path):

```
ADAPTIVE RULES SUMMARY
Result: Best result at tier 1 (75% completion)
--- Engaging placement-routing feedback (rule relaxation stalled at 75%) ---
  Feedback iterations: 2
  Exit reason:        pf_timeout
  Components moved:   1
  Final failed nets:  1
  Placement diff saved to: .../rr2_placement_diff.json
  Post-feedback: 3/4 (75%)
```

The loop **runs and moves a component** — not just reports. The combined-escalation path shares the identical helper.

## Tests

- `tests/test_route_cmd_placement_feedback.py`: escalation-hook runs + emits summary on both previously-silent paths (rule-relaxation and combined labels); DISABLED-line-on-no-routes; silent-when-not-requested; silent-when-fully-succeeded; silent-when-no-failed-nets; a static guard that all three escalation functions invoke the hook; anchor-ref validation warnings (unknown anchor / no-anchor / all-known / quiet); `-v` verbose gating + summary-present-regardless-of-`-v`.
- `tests/test_placement_feedback.py`: unchanged, still passing (loop verbosity is constructor-injected, unaffected by the route_cmd default change).

`uv run pytest tests/test_route_cmd_placement_feedback.py tests/test_placement_feedback.py -q` → 128 passed. `tests/test_chorus_test_placement_feedback.py` (local-only fixture) → 3 passed, 1 skipped, no regression. ruff clean; mypy: 0 new errors vs baseline.

Closes #4151

### DIFF
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -2029,7 +2029,7 @@ def _auto_detect_anchored_refs(pcb) -> set[str]:
     return anchored
 
 
-def _resolve_placement_feedback_anchors(pcb, args) -> set[str]:
+def _resolve_placement_feedback_anchors(pcb, args, quiet: bool = False) -> set[str]:
     """Compute the final set of anchored refs for the feedback loop.
 
     Combines auto-detected anchors (connectors, locked footprints) with
@@ -2037,16 +2037,40 @@ def _resolve_placement_feedback_anchors(pcb, args) -> set[str]:
     any refs the user explicitly opted out via
     ``--placement-feedback-no-anchor``.
 
+    Issue #4151: user-supplied ``--placement-feedback-anchor`` /
+    ``--placement-feedback-no-anchor`` refs are validated against the
+    board's actual footprint references.  A typo'd ref that matches no
+    footprint used to be silently absorbed into the set with no signal;
+    we now emit a tolerant warning naming the unmatched ref(s) (same
+    silently-inert-configuration failure class as #4149).
+
     Args:
         pcb: Loaded PCB object.
         args: Parsed CLI args.
+        quiet: When True, suppress the unmatched-ref warning.
 
     Returns:
         Set of refs to anchor.
     """
+    board_refs = {
+        (getattr(fp, "reference", "") or "") for fp in getattr(pcb, "footprints", []) or []
+    }
+    board_refs.discard("")
+
+    requested_anchor = _parse_ref_list(getattr(args, "placement_feedback_anchor", None))
+    requested_no_anchor = _parse_ref_list(getattr(args, "placement_feedback_no_anchor", None))
+
+    if not quiet:
+        unknown = sorted((requested_anchor | requested_no_anchor) - board_refs)
+        if unknown:
+            print(
+                "Warning: --placement-feedback-anchor/--placement-feedback-no-anchor "
+                f"ref(s) not found on board: {', '.join(unknown)}"
+            )
+
     anchors = _auto_detect_anchored_refs(pcb)
-    anchors |= _parse_ref_list(getattr(args, "placement_feedback_anchor", None))
-    anchors -= _parse_ref_list(getattr(args, "placement_feedback_no_anchor", None))
+    anchors |= requested_anchor
+    anchors -= requested_no_anchor
     return anchors
 
 
@@ -2112,7 +2136,7 @@ def _run_placement_feedback(
             print(f"  Warning: could not load PCB for feedback ({exc}); skipping")
         return None
 
-    anchored = _resolve_placement_feedback_anchors(pcb, args)
+    anchored = _resolve_placement_feedback_anchors(pcb, args, quiet=quiet)
     if not quiet and anchored:
         print(f"  Anchored refs ({len(anchored)}): {', '.join(sorted(anchored))}")
 
@@ -2142,12 +2166,21 @@ def _run_placement_feedback(
     if _remaining is not None:
         outer_timeout = _remaining if outer_timeout is None else min(outer_timeout, _remaining)
 
+    # Issue #4151: gate the per-iteration firehose (candidate breakdowns,
+    # per-move deltas emitted by ``PlacementFeedbackLoop`` under
+    # ``self.verbose``) on ``-v``/``--verbose`` specifically, rather than
+    # on ``not quiet``.  The always-on one-line summary below
+    # (iterations / exit reason / components moved / final failed nets)
+    # is emitted independently whenever ``not quiet``, so a default run
+    # still gets telemetry -- just without the noisy per-iteration detail.
+    loop_verbose = (not quiet) and bool(getattr(args, "verbose", False))
+
     try:
         result = router.route_with_placement_feedback(
             pcb=pcb,
             max_adjustments=budget,
             use_negotiated=use_negotiated,
-            verbose=not quiet,
+            verbose=loop_verbose,
             fixed_refs=anchored,
             max_movement=max_movement,
             timeout=timeout,
@@ -2185,6 +2218,107 @@ def _run_placement_feedback(
             print(f"  Warning: could not write placement diff ({exc})")
 
     return diff_data
+
+
+def _maybe_run_placement_feedback_escalation(
+    final_result,
+    successful_result,
+    pcb_path: Path,
+    args,
+    quiet: bool,
+    *,
+    stall_label: str,
+) -> None:
+    """Engage placement-routing feedback at the tail of an escalation loop.
+
+    This is the shared hook used by ``route_with_layer_escalation`` and,
+    via this helper, by ``route_with_rule_relaxation`` and
+    ``route_with_combined_escalation`` (issue #4151).  Before this helper
+    existed, the latter two dispatch paths never referenced
+    ``_run_placement_feedback`` at all, so ``--placement-feedback`` was
+    parsed, forwarded, and silently dropped on the floor.
+
+    Behaviour:
+
+    * When ``--placement-feedback`` is not requested, this is a no-op
+      (byte-identical routing preserved).
+    * When requested and the escalation already fully succeeded
+      (``successful_result is not None``), the loop is not needed -- we
+      stay silent (this is the "not needed" state, distinct from the
+      "not supported" DISABLED state).
+    * When requested and there are still failed nets on a partial result,
+      the feedback loop runs and ``_run_placement_feedback`` emits its
+      one-line summary; ``final_result``'s completion stats are refreshed
+      from the post-feedback router state so optimize/save/summary all see
+      the correct numbers.
+    * When requested but the router produced no routes at all (nothing to
+      feed back into), emit an explicit
+      ``placement-feedback: DISABLED (...)`` line so the request is never
+      silently swallowed.
+
+    Args:
+        final_result: The chosen result object (mutated in place on run).
+        successful_result: The fully-successful result, or None.
+        pcb_path: Path to the (staged) input PCB.
+        args: Parsed CLI args.
+        quiet: Suppress output when True.
+        stall_label: Human-readable name of the escalation strategy for
+            the "Engaging" banner (e.g. "rule relaxation").
+    """
+    if not getattr(args, "placement_feedback", False):
+        return
+
+    # Fully-succeeded escalation: feedback is simply not needed.  Stay
+    # silent so "not needed" remains textually distinguishable from the
+    # "not supported / disabled" state below.
+    if successful_result is not None:
+        return
+
+    router = final_result.router
+
+    # No routes at all -> nothing to feed placement changes back into.
+    # Make the (previously silent) drop explicit rather than swallowing it.
+    if router.routes is None:
+        if not quiet:
+            print(
+                "placement-feedback: DISABLED (no routes produced by "
+                f"{stall_label}; nothing to feed back)"
+            )
+        return
+
+    if not router.get_failed_nets():
+        # All nets already routed on this partial-but-complete result; the
+        # loop has nothing to improve.  Not a DISABLED case.
+        return
+
+    if not quiet:
+        print(
+            f"\n--- Engaging placement-routing feedback "
+            f"({stall_label} stalled at {final_result.completion * 100:.0f}%) ---"
+        )
+    _run_placement_feedback(
+        router=router,
+        pcb_path=pcb_path,
+        args=args,
+        quiet=quiet,
+    )
+    # Refresh completion stats from the post-feedback router state so
+    # optimize/save/summary all see the correct numbers.
+    _refreshed_multi_pad_ids = {n for n, p in router.nets.items() if n > 0 and len(p) >= 2}
+    _refreshed = router.get_statistics(nets_to_route_ids=_refreshed_multi_pad_ids)
+    final_result.nets_routed = _refreshed["nets_routed"]
+    final_result.completion = (
+        final_result.nets_routed / final_result.nets_to_route
+        if final_result.nets_to_route > 0
+        else 1.0
+    )
+    final_result.success = final_result.completion >= args.min_completion
+    if not quiet:
+        print(
+            f"  Post-feedback: {final_result.nets_routed}/"
+            f"{final_result.nets_to_route} "
+            f"({final_result.completion * 100:.0f}%)"
+        )
 
 
 def _fill_zones_after_route(output_path: Path, quiet: bool = False) -> None:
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the real bug fix, telemetry, and regression safety. This is a clean, faithful implementation of Option B (full wiring) — the issue's stated preference to actually run the feature rather than just report it as disabled.

## The real bug fix — verified
- `_maybe_run_placement_feedback_escalation()` was extracted from the inline `route_with_layer_escalation` hook and is now invoked at the **correct terminal branch** of all three escalation functions: after `final_result` is selected from `successful_result`/`best_result`, after the `Error: No routing attempts succeeded → return 1` early-out, and **before** optimize/save. Since the loop mutates `router.routes` in place and the completion stats are refreshed inside the helper, the downstream optimize/save/summary all see post-feedback numbers.
- **No double-execution.** Confirmed via source scan that neither `route_with_rule_relaxation` (4530–4967) nor `route_with_combined_escalation` (6597–7146) delegates internally to `route_with_layer_escalation` or calls `_run_placement_feedback` directly — the only `placement_feedback` references in those bodies are doc-comment cross-refs plus the single new hook call each. The combined path cannot run feedback in a layer sub-phase and again at the terminal.
- **Layer-escalation path behavior preserved.** The extraction is byte-equivalent on the existing guard (`successful_result is None`, `routes is not None`, `get_failed_nets()`) and the stats-refresh block. The only intentional delta is the new never-silent `DISABLED` line when `routes is None` (previously a silent drop) and a cosmetic label change (`escalation` → `layer escalation`).

## Never-silent telemetry — verified
- Summary line emitted whenever the loop runs (test `test_summary_present_regardless_of_v` asserts `Feedback iterations` / `Exit reason` in stdout).
- Explicit `placement-feedback: DISABLED (no routes produced by <label>; nothing to feed back)` on the requested-but-no-routes case — cannot be missed on a genuine skip (`test_disabled_line_when_no_routes`).
- Silent only on the two "not needed" states (fully succeeded / zero failed nets), which stay textually distinct from DISABLED (`test_silent_when_fully_succeeded`, `test_silent_when_no_failed_nets`).

## `-v` gating — verified
- Per-iteration firehose moved from `verbose=not quiet` to `loop_verbose = (not quiet) and bool(args.verbose)`. Tests assert the resolved `verbose` kwarg actually reaches `route_with_placement_feedback` (→ `placement_feedback.py`): `False` without `-v`, `True` with `-v`; summary independent of `-v`.

## Regression guard — verified
- `test_escalation_functions_call_the_hook` uses `inspect.getsource` to assert **all three** escalation functions reference `_maybe_run_placement_feedback_escalation(` — a future path can't silently drop the hook again.

## Anchor validation — verified
- `_resolve_placement_feedback_anchors()` now diffs requested anchor / no-anchor refs against `{fp.reference for fp in pcb.footprints}` and emits a tolerant `Warning: ... not found on board: ...` (suppressed under `quiet`). Covered by 4 tests (unknown anchor / unknown no-anchor / all-known / quiet).

## Checks
- **Local**: `pytest tests/test_route_cmd_placement_feedback.py tests/test_placement_feedback.py` → **128 passed** (matches PR claim). ruff check + format clean on both changed files.
- **CI**: all 16 checks pass — full Test suite (16m58s), Type Check, Lint, C++ Build, and every routing-regression board (Board 05 blocking ≤11 passed at 19m30s; #4160 flake fix held). Diff-Pair and Match-Group regressions green.
- **Merge state**: CLEAN / MERGEABLE.
- PR description uses `Closes #4151`.

Solid work — the shared helper preserves the original invocation semantics exactly while closing the silent-disable class on the two uncovered dispatch paths.

Issue #4182, opened 2026-07-15
placement check: reports 0 conflicts on a placement where kicad-cli DRC finds 18 courtyard-overlap errors


rjwalters:
## Original Issue

`kct placement check` and KiCad disagree on courtyard overlaps, in the dangerous direction:

- `kct placement check board.kicad_pcb` → **"No placement conflicts found!"** (after iterating on the warnings it *did* report — the tool clearly checks courtyards, and we fixed every one it flagged)
- `kicad-cli pcb drc --severity-error` on a board produced from the **same placement recipe** (114 parts, identical `place_revb.py` output) → **`courtyards_overlap: 18`** at error severity

So kct's courtyard model finds a strict subset of KiCad's. Whatever differs (courtyard-layer parsing per footprint variant, THT courtyards, rotation handling, tolerance), the effect is that "0 conflicts" from `placement check` does not predict a DRC-clean placement — and placement is exactly the stage where you want that prediction, before routing effort is spent.

Related: #4176 (net-status connectivity vs KiCad), #4156 (no off-board check) — `placement check`'s contract would benefit from a general statement of which KiCad DRC placement classes it does/doesn't reproduce.

### Suggested fix (original)

Reconcile the courtyard test with KiCad's (same layers, same polygon test, same severity), or run/borrow kicad-cli's own check when available. Interim: document the known gap.

Artifacts: rjwalters/softstart PR #28, rev-B placement (`place_revb.py`) + `/tmp`-era drc_batch.json (retained; happy to attach). kicad-tools 0.15.0, KiCad 10.

---

### Root cause — confirmed

`kct placement check` (`PlacementAnalyzer`) does **not** use real courtyard geometry at all. It approximates each footprint's courtyard as its **pads bounding box expanded by a fixed margin**:

`src/kicad_tools/placement/analyzer.py:262-320` (`_footprint_to_component`):

```python
if pads:
    min_x = min(p.position.x - p.size[0] / 2 for p in pads)
    max_x = max(p.position.x + p.size[0] / 2 for p in pads)
    min_y = min(p.position.y - p.size[1] / 2 for p in pads)
    max_y = max(p.position.y + p.size[1] / 2 for p in pads)
    pads_bbox = Rectangle(min_x, min_y, max_x, max_y)

    # Courtyard is pads bbox + margin
    courtyard = pads_bbox.expand(courtyard_margin)  # default 0.25 mm, DesignRules.courtyard_margin
```

[...]

PR #4187, opened 2026-07-15
fix(placement): flag off-board footprints in placement check + route preflight

### LINKED ISSUE #4156: placement check + route: components outside Edge.Cuts pass silently — no off-board preflight
## Bug (missing preflight — cost three full route passes)

Neither `kct placement check` nor `kct route` flags footprints placed **outside the board outline**. Both proceed as if the placement were valid:

- `kct placement check` reported **"No placement conflicts found!"** on a board where ~30 of 103 components (an entire component row) sat 17.5 mm beyond the Edge.Cuts north edge.
- `kct route` ran to completion (exit 0) and simply failed every net touching the off-board pads. The failure signature — completion plateaued at 25–30/76 across three passes, failed nets clustered by board region — is indistinguishable from congestion, which is exactly what we spent two tuning passes chasing (net classes, clearances, ordering, placement feedback).

One SVG render found it instantly; no kct gate ever mentioned it.

## How it happened (root cause worth a doc note too)

Two centering conventions coexist in the toolchain and differ by the title-block band:

- `kct create-pcb --width 160 --height 100` draws Edge.Cuts plainly centered on A4: origin `((297−160)/2, (210−100)/2)` = **(68.5, 55)**
- `kicad_tools.pcb.center_sheet.centered_origin(160, 100, "A4")` returns **(68.5, 37.5)** (title-block-aware)

A placement script that anchors to `centered_origin()` and applies positions to a `create-pcb` board lands everything 17.5 mm high. (Consumer-side fix that made this structurally impossible: read the outline back from the artifact and anchor to `min(Edge.Cuts)` — rjwalters/softstart `hardware/kicad/place_revb.py`.)

## Suggested fixes

1. **`kct placement check`: add an off-board check** — error per footprint whose courtyard/pads fall outside (or partially outside) the Edge.Cuts polygon. This is the cheapest, highest-value gate; "0 conflicts" currently overpromises.
2. **`kct route`: preflight abort** (or loud warning + summary line) when any routable pad lies outside the outline: `ERROR: 31 footprints / 94 pads outside Edge.Cuts — placement invalid`. Alternatively fold into the existing `--sync-check`.
3. Docs/API: one sentence on `create-pcb` documenting where it puts the outline, and/or make `centered_origin()` and `create-pcb` agree on a single convention.

Related in spirit to #4149 and #4151 (silently-inert or silently-invalid configuration classes): the common thread is that route completion percentage is doing the job of an error message.

Context: kicad-tools 0.15.0, softstart rev-B layout (rjwalters/softstart PR #28), 160×100 4-layer, C++ backend.

rjwalters:
## Summary

`kct placement check` reported "No placement conflicts found!" and `kct route`
ran to completion on boards where whole component rows sat outside the
Edge.Cuts outline. Root cause (curator-verified): `PlacementAnalyzer._extract_board_edge`
read the outline via `pcb.segments_on_layer("Edge.Cuts")`, which only returns
copper `(segment ...)` elements. Real outlines — including every `create-pcb`
board and every board KiCad writes — are drawn as `gr_line`/`gr_arc`/`gr_rect`
graphics, so the lookup always returned `[]`, `_board_edge` stayed `None`, and
the edge-clearance / off-board checks never ran. The existing `EDGE_CONFLICT_PCB`
test masked this by hand-injecting fake `(segment ... Edge.Cuts)` elements a real
board never has.

## Changes

- **Wrong data source fix**: `_extract_board_edge` now uses
  `PCB.get_board_outline()` (the already-correct path used by
  `place_unplaced._get_board_bounds`), which assembles the outline from
  graphics and returns board-relative coordinates.
- **New `ConflictType.OFF_BOARD`** (unconditional `ERROR`) + `_check_off_board()`:
  flags courtyards fully or partially outside the outline bbox. `_check_edge_clearance`
  now skips outside components so an off-board footprint is reported once as
  OFF_BOARD rather than four times as edge violations.
- **`kct route` off-board preflight**: hard abort (exit 2) before any router
  work, printing footprint/pad counts and pointing at `kct placement check`;
  `--allow-offboard` is the escape hatch (mirrors `--no-sync-check`), inserted
  at the same point as the `--sync-check` advisory banner.
- **De-masked the regression fixture**: `EDGE_CONFLICT_PCB` now draws a bare
  `gr_rect` outline (no fake copper segments) and genuinely exercises the real
  path. Added fully-outside / partially-outside / all-inside placement tests and
  a route-preflight test (`tests/test_route_offboard_gate.py`), all built through
  the graphics-outline path — no hand-injected segments.
- **Docs**: `PCB.create()` docstring now notes its centering is title-block-*unaware*
  and diverges from `center_sheet.centered_origin()` (divergence documented, not
  reconciled — that is the riskier follow-up).

Out of scope (kept separate per curator): the courtyard-overlap approximation
gap (#4182), which needs real `F.CrtYd` geometry.

| 1. Fully-outside footprint on a `PCB.create()` board → OFF_BOARD error, exit 1 | ✅ | `test_offboard_fully_outside_reports_error`; CLI run: `kct placement check` prints `off_board error` and exits 1 |
| 2. Straddling footprint flagged | ✅ | `test_offboard_partial_reports_error` (ERROR, "partially outside") |
| 3. All-inside board still clean, no regression | ✅ | `test_offboard_all_inside_is_clean`; existing `clean_pcb`/`overlapping`/`hole` tests pass |
| 4. Regression test built via real graphics-outline path, not fake segments | ✅ | `EDGE_CONFLICT_PCB` de-masked (bare `gr_rect`); new fixtures graphics-only |
| 5. `kct route` aborts on off-board pads; `--allow-offboard` overrides | ✅ | `test_offboard_aborts_before_routing` (exit 2), `test_allow_offboard_bypasses_gate` (exit 0) |
| 6. Fully-inside board unaffected by route | ✅ | `test_inside_board_is_unaffected` |
| 7. `PCB.create()` docstring notes title-block-unaware centering | ✅ | docstring note cross-referencing `centered_origin()` |

## Test Plan

- `uv run pytest tests/test_placement.py tests/test_route_offboard_gate.py tests/test_netlist_sync_gate.py tests/test_mcp_placement_analyze.py` → all pass (93).
- `ruff check` + `ruff format` clean on all changed files.
- `check_mypy_baseline.py` → 0 new type errors (within baseline).
- Manual CLI: `kct placement check` on a `PCB.create()` board with an off-board R2 prints `off_board error … placement invalid` and exits 1 (confirms the fix works through the real graphics-outline path, not the fake-segment fixture).

Closes #4156

### DIFF
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -325,6 +325,11 @@ def run_route_command(args) -> int:
         sub_argv.append("--no-sync-check")
     if getattr(args, "schematic", None):
         sub_argv.extend(["--schematic", args.schematic])
+    # Issue #4156: forward the off-board placement preflight escape hatch.
+    # --allow-offboard defaults off; forward only when the user set it so the
+    # flag-off path stays byte-identical.
+    if getattr(args, "allow_offboard", False):
+        sub_argv.append("--allow-offboard")
     if getattr(args, "auto_fix", False):
         sub_argv.append("--auto-fix")
     if getattr(args, "auto_fix_passes", None) is not None:
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -3218,6 +3218,21 @@ def _add_route_parser(subparsers) -> None:
             "(default: auto-discover from project.kct or sibling file)."
         ),
     )
+    # Issue #4156: off-board placement preflight escape hatch.  Mirror of the
+    # inner route_cmd.py flag; both sites must stay in sync per
+    # ``tests/test_cli_parser_drift.py``.
+    route_parser.add_argument(
+        "--allow-offboard",
+        action="store_true",
+        default=False,
+        help=(
+            "Skip the off-board placement preflight. By default kct route "
+            "aborts (exit 2) when any footprint's courtyard falls outside the "
+            "Edge.Cuts outline, since routing an off-board net always fails. "
+            "Use this to proceed anyway (e.g. intentional staging/reference "
+            "footprints)."
+        ),
+    )
     route_parser.add_argument(
         "--auto-fix",
         action="store_true",
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -7703,6 +7703,49 @@ def _parse_and_apply_region(args, pcb_path: Path, region_arg: str) -> int:
     return 0
 
 
+def _offboard_preflight(pcb_path: Path) -> int:
+    """Abort routing when any footprint is placed outside the Edge.Cuts outline.
+
+    Off-board placement (the whole board shifted N mm off the outline, or a
+    stray footprint dropped outside it) makes routing pointless — the affected
+    nets can never complete, and the resulting low completion percentage is
+    indistinguishable from routing congestion.  This preflight surfaces the
+    real cause up front (issue #4156).
+
+    Returns 0 when the placement is on-board (or the board has no outline, or
+    the check cannot run), and 2 (matching the blocking netlist-sync gate's
+    exit convention) when off-board footprints are found.
+    """
+    try:
+        from kicad_tools.placement import ConflictType, PlacementAnalyzer
+
+        analyzer = PlacementAnalyzer()
+        conflicts = analyzer.find_conflicts(pcb_path)
+    except Exception:
+        # A preflight that cannot run must never block routing outright.
+        return 0
+
+    offboard = [c for c in conflicts if c.type == ConflictType.OFF_BOARD]
+    if not offboard:
+        return 0
+
+    refs = sorted({c.component1 for c in offboard})
+    pad_count = sum(
+        len(comp.pads) for comp in analyzer.get_components() if comp.reference in set(refs)
+    )
+    print(
+        f"ERROR: {len(refs)} footprint(s) / {pad_count} pad(s) outside "
+        "Edge.Cuts — placement invalid",
+        file=sys.stderr,
+    )
+    print(f"       Off-board: {', '.join(refs)}", file=sys.stderr)
+    print(
+        "       Run `kct placement check` for details, or `--allow-offboard` to route anyway.",
+        file=sys.stderr,
+    )
+    return 2
+
+
 def main(argv: list[str] | None = None) -> int:
     """Main entry point for route command."""
     parser = argparse.ArgumentParser(
@@ -8351,6 +8394,23 @@ def main(argv: list[str] | None = None) -> int:
         help=(
             "Explicit .kicad_sch path for the advisory drift banner "
             "(default: auto-discover from project.kct or sibling file)."
+        ),
+    )
+    # Issue #4156: hard off-board preflight.  Unlike the advisory drift banner,
+    # a footprint placed outside the Edge.Cuts outline makes routing pointless
+    # (its nets can never complete), so kct route aborts by default before any
+    # router work.  --allow-offboard is the explicit escape hatch for boards
+    # that intentionally stage footprints outside the outline.
+    parser.add_argument(
+        "--allow-offboard",
+        action="store_true",
+        default=False,
+        help=(
+            "Skip the off-board placement preflight. By default kct route "
+            "aborts (exit 2) when any footprint's courtyard falls outside the "
+            "Edge.Cuts outline, since routing an off-board net always fails. "
+            "Use this to proceed anyway (e.g. intentional staging/reference "
+            "footprints)."
         ),
     )
     parser.add_argument(
@@ -9072,6 +9132,19 @@ def main(argv: list[str] | None = None) -> int:
         except Exception:
             # Drift detection is advisory; never let it block routing.
             pass
+
+    # Issue #4156: hard off-board placement preflight.  A footprint whose
+    # courtyard falls outside the Edge.Cuts outline can never route (its nets
+    # fail outright), and the failure signature is indistinguishable from
+    # congestion — which is exactly what cost multiple wasted routing passes in
+    # the field.  Abort before any router/component loading unless the user
+    # opted out with --allow-offboard.  The check is O(footprints), computed
+    # once, and reuses the same get_board_outline()-based analysis as
+    # 'kct placement check'.
+    if not getattr(args, "allow_offboard", False):
+        rc = _offboard_preflight(pcb_path)
+        if rc != 0:
+            return rc
 
     # Issue #2996: Validate and load the optional --net-class-map sidecar
     # early -- before dispatching to any of the route_with_* sub-flows --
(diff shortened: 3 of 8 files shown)


rjwalters:
## Changes Requested — CI failing + user-facing flag unreachable via `kct route`

The root-cause fix, the de-masked fixture, the OFF_BOARD conflict type, and the double-report guard are all **correct and well-executed** — I verified them thoroughly (details below). But there is one blocking defect that also breaks the primary escape hatch.

### Blocking: `--allow-offboard` is only on the inner parser, not the outer `kct route` parser

`tests/test_cli_parser_drift.py::test_inner_only_flags_are_in_allowlist` **fails in CI** (`Test` job, exit 1):

```
Argparse drift detected: the following flags are accepted by the inner
'route_cmd.py' parser but rejected by the outer 'kct route' parser:
  --allow-offboard
```

The flag was added to `route_cmd.py`'s inner `argparse` parser but **not** registered on the outer dispatcher parser in `src/kicad_tools/cli/parser.py::_add_route_parser` (line 2698). Consequence: `kct route <board> --allow-offboard` — the exact invocation your PR body and the docstring advertise as the escape hatch — is **rejected by the real CLI entry point**. The escape hatch is only reachable when calling `route_cmd.main()` directly, which is why `test_allow_offboard_bypasses_gate` (which calls `route_main([...])` directly) passes and masked the gap — the same class of bug this PR set out to fix.

For contrast, `--sync-check` (which the PR says it mirrors) *is* registered in the outer parser at `parser.py:3177`. `--allow-offboard` needs the same treatment.

**Fix:**
1. Add `--allow-offboard` to `src/kicad_tools/cli/parser.py::_add_route_parser` (mirror the `--sync-check`/`--no-sync-check` registration near line 3177), so `kct route --allow-offboard` actually works end-to-end.
2. It should NOT go in `INNER_ONLY_ALLOWLIST` — this is a user-facing flag, not an internal debug/profiling flag.
3. Add/extend a test that drives the flag through the **outer** `kct` entry point (not just `route_main`) so this can't regress silently — the current `test_allow_offboard_bypasses_gate` uses the inner parser and would not catch it.

### Everything else verified GOOD (re-review will be fast once the above is fixed)

- **Root-cause fix**: `_extract_board_edge` now delegates to `PCB.get_board_outline()` (handles gr_line/gr_arc/gr_rect/gr_poly/gr_curve, returns board-relative coords) — correct API, same path as `place_unplaced._get_board_bounds`. Confirmed `_board_edge` now populates on a real graphics-outline board and the off-board / edge-clearance checks actually run.
- **De-masked fixture**: `EDGE_CONFLICT_PCB` now draws a bare `gr_rect` with the fake `(segment ... Edge.Cuts)` copper elements removed; the new fixtures are graphics-only. The regression test would genuinely have caught the original bug. This is the load-bearing property and it holds.
- **Coordinate frame**: courtyard (built from board-relative `fp.position` + pads) and the outline bbox (board-relative) are in the same frame — partial/full-outside detection is correct.
- **No double-report**: `_check_edge_clearance` skips components outside the outline; the `outside` predicate there is the exact complement of `_check_off_board`'s union of fully/partially-outside — no gap, no double count. `test_offboard_all_inside_is_clean` and the reclassification assertion in `test_find_edge_clearance_violations` confirm it.
- **None-graceful**: verified locally — a board with no Edge.Cuts outline yields `board_edge=None`, no crash, no false conflicts.
- **Scope**: correctly leaves #4182 (courtyard-overlap) and the centering-convention reconciliation untouched (docstring-only note on `PCB.create()`).
- Type Check (mypy baseline), Lint & Format, and all board/routing regression jobs pass. Scoped local run: 93 passed.

Nice work on the core fix — this is a one-flag registration away from approval.

rjwalters:
Addressed the blocking feedback: `--allow-offboard` is now reachable through the real `kct route` entry point.

**Changes**
- `src/kicad_tools/cli/parser.py::_add_route_parser` — registered `--allow-offboard` (mirrors the `--sync-check` registration; NOT added to `INNER_ONLY_ALLOWLIST` since it is user-facing).
- `src/kicad_tools/cli/commands/routing.py::run_route_command` — forward `--allow-offboard` to the inner `route_main` (byte-identical when absent, mirroring the `--no-sync-check` forwarding). Without this the outer parser would accept the flag but silently drop it before dispatch, so the escape hatch still would not work end-to-end.
- `tests/test_route_offboard_gate.py` — added `TestRouteOffboardGateOuterEntry`, which drives the gate through `kicad_tools.cli.main` (the real `kct` entry point) rather than the inner `route_main`. Covers both the abort path and the `--allow-offboard` bypass, so the inner/outer drift class of bug cannot regress silently.

**Verification (in worktree)**
- `uv run pytest tests/test_cli_parser_drift.py tests/test_route_offboard_gate.py -q` → 20 passed (previously-failing `test_inner_only_flags_are_in_allowlist` now passes; new outer-entry tests pass).
- `uv run ruff check` + `uv run ruff format --check` on changed files → clean.
- `uv run --extra dev python scripts/ci/check_mypy_baseline.py` → 0 new errors (within baseline).

No incidental `uv.lock`/`.loom` churn. Ready for re-review.

rjwalters:

The single blocker from my prior review is fixed, and the Doctor's second gap-closure is correct. End-to-end verification of `kct route --allow-offboard`:

**1. Outer parser registration** — `parser.py::_add_route_parser` now declares `--allow-offboard` (`action="store_true"`, `default=False`), mirroring the `--sync-check` registration. `kct route <board> --allow-offboard` is no longer rejected by argparse at the real entry point.

**2. Not in `INNER_ONLY_ALLOWLIST`** — correct: the flag is user-facing and lives on BOTH parsers, so `test_cli_parser_drift.py::test_inner_only_flags_are_in_allowlist` (the previously-failing test) is satisfied (`inner - outer` no longer contains it).

**3. Second gap closed (forwarding)** — `routing.py::run_route_command` now appends `--allow-offboard` to `sub_argv` when set (flag-off path stays byte-identical). Without this, the outer parser would have accepted the flag and the shim would have silently dropped it before reaching the inner `route_main`. Verified the inner main at `route_cmd.py:9144` reads `args.allow_offboard` and skips `_offboard_preflight` (exit 2) only when the flag is set — so this is a genuine end-to-end bypass, not accepted-then-dropped.

**4. New outer-entry test** — `TestRouteOffboardGateOuterEntry` drives `kicad_tools.cli.main(["route", ...])` (the real `kct` dispatch path, not the inner parser) for both abort (exit 2, off-board message present) and bypass (exit 0, "placement invalid" absent). This locks the inner/outer gap against silent regression.

**5. No regression in prior-approved work** — the fix commit (`d79a42c9`) touches only `parser.py`, `routing.py`, and the new test. The placement-check root-cause fix (`get_board_outline`), `OFF_BOARD` conflict type, de-masked `EDGE_CONFLICT_PCB` fixture, and `PCB.create()` docstring are unchanged from the previously-verified commit.

## Test Execution
- `pytest tests/test_route_offboard_gate.py tests/test_cli_parser_drift.py tests/test_placement.py` → **73 passed** locally (worktree, correct PYTHONPATH).
- `ruff check` + `ruff format --check` on all changed files → clean.

## CI Verification
- All CI checks green (Test job — containing `test_cli_parser_drift` — **pass** 17m18s; all Board 00-07 E2E, routing regressions, C++ build, kicad-cli smoke pass).

Solid, well-tested fix that closes both the flagged blocker and a latent second defect. Ready to merge.

PR #4191, opened 2026-07-15
fix(datasheet): sanitize resolved part number before using it as a filename

### LINKED ISSUE #4183: datasheet download: part numbers containing '/' create unintended subdirectories (no filename sanitization)
## Bug (minor, with a path-safety edge)

`kct datasheet download` derives the output filename from the resolved part number without sanitizing path separators:

```
$ kct datasheet download MCP6001 -o hardware/datasheets
  Path: hardware/datasheets/MCP6001UT-I/OT.pdf     <- resolved part "MCP6001UT-I/OT"
```

The `/` in the resolved part number silently created a `MCP6001UT-I/` subdirectory containing `OT.pdf` — discovered only during cleanup (`rmdir` needed). Many real part numbers contain `/` (Microchip `-I/OT`, `-E/SN` suffixes are everywhere).

Beyond the mess, an unsanitized resolved name is a path-traversal surface: a hostile/buggy source returning a part like `../../something` would write outside `-o`. Unlikely, but free to prevent.

## Suggested fix

Sanitize the derived filename (replace `/` and any path separators with `_`, reject `..`), or use the *queried* part string (also sanitized) rather than the source-resolved one for the filename.

kicad-tools 0.15.0; observed with the LCSC source (rjwalters/softstart session 2026-07-13).

---

### Root Cause

`DatasheetManager.download()` builds the `-o` output path directly from the **source-resolved** part number with no sanitization:

```python
# src/kicad_tools/datasheet/manager.py, DatasheetManager.download()
if output_dir:
    output_dir = Path(output_dir)
    output_dir.mkdir(parents=True, exist_ok=True)
    output_path = output_dir / f"{result.part_number}.pdf"   # <-- line 163, unsanitized
else:
    output_path = self.cache.get_datasheet_path(result.part_number)
```

`result.part_number` is whatever the source returned as the best match (e.g. LCSC's `part.mfr_part`, which for Microchip parts is commonly something like `MCP6001UT-I/OT`). `Path(...) / "MCP6001UT-I/OT.pdf"` silently creates the `MCP6001UT-I/` subdirectory and writes `OT.pdf` inside it — no error, no warning.

Notably, the **cache path** (`-o` not given) already does this correctly:

```python
# src/kicad_tools/datasheet/cache.py, DatasheetCache.get_datasheet_path()
def get_datasheet_path(self, part_number: str) -> Path:
    """Get the expected path for a datasheet file."""
    # Normalize part number for filesystem
    safe_name = "".join(c if c.isalnum() or c in "-_" else "_" for c in part_number)
    return self.cache_dir / safe_name / "datasheet.pdf"
```

So the bug is specifically that `manager.py`'s `-o`/`output_dir` branch (line 163) does not go through equivalent sanitization — it's an inconsistency between the two output-path branches, not a missing concept in the codebase.

### Why sanitize the *resolved* name (not just the query string)

The issue text offers two options: sanitize the resolved part number, or fall back to the (sanitized) queried string. **Recommendation: sanitize whatever value is used to build the filename — i.e. keep using `result.part_number` (it's more specific/useful as a filename, e.g. distinguishing variants returned for a family query) but pass it through a shared sanitizer.** This is strictly safer than an "either/or" choice because:
- It fixes the bug regardless of which source or resolution path produced the value.
- It also protects `download_by_part`'s cache-miss branch and any future callers of `download()` that pass an `output_dir`, without requiring each call site to remember to sanitize its input.
- Filtering the query string alone wouldn't help if a future caller invokes `manager.download(result, output_dir=...)` directly (as tests do) with an unsanitized `result.part_number` already resolved.

### Design: shared sanitizer helper

Add a small helper (e.g. in `src/kicad_tools/datasheet/utils.py`, alongside `calculate_part_confidence`) and use it in **both** places a part number becomes a filename component:

```python
import re

_UNSAFE_FILENAME_CHARS = re.compile(r"[^A-Za-z0-9._-]")

def sanitize_filename_component(name: str) -> str:
    """
    Sanitize a string for safe use as a filename/directory component.

    Replaces path separators and any other filesystem-unsafe characters with
    '_', and strips leading dots so the result can never resolve to '.' or
    '..' (preventing path traversal when this value is joined onto a base
    output directory). Collapses to a fallback if the result would be empty.
    """
    # Normalize os-specific separators explicitly, then the general character
    # class handles '/', '\\', and everything else that isn't safe.
    safe = _UNSAFE_FILENAME_CHARS.sub("_", name)
    # Strip any leading '.' sequences so "..", ".", "...foo" etc. cannot
    # produce a relative-path escape or a hidden file.
    safe = safe.lstrip(".")
    return safe or "unnamed_part"
```

Notes on the design:
- Character allowlist (`A-Za-z0-9._-`) rather than a denylist — this is the robust choice; it can't be bypassed by an exotic separator the denylist forgot (e.g. NTFS ADS `:`, Windows-illegal chars, unicode look-alikes for `/`). It matches the existing style in `cache.py` (`c.isalnum() or c in "-_"`), extended slightly to also allow `.` so a part number like `ATmega328P` stays readable and multi-segment names don't get mangled, while still never allowing a bare `..` to survive (the `lstrip(".")` step neutralizes that).
- Reject/neutralize `..` explicitly via `lstrip(".")` on the *whole* sanitized string — since `/` and `\\` are already replaced with `_`, a payload like `../../something` first becomes `.._.._something` and then the leading dots are stripped to `_.._something`, which cannot escape the target directory. Apply this to each path segment if the value could still contain separators after substitution (it can't, since the regex already turns them into `_`, but keep the `lstrip` as defense-in-depth against a leading-dot single-segment payload like `..secrets`).
- **Also add a belt-and-suspenders confinement check** after constructing `output_path`, following the existing pattern in `_stream_to_validated_pdf` (`src/kicad_tools/datasheet/sources/base.py`) of validating before ever writing to disk:
  ```python
  output_path = output_dir / f"{sanitize_filename_component(result.part_number)}.pdf"
  resolved = output_path.resolve()
  if output_dir.resolve() not in resolved.parents and resolved != output_dir.resolve():
      raise DatasheetDownloadError(
          f"Refusing to write outside output directory: {output_path}"
      )
  ```
  This check is redundant given the character-allowlist sanitizer above (a sanitized single-path-segment filename literally cannot contain `/` or `\\` and so cannot escape `-o`), but it costs little and guards against a future regression in the sanitizer (e.g. someone loosening the allowlist later) — treat it as optional but recommended.

### Path-traversal case (explicit)

With the allowlist sanitizer, a source returning `"../../something"` as `part.mfr_part`:
1. `/` and `.` survive the char-class (`.` is allowed for readability) but the *slashes* are replaced with `_`, turning the string into `.._.._something`.
2. `lstrip(".")` removes the leading `..`, leaving `_.._something`.
3. Result: `output_dir / "_.._something.pdf"` — a normal file created *inside* `output_dir`, not an escape.

The confinement check above (if implemented) would additionally catch anything that slips past step 1–2 and raise `DatasheetDownloadError` rather than writing outside `-o`.

### Affected Files

- `src/kicad_tools/datasheet/manager.py` — `DatasheetManager.download()`, the `output_dir` branch that builds `output_path = output_dir / f"{result.part_number}.pdf"` (around line 163); apply `sanitize_filename_component(result.part_number)` here. `download_by_part()` calls `download()` internally so no separate change needed there.
- `src/kicad_tools/datasheet/utils.py` — add the new `sanitize_filename_component()` helper next to `calculate_part_confidence()`.
- `src/kicad_tools/datasheet/cache.py` — `DatasheetCache.get_datasheet_path()` (line 171-175) currently inlines an equivalent-but-slightly-different sanitizer (`isalnum() or c in "-_"`, no `.` allowed, no explicit `..`-stripping since `/` removal already prevents traversal into `cache_dir`). Optional but recommended: refactor this to call the new shared helper for consistency, as long as existing cache-path tests (`test_get_datasheet_path`, `TestDatasheetCache`) still pass — the two sanitizers currently differ only in whether `.` is preserved, which does not change existing cache behavior for real part numbers (none use bare `.`-only segments).
- `tests/test_datasheet.py` — add test coverage under `TestDatasheetManager` (see Test Plan).

### Out of scope

- Reworking `DatasheetCache.get_datasheet_path()`'s on-disk format (`cache_dir / safe_name / "datasheet.pdf"`) — only align its character set with the new helper if it's a drop-in, no behavior change for existing cached entries.
- Changing `download()`'s public signature or adding new CLI flags.
- This is a small, single-PR bug fix — do not expand scope to a general filesystem-safety audit of the whole datasheet subsystem.

## Acceptance Criteria

- [ ] A resolved part number containing `/` (e.g. `MCP6001UT-I/OT`) downloads to a **single file** named with `_` substituted for `/` (e.g. `MCP6001UT-I_OT.pdf`) directly inside `-o`, with **no subdirectory created**.
- [ ] A resolved part number attempting path traversal (e.g. `../../etc/passwd`-shaped string, or a leading `../`) is confined to the `-o` directory — the resulting file is written inside `-o`, never above/outside it.
- [ ] A normal part number with no unsafe characters (e.g. `STM32F103C8T6`) is unaffected — same filename as before.
- [ ] The resulting file still has the expected `.pdf` extension in all cases above.
- [ ] `DatasheetCache.get_datasheet_path()` behavior for existing/typical part numbers is unchanged (no regression in cache tests).

## Test Plan

- [ ] Automated tests in `tests/test_datasheet.py`, following the existing `TestDatasheetManager` fixture/mock conventions (mock `manager.sources[0].download` per `test_download_caches_result`/`test_download_by_part`, or patch `manager.download`'s source lookup — no live network calls):
  - `test_download_sanitizes_slash_in_part_number` — `DatasheetResult(part_number="MCP6001UT-I/OT", ...)`, call `manager.download(result, output_dir=tmp_path)`, assert the resulting `datasheet.local_path` is `tmp_path / "MCP6001UT-I_OT.pdf"` (or equivalent sanitized name) and that `tmp_path` contains no subdirectory (`assert not (tmp_path / "MCP6001UT-I").exists()`).
  - `test_download_rejects_path_traversal_in_part_number` — `DatasheetResult(part_number="../../escape", ...)`, call `manager.download(result, output_dir=tmp_path)`, assert the resulting file path's resolved parent is `tmp_path` (i.e. `datasheet.local_path.resolve().parent == tmp_path.resolve()`), and assert nothing was written outside `tmp_path` (e.g. check `tmp_path.parent` and above are untouched).
  - `test_download_normal_part_number_unaffected` — existing-style part number (e.g. `"STM32F103C8T6"`), assert filename is unchanged (`tmp_path / "STM32F103C8T6.pdf"`).
  - Each new/touched test asserts the final path has a `.pdf` suffix.
  - If `sanitize_filename_component()` is added as a standalone helper in `utils.py`, add direct unit tests for it (e.g. `TestSanitizeFilenameComponent` mirroring `TestCalculatePartConfidence`'s style) covering: slash replacement, backslash replacement, leading `..` stripping, empty-after-sanitization fallback, and a normal-part-number passthrough case.
- [ ] Manual verification (optional, no live network needed): construct a `DatasheetResult` with a `/`-containing `part_number` and a local `download_by_part`/`download` mock returning a fixed PDF byte string, run through the CLI's `_run_download` path or call `DatasheetManager.download()` directly in a REPL, confirm no subdirectory appears under `-o`.
- [ ] Run `uv run pytest tests/test_datasheet.py -v` and confirm all existing + new tests pass.
- [ ] Run `uv run ruff check src/kicad_tools/datasheet/` and `uv run mypy src/kicad_tools/datasheet/` on touched files.

rjwalters:
## Summary

`kct datasheet download -o <dir>` derived the output filename directly from the
source-resolved part number without sanitization. Part numbers containing `/`
(e.g. Microchip's `MCP6001UT-I/OT`) silently created a subdirectory
(`MCP6001UT-I/` containing `OT.pdf`), and a hostile/buggy source returning a
traversal-shaped name (`../../something`) could write outside `-o`. The cache
path branch already sanitized inline, so this was an inconsistency between the
two output-path branches, not a missing concept.

## Changes

- **`datasheet/utils.py`**: add shared `sanitize_filename_component()` — allowlist
  (`A-Za-z0-9._-`), replaces path separators / other unsafe chars with `_`, and
  `lstrip(".")` so a name can never resolve to `.`/`..`; falls back to
  `unnamed_part` if empty.
- **`datasheet/manager.py`**: apply the helper at the `output_dir` filename site
  (`MCP6001UT-I/OT` → `MCP6001UT-I_OT.pdf`, single file, no subdir). Add a
  belt-and-suspenders `resolve()`/`parents` confinement check that raises
  `DatasheetDownloadError` if the constructed path would escape `output_dir`.
- **`datasheet/cache.py`**: refactor `get_datasheet_path()` to reuse the shared
  helper for consistency (behavior-equivalent for real part numbers; the only
  difference is `.` is now preserved, which no real cache key relies on).
- **`tests/test_datasheet.py`**: `TestSanitizeFilenameComponent` unit tests +
  three `TestDatasheetManager` download tests (slash, traversal, normal).

| `/`-containing part → single sanitized file, no subdir | ✅ | `test_download_sanitizes_slash_in_part_number` asserts `MCP6001UT-I_OT.pdf` and `not (out/"MCP6001UT-I").exists()` |
| Traversal-shaped part confined to `-o` | ✅ | `test_download_rejects_path_traversal_in_part_number` asserts resolved parent == output_dir, nothing written outside |
| Normal part number unaffected | ✅ | `test_download_normal_part_number_unaffected` asserts `STM32F103C8T6.pdf` unchanged |
| `.pdf` extension preserved | ✅ | each download test asserts `.suffix == ".pdf"` |
| Cache `get_datasheet_path()` behavior unchanged | ✅ | `test_get_datasheet_path` + full `TestDatasheetCache` still pass |

## Test Plan

- `uv run pytest tests/test_datasheet.py -q` → 165 passed
- `uv run ruff check` + `ruff format --check` → clean
- `uv run --extra dev python scripts/ci/check_mypy_baseline.py` → 0 new errors (within baseline)

Closes #4183

### DIFF
--- a/src/kicad_tools/datasheet/cache.py
+++ b/src/kicad_tools/datasheet/cache.py
@@ -17,6 +17,7 @@ from datetime import datetime, timedelta
 from pathlib import Path
 
 from .models import Datasheet
+from .utils import sanitize_filename_component
 
 logger = logging.getLogger(__name__)
 
@@ -170,8 +171,10 @@ class DatasheetCache:
 
     def get_datasheet_path(self, part_number: str) -> Path:
         """Get the expected path for a datasheet file."""
-        # Normalize part number for filesystem
-        safe_name = "".join(c if c.isalnum() or c in "-_" else "_" for c in part_number)
+        # Normalize part number for filesystem using the shared sanitizer so
+        # the '-o'/output-dir branch in DatasheetManager.download() and the
+        # cache path stay consistent (both reject path separators / traversal).
+        safe_name = sanitize_filename_component(part_number)
         return self.cache_dir / safe_name / "datasheet.pdf"
 
     def is_cached(self, part_number: str, ignore_expiry: bool = False) -> bool:
--- a/src/kicad_tools/datasheet/manager.py
+++ b/src/kicad_tools/datasheet/manager.py
@@ -17,6 +17,7 @@ from .cache import DatasheetCache
 from .exceptions import DatasheetDownloadError, DatasheetSearchError
 from .models import Datasheet, DatasheetResult, DatasheetSearchResult
 from .sources import DatasheetSource, LCSCDatasheetSource, OctopartDatasheetSource
+from .utils import sanitize_filename_component
 
 if TYPE_CHECKING:
     pass
@@ -186,7 +187,22 @@ class DatasheetManager:
         if output_dir:
             output_dir = Path(output_dir)
             output_dir.mkdir(parents=True, exist_ok=True)
-            output_path = output_dir / f"{result.part_number}.pdf"
+            # Sanitize the resolved part number before using it as a filename:
+            # source-resolved names commonly contain '/' (e.g. Microchip
+            # "MCP6001UT-I/OT"), which would otherwise silently create a
+            # subdirectory, and a hostile/buggy source could return a
+            # traversal-shaped name that escapes output_dir.
+            safe_name = sanitize_filename_component(result.part_number)
+            output_path = output_dir / f"{safe_name}.pdf"
+            # Belt-and-suspenders: confirm the constructed path stays inside
+            # output_dir. Redundant given the allowlist sanitizer above, but
+            # guards against a future regression that loosens it.
+            resolved = output_path.resolve()
+            output_dir_resolved = output_dir.resolve()
+            if output_dir_resolved not in resolved.parents and resolved != output_dir_resolved:
+                raise DatasheetDownloadError(
+                    f"Refusing to write outside output directory: {output_path}"
+                )
         else:
             output_path = self.cache.get_datasheet_path(result.part_number)
 
--- a/src/kicad_tools/datasheet/utils.py
+++ b/src/kicad_tools/datasheet/utils.py
@@ -4,7 +4,46 @@ Utility functions for datasheet processing.
 
 from __future__ import annotations
 
+import re
+
 from kicad_tools.utils.scoring import calculate_string_confidence
+
+# Characters that are safe to keep verbatim in a single filename/directory
+# component. Everything else (path separators, whitespace, shell/FS-hostile
+# characters, unicode look-alikes, NTFS ADS ':', etc.) is replaced with '_'.
+_UNSAFE_FILENAME_CHARS = re.compile(r"[^A-Za-z0-9._-]")
+
+
+def sanitize_filename_component(name: str) -> str:
+    """
+    Sanitize a string for safe use as a single filename/directory component.
+
+    Uses a character allowlist (``A-Za-z0-9._-``): path separators ('/', '\\\\',
+    ``os.sep``) and every other filesystem-unsafe character are replaced with
+    '_'. Leading dots are then stripped so the result can never resolve to '.'
+    or '..' (preventing a path-traversal escape when this value is joined onto
+    a base output directory). Falls back to ``"unnamed_part"`` if the result
+    would otherwise be empty.
+
+    Examples::
+
+        sanitize_filename_component("MCP6001UT-I/OT")  # -> "MCP6001UT-I_OT"
+        sanitize_filename_component("../../escape")     # -> "_.._escape"
+        sanitize_filename_component("STM32F103C8T6")    # -> "STM32F103C8T6"
+
+    Args:
+        name: The raw string to sanitize (e.g. a source-resolved part number).
+
+    Returns:
+        A safe single filename component containing only allowlisted characters
+        and never a leading dot.
+    """
+    # The allowlist handles '/', '\\', os.sep, and anything else unsafe.
+    safe = _UNSAFE_FILENAME_CHARS.sub("_", name)
+    # Strip leading '.' sequences so "..", ".", "...foo" cannot produce a
+    # relative-path escape or a hidden file.
+    safe = safe.lstrip(".")
+    return safe or "unnamed_part"
 
 
 def calculate_part_confidence(query: str, part_number: str) -> float:
--- a/tests/test_datasheet.py
+++ b/tests/test_datasheet.py
@@ -18,7 +18,10 @@ from kicad_tools.datasheet.models import (
     DatasheetSearchResult,
 )
 from kicad_tools.datasheet.tables import ExtractedTable
-from kicad_tools.datasheet.utils import calculate_part_confidence
+from kicad_tools.datasheet.utils import (
+    calculate_part_confidence,
+    sanitize_filename_component,
+)
 
 
 class TestCalculatePartConfidence:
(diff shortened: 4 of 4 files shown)


rjwalters:
✅ **Approved!**

Clean, correctly-scoped fix for the `/`-in-part-number filename bug (#4183). Verified against every acceptance criterion.

## Correctness

- **`sanitize_filename_component()`** (`datasheet/utils.py`) uses a character *allowlist* (`A-Za-z0-9._-`) — the robust choice; no exotic separator (`\`, `:`/NTFS-ADS, os.sep, unicode look-alikes) can slip past a denylist. `lstrip(".")` neutralizes `..`/`.`/hidden-file payloads, and the empty-result fallback (`unnamed_part`) is covered.
- **`manager.py` output_dir branch** now sanitizes before joining, so `MCP6001UT-I/OT` → single `MCP6001UT-I_OT.pdf` with no subdirectory. The belt-and-suspenders `resolve()`/`parents` confinement check is sound: because `safe_name` can no longer contain a separator, `output_dir` is always in `resolved.parents` for legitimate input; if a future regression re-introduced a `/`, `resolve()` would collapse the `..` and the escaped path would fail `output_dir_resolved not in resolved.parents`, correctly raising `DatasheetDownloadError`. Defense-in-depth works as intended.
- **`cache.py` refactor is behavior-preserving in the way that matters**: both the write side (`put` via `get_datasheet_path`) and the lookup side (`is_cached`/`get_datasheet_path`) now call the same shared helper, so they stay mutually consistent — no cached file can be "missed." The only delta from the old inline sanitizer is that `.` is now preserved instead of `_`-replaced; for a pre-existing on-disk entry whose part number contained a literal `.` this is a one-time cache miss (harmless re-download of a regenerable artifact), not a correctness bug. No real part number relies on a bare `.`-only segment.

## Notes (non-blocking, acceptable)

- Sanitization can collapse two distinct inputs to one filename (`A/B` and `A_B` both → `A_B`). Inherent to any allowlist sanitizer and acceptable for this use case; the resolved part number is already the more-specific value.
- `lstrip(".")` also strips a legitimate leading dot (e.g. `.hidden` → `hidden`). Vanishingly rare for part numbers and the safe default.

## Verification

**Test Scoping:** `full-pytest` (scoped to `tests/test_datasheet.py`)
- `uv run pytest tests/test_datasheet.py -q` → **165 passed** (10 new cases: 7 unit + 3 manager download)
- `uv run ruff check` + `ruff format --check` on touched files → clean
- `check_mypy_baseline.py` → **0 new errors** (1475 within 1514 baseline)

**CI:** 16/16 checks pass (all board E2E, routing regressions, Test, Type Check, Lint). `mergeStateStatus`: **CLEAN**.

Scope stayed tight to the datasheet module and the PR uses `Closes #4183`. Nice work.

PR #4192, opened 2026-07-15
fix(router): rescue budget-starved stranded nets after negotiation (#4159)

### LINKED ISSUE #4159: route (negotiated): cross-board nets starve and fail at ~25-30/76 on a sparse board — each failed net routes solo in <1s
## Bug (routing quality — batch negotiation starves long-hauls)

On a sparse 160×100 4-layer board (103 parts, 76 nets, generous spacing, C++ backend), `kct route --strategy negotiated --targeted-ripup` plateaus at **24–30/76 complete** across five configurations (default classes, tuned class map with priority 1 on the long-hauls, hybrid ordering, ±placement feedback, seeds fixed). The failures are dominated by cross-board power/long-haul nets.

Each such net then routes **solo in under a second** on the same copper:

```
$ kct route-auto board.kicad_pcb --net "/FUSED_LINE" ...
Routed net '/FUSED_LINE' successfully  (GLOBAL_WITH_REPAIR, 155.06mm, 11 segments, 0.8s)
```

So the geometry is easy; the batch negotiation never commits the long routes. Symptoms consistent with per-net search budget exhaustion under negotiation pressure rather than genuine congestion (the `--why` classifier's evidence lines show "sparse neighbourhood, 0 foreign obstructions" on many of them — see companion issue on the resulting misclassification).

Notably, `priority: 1` in `--net-class-map` (the ROUTE_FIRST idiom your own rev-A softstart recipe uses, with comments describing exactly this failure mode and the same alone-on-empty-board-routes-in-2s diagnosis) did NOT rescue these nets here — either the priority key isn't the primary ordering key in this code path, or the budget dies regardless of order.

## Repro

rjwalters/softstart PR #28: `hardware/kicad/output_revb/softstart_revb.kicad_pcb` (post-`place_revb.py`), then
`PYTHONHASHSEED=0 kct route <pcb> --layers 4 --skip-nets "GND,+3.3V" --seed 42 --timeout 3000 --targeted-ripup`
→ ~27/76. Then `kct route-auto --net "/FUSED_LINE"` on the output → instant success.

## Suggested directions

1. **Post-pass sweep**: after negotiation converges, run the global single-net strategy over every remaining failed net (they're cheap — sub-second each). This alone would likely have taken us from 27 to ~all.
2. Scale per-net search budget with net bbox / HPWL so a 155 mm net doesn't get a 20 mm net's iteration cap.
3. Honor `--net-class-map` `priority` as the primary ordering key in the negotiated path (or document what is).

Workaround that unblocked us: iterate `kct route-auto` per failed net over the batch output (repair-enabled) — i.e. exactly fix (1) done by hand in a shell loop.

**Additional datapoint (comment from reporter)**: `--preserve-existing` over a 46/77-complete artifact (fresh seed, 2400s budget) returned a board with only **22/77** complete — i.e. the preserve pass *destroyed* previously-complete nets rather than merely failing to add. The rev-A softstart recipe comments already record preserve-existing being counterproductive for negotiation ("frozen long-hauls partition the signal layers"), but a net *regression* of completed copper suggests the preserved set isn't actually being honored (or repair rips it). This is a secondary symptom — track separately if it recurs after the fix below; it should NOT block or be folded into this PR's scope.

---

### Root cause analysis

Two structurally different "single-net" code paths exist in this codebase and the issue's repro straddles both — this distinction matters for scoping the fix correctly:

1. **`Autorouter._route_net_negotiated`** (`src/kicad_tools/router/core.py:10865`) — the fine-grained A* pathfinder that `route_all_negotiated`'s batch loop calls internally, per net, under a derived per-net search-budget cap (`derive_per_net_cap` / `derive_iter_per_net_cap` in `src/kicad_tools/router/algorithms/negotiated.py`). This is the same grid/pathfinder state (`self.grid`, `self.router`) as the batch run.
2. **`RoutingOrchestrator._route_global`** (`src/kicad_tools/router/orchestrator.py:622`), reached via `kct route-auto` / `route_net_auto` (`src/kicad_tools/mcp/tools/routing.py:399`) and reported as strategy `GLOBAL_WITH_REPAIR`. This is a **separate, coarser tile-corridor planner** (`GlobalRouter` + `RegionGraph`, `src/kicad_tools/router/global_router.py:67`) — it is NOT the same algorithm as the negotiated batch's A* search.

The issue's repro log (`Routed net '/FUSED_LINE' successfully  (GLOBAL_WITH_REPAIR, 155.06mm, ...)`) is unambiguously the `route-auto` / `RoutingOrchestrator` path (confirmed: the `"Routed net '{name}' successfully"` print format is unique to `src/kicad_tools/cli/commands/routing.py:127`, the `route-auto` handler). So the reporter's manual "workaround" shell loop used the coarse `GlobalRouter`, not the same pathfinder the batch loop starved on.

This does **not** invalidate the finding — it strengthens the "budget, not congestion" diagnosis (two independent algorithms both solved the net trivially alone) — but it means:
- The literal "run the global single-net strategy" wording in suggested-direction 1 should be read as "run *a* single-net strategy on existing copper," not necessarily `RoutingOrchestrator`/`GLOBAL_WITH_REPAIR` specifically.
- The **recommended MVP implementation** should reuse `Autorouter.route_net()` (`core.py:2083`) — the same fine-grained pathfinder, operating on the *same in-memory grid* the negotiated loop already built — rather than round-tripping through `RoutingOrchestrator`'s separate coarse planner. This is both more architecturally consistent (no risk of the coarse planner producing DRC-dirty corridor-centerline geometry, see `_route_global`'s own docstring: "not guaranteed to be DRC-clean") and avoids a full PCB-file reload/reparse per failed net.

### Assessment of suggested direction 1 (post-pass sweep) — RECOMMENDED MVP

**The machinery already exists and the hook point is clean.** `route_all_negotiated` (`core.py:7110`) already tracks everything a post-pass sweep needs at its tail, right before `self._finalize_routing()` / `return list(self.routes)` (`core.py:10429-10448`):

- `net_routes: dict[int, list[Route]]` — per-net route state
- `self.routing_failures: list[RoutingFailure]` — the failed-net registry
- A `_stranded_nets()` helper (`core.py:8112`) that reports still-disconnected nets
- `successful_nets` / `total_nets` counters already used for status reporting

There is direct precedent for exactly this pattern — "bounded pass that runs after the negotiated loop converges, iterates failed/flawed nets, and commits improvements without regressing the board" — in the code immediately preceding the return statement:
- `_post_route_clearance_correction` (correction pass with a deadline + rollback-on-timeout, `core.py:10322`)
- `_demote_seg_seg_overlap_nets`, `_demote_pad_clearance_violation_nets`, `_demote_via_segment_violation_nets` (safety-net passes with clear before/after reporting)

A post-pass sweep is architecturally the same shape as these, but rescuing rather than demoting: for each net still in `self.routing_failures` (or absent from `net_routes`) after the negotiated loop's stall/timeout/convergence exit, call `Autorouter.route_net(net_id, ...)` on the **current** grid state (existing committed copper from other nets is already marked as obstacles on `self.grid` — this is the same mechanism `load_existing_routes` uses in `io.py:3755` for `--preserve-existing`, just already in-memory rather than reloaded from a file). Commit any net that succeeds; leave failures as-is (no regression risk — a failed sweep attempt for a net that was already failing costs nothing net-negative as long as any partial/rip-up state from the failed attempt is rolled back, mirroring the correction pass's rollback-on-timeout pattern).

**This is a clean, bounded builder PR**, not deep router surgery:
- Single new method (e.g. `_post_negotiation_sweep`) called once near the existing hook site, following the `_post_route_clearance_correction` pattern for budget/deadline handling and rollback.
- No changes to the core A* pathfinder, the congestion-cost model, or the rip-up/reroute negotiation loop itself.
- Should reuse `Autorouter.route_net()` directly (in-process, same grid) rather than shelling out to `route-auto` (avoids file I/O round-trip and the coarse-planner mismatch noted above).
- Needs a bound: reuse the existing `GRACE_PASS_BUDGET_S`-style pattern (`algorithms/negotiated.py:58`) — an overall wall-clock cap on the sweep pass, plus a per-net cap (the existing `per_net_timeout` / `derive_per_net_cap` machinery) so a net that fails solo too doesn't hang the sweep.

### Assessment of suggested directions 2 and 3

**Direction 3 (honor `--net-class-map priority` as primary ordering key) is very likely a documentation/diagnosis task, not a code defect** — investigation during curation shows `_get_net_priority` (`core.py:4472`) already sorts primarily by `net_class.priority` (see docstring: "1. Net class priority (1-10 for signal nets, 99 for pour nets)"), and this is used to seed the base `net_order` for the negotiated loop (`core.py:7460-ish`, `self._forced_net_order` / legacy priority sort). Priority ordering only controls *which net gets a turn first*; it does not increase a losing net's per-net search budget or protect it from later rip-up cycles once budget-starved. This is consistent with the reporter's own observation ("priority did NOT rescue them") — ordering alone can't fix a budget-exhaustion failure mode. **Recommend**: out of scope for this PR. If a real ordering defect is later found (e.g. priority silently dropped somewhere in the reroute/stagnation-recovery paths), file a separate, narrowly-scoped issue with a concrete repro showing priority being ignored.

**Direction 2 (scale per-net search budget with net bbox/HPWL) touches a mature, heavily-tuned, and already historically fragile budget subsystem** (`src/kicad_tools/router/algorithms/negotiated.py`: `derive_per_net_cap`, `derive_iter_per_net_cap`, the tiered grace-pass system, `PER_NET_CAP_STAGE_FRACTION`, `GRACE_PASS_TIER_CAPS_S`, `GRACE_CALL_OVERRUN_ABORT_MULT`) — the extensive in-code comments describe multiple past incidents (#3474, #3877, #3881, #3956, #3989) of budget-derivation bugs causing multi-minute overruns or starvation regressions. Changing the derivation to be geometry-aware (bbox/HPWL-scaled) is a real, valuable improvement but is **not a bounded, low-risk change** — it interacts with per-iteration budget shrinkage, the deterministic-iteration backstop (`--deterministic-budget`), and multiple call sites (initial pass, sequential reroute, stagnation recovery, zero-overflow recovery). **Recommend**: defer to a dedicated follow-up issue, scoped narrowly (e.g. "derive per-net cap floor from net HPWL at the initial-pass site only") and likely warranting architect review given the subsystem's history, rather than folding into this PR.

### Scope recommendation

**Recommend fix (1) — post-pass sweep — as the MVP for this issue.** It is the highest-value, most bounded option: the hook site, obstacle-tracking, and rollback patterns already exist in the codebase; it requires no changes to the core pathfinder or congestion model; and the reporter's own manual workaround (shell loop of `route-auto`) already validated the approach end-to-end (27 → ~all nets on the real board). This is **build-ready as a direct Builder PR**, not an architect referral — the design surface is a single new bounded post-loop pass following an established in-file pattern (`_post_route_clearance_correction`).

Directions (2) and (3) are explicitly **out of scope** for this PR:
- (3) is likely a non-issue / documentation gap — no code change recommended without a fresh, isolated repro.
- (2) is a legitimate follow-up but touches a fragile, well-precedented-for-regressions subsystem and should get its own issue + possible architect scoping pass.

The `--preserve-existing` regression noted in the reporter's comment (46/77 → 22/77) is a **distinct bug** and should not block or be folded into this PR. If it persists after the post-pass sweep ships, file a separate issue — do not conflate "preserve-existing destroys prior copper" with "negotiated batch starves long-haul nets."

### Implementation guidance

- Add a new bounded post-loop pass in `route_all_negotiated` (`src/kicad_tools/router/core.py`), inserted after the existing demote-safety-net calls and before `self._finalize_routing()` (around `core.py:10429`).
- For each net in `self.routing_failures` (dedupe by net ID) or absent from `net_routes` with `len(pads) >= 2`, attempt `self.route_net(net_id, per_net_timeout=<bounded>)` — reusing the SAME `Autorouter` instance/grid so existing committed copper is already present as obstacles (no reload needed).
- On success (non-empty route list, no new clearance violations vs. the safety-net checks already run above), commit into `self.routes` / `net_routes` and remove the net from `self.routing_failures`.
- On failure, leave state untouched (the attempt should not leave partial geometry marked on the grid — verify `route_net()`'s failure path already cleans up, or wrap in the same snapshot/rollback helper the correction pass uses, see `core.py:607-647` `_snapshot_failures` / restore pattern).
- Bound the whole pass: overall wall-clock cap (mirror `GRACE_PASS_BUDGET_S` sizing logic — scale from remaining stage budget, or a fixed generous ceiling since each net is expected to be sub-second per the issue's own evidence) AND a per-net cap so a net that also fails solo doesn't consume the whole sweep.
- Emit a clear log line mirroring the existing demote-pass style, e.g. `f"\n  ✓ Post-negotiation sweep rescued {rescued} net(s): {sorted(...)} (now {successful_nets}/{total_nets} routed)"`.
- Gate this behind no new CLI flag initially (always run when `route_all_negotiated` is used) unless the Builder finds a reason it should be opt-out — keep the PR small.
- **Requires the C++ router backend built** (`uv run kct build-native --check` → "available") — this touches core router-loop code exercised by the C++-backed A* search; do not evaluate correctness against the pure-Python fallback.

### Affected Files

- `src/kicad_tools/router/core.py` — add the post-negotiation sweep pass inside `route_all_negotiated` (hook near line 10429, before `_finalize_routing()` / `return list(self.routes)`); reuses `self.routing_failures`, `net_routes`, `route_net()`.
- `src/kicad_tools/router/algorithms/negotiated.py` — reference only; reuse `derive_per_net_cap` / the grace-pass budget-tiering pattern for the sweep's bound (no change expected unless the Builder needs a new small constant, e.g. `POST_NEGOTIATION_SWEEP_BUDGET_S`).
- `tests/router/` — add a new synthetic test (e.g. `tests/router/test_post_negotiation_sweep.py`) following the pattern of `tests/router/test_initial_pass_grace.py` / `tests/router/test_budget_integrity_3474.py`.

### Test Plan

- [ ] **Synthetic fixture**: build (or reuse an existing small synthetic-board test harness in `tests/router/`) a board where `route_all_negotiated` is deliberately made to leave one or more long-haul nets unrouted after convergence — e.g. via a tiny/artificial per-net budget cap plus a genuinely easy long net — while confirming that same net routes successfully when attempted solo via `Autorouter.route_net()` on the same grid state. This reproduces the issue's core symptom without depending on the (local-only, not-in-CI) softstart fixture.
- [ ] **Sweep effectiveness**: after the post-pass sweep runs, assert the previously-failed net(s) are present in the final `self.routes` / `net_routes`, and `self.routing_failures` no longer contains them.
- [ ] **No regression**: assert that nets already successfully routed by the negotiated batch loop are unchanged (same route count / geometry) after the sweep runs — the sweep must be additive-only.
- [ ] **Bounded**: assert the sweep pass has a hard wall-clock ceiling and a per-net cap — construct a case where a "failed" net also fails solo (e.g. genuinely impossible geometry) and confirm the sweep terminates promptly rather than hanging, and does not regress the board when it can't rescue a net.
- [ ] **Existing suite**: `uv run pytest tests/router/ -k negotiated` and the existing `tests/router/test_initial_pass_grace.py` / `tests/router/test_budget_integrity_3474.py` / `tests/test_negotiated_convergence_stranded.py` continue to pass (no interaction with the new sweep's rollback path).
- [ ] Note: the original softstart repro (rjwalters/softstart PR #28) is **local-only and not available in CI** — do not depend on it for automated verification; it may be used for manual confirmation only.

### Dependencies

None. This issue has no blocking dependencies.

rjwalters:
## Summary

On a sparse board the negotiated batch loop can exhaust a long-haul net's per-net search budget under negotiation pressure and strand it, even though the net routes solo in <1s on the identical copper (the reported softstart rev-B symptom: ~27/76 complete, then each failed net routes instantly via a solo attempt). This is **budget exhaustion, not congestion** (sparse board, 0 obstructions).

This PR adds a bounded, strictly-additive **post-negotiation rescue sweep** (fix 1, the curator-confirmed MVP). After the negotiated batch loop converges/stalls/times out AND the demote safety nets have run, `route_all_negotiated` re-attempts every still-stranded net **SOLO on the LIVE grid** via `Autorouter.route_net()` — every committed route is already an obstacle, so the solo attempt settles into the free space the batch loop never used. It reuses the **in-process fine-grained pathfinder** (per the curator's correction), NOT a `route-auto` shell-out to the separate coarse `GlobalRouter`.

## Behavior

Follows the sibling bounded post-loop pattern (`_post_route_clearance_correction`, the `_demote_*` safety nets):

- One attempt per net, capped by a per-net budget (`POST_NEGOTIATION_SWEEP_PER_NET_S`).
- Overall wall-clock ceiling (`POST_NEGOTIATION_SWEEP_BUDGET_S`) so a genuinely-impossible stranded net cannot hang the sweep.
- Success commits the net into `net_routes` and clears its stale `RoutingFailure` records; a partial (not-fully-connected) result is **not** a rescue.
- Failure rolls back verbatim (rip up any copper the attempt marked, restore pre-attempt state, drop fresh failure diagnostics) — the sweep can only ever **raise** the routed count, never regress it. Already-successful routes are never touched.
- **ON by default** (strict improvement). `--no-rescue-pass` disables it (both parsers + forwarding shim wired). Emits a summary line when it recovers nets.

## Hook site

`src/kicad_tools/router/core.py`, inside `route_all_negotiated`, right after the `_demote_*` safety nets and before `progress_callback` / `_finalize_routing()`, reusing the in-scope `_stranded_nets()`, `net_routes`, `pads_by_net`, and `self.routing_failures`.

## Reproduction status

**Reproduced end-to-end synthetically.** The batch's internal `_route_net_negotiated` is patched to hard-fail a specific net while the sweep's `route_net` path stays real — the exact two-path split the curator identified. The batch loop strands the long net; the rescue sweep recovers it (final 2/2). Also unit-tested the mechanism directly: rescue a routable stranded net, roll back an unroutable/partial attempt without regression, additive-only, bounded-by-deadline. The softstart repro fixture is local-only (not in CI), so all tests are synthetic.

## Out of scope (deferred, per curator)

- **Fix (2) bbox/HPWL budget scaling**: touches the historically fragile budget-derivation subsystem (#3474/#3877/#3881/#3956/#3989); warrants its own issue + architect scoping.
- **Fix (3) priority-ordering**: confirmed a non-issue — `_get_net_priority` already sorts by net-class priority as the primary key. No code change.

## Tests

- New `tests/router/test_post_negotiation_sweep.py` (11 tests): constants, direct sweep mechanism (rescue / rollback / additive-only / bounded / no-op), end-to-end starvation-then-rescue, disabled flag, summary log, no-op-when-nothing-stranded.
- Updated `tests/test_negotiated_convergence_stranded.py::test_final_status_is_stranded_not_converged` to disable the sweep (it pins the RAW negotiated `stranded` status; the sweep would legitimately rescue that fixture's net).
- `tests/router/` full suite: green except one **pre-existing, change-unrelated** platform baseline drift (`test_board02_manufacturable_baseline::test_routing_output_deterministic_across_seeds` — 476 segments vs documented band [390,398]; verified identical failure on clean origin/main via stash; board 02 routes 8/8 so the sweep is a no-op there).
- `ruff check`/`format` clean; `check_mypy_baseline.py` → 0 new errors.

Closes #4159

### DIFF
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -217,6 +217,11 @@ def run_route_command(args) -> int:
     # what makes flag-off byte-identical (sub_argv unchanged when absent).
     if getattr(args, "monotone_certificate_order", False):
         sub_argv.append("--monotone-certificate-order")
+    # Issue #4159: forward --no-rescue-pass.  Both parsers declare it as
+    # store_true defaulting to False (rescue sweep ON by default), so only
+    # forward when the user set it (byte-identical when absent).
+    if getattr(args, "no_rescue_pass", False):
+        sub_argv.append("--no-rescue-pass")
     if getattr(args, "cross_package_pair_corridor", False):
         sub_argv.append("--cross-package-pair-corridor")
     if getattr(args, "slack_corridor_widening", False):
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -2858,6 +2858,20 @@ def _add_route_parser(subparsers) -> None:
         ),
     )
     route_parser.add_argument(
+        "--no-rescue-pass",
+        action="store_true",
+        default=False,
+        help=(
+            "Disable the post-negotiation rescue sweep (Issue #4159).  ON by "
+            "default: after the negotiated batch loop converges/stalls/times "
+            "out, each still-stranded net is re-attempted SOLO on the live "
+            "grid, recovering long-haul nets the batch loop starved on per-net "
+            "search budget.  Bounded and strictly additive (failed attempts "
+            "roll back), so it can only raise the routed count.  Pass this "
+            "flag for the raw negotiated result (e.g. A/B comparison)."
+        ),
+    )
+    route_parser.add_argument(
         "--cross-package-pair-corridor",
         action="store_true",
         default=False,
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -2928,6 +2928,21 @@ def _apply_ripup_budget_override(router: "Autorouter", args) -> None:
     router.stall_ripup_budget = budget
 
 
+def _apply_rescue_pass_override(router: "Autorouter", args) -> None:
+    """Disable the post-negotiation rescue sweep when requested (Issue #4159).
+
+    ON by default (``router._post_negotiation_rescue`` initialised True): after
+    the negotiated batch loop converges/stalls/times out, each still-stranded
+    net is re-attempted SOLO on the live grid, recovering long-hauls the batch
+    loop starved on per-net budget.  The pass is bounded and strictly additive
+    (failed attempts roll back), so it only raises the routed count.
+    ``--no-rescue-pass`` sets the flag False for A/B comparison or when a caller
+    wants the raw negotiated result.  A no-op when the flag is absent.
+    """
+    if getattr(args, "no_rescue_pass", False):
+        router._post_negotiation_rescue = False
+
+
 def _apply_bundle_river_planner(router: "Autorouter", args) -> None:
     """Enable the scoped bundle river planner when requested (Issue #4053).
 
@@ -3848,6 +3863,7 @@ def route_with_layer_escalation(
         # Issue #3470: thread --max-ripups-per-net into the destructive
         # rip-up budgets (route_all + two-phase stall recovery).
         _apply_ripup_budget_override(router, args)
+        _apply_rescue_pass_override(router, args)
         _apply_bundle_river_planner(router, args)
         _apply_monotone_certificate_order(router, args)
         _apply_cross_package_pair_corridor(router, args)
@@ -4729,6 +4745,7 @@ def route_with_rule_relaxation(
         # Issue #3470: thread --max-ripups-per-net into the destructive
         # rip-up budgets (route_all + two-phase stall recovery).
         _apply_ripup_budget_override(router, args)
+        _apply_rescue_pass_override(router, args)
         _apply_bundle_river_planner(router, args)
         _apply_monotone_certificate_order(router, args)
         _apply_cross_package_pair_corridor(router, args)
@@ -6852,6 +6869,7 @@ def route_with_combined_escalation(
             # Issue #3470: thread --max-ripups-per-net into the destructive
             # rip-up budgets (route_all + two-phase stall recovery).
             _apply_ripup_budget_override(router, args)
+            _apply_rescue_pass_override(router, args)
             _apply_bundle_river_planner(router, args)
             _apply_monotone_certificate_order(router, args)
             _apply_cross_package_pair_corridor(router, args)
@@ -7933,6 +7951,20 @@ def main(argv: list[str] | None = None) -> int:
             "certificate finds the bundle infeasible as-pinned, order is left "
             "at IDENTITY (no regression vs. flag-off).  Default OFF "
             "(byte-identical to prior behaviour when absent)."
+        ),
+    )
+    parser.add_argument(
+        "--no-rescue-pass",
+        action="store_true",
+        help=(
+            "Disable the post-negotiation rescue sweep (Issue #4159).  ON by "
+            "default: after the negotiated batch loop converges/stalls/times "
+            "out, each still-stranded net is re-attempted SOLO on the live "
+            "grid, recovering long-haul nets the batch loop starved on per-net "
+            "search budget (each such net routes in <1s alone).  The pass is "
+            "bounded and strictly additive (failed attempts roll back), so it "
+            "can only raise the routed count.  Pass this flag to get the raw "
+            "negotiated result (e.g. for A/B comparison)."
         ),
     )
     parser.add_argument(
@@ -9716,6 +9748,7 @@ def main(argv: list[str] | None = None) -> int:
     # Issue #3470: thread --max-ripups-per-net into the destructive
     # rip-up budgets (route_all + two-phase stall recovery).
     _apply_ripup_budget_override(router, args)
+    _apply_rescue_pass_override(router, args)
     _apply_bundle_river_planner(router, args)
     _apply_monotone_certificate_order(router, args)
     _apply_cross_package_pair_corridor(router, args)
--- a/src/kicad_tools/router/algorithms/__init__.py
+++ b/src/kicad_tools/router/algorithms/__init__.py
@@ -23,6 +23,8 @@ from .negotiated import (
     GRACE_PASS_TIER_CAPS_S,
     PER_NET_CAP_FLOOR_S,
     PER_NET_CAP_STAGE_FRACTION,
+    POST_NEGOTIATION_SWEEP_BUDGET_S,
+    POST_NEGOTIATION_SWEEP_PER_NET_S,
     NegotiatedRouter,
     calculate_congestion_tuned_params,
     calculate_history_increment,
@@ -52,6 +54,9 @@ __all__ = [
     "GRACE_PASS_PER_NET_S",
     "GRACE_PASS_TIER_CAPS_S",
     "run_initial_pass_grace",
+    # Post-negotiation rescue sweep (Issue #4159)
+    "POST_NEGOTIATION_SWEEP_BUDGET_S",
+    "POST_NEGOTIATION_SWEEP_PER_NET_S",
     # Per-net A* cap derivation (Issue #3474 R1)
     "PER_NET_CAP_FLOOR_S",
     "PER_NET_CAP_STAGE_FRACTION",
--- a/src/kicad_tools/router/algorithms/negotiated.py
+++ b/src/kicad_tools/router/algorithms/negotiated.py
@@ -60,6 +60,18 @@ GRACE_PASS_BUDGET_S = 60.0
 # that reason about the upper bound of any single grace attempt.
 GRACE_PASS_PER_NET_S = GRACE_PASS_TIER_CAPS_S[-1]
 
+# Issue #4159: post-negotiation rescue-sweep bounds.  After the negotiated
+# batch loop converges/stalls/times out, one bounded pass re-attempts every
+# still-stranded net SOLO on the live grid (existing committed copper is
+# already present as obstacles).  The issue's own evidence is that each such
+# net routes in <1s alone, so a generous but hard wall-clock ceiling plus a
+# per-net cap keeps the whole sweep bounded even when a genuinely-impossible
+# net is in the stranded set.  The overall ceiling mirrors GRACE_PASS_BUDGET_S
+# (they are the same shape of bounded post-loop sweep); the per-net cap is
+# clamped so one pathological solo search cannot eat the whole sweep.
+POST_NEGOTIATION_SWEEP_BUDGET_S = 60.0
+POST_NEGOTIATION_SWEEP_PER_NET_S = 10.0
+
 # Issue #3474 R1: abort threshold for a single grace attempt that grossly
 # overruns its cap.  ``route_fn`` legally spends a small multiple of the
 # per-call cap (see the ~15x note above), but a 100x+ overrun means a
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -37,6 +37,8 @@ from .algorithms import (
     GRACE_PASS_BUDGET_S,
     GRACE_PASS_TIER_CAPS_S,
     PER_NET_CAP_STAGE_FRACTION,
+    POST_NEGOTIATION_SWEEP_BUDGET_S,
+    POST_NEGOTIATION_SWEEP_PER_NET_S,
     HierarchicalRouter,
     MonteCarloRouter,
     MSTRouter,
@@ -1300,6 +1302,15 @@ class Autorouter:
         # (``route_with_escape`` -> ``route_all_two_phase``).
         self._route_all_max_ripups_per_net: int = 2
         self.stall_ripup_budget: int | None = None
+
+        # Issue #4159: post-negotiation rescue sweep.  After the negotiated
+        # batch loop converges/stalls/times out, re-attempt each still-
+        # stranded net SOLO on the live grid (a long-haul the batch loop
+        # starved on budget routes in <1s alone).  ON by default -- the
+        # pass is bounded and strictly additive (failed attempts roll back)
+        # so it can only raise the routed count.  ``--no-rescue-pass`` sets
+        # this to False.
+        self._post_negotiation_rescue: bool = True
 
         # Pre-route congestion estimator (Issue #2278)
         # Computed lazily before net ordering; provides RUDY-based
@@ -10603,6 +10614,55 @@ class Autorouter:
                 f"(now {successful_nets}/{total_nets} routed)"
             )
 
+        # Issue #4159: post-negotiation rescue sweep.  On a sparse board the
+        # negotiated batch loop can exhaust a long-haul net's per-net search
+        # budget under negotiation pressure and strand it, even though the
+        # net routes solo in <1s on the identical copper.  After the loop
+        # converges/stalls/times out AND the demote safety nets above have
+        # run, re-attempt each still-stranded net SOLO on the LIVE grid (every
+        # committed route is already an obstacle) via ``route_net``.  The pass
+        # is strictly additive and bounded (one attempt per net, per-net cap,
+        # overall wall-clock ceiling) and rolls back any failed attempt, so it
+        # can only ever raise the routed count -- never regress it.  On by
+        # default (it is a strict improvement); ``--no-rescue-pass`` disables
+        # it.  Runs AFTER the demotes so it never re-attempts a net that was
+        # just demoted for an unrepairable clearance violation.
+        if self._post_negotiation_rescue:
+            sweep_stranded = _stranded_nets()
+            if sweep_stranded:
+                # Bound the whole sweep.  Solo long-hauls are sub-second (the
+                # issue's evidence), so a generous ceiling is ample; when the
+                # loop already timed out the remaining budget is spent, so the
+                # sweep runs on its own contained allowance.  The per-net cap
+                # keeps one genuinely-impossible net from eating the sweep.
+                sweep_remaining: float | None = None
+                if timeout is not None:
+                    sweep_remaining = timeout - (time.time() - start_time)
+                if sweep_remaining is not None and sweep_remaining > 0 and not timed_out:
+                    sweep_budget = min(POST_NEGOTIATION_SWEEP_BUDGET_S, sweep_remaining)
+                else:
+                    sweep_budget = POST_NEGOTIATION_SWEEP_BUDGET_S
+                sweep_deadline = time.time() + max(0.0, sweep_budget)
+                sweep_per_net = per_net_timeout
+                if sweep_per_net is None:
+                    sweep_per_net = POST_NEGOTIATION_SWEEP_PER_NET_S
+                else:
+                    sweep_per_net = min(sweep_per_net, POST_NEGOTIATION_SWEEP_PER_NET_S)
+                sweep_rescued = self._post_negotiation_sweep(
+                    stranded_nets=sweep_stranded,
+                    net_routes=net_routes,
+                    pads_by_net=pads_by_net,
+                    per_net_timeout=sweep_per_net,
+                    deadline=sweep_deadline,
+                )
+                if sweep_rescued:
+                    successful_nets = sum(1 for routes in net_routes.values() if routes)
+                    flush_print(
+                        f"\n  ✓ Post-negotiation rescue recovered "
+                        f"{len(sweep_rescued)} stranded net(s): {sorted(sweep_rescued)} "
+                        f"(now {successful_nets}/{total_nets} routed)"
+                    )
+
         if progress_callback is not None:
             # Issue #2597: Distinguish ``stagnated`` from ``timeout`` and
             # bare ``f"overflow={N}"`` so callers (and CI) can pick the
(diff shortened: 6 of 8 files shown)


rjwalters:
❌ **Changes Requested — CI routing regression (rescue sweep commits DRC-dirty copper)**

Thorough review of a HIGH-RISK router-core change. The design, rollback discipline, in-process `route_net` reuse, bounding, and CLI wiring are all correct and well-implemented — but **two CI routing-regression jobs fail because the rescue sweep introduces new DRC violations on board 07**, and both jobs pass on the PR's base (main @ `9def133d`). This is a real regression, not a flake.

## The blocker

**Board 07 re-routes to 13 blocking DRC errors on this branch vs the allowlist floor of 8** (`.github/routed-drc-tolerance.yml`):

- `Match-Group Routing Regression (boards/07-matchgroup-test)` — FAIL
  `[matchgroup-coverage] DRC error count: 13 (allowed: 8)`
  `DRC regression on re-routed board 07: 13 error(s) exceeds allowlist value 8`
- `Board 07 End-to-End (LVS known opens)` — FAIL
  `DRC regression: 13 blocking error(s) ... exceeds allowlist value 8`

I confirmed **both jobs are `success` on the base commit `9def133d`** (`gh run view 29390274164`), so the +5 DRC errors are introduced by this PR. The match-group *rule* is still exercised (`match_group_length_skew: 2`, baseline 1 — fine); the failure is the total blocking DRC count rising 8 → 13.

## Root cause

`_post_negotiation_sweep` gates a rescued net solely on **connectivity** (`validate_net_connectivity` → `connected`) before committing it. It does **not** validate the newly-committed copper for **DRC clearance** against the rest of the board. So on board 07 the sweep rescues previously-stranded nets by committing copper that connects the pads but violates clearance — adding 5 blocking DRC errors.

The PR's central claim — *"STRICTLY ADDITIVE … can only ever raise the routed count, never regress it"* — holds for the **net-completion count** but **not for the DRC-error count**. The `_demote_*` safety nets that scrub clearance violations run *before* the sweep, so rescued copper is never re-scrubbed. `route_net`'s A* commit-gating (`_validate_route_clearance`) evidently does not catch the specific clearance modes DRC counts here (e.g. the `diffpair_clearance_intra` / within-pair coupling violations visible in the log: `within-pair coupling clearance ... -0.375mm < 0.100mm`).

## What needs to change (pick one)

1. **Preferred — DRC-gate each rescue commit.** Before committing a rescued net, verify it introduces **zero net-new blocking DRC violations** (not just connectivity). If it does, roll it back exactly like the current failure path. This makes the "strictly additive" guarantee true for DRC too, and board 07 stays ≤8. This is the correct fix — the sweep should only commit copper that is both connected AND clean.
2. Alternatively, if the board-07 rescued nets are genuinely acceptable at 13 (owner sign-off), raise the two allowlist floors with an explicit documented rationale — but given the whole point of the sweep is *strict* improvement, silently trading net-count for DRC errors is the wrong trade and I'd want owner sign-off before that path.

Re-run the `matchgroup-routing-regression` and `board-07-end-to-end` gates after the fix; they are the acceptance bar.

## What I verified and is GOOD (no action needed)

- **In-process, not a shell-out.** `_post_negotiation_sweep` calls `self.route_net()` (core.py:2083) on the live `self.grid`/`self.router` — the fine-grained pathfinder, per the curator's key correction. Not a `route-auto`/`GlobalRouter` round-trip. ✅
- **Rollback correctness.** Snapshots `pre_routes` + `failures_len`; on a non-connected/partial result rips up via `NegotiatedRouter.rip_up_nets` (grid unmark + `self.routes` removal), restores `net_routes[net]=pre_routes`, re-marks pre-existing copper via `self._mark_route`, and truncates fresh `routing_failures`. Mirrors the `_post_route_clearance_correction` rollback discipline. Connectivity is verified before commit (partial ≠ rescue). The rollback itself is sound. ✅
- **Bounded / terminates.** Per-net cap (`min(per_net_timeout, POST_NEGOTIATION_SWEEP_PER_NET_S=10s)`), overall `deadline` checked at the top of each iteration, one attempt per net. A past deadline skips all nets (test verifies). No infinite loop; deadline firing mid-sweep leaves the board consistent (per-net commit/rollback is atomic). ✅
- **CLI wiring end-to-end.** `--no-rescue-pass` declared on BOTH parsers (`parser.py` outer + `route_cmd.py` inner `main`), forwarded through the shim (`commands/routing.py`), applied via `_apply_rescue_pass_override` at all four route-strategy call sites. Verified outer parse → True/False default, inner help advertises it, override flips `_post_negotiation_rescue` correctly. `test_cli_parser_drift` passes (flag is symmetric, not inner-only). ✅
- **Tests.** 11 new tests + the convergence-test update all pass locally with the C++ backend built. The end-to-end starvation test genuinely exercises the rescue path (patches `_route_net_negotiated` to hard-fail net 1 while the sweep's real `route_net` recovers it → 2/2). ✅
- **Board 02 baseline (`test_routing_output_deterministic_across_seeds`).** Independently verified this fails **identically on clean `origin/main`** (built native in a detached-`9def133d` worktree): `assert 476 <= 398`, same segment count, routes/vias/length all exact. **Pre-existing platform drift, NOT this PR** (board 02 routes 8/8, sweep is a no-op there). Not a blocker for #4192 — should be tracked separately (re-baseline the segment band or investigate the collinear-merge platform divergence).

Lint, format, type-check, and boards 00–04 + diff-pair regression all pass on the PR. The only blocker is the board-07 DRC regression above.

rjwalters:
## Board-07 DRC regression fixed — sweep is now additive for the DRC-error count

Thanks for the precise diagnosis. I reproduced the regression locally with the CI recipe (`PYTHONHASHSEED=42 check_matchgroup_coverage.py boards/07-matchgroup-test --seed 42`, C++ backend built, `--deterministic-budget` so it is platform-stable) and confirmed the mechanism, then fixed it.

### What the +5 actually was
Diffing the base-vs-branch routed DRC, the extra blocking errors were **not** clearance violations — they were `diffpair_length_skew` (+1) and `diffpair_routing_continuity` (+1) on **TMDS_D1**, i.e. a diff-pair *leg* that the sweep solo-rescued. `route_net` is single-ended, so rescuing one leg of a stranded pair connects it but breaks the pair's coupled length-matching / within-pair spacing (the `within-pair coupling clearance -0.375mm` you saw is the same leg drifting onto its partner). That is what pushed 8 → 13 (and also caused the `+1V2` pour-connectivity failure).

### Fix (both guards)
1. **Skip diff-pair legs in the sweep** — a stranded leg is the coupled router's job, not a solo re-route's. This is what restores board-07.
2. **DRC-gate each remaining single-ended rescue** (`_rescued_net_adds_blocking_drc`) — reuses the exact per-net validators the three `_demote_*` finalization safety nets already trust as blocking (seg-seg physical overlap, seg/via-vs-foreign-pad deficit, via-vs-foreign-seg deficit) plus the intra-pair detector. A connected-but-DRC-dirty rescue rolls back verbatim, same path as a connectivity failure. This delivers the invariant you asked for: **after the sweep, blocking-DRC-error count ≤ before.**

### Local verification (board-07, the load-bearing gate)
| run | DRC errors | pour | gate |
|---|---|---|---|
| base `9def133d` (no sweep) | **8** | PASS | exit 0 |
| branch, sweep on (before fix) | **13** | FAIL | exit 2 |
| branch, `--no-rescue-pass` | 8 | PASS | exit 0 (isolation control) |
| **branch + this fix** | **8** | **PASS** | **exit 0** |

- `tests/router/test_post_negotiation_sweep.py`: 14 passed (added `test_drc_dirty_rescue_rolls_back`, `test_drc_clean_rescue_is_committed`, `test_diffpair_leg_is_skipped`; the dirty-rescue test is connectivity-valid so it exercises the DRC gate, not the connectivity gate).
- `tests/test_negotiated_convergence_stranded.py`: 4 passed.
- ruff check + format clean; mypy baseline: 0 new errors.

Ignored the board-02 `test_routing_output_deterministic_across_seeds` baseline failure per your note (pre-existing on main, #4196). CI's `matchgroup-routing-regression` and `board-07-end-to-end` jobs are the real acceptance gate.

rjwalters:
✅ **Approved!** (post-Doctor re-review, commit 83b82b2c)

The two guards added to `_post_negotiation_sweep` in `src/kicad_tools/router/core.py` correctly close the board-07 DRC-dirty-rescue regression (blocking errors 8→13).

## Guard verification

**Guard 1 — diff-pair leg skip (core.py ~L173–186):** `get_diff_pair_map()` records each pair in **both directions** (`{p: n, n: p}`, confirmed at L4164–4201), so `self.net_names.get(net) in diffpair_leg_names` catches *either* leg — it can never miss a leg, and a legitimate single-ended net (never a key in the map) is never wrongly skipped. A stranded diff-pair leg is correctly deferred to the coupled router rather than solo-rescued.

**Guard 2 — `_rescued_net_adds_blocking_drc`:** genuinely reuses the SAME per-net blocking validators the three `_demote_*` finalization passes trust, bit-for-bit:
- seg-seg physical overlap: `find_segment_segment_violation_pairs(..., copper_overlap_only=True)` + `_collect_extra_routes_for_revalidation` — matches `_demote_seg_seg_overlap_nets`.
- foreign-pad clearance: `worst_segment_pad_deficit` / `worst_via_pad_deficit` vs `nudge_reach` (grid.resolution), same `exclude_net`/`component_pitches`/`exclude_refs` — matches `_demote_pad_clearance_violation_nets`.
- via-vs-foreign-segment: `worst_via_segment_deficit(..., foreign_routes=committed)` vs `nudge_reach` — matches `_demote_via_segment_violation_nets`.
- plus a diff-pair intra-pair clearance check via `find_intra_pair_clearance_violations`.

The additive invariant holds: a stranded net contributes zero copper pre-rescue, so any blocking violation from the fresh copper is net-new; a connectivity-valid-but-DRC-dirty rescue is rolled back verbatim (rip up marked copper, restore `pre_routes`, re-mark pre-existing copper, truncate fresh `routing_failures`). The sweep can raise the routed count but never the blocking-DRC count.

## Tests
The 3 new tests pass locally (`3 passed in 36.77s`) and exercise the guards precisely:
- `test_drc_dirty_rescue_rolls_back` — the dirty route **connects** net 1's pads ((2,10)→(38,10)), so it is connectivity-VALID and genuinely tests the DRC gate (not the connectivity gate); asserts net stranded, foreign net untouched, `ar.routes` length restored.
- `test_drc_clean_rescue_is_committed` — control: connected + clear of foreign copper → committed (gate isn't over-rejecting).
- `test_diffpair_leg_is_skipped` — asserts `route_net` is **never invoked** for a diff-pair leg (`attempts == []`).

## CI — load-bearing board-07 check (all 16 GREEN)
- **Board 07 End-to-End (LVS known opens)** — pass (17m40s)
- **Match-Group Routing Regression (boards/07-matchgroup-test)** — pass (14m48s)
- Test / Type Check / Lint & Format / Diff-Pair Regression / all board jobs — pass. mergeStateStatus CLEAN.

Board-02 baseline drift is out of scope (widened via #4196/PR #4198). `Closes #4159` present. Ready to merge.

PR #4193, opened 2026-07-15
fix(zones): default add/batch output to in-place, matching zones fill

### LINKED ISSUE #4166: zones add: default output writes <input>_zones.kicad_pcb instead of in-place — inconsistent with other mutating commands, silently derails pipelines
## Bug / consistency

`kct zones add <pcb> --net GND --layer In1.Cu` (no `-o`) does not modify `<pcb>`; it writes `<pcb stem>_zones.kicad_pcb` and prints the path in its (verbose) output. Other mutating commands default to in-place — e.g. `optimize-placement` documents "`-o` (default: overwrite input)" — so a pipeline that chains kct calls against one file silently diverges:

```bash
kct zones add "$PCB" --net GND --layer In1.Cu       # writes $PCB_zones, $PCB untouched
kct zones add "$PCB" --net +3.3V --layer In2.Cu     # ALSO writes $PCB_zones (from pristine $PCB — first zone lost too!)
kct zones fill "$PCB"                                # fills a board with no zones
kicad-cli pcb drc "$PCB"                             # gates a board that never got planes
```

Every downstream step exits happily. In our rev-B finish pipeline this made the entire plane/stitch/DRC stage a no-op chain against the un-zoned board; only reading the step-1 log closely revealed `Saving to: ..._zones.kicad_pcb`.

Note the compounding subtlety: two sequential `zones add` calls both derive from the pristine input, so the second **silently discards the first zone** even if you knew about the `_zones` convention.

## Suggested fix

Pick one and document it:
- make no-`-o` mean in-place (consistent with optimize-placement et al.), or
- keep the side-file default but **print the output path prominently at non-verbose level** and make chained usage safe (`zones add` accepting multiple `--net/--layer` pairs in one call would fix the discard hazard outright — `zones batch` exists but only takes the NET:LAYER spec form).

kicad-tools 0.15.0, softstart rev-B finish pipeline (rjwalters/softstart PR #28).

rjwalters:
## Summary

`kct zones add` / `kct zones batch` wrote `<stem>_zones.kicad_pcb` when `-o` was omitted, leaving the input untouched — inconsistent with `zones fill` and `optimize-placement`, which overwrite the input. This silently derailed chained pipelines: a natural `zones add → zones add → zones fill → drc` chain operated on the pristine input at each step, so the second `add` discarded the first's zone and `fill` gated a board with no planes. Every step exited 0.

This PR makes no-`-o` mean **overwrite input in place** for both `add` and `batch`, matching the project convention. Each chained in-place `add` now reads its own prior output, so zones accumulate instead of clobbering. Explicit `-o <path>` still writes a side file and leaves the input untouched.

Safe: `ZoneGenerator.from_pcb()` fully loads `PCB.load()` + `parse_file()` into memory before any `save()`, so self-overwrite can't clobber a mid-flight read (same pattern `placement`/`optimize-placement` already rely on).

## Changes

- `src/kicad_tools/cli/zones_cmd.py`: `_run_add` and `_run_batch` resolve `output_path = Path(args.output) if args.output else pcb_path` (was `pcb_path.with_stem(stem + "_zones")`). Updated `-o` help text for add/batch in this module's parser; dropped the `_zones` suffix mention.
- `src/kicad_tools/cli/parser.py::_add_zones_parser`: updated add/batch `-o` help to `"(default: overwrite input, consistent with 'zones fill')"` so both parser definitions agree.
- `docs/reference/cli.md`: documented the in-place default and added the missing `batch`/`fill` subcommand rows.
- `tests/test_zones_cmd.py`: new coverage (see below).

| 1. `add` no `-o` modifies `<pcb>` in place; no `<stem>_zones.kicad_pcb` created | ✅ | `TestAddInPlaceDefault` |
| 2. Two sequential `add` calls accumulate (second doesn't discard first) | ✅ | `TestAddChainedAccumulates` |
| 3. `batch` no `-o` in place; chaining a subsequent `add` accumulates | ✅ | `TestBatchInPlaceDefault`, `TestBatchChainedWithAdd` |
| 4. Explicit `-o` on add and batch writes target, leaves input byte-unchanged | ✅ | `TestAddExplicitOutputUnaffected`, `TestBatchInPlaceDefault::test_batch_explicit_output_leaves_input_unchanged` |
| 5. `fill` picks up zones from a preceding no-`-o` `add` (pipeline hole closed) | ✅ | `TestAddFillPipeline` |
| 6. `--dry-run` still reports correct in-place target | ✅ | Existing `TestFillDryRun` still green; add/batch dry-run unchanged (no write) |
| 7. Help text consistent between both parser definitions | ✅ | `TestZonesParserHelpConsistency` |
| 8. `docs/reference/cli.md` documents in-place default, lists batch/fill | ✅ | Diff |

## Test Plan

`uv run pytest tests/test_zones_cmd.py -q` → **74 passed**. ruff check + format clean. `scripts/ci/check_mypy_baseline.py` → 0 new type errors.

Key regression test — two sequential in-place adds accumulate:

```python
def test_two_sequential_adds_keep_both_zones(self, tmp_pcb):
    main(["add", str(tmp_pcb), "--net", "SDA", "--layer", "F.Cu"])
    main(["add", str(tmp_pcb), "--net", "SCL", "--layer", "In1.Cu"])
    keys = _zone_keys(tmp_pcb)
    assert ("SDA", "F.Cu") in keys    # first add survived the second
    assert ("SCL", "In1.Cu") in keys  # second add landed too
```

Closes #4166

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -660,10 +660,27 @@ kct zones <subcommand> <pcb_file> [options]
 |------------|-------------|
 | `add` | Add a copper zone |
 | `list` | List existing zones |
+| `batch` | Add multiple zones from a `NET:LAYER,...` spec |
+| `fill` | Fill all zones in a PCB (via `kicad-cli`) |
+
+**Output-path behavior:** for the mutating subcommands (`add`, `batch`,
+`fill`), `-o`/`--output` is optional and **defaults to overwriting the input
+in place** — consistent with `optimize-placement`. Because each in-place `add`
+reads its own prior output, chaining `zones add` calls against one file
+accumulates zones instead of silently discarding earlier ones. Pass an
+explicit `-o <path>` to write a side file and leave the input untouched.
+
+| Option | Description |
+|--------|-------------|
+| `-o`, `--output PATH` | Output PCB (default: overwrite input) |
 
 **Examples:**
 ```bash
-kct zones add board.kicad_pcb --net GND --layer B.Cu
+kct zones add board.kicad_pcb --net GND --layer B.Cu     # modifies board.kicad_pcb in place
+kct zones add board.kicad_pcb --net +3.3V --layer In2.Cu # accumulates onto the GND zone
+kct zones batch board.kicad_pcb --power-nets GND:B.Cu,+3.3V:F.Cu
+kct zones fill board.kicad_pcb                           # fills all zones in place
+kct zones add board.kicad_pcb --net GND --layer B.Cu -o with_zones.kicad_pcb  # side file
 kct zones list board.kicad_pcb
 ```
 
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -2452,7 +2452,11 @@ def _add_zones_parser(subparsers) -> None:
     # zones add
     zones_add = zones_subparsers.add_parser("add", help="Add a copper zone")
     zones_add.add_argument("pcb", help="Path to .kicad_pcb file")
-    zones_add.add_argument("-o", "--output", help="Output file path")
+    zones_add.add_argument(
+        "-o",
+        "--output",
+        help="Output file path (default: overwrite input, consistent with 'zones fill')",
+    )
     zones_add.add_argument("--net", required=True, help="Net name (e.g., GND, +3.3V)")
     zones_add.add_argument("--layer", required=True, help="Copper layer (e.g., B.Cu, F.Cu)")
     zones_add.add_argument("--priority", type=int, default=0, help="Zone fill priority")
@@ -2473,7 +2477,11 @@ def _add_zones_parser(subparsers) -> None:
     # zones batch
     zones_batch = zones_subparsers.add_parser("batch", help="Add multiple zones from spec")
     zones_batch.add_argument("pcb", help="Path to .kicad_pcb file")
-    zones_batch.add_argument("-o", "--output", help="Output file path")
+    zones_batch.add_argument(
+        "-o",
+        "--output",
+        help="Output file path (default: overwrite input, consistent with 'zones fill')",
+    )
     zones_batch.add_argument(
         "--power-nets",
         required=True,
--- a/src/kicad_tools/cli/zones_cmd.py
+++ b/src/kicad_tools/cli/zones_cmd.py
@@ -113,7 +113,7 @@ def main(argv: list[str] | None = None) -> int:
     add_parser.add_argument(
         "-o",
         "--output",
-        help="Output file path (default: <input>_zones.kicad_pcb)",
+        help="Output file path (default: overwrite input, consistent with 'zones fill')",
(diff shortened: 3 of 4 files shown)


rjwalters:
✅ **Approved!**

Thorough review against issue #4166's acceptance criteria — all 8 met, and the core data-loss fix is genuinely load-bearing.

## Verification

**Code (`_run_add` / `_run_batch`)** — both resolve `output_path = Path(args.output) if args.output else pcb_path`, exactly the issue-specified in-place idiom, matching `zones fill` (None → overwrite) and `optimize-placement`.

**Self-overwrite safety** — confirmed no mid-flight read clobber: `ZoneGenerator.from_pcb()` fully loads `PCB.load()` + `parse_file()` into memory (generator.py:293-294) before any `save()`. Partial-failure safety is also sound: `add_zone` runs entirely in memory and returns early on error *before* `save()`; `serialize_sexp()` builds the full string in memory before `write_text()`. The write is non-atomic, but that's a pre-existing property shared by every in-place command (`fill`, `optimize-placement`) — not a regression introduced here.

**The key regression test** — I verified `TestAddChainedAccumulates::test_two_sequential_adds_keep_both_zones` genuinely FAILS under the old `_zones`-side-file default and PASSES under the fix. Reverted the two `output_path` lines locally: 7 of the new tests fail (including the two-sequential-adds accumulation test — the exact data-loss scenario), confirming the tests actually guard the bug rather than trivially passing.

**Help text** — updated and consistent in BOTH parser definitions (`parser.py::_add_zones_parser` add+batch, and `zones_cmd.py` add+batch); `_zones` suffix mention removed; `TestZonesParserHelpConsistency` asserts `"overwrite input" in help` and `"_zones" not in help` for both parsers.

**Docs** — `docs/reference/cli.md` documents the in-place default, the accumulation semantics, and adds the previously-missing `batch`/`fill` subcommand rows.

**Backward-compat** — no test/doc/example relied on the `zones add/batch` `_zones` suffix as a contract. The `board_with_zones.kicad_pcb` references in `zones/__init__.py` / `generator.py` are illustrative docstrings, not the suffix convention.

## Is the default-behavior change justified?

Yes. The old side-file default was a demonstrable data-loss bug (second `add` discards first), the owner explicitly requested the in-place direction in the issue, and explicit `-o <path>` fully preserves side-file behavior for anyone who wants the original untouched — so no capability is lost.

## Local checks
- `pytest tests/test_zones_cmd.py` → **74 passed**
- `ruff check` + `ruff format --check` → clean
- `mypy` on changed source files → clean
- **CI: 16/16 checks pass**, `mergeStateStatus: CLEAN`

## Follow-up (non-blocking)

Filed #4195: the sibling `pcb add-zone` command (`commands/pcb.py:919`) still uses the old `_zones` side-file default and carries the same compounding discard hazard. Correctly out of scope for #4166 (which targeted `zones_cmd.py`), but worth tracking.

PR #4150, opened 2026-07-14
feat(check): add courtyard-overlap sub-check with pair-level waiver file

rjwalters:
## Summary

Adds a pure-Python **courtyard-overlap** sub-check to `kct check`, gated by a
versioned pair-level waiver file (`.courtyard_waivers.json`). This makes the
courtyard gate fully headless and independent of `kicad-cli`'s per-instance
`drc_exclusions` (not honored by kicad-cli 10.0.4 headless), directly
unblocking the chorus consumer (project-shamrock/chorus#18).

Scope is **option (1) only** per the curator guidance; option (2)
(`drc_exclusions`-aware `kct drc` filtering) is explicitly deferred to a
separate follow-up issue.

## Changes

- **`schema/pcb.py`**: parse `fp_poly` into `FootprintGraphic.points` (new
  `points` field), closing the gap where courtyards drawn as polygons were
  invisible to the schema parser. `fp_poly` is now added to the graphics parse
  loop.
- **`validate/violations.py`**: add orthogonal `waived` / `waiver_reason` /
  `waiver_issue` fields to `DRCViolation` (severity stays `error`), plus
  `is_waived` and a `DRCResults.waived_count`. `error_count` / `warning_count`
  / `info_count` now exclude waived findings, so a waived overlap never fails
  the gate. `to_dict()` emits a `status` field (`"waived"` vs severity).
- **`validate/rules/courtyard.py`** (new): `CourtyardOverlapRule` extracts real
  F.CrtYd/B.CrtYd geometry (single `fp_rect`, closed `fp_line` chains via
  endpoint matching, or `fp_poly`), transforms to board space honoring KiCad's
  negated-rotation convention (`_fp_transform`, per #3739), and tests pairwise
  positive-area intersection per board side. Unresolvable courtyards emit an
  `info` "could not resolve outline" finding; stale waivers emit an `info`
  "unused waiver" finding.
- **`validate/rules/courtyard_waivers.py`** (new): versioned (`version == 1`)
  schema loader/validator with order-insensitive `refs` pair matching and
  sidecar discovery (`.courtyard_waivers.json`, probing pcb dir then
  `output/`).
- **`validate/checker.py`**: `check_courtyard_overlap` method, added to
  `CHECK_ALL_METHODS`, plus a `courtyard_waivers` constructor kwarg.
- **`cli/check_cmd.py`**: register `courtyard_overlap` in `CHECK_CATEGORIES`
  and the `check_methods` dict; add `--courtyard-waivers` flag with the same
[...]

Issue #4137, opened 2026-07-14
check/drc: headless per-instance DRC waivers — courtyard-overlap sub-check with waiver file, or drc_exclusions-aware report filtering


rjwalters:
## Motivation (consumer: project-shamrock/chorus#18)

chorus needs `courtyards_overlap` restored to **error** severity with exactly 2 intentional, EE-mandated exceptions (C52↔U10, C53↔U10 tight decoupling, caps ≤2 mm from the VCC pin). The KiCad-native mechanism — per-instance `drc_exclusions` in `.kicad_pro` — is **not honored by kicad-cli 10.0.4 headless** (it appears to require exact internal-IU marker positions not exposed by kicad-cli output). Result: the severity restore is blocked, and the board-wide `warning` demotion keeps masking any *new* courtyard overlap a future edit introduces — a silent class of placement error.

## Current kct behavior (v0.15.1, verified)

- `kct check` sub-checks are clearance / dimensions / edge / silkscreen — **no courtyard-overlap check**, so the courtyard gate currently exists only on the kicad-cli side, where exclusions don't work headlessly.
- `kct drc <report>` parses kicad-cli reports and filters by type/net, but has no exclusion/waiver awareness (`core/severity.py` recognizes an EXCLUSION severity for ERC only; `core/project_file.py` writes an empty `drc_exclusions` list when scaffolding).

## Requested feature (either or both; (1) preferred)

1. **Courtyard-overlap sub-check in `kct check` with a versioned waiver file.** Pure-Python courtyard intersection test (courtyard polygons are in the board file), with waivers expressed at the pair level — e.g. `{rule: courtyards_overlap, refs: [C52, U10], reason: "EE-mandated tight decoupling", issue: "chorus#13"}` — in a JSON/YAML file committed next to the board (like `.net_class_map.json`). Waived findings report as WAIVED (visible, counted, not errors). This makes the gate fully headless and independent of kicad-cli's exclusion handling.

2. **Exclusion-aware kicad-cli report filtering in `kct drc`**: read `.kicad_pro` `drc_exclusions` and apply them to a kicad-cli JSON report with position-tolerant matching (rule type + the two object refs + position within epsilon), replicating GUI semantics well enough for cross-gating.

## Acceptance sketch

[...]

Issue #4176, opened 2026-07-14
net-status: PCB connectivity model diverges from KiCad — reports 75/77 complete on a board where kicad-cli DRC finds 58 unconnected items


rjwalters:
## Bug (PCB-side sibling of #4143)

`kct net-status` and KiCad disagree about what is connected, in the dangerous direction. Same artifact, same moment:

```
$ kct net-status board.kicad_pcb
  Complete:   75 (100% connected)   # all signal nets "complete"
  Incomplete: 2                     # only the two un-poured plane nets

$ kicad-cli pcb drc --severity-error board.kicad_pcb
  Found 58 unconnected items
```

kct's model appears to union copper on endpoint proximity/tolerance, while KiCad requires actual geometric copper contact. A segment whose endpoint lands *near* a pad (or near another segment) satisfies kct but not KiCad — so `net-status` blesses boards that are electrically open.

This matters because `net-status` is the toolchain's own completion oracle: route loops, `--why` diagnostics, and CI checks all key off it. #4143 documented the same class of divergence in the schematic model (there: over-merging via collinear overlap; here: over-connecting via endpoint tolerance). Both sides need the same fix: match KiCad's connectivity semantics, or at minimum document the tolerance and expose a `--strict` mode that uses real geometry.

Repro: rjwalters/softstart PR #28 rev-B artifacts (75/77-complete board with 58 kicad-cli unconnected items), kicad-tools 0.15.0, KiCad 10.

---

### Root cause (confirmed by inspection)

The bug is in `NetStatusAnalyzer._build_connectivity_graph()` and its helpers in
`src/kicad_tools/analysis/net_status.py`:

- **`_points_close()`** (`net_status.py:1091-1099`) is the single connectivity
  primitive for everything that isn't zone/pour copper: it unions two points
  (segment endpoint, via center, pad center) whenever they land within
  `POSITION_TOLERANCE = 0.01` mm (`net_status.py:373`) of each other — a
  Euclidean radius check, not a geometric-contact test.
- **`_build_segment_components()`** (`net_status.py:1025-1069`) chains two
  trace segments into the same copper island purely on `_points_close(seg_a.*,
  seg_b.*)` — i.e. "their endpoints are within 0.01 mm", not "their copper
  polygons (segment width, pad size) actually overlap or touch". A segment
  endpoint that lands 0.009 mm outside a pad's copper edge, or two segment
  endpoints 0.009 mm apart with no shared copper, both satisfy this and get
[...]

PR #4194, opened 2026-07-15
fix(placement): use real courtyard polygons for placement-check overlaps

### LINKED ISSUE #4182: placement check: reports 0 conflicts on a placement where kicad-cli DRC finds 18 courtyard-overlap errors

rjwalters:
## Summary

`kct placement check` reported far fewer courtyard overlaps than KiCad's DRC on the same placement (the reported case: 0 placement conflicts vs. 18 `courtyards_overlap` DRC errors). Root cause: `PlacementAnalyzer._footprint_to_component` approximated each footprint's courtyard as its **pads bounding box expanded by a fixed 0.25 mm margin**, never reading the real `F.CrtYd`/`B.CrtYd` graphics. That axis-aligned bbox is a strict, non-conservative subset of the true courtyard polygon, so `placement check` found only a subset of KiCad's overlaps — the dangerous direction for a pre-routing manufacturability gate.

Meanwhile `kct check`'s `CourtyardOverlapRule` (#4137/PR #4150) already extracts real courtyard polygons and does positive-area intersection. This PR reconciles the two by sharing that geometry.

## Changes

- **New shared module `src/kicad_tools/geometry/courtyard.py`**: the checker-agnostic courtyard-polygon helpers (`_fp_transform`, `_courtyard_side`, `_rect_ring`, `_chain_lines`, `_courtyard_polygon`, `_has_courtyard_geometry`, plus a small `_side_has_geometry`), moved verbatim out of `validate/rules/courtyard.py`.
- **`validate/rules/courtyard.py`**: re-imports the helpers from the shared module (re-exported via `__all__` for backward compatibility). No behavior change — `CourtyardOverlapRule` is untouched and its tests pass unmodified.
- **`PlacementAnalyzer`**: `_footprint_to_component` now builds the component's real courtyard polygon from `F.CrtYd`/`B.CrtYd` artwork (per the component's placed side, honoring position + rotation), stored on `ComponentInfo.courtyard_polygon`. `_check_courtyard_overlap` uses a positive-area polygon intersection (`_check_courtyard_overlap_polygon`) when both components resolve a polygon, matching `CourtyardOverlapRule` and KiCad. Footprints with no resolvable courtyard geometry fall back to the pads-bbox+margin approximation.
- `ComponentInfo.courtyard` (the bbox `Rectangle`) becomes the polygon's bounds when a polygon exists, so `_check_off_board` / `_check_edge_clearance` (#4156/PR #4187) get more precise for free without breaking.
- **`docs/placement-scoring.md`**: corrected the metric description (was "courtyard-expanded polygon" — aspirational until now).

| Real `F.CrtYd`/`B.CrtYd` polygon when resolvable, honoring position + rotation | ✅ | `_resolve_courtyard_polygon` via shared helpers; rotation tests |
| Falls back to pads-bbox+margin when no courtyard geometry | ✅ | `test_pads_only_footprint_uses_bbox_fallback`, existing `test_placement.py` fixtures |
| Real overlapping courtyards the bbox misses now flagged | ✅ | `test_courtyard_larger_than_pads_overlaps_but_pads_do_not`, `test_offset_noncentered_courtyard_overlap` |
| Non-overlapping → zero conflicts (no false positive) | ✅ | `test_moved_apart_no_conflict`, `test_touching_zero_area_no_conflict` |
| Rotated-footprint case | ✅ | `test_rotation_brings_courtyard_into_overlap`, `test_rotated_footprint_overlap_detected` |
| Existing `test_placement.py` courtyard tests pass unmodified | ✅ | 4 pads-only tests green |
| `placement check` and `kct check` agree on shared fixture | ✅ | `test_placement_and_drc_agree_on_shared_fixture` |
| `docs/placement-scoring.md` corrected | ✅ | metric table + section 1 updated |
| No new hard dependency (shapely already core) | ✅ | uses `has_shapely`/`require_shapely` |

## Test Plan

- `uv run pytest tests/test_placement.py tests/test_placement_drc.py tests/test_placement_iterative_fix.py tests/test_placement_nudge.py tests/test_validate_courtyard_overlap.py tests/test_collision_detection.py tests/test_placement_courtyard_geometry.py -q` → 178 passed.
- Broader placement-adjacent suites (route off-board gate, optim, suggestions, place-route optimizer, `tests/placement/`) → 338 passed, 2 skipped.
- `uv run --extra dev python scripts/ci/check_mypy_baseline.py` → 0 new errors.
- New `tests/test_placement_courtyard_geometry.py` cross-checks each fixture through both `PlacementAnalyzer.find_conflicts` and `CourtyardOverlapRule.check`, asserting identical overlapping-pair verdicts.

**Board-fixture baseline shifts:** none observed. All new fixtures are synthetic in-memory `PCB(SExp(...))` objects (chorus/softstart are local-only, not in CI). Existing pads-only fixtures use the preserved bbox fallback and are unchanged.

Closes #4182

### DIFF
--- a/docs/placement-scoring.md
+++ b/docs/placement-scoring.md
@@ -17,7 +17,7 @@ surface measures, why they differ, and which one to trust for which purpose.
 
 | Surface | Command | Overlap metric | DRC metric | Layout scored |
 |---------|---------|----------------|------------|---------------|
-| Diagnostics | `kct placement check` | courtyard-expanded polygon (margin, default 0.25 mm) — boolean/severity per pair | real KiCad DRC violations | **actual on-disk placement** |
+| Diagnostics | `kct placement check` | real `F.CrtYd`/`B.CrtYd` polygon (positive-area intersection); pads-bbox + margin (default 0.25 mm) fallback when no courtyard artwork — boolean/severity per pair | real KiCad DRC violations | **actual on-disk placement** |
 | Optimizer objective (dry-run) | `kct optimize-placement --dry-run` | AABB overlap *area* (mm²), no margin | bbox clearance *count* | **actual on-disk placement** (fixed in #3940) |
 | Optimizer objective (search) | `kct optimize-placement` initial/seed eval | AABB overlap *area* (mm²), no margin | bbox clearance *count* | seed / candidate populations during the search |
 | Interactive energy | `kct placement refine` session | none (spacing energy proxy) | none explicit | live positions in the session |
@@ -29,11 +29,18 @@ There are **two independent axes** of difference:
 ### 1. Different geometry (courtyard vs. raw bounding box)
 
 - `kct placement check`
-  ([`PlacementAnalyzer`](../src/kicad_tools/placement/analyzer.py)) expands
-  each footprint's pad bounding box by `courtyard_margin` before testing for
-  overlap, and reports each conflicting pair as a human-readable
-  warning/error. A pair that merely *touches* (0 mm gap) is flagged because
-  the courtyard margins overlap.
+  ([`PlacementAnalyzer`](../src/kicad_tools/placement/analyzer.py)) reads each
+  footprint's **real** `F.CrtYd`/`B.CrtYd` courtyard polygon (via the shared
+  [`geometry.courtyard`](../src/kicad_tools/geometry/courtyard.py) helpers, the
+  same geometry `kct check`'s courtyard-overlap DRC rule and KiCad use) and
+  flags a pair when their courtyard polygons intersect with strictly positive
+  area, so `kct placement check` and `kct check` agree on courtyard overlaps
+  (issue #4182). For footprints with **no** resolvable courtyard artwork it
+  falls back to the legacy approximation: the pad bounding box expanded by
+  `courtyard_margin`, tested as an axis-aligned rectangle overlap. Under that
+  fallback a pair that merely *touches* (0 mm gap) is flagged because the
+  courtyard margins overlap; under the real-polygon path exactly-touching
+  courtyards (zero-area intersection) do **not** conflict, matching KiCad.
 - The optimizer objective
   ([`evaluate_placement`](../src/kicad_tools/placement/cost.py)) measures the
   raw axis-aligned bounding-box overlap **area** in mm², with **no** margin.
--- a/src/kicad_tools/geometry/__init__.py
+++ b/src/kicad_tools/geometry/__init__.py
@@ -1,0 +1,1 @@
+"""Checker-agnostic geometry helpers shared across kicad-tools subsystems."""
--- a/src/kicad_tools/geometry/courtyard.py
+++ b/src/kicad_tools/geometry/courtyard.py
@@ -1,0 +1,180 @@
+"""Shared, checker-agnostic courtyard-polygon geometry (Issue #4182).
+
+This module is the single source of truth for "what is this footprint's
+courtyard" — the real ``F.CrtYd`` / ``B.CrtYd`` polygon geometry, honoring
+the footprint's position and rotation.  It is consumed by:
+
+* :mod:`kicad_tools.validate.rules.courtyard` — the ``kct check``
+  courtyard-overlap DRC rule (extracted from there verbatim in #4182; behavior
+  unchanged).
+* :mod:`kicad_tools.placement.analyzer` — the ``kct placement check``
+  ``PlacementAnalyzer``, which previously approximated each courtyard as a
+  pads-bounding-box expanded by a fixed margin and therefore found only a
+  strict subset of KiCad's courtyard overlaps.
+
+Having both checkers resolve courtyards through the same helpers is exactly
+what the reconciliation issue (#4182) asks for: ``kct placement check`` and
+``kct check`` should agree on courtyard overlaps.
+
+Courtyard-outline extraction supports three representations:
+
+* ``fp_rect`` — a single axis-aligned rectangle (fast path).
+* ``fp_poly`` — a closed polygon whose vertices are recorded in
+  :attr:`FootprintGraphic.points`.
+* ``fp_line`` — a closed loop of line segments chained by endpoint matching.
+
+A courtyard that cannot be resolved from these (no CrtYd geometry, a
+non-closing ``fp_line`` chain, or fewer than three vertices) resolves to
+``None`` so callers can fall back or surface the gap explicitly rather than
+silently skipping the footprint.
+"""
+
+from __future__ import annotations
+
+import math
+from typing import TYPE_CHECKING, Any
+
+from kicad_tools.core.types import Layer
+
+if TYPE_CHECKING:
+    from kicad_tools.schema.pcb import Footprint, FootprintGraphic
+
+# Endpoint-matching tolerance (mm) when chaining ``fp_line`` segments into a
+# closed courtyard ring.  Courtyard vertices are authored on a coarse grid;
+# 1e-3 mm is comfortably below any real feature yet absorbs float noise.
+_CHAIN_EPSILON_MM = 1e-3
+
+
+def _fp_transform(footprint: Footprint):
+    """Return a ``(x, y) -> (X, Y)`` local->board transform for a footprint.
+
+    Mirrors ``validate.rules.silkscreen._fp_transform``: KiCad negates the
+    footprint orientation angle relative to CCW math (verified in #3739), so
+    the rotation applied here is ``radians(-rotation)``.
+    """
+    fp_x, fp_y = footprint.position
+    fp_rotation = math.radians(-footprint.rotation)
+    cos_rot = math.cos(fp_rotation)
+    sin_rot = math.sin(fp_rotation)
+
+    def transform(point: tuple[float, float]) -> tuple[float, float]:
+        lx, ly = point
+        return (
+            fp_x + (lx * cos_rot - ly * sin_rot),
+            fp_y + (lx * sin_rot + ly * cos_rot),
+        )
+
+    return transform
+
+
+def _courtyard_side(layer: str) -> str | None:
+    """Return ``"F"`` / ``"B"`` for a courtyard layer, else ``None``."""
+    if layer == Layer.F_CRTYD.value:
+        return "F"
+    if layer == Layer.B_CRTYD.value:
+        return "B"
+    return None
+
+
+def _rect_ring(graphic: FootprintGraphic) -> list[tuple[float, float]]:
+    """Return the 5-point closed ring for an ``fp_rect`` graphic (local space)."""
+    sx, sy = graphic.start
+    ex, ey = graphic.end
+    return [(sx, sy), (ex, sy), (ex, ey), (sx, ey), (sx, sy)]
+
+
+def _chain_lines(
+    segments: list[tuple[tuple[float, float], tuple[float, float]]],
+) -> list[tuple[float, float]] | None:
+    """Chain line segments into a single closed ring by endpoint matching.
+
+    Returns the ordered ring of vertices (first == last) when the segments
+    form exactly one closed loop, else ``None`` (open chain, branching, or
+    disconnected components -- not a resolvable simple courtyard).
+    """
+    if not segments:
+        return None
+
+    def close(a: tuple[float, float], b: tuple[float, float]) -> bool:
+        return math.hypot(a[0] - b[0], a[1] - b[1]) <= _CHAIN_EPSILON_MM
+
+    remaining = list(segments)
+    start, current = remaining.pop(0)
+    ring: list[tuple[float, float]] = [start, current]
+
+    while remaining:
+        for idx, (a, b) in enumerate(remaining):
+            if close(a, current):
+                current = b
+                ring.append(current)
+                remaining.pop(idx)
+                break
+            if close(b, current):
+                current = a
+                ring.append(current)
+                remaining.pop(idx)
+                break
+        else:
+            # No segment continues the chain -> not a single closed loop.
+            return None
+
+    # A closed loop returns to its start.
+    if not close(ring[0], ring[-1]):
+        return None
+    return ring
+
+
+def _courtyard_polygon(footprint: Footprint, side: str, Polygon: Any):
+    """Build a shapely polygon for a footprint's courtyard on ``side``.
+
+    ``side`` is ``"F"`` or ``"B"``.  Returns ``None`` when no courtyard
+    geometry on that side can be resolved into a valid closed polygon.
+    """
+    target_layer = Layer.F_CRTYD.value if side == "F" else Layer.B_CRTYD.value
+    transform = _fp_transform(footprint)
+
+    rects: list[FootprintGraphic] = []
+    polys: list[FootprintGraphic] = []
+    line_segments: list[tuple[tuple[float, float], tuple[float, float]]] = []
+    for graphic in footprint.graphics:
+        if graphic.layer != target_layer:
+            continue
+        if graphic.graphic_type == "rect":
+            rects.append(graphic)
+        elif graphic.graphic_type == "poly":
+            polys.append(graphic)
+        elif graphic.graphic_type == "line":
+            line_segments.append((graphic.start, graphic.end))
+        # circle / arc courtyards are not modeled (rare for courtyards).
+
+    ring: list[tuple[float, float]] | None = None
+    if rects:
+        # A single rect fully describes the courtyard; if multiple exist we
+        # take the first (the common single-rect case).
+        ring = _rect_ring(rects[0])
+    elif polys and len(polys[0].points) >= 3:
+        ring = list(polys[0].points)
+    elif line_segments:
+        ring = _chain_lines(line_segments)
+
+    if ring is None or len(ring) < 3:
+        return None
+
+    board_ring = [transform(p) for p in ring]
+    polygon = Polygon(board_ring)
+    if not polygon.is_valid:
+        polygon = polygon.buffer(0)
+    if polygon.is_empty or polygon.area <= 0:
+        return None
+    return polygon
+
+
+def _has_courtyard_geometry(footprint: Footprint) -> bool:
+    """True if the footprint has any F/B CrtYd graphic at all."""
+    return any(_courtyard_side(graphic.layer) is not None for graphic in footprint.graphics)
+
+
+def _side_has_geometry(footprint: Footprint, side: str) -> bool:
+    """True if the footprint has any CrtYd graphic on the given ``side``."""
+    target_layer = Layer.F_CRTYD.value if side == "F" else Layer.B_CRTYD.value
+    return any(graphic.layer == target_layer for graphic in footprint.graphics)
--- a/src/kicad_tools/placement/analyzer.py
+++ b/src/kicad_tools/placement/analyzer.py
@@ -52,17 +52,24 @@ class PlacementAnalyzer:
         conflicts = analyzer.find_conflicts("board.kicad_pcb", rules=rules)
 
     .. note::
-        This analyzer powers ``kct placement check``. Its overlap metric
-        expands each footprint's pad bounding box by ``courtyard_margin``
-        (default 0.25 mm) and yields per-violation, human-readable
-        diagnostics. It is a **different metric** from the optimizer
-        objective in :func:`kicad_tools.placement.cost.evaluate_placement`,
-        which uses raw axis-aligned bounding-box overlap area with no
-        courtyard margin. The courtyard expansion means this check can flag
-        a "touching" pair that the optimizer objective scores as zero
-        overlap. Keeping the two separate is intentional (actionable
-        diagnostics vs. a smooth optimizer search space); see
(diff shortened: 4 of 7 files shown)


rjwalters:
## Judge Review: APPROVED ✅

Reviewed against #4182's curated acceptance criteria. All CI checks pass (16/16), mergeStateStatus CLEAN, mergeable MERGEABLE.

### Verified

**Verbatim extraction (no behavior change to `kct check`)** — `_fp_transform`, `_courtyard_side`, `_rect_ring`, `_chain_lines`, `_courtyard_polygon`, `_has_courtyard_geometry` moved to `src/kicad_tools/geometry/courtyard.py` with byte-identical bodies (including `_CHAIN_EPSILON_MM = 1e-3` and the `radians(-rotation)` #3739 negation convention). `CourtyardOverlapRule` re-imports them and re-exports via `__all__`. The only addition is `_side_has_geometry`, which factors out the inline per-side geometry check the rule already had (no semantic change). `test_validate_courtyard_overlap.py` passes unmodified.

**Real polygon in placement + bbox fallback** — `_footprint_to_component` now calls `_resolve_courtyard_polygon(fp)`; `_check_courtyard_overlap` dispatches to positive-area polygon intersection only when *both* components resolve a polygon, else falls back to the pads-bbox+margin AABB test. Mixed (one polygon, one pads-only) and unresolvable-courtyard (non-closing line chain) cases correctly fall back — both covered by tests, verified no crash.

**Rotation convention matches KiCad/DRC** — placement uses the same shared `_fp_transform`, so the #4137 angle-negation is honored identically. Confirmed by the rotation tests and the agreement test.

**Cross-side (F vs B) correctness** — I probed a front/back pair with XY-overlapping courtyards directly: placement reports 0, DRC reports 0 — they agree. Pair generation is gated by `_same_layer`, so cross-side SMD pairs are never courtyard-checked, matching `CourtyardOverlapRule`'s per-side (F/F, B/B) comparison. No regression.

**Polygon-bounds-as-Rectangle for off-board/edge-clearance (#4187)** — `_check_off_board`/`_check_edge_clearance` are documented AABB tests; feeding them the polygon's `.bounds` (tightest AABB containing the true courtyard) is strictly more accurate than the old pads-bbox+0.25mm and can never introduce an off-board false-negative (bounds fully contain the courtyard). More precise, not broken.

**Agreement test genuinely proves it** — `test_placement_and_drc_agree_on_shared_fixture` runs the same 3-footprint fixture through `PlacementAnalyzer.find_conflicts` and `CourtyardOverlapRule.check`, comparing overlapping-pair *sets* (frozensets), asserting `placement_pairs == drc_pairs == {frozenset({"U1","U2"})}`. This is the core value of the issue and it's tested well.

**Zero-area/touching** — `_check_courtyard_overlap_polygon` returns `None` on `inter.area <= 0`; `test_touching_zero_area_no_conflict` confirms exactly-touching courtyards are NOT flagged (matching KiCad), both checkers 0.

### Checks run locally (worktree at PR head f978c03d)
- `tests/test_placement.py` + drc + iterative_fix + nudge + validate_courtyard_overlap + collision_detection + placement_courtyard_geometry → **178 passed**
- ruff check + ruff format --check on all changed files → clean
- mypy baseline → **0 new errors** (39 fewer, in fact)
- Docs (`placement-scoring.md`) corrected accurately, including the honest note that touching-flags under fallback but not under the real-polygon path.

Scope respected: courtyards only, #4176 untouched. Nice work — clean refactor with a real agreement guard.

PR #4198, opened 2026-07-15
test(router): widen board-02 segment-count band to cover macOS-arm64 476

### LINKED ISSUE #4196: Board 02 baseline: test_routing_output_deterministic_across_seeds fails on macOS-arm64 (476 segments vs band [390,398])
## Original Issue

`tests/router/test_board02_manufacturable_baseline.py::test_routing_output_deterministic_across_seeds` fails on macOS-arm64 (Python 3.14, uv.lock-synced venv, C++ backend) with:

```
AssertionError: Board 02 segment count 476 outside the documented platform band [390, 398]
assert 476 <= 398
```

Routes (22), vias (24), and total length (327.93mm) all match the exact pins — **only the collinear-segment count drifted** (476 vs the documented 390–398 band). This is the same "identical geometry, different collinear segmentation" mode the test's own PLATFORM NOTE (#3545) documents, but the drift now exceeds the tolerance band.

### Independently verified as pre-existing (not caused by any open PR)

Discovered while reviewing PR #4192. Judge built the C++ backend in a detached worktree at clean `origin/main` (`9def133d`) and ran the test in isolation — it fails identically there (`476 <= 398`), so this is **platform baseline drift on `main`**, not introduced by any feature branch. It did not surface in CI because CI runs on Linux-x86_64 (band value 399) where the collinear merge lands differently.

### Suggested action (original)

Either:
- Re-measure the segment band on current macOS-arm64 + Linux-x86_64 and widen `EXPECTED_SEGMENTS_RANGE` with an updated PLATFORM NOTE, or
- Investigate whether a recent `merge_collinear` / `path_is_clear` change (since the 2026-07-09 sheet-centering re-baseline) regressed collinear merging on this platform (476 is a large jump from ~393–397, larger than the prior ±3 platform skew — worth confirming it's benign resegmentation and not a real merge regression before widening the band).

Routes/vias/length are pinned exactly and still pass, so this is scoped to the segment-count band only.

Discovered during evaluation of PR #4192.

---

### Root-cause determination: STALE BASELINE (deterministic, platform-specific), not a router regression

Independently re-reproduced on the current tip of `origin/main` (`a8ed3b2b`, 2026-07-14), which has **zero router-touching commits** since the judge's tested commit `9def133d` (`git log --oneline 9def133d..HEAD -- src/kicad_tools/router/` is empty) — so this is not a fresh regression, it is the same drift the judge found, still present.

Direct reproduction (two independent `kct route` invocations, seed 42, `--backend cpp`, jlcpcb-tier1, macOS-arm64, C++ backend built via `uv run kct build-native`):

```
Run 1: Routes 22, Segments 476, Vias 24, Total length 327.93mm, Nets routed 8/8
Run 2: Routes 22, Segments 476, Vias 24, Total length 327.93mm, Nets routed 8/8
```

Both runs produced **bit-identical** output. This confirms the judge's characterization: **476 is deterministic on this platform**, not a random draw. It is not the "genuine cross-run/cross-seed nondeterminism" failure mode the test's own docstring warns about (see `test_routing_output_deterministic_across_seeds`'s cross-seed equality assertion, which is separate from and passes independently of the segment-count band check below it).

**Verdict: this is case (a) from the task brief — deterministic-but-platform-drifted (stale band), not case (b) genuine nondeterminism. No escalation to architect needed; no router bug.**

### What the test actually guarantees (read from `tests/router/test_board02_manufacturable_baseline.py:460-651`)

`test_routing_output_deterministic_across_seeds` bundles three distinct guarantees in one test body:

1. **Cross-seed determinism** (lines 540-552): seeds 42/43/44 must produce byte-identical `(routes, segments, vias, length)` tuples. This is an exact-equality assertion across seeds and is **unaffected** by this issue — it still passes (all three seeds converge to the same 476-segment result on this platform).
2. **Exact regression pin on routes/vias/length** (lines 620-641): `EXPECTED_ROUTES=22`, `EXPECTED_VIAS=24`, `EXPECTED_LENGTH=327.93` — asserted via exact tuple equality. **Still passes.**
3. **Segment-count band** (lines 628, 642-650): `EXPECTED_SEGMENTS_RANGE = (390, 398)`, with an explicit comment (line 623-627) claiming "393 on macOS-arm64" with "the same +/-3 envelope" applied from a pre-centering +2 platform skew observation. **This is the check that fails** — the documented macOS-arm64 baseline of 393 has drifted to 476 (a +83 delta), which is far outside the "harmless collinear resplit" magnitude (single-digit deltas) seen in every prior re-baseline entry in the docstring history (155→206→299→193→387→397→393, all ≤14-segment steps).

The module docstring (lines 1-122) is a meticulous, self-documenting history of *every* prior segment-count re-baseline (10+ entries since #3433), each time re-affirming: routes/vias/length/reach are the regression-sensitive metrics (pinned exactly), while segment count is explicitly a platform-and-version-sensitive collinear-resegmentation artifact (given a tolerance band). This +83 delta is consistent with that established pattern in kind, but is a much larger magnitude jump than any prior entry — worth noting explicitly in the re-baseline, not just silently widening the band.

### Why 476 is plausible as "still benign" despite the unusually large jump

- Routes (22), vias (24), length (327.93mm) and reach (8/8) are **byte-identical** to the currently-documented "2026-07-09 sheet-centering" pin — i.e., the exact geometric solution the router converges to has **not changed** since the last verified-good re-baseline.
- Only the segment count changed, and only in the direction of *more* segments (finer collinear splitting), which is the same failure mode as every prior platform-note entry (never a *coarsening* / fewer-segments direction, which would suggest a merge-logic bug swallowing real geometry).
- Determinism (both cross-seed and cross-run, per the reproduction above) is intact.

This combination (identical routes/vias/length/reach + exact cross-run reproducibility + directionally-consistent-with-history delta) is strong evidence for "stale platform band," not "collinear-merge regression." However, the curator has **not** bisected `merge_collinear`/`path_is_clear` to rule out a genuine (but currently harmless) behavior change in that pass — the Builder should do a lightweight sanity check (see Test Plan) before finalizing the new band, per the original issue's own caution ("investigate... before widening").

### CI status: GREEN, not exercising this failure mode

- The CI `test` job (`.github/workflows/ci.yml:72-142`) runs on `runs-on: ubuntu-latest` (Linux-x86_64) inside the `kicad/kicad:10.0` container, invoking `uv run pytest -n auto -o addopts= --benchmark-disable --timeout=60 -m "not slow"`.
- `test_routing_output_deterministic_across_seeds` carries no `slow`, `skip`, or `skipif` marker (grep confirms `pytest.skip` appears only in the `unrouted_pcb_path` fixture, gated on the committed PCB file existing) — it **is** collected and run in this CI job.
- The documented Linux-x86_64 band value is 399 (inside `[390, 398]`... actually just outside by 1, but the docstring's "+/-3 envelope" language suggests the band was hand-picked to cover both platform values with margin) — either way, CI on Linux is currently **green**: `gh run list` shows the CI run for the judge's exact tested commit `9def133d` completed with `conclusion: "success"`.
- Conclusion: **this is a macOS-arm64-only local failure.** CI does not catch it because the Linux collinear-merge landing differs from the macOS one, and the current band happens to still cover the Linux value. This is exactly the platform-divergence mechanism the test's own PLATFORM NOTE (#3545) already documents — it has just grown past the macOS side of the tolerance window.

### Recommended fix (stale-band re-baseline, NOT an architect escalation)

1. Re-measure the macOS-arm64 segment count (this issue's evidence: 476, reproduced twice, bit-identical) and the current Linux-x86_64 CI value (should be re-confirmed from a CI log rather than assumed — the docstring's 399 figure predates this drift and may itself be stale).
2. Update `EXPECTED_SEGMENTS_RANGE` in `test_routing_output_deterministic_across_seeds` to a band that covers both current platform values with the same style of small margin used historically, OR loosen the check's *intent* rather than just widening the number blindly:
   - Preferred: keep the exact `EXPECTED_ROUTES`/`EXPECTED_VIAS`/`EXPECTED_LENGTH` pins (these are the manufacturability-sensitive metrics and must stay tight), keep the cross-seed exact-equality check (this is the real determinism guarantee), but replace the narrow historical-magnitude band on segment count with either (a) a wider band justified by a fresh two-platform measurement, or (b) a looser sanity bound (e.g., a generous multiple of routes, since segment count is a router-internal collinear-splitting artifact with no manufacturability meaning by itself) plus a comment explicitly stating segment count is platform/KiCad-version-sensitive and NOT itself a manufacturability or correctness signal.
3. Add a fresh dated docstring entry (matching the existing "Re-baselined YYYY-MM-DD for Issue #NNNN" convention at lines 53-70, 554-619) explaining the new macOS-arm64/Linux-x86_64 values and why the jump is large but benign (identical routes/vias/length/reach, deterministic, directionally consistent with prior "finer resplitting" drift).
4. Before finalizing, do a quick sanity pass confirming `merge_collinear`/`path_is_clear` (in `src/kicad_tools/router/optimizer/algorithms.py`) haven't regressed in a way that produces genuinely worse (not just differently-segmented) output — e.g. confirm no new DRC violations and confirm the `test_all_signal_nets_routed` / `test_drc_clean_at_jlcpcb_tier1` tests in the same file still pass (they were not reported as failing by the judge, and this curator pass did not re-run them, but the Builder should include them in the verification sweep since they share the `route_stdout` fixture in the same module).

### Acceptance Criteria

- [ ] `test_routing_output_deterministic_across_seeds` passes on macOS-arm64 (this platform) with the current router build (no source changes to the router itself required, or if a source change is made, it is justified separately and not conflated with this baseline fix).
- [ ] `test_routing_output_deterministic_across_seeds` continues to pass on CI (Linux-x86_64, `ubuntu-latest` + `kicad/kicad:10.0` container).
- [ ] The cross-seed exact-equality assertion (lines 540-552) remains unchanged and still exact — this issue must NOT loosen the real determinism guarantee, only the platform-sensitive segment-count band.
- [ ] `EXPECTED_ROUTES`, `EXPECTED_VIAS`, `EXPECTED_LENGTH` exact pins remain unchanged (22, 24, 327.93) unless a genuine geometry change is independently justified (not expected here).
- [ ] `EXPECTED_SEGMENTS_RANGE` is updated with a dated docstring entry explaining the new macOS-arm64 (and re-confirmed Linux-x86_64) values, following the established "Re-baselined YYYY-MM-DD for Issue #4196" convention already used throughout this file's docstring.
- [ ] `test_all_signal_nets_routed` and `test_drc_clean_at_jlcpcb_tier1` (same module, same `route_stdout` fixture) are re-run and confirmed still passing as part of this fix, to rule out a coincident manufacturability regression riding along with the segment-count drift.
- [ ] If, contrary to this curation's assessment, the Builder finds cross-run or cross-seed nondeterminism (i.e. repeated identical invocations produce *different* segment counts, or seeds 42/43/44 stop matching each other), STOP and escalate to `loom:architect` instead of widening the band — that would indicate a genuine router determinism bug, not a stale baseline.

### Affected Files

- `tests/router/test_board02_manufacturable_baseline.py` — update `EXPECTED_SEGMENTS_RANGE` (currently `(390, 398)` at line 628) and add a dated docstring re-baseline entry (module docstring lines 1-122 and/or the inline comment block at lines 554-627) documenting the new macOS-arm64/Linux-x86_64 measured values and rationale.

### Test Plan

- [ ] Manual verification: `uv run kct build-native --check` confirms C++ backend available, then `uv run pytest tests/router/test_board02_manufacturable_baseline.py -v` passes all tests (not just the one being fixed) on macOS-arm64.
- [ ] Automated tests: the fix IS the test change itself (`test_routing_output_deterministic_across_seeds`); also re-run `test_all_signal_nets_routed` and `test_drc_clean_at_jlcpcb_tier1` in the same file to confirm no coincident regression.
- [ ] CI: confirm the updated band also covers the current Linux-x86_64 CI value — either by re-reading a recent CI log's segment count for this test, or by widening generously enough that a reasonable Linux value is covered, with a follow-up note if CI unexpectedly fails after the change.
- [ ] Edge cases: repeat the macOS-arm64 `kct route` invocation (seed 42, `--backend cpp`, jlcpcb-tier1) 2-3 times independently to reconfirm bit-identical segment counts (this curation already did this twice: 476/476) before locking in the new band value, to guard against the genuine-nondeterminism escalation path.

### Related

- #3545 — origin of the "PLATFORM NOTE" mechanism this issue extends (macOS-arm64 vs Linux-x86_64 collinear-merge float-rounding divergence).
- #3799 (closed) — a distinct, previously-resolved genuine A* nondeterminism issue; not a duplicate of this one (that issue was about non-reproducibility despite `--seed`; this issue's reproduction confirms determinism is intact here).
- PR #4192 — the PR whose Judge review discovered this pre-existing failure on clean `origin/main`.

rjwalters:
## Summary

`tests/router/test_board02_manufacturable_baseline.py::test_routing_output_deterministic_across_seeds` failed on macOS-arm64 (C++ backend) because the segment-count band `EXPECTED_SEGMENTS_RANGE = (390, 398)` no longer covered this platform's deterministic value. The router emits **476 segments** on macOS-arm64 while routes (22), vias (24), length (327.93mm) and reach (8/8) are byte-identical to the 2026-07-09 sheet-centering pin — only the collinear-segment resplit count moved, in the benign "more segments" direction the existing #3545 PLATFORM NOTE documents.

Per the curator's STALE-BASELINE verdict, this is a platform-specific deterministic drift, not a router bug. Verified deterministic locally: two independent full-test runs, plus the cross-seed 42/43/44 exact-equality assertion, all produced bit-identical 476-segment output. This is NOT genuine nondeterminism (no escalation warranted).

CI runs on Linux-x86_64 where the band still holds and the test is green, so the band is **widened** (not moved) to cover both platforms.

## Changes

- Widen `EXPECTED_SEGMENTS_RANGE` from `(390, 398)` to `(390, 490)` — covers the Linux-x86_64 ~390-399 value (keeps CI green) AND the macOS-arm64 476 value, with a small margin above 476.
- **Unchanged:** the cross-seed exact-equality determinism assertion (the real determinism guarantee).
- **Unchanged:** the exact `EXPECTED_ROUTES=22` / `EXPECTED_VIAS=24` / `EXPECTED_LENGTH=327.93` manufacturability pins (platform-invariant, the actual regression guard).
- Add a dated `#4196` re-baseline note to the module docstring and the inline comment/assertion message explaining that segment count is platform/KiCad-version-sensitive collinear resegmentation and NOT itself a manufacturability signal, referencing the #3545 PLATFORM NOTE mechanism.

| Target test passes on macOS-arm64 | ✅ | `pytest ...::test_routing_output_deterministic_across_seeds` passes (476 within new band) |
| Test continues to pass on CI (Linux-x86_64) | ✅ | Band still covers Linux ~390-399; lower bound 390 unchanged, only upper bound widened |
| Cross-seed exact-equality assertion unchanged | ✅ | No edit to lines asserting seed 42/43/44 tuple equality; still exact |
| Exact routes/vias/length pins unchanged | ✅ | `EXPECTED_ROUTES=22`, `EXPECTED_VIAS=24`, `EXPECTED_LENGTH=327.93` untouched |
| `EXPECTED_SEGMENTS_RANGE` updated with dated docstring entry | ✅ | New `(390, 490)` band + `Re-baselined 2026-07-14 for Issue #4196` note |
| `test_all_signal_nets_routed` + `test_drc_clean_at_jlcpcb_tier1` still pass | ✅ | Full-file run: 3 passed |
| Escalate if nondeterminism found | ✅ (N/A) | Two runs + cross-seed all bit-identical 476 — deterministic, no escalation |

## Test Plan

- `uv run kct build-native --check` → C++ backend available (version 1.0.0), built via `kct build-native`.
- `uv run pytest tests/router/test_board02_manufacturable_baseline.py -q` → **3 passed**.
- Target test run twice independently → segment count 476 both times (bit-identical), confirming determinism.
- `ruff check` + `ruff format --check` → clean.
- `scripts/ci/check_mypy_baseline.py` → 0 new type errors (1475 within 1514 baseline).

Closes #4196

### DIFF
--- a/tests/router/test_board02_manufacturable_baseline.py
+++ b/tests/router/test_board02_manufacturable_baseline.py
@@ -13,10 +13,17 @@ Baseline measurement at HEAD (worst-of-3 across seeds 42/43/44 with
 - **DRC: 0 errors, 0 warnings** at ``jlcpcb-tier1`` profile
 - **Deterministic output**: 22 routes / 24 vias / 327.93mm total
   length identical across seeds 42/43/44 -- this small 2-layer board
-  has fully converged.  Segment count is 393 on macOS-arm64
-  (platform-dependent collinear resplit at identical geometry; see
-  the #3545 PLATFORM NOTE in
-  ``test_routing_output_deterministic_across_seeds``).  (397/328.16mm
+  has fully converged.  Segment count is 476 on macOS-arm64 as of the
+  2026-07-14 #4196 re-baseline (up from 393; routes/vias/length/reach
+  are byte-identical, only the collinear resplit count moved -- see the
+  #4196 note in ``test_routing_output_deterministic_across_seeds``).
+  Segment count is platform- and KiCad-version-sensitive collinear
+  resplitting at identical geometry and is NOT itself a
+  manufacturability signal; it is guarded by a loose band while the
+  exact routes/vias/length pins and the cross-seed equality assertion
+  are the real regression + determinism guards (see the #3545 PLATFORM
+  NOTE in ``test_routing_output_deterministic_across_seeds``).
+  (397/328.16mm
   before the 2026-07-09 sheet-centering translation of the unrouted
   artifact -- kct pcb center-on-sheet; 387 segments
   before the #3545 static-halo rip-up survival; 193/325.08mm before
@@ -620,12 +627,28 @@ def test_routing_output_deterministic_across_seeds(unrouted_pcb_path: Path) -> N
     EXPECTED_ROUTES = 22
     EXPECTED_VIAS = 24
     EXPECTED_LENGTH = 327.93
-    # Measured: 393 on macOS-arm64.  The pre-centering baseline showed
-    # a +2 platform skew on CI Linux-x86_64 (397 vs 399; see PLATFORM
-    # NOTE above), so the band keeps the same +/-3 envelope around the
-    # macOS value and its assumed Linux sibling (395).  Deliberately
-    # tight so a real collinear-handling regression still trips it.
-    EXPECTED_SEGMENTS_RANGE = (390, 398)
+    # Re-baselined 2026-07-14 for Issue #4196: the macOS-arm64 segment
+    # count drifted from 393 to 476 (a +83 delta) while routes (22),
+    # vias (24), total length (327.93mm) and reach (8/8) stayed
+    # BYTE-IDENTICAL to the 2026-07-09 sheet-centering pin -- i.e. the
+    # geometric solution the router converges to is unchanged; only the
+    # collinear-segment SPLITTING count moved, and only in the
+    # finer-splitting direction (more segments), the same benign mode
+    # every prior PLATFORM NOTE entry documents.  Verified deterministic:
+    # two independent local runs (and cross-seed 42/43/44) all produced
+    # bit-identical 476-segment output, so this is a stale platform band,
+    # NOT genuine nondeterminism.  CI Linux-x86_64 still lands in the
+    # historical ~390-399 window (this test is GREEN in CI on the same
+    # commit), so the band is WIDENED -- not moved -- to comfortably cover
+    # BOTH platforms' deterministic values: the Linux ~390-399 side and
+    # the macOS-arm64 476 side, with a small margin above 476.  The
+    # segment count is a router-internal collinear-splitting artifact with
+    # no manufacturability meaning on its own; the exact routes/vias/length
+    # pins above and the cross-seed equality assertion earlier are the real
+    # regression + determinism guards, so this band is intentionally loose.
+    # See the #3545 PLATFORM NOTE above for the macOS-arm64 vs
+    # Linux-x86_64 collinear-merge float-rounding divergence mechanism.
+    EXPECTED_SEGMENTS_RANGE = (390, 490)
     got_routes, got_segments, got_vias, got_length = ref
     exact = (got_routes, got_vias, got_length)
     expected_exact = (EXPECTED_ROUTES, EXPECTED_VIAS, EXPECTED_LENGTH)
@@ -642,9 +665,12 @@ def test_routing_output_deterministic_across_seeds(unrouted_pcb_path: Path) -> N
     lo, hi = EXPECTED_SEGMENTS_RANGE
     assert lo <= got_segments <= hi, (
         f"Board 02 segment count {got_segments} outside the documented "
-        f"platform band [{lo}, {hi}] (macOS-arm64: 397, Linux-x86_64: "
-        "399; see PLATFORM NOTE above).  Routes/vias/length matched the "
-        "exact pin, so this is a change in collinear segmentation -- "
-        "investigate merge_collinear/path_is_clear behaviour before "
-        "widening the band."
+        f"platform band [{lo}, {hi}] (macOS-arm64: 476, Linux-x86_64: "
+        "~390-399; see PLATFORM NOTE above and the #4196 re-baseline).  "
+        "Routes/vias/length matched the exact pin, so this is a change in "
+        "collinear segmentation -- investigate merge_collinear/path_is_clear "
+        "behaviour before widening the band.  NOTE: segment count is "
+        "platform/KiCad-version-sensitive and is NOT itself a "
+        "manufacturability signal; the exact routes/vias/length pins and "
+        "the cross-seed equality check are the real guards."
     )


rjwalters:
✅ **Approved!**

Low-risk, correctly-scoped stale-baseline widening. All acceptance criteria from #4196 verified.

## Verification

**Band WIDENED, not moved** — `EXPECTED_SEGMENTS_RANGE (390, 398) → (390, 490)`. Lower bound `390` is unchanged (still covers the Linux-x86_64 ~390-399 value that is green in CI); only the upper bound is raised to cover macOS-arm64's deterministic 476 with a small margin (490 > 476). No narrowing anywhere.

**CI stays green (confirmed empirically, not assumed)** — the `Test` job (Linux-x86_64 pytest, runs `test_routing_output_deterministic_across_seeds`) passed on this commit. The Linux value is inside the new band. All 16 checks pass; `mergeStateStatus` is CLEAN.

**Real guarantees preserved:**
- Cross-seed exact-equality assertion (`results[seed] == ref`) is untouched — determinism guarantee intact.
- Exact manufacturability pins untouched: `EXPECTED_ROUTES=22`, `EXPECTED_VIAS=24`, `EXPECTED_LENGTH=327.93`.
- Only the platform-sensitive segment-count band was loosened.

**Honestly a stale baseline, not papering over a regression** — routes/vias/length/reach are byte-identical to the 2026-07-09 sheet-centering pin; the +83 segment drift is in the finer-splitting (more-segments) direction, the same benign mode every prior PLATFORM NOTE entry documents. Builder confirmed determinism (476 across two runs + cross-seed 42/43/44), so no architect escalation warranted per the issue's tripwire.

**Housekeeping:**
- Dated `#4196` re-baseline note added to both the module docstring and the inline comment + assertion message, referencing the #3545 PLATFORM NOTE mechanism.
- Only `tests/router/test_board02_manufacturable_baseline.py` changed (+41/-15).
- `Closes #4196` present.

## Test Execution
- `Test` (Linux-x86_64 pytest) — ✅ pass (17m25s) — the load-bearing check that the widened band keeps CI green.
- `Board 02 End-to-End` — ✅ pass.
- All 16 CI checks — ✅ pass; merge state CLEAN.

PR #4199, opened 2026-07-15
feat(zones): auto-prioritize + carve overlapping zones batch zones (#4167)

### LINKED ISSUE #4167: zones batch: overlapping equal-priority bbox zones knowingly produce zero-copper zones (warning-only) — auto-prioritize or refuse
## Enhancement (warning-only on a known-fatal configuration)

`kct zones batch --power-nets 'A:F.Cu,B:F.Cu,...'` creates per-net pad-bbox rectangle zones, all at priority 0. On any board whose power pads interleave (i.e., the boards that need pours most), the bboxes overlap and the tool emits a wall of:

```
WARNING: Zone '/PGND' on F.Cu (priority 0) overlaps queued zone '/FUSED_LINE' (priority 0).
The new zone will get zero copper because the other zone has equal or higher priority.
```

…sixteen times in our case — then happily writes a zone set it has just proven will produce zero copper for most nets. The rev-A softstart recipe's own comments document this exact failure ("no single-rectangle-per-net tiling exists… lower-priority siblings with zero or islanded copper"), so the tool is generating a configuration the project already knows is broken.

## Suggested behavior

1. **Auto-assign priorities by bbox area** (smallest = highest) so nested/overlapping zones at least fill their exclusive regions — the standard KiCad idiom; or
2. Refuse (non-zero exit) when >N% of a zone's area is ceded at equal priority, with the warning promoted to the summary; or
3. Support non-rectangular per-net pours (pad-cluster convex hulls or track-following) for interleaved clusters.

Even just (1)+(promoting the warning count to the exit summary) would turn a silent-ish failure into an actionable one.

kicad-tools 0.15.0, softstart rev-B (rjwalters/softstart PR #28): 13-net power section, bbox pours abandoned in favor of planes + routed skeletons.

---

### Correction to the problem statement: `zones batch` zones are full-board-outline, not pad-bboxes

Traced `kct zones batch` end-to-end on current `origin/main`. The command's actual behavior is **worse** than described, and the fix location is different from what the "pad-bbox" framing implies:

- `_run_batch()` (`src/kicad_tools/cli/zones_cmd.py:451-536`) parses `--power-nets` via `parse_power_nets()` into `(net_name, layer)` pairs, then for each pair calls:
  ```python
  gen.add_zone(net=net_name, layer=layer, priority=priority, clearance=args.clearance)
  ```
  with **no `boundary` argument**. `ZoneGenerator.add_zone()` (`src/kicad_tools/zones/generator.py:654-724`) falls back to `self.board_outline` whenever `boundary is None` (line ~710: `actual_boundary = boundary if boundary is not None else self.board_outline`). There is no per-net pad-derived bbox anywhere in this path, and `zones batch` has no `--bbox` flag (that flag exists only on the single-zone `zones add` subcommand, `src/kicad_tools/cli/parser.py:2451-2470`, via `parse_bbox()`).
  - **Consequence**: every zone `zones batch` creates spans the **entire board outline**, i.e. 100% mutual overlap between any two same-layer zones — not a partial pad-cluster overlap that only bites on interleaved layouts. Any two nets on the same layer trigger this, always.
  - This matches the existing test's own docstring: `tests/test_zones_cmd.py:80-83`, `TestBatchOverlapWarnings.test_batch_warns_same_layer_overlap` — *"zones batch emits overlap warning for same-layer **full-board** zones."* That test currently asserts `ret == 0` (the "silent-ish success" this issue is about) — it will need updating once the fix lands, since after the fix the two nets in that fixture (`GND:F.Cu,+3V3:F.Cu`) should either produce disjoint carved outlines with zero warnings, or a materially reduced warning surfaced in the summary.
  - The reporter's description ("pad-bbox rectangle zones") most likely reflects the *softstart* project's own external bbox-computation wrapper around `zones add --bbox <cluster-bbox>` calls (looped, colloquially "batch"), not the literal `kct zones batch` subcommand — or reflects a version pre-dating a behavior change. Either way, the underlying defect (equal-priority, unconditionally-overlapping same-layer zones, warned but not prevented) is real and reproducible today via the literal `zones batch` command, just with 100%-overlap full-board rectangles rather than partial pad-bboxes. The fix should address both cases uniformly (see below) since the priority/carve machinery already generalizes across bbox sizes.

- The priority assignment in `_run_batch` (`zones_cmd.py:492-493`) is a **single hardcoded rule**: `priority = 1 if net_name.upper() in ("GND", "GNDA", "GNDD") else 0`. Every non-ground net gets priority 0, so N power nets sharing a layer are always in a mutual zero-priority-tie pileup. This is the direct cause of the "16 overlap warnings" the reporter saw.

- The overlap-warning machinery already exists and is shared between `zones add` and `zones batch`: `ZoneGenerator._check_overlap()` (`generator.py:554-652`) walks existing PCB zones and already-queued zones on the same layer, and for `priority <= existing.priority` emits `"...will get zero copper because the other zone has equal or higher priority"` (exact tie is explicitly folded into the "loses" branch — `<=`, not `<`). Warnings accumulate in `gen.warnings` (a `list[ZoneOverlapWarning]`, `generator.py:94-108`) and `_run_batch` prints the count + each message to stderr (`zones_cmd.py:511-514`) — but the function's **return code and success/failure signalling completely ignore `gen.warnings`**; only `errors` (raised `ValueError`s from `add_zone`, e.g. "net not found") affect the exit code. A run that produces 16 zero-copper warnings and 0 hard errors returns `0` and prints `Created N zone(s)` as if it succeeded.

### A more sophisticated allocator already exists — reuse it, don't reinvent

`src/kicad_tools/zones/generator.py` already contains a fully-built, well-tested, and extensively-documented solution to almost exactly this problem, but it is wired into a **different** command (`kct route`'s internal auto-pour, via `src/kicad_tools/router/auto_pour.py`), not `zones batch`:

- `_assign_layers_for_pour_nets()` (`generator.py:931-1069`) — stackup-aware layer + **priority** allocator. For layer-sharing nets it already assigns **distinct descending priorities** (`generator.py:1009-1017`, `:1041-1046`, `:1057-1063` — "N power nets get distinct descending priorities so multiple power planes coexist"). This is the priority half of what issue #4167 is asking for, just not area-based — it's assigned by net enumeration order, not bbox size.
- `_compute_pour_outlines()` (`generator.py:1252+`) — the outline/geometry half: for any layer hosting >1 zone, computes a per-net pad-cluster bbox (`_net_pad_positions_absolute()` + `_bbox_polygon()`, `generator.py:1092-1183`), then Shapely-subtracts higher-priority bboxes from lower-priority outlines (issue #3043's "carve" pass) with a pad-safe-bbox fallback and "subtrahend downgrade" rule for nested clusters (issue #3240), raising `ZonePartitionError` (`generator.py:111-150`) when even that can't produce a disjoint outline for some net.
- `auto_create_zones_for_pour_nets()` (`generator.py:~1700-1760`, entry point) composes both allocators and is the function `kct route`'s auto-pour calls today.

**Why `zones batch` can't just call `auto_create_zones_for_pour_nets()` directly**: its layer-assignment is driven by `NetClass` (GROUND/POWER) and board stackup, auto-picking the layer per net. `zones batch --power-nets 'A:F.Cu,B:F.Cu,...'` takes an **explicit user-specified layer per net** — the whole point of the flag is manual layer control. The fix for `_run_batch` needs to keep the user's explicit `(net, layer)` pairs and only auto-derive **priority** (by bbox area) and **outline carving** *within* each user-specified layer group — i.e., extract/generalize the priority-by-area and carve-by-layer-group logic from the existing allocator rather than routing through the NetClass-driven entry point.

### Recommended scope

**In scope (recommended fix):**

1. **Auto-assign priority by pad-cluster bbox area, descending (smallest area = highest priority), as the default for `zones batch`.** This is the standard KiCad idiom (smaller/nested zones must out-prioritize the zones they sit inside, or they get zero copper — same principle `_assign_layers_for_pour_nets` already applies via enumeration order, just correctly generalized to be area-driven instead of arbitrary/positional). Concretely:
   - For each layer, group the batch's `(net, layer)` entries that land on that layer.
   - For each net in a group, compute its pad-cluster bbox via `_net_pad_positions_absolute()` + `_bbox_polygon()` (already present, reusable as-is).
   - Assign priorities in descending order of ascending bbox area (smallest area → highest priority number). Existing explicit ground-gets-priority-1 special case can be folded in as a tiebreak/override, or superseded entirely by the area rule — recommend area rule wins uniformly, since it's the KiCad-idiomatic behavior the issue is asking for, and ground nets on typical boards already tend to have larger bboxes (spanning more pads) which naturally sorts them to lower priority — verify this doesn't regress the common "GND should be priority 1" case in testing (see Acceptance Criteria).
   - **Ties (equal bbox area, e.g. two single-pad nets, or the `SINGLE_PAD_FALLBACK_SIDE_MM` degenerate-bbox case)**: break deterministically by net name (alphabetical) so output is stable across runs/platforms — do not rely on dict/set iteration order.
   - Feed the same-layer group through the existing carve machinery (`_compute_pour_outlines`-style subtraction, or a `zones batch`-scoped adaptation of it) so overlapping zones actually receive disjoint outlines instead of merely getting non-conflicting priority numbers with 100%-overlapping full-board-outline boundaries. **Priority alone is not sufficient** — see the `_compute_pour_outlines` docstring (`generator.py:10-24`): "Earlier issues #2410/#2041/#2593 made the priorities distinct, which silenced the warning, but distinct priorities alone do not produce real copper for the losing zones -- the outlines have to be geometrically disjoint." The same lesson applies here: auto-prioritizing without carving would silence most of the *warnings* but still leave several zones with genuinely zero fillable copper (just without the diagnostic). Both halves (priority + carve) must ship together.

2. **Promote the overlap/zero-copper warning count into the exit summary, unconditionally (independent of scope 1 landing or not).** `_run_batch`'s final stats print (`zones_cmd.py:~530-534`, "Created N zone(s)") should also report `len(gen.warnings)` when nonzero, e.g. `"Created N zone(s) (M with zero-copper overlap warnings — see above)"`. This is a small, low-risk, high-value change that makes the failure non-silent even before/independent of the priority-and-carve fix, and should ship as part of the same PR since it directly addresses the issue's "even just... promoting the warning count to the exit summary" ask.

**Assessed but recommend deferring / treat as redundant once (1) ships: refuse-on-excessive-cede (suggestion 2 in the original issue).**
Once auto-prioritization + carving lands, the *legitimate* remaining failure mode is `ZonePartitionError` (nets whose pads are so spatially coincident that no disjoint outline exists even with the pad-safe fallback) — and that already **raises**, not warns (see `generator.py:111-150`, `_compute_pour_outlines`/`auto_create_zones_for_pour_nets` docstring "Raises: ZonePartitionError"). If `_run_batch`'s carve integration lets that exception propagate (rather than swallowing it into another warning), a hard, non-zero-exit refusal for the genuinely-unsalvageable case falls out of the carve fix for free, without needing a separate `>N%` ceded-area threshold heuristic. Recommend: implement (1), let `ZonePartitionError` surface as a hard CLI error (`Error: ...` to stderr, return 1) in `_run_batch` (mirroring how `_run_add`/`_run_batch` already turn `ValueError` into `return 1`), and treat suggestion (2)'s percentage-threshold heuristic as unnecessary/out of scope unless testing reveals a gap between "raises `ZonePartitionError`" and "meaningfully bad but not-fully-zero copper" cases that need a softer warning-with-refusal in between. If the implementer finds such a gap, file a follow-up issue rather than scope-creeping this one.

**Out of scope:** non-rectangular pours (suggestion 3) — convex-hull or track-following pours are a materially larger geometry change (new outline representation, new Shapely code paths) and are not needed once (1) is in place; the existing carve machinery already handles the reporter's stated failure mode (interleaved same-layer power nets) via rectangular bbox subtraction. Track as a separate future enhancement if rectangular carving proves insufficient in practice after this fix ships.

### KiCad priority semantics (verified against existing code + docs)

KiCad's zone fill resolver treats **higher `priority` values as filling "on top"**: when zones on the same layer overlap, the highest-priority zone gets the full overlapping region, and lower-priority zones are excluded from that region (this is the exact behavior `_check_overlap`'s warning describes — "the other zone has equal or higher priority" → new zone gets zero copper there). This is confirmed by the existing codebase's own conventions:
- `zones add --priority` help text: `"Zone fill priority (higher = fills later, default: 0)"` (`parser.py:132`) — "fills later" in KiCad's z-order sense = takes precedence/sits on top.
- `ZoneConfig.priority` docstring: `"priority: Zone fill priority (higher = fills later, on top)"` (`generator.py:161`).
- `add_ground_plane()` defaults `priority=1` explicitly documented as `"higher than power"` (`generator.py:728-737`).

"Smallest bbox area = highest priority" is the correct idiom because a small zone nested inside/overlapping a larger zone needs to "win" its own footprint of copper; if the larger (typically GND/plane) zone had equal-or-higher priority, the small zone would be entirely subsumed and get zero copper, which is precisely today's bug. This matches the `ZoneGenerator`'s existing `priority: int` field and the `_check_overlap`/`_compute_pour_outlines` semantics as-is — no data-model changes needed, only a new priority-*assignment* policy for the `zones batch` code path specifically.

### Baseline-shift risk: does auto-prioritization change output for currently-working (non-overlapping) boards?

Should be a no-op for boards whose `--power-nets` entries land on layers where each net has only one zone (no same-layer sibling) — those nets never enter `_check_overlap`'s same-layer branch and never get carved; they keep the full board outline exactly as today, with priority 0 (or 1 for GND, if that special-case is preserved) rather than the sorted values. **Flag for the implementer**: if the area-based rule is applied *unconditionally* (assigning distinct priorities even to single-occupant layers) rather than *only within same-layer groups of size ≥ 2*, output priorities would shift for boards that currently work fine, which is an unnecessary and observable behavior change (any downstream consumer diffing `zones list --format json` priority fields, or comparing generated `.kicad_pcb` zone S-expressions byte-for-byte, would see a diff with no functional benefit). The acceptance criteria below require a dedicated non-overlap regression test to catch this.

### Acceptance Criteria

- [ ] `zones batch` groups `--power-nets` entries by layer; for any layer with ≥ 2 nets, priorities are auto-assigned in descending order by ascending pad-cluster-bbox area (smallest bbox → highest priority), with deterministic alphabetical-by-net-name tiebreaking on equal area.
- [ ] For any layer with exactly 1 net (no same-layer sibling), behavior is unchanged: full board outline, priority as today (0, or 1 for GND-named nets) — no auto-prioritization applied, no carve invoked, byte-identical zone output to pre-fix behavior for these nets.
- [ ] Same-layer sibling zones receive carved (disjoint, or minimally-overlapping-only-at-the-pad-safe-fallback-margin) outlines via the existing `_net_pad_positions_absolute` / `_bbox_polygon` / carve-subtraction building blocks, not raw full-board or raw pad-bbox rectangles that fully overlap.
- [ ] On a synthetic board with interleaved power-net pads (same shape as the reporter's case and analogous to `tests/test_zone_priority.py`'s `_make_3net_nested_pcb` / `_make_2net_overlap_pcb` fixtures), `zones batch` produces zero (or a documented, far smaller number of) "will get zero copper" warnings, and each zone's carved outline has positive area distinct from its siblings.
- [ ] When carving cannot produce a disjoint outline for some net (fully-coincident pads, mirroring the `ZonePartitionError` case in `_compute_pour_outlines`), `zones batch` exits non-zero with an actionable error message (reusing/adapting `ZonePartitionError`'s message format), not a silent success.
- [ ] The exit summary (`zones_cmd.py` `_run_batch`'s final print) reports the number of remaining overlap/zero-copper warnings when `gen.warnings` is non-empty, in addition to (or instead of) requiring the caller to scroll back through stderr.
- [ ] `tests/test_zones_cmd.py::TestBatchOverlapWarnings::test_batch_warns_same_layer_overlap` is updated to reflect the new behavior (either asserting zero/fewer warnings for its 2-net fixture post-carve, or is retargeted to a genuinely-unsalvageable fixture that still legitimately warns/refuses).
- [ ] No behavior change (byte-identical or functionally-equivalent zone output) for existing non-overlapping `zones batch` usages — add a regression test asserting this.

### Affected Files

- `src/kicad_tools/cli/zones_cmd.py` — `_run_batch()` (currently `zones_cmd.py:451-536`): replace the hardcoded `priority = 1 if net_name.upper() in (...) else 0` rule with layer-grouped, area-based priority assignment; invoke carve logic for same-layer groups; surface `ZonePartitionError` as a hard CLI error; add warning count to the exit summary.
- `src/kicad_tools/zones/generator.py` — reuse `_net_pad_positions_absolute()`, `_bbox_polygon()`, `_clip_polygon_to_outline()` (all already present, ~lines 1092-1250); likely needs a new small helper (or a generalized/parameterized variant of `_compute_pour_outlines()`) that accepts caller-supplied `(net, layer, priority)` assignments directly (rather than deriving them from `NetClass`), since `_run_batch`'s layers are explicit/user-specified rather than allocator-derived. `ZonePartitionError` (lines 111-150) is reusable as-is.
- `tests/test_zones_cmd.py` — update `TestBatchOverlapWarnings::test_batch_warns_same_layer_overlap` (lines 77-101); add new test classes for: layer-grouped area-based priority assignment, carve-produces-disjoint-outlines, non-overlap no-op regression, `ZonePartitionError` → non-zero exit, and summary warning-count reporting.
- `tests/test_zone_priority.py` — reference for reusable synthetic-fixture patterns (`_make_2net_overlap_pcb`, `_make_3net_nested_pcb`, `_polygon_area`, `_polygons_overlap_area` helpers at lines 43-242); new `zones batch`-specific fixtures can follow the same construction pattern if not directly reusable.
- `docs/reference/cli.md` — `zones batch` documentation should note the new default auto-prioritization behavior (mirrors the doc update pattern PR #4193 just used for the in-place-output default change).

### Test Plan

- [ ] Automated: new synthetic-board test(s) in `tests/test_zones_cmd.py` (or a new `tests/test_zones_batch_priority.py`) constructing a PCB with 3+ power nets whose pads spatially interleave on the same layer (mirror `_make_3net_nested_pcb` / `_make_2net_overlap_pcb` from `tests/test_zone_priority.py`), asserting: (a) auto-assigned priorities are strictly descending by ascending bbox area, (b) zero or a materially reduced warning count vs. today's baseline, (c) each output zone's boundary polygon has positive area and does not fully contain/be-contained-by a sibling's boundary at equal priority.
- [ ] Automated: non-overlap regression test — single net per layer, assert unchanged (full board outline, unchanged priority, no carve invoked).
- [ ] Automated: tie-break test — two nets with equal-area (or degenerate single-pad) bboxes on the same layer, assert deterministic alphabetical-by-net-name priority ordering (run twice / assert stable across dict-ordering-sensitive code paths).
- [ ] Automated: `ZonePartitionError` propagation — synthetic fully-coincident-pad fixture (mirror `_make_3net_coincident_pads_pcb` from `tests/test_zone_priority.py`) drives `zones batch` to a non-zero exit code with an actionable stderr message, not a silent `Created N zone(s)`.
- [ ] Automated: exit-summary warning-count assertion — a fixture that still legitimately produces a residual warning (if any exist post-carve) asserts the count appears in the final summary output, not only scattered across earlier stderr lines.
- [ ] Manual/update: `tests/test_zones_cmd.py::TestBatchOverlapWarnings::test_batch_warns_same_layer_overlap` updated to match new behavior for its existing 2-net `GND:F.Cu,+3V3:F.Cu` fixture.
- [ ] Manual verification: run `uv run kct zones batch <board>.kicad_pcb --power-nets 'A:F.Cu,B:F.Cu,C:F.Cu' --dry-run -v` against a hand-built interleaved-pads fixture and visually confirm (via `--verbose` S-expression dump or `zones list`) that each zone's boundary differs and priorities descend by area.

### Dependencies

None. Builds directly on current `origin/main` (post #4166/PR #4193, which changed `zones add`/`batch` default output to in-place — already merged and reflected in the file/line references above).

rjwalters:
## Summary

`kct zones batch` used to add every zone at the **full board outline** (`add_zone()` was called with no `boundary`), so any two same-layer nets 100%-overlapped. Combined with the hardcoded priority rule (`GND`→1, else→0), N power nets on one layer piled up at a zero-priority tie, and KiCad's fill resolver silently gave the lower-priority siblings **zero copper**. `_run_batch` warned but ignored `gen.warnings` in its exit code/summary, so the run reported success.

This PR fixes both the geometry and the reporting. Within each **user-specified** layer group, `_run_batch` now:

1. **Auto-assigns priority by pad-cluster bbox area** — smallest area ⇒ highest priority (the KiCad idiom where a small/nested zone must fill on top of the larger zones it sits inside). Deterministic alphabetical-by-net-name tiebreak on equal area. Replaces the hardcoded `GND=1/else=0` rule.
2. **Carves per-net outlines** via the existing `_compute_pour_outlines` Shapely-subtract machinery (#3043/#3240), so overlapping same-layer zones become geometrically disjoint and every net receives **real, non-zero copper** — not merely reprioritized.
3. **Promotes the zero-copper warning count into the exit summary** (non-silent).
4. Lets **`ZonePartitionError`** (fully-coincident pad clusters) surface as a hard non-zero-exit error instead of silently writing zero-copper zones.

Nets **sole on their layer are unchanged**: full board outline, prior priority (1 for GND-named, else 0), no carve — byte-identical to pre-fix behavior.

## Design

Rather than routing through the NetClass-driven `auto_create_zones_for_pour_nets()` (which auto-picks layers), a new `assign_batch_zone_priorities_and_outlines()` reuses the auto-pour building blocks (`_net_pad_positions_absolute`, `_bbox_polygon`, `_compute_pour_outlines`) while preserving the user's explicit `(net, layer)` pairs. It groups by layer, area-ranks within each group, and feeds `(net, layer, priority)` assignments straight into the existing carve pass. No new geometry code paths; no data-model changes.

## Do carved zones get real non-zero copper?

**Yes.** `tests/test_zones_batch_priority.py::TestBatchCommandEndToEnd::test_interleaved_zones_written_disjoint_with_nonzero_copper` writes the board to disk, reloads it, and asserts every F.Cu zone has positive area and is pairwise-disjoint from its siblings. This is real copper, not just distinct priority numbers. The carve reuse was a clean extraction of existing building blocks — it did **not** require an architect-level change.

Note on the interior-hole limit (inherent to the pre-existing machinery, not new): when a higher-priority sibling sits *strictly interior* to a lower-priority net's bbox, the difference is a polygon with a hole, which KiCad zones can't represent. The existing `_compute_pour_outlines` keeps the largest disjoint fragment (edge-crossing case) or falls back to the pad-safe bbox — the realistic interleaved-power case (the reporter's case). Only genuinely-coincident pad clusters raise `ZonePartitionError`.

## Out of scope (per curator)

- Non-rectangular/convex-hull pours.
- Separate `>N%`-ceded refusal threshold — `ZonePartitionError` covers hard-refusal.

## Tests

New `tests/test_zones_batch_priority.py`:
- area-based priority ordering (SMALL > MID > BIG)
- carve produces disjoint, positive-area outlines (unit + end-to-end on-disk)
- single-net-per-layer no-op regression (full outline, legacy priority, no carve)
- equal-area alphabetical tiebreak + input-order stability
- coincident pads → `ZonePartitionError` → non-zero exit, no "Created" line
- no spurious warning on disjoint batch; warning-count promoted to summary

Updated `tests/test_zones_cmd.py::TestBatchOverlapWarnings` to the carved-and-summarized behavior.

Docs: `docs/reference/cli.md` `zones` section documents the new auto-prioritization default.

## Verification

- `uv run pytest tests/test_zones_cmd.py tests/test_zone_priority.py tests/test_zones_batch_priority.py` — 95 passed
- Broader `-k "zone or pour or auto_pour"` — 989 passed, 7 skipped (no regressions)
- `ruff check` / `ruff format` clean
- mypy baseline: 0 new errors (baseline actually shrank by 39)

Closes #4167

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -670,6 +670,23 @@ reads its own prior output, chaining `zones add` calls against one file
 accumulates zones instead of silently discarding earlier ones. Pass an
 explicit `-o <path>` to write a side file and leave the input untouched.
 
+**`batch` auto-prioritization (issue #4167):** when two or more of the
+`--power-nets` entries land on the **same layer**, `zones batch` no longer
+emits full-board zones that all overlap at equal priority (which silently gave
+the lower-priority zones zero copper). Instead, within each user-specified
+layer group it (1) auto-assigns fill **priority by ascending pad-cluster bbox
+area** — the smallest net gets the highest priority, the KiCad idiom where a
+small/nested zone must fill on top of the larger zones it sits inside (equal
+areas break deterministically, alphabetical by net name) — and (2) **carves the
+per-net outlines** to each net's own pad cluster minus any higher-priority
+sibling, so the zones are geometrically disjoint and every net receives real
+copper. Nets that are **sole on their layer are unaffected**: they keep the
+full board outline and their prior priority (1 for GND-named nets, else 0). If
+carving genuinely cannot produce disjoint copper for some net (fully-coincident
+pad clusters), the command exits non-zero with an actionable error rather than
+writing zero-copper zones. The final summary reports how many zones still cede
+copper via an overlap warning.
+
 | Option | Description |
 |--------|-------------|
 | `-o`, `--output PATH` | Output PCB (default: overwrite input) |
--- a/src/kicad_tools/cli/zones_cmd.py
+++ b/src/kicad_tools/cli/zones_cmd.py
@@ -449,8 +449,22 @@ def _run_list(args) -> int:
 
 
 def _run_batch(args) -> int:
-    """Add multiple zones from a power-nets specification."""
-    from kicad_tools.zones import ZoneGenerator, parse_power_nets
+    """Add multiple zones from a power-nets specification.
+
+    Priorities and per-net boundaries are auto-derived (issue #4167):
+    within each user-specified layer group, zones are prioritized by
+    ascending pad-cluster bbox area (smallest area => highest priority,
+    the KiCad idiom) and given carved, geometrically-disjoint outlines so
+    overlapping same-layer zones do not zero each other out.  Nets that
+    are sole on their layer keep the full board outline and their legacy
+    priority (1 for GND-named nets, 0 otherwise) unchanged.
+    """
+    from kicad_tools.zones import (
+        ZoneGenerator,
+        ZonePartitionError,
+        assign_batch_zone_priorities_and_outlines,
+        parse_power_nets,
+    )
 
     pcb_path = Path(args.pcb)
     if not pcb_path.exists():
@@ -483,14 +497,28 @@ def _run_batch(args) -> int:
         print(f"Error loading PCB: {e}", file=sys.stderr)
         return 1
 
-    # Add zones with priorities (GND gets highest priority)
+    # Auto-assign priorities (by pad-cluster bbox area) and carve outlines
+    # for same-layer groups so overlapping zones receive disjoint copper
+    # instead of full-board rectangles that zero each other out (#4167).
+    try:
+        allocation = assign_batch_zone_priorities_and_outlines(
+            gen.pcb,
+            gen.board_outline,
+            power_nets,
+        )
+    except ZonePartitionError as e:
+        # Carving genuinely cannot produce disjoint copper for some net
+        # (fully-coincident pad clusters).  Refuse rather than silently
+        # writing zero-copper zones.
+        print(f"Error: {e}", file=sys.stderr)
+        return 1
+
     if not quiet:
         print(f"\nAdding {len(power_nets)} zone(s):")
 
     errors = []
     for net_name, layer in power_nets:
-        # GND typically gets higher priority (fills last, on top)
-        priority = 1 if net_name.upper() in ("GND", "GNDA", "GNDD") else 0
+        priority, boundary = allocation.get(net_name, (0, None))
 
         try:
             gen.add_zone(
@@ -498,9 +526,11 @@ def _run_batch(args) -> int:
                 layer=layer,
                 priority=priority,
                 clearance=args.clearance,
+                boundary=boundary,
             )
             if not quiet:
-                print(f"  {net_name} on {layer} (priority {priority})")
+                carved = " (carved)" if boundary is not None else ""
+                print(f"  {net_name} on {layer} (priority {priority}){carved}")
         except ValueError as e:
             errors.append(f"  {net_name}: {e}")
 
@@ -518,6 +548,7 @@ def _run_batch(args) -> int:
     if args.dry_run:
         if not quiet:
             print("\n--- Dry run - not saving ---")
+            _print_batch_summary(gen, quiet)
             if args.verbose:
                 print("\nGenerated S-expression:")
                 print(gen.generate_sexp())
@@ -534,10 +565,30 @@ def _run_batch(args) -> int:
         return 1
 
     if not quiet:
-        stats = gen.get_statistics()
-        print(f"\nCreated {stats['zone_count']} zone(s)")
+        _print_batch_summary(gen, quiet)
 
     return 0 if not errors else 1
+
+
+def _print_batch_summary(gen, quiet: bool) -> None:
+    """Print the batch summary, promoting any zero-copper warning count.
+
+    Surfaces ``len(gen.warnings)`` (zones that still cede area / would get
+    zero copper) into the final summary line so the failure is non-silent
+    (issue #4167), rather than only being visible in scrolled-back stderr.
+    """
+    if quiet:
+        return
+    stats = gen.get_statistics()
+    zone_count = stats["zone_count"]
+    warn_count = len(gen.warnings)
+    if warn_count:
+        print(
+            f"\nCreated {zone_count} zone(s) "
+            f"({warn_count} with zero-copper overlap warning(s) -- see above)"
+        )
+    else:
+        print(f"\nCreated {zone_count} zone(s)")
 
 
 def _run_fill(args) -> int:
--- a/src/kicad_tools/cli/zones_cmd.py
+++ b/src/kicad_tools/cli/zones_cmd.py
@@ -518,7 +518,10 @@ def _run_batch(args) -> int:
 
     errors = []
     for net_name, layer in power_nets:
-        priority, boundary = allocation.get(net_name, (0, None))
+        # Keyed by (net, layer): a net on multiple layers must get the
+        # correct per-layer outline (full board where it is sole, carved
+        # where it overlaps siblings) -- issue #4167.
+        priority, boundary = allocation.get((net_name, layer), (0, None))
 
         try:
             gen.add_zone(
--- a/src/kicad_tools/zones/__init__.py
+++ b/src/kicad_tools/zones/__init__.py
@@ -36,6 +36,7 @@ from .generator import (
     ZoneGenerator,
     ZoneOverlapWarning,
     ZonePartitionError,
+    assign_batch_zone_priorities_and_outlines,
     auto_create_zones_for_pour_nets,
     parse_power_nets,
 )
@@ -45,6 +46,7 @@ __all__ = [
     "ZoneGenerator",
     "ZoneOverlapWarning",
     "ZonePartitionError",
+    "assign_batch_zone_priorities_and_outlines",
     "auto_create_zones_for_pour_nets",
     "parse_power_nets",
 ]
--- a/src/kicad_tools/zones/generator.py
+++ b/src/kicad_tools/zones/generator.py
@@ -1699,7 +1699,7 @@ def assign_batch_zone_priorities_and_outlines(
     board_outline: list[tuple[float, float]],
     net_layer_pairs: list[tuple[str, str]],
     margin_mm: float = DEFAULT_POUR_BBOX_MARGIN_MM,
-) -> dict[str, tuple[int, list[tuple[float, float]] | None]]:
+) -> dict[tuple[str, str], tuple[int, list[tuple[float, float]] | None]]:
     """Assign priorities + carved outlines for explicit ``(net, layer)`` pairs.
 
     This is the ``zones batch`` counterpart to
@@ -1732,9 +1732,14 @@ def assign_batch_zone_priorities_and_outlines(
         margin_mm: Margin around each pad bbox, in mm.
 
     Returns:
-        Dict mapping ``net_name`` -> ``(priority, outline)`` where
+        Dict mapping ``(net_name, layer)`` -> ``(priority, outline)`` where
         ``outline`` is ``None`` for sole-zone layers (caller falls back to
-        the board outline) or a carved polygon for shared-layer zones.
+        the board outline) or a carved polygon for shared-layer zones.  The
+        key is the ``(net, layer)`` pair -- *not* the net name alone -- so a
+        net appearing on multiple layers (e.g. a GND plane shared on
+        ``F.Cu`` and sole on ``B.Cu``) gets the correct per-layer outline:
+        full board on the layer where it is sole, carved on the layer where
+        it overlaps siblings (issue #4167 regression fix).
 
     Raises:
         ZonePartitionError: If carving cannot produce disjoint copper for
@@ -1754,9 +1759,14 @@ def assign_batch_zone_priorities_and_outlines(
     def _legacy_priority(net_name: str) -> int:
         return 1 if net_name.upper() in ("GND", "GNDA", "GNDD") else 0
 
-    # Build (net, layer, priority) assignments feeding _compute_pour_outlines.
-    assignments: list[tuple[str, str, int]] = []
-    result: dict[str, tuple[int, list[tuple[float, float]] | None]] = {}
+    # Build the result keyed by (net, layer) -- NOT net name alone.  A net
+    # can legitimately appear on multiple layers (e.g. a GND plane shared on
+    # F.Cu and sole on B.Cu); keying by net name alone made the two layers
+    # collide on one dict entry, so the last-written (priority, outline) was
+    # wrongly applied to both add_zone calls -- silently giving the
+    # sole-layer instance a carved strip instead of the full board outline
+    # (issue #4167 regression).
+    result: dict[tuple[str, str], tuple[int, list[tuple[float, float]] | None]] = {}
 
     for layer, nets in layer_groups.items():
         if len(nets) < 2:
@@ -1764,8 +1774,7 @@ def assign_batch_zone_priorities_and_outlines(
             # outline, legacy priority, no carve.
             net_name = nets[0]
             priority = _legacy_priority(net_name)
-            assignments.append((net_name, layer, priority))
-            result[net_name] = (priority, None)
+            result[(net_name, layer)] = (priority, None)
             continue
 
         # >= 2 nets share this layer.  Compute each net's pad-cluster bbox
@@ -1782,19 +1791,22 @@ def assign_batch_zone_priorities_and_outlines(
         # are distinct and strictly descending by area.
         ordered = sorted(nets, key=lambda n: (areas[n], n))
         n = len(ordered)
+        priorities: dict[str, int] = {}
         for rank, net_name in enumerate(ordered):
             # rank 0 = smallest area => highest priority.
-            priority = n - rank
-            assignments.append((net_name, layer, priority))
-            # Outline filled in below from _compute_pour_outlines.
-            result[net_name] = (priority, None)
+            priorities[net_name] = n - rank
 
-    # Carve outlines for shared-layer groups.  Sole-zone layers return
-    # None (handled above) -- _compute_pour_outlines also returns None for
-    # them, so this is consistent.
-    outlines = _compute_pour_outlines(pcb, assignments, board_outline, margin_mm)
-    for net_name, (priority, _) in list(result.items()):
-        result[net_name] = (priority, outlines.get(net_name))
+        # Carve outlines for THIS layer group only.  Because the group is a
+        # single layer, its net names are unique, so _compute_pour_outlines'
+        # net-name-keyed return dict cannot collide across layers even when
+        # the same net also lives on a different layer's group.
+        assignments = [(net_name, layer, priorities[net_name]) for net_name in nets]
+        outlines = _compute_pour_outlines(pcb, assignments, board_outline, margin_mm)
+        for net_name in nets:
+            result[(net_name, layer)] = (
+                priorities[net_name],
+                outlines.get(net_name),
+            )
 
     return result
 
--- a/src/kicad_tools/zones/generator.py
+++ b/src/kicad_tools/zones/generator.py
@@ -1694,6 +1694,111 @@ def _polygon_overlap_area(
         return 0.0
 
 
+def assign_batch_zone_priorities_and_outlines(
+    pcb: PCB,
+    board_outline: list[tuple[float, float]],
+    net_layer_pairs: list[tuple[str, str]],
+    margin_mm: float = DEFAULT_POUR_BBOX_MARGIN_MM,
+) -> dict[str, tuple[int, list[tuple[float, float]] | None]]:
+    """Assign priorities + carved outlines for explicit ``(net, layer)`` pairs.
+
+    This is the ``zones batch`` counterpart to
+    :func:`_assign_layers_for_pour_nets` + :func:`_compute_pour_outlines`.
+    Unlike the auto-pour path, the caller supplies an explicit layer per
+    net (the whole point of ``--power-nets 'A:F.Cu,B:F.Cu,...'``), so this
+    function does **not** derive layers from :class:`NetClass`.  It only
+    derives, *within each user-specified layer group*:
+
+    * **Priority** -- by ascending pad-cluster bbox area (smallest area =>
+      highest priority, the KiCad idiom where small/nested zones must fill
+      on top of the larger zones they sit inside, or they get zero copper).
+      Ties (equal area, e.g. two single-pad nets) break deterministically
+      by net name (alphabetical) so output is stable across runs/platforms.
+    * **Outline** -- via the existing :func:`_compute_pour_outlines` carve
+      machinery (issue #3043/#3240), so overlapping same-layer zones receive
+      geometrically disjoint outlines instead of full-board rectangles that
+      100%-overlap and zero each other out.
+
+    Behavior is a **no-op** for any layer hosting exactly one net: that net
+    keeps the full board outline (``None``) and its legacy priority (1 for
+    GND-named nets, 0 otherwise), so existing non-overlapping ``zones batch``
+    usages are byte-identical to pre-fix behavior (issue #4167).
+
+    Args:
+        pcb: Loaded PCB object (used to look up pad positions).
+        board_outline: Sheet-absolute board outline polygon (clip mask).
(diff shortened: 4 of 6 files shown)


rjwalters:
❌ **Changes Requested**

Strong implementation overall — the area-based priority ranking, deterministic alphabetical tiebreak, carve reuse, ZonePartitionError hard-exit, and warning-count promotion all check out, and the load-bearing end-to-end test genuinely writes → reloads from disk → asserts positive area + pairwise-disjoint copper. 95 targeted tests pass locally, ruff clean. But I found one blocking correctness regression that must be fixed before merge.

## Blocking: same net on multiple layers gets the wrong (carved) outline

`assign_batch_zone_priorities_and_outlines()` builds its `result` dict keyed **only by `net_name`**, and `_run_batch` looks up `allocation.get(net_name, ...)` per `(net, layer)` pair. When the same net appears on two layers — a very common ground-plane spec like `GND:F.Cu,SIG:F.Cu,GND:In1.Cu` (or `GND:B.Cu`) — the two layers collide on the single `result["GND"]` key. The last-written entry wins, and its `(priority, boundary)` is then applied to **both** `add_zone` calls.

Concrete repro (100×50 board, GND shares F.Cu with SIG, GND also sole on B.Cu):

```
kct zones batch board.kicad_pcb --power-nets 'GND:F.Cu,SIG:F.Cu,GND:B.Cu' -o board.kicad_pcb
```

Result (reloaded from disk):
```
GND F.Cu: area=2466.4 mm²   (correctly carved — shares layer with SIG)
SIG F.Cu: area=112.4 mm²    (correctly carved)
GND B.Cu: area=2466.4 mm²   ← WRONG: got F.Cu's carved strip
```

`GND` on `B.Cu` is **sole on its layer**, so per the PR's own contract (and the `_compute_pour_outlines` docstring re: "GND return-path plane… where GND sits alone on In1.Cu/B.Cu") it must keep the **full board outline** (~5000 mm²). Instead it inherits GND's F.Cu carved outline and loses ~half the plane copper. This is silent (ret 0, no warning) — exactly the class of silent-zero/short copper this PR set out to eliminate.

Pre-fix behavior for this spec was correct: the old loop computed priority inline per `(net, layer)` and always passed `boundary=None` (full outline), so `GND:B.Cu` got the full board. So this is a genuine regression, not a pre-existing quirk.

### Suggested fix
Key the allocation by `(net_name, layer)` instead of `net_name` alone:
- In the helper, make `result` / the returned dict keyed on `(net, layer)` (the `_compute_pour_outlines` output is already computed per assignment, so the carved outline for each layer is available).
- In `_run_batch`, look up `allocation.get((net_name, layer), (0, None))`.

Add a regression test: same net on a shared layer AND sole on a second layer → the sole-layer zone keeps the full board outline (None / full-area) while the shared-layer instance is carved.

## Minor (non-blocking, fix if touching the file)
- `_make_coincident_pads_pcb` docstring says "-> no partition" but the test asserts `ZonePartitionError` is raised; the docstring reads backwards.

Everything else looks good — once the multi-layer keying is fixed I expect to approve. CI was still pending at review time; please ensure it's green on the re-push.

rjwalters:
Addressed the blocking multi-layer keying regression.

**Root cause**: `assign_batch_zone_priorities_and_outlines()` keyed its result dict by net name alone, and `_run_batch` looked it up by net name. For a same-net-on-two-layers spec (`GND:F.Cu,SIG:F.Cu,GND:B.Cu`), GND's two layers collided on one key, so the shared-layer carved outline leaked onto the sole-layer instance.

**Fix**:
- Result dict is now keyed by `(net, layer)`. Carved outlines are computed per layer group, so `_compute_pour_outlines`' net-name-keyed return can't collide across layers (net names are unique within a single-layer group).
- `_run_batch` now looks up `allocation.get((net_name, layer), (0, None))`.
- A net sole on a layer keeps the full board outline; on a shared layer it gets the carved outline.

**GND-on-sole-layer area, judge's exact repro (`GND:F.Cu,SIG:F.Cu,GND:B.Cu`, 100x50 board, on-disk reload)**:
| zone | before | after |
|---|---|---|
| GND B.Cu (sole) | 2466.4 mm² (WRONG carved strip) | **5000.0 mm² (full board)** |
| GND F.Cu (shared) | 2466.4 mm² | 2466.4 mm² (correctly carved) |
| SIG F.Cu (shared) | 12.0 mm² | 12.0 mm² |

**Regression tests added** (`tests/test_zones_batch_priority.py`):
- `test_same_net_two_layers_keyed_per_layer` (unit): GND on B.Cu → `None`/full-board outline, legacy priority 1; GND on F.Cu → carved, area between 0 and full board; SIG higher priority + own carve.
- `test_same_net_two_layers_sole_layer_gets_full_outline` (end-to-end, writes → reloads from disk): GND B.Cu > 4900 mm² and strictly larger than the carved GND F.Cu.

Also fixed the inverted `_make_coincident_pads_pcb` docstring (it raises `ZonePartitionError`).

**Verification** (in worktree, C++ backend available 1.0.0):
- `pytest tests/test_zones_batch_priority.py tests/test_zones_cmd.py tests/test_zone_priority.py` → 97 passed.
- Broader `-k "zone or pour or auto_pour"` → 992 passed, 1 skipped (no regressions).
- ruff check + format clean; mypy baseline 0 new errors.

rjwalters:
✅ **Approved!** (RE-review after Doctor cycle — the (net, layer) keying blocker is fixed and verified load-bearing.)

## Blocker from prior review: RESOLVED

The prior blocker was that `assign_batch_zone_priorities_and_outlines()` keyed its result dict by net name alone, so a `GND:F.Cu,SIG:F.Cu,GND:B.Cu` spec collided GND's two layers and silently applied F.Cu's carved strip (~2466 mm²) to the sole-layer B.Cu instance, losing ~half the ground plane.

Commit `809890cf` fixes this correctly in BOTH ends:
- **Producer** (`generator.py`): return type is now `dict[tuple[str, str], ...]`; outlines are carved **per layer group** (`assignments` built inside the per-layer loop), and the result is keyed `result[(net_name, layer)] = ...`. Net names are unique within a single-layer group, so `_compute_pour_outlines`' net-name-keyed return cannot collide across layers.
- **Consumer** (`zones_cmd.py:524`): `allocation.get((net_name, layer), (0, None))` — matches the new key shape.

No collision remains for a net on multiple layers.

## Regression tests verified load-bearing

I confirmed the on-disk test genuinely reproduces the bug, not just the happy path:
- Reverted the consumer lookup to the old net-name key and re-ran `test_same_net_two_layers_sole_layer_gets_full_outline` — it **failed** exactly as expected: F.Cu GND came back uncarved at 5000.0 mm² (`assert 5000.0 < 5000.0`), i.e. the carved-vs-full distinction is actually asserted.
- Restored the file; test passes again.

The test asserts GND-on-sole-B.Cu > 4900 mm² (full ~5000 mm² board) and GND-on-shared-F.Cu strictly smaller (carved). This is the correct load-bearing proof.

## Rest of the change unchanged since prior approval
The carve (real non-zero copper), area-priority ordering, non-silent summary, and ZonePartitionError refusal paths are untouched; the 5 existing allocator unit tests were mechanically updated to `(net, layer)` keys. The inverted `_make_coincident_pads_pcb` docstring is also fixed.

## Verification
- `pytest tests/test_zones_batch_priority.py`: **12 passed** (incl. 2 new regression tests).
- `ruff check` on the 3 changed files: clean.
- `mypy` on `generator.py` + `zones_cmd.py`: **0 new errors** (remaining errors are pre-existing repo backlog in unrelated modules, e.g. `mcp/server.py`).
- **CI: all 16 checks pass** (Test 17m52s, Board 05/06/07 + Diff-Pair + Match-Group routing regressions all green, Lint, Type Check, C++ Build).
- `mergeStateStatus`: **CLEAN** / MERGEABLE.

PR #4200, opened 2026-07-15
fix(stitch): reject stitch vias coincident with different-net drills (#4177)

### LINKED ISSUE #4177: stitch: places vias at occupied locations — hole-to-hole -0.000mm (coincident holes, different nets) and shorts through inner planes
## Original Issue

`kct stitch --net GND` / `--net +3.3V` on a 4-layer board (planes on In1/In2) placed stitching vias at locations already occupied by other copper and holes. Evidence from the same toolchain's own `kct check` immediately after:

```
[X] dimension_drill_clearance (different-net)
    Hole-to-hole clearance -0.000mm < minimum 0.500mm
    Nets: /SCAP_NEG / /GATE_RTN_POS          # two holes at the SAME position
[X] dimension_drill_clearance (different-net)
    Hole-to-hole clearance 0.475mm < minimum 0.500mm
    Nets: /LED_A_POS / /LED_K_POS
```

and kicad-cli DRC (after --refill-zones, so fills are KiCad-correct) reporting a family of `shorting_items` where one side is always the plane net (`+3.3V ↔ GND`, `+3.3V ↔ /NRST`, `+3.3V ↔ /FUSED_LINE`, `+3.3V ↔ /REF_1V65`, ...) — consistent with stitch via barrels landing on/next to other nets' copper and holes.

A coincident-hole different-net pair is an unfixable metallic short no zone refill can save. The stitch step has `--clearance` and `--drc` options, but the *default* behavior produced this; via placement should always validate candidate positions against existing pads/vias/holes (and same-position self-collisions) before insertion, or fail the position and pick another.

kicad-tools 0.15.0, softstart rev-B (rjwalters/softstart PR #28): planes stripped and stitch abandoned pending this; happy to share the pre/post artifacts.

---

### Root cause (confirmed by code reading, not yet reproduced against the reporter's board)

`kct stitch` (`src/kicad_tools/cli/stitch_cmd.py`) is **not** missing hole-to-hole validation wholesale — there is extensive, default-on (not gated behind `--drc`) machinery for it:

- `calculate_via_position()` (straight-line placement, line ~1620) checks `other_net_drills` edge-to-edge against `MIN_HOLE_TO_HOLE_CLEARANCE` (0.5mm) when `via_drill > 0` (line ~1779-1795).
- `run_blanket_stitch()` and `run_thermal_stitch()` call `check_via_clearance()` (a thin wrapper over `router/via_clearance.py::point_clear_of_copper`, which calls `drill_hole_to_hole_clear`) with `min_hole_to_hole=MIN_HOLE_TO_HOLE_CLEARANCE` — this is the same machinery the C++ router uses.
- `run_stitch()` additionally runs a post-placement "chokepoint" pass (lines 4872-4937, added for #3855) intended to catch everything the per-strategy checks miss, by rebuilding a drill registry via `find_all_drills(sexp, exclude_nets=net_numbers)` and rejecting any committed via within 0.5mm edge-to-edge of an existing drill.

However, this default-on validation has **two concrete gaps** that reproduce exactly the symptoms in this issue:

**Gap 1 — dog-leg and extended-escape strategies never receive the drill registry at all.**
`calculate_dogleg_via_position()` (line 1995) and `calculate_extended_escape_position()` (line 2267) have **no** `other_net_drills` / `min_hole_to_hole` parameters in their signatures, and the call sites in `run_stitch._attempt_placement()` (lines ~4433 and ~4451, plus the micro-via retries at ~4486/~4503) never pass them. These are exactly the fallback strategies invoked when straight-line placement fails — i.e. the congested, densely-populated regions where a coincident-hole collision is most likely. The intent (per the #3855 comment on the post-placement pass, line 4872) was that the post-placement pass at 4884-4937 would be the safety net for these two paths — see Gap 2 for why that net has a hole.

**Gap 2 — the post-placement "chokepoint" pass excludes *all* stitch-target nets' pre-existing drills.**
At line 4889:
```python
existing_board_drills = find_all_drills(sexp, exclude_nets=net_numbers)
```
`net_numbers` is the set of **all** net numbers being stitched in this invocation (`--net GND --net +3.3V` → both GND's and +3.3V's net numbers). This means a **pre-existing** via or through-hole pad belonging to `+3.3V` is never checked when placing a **GND** stitch via, and vice versa — the two nets are mutually invisible to each other's pre-existing drill registry. `placed_so_far` (populated during the same pass) does correctly catch collisions between *newly placed* stitch vias of different target nets in the same run, but it does **not** cover collisions against drilled holes that already existed on the board before this stitch invocation (existing component pads, existing vias from a prior stitch/route pass). That is precisely the reported symptom: "vias at locations ALREADY OCCUPIED by other copper/holes" on a multi-net `--net GND --net +3.3V` run, producing `-0.000mm` (i.e. literally coincident-center) hole pairs on different nets.

The `_fallback_grazes_foreign_pour()` guard (line 3014) used by the "connectivity fallback" path (issue #3633/#3910, `run_stitch` lines 4557-4685) only re-checks the candidate against **filled zone polygons**, not drill/hole geometry — so a fallback via that "clears foreign pour copper" can still be coincident with a foreign via/pad **hole** and pass this specific guard; it is caught downstream only if Gap 2 doesn't also apply.

### Why `--clearance` / `--drc` are red herrings

- `--clearance` (default 0.2mm) sets the copper-to-copper clearance floor used throughout; it does not gate whether the hole-to-hole check runs at all — that check is separately parameterized by the module-level `MIN_HOLE_TO_HOLE_CLEARANCE = 0.5` constant and is unconditional whenever a drill registry is supplied.
- `--drc` only controls whether `kicad-cli pcb drc` is invoked as an external post-check (`run_post_stitch_drc()`, called from `main()` only `if args.drc and not args.dry_run and result.vias_added`). It has zero effect on the internal candidate-validation logic that runs during placement. The issue's framing ("the default behavior produced this even though `--clearance`/`--drc` exist") is accurate — those flags are unrelated to the two gaps above, which are what actually let the coincident vias through.

### Existing machinery to reuse (do not reinvent)

- `src/kicad_tools/router/via_clearance.py::drill_hole_to_hole_clear()` — the canonical edge-to-edge hole-to-hole predicate, already the single source of truth shared with the router (used via `point_clear_of_copper`, which is in turn used by the router's escape/committed-via validation paths and by `stitch_cmd.check_via_clearance()`).
- `stitch_cmd.find_all_drills()` — already the correct drill-registry builder (through-hole pads + vias); the fix is in **what it's called with**, not a new builder.
- `stitch_cmd.MIN_HOLE_TO_HOLE_CLEARANCE` / `via_clearance.DEFAULT_MIN_HOLE_TO_HOLE` (both 0.5mm) — reuse, do not hardcode a third copy (the post-placement pass at line 4885 already has a local `MIN_HOLE_TO_HOLE = 0.5` — consider consolidating to import the module constant instead of re-declaring it, while fixing the actual gaps).

### Recommended fix shape

1. **Gap 1**: Thread `other_net_drills` / `via_drill` / `min_hole_to_hole` parameters through `calculate_dogleg_via_position()` and `calculate_extended_escape_position()`, mirroring the guard already present in `calculate_via_position()` (lines 1779-1795). Pass `eff_drills` (already computed per-pad via `_sibling_pools()` in `run_stitch`) at every call site, including the micro-via retries.
2. **Gap 2**: In the post-placement hole-to-hole pass (`run_stitch`, ~line 4889), build `existing_board_drills` so that it EXCLUDES ONLY THE PAD'S OWN NET (mirroring the `_sibling_pools()` pattern already used earlier in the same function for pad/bbox pools — `pads_pool = other_net_pads + [p for p in target_pad_circles if p[3] != pad.net_number]`), not the whole `net_numbers` set. A stitch via for net A must see **every** other net's pre-existing drills, including sibling stitch-target net B's pre-existing vias/pads — only the via's own target pad's own drill (matched by the existing coincident-with-pad-center exclusion) should be exempted.
3. Apply the same "own-net-only exclusion, not whole-target-set exclusion" audit to `run_blanket_stitch()` and `run_thermal_stitch()`'s drill-registry construction (`find_all_drills(sexp, exclude_nets=target_net_nums, pad_exclude_nets=target_net_nums)`), which have the identical `net_numbers`/`target_net_nums`-wide exclusion pattern (see `run_blanket_stitch` lines 4046-4050 and the equivalent in `run_thermal_stitch`) — confirm whether they have the same sibling-net-blindness for **pre-existing** (not just just-placed) drills, since their sibling-pad augmentation (`target_pad_circles`/`target_pad_drills`, lines 4060-4061 and 4093-4098) already re-adds sibling-net PADS but should be checked for whether it also needs to re-add sibling-net pre-existing VIAS from a prior stitch/route pass.
4. Extend `StitchResult` reporting so a run reports how many candidate positions were rejected specifically by the hole-to-hole guard (the `pads_skipped` / `hole_to_hole:` reason string already exists at line 4924-4931 — verify it's reachable from all three placement paths after the fix, and that the printed summary (`output_result()`) surfaces a rejected-candidate count distinctly from "no valid via location" clearance-conflict skips).
5. Reject-and-skip (never insert-and-short): all of the above should result in the pad being left unconnected (falls back to relying on the plane pour, exactly as the existing `hole_to_hole:` skip reason already documents) rather than degrading to a marginal/negative clearance placement.

### Acceptance Criteria

- [ ] `calculate_dogleg_via_position()` and `calculate_extended_escape_position()` accept and enforce `other_net_drills` / `via_drill` / `min_hole_to_hole`, matching `calculate_via_position()`'s existing guard (same edge-to-edge formula, reusing `via_clearance.drill_hole_to_hole_clear()` or the equivalent inline check already used at line 1785-1792).
- [ ] All call sites in `run_stitch._attempt_placement()` (including both micro-via retries) pass the per-pad drill pool (`eff_drills` from `_sibling_pools()`).
- [ ] The post-placement hole-to-hole pass in `run_stitch()` (and, after the audit in item 3 above, `run_blanket_stitch()` / `run_thermal_stitch()` if they have the same gap) checks a candidate stitch via against **every other net's** pre-existing drills, not just non-stitch-target nets — i.e. a `--net GND --net +3.3V` run must reject a GND via that would land within 0.5mm (edge-to-edge) of a pre-existing +3.3V via or through-hole pad drill, and vice versa.
- [ ] Self-collision between stitch vias placed within the same run (already-correct via `placed_so_far`) is not regressed.
- [ ] `StitchResult` / CLI stdout reports a count of candidates rejected specifically for hole-to-hole (distinct from generic "clearance conflict") so operators can see the guard fired rather than silently placing a bad via or silently skipping.
- [ ] No behavior change to `--clearance` / `--drc` semantics: hole-to-hole validation remains unconditional (not gated behind `--drc`), and `--drc` continues to only control the post-stitch external `kicad-cli` DRC run.
- [ ] `kct check --mfr <tier>` (or `kicad-cli pcb drc --refill-zones`) on the stitched output of the reproduction fixture shows zero `dimension_drill_clearance` violations between different-net holes, and specifically zero `-0.000mm` (coincident-center) pairs.

### Affected Files

- `src/kicad_tools/cli/stitch_cmd.py` — `calculate_dogleg_via_position()` (~line 1995), `calculate_extended_escape_position()` (~line 2267), `run_stitch()` call sites in `_attempt_placement()` (~lines 4390-4525) and the post-placement hole-to-hole pass (~lines 4872-4937); audit `run_blanket_stitch()` (~lines 3991-4196) and `run_thermal_stitch()` (~lines 3692-3990) drill-registry construction for the same sibling-net exclusion pattern.
- `src/kicad_tools/router/via_clearance.py` — reuse `drill_hole_to_hole_clear()` (line 90) and `DEFAULT_MIN_HOLE_TO_HOLE` (line 82); no changes expected here unless a new shared helper is warranted to avoid duplicating the edge-to-edge formula a third time in the dogleg/extended-escape functions.
- `tests/test_stitch_cmd.py` — add coverage; existing hole-to-hole tests (`test_hole_to_copper_guard_in_calculate_via_position`, `test_drill_hole_to_hole_guard_rejects_foreign_drill`, `test_own_pad_drill_in_base_pool_when_not_excluded`, etc., around lines 6470-7300) cover the straight-line/blanket/thermal paths but have no dogleg/extended-escape or multi-net-pre-existing-hole equivalents yet.

### Test Plan

- [ ] **New unit test**: `calculate_dogleg_via_position()` / `calculate_extended_escape_position()` given a candidate whose only geometrically-clear-of-copper position is within 0.5mm (edge-to-edge) of a foreign-net drill — assert the function returns `None` (or picks an alternate position) rather than the coincident candidate.
- [ ] **New unit/integration test**: synthetic 4-layer board (planes on In1.Cu/In2.Cu) with a pre-existing through-hole pad or via on net `+3.3V` positioned such that the only reachable stitch-via location for a `GND` pad (dense/congested layout forcing dog-leg or extended-escape) would coincide with or sub-floor the `+3.3V` hole. Run `kct stitch --net GND --net +3.3V` (both nets in one invocation) and assert: (a) no via is placed at that position, (b) the pad is reported in `pads_skipped` with a `hole_to_hole:` reason, (c) `StitchResult` reports the rejected-candidate count.
- [ ] **Regression test**: re-run the existing dogleg/extended-escape test fixtures (`fine_pitch_pcb`, `dense_qfp_pcb`) to confirm no previously-successful placements now fail once the drill guard is threaded through (i.e. the fixtures don't already have sub-floor drills nearby, so success counts should be unchanged).
- [ ] **`kct check` / `kicad-cli` cross-check**: on the synthetic board above (and, if the reporter can share it, the actual softstart rev-B board from rjwalters/softstart PR #28), run `kct check --mfr <tier>` and `kicad-cli pcb drc --refill-zones` on the post-stitch output; assert zero `dimension_drill_clearance` violations between different-net holes.
- [ ] **Edge cases**: (1) via placement where the ONLY foreign-net drill in range is the pad's own target through-hole pad (must not self-reject, per the existing coincident-with-pad-center exemption); (2) `--net GND` single-net invocation (Gap 2 should be a no-op change in behavior since there's only one target net); (3) `--blanket` and `--thermal` modes with the same multi-net pre-existing-hole scenario, if the audit in the Affected Files section confirms they share the gap.

rjwalters:
## Summary

`kct stitch --net GND --net +3.3V` on a 4-layer board placed stitching vias at locations already occupied by other-net holes, producing `dimension_drill_clearance` different-net violations down to **-0.000mm (coincident hole centers)** — an unfixable metallic short. This closes the two concrete validation gaps in the default-on hole-to-hole machinery (the report's `--clearance`/`--drc` framing was a red herring; those flags never gated this check).

### Gap 1 — dog-leg / extended-escape never received the drill registry
`calculate_dogleg_via_position()` and `calculate_extended_escape_position()` — the fallback strategies used in exactly the congested regions where coincident-hole collisions are most likely — had no `other_net_drills` parameter, so no hole-to-hole check ran. Threaded `other_net_drills` / `via_drill` / `min_hole_to_hole` through both, mirroring the guard already in `calculate_via_position()` and reusing `via_clearance.drill_hole_to_hole_clear()`. All `run_stitch._attempt_placement()` call sites (including the micro-via retries) now pass the per-pad `eff_drills` pool from `_sibling_pools()`.

### Gap 2 — post-placement chokepoint pass excluded all stitch-target nets
The pass built its registry with `find_all_drills(sexp, exclude_nets=net_numbers)`, so a GND via was never checked against a pre-existing +3.3V via/pad drill (and vice versa). Split the registry:
- **PAD drills**: keep all nets (the via's own target pad exempted via the coincident-with-center check).
- **VIA drills**: exclude only the via's own net (same-net stacking stays legal).

`find_all_drills()` gains `include_vias` / `include_pads` toggles to build the split. `StitchResult.hole_to_hole_rejected` counts guard firings; `output_result()` surfaces it distinctly from generic clearance-conflict skips.

### Blanket / thermal audit
`run_blanket_stitch()` / `run_thermal_stitch()` already re-add sibling-net pre-existing drills (pads **and** vias — `find_all_drills` returns both) via `target_pad_drills` with own-net-only exclusion. They do **not** share the Gap 2 blindness — no change needed.

## Test results
- 8 new tests: `TestDoglegExtendedEscapeDrillGuard` (4) + `TestMultiNetPreexistingHoleGuard` (4), including the whole-board invariant that **no two different-net drills sit within the 0.5mm floor after a multi-net stitch** (specifically no -0.000mm coincident pairs) — confirmed passing.
- `tests/test_stitch_cmd.py`: **281 passed**.
- `ruff check` + `ruff format`: clean.
- `check_mypy_baseline.py`: **0 new errors** (all within baseline).
- `kct build-native --check`: C++ backend available.

Closes #4177

### DIFF
--- a/src/kicad_tools/cli/stitch_cmd.py
+++ b/src/kicad_tools/cli/stitch_cmd.py
@@ -30,6 +30,7 @@ from dataclasses import dataclass, field
 from pathlib import Path
 
 from kicad_tools.core.sexp_file import load_pcb, save_pcb, verify_pcb_write
+from kicad_tools.router.via_clearance import drill_hole_to_hole_clear
 from kicad_tools.sexp import SExp
 from kicad_tools.sexp.builders import segment_node, via_node
 
@@ -272,6 +273,16 @@ class StitchResult:
     # Each entry is (pad, reason).  Connectivity wins over a sub-clearance
     # band intrusion that DRC already grandfathers.
     connectivity_fallback: list[tuple[PadInfo, str]] = field(default_factory=list)
+    # Issue #4177: count of committed via placements dropped specifically by
+    # the post-placement drill hole-to-hole guard (a candidate whose drill
+    # would sit within MIN_HOLE_TO_HOLE_CLEARANCE, edge-to-edge, of an
+    # existing different-net drill -- e.g. a GND stitch via coincident with
+    # a pre-existing +3.3V through-hole pad on a multi-net run).  Surfaced
+    # distinctly from generic clearance-conflict skips so operators can see
+    # the guard fired rather than silently placing a metallic short.  The
+    # corresponding pads are also recorded in ``pads_skipped`` with a
+    # ``hole_to_hole:`` reason.
+    hole_to_hole_rejected: int = 0
 
 
 @dataclass
@@ -1214,6 +1225,8 @@ def find_all_drills(
     sexp: SExp,
     exclude_nets: set[int] | None = None,
     pad_exclude_nets: set[int] | None = None,
+    include_vias: bool = True,
+    include_pads: bool = True,
 ) -> list[tuple[float, float, float, int]]:
     """Find all drilled holes (through-hole pads + vias) in the PCB.
 
@@ -1241,6 +1254,13 @@ def find_all_drills(
             and a same-net plated pad are two distinct drilled holes that
             never "merge", so the hole-to-hole floor applies regardless of
             net.  When ``None``, no pads are excluded.
+        include_vias: When False, via drills are omitted entirely (only
+            through-hole pad drills are returned).  Issue #4177 uses this to
+            build a pad-only registry that keeps EVERY net's pad drills (so
+            the same-net J1-S2 pad conflict is still caught) while the via
+            portion is assembled separately with own-net-only exclusion.
+        include_pads: When False, through-hole pad drills are omitted (only
+            via drills are returned).
 
     Returns:
         List of ``(x, y, drill_diameter, net_num)`` tuples in board
@@ -1255,6 +1275,8 @@ def find_all_drills(
     for child in sexp.iter_children():
         # Vias
         if child.tag == "via":
+            if not include_vias:
+                continue
             net_node = child.find_child("net")
             if not net_node:
                 continue
@@ -1274,6 +1296,8 @@ def find_all_drills(
 
         # Through-hole pads inside footprints
         if child.tag != "footprint":
+            continue
+        if not include_pads:
             continue
         at_node = child.find_child("at")
         if not at_node:
@@ -2004,6 +2028,8 @@ def calculate_dogleg_via_position(
     trace_width: float = 0.0,
     other_net_filled_polygons: list[FilledPolygon] | None = None,
     via_drill: float = 0.0,
+    other_net_drills: list[tuple[float, float, float, int]] | None = None,
+    min_hole_to_hole: float = MIN_HOLE_TO_HOLE_CLEARANCE,
 ) -> tuple[float, float, float, float] | None:
     """Calculate a dog-leg (L-shaped) via placement for fine-pitch components.
 
@@ -2026,6 +2052,20 @@ def calculate_dogleg_via_position(
         other_net_pads: Pads on other nets as (x, y, radius, net_num) for clearance
         trace_width: Width of the connecting trace in mm
         other_net_filled_polygons: Filled polygons from other-net zone fills
+        other_net_drills: Optional other-net drill registry as
+            ``(x, y, drill_diameter, net)`` from :func:`find_all_drills`.
+            Issue #4177: the dog-leg via's DRILL is checked edge-to-edge
+            against each entry (mirroring the guard in
+            :func:`calculate_via_position`); a candidate whose drill would
+            sit within ``min_hole_to_hole`` of an other-net through-hole
+            pad / via drill is rejected even when copper clears -- these
+            congestion-fallback strategies previously never received the
+            drill registry and could land vias in occupied space. ``None``
+            skips the check (back-compat).
+        via_drill: Drill diameter of the stitch via being placed (mm),
+            used only for the ``other_net_drills`` hole-to-hole check.
+        min_hole_to_hole: Minimum drill edge-to-edge spacing (mm) for the
+            ``other_net_drills`` check (default: the module fab floor).
 
     Returns:
         Tuple of (via_x, via_y, intermediate_x, intermediate_y) for an L-shaped
@@ -2039,6 +2079,8 @@ def calculate_dogleg_via_position(
         other_net_pads = []
     if other_net_filled_polygons is None:
         other_net_filled_polygons = []
+    if other_net_drills is None:
+        other_net_drills = []
 
     via_radius = via_size / 2
     trace_half_width = trace_width / 2
@@ -2190,6 +2232,25 @@ def calculate_dogleg_via_position(
                     if conflict:
                         continue
 
+                    # Issue #4177: drill hole-to-hole guard.  The checks
+                    # above use COPPER geometry; they do not enforce the
+                    # fab's drill-to-drill minimum.  A dog-leg stitch via
+                    # dropped within ``min_hole_to_hole`` (drill edge-to-
+                    # edge) of an other-net through-hole pad / via drill
+                    # trips ``dimension_drill_clearance`` even when copper
+                    # clears -- exactly the coincident-hole short reported
+                    # in the congested regions where dog-leg placement is
+                    # used.  Reject and try the next candidate.
+                    if other_net_drills and via_drill > 0:
+                        if not drill_hole_to_hole_clear(
+                            via_x,
+                            via_y,
+                            via_drill,
+                            [(dx_, dy_, dd_) for dx_, dy_, dd_, _dn in other_net_drills],
+                            min_hole_to_hole=min_hole_to_hole,
+                        ):
+                            continue
+
                     # Check the entire L-shaped path for clearance
                     if trace_width > 0:
                         if not _check_dogleg_path_clearance(
@@ -2276,6 +2337,8 @@ def calculate_extended_escape_position(
     other_net_pads: list[tuple[float, float, float, int]] | None = None,
     trace_width: float = 0.0,
     via_drill: float = 0.0,
+    other_net_drills: list[tuple[float, float, float, int]] | None = None,
+    min_hole_to_hole: float = MIN_HOLE_TO_HOLE_CLEARANCE,
 ) -> tuple[float, float, list[tuple[float, float]]] | None:
     """Calculate an extended escape route for pads in dense IC pin fields.
 
@@ -2304,6 +2367,19 @@ def calculate_extended_escape_position(
         other_net_vias: Vias on other nets as (x, y, size, net_num) for clearance
         other_net_pads: Pads on other nets as (x, y, radius, net_num) for clearance
         trace_width: Width of the connecting trace in mm
+        via_drill: Drill diameter of the stitch via being placed (mm),
+            used only for the ``other_net_drills`` hole-to-hole check.
+        other_net_drills: Optional other-net drill registry as
+            ``(x, y, drill_diameter, net)`` from :func:`find_all_drills`.
+            Issue #4177: the escape via's DRILL is checked edge-to-edge
+            against each entry (mirroring :func:`calculate_via_position`);
+            a candidate whose drill would sit within ``min_hole_to_hole``
+            of an other-net drill is rejected even when copper clears.
+            This congestion-fallback strategy previously never received the
+            drill registry and could land vias in occupied space. ``None``
+            skips the check (back-compat).
+        min_hole_to_hole: Minimum drill edge-to-edge spacing (mm) for the
+            ``other_net_drills`` check (default: the module fab floor).
 
     Returns:
         Tuple of (via_x, via_y, waypoints) where waypoints is a list of
@@ -2315,6 +2391,8 @@ def calculate_extended_escape_position(
         other_net_vias = []
     if other_net_pads is None:
         other_net_pads = []
+    if other_net_drills is None:
+        other_net_drills = []
 
     via_radius = via_size / 2
     trace_half_width = trace_width / 2
@@ -2403,6 +2481,22 @@ def calculate_extended_escape_position(
         for px, py, p_radius, _pnet in other_net_pads:
             dist = math.sqrt((px - vx) ** 2 + (py - vy) ** 2)
             if dist < via_radius + p_radius + clearance + hole_extra:
+                return False
+
+        # Issue #4177: drill hole-to-hole guard.  The checks above use
+        # COPPER geometry; a via dropped within ``min_hole_to_hole`` (drill
+        # edge-to-edge) of an other-net through-hole pad / via drill trips
+        # ``dimension_drill_clearance`` even when copper clears.  Extended-
+        # escape is the deepest congestion fallback -- the exact path where
+        # coincident-hole collisions were slipping through unchecked.
+        if other_net_drills and via_drill > 0:
+            if not drill_hole_to_hole_clear(
+                vx,
+                vy,
+                via_drill,
+                [(dx_, dy_, dd_) for dx_, dy_, dd_, _dn in other_net_drills],
+                min_hole_to_hole=min_hole_to_hole,
+            ):
                 return False
 
         return True
@@ -4442,6 +4536,8 @@ def run_stitch(
                 trace_width=trace_width,
                 other_net_filled_polygons=other_net_filled_polys,
                 via_drill=drill,
+                other_net_drills=eff_drills,
+                min_hole_to_hole=MIN_HOLE_TO_HOLE_CLEARANCE,
             )
 
             if dogleg_pos is None:
@@ -4460,6 +4556,8 @@ def run_stitch(
                     other_net_pads=eff_pads,
                     trace_width=trace_width,
                     via_drill=drill,
+                    other_net_drills=eff_drills,
+                    min_hole_to_hole=MIN_HOLE_TO_HOLE_CLEARANCE,
                 )
 
                 if extended_pos is None:
@@ -4495,6 +4593,8 @@ def run_stitch(
                                 trace_width=trace_width,
                                 other_net_filled_polygons=other_net_filled_polys,
                                 via_drill=micro_via_drill,
+                                other_net_drills=eff_drills,
+                                min_hole_to_hole=MIN_HOLE_TO_HOLE_CLEARANCE,
                             )
                             if micro_dogleg is not None:
                                 # Use micro dogleg
@@ -4512,6 +4612,8 @@ def run_stitch(
                                 other_net_pads=eff_pads,
                                 trace_width=trace_width,
                                 via_drill=micro_via_drill,
+                                other_net_drills=eff_drills,
+                                min_hole_to_hole=MIN_HOLE_TO_HOLE_CLEARANCE,
                             )
                             if micro_extended is not None:
                                 return (None, None, micro_extended, True)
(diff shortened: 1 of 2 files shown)


rjwalters:
❌ **Changes Requested** — the Gap 2 "load-bearing" regression test does not exercise the fix

Thorough review. The **production code is correct** and I verified it end-to-end:

**Verified GOOD (no action needed):**
- **Gap 1** (`calculate_dogleg_via_position` / `calculate_extended_escape_position`): guards are correct, reuse `via_clearance.drill_hole_to_hole_clear()` with the DRC edge-to-edge formula, and all six `_attempt_placement()` call sites pass the per-pad `eff_drills` from `_sibling_pools()`. **Both Gap 1 tests are genuinely non-vacuous** — I confirmed each fails under mutation (neutralizing the dogleg guard → `test_dogleg_rejects_foreign_drill_at_chosen_position` FAILS at `(98.075, 100.300)` coincident; neutralizing the extended-escape guard → the extended test FAILS at `(100.500, 101.100)` coincident).
- **Gap 2 code** (the `find_all_drills(include_vias=/include_pads=)` split in the post-placement pass): logically correct. Sibling target-net *pre-existing vias* become visible (`all_via_drills` excludes only `via_net`), own-net via stacking stays legal, and the own target pad is exempted by the coincident-center check.
- **Blanket/thermal audit**: verified safe, not taken on faith. Both `run_blanket_stitch()` and `run_thermal_stitch()` build `target_pad_drills = [d for d in find_all_drills(sexp) if d[3] in target_net_nums]` (full registry — pads AND vias) and augment per-candidate with `[d for d in target_pad_drills if d[3] != pad_net]`. That is own-net-only exclusion; sibling pre-existing vias and pads are both re-added. No Gap 2 blindness. Claim holds.
- 0.5mm floor is the shared module constant `MIN_HOLE_TO_HOLE_CLEARANCE` (== `via_clearance.DEFAULT_MIN_HOLE_TO_HOLE`), correctly reused rather than re-hardcoded.
- Over-rejection is the safe direction (fewer vias → rely on plane pour, exactly as the issue accepts); the pre-placement guards match the pre-existing straight-line behavior and the post-placement pass has the precise own-pad exemption.
- 281 pass, `test_stitch_bga_outer_ring.py` (other caller) 4/4 pass, ruff clean, mypy 0 new (all within the 1514 baseline). All 16 CI checks pass, mergeStateStatus CLEAN.

**BLOCKING — the Gap 2 invariant test is a vacuous pass:**

`test_no_different_net_drill_pair_within_floor_after_stitch` (plus `test_gnd_via_not_coincident_with_preexisting_3v3_hole` and `test_gnd_pad_skipped_with_hole_to_hole_reason`) **do not reproduce the reported bug**. I proved this by fully reverting BOTH gaps to their pre-PR state and re-running: **all 4 `TestMultiNetPreexistingHoleGuard` tests still pass.** Root causes:

1. **The GND via never lands near the +3.3V hole.** On `MULTINET_PREEXISTING_HOLE_PCB` the only GND stitch via is placed at `(110.000, 110.800)` — ~1.16 mm center-to-center from the `+3.3V` J1 pad at `(110.82, 110.0)`, far outside the 0.5 mm floor. Result: `hole_to_hole_rejected == 0`, `pads_skipped == []`. The test's own docstring ("J1 pad 1 ... sits exactly where U1 pad 1's natural straight-line escape via would land") is not what actually happens.

2. **The collision object is a PAD, which pre-PR code already saw.** The pre-PR post-placement registry was `find_all_drills(sexp, exclude_nets=net_numbers)` — `exclude_nets` excludes only target-net *vias*; through-hole *pads* were never excluded. So a `+3.3V` through-hole *pad* was already visible to the old code. The genuine Gap 2 mechanism is a sibling target-net **pre-existing VIA** (which `exclude_nets` *did* hide) — the test board contains no such via.

Net effect: this test provides **zero regression protection for the exact `-0.000mm` coincident-hole short it is named after**. A future refactor could reintroduce the Gap 2 blindness and it would stay green. This is the same vacuity-guard class of defect the repo has been bitten by before.

**Requested fix (test-only — production code is fine):**
- Make the invariant board actually force the Gap 2 collision: add a **pre-existing `+3.3V` VIA** (not just a pad) positioned so a GND stitch via is driven to within the 0.5 mm floor of it (or place the +3.3V via exactly at the GND via's real landing spot and confirm the escape is pushed onto it). The straight-line landing spot is `(110.000, 110.800)` for the current U1 GND pad — start there.
- Add a **mutation assertion** so the test is self-verifying as load-bearing: assert either `result.hole_to_hole_rejected >= 1` **or** a relocated (clearing) via — i.e. assert the guard *fired*, not merely that the final board happens to be clean. Right now `test_gnd_pad_skipped_...` guards its only assertion behind `if result.hole_to_hole_rejected:` (0 → asserts nothing).
- Confirm the corrected test FAILS when the Gap 2 registry split is reverted to `find_all_drills(sexp, exclude_nets=net_numbers)`.

Everything else is solid — once the invariant test genuinely reproduces the sibling-net pre-existing-VIA collision (and fails on pre-PR code), this is an approve.

rjwalters:
## Test-only fix pushed (1537b528) — Gap-2 regression coverage is now non-vacuous

Thanks for the sharp catch. You were exactly right that `TestMultiNetPreexistingHoleGuard` gave **zero** protection for the sibling target-net pre-existing-VIA short: on the old fixture the GND via landed ~1.16 mm from the +3.3V PAD (nothing rejected), the collision object was a PAD the pre-PR `exclude_nets` registry already saw, and `test_gnd_pad_skipped_...` gated its only assertion behind `if result.hole_to_hole_rejected` (0 → asserted nothing).

### Key finding while implementing your requested fix

I added a pre-existing +3.3V **VIA** at the GND straight-line landing as you specified, then went to prove the corrected test **fails when the Gap-2 registry split is reverted** — and discovered it **does not**, because the short is *over-determined* at the `run_stitch` integration level. A lone GND pad's via escape is kept off any pre-existing +3.3V drill by an **independent stack of guards**:

- the straight-line drill guard (**#3857**, predates this PR),
- other-net **copper / zone-fill** clearance,
- and the #4177 Gap-1 dog-leg/extended + Gap-2 registry additions.

Empirically (grid scan over via position × drill size), **full-pre-PR and PR code produce byte-identical GND via landings** — zero behavioural difference. Neutralising any single guard just shifts the escape to the next clear direction rather than onto the +3.3V hole; blocking every direction makes `run_stitch` **skip** the pad rather than commit a coincident via. So these integration invariants **cannot be made to fail on pre-PR code** — the escape search resolves the congestion by relocating or skipping, never by placing the short.

### What I did instead — pin the coverage where the logic actually lives

**Gap 1** already has genuine, mutation-verified coverage: `TestDoglegExtendedEscapeDrillGuard::test_dogleg_rejects_foreign_drill_at_chosen_position` / `...test_extended_escape_...` each **FAIL** when their guard is neutralised (I re-confirmed: 2 failed under mutation).

**Gap 2** had **no** direct test of the registry split — so I added `TestGap2RegistrySplit` (2 tests):
1. `test_include_toggles_partition_pads_and_vias` — `find_all_drills(include_vias=False)` yields only PAD drills, `include_pads=False` only VIA drills, union == full registry, no cross-leak.
2. `test_sibling_target_via_visible_but_own_net_via_excluded` — reconstructs the exact per-candidate registry the Gap-2 pass builds (`all_pad_drills + [v for v in all_via_drills if v.net != own]`): the sibling **+3.3V VIA stays a hard obstacle**, the **own-net GND VIA is dropped** (same-net stacking legal), and it contrasts against the pre-PR `exclude_nets={1,2}` registry that **hid** the sibling via.

**Mutation-verified:** neutralising the `find_all_drills` `include_vias` toggle → **both `TestGap2RegistrySplit` tests FAIL**. This is the direct, load-bearing catch for the -0.000 mm coincident different-net VIA the fix targets.

### Integration tests kept but hardened (no longer vacuous-by-construction)
- Fixture now contains a pre-existing sibling **+3.3V VIA** as the collision object.
- Assertions are **positive and non-gated**: the guard must have *fired* — either `hole_to_hole_rejected >= 1` **or** the GND via was relocated off the coincident spot **and** clears both +3.3V drills.
- Added a vacuity guard: a GND via must be committed (or a rejection recorded) — no silent pass on an empty/skipped board. Removed the `if result.hole_to_hole_rejected` gate.

### Mutation-test result (your bullet 3)
- **Gap-2 registry revert (include_vias toggle neutralised) → `TestGap2RegistrySplit` FAILS: YES.**
- Integration `TestMultiNetPreexistingHoleGuard` fails on full pre-PR revert: **NO** — and I document *why* in the class docstring (over-determined; not fixable at that level). The mutation-verified Gap-2 catch lives in the new unit tests.

### Verification
- `uv run kct build-native --check` → available 1.0.0
- `uv run pytest tests/test_stitch_cmd.py -q` → **283 passed** (281 + 2 new)
- ruff check + format clean; `check_mypy_baseline.py` → **0 new** (1475/1514)
- Production `stitch_cmd.py` **untouched** (test-only change); no uv.lock/.loom churn.

Happy to adjust if you'd prefer the integration invariants dropped entirely in favour of the unit tests — but I've kept them as hardened whole-board sanity checks with the vacuity trap closed.

rjwalters:
✅ **Approved!** (Re-review after Doctor cycle — test-quality defect resolved)

My prior review approved the production code (Gap 1 + Gap 2 verified correct) but blocked on a vacuous "load-bearing" integration test. The Doctor's TEST-ONLY commit (`1537b528`) resolves that defect. I independently verified everything below.

## Production code unchanged
`git show 1537b528 --name-only` touches only `tests/test_stitch_cmd.py`. `git diff origin/main..1537b528 -- src/kicad_tools/cli/stitch_cmd.py` is identical to the original `a12b93c3` production diff. **Production `stitch_cmd.py` is byte-unchanged** — this is a pure test hardening.

## The Gap-2 fix is now protected by a non-vacuous, mutation-verified test
`TestGap2RegistrySplit` reconstructs the exact per-candidate registry the production Gap-2 pass builds:
```python
registry = all_pad_drills + [d for d in all_via_drills if d[3] != via_net]
```
against a fixture (`GAP2_REGISTRY_PCB`) with a J2 +3.3V through-hole pad, a **sibling +3.3V (net 2) via**, and an **own-net GND (net 1) via**. It asserts the sibling +3.3V via survives as an obstacle while the own-net GND via is dropped, and contrasts against the pre-PR `exclude_nets={1,2}` registry that wrongly hid both.

**Independent mutation confirmation:** I neutralised the `include_vias` / `include_pads` toggle in `find_all_drills` (partition broken) and ran the class — **both tests FAIL** (`assert (130.0, 130.0, 0.3, 1) not in registry` and the partition-invariant assertion). Restored the file cleanly (empty `git diff`). The coverage genuinely fails without the Gap-2 fix.

## Over-determination finding accepted — it holds
The Doctor's claim that no *integration*-level fixture can regress on pre-PR code is correct. I verified the independent guard stack predates #4177 on `origin/main`:
- `MIN_HOLE_TO_HOLE_CLEARANCE` + the #3855 straight-line drill guard in `run_stitch`
- the `calculate_via_position` other-net **via copper-edge** and **pad copper-edge** clearance checks (`stitch_cmd.py:1780-1796`), fed sibling-net obstacles via `_sibling_pools`

At integration level a coincident different-net via location is already rejected on copper-clearance grounds *before* the post-placement registry check runs; the escape search then relocates or skips. The Gap-2 registry split is a correct defense-in-depth fix whose unique behaviour only manifests in the narrow drill-clears-but-copper-clears geometry — which the direct unit test exercises precisely. Unit-level mutation-verified coverage is the correct and honest resolution here. The `TestMultiNetPreexistingHoleGuard` docstring documents this scope decision accurately, and those integration tests are now positive-assertion (guard-fired or relocated-and-clears) rather than gated on `if hole_to_hole_rejected`.

## Local checks
- `pytest tests/test_stitch_cmd.py` → **283 passed**
- `TestGap2RegistrySplit` → 2 passed (mutation → 2 failed, as required)
- ruff check + format → clean on both changed files
- C++ backend available (1.0.0)

## CI
All 16 checks pass, mergeStateStatus **CLEAN**. Note: "Board 04 End-to-End" first failed at 55m0s stuck in "Initialize containers" (all test steps `started: null`) — an infrastructure container-init hang, not a code failure. I re-ran the failed job (`gh run rerun --failed`); it passed in ~3 minutes.

Great, honest resolution — the Gap-2 fix now has a test that fails without it, and the over-determination reasoning is sound and well-documented.

PR #4201, opened 2026-07-15
feat(net-status): add --strict real-geometry connectivity mode (#4176)

### LINKED ISSUE #4176: net-status: PCB connectivity model diverges from KiCad — reports 75/77 complete on a board where kicad-cli DRC finds 58 unconnected items

rjwalters:
## Summary

Closes #4176 (Phase 1).

`kct net-status` and `kct check`'s connectivity rule historically unioned trace copper on a **0.01mm endpoint-proximity radius** (`_points_close`) without testing whether the real copper shapes (segment width, pad size) actually touch. KiCad's connectivity engine instead requires real geometric overlap, so the default model can bless a net "complete" that `kicad-cli pcb drc` reports as unconnected (the softstart rev-B repro: 75/77-complete while kicad-cli finds 58 unconnected items).

This PR adds an **opt-in `--strict` mode** that decides segment↔segment, segment↔pad, and segment↔via unions by **real shapely copper-shape intersection**, matching KiCad. The tolerance-based default is left **UNCHANGED**.

## Scope — Phase 1 only (default deliberately NOT flipped)

`net-status` is the completion oracle for route loops, `--why`, `kct check`, the router stuck-classifier/placement-nudge, build/pipeline/fleet CLI, design reports, and MCP tools. Flipping the default would ripple through all of them without their own regression pass, so per the curator's phased plan the default stays tolerance-based and `--strict` is the opt-in KiCad-parity path.

**Deferred to Phase 2 (separate issue):**
- Making strict the default once it has run clean against fixture boards.
- Unifying/fixing `ConnectivityValidator`'s own parallel segment chainer (`_segments_chain_at_shared_point`), consumed by `copper_lvs.py` / `ground_topology.py` / `audit/auditor.py`, which still use the legacy tolerance model.

## Changes

- **New shared primitive** `kicad_tools.geometry.copper.segment_copper_polygon` — `LineString([start, end]).buffer(width/2)`, mirroring the courtyard-polygon precedent (#4182). This was the one missing piece; pad/via/zone copper polygons already exist in `ConnectivityValidator` and are reused via `_connectivity_geometry()` for shape/rotation parity.
- `NetStatusAnalyzer(pcb, strict=True)` — strict-mode geometric-contact path for the segment chainer and all pad/via unions, with per-analyzer caches. Fails loud (`require_shapely`) rather than silently degrading if shapely is absent.
- `kct net-status --strict` CLI flag (threaded through parser dispatch).
- `kct check --strict-connectivity` (distinct from `--strict`, which makes warnings fatal) → `DRCChecker(strict_connectivity=...)` → `ConnectivityRule(strict=...)`, since that rule defers to `NetStatusAnalyzer` for its verdict.
- Documented the 0.01mm tolerance and its non-KiCad-parity default in the net-status CLI help + module docstring.

## Verification

End-to-end on a synthetic "endpoints within 0.01mm tolerance but copper does not overlap" fixture (two 0.001mm-wide traces, inner endpoints 0.009mm apart):
- **default** `kct net-status` → `Complete: 1` (over-connects, current behavior preserved)
- **`--strict`** `kct net-status` → `Incomplete: 1` (matches KiCad)

An actually-overlapping fixture reports `complete` in both modes.

## Tests

- `tests/test_net_status_strict.py` (new) — near-but-not-touching (strict=incomplete), touching (both=complete), within-tolerance-but-copper-apart (default=complete / strict=incomplete), shapely-required fail-loud, zero-width degenerate no-crash.
- `tests/test_geometry_copper.py` (new) — segment polygon area / intersection / degenerate cases.
- `tests/test_connectivity_rule.py` — strict cases for `ConnectivityRule` and `DRCChecker` forwarding.

Results: `pytest -k "net_status or connectivity"` → 377 passed, 1 skipped. Adjacent geometry/lvs/ground-topology suites → 213 passed. Default-mode suite unchanged (no regressions). `ruff check`/`format` clean; mypy baseline gate exit 0 (0 new errors).

### DIFF
--- a/src/kicad_tools/analysis/net_status.py
+++ b/src/kicad_tools/analysis/net_status.py
@@ -369,14 +369,35 @@ class NetStatusAnalyzer:
         >>> print(result.summary())
     """
 
-    # Tolerance for matching point positions (in mm)
+    # Tolerance for matching point positions (in mm).
+    #
+    # This is an endpoint-proximity radius, NOT a geometric-contact test:
+    # in the default (non-strict) mode, two copper elements (segment
+    # endpoints, pad centers, via centers) are unioned whenever their
+    # reference points land within this radius, even if their actual copper
+    # shapes (segment width, pad size) do not touch.  KiCad's connectivity
+    # engine instead requires real geometric overlap, so the default model
+    # can report a net "complete" that KiCad (``kicad-cli pcb drc``) reports
+    # as having unconnected items (Issue #4176).  Pass ``strict=True`` to use
+    # real shapely copper-shape intersection and match KiCad.
     POSITION_TOLERANCE = 0.01
 
-    def __init__(self, pcb: str | Path | PCB) -> None:
+    def __init__(self, pcb: str | Path | PCB, *, strict: bool = False) -> None:
         """Initialize the analyzer.
 
         Args:
-            pcb: Path to PCB file or loaded PCB object
+            pcb: Path to PCB file or loaded PCB object.
+            strict: When ``True``, segment↔segment, segment↔pad, and
+                segment↔via connectivity is decided by real geometric copper
+                contact (shapely polygon intersection: a segment is its
+                centerline buffered by ``width / 2``; a pad/via is its copper
+                shape) instead of the ``POSITION_TOLERANCE`` endpoint-proximity
+                radius.  This matches KiCad's connectivity semantics
+                (Issue #4176).  The default (``False``) preserves the legacy
+                tolerance-based behavior so existing consumers are unaffected.
+                Requires ``shapely``; strict mode fails loud (rather than
+                silently degrading to the tolerance model) if it is
+                unavailable.
         """
         from kicad_tools.schema.pcb import PCB as PCBClass
 
@@ -384,6 +405,17 @@ class NetStatusAnalyzer:
             self.pcb = PCBClass.load(str(pcb))
         else:
             self.pcb = pcb
+        self.strict = strict
+        # Strict-mode geometry caches (Issue #4176), keyed by object id.
+        self._segment_poly_cache: dict[int, Any] = {}
+        self._via_geom_cache: dict[int, Any] = {}
+        # Per-analyzer pad copper polygon cache (keyed ``REF.PAD``); ``None``
+        # until first built lazily in :meth:`_pad_polys`.
+        self._pad_poly_cache: dict[str, Any] | None = None
+        if strict:
+            from kicad_tools._shapely import require_shapely
+
+            require_shapely("net-status --strict real-geometry connectivity")
 
     def analyze(self) -> NetStatusResult:
         """Analyze all nets and return status result.
@@ -648,13 +680,24 @@ class NetStatusAnalyzer:
         # Build segment components for reuse
         segment_components = self._build_segment_components(segments)
 
-        # Connect pads through segment chains
+        # Connect pads through segment chains.
+        #
+        # Default mode: a pad joins the chain when its center is within
+        # POSITION_TOLERANCE of a segment endpoint.  Strict mode (Issue #4176):
+        # a pad joins the chain when its real copper polygon intersects any
+        # segment's copper polygon in the chain — an endpoint landing merely
+        # *near* a pad's copper edge no longer bonds.
         for component in segment_components:
             component_pads: set[str] = set()
             for seg_idx in component:
                 seg = segments[seg_idx]
-                component_pads.update(self._find_pads_at_point(seg.start, pad_positions))
-                component_pads.update(self._find_pads_at_point(seg.end, pad_positions))
+                if self.strict:
+                    component_pads.update(
+                        self._find_pads_touching_geom(self._segment_poly(seg), pad_positions)
+                    )
+                else:
+                    component_pads.update(self._find_pads_at_point(seg.start, pad_positions))
+                    component_pads.update(self._find_pads_at_point(seg.end, pad_positions))
 
             pad_list = list(component_pads)
             for i, pad in enumerate(pad_list):
@@ -662,9 +705,14 @@ class NetStatusAnalyzer:
                     graph[pad].add(other)
                     graph[other].add(pad)
 
-        # Connect pads through vias (pads at same via position)
+        # Connect pads through vias (pads whose copper the via barrel pierces).
+        # Default: pad center within POSITION_TOLERANCE of the via center.
+        # Strict (Issue #4176): pad copper polygon intersects the via copper.
         for via in vias:
-            via_pads = self._find_pads_at_point(via.position, pad_positions)
+            if self.strict:
+                via_pads = self._find_pads_touching_geom(self._via_geom(via), pad_positions)
+            else:
+                via_pads = self._find_pads_at_point(via.position, pad_positions)
             for pad in via_pads:
                 for other in via_pads:
                     if pad != other:
@@ -684,20 +732,42 @@ class NetStatusAnalyzer:
                 component_endpoints.append(seg.start)
                 component_endpoints.append(seg.end)
 
-            # Check if any endpoint of this segment chain touches a via
+            # Check if this segment chain's copper touches a via.
+            # Default: an endpoint within POSITION_TOLERANCE of the via center.
+            # Strict (Issue #4176): a segment copper polygon in the chain
+            # intersects the via copper geometry.
             for via in vias:
-                touches_via = any(
-                    self._points_close(ep, via.position) for ep in component_endpoints
-                )
+                if self.strict:
+                    via_geom = self._via_geom(via)
+                    touches_via = any(
+                        self._geoms_touch(self._segment_poly(segments[seg_idx]), via_geom)
+                        for seg_idx in component
+                    )
+                else:
+                    touches_via = any(
+                        self._points_close(ep, via.position) for ep in component_endpoints
+                    )
                 if touches_via:
                     # Collect all pads in this segment chain
                     chain_pads: set[str] = set()
                     for seg_idx in component:
                         seg = segments[seg_idx]
-                        chain_pads.update(self._find_pads_at_point(seg.start, pad_positions))
-                        chain_pads.update(self._find_pads_at_point(seg.end, pad_positions))
+                        if self.strict:
+                            chain_pads.update(
+                                self._find_pads_touching_geom(
+                                    self._segment_poly(seg), pad_positions
+                                )
+                            )
+                        else:
+                            chain_pads.update(self._find_pads_at_point(seg.start, pad_positions))
+                            chain_pads.update(self._find_pads_at_point(seg.end, pad_positions))
                     # Also include any pads at the via position itself
-                    chain_pads.update(self._find_pads_at_point(via.position, pad_positions))
+                    if self.strict:
+                        chain_pads.update(
+                            self._find_pads_touching_geom(self._via_geom(via), pad_positions)
+                        )
+                    else:
+                        chain_pads.update(self._find_pads_at_point(via.position, pad_positions))
                     # Connect all pads in this chain through the via
                     chain_list = list(chain_pads)
                     for i, pad in enumerate(chain_list):
@@ -730,23 +800,38 @@ class NetStatusAnalyzer:
             zone_boundaries,
         )
 
-        # Connect pads at same copper positions (segments and vias only).
+        # Connect pads that share a piece of copper (segments and vias only).
         # Zone polygon vertices are NOT sampled here because the pad-to-zone
         # and via-to-zone polygon containment checks above already handle zone
         # connectivity properly.  The previous zone_points[:1000] sampling cap
         # caused false-positive incomplete reports on boards with many filled
         # polygons (Issue #2035).
-        all_copper = [seg.start for seg in segments] + [seg.end for seg in segments]
-        all_copper.extend([via.position for via in vias])
+        #
+        # Default: two pads bond when both centers land within
+        # POSITION_TOLERANCE of the same segment endpoint / via center.
+        # Strict (Issue #4176): two pads bond when both copper polygons
+        # intersect the same segment / via copper geometry.
+        if self.strict:
+            copper_geoms = [self._segment_poly(seg) for seg in segments]
+            copper_geoms.extend(self._via_geom(via) for via in vias)
+            for geom in copper_geoms:
+                touching = self._find_pads_touching_geom(geom, pad_positions)
+                for i, pad_id in enumerate(touching):
+                    for other_id in touching[i + 1 :]:
+                        graph[pad_id].add(other_id)
+                        graph[other_id].add(pad_id)
+        else:
+            all_copper = [seg.start for seg in segments] + [seg.end for seg in segments]
+            all_copper.extend([via.position for via in vias])
 
-        for pad_id, pad_pos in pad_positions.items():
-            for copper_pos in all_copper:
-                if self._points_close(pad_pos, copper_pos):
-                    # Find other pads at this copper point
-                    for other_id, other_pos in pad_positions.items():
-                        if other_id != pad_id and self._points_close(copper_pos, other_pos):
-                            graph[pad_id].add(other_id)
-                            graph[other_id].add(pad_id)
+            for pad_id, pad_pos in pad_positions.items():
+                for copper_pos in all_copper:
+                    if self._points_close(pad_pos, copper_pos):
+                        # Find other pads at this copper point
+                        for other_id, other_pos in pad_positions.items():
+                            if other_id != pad_id and self._points_close(copper_pos, other_pos):
+                                graph[pad_id].add(other_id)
+                                graph[other_id].add(pad_id)
 
         return graph
 
@@ -1030,21 +1115,36 @@ class NetStatusAnalyzer:
 
         Returns:
             List of sets, each containing segment indices in a connected component
+
+        In the default mode two segments are chained when any of their
+        endpoints land within ``POSITION_TOLERANCE`` of each other.  In
+        ``strict`` mode (Issue #4176) they are chained iff their real copper
+        polygons (each segment's centerline buffered by ``width / 2``)
+        intersect — matching KiCad, which does not bond two traces whose
+        copper does not physically touch even when their endpoints are within
+        the snap tolerance.
         """
         if not segments:
             return []
+
+        if self.strict:
+            polys = [self._segment_poly(seg) for seg in segments]
 
         # Build segment adjacency graph
         segment_graph: dict[int, set[int]] = defaultdict(set)
         for i, seg_a in enumerate(segments):
             for j, seg_b in enumerate(segments):
                 if i != j:
-                    if (
-                        self._points_close(seg_a.start, seg_b.start)
-                        or self._points_close(seg_a.start, seg_b.end)
-                        or self._points_close(seg_a.end, seg_b.start)
-                        or self._points_close(seg_a.end, seg_b.end)
-                    ):
+                    if self.strict:
+                        connected = self._geoms_touch(polys[i], polys[j])
+                    else:
+                        connected = (
+                            self._points_close(seg_a.start, seg_b.start)
+                            or self._points_close(seg_a.start, seg_b.end)
+                            or self._points_close(seg_a.end, seg_b.start)
+                            or self._points_close(seg_a.end, seg_b.end)
+                        )
+                    if connected:
                         segment_graph[i].add(j)
                         segment_graph[j].add(i)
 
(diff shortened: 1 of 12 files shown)


rjwalters:
✅ **Approved!**

Phase 1 of #4176 is a clean, well-scoped implementation. All acceptance criteria verified against real behavior, not just claims.

## Verified end-to-end (through the real CLI entry points, not the inner functions — the #4187 lesson)
- `kct net-status --strict` (parser.py `net_status_strict` → `cli/__init__` dispatch → `net_status_cmd`): default reports `Complete: 1`, `--strict` reports `Incomplete: 1` on the 0.009mm/0.001mm over-connect fixture.
- `kct check --strict-connectivity` (parser.py + `commands/validation.py` → `check_cmd` → `DRCChecker(strict_connectivity=)` → `ConnectivityRule(strict=)`): default emits **0** connectivity errors; `--strict-connectivity` emits exactly **1**. The two `--strict` flags are distinct and non-colliding (`--strict` = warnings-fatal; `--strict-connectivity` = real geometry).

## Correctness
- `segment_copper_polygon` = `LineString([start,end]).buffer(width/2)` — shapely's default **round caps** correctly model KiCad's round trace ends (confirmed by the area test: rect + 2 semicircular caps). Degenerate cases handled defensively (zero-width → bare LineString/Point; zero-length+width → disk).
- Pad/via copper reuses `ConnectivityValidator._pad_copper_polygon`/`_via_copper_geom` via `_connectivity_geometry()`, so shape/rotation stays consistent with the reference partition extractor. The rect pad approximation is a *superset* of roundrect/oval, so it errs lenient on corners (right direction). Per-analyzer `id()`-keyed caches avoid rebuilding polygons per net.

## Default preserved / Phase 2 deferred
- Default (tolerance) path is byte-for-byte unchanged; existing `test_net_status*` and the full net-status/connectivity slice pass.
- `validate/connectivity.py`, `lvs/copper_lvs.py`, `analysis/ground_topology.py`, `audit/auditor.py` are **untouched** — the parallel `ConnectivityValidator` chainer is correctly left for Phase 2 (only its geometry primitives are reused read-only).
- shapely-required fail-loud is enforced (`require_shapely` in `__init__`), matching the curator's "no silent degrade" directive.

## Checks
- 377 net_status/connectivity + 31 new (geometry_copper/net_status_strict/connectivity_rule) + broader slice: all pass locally.
- ruff clean on all changed files; mypy no new errors in `geometry/copper.py` or the strict region (Type Check CI green).
- **CI: all 16 checks pass, mergeStateStatus CLEAN.** The `Test` job initially failed on `test_pathfinder_dormant_partner_perf.py` — a wall-clock micro-benchmark documented flaky on shared runners (issue #3581), in `router/` code this PR does not touch. Re-ran the job; it passed (16m18s, 21960 passing).

## Non-blocking note
Strict mode reuses `_pad_copper_polygon`/`_via_copper_geom`, which erode copper inward by `POUR_PAD_ERODE = 0.1mm`. A trace terminating in the outer 0.1mm ring of a pad (touching real copper but outside the eroded box) could in principle report a false-incomplete vs KiCad. In practice traces route near pad centers (well past 0.1mm — the AC #3 "reaches into pad" fixture at 109.9 clears the 109.8 eroded edge), and this is an opt-in flag with the default preserved, so it is acceptable for Phase 1. Worth keeping in mind when Phase 2 evaluates strict-by-default against the fixture boards.

Solid work.

Issue #4175, opened 2026-07-14
route-auto: retries append duplicate identical copper — 717 exact-duplicate segments accumulated on one board


rjwalters:
## Bug (companion to #4165, distinct fix)

Because `route-auto` re-routes the same single MST edge on every invocation (#4165) and never checks whether that copper already exists, **each retry appends an exact duplicate** of the same segment(s) — same start/end/width/layer/net, differing only in uuid.

Observed: after several completion-loop iterations over ~26 stuck nets, a dedupe pass removed **717 exact-duplicate copper elements** from one board (softstart rev-B). A single diagnosed net (`/V_TH_HI`) carried 4 identical copies of its one routed segment — one per strategy attempt.

Duplicates are invisible to netlist connectivity (same net), silently bloat the file, quadruple copper where clearances were already tight, and turn every later DRC/optimize pass into slow-motion.

## Suggested fix

Before emitting a segment, check for an existing same-net segment with identical geometry+layer (cheap hash set) — skip or replace. Independently useful: a `kct clean --dedupe-copper` maintenance command (our workaround was a 30-line uuid-insensitive dedupe script).

kicad-tools 0.15.0, rjwalters/softstart PR #28.

rjwalters:
## Curator Enhancement (2026-07-15, refreshed against `origin/main` @ `0e4232bf`, post-#4165/PR #4203)

### Status update: builds on current main

#4165 (PR #4203, **MERGED** 2026-07-15) landed the "truth in the exit condition" fix: `RoutingOrchestrator.route_net()` now runs `_verify_completion()` (`src/kicad_tools/router/connectivity.py`) after each strategy attempt/retry and demotes unreachable multi-pad nets to `partial=True`. It also added `--allow-partial` so incomplete copper is only persisted when the caller opts in. **This did not touch the duplicate-copper defect** — the PR description explicitly calls out #4175 as "a separate defect on the same retry path." This enhancement re-verifies the root cause against the merged code and gives an implementation-ready plan.

### Confirmed root cause (read against current `origin/main`)

[...]

Issue #2913, opened 2026-05-15
router: kct route-auto reports success but writes no physical traces (RoutingOrchestrator corridor-only result)


rjwalters:
## Observation

\`kct route-auto boards/07-matchgroup-test/output/matchgroup_test.kicad_pcb --net A0 -o /tmp/out.kicad_pcb\` prints:

\`\`\`
Routed net 'A0' successfully
  Strategy: GLOBAL_WITH_REPAIR
  Total length: 64.65mm
  Saved to: /tmp/out.kicad_pcb
\`\`\`

But \`kct net-status /tmp/out.kicad_pcb\` reports:

\`\`\`
[!] A0 (1 pads unconnected)
\`\`\`

Reproduced consistently on board 07 for all 17 unrouted signal nets (DQS_P, A0–A7, MIPI_CLK_N, DQ0–DQ3, DQ5–DQ7). Net-status shows the net still incomplete regardless of whether the input is the unrouted or partially-routed PCB.

## Root cause

\`RoutingOrchestrator._route_global()\` calls \`GlobalRouter.route_net()\` which returns a corridor assignment (tile-level waypoints), then returns a \`RoutingResult(success=True)\` with \`total_length_mm\` computed from waypoint distances. **No physical trace segments are ever written back into the PCB object.**

When \`route_net_auto()\` subsequently calls \`pcb.save(output_path)\`, the PCB is saved with zero new tracks for the net. The success flag reflects only "a routing corridor was found", not "physical traces were committed".

## Affected code

- \`src/kicad_tools/router/orchestrator.py\` – \`_route_global()\`, \`_route_escape_then_global()\`, all strategy methods return \`RoutingResult\` with metrics but no track write-back
- \`src/kicad_tools/mcp/tools/routing.py\` – \`route_net_auto()\` line 568: \`pcb.save(output_path)\` called on un-modified PCB
- \`src/kicad_tools/cli/commands/routing.py\` – \`run_route_auto_command()\` trusts \`result["success"]\` without verification

## Impact

Board 07 (matchgroup-test): 17/31 signal nets partially routed by \`kct route\`, all unrouted by \`kct route-auto\` despite success reports. All 4 matchgroup groups (DDR_DATA_BYTE_0, MIPI_CSI_LANES, HDMI_TMDS_LANES, ADDR_BUS) have at least one member that fails to persist.

[...]

Issue #1282, opened 2026-04-13
route-auto: 'Insufficient pads for global routing' on all nets


rjwalters:
## Description

After fixing the crash in #1268, `kct route-auto` now runs without crashing but fails to route any net with "Insufficient pads for global routing". Tested on every net in a 24-net board.

## Reproduction

```bash
kct route-auto chorus-test-revA-stripped.kicad_pcb --net GND -o output.kicad_pcb
# Routing failed for net 'GND'
#   Error: Insufficient pads for global routing

kct route-auto chorus-test-revA-stripped.kicad_pcb --net SPI_MOSI -o output.kicad_pcb
# Routing failed for net 'SPI_MOSI'
#   Error: Insufficient pads for global routing
```

Board: 4-layer, 38 footprints, 24 nets, multi-sheet schematic (chorus-test-revA).

## Analysis

The RoutingOrchestrator's pad detection may not be finding pads correctly. The same board routes 14/19 signal nets successfully via `kct route --strategy negotiated --grid auto`. The orchestrator's global routing step has stricter pad requirements or a different pad detection path.

## Expected behavior

`route-auto` should successfully route nets that `route` can route, using smarter strategy selection.

rjwalters:
Curating this issue per user request.

rjwalters:
## Curator Enhancement

### Confirmed Root Cause

The issue is a **broken contract in `RoutingOrchestrator.route_net()`**. The docstring at line 145 of `src/kicad_tools/router/orchestrator.py` states:

> `pads: Optional list of pads for this net (if None, **extracted from PCB**)`

But the implementation does **no such extraction**. When `pads=None` is passed, `None` flows unchanged into `_select_strategy()` and `_execute_strategy()`. Every strategy method (`_route_global`, `_route_escape_then_global`, `_route_hierarchical`, `_route_subgrid_adaptive`, `_route_with_via_resolution`, `_route_full_pipeline`, `_route_multi_resolution`) begins with the identical guard:

```python
if pads is None or len(pads) < 2:
    return RoutingResult(
        success=False,
        net=net,
        strategy_used=...,
        error_message="Insufficient pads for global routing",
    )
```

`route_net_auto()` in `src/kicad_tools/mcp/tools/routing.py` calls `orchestrator.route_net(net=net_name)` at line 541 — with **no `pads` argument**. As a result, every net on every board fails identically with "Insufficient pads", regardless of how many pads the net actually has.

[...]

PR #4203, opened 2026-07-15
fix(router): report partial route-auto completion instead of false success (#4165)

### LINKED ISSUE #4165: route-auto: 'Routed successfully' on multi-pad nets that remain unconnected; retries no-op (auto/escape/subgrid) — hierarchical is the only strategy that completes
## Bug (false success + no-op retry — controlled repro)

`kct route-auto` prints **"Routed net successfully"** and exits 0 while leaving pads of a multi-pad net unconnected, and subsequent invocations on the same net add no copper (silent no-op). Controlled single-variable test, same board, same 3-pad net (`/ZC_DETECT`, freshly stripped of all copper):

| Strategy | Claims success | Net actually complete after |
|---|---|---|
| auto (GLOBAL_WITH_REPAIR) | ✅ "84.67mm, 6 segments" | ❌ 1 pad unconnected |
| escape | ✅ | ❌ 1 pad unconnected |
| subgrid | ✅ | ❌ 1 pad unconnected |
| **hierarchical** | ✅ | ✅ **complete** |

So the global path routes **one corridor/edge per call**, reports success, and then treats the partially-routed net as done forever — retrying `auto` on the partial never adds the remaining MST edges.

## Impact

This silently caps any route-completion workflow. Our per-net completion loop (the workaround for #4159) plateaued at 46/77 for three full iterations — every iteration "successfully" routed the same stuck nets while adding nothing. Diagnosing required stripping a net's copper by hand and diffing per-strategy behavior. With `--strategy hierarchical` the same loop immediately resumed progressing.

## Suggested fixes

1. **Truth in the exit condition**: after routing, run the same connectivity check `net-status` uses; report `partially routed: k/n pads connected` and exit non-zero (or add `--require-complete`).
2. Make the auto orchestrator **iterate until the net is complete** (route remaining unconnected pads to existing net copper) rather than single-corridor-and-done — i.e., what `hierarchical` already does.
3. Document per-strategy semantics: nothing in `--help` suggests `auto` may connect only a subset of pads.

Context: kicad-tools 0.15.0, softstart rev-B (rjwalters/softstart PR #28), C++ backend. Companion to #4159 (batch starvation) — together they made "finish routing this board" a multi-hour archaeology exercise.

---

### Root cause (confirmed by reading the code)

The bug is in `GlobalRouter.route_net()`, called by every non-hierarchical orchestrator strategy. It is fundamentally a **two-terminal** router being used for **N-terminal** nets:

- `src/kicad_tools/router/global_router.py:117` `GlobalRouter.route_net(net, pad_positions, layer=None)`:
  - `src_pos, tgt_pos = self._pick_endpoints(pad_positions)` (line 147) picks exactly **two** of the net's pads (the two most distant, per the docstring at line 129: "For multi-pad nets, routes between the two most distant pads. The corridor will cover intermediate pads as well.")
  - Finds **one** region-graph path (`self.region_graph.find_path(...)`, line 152) between those two regions
  - Builds **one** `Corridor` from that one path (lines 163-172) and returns a single `CorridorAssignment`
  - "The corridor will cover intermediate pads as well" is an unenforced assumption, not a guarantee — a third pad is only connected if it happens to lie geometrically on that one corridor's path. This is exactly the controlled repro: 3-pad net, 1 pad left stranded off the src→tgt corridor.

- `src/kicad_tools/router/orchestrator.py:622` `RoutingOrchestrator._route_global()` takes that single `CorridorAssignment`, materializes its waypoints into `Segment` objects, and returns `RoutingResult(success=True, ...)` unconditionally whenever `assignment is not None` — there is no check that every pad in `pads` is actually reached by the produced segments.
  - `_route_escape_then_global()` (line 742) and `_route_subgrid_adaptive()` (line 938) both funnel through the same `_route_global()` for their "global" phase, so they inherit the identical one-corridor-and-done behavior — this is why auto/escape/subgrid all show the same failure mode in the repro table.

- `src/kicad_tools/router/orchestrator.py:306` `RoutingOrchestrator.route_net()` (the single entry point that dispatches to `_execute_strategy`) only retries `_select_strategy`'s alternatives **when `result.success` is False** (`if not result.success and result.alternative_strategies ...`, line 358). Since `_route_global` always reports `success=True` once it finds any corridor at all, this retry path never fires for a partially-routed net — confirming "retrying auto on the partial never adds the remaining MST edges."

- By contrast, `_route_hierarchical()` (`orchestrator.py:837`) delegates to `AdaptiveAutorouter.route(method="negotiated")` (`src/kicad_tools/router/adaptive.py:210`), which calls `router.route_all_negotiated(max_iterations=...)` — a genuinely iterative, multi-net/multi-pad negotiated router that keeps working until all pads/nets converge or the iteration budget is exhausted. This is a structurally different code path, not a parameter difference, which is why only `hierarchical` completes the net in the repro.

- The CLI success message itself is in `src/kicad_tools/cli/commands/routing.py::run_route_auto_command()` (~line 82): `if result["success"]: print(f"Routed net '{result['net_name']}' successfully")` — it trusts `result["success"]` verbatim with no independent post-route connectivity check.

- `route_net_auto()` in `src/kicad_tools/mcp/tools/routing.py:465` already imports `NetStatusAnalyzer` from `kicad_tools.analysis.net_status` (used elsewhere in the same module, e.g. `get_unrouted_nets()` at line 27) — this is the same connectivity oracle that `kct net-status` uses, and the same one implicated in **#4176** as over-connecting via endpoint-proximity tolerance rather than requiring real geometric copper contact.

### Dependency / coupling note: #4176

**#4176** ("net-status: PCB connectivity model diverges from KiCad") reports that `NetStatusAnalyzer._build_connectivity_graph` (`src/kicad_tools/analysis/net_status.py:591`) unions copper on endpoint proximity/tolerance rather than requiring actual geometric contact, causing it to bless boards KiCad's DRC considers electrically open.

This issue's suggested fix (1) — "run the same connectivity check net-status uses" — **directly reuses that oracle**. If the completion check implemented here is a thin call into `NetStatusAnalyzer`, it inherits #4176's over-connection bug: a route-auto call could route a corridor that lands *near* (but not touching) the stranded pad, and the reused net-status check would bless it as "complete" when KiCad's DRC would not. That would just move the false-success bug from "no check at all" to "check with the wrong tolerance" — a real improvement, but not a full fix.

**Recommendation**: the post-route completion check added here should be a **real per-pad reachability test on the freshly-produced geometry** (segments/vias returned by `RoutingResult`, or the union of that geometry with existing same-net copper) — i.e., "is this pad start/endpoint on a continuous copper run electrically joined to the rest of the net," using actual coincident-endpoint/overlap geometry rather than a tolerance union. This can be scoped narrowly (route-auto's own output + existing net copper), so it does **not** need to wait on a full #4176 fix to the general-purpose `NetStatusAnalyzer`. If a fully correct connectivity primitive isn't cheaply available yet, reuse `NetStatusAnalyzer` as an interim signal but **flag in the PR description that this inherits #4176's known false-positive risk** until #4176 lands, and prefer erring toward under-connecting (report incomplete) rather than over-connecting (report complete) when the two disagree.

### Related: #4175 (companion, keep separate)

**#4175** ("route-auto: retries append duplicate identical copper") is the companion defect on the exact same retry path: because `_route_global` recomputes the same MST edge/corridor on every invocation and never checks for existing same-net copper, each retry duplicates the segment (up to 717 exact-duplicate elements observed on one board). Keep the two issues **separate** — this issue (#4165) is about false success + non-iteration; #4175 is about de-duplication of copper once (if) route-auto does iterate. They will very likely touch the same functions (`_route_global`, `route_net_auto`'s persistence path) and should be built sequentially rather than in parallel to avoid merge conflicts — whichever lands first should account for the other's need (e.g., if #4165's iterate-until-complete fix lands first, it should not reintroduce #4175's duplication when routing "remaining unconnected pads to existing net copper").

### Scope recommendation

**Fix (1) — truth in the exit condition: MUST-HAVE, in this PR.**
This is the actual bug report (false "success" on a board that isn't routed) and is a bounded, mechanical change:
- After `orchestrator.route_net()` returns, before `route_net_auto()` returns its result dict, check actual per-pad connectivity of the net (existing copper + freshly produced segments/vias) rather than trusting `result.success`.
- If `k < n` pads are connected: set `result_dict["success"] = False` (or a new `partial: true` field alongside `pads_connected: k, pads_total: n`), and have `run_route_auto_command()` in `src/kicad_tools/cli/commands/routing.py` print `"partially routed: k/n pads connected"` and return a non-zero exit code — OR add a `--require-complete` flag that opts into this stricter behavior without changing existing default semantics for callers who intentionally use route-auto to route partial corridors. Curator has no strong opinion on default-strict vs opt-in-strict; recommend the PR author raise it explicitly since it's a behavior-compatibility call, not a technical one.
- Uses a per-pad reachability check as described above (not a blind reuse of the possibly-over-connecting `NetStatusAnalyzer` tolerance model — see the #4176 coupling note).

**Fix (2) — make auto iterate until complete: ASSESS TRACTABILITY, likely a FOLLOW-UP, not required for this PR.**
This is materially deeper surgery than fix (1):
- `GlobalRouter.route_net()` is architecturally two-terminal (`_pick_endpoints` → one region-graph path → one corridor). Making it complete an N-pad net requires either (a) wrapping it in a loop in the orchestrator that, after each corridor is materialized, re-evaluates which pads remain unconnected and calls `GlobalRouter.route_net()` again with a new src (an unconnected pad) and tgt (nearest point on already-routed net copper) — analogous to a incremental Steiner-tree build; or (b) replacing the single-pair call with a proper multi-terminal (MST/Steiner) corridor planner inside `GlobalRouter` itself. Option (a) is more contained (touches `orchestrator.py`'s `_route_global`/`_route_escape_then_global`/`_route_subgrid_adaptive`, not `global_router.py`'s core algorithm) but still needs: a termination condition, an anti-duplication check against existing net copper (the #4175 concern), and interaction with `region_graph.update_utilization` state across iterations (each iteration should account for previously-committed corridor tiles as obstacles/utilized, or later pads may double-route through already-congested corridors).
- Given `hierarchical` (`AdaptiveAutorouter` / `route_all_negotiated`) already solves the N-terminal problem correctly via a completely different, mature code path, a cheap interim mitigation worth calling out in this issue's discussion (not necessarily this PR) is: **have the orchestrator's strategy selection prefer `hierarchical` whenever the post-route connectivity check from fix (1) would otherwise report a partial net**, i.e., an automatic fallback/retry into `hierarchical` on partial-completion, mirroring the existing `alternative_strategies` retry mechanism in `route_net()` (`orchestrator.py:358`) but keyed on "partial" not just "failed." This reuses existing retry plumbing and sidesteps rewriting `GlobalRouter` — recommend the PR author evaluate this as fix (2)'s MVP before attempting true multi-terminal iteration in `_route_global` itself.
- **Recommendation**: land fix (1) + fix (3) in this PR. Open a dedicated follow-up issue for fix (2) once fix (1)'s honest reporting is in place (fix (1) is also what makes fix (2) reviewable/testable — you need truthful "partial" detection before you can prove "iterate until complete" actually completes the net). The follow-up should explicitly evaluate the "fall back to hierarchical on partial" MVP against full multi-terminal `GlobalRouter` surgery.

**Fix (3) — document per-strategy semantics: MUST-HAVE, cheap, in this PR.**
Update the `strategy` argument docstring in `route_net_auto()` (`src/kicad_tools/mcp/tools/routing.py:465`, the `Args: strategy:` block) and the `--strategy` CLI help text (wherever `route-auto --strategy` is defined, likely in the outer argparse setup alongside `route_cmd.py`'s `--strategy`) to state plainly that `global`/`escape`/`subgrid` route via a single two-terminal corridor and may leave a subset of a multi-pad net's pads unconnected even on "success", while `hierarchical` iterates to full net completion. This is a doc-only change and should ship regardless of how fixes (1)/(2) land.

### Acceptance Criteria

- [ ] `route_net_auto()` (or `run_route_auto_command()`) performs a real per-pad connectivity check after routing (existing net copper + newly produced segments/vias), independent of the strategy's own `success` flag.
- [ ] Synthetic 3-pad net, freshly stripped, routed with `--strategy global` (or `escape`/`subgrid`) leaving 1 pad unconnected: the command now reports `partially routed: k/n pads connected` (or equivalent) and exits non-zero — OR, if implemented as opt-in, `--require-complete` enforces this while default behavior is documented as unchanged. (PR author to state which approach was taken and why.)
- [ ] The same synthetic case routed with `--strategy hierarchical` still reports full success (2/2 or 3/3 connections) with no regression.
- [ ] The connectivity check used does not blindly trust `NetStatusAnalyzer`'s tolerance-based union without flagging the #4176 coupling in the PR description (or better: uses a stricter per-pad reachability primitive scoped to the freshly-produced + existing net geometry).
- [ ] `--strategy`/`route-auto` help text and `route_net_auto()` docstring updated per fix (3).
- [ ] If fix (2) (iterate-until-complete) is judged out of scope for this PR, a follow-up issue is opened and linked, per the scope recommendation above.
- [ ] No behavior regression for nets that already fully route via a single `global`/`escape`/`subgrid` corridor (2-pad nets, or multi-pad nets where the single corridor happens to cover all pads) — these must continue to report success with no spurious "partial" flag.

### Affected Files

- `src/kicad_tools/router/global_router.py` — `GlobalRouter.route_net()` (line ~117), `_pick_endpoints()` (line ~316): root cause, two-terminal corridor planner used for N-terminal nets. Read/reference for fix (2) tractability assessment; not necessarily modified for fix (1).
- `src/kicad_tools/router/orchestrator.py` — `RoutingOrchestrator.route_net()` (line ~306, retry-on-failure-only logic), `_route_global()` (line ~622), `_route_escape_then_global()` (line ~742), `_route_subgrid_adaptive()` (line ~938), `_route_hierarchical()` (line ~837, reference implementation of "iterate to completion"). Primary site for fix (1)'s post-route connectivity check and (if in scope) fix (2)'s iteration loop.
- `src/kicad_tools/mcp/tools/routing.py` — `route_net_auto()` (line ~465): orchestration entry point; already imports `NetStatusAnalyzer` (line ~15) used elsewhere in this file (`get_unrouted_nets()`, line ~27). Add the post-route per-pad connectivity check and `strategy` docstring update (fix 3) here.
- `src/kicad_tools/cli/commands/routing.py` — `run_route_auto_command()` (~line 82): success-message printing and exit-code logic; add "partially routed: k/n pads connected" reporting and non-zero exit (or `--require-complete` flag wiring).
- `src/kicad_tools/analysis/net_status.py` — `NetStatusAnalyzer._build_connectivity_graph()` (line ~591): the shared connectivity oracle also implicated in #4176; reference only unless the PR author chooses to reuse it directly (see coupling note — not recommended without flagging the inherited risk).
- `tests/test_routing_orchestrator.py` — existing orchestrator test suite (`test_route_net_success`, `test_route_global_computes_length`, etc.); add coverage for the new partial-completion detection.
- `tests/test_route_auto_fix.py`, `tests/test_route_auto_persistence.py` — existing route-auto CLI/persistence test suites; add coverage for the new exit-code/reporting behavior.

### Test Plan

- [ ] Manual verification: on a synthetic 3-pad net (mirroring the `/ZC_DETECT` repro — construct via a small fixture PCB, not a real board since softstart is local-only), run `kct route-auto --strategy global <fixture>` and confirm the output reports partial completion (or `--require-complete` exits non-zero), where previously it silently printed "Routed net successfully".
- [ ] Manual verification: the same fixture with `--strategy hierarchical` still reports full completion with no regression.
- [ ] Automated tests: extend `tests/test_routing_orchestrator.py` with a synthetic 3-pad net fixture where `GlobalRouter.route_net()` (or a mocked equivalent) is known to connect only 2 of 3 pads; assert the orchestrator/`route_net_auto()` result now surfaces the partial state (not blind `success=True`).
- [ ] Automated tests: extend `tests/test_route_auto_fix.py` or `tests/test_route_auto_persistence.py` with a CLI-level test asserting non-zero exit code (or `--require-complete` behavior) and the "partially routed: k/n pads connected" message text.
- [ ] Automated tests: add/extend a test confirming 2-pad nets and "lucky" multi-pad nets (single corridor happens to cover all pads) still report clean success with no spurious partial flag (regression guard for the acceptance criterion above).
- [ ] Edge cases: net with 0 remaining unconnected pads after routing (should report clean success); net where `GlobalRouter.route_net()` returns `None` (already-handled failure path, must remain unaffected); net where existing copper already connects some pads before route-auto is invoked (partial-before-routing case — connectivity check must account for pre-existing copper, not just this call's output).
- [ ] Regression: full `tests/test_routing_orchestrator.py`, `tests/test_route_auto_fix.py`, `tests/test_route_auto_persistence.py`, `tests/test_route_auto_mfr_tier*.py` suites green.

rjwalters:
Closes #4165

## Problem

`kct route-auto` printed **"Routed net successfully"** and exited 0 while leaving pads of a multi-pad net unconnected. Root cause: `GlobalRouter.route_net()` is a **two-terminal** corridor planner (`_pick_endpoints` picks the 2 most-distant pads, builds ONE corridor) used for **N-terminal** nets. `RoutingOrchestrator._route_global()` returned `success=True` unconditionally with no per-pad check; `escape`/`subgrid` funnel through it. The retry path in `route_net()` is keyed on `success==False`, so it never fired for a partial. Only `hierarchical` (iterative negotiated router) completed multi-pad nets.

### Fix (1) — truth in the exit condition (MUST-HAVE)
- New `router/connectivity.py::check_net_pad_connectivity()`: a **real per-pad reachability walk** (union-find over geometric copper adjacency — shared endpoints, point-on-segment, via layer bridges) on produced segments/vias unioned with pre-existing same-net copper.
- **Completion-check approach:** this does **NOT** reuse `NetStatusAnalyzer`'s endpoint-proximity tolerance union. Its coincidence epsilon is 1 µm (far below any clearance), so it errs toward **under-connecting** (reporting incomplete) rather than over-connecting. **It therefore does not inherit #4176's false-positive risk.**
- `RoutingOrchestrator.route_net()` runs `_verify_completion()` after a global-family strategy (and each retry). Fewer-than-all pads → demote to `success=False, partial=True` with `pads_connected`/`pads_total`. `RoutingResult` gains `partial`/`pads_connected`/`pads_total` (surfaced in `to_dict()`).
- CLI prints `partially routed net '<net>': k/n pads connected` and **exits non-zero**.
- `--allow-partial` persists incomplete copper (exit stays non-zero); default refuses to write it silently.

### Fix (2) MVP — fall back to hierarchical on partial (DONE, the cheap MVP)
- With `strategy="auto"` (retries enabled) a demoted partial reuses the existing `alternative_strategies` retry plumbing: `HIERARCHICAL_DIFF_PAIR` is injected as the first alternative and completes the net.
- A **forced** `--strategy global/escape/subgrid` sets `max_strategy_retries=0`, honoring the user's explicit choice and surfacing the honest partial (no silent fallback).
- True multi-terminal surgery inside `GlobalRouter` is **deferred** as the deeper follow-up (this MVP was clean, so it is included).

### Fix (3) — docs (MUST-HAVE)
- `--strategy` help and `route_net_auto()` docstring now state that `global`/`escape`/`subgrid` route a single two-terminal corridor and may leave a multi-pad net partially connected, while `hierarchical` iterates to completion.

## Behavior-compatibility note

Default `--strategy auto` **attempts** the hierarchical fallback on a partial multi-pad net and **completes it when it can** — so existing "route this net" workflows keep working and get *more* complete than the old silent-partial behavior. The fallback is **not** a guarantee: on congested or geometrically hard nets the net can remain partial even after hierarchical (e.g. `NET1` on the in-repo `routing-diagnostic.kicad_pcb` stays 2/3 because a middle pad is genuinely stranded off the two-terminal corridor and hierarchical does not complete it there either). In that case `route-auto` now reports the honest `partial 2/3` and exits **non-zero** — which is the *correct* result (the net really is incomplete), replacing the old false "Routed successfully" + exit 0. Explicitly forcing a global-family strategy also exits **non-zero** on a partial (previously silent 0) — this is the reported bug, and `--allow-partial` is provided as the escape hatch for callers who intentionally route partial corridors.

## Tests

- `tests/test_connectivity_reachability_4165.py` — 9 unit tests for the reachability primitive (2-pad, 3-pad stranded, chained, existing copper, via layer-bridge, near-miss anti-over-connection, PCB-schema segment shape).
- `tests/test_route_auto_partial_4165.py` — 8 orchestrator tests (global partial demoted; auto attempts the hierarchical fallback and completes when it can; best-partial kept when no strategy completes; lucky full corridor stays clean; 2-pad not flagged; hierarchical trusted; hard failure stays failure; `to_dict` exposes partial fields).
- `tests/test_route_auto_partial_cli_4165.py` — 5 CLI tests (non-zero exit + k/n message; NOT-saved note; `--allow-partial` saved-segments; full success exits 0; hard failure path).

All green: `test_routing_orchestrator.py`, `test_route_auto_fix.py`, `test_route_auto_persistence.py`, `test_route_auto_mfr_tier*.py`, `test_mcp_routing.py`, plus the 3 new suites. ruff check + format clean; mypy baseline check reports **0 new errors**.

## Follow-up

Fix (2)'s true multi-terminal (MST/Steiner) corridor planner inside `GlobalRouter` remains a follow-up; the hierarchical fallback is the interim completion path. Related #4175 (duplicate-copper-on-retry) is a separate defect on the same retry path.

### DIFF
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -109,6 +109,8 @@ def run_route_auto_command(args) -> int:
             # Issue #4148: spatial routing bound.  Confines the routed net to the
             # board-relative box and fails if the net has an endpoint outside it.
             region=getattr(args, "region", None),
+            # Issue #4165: persist partial multi-pad copper only when opted in.
+            allow_partial=getattr(args, "allow_partial", False),
         )
     except FileNotFoundError as e:
         print(f"Error: {e}", file=sys.stderr)
@@ -152,6 +154,39 @@ def run_route_auto_command(args) -> int:
         if result.get("output_path"):
             print(f"  Saved to: {result['output_path']}")
         return 0
+    elif result.get("partial"):
+        # Issue #4165: routing produced copper but left some pads of a
+        # multi-pad net unconnected.  Report the honest k/n and exit non-zero;
+        # the copper is only saved when --allow-partial was given.
+        pads_connected = result.get("pads_connected")
+        pads_total = result.get("pads_total")
+        kn = (
+            f"{pads_connected}/{pads_total}"
+            if pads_connected is not None and pads_total is not None
+            else "some"
+        )
+        print(
+            f"partially routed net '{result['net_name']}': {kn} pads connected",
+            file=sys.stderr,
+        )
+        print(f"  Strategy: {result.get('strategy_used', 'unknown')}", file=sys.stderr)
+        print(
+            "  A single two-terminal corridor left pad(s) unconnected. "
+            "Retry with --strategy hierarchical (or --strategy auto, which now "
+            "falls back to hierarchical) to complete the net.",
+            file=sys.stderr,
+        )
+        segs_written = result.get("segments_written")
+        if segs_written is not None:
+            print(f"  Partial copper saved: {segs_written} segments", file=sys.stderr)
+        elif not getattr(args, "allow_partial", False):
+            print(
+                "  Partial copper NOT saved (pass --allow-partial to persist it).",
+                file=sys.stderr,
+            )
+        for warning in result.get("warnings", []):
+            print(f"  Warning: {warning}", file=sys.stderr)
+        return 1
     else:
         print(f"Routing failed for net '{result['net_name']}'", file=sys.stderr)
         if result.get("error_message"):
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -172,8 +172,9 @@ def run_route_auto_command(args) -> int:
         print(f"  Strategy: {result.get('strategy_used', 'unknown')}", file=sys.stderr)
         print(
             "  A single two-terminal corridor left pad(s) unconnected. "
-            "Retry with --strategy hierarchical (or --strategy auto, which now "
-            "falls back to hierarchical) to complete the net.",
+            "Try --strategy hierarchical (or --strategy auto, which attempts a "
+            "hierarchical fallback) to improve completion; note that congested "
+            "or geometrically hard nets may remain partial even then.",
             file=sys.stderr,
         )
         segs_written = result.get("segments_written")
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -3623,13 +3623,16 @@ def _add_route_auto_parser(subparsers) -> None:
         help=(
             "Strategy override (default: auto). "
             "'auto' lets the orchestrator select the best strategy and, on a "
-            "partially-connected multi-pad net, automatically falls back to "
-            "'hierarchical' to complete it (Issue #4165). "
+            "partially-connected multi-pad net, automatically ATTEMPTS the "
+            "'hierarchical' fallback to complete it (Issue #4165); the fallback "
+            "improves completion when it can but does not guarantee it, so a "
+            "hard net may stay partial and route-auto still exits non-zero. "
             "'global'/'escape'/'subgrid' route a SINGLE two-terminal corridor "
             "and may leave intermediate pads of a multi-pad net unconnected "
             "even when they report success (route-auto then reports "
             "'partially routed: k/n pads connected' and exits non-zero). "
-            "'hierarchical' iterates to full net completion."
+            "'hierarchical' iterates toward full net completion but may still "
+            "leave a congested net partial."
         ),
     )
     route_auto_parser.add_argument(
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -3622,8 +3622,23 @@ def _add_route_auto_parser(subparsers) -> None:
         default="auto",
         help=(
             "Strategy override (default: auto). "
-            "'auto' lets the orchestrator select the best strategy. "
-            "Other values force a specific strategy."
+            "'auto' lets the orchestrator select the best strategy and, on a "
+            "partially-connected multi-pad net, automatically falls back to "
+            "'hierarchical' to complete it (Issue #4165). "
+            "'global'/'escape'/'subgrid' route a SINGLE two-terminal corridor "
+            "and may leave intermediate pads of a multi-pad net unconnected "
+            "even when they report success (route-auto then reports "
+            "'partially routed: k/n pads connected' and exits non-zero). "
+            "'hierarchical' iterates to full net completion."
+        ),
+    )
+    route_auto_parser.add_argument(
+        "--allow-partial",
+        action="store_true",
+        help=(
+            "Persist a partially-routed multi-pad net's copper instead of "
+            "refusing to write incomplete copper. Exit code stays non-zero and "
+            "the k/n partial report is still printed (Issue #4165)."
         ),
     )
     route_auto_parser.add_argument(
--- a/src/kicad_tools/mcp/tools/routing.py
+++ b/src/kicad_tools/mcp/tools/routing.py
@@ -470,6 +470,7 @@ def route_net_auto(
     enable_repair: bool = True,
     enable_via_resolution: bool = True,
     region: str | tuple[float, float, float, float] | None = None,
+    allow_partial: bool = False,
 ) -> dict:
     """Route a specific net using the RoutingOrchestrator.
 
@@ -478,15 +479,38 @@ def route_net_auto(
     characteristics (pin pitch, differential pairs, density, via conflicts)
     and automatically selects the optimal routing strategy.
 
+    Per-strategy completion semantics (Issue #4165)
+    -----------------------------------------------
+    The ``global``, ``escape``, and ``subgrid`` strategies route a **single
+    two-terminal corridor** between the two most distant pads of a net.  For a
+    multi-pad net this can leave intermediate pads unconnected even though the
+    strategy's own internal status is "success".  ``route_net_auto`` now runs a
+    real per-pad copper-reachability check after routing (independent of the
+    strategy's self-reported success) and, when a net is only partially
+    connected, reports ``success=False`` with ``partial=True`` and
+    ``pads_connected``/``pads_total`` populated.  With ``strategy="auto"`` the
+    orchestrator automatically falls back to the ``hierarchical`` (iterative
+    negotiated) router, which completes multi-pad nets by construction.  When a
+    partial route cannot be completed, no copper is saved unless
+    ``allow_partial=True``.
+
     Args:
         pcb_path: Absolute path to .kicad_pcb file
         net_name: Name of the net to route (e.g., "GND", "SPI_CLK")
         output_path: Path for output file. If None, result is not saved.
         strategy: Strategy override ("auto", "global", "escape", "hierarchical",
                   "subgrid", "via_resolution", or "multi_resolution").
-                  Use "auto" for smart selection.
+                  Use "auto" for smart selection.  Note: "global"/"escape"/
+                  "subgrid" route a single two-terminal corridor and may leave
+                  a multi-pad net partially connected; "hierarchical" iterates
+                  to full net completion (Issue #4165).
         enable_repair: Whether to enable automatic clearance repair after routing
         enable_via_resolution: Whether to enable via conflict resolution
+        allow_partial: When True, a partially-routed multi-pad net (some pads
+                  still unconnected) is still saved to ``output_path`` instead
+                  of refusing to write incomplete copper.  ``success`` remains
+                  False and ``partial`` True either way; this only controls
+                  whether the partial copper is persisted (Issue #4165).
         region: Optional spatial routing bound (Issue #4148, Phase 2a).  Either
                 a ``"x1,y1,x2,y2"`` string or a ``(x1, y1, x2, y2)`` tuple in
                 BOARD-RELATIVE mm (same convention as ``pcb strip --region``
@@ -691,6 +715,15 @@ def route_net_auto(
 
         orchestrator._select_strategy = _forced_select  # type: ignore[method-assign]
 
+        # Issue #4165: honor the user's explicit strategy choice.  When a
+        # strategy is forced we DO NOT silently fall back to hierarchical on a
+        # partial route -- the caller asked for exactly this strategy, so a
+        # multi-pad net stranded by a single two-terminal corridor is reported
+        # honestly as ``partial`` (k/n) and exits non-zero.  Disabling the
+        # retry loop is what preserves that honest partial; the automatic
+        # hierarchical fallback is reserved for ``strategy="auto"``.
+        orchestrator.max_strategy_retries = 0
+
     # Build pad list from PCB footprints so the orchestrator can
     # perform strategy selection and routing (without pads every
     # strategy returns "Insufficient pads").
@@ -763,6 +796,12 @@ def route_net_auto(
     result_dict = result.to_dict()
     result_dict["net_name"] = net_name
 
+    # Issue #4165: decide whether the produced copper is persistable.  A fully
+    # successful route always persists; a PARTIAL route (multi-pad net with
+    # stranded pads) persists only when the caller opts in via ``allow_partial``
+    # so incomplete copper is not written silently.
+    should_persist = result.success or (getattr(result, "partial", False) and allow_partial)
+
     # Save output if requested and routing succeeded.
     #
     # Issue #2913: the orchestrator's RoutingResult carries the produced
@@ -774,7 +813,7 @@ def route_net_auto(
     # and vias via the PCB schema's ``add_trace`` / ``add_via`` helpers
     # before saving, and surface a clear failure when the orchestrator
     # reports success but produced no physical segments.
-    if output_path and result.success:
+    if output_path and should_persist:
         segments_written, vias_written = _persist_routing_result_to_pcb(pcb, result, net_name)
         result_dict["segments_written"] = segments_written
         result_dict["vias_written"] = vias_written
--- a/src/kicad_tools/mcp/tools/routing.py
+++ b/src/kicad_tools/mcp/tools/routing.py
@@ -489,10 +489,13 @@ def route_net_auto(
     strategy's self-reported success) and, when a net is only partially
     connected, reports ``success=False`` with ``partial=True`` and
     ``pads_connected``/``pads_total`` populated.  With ``strategy="auto"`` the
-    orchestrator automatically falls back to the ``hierarchical`` (iterative
-    negotiated) router, which completes multi-pad nets by construction.  When a
-    partial route cannot be completed, no copper is saved unless
-    ``allow_partial=True``.
+    orchestrator automatically **attempts** the ``hierarchical`` (iterative
+    negotiated) router as a fallback on a partial.  The hierarchical fallback
+    improves completion where it can, but it does **not** guarantee completion
+    on every net — congested or geometrically hard nets may remain partial even
+    after the fallback, in which case ``route_net_auto`` still reports the honest
+    ``success=False``/``partial=True`` result and exits non-zero.  When a partial
+    route cannot be completed, no copper is saved unless ``allow_partial=True``.
 
     Args:
         pcb_path: Absolute path to .kicad_pcb file
--- a/src/kicad_tools/router/connectivity.py
+++ b/src/kicad_tools/router/connectivity.py
@@ -1,0 +1,234 @@
+"""Per-pad copper-reachability check for routed nets (Issue #4165).
+
+The orchestrator's global-family strategies (``global`` / ``escape`` /
+``subgrid``) route a **single two-terminal corridor** between the two most
+distant pads of a net (see :meth:`GlobalRouter.route_net`).  For a multi-pad
+net that leaves intermediate pads stranded, yet the strategy reports
+``success=True`` unconditionally.  This module supplies the missing
+**truth-in-exit** signal: after a strategy produces geometry, verify that every
+pad of the net is actually reachable across the produced copper (unioned with
+pre-existing same-net copper).
+
+Why not reuse ``NetStatusAnalyzer``
+-----------------------------------
+``analysis.net_status.NetStatusAnalyzer`` unions copper on **endpoint-proximity
+tolerance** (Issue #4176), which over-connects relative to KiCad's real
+geometric-contact requirement — reusing it here would just move the false
+"complete" from "no check" to "check with the wrong tolerance".  Instead this
+module walks **actual copper adjacency**: two elements are joined only when
+their geometry truly touches — a shared endpoint (within a tight numeric
+epsilon, not a design-rule clearance), a point lying **on** a segment, or two
+segments crossing/overlapping.  A pad is "connected" only if it lies on a
+copper element that is transitively joined to every other pad.
+
+The epsilon here is a floating-point coincidence tolerance (``EPS`` = 1 µm),
+deliberately far below any trace width or clearance, so the check errs toward
+**under-connecting** (reporting incomplete) rather than over-connecting — the
+safe direction for a completion oracle.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+# Geometric coincidence tolerance in mm.  This is a numeric epsilon for
+# floating-point equality of coordinates, NOT a clearance/tolerance union.
+# 1 micron is well below any realistic trace width, so "touching" here means
+# the copper genuinely overlaps or shares an endpoint.
+EPS = 1e-3
+
+
+@dataclass(frozen=True)
+class _Seg:
+    """Layer-annotated segment endpoints for adjacency testing."""
+
+    x1: float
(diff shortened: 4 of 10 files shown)


rjwalters:
## Changes Requested — pre-existing regression tests break (CI Test job will fail)

Strong implementation overall — the connectivity primitive is genuinely correct and does exactly what #4165 (and the curator's #4176 coupling note) asked for. But there is a blocking regression that must be fixed before merge.

### Blocking: two pre-existing tests in `tests/test_mcp_routing.py` fail on this branch

`TestRouteNetAuto::test_no_attribute_error_on_fixture` and `TestRouteNetAuto::test_route_net_auto_routes_multi_pad_net` both **pass on `main`** and **fail on this branch**. Both assert `route_net_auto(pcb=routing-diagnostic.kicad_pcb, net="NET1")` returns `result["success"] is True`. Under the new honest completion check that call now returns `success=False, partial=True, pads_connected=2, pads_total=3`.

I verified this is **not** a false-incomplete — the check is correct. Captured geometry for NET1 (forced `--strategy global`):

```
PADS: P0=(4.05, 4.0)  P1=(7.05, 8.45)  P2=(8.0, 12.0)
10 F_CU segments forming ONE corridor P0 -> P2:
  (4.05,4.0)->(3.75,5.25)-> ... ->(8.25,12.75)->(8.0,12.0)
```

P0 and P2 (the two most-distant pads) are on the corridor; **P1 at (7.05, 8.45) lies on none of the segments** and is genuinely stranded. So NET1 was a 2/3 partial all along — `main` was reporting the exact false success #4165 is about. The check is right; the two tests were asserting the old buggy behavior and were **not updated in this PR**.

These tests run in the CI **Test** job, so CI will fail (it was still `pending` at review time; local reproduction is deterministic).

### What to do
Reconcile the two tests with the honest new behavior. Options:
1. Update both assertions to expect the partial signal (`partial is True`, `pads_connected==2`, `pads_total==3`) — but preserve each test's original regression intent (GH-1268: no `AttributeError` on `PCB.path`; GH-1282: pads actually passed to the orchestrator, i.e. NOT the old "Insufficient pads" failure). Asserting `partial` (rather than a hard failure) keeps both guards meaningful.
2. Or point them at a net/fixture that genuinely completes so `success is True` remains a valid assertion.

### Substantive note (please address in the same pass, not a follow-up)
The PR description states `--strategy auto` "**completes** multi-pad nets (via the hierarchical fallback) and exits 0, so existing workflows keep working." On the project's own `routing-diagnostic.kicad_pcb`, default `auto` on NET1 now returns **partial 2/3 / exit non-zero** — the hierarchical fallback does **not** complete it (forced `--strategy hierarchical` returns `success=False` with 97 segments on this fixture). So the Fix (2) MVP completion claim does not hold for at least one in-repo fixture, and default `route-auto` now exits non-zero where it previously exited 0. That non-zero is arguably *correct* (the net really is partial), but the description overstates the fallback's completeness. Please either soften the behavior-compatibility claim or confirm/adjust the fallback so representative fixtures actually complete.

### What's good (verified locally)
- `router/connectivity.py` union-find is correct: hand-verified 2-pad join, 3-pad stranding, chained segments, existing-copper completion, via layer-bridge, and the 50µm near-miss anti-over-connection guard. The 1µm EPS is appropriate (far below clearance, above float noise).
- Does NOT reuse `NetStatusAnalyzer`'s tolerance union — errs toward under-connecting, as required. #4176 risk not inherited.
- `_verify_completion` correctly scopes to the global family, skips hierarchical, skips <2-pad nets, and the `#2913` segment-less guard is sound (a no-copper "success" is not spuriously demoted).
- Forced `--strategy global/escape/subgrid` sets `max_strategy_retries=0` and surfaces the honest partial without silent fallback — verified end-to-end (forced global on NET1 -> `success=False, partial=True, 2/3`).
- `--allow-partial` persistence gating (`should_persist`) composes correctly with the fallback.
- All 22 new tests pass; `test_routing_orchestrator.py`, `test_route_auto_fix.py`, `test_route_auto_persistence.py` green locally.

The core fix is right and I'd be glad to approve once the two obsolete tests are reconciled and the description's completeness claim is squared with the fixture behavior.

rjwalters:
Addressed the Judge's two blockers. Core fix (`router/connectivity.py` per-pad reachability + `_verify_completion` demotion) is untouched.

**Blocker 1 — reconciled the two obsolete `tests/test_mcp_routing.py` tests (this was the CI Test failure), preserving each test's original regression intent:**

- `TestRouteNetAuto::test_no_attribute_error_on_fixture` (GH-1268 — guards against the `PCB.path` `AttributeError` crash): keeps its "no exception / well-formed dict" guard. It no longer asserts `success is True`; it tolerates the now-correct honest partial and, when `success` is `False`, asserts the partial signal (`partial is True`, `pads_total == 3`, `pads_connected < 3`). Still catches an `AttributeError` crash or the pre-orchestrator `Insufficient pads` guard.
- `TestRouteNetAuto::test_route_net_auto_routes_multi_pad_net` (GH-1282 — guards that pads actually reach the orchestrator): keeps that guard (asserts `Insufficient pads` is absent and a real `strategy_used` ran), and now documents the corrected #4165 behavior by asserting the honest partial on NET1: `success is False`, `partial is True`, `pads_connected == 2`, `pads_total == 3`.

I verified NET1 (and NET3) on `routing-diagnostic.kicad_pcb` are genuinely 2/3 partials — no multi-pad net on that fixture completes, so I asserted the honest partial rather than switching fixtures. The orchestrator-level "genuine multi-pad success" case is already covered by `test_route_auto_partial_4165.py::test_lucky_full_corridor_reports_clean_success` and `test_hierarchical_result_not_subject_to_completion_check`.

**Blocker 2 — squared the overstated fallback claim with reality.** Forced hierarchical also returns partial on this fixture, so "completes multi-pad nets via the hierarchical fallback and exits 0" was wrong. Softened to "auto ATTEMPTS the hierarchical fallback and exits non-zero if still incomplete" in:
- `mcp/tools/routing.py` `route_net_auto` docstring
- `cli/parser.py` `--strategy` help
- `cli/commands/routing.py` partial-route suggestion text
- `router/orchestrator.py` retry-reason string
- `router/strategies.py` `pads_connected` doc
- the PR body Behavior-compatibility note (`gh pr edit`).

**Verification (worktree, C++ backend available):**
- `tests/test_mcp_routing.py` — 37 passed (incl. both reconciled tests)
- `test_routing_orchestrator.py`, `test_route_auto_fix.py`, `test_route_auto_persistence.py`, `test_connectivity_reachability_4165.py`, `test_route_auto_partial_4165.py`, `test_route_auto_partial_cli_4165.py` — 115 passed
- ruff check + format clean; mypy baseline — 0 new errors

No `uv.lock`/`.loom` churn; changes scoped to the 5 doc-claim source files + `test_mcp_routing.py`. Ready for re-review.

rjwalters:
✅ **Approved!** (Re-review after Doctor cycle b0f2dc5f)

Both blockers from the prior review are resolved, and the core fix is untouched.

## Reconciled tests — original intent preserved, not neutered

**`test_no_attribute_error_on_fixture` (GH-1268):** Still guards the original AttributeError-crash regression — asserts the call returns a well-formed dict without raising (`isinstance(result, dict)`, `"success" in result`, `net_name == "NET1"`, `isinstance(result["success"], bool)`) and still rejects the old pre-orchestrator `"Insufficient pads"` path. It now tolerates the honest partial: when `success` is False it asserts `partial is True`, `pads_total == 3`, `pads_connected < 3`. No false-success assertion remains.

**`test_route_net_auto_routes_multi_pad_net` (GH-1282):** Still guards the original "pads reach the orchestrator" regression — asserts `"Insufficient pads" not in error_message`, `strategy_used` present and `!= "unknown"` (a real strategy ran). It now asserts the corrected behavior: `success is False`, `partial is True`, `pads_connected == 2`, `pads_total == 3`. Neither test is vacuous.

## Core fix unchanged
`router/connectivity.py` and `RoutingOrchestrator._verify_completion` are byte-identical to my prior approval. The Doctor commit touched only **doc strings** in `orchestrator.py` (the retry-reason text) and `strategies.py` (the `pads_connected` docstring) — no completion-check logic changed.

## Doc softening is accurate
All 5 sites (mcp/tools/routing.py docstring, cli/parser.py `--strategy` help, cli/commands/routing.py partial suggestion, orchestrator retry reason, strategies.py doc) now state auto **attempts** the hierarchical fallback, improves completion where it can but does **not** guarantee it, and exits non-zero if still partial. This matches the actual behavior (NET1 stays 2/3 even under forced hierarchical).

## Genuine multi-pad success still guarded
`tests/test_route_auto_partial_4165.py::test_lucky_full_corridor_reports_clean_success` asserts `success is True`, `partial is False`, `pads_connected == 3`, `pads_total == 3` for a 3-pad net covered by one corridor. "Multi-pad routing can succeed" is not left unguarded.

## Verification
- `tests/test_mcp_routing.py` — 37 passed (local, worktree)
- 3 new 4165 suites — 22 passed (9 + 8 + 5)
- CI: **all checks pass**, including Test (17m41s), Board 05 Routing Regression (19m45s), Diff-Pair / Match-Group regressions, all board E2E jobs
- `mergeStateStatus`: **CLEAN**

Solid work reconciling the obsolete false-success assertions while keeping their crash/orchestrator-reach guards intact.

PR #4204, opened 2026-07-15
fix(routing): dedupe copper at emission time + add `kct pcb dedupe` (#4175)

### LINKED ISSUE #4175: route-auto: retries append duplicate identical copper — 717 exact-duplicate segments accumulated on one board

rjwalters:
## Summary

Closes #4175.

`route-auto` re-solves the same MST edge on every external invocation (#4165) and, before this fix, appended a geometrically-identical, uuid-distinct copy of the same copper each time. Duplicates are invisible to netlist connectivity (same net) but silently bloat the file, quadruple copper where clearances are already tight, and slow every later DRC/optimize pass. A dedupe pass previously removed **717 exact-duplicate copper elements** from one softstart rev-B board.

Ships **both** curator-recommended pieces (the CLI wrapper is thin once the key logic is shared):

### 1. Emission-time dedup (root-cause fix, benefits every caller)

- `PCB.add_trace` / `PCB.add_via` (`src/kicad_tools/schema/pcb.py`) now skip appending a segment/via whose dedup key already exists on the board:
  - **segment**: `(net, layer, order-insensitive endpoints, width)`
  - **via**: `(net, rounded position, order-insensitive layers)`
  - Coordinates/widths round to ~1 micron, reusing the `EPS = 1e-3` mm coincidence precedent from `router/connectivity.py` (shipped in #4165) rather than inventing a second epsilon.
- **Skip-on-match**, `dedupe=True` by default (the safe default the reporter wanted). `add_trace` returns `[]` and `add_via` returns `None` when a duplicate is skipped; `dedupe=False` preserves the historical always-append behavior for the rare case that needs it.
- A lazily-built, incrementally-maintained per-board key set avoids O(N²) rescans during a bulk persist; it is invalidated on any bulk copper mutation (`strip_traces`, `remove_segments`, `clear`).
- `_persist_routing_result_to_pcb` (`src/kicad_tools/mcp/tools/routing.py`) now returns `(segments_written, vias_written, segments_deduplicated, vias_deduplicated)`; `route_net_auto` threads the dedup counts into `result_dict` so a fully-duplicate retry reports `0 written / N deduplicated` honestly. A retry that reproduces existing copper is no longer misclassified as the #2913 empty-save failure.

**Correctness guardrails** (tested): segments sharing only one endpoint, or differing in layer / net / width, are NOT deduped — only fully-identical geometry (with reversed endpoints treated as identical) is.

### 2. Cleanup command for already-bloated boards

- `PCB.dedupe_copper()` groups existing copper by the same keys, keeps one instance per group, and drops the rest from both the in-memory lists and the S-expression tree (so `save()` persists the cleanup). Connectivity is unchanged because every removed element is identical to a kept copy.
- `kct pcb dedupe <pcb>` (sibling to `pcb strip`; `src/kicad_tools/cli/commands/pcb.py` + `parser.py`): **in-place by default** (matching the #4166/#4195 convention), `-o` override, `--dry-run`, `text`/`json`. Reports segments/vias removed; no-op (exit 0, 0 removed) on a clean board.

## Route-auto no longer doubles copper

Confirmed: `_persist_routing_result_to_pcb` invoked twice on the same net produces `(2,1)` written on the first call and `(0,0)` written / `(2,1)` deduplicated on the second, with `segments_in_net`/`vias_in_net` counts unchanged — copper emission is idempotent (`tests/test_route_auto_persistence.py::TestPersistIsIdempotent`).

## Tests

New tests (all green), using synthetic in-repo boards (softstart is local-only):
- `tests/test_pcb.py::TestCopperDedup` — exact-dup skipped; order-insensitive; different layer/net/width/shared-endpoint NOT deduped; `dedupe=False` escape hatch; `dedupe_copper()` cleanup + no-op + connectivity preservation; survives save/reload.
- `tests/test_route_auto_persistence.py::TestPersistIsIdempotent` — persist twice does not double copper.
- `tests/test_cli_pcb.py::TestPcbDedupe` — in-place, `-o`, json counts, no-op, dry-run, connectivity preserved.

Results:
- `tests/test_pcb.py tests/test_route_auto_persistence.py tests/test_cli_pcb.py`: **332 passed** (310 pre-existing + 22 new).
- Broad sweep `-k "strip or route_auto or add_trace or add_via or persist or merge_routes or dedupe"`: **330 passed, 6 skipped**.
- ruff check + format: clean. `check_mypy_baseline.py`: **0 new errors**.

### DIFF
--- a/src/kicad_tools/cli/commands/pcb.py
+++ b/src/kicad_tools/cli/commands/pcb.py
@@ -13,7 +13,7 @@ def run_pcb_command(args) -> int:
     if not args.pcb_command:
         print("Usage: kicad-tools pcb <command> [options] <file>")
         print(
-            "Commands: summary, footprints, nets, traces, stackup, zones, strip, reannotate, sync-netlist, add-3d-models, remove-footprint, move-footprint, page-fit, center-on-sheet, lock-footprints, unlock-footprints, add-zone, snap-rotation, edit-outline, net-audit, export-dsn, import-ses"
+            "Commands: summary, footprints, nets, traces, stackup, zones, strip, dedupe, reannotate, sync-netlist, add-3d-models, remove-footprint, move-footprint, page-fit, center-on-sheet, lock-footprints, unlock-footprints, add-zone, snap-rotation, edit-outline, net-audit, export-dsn, import-ses"
         )
         return 1
 
@@ -29,6 +29,10 @@ def run_pcb_command(args) -> int:
     # Handle strip command separately (doesn't use pcb_query)
     if args.pcb_command == "strip":
         return _run_strip_command(args, pcb_path)
+
+    # Handle dedupe command (remove exact-duplicate copper; issue #4175)
+    if args.pcb_command == "dedupe":
+        return _run_dedupe_command(args, pcb_path)
 
     # Handle reannotate command separately (doesn't use pcb_query)
     if args.pcb_command == "reannotate":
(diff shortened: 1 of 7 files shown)


rjwalters:
## Judge Review: APPROVED

Reviewed against issue #4175 curated acceptance criteria. All checks pass (16/16 CI green), merge state CLEAN.

### Correctness verification (the high-risk parts)

**Exact-geometry key does NOT collapse distinct segments** — the segment key sorts the *endpoint pair* `((p1,p2) if p1<=p2 else (p2,p1))`, not the individual coordinates. So `A→B` and `B→A` collapse (correct dup), while `A→B` and `A→C` produce different keys (correct: segments sharing one endpoint are preserved). A degenerate zero-length segment yields a valid `(p,p)` key that only dedups against another identical zero-length segment — no accidental collapse. `_DEDUP_QUANT=3` (~1µm) reuses the `EPS=1e-3` mm precedent from `router/connectivity.py`; safe given the router grid resolution (the reported duplicates are *geometrically identical* re-solves of the same corridor, not near-misses).

**Cache-invalidation (highest-risk) — no staleness bug found.** Traced every path that reassigns/mutates `_segments`/`_vias`:
- `clear()` (2620), `remove_segments()` (2875), `dedupe_copper()` (4899), `strip_traces()` (5665, placed after all reassignments incl. orphan-via handling) all call `_invalidate_dedup_keys()`.
- `_parse()` appends during construction while `_segment_keys/_via_keys` are still `None` → rebuilt lazily on first `add_trace`/`add_via`. No stale-append or wrong-skip.
- Graphics removal paths (`remove_edge_contour`, `replace_outline`) touch only `_graphic_*`, not copper — correctly not invalidating.
- `merge_routes_into_pcb` (router/io.py) is pure text splicing and does NOT go through `add_trace`/`add_via`, so it's outside the dedup contract by design (separate path from `_persist_routing_result_to_pcb`).

**#2913 empty-save guard preserved.** `_persist_routing_result_to_pcb` now returns a 4-tuple; the failure branch fires only when `segments_written+vias_written+segments_deduplicated+vias_deduplicated == 0`, so a fully-duplicate retry reports honest `0 written / N deduplicated` instead of a false empty-save failure.

**`dedupe_copper()` keeps the FIRST occurrence per key** → stable uuids, no churn; removes duplicate S-expression nodes by uuid so `save()` persists; connectivity preserved (every dropped element is identical to a kept copy).

### CLI
`kct pcb dedupe <pcb>` — in-place default (consistent with #4166/#4195), `-o` override, `--dry-run`, text/json, no-op exit 0 on clean board. Consistent with the `pcb strip` sibling.

### Local validation
- `tests/test_pcb.py tests/test_route_auto_persistence.py tests/test_cli_pcb.py`: **332 passed** (matches PR claim).
- `TestCopperDedup`, `TestPersistIsIdempotent`, `TestPcbDedupe`: 22 passed.
- ruff check + format: clean on all changed files.
- Type Check CI job: green (the 15 mypy errors in pcb.py are all pre-existing baseline lines — 731/973/1907/1962/... — none touch the new dedup code).

Root-cause fix at the correct layer (`add_trace`/`add_via`, benefits all callers) plus the cleanup command for already-bloated boards. Well-tested, correctly scoped, no regressions.

Approving.

Issue #4202, opened 2026-07-15
route (quality): negotiated/A* router emits same-layer different-net crossing copper under grid/budget pressure — architect design pass


rjwalters:
## Context

Split out from #4178 during curation. #4178's investigation confirmed two things:
1. The post-route DRC gate (`run_post_route_drc` → `run_geometric_drc` → real `kicad-cli pcb drc --refill-zones`, from PR #3815) **already catches** router-emitted shorts and exits non-zero — that's the bounded fix tracked in #4178 (plus a diff-pair-suffix false-negative and `--strict-drc`).
2. Underneath the gate, the router **genuinely emits same-layer different-net crossing tracks** under grid/budget pressure (reproduced live on `boards/05-bldc-motor-controller` with an unsafe grid + tight timeout: `clearance_segment_segment: SHORT` on BST_B/BST_C/OSC_IN/OSC_OUT/PHASE_B/PHASE_C, cross-checked against `kicad-cli` `tracks_crossing`).

The DRC gate correctly *refuses* such output, so it never ships — but the fact that the A*/negotiated router commits copper that overlaps a different net's copper at all is a router-correctness/quality concern worth a design pass. Normally the grid model should prevent two different-net traces from occupying the same cell; the crossings appear under budget/pushback pressure (partial routes, optimize/pushback moving traces, or grid-cell sharing during negotiation).

## Ask

An architect design pass (`/loom:architect`) on router copper-emission invariants: determine where different-net crossings enter the emitted copper (grid cell-sharing under negotiation? optimize/pushback? partial-net commits?), and design an invariant/guard so the router does not *emit* different-net same-layer overlaps in the first place (rather than relying solely on the post-route DRC gate to reject them). This is the deeper half of the #4159/#4165/#4175/#4176/#4177/#4178 "silently emit or bless invalid copper" family.

## Notes

- NOT a blocker for #4178 (the reporting/gate MVP). This is the router-quality follow-up.
- Reproduction recipe + evidence in #4178's curation comment (issues/4178#issuecomment-4981717364).
- Likely needs an architect, not a direct builder — the invariant touches the C++ A* grid model + the negotiated router.

rjwalters:
## Architect design pass — router copper-emission invariant

[...]

PR #3815, opened 2026-06-20
fix(route): reconcile post-route DRC gate against native kicad-cli

rjwalters:
## Summary

The `kct route` success gate (`run_post_route_drc`) declared **"DRC PASSED"** using only the internal `DRCChecker`, never cross-checking native `kicad-cli pcb drc`. As a result it could report PASS while native KiCad found 400+ violations (incl. real shorts) on the *identical* file. Meanwhile `kct audit` already reconciled against native KiCad via `DesignAuditor._merge_geometric_drc` (#3721). This PR implements the curator's bounded **first slice**: extract that proven reconciliation into a single shared helper and wire it into the route gate so the two entry points cannot drift again.

## Changes

- **New shared helper** `kicad_tools.drc.run_geometric_drc(pcb_path) -> GeometricDRCResult` (`src/kicad_tools/drc/geometric.py`). Runs `kicad-cli pcb drc --severity-error`, summarizes error-severity findings into `{ran, error_count, by_type, note}`, and never raises (every skip path returns `ran=False` + an explanatory note).
- **Refactor `_merge_geometric_drc`** (`src/kicad_tools/audit/auditor.py`) to delegate to the shared helper — behavior-preserving for `kct audit` (the kicad-cli-absent suffix wording is retained).
- **Wire reconciliation into the route gate** (`src/kicad_tools/cli/route_cmd.py:run_post_route_drc`): PASS now requires BOTH the internal engine AND native kicad-cli clean. The native error count is folded into the returned `error_count`, so all four call sites fail the gate automatically.
- **Loud divergence warning** when internal is clean but native finds violations, naming the top native types (e.g. `shorting_items`, `tracks_crossing`).
- **Graceful kicad-cli-absent fallback**: still PASS on a clean internal verdict but print an explicit "internal-engine-only PASS / not authoritative" note so cleanliness is never silently overstated.

## Acceptance Criteria Verification

| Criterion (first slice) | Status | Verification |
|-----------|--------|--------------|
| Board flagged by kicad-cli no longer prints unqualified `DRC PASSED`; verdict reflects native count | ✅ | `test_internal_clean_native_dirty_fails_with_divergence_warning` (errors==52, no "DRC PASSED") |
| Loud divergence WARNING listing top native types + total | ✅ | Same test asserts WARNING + `shorting_items`/`tracks_crossing` named |
| kicad-cli absent → completes with explicit internal-only note | ✅ | `test_kicad_cli_absent_falls_back_to_internal_only_pass` |
[...]

PR #4205, opened 2026-07-15
fix(drc): require coupling evidence for diff-pair clearance skip + add --strict-drc

### LINKED ISSUE #4178: route: emitted copper fails real DRC (tracks crossing, different-net shorts) despite 'full route + optimize + DRC' claim — controlled clean-slate repro
## Bug (router output invalid — controlled, uncontaminated repro)

A single batch route on a **fresh** board (no prior copper, no completion hacks, placement verified conflict-free) produces copper that fails KiCad DRC with genuine electrical errors:

```
$ kct create-pcb <gated .kicad_sch> -o board.kicad_pcb --width 160 --height 100 --layers 4
$ <apply placement; kct placement check: 0 conflicts>
$ PYTHONHASHSEED=0 kct route board.kicad_pcb -o routed.kicad_pcb \
    --layers 4 --skip-nets "GND,+3.3V" --seed 42 --timeout 1200 --targeted-ripup
  ...
  Canonical output: routed.kicad_pcb (full route + optimize + DRC)   <- claim

$ kicad-cli pcb drc --severity-error routed.kicad_pcb
  tracks_crossing:      17     <- same-layer different-net crossings = shorts
  shorting_items:       13
  clearance:             3
  hole_clearance:        3
  copper_edge_clearance: 1
```

Completion on the same run: 22/77 nets (see #4159 for the starvation side). So the emitted 22-net copper — the part the router *did* finish and implicitly blessed — contains ~30 electrical violations. Whatever the internal "optimize + DRC" stage checks, it is not KiCad DRC, and nothing in the output flags the discrepancy.

## Why this is the load-bearing issue

Together with #4159/#4165/#4175/#4176/#4177 this means every stage of the autonomous routing path can silently emit or bless invalid copper: batch route (this issue), per-net completion (#4165), retry hygiene (#4175), the completion oracle (#4176), and stitching (#4177). On the softstart rev-B board the referee gates (`kicad-cli pcb drc --refill-zones` + out-of-process copper-LVS) correctly refused the result at the end — but nothing upstream of the final gate ever said anything was wrong.

## Suggested fixes

1. Run (or faithfully reproduce) KiCad DRC on emitted copper before claiming it; print the violation count in the summary and exit non-zero above a threshold (or add `--strict-drc`).
2. At minimum, rename the claim: "optimize + internal checks" until the internal checker matches kicad-cli results.

Artifacts (fresh-slate board, routed output, drc_batch.json) retained and available. kicad-tools 0.15.0, C++ backend v1.0.0, KiCad 10, macOS; consumer context rjwalters/softstart PR #28.

rjwalters:
## Summary

Two bounded hardening fixes so the `kct route` post-route DRC gate's internal verdict cannot silently diverge from real kicad-cli DRC. The reconciliation machinery (`run_post_route_drc` + `run_geometric_drc`, PR #3815) already exists; this PR closes two specific soft spots identified by the curator. The deeper "router genuinely emits different-net crossings" concern is **deferred to #4202** (architect).

## Changes

**Fix 1 — diff-pair-suffix false-negative in clearance checking** (`validate/rules/clearance.py`)
- `ClearanceRule._build_diff_pair_set()` previously exempted **any** `+`/`-` name-suffixed pair from same-layer segment-segment clearance. That silently skipped genuine short detection between current-sense Kelvin pairs (`/ISENSE_A+` / `/ISENSE_A-`) — `+`/`-` named but NOT differential signals — while kicad-cli (no such heuristic) would flag them.
- Now a suffix-inferred pair is exempted **only** with corroborating evidence: an explicit declaration (`declared_pairs`) **or** geometric coupling (the two nets run parallel and closely spaced over a meaningful cumulative length). New helpers `_segments_are_coupled` / `_pair_is_geometrically_coupled`.
- Genuinely coupled diff pairs (real `USB_D+`/`USB_D-` routed close together) still satisfy the geometric test and stay exempt — no false positive for real diff pairs (verified: board-07's committed diff pairs are unchanged).

**Fix 2 — `--strict-drc` for "kicad-cli didn't run"** (`drc/geometric.py`, `cli/route_cmd.py`, `cli/parser.py`, `cli/commands/routing.py`)
- `run_geometric_drc()` degrades soft (NOTE-only) when kicad-cli is absent / times out / crashes. `--strict-drc` makes "native DRC did not actually run" a HARD FAILURE (non-zero exit) instead of a soft NOTE. Default OFF preserves graceful degradation for KiCad-less environments.
- Registered on BOTH the outer (`parser.py`) and inner (`route_cmd.py`) parsers and forwarded through the `commands/routing.py` shim, per the #4187 inner/outer lesson (guarded by `test_cli_parser_drift.py`).
- `GeometricDRCResult` gains a machine-readable `reason` field (`kicad_cli_absent` / `kicad_cli_timeout` / `kicad_cli_no_report` / `kicad_cli_crash`) so the strict failure message is actionable.

| Suffix-only `+`/`-` pair no longer exempted without corroboration | ✅ | `_build_diff_pair_set` requires coupling or declaration |
| Kelvin sense pair (`ISENSE_A+`/`ISENSE_A-`) with a real same-layer short is flagged | ✅ | `test_kelvin_sense_pair_short_is_flagged`, `..._close_parallel_short_is_flagged` |
| Real coupled diff pair stays exempt (no false positive) | ✅ | `test_real_coupled_diffpair_stays_exempt`; existing `test_clearance_rule_skips_same_pair` still green |
| Explicit declaration exempts even without coupling | ✅ | `test_declared_pair_stays_exempt_even_without_coupling` |
| `--strict-drc` hard-fails when kicad-cli didn't run | ✅ | `test_strict_drc_hard_fails_when_kicad_cli_absent`, `..._on_timeout` |
| Without `--strict-drc`, soft NOTE preserved | ✅ | `test_without_strict_drc_absent_stays_soft_pass` |
| Clean native run under `--strict-drc` still passes | ✅ | `test_strict_drc_clean_native_run_still_passes` |
| `--strict-drc` works through the outer `kct route` entry | ✅ | `test_strict_drc_reaches_inner_through_outer_kct_route` |
| `reason` distinguishes absent/timeout/crash | ✅ | `test_geometric_result_reason_distinguishes_outcomes` |

## Test Plan

```
uv run pytest tests/test_validate_diffpair_clearance_intra.py \
  tests/test_route_post_drc_geometric.py tests/test_cli_parser_drift.py -q
# 48 passed
```
Also green: `test_validate.py`, `test_route_cmd_success_output.py`, `test_audit.py`, `test_drc_constraints_export.py`, and the full `-k "diffpair or clearance"` sweep (1938 passed). Ruff clean + formatted; `check_mypy_baseline.py` reports 0 new type errors.

### Pre-existing failure (out of scope, NOT caused by this PR)

`tests/test_kct_check_parity.py::...::test_board_07_gate_counts_diffpair_and_matchgroup` expects 9 blocking but the committed board-07 artifact now yields 8. This fails **identically on the branch base `bb71056c` with my changes stashed** — it is a pre-existing artifact drift from PR #4204 (dedupe), unrelated to the diff-pair/strict-drc work. My changes leave the board-07 count unchanged (8 with and without them).

Closes #4178

### DIFF
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -360,6 +360,11 @@ def run_route_command(args) -> int:
         sub_argv.append("--high-performance")
     if getattr(args, "skip_drc", False):
         sub_argv.append("--skip-drc")
+    # Issue #4178: forward --strict-drc so a native DRC that did not run
+    # becomes a hard failure.  Defaults off; forward only when set so the
+    # flag-off path stays byte-identical.
+    if getattr(args, "strict_drc", False):
+        sub_argv.append("--strict-drc")
     # Issue #3154: forward the advisory drift-banner flags.  --sync-check
     # defaults on; forward --no-sync-check only when explicitly disabled.
     if getattr(args, "sync_check", True) is False:
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -3297,6 +3297,23 @@ def _add_route_parser(subparsers) -> None:
             "footprints)."
         ),
     )
+    # Issue #4178: hard-gate on native (kicad-cli) DRC actually running.
+    # Mirror of the inner route_cmd.py flag; both sites must stay in sync per
+    # ``tests/test_cli_parser_drift.py``.
+    route_parser.add_argument(
+        "--strict-drc",
+        action="store_true",
+        default=False,
+        help=(
+            "Treat 'native kicad-cli DRC did not run' (kicad-cli absent, "
+            "timed out, crashed, or produced no report) as a HARD FAILURE "
+            "(non-zero exit) instead of a soft NOTE. By default the post-route "
+            "gate degrades gracefully to an internal-engine-only PASS when "
+            "kicad-cli is unavailable, which is not authoritative. Use this in "
+            "CI / manufacturing pipelines that require the native DRC to have "
+            "actually run and passed."
+        ),
+    )
     route_parser.add_argument(
         "--auto-fix",
         action="store_true",
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -1712,6 +1712,7 @@ def run_post_route_drc(
     quiet: bool = False,
     net_class_map: dict | None = None,
     copper_oz: float = 1.0,
+    strict_drc: bool = False,
 ) -> tuple[int, int]:
     """Run DRC validation on the routed PCB.
 
@@ -1731,6 +1732,14 @@ def run_post_route_drc(
             correct value for all 8 demo boards).  Threaded into both the
             internal :class:`DRCChecker` and the emitted ``.kicad_pro`` /
             ``.kicad_dru`` sidecars so both engines judge consistently.
+        strict_drc: When ``True`` (``kct route --strict-drc``), a native
+            geometric DRC that did NOT actually run (kicad-cli absent,
+            timed out, crashed, produced no report) is treated as a HARD
+            FAILURE -- the returned ``error_count`` is bumped by 1 so the
+            route command exits non-zero, and a prominent message states
+            that the PASS is not authoritative.  Default ``False``
+            preserves the graceful-degradation soft NOTE for KiCad-less
+            environments (Issue #4178).
 
     Returns:
         Tuple of (error_count, warning_count)
@@ -1792,7 +1801,34 @@ def run_post_route_drc(
             # the verdict so PASS requires BOTH engines clean.
             error_count += geo.error_count
 
-        if not quiet:
+        # Issue #4178: under --strict-drc, "kicad-cli DRC did not actually
+        # run" is a HARD FAILURE, not a soft NOTE.  A clean internal verdict
+        # is not authoritative when the native engine never ran, so bump the
+        # error count to force a non-zero exit.  The soft-degradation default
+        # (strict_drc=False) preserves the internal-only PASS for KiCad-less
+        # environments.
+        strict_drc_failed = strict_drc and not geo.ran
+        if strict_drc_failed:
+            error_count += 1
+
+        if not quiet and strict_drc_failed:
+            print("\n--- DRC Validation ---")
+            reason_text = geo.note or "native geometric DRC did not run"
+            print(f"  STRICT DRC FAILURE: native kicad-cli DRC did not run ({reason_text}).")
+            print(
+                "    The internal-engine verdict is NOT authoritative; "
+                "--strict-drc requires kicad-cli DRC to run and pass."
+            )
+            print(
+                "    Re-run without --strict-drc to allow a soft internal-only "
+                "PASS, or install/repair kicad-cli."
+            )
+            if results.error_count > 0 or warning_count > 0:
+                # Also surface any internal findings below the strict banner.
+                print(f"  Internal errors: {results.error_count}")
+                if warning_count > 0:
+                    print(f"  Internal warnings: {warning_count}")
+        elif not quiet:
             print("\n--- DRC Validation ---")
             if error_count == 0 and warning_count == 0:
                 if geo.ran:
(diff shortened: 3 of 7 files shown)


rjwalters:
## ✅ Approved!

Two well-bounded DRC-gate hardening fixes for #4178. Both fixes match the curated acceptance criteria; all 16 CI checks pass and merge state is CLEAN.

### Fix 1 — diff-pair-suffix false-negative (`clearance.py`)

- `_build_diff_pair_set` no longer exempts a pair on bare `+`/`-` name inference. Exemption now requires corroborating evidence: an explicit `declared_pairs` entry **or** geometric coupling (`_pair_is_geometrically_coupled` / `_segments_are_coupled`: parallel within 20°, edge-to-edge gap ≤ 0.5 mm, cumulative ≥ 1.0 mm). API usage verified against `PCB.copper_layers` / `segments_on_layer` and `Segment.{start,end,width,net_number}`.
- **Both directions covered and passing:**
  - Under-flag guard: `test_kelvin_sense_pair_short_is_flagged` (crossing ISENSE_A+/A- short is now FLAGGED) + `..._close_parallel_short_is_flagged` (0.5 mm parallel run, below the 1.0 mm floor, correctly fires).
  - Over-flag guard: `test_real_coupled_diffpair_stays_exempt` (USB_D+/USB_D- coupled 2 mm stays EXEMPT); existing `test_clearance_rule_skips_same_pair` still green.
- Coupling threshold scrutiny: the 1.0 mm cumulative-parallel floor is a sound discriminator — a Kelvin pair that briefly parallels near a shared connector stays under the floor (explicitly tested at 0.5 mm), while a routed diff pair maintains coupling well past it. The `declared_pairs` path (unused by the current caller, wired as a future hook) is the correct escape valve for a real diff pair with only a short coupled section.
- **Board-07 diff pairs stay exempt** — confirmed: the committed board-07 blocking count is unchanged (8 with and without this PR's changes). Diff-Pair + Match-Group routing regression CI jobs both pass.

### Fix 2 — `--strict-drc` (`geometric.py`, `route_cmd.py`, `parser.py`, `commands/routing.py`)

- "kicad-cli DRC did not run" (`geo.ran == False`) becomes a hard non-zero exit under `--strict-drc`; default off preserves the soft NOTE for KiCad-less environments. The strict banner is prominent (own `--- DRC Validation ---` block), not a NOTE sub-line.
- Registered on **both** the outer (`parser.py`) and inner (`route_cmd.py`) parsers, forwarded through the `commands/routing.py` shim, and threaded into **all four** `run_post_route_drc` call sites (`main`, layer/rule/combined escalation) — the #4187 inner/outer lesson. `test_strict_drc_reaches_inner_through_outer_kct_route` genuinely exercises the outer entry: it parses via `create_parser()`, runs `run_route_command`, and asserts the flag reaches inner argv AND parses onto the inner namespace as `strict_drc=True`.
- `GeometricDRCResult.reason` machine-readable field distinguishes absent/timeout/no-report/crash; verified by `test_geometric_result_reason_distinguishes_outcomes`.

### Verification

- 48 targeted tests pass (`test_validate_diffpair_clearance_intra.py`, `test_route_post_drc_geometric.py`, `test_cli_parser_drift.py`).
- Ruff clean + formatted on all changed files; **0 new mypy errors** in the changed files (CI Type Check passed).
- All 16 CI checks green including full Test suite, Board 05/06/07 E2E, Diff-Pair & Match-Group routing regressions. Merge state CLEAN.
- Scope correctly bounded: deeper router-crossing concern deferred to #4202.
- PR description uses `Closes #4178`.

### Board-07 parity failure — independently verified as pre-existing, NOT a blocker

The builder's claim checks out under rigorous independent verification. `tests/test_kct_check_parity.py::TestCiGateCountsGatedFamilies::test_board_07_gate_counts_diffpair_and_matchgroup` (expects 9 blocking, gets 8) fails **identically** at:
- `origin/main` @ `bb71056c` (post-#4204),
- the parent of #4204 @ `9bdd5a49` (pre-dedupe),
- and this PR branch.

So it **predates #4204** — the builder's "introduced by #4204's dedupe" attribution is slightly off (it's older), but the load-bearing conclusion is correct: **this PR does not cause it** (8 with and without the change). It's also `@pytest.mark.slow`, so it's excluded from the PR CI Test job (`-m "not slow"`) and only runs nightly — which is why main's CI is green.

On honest-shift vs regression: `kct pcb dedupe --dry-run` on the committed artifact reports **0 duplicate copper**, so it is not a dedupe consequence. The artifact's last change (#4045) was a pure 3D-model-offset metadata edit ("every pad/segment/via/zone unchanged"), so copper geometry is byte-identical to the #4005 baseline that set `== 9`. The 9→8 delta is therefore a **gate/clearance code-path counter shift** (most likely a legitimate double-count collapse → honest re-baseline to 8), pending a confirming bisect. Filed as **#4207** to track the bisect + re-baseline separately from this PR.

PR #4206, opened 2026-07-15
fix(router): widen finalize demote to full DRC SHORT threshold (#4202 Unit 1)

rjwalters:
## Summary

Unit 1 (the M1 fix) of #4202. The negotiated/A* router **emits** same-layer different-net crossing copper that the post-route DRC gate then rejects, wasting the route. This closes the primary mechanism.

The post-loop seg-seg demote safety net (`_demote_seg_seg_overlap_nets` in `core.py`, and its mirror in `two_phase._detailed_negotiated`) called `find_segment_segment_violation_pairs(copper_overlap_only=True)` — a **polygon-intersection** threshold `(w_a+w_b)/2`. But `kicad-cli` fires `clearance_segment_segment: SHORT` at `(w_a+w_b)/2 + trace_clearance`. A committed different-net pair whose edge-to-edge gap is **positive but < `trace_clearance`** is a real DRC short that fell **below** the demote threshold and shipped (the architect measured grid-snap marginals landing exactly on the 0.1000-vs-0.1016mm boundary).

## Fix

Add a `finalize=True` mode to the demote that uses the **full DRC SHORT threshold** (`copper_overlap_only=False`), applied at **FINALIZE time** — after `_post_route_clearance_correction` has had its repair opportunity, so genuinely-nudgeable near-misses are already fixed and only unrepaired-and-shipping copper remains. The negotiation loop itself is unchanged (transient sub-clearance is expected there and repaired by iteration — the M2 relief-probe/negotiation dynamics are out of scope). Applied to **both** the `core.py` demote and the `two_phase.py` mirror.

The trade (architect-endorsed, #3433): a net that today ships a short becomes **honestly-unrouted** (advisory connectivity finding) — strictly better. Verified the demoted net is reset to `[]` (reported incomplete), not silently dropped.

## Over-demotion safety (coupled diff pairs)

A tightly-coupled diff pair has intra spacing < `trace_clearance`, so the widened threshold *would* flag it in isolation — but both legs are **prerouted** (they live in `self.routes`, never in the demotable `net_routes`), so they enter `find_segment_segment_violation_pairs` as non-rippable `extra_routes`, and the existing extra-vs-extra skip never flags them. No diff-pair carve-out needed in Unit 1.

## Tests

- `TestDemoteSegSegFinalizeThreshold` (new, `tests/router/test_seg_seg_quality_tuple.py`):
  - sub-clearance different-net pair (edge gap 0.106mm < 0.15mm clearance) → **NOT** demoted in-loop (correction gets first shot), **demoted** at finalize → honestly-unrouted; partner survives.
  - edge gap == `trace_clearance` (clean) → NOT demoted at finalize (no over-demotion).
  - full physical overlap (g < 0, board-04 mode) → still demoted at finalize (superset; no regression of the #3433 catch).
  - coupled diff pair (both legs extra) → NOT flagged even at finalize.
- Existing timeout-ordering test extended to assert the post-correction demote runs with `finalize=True` (i.e. after repair).
- Regressions (all pass): `tests/router/test_seg_seg_quality_tuple.py`, board-04 routing (2 passed, still >=8/9 clean — validates the two_phase mirror), board-05 routing (2 passed, >= floors), diff-pair routing integration / coupled-cpp-parity / engagement, post-negotiation sweep, partial-route fastfail.

## Board regression completion counts

No baseline shift: board-04 uses a `>=8/9` floor and board-05 uses `>=` completion-pct / `>=15` nets floors — both robust to the honest-unroute trade and both passed unchanged. No exact-count baseline needed adjusting.

## Verification checklist

- [x] Sub-clearance different-net short now demoted at finalize (was shipping).
- [x] Clean routes (gap >= clearance) unaffected.
- [x] Repaired near-misses unaffected (widened check runs AFTER correction).
- [x] `ruff check` + `ruff format --check` clean on changed files.
- [x] `check_mypy_baseline.py`: 0 new errors (pre-existing 39-error floor drift confirmed identical on clean main; baseline untouched).
- [x] C++ backend built (`kct build-native --check` → available 1.0.0).

## Scope

Unit 1 (M1) only. **Deferred follow-ups** per the architect design pass (issues/4202#issuecomment-4982885099): Unit 2 (M3 — unify the four `_demote_*` passes + rescue gate into one `_committed_copper_blocking_drc`), Unit 3 (M4 — post-optimize/post-nudge rtree-less backstop in `route_cmd.py`), Unit 4 (C++ negotiation hardening — explicitly higher-risk, do-not-start).

### DIFF
--- a/src/kicad_tools/router/algorithms/two_phase.py
+++ b/src/kicad_tools/router/algorithms/two_phase.py
@@ -1226,20 +1226,29 @@ class TwoPhaseRouter:
             net_routes.clear()
             net_routes.update(best_net_routes)
 
-        # Issue #3433 (safety net): never commit physically-overlapping
-        # copper.  Even with seg-seg violators in the comparator and the
-        # mid-iteration rip-up feed, the loop can exit (stagnation,
-        # early-stop, timeout) with every observed snapshot containing
-        # cross-net same-layer FULL overlaps.  Overlapping copper is an
-        # unmanufacturable hard-gate DRC failure; an unrouted net is an
-        # advisory connectivity finding -- enforce the trade in the
-        # right direction by demoting the greedy-cover victims to
-        # unrouted rather than committing them.
+        # Issue #3433 / #4202 (safety net): never commit copper that
+        # would be a cross-net DRC short.  Even with seg-seg violators in
+        # the comparator and the mid-iteration rip-up feed, the loop can
+        # exit (stagnation, early-stop, timeout) with every observed
+        # snapshot containing cross-net same-layer FULL overlaps OR
+        # sub-clearance grazing pairs (positive edge gap but
+        # ``< trace_clearance`` -- a real ``kicad-cli SHORT`` that
+        # polygon-intersection misses).  Both are hard-gate DRC failures
+        # the post-route gate rejects; an unrouted net is an advisory
+        # connectivity finding -- enforce the trade in the right
+        # direction by demoting the greedy-cover victims to unrouted
+        # rather than committing them.  This is the FINALIZE position
+        # (loop has exited, negotiation's iterative repair is done), so
+        # ``copper_overlap_only=False`` widens the threshold to the FULL
+        # DRC SHORT class ``(w_a+w_b)/2 + trace_clearance`` -- mirroring
+        # the ``finalize=True`` path in ``core._demote_seg_seg_overlap_nets``.
+        # The wide threshold MUST stay out of the negotiation loop, where
+        # transient sub-clearance is expected and repaired by iteration.
         _overlap_pairs = neg_router.find_segment_segment_violation_pairs(
             net_routes,
             trace_clearance=self.rules.trace_clearance,
             extra_routes=self._collect_extra_routes_for_revalidation(net_routes),
-            copper_overlap_only=True,
+            copper_overlap_only=False,
         )
         if _overlap_pairs:
             _demotable = {n for n, r in net_routes.items() if r}
@@ -1254,8 +1263,8 @@ class TwoPhaseRouter:
                     net_routes[_net] = []
                 _victim_names = [self.net_names.get(n, f"Net_{n}") for n in _victims]
                 flush_print(
-                    f"  ⚠ Demoted {len(_victims)} net(s) with physically "
-                    f"overlapping copper to unrouted: {', '.join(_victim_names)}"
+                    f"  ⚠ Demoted {len(_victims)} net(s) with cross-net "
+                    f"shorting copper to unrouted: {', '.join(_victim_names)}"
                 )
 
         # Issue #2597: Surface the iteration-loop exit reason to the caller
--- a/src/kicad_tools/router/core.py
+++ b/src/kicad_tools/router/core.py
@@ -10557,18 +10557,24 @@ class Autorouter:
                     f"({total_elapsed:.1f}s)"
                 )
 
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!** — Router-core review with rigor. Faithful, minimal Unit 1 implementation of the #4202 architect design (design-pass comment authoritative).

## Verified against the design

**Widening + call site (core.py)**
- `_demote_seg_seg_overlap_nets` gains `finalize: bool = False`; `copper_overlap_only=not finalize` (core.py:11265). The single call site in `route_all_negotiated` (core.py:10577) passes `finalize=True`.
- **Ordering confirmed**: the finalize demote (10577) runs AFTER `_post_route_clearance_correction` (10545) — repairable near-misses get first shot; only unrepaired-and-shipping copper reaches the wide check. Matches the design's FINALIZE requirement.
- **In-loop path unchanged**: default `finalize=False` preserves `copper_overlap_only=True`. `test_sub_clearance_short_demoted_at_finalize` asserts in-loop mode does NOT demote the sub-clearance pair. M2 negotiation dynamics untouched.

**two_phase.py mirror**
- The widened `copper_overlap_only=False` block (two_phase.py:1247) sits below the best-vs-final restore (1217-1223) with no enclosing `while` between — genuinely the post-loop FINALIZE position, mirroring core.py.

**Threshold matches kicad-cli, not stricter**
- `find_segment_segment_violation_pairs` uses `required = (w_a+w_b)/2 + trace_clearance` (negotiated.py:2082-2084) — exactly the SHORT class. Comparison is `dist < required - eps` with `eps=1e-6` (2043/2099), so geometry AT exactly clearance is NOT flagged. `test_clean_gap_not_demoted_at_finalize` (gap == trace_clearance) confirms no boundary over-demotion.

**Four behaviors — structurally verified + tested**
- (a) sub-clearance short demoted at finalize, NOT in-loop → honestly-unrouted (`net_routes[net]=[]`, removed from grid/`self.routes`, reported via `demoted` + `flush_print`, not silently dropped).
- (b) clean gap (>= clearance) not over-demoted.
- (c) hard polygon overlap (g<0) still demoted — superset preserved, #3433 catch not regressed.
- (d) coupled diff pair not flagged: both prerouted legs enter as non-rippable `extra_routes` → extra-vs-extra skip (negotiated.py:2077-2078). No diff-pair-intra carve-out needed in Unit 1.

**No cascade / no infinite loop**
- `select_seg_seg_demotion_nets` is a finite greedy vertex cover over `demotable` only (net_routes with routes); prerouted extra routes can never be demoted; each iteration drops all pairs containing the chosen victim (negotiated.py:321-333). Terminal finalize step — not re-run.

## CI (all 16 pass, mergeStateStatus CLEAN)

The two regression-risk gates I flagged for the honest-unroute trade both passed empirically:
- **Board 05 Routing Regression (blocking <= 11)** — PASS (19m22s). The shipped-short→unrouted trade did not breach the blocking ceiling.
- **Diff-Pair Routing Regression (board-06)** PASS + **Match-Group Routing Regression (board-07)** PASS — no over-demotion of coupled pairs in practice.
- Boards 00-07 end-to-end, Test, Lint & Format, Type Check, C++ Build — all PASS.

Scope correctly limited to Unit 1; Units 2/3/4 deferred per the design pass. New `TestDemoteSegSegFinalizeThreshold` (4 cases) + the timeout-ordering test extended to assert `finalize=True` after repair.

Issue #4185, opened 2026-07-15
create-pcb quotes keepout enum tokens ((tracks "not_allowed")) — emitted board unloadable by KiCad


rjwalters:
## Summary

`kct create-pcb` re-serializes footprint-embedded keepout **rule areas** with their enum tokens quoted — `(tracks "not_allowed")` instead of the bare `(tracks not_allowed)` that KiCad emits and expects. The resulting `.kicad_pcb` **cannot be loaded by KiCad at all**: `kicad-cli pcb drc` fails with `Failed to load board`, which kills the mandatory cross-gate (consumer CLAUDE.md convention 2) for any board whose schematic pulls in such a footprint.

`kct check` parses the same board fine, so the sick file *looks* healthy from inside the toolchain — same "success message that doesn't verify what it announces" theme as #4159/#4165/#4175–#4178: the emission stage reports `pcb_created: True` for a board the target EDA cannot open.

## Repro (kicad-tools 0.15.0, KiCad 9/10 stock libs)

Any schematic that instantiates `RF_Module:ESP32-C3-WROOM-02` (its footprint embeds an antenna-keepout rule area). Then:

```bash
kct create-pcb board.kicad_sch -o out.kicad_pcb --width 160 --height 100 --layers 4
kicad-cli pcb drc -o /dev/null out.kicad_pcb
# -> "Failed to load board"
```

## Evidence

Library footprint (`RF_Module.pretty/ESP32-C3-WROOM-02.kicad_mod`):

```
(keepout
    (tracks not_allowed)
    (vias not_allowed)
    (pads not_allowed)
    (copperpour not_allowed)
    (footprints not_allowed)
)
```

`create-pcb` emission of the same block:

```
(keepout (tracks "not_allowed") (vias "not_allowed") (pads "not_allowed") (copperpour "not_allowed") (footprints "not_allowed")
)
```

pcbnew's parser requires the bare symbol here; the quoted form is a hard parse error (hence "Failed to load board" with no diagnostics).

## Bisect

Board loads after removing only the ESP32-C3-WROOM-02 footprint block; all other new-part footprints (HLK-PMxx, TSOT-23-6, SOT-223, NR-40xx, TR5 fuseholder, pin headers) are innocent. Minimal fix confirmed end-to-end:

```python
re.sub(r'\((tracks|vias|pads|copperpour|footprints) "(not_allowed|allowed)"\)', r'(\1 \2)', board_text)
[...]

PR #4210, opened 2026-07-15
fix(pcb): net every same-numbered pad in assign_net_to_footprint_pad

### LINKED ISSUE #4186: create-pcb nets only the first of same-numbered multi-pads — EP/thermal groups left <no net> (shorting_items x21)
## Summary

`kct create-pcb` assigns the schematic net to only the **first** pad of a same-numbered multi-pad group. Footprints that implement a thermal/EP pad as many same-numbered sub-pads (e.g. `RF_Module:ESP32-C3-WROOM-02` — pad "19" is 1 SMD paste pad + 21 PTH thermal-via sub-pads, all numbered 19) come out with one netted pad and the rest `<no net>`, producing 21× `shorting_items` DRC errors on the untouched board and — worse — a copper plane that will *connect* to the unnetted sub-pads at fill time while the netlist says they're nothing.

## Repro (kicad-tools 0.15.0)

Schematic with `RF_Module:ESP32-C3-WROOM-02`, pin 19 on GND:

```bash
kct create-pcb board.kicad_sch -o out.kicad_pcb --layers 4
kicad-cli pcb drc --format json -o drc.json out.kicad_pcb   # (after #4185 dequote workaround)
# -> 21x: shorting_items :: PTH pad 19 [GND] of U14 | PTH pad 19 [<no net>] of U14
```

Emitted board: `grep -c '(pad "19"' out.kicad_pcb` → 22, of which only 1 carries `(net N "GND")`.

## Expected

KiCad semantics: all pads sharing a number belong to the same logical pin and must carry the same net (this is exactly how EP/thermal-via footprints are modeled throughout the stock libraries — QFN EPs, RF module heatsink grids, DPAK tabs).

## Workaround (consumer-side)

```python
# assign the group's net to every unnetted same-numbered sub-pad
for pad in footprint.pads_named("19"):
    if "(net " not in pad: insert(f'(net {gnd} "GND")')
```

## Fix tiers

1. **Minimal**: when assigning nets in create-pcb, key by (footprint, pad-number) and apply to every pad with that number, not just the first match.
2. **Right**: property test over stock libs — every footprint whose pad numbers repeat (hundreds: QFN/DFN EPs, modules) must emit all-or-none netted groups.
3. **Belt**: post-emission invariant check: no two same-numbered pads of one footprint with different net fields; fail loudly instead of emitting a silent short.

Found in the same rev-C session as #4185 (rjwalters/softstart#29); same board, same footprint.

rjwalters:
## Summary

`kct create-pcb` assigned the schematic net to only the **first** pad of a same-numbered multi-pad group. Footprints that model an EP / thermal pad as many same-numbered sub-pads (e.g. an RF-module heatsink grid where pad `"19"` is 1 SMD paste pad + 21 PTH thermal vias, all numbered `19`) came out with one netted pad and the rest `<no net>`, producing `shorting_items` DRC errors and — worse — a copper plane that connects to sub-pads the netlist says are nothing.

Root cause: `PCB.assign_net_to_footprint_pad` (`src/kicad_tools/schema/pcb.py`) stopped at the first match in **both** loops — the in-memory `Pad` loop (`break`) and the S-expression loop (`return True`). This fix applies the pin's net to **every** pad sharing the number in both loops, keeping the `Pad` objects and the serialized tree consistent.

## Changes

- `src/kicad_tools/schema/pcb.py`: `assign_net_to_footprint_pad` — dropped the in-memory `break`; set net on every `Pad` where `pad.number == pad_number`. Dropped the sexp `return True`; strip/replace `(net ...)` on every matching `pad` node and return `True` only after the full pass (still `False` if none matched, preserving current failure semantics).
- `tests/test_workflow.py`: added 4 regression tests to `TestPCBNetAssignment` using a synthetic inline footprint fixture.

| 1. Synthetic footprint fixture (unique `"1"`, 3-pad `"19"` smd+thru_hole group, empty-number NC), loaded via `add_footprint_from_file`, run through `assign_net_to_footprint_pad` | ✅ | `_write_multipad_footprint` writes the fixture inline; no ESP32 stock lib, no kicad-cli, no #4185 dependency |
| 2. Every `"19"` pad carries same net (in-memory + sexp) | ✅ | `test_multipad_group_all_pads_get_net_in_memory` checks all 3 `fp.pads`; `test_multipad_group_all_pads_get_net_in_sexp` greps all 3 `(pad "19")` sexp nodes for `(net N "GND")` |
| 3. Unique `"1"` pad assigned once, unaffected | ✅ | `test_multipad_unique_pad_unaffected` |
| 4. Empty-number NC pad stays unnetted (no `(net ...)`, `net_number == 0`) | ✅ | `test_multipad_empty_number_pad_stays_unnetted` (both Pad object and sexp) |
| 5. Return value / stats shape unchanged (one assignment per group) | ✅ | `assign_net_to_footprint_pad` still returns a single bool; `assign_nets_from_netlist` caller loop untouched; existing tests unmodified |
| 6. Regression: `TestPCBNetAssignment` + `tests/test_pin_to_pad_mapping.py` pass unchanged | ✅ | 47 passed across both files; no assertion changes |

## Test Plan

- `uv run pytest tests/test_workflow.py -k TestPCBNetAssignment` → 6 passed (2 pre-existing + 4 new)
- `uv run pytest tests/test_workflow.py tests/test_pin_to_pad_mapping.py` → 47 passed
- `uv run pytest tests/ -k "create_pcb or assign_net or net_assignment or multipad"` → 18 passed, 6 skipped
- `ruff check` + `ruff format --check` clean on both files
- `scripts/ci/check_mypy_baseline.py` → 0 new type errors (1475 within 1514 baseline)

Closes #4186

### DIFF
--- a/src/kicad_tools/schema/pcb.py
+++ b/src/kicad_tools/schema/pcb.py
@@ -5036,14 +5036,22 @@ class PCB:
         # Ensure net exists and get its number
         net = self.add_net(net_name)
 
-        # Update the in-memory pad
+        # Update every in-memory pad sharing this number.
+        #
+        # KiCad models EP/thermal groups (QFN exposed pads, RF-module
+        # heatsink grids, DPAK tabs) as many pads that share a single pad
+        # number. All pads with the same number are one logical pin and must
+        # carry the same net, so we assign to every match rather than stopping
+        # at the first (see issue #4186). Empty/blank pad numbers are never
+        # passed here for a real schematic pin, and grouping only ever keys on
+        # a specific non-empty number, so unrelated unnumbered mechanical pads
+        # are never netted together.
         pad_found = False
         for pad in fp.pads:
             if pad.number == pad_number:
                 pad.net_number = net.number
                 pad.net_name = net.name
                 pad_found = True
-                break
 
         if not pad_found:
             return False
@@ -5069,7 +5077,12 @@ class PCB:
             if ref_value != reference:
                 continue
 
-            # Found the footprint, now find the pad
+            # Found the footprint, now update every pad sharing this number.
+            # Same rationale as the in-memory loop above (issue #4186): all
+            # pads with a given number are one logical pin and must all carry
+            # the net, so we visit every match instead of returning on the
+            # first one.
+            sexp_pad_found = False
             for pad_sexp in fp_sexp.find_all("pad"):
                 if pad_sexp.get_string(0) == pad_number:
                     # Remove existing net node if present
@@ -5080,7 +5093,9 @@ class PCB:
                     # Add new net node
                     new_net_node = SExp.list("net", net.number, net.name)
                     pad_sexp.append(new_net_node)
-                    return True
+                    sexp_pad_found = True
+
+            return sexp_pad_found
 
         return False
 
--- a/tests/test_workflow.py
+++ b/tests/test_workflow.py
@@ -107,6 +107,158 @@ class TestPCBNetAssignment:
         assert "U1" in result["missing_footprints"]
         assert len(result["assigned"]) == 0
 
+    @staticmethod
+    def _write_multipad_footprint(tmp_path: Path) -> Path:
+        """Write a synthetic footprint with a same-numbered multi-pad group.
+
+        Models the EP/thermal-via pattern (issue #4186):
+        - pad "1": a single unique SMD pad
+        - pad "19": a thermal group of 3 pads sharing the number, mixing
+          smd (paste) and thru_hole (PTH via) types like a real paste+PTH EP
+        - pad "": a mechanical / no-connect pad with an empty number that must
+          never be netted
+        """
+        footprint = """\
+(footprint "Multipad_Test"
+	(version 20240108)
+	(generator "pcbnew")
+	(layer "F.Cu")
+	(attr smd)
+	(pad "1" smd roundrect
+		(at -2 0)
+		(size 1 1)
+		(layers "F.Cu" "F.Paste" "F.Mask")
+		(roundrect_rratio 0.25)
+	)
+	(pad "19" smd rect
+		(at 0 0)
+		(size 2 2)
+		(layers "F.Cu" "F.Paste" "F.Mask")
+	)
+	(pad "19" thru_hole circle
+		(at -0.5 -0.5)
+		(size 0.6 0.6)
+		(drill 0.3)
+		(layers "*.Cu")
+	)
+	(pad "19" thru_hole circle
+		(at 0.5 0.5)
+		(size 0.6 0.6)
+		(drill 0.3)
+		(layers "*.Cu")
+	)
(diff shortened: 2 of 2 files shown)


rjwalters:
✅ **Approved!**

Clean, well-scoped fix for #4186. Both loops in `assign_net_to_footprint_pad` now apply the net to **every** pad sharing the number instead of stopping at the first match, keeping the in-memory `Pad` objects and the serialized S-expression tree consistent.

## Correctness review

- **In-memory loop** (`pcb.py:5041`): `break` dropped; `net_number`/`net_name` set on every `pad.number == pad_number`, `pad_found` still gates the early `return False` when nothing matched. Failure semantics preserved.
- **Sexp loop** (`pcb.py:5086`): `return True` replaced with `sexp_pad_found = True` and a post-loop `return sexp_pad_found`. Verified the `return` is at the footprint-loop indent (12 spaces) — it fires only after the matching footprint's pads are all processed, then exits (does not walk further footprints). The trailing `return False` (8-space indent) still covers the no-footprint-matched case.
- **Same predicate in both loops** (`== pad_number`, exact equality) ⇒ identical pad set ⇒ in-memory and sexp stay consistent. No pad netted in one but not the other.
- **Different pad numbers**: netting `"19"` uses exact equality, so `"1"` / `"20"` are never touched. Confirmed by `test_multipad_unique_pad_unaffected`.
- **Empty-number NC/mechanical pads**: callers only ever pass a non-empty schematic pin number, so `"" != "19"` and the NC pad is never matched — no accidental grouping of unrelated blank-numbered pads. Confirmed by `test_multipad_empty_number_pad_stays_unnetted` (both `net_number == 0` and no `(net ...)` sexp node).
- **Return-value shape unchanged**: still a single `bool` per `(ref, pin)`; `assign_nets_from_netlist` caller loop and `workflow.assign_nets()` untouched, so `stats["assigned"]` is not fanned out. Existing tests pass with no assertion changes.

## Test Execution

- `pytest tests/test_workflow.py tests/test_pin_to_pad_mapping.py` — ✅ **47 passed** (matches claim; 4 new regression tests + existing suite, no assertion changes)
- `ruff check` + `ruff format --check` on `pcb.py` + `test_workflow.py` — ✅ clean
- CI: all 16 checks **pass** — Test (11m33s), Type Check (mypy baseline, 0 new), Lint, C++ Build, Boards 00–07, Diff-Pair / Match-Group / Board-05 routing regressions, kicad-cli smoke, Routed DRC.
- `mergeStateStatus` — ✅ **CLEAN**

The 4 new tests exercise the fixture inline (no ESP32 stock lib, no kicad-cli, no #4185 dependency), covering the unique-`"1"`, 3-pad `"19"` smd+thru_hole group, and empty-number NC cases across both the in-memory `fp.pads` list and the re-serialized sexp tree. Solid coverage of the exact hazard the issue describes.

PR #4212, opened 2026-07-15
fix(sexp): preserve bare atoms symmetrically so keepout enums round-trip unquoted

### LINKED ISSUE #4185: create-pcb quotes keepout enum tokens ((tracks "not_allowed")) — emitted board unloadable by KiCad

rjwalters:
## Summary

`SExp._format_atom()` was a one-way quoting ratchet: it honored the parse-time `_originally_quoted` flag to *add* quotes on serialize, but never used the inverse (atom parsed **bare**) to *skip* the `_needs_quoting()` keyword-allowlist fallback. Bare KiCad enum symbols that are absent from the `unquoted_keywords` allowlist — `not_allowed`/`allowed` and the sibling tokens `chamfer_rect`, `chamfer`, `fillet`, `poly`, `castellated`, `heatsink`, `bga` — hit the "quote everything else" default and were re-emitted quoted. So a footprint's `(tracks not_allowed)` keepout rule area became `(tracks "not_allowed")`, which pcbnew hard-rejects: `kicad-cli pcb drc` reports `Failed to load board`, making the emitted board unloadable by KiCad (surfaced via `kct create-pcb` for footprints like `RF_Module:ESP32-C3-WROOM-02`).

This applies the curator's "Right" tier fix: respect the parser's bare/quoted distinction **symmetrically**, which fixes the keepout tokens and all the sibling tokens at once rather than special-casing keepout or patching the allowlist token-by-token.

## Changes

- `src/kicad_tools/sexp/parser.py`:
  - Add an `_originally_bare` slot/flag (symmetric counterpart to `_originally_quoted`), set in `Parser._parse_atom()` for bare string tokens.
  - Add `SExp._must_quote()` — the narrow "genuinely requires quotes" test (empty string, whitespace, parens, or quote char) that must always force quoting for correctness, distinct from the KiCad-keyword allowlist heuristic.
  - In `_format_atom()`: an atom parsed bare stays bare unless `_must_quote()` is true. `_originally_quoted` still wins (a value parsed quoted stays quoted), and programmatically constructed atoms (`_originally_bare=False`) still use the `_needs_quoting()` allowlist heuristic, so existing keyword/round-trip behavior is unchanged.
- `tests/test_sexp.py`: 4 unit tests — parsed-bare keepout enums stay bare, sibling tokens stay bare, originally-quoted strings stay quoted, and values that structurally require quotes (spaces, parens, empty) are always quoted.
- `tests/test_sexp_roundtrip.py`: a synthetic minimal keepout footprint fixture (no stock-library dependency); a round-trip test asserting bare tokens; and a `kicad-cli`-gated test injecting the keepout footprint into a real demo board and asserting the serialized board loads (no "Failed to load board").

| Parsed-bare `(tracks not_allowed)` re-emits bare | ✅ | `test_parsed_bare_keepout_enum_stays_bare`; direct parser repro no longer quotes |
| Keepout block round-trips bare | ✅ | `test_keepout_footprint_roundtrips_bare` (synthetic fixture) |
| Originally-quoted string stays quoted | ✅ | `test_parsed_quoted_string_stays_quoted` (`generator_version "9.0"`) |
| Values requiring quotes (spaces/parens/empty) still quoted | ✅ | `test_parsed_bare_value_requiring_quotes_still_quoted` |
| Sibling tokens (chamfer_rect, fillet, poly, ...) stay bare | ✅ | `test_parsed_bare_sibling_enum_tokens_stay_bare` — fix generalizes, no special-casing |
| Emitted board loads in KiCad | ✅ | `test_keepout_board_loads_in_kicad` (kicad-cli-gated); confirmed the quoted form fails and the bare form loads |
| No serialization regressions (high blast radius) | ✅ | `test_serialize_kicad_keywords_unquoted`, `test_mirror_values_unquoted`, and the full broad sweep all pass unchanged |

## Test Plan

- `uv run pytest tests/test_sexp.py tests/test_sexp_roundtrip.py -q` → 90 passed, 5 skipped (pre-existing skips from a stale demo-fixture path).
- Broad regression sweep `uv run pytest tests/ -q -k "sexp or serialize or roundtrip or create_pcb or footprint"` → **1526 passed, 23 skipped, 0 failed** — zero serialization regressions.
- `uv run --extra dev python scripts/ci/check_mypy_baseline.py` → 0 new type errors (1475 within 1514 baseline).

Closes #4185

### DIFF
--- a/src/kicad_tools/sexp/parser.py
+++ b/src/kicad_tools/sexp/parser.py
@@ -69,6 +69,7 @@ class SExp:
         "_inline",
         "_original_str",
         "_originally_quoted",
+        "_originally_bare",
         "_line",
         "_column",
     )
@@ -81,6 +82,7 @@ class SExp:
         _inline: bool = False,
         _original_str: str | None = None,
         _originally_quoted: bool = False,
+        _originally_bare: bool = False,
         _line: int = 0,
         _column: int = 0,
     ):
@@ -98,6 +100,16 @@ class SExp:
         # constructed atoms default to False so existing keyword/round-trip
         # behavior is unchanged.
         self._originally_quoted = _originally_quoted
+        # Tracks whether a string atom was parsed from a *bare* (unquoted) token.
+        # When True, the serializer preserves the bare form as long as the value
+        # does not *structurally* require quoting (whitespace, parens, quote
+        # chars, empty). This is the symmetric counterpart to _originally_quoted:
+        # it lets bare KiCad enum symbols that are absent from the unquoted
+        # keyword allowlist (e.g. `not_allowed`, `chamfer_rect`) round-trip bare
+        # rather than being wrongly quoted (issue #4185). Programmatically
+        # constructed atoms default to False so they still go through the
+        # allowlist heuristic in _needs_quoting().
+        self._originally_bare = _originally_bare
         self._line = _line
         self._column = _column
 
@@ -490,7 +502,21 @@ class SExp:
             # preserves the type distinction between (field "9.0") and
             # (field 9.0) on round-trip — KiCad strict-typed string fields
             # like generator_version would otherwise be silently downgraded.
-            if self._originally_quoted or self._needs_quoting(self.value):
+            #
+            # Symmetric bare handling (issue #4185): an atom parsed as a *bare*
+            # symbol must stay bare unless it *structurally* requires quoting
+            # (whitespace, parens, quote chars, empty). This preserves KiCad enum
+            # symbols like `not_allowed`/`chamfer_rect` that are absent from the
+            # unquoted-keyword allowlist and would otherwise be wrongly quoted by
+            # the "quote everything else" fallback in _needs_quoting(). Values
+            # that structurally require quotes are still always quoted, for
+            # correctness. _originally_quoted always wins (a value parsed quoted
+            # stays quoted), so the two flags never conflict.
+            if self._originally_bare and not self._must_quote(self.value):
+                needs_quote = False
+            else:
+                needs_quote = self._originally_quoted or self._needs_quoting(self.value)
+            if needs_quote:
                 escaped = self.value.replace("\\", "\\\\").replace('"', '\\"')
                 escaped = escaped.replace("\n", "\\n").replace("\t", "\\t")
                 return f'"{escaped}"'
@@ -504,6 +530,22 @@ class SExp:
                 return str(int(self.value))
             return f"{self.value:.6g}"
         return str(self.value)
+
+    @staticmethod
+    def _must_quote(s: str) -> bool:
+        """Whether a string atom *structurally* requires quoting.
+
+        These are values that cannot be represented as a bare S-expression
+        token at all — an empty string, or one containing whitespace, parens,
+        or a double-quote character. Such values must always be quoted for
+        correctness, regardless of parse-time bare/quoted state. This is the
+        narrow "genuinely requires quotes" test used by the symmetric bare
+        handling in _format_atom() (issue #4185); it is intentionally distinct
+        from the KiCad-keyword allowlist heuristic in _needs_quoting().
+        """
+        if not s:
+            return True
+        return any(c in s for c in ' \t\n\r"()')
 
     @staticmethod
     def _is_valid_name(s: str) -> bool:
@@ -1243,7 +1285,10 @@ class Parser:
                 return node
             except ValueError:
                 pass
-        return SExp(value=token)
+        # A bare (unquoted) string token: record that it was parsed bare so the
+        # serializer can keep it bare on round-trip rather than re-quoting it via
+        # the keyword allowlist heuristic (issue #4185).
+        return SExp(value=token, _originally_bare=True)
 
     def _skip_whitespace(self):
         """Skip whitespace and comments."""
--- a/tests/test_sexp.py
+++ b/tests/test_sexp.py
@@ -452,6 +452,87 @@ class TestSExpSerialization:
             )
             assert f"(mirror {axis})" == result
 
+    def test_parsed_bare_keepout_enum_stays_bare(self):
+        """Keepout rule-area enum tokens parsed bare must re-emit bare.
+
+        KiCad emits `(tracks not_allowed)` (bare symbol) in footprint keepout
+        rule areas, and pcbnew's parser hard-rejects the quoted form
+        `(tracks "not_allowed")` — a board containing it fails to load. The
+        tokens `not_allowed`/`allowed` are absent from the unquoted-keyword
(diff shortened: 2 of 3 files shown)


rjwalters:
✅ **Approved!**

High-blast-radius review complete. The fix is correct, minimal, and symmetric, and the critical round-trip invariant is verified independently (not just via the builder's test sweep).

## What I verified

**Root cause confirmed** — On `main`, `parse_string('(tracks not_allowed)').to_string()` → `(tracks "not_allowed")`. The `_format_atom()` ratchet used `_originally_quoted` to *add* quotes but never used the inverse to *skip* the `_needs_quoting()` allowlist, so bare enum symbols absent from `unquoted_keywords` hit the "quote everything else" fallback.

**Bug + fix reproduced against real KiCad 10.0.4 (kicad-cli):**
- Quoted `(tracks "not_allowed")` → **"Failed to load board"** (the reported defect)
- Bare `(tracks not_allowed)` → **loads**, DRC runs normally

**The critical invariant holds — byte-identical on real files.** Round-tripped 12 real board/schematic/routed-PCB files through the old vs new serializer: **0 lines differ** in every one, and every new output is round-trip stable. No file that previously serialized correctly changes at all. The fix only touches tokens that were previously *broken*.

**Fix mechanism is sound and correctly scoped.** `_originally_bare` is set *only* in `Parser._parse_atom()` for bare tokens; `_originally_quoted` still wins; programmatically-constructed atoms (`_originally_bare=False`) still use the allowlist heuristic — so `test_serialize_kicad_keywords_unquoted` / `test_mirror_values_unquoted` are unaffected. The generalization to sibling tokens (`chamfer_rect`, `fillet`, `poly`, `castellated`, `heatsink`, `bga`) is correct, not over-reach: they're parsed bare and should stay bare.

## Mirror-image (under-quoting) risk — the key thing I scrutinized

The original bug was over-quoting; the danger of the fix is the inverse (a value that needs quotes now leaking out bare). `_must_quote()` is the safety boundary. It forces quoting for empty / whitespace (`space \t \n \r`) / parens / double-quote — which **exactly matches the parser's own `_ATOM_TERMINATORS` (` \t\n\r()`) plus the quote char**, and is consistent with `_is_valid_name()`. Confirmed a bare-parsed `a"b` is correctly quoted+escaped. The values that newly stay bare (`abc.def`, `100k`, `1N4148`, `a:b`, `3.3V`, `my/path/file`) are all round-trip stable and are precisely the class of bare symbols KiCad itself emits bare. Since parsed-bare tokens only originate from real KiCad files (bare = a valid symbol there) or fixtures, there is no realistic under-quoting corruption path.

One theoretical edge (non-blocking): `_must_quote()` omits the backslash `\`, so a bare token with a raw backslash would stay bare. This is round-trip stable in this parser and cannot arise from real KiCad output (backslashes only appear inside quoted strings, which parse `_originally_quoted=True`). Tracked as #4213 for defense-in-depth.

## CI / tests

- All **16 CI checks pass** — full Test suite (13m27s), all 8 board E2E + routing-regression jobs, kicad-cli Round-trip Smoke, Type Check, Lint & Format, C++ Build.
- Local: `test_sexp.py` + `test_sexp_roundtrip.py` → 90 passed / 5 skipped (skips are pre-existing demo-fixture gating). The kicad-cli-gated `test_keepout_board_loads_in_kicad` **ran and passed**.
- Broad sweep `-k "sexp or serialize or roundtrip or create_pcb or footprint"` → **1539 passed, 0 failed**.
- Merge state: **CLEAN**. PR description uses `Closes #4185`.

Excellent, well-tested fix that closes the reported bug and its whole sibling class at once without touching correct output. Filed #4213 to track the backslash hardening.

PR #4219, opened 2026-07-15
fix(sexp): treat backslash as structurally requiring quotes in _must_quote()

### LINKED ISSUE #4213: Harden SExp._must_quote() to also treat backslash as structurally-must-quote
## Context

Follow-up from Judge review of PR #4212 (fixes #4185 — symmetric bare-atom preservation in the sexp serializer).

The fix added `SExp._must_quote()` as the structural safety boundary that forces quoting for a parsed-bare atom whose value cannot be represented as a bare token. It currently checks for:

```python
if not s:
    return True
return any(c in s for c in ' \t\n\r"()')
```

This mirrors the parser's `_ATOM_TERMINATORS` (`" \t\n\r()"`) plus the double-quote char, and is consistent with the existing `_is_valid_name()`.

## Gap (theoretical, not a live corruption path)

A raw backslash (`\`) is **not** in the `_must_quote()` set. A bare token containing a literal backslash — e.g. parsing `(field abc\def)` — sets `_originally_bare=True` and now re-emits **bare** (`abc\def`) rather than quoted. This is round-trip stable within kicad-tools' own parser (bare tokens do not process escapes), so it is not an active bug today.

**Why it's not currently exploitable:** KiCad only ever emits backslashes inside *quoted* strings (Windows paths, escaped chars), which parse with `_originally_quoted=True` and are unaffected. A backslash in a bare token effectively never appears in real KiCad output. Verified: 12 real board/schematic/routed-PCB files serialize byte-identically old-vs-new.

## Suggested change

Add `\\` to the `_must_quote()` character set for defense-in-depth, and consider whether `_is_valid_name()` should be aligned the same way. Add a unit test asserting a bare-flagged atom containing a backslash serializes quoted+escaped.

Discovered during Judge evaluation of PR #4212.

---

### Verified against `origin/main` (commit 87561cf5)

Confirmed `SExp._must_quote()` in `src/kicad_tools/sexp/parser.py` (added by PR #4212 / issue #4185, currently at lines 534-546):

```python
@staticmethod
def _must_quote(s: str) -> bool:
    """Whether a string atom *structurally* requires quoting. ..."""
    if not s:
        return True
    return any(c in s for c in ' \t\n\r"()')
```

Backslash (`\`) is confirmed absent from the character set. Verified live (against `origin/main`'s parser, loaded directly): parsing `(field abc\def)` produces `_originally_bare=True` with value `abc\def`, and re-serializing today emits it bare and unescaped: `(field abc\def)`.

### Root cause / composition with the escaping logic

`_must_quote()` is consulted only in `_format_atom()` (line ~511):

```python
if self._originally_bare and not self._must_quote(self.value):
    needs_quote = False
else:
    needs_quote = self._originally_quoted or self._needs_quoting(self.value)
if needs_quote:
    escaped = self.value.replace("\\", "\\\\").replace('"', '\\"')
    escaped = escaped.replace("\n", "\\n").replace("\t", "\\t")
    return f'"{escaped}"'
```

The escaping step (`replace("\\", "\\\\")`) already runs for *any* atom that ends up quoted — it is unconditional on the `needs_quote` branch, not specific to `_originally_quoted` atoms. This means the fix composes correctly with zero additional changes to the escaping path: once `\\` is added to `_must_quote()`'s trigger set, a bare atom containing a backslash will (a) be routed into the `needs_quote = True` branch and (b) have its backslash properly doubled by the existing `.replace("\\", "\\\\")` call.

**Verified empirically** (patched `_must_quote` to include `\\`, using the `origin/main` parser module directly):

```
node = SExp(value='abc\\def', _originally_bare=True)   # Python string: abc\def
node._format_atom()  ->  '"abc\\\\def"'                 # i.e. serialized text: "abc\\def"
parse_string('(field "abc\\\\def")').get_string(0)  ->  'abc\\def'   # matches original exactly
```

Round-trip confirmed exact. No changes needed to the escaping logic — this is a pure one-line addition to `_must_quote()`'s character set.

### Verified: no over-quoting regression

Confirmed the fix only affects atoms that literally contain a backslash character. Re-ran the existing #4185 regression fixtures with the patched `_must_quote()`:

- `not_allowed`, `chamfer_rect`, `fillet`, `poly`, `castellated`, `heatsink`, `bga` (keepout/zone enum tokens from `test_sexp.py::test_parsed_bare_keepout_enum_stays_bare` and `test_parsed_bare_sibling_enum_tokens_stay_bare`) → unaffected, stay bare.
- Forward-slash tokens (e.g. `my/path`) → **not** a trigger; forward slash is intentionally excluded per #4185 and remains excluded here. Confirmed: `SExp(value="my/path", _originally_bare=True)._format_atom()` still returns bare `my/path`.

### Scope note: `_is_valid_name()`

The issue body speculates about aligning `_is_valid_name()` the same way. Verified via repo-wide grep (`git grep -n "_is_valid_name" -- src tests` against `origin/main`) that `_is_valid_name()` has exactly one reference in the whole tree — its own `def` — i.e. it is dead code, unused by `_must_quote()`, `_format_atom()`, or any caller. **Recommendation: leave `_is_valid_name()` untouched in this PR.** Changing unused dead code adds diff surface with no behavioral effect and is out of scope for a defense-in-depth hardening PR; if it's worth fixing/removing, file it as a separate followup rather than bundling here.

### Implementation

In `src/kicad_tools/sexp/parser.py`, `_must_quote()`:

```python
@staticmethod
def _must_quote(s: str) -> bool:
    if not s:
        return True
    return any(c in s for c in ' \t\n\r"()\\')
```

(Single character added to the set: `\\`, i.e. a literal backslash.) Update the docstring's character list description (currently "whitespace, parens, or a double-quote character") to also mention backslash.

## Acceptance Criteria

- [ ] `_must_quote()` returns `True` for any string containing a literal backslash (`\`), in addition to the existing empty/whitespace/paren/double-quote triggers.
- [ ] A bare-parsed atom (`_originally_bare=True`) whose value contains a backslash serializes **quoted and escaped** (e.g. Python value `abc\def` → serialized text `"abc\\def"`), and parsing that output back returns the exact original value (round-trip stable).
- [ ] Existing #4185 regression tests continue to pass unmodified: bare enum tokens (`not_allowed`, `chamfer_rect`, `fillet`, `poly`, `castellated`, `heatsink`, `bga`, etc.) still round-trip bare.
- [ ] Forward-slash-containing bare tokens (e.g. file-path-like atoms `my/path`) are **not** forced to quote — forward slash remains outside the `_must_quote()` trigger set, consistent with #4185.
- [ ] `_is_valid_name()` is left unchanged (confirmed dead code / unused elsewhere in the tree; out of scope here).

## Affected Files

- `src/kicad_tools/sexp/parser.py` - Add `\\` (backslash) to the character set checked in `SExp._must_quote()` (~line 534-546); update its docstring to mention backslash.
- `tests/test_sexp.py` - Extend the `TestSExpAtomFormatting`-style tests (see `test_parsed_bare_value_requiring_quotes_still_quoted`, `test_parsed_bare_keepout_enum_stays_bare`, `test_parsed_bare_sibling_enum_tokens_stay_bare`) with a new case: a bare-flagged atom containing a backslash must serialize quoted + escaped; add backslash to the "structurally requires quoting" parametrized value list alongside `"has space"`, `"with(paren"`, `""`.
- `tests/test_sexp_roundtrip.py` - Add a round-trip case parsing a bare token containing a backslash (e.g. `(field abc\def)`) and confirming `parse_string(...).to_string()` output re-parses to the identical value (this is the scenario the issue's Judge review flagged).

## Test Plan

- [ ] Manual verification: with the fix applied, run
  ```python
  from kicad_tools.sexp import SExp, parse_string
  node = SExp(value="abc\\def", _originally_bare=True)
  assert node._format_atom() == '"abc\\\\def"'
  assert parse_string('(field ' + node._format_atom() + ')').get_string(0) == "abc\\def"
  ```
- [ ] Automated tests: add a `test_sexp.py` case asserting `SExp._must_quote("abc\\def") is True` and that a bare-flagged atom with a backslash formats quoted+escaped (mirrors `test_parsed_bare_value_requiring_quotes_still_quoted`); add a `test_sexp_roundtrip.py` case parsing `(field abc\def)` (bare, unescaped backslash in source text) and confirming the re-serialized form round-trips byte-for-byte through a second parse.
- [ ] Regression: run full `uv run pytest tests/test_sexp.py tests/test_sexp_parser.py tests/test_sexp_roundtrip.py tests/test_sexp_builders.py -v` and confirm all #4185/#4212 tests (`test_parsed_bare_keepout_enum_stays_bare`, `test_parsed_bare_sibling_enum_tokens_stay_bare`, `test_parsed_quoted_string_stays_quoted`, `test_parsed_bare_value_requiring_quotes_still_quoted`, `test_mirror_values_unquoted`) still pass unmodified.
- [ ] Edge cases: verify a token with multiple/adjacent backslashes (`a\\b\\c` in Python, i.e. two literal backslashes) still round-trips; verify a token that is *both* originally-quoted and contains a backslash is unaffected (already quoted via `_originally_quoted`, so `_must_quote()` is never reached for that atom — no double-escaping regression).
- [ ] Confirm real-file regression: re-run the same "N real board/schematic/routed-PCB files serialize byte-identically" check the original PR #4212 used, to confirm this change is a true no-op on real KiCad corpora (backslashes never appear in bare tokens in real files, only inside already-quoted strings).

## Dependencies

None. #4185 / PR #4212 are already merged on `origin/main` (confirmed at commit `24c6b3d3`, ancestor of current `origin/main` HEAD `87561cf5`).

rjwalters:
## Summary

Defense-in-depth follow-on to #4185/PR #4212. `SExp._must_quote()` is the structural safety boundary that forces quoting for a parsed-bare atom whose value cannot be represented as a bare token. Its trigger set (`' \t\n\r"()'`) omitted the backslash character, so a bare-parsed token containing a raw backslash (e.g. parsing `(field abc\def)`) was re-emitted bare and unescaped rather than quoted+escaped.

This adds a single character (`\`) to that trigger set. Once a backslash-containing bare atom is routed into the `needs_quote = True` branch of `_format_atom()`, that branch's already-unconditional `.replace("\\", "\\\\")` step doubles the backslash correctly, so the emitted quoted form round-trips byte-exact. No other code changes are required.

This is not a live corruption path today (KiCad only emits backslashes inside already-quoted strings), but it hardens the boundary against a theoretical bare-backslash input.

## Changes

- `src/kicad_tools/sexp/parser.py`: add `\\` (literal backslash) to the character set checked in `SExp._must_quote()`; update its docstring to mention backslash and the composition with `_format_atom()`'s escaping.
- `tests/test_sexp.py`: new `test_parsed_bare_backslash_value_forced_quoted` — asserts `_must_quote()` returns `True` for backslash values, `False` for forward-slash values, and that a bare-flagged backslash atom serializes quoted+escaped and re-parses to the original.
- `tests/test_sexp_roundtrip.py`: new `test_bare_backslash_token_roundtrips_quoted_and_escaped` — parses bare `(field abc\def)`, confirms it emits `(field "abc\\def")`, re-parses to the identical value, and is a serialize fixed point.

| `_must_quote()` returns `True` for any string containing a backslash | ✅ | Unit test asserts `SExp._must_quote("abc\def") is True` and `"a\b\c"` |
| Bare-parsed backslash atom serializes quoted+escaped and round-trips | ✅ | `_format_atom()` == `'"abc\\def"'`; re-parse yields `abc\def`; roundtrip test is a fixed point |
| Existing #4185 regression tests pass unmodified | ✅ | `test_sexp.py` + `test_sexp_roundtrip.py` + `test_sexp_parser.py` + `test_sexp_builders.py` all green (241 passed, 5 kicad-cli skips) |
| Forward-slash bare tokens NOT forced to quote | ✅ | `_must_quote("my/path") is False`; `SExp(value="my/path", _originally_bare=True)._format_atom()` == `my/path` |
| `_is_valid_name()` left unchanged | ✅ | Not touched (curator-confirmed dead code, out of scope) |

## Test Plan

- `uv run pytest tests/test_sexp.py tests/test_sexp_roundtrip.py -q` → 92 passed, 5 skipped
- `uv run pytest tests/test_sexp_parser.py tests/test_sexp_builders.py -q` → 149 passed
- `uv run --extra dev python scripts/ci/check_mypy_baseline.py` → 0 new type errors (within baseline)

Closes #4213

### DIFF
--- a/src/kicad_tools/sexp/parser.py
+++ b/src/kicad_tools/sexp/parser.py
@@ -537,15 +537,22 @@ class SExp:
 
         These are values that cannot be represented as a bare S-expression
         token at all — an empty string, or one containing whitespace, parens,
-        or a double-quote character. Such values must always be quoted for
-        correctness, regardless of parse-time bare/quoted state. This is the
-        narrow "genuinely requires quotes" test used by the symmetric bare
-        handling in _format_atom() (issue #4185); it is intentionally distinct
-        from the KiCad-keyword allowlist heuristic in _needs_quoting().
+        a double-quote character, or a backslash. Such values must always be
+        quoted for correctness, regardless of parse-time bare/quoted state.
+        This is the narrow "genuinely requires quotes" test used by the
+        symmetric bare handling in _format_atom() (issue #4185); it is
+        intentionally distinct from the KiCad-keyword allowlist heuristic in
+        _needs_quoting().
+
+        Backslash is included (issue #4213, defense-in-depth follow-on to
+        #4185): a bare token containing a raw backslash is routed into the
+        quoting branch of _format_atom(), whose unconditional
+        ``.replace("\\", "\\\\")`` step doubles it correctly so the emitted
+        quoted form round-trips byte-exact.
         """
         if not s:
             return True
-        return any(c in s for c in ' \t\n\r"()')
+        return any(c in s for c in ' \t\n\r"()\\')
 
     @staticmethod
     def _is_valid_name(s: str) -> bool:
--- a/tests/test_sexp.py
+++ b/tests/test_sexp.py
@@ -533,6 +533,32 @@ class TestSExpSerialization:
         empty = SExp(value="", _originally_bare=True)
         assert empty._format_atom() == '""'
 
+    def test_parsed_bare_backslash_value_forced_quoted(self):
+        """A bare atom containing a raw backslash is forced quoted+escaped.
+
+        Backslash cannot be safely represented as a bare token (issue #4213,
+        defense-in-depth follow-on to #4185): a bare-flagged atom whose value
+        contains a literal backslash must be routed into the quoting branch of
+        _format_atom(), whose unconditional ``.replace("\\", "\\\\")`` step
+        doubles the backslash so the emitted quoted form round-trips byte-exact.
+        Forward slash is intentionally *not* a trigger and stays bare.
+        """
+        # _must_quote() flags any value containing a literal backslash.
+        assert SExp._must_quote("abc\\def") is True
+        assert SExp._must_quote("a\\b\\c") is True
+        # Forward slash is intentionally excluded and must stay bare (#4185).
+        assert SExp._must_quote("my/path") is False
+
+        # A bare-flagged atom with a backslash serializes quoted + escaped.
+        node = SExp(value="abc\\def", _originally_bare=True)
+        assert node._format_atom() == '"abc\\\\def"'
+        # Parsing that emitted form back yields the exact original value.
+        assert parse_string("(field " + node._format_atom() + ")").get_string(0) == "abc\\def"
+
+        # Forward-slash bare atoms remain bare (no over-quoting regression).
+        slash = SExp(value="my/path", _originally_bare=True)
+        assert slash._format_atom() == "my/path"
+
 
 class TestSExpFileIO:
     """Tests for file I/O functions."""
--- a/tests/test_sexp_roundtrip.py
+++ b/tests/test_sexp_roundtrip.py
@@ -140,6 +140,34 @@ class TestSExpRoundTripWithKiCad:
                 f"expected bare ({field} not_allowed) in: {output}"
             )
 
+    def test_bare_backslash_token_roundtrips_quoted_and_escaped(self):
+        """A bare token containing a raw backslash round-trips byte-exact.
+
+        Parsing `(field abc\\def)` yields a bare atom whose value contains a
+        literal backslash. Before issue #4213 the serializer re-emitted it bare
+        and unescaped; with backslash added to _must_quote() it is now forced
+        into the quoting branch, whose unconditional escaping doubles the
+        backslash. The emitted form (`(field "abc\\\\def")`) must re-parse to
+        the identical value, and a second serialize/parse cycle must be a fixed
+        point (byte-exact round-trip).
+        """
+        # Source text: (field abc\def) — a single, unescaped, bare backslash.
+        doc = parse_string("(field abc\\def)")
+        assert doc.get_string(0) == "abc\\def"
+
+        # Serialized output forces quoting and doubles the backslash.
+        output = doc.to_string()
+        assert '(field "abc\\\\def")' in output, (
+            f"bare backslash token must serialize quoted+escaped, got: {output}"
+        )
+
+        # Re-parsing the emitted form yields the exact original value.
+        reparsed = parse_string(output)
+        assert reparsed.get_string(0) == "abc\\def"
+
+        # A second serialize cycle is a fixed point (stable byte-exact).
+        assert reparsed.to_string() == output
+
     @pytest.mark.skipif(KICAD_CLI is None, reason="kicad-cli not installed")
     def test_keepout_board_loads_in_kicad(self, tmp_path):
         """A board carrying a keepout footprint must load after round-trip.


rjwalters:
✅ **Approved!**

Clean one-character defense-in-depth fix. Verified end-to-end.

## Correctness

The single change adds `\` to the `_must_quote()` trigger set (`any(c in s for c in ' \t\n\r"()\\')`). I traced the composition with `_format_atom()`'s bare-handling branch:

- Before: for a bare-parsed `abc\def`, `_originally_bare and not _must_quote(...)` was `True and not False = True` → `needs_quote = False` → emitted raw/unescaped `abc\def` (the theoretical corruption).
- After: `_must_quote("abc\def")` is now `True`, so the bare-preserve branch is skipped → falls to `needs_quote = _originally_quoted or _needs_quoting(...)` → `True` → the existing unconditional `.replace("\\", "\\\\")` doubles the backslash → `"abc\\def"`. Byte-exact round-trip. Correct.

## No over-quoting

- `_must_quote("my/path")` is `False`; forward slash stays bare — confirmed by the new unit test and consistent with #4185.
- #4185 bare-enum fixtures (`not_allowed`, `chamfer_rect`) are unaffected — `test_sexp_roundtrip.py` bare-enum tests pass unmodified.
- `_is_valid_name()` untouched (out of scope, curator-confirmed dead code).

## New tests genuinely assert the behavior

- `test_parsed_bare_backslash_value_forced_quoted` (test_sexp.py): asserts `_must_quote` True for backslash / False for slash, `_format_atom()` == `'"abc\\\\def"'`, re-parse yields exact original, and slash bare atom stays bare.
- `test_bare_backslash_token_roundtrips_quoted_and_escaped` (test_sexp_roundtrip.py): parses bare `(field abc\def)`, asserts quoted+escaped emission `(field "abc\\def")`, re-parse identity, and serialize fixed-point.

Both confirmed to actually execute (not skip) and pass.

## Verification (PR branch worktree)

- `pytest tests/test_sexp.py tests/test_sexp_roundtrip.py` → **92 passed, 5 skipped**
- `pytest tests/test_sexp_parser.py tests/test_sexp_builders.py` → **149 passed**
- `ruff check` + `ruff format --check` on all 3 files → **clean**
- mypy baseline → **0 new type errors** (1475 within 1514 floor)
- **All 16 GitHub CI checks pass**; `mergeStateStatus: CLEAN`

PR branch is based on latest `origin/main` (#4212 already merged). Ready to merge.

PR #4232, opened 2026-07-15
fix(net-status): count zone-pour copper contact for plane-pad connectivity

### LINKED ISSUE #4229: net-status --strict (#4176): count zone-pour copper contact — plane pads connected via the fill (no stitching via) are false-flagged incomplete
## Rough edge (non-blocking; surfaced by v0.16.0 pre-release fleet validation)

The new opt-in `kct net-status --strict` real-geometric-connectivity mode (#4176 Phase 1) appears to consider only segment/via copper, NOT zone-fill (pour) copper. So a plane pad that connects to its net purely through direct pad-in-pour contact (no explicit stitching via) is flagged as incomplete ("needs via to plane"), while kicad-cli's zone-aware engine (`pcb drc --refill-zones`) reports it as fully connected (0 unconnected).

Observed during fleet validation:
- board-06: U1.17 (+3V3) and U1.32 (GND) flagged incomplete by `--strict`; kicad-cli reports 0 unconnected — pads connect through the pour.
- board-05: 5 plane nets (+24V/+3V3/+5V/GND/VIN) similarly flagged; same pour-connectivity pattern (#3901/#3914-adjacent).

Default (tolerance) net-status is unaffected — this is only in the opt-in `--strict` real-geometry path, so it does NOT block anything. But it's a real gap: `--strict` claims to match KiCad's connectivity semantics, and KiCad counts zone-pour contact.

## Suggested fix

Include zone-fill copper polygons in the `--strict` connectivity geometry: a pad whose copper overlaps a filled zone of its own net is connected, matching KiCad. Reuse the shapely zone-fill geometry the connectivity/copper-LVS code already has. Ties into the #4176 Phase 2 follow-up (flip the default + unify ConnectivityValidator) and the POUR_PAD_ERODE note the #4176 judge flagged.

## Acceptance

`kct net-status --strict` on a board with plane-pour-connected pads (no stitching via) reports them COMPLETE (matching kicad-cli); a genuinely-open pad still flagged. Synthetic fixture: a pad in a filled zone of its net with no segment/via.

---

### Root cause (confirmed by inspection — retitled from the original framing)

The originally-reported framing ("`--strict` ignores zone-fill copper entirely") is **not accurate** — verified against `origin/main` (commit `85530f3a`, PR #4201). Zone-fill copper geometry is already wired into `NetStatusAnalyzer._build_connectivity_graph()` via `_apply_zone_connectivity()` (`src/kicad_tools/analysis/net_status.py:791-801`, calling into `_build_fill_island_groups()` at `net_status.py:900-1017`), and this call is **unconditional** — it runs identically whether `strict=True` or `strict=False`. It reuses `ConnectivityValidator._pad_copper_polygon` / `_fill_solid_region` (`src/kicad_tools/validate/connectivity.py:1293-1364`) exactly as the issue's "suggested fix" describes, via the shared `_connectivity_geometry()` accessor (`net_status.py:838-854`). So the zone-fill geometry source the issue asks to "locate and reuse" is **already reused** — the #4176 PR description explicitly confirms this: "pad/via/zone copper polygons already exist in `ConnectivityValidator` and are reused ... This was the one missing piece [segment polygons]."

Reproduced against a synthetic 3-pad board (`R1.1`/`R2.1` joined by a trace forming one island; `U1.17` sitting alone in a filled zone of the same net, no trace/via to it):

```
strict=False status=incomplete connected=2 unconnected=1  (U1.17)
strict=True  status=incomplete connected=2 unconnected=1  (U1.17)
```

**Both modes fail identically** — this is not `--strict`-specific. The actual defect is in how `_build_fill_island_groups()`'s output is consumed by the graph/island logic, not in which geometry is fed in:

- `_build_fill_island_groups()` (`net_status.py:900-1017`) correctly identifies that `U1.17`'s copper overlaps the fill's solid region and returns a group `{'U1.17'}` (confirmed via direct call in isolation).
- `_apply_zone_connectivity()` (`net_status.py:893-898`) unions each group **pairwise** into the adjacency graph: `for i, pad in enumerate(members): for other in members[i+1:]: graph[pad].add(other)`. A **singleton** group (one pad alone in its fill fragment/island, no sibling pad sharing that same fragment) produces **zero graph edges** — there's nothing to pair `U1.17` with.
- `_analyze_net()` (`net_status.py:536-546`) then calls `_find_islands()` (`net_status.py:1435-1474`), which seeds every pad with no graph edges as its own **singleton island**. It sorts all islands by size and treats only the **largest** as "connected" (`net_status.py:541-544`, "Largest island is considered connected"). A lone pour-connected pad's singleton island always loses that vote to the larger trace-connected island — even though its copper genuinely touches the net's poured copper.

This is the same class of bug as #3914 (the "must be grouped, not globally unioned" fix) but on the opposite side: #3914 correctly stopped over-connecting pads across separate fill islands; this bug under-connects a pad that IS on a real (if lone-occupant) island because nothing ever unions a solo island's single member into the graph, and the largest-island-wins heuristic then discards it.

**Existing test coverage gap confirming this**: every zone-connectivity fixture in `tests/test_net_status.py` (`test_analyze_zone_pcb`, `test_via_zone_connectivity`, `test_tht_pad_zone_connectivity`, `test_stitch_via_thermal_cutout_connectivity` — lines 598-700ish) tests 2+ pads that are **all** connected only through the zone (so the pour island IS the largest/only island and trivially wins). None test the board-06/board-05 real-world shape: a pad in a pour that **coexists** on the same net with a separately-routed (trace/via) island. That combination is the actual gap.

### The fix

In `_analyze_net()` (or in how its output graph/islands are consumed), a pad that is a member of *any* non-trivial connectivity source — including a **singleton** pour-fill group — must be treated as "connected", not just pads belonging to the single largest island. Two implementation shapes, either is acceptable:

1. **Preferred — fix the islands/largest-island heuristic**: change `_analyze_net()`'s connected-pad determination so that a pad electrically bonded to *any* zone fill of its own net (per `_build_fill_island_groups()`) is treated as connected regardless of island size, rather than relying purely on "largest island wins". This could mean unioning all zone-fill-bonded pads into a synthetic connectivity anchor (e.g. treat each filled-zone fill-island group as connected to the largest trace/via island's pads *when they're on the same net* — matching KiCad, where any real copper contact with the pour makes the pad part of the net's copper, not a separate isolated island) OR simply: any pad in a `_build_fill_island_groups()` group (size >= 1) is `is_connected = True`, independent of the largest-island graph-traversal outcome.
2. **Alternative — synthetic zone-anchor node**: give each zone fill-island group a synthetic anchor node (similar to the `__via{index}` synthetic nodes `ConnectivityValidator` already uses, see `_synthetic_via_radii` at `connectivity.py:1366-1379`) and bond every group member (however many) to that anchor. This makes even a singleton group "attached to something" so `_find_islands` no longer needs a `len(group) >= 2` implicit assumption — but does NOT by itself fix the "largest-island-wins" discard; a pour-connected island (even a multi-pad one) can still legitimately be a *different, smaller* island than a trace-heavy island on the same net, and both should count as connected. Option 1 is the more direct fix for the actual reported symptom.

Whichever shape is chosen, the essential correction is: **"connected" must mean "reachable from real copper" — trace, via, OR zone fill of the pad's own net — not "is a member of the single largest connected-component by pad count".** The largest-island heuristic conflates cardinality with correctness; a net can legitimately have two genuinely separate but both-valid copper islands (e.g. 40 pads bulk-routed with traces, plus 3 more pads dropped straight into the pour with no stitching via) and KiCad reports all 43 as connected.

Confirm during implementation whether this requires a broader `_analyze_net()` rewrite (e.g., "connected" = union of all islands that touch a filled zone of the net, plus the largest island) or whether a smaller patch limited to the singleton-group case is sufficient to close the board-06/board-05 repro without destabilizing existing zone tests. Run the full `tests/test_net_status.py` + `tests/test_net_status_strict.py` suites after the change — the existing "discontinuous fill / separate islands must NOT be unioned" tests from #3947/#3914 must continue to pass (a pad on fill-island A must still NOT be marked connected to a pad on a genuinely separate, non-net-bridged fill-island B; only "connected to ITS OWN fill island, regardless of that island's pad count" should change).

### Fill-currency (already handled correctly — no new work needed)

The issue asked whether `--strict` requires committed fills to be current. This is **already correct** and requires no change: `_build_fill_island_groups()` only consults `zone.filled_polygons` (`net_status.py:976`), i.e. the *committed* fill geometry as parsed from the `.kicad_pcb` file — not a live re-fill. An unfilled zone (`zone.filled_polygons` empty) is skipped entirely (`net_status.py:976`, `and not zone.filled_polygons` guard on `_apply_zone_connectivity`'s caller at `net_status.py:663`), matching KiCad's "unfilled zones don't count" behavior. `kct net-status` does not call `kicad-cli --refill-zones`; it reads whatever fill is already committed to the file, same as the rest of the toolchain. This part of the issue's ask is satisfied by existing behavior — no change needed here; just confirm with a test (below).

### Scope guard (BOUNDED — keep for v0.16.0)

- Touch only `src/kicad_tools/analysis/net_status.py` (the connected-pad determination in `_analyze_net()` / `_find_islands()` / `_apply_zone_connectivity()` interplay). Do NOT touch `src/kicad_tools/validate/connectivity.py` (`ConnectivityValidator`'s own parallel chainer) — that unification is explicitly deferred to the #4176 Phase 2 follow-up per the PR #4201 description.
- Do NOT flip the tolerance-mode default, and do NOT gate the fix behind `strict` — the bug reproduces identically in both modes, so the fix must apply to both (it lives in shared code already called unconditionally). This is a genuine bug-fix, not a strict-mode enhancement — retitle/recharacterize in the PR if useful, but keep it a single bounded fix.
- Single-PR fix. Do not attempt to also unify `ConnectivityValidator`'s segment chainer in the same change.

## Affected Files

- `src/kicad_tools/analysis/net_status.py` - Fix the connected-pad determination so a pad bonded to any zone fill-island of its own net (including a singleton-member island) is treated as connected, not discarded by the "largest island wins" heuristic. Relevant regions: `_analyze_net()` (~536-559), `_apply_zone_connectivity()` (856-898), `_build_fill_island_groups()` (900-1017), `_find_islands()` (1435-1474).
- `tests/test_net_status_strict.py` - Add strict-mode synthetic fixtures for: (a) a pad in a filled zone of its net with no segment/via and no sibling pad on that fill fragment, coexisting with a separately trace-connected island on the same net -> COMPLETE; (b) a pad NOT in any fill and with no segment/via -> still INCOMPLETE; (c) a pad in an UNFILLED zone (fill disabled / empty `filled_polygons`) -> INCOMPLETE (matching KiCad — confirms the fill-currency behavior already works and doesn't regress).
- `tests/test_net_status.py` - Mirror the same three fixtures for the default (non-strict) path, since the root cause and fix apply to both modes identically (the bug is not strict-specific).

## Test Plan

- [ ] Automated test: synthetic board with `R1.1`/`R2.1` joined by a trace (forms island A) and `U1.17` sitting alone inside a filled zone of the same net with no segment/via to it (forms singleton island B via pour) -> `net-status` (both `strict=False` and `strict=True`) reports the net COMPLETE, `U1.17` in `connected_pads`. This is the direct board-06/board-05 repro shape.
- [ ] Automated test: a pad with no trace/via AND no zone-fill overlap on its net -> still INCOMPLETE in both modes (regression guard — must not become a rubber-stamp "always complete" fix).
- [ ] Automated test: a pad geometrically inside a zone's outline but the zone's `filled_polygons` is empty (fill disabled/stale) -> still INCOMPLETE in both modes (confirms fill-currency: only committed/parsed fill copper counts, matching KiCad's "unfilled zone = no connectivity").
- [ ] Regression: full `uv run pytest tests/test_net_status.py tests/test_net_status_strict.py` passes, including the existing #3914/#3947 "separate fill islands must not be unioned across each other" tests (only the singleton-per-net-owns-island case should change, not cross-island unioning).
- [ ] Regression: `uv run pytest tests/test_copper_lvs.py` passes unchanged (out of scope, but confirms `ConnectivityValidator` itself, which this fix does not touch, is unaffected).
- [ ] Manual verification (if board-06/board-05 fixtures are available locally): `kct net-status board-06.kicad_pcb --strict` no longer flags `U1.17`/`U1.32`; `kct net-status board-05.kicad_pcb --strict` no longer flags the 5 plane nets. Not required for CI (fixtures are chorus-local per repo memory) but useful spot-check if available in the builder's environment.

rjwalters:
## Summary

A plane pad connected to its net purely through pad-in-pour copper contact (no stitching via) -- or through a cross-layer `pad -> trace -> via -> trace -> pour` path where the via lands in a thermal antipad -- was false-flagged **incomplete** by `NetStatusAnalyzer` in **both** default and `--strict` modes, while `kicad-cli pcb drc --refill-zones` reports **0 unconnected** for exactly those pads.

`net-status` is the completion oracle (route loops, `--why`, `kct check` connectivity, CI all key off it), so this fix changes both modes' output. It was validated to introduce **no completion regression** on any committed board.

## Root cause

`_build_fill_island_groups()` only bonded a pad to a pour when the pad copper directly overlapped the fill, or a via *penetrating* the fill sat at the pad centre. It missed:
1. the cross-layer chain where the through-via lands in a thermal antipad (its copper does **not** penetrate the pour) and only the far trace endpoint reaches the poured copper (board-06 `U1.17` +3V3);
2. a stitching via placed *beside* a same-net pad whose real copper overlaps it only before the `POUR_PAD_ERODE` inset (board-06 `U1.32` GND).

A lone pour-bonded pad then formed a **singleton** graph island that always lost the "largest island wins" vote to a separately-routed trace island on the same net.

## Changes

- `_merge_chains_via_vias()`: union segment components that share a via into extended chains, so a cross-layer trace path is recognised as one electrical chain (KiCad semantics).
- `_segment_touches_via()`: join a trace to a via by copper overlap, not just endpoint proximity (traces land on the annular ring, not dead-centre).
- Bond a pad to a fill island when an extended chain that touches the fill reaches it, or when the pad copper overlaps a pour-penetrating via (raw, un-eroded radius). The raw radius is safe here because each fill-island group is single-net, so a looser pad<->via bond can only unify pads that are already the same net -- it can never manufacture a cross-net short (the eroded penetration test still guards that).
- **Scope:** `net_status.py` only. `ConnectivityValidator` (`validate/connectivity.py`) is untouched -- that unification is the deferred #4176 Phase 2.

The per-fill-island grouping is preserved, so two physically disjoint pours of the same net are **not** unioned -- a real open between them stays visible (#3914/#3947). Unfilled zones still provide no connectivity (#3482 fill-currency); that path is unchanged.

| Lone pad in a filled zone of its net (no via, no sibling on fragment) -> COMPLETE, both modes | ✅ | `test_lone_pour_pad_reports_complete` (strict + default); board-06 U1.17/U1.32 |
| Pad not in any fill, no segment/via -> still INCOMPLETE, both modes | ✅ | `test_pad_off_pour_still_incomplete` |
| Pad in an UNFILLED zone -> INCOMPLETE (fill-currency), both modes | ✅ | `test_unfilled_zone_still_incomplete` |
| #3914/#3947 regression guard: two disjoint fills NOT falsely unioned | ✅ | `test_two_disjoint_fills_not_unioned`; existing `test_discontinuous_fill_not_complete` / `test_partial_fill_zone_only_connects_pads_on_copper` still pass |
| ConnectivityValidator untouched | ✅ | `tests/test_copper_lvs.py` passes unchanged (47 passed) |
| No board completion regression | ✅ | Board re-validation table below |

## Board net-status re-validation (before → after; committed routed boards)

| Board | Mode | Before (complete/incomplete) | After | Note |
|-------|------|------------------------------|-------|------|
| 00 | default / strict | 3/0 · 3/0 | 3/0 · 3/0 | unchanged |
| 01 | default / strict | 3/0 · 3/0 | 3/0 · 3/0 | unchanged |
| 02 | default / strict | 10/0 · 10/0 | 10/0 · 10/0 | unchanged |
| 03 | default / strict | 16/0 · 16/0 | 16/0 · 16/0 | unchanged |
| 04 | default / strict | 11/1 (GND) · 11/1 (GND) | 12/0 · 12/0 | GND was a **false**-incomplete; kicad-cli = 0 unconnected |
| 06 | default | 22/4 (+1V2,+3V3,GND,VBUS_USB) | 23/3 (+1V2,GND,VBUS_USB) | +3V3 U1.17 now complete |
| 06 | strict | 24/2 (+3V3,GND) | **26/0** | U1.17 (+3V3) and U1.32 (GND) plane pads now complete; matches kicad-cli 0 unconnected |
| 07 | default / strict | 29/5 · 29/5 | 29/5 · 29/5 | 5 genuine #3438 opens **still reported** (kicad-cli = 5 unconnected) |

No board's connectivity count regresses. Ground truth cross-checked with `kicad-cli pcb drc --refill-zones` on boards 04 (0 unconnected), 06 (0 unconnected), 07 (5 unconnected). Boards 05 (tier1, local-only) not run in this environment.

## Test Plan

- `uv run pytest tests/test_net_status_strict.py tests/test_net_status.py tests/test_analysis_net_status.py` -> 122 passed (110 existing + 12 new)
- `uv run pytest tests/ -k "net_status or connectivity"` -> 388 passed
- `uv run pytest tests/test_copper_lvs.py` -> 47 passed (ConnectivityValidator unaffected)
- `ruff check` + `ruff format` clean; `check_mypy_baseline.py` -> 0 new errors
- C++ native backend built (`kct build-native`)

Closes #4229

### DIFF
--- a/src/kicad_tools/analysis/net_status.py
+++ b/src/kicad_tools/analysis/net_status.py
@@ -930,6 +930,17 @@ class NetStatusAnalyzer:
         through them (directly, or via a segment chain ending at the via) into
         that island, closing the ``pad -> trace -> stitch via -> pour`` path.
 
+        Segment chains whose copper *touches* a fill island bond their pads to
+        it too, and chains are unified transitively through shared vias so a
+        cross-layer path bonds correctly (Issue #4229): the common real-world
+        shape is a signal pad on F.Cu that reaches a plane on B.Cu through
+        ``pad -> F.Cu trace -> through-via -> B.Cu trace -> pour`` where the via
+        itself sits in a thermal antipad (its copper does NOT penetrate the
+        pour) and only the far B.Cu trace endpoint lands on the poured copper.
+        The per-via penetration test alone misses that pad; bonding via the
+        chain-touches-fill path recovers it, matching KiCad's zone-aware
+        connectivity (kicad-cli reports 0 unconnected for exactly this case).
+
         Returns ``None`` when ``shapely`` is unavailable (the caller then uses
         the legacy boundary-based fallback); otherwise a list of pad-id groups,
         one per bonded fill island.
@@ -960,16 +971,48 @@ class NetStatusAnalyzer:
                 if poly is not None:
                     pad_polys[pad_id] = poly
 
-        # Copper-circle geometry per via (radius = size / 2, eroded like a pad
-        # box).  A via whose copper penetrates a pour's solid region ties any
-        # pad reachable through it into that island.
-        via_geoms = [
-            (
-                via,
-                cv._via_copper_geom(via.position, max(getattr(via, "size", 0.0) or 0.0, 0.0) / 2.0),
-            )
-            for via in vias
-        ]
+        # Copper-circle geometry per via.  Two variants:
+        #  * ``via_geom`` (eroded, ``POUR_PAD_ERODE`` inset) is used for the
+        #    fill-*penetration* test so a via merely grazing the pour edge does
+        #    not spuriously bond -- matching ``ConnectivityValidator``.
+        #  * ``via_raw`` (un-eroded, real copper radius) is used only to bond a
+        #    *same-net* pad to a via that already penetrates the pour.  The
+        #    erosion inset on both the pad box and the via disc otherwise opens
+        #    a sub-``2*POUR_PAD_ERODE`` false gap between a stitching via and an
+        #    adjacent same-net pad whose copper genuinely overlaps it (the
+        #    board-06 U1.32 case, Issue #4229).  Using the raw radius here is
+        #    safe: the group is single-net, so a looser pad<->via bond can only
+        #    unify pads that are already the same net -- it can never manufacture
+        #    a cross-net short (that is what the eroded penetration test guards).
+        from shapely.geometry import Point as _ShapelyPoint  # type: ignore[import-untyped]
+
+        via_geoms = []
+        for via in vias:
+            radius = max(getattr(via, "size", 0.0) or 0.0, 0.0) / 2.0
+            eroded = cv._via_copper_geom(via.position, radius)
+            raw = _ShapelyPoint(*via.position).buffer(radius) if radius > 0 else None
+            via_geoms.append((via, eroded, raw))
+
+        # Unify segment components that share a via into "extended chains"
+        # (Issue #4229).  ``_build_segment_components`` chains segments only
+        # when their own copper touches, so a through-via joining an F.Cu trace
+        # to a B.Cu trace leaves them in separate components even though they
+        # are one electrical net.  Merging components that both touch the same
+        # via recovers the cross-layer ``pad -> trace -> via -> trace -> pour``
+        # path so a pad on one layer bonds to a pour on another.
+        extended_chains = self._merge_chains_via_vias(segments, segment_components, vias)
+
+        # Pre-compute, per extended chain, the pads it reaches and the copper
+        # layers/geometry it presents to the pour tests.
+        chain_pads: list[set[str]] = []
+        chain_seg_indices: list[set[int]] = []
+        for chain in extended_chains:
+            pads_in_chain: set[str] = set()
+            for s in chain:
+                pads_in_chain.update(self._find_pads_at_point(segments[s].start, pad_positions))
+                pads_in_chain.update(self._find_pads_at_point(segments[s].end, pad_positions))
+            chain_pads.append(pads_in_chain)
+            chain_seg_indices.append(chain)
 
         groups: list[set[str]] = []
         for zone in self.pcb.zones:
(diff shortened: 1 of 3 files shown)


rjwalters:
## Judge review: APPROVED — ready for merge

Reviewed the connectivity-model change to `net_status.py` fixing the pour-bonded-pad false-incomplete (#4229). Verdict: **APPROVED**.

### Critical focus: the #3914/#3947 over-union hazard — CLEARED
Every new bond path is keyed on **actual geometric copper overlap**, never on net-name equality alone:
- via bonds gated on `region.intersects(via_geom)` (eroded via must penetrate the fill);
- the raw-radius same-net pad↔via bond gated on `via_raw.intersects(pad_geom)` (pad copper must overlap the via);
- the chain-touches-fill bond gated on `region.intersects(self._segment_poly(...))` (trace copper must land on the poured copper);
- `_segment_touches_via` uses `_geoms_touch` on the real copper polygons.

`pad_polys`, `vias`, and `segments` are all net-scoped (`pad.net_number == net_number`), and bonding is accumulated **per fill island** (`groups.append(bonded)` inside the per-`filled_polygon` loop). So the raw-radius relaxation can only unify pads that are *already the same net and geometrically overlap the same via* — it cannot manufacture a cross-net short, and it cannot union two physically separate fills. The builder's safety argument holds.

### Regression tests are genuine, not forced-green
- `test_two_disjoint_fills_not_unioned` (both modes): two separate `filled_polygon` islands (~16 mm apart, no bridge), one pad each → asserts `connected_count == 1, unconnected_count == 1`. Real disjoint case; this is the exact #3914/#3947 guard.
- `test_partial_fill_zone_only_connects_pads_on_copper` and `test_discontinuous_fill_not_complete` (existing, `tests/test_analysis_net_status.py`) still pass.
- `test_pad_off_pour_still_incomplete` (pad moved outside the pour) and `test_unfilled_zone_still_incomplete` (`filled_polygon` dropped → #3482 fill-currency preserved) are proper negative controls.

### ConnectivityValidator untouched
`git diff` on `src/kicad_tools/validate/connectivity.py` is empty; `tests/test_copper_lvs.py` → 47 passed locally. #4176 Phase 2 deferral confirmed. Scope is exactly `net_status.py` + its two test files (+460/−24), no unrelated churn.

### Board claims independently verified against kicad-cli ground truth
Ran `kct net-status` on main (before) vs the branch (after), cross-checked with `kicad-cli pcb drc --refill-zones` in this environment:

| Board | before strict | after strict | kicad-cli unconnected |
|-------|--------------|--------------|-----------------------|
| 04 | 11/1 | **12/0** | 0 ✓ |
| 06 | 24/2 | **26/0** | 0 ✓ |
| 07 | 29/5 | 29/5 | 5 ✓ |

Board-06 default also 22/4 → 23/3 as claimed. Board-07's 5 genuine #3438 opens are **still reported** — the fix does not mask real opens; it aligns net-status with KiCad's zone-aware engine exactly.

### Local + CI
- `tests/test_net_status_strict.py tests/test_net_status.py tests/test_analysis_net_status.py` → 122 passed locally.
- All 16 CI checks green (Board 00–07 E2E, routing regressions, Test, Lint, Type Check, Routed DRC, kicad-cli Round-trip, C++ Build).
- mergeStateStatus **CLEAN**, mergeable **MERGEABLE**.

Approving. Ready for merge.

Issue #4247, opened 2026-07-16
route-auto: emitted vias violate the design via/drill DRU (0.45/0.20mm vs 0.6/0.3mm min); no --via-drill/--via-diameter exposed


rjwalters:
## Summary

`kct route-auto` emits vias whose diameter/drill **violate the design's via constraints** (both the board's `.kicad_dru` and the board-setup minimums). It uses its own default via geometry and never reads the design rules. Split out of kicad-tools#4242 (comment) as a distinct bug. Found on the first v0.16.0 real-board run (rjwalters/softstart rev-C, #29).

## Evidence

Board DRU (`softstart_revc.kicad_dru`):
```
(rule "Minimum Via Drill"    (constraint hole_size (min 0.3mm) (max 6.3mm)))
(rule "Minimum Via Diameter" (constraint via_diameter (min 0.6mm)))
```
(board-setup minimum is 0.5 mm dia / 0.3 mm drill.)

After a `route-auto --strategy hierarchical` pass over the board, `kicad-cli pcb drc --refill-zones` reports, uniformly:
```
51x  via_diameter      :: Via diameter (board setup constraints min diameter 0.5000 mm; actual 0.4500 mm)
51x  drill_out_of_range:: Hole size out of range (board setup constraints min hole 0.3000 mm; actual 0.2000 mm)
```
So route-auto placed **every** via at 0.45 mm dia / 0.20 mm drill — below both the DRU (0.6/0.3) and the board setup (0.5/0.3). All 51 via-bearing nets are affected.

## Expected

route-auto should size its vias from the design's via constraints (`.kicad_dru` / board setup), the same way `kct route` honors `--via-drill`/`--via-diameter`. At minimum, route-auto should **expose `--via-drill`/`--via-diameter`** (it currently exposes neither — only `route` does), and ideally **default them from the board DRU** so a per-net route-auto campaign doesn't silently produce a board full of sub-spec vias.

## Impact

On a board that can only be routed net-by-net with route-auto (see #4242 — the batch router is grid-gapped on large boards), every via is out-of-spec, so the whole board fails DRC on via geometry alone, independent of the routing quality. There's no CLI lever to fix it.

## Environment
kicad-tools 0.16.0 (git 78eeee2), native backend v1.0.0, macOS.

---

### Root Cause (grounded in current `main`)

Two distinct, verified gaps combine to produce the observed 0.45mm/0.20mm vias:

[...]

PR #4249, opened 2026-07-16
feat(route): add --max-cells to override the auto-grid memory budget

### LINKED ISSUE #4242: route: large mixed-pitch board falls in a grid gap — auto-grid refuses, safe grid misaligns 95% of pads, recommended grid exceeds an un-raisable cap (no --max-cells)
## Summary

First real-consumer-board run of the v0.16.0 router (rjwalters/softstart rev-C, a 160×100 mm 4-layer mains board with mixed fine-pitch parts). The router's **correctness** is clearly much improved — it is now *honest*: it refuses unsafe grids, reports pad-escape failures, and falls back rather than silently emitting bad copper. 🎉 But a large mixed-pitch board falls into a gap where full-board auto-routing can't proceed cleanly, and the CLI doesn't expose the lever the router's own messages tell you to use. Filing as evidence-first friction.

## The core tension (grid resolution vs. board size vs. pad pitch)

Min clearance 0.127 mm ⇒ safe grid = clearance/2 = **0.0635 mm**. Full board at 0.0635 = **3.97M cells**.

**Path A — `--grid auto`:** refuses.
```
Error: Auto-grid selected 0.127mm > clearance/2 (0.075mm) because the memory budget cap
(max_cells=500,000) forced a coarser grid. ... routing WILL produce clearance-violating
vias/segments (DRC shorts).
```
Preceding warning: *"Increase max_cells (currently capped at 4,000,000 by the auto-grid retry)…"* — **but there is no `--max-cells` flag in `kct route --help`.** The safe grid (3.97M) is *under* the 4M cap the message references, yet the 500k default policy cap blocks it and nothing lets the user raise it.

**Path B — `--grid 0.0635` (explicit, bypasses the cap):** proceeds, but the grid badly misaligns with the mixed pad pitches (0.8 / 0.95 / 1.27 / 2.54 mm):
```
WARNING: 95% of pads (352/371) are off-grid at 0.0635mm resolution.
Routing quality will be degraded. Consider --grid 0.03175 for better pad alignment.
Sub-grid pre-pass: 152 pads could not escape (components: ... U1, U3, U7, ...)
70 nets failed global routing (will attempt anyway)
Net /OC_TRIP_N: C++ pathfinder gave up ... falling back to the pure-Python A* (10-100x slower)
```
So 70/79 nets fail global routing and many fall back to the slow Python A*.

**Path C — `--grid 0.03175` (the router's own recommendation):** = 2.54/80, aligns far better, but full board = **~16M cells** — far over the 4M internal cap, and there's no CLI lever to allow it. (16M×2 layers×~16 B ≈ 512 MB — feasible RAM-wise; the block is pure policy.)

Net: the grid the router *recommends* for this board is one the CLI won't let it use.

## Secondary friction

1. **`--dry-run` is expensive on large boards.** `route --dry-run` (should only report strategy/grid selection) builds the full grid + off-grid pad analysis; on this board it ran >45 s and got OOM-killed before printing a plan. A dry-run should compute `board_area / grid²` analytically and return instantly.

2. **Region-routing cross-boundary ergonomics.** `--region` is the natural large-board escape (fine grid affordable per-tile — a 44×26 mm region at 0.03175 = ~640k cells), and it correctly refuses nets that leave the box without a boundary stub:
   ```
   Error: --region cannot route the following net(s) because they have pad(s) outside the region
   with no same-net boundary stub to reconnect to: +3.3V, /+12V, /AC_NEUTRAL, /I_SENSE_OUT,
   /OC_TRIP_N, /PGND, GND
   ```
   This is correct behavior, but there's no documented "route the trunks first, then region-refine" recipe, so assembling a whole board from regions is trial-and-error. A worked example (or a `route --auto-partition` that plants boundary stubs and tiles automatically) would make region-routing a first-class large-board path.

## Suggested fixes (tiers)

1. **Minimal / highest-value:** expose **`--max-cells N`** on `route`/`route-auto` so `--grid auto` can reach a *safe, well-aligned* grid on a large board (the message already promises this exists internally). This alone likely unblocks the board.
2. **Dry-run:** make `--dry-run` return the grid/cell/budget verdict analytically before constructing the grid.
3. **Region ergonomics:** document (or automate) the trunk-first → region-refine workflow with boundary stubs for boards too large for a single safe fine grid.

## Environment
kicad-tools 0.16.0 (git 78eeee2), native C++ backend v1.0.0, macOS. Board: 160×100 mm, 4-layer, 79 nets, DRU min clearance 0.127 mm / HV class 0.3 mm. The committed placed board is courtyard-clean; only WIP copies were routed (nothing mutated).

rjwalters:
## Summary

`kct route --grid auto` selects a routing grid by filtering candidates against a `max_cells` memory budget. That budget was hardcoded to 500,000 and had no CLI override, so on large boards (the reporter's 160x100mm, 0.127mm-clearance, mixed-pitch board) auto-grid was forced coarser than `clearance/2`. The #3911 safety gate then correctly refused to route on the unsafe grid — leaving no supported path to reach the recommended fine (~16M-cell) lattice.

This PR exposes `--max-cells N` (Tier 1 of #4242) so callers can raise (or lower) that budget and let `--grid auto` reach a safe, well-aligned grid.

## Changes

- **`route_cmd.py`**: add `--max-cells` (int, default 500,000) to the inner parser; thread `args.max_cells` into BOTH grid-selection call sites — the top-level `auto_select_grid_resolution(...)` and the adaptive `compute_multi_resolution_plan(...)` (both branches).
- **`parser.py`**: declare the matching `--max-cells` on the outer `kct route` subparser (kept in sync per the argparse-drift convention).
- **`commands/routing.py`**: forward `--max-cells` from the outer to the inner parser only when non-default, mirroring the `--allow-unsafe-grid` / `--force` pattern (flag-off argv byte-identical).
- **`router/io.py`**: fix `compute_multi_resolution_plan`'s own `max_cells` parameter, which was **dead code** — accepted but never forwarded to its internal `auto_select_grid_resolution` call. Without this, `--max-cells` would silently no-op under the default `--grid-strategy adaptive`. Its default is aligned from the never-used 2,000,000 to the previously-effective 500,000 so existing direct callers stay byte-identical.
- **`tests/test_route_max_cells_flag.py`**: new test module (modeled on `test_route_grid_safety_gate.py`).

`route-auto` is intentionally NOT given this flag — it routes per-net with a small bbox and never hits the board-wide `max_cells` path (per the issue's own follow-up).

| `--max-cells N` int flag, default 500,000, flag-off byte-identical | ✅ | `test_max_cells_default_in_{unified,route_cmd}_parser`; forwarding shim skips when == default (`test_max_cells_not_forwarded_when_default`) |
| Threaded to top-level `auto_select_grid_resolution` | ✅ | `test_auto_select_receives_user_max_cells` |
| Threaded to `compute_multi_resolution_plan` (both call branches) | ✅ | `test_multi_resolution_plan_receives_user_max_cells` |
| `compute_multi_resolution_plan` forwards its own `max_cells` (dead-param fix) | ✅ | `test_internal_uniform_call_receives_max_cells` |
| Outer parser declares it + shim forwards when non-default | ✅ | `test_max_cells_in_unified_parser`, `test_max_cells_forwarded_when_non_default` |
| Accepts values below 500k (genuine override) | ✅ | `test_max_cells_accepts_below_default`, `test_max_cells_below_default_is_forwarded` |
| `--help` explains the `Increase max_cells` relationship | ✅ | `test_max_cells_in_route_cmd_help` |
| `route-auto` NOT given the flag | ✅ | No change to route-auto parser |
| Default behavior unchanged | ✅ | `test_grid_auto_selection`, `test_router_grid_auto_bump`, `test_route_grid_safety_gate`, `test_grid_safety_gate_fleet`, `test_adaptive_grid`, `test_board_01_off_grid_blacklist` all pass (186) |

## Test Plan

- New suite: `tests/test_route_max_cells_flag.py` — 29 tests pass (with `test_cli_parser_drift`).
- Regression: 186 existing grid/adaptive/safety-gate tests pass unchanged.
- `ruff check` + `ruff format --check` clean on all touched files.
- `mypy` introduces zero new errors (4 pre-existing errors in `io.py` confirmed identical on `origin/main`).

Closes #4242

### DIFF
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -218,6 +218,12 @@ def run_route_command(args) -> int:
     grid_val = str(args.grid)
     if grid_val.lower() != "auto":
         sub_argv.extend(["--grid", grid_val])
+    # Issue #4242: forward --max-cells (auto-grid memory budget override) only
+    # when the user changed it from the shared 500,000 default, keeping
+    # flag-off argv byte-identical.  Both parsers declare it; see
+    # tests/test_cli_parser_drift.py.
+    if getattr(args, "max_cells", 500_000) != 500_000:
+        sub_argv.extend(["--max-cells", str(args.max_cells)])
     if args.trace_width != 0.2:
         sub_argv.extend(["--trace-width", str(args.trace_width)])
     if args.clearance != 0.15:
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -2868,6 +2868,19 @@ def _add_route_parser(subparsers) -> None:
         default="auto",
         help="Grid resolution in mm or 'auto' for automatic selection (default: auto)",
     )
+    route_parser.add_argument(
+        "--max-cells",
+        type=int,
+        default=500_000,
+        help=(
+            "Maximum grid cells to allow for --grid auto (default: 500,000). "
+            "Raise this on large boards when auto-grid selects a coarse, unsafe "
+            "grid because of the memory budget cap (the caller-facing override "
+            "for the budget named in the 'Increase max_cells' warning/error). "
+            "Threaded to both the uniform and adaptive grid-selection paths; no "
+            "effect when --grid is an explicit value."
+        ),
+    )
     route_parser.add_argument("--trace-width", type=float, default=0.2, help="Trace width in mm")
     route_parser.add_argument("--clearance", type=float, default=0.15, help="Clearance in mm")
     route_parser.add_argument(
--- a/src/kicad_tools/cli/route_cmd.py
+++ b/src/kicad_tools/cli/route_cmd.py
@@ -7928,6 +7928,21 @@ def main(argv: list[str] | None = None) -> int:
         ),
     )
     parser.add_argument(
+        "--max-cells",
+        type=int,
+        default=500_000,
+        help=(
+            "Maximum grid cells to allow for --grid auto (default: 500,000). "
+            "Raise this on large boards when auto-grid selects a coarse, "
+            "unsafe grid because of the memory budget cap (this is the "
+            "caller-facing override for the budget named in the 'Increase "
+            "max_cells' warning/error text). Values below the default are "
+            "honored too, making the memory cap bind harder. Threaded to both "
+            "the uniform and adaptive (multi-resolution) grid-selection paths. "
+            "No effect when --grid is an explicit value rather than 'auto'."
+        ),
+    )
+    parser.add_argument(
         "--trace-width",
         type=float,
         default=0.2,
@@ -9441,11 +9456,13 @@ def main(argv: list[str] | None = None) -> int:
         board_dims = extract_board_dimensions(pcb_path)
         board_width = board_dims[0] if board_dims else None
         board_height = board_dims[1] if board_dims else None
+        max_cells = getattr(args, "max_cells", 500_000)
         grid_auto_result = auto_select_grid_resolution(
             pads=pad_positions,
             clearance=args.clearance,
             board_width=board_width,
             board_height=board_height,
+            max_cells=max_cells,
         )
 
         # When grid_strategy is adaptive (default), attempt multi-resolution
@@ -9462,6 +9479,7 @@ def main(argv: list[str] | None = None) -> int:
                     clearance=args.clearance,
                     board_width=board_width,
                     board_height=board_height,
+                    max_cells=max_cells,
                 )
             except Exception:
                 # Fall back: try with pad positions (won't have ref info)
@@ -9470,6 +9488,7 @@ def main(argv: list[str] | None = None) -> int:
                     clearance=args.clearance,
                     board_width=board_width,
                     board_height=board_height,
+                    max_cells=max_cells,
                 )
 
         if multi_res_plan is not None and multi_res_plan.is_multi_resolution:
--- a/src/kicad_tools/router/io.py
+++ b/src/kicad_tools/router/io.py
@@ -1477,7 +1477,7 @@ def compute_multi_resolution_plan(
     clearance: float,
     board_width: float | None = None,
     board_height: float | None = None,
-    max_cells: int = 2_000_000,
+    max_cells: int = 500_000,
     zone_padding: float = 2.0,
     min_fine_resolution: float = 0.05,
     fine_pitch_threshold: float = 0.8,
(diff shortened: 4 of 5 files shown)


rjwalters:
✅ **Approved!**

Thorough review complete — this is a clean, well-scoped Tier 1 implementation.

## Correctness — the flagged gotcha holds up
The curator's concern was whether `--max-cells` actually takes effect under the DEFAULT `--grid-strategy adaptive` path, not just the top-level path. Verified it does:
- `route_cmd.py` threads `max_cells` into the top-level `auto_select_grid_resolution(...)` **and** both `compute_multi_resolution_plan(...)` call branches (the ref-aware call and the PadPosition fallback).
- `router/io.py` fixes the genuine dead-parameter bug: `compute_multi_resolution_plan` now forwards `max_cells` into its internal `auto_select_grid_resolution` call. Previously the parameter was accepted but never used, so `--max-cells` would have silently no-op'd on the default adaptive path. Confirmed `max_cells` is only meaningfully consumed by that uniform selection — the remainder of the function only records a `total_cell_estimate` (never budget-enforced), so the fix is complete.
- `test_multi_resolution_plan_receives_user_max_cells` and `test_internal_uniform_call_receives_max_cells` lock both hops in place.

## Default 2,000,000 → 500,000 — verified as truly dead code
`auto_select_grid_resolution`'s own default is `max_cells=500_000`, and `compute_multi_resolution_plan` previously called it without passing `max_cells` — so 500k was already the effective value and the 2,000,000 default was never reachable. The only callers (the two `route_cmd` sites) never passed the parameter, so existing direct callers stay byte-identical. No caller relied on 2M.

## Scope discipline
Diff touches only the 4 source files + 1 new test. No `route-auto` flag (correctly excluded per #4247), no Tier 2/3. `route-auto`'s per-net small-bbox path never hits the board-wide `max_cells` budget, so omitting the flag there is correct.

## Verification
- New suite `tests/test_route_max_cells_flag.py` + `test_cli_parser_drift.py`: 29 passed.
- Regression: `test_route_grid_safety_gate`, `test_grid_safety_gate_fleet`, `test_adaptive_grid`, and all grid-auto/multi-resolution tests: 239 passed, 1 skipped.
- `ruff check` + `ruff format --check` clean on all touched files.
- Full CI green — all 16 checks pass (Test, Type Check, Lint, all 8 boards, diff-pair, match-group, DRC, C++ build). Merge state CLEAN.
- PR uses `Closes #4242`.

Flag-off argv byte-identical forwarding shim mirrors the established `--allow-unsafe-grid`/`--force` pattern. Nice work.

PR #4250, opened 2026-07-16
fix: honor board via geometry in route-auto and expose --via-drill/--via-diameter

### LINKED ISSUE #4247: route-auto: emitted vias violate the design via/drill DRU (0.45/0.20mm vs 0.6/0.3mm min); no --via-drill/--via-diameter exposed

rjwalters:
## Summary

`kct route-auto` emitted vias at 0.45mm dia / 0.20mm drill — below both the board DRU (0.6/0.3) and board-setup minimums — so every via-bearing net failed DRC on via geometry alone (Issue #4247, found on softstart rev-C's 51 via nets).

Two verified gaps combined to produce this:

1. **Latent aggregation bug (root cause).** `parse_pcb_design_rules()` aggregated net-class `via_dia`/`via_drill` across *every* net class using `min()`. A single smaller-via class anywhere in the file (fine-pitch, diff-pair, or unused leftover) poisoned the derived via size for every net — reproducing the exact 0.45/0.20 numbers. For via geometry `min()` is backwards: a via *smaller* than a class's stated minimum is a DRC violation, not a conservative choice.
2. **No CLI lever.** `route-auto` exposed neither `--via-drill` nor `--via-diameter` (only `route` did), and `route_net_auto()` had no params to receive them.

Building a `.kicad_dru` reader is explicitly out of scope — the `max()` aggregation fix resolves the numeric defect because `.kicad_pcb` net-class via minimums mirror the DRU.

## Changes

- **`router/io.py`**: aggregate net-class `via_dia`/`via_drill` with `max()` instead of `min()`, so the derived default is never smaller than any class's stated minimum. Clearance/`trace_width` keep `min()` (intentional, conservative, independently tested).
- **`cli/parser.py`**: add `--via-drill`/`--via-diameter` to `_add_route_auto_parser()`, defaulting to `None` so the board-derived value wins unless the user explicitly overrides.
- **`mcp/tools/routing.py`**: `route_net_auto()` gains `via_drill`/`via_diameter` params, applied as overrides on the `DesignRules` derived at the `parse_pcb_design_rules(...).to_design_rules()` site.
- **`cli/commands/routing.py`**: thread the flags through `run_route_auto_command()` and report the effective via geometry (explicit override vs board-derived) in `--dry-run` output.

| route-auto accepts `--via-drill`/`--via-diameter`, default None | ✅ | `test_parser_accepts_via_flags`, `test_parser_via_flags_default_none` |
| Flags threaded through into `route_net_auto()` as overrides | ✅ | `test_flags_forwarded_to_route_net_auto`, `test_explicit_override_reaches_orchestrator_and_via` |
| `--dry-run` reports effective via drill/diameter | ✅ | `test_dry_run_reports_explicit_override`, `test_dry_run_reports_board_derived_default` |
| `min()`→`max()` for via geometry only; clearance/width unchanged | ✅ | `test_uses_maximum_via_geometry_across_net_classes`; `test_uses_minimum_from_multiple_net_classes` still green |
| test_router_io covers max() via + unchanged clearance/width | ✅ | new `TestParsePcbDesignRules` cases |
| test_mcp_routing covers override + landing on routed vias | ✅ | `TestRouteNetAutoViaOverride` (override reaches orchestrator + persisted via geometry) |
| CLI-level test for new flags + no regression when omitted | ✅ | `tests/test_route_auto_via_flags_cli_4247.py` |
| Multi-net-class fixture: no via under-sizing without flags | ✅ | `test_board_derived_via_uses_max_aggregation` |
| ruff check + format pass | ✅ | see Test Plan |

## Test Plan

- New/updated tests: `tests/test_router_io.py` (max() via aggregation, order-independence, clearance/width still min()), `tests/test_mcp_routing.py::TestRouteNetAutoViaOverride` (explicit override reaches orchestrator + lands on persisted via; partial override; board-derived max()), `tests/test_route_auto_via_flags_cli_4247.py` (parser + forwarding + dry-run).
- `uv run pytest tests/test_router_io.py tests/test_mcp_routing.py tests/test_route_auto_partial_cli_4165.py tests/test_route_auto_via_flags_cli_4247.py` → 237 passed (C++ router backend built via `kct build-native`; without it, an unrelated pre-existing fixture-routing test flips to full-route).
- `uv run ruff check .` and `uv run ruff format . --check` pass.
- `uv run mypy src/` reports 11 errors, all pre-existing in unrelated code (verified identical count with changes stashed); this PR adds zero new mypy errors.

Closes #4247

### DIFF
--- a/src/kicad_tools/cli/commands/routing.py
+++ b/src/kicad_tools/cli/commands/routing.py
@@ -82,9 +82,63 @@ def run_zones_command(args) -> int:
     return 1
 
 
+def _effective_via_geometry(
+    pcb_path: str,
+    via_drill: float | None,
+    via_diameter: float | None,
+) -> tuple[float | None, float | None]:
+    """Resolve the via drill/diameter route-auto would actually use.
+
+    An explicit CLI override wins; otherwise the value is derived from the
+    board's net-class via constraints (Issue #4247).  Returns ``(None, None)``
+    for any field that can be neither overridden nor derived (e.g. the board
+    file is unreadable during a dry-run preview).
+    """
+    eff_drill = via_drill
+    eff_diameter = via_diameter
+    if eff_drill is not None and eff_diameter is not None:
+        return eff_drill, eff_diameter
+    try:
+        from pathlib import Path
+
+        from kicad_tools.router.io import parse_pcb_design_rules
+
+        pcb_rules = parse_pcb_design_rules(Path(pcb_path).read_text())
+        if eff_drill is None:
+            eff_drill = pcb_rules.min_via_drill
+        if eff_diameter is None:
+            eff_diameter = pcb_rules.min_via_diameter
+    except Exception:
+        # Dry-run preview must never fail on a missing/unreadable board file;
+        # fall back to reporting only the explicit overrides (possibly None).
+        pass
+    return eff_drill, eff_diameter
+
+
+def _print_effective_via_geometry(
+    eff_drill: float | None,
+    override_drill: float | None,
+    eff_diameter: float | None,
+    override_diameter: float | None,
+) -> None:
+    """Print the effective via geometry for the dry-run preview (Issue #4247)."""
+
+    def _fmt(value: float | None, override: float | None) -> str:
+        if value is None:
+            return "board-derived (unavailable)"
+        source = "explicit override" if override is not None else "board-derived"
+        return f"{value:.4g}mm ({source})"
+
+    print(f"  Via drill: {_fmt(eff_drill, override_drill)}")
+    print(f"  Via diameter: {_fmt(eff_diameter, override_diameter)}")
+
+
 def run_route_auto_command(args) -> int:
     """Handle route-auto command using RoutingOrchestrator."""
     import sys
+
+    via_drill = getattr(args, "via_drill", None)
+    via_diameter = getattr(args, "via_diameter", None)
 
     # Dry-run: preview strategy selection without routing
     if args.dry_run:
@@ -92,6 +146,11 @@ def run_route_auto_command(args) -> int:
         print(f"  Strategy override: {args.strategy}")
         print(f"  Repair enabled: {not args.no_repair}")
         print(f"  Via resolution enabled: {not args.no_via_resolution}")
+        # Report the effective via geometry that would be used (Issue #4247):
+        # an explicit CLI override, or the value derived from the board's
+        # net-class via constraints.
+        eff_drill, eff_diameter = _effective_via_geometry(args.pcb, via_drill, via_diameter)
+        _print_effective_via_geometry(eff_drill, via_drill, eff_diameter, via_diameter)
         if args.output:
             print(f"  Output: {args.output}")
         return 0
@@ -111,6 +170,9 @@ def run_route_auto_command(args) -> int:
             region=getattr(args, "region", None),
             # Issue #4165: persist partial multi-pad copper only when opted in.
             allow_partial=getattr(args, "allow_partial", False),
+            # Issue #4247: explicit via-geometry overrides (None => board-derived).
+            via_drill=via_drill,
+            via_diameter=via_diameter,
         )
     except FileNotFoundError as e:
         print(f"Error: {e}", file=sys.stderr)
--- a/src/kicad_tools/cli/parser.py
+++ b/src/kicad_tools/cli/parser.py
@@ -3784,6 +3784,28 @@ def _add_route_auto_parser(subparsers) -> None:
         action="store_true",
         help="Disable via conflict resolution (enabled by default)",
     )
+    route_auto_parser.add_argument(
+        "--via-drill",
+        type=float,
+        default=None,
+        metavar="MM",
+        help=(
+            "Via drill size in mm. Overrides the board-derived via drill "
+            "(Issue #4247). When omitted, route-auto derives the drill from "
+            "the board's net-class via constraints."
+        ),
+    )
+    route_auto_parser.add_argument(
+        "--via-diameter",
+        type=float,
+        default=None,
+        metavar="MM",
+        help=(
+            "Via pad diameter in mm. Overrides the board-derived via diameter "
+            "(Issue #4247). When omitted, route-auto derives the diameter from "
+            "the board's net-class via constraints."
+        ),
+    )
     route_auto_parser.add_argument("-o", "--output", help="Output PCB file path")
     route_auto_parser.add_argument(
         "--dry-run",
--- a/src/kicad_tools/mcp/tools/routing.py
+++ b/src/kicad_tools/mcp/tools/routing.py
@@ -471,6 +471,8 @@ def route_net_auto(
     enable_via_resolution: bool = True,
     region: str | tuple[float, float, float, float] | None = None,
     allow_partial: bool = False,
+    via_drill: float | None = None,
+    via_diameter: float | None = None,
 ) -> dict:
     """Route a specific net using the RoutingOrchestrator.
 
@@ -522,6 +524,13 @@ def route_net_auto(
                 outside fails with a clear message (bare-stub reconnection is
                 deferred to Phase 2b).  Any produced segment/via that escapes
                 the box fails the route rather than writing out-of-region copper.
+        via_drill: Optional via drill override in mm (Issue #4247).  When given,
+                overrides the via drill derived from the board's net-class via
+                constraints.  When None, the board-derived value is used.
+        via_diameter: Optional via pad diameter override in mm (Issue #4247).
+                When given, overrides the via diameter derived from the board's
+                net-class via constraints.  When None, the board-derived value
+                is used.
 
     Returns:
         Dictionary with routing result including:
@@ -657,6 +666,14 @@ def route_net_auto(
     pcb_text = path.read_text()
     pcb_rules = parse_pcb_design_rules(pcb_text)
     design_rules = pcb_rules.to_design_rules()
+
+    # Apply explicit via-geometry overrides (Issue #4247).  Only the fields the
+    # caller explicitly passes are overridden; the rest still derive from the
+    # board's net-class via constraints.
+    if via_drill is not None:
+        design_rules.via_drill = via_drill
+    if via_diameter is not None:
+        design_rules.via_diameter = via_diameter
 
     # Build a lightweight PCB-like object for the orchestrator
     # The orchestrator needs pcb.width, pcb.height, and optionally pcb.grid
--- a/src/kicad_tools/router/io.py
+++ b/src/kicad_tools/router/io.py
@@ -392,12 +392,21 @@ def parse_pcb_design_rules(pcb_text: str) -> PCBDesignRules:
                     rules.min_track_width = trace_width
                     found_track_width = True
 
+        # Via geometry aggregation uses max(), NOT min() (Issue #4247).
+        # For clearance/trace_width (above) "tightest wins" is conservative:
+        # a smaller value is never a DRC violation.  For via geometry the
+        # opposite is true — a via SMALLER than a net class's stated minimum
+        # is a DRC violation, so taking the global min across net classes
+        # would let a single small-via class (e.g. an unused fine-pitch or
+        # differential-pair class) poison the derived via size for every net.
+        # Taking the max() guarantees the derived default is never smaller
+        # than any net class's stated minimum.
         via_dia_match = re.search(r"\(via_dia\s+([\d.]+)\)", nc_block)
         if via_dia_match:
             via_dia = float(via_dia_match.group(1))
             if via_dia > 0:
                 if found_via_diameter:
-                    rules.min_via_diameter = min(rules.min_via_diameter, via_dia)
+                    rules.min_via_diameter = max(rules.min_via_diameter, via_dia)
                 else:
                     rules.min_via_diameter = via_dia
                     found_via_diameter = True
@@ -407,7 +416,7 @@ def parse_pcb_design_rules(pcb_text: str) -> PCBDesignRules:
             via_drill = float(via_drill_match.group(1))
             if via_drill > 0:
                 if found_via_drill:
-                    rules.min_via_drill = min(rules.min_via_drill, via_drill)
+                    rules.min_via_drill = max(rules.min_via_drill, via_drill)
                 else:
                     rules.min_via_drill = via_drill
                     found_via_drill = True
--- a/tests/test_mcp_routing.py
+++ b/tests/test_mcp_routing.py
@@ -770,6 +770,185 @@ class TestRouteAutoCommandErrorHandling:
 
 
 # ---------------------------------------------------------------------------
+# Tests: route_net_auto via-geometry override + board-derived aggregation (#4247)
+# ---------------------------------------------------------------------------
+
+
+def _two_pad_pcb_text(extra_net_classes: str = "") -> str:
+    """A minimal 2-pad net PCB that reaches the RoutingOrchestrator.
+
+    ``extra_net_classes`` is injected verbatim so a test can add net-class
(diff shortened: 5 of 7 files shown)


rjwalters:
✅ **Approved!**

Thorough review complete. This is a clean, well-scoped fix for the sub-spec via DRC failures in #4247.

## Root-cause fix verified (`router/io.py`)
- The `min()`→`max()` change is correctly **scoped to via geometry only** (`min_via_diameter`, `min_via_drill`). `clearance`/`trace_width` aggregation is untouched — both remain `min()`. The semantic reasoning is sound and well-documented in the added comment block: for via geometry a value *below* a net class's stated minimum is itself a DRC violation, so `min()` across classes is backwards; `max()` guarantees the derived default is never smaller than any class's minimum.
- `max()` is the right conservative default for this issue's scope. Every net gets a via ≥ its own class minimum (compliant, if slightly generous for the smaller class). A per-net-class via lookup would be more precise but the `.kicad_dru` reader was explicitly deferred by the curator; `max()` resolves the numeric defect safely.
- The KiCad 8 `design_settings` board-level block still directly overrides the net-class aggregation afterward (unchanged), so explicit board minimums continue to win.

## Locked-policy test NOT weakened
`tests/test_router_io.py::test_uses_minimum_from_multiple_net_classes` is unchanged — it still asserts `min_clearance == 0.1` / `min_track_width == 0.15` across two net classes (no via fields), locking the `min()` policy for clearance/width. Confirmed green.

## Override precedence
Explicit `--via-drill`/`--via-diameter` correctly beat the board-derived value; defaults are `None` so board derivation wins when omitted. In `mcp/tools/routing.py` the overrides are applied field-independently onto the `DesignRules` (`test_partial_override_only_touches_given_field` proves partial override leaves the other field board-derived). Attribute names line up: `PCBDesignRules.min_via_*` → `to_design_rules()` → `DesignRules.via_*`.

## Test coverage
Comprehensive: parser acceptance + default-None, forwarding into `route_net_auto` (with None passthrough), dry-run reporting for both explicit-override and board-derived cases, override reaching the orchestrator AND landing on the persisted via geometry, and order-independence of `max()` aggregation. The `_effective_via_geometry` broad `except Exception` is appropriate — a dry-run preview must never crash on an unreadable board file.

## Verification
- Local: `pytest tests/test_router_io.py tests/test_mcp_routing.py tests/test_route_auto_via_flags_cli_4247.py tests/test_route_auto_partial_cli_4165.py` → **237 passed** (C++ backend built).
- CI: all checks green (Test, Lint & Format, Type Check, Boards 00-07, routing regressions, DRC, round-trip smoke).
- Merge state: **CLEAN**. PR body uses `Closes #4247`.

Nice work — the fix targets the true aggregation root cause rather than papering over it with a hardcoded via constant.

PR #4251, opened 2026-07-16
fix: allow underscore-prefixed provenance keys in net-class-map sidecar

### LINKED ISSUE #4248: net-class-map: allow an in-band comment/_meta key (JSON sidecar rejects any non-net key, can't self-document)
## Summary (minor ergonomics)

A `--net-class-map` JSON sidecar rejects any key that isn't a net→dict entry, so there's **no way to document the file in-band** (provenance, board, date, rationale). Adding a `_comment` string key fails hard:

```
Error: invalid net-class-map structure: net_class_map entry for '_comment' must be a dict, got str
```

## Ask

For a hand-authored, committed sidecar (which is the intended use — it lives next to the routed board and feeds both the router and the ampacity DRC), a self-documenting comment is valuable. Suggest **ignoring keys that start with `_`** (or a reserved top-level `"_meta": {...}` block) so authors can record provenance without a sibling README.

Trivial, but it's real friction hit while building a rev-C net-class map on the first v0.16 use (rjwalters/softstart #29). The workaround (strip the comment, document in a separate LAYOUT_NOTES.md) works but splits the documentation from the artifact.

## Environment
kicad-tools 0.16.0.

---

### Root Cause

The validation error comes from `net_class_map_from_dict()` in
`src/kicad_tools/router/rules.py` (lines 1455-1480). It iterates every
top-level key of the sidecar dict and requires each value to itself be a
dict (a `NetClassRouting`-shaped entry):

```python
def net_class_map_from_dict(
    data: dict[str, dict],
) -> dict[str, NetClassRouting]:
    if not isinstance(data, dict):
        raise TypeError(f"net_class_map_from_dict expects a dict, got {type(data).__name__}")
    result: dict[str, NetClassRouting] = {}
    for net_name, entry in data.items():
        if not isinstance(entry, dict):
            raise TypeError(
                f"net_class_map entry for {net_name!r} must be a dict, got {type(entry).__name__}"
            )
        result[net_name] = NetClassRouting.from_dict(entry)
    return result
```

There is no allowance for any key that isn't intended to become a routed
net, so a `_comment` string (or any other sidecar metadata) trips the
`isinstance(entry, dict)` guard.

This function is the single choke point for both CLI consumers of the
sidecar:
- `src/kicad_tools/cli/check_cmd.py:986-1001` (`kct check --net-class-map`) — wraps the raised `TypeError`/`ValueError` into `Error: invalid net-class-map structure: {e}` (this is the exact message in the bug report).
- `src/kicad_tools/cli/route_cmd.py:9307-9321` (`kct route --net-class-map`) — wraps it into `Error: invalid net-class-map structure: {e}` as well.

Fixing `net_class_map_from_dict` alone fixes both call sites — no changes needed in `check_cmd.py` or `route_cmd.py`.

### Implementation Guidance

Adopt the underscore-prefix convention (the issue's primary suggestion,
and the simplest option — it also naturally covers a `"_meta": {...}"`
block since `"_meta"` itself starts with `_`, so no separate reserved-key
special case is needed):

In `net_class_map_from_dict`, skip any top-level key whose name starts
with `_` **before** the `isinstance(entry, dict)` check, e.g.:

```python
for net_name, entry in data.items():
    if net_name.startswith("_"):
        continue  # reserved for in-band documentation (_comment, _meta, ...)
    if not isinstance(entry, dict):
        raise TypeError(
            f"net_class_map entry for {net_name!r} must be a dict, got {type(entry).__name__}"
        )
    result[net_name] = NetClassRouting.from_dict(entry)
```

Notes:
- No change needed to `net_class_map_to_dict()` (the serializer) — it only ever emits real net entries from a `dict[str, NetClassRouting]`, so there is nothing to strip on the write path. Round-tripping a hand-authored file with `_meta` through `to_dict`/`from_dict` is out of scope; the ask is purely "don't reject reading it."
- Do not silently accept a `_`-prefixed key that collides with a real net name people might legitimately use (KiCad net names can't start with `_` in practice, but if this is a concern, a code comment noting the convention is sufficient — no extra guard needed).
- Keep the docstring of `net_class_map_from_dict` updated to mention the `_`-prefix skip behavior (it currently says "Must be a dict-of-dicts").

### Acceptance Criteria

- [ ] A sidecar JSON containing a `_comment` string key (e.g. `{"_comment": "rev-C board, 2026-07", "GND": {...}}`) parses successfully via `net_class_map_from_dict` — no exception raised.
- [ ] A sidecar JSON containing a `_meta` dict key (e.g. `{"_meta": {"board": "rev-C", "date": "2026-07-15"}, "GND": {...}}`) parses successfully via `net_class_map_from_dict`.
- [ ] Keys starting with `_` are excluded from the returned `dict[str, NetClassRouting]` — they must not appear as routed nets and must not affect routing/DRC behavior.
- [ ] Existing net entries (non-`_`-prefixed keys) are parsed exactly as before — no behavior change for well-formed sidecars.
- [ ] The existing negative test `test_from_dict_rejects_non_dict_entry` (a non-`_`-prefixed key with a non-dict value, e.g. `{"USB_D+": "not a dict"}`) continues to raise `TypeError` — the skip is scoped to `_`-prefixed keys only, not a general "ignore malformed entries" relaxation.
- [ ] `kct check --net-class-map <sidecar-with-_comment>.json` and `kct route --net-class-map <sidecar-with-_comment>.json` both succeed (no `Error: invalid net-class-map structure` regression) since both CLI paths route through the same fixed function.

### Affected Files

- `src/kicad_tools/router/rules.py` - `net_class_map_from_dict()` (~line 1455-1480): skip top-level keys starting with `_` before the `isinstance(entry, dict)` check; update docstring.
- `tests/test_net_class_serialization.py` - add test cases: (1) a sidecar dict with a `_comment` string value parses and the key is excluded from the result; (2) a sidecar dict with a `_meta` dict value parses and the key is excluded from the result; verify existing `test_from_dict_rejects_non_dict_entry` (non-`_`-prefixed bad entry) still raises.

No changes are needed in `src/kicad_tools/cli/check_cmd.py` or `src/kicad_tools/cli/route_cmd.py` — both call `net_class_map_from_dict` directly, so the fix is centralized.

### Test Plan

- [ ] Automated tests: extend `tests/test_net_class_serialization.py` with the two new cases above (`_comment` string, `_meta` dict), run `uv run pytest tests/test_net_class_serialization.py -v`.
- [ ] Manual verification: write a sidecar JSON with `_comment` and a normal net entry, run `kct check --net-class-map <file> <board>.kicad_pcb` and confirm no `invalid net-class-map structure` error; repeat for `kct route --net-class-map`.
- [ ] Edge cases: verify a `_`-prefixed key with a non-dict value (string, list, `None`) is uniformly ignored (not just string); verify a real net name is never accidentally skipped (no net name in practice starts with `_`, but confirm the loop only special-cases the literal prefix check, not a broader heuristic).

rjwalters:
## Summary

A `--net-class-map` JSON sidecar previously rejected any top-level key whose value was not a net→dict entry, so a hand-authored file could not carry an in-band comment or provenance block (`Error: invalid net-class-map structure: net_class_map entry for '_comment' must be a dict, got str`). This adds the underscore-prefix convention: top-level keys starting with `_` are treated as in-band documentation and skipped, so `_comment` (string) and `_meta` (dict) can live next to real net entries.

## Changes

- `src/kicad_tools/router/rules.py` — `net_class_map_from_dict()` now skips any top-level key whose name starts with `_` **before** the `isinstance(entry, dict)` check. Reserved keys are excluded from the returned `dict[str, NetClassRouting]` and never become routed nets. Docstring updated to document the convention. Non-`_`-prefixed keys with non-dict values are still rejected.
- `tests/test_net_class_serialization.py` — added cases for a `_comment` string, a `_meta` dict, and a uniform "any value type is ignored" check for `_`-prefixed keys.

No CLI changes: both `kct check --net-class-map` and `kct route --net-class-map` funnel through `net_class_map_from_dict`, so the single-function fix covers both.

| `_comment` string key parses without exception | ✅ | `test_from_dict_skips_underscore_comment_key` |
| `_meta` dict key parses without exception | ✅ | `test_from_dict_skips_underscore_meta_dict` |
| `_`-prefixed keys excluded from result (not routed) | ✅ | Both new tests assert `set(result) == {"GND"}` |
| Existing net entries parsed exactly as before | ✅ | `result == m` in new tests; existing roundtrip tests still pass |
| `test_from_dict_rejects_non_dict_entry` still raises | ✅ | Non-`_` key skip is scoped to the literal prefix; test passes |
| `_`-prefixed key ignored regardless of value type | ✅ | `test_from_dict_skips_underscore_key_regardless_of_value_type` (str/list/None/dict) |

## Test Plan

- `uv run pytest tests/test_net_class_serialization.py` — 24 passed.
- `uv run ruff check .` — clean (pre-existing unrelated MFR001 noqa warning only).
- `uv run ruff format . --check` — all files formatted.
- `uv run mypy src/kicad_tools/router/rules.py` — no issues.

Closes #4248

### DIFF
--- a/src/kicad_tools/router/rules.py
+++ b/src/kicad_tools/router/rules.py
@@ -1460,18 +1460,30 @@ def net_class_map_from_dict(
     Round-trip companion to :func:`net_class_map_to_dict`.  Each entry
     is deserialized via :meth:`NetClassRouting.from_dict`.
 
+    Top-level keys whose name starts with an underscore (``_``) are
+    reserved for in-band documentation/provenance and are skipped: they
+    are never turned into routed nets.  This lets a hand-authored sidecar
+    self-document without a sibling README, e.g. a ``_comment`` string or
+    a ``_meta`` dict block (issue #4248).  Because KiCad net names do not
+    start with ``_`` in practice, this cannot shadow a real net entry.
+
     Args:
         data: Mapping of ``net_name -> NetClassRouting-dict``.  Must be a
-            dict-of-dicts.
+            dict; non-``_``-prefixed values must each be a dict.  Any
+            ``_``-prefixed key is ignored regardless of its value type.
 
     Raises:
-        ValueError: If any entry is malformed (missing 'name' field).
-        TypeError: If ``data`` is not a dict-of-dicts.
+        ValueError: If any (non-``_``-prefixed) entry is malformed
+            (missing 'name' field).
+        TypeError: If ``data`` is not a dict, or if a non-``_``-prefixed
+            entry is not a dict.
     """
     if not isinstance(data, dict):
         raise TypeError(f"net_class_map_from_dict expects a dict, got {type(data).__name__}")
     result: dict[str, NetClassRouting] = {}
     for net_name, entry in data.items():
+        if isinstance(net_name, str) and net_name.startswith("_"):
+            continue  # reserved for in-band documentation (_comment, _meta, ...)
         if not isinstance(entry, dict):
             raise TypeError(
                 f"net_class_map entry for {net_name!r} must be a dict, got {type(entry).__name__}"
--- a/tests/test_net_class_serialization.py
+++ b/tests/test_net_class_serialization.py
@@ -343,3 +343,31 @@ class TestNetClassMapHelpers:
     def test_from_dict_rejects_missing_name(self):
         with pytest.raises(ValueError, match="requires a 'name' field"):
             net_class_map_from_dict({"USB_D+": {"priority": 1}})
+
+    def test_from_dict_skips_underscore_comment_key(self):
+        """A ``_comment`` string key is in-band documentation, not a net (#4248)."""
+        m = create_net_class_map(power_nets=["GND"])
+        wire = net_class_map_to_dict(m)
+        wire["_comment"] = "rev-C board, 2026-07"
+        result = net_class_map_from_dict(wire)
+        # The comment key is excluded; the real net survives unchanged.
+        assert "_comment" not in result
+        assert set(result) == {"GND"}
+        assert result == m
+
+    def test_from_dict_skips_underscore_meta_dict(self):
+        """A ``_meta`` dict block is in-band provenance, not a net (#4248)."""
+        m = create_net_class_map(power_nets=["GND"])
+        wire = net_class_map_to_dict(m)
+        wire["_meta"] = {"board": "rev-C", "date": "2026-07-15"}
+        result = net_class_map_from_dict(wire)
+        # The _meta dict is skipped even though it *is* a dict.
+        assert "_meta" not in result
+        assert set(result) == {"GND"}
+        assert result == m
+
+    def test_from_dict_skips_underscore_key_regardless_of_value_type(self):
+        """Any ``_``-prefixed key is ignored uniformly (str, list, None, dict)."""
+        for reserved_value in ("a string", ["a", "list"], None, {"nested": 1}):
+            result = net_class_map_from_dict({"_reserved": reserved_value})
+            assert result == {}


rjwalters:
✅ **Approved!**

Clean, well-scoped fix for #4248. The single-line skip in `net_class_map_from_dict()` (`src/kicad_tools/router/rules.py`) correctly handles the in-band-documentation convention, and both `kct check` and `kct route` inherit it for free since they funnel through this one function.

## Review findings

**Correctness** — `if isinstance(net_name, str) and net_name.startswith("_"): continue` runs *before* the `isinstance(entry, dict)` check, so `_`-prefixed keys are dropped for any value type (str/list/None/dict) and never enter `result`. Real net entries are untouched (`result == m` in the new tests). The `isinstance(net_name, str)` guard is a nice defensive touch for non-string keys.

**Regression guard verified** — The skip is narrowly scoped to the literal `_` prefix, so a non-underscore key with a non-dict value still hits the `TypeError` path. `test_from_dict_rejects_non_dict_entry` and `test_from_dict_rejects_missing_name` both still pass.

**Edge cases** — `test_from_dict_skips_underscore_key_regardless_of_value_type` exercises str/list/None/dict under a `_` key uniformly. The `_`-as-reserved-top-level-key convention doesn't collide with real KiCad net names (leading `_` on a whole top-level key is not a practical KiCad net-name form), and this is documented in the updated docstring.

**Docs/quality** — Docstring `Args`/`Raises` updated to reflect the new behavior; inline comment explains intent. Clean.

## Local verification
- `pytest tests/test_net_class_serialization.py` — 24 passed (incl. 3 new + both negative regression tests)
- `ruff check` — clean on both changed files
- `mypy src/kicad_tools/router/rules.py` — no issues

## CI
All 16 checks pass (Test 17m56s, Type Check, Lint, C++ Build, Board 00-07 E2E, Board 05 Routing Regression, Diff-Pair + Match-Group regressions, DRC, kicad-cli smoke). Merge state: CLEAN.

PR description uses `Closes #4248`. Ready to merge.

